PR #1028, opened 2017-06-01
[RFC] Enable sosreport to capture kernel crash report/summary

ankitkumar2391:
This patchset enables sosreport to capture summary of kernel dump.

This plugin runs crash tool on latest kernel dump and generates summary. In our experience with this summary we are able to debug ~60% of the issues.

Advantage :

    With single sosreport output we will be able to debug most of the kernel issues as well.
    We can possibly avoid shipping kernel dump from customer site

Limitations :

    crash tool needs debug kernel on running system. So if debug kernel is not installed we just come out with error message.
    Running crash tool on kernel dump takes some time.

### DIFF
--- a/sos/plugins/crash_report.py
+++ b/sos/plugins/crash_report.py
@@ -1,0 +1,225 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
+
+import os
+import time
+from datetime import datetime
+from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+from sos.utilities import sos_get_command_output, find
+
+
+class CrashReport(Plugin):
+    """Crash kernel report
+    """
+    plugin_name = "crash_report"
+    profiles = ('system', 'debug')
+
+    dbg_vmlinux_path = '/usr/lib/debug/'
+    vmcore_stnd_path = '/var/crash'
+
+    crash_script = ""
+
+    # list of crash command
+    crash_cmd_list = [
+        ("sys", "SYSTEM INFORMATION:"),
+        ("help -m", "STACK LOCATION INFO:"),
+        ("bt", "STACK TRACE OF CURRENT CONTEXT:"),
+        ("bt -a", "STACK TRACES OF ACTIVE TASKS:"),
+        ("kmem -i", "MEMORY USAGE:"),
+        ("kmem -s", "KMALLOC SLAB DATA:"),
+        ("mod", "MODULES:"),
+        ("ps", "PROCESS STATUS:"),
+        ("log", "SYSTEM MESSAGE BUFFER:"),
+        ("files", "OPEN FILES OF CURRENT CONTEXT:"),
+        ("dev -p", "PCI DEVICE DATA:"),
+        ("runq", "RUN QUEUE TASK:"),
+        ("mach -o", "OPALMSG LOG:"),
+        ("irq -s", "DUMP KERNEL IRQ STATE:")]
+
+    def setup(self):
+        path = self.get_cmd_output_path()
+        CrashReport.output_file = os.path.join(path, 'crash_report.txt')
+        self.get_crash_report()
+
+    def form_crash_script(self):
+        """
+        parse crash_cmd_list and form crash_script which will
+        be given as an input to crash command
+        """
+        for (cmd, header) in self.crash_cmd_list:
+            self.crash_script += "!echo %s >>{OUTPUT}\n" % header
+            self.crash_script += "!echo >> {OUTPUT}\n"
+            self.crash_script += "%s >> {OUTPUT}\n" % cmd
+            self.crash_script += "!echo >> {OUTPUT}\n"
+            self.crash_script += "!echo >> {OUTPUT}\n"
+        self.crash_script += "quit\n"
+        self.crash_script += "!echo >> {OUTPUT}\n"
+        self.crash_script += "!echo >> {OUTPUT}\n"
+
+    def is_dump_for_crashing_kernel(self, file_path, crashing_time):
+        file_creation_info = time.ctime(os.path.getctime(file_path))
+        """
+        used to find dump for crashed kernel.
+        it compares the time of dump file with last crashed time.
+        It looks for dump(vmcore) file whose creation time is newer
+        than the time of system crashed, In this case (t1 >= t2) and
+        hence difference.days >= 0.
+        return
+            True  : In case of success.
+            False : In case of failure.
+        """
+
+        """
+        get the actual time string using split on complete line
+        ctime returns string as 'Mon Mar 13 02:54:32 2017'
+        we are interested only in Month date hrs:min field
+        """
+        vmcore_creation_time = file_creation_info.rsplit(':', 1)[0].split(
+                                                          ' ', 1)[1]
+        t1 = datetime.strptime(vmcore_creation_time, "%b %d %H:%M")
+        t2 = datetime.strptime(crashing_time, "%b %d %H:%M")
+        difference = t1 - t2
+        return True if difference.days >= 0 else False
+
+    def get_crash_time(self, line):
+        """
+        Here we are trying to get crash time(m/d/hr/min) from last cmd output
+        Sample example of lines are:
+        root     pts/0        gateway       Fri Aug 18 04:35 - crash  (00:05)
+        root     hvc0                       Fri Aug 18 04:41 - crash  (00:06)
+        """
+        crashing_time = (line.split()[4] + ' ' + line.split()[5] +
+                         ' ' + line.split()[6])
+        try:
+            t2 = datetime.strptime(crashing_time, "%b %d %H:%M")
+        except ValueError:
+            t2 = None
+            crashing_time = (line.split()[3] + ' ' + line.split()[4] +
+                             ' ' + line.split()[5])
+            try:
+                t2 = datetime.strptime(crashing_time, "%b %d %H:%M")
+            except ValueError:
+                t2 = None
+                return ""
+        return crashing_time
+
+    def get_core_info(self):
+        # Collect data from last command
+        last_cmd_res = sos_get_command_output('last')
+        if last_cmd_res['status']:
+            self._log_error("last command failed")
+            return -1
+
+        crash_found = False
+
+        for line in last_cmd_res['output'].splitlines():
+            """
+            Sample example of last command output:
+            root     pts/0        gateway   Mon Aug 21 23:52   still logged in
+            root     hvc0                   Fri Aug 18 04:41 - crash(00:06)
+            reboot   system boot  3.10.0-514.el7.p Mon Aug 21 01:28 - 00:41
+            Below loop reads line by line and get crash kernel name,
+            crashing time info
+            """
+            line = line.rstrip()
+            if crash_found is True:
+                reboot_str = (line.split()[0] + ' ' + line.split()[1] + ' ' +
+                              line.split()[2])
+                if reboot_str == 'reboot system boot':
+                    self.crashed_kernel_name = line.split()[3]
+                    break
+            elif "crash" in line:
+                line_contain_crash = line
+                crash_found = True
+        if not crash_found or self.crashed_kernel_name == "":
+            self._log_error("Failed in finding crash from last cmd output")
+            return -1
+        crashing_time = self.get_crash_time(line_contain_crash)
+        if not crashing_time:
+            self._log_error("Failed in locating time for last crash:[%s]"
+                            % line_contain_crash)
+            return -1
+        ret = self.get_vmlinux_path(self.dbg_vmlinux_path,
+                                    self.crashed_kernel_name)
+        if ret:
+            return -1
+        ret = self.get_vmcore_path(self.vmcore_stnd_path, crashing_time)
+        if ret:
+            return -1
+        return 0
+
+    def get_vmlinux_path(self, dbg_vmlinux_path, kernel_name):
+        for filename in find("vmlinu*", dbg_vmlinux_path):
+            """
+            in rhel : 3.10.0-327.el7.ppc64/vmlinux
+            in ubuntu: vmlinux-4.4.0-21-generic
+            hence below check confirms that vmlinux is of crashed kernel
+            """
+            if filename.find(kernel_name) != -1:
+                self.vmlinux_path = filename
+                return 0
+        self._log_error("Debug vmlinux not found['%s'] inside path:['%s']"
+                        % (self.crashed_kernel_name, self.dbg_vmlinux_path))
+        return -1
+
+    def get_vmcore_path(self, vmcore_stnd_path, crashing_time):
+        for filename in find(self.dump_file_prefix, vmcore_stnd_path):
+            ret = self.is_dump_for_crashing_kernel(filename, crashing_time)
+            if ret:
+                self.vmcore_file_path = filename
+                return 0
+        self._log_error("vmcore not found for kernel['%s'] inside path: ['%s']"
+                        % (self.crashed_kernel_name, self.vmcore_stnd_path))
+        return -1
+
+    def get_crash_report(self):
+        result = self.get_core_info()
+        if result:
+            return
+
+        self.form_crash_script()
+        crash_full_command = ("crash" + " -s " + self.vmlinux_path + " " +
+                              self.vmcore_file_path)
+        self._log_info("crash command: '%s'" % crash_full_command)
+        result = sos_get_command_output(
+                crash_full_command, stderr=True,
+                input=self.crash_script.format(
+                 OUTPUT=self.output_file)
+                )
+
+        if result['status']:
+            self._log_error("crash command failed")
+        return
+
+
+class RedHatCrashReport(CrashReport, RedHatPlugin):
+
+    def default_enabled(self):
+        return False
+
+    def setup(self):
+        CrashReport.dump_file_prefix = 'vmcore'
+        super(RedHatCrashReport, self).setup()
+
+
+class DebianCrashReport(CrashReport, DebianPlugin, UbuntuPlugin):
+
+    def default_enabled(self):
+        return False
+
+    def setup(self):
+        CrashReport.dump_file_prefix = 'dump.*'
+        super(DebianCrashReport, self).setup()
+
+# vim: set et ts=4 sw=4 :
--- a/sos/utilities.py
+++ b/sos/utilities.py
@@ -110,7 +110,7 @@ def is_executable(command):
 
 
 def sos_get_command_output(command, timeout=300, stderr=False,
-                           chroot=None, chdir=None, env=None):
+                           chroot=None, chdir=None, env=None, input=None):
     """Execute a command and return a dictionary of status and output,
     optionally changing root or current working directory before
     executing command.
@@ -149,10 +149,11 @@ def sos_get_command_output(command, timeout=300, stderr=False,
             expanded_args.append(arg)
     try:
         p = Popen(expanded_args, shell=False, stdout=PIPE,
+                  stdin=PIPE if input else None,
                   stderr=STDOUT if stderr else PIPE,
                   bufsize=-1, env=cmd_env, close_fds=True,
                   preexec_fn=_child_prep_fn)
-        stdout, stderr = p.communicate()
+        stdout, stderr = p.communicate(input)
     except OSError as e:
         if e.errno == errno.ENOENT:
             return {'status': 127, 'output': ""}


sos/plugins/crash_report.py
```diff
@@ -0,0 +1,249 @@
...
+    vmcore_file_path = ''
+    crash_script = ""
+    # crash result output file
+    crashed_output_file = '/var/log/os_crash_report.txt'
+
+    # list of crash command
+    crash_cmd_list = [
+                    ("sys", "\"SYSTEM INFORMATION:\""),
+                    ("mach", "\"MACHINE SPECIFIC DATA:\""),
+                    ("mach", "\"MACHINE SPECIFIC DATA:\""),
```
robbmanes: Do we need to call `mach` twice?  Also, `help -m` prints all this and more in crash AFAIK.

sos/plugins/crash_report.py
```diff
@@ -0,0 +1,249 @@
...
+                    ("bt", "\"STACK TRACE OF CURRENT CONTEXT:\""),
+                    ("bt -a", "\"STACK TRACES OF ACTIVE TASKS:\""),
+                    ("kmem -i", "\"MEMORY USAGE:\""),
+                    ("kmem -s", "\"KMALLOC SLAB DATA:\""),
```
robbmanes: Is there a concern on large memory systems about how long any kmem commands take?  I know upstream crash did some improvements when walking SLAB tables so this would be better than it used to be, but I'd suspect anyone who's done core analysis on a system that had a large SLAB table or some type of memory-related issue knows how long `kmem` can take when it is trying to wrestle with huge indexes.  Two examples, an ideal and a problematic one:

```bash
$ cat crash_commands 
sys
mach
bt
bt -a
kmem -i
kmem -s
mod
ps
log
files
dev -p
runq
mach -o
irq -s
exit

$ time crash -i crash_commands vmcore /usr/lib/debug/lib/modules/2.6.32-358.el6.x86_64/vmlinux -s
- - - - - - - - 8< - - - - - - - - 
real	0m7.662s
user	0m7.343s
sys	0m0.302s
```

The above core had 16GB of ram and 2 CPU's so that seems a reasonable amount of time to me.  But depending on the issue this can take a very long time, such as any core hitting an [anon_vma_chain](https://lkml.org/lkml/2011/6/17/3) issue:

```bash
$ time crash -i crash_commands vmcore /usr/lib/debug/lib/modules/2.6.32-131.21.1.el6.x86_64/vmlinux -s
- - - - - - - - 8< - - - - - - - - 
real	23m28.618s
user	2m3.791s
sys	0m18.660s
```

The above was with a core with 250GB of RAM and 32 CPU's, with a massive amount of SLAB objects in the table due to the aforementioned issue - and it didn't finish, I actually killed it.  The vast majority of time was in `kmem -s` as it indexed the SLAB table and it would have eventually completed but probably only after an hour or so, myabe more.

So I have a real concern about how long this could take, or if we have a way to bail out if it takes an inordinate amount of time.  I'm playing devil's advocate here with a worst-case scenario but I still think it's a real possibility that would prevent `sosreport` from being effective if this was a default plugin.

sos/plugins/crash_report.py
```diff
@@ -0,0 +1,249 @@
...
+                    ("mod", "\"MODULES:\""),
+                    ("ps", "\"PROCESS STATUS:\""),
+                    ("log", "\"SYSTEM MESSAGE BUFFER:\""),
+                    ("files", "\"OPEN FILES OF CURRENT CONTEXT:\""),
+                    ("dev -p", "\"PCI DEVICE DATA:\""),
+                    ("runq", "\"RUN QUEUE TASK:\""),
+                    ("mach -o", "\"OPALMSG LOG:\""),
```
robbmanes: `mach -o` only works on PPC64 systems, maybe we can do a check with the output of `help -m` to determine architecture type and run include it in the commands list?  That might be a lot of work so maybe it's better just to get the garbage error output of `mach` for all other architectures, or not include it at all since the majority of `sos` consumers are running x86_64 to my knowledge.

sos/plugins/crash_report.py
```diff
@@ -0,0 +1,249 @@
...
+             '[ ' + line_contain_crash + ' ]' + '\n\n')
+            return -1
+        ret = self.retrieve_path(self.dbg_vmlinux_path, "vmlinu",
+                                 self.crashed_kernel_name, crashing_time)
+        if ret:
+            self.message_to_log = (
+             '\n\n' + 'Failed in retrieving debug linux:' +
+             '[ ' + self.crashed_kernel_name + ' ]' + ' inside path: ' +
+             '[ ' + self.dbg_vmlinux_path + ' ]' +
+             '\nPlease install debug linux and rerun sosreport\n\n')
```
robbmanes: Pedantic on my part, but it is `kernel-debuginfo` in Red Hat/CentOS/Fedora and `linux-image-$(uname -r)-dbgsym` in Debian/Ubuntu that provides the symbols needed for crash/kgdb, which might be confused with `kernel-debug` which is usually a build of kernel with extra compiler flags for debugging either live or adding more fields in structs for post-mortem debugging. Maybe change the message to reflect this like "Please install the appropriate debug symbols to proceed for your currently-running kernel, usually kernel-debuginfo for RHEL and linux-image-$(uname -r)-dbgsym for Ubuntu?"

sos/plugins/crash_report.py
battlemidget: Anything potentially long running needs to be behind an option flag

sos/plugins/crash_report.py
battlemidget: Also this requires the user to install something else on their system to retrieve this data. sosreport should not halt if it can't find the debug symbols and continue on.

robbmanes:
> It sounds like this plugin could potentially take a long time to run. We should think about making this plugin disabled and having the user manually enable it during a sosreport run.

+1.  For my two cents, this is the only way I think this should be implemented - most of customers do not want/have crash or debug symbols installed on their systems, instead wishing the core analysis to be done remotely on either another system of theirs or by their support vendor.  Alternatively they may not want to start such a possibly heavy process just while doing regular data collection via `sosreport`.  I think it would be trivial for the vendor/support engineer to ask for `sosreport -e crash` instead of having it enabled by default.

sos/plugins/crash_report.py
ankitkumar2391: 1. "mach" and "mach -o" gives different info.
"mach" provides machine specific details.
"mach -o" provides firmware information and only supported on OpenPower.
Hence both commands are needed.
On other arch , mach -o prints error in output file and hence can be identified that log is from other arch .

2. We collect log of various command in a single file one after other.
We append description and then add "output result of command" to identify log, else it will be difficult to identify after seeing log.

Example:
I am pasting partial log of file here. Below is the output of sys command.
We added "SYSTEM INFORMATION:" first to identify that log is from sys command and then output result.
----
SYSTEM INFORMATION:

      KERNEL: 4.8.0-39/vmlinux-4.8.0-39-generic
    DUMPFILE: 4.8.0-39/dump.201705251505  [PARTIAL DUMP]
        CPUS: 160
        DATE: Thu May 25 15:06:14 2017

sos/plugins/crash_report.py
ankitkumar2391: Hi Robb && Adam,
Thank you so much for review.
Sure will work on it and make this plugin not part of default plugin. (will keep it under option flag) and send next version soon.

How about below approach ?
While running sosoreport if we find a latest crash available.. print a message to user saying he could run sosreport -e crash to capture crash summary.
So that If user is interested can rerun sosreport to collect crash report.

sos/plugins/crash_report.py
ankitkumar2391: Hi Robb,
Yes, mach -o is supported only on PPC64 systems.

I prefer logging error in case of other architecture rather checking help command (check will increase complexity of this unnecessarily ).

We should keep "mach -o" part of this plugin as it is needed on PPC64 architecture and even on PPC64 sos is an important tool to collect data.

sos/plugins/crash_report.py
ankitkumar2391: Hi Robb,
Will modify message as suggested by you.

Adam,
We require only debuginfo and dump data.
sosreport won't halt if debug symbol is not found as crash results error in that case and error will be logged to output file.

sos/plugins/crash_report.py
bmr-cymru: > While running sosoreport if we find a latest crash available.. print a message to user saying he could 
> run sosreport -e crash to capture crash summary.

Absolutely not. Plugins must not attempt to do any form of terminal IO during a run; this breaks `--batch` and `--quiet` and is not supported by the plugin class API. There have been proposals to extend the API to allow interactive prompting when appropriate, but this is quite a large and complex piece of work and not currently high priority.

sos/plugins/crash_report.py
robbmanes: I mean that `mach` is literally called twice with the same arguments:
```
                    ("mach", "\"MACHINE SPECIFIC DATA:\""),
                    ("mach", "\"MACHINE SPECIFIC DATA:\""),
```
I know `mach` provides different data based on flags but this looks like it just a duplicated line, since we call `mach -o` later, no?
```
                    ("mach -o", "\"OPALMSG LOG:\""),
```
Also, pretty sure everything that is output in `mach` without arguments is also in `help -m`, which also includes stack locations and various settings, but we don't have to include it if it's not needed.

ankitkumar2391: Hi Robb,

Thank you so much for your time in reviewing the patch.

Ahh, Yes "mach" is duplicate here.
I will remove in next version.

Yes, "help -m" prints the information present in "mach" and even more.

What I am thinking is lets get this patch part of sos-report source code.
Later we can add other command needed.

Robb,What you say ?

sos/plugins/crash_report.py
ankitkumar2391: Hi Bryn,

Sure.

sos/plugins/crash_report.py
ankitkumar2391: What I meant is ,
I will replace "mach" with "help -m" .

ankitkumar2391: Hi Robb,
Thank you so much for your time in reviewing the patch.

Ahh, Yes "mach" is duplicate here.
I will remove in next version.

Yes, "help -m" provided the same output as mach provides and even more.
Hence will replace mach with "help -m".
Thank you.

sos/plugins/crash_report.py
ankitkumar2391: Hi Robb,
I sent next version after addressing review comment.
Please have a look.

ankitkumar2391:
Hi Bryn,
I sent next version after addressing review comment.
Please have a look.

sos/plugins/crash_report.py
robbmanes: This is fixed in latest squashed commit, thanks.

sos/plugins/crash_report.py
robbmanes: This is fixed in submitted/squashed version, thanks.

sos/plugins/crash_report.py
robbmanes: The message is fixed in submitted/squashed version, thanks again.

ankitkumar2391:
Hi Bryn,
Did you get a chance to look into new changes.

sos/plugins/crash_report.py
```diff
@@ -0,0 +1,256 @@
...
+        self.form_crash_script()
+        crash_full_command = ("crash" + " -s " + self.vmlinux_path + " " +
+                              self.vmcore_file_path)
+        self.message_to_log = 'crash command: ' + crash_full_command
+        self.dump_log_data()
+        print("\n\nCollecting crash report. This may take a while !!! ...\n\n")
+        result = sos_get_command_output(
+                crash_full_command, stderr=True,
+                input=self.crash_script.format(
+                 OUTPUT=self.crashed_output_file))
```
bmr-cymru: Closing paren should be on the following line, on its own (indented to the same level as `result =`).

sos/plugins/crash_report.py
```diff
@@ -0,0 +1,256 @@
...
+            sys.stderr.write('\n\nFailed opening crash log file:[%s]\n\n'
+                             % self.crashed_output_file)
+            return -1
+        try:
+            f.write("\n\n[%s]\n\n" % self.message_to_log)
+        except:
+            sys.stderr.write('write to log file failed\n')
+            f.close()
+            return -1
+        f.close()
```
bmr-cymru: Close can raise `IOError` exceptions.

bmr-cymru:
I'm trying to go through the current patches and add some review comments to this thread. Thank you for all the work so far - I will try to explain any concerns we have, as well as minor problems and style/standards errors in the current patches.

First a few minor points:

 * The commit tag for the new plugin should be just `[crash_report]` (it's always the plugin name).
 * The subject for the main commit is too long. Keep it under 78 chars please (there is no need to use the words plugin, crash and report twice each in a brief summary line!).

More generally, I have concerns that this plugin is duplicating functionality provided by other components in some distributions - most notable the "ABRT" package in Red Hat derived distributions, and the generic `kerneloops` client tool and (oops.kernel.org)[oops.kernel.org] service. I am not sure that the module proposed here adds significantly to the capabilities these services already offer. This isn't necessarily a reason to not merge this as an optional component but I do think that it is something that should be addressed during the review process (especially since currently we may not be taking advantage of ABRT and kerneloops data when it does already exist on a machine).

One more technical matters there are still a large number of fixes neede - mostly minor/trivial:

 * plugin class name (`CrashLog`) and `plugin_name` (`crash_report`) do not match.
 * Incorrect superclass initialisation order
 * Mis-quoted strings with embedded `"`
 * Writes to temporary files outside `$TMP_DIR/sos*`
 * Odd iteration over command array in `form_crash_script()`
 * Odd time check on vmcore validity
 * Log messages need to be a single line (unless they are dumping multi-line data in `-vvv` or `--debug` mode).

But there are some more complex problems that would need to be resolved first. In some ways I think the plugin has gotten ahead of itself a little: it may be useful at this point to treat it as a prototype or proof-of-concept. If I was creating a plugin to solve this problem I would first focus on the basic infrastructure: just come up with a plugin that can run _one_ crash command, and then build up from there. It's much easier than trying to solve all the little problems once the plugin has grown to this size.

sos/plugins/crash_report.py
```diff
@@ -0,0 +1,256 @@
...
+    vmcore_stnd_path = '/var/crash'
+
+    dump_file_prefix = ''
+    crashed_kernel_name = ''
+    vmlinux_path = ''
+    message_to_log = ''
+    vmcore_file_path = ''
+    crash_script = ""
+    # crash result output file
+    crashed_output_file = '/var/log/os_crash_report.txt'
```
ankitkumar2391: Hi Bryn,
Agree, we should not modify system configuration and no plug-in does that in sosreport.

But I really wonder a way to create file from our plugin under sos temporary archive (/var/tmp/sos.XYZ/)
Because i don't see any plugin using this variable(self.tmpdir) or importing sos.sosreport.

Please advice your approach, sothat i will go ahead with that.
~Ankit

bmr-cymru: Use `Plugin.get_cmd_output_path()` - grep in `plugins/` for usage examples.

ankitkumar2391: Thanks Bryn for quick response.
Yes I do get sos archive path from "get_cmd_output_path".

But If i create any file under my plugin directory inside **sos archive** 
             (/var/tmp/sos.XYZ/sos_commands/crash_report/file_created). 
This file will become part of tarball only if i pass file as an input to below call:
1. add_cmd_output
2. add_copy_spec

Issue i am facing to use above two calls are:
1. add_cmd_output:
-----------------------------
1. Got sos archive path for my plugin using "get_cmd_output_path" 
                                         -- /var/tmp/sos.xx/sos_commands/plugin_name/
2. passed my command and file name(in which i want to save output) to "add_cmd_output" call.
Issue in above call in my plug-in is:
When i run crash command, it provides it's command prompt on which i type command which i want to execute on crash prompt.
function "add_cmd_output" doesn't provide an interface to provide comand as aninput and hence we won't be able to run command on crash prompt and it hangs.
Example:
How do we provide input on crash prompt:
1. run crash command:
     crash -s <PATH OF VMLINUX>   <PATH OF VMCORE FILE>
2. Above command opens a crash command prompt on which we will trigger crash command
    crash> 
3. Enter command:
    crash> sys


1. add_copy_spec:
--------------------------------
Steps followed:
1. Got sos archive path for my plugin using "get_cmd_output_path" 
                                         -- /var/tmp/sos.xx/sos_commands/plugin_name/
2. appended my file string to above path to create file
3. created file.
4. paased my file name as an input to "add_copy_spec" 
Issue here is:
 - It adds my file under tarball with complete location 
                  sosreport.../var/tmp/sos_zyz/sos_commands/crash_report/filename
    Which we of course don't want.
          File should be under sosreport.../sos_commands/crash_report/filename

bmr-cymru: > This file will become part of tarball only if i pass file as an input to below call:
> A. add_cmd_output
> B. add_copy_spec

No: everything in that location is added to the tar archive (read `sos/archive.py` and the `Archive.finalize()` method).

Just look at the examples I suggested...

ankitkumar2391: Thank you Bryn.
Ahh due to postproc definition , this file was getting cleanedup.(My bad).
Now i am able to create file under sos_commands/crash_report

sos/plugins/crash_report.py
```diff
@@ -0,0 +1,256 @@
...
+                    ("kmem -i", "\"MEMORY USAGE:\""),
+                    ("kmem -s", "\"KMALLOC SLAB DATA:\""),
+                    ("mod", "\"MODULES:\""),
+                    ("ps", "\"PROCESS STATUS:\""),
+                    ("log", "\"SYSTEM MESSAGE BUFFER:\""),
+                    ("files", "\"OPEN FILES OF CURRENT CONTEXT:\""),
+                    ("dev -p", "\"PCI DEVICE DATA:\""),
+                    ("runq", "\"RUN QUEUE TASK:\""),
+                    ("mach -o", "\"OPALMSG LOG:\""),
+                    ("irq -s", "\"DUMP KERNEL IRQ STATE:\"")]
```
ankitkumar2391: Thank you so much for your time.

Yes i could have used single quote in place of " and then we don't require \
something like this: '"PCI DEVICE DATA:"'

But I feel " is not required .
And hence will remove and keep it simple as "PCI DEVICE DATA".

sos/plugins/crash_report.py
```diff
@@ -0,0 +1,256 @@
...
+
+    """
+    parse crash_cmd_list and form crash_script which will
+    be given as an input to crash command
+    """
+    def form_crash_script(self):
+        for idx in range(len(self.crash_cmd_list)):
```
ankitkumar2391: Will make changes in next version and send out in a day or two.

sos/plugins/crash_report.py
```diff
@@ -0,0 +1,256 @@
...
+            self.crash_script += "!echo >> {OUTPUT}\n"
+            self.crash_script += "!echo >> {OUTPUT}\n"
+        self.crash_script += "quit\n"
+        self.crash_script += "!echo >> {OUTPUT}\n"
+        self.crash_script += "!echo >> {OUTPUT}\n"
+
+    """
+    used to find dump for crashed kernel.
+    it compares the time of dump file with last crashed time.
+    return
```
ankitkumar2391: Will make changes in next version and send out in a day or two.

sos/plugins/crash_report.py
```diff
@@ -0,0 +1,256 @@
...
+        get the actual time string using split on complete line
+        ctime returns string as 'Mon Mar 13 02:54:32 2017'
+        we are interested only in Month date hrs:min field
+        """
+        vmcore_creation_time = file_creation_info.rsplit(':', 1)[0].split(
+                                                          ' ', 1)[1]
+        t1 = datetime.strptime(vmcore_creation_time, "%b %d %H:%M")
+        t2 = datetime.strptime(crashing_time, "%b %d %H:%M")
+        difference = t1 - t2
+        return True if difference.days >= 0 else False
```
ankitkumar2391: Yes logic is correct.
I will add more and proper comment to explain things clearly.

sos/plugins/crash_report.py
```diff
@@ -0,0 +1,256 @@
...
+
+    # retrieve path of vmlinux/vmcore
+    def retrieve_path(self, start_path, file_type, file_opt, crashing_time):
+        if file_type == "vmcore":
+            for filename in find(self.dump_file_prefix, start_path):
+                ret = self.is_dump_for_crashing_kernel(filename, crashing_time)
+                if ret:
+                    self.vmcore_file_path = filename
+                    return 0
+        elif file_type == "vmlinu":
```
ankitkumar2391: ok

sos/plugins/crash_report.py
```diff
@@ -0,0 +1,256 @@
...
+                in ubuntu: vmlinux-4.4.0-21-generic
+                hence below check confirms that vmlinux is of crashed kernel
+                """
+                if filename.find(file_opt) != -1:
+                    self.vmlinux_path = filename
+                    return 0
+        return -1
+
+    def get_crash_time(self, line):
+        crashing_time = (line.split()[4] + ' ' + line.split()[5] +
```
ankitkumar2391: ok

sos/plugins/crash_report.py
```diff
@@ -0,0 +1,256 @@
...
+            crashing_time = (line.split()[3] + ' ' + line.split()[4] +
+                             ' ' + line.split()[5])
+            try:
+                t2 = datetime.strptime(crashing_time, "%b %d %H:%M")
+            except ValueError:
+                t2 = None
+                return ""
+        return crashing_time
+
+    def get_vmlinux_vmcore_path(self):
```
ankitkumar2391: sure.

sos/plugins/crash_report.py
```diff
@@ -0,0 +1,256 @@
...
+        # retrieve data from last command
+        last_cmd_res = sos_get_command_output('last')
+        if last_cmd_res['status']:
+            self.message_to_log = '\n\n' + 'last cmd failed' + '\n\n'
+            self.dump_log_data()
+            return -1
+
+        crash_found = False
+
+        for line in last_cmd_res['output'].splitlines():
```
ankitkumar2391: sure

sos/plugins/crash_report.py
```diff
@@ -0,0 +1,256 @@
...
+        if not crashing_time:
+            self.message_to_log = (
+             '\n\n' + 'Failed in retrieving time for last crash:' +
+             '[ ' + line_contain_crash + ' ]' + '\n\n')
+            return -1
+        ret = self.retrieve_path(self.dbg_vmlinux_path, "vmlinu",
+                                 self.crashed_kernel_name, crashing_time)
+        if ret:
+            self.message_to_log = (
+             '\n\n' + 'Failed in retrieving debug linux:' +
```
ankitkumar2391: Sure will remove newline ('\n') from string.

I have a question over here. In my plugin at some places message becomes long. so i should keep it in same line ?

Example:
self._log_error("Failed in locating debug linux:['%s'] inside path:['%s']. Please i    nstall the appropriate debug symbols to proceed for your currently-running kernel, usually kernel-debuginfo for RHEL and linux-image-$(uname -r)-dbgsym for Ubuntu"
 % (self.crashed_kernel_name, self.dbg_vmlinux_path))

sos/plugins/crash_report.py
```diff
@@ -0,0 +1,256 @@
...
+             '[ ' + self.crashed_kernel_name + ' ]' + ' inside path: ' +
+             '[ ' + self.dbg_vmlinux_path + ' ]' +
+             '\nPlease install the appropriate debug symbols to proceed ' +
+             'for your currently-running kernel, usually kernel-debuginfo' +
+             ' for RHEL and linux-image-$(uname -r)-dbgsym for Ubuntu\n\n')
```
ankitkumar2391: In next version will use stnd logging mechanism for logging messages.
Will remove message_to_log from plugin.

sos/plugins/crash_report.py
```diff
@@ -0,0 +1,256 @@
...
+            return -1
+        ret = self.retrieve_path(self.vmcore_stnd_path, "vmcore", "",
+                                 crashing_time)
+        if ret:
+            self.message_to_log = (
+             '\n\n' + 'Failed in retrieving vmcore for kernel' +
```
ankitkumar2391: sure

sos/plugins/crash_report.py
```diff
@@ -0,0 +1,256 @@
...
+             '[ ' + self.vmcore_stnd_path + ' ]' + '\n\n')
+            return -1
+        return 0
+
+    # dump error log to output file
+    def dump_log_data(self):
+        try:
+            f = open(self.crashed_output_file, "a")
+        except IOError:
+            sys.stderr.write('\n\nFailed opening crash log file:[%s]\n\n'
```
ankitkumar2391: sure

sos/plugins/crash_report.py
ankitkumar2391: ok

sos/plugins/crash_report.py
```diff
@@ -0,0 +1,256 @@
...
+        except:
+            sys.stderr.write('write to log file failed\n')
+            f.close()
+            return -1
+        f.close()
+        return 0
+
+    def postproc(self):
+        if os.path.exists(self.crashed_output_file):
+            os.remove(self.crashed_output_file)
```
ankitkumar2391: sure will do that

sos/plugins/crash_report.py
```diff
@@ -0,0 +1,256 @@
...
+        if result:
+            self.dump_log_data()
+            return
+
+        self.form_crash_script()
+        crash_full_command = ("crash" + " -s " + self.vmlinux_path + " " +
+                              self.vmcore_file_path)
+        self.message_to_log = 'crash command: ' + crash_full_command
+        self.dump_log_data()
+        print("\n\nCollecting crash report. This may take a while !!! ...\n\n")
```
ankitkumar2391: ok

sos/plugins/crash_report.py
ankitkumar2391: sure

sos/plugins/crash_report.py
```diff
@@ -0,0 +1,256 @@
...
+
+    def default_enabled(self):
+        return False
+
+    def setup(self):
+        CrashLog.dump_file_prefix = 'vmcore'
+        self.get_crash_report()
+        self.add_copy_spec([
+            CrashLog.crashed_output_file
+        ])
```
ankitkumar2391: sure

sos/plugins/crash_report.py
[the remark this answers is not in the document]
ankitkumar2391: Hi Bryn,
I have worked on other review comments.
Please respond whether approach is ok ?
If you are ok then will send patches(update patches in pull request) .

sos/plugins/crash_report.py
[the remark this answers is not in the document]
ankitkumar2391: Hi Bryn,
Please suggest whether it is okay to keep this lines ?
Or should i remove this line ?
But keeping this line help user to wait so that user doesn't think that it is hanged or something .
Please suggest.
Will send patch once you respond.

ankitkumar2391:
Hi Bryn,
Worked on review comment and pushed changes.
Please review .
1. I removed log which i was printing earlier on stdout 
            "This pluggin may take some time"

ankitkumar2391:
Hi Bryn,
Did you get chance to look into  ?

hegdevasant:
Hi Bryn,

Any update?

-Vasant

bmr-cymru:
Hey guys,

Now that 3.5 is out we're finishing off that and other packaging work for RHEL7.5 at the moment. This is on my priority list for 3.6 (due around first quarter next year).

There's been some progress towards some of the points raise earlier but I've not seen any reply to the comments that were made in August: when I follow the two commit links at the bottom of this PR I'm still seeing the commits that I commented on that month.

Looking at the last comment it just mentions removing a log message: there were a bunch of other items raised in the previous comment but it's not clear what was and wasn't changed.

Is the branch up-to-date?

ankitkumar2391:
Hi Bryn,
I have worked on all review comments suggested by you.

Most of the comments were also mentioned in above respective section .
I had replied to those sections and hence I din't replied in last section to avoid duplicacy but had worked on all review comments.

hegdevasant:
Bryn,

Can you look into this issue?

-Vasant

bmr-cymru:
Hi Vasant,

Unfortunately I'm tied up with other project work at the moment - the next sos release (3.6) is scheduled for April and we intend to clear the review backlog by then. The last time I looked at the plugin is was in better shape so there is a good chance we will be able to merge it by then.

Thanks,

bmr-cymru:
Hi Vasant,

3.6 is overdue by a little over a month - we expect to make the final release some time between now and the middle of June, once the review backlog has been dealt with. Some of our colleagues at Red Hat began a project this year to identify gaps in data collection and this has left us with a very large queue of pulls to go through for the release. This is still on the list, but as it's one of the larger patches to review it's currently not at the top of the list but it will get looked at prior to release and from the last detailed check I don't expect it to be held up.

bmr-cymru:
I've looked at the current patch today, and there are still a few problems with the current implementation:

* Does not work on current Fedora, reports that it cannot find `vmlinux` despite it being present:

```
 Setting up archive ...
[archive:TarFileArchive] initialised empty FileCacheArchive at '/var/tmp/sos.ORaIeE/sosreport-localhost-2018-05-29-nlgqjaa'
[sos.sosreport:setup] executing 'sosreport -v --batch --build -o crash_report'
 Setting up plugins ...
[plugin:crash_report] Debug vmlinux not found['4.14.14-200.fc26'] inside path:['/usr/lib/debug/']
```

vs:

```
# ls /usr/lib/debug/usr/lib/modules/4.11.8-300.fc26.x86_64/
extra  kernel  vdso  vmlinux
```

The plugin is using integer return values to signal success/failure. This isn't how Python normally does this: you should return data (or nothing) on success, and raise an exception on failure. The only place in `sos` that _should_ be doing this is the command line interface in `sos.sosreport`: this needs to use integers in order to return the command exit status to the shell. We do still have some other parts of `sos` that are incorrectly using success/error returns but we are trying to clean those up and not add more. It's also fine to return a value on success and `None` or `-1` on not found or error (but it must be either `NoneType`, or `int(-1)` not `0` or some other type value).

Fixing this will make the plugin simpler and easier to understand: for now, it's a bit confusing that some of the `get_...` methods (e.g. `get_crash_time()`) return a value, while others (`get_core_info()`) set values in the plugin object and return an integer.

Let's make it all consistent:

  * Have all `get_*` methods return a value
  * Have `get_core_info()` (the top-level method) collect the values from the other methods, and return them as a tuple
  * Then the main `get_crash_report()` method can assign them all to the relevant members in one place (making the code easier to navigate), OR, the whole of that method can just be merged into `setup()` (they are both pretty short).

There are also some general style errors, e.g.:

```python
        result = sos_get_command_output(
                crash_full_command, stderr=True,
                input=self.crash_script.format(
                 OUTPUT=self.output_file)
                )
```

Should be:

```python
        result = sos_get_command_output(
            crash_full_command, stderr=True,
            input=self.crash_script.format(
            OUTPUT=self.output_file)
        )
```

Lastly please don't log command execution state:

```python
        if result['status']:
            self._log_error("crash command failed")
```

We're already logging the fact that we run these commands in the report logs, and we store the result (including any actual error messages) in the tarball (same for all the other commands, e.g. `last`).

bmr-cymru:
OK; it's reporting that it can't find the debuginfo for the current kernel. This output is quite confusing since it is logged at "error" level: it will always be displayed to the user, even though it may not be relevant (in this case, there is no crash dump to process anyway).

I would suggest logging this at "info" level or lower: not being able to find anything to work with isn't an error.

TurboTurtle:
I'm closing this out currently, as it has been several years since the last work iteration. Since then there's been a large overhaul of the sos project, and the plugin API, which would mean a significant re-write of this would be needed as well.

If this is still something that is needed by support teams, please update the branch to be based on the current `master`, and open a new PR for a fresh review cycle.

PR #1030, opened 2017-06-06
[sosreport] Add check for root user when listing plugins. 

illobo:
See also: https://bugs.launchpad.net/ubuntu/+source/sosreport/+bug/1685274

This fix will alert an user when trying to list plugins without root privileges. 

Signed-off-by: Fabio Bardella <fabio.bardella@canonical.com>

### DIFF
--- a/sos/plugins/docker.py
+++ b/sos/plugins/docker.py
@@ -122,13 +122,12 @@ class UbuntuDocker(Docker, UbuntuPlugin):
 
     packages = ('docker.io', 'docker-engine')
 
-    # Name collision with another package requires docker binary rename
-    docker_cmd = 'docker.io'
-
     def setup(self):
         super(UbuntuDocker, self).setup()
         self.add_copy_spec([
-            "/etc/default/docker.io"
+            "/etc/default/docker",
+            "/etc/docker/daemon.json",
+            "/var/run/docker/libcontainerd/containerd/events.log"
         ])
 
 # vim: set et ts=4 sw=4 :


BryanQuigley:
Thanks for contributing this upstream!  Please do follow the https://github.com/sosreport/sos/wiki/Contribution-Guidelines.

illobo:
@BryanQuigley thanks - I've amended some changes and CI is going 👍  now

bmr-cymru:
Please follow the `[subsystem, plugin, or module]` convention for tagging issues, pulls, and commits: it should describe the part of `sos` that is affected, so in this case `[sosreport]` (the command line program). Using free-form text here defeats the whole purpose - to allow grouping, filtering, sorting and other operations on the lists of issues, commits, and pulls, according to the part of `sos` that is affected.

bmr-cymru:
Issue and pull descriptions should also be in the present tense (as should commit subjects generally but we are less strict about that).

bmr-cymru:
Please also squash any "fixup" commits together when pushing branches - user either `git commit --amend`, or `git rebase -i` to reformat, and combine commits: no "fixups" are permitted in merges.

sos/sosreport.py
```diff
@@ -1114,6 +1114,10 @@ def _report_profiles_and_plugins(self):
         self.ui_log.info("")
 
     def list_plugins(self):
+	if  not os.geteuid() == 0:
+            self.ui_log.info(_("Plugin list unavailable, root access required."))
+            return
```
bmr-cymru: Nack. This completely ignores the entire `requires_root` infrastructure.

bmr-cymru:
There is existing infrastructure for plugins to declare whether or not they require root privileges (and a small number of plugins that set this).

If you would like to help complete that work we can give you some pointers to get started but as it stands this request is just a hack that drives a truck through efforts to handle this in the project.

bmr-cymru:
Wouldn't it have been better to have tried to fix the bug that was actually reported on LaunchPad?

I.e. that `sos` gives a confusing error message in the case that a non-root user attempts to enable a plugin that is marked as requiring root privileges? This **IS** a bug and *should* be fixed. Crippling plugin listings is about the worst possible approach to that I can imagine.

Rather than a just disabling any listing of plugins in this case?

illobo:
Hi @bmr-cymru, thanks for your suggestions! 

I am still a rookie with code and contributing to open source software, so I will take them in account for next PRs towards sosreport. Forgive me if my PR seemed confused. 

I'm going to close it and find a better way rather than drive a truck over other people's work. Unfortunately my python skills are limited, as I am now discovering with your feedback. I will submit a new PR with a better solution then. Thanks again.

bmr-cymru:
You can keep the same PR open, and just push a new branch with `git push --force` - this is preferred to churning through PR numbers.

For something like this, if there's clearly existing infrastructure (`requires_root`), even if it's incomplete, it's usually better to talk to us via an issue, or the mailing list, to get a sense of the direction to go in if it's not completely clear.

The best solution for this specific problem will be to detect that a plugin requires root, and to print that specific error message when a non-root user tries to enable or activate it. That would also get us a step closer to being able to run `sos` without root privileges which would be a welcome and useful addition.

illobo:
@bmr-cymru thanks again - I'll try my best and amend the PR then, I'm looking at plugin_class.requires_root right now and subscribing to the mailing list as well.

bmr-cymru:
Sure, no problem - or we can get the issues worked out in this one, and then close it out and start a new one when you have something that's ready to merge. If there's anything you're unsure of post a question here and someone will pick it up.

It would be great to get non-root support a bit closer to working (none of the distros really care that much, as for support they always want the root sosreport, but it's still a potentially useful feature).

battlemidget:
I'm not sure what happened here but the PR doesn't have any repos attached to it. I assume that the committer deleted their branch and this PR is now left in limbo.

I'm going to close this out for now, please open a new PR with any suggestions applied from this request.

PR #1031, opened 2017-06-06
[block] Add option to get data from all block dev. 

rrosatti:
A command line called 'all' was defined to this plugin.
This option aims to get the data from all block devices in /sys/block/.

The block plugin will continue to work just as before. The
only difference is that now the user has the option to get info from all
devices.

Signed-off-by: Rodrigo R. Galvão <rodrigorosattig@gmail.com>


rrosatti:
That's odd .. the check on the third commit (ac750b2) didn't run.

sos/plugins/block.py
```diff
@@ -22,6 +22,15 @@ class Block(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
 
     plugin_name = 'block'
     profiles = ('storage', 'hardware')
+    option_list = [("all", "get all block info", "fast", False)]
+
+    def get_info_for_all_devices(self):
+        block_devices = [blk for blk in os.listdir('/sys/block/')]
+        for blk in block_devices:
+            all_files = self.call_ext_prog("find /sys/block/%s/ -type f" % blk)
```
battlemidget: Why not just use add_copy_spec here?

sos/plugins/block.py
```diff
@@ -32,6 +41,9 @@ def setup(self):
             "ls -lanR /sys/block"
         ])
 
+        if self.get_option("all"):
```
battlemidget: If this is a fast thing to do why not just run it regardless and not make it optionable? I tend to think the slow tasks should be optional and everything else run by default.

rrosatti: I just thought it would be a good idea to let it be optional, so the user can choose whether he wants to get all the data in /sys/block/ or not. I mean, in case the machine has lots of paths it would make this task slow, right?

bmr-cymru: It always seems a good idea, until suddenly half the functionality is hidden behind magic options that most users and support engineers are unaware of: then you only get half the data by default, and have to make additional round-trips to request all the extra options be turned on.

There are very few reasons to make something non-default in `sos`:

* It really is very expensive or of limited use
* It requires authentication, special files, or other set up before it is useful and those are not guaranteed to be present

If it satisfies both of those, and it's still so slow you want to consider making it optional it's usually a sign that the feature needs to be re-thought, or some of the work moved out of sos (e.g. having a tool to do the processing for us, rather than text scraping and processing in Python - but those cases are rare).

battlemidget: For me I always like to try and get everything I could possibly want to help debug a problem in the first go around rather than going back to the customer to ask for more data.

sos/plugins/block.py
rrosatti: My idea here was to get only the files inside /sys/block and save it in a list. Is there a better way to do this?

battlemidget: You can just use add_copy_spec with file globbing here and remove the need to do the code traversal

bmr-cymru: This has been tried before however, and gets pretty slow on large systems. You may find that using `os.path.walk` and pre-filtering `sysfs` attributes that are unreadable, irrelevant or otherwise not worth collecting gives you better results.

bmr-cymru:
This branch needs some tidying up to be suitable for review - the NVME stuff doesn't really seem to belong to htis PR (`[block] Add option to get data from all block dev`) and any development commits like "add this" ... "delete this" need to be merged/rewritten as appropriate.

sos/plugins/nvme.py
```diff
@@ -1,119 +1,98 @@
...
+
+                if pci_slot_location:
+                    pci_loc = pci_slot_location[0]
+                    slot_loc = pci_slot_location[3]
+                    self.add_string_as_file(pci_loc, "pci_loc.%s" % dev)
+                    self.add_string_as_file(slot_loc, "slot_loc.%s" % dev)
+
+            # runs nvme-cli commands
+            self.add_cmd_output([
+                                "nvme list",
```
bmr-cymru: These lines should only be indented *one* level from the preceding line (same as any other Python block indent).

bmr-cymru:
There's good potential in this work but it currently needs some re-thinking and tidying up of the branch. Moving out to 3.7.

pmoravec:
Some updates to the PR, @rrosatti ?

TurboTurtle:
Given the long period of inactivity on this, I am closing this out now. If you would like to resubmit these changes please factor in the reviews above, rebase on master, and open a new PR.

PR #1032, opened 2017-06-06
[docker-distribution] Add New Plugin

TurboTurtle:
Adds a new plugin for docker-distribution (registry v2).

Collects the journal, registry config and a view of the contents of
images in the registry.

### DIFF
--- a/sos/plugins/docker_distribution.py
+++ b/sos/plugins/docker_distribution.py
@@ -1,0 +1,43 @@
+# Copyright (C) 2017 Red Hat, Inc. Jake Hunsaker <jhunsake@redhat.com>
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+from sos.plugins import Plugin, RedHatPlugin
+import os
+
+
+class DockerDistribution(Plugin):
+
+    """Docker Distribution"""
+
+    plugin_name = "docker_distribution"
+
+    def setup(self):
+        self.add_copy_spec('/etc/docker-distribution/')
+        self.add_journal('docker-distribution')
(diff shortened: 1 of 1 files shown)


sos/plugins/docker-distribution.py
```diff
@@ -0,0 +1,43 @@
...
+
+from sos.plugins import Plugin, RedHatPlugin
+import os
+
+
+class DockerDistribution(Plugin):
+
+    """Docker Distribution"""
+
+    plugin_name = "docker-distribution"
```
bmr-cymru: Need to use an underscore here, not a hyphen (also in the file name - otherwise it breaks name mapping in the plugins module).

sos/plugins/docker-distribution.py
```diff
@@ -0,0 +1,43 @@
...
+    def setup(self):
+        self.add_copy_spec('/etc/docker-distribution/')
+        self.add_journal('docker-distribution')
+        if os.path.exists('/etc/docker-distribution/registry/config.yml'):
+            with open('/etc/docker-distribution/registry/config.yml') as f:
+                for line in f:
+                    if 'rootdirectory' in line:
+                        loc = line.split()[1]
+                        self.add_cmd_output('tree ' + loc)
+
```
bmr-cymru: Ack.

sos/plugins/docker-distribution.py
```diff
@@ -0,0 +1,43 @@
...
+
+class RedHatDockerDistribution(DockerDistribution, RedHatPlugin):
+
+    packages = ('docker-distribution',)
+
+    def setup(self):
+        self.add_forbidden_path('/etc/docker-distribution/registry/*passwd')
+        super(RedHatDockerDistribution, self).setup()
```
bmr-cymru: Ack.

TurboTurtle:
Updated to use an underscore - interesting that the hyphen breaks name mapping since everything appeared to work normally when I was testing this.

bmr-cymru:
Huh: we appear to have unintentionally fixed something - previously, you would get an error when using plugin options (`-kplug-name.plugopt`). I just added one to your original module, and I'm not now getting an error (although maybe I forgot something.. it's been a while since this came up).

But regardless: we adopted this restriction from `pep8`, which includes it for two reasons: not all file systems support the hyphen in file names, and more importantly, the syntactic ambiguity that arises when you allow hyphens to be embedded in names. Should the expression `foo-bar` evaluate as "foo-dash-bar", or as "foo minus bar"?

We got into this originally around the time the OpenStack plugins were being added - they were the first big user in the tree to want to be able to qualify plugin names with a prefix.

bmr-cymru:
Merged as commit 32e462a5e1495f36e4107324c6f656c5f7db06ec - thanks.

bmr-cymru:
So GitHub only bothers with commit abbreviation/linking if it exists upstream at the time of the comment?

Well OK GitHub: commit 32e462a5e1495f36e4107324c6f656c5f7db06ec then!

PR #1033, opened 2017-06-08
[systemd] Collect also output of systemd-cgls

msekletar:
systemd-cgls is a tool that reports current content of cgroup
tree. Output of the tool is very useful for figuring out in which
cgroup given process runs. Process cgroup placement may have
implications on process lifetime, e.g. user sessions are terminated
early on shutdown, before "normal" system services.

### DIFF
--- a/sos/plugins/systemd.py
+++ b/sos/plugins/systemd.py
@@ -44,6 +44,7 @@ class Systemd(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
             "systemctl show-environment",
             "systemd-delta",
             "systemd-analyze",
+            "systemd-cgls",
             "journalctl --list-boots",
             "ls -lR /lib/systemd",
             "timedatectl"


bmr-cymru:
Thanks for the PR, but you should check out the `cgroups` plugin:

```python
class Cgroups(Plugin, DebianPlugin, UbuntuPlugin):
    """Control groups subsystem
    """

    plugin_name = "cgroups"
    profiles = ('system',)
    files = ('/proc/cgroups',)

    def setup(self):
        self.add_copy_spec([
            "/proc/cgroups",
            "/sys/fs/cgroup"
        ])

        self.add_cmd_output("systemd-cgls")

        return
```

After lengthy debate we chose to add this to `cgroups` instead of `systemd` since it is more functionally related to the former. This plugin is part of the `system` profile so it will execute by default, or whenever that profile is selected by the user.

For future pulls, please include a "Signed-off-by" (DCO) for each commit.

msekletar:
Ahh, obviously I was not aware of this.

PR #1034, opened 2017-06-09
[landscape] add /sys/class/dmi/id/sys_vendor

davecore82:
The get_vm_info and _get_vm_by_vendor functions in the Landscape
client's vm_info.py script rely on /sys/class/dmi/id/sys_vendor
for VM type detection.

Sometimes Landscape doesn't detect the VM type, because of an
unknown string in that file. Adding this file will help us
understand why Landscape didn't detect the computer as a VM.

The data included in this file is only a few characters long and
will not have an impact on the time or size of the report.

Signed-off-by: David Coronel <david.coronel@canonical.com>

PR #1037, opened 2017-06-13
[hardware] add /sys/class/dmi/id/*

davecore82:
Collect the entire /sys/class/dmi/id/ directory from the
hardware plugin. Today, this is really the "native" format for
this information, as far as Linux is concerned. The dmidecode
program (which historically at least read the data directly
from firmware tables, as no proc or sysfs interface existed) is
a legacy formatting of this information which we are
unfortunately obliged to keep around a little longer yet, as
many users have grown dependencies on its format.

The data included in this directory is small (about 500 bytes
on my system) and will not have an impact on the time or size
of the report.

Original discussion at:
https://github.com/sosreport/sos/pull/1034

Signed-off-by: David Coronel <david.coronel@canonical.com>

### DIFF
--- a/sos/plugins/hardware.py
+++ b/sos/plugins/hardware.py
@@ -29,7 +29,8 @@ class Hardware(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
             "/proc/dma",
             "/proc/devices",
             "/proc/rtc",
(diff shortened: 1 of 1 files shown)


bmr-cymru:
Ack, looks good - thanks David.

PR #1039, opened 2017-06-14
[origin] Collect OVS Flows for origin bridge

robbmanes:
Using ovs-ofctl, collect the flows and port descriptions relevant to
troubleshooting connectivity issues on nodes.

---

[X] Is the commit message split over multiple lines and hard-wrapped at 72 characters?
[X] Is the subject and message clear and concise?
[X] Does the subject start with **[plugin_name]** if submitting a plugin patch or a **[section_name]** if part of the core sosreport code?
[X] Does the commit contain a **Signed-off-by: First Lastname <email@example.com>**?

### DIFF
--- a/sos/plugins/origin.py
+++ b/sos/plugins/origin.py
@@ -162,6 +162,12 @@ class OpenShiftOrigin(Plugin):
             ])
             self.add_journal(units="atomic-openshift-node")
 
+            # The node bridge is always named br0, collect OVS flows
+            self.add_cmd_output([
+                "ovs-ofctl -O OpenFlow13 dump-flows br0",
+                "ovs-ofctl -O OpenFlow13 dump-ports-desc br0"
+            ])
+
     def postproc(self):
         # Clear env values from objects that can contain sensitive data
         # Sample JSON content:


robbmanes:
Passed local tests and pep8, tested on RHEL7 OSE3 environment.  Need for this was brought up by @rjhowe who often asks for this information in troubleshooting.

rjhowe:
These commands are needed to make sure the OpenShift OVS networking is working. 

When dealing with OpenShift network issues these two commands are always added when requesting a sosreport.
https://docs.openshift.com/container-platform/3.5/admin_guide/sdn_troubleshooting.html#is-the-open-vswitch-configured-correctly

The ovs bridge name is always configured as br0 with the Redhat OVS Network Plugins. 

The openvswitch plugin will not gather this information due to the OpenFlow version needed to be specified  (-O OpenFlow13) with OpenShift. 
https://github.com/sosreport/sos/blob/master/sos/plugins/openvswitch.py#L91

bmr-cymru:
Thanks for filing the pull & taking time to fill out the template. From the admin side all looks fine.

Just trying to understand the need for, and behaviour of the `-O` switch here? Can you point me at any docs that would save me installing this to test?

Generally when there's a request to add a new command to one plugin, where we already collect something very similar in another, there's a need to understand if the two cannot share more of the implementation: it looks odd having these two commands in `origin` when very other occurrence of `ovs-ofctl` is in the `openvswitch` plugin.

robbmanes:
Sure - the `-O` flag here is the protocol version, you can substitute `--protocols` for it too.  Notably origin uses OpenFlow13 on its bridge (which is always named br0) - I debated adding this to the openvswitch plugin and having it loop through every vswitch for every possible flow type (versions [10..13] are currently in use) but this seemed more targeted and I am not sure of the total ramifications of running that on a heavy-switching system with every possible flow.

If you do think that we should do the above to the openvswitch plugin (I agree that would be cleaner) then before we actually implement something like that I'd have to test to make sure dumping flows for every bridge for every protocol on a busy system doesn't lead to any misbehavior, and that will take me some time to test.

robbmanes:
Okay, maybe something like this would suffice for what we need but still not have a weird ovs-ofctl command in origin plugin:

```
diff --git a/sos/plugins/openvswitch.py b/sos/plugins/openvswitch.py
index 177da89..ee9a27f 100644
--- a/sos/plugins/openvswitch.py
+++ b/sos/plugins/openvswitch.py
@@ -83,6 +83,17 @@ class OpenVSwitch(Plugin):
                     "ovs-ofctl show %s" % br
                 ])
 
+                # If possible, grab specific protocol information
+                flow_versions = ["OpenFlow10",
+                                 "OpenFlow11",
+                                 "OpenFlow12",
+                                 "OpenFlow13"]
+
+                for flow in flow_versions:
+                        "ovs-ofctl --protocols=%s dump-flows %s" % (flow, br)
+                    ])
+
 
 class RedHatOpenVSwitch(OpenVSwitch, RedHatPlugin):
```

I can get to testing this today, it will take me a bit to get a bunch of flows and switches busy enough to make sure there's no impact, but we kinda already collect dump-flows so I don't think that would be very bad at all.

bmr-cymru:
Thanks for the explanation - it's kinda as I feared :-) (that there are legitimate concerns with putting this into the main flow for all `openvswitch` users).

If you are happy to do that work, that would be great - it's really appreciated, but if it turns into a major hassle, let us know and we will find another way forward.

I'm not completely opposed to putting this in `origin`, but if we can find a way to make it all uniform I think it would be an improvement.

robbmanes:
Closing in favor of [PR#1040](https://github.com/sosreport/sos/pull/1040) as this adds this functionality to the OpenVSwitch plugin instead.

Issue #782, opened 2016-03-03
[openvswitch] plugin fails to collect information sometimes


codificat:
In an OpenShift environment:

```
# cat sos_commands/openvswitch/ovs-ofctl_dump-flows_br0
2016-03-01T21:00:15Z|00001|vconn|WARN|unix:/var/run/openvswitch/br0.mgmt: version negotiation failed (we support version 0x01, peer supports version 0x04)
ovs-ofctl: br0: failed to connect to socket (Broken pipe)
```

Specifying `-O OpenFlow13` works, but modifying the plugin to specify the version doesn't sound right. Well, maybe adding an extra line with the version explicitly might be an acceptable hack, but is there a better way?

bmr-cymru:
First question is why is it happening?

I am assuming this is similar to the problems we've had with NetworkManager protocol version checks - the copy in the file system has been updated but the running daemon still uses the old version?

(although in the output above it seems like the client has the newer version?)

There are some valid reasons for these services not restarting themselves (e.g. in NM's case it would kill all running VPN sessions etc.) but this is somewhat painful for us to deal with in sos.

Let's confirm the cause in this instance and then we can try to work out a strategy for handling it.

codificat:
The man page for `ovs-ofctl(8)` in `openvswitch-2.4.0-1.el7` says:

> Because support for OpenFlow 1.1 and later is still experimental in Open vSwitch,
> it is necessary to explicitly enable these protocol  versions  in  ovs-ofctl
> (using -O) and in the switch itself (with the protocols column in the Bridge table).
> For more information, see ``Q: What versions of OpenFlow does Open vSwitch support?'' in the Open vSwitch FAQ.

This seems to be [the FAQ in question](https://github.com/openvswitch/ovs/blob/master/FAQ.md#q-what-versions-of-openflow-does-open-vswitch-support), and the info there doesn't seem to match the man page above, as it suggests that OF1.3 is enabled by default starting in ovs 2.3.

So, there's some conflicting documentation here, would have to dig the sources/package to determine exact details.

Listing all protocols explicitly in `-O` should avoid relying on the (apparently fragile) auto-negotiation. Should we? (at the risk of specifying protocol versions not supported by the installed ovs...)

pmoravec:
"Listing all protocols explicitly in -O" - will the negotiation choose some of the supported version? I.e. specifying `-O X,Y` and ovs supporting just Y, will the protocol version Y be chosen?

[...]

PR #1040, opened 2017-06-15
[openvswitch] get specific flow version commands

robbmanes:
The output of ovs-ofctl varies depending on the version of OpenFlow you
query it with.  Using the currently supported versions, retrieve
ovs-ofctl output for each ovs-bridge depending on which specified
versions of OpenFlow are in use.

---

This is a continuation of the discussion [here](https://github.com/sosreport/sos/pull/1039), where we need this information when troubleshooting origin issues.  However, it was discussed that collection of OpenFlow-specific version information probably is better suited in the general OpenVSwitch plugin rather than one-off collecting it in origin.  This should meet that need and negate PR#1039.

The above patch has been tested under load in OSE and regular vswitch environments and does not appear to show any regression.  Tests passed locally, including pep8 as well.

### DIFF
--- a/sos/plugins/openvswitch.py
+++ b/sos/plugins/openvswitch.py
@@ -83,6 +83,29 @@ class OpenVSwitch(Plugin):
                     "ovs-ofctl show %s" % br
                 ])
 
+                # Flow protocols currently supported
+                flow_versions = ["OpenFlow10",
+                                 "OpenFlow11",
+                                 "OpenFlow12",
+                                 "OpenFlow13"]
+
+                # List protocols currently in use, if any
+                br_info_file = self.get_cmd_output_now(
+                                   "ovs-vsctl list bridge %s" % br)
+                br_info = open(br_info_file).read()
+                for line in br_info.splitlines():
+                    if "protocols" in line:
+                        br_protos_ln = line[line.find("[")+1:line.find("]")]
+                        br_protos = br_protos_ln.replace('"', '').split(", ")
+
+                # Collect flow information for relevant protocol versions only
+                for flow in flow_versions:
+                    if flow in br_protos:
+                        self.add_cmd_output([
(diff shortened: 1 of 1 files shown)


codificat:
Thanks for the PR! This should fix #782

bmr-cymru:
Bonus! 👍

sos/plugins/openvswitch.py
```diff
@@ -83,6 +83,29 @@ def setup(self):
                     "ovs-ofctl show %s" % br
                 ])
 
+                # Flow protocols currently supported
+                flow_versions = ["OpenFlow10",
```
bmr-cymru: I'd prefer this in the conventional sos list style:

```python
   foo = [
        "bar",
        "baz",
       "qux"
    ]
```

But I'll fix it up when I merge - this PR has been hanging around long enough already.

sos/plugins/openvswitch.py
```diff
@@ -83,6 +83,29 @@ def setup(self):
...
+                                 "OpenFlow11",
+                                 "OpenFlow12",
+                                 "OpenFlow13"]
+
+                # List protocols currently in use, if any
+                br_info_file = self.get_cmd_output_now(
+                                   "ovs-vsctl list bridge %s" % br)
```
bmr-cymru: Ditto parens/newlines here.

bmr-cymru:
Merged in commit c17c04df0b33fc502dce97d31c03699799d54d24 - thanks!

robbmanes:
Re: style - my bad, I just assumed pep8 and will pay attention to other calls elsewhere.  Thanks!

bmr-cymru:
No problem - it's a "twist" on pep8 (in that it passes the pep8 checker, and is allowed but not required by pep8 -  we just prefer the visual style over the vanilla version :-) ). It helps especially for plugin code as it suits quickly eyeballing multi-line lists and checking for things like missing commas.

PR #1041, opened 2017-06-19
[gluster_block] Added new plugin gluster_block

vredara:
gluster_block plugin collects config and log files related to blocks

Signed-off-by: venkata edara <redara@redhat.com>

---

### DIFF
--- a/sos/plugins/gluster_block.py
+++ b/sos/plugins/gluster_block.py
@@ -1,0 +1,39 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+import glob
+from sos.plugins import Plugin, RedHatPlugin
+
+
+class GlusterBlock(Plugin, RedHatPlugin):
+    """Gluster Block"""
+
+    plugin_name = 'gluster_block'
+    profiles = ('storage',)
+    packages = ("gluster-block",)
+    files = ("/usr/sbin/gluster-block",)
+
+    def setup(self):
+
+        # collect logs - apply log_size for any individual file
+        # all_logs takes precedence over logsize
+        if not self.get_option("all_logs"):
+            limit = self.get_option("log_size")
+        else:
+            limit = 0
+
+        if limit:
+            for f in glob.glob("/var/log/gluster-block/*.log"):
+                self.add_copy_spec(f, limit)
+        else:
+            self.add_copy_spec("/var/log/gluster-block")


bmr-cymru:
In future please fill out the pull request template - it is there to save everyone time (including you, by avoiding unnecessary requests to respin the branch).

sos/plugins/gluster_block.py
```diff
@@ -0,0 +1,45 @@
...
+from sos.plugins import Plugin, RedHatPlugin
+
+
+class GlusterBlock(Plugin, RedHatPlugin):
+    """Gluster Block"""
+
+    plugin_name = 'gluster_block'
+    profiles = ('storage',)
+    packages = ["gluster-block"]
+    files = ["/usr/sbin/gluster-block"]
```
bmr-cymru: Why lists for these? The normal `sos` style is to use a tuple (since they are invariant and do not need the baggage of a mutable list type).

sos/plugins/gluster_block.py
```diff
@@ -0,0 +1,45 @@
...
+    config_to_collect = ["/sys/kernel/config/target", "/etc/target"]
+
+    def setup(self):
+        self.add_cmd_output("targetcli ls")
```
bmr-cymru: Nack. We have the `targetcli` plugin for that.

sos/plugins/gluster_block.py
```diff
@@ -0,0 +1,45 @@
...
+
+        # collect config files
+        for config_file in self.config_to_collect:
+            self.add_copy_spec(config_file)
```
bmr-cymru: Nack. Please look at the API documentation. The `add_copy_spec()` method accepts either a single string, or a list of strings. This is just pointless make-work.

sos/plugins/gluster_block.py
```diff
@@ -0,0 +1,45 @@
...
+
+        # collect logs - apply log_size for any individual file
+        # all_logs takes precedence over logsize
+        if not self.get_option("all_logs"):
+            limit = self.get_option("log_size")
+        else:
+            limit = 0
+
```
bmr-cymru: Ack, thanks for implementing log limiting.

sos/plugins/gluster_block.py
```diff
@@ -0,0 +1,45 @@
...
+        if limit:
+            for f in glob.glob("/var/log/gluster-block/*.log"):
```
bmr-cymru: Ack. This is ugly, but it is in keeping with other plugins that try to work around the current limits API shortcomings. We can clean it up later when those are addressed.

sos/plugins/gluster_block.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class GlusterBlock(Plugin, RedHatPlugin):
+    """Gluster Block"""
+
+    plugin_name = 'gluster_block'
+    profiles = ('storage',)
+    packages = ["gluster-block"]
+    files = ["/usr/sbin/gluster-block"]
+    config_to_collect = ["/sys/kernel/config/target", "/etc/target"]
```
bmr-cymru: Nack: you're just duplicating the `targetcli` plugin under a different name!

sos/plugins/gluster_block.py
vredara: ok. I am changing lists to tuples.

sos/plugins/gluster_block.py
vredara: Didnt observed the targetcli plugin,  I will remove targetcli from gluster-block plugin.

sos/plugins/gluster_block.py
vredara: ok, will collect only "/sys/kernel/config/target" which is not being collected by targetcli.

sos/plugins/gluster_block.py
vredara: Sure. removed this cmd.

sos/plugins/gluster_block.py
bmr-cymru: It need to go into the `targetcli` plugin - you can either file a separate pull request for that, or just amend your existing commit (to address the review items), and then add a new commit with the change for `targetcli`. Once that's done you can update the branch for the PR with `git push --force`.

Note that we don't accept 'fixup' commits in master so you'll need to respin the branch anyway for the review items.

vredara:
I changed the code in latest commit. please review the latest changes.

bmr-cymru:
Please don't push fixup commits (`"update gluster_block.py"` - and even without being a fixup, that subject does not comply with the commit formatting guidelines - every commit must be in `[component] short description of change` format).

Use `git reset` and `git commit --amend`, or `git rebase -i` to produce a single commit with the changes you want to merge, and then push the result with `git push --force`.

sos/plugins/gluster_block.py
```diff
@@ -20,16 +20,13 @@ class GlusterBlock(Plugin, RedHatPlugin):
...
+    files = ("/usr/sbin/gluster-block",)
 
     def setup(self):
-        self.add_cmd_output("targetcli ls")
 
-        # collect config files
-        for config_file in self.config_to_collect:
-            self.add_copy_spec(config_file)
+        # collect config
+        self.add_copy_spec("/sys/kernel/config/target")
```
bmr-cymru: As discussed, this needs to go in the `targetcli` plugin.

sos/plugins/gluster_block.py
```diff
@@ -43,3 +40,4 @@ def setup(self):
                 self.add_copy_spec(f, limit)
         else:
             self.add_copy_spec("/var/log/gluster-block")
+
```
bmr-cymru: Other than the remaining `targetcli` line these changes look fine, once they are folded into the original commit and the branch updated.

vredara:
Sorry for too many updates. I am contributing to sos report for 1st time. I made code changes according to guidelines and submitted new commit. the new commit also has target cli changes in targetcli plugin only. so gluster block plugin only collects logs related to it.

vredara:
can you please review my latest commit. I incorporated all the comments.

bmr-cymru:
You should either note that `targetcli` change in the commit message or (better, but I won't insist on it), in a separate commit (although technically, to comply with our commit formatting, you'd then need to have a subject of `[targetcli, gluster_block] add new plugin and one item to targetcli`, which just looks weird..).

It's obfuscating to hide a change to another plugin inside a commit that say "add a new plugin"; it's potentially confusing for people reading the logs and trying to track down changes.

The point is to make it easy for folks reviewing for changes, or filtering commits by affected plugins.

vredara:
I migrated all changes of targetcli to other commit of branch 
"target-cli-collect-kernel-config". I created new pull request for 
targetcli separately.

gluster-block-plugin has only gluster-block changes.

Let me know if the commit changes are good.

-Venkat

bmr-cymru:
Looks good, thanks for re-working. 👍

PR #1042, opened 2017-06-21
[tripleo] Add ui logs

honza:
---

### DIFF
--- a/sos/plugins/openstack_instack.py
+++ b/sos/plugins/openstack_instack.py
@@ -75,6 +75,12 @@ class OpenStackInstack(Plugin):
                             deployment + ".log",
                             timeout=600)
 
+            # get tripleo-ui logs from swift
+            cmd = ("openstack object tripleo-ui-logs tripleo-ui.log "
+                   "--filename /var/log/tripleo-ui.log")
+            self.call_ext_prog(cmd)
+            self.add_cmd_output("/var/log/tripleo-ui.log")
+
     def postproc(self):
         protected_keys = [
             "UNDERCLOUD_TUSKAR_PASSWORD", "UNDERCLOUD_ADMIN_PASSWORD",


sos/plugins/openstack_instack.py
```diff
@@ -75,6 +75,12 @@ def setup(self):
                             deployment + ".log",
                             timeout=600)
 
+            # get tripleo-ui logs from swift
+            cmd = ("openstack object tripleo-ui-logs tripleo-ui.log "
+                   "--filename /var/log/tripleo-ui.log")
+            self.call_ext_prog(cmd)
+            self.add_cmd_output("/var/log/tripleo-ui.log")
+
```
bmr-cymru: This pattern (generating a file by running a command in `setup()` and then collecting it in the same method) is frowned upon for several reasons:

* It breaks the promises we make in the `sos` disclaimer that we will not modify system configuration
    - We either have to leave the file lying around, or clean it up (using a postproc hook). Neither is very clean/elegant.
* Command execution and collection are asynchronous and likely to become more so over time; the `sos` APIs make no guarantee that the call to the program and the file collection occur in the order you expect.

Today that is the case for the code you've written (since `call_ext_prog()` runs the commands synchronously), but it's not something we encourage users to rely on - generally, any events in any `sos` phase (other than the flow of code in a given method or function) can be re-ordered, and within limits, the phases themselves can be re-ordered.

If this path (`/var/log/tripleo-ui.log`) normally exists on the system, then that's acceptable - but it does need a comment in the `sos` code to indicate that this is the case, and that that is why it's OK for us to leave the file there.

If it's not normally present then it would be better if we can find another way to do this - most UNIX programs will accept `-` or some other special indicator to direct them to write to `stdout`. That would be fine here as we could just collect the data via `add_cmd_output()`.

If this isn't really urgent I'd suggest taking a bit of time to see if we can't find a better way to get these logs.

honza:
Awesome! Thanks for the detailed review. I'll abandon this change for now, and seek to make the suggested changes to `openstack object`.

bmr-cymru:
👍 cool - let us know how things go. If it is a problem on the OpenStack side then we can get together and figure something out (we did have something like this in Gluster for a while, which was even messier - magic signals etc., but we do try to avoid it where possible).

PR #1043, opened 2017-06-22
[etcd] Add available port check to get information correctly.

knoha-rh:
etcd v3 no longer listens on port 4001 for backward compatibility of etc
d v2.0 earlier. To collect information, the plugin should check which
port is available to the running etcd. In this patch, etcd plugin will
run port check then curl commands will access to the available port.

Signed-off-by: Keigo Noha <knoha@redhat.com>

---

### DIFF
--- a/sos/plugins/etcd.py
+++ b/sos/plugins/etcd.py
@@ -16,7 +16,6 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
 from sos.plugins import Plugin, RedHatPlugin
-import socket
 
 
 class etcd(Plugin, RedHatPlugin):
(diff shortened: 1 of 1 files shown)


sos/plugins/etcd.py
```diff
@@ -16,21 +16,39 @@
...
+        """
+        self.sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        try:
+            self.sock.connect(("localhost", 2379))
+        except socket.error, msg:
+            self.sock.close()
+            return "4001"
+        self.sock.close()
+        return "2379"
+
```
bmr-cymru: I am not at all keen on adding `socket` usage to plugins. This seems like a fragile, and potentially risky way of doing things (we are assuming, based on nothing more than a lack of exception, that whatever socket responds is really `etcd`, and then later we send a bunch of HTTP requests to that port).

Surely this cannot be the way that other components on the system obtain this port value?

Why can't we either parse configuration, version information, or even e.g. `netstat` output and make a better determination?

sos/plugins/etcd.py
```diff
@@ -16,21 +16,39 @@
...
-        self.add_cmd_output("curl http://localhost:4001/v2/members")
-        self.add_cmd_output("curl http://localhost:4001/v2/stats/leader")
-        self.add_cmd_output("curl http://localhost:4001/v2/stats/self")
-        self.add_cmd_output("curl http://localhost:4001/v2/stats/store")
+        curl_command = "curl -s http://localhost:" + str(self.port_report())
+        self.add_cmd_output(str(curl_command) + "/version")
+        self.add_cmd_output(str(curl_command) + "/v2/members")
+        self.add_cmd_output(str(curl_command) + "/v2/stats/leader")
+        self.add_cmd_output(str(curl_command) + "/v2/stats/self")
+        self.add_cmd_output(str(curl_command) + "/v2/stats/store")
```
bmr-cymru: If `curl_command` is not a string here (requiring coercion with `str()`) then something is already horribly wrong. It's also wasteful to do it five times over.

sos/plugins/etcd.py
```diff
@@ -16,21 +16,39 @@
...
+
     def setup(self):
         self.add_copy_spec("/etc/etcd")
 
-        self.add_cmd_output("curl http://localhost:4001/version")
-        self.add_cmd_output("curl http://localhost:4001/v2/members")
-        self.add_cmd_output("curl http://localhost:4001/v2/stats/leader")
-        self.add_cmd_output("curl http://localhost:4001/v2/stats/self")
-        self.add_cmd_output("curl http://localhost:4001/v2/stats/store")
+        curl_command = "curl -s http://localhost:" + str(self.port_report())
```
bmr-cymru: Please use a format string instead of coercion to string and concatenation (either old or new style formatting is acceptable).

knoha-rh:
Hi,

Thank you for your review. I modified the patch to use etcd.conf for identifying the address and port.
I confirmed its works with etcd-0.4.6-0.12.el7, etcd-2.0.9-2.el7, etcd-2.3.7-4.el7, etcd-3.1.7-1.el7.

sos/plugins/etcd.py
```diff
@@ -16,40 +16,38 @@
...
+        for line in f:
+            if line.startswith('ETCD_LISTEN_CLIENT_URLS='):
+                access_url_port = line[len('ETCD_LISTEN_CLIENT_URLS='):-1]
+            elif line.startswith('[peer]'):
+                peer_flag = True
+            elif line.startswith('addr =') and (peer_flag is False):
+                access_url_port = line[len('addr ='):-1]
+
+        f.close()
+        return access_url_port
```
bmr-cymru: This will raise an exception on an unbound local variable in the case that the configuration file is malformed (e.g. no `ETCD_LISTEN_CLIENT_URLS` present, or no `addr` statement other than within a `[peer]` block.

Since `sos` is designed to help debug situations where configuration may be invalid, or missing entirely, this needs to be more robust.

Also, when making changes to a branch for a pull request, please merge any "fixup" commits back into the main commit (use `git commit --amend` or `git rebase -i`, and then update the branch with `git push --force`).

bmr-cymru: This will assume the port is `4001` in the case that configuration is missing or invalid, is that really the best solution? (e.g. why not use 4001 if the version is <= 2, and 2379 if it is 2 or greater).

Also, please fold any "fixup" commits on PR branches back into the main commit (use `git commit --amend` or `git rebase -i`, and then update the branch with `git push --force`).

sos/plugins/etcd.py
```diff
@@ -16,40 +16,38 @@
...
-            self.sock.connect(("localhost", 2379))
-        except socket.error, msg:
-            self.sock.close()
-            return "4001"
-        self.sock.close()
-        return "2379"
+    def report_url_and_port(self):
+        access_url_port = 'http://localhost:4001'
+        peer_flag = False
+        f = open('/etc/etcd/etcd.conf', 'r')
```
bmr-cymru: You need to handle exceptions whenever you are doing file IO in a plugin; otherwise a single exception during setup or collection will lead to any remaining actions for that plugin being skipped.

bmr-cymru:
Thanks for working on this but I prefer the approach taken by @TurboTurtle in #1059.

PR #1045, opened 2017-06-28
[smartcard] Stop collecting pkcs11_inspect debug

pmoravec:
The tool always requires PIN to be entered manually. That contradicts the aim
of sosreport to collect data as automatically as possible.

Once smartcar community allows providing the PIN e.g. via an env.variable,
we can add the command back.

Resolves: #1045

---


bmr-cymru:
Merged in commit 08458f8637528b71beed573c1f13e36d15c4d18a - thanks!

PR #1046, opened 2017-06-28
[plugins] Tripleo specific containerized services logs

bogdando:
Add Tripleo Pike opinionated logs paths to be collected
for services, when running in containers. This is a temporary
and will be reworked for Queens, like switching those to syslog
or fluentd shipping logs to Elasticsearch cluster.

Partial bug: https://bugs.launchpad.net/tripleo/+bug/1700909
Related blueprint: https://review.openstack.org/#/c/462900

Signed-off-by: Bogdan Dobrelya <bdobreli@redhat.com>

---

### DIFF
--- a/sos/plugins/mongodb.py
+++ b/sos/plugins/mongodb.py
@@ -30,7 +30,8 @@ class MongoDb(Plugin, DebianPlugin, UbuntuPlugin):
     def setup(self):
         self.add_copy_spec([
             "/etc/mongodb.conf",
-            "/var/log/mongodb/mongodb.log"
+            "/var/log/mongodb/mongodb.log",
+            "/var/log/containers/mongodb/mongodb.log"
         ])
 
     def postproc(self):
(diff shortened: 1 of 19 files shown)


bmr-cymru:
Thanks for the pull request - the patch itself looks fine (it's a pity that a number of those plugins do not correctly implement log limiting, but that's not related to this change and we can address it later).

Could you please respin your branch to fix the DCO (Signed-off-by) formatting?

The accepted format for these lines is: `Signed-off-by: Firstname Lastname <mail@example.com>` (with `<...>` surrounding the email address).

It's important since there are many scripts that understand this notation and not having the correct format may confuse them.

You can fix this by hand with `git commit --amend` and then `git commit --force`, or just use `git commit -s --amend` and then force push (git knows how to format the DCO from your stored user information).

bmr-cymru:
Huh. Never mind - the GitHub mail with the commit message seems to be mis-formatted. The actual commit is correct.

PR #1047, opened 2017-06-28
[gnocchi] Tripleo specific containerized services logs

stuggi:
This is an addition to pull request 1046 to collect gnocchi logs
correct when service is running in a container.

Signed-off-by: Martin Schuppert mschuppert@redhat.com

---

### DIFF
--- a/sos/plugins/gnocchi.py
+++ b/sos/plugins/gnocchi.py
@@ -38,11 +38,17 @@ class GnocchiPlugin(Plugin, RedHatPlugin):
 
         self.limit = self.get_option("log_size")
         if self.get_option("all_logs"):
-            self.add_copy_spec("/var/log/gnocchi/",
-                               sizelimit=self.limit)
+            self.add_copy_spec([
+                "/var/log/gnocchi/*",
+                "/var/log/containers/gnocchi/*"],
+                sizelimit=self.limit
+            )
         else:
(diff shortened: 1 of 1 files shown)


sos/plugins/gnocchi.py
```diff
@@ -38,11 +38,17 @@ def setup(self):
 
         self.limit = self.get_option("log_size")
         if self.get_option("all_logs"):
-            self.add_copy_spec("/var/log/gnocchi/",
-                               sizelimit=self.limit)
+            self.add_copy_spec([
+                "/var/log/gnocchi/",
+                "/var/log/containers/gnocchi/"],
+                sizelimit=self.limit
```
bmr-cymru: There's a bug here in the original code (not that you're adding one, but I'd rather we fixed it up as we have the opportunity).

The `limit` parameter does not work effectively when passed a directory (since the size check is applied to the directory inode, which is then passed whole to the archival routines).

Any `add_copy_spec()` call that defines a limit needs to pass a glob that will expand to a set of files (not directories).

pmoravec:
+1

bogdando:
:+1:

bmr-cymru:
Merged as commit f379776da6d2d3f4e3f31bf34b12908674dc122d - thanks!

PR #1049, opened 2017-06-29
[cs] Add Certificate System 9 data capture

83problems:
Red Hat Certificate System 9 (Dogtag 10) has been out for a while.
This commit adds the ability to capture those logs and config
files for each subsystem.

Added pki-base because redhat-pki or dogtag-pki do not have to be
installed to build a Certificate Authority.

Removed "/usr/share/java/pki" check because CS 9 uses the same
directory.

Signed-off-by: Brian Gribble <bgribble@redhat.com>

---

### DIFF
--- a/sos/plugins/cs.py
+++ b/sos/plugins/cs.py
@@ -31,7 +31,10 @@ class CertificateSystem(Plugin, RedHatPlugin):
     packages = (
         "redhat-cs",
         "rhpki-common",
-        "pki-common"
+        "pki-common",
+        "redhat-pki",
+        "dogtag-pki",
+        "pki-base"
     )
 
     files = (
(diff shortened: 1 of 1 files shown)


83problems:
I have a report archive I can attach if requested.

bmr-cymru:
Hi Brian, thanks for the PR - the patch looks fine. If you are OK with what you are seeing in the report tarballs (as a cs/dogtag expert) then I'm happy to merge this - the changes all look logical based on your description & fit into the existing plugin structure.

Will get this pulled in with the next round of merges.

83problems:
Good morning,

Thank you.  The report captures all the logs/configs we would need so I'm happy with it.  

Brian

PR #1050, opened 2017-07-04
[openstack_aodh] OpenStack Alarm support

psachin:
New plugin to collect OpenStack Aodh configurations and logs.

Signed-off-by: Sachin Patil <psachin@redhat.com>

---

### DIFF
--- a/sos/plugins/openstack_aodh.py
+++ b/sos/plugins/openstack_aodh.py
@@ -1,0 +1,78 @@
+# Copyright (C) 2017 Red Hat, Inc., Sachin Patil <psachin@redhat.com>
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+import os
+from sos.plugins import Plugin, RedHatPlugin
+
+
+class OpenStackAodh(Plugin, RedHatPlugin):
+    """OpenStack Alarm service"""
+    plugin_name = "openstack_aodh"
+    profiles = ('openstack', 'openstack_controller')
+
+    packages = (
+        'openstack-aodh-api',
+        'openstack-aodh-listener',
+        'openstack-aodh-notifier',
+        'openstack-aodh-evaluator,'
+        'openstack-aodh-common')
+
+    requires_root = False
+
+    def setup(self):
+        self.add_copy_spec("/etc/aodh/")
+
+        self.limit = self.get_option("log_size")
+        if self.get_option("all_logs"):
+            self.add_copy_spec("/var/log/aodh/",
+                               sizelimit=self.limit)
+        else:
+            self.add_copy_spec("/var/log/aodh/*.log",
+                               sizelimit=self.limit)
+
+        vars_all = [p in os.environ for p in [
+            'OS_USERNAME', 'OS_PASSWORD', 'OS_AUTH_TYPE'
+        ]]
+
+        vars_any = [p in os.environ for p in [
+            'OS_TENANT_NAME', 'OS_PROJECT_NAME'
+        ]]
+
+        if not (all(vars_all) and any(vars_any)):
+            self.soslog.warning("Not all environment variables set. Source "
+                                "the environment file for the user intended "
+                                "to connect to the OpenStack environment.")
+        else:
+            self.add_cmd_output([
+                "aodh --version",
+                "aodh capabilities list",
+                "aodh alarm list"
+            ])
+
+    def postproc(self):
+        self.do_file_sub(
+            "/etc/aodh/aodh.conf",
+            r"(password[\t\ ]*=[\t\ ]*)(.+)",
+            r"\1********"
+        )
+
+        self.do_file_sub(
+            "/etc/aodh/aodh.conf",
+            r"(auth_url[\t\ ]*=[\t\ ]*)(.+)",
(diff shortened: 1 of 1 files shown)


sos/plugins/openstack_aodh.py
```diff
@@ -0,0 +1,75 @@
...
+
+        if not (all(vars_all) and any(vars_any)):
+            self.soslog.warning("Not all environment variables set. Source "
+                                "the environment file for the user intended "
+                                "to connect to the OpenStack environment.")
+        else:
+            self.add_cmd_output([
+                "aodh --version",
+                "aodh capabilities list",
+                "aodh alarm list"])
```
bmr-cymru: The closing `])` should be on a line by itself per `sos` house style (indented to the same level as the opening `self.add_cmd...`).

sos/plugins/openstack_aodh.py
```diff
@@ -0,0 +1,75 @@
...
+        if self.get_option("all_logs"):
+            self.add_copy_spec("/var/log/aodh/",
+                               sizelimit=self.limit)
+        else:
+            self.add_copy_spec("/var/log/aodh/*.log",
+                               sizelimit=self.limit)
+
+        vars_all = [p in os.environ for p in [
+            'OS_USERNAME', 'OS_PASSWORD', 'OS_AUTH_TYPE']]
+
```
bmr-cymru: I'm not a fan of this `[...[\nblah]]` style, but I think we let another one into the tree recently, so I'm happy to let this slide and then fix it up later.

For reference, what I'd prefer is either:

- Assign the list of names to check to a variable (so it all fits onto one line)
- If you have to do it in an open list comprehension, use the `sos` multi-line list convention:

```python
    vars_all = [p in os.environ for p in [
        "FOO", "BAR", "BAZ", "QUX"
    ]]
````

It draws attention to the different parts of the statement better and makes bracketing errors when modifying the lists less likely.

File: sos/plugins/openstack_aodh.py [hunk 3]
```diff
@@ -0,0 +1,75 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+import os
```
bmr-cymru: In future we should probably consider exposing the environment through the `Plugin` class (we are starting to get a lot of plugins importing `os` for this & essentially doing their own thing now).

It's nothing to worry about for this PR though, just noting the concern.

sos/plugins/openstack_aodh.py
```diff
@@ -0,0 +1,75 @@
...
+                               sizelimit=self.limit)
+        else:
+            self.add_copy_spec("/var/log/aodh/*.log",
+                               sizelimit=self.limit)
+
+        vars_all = [p in os.environ for p in [
+            'OS_USERNAME', 'OS_PASSWORD', 'OS_AUTH_TYPE']]
+
+        vars_any = [p in os.environ for p in [
+            'OS_TENANT_NAME', 'OS_PROJECT_NAME']]
```
bmr-cymru: Ditto.

sos/plugins/openstack_aodh.py
```diff
@@ -0,0 +1,75 @@
...
+            "/etc/aodh/aodh.conf",
+            r"(password[\t\ ]*=[\t\ ]*)(.+)",
+            r"\1********"
+        )
+
+        self.do_file_sub(
+            "/etc/aodh/aodh.conf",
+            r"(auth_url[\t\ ]*=[\t\ ]*)(.+)",
+            r"\1********",
+        )
```
bmr-cymru: Ack.

sos/plugins/openstack_aodh.py
psachin: I' d prefer `sos` multi-line over using a variable. Thanks

sos/plugins/openstack_aodh.py
psachin: Sure

sos/plugins/openstack_aodh.py
psachin: Done

sos/plugins/openstack_aodh.py
psachin: Now here's where I got  issue. I actually want to use
```
r"(^auth_url[\t\ ]*=[\t\ ]*)(.+)",
```  
over
```
r"(auth_url[\t\ ]*=[\t\ ]*)(.+)",
```  

to skip comments but I see `codecs.open(path, "r", encoding='utf-8').read()` skips `r"(^auth_url[\t\ ]*=[\t\ ]*)(.+)"`  totally. However contextmanager works fine
```
with codecs.open("/etc/aodh/aodh.conf", "r", encoding='utf-8') as infile:
    for line in infile:
        result, replacement = re.subn(r"(^auth_url[\t\ ]*=[\t\ ]*)(.+)", r"\1****", line)
            if replacement:
            print(result)
```

Nothing specific to this plugin but I want to bring this up.

sos/plugins/openstack_aodh.py
psachin: Agreed. For openstack, "environment variable" check should be moved.

sos/plugins/openstack_aodh.py
bmr-cymru: That looks like a bug - thanks, we'll look into it.

sos/plugins/openstack_aodh.py
```diff
@@ -0,0 +1,78 @@
...
+    """OpenStack Alarm service"""
+    plugin_name = "openstack_aodh"
+    profiles = ('openstack', 'openstack_controller')
+
+    packages = (
+        'openstack-aodh-api',
+        'openstack-aodh-listener',
+        'openstack-aodh-notifier',
+        'openstack-aodh-evaluator,'
+        'openstack-aodh-common')
```
bmr-cymru: Should be in sos list style (which you use pretty consistently throughout the rest of the plugin...). I'll fix during merge.

bmr-cymru:
Merged as commit ed000dcc49c872d28ad5352eedafcb40c9e5da11 - thanks!

PR #1027, opened 2017-05-31
Create plugin to collect NVMe data

rrosatti:
This plugin collects configuration and system information about NVMe devices.
Some of this information is located in /sys/block/<nvme-device>/,
inside the subdirectories: queue/, device/, integrity/ and mq/

The nvme-cli tool is used to get relevant information about firmware,
supported commands, serial, model, usage, etc..

Also some extra commands were used to get information about block-size,
pci location, slot location, vendor id, subsystem id and part number.

PR #1053, opened 2017-07-07
[nvme] Create nvme plugin

rrosatti:
This plugin collects configuration and system information about NVMe
devices.
The nvme-cli tool is used to get relevant information about firmware,
supported commands, serial, model, usage, etc..

Signed-off-by: Rodrigo R. Galvao <rosattig@br.ibm.com>

---

### DIFF
--- a/sos/plugins/nvme.py
+++ b/sos/plugins/nvme.py
@@ -1,0 +1,98 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
+
+from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+import os
+
+
+class Nvme(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
+    """Collect config and system information about NVMe devices"""
+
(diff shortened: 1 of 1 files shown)


rrosatti:
Hey @bmr-cymru , I've closed the pull request #1027 and tried to make it more organized. In this pull request I've cleaned up my fork and restart it from the upstream. Then, I created a new branch and made the changes only for the nvme plugin.

Could you please review this plugin and give me a feedback about what need to be changed?

pmoravec:
+1 from my review.

Technically, the comment lacks "X" inside each '[]' as confirmation you followed Contributor Guidelines, but all the conditions from the Guideliness are met.

bmr-cymru:
Is this the (current) working version of the NVME patch?

rrosatti:
@bmr-cymru yes, it is! 

I'll attach the results when running this plugin.

[sosreport-nvme.tar.gz](https://github.com/sosreport/sos/files/1299317/sosreport-nvme.tar.gz)

bmr-cymru:
Thanks @rrosatti - and thanks for the tarball: it's always helpful to see the results.

bmr-cymru:
I'm going to make a few changes to this as I commit: there's some stuff that really doesn't belong here (that was mentioned in the other long review thread - e.g. slicing up tiny bits of `lsblk` output to store as strings. There's just no need for this as we collect the complete `lsblk` output in the block plugin).

bmr-cymru:
(also, the lsblk stuff looks weird: `get_block_size()` is returning the content of column 3 of lsblk, but that column is SIZE: the closest thing that `lsblk` offers to "block size" are the IO limits and sector size fields:

```
      MIN-IO  minimum I/O size
      OPT-IO  optimal I/O size
     PHY-SEC  physical sector size
     LOG-SEC  logical sector size
```
)

bmr-cymru:
Also the `lscfg` command seems to be a 3rd party tool from IBM that is not packaged in distributions (it is only available via a Yum repository hosted by IBM). We can't include that in sos since it is not universally available to users of NVMe devices (nor will it work everywhere we would want NVMe data, since it appears to be Power-specific).

bmr-cymru:
I've merged the parts that are generic now, thanks! If the remaining pieces have a use for certain systems then we'll need to find a proper place for them in future.

PR #1054, opened 2017-07-09
[openstack plugins] Tripleo specific containerized services configs

stuggi:
Add Tripleo Pike opinionated config paths to be collected
for services, when running in containers. Each service has a
config dir in /var/lib/config-data on the host which gets bind
mounted read only into the container.

---

### DIFF
--- a/sos/plugins/gnocchi.py
+++ b/sos/plugins/gnocchi.py
@@ -33,8 +33,17 @@ class GnocchiPlugin(Plugin, RedHatPlugin):
 
     requires_root = False
 
+    var_puppet_gen = "/var/lib/config-data/puppet-generated/gnocchi"
+
     def setup(self):
-        self.add_copy_spec("/etc/gnocchi/")
+        self.add_copy_spec([
+            "/etc/gnocchi/*",
+            self.var_puppet_gen + "/etc/gnocchi/*",
+            self.var_puppet_gen + "/etc/httpd/conf/*",
+            self.var_puppet_gen + "/etc/httpd/conf.d/*",
+            self.var_puppet_gen + "/etc/httpd/conf.modules.d/wsgi.conf",
+            self.var_puppet_gen + "/etc/my.cnf.d/tripleo.cnf"
+        ])
 
         self.limit = self.get_option("log_size")
         if self.get_option("all_logs"):
@@ -70,5 +79,12 @@ class GnocchiPlugin(Plugin, RedHatPlugin):
             r"password=(.*)",
             r"password=*****",
         )
+        self.do_file_sub(
+            self.var_puppet_gen + "/etc/gnocchi/"
+            "gnocchi.conf",
+            r"password=(.*)",
+            r"password=*****",
+        )
+
 
 # vim: set et ts=4 sw=4 :
--- a/sos/plugins/mongodb.py
+++ b/sos/plugins/mongodb.py
@@ -25,11 +25,18 @@ class MongoDb(Plugin, DebianPlugin, UbuntuPlugin):
     profiles = ('services',)
 
     packages = ('mongodb-server',)
-    files = ('/etc/mongodb.conf',)
+    var_puppet_gen = "/var/lib/config-data/puppet-generated/mongodb"
+
+    files = (
+        '/etc/mongodb.conf',
+        var_puppet_gen + '/etc/mongod.conf'
+    )
 
     def setup(self):
         self.add_copy_spec([
             "/etc/mongodb.conf",
+            self.var_puppet_gen + "/etc/",
+            self.var_puppet_gen + "/etc/systemd/system/mongod.service.d/",
             "/var/log/mongodb/mongodb.log",
             "/var/log/containers/mongodb/mongodb.log"
         ])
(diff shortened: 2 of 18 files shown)


pmoravec:
`/var/lib/config-data` directory can exist only on a some node of a (containerized) OpenStack instance, am I right?

Shall not a plugin relevant for such node instance collect content of that dir, instead of collecting it some-bits in `mongodb`, some-bits in `mysql` or `rabbitmq`? For example collecting `/var/lib/config-data/mysql` shall be triggered at situation, when the system has:

- `mysql` installed
- is some particular OSP instance

I *guess* the 2nd condition can be formalized as "a sos plugin to be run", so the dir shall be collected when *both* plugins should be run / enabled. Methodological question: shall we collect it right in the `mysql` plugin or in the other one?

bmr-cymru:
It depends how you look at the content ownership - in general, I see the "leaf" application (mysql, mongo etc.) as the owner, in which case it is right that these should be in the individual plugins (they are being driven by the OpenStack container bits, but that is not the "leaf" of the ownership tree).

I am not convinced we have the abstraction right yet: I am more starting  to think of these as either warranting a new plugin mechanism (something like "path prefixes" or prefix overrides), or of elevating the OpenStack bits to a new policy/distribution and handling these variations through subclassing.

If time is short for this then we can begin collecting things as we are now and introduce a better approach when it appears: the collected path locations will not change.

stuggi:
Sorry for the delay in response here. /var/lib/config-data can exist on nodes running controller or compute service. Almost all run in container with the next release. Only the undercloud has no container services running (at the moment). Since we started to include the logs for containers in the existing plugins I used the same for the configs. Let me know if things should change and I can start to work on a general change here. But for now I think it would be good if we can start with what we have.

sos/plugins/gnocchi.py
```diff
@@ -34,7 +34,8 @@ class GnocchiPlugin(Plugin, RedHatPlugin):
     requires_root = False
 
     def setup(self):
-        self.add_copy_spec("/etc/gnocchi/")
+        self.add_copy_spec(["/etc/gnocchi/",
+                            "/var/lib/config-data/gnocchi/etc/gnocchi/"])
```
bmr-cymru: Use the sos multi-line list style please.

stuggi:
As discussed, did format cleanup and also introduced a var_puppet_gen variable with the path prefix for better reading on what files is worked on to shorten them when used.

sos/plugins/mongodb.py
```diff
@@ -25,11 +25,18 @@ class MongoDb(Plugin, DebianPlugin, UbuntuPlugin):
...
+    files = (
+        '/etc/mongodb.conf',
+        var_puppet_gen + '/etc/mongod.conf'
+    )
 
     def setup(self):
         self.add_copy_spec([
             "/etc/mongodb.conf",
+            self.var_puppet_gen + "/etc/",
+            self.var_puppet_gen + "/etc/systemd/system/mongod.service.d/"
```
pmoravec: Missing comma at the end of line.

sos/plugins/openstack_glance.py
```diff
@@ -38,15 +39,22 @@ def setup(self):
...
-        self.add_copy_spec("/etc/glance/")
+            self.add_copy_spec([
+                "/var/log/glance/*.log",
+                "/var/log/containers/glance/*.log"
+            ], sizelimit=self.limit)
+
+        self.add_copy_spec([
+            "/etc/glance/",
+            self.var_puppet_gen + "/etc/glance/",
+            self.var_puppet_gen + "/etc/my.cnf.d/"
```
pmoravec: Missing comma or '+' at the end of line? (or missing path for tripleo.cnf?)

pmoravec:
2 comments about typos, otherwise great!

sos/plugins/mongodb.py
stuggi: thx fixed

sos/plugins/openstack_glance.py
stuggi: thx fixed

stuggi:
fixed the 2 issues mentioned above. thx for the pointer.

mandre:
Hello, thanks for adding the path for the configuration files used by TripleO containers, this is great! I'm a bit late to the party but I noticed we're missing the missing config file collection for a few files containerized services. Potentially (I still need to confirm):

- ceph
- cron
- collectd
- etcd
- libvirt
- logrotate
- openstack_aodh
- openstack_cinder
- openstack_ironic
- pacemaker?
- puppet?
- targetcli?

Would it be possible to include them in a subsequent patch?

bmr-cymru:
> Would it be possible to include them in a subsequent patch?

Yes, but they may not make 3.5 at this stage. We still have a backlog of PRs to merge and the release is overdue already.

Generally, this entire approach of scattering this around among many plugins needs to be re-thought, but that can be done after the next release is out.

bmr-cymru:
> Would it be possible to include them in a subsequent patch?

Yes, but they may not make 3.5 at this stage. We still have a backlog of PRs to merge and the release is overdue already.

Generally, this entire approach of scattering this around among many plugins needs to be re-thought, but that can be done after the next release is out.

bogdando:
related LP bug https://bugs.launchpad.net/tripleo/+bug/1700909

PR #1057, opened 2017-07-13
[lustre] Add Lustre filesystem plugin

utopiabound:
This collects some basic information about the configuration
of Lustre on either a client or a server, including:
* Interal ring buffer log
* Basic configuration information
* Quota information
* Configuration of LNet
* JobID info
* LDLM connection states

---

### DIFF
--- a/sos/plugins/lustre.py
+++ b/sos/plugins/lustre.py
@@ -1,0 +1,44 @@
+### This program is free software; you can redistribute it and/or modify
+## it under the terms of the GNU General Public License as published by
+## the Free Software Foundation; either version 2 of the License, or
+## (at your option) any later version.
+
+## This program is distributed in the hope that it will be useful,
+## but WITHOUT ANY WARRANTY; without even the implied warranty of
+## MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+## GNU General Public License for more details.
+
+from sos.plugins import Plugin, RedHatPlugin
+
+class Lustre(Plugin, RedHatPlugin):
+    '''Lustre filesystem'''
+
+    plugin_name = 'lustre'
+    profiles = ('storage','network','cluster',)
+    packages = ('lustre', 'lustre-client',)
+
+    def setup(self):
+        self.add_cmd_output("lctl debug_kernel")
+        self.add_cmd_output("lctl device_list")
+
+        self.get_params("basic", [ "version", "health_check", "debug" ])
+        self.get_params("lnet", [ "peers", "routes", "routers", "nis" ])
+        self.get_params("ldlm-states", [ "*.*.state" ])
+        self.get_params("jobid", [ "jobid_name", "jobid_var" ])
+
+        if (self.is_installed('lustre-client')):
+            self.add_cmd_output("lfs df")
+            self.add_cmd_output("lfs df -i")
+
+        else:
+            self.get_params("osd", [ "osd-*.*.{mntdev,files*,kbytes*,blocksize,brw_stats}" ])
+            self.get_params("quota", [ "osd-*.*.quota_slave.{info,limit_*,acct_*}" ])
+
+    def get_params(self, name, param_list):
+        ''' Use lctl get_param to collect a selection of parameters into a file.
+        '''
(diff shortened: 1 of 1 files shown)


sos/plugins/lustre.py
```diff
@@ -0,0 +1,44 @@
...
+
+        else:
+            self.get_params("osd", [ "osd-*.*.{mntdev,files*,kbytes*,blocksize,brw_stats}" ])
+            self.get_params("quota", [ "osd-*.*.quota_slave.{info,limit_*,acct_*}" ])
+
+    def get_params(self, name, param_list):
+        ''' Use lctl get_param to collect a selection of parameters into a file.
+        '''
+        self.add_cmd_output("lctl get_param "+" ".join(param_list),
+                            suggest_filename="params-"+name,
```
battlemidget: Probably want to use the .format() way here:

```
self.add_cmd_output("lctl get_param {}".format(join(param_lists))
```

bmr-cymru: Or a `%` tuple format (I am still really not sold on Python's new formatting syntax unless it is actually required :-) ).

bmr-cymru:
Tag for this plugin should be the plugin name, i.e. `[lustre] Add new file system plugin` or something like that (we use `plugins` for broad changes, e.g. converting a bunch of similar API uses in one go, and `Plugin` for changes to the base plugin classes).

sos/plugins/lustre.py
```diff
@@ -0,0 +1,44 @@
...
+    '''Lustre filesystem'''
+
+    plugin_name = 'lustre'
+    profiles = ('storage','network','cluster',)
+    packages = ('lustre', 'lustre-client',)
+
+    def setup(self):
+        self.add_cmd_output("lctl debug_kernel")
+        self.add_cmd_output("lctl device_list")
+
```
bmr-cymru: You can do these as a single `add_cmd_output()` call by passing a list of strings, something like:

```python
    self.add_cmd_output([
        "lctl debug_kernel",
        "lctl device_list",
    ])
```

The call overhead isn't an issue - it just looks slightly neater.

sos/plugins/lustre.py
```diff
@@ -0,0 +1,44 @@
...
+        self.get_params("basic", [ "version", "health_check", "debug" ])
+        self.get_params("lnet", [ "peers", "routes", "routers", "nis" ])
+        self.get_params("ldlm-states", [ "*.*.state" ])
+        self.get_params("jobid", [ "jobid_name", "jobid_var" ])
+
+        if (self.is_installed('lustre-client')):
+            self.add_cmd_output("lfs df")
+            self.add_cmd_output("lfs df -i")
```
bmr-cymru: Ditto for this.

sos/plugins/lustre.py
```diff
@@ -0,0 +1,44 @@
...
+
+        else:
+            self.get_params("osd", [ "osd-*.*.{mntdev,files*,kbytes*,blocksize,brw_stats}" ])
+            self.get_params("quota", [ "osd-*.*.quota_slave.{info,limit_*,acct_*}" ])
+
+    def get_params(self, name, param_list):
```
bmr-cymru: For style consistency with other `sos` plugin methods it might be nice to accept a list here: I'll leave that to your judgement however, as unlike commands and paths we would need to pass a tuple of `(name, [ param_list ])` in this case.

Also style-wise we tend to put utility functions at the top of classes, above the `setup()` method, but it isn't a hard and fast rule (lots of plugins probably break it).

utopiabound: I'll push a new branch with the same name and redo the pull request.

bmr-cymru:
In future please don't burn through pull requests just to update the branch - apart from the churn in PR numbers it loses the context of the prior discussion.

Just make changes using `git commit --amend` or `git rebase -i`, and then re-push the result with `git push --force`.

bmr-cymru:
In future please don't burn through pull requests just to update the branch - apart from the churn in PR numbers it loses the context of the prior discussion.

Just make changes using `git commit --amend` or `git rebase -i`, and then re-push the result with `git push --force`.

PR #1058, opened 2017-07-13
[lustre] Add Lustre filesystem plugin

utopiabound:
This collects some basic information about the configuration
of Lustre on either a client or a server, including:
* Interal ring buffer log
* Basic configuration information
* Quota information
* Configuration of LNet
* JobID info
* LDLM connection states

---

### DIFF
--- a/sos/plugins/lustre.py
+++ b/sos/plugins/lustre.py
@@ -1,0 +1,57 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+from sos.plugins import Plugin, RedHatPlugin
+from string import join
+
+
+class Lustre(Plugin, RedHatPlugin):
+    '''Lustre filesystem'''
+
+    plugin_name = 'lustre'
+    profiles = ('storage', 'network', 'cluster', )
+    packages = ('lustre', 'lustre-client', )
+
+    def get_params(self, name, param_list):
+        '''Use lctl get_param to collect a selection of parameters into a
+        file.
+
+        '''
+        self.add_cmd_output("lctl get_param %s" % join(param_list, " "),
+                            suggest_filename="params-%s" % name,
+                            stderr=False)
+
+    def setup(self):
+        self.add_cmd_output([
+            "lctl debug_kernel",
+            "lctl device_list",
+            "lctl list_nids",
+            "lctl route_list",
+            "lnetctl net show -v"
+        ])
+
+        self.get_params("basic", ["version", "health_check", "debug"])
+        self.get_params("lnet", ["peers", "routes", "routers", "nis"])
+        self.get_params("ldlm-states", ["*.*.state"])
+        self.get_params("jobid", ["jobid_name", "jobid_var"])
+
+        # Client Specific
+        self.add_cmd_output([
+                "lfs df",
+                "lfs df -i"
+            ])
+
+        # Server Specific
+        self.get_params("osd", ["osd-*.*.{mntdev,files*," +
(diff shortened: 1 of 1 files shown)


utopiabound:
This is the fixed version of Pull request #1057

bmr-cymru:
I'm just heading out for the day but based on a quick read this looks good - thanks for making the changes.

I see Adam's added his ack so I'll try to get a proper review done in the morning so we can go ahead and merge.

sos/plugins/lustre.py
```diff
@@ -0,0 +1,50 @@
...
+
+        '''
+        self.add_cmd_output("lctl get_param %s" % join(param_list, " "),
+                            suggest_filename="params-%s" % name,
+                            stderr=False)
+
+    def setup(self):
+        self.add_cmd_output(["lctl debug_kernel",
+                             "lctl device_list"])
+
```
bmr-cymru: The conventional style in `sos` is as in the example I gave:

```python
    self.add_cmd_output([
        "command one",
        "command two",
        "command ..."
    ])
```

If you wouldn't mind respinning this and updating the branch, it helps to have as consistent a style throughout as we can (otherwise it's a minor tweak I can fix up when I merge).

sos/plugins/lustre.py
```diff
@@ -0,0 +1,50 @@
...
+        self.get_params("basic", ["version", "health_check", "debug"])
+        self.get_params("lnet", ["peers", "routes", "routers", "nis"])
+        self.get_params("ldlm-states", ["*.*.state"])
+        self.get_params("jobid", ["jobid_name", "jobid_var"])
+
+        if (self.is_installed('lustre-client')):
+            self.add_cmd_output(["lfs df",
+                                 "lfs df -i"])
```
bmr-cymru: Ditto here.

sos/plugins/lustre.py
```diff
@@ -0,0 +1,50 @@
...
+
+        else:
+            self.get_params("osd", ["osd-*.*.{mntdev,files*," +
+                                    "kbytes*,blocksize,brw_stats}"])
+            self.get_params("quota", ["osd-*.*.quota_slave." +
```
bmr-cymru: These are OK as they are; if anything we'd pregenerate the list in a variable, but it does not seem worth it here since there is no real commonality beyond the "osd-*.*." prefix.

sos/plugins/lustre.py
utopiabound: Sorry, I missed the formatting; I'll do that.

utopiabound:
This new patch should address all the previous comments.

utopiabound:
Is there anything else that's needed for this to land?

sos/plugins/lustre.py
```diff
@@ -0,0 +1,54 @@
...
+            "lctl debug_kernel",
+            "lctl device_list"
+        ])
+
+        self.get_params("basic", ["version", "health_check", "debug"])
+        self.get_params("lnet", ["peers", "routes", "routers", "nis"])
+        self.get_params("ldlm-states", ["*.*.state"])
+        self.get_params("jobid", ["jobid_name", "jobid_var"])
+
+        if (self.is_installed('lustre-client')):
```
bmr-cymru: There's no real reason to make this conditional - all command execution in `sos` is speculative (it's not a big deal, but it tends to make plugins cleaner/simpler if they follow the "forgiveness rather than permission" approach).

sos/plugins/lustre.py
bmr-cymru: These may want to be protected by an `if not self.is_installed():` if the previous conditional is removed and these commands will spit error messages - in this case, because the command does exist on the system we will be able to run it, and will collect whatever output it generates in the report.

jgrund:
Is this able to land? Would like to have this in a sos report.

bmr-cymru:
We will try to pull it in for 3.5 (a few weeks overdue currently).

utopiabound:
I've updated the plugin with more information collection.  I removed the server/client distinction and just collect everything as that will be cleaner if for instance, the modules aren't installed via the "normal" rpms.

utopiabound:
This seems to have missed 3.5, can we land it for 3.6?

bmr-cymru:
> This seems to have missed 3.5, can we land it for 3.6?

It just got to the front of the review queue, so I hope so - we're expecting to release 3.6 next month.

PR #1059, opened 2017-07-16
[etcd] Use etcdctl and determine proper listening address

TurboTurtle:
This moves several collected outputs over from using curl to using the
etcdctl command, as well as adding some addition collection.

Only statistic information is now obtained via curl.

Also we now attempt to determine the actual listening address based
on etcd.conf, and if that is not possible the plugin assumes
version-specifc defaults.

---

### DIFF
--- a/sos/plugins/etcd.py
+++ b/sos/plugins/etcd.py
@@ -23,15 +23,53 @@ class etcd(Plugin, RedHatPlugin):
     """etcd plugin
     """
 
-    def setup(self):
-        self.add_copy_spec("/etc/etcd")
+    packages = ('etcd',)
 
-        self.add_cmd_output("curl http://localhost:4001/version")
-        self.add_cmd_output("curl http://localhost:4001/v2/members")
-        self.add_cmd_output("curl http://localhost:4001/v2/stats/leader")
-        self.add_cmd_output("curl http://localhost:4001/v2/stats/self")
-        self.add_cmd_output("curl http://localhost:4001/v2/stats/store")
+    def setup(self):
+        address = self.get_address()
+        self.cmd = 'etcdctl'
+
+        self.add_copy_spec('/etc/etcd')
+
+        subcmds = [
+                   '--version',
+                   'member list',
+                   'cluster-health',
+                   'ls --recursive',
+                   ]
+
+        self.add_cmd_output([
+                             '%s %s' % (self.cmd, sub) for sub in subcmds
+                             ]
+                            )
+
+        urls = [
+                '/v2/stats/leader',
+                '/v2/stats/self',
+                '/v2/stats/store',
+                ]
+
+        if address:
+            self.add_cmd_output([
+                                 'curl -s %s%s' % (address, u) for u in urls
+                                 ]
+                                )
+
         self.add_cmd_output("ls -lR /var/lib/etcd/")
 
+    def get_address(self):
+        try:
+            with open('/etc/etcd/etcd.conf', 'r') as ef:
+                for line in ef:
+                    if line.startswith('ETCD_LISTEN_CLIENT_URLS'):
+                        return line.split('=')[1].replace('"', '').strip()
+        # If we can't read etcd.conf, assume defaults by etcd version
+        except:
+            ver = self.policy().package_manager.get_pkg_list()['etcd']
+            ver = ver['version'][0]
+            if ver == '2':
+                return 'http://localhost:4001'
+            if ver == '3':
+                return 'http://localhost:2379'
 
 # vim: et ts=5 sw=4


sos/plugins/etcd.py
```diff
@@ -23,15 +23,46 @@ class etcd(Plugin, RedHatPlugin):
...
+
+        subcmds = ['--version',
+                   'member list',
+                   'cluster-health',
+                   'ls --recursive',
+                   ]
+
+        for sub in subcmds:
+            self.add_cmd_output("%s %s" % (self.cmd, sub))
+
```
bmr-cymru: Why not form a list of actual commands from a list comprehension on the `subcms` list, and then pass the whole lot into a single `add_cmd_output()`? I think that would be more readable as well as being more terse.

Also please format the list in the conventional `sos` style (opening `[` on the first line by itself, closing `]` on the last line by itself and list content indented one level further than each).

sos/plugins/etcd.py
```diff
@@ -23,15 +23,46 @@ class etcd(Plugin, RedHatPlugin):
...
+        api_urls = ['/v2/stats/leader',
+                    '/v2/stats/self',
+                    '/v2/stats/store',
+                    ]
+
```
bmr-cymru: Ditto list style.

sos/plugins/etcd.py
```diff
@@ -23,15 +23,46 @@ class etcd(Plugin, RedHatPlugin):
...
+        for url in api_urls:
+            self.add_cmd_output("curl -s %s%s" % (self.address, url))
```
bmr-cymru: Ditto list formation rather than driving `add_cmd_putput()` in a loop. This is why these calls all accept a list as argument.

sos/plugins/etcd.py
```diff
@@ -23,15 +23,46 @@ class etcd(Plugin, RedHatPlugin):
...
         self.add_cmd_output("ls -lR /var/lib/etcd/")
 
+    def get_address(self):
+        try:
+            with open('/etc/etcd/etcd.conf', 'r') as ef:
+                for line in ef:
+                    if line.startswith('ETCD_LISTEN_CLIENT_URLS'):
+                        return line.split('=')[1].replace('"', '').strip()
+        # If we can't read etcd.conf, assume defaults by etcd version
+        except IOError:
```
bmr-cymru: What other exceptions can be raised here?

sos/plugins/etcd.py
```diff
@@ -23,15 +23,46 @@ class etcd(Plugin, RedHatPlugin):
...
+            ver = self.get_command_output('etcdctl --version')['output']
+            ver = ver.split('\n')[0].split(':')[1].strip()
+            if ver.startswith('2'):
```
bmr-cymru: Please use the version information that's already in memory, via the `PackageManager` accessible through the `Plugin` instance (this should be sufficient on all distributions now; if it's not we need to fix it).

sos/plugins/etcd.py
```diff
@@ -23,15 +23,46 @@ class etcd(Plugin, RedHatPlugin):
...
+                return 'http://localhost:4001'
+            if ver.startswith('3'):
+                return 'http://localhost:2379'
```
bmr-cymru: There should either be a default case here (in case for whatever reason the version information we receive is garbled or missing), or the calling code needs to know how to deal with a `None` return here: right now the main `setup()` code will just carry on merrily formatting `None` into a bunch of URLs...

sos/plugins/etcd.py
TurboTurtle: Ack. I did the for loop since that's what I used in plugins like `docker`. Fixed.

sos/plugins/etcd.py
TurboTurtle: Ack. Fixed.

sos/plugins/etcd.py
TurboTurtle: Ack. Fixed.

sos/plugins/etcd.py
TurboTurtle: I can't really think of any other we would hit here, so I made it a generic `except` catch.

sos/plugins/etcd.py
TurboTurtle: I didn't see one in `Plugin`, but I did see it `Policy` after taking another look. Used that, please let me know if I'm being blind.

sos/plugins/etcd.py
TurboTurtle: Ack. Fixed this, won't use an empty/`None` value for `address` anymore.

bmr-cymru:
Nope: that's the one (`Plugin -> Policy -> PackageManager)`, thanks.

Some day we should make the abstractions in that class a bit less crazy; they date back to very old `sos`.

pmoravec:
An example how to check package version: https://github.com/sosreport/sos/blob/master/sos/plugins/networking.py#L340-L342

pmoravec:
Hi @TurboTurtle , could you please incorporate the comments such that the PR is merged before sos 3.5 is out in a week-ish timeframe?

TurboTurtle:
@pmoravec the PR was already updated on July 17th with that change.

PR #1060, opened 2017-07-16
[sosreport] Replace optparse with argparse

### LINKED ISSUE #738: [sosreport] replace deprecated optparse with argparse
Link [https://docs.python.org/2/library/optparse.html](https://docs.python.org/2/library/optparse.html)

TurboTurtle:
Optparse is deprecated, this patch converts sos to use argparse instead.

Resolves #738

---

### DIFF
--- a/sos/sosreport.py
+++ b/sos/sosreport.py
@@ -26,7 +26,8 @@ import traceback
 import os
 import errno
 import logging
-from optparse import OptionParser, Option
+
+from argparse import ArgumentParser, Action
 from sos.plugins import import_plugin
 from sos.utilities import ImporterHelper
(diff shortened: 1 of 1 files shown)


bmr-cymru:
Assuming all the packagers agree (we only really support 2.7 and later runtimes upstream today, but downstream may still be supporting older stuff), I don't see a problem with this, so long as it's functionally equivalent.

bmr-cymru:
Nobody has answered to say that they are still supporting 2.6 or older with new upstream releases, so I think we can go ahead with this.

I've had a quick look through and all seems fine - I want to review the `SoSListOption` changes a bit more closely tomorrow but I think we can pull this in then (I've not really done custom options with ArgParse so far).

bmr-cymru:
Closed in commit 4d996d4c91f4d3172109d27a81bd2b0ca4d3bfdf

PR #1063, opened 2017-07-16
[ovirt] Capture domain information for v4 and resolve #947

### LINKED ISSUE #947: sosreport emits: Unable to get ovirt-engine pid
Initially reported on https://bugzilla.redhat.com/show_bug.cgi?id=1427490

currently, the sos plugin 'ovirt.py' is activated by existence of any of several packages, including 'ovirt-engine' and 'ovirt-engine-dwh'.

If dwh is installed on a separate machine (which is a supported configuration), sosreport will emit 'Unable to get ovirt-engine pid'.

We should either check the presence of the package 'ovirt-engine' before checking the pid (and emitting  this error if engine is down), or split this plugin to several different ones, where only an engine-specific one will check engine pid etc.

TurboTurtle:
In ovirt 4.x, the engine-manage-domains tool is no longer used, being
replaced by ovirt-aaa-jdbc-tool. This patch uses the correct utility
based on which is available on the ovirt manager.

Additionally, the jbosstrace option will now only work where
ovirt-engine is installed (instead of just dwh), resolving #947.

---

### DIFF
--- a/sos/plugins/ovirt.py
+++ b/sos/plugins/ovirt.py
@@ -22,6 +22,7 @@ import signal
 
 
 from sos.plugins import Plugin, RedHatPlugin
+from sos.utilities import is_executable
 
 
 # Class name must be the same as file name and method names must not change
@@ -66,7 +67,7 @@ class Ovirt(Plugin, RedHatPlugin):
     ]
 
     def setup(self):
-        if self.get_option('jbosstrace'):
+        if self.get_option('jbosstrace') and self.is_installed('ovirt-engine'):
             engine_pattern = "^ovirt-engine\ -server.*jboss-modules.jar"
             pgrep = "pgrep -f '%s'" % engine_pattern
             lines = self.call_ext_prog(pgrep)['output'].splitlines()
@@ -85,7 +86,26 @@ class Ovirt(Plugin, RedHatPlugin):
         self.add_forbidden_path('/etc/rhevm/.pgpass')
         # Copy all engine tunables and domain information
         self.add_cmd_output("engine-config --all")
-        self.add_cmd_output("engine-manage-domains list")
+
+        # 3.x line uses engine-manage-domains, 4.x uses ovirt-aaa-jdbc-tool
+        if is_executable('engine-manage-domains'):
+            self.add_cmd_output('engine-manage-domains list')
+        if is_executable('ovirt-engine-extensions-tool'):
+            self.add_cmd_output(
+                'ovirt-engine-extensions-tool info list-extensions'
+            )
+        if is_executable('ovirt-aaa-jdbc-tool'):
+            cmd = 'ovirt-aaa-jdbc-tool'
+            subcmds = [
+                'query --what=user',
+                'query --what=group',
+                'settings show'
+            ]
+
+            self.add_cmd_output([
+                '%s %s' % (cmd, sub) for sub in subcmds
+            ])
+
         # Copy engine config files.
         self.add_copy_spec([
             "/etc/ovirt-engine",


TurboTurtle:
Regarding #947, ovirt/rhv is apparently looking to move away from dwh (in fact in 4.x while it is possible to still have dwh running, the product doesn't actually do anything with it at this time. jasperreports was removed in the transition from 3.6 to 4.0) so I don't believe splitting it out into a separate plugin is worth the time. See http://rhelblog.redhat.com/2017/05/23/red-hat-virtualization-reporting-evolution-transitioning-to-metrics-store/


Also, from a support perspective I have only ever seen a single deployment where dwh was running on a non-manager system over ~4 years of supporting RHEV. 

There is a situation in which *both* `engine-manage-domains` and `ovirt-aaa-jdbc-tool` will be present on an ovirt/rhev 3.6 system, as 3.6 was where the conversion to the aaa extension was introduced. There's not a good way for sos to determine if a particular system has been converted yet or not, and in any event both utilities can/will produce usable output.

sos/plugins/ovirt.py
```diff
@@ -85,7 +86,23 @@ def setup(self):
...
+                       'query --what=user',
+                       'query --what=group',
+                       'settings show'
+                       ]
+
+            self.add_cmd_output([
+                                 '%s %s' % (cmd, sub) for sub in subcmds
+                                 ]
+                                )
+
```
mwperina: You cannot easily list domains provided by AAA extensions, but you can list configured extensions using:
```
ovirt-engine-extensions-tool info list-extensions
```
which will produce following content (default installation, no aaa-ldap configured):
```
2017-09-07 11:19:50,789+02 INFO    ========================================================================
2017-09-07 11:19:50,806+02 INFO    ============================ Initialization ============================
2017-09-07 11:19:50,806+02 INFO    ========================================================================
2017-09-07 11:19:50,813+02 INFO    Loading extension 'internal-authz'
2017-09-07 11:19:50,815+02 INFO    Extension 'internal-authz' loaded
2017-09-07 11:19:50,816+02 INFO    Loading extension 'internal-authn'
2017-09-07 11:19:50,817+02 INFO    Extension 'internal-authn' loaded
2017-09-07 11:19:50,817+02 INFO    Initializing extension 'internal-authz'
2017-09-07 11:19:50,817+02 INFO    Extension 'internal-authz' initialized
2017-09-07 11:19:50,818+02 INFO    Initializing extension 'internal-authn'
2017-09-07 11:19:50,819+02 INFO    Extension 'internal-authn' initialized
2017-09-07 11:19:50,819+02 INFO    Start of enabled extensions list
2017-09-07 11:19:50,820+02 INFO    Instance name: 'internal-authn', Extension name: 'Internal Authn (Built-in)', Version: 'N/A', Notes: '', License: 'ASL 2.0', Home: 'http://www.ovirt.org', Author 'The oVirt Project', Build interface Version: '0',  File: '/home/mperina/ovirt-engine/etc/ovirt-engine/extensions.d/internal-authn.properties', Initialized: 'true'
2017-09-07 11:19:50,820+02 INFO    Instance name: 'internal-authz', Extension name: 'Internal Authz (Built-in)', Version: 'N/A', Notes: '', License: 'ASL 2.0', Home: 'http://www.ovirt.org', Author 'The oVirt Project', Build interface Version: '0',  File: '/home/mperina/ovirt-engine/etc/ovirt-engine/extensions.d/internal-authz.properties', Initialized: 'true'
2017-09-07 11:19:50,820+02 INFO    End of enabled extensions list
2017-09-07 11:19:50,821+02 INFO    ========================================================================
2017-09-07 11:19:50,821+02 INFO    ============================== Execution ===============================
2017-09-07 11:19:50,821+02 INFO    ========================================================================
2017-09-07 11:19:50,825+02 INFO    Extension instance='internal-authn' name='Internal Authn (Built-in)' version='N/A'
2017-09-07 11:19:50,825+02 INFO    Extension instance='internal-authz' name='Internal Authz (Built-in)' version='N/A'
```

Using `ovirt-aaa-jdbc-tool` with above commands IMO reveals too much sensitive informations about customer setup and don't cover other extensions like aaa-ldap

bmr-cymru:
Please fix your list formatting! To repeat again, in `sos` we do this for a multi-line list that is passed into a function or method call:

```python
   foo([
        "entry1",
        "entry2",
        "entry3"
    ])
```

For a simple list definition, where it will not fit neatly on one line, it is:

```python
    some_list = [
        "foo",
        "bar",
        "baz",
        "qux"
    ]
```

In summary:

* Opening parens and brackets on the same line
* Each entry on a line by itself, indented one level from the parent line
* All terminating parens and brackets on the same line, on their own, at the outer indentation level

I know this isn't the most widely used convention, but for our purposes, especially in plugin code it works very well (it is easier to eyeball the entries in the list when they are all at the same indent, and it is easier to check for missing commas, parens, and brackets when everything is consistently formatted in this way).

TurboTurtle:
Fixed the formatting - not sure why I have an issue groking that style.

bmr-cymru:
Thanks! 👍

sos/plugins/ovirt.py
```diff
@@ -66,7 +67,7 @@ class Ovirt(Plugin, RedHatPlugin):
     ]
 
     def setup(self):
-        if self.get_option('jbosstrace'):
+        if self.get_option('jbosstrace') and 'ovirt-engine' in self.packages:
```
bmr-cymru: Is this test correct? Given how the plugin is currently written, `'ovirt-engine` in self.packages` will always evaluate to `True` (it's in the hard-coded packages list in the class declaration - normally when you have things checking that list, it's because there are separate `RedHatPlugin` `UbuntuPlugin` etc. versions with variant package lists).

bmr-cymru: I am guessing you actually wanted to do a test against `self.policy().package_manager`...

TurboTurtle: This worked for me, but admittedly I only tested this part once. I'll need to rebuild a test system for this. I'll try and get that done today.

bmr-cymru: Here is the problem. The `oVirt.packages` tuple is assigned exactly once in this file, near the top of the class definition, and containing the `"ovirt-engine"` string:

```python
# Class name must be the same as file name and method names must not change
class Ovirt(Plugin, RedHatPlugin):
    """oVirt Engine"""

    plugin_name = "ovirt"
    profiles = ('virt',)

    packages = (
        'ovirt-engine',
        'ovirt-engine-dwh',
        'ovirt-engine-reports',
        'ovirt-scheduler-proxy',
        'rhevm',
        'rhevm-dwh',
        'rhevm-reports'
    )
```

The `packages` member is never re-assigned or modified in this plugin. This means that the 2nd clause in the test:


```python
        if self.get_option('jbosstrace') and 'ovirt-engine' in self.packages:
```

Can never evaluate `False`. It is exactly equivalent to the line that it replaces. To check whether a package is *installed* (rather than a trigger for the current plugins), you either need to reach into the policy and package manager classes:

```python
        self.policy().package_manager.all_pkgs()['net-tools']
```

Which I admit is hideously ugly (patches welcome! ;-), or you can use the `Plugin.is_installed(package)` method:

```python
        if self.get_option('jbosstrace') and self.is_installed('ovirt-engine'):
```

Which looks even better and more readable than the original.

TurboTurtle: *double checks local branches*
I'm an idiot. I had a different set of changes between my laptop and desktop, and apparently never actually committed from my laptop, which is using `is_installed` (and was the only place I was looking for this one).

Will fix this is about a minute.

bmr-cymru: Np, that's what reviews are for :)

Thanks for fixing it up.

sos/plugins/ovirt.py
```diff
@@ -85,7 +86,22 @@ def setup(self):
...
+        # 3.x line uses engine-manage-domains, 4.x uses ovirt-aaa-jdbc-tool
+        if is_executable('engine-manage-domains'):
+            self.add_cmd_output("engine-manage-domains list")
+        if is_executable('ovirt-aaa-jdbc-tool'):
+            cmd = 'ovirt-aaa-jdbc-tool'
+            subcmds = [
+                'query --what=user',
+                'query --what=group',
+                'settings show'
+            ]
```
mwperina: You cannot easily list domains provided by AAA extensions, but you can list configured extensions using:
```
ovirt-engine-extensions-tool info list-extensions
```
which will produce following content (default installation, no aaa-ldap configured):
```
2017-09-07 11:19:50,789+02 INFO    ========================================================================
2017-09-07 11:19:50,806+02 INFO    ============================ Initialization ============================
2017-09-07 11:19:50,806+02 INFO    ========================================================================
2017-09-07 11:19:50,813+02 INFO    Loading extension 'internal-authz'
2017-09-07 11:19:50,815+02 INFO    Extension 'internal-authz' loaded
2017-09-07 11:19:50,816+02 INFO    Loading extension 'internal-authn'
2017-09-07 11:19:50,817+02 INFO    Extension 'internal-authn' loaded
2017-09-07 11:19:50,817+02 INFO    Initializing extension 'internal-authz'
2017-09-07 11:19:50,817+02 INFO    Extension 'internal-authz' initialized
2017-09-07 11:19:50,818+02 INFO    Initializing extension 'internal-authn'
2017-09-07 11:19:50,819+02 INFO    Extension 'internal-authn' initialized
2017-09-07 11:19:50,819+02 INFO    Start of enabled extensions list
2017-09-07 11:19:50,820+02 INFO    Instance name: 'internal-authn', Extension name: 'Internal Authn (Built-in)', Version: 'N/A', Notes: '', License: 'ASL 2.0', Home: 'http://www.ovirt.org', Author 'The oVirt Project', Build interface Version: '0',  File: '/home/mperina/ovirt-engine/etc/ovirt-engine/extensions.d/internal-authn.properties', Initialized: 'true'
2017-09-07 11:19:50,820+02 INFO    Instance name: 'internal-authz', Extension name: 'Internal Authz (Built-in)', Version: 'N/A', Notes: '', License: 'ASL 2.0', Home: 'http://www.ovirt.org', Author 'The oVirt Project', Build interface Version: '0',  File: '/home/mperina/ovirt-engine/etc/ovirt-engine/extensions.d/internal-authz.properties', Initialized: 'true'
2017-09-07 11:19:50,820+02 INFO    End of enabled extensions list
2017-09-07 11:19:50,821+02 INFO    ========================================================================
2017-09-07 11:19:50,821+02 INFO    ============================== Execution ===============================
2017-09-07 11:19:50,821+02 INFO    ========================================================================
2017-09-07 11:19:50,825+02 INFO    Extension instance='internal-authn' name='Internal Authn (Built-in)' version='N/A'
2017-09-07 11:19:50,825+02 INFO    Extension instance='internal-authz' name='Internal Authz (Built-in)' version='N/A'
```

Using `ovirt-aaa-jdbc-tool` with above commands IMO reveals too much sensitive informations about customer setup and don't cover other extensions like aaa-ldap. So could you please change the code to use `ovirt-engine-extensions-tool` instead?

TurboTurtle: The output of `ovirt-engine-extensions-tool` is not really useful here.

I also don't see how the output from `ovirt-aaa-jdbc-tool` is too sensitive. There are no passwords or anything similar in it.

mwperina: > The output of ovirt-engine-extensions-tool is not really useful here.

It is useful, it's the most closer output we can get in 4.x if you want to list domains similarly as you did with `engine-manage-domains list`

TurboTurtle: I've added `ovirt-engine-extensions-tool` to the patch, but I'd very much like to leave `ovirt-aaa-jdbc-tool` in place as those commands are in there because they've been useful for myself and other support engineers numerous times.

mwperina: OK, so let's have both there, thanks.

PR #1064, opened 2017-07-19
[kernel] Just use modinfo as filename

BryanQuigley:
We've found the modinfo autogenerated name ends up being too long
for some filesystems (specifically ecryptfs).  This just changes
the name to modinfo.  It used to be something like:
modinfo_cpuidle_pcie_aspm_ccp_kernel_pstore_edac_core_pci_hotplug_
rcupdate_mousedev_vt_x_tables_processor_lp_r8169_8250_snd_hda_
intel_xen_netfront_crypto_simd_efivars_i2c_piix4_sysimgblt
_pciehp_macvlan_i2c_designware_core_gpio_generic_uvcvideo_usbhid_ther

---
 [x] Is the commit message split over multiple lines and hard-wrapped at 72 characters?
 [x] Is the subject and message clear and concise?
 [x] Does the subject start with [plugin_name] if submitting a plugin patch or a [section_name] if part of the core sosreport code?
 [x] Does the commit contain a Signed-off-by: First Lastname email@example.com?

### DIFF
--- a/sos/plugins/kernel.py
+++ b/sos/plugins/kernel.py
@@ -34,7 +34,8 @@ class Kernel(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
 
         try:
             modules = os.listdir(self.sys_module)
-            self.add_cmd_output("modinfo " + " ".join(modules))
+            self.add_cmd_output("modinfo " + " ".join(modules),
+                                suggest_filename="modinfo_ALL_MODULES")
         except OSError:
             self._log_warn("could not list %s" % self.sys_module)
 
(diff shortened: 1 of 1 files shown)


bmr-cymru:
Yeah, this has become pretty gross. Otoh, I'd like to not lose the information that it bears (I'm not sure we have another capture for "list of all _present_ kernel modules"?). It may be worth inserting this into `sos_strings` if that's not the case.

Also I think something like `modinfo_ALL_MODULES` is a better name than simply `modinfo` (which is just too close to the cooked name we would give a plain 'modinfo' invocation for my taste).

sos/plugins/kernel.py
```diff
@@ -34,7 +34,8 @@ def setup(self):
 
         try:
             modules = os.listdir(self.sys_module)
-            self.add_cmd_output("modinfo " + " ".join(modules))
+            self.add_cmd_output("modinfo " + " ".join(modules),
+                                suggest_filename="modeinfo")
```
bmr-cymru: 'modeinfo'?? ;-)

Since you're going to need to respin this anyway, let's go with "_ALL_MODULES" or some other uniquifier that the other maintainers agree on.

BryanQuigley:
Updated commit with suggested changes.    Helped /sys/modules to collect all modules (it was missing 10 on my machine).   They are also found in modules.dep afaict - by filename.

PR #1065, opened 2017-07-24
[block] Add the output of 'lsblk -t'

jcastill:
The output of 'lsblk -t' provides very useful information
about the device's topology, like alignment offset,
minimum io size, optimal io size, and others. It is
equivalent to running lsblk with options:

-o NAME,ALIGNMENT,MIN-IO,OPT-IO,PHY-SEC,LOG-SEC,ROTA,SCHED,
RQ-SIZE,WSAME

With this improvement to the block plugin  we can see at one
glance data available in different places like sysfs,
fdisk and/or parted, as well as data not available in the
sosreport nowadays, like
/sys/block/<device>/queue/rotational .

---

### DIFF
--- a/sos/plugins/block.py
+++ b/sos/plugins/block.py
@@ -26,6 +26,7 @@ class Block(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
     def setup(self):
         self.add_cmd_output([
             "lsblk",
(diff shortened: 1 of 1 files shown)


pmoravec:
ACK, useful debug info added.

PR #1066, opened 2017-07-25
[plugins] add unpackaged plugin

pierg75:
This adds a plugin named 'unpackaged'.
It collects a list of files that are not handled by the package manager (it support for now only RHEL
and Fedora distros).
If a file is a link, it will be listed as 'link -> target'.

---

### DIFF
--- a/sos/plugins/unpackaged.py
+++ b/sos/plugins/unpackaged.py
@@ -73,3 +73,5 @@ class Unpackaged(Plugin, RedHatPlugin):
         not_packaged = [x for x in all_fsystem if x not in all_frpm]
         not_packaged_expanded = format_output(not_packaged)
         self.add_string_as_file('\n'.join(not_packaged_expanded), 'unpackaged')
+
+# vim: set et ts=4 sw=4 :
--- a/sos/plugins/unpackaged.py
+++ b/sos/plugins/unpackaged.py
@@ -43,7 +43,8 @@ class Unpackaged(Plugin, RedHatPlugin):
 
             for root, dirs, files in os.walk(path, topdown=True):
                 if exclude:
-                    dirs[:] = [d for d in dirs if dirs not in exclude]
+                    for e in exclude:
+                        dirs[:] = [d for d in dirs if d not in e]
                 for name in files:
                     file_list.append(os.path.join(root, name))
                 for name in dirs:
--- a/sos/plugins/unpackaged.py
+++ b/sos/plugins/unpackaged.py
@@ -66,8 +66,8 @@ class Unpackaged(Plugin, RedHatPlugin):
             return expanded
 
         all_fsystem = []
-        all_frpm = set(self.policy().mangle_package_path(
-                       self.policy().package_manager.files))
+        all_frpm = set(self.policy.mangle_package_path(
+                       self.policy.package_manager.files))
         for d in get_env_path_list():
             all_fsystem += all_files_system(d)
         not_packaged = [x for x in all_fsystem if x not in all_frpm]
--- a/sos/plugins/unpackaged.py
+++ b/sos/plugins/unpackaged.py
@@ -1,0 +1,75 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
+
+import os
+
+from sos.plugins import Plugin, RedHatPlugin
+
+
+class Unpackaged(Plugin, RedHatPlugin):
+    '''
+    Collects a list of files that are not handled by the package
+    manager
+    '''
+
+    def setup(self):
+
+        def get_env_path_list():
+            """
+            It returns a list of directories contained in the $PATH var
+            """
+            return os.environ['PATH'].split(':')
+
+        def all_files_system(path, exclude=None):
+            """
+            path: string
+            exclude: list
+
+            It returns a list of all files installed on a system.
+            It excludes all the files in the directories listed in 'exclude'
+            """
+            file_list = []
+
+            for root, dirs, files in os.walk(path, topdown=True):
+                if exclude:
+                    dirs[:] = [d for d in dirs if dirs not in exclude]
+                for name in files:
+                    file_list.append(os.path.join(root, name))
+                for name in dirs:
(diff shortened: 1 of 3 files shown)


bmr-cymru:
Can we add some performance numbers to this pull request? I.e. runtime tests with and without the new plugin enabled, and ideally on a variety of systems (e.g. a matrix of "everything install"..."minimal", crossed with "no unpackaged"..."many unpackaged").

Alternately, a detailed discussion of the plugin's expected performance characteristics, along with less extensive testing would be adequate.

There was a prior attempt to implement this functionality which was abandoned due to severe performance overheads.

You'll also need to tidy up the branch a bit and update with `git push --force`:

* no merge commits are allowed in `sos` master
* commits like "comment out print statement" are development: they should be dropped, or merged into a parent commit (use `git rebase -i`).
* No commented-out debug statements are allowed in master
* PEP8 errors reported by Travis need fixing

sos/plugins/unpackaged.py
```diff
@@ -0,0 +1,86 @@
...
+    manager
+    '''
+
+    def setup(self):
+
+        def all_files_rpm():
+            """
+            It returns a list of files known by the package manager
+            """
+            return self.get_command_output('rpm -qal')['output'].splitlines()
```
bmr-cymru: It would be preferable to get this from the existing `Packages` dictionary in `Plugin->policy()->PackageManager`. This is already in a distribution agnostic format and it avoids re-running the `rpm` program (and all the side effects that causes).

It would also be a step toward making this distribution neutral.

sos/plugins/unpackaged.py
```diff
@@ -0,0 +1,86 @@
...
+
+        def get_env_path_list():
+            """
+            It returns a list of directories contained in the $PATH var
+            """
+            return os.environ['PATH'].split(':')
+
+        def check_usrmove():
```
bmr-cymru: Nack: this is really a `Policy` item, not `Plugin`. Plugins should either call into the policy or this should be abstracted so that they don't need to care about it.

sos/plugins/unpackaged.py
```diff
@@ -0,0 +1,86 @@
...
+            """
+            It checks if '/bin' and '/sbin' are symlink to '/usr/[s]bin'.
+            It returns True if both paths are symlinks.
+            """
+            return os.path.islink('/bin') and os.path.islink('/sbin')
+
+        def mangle_usrmove(files):
```
bmr-cymru: Ditto.

sos/plugins/unpackaged.py
```diff
@@ -0,0 +1,86 @@
...
+            """
+            path: string
+            exclude: list
+
+            It returns a list of all files installed on a system.
+            It excludes all the files in the directories listed in 'exclude'
+            """
+            file_list = []
+
+            print("path: {}".format(path))
```
bmr-cymru: No `print()`s please: either use the log facility or delete this.

sos/plugins/unpackaged.py
```diff
@@ -0,0 +1,86 @@
...
+
+            return file_list
+
+        all_fsystem = []
+        all_frpm = set(mangle_usrmove(all_files_rpm()))
+        for d in get_env_path_list():
+            all_fsystem += all_files_system(d)
+        not_packaged = [x for x in all_fsystem if x not in all_frpm]
+        not_packaged_expanded = ["{} -> {}".format(x, os.readlink(x)) for x in not_packaged if os.path.islink(x)]
+        #self.add_string_as_file('\n'.join(all_fsystem), 'all_system_files')
```
bmr-cymru: Nack. No commented out statements please - either uncomment or remove.

sos/plugins/unpackaged.py
```diff
@@ -0,0 +1,86 @@
...
+                    dirs[:] = [d for d in dirs if dirs not in exclude]
+                for name in files:
+                    file_list.append(os.path.join(root, name))
+                for name in dirs:
+                    file_list.append(os.path.join(root, name))
+
+            return file_list
+
+        all_fsystem = []
+        all_frpm = set(mangle_usrmove(all_files_rpm()))
```
bmr-cymru: I'd suggest something like a policy or package manager method `mangle_package_path()` - in the default case it simply returns the string unchanged, but for post-UsrMove distros it can implement the necessary path rewriting.

pierg75: Right. I guess this should be a linuxpolicy method...I'll look into that.

pierg75:
This is a run on my notebook with the plugin:

`sudo ./sosreport --batch  25.87s user 9.77s system 56% cpu 1:03.10 total`

And this without:

`sudo ./sosreport --batch  23.16s user 6.99s system 68% cpu 43.976 total`

I'll do more tests later.

pierg75:
Here few more tests:
### rhel7 vm without any 3rd party executable:

- without the unpackaged plugin:

```
### time sudo ./sosreport --batch
...
real    2m20.756s
user    0m42.384s
sys     0m10.286s
```
- with the unpackaged plugin:

```
### time sudo ./sosreport --batch
...
real    2m12.128s
user    0m44.846s
sys     0m10.910s
```
### rhel7 vm with 100 random executable (and the unpackaged plugin enabled):

```
real    2m3.256s
user    0m44.230s
sys     0m11.339s

```
### rhel7 vm with 1000 random executable (and the unpackaged plugin enabled):

```
real    2m13.290s
user    0m44.226s
sys     0m10.312s
```
The tests done without the plugins were done with the master branch as some changes were made to the PackageManager class defaults. All the tests were done with either a complete reboot or a complete cache drop. The fact that the latest tests were a bit quicker than the first could be do to the fact that vm disk is backed up by a dm-cache device, so something may have been served by that speeding up everything a bit.
However I don't expect this plugin to add too much overhead as the 'rpm -qla' commands doesn't take a lot to run on this system (around 1-2 seconds) and then we compare two lists.

bmr-cymru:
@pierg75 are you still working on this? The latest performance numbers look good, but it still needs a bit of work for CodingStyle (esp. the commit messages/formatting).

Let me know and I'll schedule it for re-review for 3.6 if so.

iranzo:
Will this catch also pip installed files right? so this could also close #1200

pierg75:
Apologies for the delay (it's been pretty busy lately).
I've reworked a bit the commits plus some minor changes due to some other changes in the main sos code. There's nothing changed in the main 'unpackaged' plugin logic so performance should be exactly the same.

sos/plugins/unpackaged.py
```diff
@@ -0,0 +1,75 @@
...
+            return expanded
+
+        all_fsystem = []
+        all_frpm = set(self.policy().mangle_package_path(
+                       self.policy().package_manager.files))
+        for d in get_env_path_list():
+            all_fsystem += all_files_system(d)
+        not_packaged = [x for x in all_fsystem if x not in all_frpm]
+        not_packaged_expanded = format_output(not_packaged)
+        self.add_string_as_file('\n'.join(not_packaged_expanded), 'unpackaged')
```
bmr-cymru: Need a blank line & vim modeline here ;) (see other files for an example to c'n'p).

pierg75: Sorry....pushed the fix :)

PR #1067, opened 2017-07-28
[wireless] Add new plugin for capturing wireless info

BryanQuigley:
Currently includes iw list/dev, iwconfig, and iwlist scanning.

---

### DIFF
--- a/sos/plugins/wireless.py
+++ b/sos/plugins/wireless.py
@@ -1,0 +1,34 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+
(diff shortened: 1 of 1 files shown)


bmr-cymru:
I guess the big question is: "do we make a new plugin here, or stuff this into the (increasingly bloated) networking plugin".

Although this is a break with our traditional approach for networking (it contains link level stuff for ethernet devices) I'm actually fine with this proposal - if anything, longer term, I'd like to move those ethernet bits out from `networking`, but to do that we probably need to abstract out the interface iterators so that they can be used in multiple plugins...

bmr-cymru:
I like the simplicity of this plugin compared to `networking`: for now I've merged it as-is (with a couple of minor tweaks on top). If anything I think I would like to see the rest of the `networking` plugin split up along similar lines.

PR #1068, opened 2017-08-01
[kubernetes] Add option to not collect individual namespaces.

jds2001:
When there are a large number of namespaces in a Kubernetes cluster,
collecting all of them individually is slow and inefficient.

Add an option to cause that not to happen. If we use that option, we do
want to gather --all-namespaces though, so make it implicit.

A drawback of this approach is that the 'podlogs' and 'describe' options
won't work if this option is used. However, in a system with thousands
of namespaces, it's quite unlikely that the user will want those
options.

Signed-off-by: Jon Stanley <jonstanley@gmail.com>

---

### DIFF
--- a/sos/plugins/block.py
+++ b/sos/plugins/block.py
@@ -43,7 +43,7 @@ class Block(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
 
         if os.path.isdir("/sys/block"):
             for disk in os.listdir("/sys/block"):
-                if disk in [".",  ".."] or disk.startswith("ram"):
+                if disk.startswith("ram"):
                     continue
                 disk_path = os.path.join('/dev/', disk)
                 self.add_cmd_output([


bmr-cymru:
Thanks for the PR. I'll need to sit down and look at this properly in context since it is changing quite a bit of existing code (on a quick glance it seems fine though).

As we don't permit 'fixup' commits in master could you please squash the 2nd commit into the first and then re-push the branch? (`git rebase -i` with `f` then `git push --force` to update the PR).

jds2001:
Collapsed the commits, sorry for the noise. The extensive changes is more of a conditional on the presence of the new option to a bunch of code, so it's just indentation changes, really. There's a very small amount of net new code here.

bmr-cymru:
> The extensive changes is more of a conditional on the presence of the new option to a bunch of code

Yep, it's just the way the whitespace changes cause the diff to flow - it's not really avoidable, but it does make the change harder to eyeball from a quick glance.

bmr-cymru:
Thanks for merging those two together - I'll try to get this included in this week's batch of reviews.

bmr-cymru:
Discussing this a bit more with support folks I'm inclined to suggest flipping the default around - i.e. requiring an option (say `all_namespaces` or `all_ns`) to enable the old behaviour.

I think this has less scope for annoying users - there are already a few reports of frustration with the time taken to run sos on hosts with many configured.

We can then document the use of `-k kubernetes.all_ns` for those cases where it is needed.

pmoravec:
Hello @jds2001 , do you agree with changing the defaults to not collect all namespaces by default? If so, could you please update the PR?

jds2001:
Totally agree, and my fault for not updating the PR - I'll get some time this weekend to do it.

PR #1075, opened 2017-08-20
[ovirt] Obfuscate passwords in backup config files.

### LINKED ISSUE #1074: [ovirt] Passwords are not obfuscated for backup copies of config files
If a system has a copy of a config file under /etc/ovirt-engine that is of the same name with an added string (e.g. `/etc/ovirt-engine/engine.conf.d/10-setup-pki.conf` and `/etc/ovirt-engine/engine.conf.d/10-setup-pki.conf.20170819`) then the copy does not have the passwords obfuscated.

For example:

```
[root@rhev40dbv1 engine.conf.d]# pwd
/root/sostest/sosreport-rhev40dbv1.gsslab.rdu2.redhat.com-20170820131931/etc/ovirt-engine/engine.conf.d

[root@rhev40dbv1 engine.conf.d]# ls
10-rhevm-doc.conf           10-setup-jboss.conf            11-setup-sso.conf
10-setup-database.conf      10-setup-pki.conf              20-branding-rhev.conf
10-setup-dwh-database.conf  10-setup-pki.conf.20170819tmp  50-ovirt-engine-extension-aaa-jdbc.conf
10-setup-java.conf          10-setup-protocols.conf        README


[root@rhev40dbv1 engine.conf.d]# grep -i password 10-setup-pki.conf
ENGINE_PKI_TRUST_STORE_PASSWORD=********
ENGINE_PKI_ENGINE_STORE_PASSWORD=********
ENGINE_HTTPS_PKI_TRUST_STORE_PASSWORD="${ENGINE_PKI_TRUST_STORE_PASSWORD}"
[root@rhev40dbv1 engine.conf.d]# grep -i password 10-setup-pki.conf.20170819tmp 
ENGINE_PKI_TRUST_STORE_PASSWORD="mypass"
ENGINE_PKI_ENGINE_STORE_PASSWORD="mypass"
ENGINE_HTTPS_PKI_TRUST_STORE_PASSWORD="${ENGINE_PKI_TRUST_STORE_PASSWORD}"
```

TurboTurtle:
Previously, backup copies of config files under /etc/ovirt-engine would
not be obfuscated properly. This patch corrects this by also matching
any string following '.conf' in the file name.

Resolves #1074

---

### DIFF
--- a/sos/plugins/ovirt.py
+++ b/sos/plugins/ovirt.py
@@ -48,7 +48,7 @@ class Ovirt(Plugin, RedHatPlugin):
         /etc/
         (rhevm|ovirt-engine|ovirt-engine-dwh)/
         (engine.conf|ovirt-engine-dwhd.conf)
-        (\.d/.+.conf)?
+        (\.d/.+.conf.*?)?
         $
         """
     )


bmr-cymru:
The change is fine in itself, but are these files actually of any use to us?

TurboTurtle:
They can help from a historical context. They would never be used by ovirt, but I have had a few instances where the "backups" were actually slightly different for one reason or another and having them in the sosreport showed me precisely what was being tinkered with (correct or not). 

It's a narrow use case for sure, but I think we should still capture them.

bmr-cymru:
Are they created automatically or by the users?

TurboTurtle:
Should be entirely user created.

bmr-cymru:
Then I am very reluctant to deliberately collect these in `sos` -  we have never been in the business of collecting backups or other files that the administrator may have copied or left lying around. When they are not configuration owned by a component we can't properly match them up to postprocessing hooks and we are leaving ourselves open to problems when admins carelessly leave sensitive data lying around the file system.

TurboTurtle:
I can certainly understand that. 

Again I can only point to my own experience where seeing these "extra" files showed me that the user was tinkering with something incorrectly; which can be the difference between a 5 minute "this broke because XYZ option was changed to something invalid" RCA and a much longer deep dive.

bmr-cymru:
Perhaps a directory listing could be a compromise, allowing the detection of the presence of these files without the risks of scooping up unknown content? Not as direct but it greatly reduces scope for unwanted inclusions.

TurboTurtle:
Hmm. The problem I see with the directory listing is that if a support engineer sees the "extra" files then they would need to separately request a copy of those file(s) from the user, so having them in the sosreport saves time, effort and some back-and-forth with the user (and it's extremely unlikely that the user will remove the password from the files themselves). In the case of the extra files not being present, the directory listing doesn't really net any benefit.


Re-reading your comment though, I'm thinking we might be in a corner case with ovirt where we *can* match these files for post processing since we have a well defined list of exactly what the sensitive data looks like (`DEFAULT_SENSITIVE_KEYS`)? 

In any case, I know that from the perspective of the support teams, we'd greatly prefer to see the "extra" files still collected. If that's a real show-stopper though, then we could work with the directory listing.

bmr-cymru:
I don't really see how that's different from _any other_ directory where we might see that there are "extra files"?

What's magic about these that we should just grab them all unconditionally?

If, as @sandrobonazzola says in #1074 these are _actually_ created by the service, and have predictable names and content, then we can include them.

If they are random bits of garbage that users frequently scatter around by hand then no, I do not think we can include that safely.

PR #872, opened 2016-09-07
[java] new plugin: get information about installed JARs

msrb:
This plugin collects information about JAR files installed in the
standard locations. The information can be later used when debugging
problems with missing/different than expected classes on classpath.

Signed-off-by: Michal Srb <msrb@redhat.com>

PR #1077, opened 2017-08-21
[jars] Scan only /usr/{share,lib}/java by default

msrb:
Other known locations can be added via:
-k jars.all_known_locations
    
Additional user-specified locations can be added via:
-k jars.append_locations=...
    
Signed-off-by: Michal Srb <michal@redhat.com>

---

### DIFF
--- a/sos/plugins/jars.py
+++ b/sos/plugins/jars.py
@@ -98,7 +98,7 @@ class Jars(Plugin, RedHatPlugin):
                 with zipfile.ZipFile(path) as f:
                     if "META-INF/MANIFEST.MF" in f.namelist():
                         return True
-            except IOError:
+            except (IOError, zipfile.BadZipfile):
                 pass
         return False
 
--- a/sos/plugins/jars.py
+++ b/sos/plugins/jars.py
@@ -31,7 +31,8 @@ class Jars(Plugin, RedHatPlugin):
     profiles = ("java",)
     option_list = [
         ("append_locations", "colon-separated list of additional JAR paths",
-         "fast", "")
+         "fast", ""),
+        ("all_known_locations", "scan all known paths", "slow", False)
     ]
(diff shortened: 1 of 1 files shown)


msrb:
This PR addresses the performance issue reported by @bmr-cymru in https://github.com/sosreport/sos/pull/872#issuecomment-323742775.

bmr-cymru:
Yep - this is roughly what we were thinking of, but with those additional paths available via some new plugin option (`-k jars.all_jars` or `-k jars.extra_paths` or something).

bmr-cymru:
The problem so far mainly shows up on Satellite 6 hosts - these expand millions of files into directories below `/var/lib` (although it's certainly not alone in that).

msrb:
> Yep - this is roughly what we were thinking of, but with those additional paths available via some new plugin option (-k jars.all_jars or -k jars.extra_paths or something).

Yeah, I like that idea. PR updated 😉

bmr-cymru:
Thanks! I think that gets us a long way to addressing this.

mrobson:
An an FYI, this is also problem on dense Openshift nodes. All of the container mounts etc.. sit under: 

/var/lib/origin/openshift.local.volumes/

bmr-cymru:
> /var/lib/origin/openshift.local.volumes/

We are excluding all paths beneath `/var/lib` unless explicitly whitelisted or the `-k all_known_locations` option is given, so the proposed patches will already address this for OpenShift environments.

Users can whitelist additional paths on the command line using `-k append_locations=...` or revert to the prior behaviour (which also includes paths in `/opt` and `/usr/local`) with `all_known_locations`.

bmr-cymru:
Merged in commit 6fc42802b87f95dba1d6bfda49ae158143e7799c.

PR #1078, opened 2017-08-29
[openstack_aodh] OpenStack Aodh config and logs for container env

stuggi:
Add Tripleo Pike opinionated config paths to be collected
when running in containers.

Removed obfuscation for auth_url config parameter as this is useful
and not security critical information to collect.

---

### DIFF
--- a/sos/plugins/openstack_aodh.py
+++ b/sos/plugins/openstack_aodh.py
@@ -1,4 +1,5 @@
 # Copyright (C) 2017 Red Hat, Inc., Sachin Patil <psachin@redhat.com>
+# Copyright (C) 2017 Red Hat, Inc., Martin Schuppert <mschuppert@redhat.com>
 
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


sos/plugins/openstack_aodh.py
```diff
@@ -34,15 +35,28 @@ class OpenStackAodh(Plugin, RedHatPlugin):
...
     def setup(self):
-        self.add_copy_spec("/etc/aodh/")
+        self.add_copy_spec([
+            "/etc/aodh/",
+            "/var/lib/config-data/puppet-generated/aodh/etc/aodh/",
+            "/var/lib/config-data/puppet-generated/aodh/etc/httpd/conf/",
+            "/var/lib/config-data/puppet-generated/aodh/etc/httpd/conf.d/",
+            "/var/lib/config-data/puppet-generated/aodh/etc/httpd/"
+            "conf.modules.d/*.conf",
+            "/var/lib/config-data/puppet-generated/aodh/etc/my.cnf.d/"
```
bmr-cymru: Given that these all have a common prefix, why not store the `/var/lib/config-data/puppet-generated` path in a local, e.g. `var_puppet_generated`, and then cons the rest of the string on in the list?

This would avoid the need to break paths over multiple lines which reduces the readability of the list and makes it harder to parse visually.

sos/plugins/openstack_aodh.py
```diff
@@ -71,9 +85,10 @@ def postproc(self):
         )
 
         self.do_file_sub(
-            "/etc/aodh/aodh.conf",
-            r"(auth_url[\t\ ]*=[\t\ ]*)(.+)",
-            r"\1********",
+            "/var/lib/config-data/puppet-generated/aodh/etc/aodh/aodh.conf",
```
bmr-cymru: Since we have the same path prefix here (although it does not force a linebreak) we could use e.g. `self.var_puppet_generated` or similar.

stuggi:
Thats a good idea, changed. Thanks. I'll do that also for the other pull request I sent in 1054

bmr-cymru:
👍

bmr-cymru:
Thanks - I read it while reviewing and was like "how do you have non-absolute paths here?? Oh...".

I definitely like this version better - much clearer!

bmr-cymru:
This PR no longer seems to be relevant: the majority of the changes in the one commit here are also present in commit d5ac6cd768acbd6a8d2d1de2ec26f127272089f2:

```
commit d5ac6cd768acbd6a8d2d1de2ec26f127272089f2
Author: Martin Schuppert <mschuppert@redhat.com>
Date:   Wed Oct 18 15:50:23 2017 +0200

    [openstack_*, logrotate] Tripleo additional directories
    
    Tripleo Pike opinionated config+log paths to be collected
    for services, when running in containers.
    
    This is an addition to #1054 where a couple of things got missed
    
    Fixes: #1130
    
```
Let me know if anything else was required.

PR #1079, opened 2017-08-30
[rpm] --verify option to check also SSL related packages

pmoravec:
--verify option shall also check zlib, openssl* and nss* packages

Having those packages improperly installed often causes system-wide problems as well.

Resolves: #1079

---

### DIFF
--- a/sos/plugins/rpm.py
+++ b/sos/plugins/rpm.py
@@ -32,6 +32,7 @@ class Rpm(Plugin, RedHatPlugin):
         'java.*', 'perl.*',
         'rpm', 'yum',
         'spacewalk.*',
+        'zlib', 'openssl.*', 'nss.*',
     ]
 
     def setup(self):


bmr-cymru:
I really would like to get the per-plugin verification list stuff worked out. It's something that keeps getting left behind because there are some knotty issues in exposing the verify command via the `PackageManager` class. I don't think they are that complicated though - I am tempted to take another look at that before merging this so that we can try to reduce the bloat in the `rpm` plugin and make the global `--verify` genuinely useful (e.g. you could then do `sosreport --batch --verify --profile=crypto` to *just* verify those packages.

Verification is incredibly useful in support as a "catch all" / "lowest common denominator" for finding problems but we have long refused requests to re-enable `rpm -Va` behaviour on RPM based distributions due to the costs.

sos/plugins/rpm.py
```diff
@@ -32,6 +32,7 @@ class Rpm(Plugin, RedHatPlugin):
         'java.*', 'perl.*',
         'rpm', 'yum',
         'spacewalk.*',
+        'zlib', 'openssl.*', 'nss.*',
```
bmr-cymru: Ack to the packages, but nack to the patch :-D

I kinda overtook this work this week with the new per-plugin verify changes. I think it makes more sense to rebase this on top of that and merge it to master with the rest of those changes.

I'll get a patch added in my tree.

pmoravec:
I like much the work done by @bmr-cymru , just to complete the original request, we should verify also `zlib` and `nss*` packages that should be counted to the base system package list whose malformed installation severely affects the whole system (and it can be tricky to reveal that).

So I will add a PR to add those 2 to relevant `verify_packages`.

bmr-cymru:
Hmm, I guess I should merge that branch... :-)

PR #1082, opened 2017-08-30
[plugins] add_copy_spec called with sizelimit to grab newest files

pmoravec:
Currently, files to collect within sizelimit are ordered alphabethically
by their names. That means e.g. /var/log/secure-20160101 is taken before
/var/log/secure-20170101 - and this file could already be over the size
limit.

Resolves: #1082

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -28,6 +28,7 @@ from time import time
 import logging
 import fnmatch
 import errno
+import operator
 
 # PYCOMPAT
 import six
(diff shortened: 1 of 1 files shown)


pmoravec:
I received yet simplier patch:

    -            files.sort()
    +            if sizelimit and (len(files) > 1):
    +                def getmtime(x):
    +                    try:
    +                        return os.path.getmtime(x)
    +                    except (OSError, FileNotFoundError):
    +                        return -1
    +                files.sort(key=getmtime, reverse=True)
    +            else:
    +                files.sort()

that is better in my eyes. @bmr-cymru, shall I update my PR to this one?

PR #1083, opened 2017-08-31
[apache|logs|nscd] honour --all-logs

pmoravec:
The three plug-ins partially or completely ignore --all-logs option. This PR applies the option completely to the plug-ins.

---

### DIFF
--- a/sos/plugins/apache.py
+++ b/sos/plugins/apache.py
@@ -55,22 +55,25 @@ class RedHatApache(Apache, RedHatPlugin):
 
         self.add_forbidden_path("/etc/httpd/conf/password.conf")
 
+        # determine how much logs to collect
+        self.limit = None if self.get_option("all_logs") else 5
+
         # collect only the current log set by default
-        self.add_copy_spec("/var/log/httpd/access_log", 5)
-        self.add_copy_spec("/var/log/httpd/error_log", 5)
-        self.add_copy_spec("/var/log/httpd/ssl_access_log", 5)
-        self.add_copy_spec("/var/log/httpd/ssl_error_log", 5)
+        self.add_copy_spec("/var/log/httpd/access_log", self.limit)
+        self.add_copy_spec("/var/log/httpd/error_log", self.limit)
+        self.add_copy_spec("/var/log/httpd/ssl_access_log", self.limit)
+        self.add_copy_spec("/var/log/httpd/ssl_error_log", self.limit)
         # JBoss Enterprise Web Server 2.x
-        self.add_copy_spec("/var/log/httpd22/access_log", 5)
-        self.add_copy_spec("/var/log/httpd22/error_log", 5)
-        self.add_copy_spec("/var/log/httpd22/ssl_access_log", 5)
-        self.add_copy_spec("/var/log/httpd22/ssl_error_log", 5)
+        self.add_copy_spec("/var/log/httpd22/access_log", self.limit)
+        self.add_copy_spec("/var/log/httpd22/error_log", self.limit)
+        self.add_copy_spec("/var/log/httpd22/ssl_access_log", self.limit)
+        self.add_copy_spec("/var/log/httpd22/ssl_error_log", self.limit)
         # Red Hat JBoss Web Server 3.x
-        self.add_copy_spec("/var/log/httpd24/access_log", 5)
-        self.add_copy_spec("/var/log/httpd24/error_log", 5)
-        self.add_copy_spec("/var/log/httpd24/ssl_access_log", 5)
-        self.add_copy_spec("/var/log/httpd24/ssl_error_log", 5)
-        if self.get_option("log"):
+        self.add_copy_spec("/var/log/httpd24/access_log", self.limit)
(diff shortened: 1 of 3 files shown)


bmr-cymru:
Aren't you supposed to be on vacation??? :-D

bmr-cymru:
Some funky indentation upsetting pep8, but apart from that ack:

```
sos/plugins/logs.py:33:13: E113 unexpected indentation
sos/plugins/nscd.py:34:13: E113 unexpected indentation
```

pmoravec:
Re pep8 errors: I am unable to find out what proper indentation could be; it must be multiplier of 4 but any one I tried failed with the E113 error :( .

bmr-cymru:
> Re pep8 errors: I am unable to find out what proper indentation could be; it must be multiplier of 4 but any one 
> I tried failed with the E113 error :( .

Nope :)

It's always a multiple of four, except for the times when it isn't... (e.g. a method invocation continuation line which is indented to the level of the first character following the opening parentheses).

In this case though the problem isn't the indent level per-se, but it's the fact that `foo if bar` is not a valid statement in Python (it doesn't terminate in `:` like an `if` or `else` statement, and it does not have the `else qux` that would make it a valid conditional statement.

If you must write it this way the entire conditional must be inside parens:

```python
def qux():
    foo = (None if True
           else False)

qux()
```

With the continuation line indented as for a method or function invocation.

Play around with it - I'm not a fan of split-conditional-in-parens, but maybe it looks best. The alternative is to either put one or more values into a local, or to use non-conditional assignment like:

```python
    self.limit = self.get_option("log_size")
        self.limit = None
```

It's less elegant in some ways but more readable when you have to start playing `(...)` games.

pmoravec:
Thanks for the hint, lesson learned.

PR updated accordingly.

BryanQuigley:
The [logs] pull isn't needed AFAICT.  If --all-logs is set, we only run (with no size limit):
 self.add_copy_spec("/var/log/")

bmr-cymru:
> self.add_copy_spec("/var/log/")

Only in the DebianPlugin, and fwiw I think it's a terrible idea :-)

I'm not so familiar with the default logging config on the Debian derived distros, but it's common for this path to contain several GiB of data - part of the idea of moving to finer-grained plugins is to give users more options for limiting and filtering the data collected. The logs plugin should only be dealing with things that are logged by the system logging facilities (plus a few legacy bits, esp. in the Red Hat case). So today that means `syslog`, `rsyslog`, and `journalctl`. Plenty of other subsystems write logs into `/var/log` and `logs` doesn't have any business collecting these - especially not when there is already an existing plugin doing just that!

bmr-cymru:
Maybe we should rename it "system_logs"? I don't like long names, but "syslog" is too confusing due to the logger of the same name.

BryanQuigley:
I think it makes some sense that if you say --all-logs it just collects all of /var/log.  But I get that's not what we should be striving for.  

Doesn't change the fact that the commit "[logs] honour --all-logs every time", doesn't actually do anything in it's current form.

bmr-cymru:
I don't think that's what we've ever communicated `--all-logs` to do, and it runs counter to the model we've been pushing for a good few years now: many small plugins, with strong "ownership": a plugin corresponds to packages, and the plugin "owns" the logs produced by those packages.

The man page description says:

```
Tell plugins to collect all possible log data ignoring any size limits and including logs
in non-default locations. This option may significantly increase the size of reports.
```

Which isn't all that precise one way or the other.

I don't really care that much which way we decide to go as a group, but I do care that it is consistent for users of `sos` regardless of which distribution they are using (modulo necessary changes - the point is our command line options should behave the same on every platform).

My main arguments against scooping `/var/log` are the risk of picking up such a large subtree (I have seen machines where it contains _many_ gibibytes of data), and the fact that it turns `sos` into a bit of a glorified `tar` program.

> "[logs] honour --all-logs every time", doesn't actually do anything in it's current form.

Can you elaborate? It's changing the limit logic from just getting `--log-size` to using `None` if `--all-logs` is given, so it does _something_  (really I think the limit should be 0, but that's separate).

BryanQuigley:
I agree with the first bit.  It's been on my todo to reevaluate it for a while now. 

>Can you elaborate? It's changing the limit logic from just getting --log-size to using None if --all-logs is given, so it does something (really I think the limit should be 0, but that's separate).

If --all-logs is set it never gets to the codeblocks where limit is used.  

Unless I'm missing something I read it as basically:
If --all-logs is TRUE:
set no limit

If --all-logs is FALSE:
Use the limit for these files
else (when --all-logs is TRUE):
Don't use the limit, collect all of /var/log

bmr-cymru:
Oh hah, yeah - iswym for the Debian plugin. The branch conditions are useless with the limit change.

@pmoravec mind having a look at this again if you have some time? Otherwise I'll leave it on my list to fix before release (RH stuff is on priority right now, but it is good if we don't break other distros if we can possibly help it! :).

pmoravec:
I agree the change in `logs.py` is idempotent in Debian plugin. So I will skip the changes there.

pmoravec:
Closing as merged via comits 8abdbbf and 9754966.

PR #1087, opened 2017-09-01
xfs: Add xfs_admin information

trbecker:
Some information about xfs volumes is unavailable when using
xfs_info, namely UUId an label of the volume. This information
is sometimes needed to debug other facilities.

Signed-off-by: Thiago Rafael Becker <thiago.becker@gmail.com>

---

### DIFF
--- a/sos/plugins/cifs.py
+++ b/sos/plugins/cifs.py
@@ -1,0 +1,30 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+
+
+class Cifs(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
+    """Cifs filesystem
+    """
+    packages = ('cifs-utils',)
+    plugin_name = "cifs"
+    profiles = ('storage',)
+
+    def setup(self):
+        self.add_copy_spec([
+            "/proc/fs/cifs"
+        ])
+
+# vim: set et ts=4 sw=4 :
--- a/sos/plugins/xfs.py
+++ b/sos/plugins/xfs.py
@@ -32,6 +32,7 @@ class Xfs(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
             for e in dev:
                 parts = e.split(' ')
                 self.add_cmd_output("xfs_info %s" % (parts[1]))
+                self.add_cmd_output("xfs_admin -l -u %s" % (parts[1]))
 
         if self.get_option('logprint'):
             for dev in zip(self.do_regex_find_all(ext_fs_regex, mounts)):


bmr-cymru:
Please read the [Contributor Guidelines](https://github.com/sosreport/sos/wiki/Contribution-Guidelines) and ensure that your code changes and commit message are formatted accordingly (there is another link in the template that you nuked while submitting this PR).

As well as the subject being mis-formatted, the commit message talks about `xfs_info` but the commit adds an invocation of `xfs_admin`.

All that said, I am not sure I see any point in adding this - the same information is already collected (in a file system agnostic way) from the `blkid` command by the `block` plugin:

```
$ blkid /dev/mapper/vg_hex-lv_images
/dev/mapper/vg_hex-lv_images: LABEL="IMAGES" UUID="63f758a7-3ced-4cbc-9be9-3b766dafeaf8" TYPE="xfs"
$ sudo xfs_admin -l -u /dev/mapper/vg_hex-lv_images 
label = "IMAGES"
UUID = 63f758a7-3ced-4cbc-9be9-3b766dafeaf8
```

We don't collect duplicate information without good cause so this is a nack from my side.

bmr-cymru:
Any chance of responding to the earlier review comments (and getting the commits to comply with the guidelines), before pushing more changes to the branch?

sos/plugins/cifs.py
```diff
@@ -0,0 +1,30 @@
...
+    """Cifs filesystem
+    """
+    packages = ('cifs-utils',)
+    plugin_name = "cifs"
+    profiles = ('storage',)
+
+    def setup(self):
+        self.add_copy_spec([
+            "/proc/fs/cifs"
+        ])
```
bmr-cymru: A whole plugin for a single file is overkill: if there is nothing else to put in here then please add this to the existing `filesys` plugin instead.

Also, please when adding calls to `Plugin` class methods that accept either a single item (string), or a list of items (strings), do not use the list style to add a single element list. It should be just:

```
        self.add_copy_spec("/proc/fs/cifs")
```

sos/plugins/cifs.py
```diff
@@ -0,0 +1,30 @@
...
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+
+
+class Cifs(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
+    """Cifs filesystem
+    """
+    packages = ('cifs-utils',)
```
bmr-cymru: This plugin does not use anything from this package at all. Either get rid of the package trigger, and use a file trigger that is known to exist if this plugin should run (e.g. `/proc/fs/cifs`), or add something that this plugin does actually depend on.

bmr-cymru:
OK. Since there's been no response in 6 months I'll rewrite these commits and either merge or drop them as appropriate.

sos/plugins/cifs.py
bmr-cymru: And looking at this entire directory hierarchy, it's tiny: there's no reason to not collect it all in the filesys plugin.

sos/plugins/xfs.py
```diff
@@ -32,6 +32,7 @@ def setup(self):
             for e in dev:
                 parts = e.split(' ')
                 self.add_cmd_output("xfs_info %s" % (parts[1]))
+                self.add_cmd_output("xfs_admin -l -u %s" % (parts[1]))
```
bmr-cymru: What's the value of adding a special XFS command to gather information that is already captured by the generic `blkid` and `lsblk` data?

```
# xfs_admin -l -u /dev/mapper/vg_hex-lv_images 
[sudo] password for breeves: 
label = "IMAGES"
UUID = 63f758a7-3ced-4cbc-9be9-3b766dafeaf8
```
vs.

```
# blkid /dev/mapper/vg_hex-lv_images
/dev/mapper/vg_hex-lv_images: LABEL="IMAGES" UUID="63f758a7-3ced-4cbc-9be9-3b766dafeaf8" TYPE="xfs"
```

PR #1092, opened 2017-09-05
Collect more firewall related data

pmoravec:
Collect /var/log/firewalld and also ip[|6]tables for all existing tables, not only for the default 3 ones.

---


pmoravec:
Closing as committed via c48fb67 and ac50932.

PR #1094, opened 2017-09-06
[policies] stop execution on Ctrl+C during user input

pmoravec:
When pressing Ctrl+C during user input (name and case id), sosreport
treats that as an empty string and starts to collect data. It must
terminate instead.

It is sufficient to not catch BaseException in the relevant try block.

Resolves: #1094

---


jcastill:
ACK, looks pretty straightforward and reasonable to me.

PR #1095, opened 2017-09-06
[networking] call iptables -L only when iptable_filter module is loaded

pmoravec:
If iptables_filter kernel module isn't loaded, dont call
"iptables -vnxL" that would load that plugin.

The same applies to ip6tables as well.

Kudos to @sbradley7777 to identify the cause and to provide the patch.

Resolves: #1095

---


pmoravec:
Kudos to @sbradley7777 to identify the cause and to provide the patch.

PR #1098, opened 2017-09-07
[origin] Add options to control oc adm diagnostics command invocation

codificat:
Add options to control the invocation of the "oc adm diagnostics" command.

These are useful in certain deployments that can't or don't want to have sosreport invoke the command (the `diag` option), or at least need to prevent the creation of test pods (the `diag-prevent` option).

Defaults are set to maintain the current behaviour.

Also taking the opportunity to remove the only reference to `oadm` in the node config check (replaced by `oc adm`) and clean up `oc` invocation on nodes to remove reference to the admin kubeconfig (which is only present on masters).

Fixes [RHBZ#1463509](https://bugzilla.redhat.com/show_bug.cgi?id=1463509).

---

### DIFF
--- a/sos/plugins/origin.py
+++ b/sos/plugins/origin.py
@@ -14,8 +14,8 @@
 # with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
-from sos.plugins import Plugin, RedHatPlugin
 import os.path
+from sos.plugins import Plugin, RedHatPlugin
 
 # This plugin collects static configuration and runtime information
 # about OpenShift Origin based environments, like OpenShift Enterprise 3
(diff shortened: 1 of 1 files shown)


myllynen:
As said in the RHBZ, this sounds very good, thanks a lot for working on this!

pmoravec:
+1

sos/plugins/origin.py
```diff
@@ -14,8 +14,8 @@
 # with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
-from sos.plugins import Plugin, RedHatPlugin
 import os.path
+from sos.plugins import Plugin, RedHatPlugin
```
bmr-cymru: Style note: typically we import the sos bits first, then python libs - I'll fix this up on merge.

sos/plugins/origin.py
```diff
@@ -45,14 +45,21 @@ class OpenShiftOrigin(Plugin):
     files = None  # file lists assigned after path setup below
     profiles = ('openshift',)
 
+    option_list = [
+        ("diag", "run 'oc adm diagnostics' to collect its output",
+         'fast', True),
+        ("diag-prevent", "set --prevent-modifications on 'oc adm diagnostics'",
+         'fast', False),
```
bmr-cymru: Wouldn't we actually want this `True` by default? Are there any harmful modifications possible is this option is absent?

codificat: There was a discussion about this in [bug 1463509](https://bugzilla.redhat.com/show_bug.cgi?id=1463509#c13) and the conclusion in comment 13 was that we'd introduce these options but leave their defaults to match the current behaviour.

The modifications that the option refers to involve the creation of temporary namespaces and pods to help diagnose network connectivity, and they are expected to be "safe enough" and hence are enabled by default (i.e. `--prevent-modifications` is off by default).

Basically the new option is here to allow disabling when/if these diagnostics can cause a problem (e.g. if the cluster is running disconnected and doesn't have access to the images required to run these diagnostics).

bmr-cymru: Thanks for confirming, I'll defer to that judgement in that case.

bmr-cymru:
I've dropped the import shuffling from the patch as I merged - unless there is some specific reason it's normally preferred to have the `sos` imports first, and then other auxiliary python modules.

PR #1100, opened 2017-09-10
[rabbitmq] Mask default password in rabbitmq.conf

### LINKED ISSUE #840: [rabbitmq] does not elide password in /etc/rabbitmq.config
Hi,
Sosreport as in sos-3.2-36.eo7ost.1.noarch includes in `/etc/rabbitmq/rabbitmq.conf` `default_user` and `default_pass` without being hidden.

Regards,
Pablo

jcastill:
Use do_file_sub() to mask the default password in
the config file /etc/rabbitmq/rabbitmq.conf, to resolve
 #840.

---


BryanQuigley:
Could rabbitmq.config be affected to?  If so, it would be helpful to do both files.  Thanks!

PR #1104, opened 2017-09-19
[zfs] Get zpool properties

aldenso:
Get properties output for all zpools.

Signed-off-by: Aldo Sotolongo <aldenso@gmail.com>

---


sos/plugins/zfs.py
```diff
@@ -32,5 +32,11 @@ def setup(self):
             "zpool list",
             "zpool status -x"
         ])
+        zpools = self.call_ext_prog("zpool list -H -o name")
+        zpools_list = zpools['output'].splitlines()
+        for zpool in zpools_list:
+            self.add_cmd_output("zpool get all {}".format(zpool),
```
bmr-cymru: Ack to the rest of the patch, but please don't use new-style string formatting unless it is either really needed, or provides a genuine readability or formatting advantage - in this case it is just adding ~6 characters to the line unnecessarily. As a project `sos` pre-dates the new style and some distributions are still shipping Python 2.6 which requires positional indexes for new-style format strings (we are hoping to drop support for it in the next release, but still awaiting confirmation from some downstream maintainers).

There are a handful of uses in the tree that got past review (38 vs. 1325 old-style) but we'd prefer to make a change like this as a planned conversion across the whole code base rather than as a piecemeal effort.

aldenso:
thanks for the info, I also added a change, because I needed to verify the status first, but I think I made a mistake with rebase, any tips?

aldenso:
or should I close this PR and create another?

aldenso:
got it, I think is ready now

bmr-cymru:
Once you have your branch rebased you can go ahead and just update the branch in this PR with `git push --force` - no need to start a new one.

I'll be away from this evening until next Monday, but I'll be starting merges for 3.5 then with a view to getting this included.

Thanks,

aldenso:
yes, thank you, it's ready now

PR #1106, opened 2017-09-19
[pacemaker] Clean up Pacemaker plugin and grab new log locations

kgaillot:
Pacemaker upstream is planning to change the default log location with its 2.0 release, so prepare for that now. Also clean up a few minor issues with the plugin.
---

### DIFF
--- a/sos/plugins/pacemaker.py
+++ b/sos/plugins/pacemaker.py
@@ -36,11 +36,19 @@ class Pacemaker(Plugin, DebianPlugin, UbuntuPlugin):
 
     def setup(self):
         self.add_copy_spec([
+            # Pacemaker cluster configuration file
             "/var/lib/pacemaker/cib/cib.xml",
-            self.defaults,
+
+            # Pacemaker 2.x default log locations
+            "/var/log/cluster/pacemaker.log",
+            "/var/log/cluster/bundles/*/",
+
+            # Pacemaker 1.x default log locations
             "/var/log/pacemaker.log",
-            "/var/log/pcsd/pcsd.log",
             "/var/log/pacemaker/bundles/*/",
+
+            self.defaults,
+            "/var/log/pcsd/pcsd.log",
         ])
         self.add_cmd_output([
             "crm_mon -1 -A -n -r -t",
@@ -76,9 +84,9 @@ class Pacemaker(Plugin, DebianPlugin, UbuntuPlugin):
                             (crm_scrub, crm_dest, crm_from),
                             chroot=self.tmp_in_sysroot())
 
-        # collect user-defined logfiles, matching pattern:
-        # PCMK_loggfile=filename
-        # specified in the pacemaker defaults file.
+        # collect user-defined logfiles, matching a shell-style syntax:
+        #   PCMK_logfile=filename
+        # specified in the pacemaker start-up environment file.
         pattern = '^\s*PCMK_logfile=[\'\"]?(\S+)[\'\"]?\s*(\s#.*)?$'
         if os.path.isfile(self.defaults):
             with open(self.defaults) as f:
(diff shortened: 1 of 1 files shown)


sos/plugins/pacemaker.py
```diff
@@ -105,8 +118,9 @@ def setup(self):
         #   PCMK_logfile=filename
         # specified in the pacemaker start-up environment file.
         pattern = '^\s*PCMK_logfile=[\'\"]?(\S+)[\'\"]?\s*(\s#.*)?$'
-        if os.path.isfile(self.defaults):
-            with open(self.defaults) as f:
+        if os.path.isfile(self.envfile):
+            self.add_copy_spec([self.envfile])
```
bmr-cymru: No need to pass a single-entry list here. The `add_copy_spec()` method accepts either a string, or a list of strings - this should be documented in the method docstring: I'll get that fixed up.

sos/plugins/pacemaker.py
```diff
@@ -18,22 +18,44 @@
...
 
     option_list = [
         ("crm_from", "specify the start time for crm_report", "fast", False),
         ("crm_scrub", "enable password scrubbing for crm_report", "", True),
     ]
 
+    envfile = ""
+
+    def setup_crm_mon(self):
+        self.add_cmd_output([
```
bmr-cymru: Ditto for `add_cmd_output()`: you can just pass the "crm_mon" command directly in a single line call.

kgaillot:
Cool, do you prefer a new commit with the additional changes, or replace the existing commits in the request?

Re: postproc, understandable. The postproc here is just password sanitization of the command output. Passwords may appear in three commands' output: crm_report, crm configure show, and pcs config. The user option pacemaker.crm_scrub is intended to determine whether the passwords are left as-is or replaced with stars. Previously, it only applied to crm_report's output -- crm configure show's output was always left as-is, and pcs config's output was always sanitized, which doesn't make any sense. With this change, either they are all sanitized or all left as-is, per the user option (which defaults to sanitized).

bmr-cymru:
Replacing is best - thanks - just push a new HEAD and we can keep everything together under the one PR.

The problem with the postprocessing change is that it conflicts with how this normally works in `sos`, and with what we've long communicated to users - we made an exception for this originally because it was requested by Pacemaker folks, and only applied to how we ran one external command (that was doing its own thing to elide secrets).

I don't have an issue with giving users a mechanism to disable this if it is important to support the components, but we need to find a way to do that that will fit in with these expectations. We've talked in the past about adding command line options to disable some or all of this postprocessing: it might be that the best way to solve this is to revive that and add something like:

```
   -P|--postproc <none|secrets|all>
```

Where 'all' is the default, none disables every `postproc()` hook, and `secrets` allows a plugin-defined behaviour where hooks will only run if they are eliding secrets from the data.

That said I can also see situations where being able to control this on a per-plugin level would also be useful: there are ways that we could approach that with the current plugin options system, for e.g. making `--postproc` be a global plugin option that can be overridden on a per-plugin basis:

```
# sosreport -v --batch --postproc=secrets -k pacemaker.postproc=none
```

We haven't had this type of global-or-per-plugin option in the past but it should be a straightforward mechanism to add.

kgaillot:
Modified:

Password scrubbing commit now always performs scrubbing on pcs/crm output, leaving pacemaker.crm_scrub option to control only crm_report behavior, as before. This actually makes more sense, since we never need passwords from pcs/crm. (BTW, the suggested approaches for disabling postproc aren't entirely useful here, as the crm_report output is sanitized as part of setup.)

The distribution-specific classing commit now avoids unnecessary single-member lists.

bmr-cymru:
> the suggested approaches for disabling postproc aren't entirely useful here, as the crm_report output is sanitized as part of setup

That's true: it would need to be a predicate that could be tested in either method.

kgaillot:
Actually, I'm glad you reminded me ... upstream decided on a different log location for 2.0.0 (which is now in release candidate phase). I'll add one more commit with that location.

kgaillot:
> I'll add one more commit with that location.

Done: 05a8ada (other commits unchanged)

bmr-cymru:
Merged, thanks!

PR #1108, opened 2017-09-20
[kubernetes] Do not collect all namespaces unless requested

jds2001:
Collecting all namespaces on a cluster can be extremely time consuming.
Do not do this unless specifically requested, repurposing the 'all'
module option.

Signed-off-by: Jon Stanley <jonstanley@gmail.com>

---

### DIFF
--- a/sos/plugins/kubernetes.py
+++ b/sos/plugins/kubernetes.py
@@ -30,10 +30,10 @@ class kubernetes(Plugin, RedHatPlugin):
     files = ("/etc/origin/master/master-config.yaml",)
 
     option_list = [
-        ("all", "also collect --all-namespaces output separately",
-            'fast', True),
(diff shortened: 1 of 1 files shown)


jds2001:
This is a continuation of #1068 that I abandoned since the changes are substantive.

One question - how to provide meaningful feedback to the user that if they use 'describe' or 'podlogs' without also using 'all' that those options will have no effect?

pmoravec:
Sounds good to me. reviewing the whole plug-in, I think we shall be more careful in:

           kn = self.get_command_output('%s get namespaces' % kube_cmd)
           knsps = [n.split()[0] for n in kn['output'].splitlines()[1:] if n]

by catching possible `kn['status'] != 0`. But that is out of scope this PR.

Issue #907, opened 2016-12-28
tailing gzipped logfiles should ungzip them first


pmoravec:
Having logfiles to collect with a limit, the latest/oldest file is truncated / only its tail is collected.

If this "border logfile" is gzip file, `sosreport` collects `tail` of gzip archive, i.e. useless sequence of bytes.

Reproducer:

    # ll /var/log/messages*
    -rw-------. 1 root root  297736 Dec 28 15:10 /var/log/messages
    -rw-r--r--. 1 root root 4411245 Dec 28 13:53 /var/log/messages-20161212.gz
    # ./sosreport -o logs -e logs --batch --log-size=1 --build
    ..
    sosreport build tree is located at : /var/tmp/sosreport-pmoravec-rhel72.gsslab.brq.redhat.com-20161228151642
    # file /var/tmp/sosreport-pmoravec-rhel72.gsslab.brq.redhat.com-20161228151642/var/log/messages-20161212.gz
    /var/tmp/sosreport-pmoravec-rhel72.gsslab.brq.redhat.com-20161228151642/var/log/messages-20161212.gz: symbolic link to `../../sos_strings/logs/var.log.messages-20161212.gz.tailed'
    # file /var/tmp/sosreport-pmoravec-rhel72.gsslab.brq.redhat.com-20161228151642/sos_strings/logs/var.log.messages-20161212.gz.tailed
    /var/tmp/sosreport-pmoravec-rhel72.gsslab.brq.redhat.com-20161228151642/sos_strings/logs/var.log.messages-20161212.gz.tailed: data
    # zcat /var/tmp/sosreport-pmoravec-rhel72.gsslab.brq.redhat.com-20161228151642/sos_strings/logs/var.log.messages-20161212.gz.tailed

    gzip: /var/tmp/sosreport-pmoravec-rhel72.gsslab.brq.redhat.com-20161228151642/sos_strings/logs/var.log.messages-20161212.gz.tailed: not in gzip format
    #

pmoravec:
Any fix has to deal with 2 problems:

1. identify file as gzip archive. Or ideally any archive (but `gz` dominates as archive manager for logs, I think).
2. efficiently (and with respect to the current code) unpack the archive

Both might be resolved by using `gzip` library, but that is reported as very slow and also not included in current `sos` prerequisities.

Very naive (and not working) patch (not relying on `gzip`):

    def tail(filename, number_of_bytes):
        """Returns the last number_of_bytes of filename"""
[...]

PR #1111, opened 2017-09-24
[plugins] Don't tail gzipped files when reaching sizelimit

pmoravec:
In case sizelimit in add_copy_spec is reached and tailit=True, we shall
skip tailing a gzip archive that would be collected as damaged.
    
Resolves: #907
    
---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -81,6 +81,15 @@ def _node_type(st):
             return t[1]
 
 
+def _file_is_archive(_file):
+    """ check if the file is an archive - used to decide if we can tail it or
+    skip if sizelimit is reached.
+    The check is done heuristically just by comparing file suffix against four
+    most often used archive algorithms.
+    """
+    return _file.endswith(('.gz', '.xz', '.bz', '.bz2'))
+
+
 class Plugin(object):
     """ This is the base class for sosreport plugins. Plugins should subclass
     this and set the class variables where applicable.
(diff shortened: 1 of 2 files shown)


bmr-cymru:
Thanks for updating the commit - I'm going to tweak this slightly when I merge, to change the name of the test function: technically (pedantically? ;) we are testing to see if the file is _compressed_ - an archive implies some kind of container structure like `tar` or `zip` that is not the case for these files.

sos/plugins/__init__.py
```diff
@@ -517,6 +517,9 @@ def add_copy_spec(self, copyspecs, sizelimit=None, tailit=True):
         file in a glob is too large it will be tailed to meet the sizelimit.
         """
 
+        if self.get_option('all_logs'):
+            sizelimit = None
+
```
bmr-cymru: I don't necessarily disagree with this change, and it potentially simplifies the use of `all_logs` and `log_size` for plugin authors, BUT, there's a big chunk of this patch missing: if we are going to automate this inside of `add_copy_spec()` then there are a _tonne_ of locations in plugin code that need updating to modify or remove `all_logs` logic handling, e.g.:

```python
        self.limit = (None if self.get_option("all_logs")
                      else self.get_option("log_size"))
```

Following this change plugins do not need to modify the size limit themselves when `all_logs` is in effect: that's a good thing, but there's a whole lot of effectively dead code that needs to be reviewed and removed following this change.

bmr-cymru:
Merged in commit 4229326964c05e9caf617ac89e1ec69ab4b5c979

PR #1115, opened 2017-09-27
[policies] Add dynamic policies for RH rebuilds.

AlexBaranowski:
Current state prints Red Hat branding for rebuilds (Scientific Linux,
Centos). Solution provided takes information from /etc/os-release when
distro has it (EL 7). In other cases it take information
from /etc/redhat-release.

Solution is not perfect, os-release hasn't got vendor information.
Older distribution has empty support url.
In both cases vendor is same as name.

Signed-off-by: aleksander.baranowski <aleksander.baranowski@yahoo.pl>

---

### DIFF
--- a/sos/policies/redhat.py
+++ b/sos/policies/redhat.py
@@ -16,12 +16,14 @@
 
 # This enables the use of with syntax in python 2.5 (e.g. jython)
 from __future__ import print_function
+
 import os
+import re
 import sys
 
+from sos import _sos as _
 from sos.plugins import RedHatPlugin
 from sos.policies import LinuxPolicy, PackageManager
-from sos import _sos as _
 
 sys.path.insert(0, "/usr/share/rhn/")
 try:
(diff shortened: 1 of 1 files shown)


bmr-cymru:
Please don't burn through PRs for no reason - if there is a problem with an existing one you can just edit the PR and push a new HEAD to the branch with `git push --force`.

bmr-cymru:
This gets a nack from me: it's over complicated and too reliant on probing things from the running machine.

The policy mechanism is simple and it is simple to introduce new classes to represent the distributions that you are concerned with. It also gives deterministic behaviour as long as the policy detection checks are unambiguous.

The whole idea is that each "variant" distribution is effectively just a class declaration with a handful of instance attributes defined.

bmr-cymru:
Rather than complicating the policy system with the notion of "dynamic" policies, let's just add policy classes for the missing distributions. That's much simpler and more in-keeping with how the framework was designed to be used. I'll file an issue to that effect.

PR #1116, opened 2017-10-02
[openvswitch] Collect all logs

pmoravec:
Do not collect just 2 logfiles but all from the log directory.

Resolves: #1116

---

### DIFF
--- a/sos/plugins/openvswitch.py
+++ b/sos/plugins/openvswitch.py
@@ -25,9 +25,17 @@ class OpenVSwitch(Plugin):
 
     def setup(self):
 
+        all_logs = self.get_option("all_logs")
+        limit = self.get_option("log_size")
+
+        if not all_logs:
+            self.add_copy_spec("/var/log/openvswitch/*.log",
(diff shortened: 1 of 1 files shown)


BryanQuigley:
Commit message doesn't quite match what you are doing.  You're adding the all_logs support (which does what the description says).   You are also adding limit support.

For the code itself, I don't think you want  sizelimit=limit when --all-logs is in play or do we?

bmr-cymru:
That's what it's documented to do:

```
  --all-logs            collect all available logs regardless of size
```

However currently a number of plugins implement this differently: ``--all-logs`` and ``--log-size`` are applied separately. I.e. ``--all-logs`` controls _which_ files we might collect, and ``--log-size`` controls only the limit on the quantity of data.

In some ways, I think this is a better definition of the options - it's cleaner and more consistent (although it does mean users need to supply both in order to collect "everything").

It's something I'd left for the time being, with one thought being to automate this behind the API so that all plugins behave exactly the same - haven't had any time to look at working on that though.

Issue #1103, opened 2017-09-18
Collect postgresql configuration files


alexmcan:
Have sosreport collect files from /etc/postgresql/VERSION/main/*

alexmcan:
@BryanQuigley have a look at this ^

BryanQuigley:
thanks! Likely Debian/Ubuntu specific

PR #1117, opened 2017-10-02
[postgresql] Run on -common package being present too

BryanQuigley:
On Ubuntu and Debian the postgresql package is a meta
package pointing to the latest postgresql package. It's
not required to be installed to have postgresql.

This adds checking for postgresql-common package which has
to be installed if they are running a server.

Closes: #1103

---

### DIFF
--- a/sos/plugins/postgresql.py
+++ b/sos/plugins/postgresql.py
@@ -30,7 +30,7 @@ class PostgreSQL(Plugin):
     plugin_name = "postgresql"
     profiles = ('services',)
 
-    packages = ('postgresql',)
+    packages = ('postgresql', 'postgresql-common')
 
     tmp_dir = None
 


slashdd:
:+1: 

```
$ sudo apt-get install postgresql
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following additional packages will be installed:
  postgresql-9.5 postgresql-client-9.5 postgresql-client-common =>postgresql-common<= postgresql-contrib-9.5
```

Took from d/control postgresql-9.5 source package from LTS Xenial/16.04
```
Package: postgresql-9.5
Architecture: any
Depends: ${misc:Depends},
 ${shlibs:Depends},
 postgresql-client-9.5,
 => postgresql-common (>= 142~), <=
```
Adding postgresql-common is definitely a good approach IMHO.

battlemidget:
@slashdd @BryanQuigley This is applied to all supported distros, is this tested on Fedora as well?

BryanQuigley:
@battlemidget wasn't tested until now - as it should run as long as one package is installed. AFAICT the postgresql plugin doesn't work on centos/6 or fedora/26, that includes stock sosreport.  This patch doesn't fix it though and I haven't been able to grep why it doesn't work... I don't think it's related to the package= line at all though.

Issue #1090, opened 2017-09-05
Support PostgreSQL 9.5 from SoftwareCollections in postgresql plugin


sandrobonazzola:
Once Issue #900 will be available, sos postgresql plugin should be adapted for allowing to collect data from PosgreSQL 9.5 (or possibly generic version in software collections) in addition to the system postgresql setup.

sandrobonazzola:
oVirt project will consume the plugin when available, tracking it on https://bugzilla.redhat.com/show_bug.cgi?id=1427827

pmoravec:
Issue #900 needs some re-work while the author is not further involved / interested. If it is prerequisite of this issue, could you pls. improve that PR?

What in particular is required to be changed? In general: in what new situation (triggered by some package or file presence) we should collect what files or what command outputs?

More technically: shall we enable `postgresql` plugin also when a different package (or file) is present? What particular configs or logs shall we collect? What command outputs? Something secret to scrub?

I.e. technical specification behind this issue is completely missing..

sandrobonazzola:
In practice what's needed is to get the same amount of information within the sos report comparing an install running with a posgresql from EL7 base and an install running postresql95 from software collections.

current result running sosreport on a postgresql base system is that postgresql sos plugin is enabled and collect database data, logs, config, etc... while running on a posgresql95 scl system the postgresql sos plugin is disabled and even forcing it enabled it doesn't collect anything.

expected result is that running sos on a  posgresql95 scl system either the postgresql plugin or a brand new plugin is enabled and collect as much information as the ones collected by postgresql plugin on the base version.

I would expect that if both postgresql base and posgresql95 scl are running on the same host, data from both instances are collected (not an hard requirement on my side).

pmoravec:
OK, so:

- how can `sos` detect `postgresql` is installed from `scl`? (it can detect a package installed or a file present, so what package or file exists (only) on such system?)
- where the logs and configs and commands to execute are placed? (if the locations are similar to current ones, we can extend the current plugin, otherwise we might need to add a new plugin)

Optionally, please provide me e.g. via mail an access to such a system to let me check/guess by myself.

sandrobonazzola:
> how can sos detect postgresql is installed from scl? (it can detect a package installed or a file present, so what package or file exists (only) on such system?)

rh-postgresql95-postgresql-server is installed on systems providing postgresql95 from SCL.

[...]

PR #900, opened 2016-12-05
Implement SCLPlugin mechanism and an example SCLNpm plugin

bkabrda:
This is initial implementation for #898. In the end I took a bit different approach since I figured out that proposal from #898 would require many more conceptual changes in how sos runs plugins. The way I took here basically records all matched SCLs (e.g. that contain certain packages/files) on the SCL Plugin class and it is then up to the plugin to traverse all these SCLs and do something with them. Hope this is still reasonable, so PTAL.

I haven't yet implemented any tests and didn't make the code completely clean. When I do so, do you prefer more commits or should I just ammend to the first commit?

Thanks!

PR #1118, opened 2017-10-03
[postgresql] Collect data for postgreSQL from RHSCL

pmoravec:
Collect postgreSQL data also when postgreSQL is installed from
Red Hat Software Collections.

Resolves: #1090

---

### DIFF
--- a/sos/plugins/postgresql.py
+++ b/sos/plugins/postgresql.py
@@ -1,3 +1,4 @@
+# Copyright (C) 2017 Red Hat, Inc., Pavel Moravec <pmoravec@redhat.com>
 # Copyright (C) 2014 Red Hat, Inc., Sandro Bonazzola <sbonazzo@redhat.com>
 # Copyright (C) 2013 Chris J Arges <chris.j.arges@canonical.com>
 # Copyright (C) 2012-2013 Red Hat, Inc., Bryn M. Reeves <bmr@redhat.com>
@@ -20,7 +21,8 @@
 import os
 import tempfile
 
-from sos.plugins import Plugin, RedHatPlugin, UbuntuPlugin, DebianPlugin
+from sos.plugins import (Plugin, RedHatPlugin, UbuntuPlugin, DebianPlugin,
+                         SCLPlugin)
 from sos.utilities import find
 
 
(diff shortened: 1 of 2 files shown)


pmoravec:
There was an idea to have a child class of `RedHatPostgreSQL` that would bring even simpler patch. But that would be against the principle that `sosreport` executes *exactly one* class from each plug-in (cf. with https://github.com/sosreport/sos/blob/master/sos/policies/__init__.py#L239, breaking this would allow two plugins of "postgresql" name, quite confusing).

sos/plugins/postgresql.py
```diff
@@ -107,29 +113,29 @@ def postproc(self):
 
 class RedHatPostgreSQL(PostgreSQL, RedHatPlugin):
 
+    packages = ('postgresql', 'rh-postgresql95-postgresql-server')
+
     def setup(self):
+        if self.is_installed('postgresql'):
+            self.pghome = self.get_option("pghome")
+        else:  # SCL postgreSQL is installed
```
dougsland: The patch looks good to me. Just one question, isn't a good idea to check if  rh-postgresql95-postgresql-server is installed as you did in line 119?

pmoravec: This class can be enabled either by:

- presence of either `postgresql` or `rh-postgresql95-postgresql-server` package (plugin/class is enabled by default when a package from `packages` is present, or when a file from `files` is present) - see https://github.com/sosreport/sos/pull/1118/files#diff-93be9976582cadc41ed1cec376205196R116 and https://github.com/sosreport/sos/blob/master/sos/plugins/__init__.py#L99-L102 . In this case, and if `postgresql` isnt installed, we can deduce ` rh-postgresql95-postgresql-server` is installed.
- or this class can be enabled manually, e.g. via cmdline option `-o postgres`. Then potentially neither package can be present. In that case, we should rather take default postgres data as the default option. So we should rather revert the test (use `scl` commands only when `scl` installed, otherwise default to normal `postgres`).

The difference is imho small, leaving on @bmr-cymru opinion if it makes sense to improve it that way.

bmr-cymru: Eh, this does not look right to me.

Having separate `pghome` and `sclpghome` plugin options is a bit of an ugly hack, and not very nice for users to try to understand.

I think it would be better to revive the `SCLPluign` mixin concept from PR #900 than to attempt to solve this in an ad-hoc and plugin specific fashion.

sandrobonazzola:
So did this miss sos 3.5?

pmoravec:
It is included in 3.5 via commits 947e708 and 62d6435.

So I am closing this PR.

PR #1119, opened 2017-10-05
[clear_containers] Add new plugin for Clear Containers

jodh-intel:
Add a new plugin for the Intel(R) Clear Containers runtime
(`cc-runtime`).

See:

  - https://github.com/clearcontainers/runtime
  - https://github.com/clearcontainers

Signed-off-by: James O. D. Hunt <james.o.hunt@intel.com>

---

### DIFF
--- a/sos/plugins/clear_containers.py
+++ b/sos/plugins/clear_containers.py
@@ -1,0 +1,88 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+import re
+
+from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin, \
+        SuSEPlugin
+
+
+class ClearContainers(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin,
+                      SuSEPlugin):
+    """Intel(R) Clear Containers configuration
+    """
+
+    plugin_name = 'clear_containers'
+    profiles = ('system', 'virt')
+
+    runtime = 'cc-runtime'
+    packages = (runtime,)
+
+    def attach_cc_config_files(self):
+
+        # start with the default file locations
+        config_files = [
+                '/etc/clear-containers/configuration.toml'
+                '/usr/share/defaults/clear-containers/configuration.toml'
+        ]
+
+        # obtain a list of config files by asking the runtime
+        cmd = '{} --cc-show-default-config-paths'.format(self.runtime)
+        configs = self.get_command_output(cmd)['output']
+
+        for config in configs.splitlines():
+            if config != "":
+                config_files.append(config)
+
+        # get a unique list of config files
+        config_files = set(config_files)
+
+        self.add_copy_spec(config_files)
+
+    def attach_cc_log_files(self):
+
+        # start with the default global log
+        log_files = [
+            '/var/lib/clear-containers/runtime/runtime.log'
+        ]
+
+        # query the runtime to find the configured global log file
+        cmd = '{} cc-env'.format(self.runtime)
+        output = self.get_command_output(cmd)['output']
+        for line in output.splitlines():
+            result = re.search(r'\bGlobalLogPath\b\s+=\s+"(.+)"', line)
+            if result:
+                global_logfile = result.group(1)
+                if global_logfile:
+                    log_files.append(global_logfile)
+                break
+
+        # get a unique list of log files
+        log_files = set(log_files)
+
+        self.add_copy_spec(log_files, self.limit)
+
+    def setup(self):
+        self.limit = self.get_option("log_size")
+
+        if self.get_option("all_logs"):
+            # no limit on amount of data recorded
+            self.limit = None
+
+        self.add_cmd_output('{} cc-env'.format(self.runtime))
+        self.attach_cc_config_files()
+
+        self.attach_cc_log_files()
+        self.add_journal(units="cc-proxy")
+        self.add_journal(identifier="cc-shim")
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -741,7 +741,8 @@ class Plugin(object):
         self.custom_text += text
 
     def add_journal(self, units=None, boot=None, since=None, until=None,
-                    lines=None, allfields=False, output=None, timeout=None):
+                    lines=None, allfields=False, output=None, timeout=None,
+                    identifier=None):
         """ Collect journald logs from one of more units.
 
         Keyword arguments:
@@ -768,6 +769,7 @@ class Plugin(object):
         until_opt = " --until %s"
         lines_opt = " --lines %s"
         output_opt = " --output %s"
+        identifier_opt = " --identifier %s"
 
         if isinstance(units, six.string_types):
             units = [units]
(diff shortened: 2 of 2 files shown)


sos/plugins/cc-runtime.py
```diff
@@ -0,0 +1,30 @@
...
+from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin, \
+        SuSEPlugin
+
+
+class ClearContainers(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin,
+                      SuSEPlugin):
+    """Intel(R) Clear Containers configuration
+    """
+
+    plugin_name = 'clear-containers'
```
bmr-cymru: The plugin name and Python module name must match, and due to ambiguity with the subtraction operator the module name cannot contain the `-` character.

I'd suggest putting this either in a file named `cc_runtime.py` and setting `plugin_name` to `"cc_runtime", or naming the file `clear_containers.py` and setting `plugin_name` to `"clear_containers"`.

sos/plugins/cc-runtime.py
```diff
@@ -0,0 +1,30 @@
...
+    profiles = ('system', 'virt')
+
+    packages = ('cc-runtime',)
+
+    def setup(self):
+        self.add_cmd_output('cc-collect-data.sh')
```
bmr-cymru: Looking at this script in the runtime repository it duplicates functionality that's already present in other plugins (Kubernetes, Docker, KVM, package listings that are provided by the rpm and dpkg plugins). It's also capturing logs from components in a way that is not in keeping with what sos's users expect: they are not subject to the global `--log-size` and `--all-logs` switches and will not appear at the conventional locations in the generated report.

There's nothing very complex going on in this script and I think it would be better to implement this natively in Python as the other plugins in the tree do today. The one aspect that's potentially useful is the log grepping: this could be done in the plugin itself, or, if the script allows a method to disable all the other unnecessary collection we could carry on calling out to it for that.

jodh-intel: Thanks for the prompt feedback @bmr-cymru. I'll rework this PR as suggested...

bmr-cymru:
Np, thank you for submitting the PR! If the existing `cc-collect-data.sh` is already a canonical format that you use for collecting and analysing data from hosts then there's no objection to keeping that in if it is useful - it'd just be nice to also fill out some of the "usual suspects" for a typical sos plugin as well.

jodh-intel:
Hi @bmr-cymru - I've reworked the plugin as suggested.

sos/plugins/clear_containers.py
```diff
@@ -0,0 +1,80 @@
...
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+import re
+
+from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin, \
+        SuSEPlugin
+
+
+class ClearContainers(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin,
```
bmr-cymru: Ack to the new class & file names.

sos/plugins/clear_containers.py
```diff
@@ -0,0 +1,80 @@
...
+        # get a unique list of config files
+        config_files = set(config_files)
+
+        self.add_copy_spec(config_files)
+
+    def attach_cc_log_files(self):
+        log_files = []
+
+        # start with the default global log
+        log_files.append('/var/lib/clear-containers/runtime/runtime.log')
```
bmr-cymru: Similar here for the initial member but again nothing major.

sos/plugins/clear_containers.py
```diff
@@ -0,0 +1,80 @@
...
+    packages = (runtime,)
+
+    def attach_cc_config_files(self):
+        config_files = []
+
+        # start with the default file locations
+        config_files.append('/etc/clear-containers/configuration.toml')
+        config_files.append(
+                '/usr/share/defaults/clear-containers/configuration.toml'
+        )
```
bmr-cymru: It's not a big deal but it might look neater to just place these in the static list initialisation, e.g.:

```python
        config_files = [
            '/etc/clear-containers/configuration.toml',
            '/usr/share/defaults/clear-containers/configuration.toml'
        ]
```
This is a pretty commonly used style for plugin metadata init.

sos/plugins/clear_containers.py
```diff
@@ -0,0 +1,80 @@
...
+            if result:
+                global_logfile = result.group(1)
+                if global_logfile:
+                    log_files.append(global_logfile)
+                break
+
+        # get a unique list of log files
+        log_files = set(log_files)
+
+        self.add_copy_spec(log_files)
```
bmr-cymru: It would be good to support the global `--log-size` and `--all-logs` switches here. The usual pattern is something like:

```python
        all_logs = self.get_option("all_logs")
        limit = self.get_option("log_size")
        if not all_logs:
            # add plugin-defined minimal set of logs
            self.add_copy_spec("crucial.log", sizelimit=limit)
        else:
            # add "all possible" (plugin-defined) logs
            self.add_copy_spec("*.log", sizelimit=limit)
```

The example just uses a glob, but the behaviour is entirely down to the plugin - it might be that this gates whether you check for extra logs in the tool output, or it may perform other filtering that makes sense for the subsystem (it's also OK for it to have no effect if it really does not make sense to slice the logs this way).

Limits are set in whole numbers of mibibytes, or 0 to disable size limiting (files that partially exceed the limit are 'tailed' in `sos_strings/` and then symlinked to their original location in the tarball).

(we should add this to the plugin writer's guide, it's been around a while now)

sos/plugins/clear_containers.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def setup(self):
+        self.add_cmd_output('{} cc-env'.format(self.runtime))
+        self.attach_cc_config_files()
+        self.attach_cc_log_files()
+        self.add_journal(units="cc-proxy")
```
bmr-cymru: Ack.

sos/plugins/clear_containers.py
jodh-intel: Fixed.

sos/plugins/clear_containers.py
jodh-intel: Ditto ;)

sos/plugins/clear_containers.py
jodh-intel: Thanks - branch updated for these 2 useful options.

sos/plugins/clear_containers.py
jodh-intel: Just an idea, but it might be useful to add a `sizelimit=` param to `add_journal()` for parity with `add_copy_spec()`?

bmr-cymru: It's tricky unfortunately - it does offer other limiting options (since/until/lines), but as the `journalctl` command doesn't offer a direct byte-limiting mode it's something we have to handle in the process IO module.

@TurboTurtle is working on an improvement to how we run external commands that should improve this, and would mean we could plumb a similar size limit through to `add_journal()` - if/when that lands we'll do a mass update of the plugins that use it to pass the command line value through.

bmr-cymru:
Thanks for updating the PR again - on a quick scan I think it all looks good now. I'll have a thorough look in the morning & set the approve flag.

jodh-intel:
Just added an `add_journal()` call for `cc-shim` which wasn't present originally.

bmr-cymru:
👍

jodh-intel:
Hi @bmr-cymru - I've just realised that that last call to `add_journal()` isn't ever going to capture anything: it's benign, but ideally I'd be able to specify the following because our shim isn't a systemd unit:

```
add_journal(identifier="cc-shim")
```

bmr-cymru:
That shouldn't be a problem - just needs hooking up to `-t|--identifier SYSLOG_IDENTIFIER` presumably.

jodh-intel:
Hi @bmr-cymru - fixed.

bmr-cymru:
Thanks for adding that @jodh-intel - I was planning to take a look so appreciate your saving me the time!

The patch looks good - I may move the argument up nearer the front (to group it with the other selection args), but that's a minor tweak I can make as I merge.

jodh-intel:
Hi @bmr-cymru - do you think it's possible this can be added as part of the next release?

bmr-cymru:
That's the current plan - we're a bit overdue with 3.5 as there is some other high priority work under way on a different project. I'm hoping to start merging the final batches of PRs this week.

bmr-cymru:
Unfortunately this missed 3.5 due to other release pressure but it's on the list now for 3.6 (later this month).

Issue #1029, opened 2017-06-02
Limit size of systemd journal collection


jbainbri:
When the systemd journal is configured to a very large max size like 4 GiB, the `logs` plugin collects the entire journal twice, once in verbose mode. This can result in very large sosreports:

    $ du -s -BM sos_commands/logs/*
    1256M   sos_commands/logs/journalctl_--all_--this-boot_--no-pager
    4090M   sos_commands/logs/journalctl_--all_--this-boot_--no-pager_-o_verbose

Suggest apply a size limitation to this command, like the existing `messages` file is limited to 15 MiB.

TurboTurtle:
Ran into an issue with this today where journal sizes were outrageous. Looking into this, I don't believe there is any way within journalctl to only read X number of bytes. The only size-related functions are related to vacuuming the journal to a specific size, which we don't want to do here since that is data-destructive.

I think the only way sos could do this is via `bufsize` here within `sos_get_command_output` in `utilities.py` which is what is actually running any given journalctl command:

```
    p = Popen(expanded_args, shell=False, stdout=PIPE,
              stderr=STDOUT if stderr else PIPE,
              bufsize=-1, env=cmd_env, close_fds=True,
              preexec_fn=_child_prep_fn)
    stdout, stderr = p.communicate()
```

However, the `-1` is set to use the OS's default of buffered or unbuffered for the output, which is usually fully buffered. I *think* this could reasonably be changed to (optionally) reflect a given `--log-size`.

Thoughts? @bmr-cymru

bmr-cymru:
Afaik it is not so simple as that when using `Pipe.communicate()`, since regardless of buffer settings this will accumulate all the output into a Python `string` object in one go. To actually convert this to a *streaming* style of IO is more complex; we have to deal with the underlying process interface and implement two-way pipe IO (I never managed to make this work without deadlocking on input to the child, but that was a few years back).

TurboTurtle:
If it is a standard `str` object and not unicode, isn't byte length the same as character length here? So we could just crop the returned string to a fixed length before converting it to unicode?

E.G.
```
    return {
        'status': p.returncode,
        'output': stdout[-131072:].decode('utf-8', 'ignore')
    }
```

To return the last, for example, 128k (or whatever we want)? I did a quick test of this and it seems to work:

```
jake@cadia sos]$ sudo ./sosreport -o logs --batch

sosreport (version 3.2)

This command will collect system configuration and diagnostic
[...]

PR #1120, opened 2017-10-06
[utilities] Allow size limits for command output

TurboTurtle:
As mentioned by @bmr-cymru  in #1119 I have been working on a size limitation for command and journal output. I believe this is now in a good state for review, as it is functional and is exposed in the proper places (that I can think of at least).

This patch adds a size_limit option to `add_cmd_output()` and `add_journal()` which defaults to using `--log-size` (which itself currently defaults to 10MB).

This changes the way command execution is handled by handing the output channel to a new thread (only way I could avoid deadlocks) that reads the output line by line, and adds it to a deque. As the output is added, the deque's size will be checked against the given `size_limit` and if the deque is larger, the oldest (LRU) line in the deque will be removed. This technically means the size_limit is line sensitive, not byte sensitive, however I believe this is a close enough approximation. If `size_limit` is set to `0`, then output will not be limited (so, how it currently works) - using `None` produces the default behavior.

Running a few tests on some RHEL boxes, one very small and one decently sized, I have seen a massive reduction in the amount of memory used. The max memory reading below is from zsh's `time` which includes memory statistics:

```
RHEL 7.4, sos 3.4, 2GB memory system - journal on disk is 300mb - verbose journal failed to collect
sosreport --batch   223.05s  user 69.98s system 46% cpu 10:31.42 total
max memory:                1375 MB


RHEL 7.4, sos 3.4 with patch, same node as above, verbose journal was collected
sosreport --batch   260.02s  user 38.55s system 80% cpu 6:09.75 total
max memory:                641 MB



RHEL 7.3, sos 3.4, 48GB memory kube node with ~20 pods. journal on disk is 2.2G
sosreport --batch   371.95s  user 93.11s system 85% cpu 9:04.12 total
max memory:                13525 MB <==== !!!!



RHEL 7.3, sos 3.4 with patch, same node as above
sosreport --batch   393.94s  user 61.12s system 111% cpu 6:46.73 total
max memory:                449 MB

RHEL 7.3, same as above but with --log-size=100
sosreport --batch --log-size=100   392.64s  user 64.53s system 108% cpu 7:00.91 total
max memory:                885 MB
```

Note that while in the first test this patch does increase run time it is mostly because of the fact that sos hit a `Memory Error` while running the `logs` plugin. In the second test the run time still increases somewhat, but to a lesser extent.

I feel that the trade off of run time here is well worth the gains. As it stands, a low memory system will likely fail to collect a number of important items - taking an extra 20 seconds versus not getting output is preferable in my mind. Similarly on larger systems, the significantly reduced memory load is still a large benefit so that running sos is far less likely to affect other running processes. It also does not appear that the run time increase gets larger as the log-size gets larger - in other words this looks like a static increase in run time.

The other major benefit to this is that sosreports will no longer collect (by default at least) absurdly large journals and in addition the larger verbose journals as in #1029 

The one main point I am unsure of for this change is the default setting of `size_limit` for `add_journal()`. Right now it will also default to `--log-size` however as a support engineer I personally think 10MB is too small for a collected journal. I would venture that a 100MB default would be sufficient, but I am open to other suggestions.

Resolves: #1029 

---

### DIFF
--- a/sos/utilities.py
+++ b/sos/utilities.py
@@ -23,8 +23,10 @@ import fnmatch
 import errno
 import shlex
 import glob
+import threading
 
 from contextlib import closing
+from collections import deque
 
 # PYCOMPAT
 import six
@@ -111,7 +113,7 @@ def is_executable(command):
 
 def sos_get_command_output(command, timeout=300, stderr=False,
                            chroot=None, chdir=None, env=None,
-                           binary=False):
+                           binary=False, sizelimit=None):
     """Execute a command and return a dictionary of status and output,
     optionally changing root or current working directory before
     executing command.
@@ -153,7 +155,11 @@ def sos_get_command_output(command, timeout=300, stderr=False,
                   stderr=STDOUT if stderr else PIPE,
                   bufsize=-1, env=cmd_env, close_fds=True,
                   preexec_fn=_child_prep_fn)
-        stdout, stderr = p.communicate()
+
+        reader = AsyncReader(p.stdout, sizelimit, binary)
+        stdout = reader.get_contents()
+        p.poll()
+
     except OSError as e:
         if e.errno == errno.ENOENT:
             return {'status': 127, 'output': ""}
@@ -165,7 +171,7 @@ def sos_get_command_output(command, timeout=300, stderr=False,
 
     return {
         'status': p.returncode,
-        'output': stdout if binary else stdout.decode('utf-8', 'ignore')
+        'output': stdout
     }
 
 
@@ -191,6 +197,55 @@ def shell_out(cmd, timeout=30, chroot=None, runat=None):
     """
     return sos_get_command_output(cmd, timeout=timeout,
                                   chroot=chroot, chdir=runat)['output']
+
+
+class AsyncReader(threading.Thread):
+    '''Used to limit command output to a given size without deadlocking
+    sos.
+
+    Takes a sizelimit value in MB, and will compile stdout from Popen into a
+    string that is limited to the given sizelimit.
+    '''
+
+    def __init__(self, channel, sizelimit, binary):
+        super(AsyncReader, self).__init__()
+        self.chan = channel
+        self.binary = binary
+        self.chunksize = 2048
+        slots = None
+        if sizelimit:
+            sizelimit = sizelimit * 1048576  # convert to bytes
+            slots = sizelimit / self.chunksize
+        self.deque = deque(maxlen=slots)
+        self.start()
+        self.join()
+
+    def run(self):
+        '''Reads from the channel (pipe) that is the output pipe for a
+        called Popen. As we are reading from the pipe, the output is added
+        to a deque. After the size of the deque exceeds the sizelimit
+        earlier (older) entries are removed.
+
+        This means the returned output is chunksize-sensitive, but is not
+        really byte-sensitive.
+        '''
+        try:
+            while True:
+                line = self.chan.read(self.chunksize)
+                if not line:
+                    # Pipe can remain open after output has completed
+                    break
+                self.deque.append(line)
+        except (ValueError, IOError):
+            # pipe has closed, meaning command output is done
+            pass
+
+    def get_contents(self):
+        '''Returns the contents of the deque as a string'''
+        if not self.binary:
+            return ''.join(ln.decode('utf-8', 'ignore') for ln in self.deque)
+        else:
+            return b''.join(ln for ln in self.deque)
 
 
 class ImporterHelper(object):
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -606,7 +606,7 @@ class Plugin(object):
 
     def get_command_output(self, prog, timeout=300, stderr=True,
                            chroot=True, runat=None, env=None,
-                           binary=False):
+                           binary=False, sizelimit=None):
         if chroot or self.commons['cmdlineopts'].chroot == 'always':
             root = self.sysroot
         else:
@@ -614,7 +614,8 @@ class Plugin(object):
 
         result = sos_get_command_output(prog, timeout=timeout, stderr=stderr,
                                         chroot=root, chdir=runat,
-                                        env=env, binary=binary)
+                                        env=env, binary=binary,
+                                        sizelimit=sizelimit)
 
         if result['status'] == 124:
             self._log_warn("command '%s' timed out after %ds"
@@ -652,14 +653,16 @@ class Plugin(object):
 
     def _add_cmd_output(self, cmd, suggest_filename=None,
                         root_symlink=None, timeout=300, stderr=True,
-                        chroot=True, runat=None, env=None, binary=False):
+                        chroot=True, runat=None, env=None, binary=False,
+                        sizelimit=None):
         """Internal helper to add a single command to the collection list."""
         cmdt = (
             cmd, suggest_filename,
             root_symlink, timeout, stderr,
-            chroot, runat, env, binary
+            chroot, runat, env, binary, sizelimit
         )
-        _tuplefmt = "('%s', '%s', '%s', %s, '%s', '%s', '%s', '%s', '%s')"
+        _tuplefmt = ("('%s', '%s', '%s', %s, '%s', '%s', '%s', '%s', '%s', "
+                     "'%s')")
         _logstr = "packed command tuple: " + _tuplefmt
         self._log_debug(_logstr % cmdt)
         self.collect_cmds.append(cmdt)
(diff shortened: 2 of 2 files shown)


pmoravec:
> This technically means the size_limit is line sensitive, not byte sensitive, however I believe this is a close enough approximation. 

Definitely no problem, a line is usually few tens of bytes only so the granularity is very similar. For a comparison, collecting multiple files with `size_limit`, the behaviour is file-sensitive, even :)

> I feel that the trade off of run time here is well worth the gains.

Additional 20s / 5% is still acceptable trade-off. We might try to improve it (or in general, I planned to review performance of sos that would help orthogonally, sadly no time).

> The one main point I am unsure of for this change is the default setting of size_limit for add_journal().

Agreed to have it here 100MB.

Travis build failed as 2 automated tests failed:

> test_output (utilities_tests.ExecutableTest) ... FAIL
> test_output_chdir (utilities_tests.ExecutableTest) ... FAIL

Will review the patch and also the build failures soon..

pmoravec:
Code changes seem OK to me.

I run on 4 different systems the current upstream and also the branch under this PR, and I see practically no difference in time execution and no regression in collected data.

sos/utilities.py
```diff
@@ -152,7 +155,10 @@ def _child_prep_fn():
                   stderr=STDOUT if stderr else PIPE,
                   bufsize=-1, env=cmd_env, close_fds=True,
                   preexec_fn=_child_prep_fn)
-        stdout, stderr = p.communicate()
+
+        reader = AsyncReader(p.stdout, size_limit)
+        stdout = reader.get_contents()
+
```
pmoravec: I suggest adding here `p.poll()`. At least this will properly set `returncode` used later on (and also this is the reason why 2 automated tests fail).

Anyway, I am not an expert on python (sub)processes.

bmr-cymru:
Thanks @TurboTurtle. Just to set expectations, I am not sure that this can make 3.5 - obviously a change of this scale needs some pretty close attention at review, as well as ideally a very broad set of testing data. I think there may also be some API considerations to take account of from what I've looked at so far (nothing huge, mostly just about consistency and making changes in the right order so as not to leave "funny" differences between otherwise like method calls).

I am still really busy with work for `boom`, but I will try to get at least a basic pass on this done this week.

sos/utilities.py
TurboTurtle: Yep - you're right. I've added `p.poll()`

TurboTurtle:
@pmoravec Thanks for taking a look. Made a slight change since I found it would fail with python3. I just moved the `decode()` from `sos_get_command_output()` to `AsyncReader.get_contents()`. As it was, with python3 we'd get a `TypeError` in the join (since it would be bytes at that point, not str).

@bmr-cymru Absolutely. I just figured it'd be best to get this up now to get the ball rolling on it since it does end up touching a whole heck of a lot.

k-hagio:
I'm interested in this PR, too, because I have encountered several cases that this sosreport behavior caused temporary system stall (delay) due to lack of memory on customer's system. All of them also ware caused by verbose journal, but regardless of its existence, I think this kind of safety mechanism is necessary especially for systems on production phase.

I also agree to the way of the patch on the whole but how about changing `readline()` to `read(4096)`? With this, CPU time can be reduced significantly on a test server. Here is the test result:

```
RHEL7.4, TurboTurtle:size-limit branch, 32GB memory, and verbose journal is about 300MB.

# journalctl --no-pager --all --boot --output verbose | wc
7775339 11461737 296754193

*** readline()
# /usr/bin/time ./sosreport --batch >/dev/null
55.93user 13.26system 1:13.24elapsed 94%CPU (0avgtext+0avgdata 224660maxresident)k
22092inputs+588984outputs (3major+2215137minor)pagefaults 0swaps

(top -bcd 1)
  PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND
31047 root      20   0  490900 108540   7548 S 100.0  0.3   0:15.98 /usr/bin/python ./sosreport --batch
 1326 root      20   0  483716 182000 181752 R  68.3  0.6   0:07.33 journalctl --no-pager --all --boot --output verbose
--> the bottleneck is sosreport.

*** read(4096)
# /usr/bin/time ./sosreport --batch >/dev/null
35.51user 13.24system 1:03.91elapsed 76%CPU (0avgtext+0avgdata 224664maxresident)k
^^^^^ 20s faster      ^^^^^^^ 10s faster
22092inputs+590176outputs (3major+2139330minor)pagefaults 0swaps

  PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND
 4389 root      20   0  247476  76468  76228 R 100.0  0.2   0:02.54 journalctl --no-pager --all --boot --output verbose
 1727 root      20   0  443476  44728   7552 S  13.9  0.1   0:04.07 /usr/bin/python ./sosreport --batch
--> sosreport is much faster than journalctl.

*** read(32768)
# /usr/bin/time ./sosreport --batch >/dev/null
34.88user 13.33system 1:02.80elapsed 76%CPU (0avgtext+0avgdata 224660maxresident)k
22092inputs+581112outputs (3major+2132073minor)pagefaults 0swaps

  PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND
 4824 root      20   0  328068 115012 114772 R 100.0  0.4   0:03.64 journalctl --no-pager --all --boot --output verbose
 2168 root      20   0  443476  44768   7548 S   9.9  0.1   0:04.05 /usr/bin/python ./sosreport --batch
```

TurboTurtle:
@k-hagio that pretty much removes the time penalty of this patch versus sos today, so that's great! I'll play around with this a bit myself as well but looking at it so far it looks like something we'd want to do here. I'll update the commit in a bit.

TurboTurtle:
I've updated the PR to use `read(4096)` as I've not found any issues with it on my end and it does pretty much remove the run time penalty this would've added otherwise.

TurboTurtle:
@bmr-cymru @pmoravec
 
I've updated the PR again to be a much simpler implementation. Following k-hagio's suggestion on `read(4096)` I thought why don't we just define a line size and then let the `deque` do it's normal stuff so we don't have to manually monitor and control the size. So, for now I've defined a line size as 2048 bytes, and use that to determine the length of the deque based on the given `sizelimit`. In my tests this still alleviates the time penalty that `read(4096)` solved, but also allows for us to lose less information from the first entry when we go over the sizelimit (and I think the chances of losing critical data in the first 2048 bytes of a 100MB+ log is negligible). Why I didn't think of this sooner is beyond me.

Also I updated `size_limit` to be `sizelimit` so that it matches the other usage of that, and I set the journal sizelimit to default to the larger of either `--log-size` or 100MB.

pmoravec:
Testing on 2 RHEL7 systems:

original sosreport on system A took 8m0.847s with peak 4581312 kB RSS.
patched sosreport on system A took 6m32.378s with peak 888388 kB RSS.

5times better memory, even faster run.

original sosreport on system B took 6m56.252s with peak 788248 kB RSS.
patched sosreport on system B took 3m47.187s with peak 485788 kB RSS.

considerable smaller memory and faster run.

+1

sos/utilities.py
```diff
@@ -192,6 +199,51 @@ def shell_out(cmd, timeout=30, chroot=None, runat=None):
...
+    sos.
+
+    Takes a sizelimit value in MB, and will compile stdout from Popen into a
+    string that is limited to the given sizelimit.
+    '''
+
+    def __init__(self, channel, sizelimit):
+        super(AsyncReader, self).__init__()
+        self.chan = channel
+        self.chunksize = 2048
```
pmoravec: Isn't using 4k / `4096` for `chunksize` more appropriate to match default kernel page size? Can't it improve performance?

pmoravec: It might be more appropriate to match kernel page size, but I havent seen any performance gain. Any value betwwen 1k - 16k (havent tried bigger) show similar performance.

bmr-cymru: It makes no real difference unless you are doing direct IO (and then, probably only on partial page writes where the next write hits the "other" part).

The kernel faults this stuff in in page sized units, so whether you ask for 1k, or 4k, the same data is going to be physically read in (and readahead provides another layer of insulation between physical device performance).

bmr-cymru: (aside from syscall and context switch overheard, that is, but I don't see that being a large effect here)

bmr-cymru: In some ways, a true ring buffer would be the ideal solution here: I'll look into how easy this is to do in Python (it's trivial in C by pointer shuffling and has the benefit that expiring entries is effectively zero-work: you just move the end pointer back, and free up some more space in the buffer).

TurboTurtle:
I've rebased the PR to accommodate the previous addition of the `binary` option. Tested to make sure the binary data wasn't lost/corrupted by running the `postgresql` plugin to get a db dump from a RHV manager. Dump restored without issue and RHV did not lose any information.

PR #1121, opened 2017-10-10
[rear] Add new plugin

TurboTurtle:
This adds a plugin for Relax and Recover, or rear, which is a backup and
recovery tool.

---

### DIFF
--- a/sos/plugins/rear.py
+++ b/sos/plugins/rear.py
@@ -1,0 +1,54 @@
+# Copyright (C) 2017 Red Hat, Inc. Jake Hunsaker <jhunsake@redhat.com>
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


sos/plugins/rear.py
```diff
@@ -0,0 +1,50 @@
...
+    packages = ('rear',)
+
+    def setup(self):
+        # don't collect recovery ISOs or tar archives
+        self.add_forbidden_path('/var/log/rear/*.iso')
+        self.add_forbidden_path('/var/log/rear/*.tar.gz')
+
+        self.add_copy_spec([
+            '/etc/rear',
+            '/var/log/rear'
```
bmr-cymru: Need for size limits here?

TurboTurtle:
Added a sizelimit, and also constrained the `add_copy_spec` a bit.

bmr-cymru:
:+1: thanks!

PR #1125, opened 2017-10-13
[logs] Add gathering of systemd journal from previous boot

rmetrich:
As a support engineer, having the journal from previous boot helps debugging issues happening during shutdown.

---

### DIFF
--- a/sos/plugins/logs.py
+++ b/sos/plugins/logs.py
@@ -34,6 +34,8 @@ class Logs(Plugin):
         self.add_copy_spec("/var/log/cloud-init*", sizelimit=self.limit)
         self.add_journal(boot="this")
         self.add_journal(boot="this", allfields=True, output="verbose")
+        self.add_journal(boot="last")
+        self.add_journal(boot="last", allfields=True, output="verbose")
         self.add_cmd_output("journalctl --disk-usage")
 
         if self.get_option('all_logs'):


bmr-cymru:
The `add_journalctl` call should use `"last"` to select the last boot - it's a documented value for the `boot` parameter and it makes the plugin code easier to read.

Either way, I'm not convinced that this should be on by default (rather than e.g. enabled by `--all-logs`). Logs for the current boot are self explanatory but `journalctl` collection is potentially expensive - on my fairly lightly logging system it doubles the time spent collecting journals for this plugin.

I'd like to get some more opinions from the other distro maintainers before we go ahead with this.

rmetrich:
OK for "last".
Everytime we have a customer which has issues with shutdown (it's very common) we have to request last boot, again and again.
However, verbose mode is likely unused and may be safely removed.

bmr-cymru:
> Everytime we have a customer which has issues with shutdown (it's very common) we have to request last 
> boot, again and again.

Which is exactly the kind of information to include in a PR or commit message, to explain the change you are making... All collection has a cost and needs to be justified - especially expensive things like `journalctl`.

The verbose output flag makes logs more suitable for machine reading, but less so for human users (although this can be mostly reversed by script). We added the verbose option at (iirc) Lennart or Harald's suggestion. I'd be happy to drop it, but I'd prefer to have consistency for boot time log collection (i.e. one or the other for both calls).

Issue #405, opened 2014-09-18
[navicli] remove interactive prompting


bmr-cymru:
```
[plugin:navicli] added copyspec '/etc/Navisphere/agent.config'
[plugin:navicli] added copyspec '/etc/Navisphere/Navimon.cfg'
[plugin:navicli] added copyspec '/etc/Navisphere/Quietmode.cfg'
[plugin:navicli] added copyspec '/etc/Navisphere/messages/[a-z]*'
[plugin:navicli] added copyspec '/etc/Navisphere/log/[a-z]*'
CLARiiON SP IP Address or [Enter] to exit: 
```

Plugins shouldn't be open-coding prompts to the user. There should either be a Plugin API for raising a prompt (possibly combined with the options interface - e.g. an unset option could prompt the user if `prompt=True` is passed) or no interactive prompts after the initial start up.

HarryRybacki:
I believe this can be avoided by using the [``--batch`` flag](https://github.com/sosreport/sos/blob/master/sos/sosreport.py#L1185-L1195).

bmr-cymru:
> I believe this can be avoided by using the --batch flag.

Have you actually tried this? Or even read the code?

Where are you seeing anything testing for the state of `--batch` in this plugin?

(hint: there isn't currently any way for plugins to know this, without digging through SoSOptions directly).

bmr-cymru:
```
[root@localhost ~]# sosreport --batch -e navicli -o navicli

sosreport (version 3.2.0a1)

This command will collect system configuration and diagnostic
information from this Fedora system. An archive containing the collected
information will be generated in /var/tmp/sos.a_v0gjl2.

For more information on the Fedora Project visit:

  https://fedoraproject.org/

The generated archive may contain data considered sensitive and its
content should be reviewed by the originating organization before being
passed to any third party.

No changes will be made to system configuration.


 Setting up archive ...
 Setting up plugins ...
CLARiiON SP IP Address or [Enter] to exit: 
```

See what I mean?

HarryRybacki:
I've tested it locally to collect sosreports [with ansible](https://review.gerrithub.io/#/c/298299/1/tasks/collect.yml). I guess I don't understand exactly what you're looking for -- a *non-interactive* mode or a *silent* mode?

bmr-cymru:
I think you've misunderstood the problem; this issue documents a bug in the `navicli` plugin, in that it attempts to prompt the user for input, regardless of the setting of the `--batch` command line switch.

[...]

PR #1126, opened 2017-10-13
[navicli] remove interactive prompting #405

rmetrich:
This resolves Issue #405 

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -148,6 +148,7 @@ class Plugin(object):
         self.copy_strings = []
         self.collect_cmds = []
         self.sysroot = commons['sysroot']
+        self.batch = commons['cmdlineopts'].batch
 
         self.soslog = self.commons['soslog'] if 'soslog' in self.commons \
             else logging.getLogger('sos')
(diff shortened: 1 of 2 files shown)


bmr-cymru:
Thanks for working on this, but I would prefer if we either:

  * Delete the interactive prompting entirely: `sos` has never offered a prompting mechanism and the navicli
    plugin is a hack (a very old hack, which is why it has persisted this long).

  * Introduce a proper mechanism for prompting when batch mode is False: e.g. extending the `get_option()` 
    interface to permit un-set options to be prompted for.

Also, when referencing PR or issue numbers in commit messages, please put them in the commit body using one of the keywords recognised by GitHub (issue/resolves/closes etc.):

  https://help.github.com/articles/closing-issues-using-keywords/

This means that auto-close and embedded links work properly (like the current 'issue' link does).

bmr-cymru:
I think we should move consideration for this to the next major release: the major question is whether we want to support interactive prompts as a means of supplying missing option values (when not in `--batch` mode).

Historically outside of the global prompts for user and case information we've avoided this, and the only plugin that has tried to do this so far has been the NaviCLI one, with the known limitations/problems that causes.

There's potentially a few uses for this, including gathering authentication info that would be of use to multiple plugins but it's a fair amount of work to implement that and then modify all the plugins that use this type of data to make use of the facility.

rmetrich:
Let's close this for now.

PR #1129, opened 2017-10-26
[openstack_ironic] update to cover recent features

dtantsur:
* ironic-discoverd was renamed to ironic-inspector in Liberty release
* collect port groups list, driver list and detailed port list
* collect introspection statuses list
* optionally collect introspection data from all nodes

Signed-off-by: Dmitry Tantsur <divius.inside@gmail.com>

---

### DIFF
--- a/sos/plugins/openstack_ironic.py
+++ b/sos/plugins/openstack_ironic.py
@@ -77,7 +77,9 @@ class OpenStackIronic(Plugin):
         vars_any = [p in os.environ for p in [
                     'OS_TENANT_NAME', 'OS_PROJECT_NAME']]
 
-        if not (all(vars_all) and any(vars_any)):
+        self.osc_available = all(vars_all) and any(vars_any)
+
+        if not self.osc_available:
             self.soslog.warning("Not all environment variables set. Source "
                                 "the environment file for the user intended "
                                 "to connect to the OpenStack environment.")
@@ -136,5 +138,25 @@ class RedHatIronic(OpenStackIronic, RedHatPlugin):
 
             self.add_journal(units="openstack-ironic-discoverd")
             self.add_journal(units="openstack-ironic-discoverd-dnsmasq")
+
+        # ironic-discoverd was renamed to ironic-inspector in Liberty
+        self.conf_list.append('/etc/ironic-inspector/*')
+        self.conf_list.append(self.var_puppet_gen + '/etc/ironic-inspector/*')
+        self.add_copy_spec('/etc/ironic-inspector/')
+        self.add_copy_spec(self.var_puppet_gen + '/etc/ironic-inspector/')
+        self.add_copy_spec('/var/lib/ironic-inspector/')
+        if self.get_option("all_logs"):
+            self.add_copy_spec('/var/log/ironic-inspector/')
+            self.add_copy_spec('/var/log/containers/ironic-inspector/')
+        else:
+            self.add_copy_spec('/var/log/ironic-inspector/*.log')
+            self.add_copy_spec('/var/log/ironic-inspector/ramdisk/')
+            self.add_copy_spec('/var/log/containers/ironic-inspector/*.log')
+            self.add_copy_spec('/var/log/containers/ironic-inspector/ramdisk/')
+
+        self.add_journal(units="openstack-ironic-inspector-dnsmasq")
+
+        if self.osc_available:
+            self.add_cmd_output("openstack baremetal introspection list")
 
 # vim: set et ts=4 sw=4 :
--- a/sos/plugins/openstack_ironic.py
+++ b/sos/plugins/openstack_ironic.py
@@ -82,8 +82,10 @@ class OpenStackIronic(Plugin):
                                 "the environment file for the user intended "
                                 "to connect to the OpenStack environment.")
         else:
+            self.add_cmd_output("openstack baremetal driver list --long")
             self.add_cmd_output("openstack baremetal node list --long")
-            self.add_cmd_output("openstack baremetal port list")
+            self.add_cmd_output("openstack baremetal port list --long")
+            self.add_cmd_output("openstack baremetal port group list --long")
 
     def postproc(self):
         protect_keys = [
--- a/sos/plugins/openstack_ironic.py
+++ b/sos/plugins/openstack_ironic.py
@@ -63,10 +63,9 @@ class OpenStackIronic(Plugin):
                 "/var/log/containers/httpd/ironic-api/*log"
             ], sizelimit=self.limit)
 
-        self.add_cmd_output('ls -laRt /var/lib/ironic/')
-        self.add_cmd_output(
-            'ls -laRt ' + self.var_puppet_gen + '/var/lib/ironic/'
-        )
+        for path in ['/var/lib/ironic', '/httpboot', '/tftpboot']:
+            self.add_cmd_output('ls -laRt %s' % path)
+            self.add_cmd_output('ls -laRt %s' % (self.var_puppet_gen + path))
 
         if self.get_option("verify"):
             self.add_cmd_output("rpm -V %s" % ' '.join(self.packages))
--- a/sos/plugins/openstack_ironic.py
+++ b/sos/plugins/openstack_ironic.py
@@ -125,6 +125,20 @@ class RedHatIronic(OpenStackIronic, RedHatPlugin):
         'openstack-ironic-discoverd-ramdisk'
     ]
 
+    def collect_introspection_data(self):
+        uuids_result = self.call_ext_prog('openstack baremetal node list '
+                                          '-f value -c UUID')
+        if uuids_result['status']:
+            self.soslog.warning('Failed to fetch list of ironic node UUIDs, '
+                                'introspection data won\'t be collected')
+            return
+
+        uuids = [uuid for uuid in uuids_result['output'].split()
+                 if uuid.strip()]
+        for uuid in uuids:
+            self.add_cmd_output('openstack baremetal introspection '
+                                'data save %s' % uuid)
+
     def setup(self):
         super(RedHatIronic, self).setup()
 
@@ -157,5 +171,7 @@ class RedHatIronic(OpenStackIronic, RedHatPlugin):
 
         if self.osc_available:
             self.add_cmd_output("openstack baremetal introspection list")
+            if self.get_option("all_logs"):
+                self.collect_introspection_data()
 
 # vim: set et ts=4 sw=4 :


sos/plugins/openstack_ironic.py
```diff
@@ -50,13 +50,16 @@ def setup(self):
         vars_any = [p in os.environ for p in [
                     'OS_TENANT_NAME', 'OS_PROJECT_NAME']]
 
-        if not (all(vars_all) and any(vars_any)):
+        self.osc_available = all(vars_all) and any(vars_any)
+
+        if not self.osc_available:
```
bmr-cymru: Commits aren't expensive - I know it's tempting to cram internal refactoring in with other changes, but including this all in one commit makes your commit message a bit misleading. There's no hint of these changes in the four bullets that are mentioned.

sos/plugins/openstack_ironic.py
```diff
@@ -95,6 +98,15 @@ class RedHatIronic(OpenStackIronic, RedHatPlugin):
         'openstack-ironic-discoverd-ramdisk'
     ]
 
+    inspector_package = 'openstack-ironic-inspector'
+
+    def add_introspection_data(self):
+        shell = (
+            "for uuid in $(openstack baremetal node list -f value -c UUID); "
+            "do openstack baremetal introspection data save $uuid; done"
+        )
```
bmr-cymru: NACK: using shell syntax in `sos` plugins is absolutely forbidden (there's one historical example in the current tree - in the `rpm` plugin but new uses are completely banned).

This is just a bit lazy anyway - there is no reason at all that this cannot be implemented properly in Python, as other plugins (e.g. `networking`) do.

If for some reason this is some kind of "preferred form" of this data from upstream then a script should be provided that users or `sos` can run.

sos/plugins/openstack_ironic.py
```diff
@@ -108,4 +120,23 @@ def setup(self):
             self.add_journal(units="openstack-ironic-discoverd")
             self.add_journal(units="openstack-ironic-discoverd-dnsmasq")
 
+        # ironic-discoverd was renamed to ironic-inspector in Liberty
+        if self.is_installed(self.inspector_package):
```
bmr-cymru: We don't put things behind `is_installed()` unless there are truly harmful effects to not doing it: for one thing, it causes the `all_logs`/`log_size` handling to propagate through the branches and it makes the code ugly and harder to read.

What's the downside here of just running these calls? I can see one journal command, that will collect an error (unit not found), but that's pretty harmless.

sos/plugins/openstack_ironic.py
dtantsur: Will do

sos/plugins/openstack_ironic.py
dtantsur: Ok, I just wasn't sure if it's fine to run commands right from setup()

sos/plugins/openstack_ironic.py
dtantsur: No downside at all, I was just following the same pattern as above with discoverd. Will remove.

bmr-cymru: OK, thanks - I'll take a look through the OpenStack plugins to see if there are more examples of this pattern - we haven't always been as strict with review as we are now.

sos/plugins/openstack_ironic.py
bmr-cymru: Take a look at the `networking` plugin - it has a bunch of examples of running commands during `setup()` and then parsing the results. There are different interfaces to use depending on whether you want to also capture the data in the report, or if you just need it in-memory to figure out what other commands to run:

The `brctl` stuff uses `Plugin.get_cmd_output_now()`, which stores the result in the report and provides a path name string on return (absolute path into the `FileCacheArchive`):

https://github.com/sosreport/sos/blob/156e39ddf0126d1dee75087f3bfbd898db8f4692/sos/plugins/networking.py#L296

You can `open()` on it and do whatever you need to.

Alternately, the code to get ethernet interfaces does not store the result - it's held in memory only and thrown away. This uses the `call_ext_prog()` method:

https://github.com/sosreport/sos/blob/156e39ddf0126d1dee75087f3bfbd898db8f4692/sos/plugins/networking.py#L86

This family of calls return a Python dictionary object (conventionally stored in a variable named `something_result`). The keys are:

  * `status` numeric command status returned by wait/waitpid
  * `output` standard output, or combined stdout/stderr if `stderr=True`

dtantsur:
Thanks for your help! I've rebuilt the whole patch chain, please take another look.

bmr-cymru:
Thanks, will do!

sos/plugins/openstack_ironic.py
dtantsur: This comment was fixed, but I don't know how to mark is as such..

dtantsur:
Rebased to fix merge conflicts.

dtantsur:
@bmr-cymru sorry for pinging you, but could you please move this forward? I'm a bit worried that we are releasing OpenStack with really outdated sos information. Thanks!

sos/plugins/openstack_ironic.py
```diff
@@ -122,8 +125,18 @@ class RedHatIronic(OpenStackIronic, RedHatPlugin):
         'openstack-ironic-discoverd-ramdisk'
     ]
 
+    def collect_introspection_data(self):
+        uuids_result = self.call_ext_prog('openstack baremetal node list '
```
pmoravec: Output of `openstack baremetal node list` is not required to collect, am I right? (it wont be when you use `call_ext_prog`)

sos/plugins/openstack_ironic.py
```diff
@@ -122,8 +125,18 @@ class RedHatIronic(OpenStackIronic, RedHatPlugin):
...
+                                          '-f value -c UUID')
+        uuids = [uuid for uuid in uuids_result['output'].split()
```
pmoravec: What if the `openstack baremetal node list` command terminates with an error? Isnt it worth checking `if uuids_result["status"] == 0` first?

sos/plugins/openstack_ironic.py
```diff
@@ -135,4 +148,24 @@ def setup(self):
             self.add_journal(units="openstack-ironic-discoverd")
             self.add_journal(units="openstack-ironic-discoverd-dnsmasq")
 
-# vim: set et ts=4 sw=4 :
+        # ironic-discoverd was renamed to ironic-inspector in Liberty
+        self.conf_list.append('/etc/ironic-inspector/*')
+        self.conf_list.append(self.var_puppet_gen + '/etc/ironic-inspector/*')
+        self.add_copy_spec('/etc/ironic-inspector/')
+        self.add_copy_spec(self.var_puppet_gen + '/etc/ironic-inspector/')
+        self.add_copy_spec('/var/lib/ironic-inspector/')
```
pmoravec: Can't be here some big amount of data in some (unexpected) use case that sosreport will all copy and create ridiculously huge tarball?

sos/plugins/openstack_ironic.py
dtantsur: It's collected in the parent class, so it does not make sense to collect it here again.

sos/plugins/openstack_ironic.py
dtantsur: As this command is called in the parent class, I'm not sure how sosreport will behave.. Anyway, this is the most basic ironic command, if it fails, it's fine to fail the whole collector.

sos/plugins/openstack_ironic.py
dtantsur: I'm not sure we even store anything there right now :) In the near future, there will be some amount of really small (several bytes) files, but that's it.

dtantsur:
Thanks for review @pmoravec I hope my answers clarify it.

sos/plugins/openstack_ironic.py
```diff
@@ -135,4 +148,24 @@ def setup(self):
             self.add_journal(units="openstack-ironic-discoverd")
             self.add_journal(units="openstack-ironic-discoverd-dnsmasq")
 
-# vim: set et ts=4 sw=4 :
+        # ironic-discoverd was renamed to ironic-inspector in Liberty
```
bmr-cymru: Please don't delete the VIM mode line from this file.

sos/plugins/openstack_ironic.py
bmr-cymru: That does not excuse you from having to perform normal validation: exceptions in plugin classes (which is the likely result of continuing to try to operate without checking the command status...) are never acceptable.

Also, if this is collected already in the base class, why are we running it _again_ here? Why wasn't the superclass changed to store the result somewhere for further processing?

sos/plugins/openstack_ironic.py
dtantsur: Weird, I'm pretty sure I did not delete it Oo

sos/plugins/openstack_ironic.py
dtantsur: > Why wasn't the superclass changed to store the result somewhere for further processing?

Probably because I'm not a sosreport guru :) I think I get it, lemme try.

dtantsur: Aha, now I remember. Because we collect nodes in the base class as a table with all fields, and here we only collect UUIDs. I don't think we should parse tables here, so I guess I'll go with "normal validation" option then.

sos/plugins/openstack_ironic.py
bmr-cymru: It is deleted by this hunk, and never replaced...

sos/plugins/openstack_ironic.py
pmoravec: Hm.. can't we merge the two calls into one, then? (definitely there can be some trade-off in case that command execution is fast every time).

dtantsur: This is how the two outputs look like: http://paste.openstack.org/show/627601/. Actually, the former is MUCH longer, but I did not use --long here. To get the latter we need to parse this table in Python, which is probably on overkill.

The command itself is one SELECT from the database.

pmoravec: That sounds acceptable to me to have one extra call, then 9if we would have to parse the output from --long output otherwise). Thanks for explanation.

pmoravec:
Final code changes seems OK to me, though it makes sense to squash:

- ace40853cc8d35f9c599dda8ac36d71231e9cc76 and a4dfdb4892086349529c0cfb8ee166a10b1c1fa6 (dont delete the `# vim: set ..` and append back)
- c4b621d18de335ba6c5334adfd0ea819e280fd92 and 4eabf48b63455d53fcf2d570beb713fe32e5d137 (let have the test of return value from beginning, not so important though)

dtantsur:
@pmoravec @bmr-cymru streamlines the patch chain, thanks.

bmr-cymru:
Most of the patches are looking good now - thanks - I'll try to make time for a thorough review in the next day or two.

The only comment so far is the unnecessary RPM command in the PXE commit.

dtantsur:
@bmr-cymru @pmoravec thanks for review, removed the redundant command.

dtantsur:
Hi folks! Any news about this PR? @pmoravec is it waiting for your review?

pmoravec:
No complains from my side. Just pending on Bryn to merge it in some next bunch of merges/commits.

bogdando:
@bmr-cymru PTAL

sos/plugins/openstack_ironic.py
```diff
@@ -63,10 +63,9 @@ def setup(self):
                 "/var/log/containers/httpd/ironic-api/*log"
             ], sizelimit=self.limit)
 
-        self.add_cmd_output('ls -laRt /var/lib/ironic/')
-        self.add_cmd_output(
-            'ls -laRt ' + self.var_puppet_gen + '/var/lib/ironic/'
-        )
+        for path in ['/var/lib/ironic', '/httpboot', '/tftpboot']:
```
bogdando: please also add the expected paths for containerized undercloud. For me, that is

```
# ls /var/lib/config-data/puppet-generated/ironic_inspector/var/lib/ironic/httpboot/
inspector.ipxe
```

dtantsur:
@bogdando I kind of understand your desire to push your personal favourite yet-to-be feature as far forward as you can. But this is a fix for a long-troubling situation, that affects all versions of OpenStack starting with Liberty (!!!). Can we please finally move on with this and then consider the containerized undercloud (preferably when it finally works and stop changing every week).

sos/plugins/openstack_ironic.py
bogdando: ... or should it be ``/var/lib/config-data/puppet-generated/ironic/var/lib/ironic/httpboot/`` ?

bogdando:
@dtantsur I do not insist, only proposed to not drop the code that seems used to be working for containerized undercloud : ``self.var_puppet_gen + '/var/lib/ironic/'``

dtantsur:
@bogdando blocking a pull request fits into my definition of insisting ;) Anyway, I don't think I'm removing the code - look at line 68. I'm even extending it, I've just rolled it into the loop.

pmoravec:
(closed via 6a23a57596e94ea3e4f4f19513c8b691af9c53ff and becab7d603b86975c8712d73c738bbf621740b01 and ff2b26dfe74e67ff969796b4f60bd5291a535277 and 9d9af248d85b576abad1f5fdb9112bafdbbc913a)

PR #1130, opened 2017-10-27
[openstack plugins+logrotate] Tripleo additional directories - patch2

stuggi:
Tripleo Pike opinionated config+log paths to be collected
for services, when running in containers.

This is an addition to #1054 where a couple of things got missed

---

### DIFF
--- a/sos/plugins/openstack_aodh.py
+++ b/sos/plugins/openstack_aodh.py
@@ -1,4 +1,5 @@
 # Copyright (C) 2017 Red Hat, Inc., Sachin Patil <psachin@redhat.com>
+# Copyright (C) 2017 Red Hat, Inc., Martin Schuppert <mschuppert@redhat.com>
 
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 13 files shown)


sos/plugins/openstack_aodh.py
```diff
@@ -72,8 +88,21 @@ def postproc(self):
 
         self.do_file_sub(
             "/etc/aodh/aodh.conf",
-            r"(auth_url[\t\ ]*=[\t\ ]*)(.+)",
+            r"(rabbit_password[\t\ ]*=[\t\ ]*)(.+)",
```
pmoravec: Isnt this a regression that auth_url value won't be obfuscated?

Or have there been a need to scrub auth_url value at all?

sos/plugins/openstack_nova.py
```diff
@@ -89,7 +93,12 @@ def setup(self):
             self.var_puppet_gen + "_placement/etc/httpd/conf.d/",
             self.var_puppet_gen + "_placement/etc/httpd/conf.modules.d/*.conf",
             self.var_puppet_gen + "_placement/etc/my.cnf.d/tripleo.cnf",
-            self.var_puppet_gen + "/../memcached/etc/sysconfig/memcached"
+            self.var_puppet_gen + "/../memcached/etc/sysconfig/memcached",
+            self.var_puppet_gen + "_libvirt/etc/libvirt/",
+            self.var_puppet_gen + "_libvirt/etc/my.cnf.d/tripleo.cnf",
+            self.var_puppet_gen + "_libvirt/etc/nova/migration/"
+            "authorized_keys",
```
pmoravec: I guess this contains just some public keys that can be shared (but no secret there).

pmoravec:
PR seems good, just 1 relevant question (and one comment that can be ignored if I am right there).

sos/plugins/openstack_nova.py
stuggi: yes, the authorized_keys only have pub keys in.

sos/plugins/openstack_aodh.py
stuggi: we should not obfuscate the auth_url. the auth_url does not password information in and for trouble shooting it is useful to know which endpoint we connect to. E.g. this is an example of an auth_url:

auth_url=http://172.17.1.19:5000

bmr-cymru: Are you certain of that? Generally, an HTTP URL may include both username and password for HTML basic auth. If we can be absolutely certain that will never be the case here then this is OK (it would also be better if it did not elide the entire URL, but rather only actual auth elements, if present).

sos/plugins/openstack_nova.py
bmr-cymru: Are those keys generated by or embedded in the product, or are they privately generated by the user?

Remember that some organisations consider IP addresses to be "confidential" (we don't scrub to that level in `sos` but there are 3rd party tools that do this).

If the public data also winds up in logs then it's less of a problem but it's not so simple as "they are public keys -> it's fine".

sos/plugins/openstack_aodh.py
stuggi: This is not a basic auth or so. You have different options for that. This information is used to query a token from keystone. The User information is passed via json of the POST request.

# Authentication URL (string value)
#auth_url = <None>
# User id (string value)
#user_id = <None>
# User's password (string value)
#password = <None>

bmr-cymru: As long as users aren't likely to modify that, it's fine (we have had other URLs that we've ended up having to obfuscate in the past for this).

sos/plugins/openstack_nova.py
stuggi: they get generated by the triplo installer. if the pub key is too sensitive data I am ok to remove that information and we collect it on a case by case situation. should we do that?

I just checked the info on the a non container env and we collect there the full migration dir, which also has the private key . I think that is an issue which we should fix. we definitely need to exclude the etc/nova/migration/identity file.

stuggi: just for reference the /etc/nova/migration subdir got added just recently and it seems we collect everything under /etc/nova . thats why we have the identity file now

bmr-cymru: If it's generated by the product and the user has no (normal/direct) control over it then that's fine - it's just product internal debug data. It's only if the users can supply their own keys that it becomes a concern.

Issue #1137, opened 2017-11-08
haproxy uses py2 specific urlparse import


karibou:
New changes in the haproxy plugin (commit 6cf81bed12f119c580f9d3a0231aa46b2b165929) uses the python2 specific import syntax for urlparse. The plugin fails to load when run in python3.

bmr-cymru:
How come this didn't pop on any of the Travis builds? Afaik everything was passing.

karibou:
Well, it only shows up in my autopkgtests when running 'sosreport -l' as non-root. No error is triggered when it is run as root but I do see the same error : 

# sosreport --batch
                           
sosreport (version 3.5)          
                        
plugin haproxy does not install, skipping: No module named 'urlparse'
This command will collect system configuration and diagnostic
information from this Debian system. An archive containing the collected
information will be generated in /tmp/sos.4ba9j0h3.

bmr-cymru:
Got it - it's because nothing in the test suite actually loads every plugin. We should probably look at fixing that.

karibou:
Well, I think that I should propose the autopkgtests scripts for inclusion somewhere upstream. Let's discuss that on the ML

bmr-cymru:
Both have value I think.

PR #1138, opened 2017-11-08
[haproxy] Fix py2 specific import syntax for urlparse

karibou:
urlparse is now part of urllib in python3. Make sure that
the proxy behaves correctly on both versions.
(Closes: #1137)

Signed-off-by: Louis Bouchard <louis@ubuntu.com>

---


pmoravec:
+1. Could you please fix the `pep8` issue (add one new blank line after the try/catch code)?

karibou:
@pmoravec Done & forced pushed to the branch.

PR #1152, opened 2017-11-23
Kubernetes plugin for IBM Cloud Private

abdasgupta:
---

### DIFF
--- a/sos/plugins/kubernetes.py
+++ b/sos/plugins/kubernetes.py
@@ -9,11 +9,13 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.plugins import Plugin, RedHatPlugin
+from sos.plugins import Plugin, RedHatPlugin, UbuntuPlugin
+import os
 from os import path
+from os.path import expanduser
 
 
-class kubernetes(Plugin, RedHatPlugin):
+class kubernetes(Plugin, RedHatPlugin, UbuntuPlugin):
 
     """Kubernetes plugin
     """
@@ -24,11 +26,18 @@ class kubernetes(Plugin, RedHatPlugin):
     profiles = ('container',)
     files = ("/etc/origin/master/master-config.yaml",)
 
+    # For IBM Cloud Private(ICP):
+    # Set the environment variable like:
+    # export icp=true
+    # Then run the sosreport like:
+    # sosreport -o kubernetes -k kubernetes.kubeconfig=/root/.kube/config
     option_list = [
         ("all", "also collect all namespaces output separately",
             'slow', False),
         ("describe", "capture descriptions of all kube resources",
             'fast', False),
+        ("kubeconfig", "kubecofig file other than the default location",
+            'fast', ""),
         ("podlogs", "capture logs for pods", 'slow', False),
     ]
 
@@ -56,9 +65,20 @@ class kubernetes(Plugin, RedHatPlugin):
             self.add_journal(units=svc)
 
         # We can only grab kubectl output from the master
-        if self.check_is_master():
+        if self.check_is_master() or "icp" in os.environ:
             kube_cmd = "kubectl "
-            if path.exists('/etc/origin/master/admin.kubeconfig'):
+            if os.getenv("icp"):
+                # If the user wants to provide kubeconfig path,
+                # fetch the path from user option, otherwise fetch
+                # from default location.
+                if self.get_option('kubeconfig'):
+                    kubeconfig_path = self.get_option('kubeconfig')
+                else:
+                    kubeconfig_path = expanduser("~/.kube/config")
+                if kubeconfig_path != "" and path.exists(kubeconfig_path):
+                    kube_cmd += "--kubeconfig=" + kubeconfig_path + " "
+
+            elif path.exists('/etc/origin/master/admin.kubeconfig'):
                 kube_cmd += "--config=/etc/origin/master/admin.kubeconfig"
 
             kube_get_cmd = "get -o json "


abdasgupta:
Any update on this??

sos/plugins/kubernetes.py
```diff
@@ -60,70 +50,73 @@ def setup(self):
...
-
-                    self.add_cmd_output('%s events' % k_cmd)
+        kube_cmd = "kubectl "
+        kubeconfig_path = ""
+        if self.get_option('kubeconfig'):
+            kubeconfig_path = self.get_option('kubeconfig')
+        else:
+            kubeconfig_path = expanduser("~/.kube/config")
+        if kubeconfig_path != "" and path.exists(kubeconfig_path):
+            kube_cmd += "--kubeconfig=" + kubeconfig_path
```
pmoravec: I suggest adding `+ " "` at the end. To have aligned kube_cmd with trailing space in either case (i.e. if someone will add another option in advance).

Or alternatively, set `kube_cmd = "kubectl"` and then `kube_cmd += " --kubeconfig="..`.

sos/plugins/kubernetes.py
```diff
@@ -60,70 +50,73 @@ def setup(self):
...
+
+        kube_get_cmd = "get -o json "
```
pmoravec: Worth moving this assignment before the variable is used (i.e. around line 80)? To help code clarity/readiness.

sos/plugins/kubernetes.py
```diff
@@ -60,70 +50,73 @@ def setup(self):
...
-                                    k_cmd = '%s %s' % (kube_cmd, knsp)
-                                    self.add_cmd_output(
-                                        '%s describe %s %s' % (k_cmd, res, k))
-
-                if self.get_option('podlogs'):
+        # nodes and pvs are not namespaced, must pull separately
+        self.add_cmd_output([
+            "{} get -o json nodes".format(kube_cmd),
+            "{} get -o json pv".format(kube_cmd)
+        ])
```
pmoravec: Worth moving this command (and it's comment) before `# get all namespaces in use`? Since getting all namespaces logically precedes the `if self.get_option('all'):` command / code below, while getting the jsons is independent?

sos/plugins/kubernetes.py
```diff
@@ -60,70 +50,73 @@ def setup(self):
...
-                            self.add_cmd_output('%s logs %s' % (k_cmd, pod))
-
-            k_cmd = '%s get --all-namespaces=true' % kube_cmd
-            for res in resources:
-                self.add_cmd_output('%s %s' % (k_cmd, res))
+                    for res in resources:
+                        r = self.get_command_output(
+                            '%s %s' % (k_cmd, res))
+                        if r['status'] == 0:
+                            k_list = [k.split()[0] for k in
```
pmoravec: Can be here a few-words-comment what the `k_list` contains? To help understanding the code as `k_list` is a split of a formatted output of a command composed of .. (and one gets lost).

pmoravec:
In general the change makes sense. Just few minor suggestions for improvement made.

We wont be backward compatible due to the change (plugin renamed and old name used for something similar though not identical), but the change is sane wrt. plugin names.

sos/plugins/kubernetes.py
abdasgupta: Yes @pmoravec . The comment would increase the readability and I added few words as comment in the update of the PR.

sos/plugins/kubernetes.py
abdasgupta: It make sense, @pmoravec . I updated that.

sos/plugins/kubernetes.py
abdasgupta: Yes. Updated with your suggestion.

sos/plugins/kubernetes.py
abdasgupta: Yes, Updated

abdasgupta:
Hi @pmoravec . I updated the PR with all your suggestion. Thank you very much for the review.

pmoravec:
#1190 made by @TurboTurtle is in merge conflict with this, and the #1090 also fixes some bug that is copied to both plugins here (this was discovered even recently). Maybe worth rebasing this PR upon #1090 (once no comments are there (I dont think there will be changes, but who knows..))?

TurboTurtle:
I'm thinking we could actually simplify this a bit easier.

I just confirmed with the RH OCP support team that they don't even use the `kubernetes` plugin, everything needed for OpenShift is in the `origin` plugin.

So I think rather than splitting this into two plugins, we could just make `kubernetes` less RH-specific and add the `kubeconfig` stuff while removing the OCP checks for a kubeconfig file.

abdasgupta:
@TurboTurtle The kubernetes plugin that updated in this PR contains no RH-specific code. So,  can we just omit the RH-specific kubernetes plugin and make the plugin from this PR the only kubernetes plugin??

TurboTurtle:
I have some other questions around this PR.

- Why drop `check_is_master()`? kubectl only gives output from a master node - if this runs on a non-master node we'd just be collecting a lot of error output. 
- Why are we removing the checks for the base kubernetes packages? It makes sense to remove the `atomic_openshift_master` package, but the other RPMs are still valid.
- On that note, we need to handle being able to check for kubernetes services running in containers, just thought of this one.
- Since `sos` runs as root, this bit seems odd:

>         else:
>             kubeconfig_path = expanduser("~/.kube/config")
>         if kubeconfig_path != "" and path.exists(kubeconfig_path):
>             kube_cmd += "--kubeconfig=" + kubeconfig_path + " "


If the `kubeconfig` option isn't passed to sos, I would imagine we should just skip doing anything with it for `kubectl`. Is `~/.kube/config` standard for Ubuntu distros? It's not present for RHEL (by default).

abdasgupta:
@TurboTurtle Different products package kubernetes in different way. So, probably there is no definite way to determine if a node is master or worker. So, in this PR, the basic approach is taken. This PR looks for kubeconfig or if the default one is not used, then it expects the user to provide it.  Are you suggesting something else?

TurboTurtle:
Ok, I see now - I was misreading the config part earlier.

If you mean different distributions package kubernetes differently, then the plugin should have different classes for each distro that needs to do things slightly different. E.G. a `UbuntuKubernetes` class that controls enablement like you need for Ubuntu, and a `RedHatKubernetes` class that sets it for RH-family distros.

However, if you're referring to _layered_ products that use kubernetes (for example, OpenShift), then really what would be needed is a plugin for that product (like the `origin` plugin).

abdasgupta:
@TurboTurtle I was having the idea of totally generic kubernetes plugin. Be it from any package or layered product or just kubernetes binaries freshly built from source code. But if this idea of mine conflict with the overall structure of SOS project then I have to consider it differently.  The problem is there are so many layered projects or distro packages, if we want to create plugins for different products with same underlying kubernetes, the project size will become unnecessarily large.

TurboTurtle:
Let me clarify a bit:

The `kubernetes` plugin itself _should_ be generic insofar as the commands it runs. If a user installed the kube packages on Ubuntu or RHEL or Suse, the commands available don't change (barring major version differences).

What isn't necessarily generic is the config file locations between distros or perhaps enablement checks. In this case we'd want to do something like in the Wiki (https://github.com/sosreport/sos/wiki/How-to-Write-a-Plugin#cross-platform-plugins):

```python
from sos.plugins import Plugin, RedHatPlugin, UbuntuPlugin

class Kubernetes(Plugin):

    name = "kubernetes"

    def setup(self):
        # stuff that is already in the plugin


class RedHatKubernetes(Kubernetes, RedHatPlugin):

    # only run on RHEL if any of these packages are present
    packages = ('kuberentes', 'kubernetes-master')

    def setup(self):
       # check the RH config location or do RH-only enablement
       super(RedHatKubernetes, self).setup()


class UbuntuKubernetes(Kubernetes, DebianPlugin):

    def setup(self):
        # check the Ubuntu config location
        super(RedHatKubernetes, self).setup()

```

So that if it runs on Ubuntu for example, sos calls `UbuntuKubernetes.setup()` which in turn calls `Kubernetes.setup()`.


However, if a layered product wraps kubernetes, like OpenShift does, then what typically happens is another plugin is made for the layered product. This is because even if the layered product is built on kubernetes, doesn't mean `kubectl` output is meaningful for the product. For example, the `origin` plugin exists to capture `oc` or `ocadm` output which itself (`oc`) wraps kubernetes, but for OpenShift output from `kubectl` is either not applicable/appropriate or would be duplicating output from `oc`.


All this to say:

Generic/standaridzed kubernetes commands, yes. One size fits all across distros, no.

abdasgupta:
@TurboTurtle okay. Let me refactor the PR according to your suggestion.

abdasgupta:
@TurboTurtle Our main push for this SOS plugin actually is for ICP (IBM cloud private) product. This is a layered product. So I will change the name of the plugin specific for ICP. I hope it aligns with your suggestion?

abdasgupta:
Modified the PR as it is best suited for ICP (IBM Cloud Private) cc @pmoravec @TurboTurtle

bmr-cymru:
This is largely a copy of the existing `kubernetes` plugin with a name change: I can't merge that.

You need to follow the suggestion @TurboTurtle made and use the existing plugin mechanisms in order to make the existing plugin work also for ICP. We're happy to take additional changes to allow that, including defining a new policy and tagging classes if required, but we cannot have a proliferation of near-identical plugins like this. It's unmaintainable and a poor experience for our users.

bmr-cymru:
(If you read the comment @TurboTurtle made again, he's suggesting that if ICP *wraps the kubernetes commands* then you should implement a new ICP plugin to collect *those wrapped commands* - not to simply duplicate what's already going on in `kubernetes.py`. If there is no wrapping going on then I don't really see why ICP cannot play nicely with the existing plugin: the code submitted is almost line-by-line identical).

abdasgupta:
@bmr-cymru I set up a kubernetes cluster through kubeadm following this guide: https://kubernetes.io/docs/setup/independent/install-kubeadm/ . But the existing kubernetes plugin was unable to collect data(like POD info etc.) from the master. So , I modified the existing plugin just to remove the constraint of installing any kubernetes package and the constraint of a specific `KUBECONFIG` file that the existing k8s plugin is using to collect data. And then it worked. It collected all the node info, POD info etc. Can you suggest me if there is any other way I should take to make this k8s plugin generic which can run in environments like kubeadm?

bmr-cymru:
So, re-reading the review and discussion to date, it seems that all that is needed here is some revision to the patch to allow the existing `kubernetes` plugin to be enabled and to collect data on IBM (and ideally other) Kubernetes deployments?

We can't merge the current patch in the PR because it duplicates the entire plugin, but it doesn't feel like there is really that much work to do here to achieve the actual goal?

abdasgupta:
@bmr-cymru so should we go with refactoring the kubernetes plugin?? But the problem is as per the guide there seems inherent problem to use the same plugin for both package based k8s offering and docker image based k8s offering. ICP is delivered through docker images.

bmr-cymru:
@abdasgupta any updates? We are trying to clear out the final pull requests before releasing 3.6.

bmr-cymru:
For e.g. It might help to just write out the changes you had to make to make the existing `kubernetes` plugin work for ICP. We can then try to understand how to make a version that will work for that without duplicating any plugin code.

abdasgupta:
Hi @bmr-cymru ,
I have updated the PR with the changes that will make the existing kubernetes plugin work for ICP. Can you check please?

abdasgupta:
any update on this? @bmr-cymru

sos/plugins/kubernetes.py
```diff
@@ -15,29 +15,38 @@
...
+        ("kubeconfig", "kubecofig file other than the default location",
+            'fast', ""),
+        ("icp", "set this option true if your installation is ICP based",
+            'fast', False),
         ("podlogs", "capture logs for pods", 'slow', False),
     ]
 
     def check_is_master(self):
+        if self.get_option('icp'):
+            return True
```
bmr-cymru: ICP can never be master? This short-circuits the rest of this method call entirely when the `kubernetes.icp` option is given.

is that really what you intend?

sos/plugins/kubernetes.py
```diff
@@ -65,6 +74,13 @@ def setup(self):
             kube_cmd = "kubectl "
             if path.exists('/etc/origin/master/admin.kubeconfig'):
                 kube_cmd += "--config=/etc/origin/master/admin.kubeconfig"
+            else:
+                if self.get_option('kubeconfig'):
+                    kubeconfig_path = self.get_option('kubeconfig')
+                else:
+                    kubeconfig_path = expanduser("~/.kube/config")
+                if kubeconfig_path != "" and path.exists(kubeconfig_path):
+                    kube_cmd += "--kubeconfig=" + kubeconfig_path + " "
```
bmr-cymru: None of these lines will run if `icp=True`... how come they are being added here?

sos/plugins/kubernetes.py
```diff
@@ -15,29 +15,38 @@
...
+    # For IBM Cloud Private(ICP), run the sosreport like:
+    # sosreport -o kubernetes -k kubernetes.icp=True
     option_list = [
         ("all", "also collect all namespaces output separately",
             'slow', False),
         ("describe", "capture descriptions of all kube resources",
             'fast', False),
+        ("kubeconfig", "kubecofig file other than the default location",
+            'fast', ""),
+        ("icp", "set this option true if your installation is ICP based",
```
bmr-cymru: We really cannot just detect this from the environment? Any plugin that cannot work properly without the user intervening to set plugin options delivers a poor user experience. Less options -> happier users.

sos/plugins/kubernetes.py
abdasgupta: Hi @bmr-cymru 
Pushed some update on the patch including the changes you requested. Can you check please??

abdasgupta:
Any update @bmr-cymru ??

bmr-cymru:
You've changed the mechanism for the user to specify that ICP is in use, but not the spirit: unless the `icp` environment variable is automatically defined in the user's shell by the ICP installation then this does not materially change the problem. A user still must be aware of what system they are using, and set environment variables and pass in configuration file locations - this is not what I requested - it needs to be automatic in the case of a default installation.

There must exist some paths, configuration entries, etc. that you can use to **automatically** distinguish an ICP installation from another.

abdasgupta:
@bmr-cymru a user does not need to provide configuration file specifically. It will fetch the default location automatically. ICP is a docker image based installation. Everything is managed there through docker containers. There is no standard location for config files or any sort of similar files in ICP. A user can install ICP from any location. And user was not needed to provide environment variable as ICP if we could write a different icp plugin. Which I wrote earlier, but since we have to use same plugin for both ICP and other package based installation, I don't know any other mechanism we can take. Do you have any suggestions?

abdasgupta:
okay @bmr-cymru . For the config file option, I will change the comment to instruct the user that they can run sosreport for ICP with only following commands:
```
    # export icp=true
    # Then run the sosreport like:
    # sosreport -o kubernetes
```
See, all the components of ICP are run as docker containers. But the concept of kubernetes is still valid for ICP. We have apiserver  in ICP as well and that needs sosreport to post the debugging report of the apiserver. sosreport does not need to support collecting info from the container in our case.

bmr-cymru:
I think you need to define this use-case better: it's clearly not as simple as just making changes to this one plugin.

I do not believe that it is impossible for you to to detect that you are running on the IBM environment and to avoid the need for users to set environment variables. That is an ugly kludge and we would not normally allow a plugin to behave this way: it's fine to obtain needed *values* from the environment, but requiring the user to set environment variables just to have basic plugin functionality work properly is not.

Look at how this is handled for other environments like `Atomic`, or `IBMKVM` which defines its own policy. That may not be needed in this case but it seems like you want the kind of behaviour that the policy-defined tagging classes provide: a means to dynamically select alternate configuration paths and other options at runtime.

pmoravec:
Bump @abdasgupta .

TurboTurtle:
Given the long period of inactivity, and the non-trivial conflicts with the current master branch I am closing this out now.

If you would still like to add ICP detection to the `kubernetes` plugin, please follow the advice in the reviews above, rebase on master, and open a new PR.

PR #1153, opened 2017-11-24
[networking] Add /etc/netplan/*.yaml collection for Ubuntu.

slashdd:
Netplan is a YAML network configuration abstraction for various
backends (NetworkManager, networkd). It reads network configuration
from /etc/netplan/*.yaml.

Related documents:
https://wiki.ubuntu.com/Netplan
https://wiki.ubuntu.com/Netplan/Design

---

### DIFF
--- a/sos/plugins/networking.py
+++ b/sos/plugins/networking.py
@@ -375,7 +375,8 @@ class UbuntuNetworking(Networking, UbuntuPlugin, DebianPlugin):
             "/etc/network/interfaces.d",
             "/etc/ufw",
             "/var/log/ufw.Log",
-            "/etc/resolv.conf"
+            "/etc/resolv.conf",
+            "/etc/netplan/*.yaml"
         ])
         self.add_cmd_output([
             "/usr/sbin/ufw status",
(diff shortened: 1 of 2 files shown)


slashdd:
Output of `python sosreport -o networking` from Ubuntu 18.04/Bionic with the above proposal.

```
# ls -altr /path_to_sosreport/etc/netplan/
-rw-r--r-- 1 root root   59 Nov 15 03:57 10-lxc.yaml

# cat /path_to_sosreport/etc/netplan/10-lxc.yaml 
network:
  ethernets:
    eth0: {dhcp4: true}
  version: 2
```

slashdd:
Will close this PR and re-submit it since I pushed 2 commits for 2 different issues into the same PR.

slashdd:
#1161 is the new good one.

bmr-cymru:
In future, if you do something like that, just fix your branch locally, and then run `git push --force`. You can do any amount of `git reset --hard`, `git commit --amend`, or `git rebase -i`: you can literally rewrite the branch entirely and then replace it with another with a simple `git push --force`.

That way we don't burn through extra PR numbers for no reason and all discussion and review comments are kept in one place & are easy to find.

PR #1154, opened 2017-11-25
set proper PATH for SCL commands & call pg_dump from SCL properly

pmoravec:
pg_dump called from SCL needs to have proper PATH reflecting where SCL deploys its binaries.

Also the destination filename of the dump should match the non-SCL dump to simplify work of tools further manipulating with the dump(s).

---

### DIFF
--- a/sos/plugins/postgresql.py
+++ b/sos/plugins/postgresql.py
@@ -34,8 +34,6 @@ class PostgreSQL(Plugin):
 
     packages = ('postgresql',)
 
-    tmp_dir = None
-
     password_warn_text = " (password visible in process listings)"
 
     option_list = [
@@ -47,11 +45,9 @@ class PostgreSQL(Plugin):
         ('dbport', 'database server port number', '', '5432')
     ]
 
-    def pg_dump(self, pg_dump_command="pg_dump", filename="sos_pgdump.tar"):
+    def do_pg_dump(self, scl=None, filename="pgdump.tar"):
         if self.get_option("dbname"):
             if self.get_option("password") or "PGPASSWORD" in os.environ:
-                self.tmp_dir = tempfile.mkdtemp()
-                dest_file = os.path.join(self.tmp_dir, filename)
                 # We're only modifying this for ourself and our children so
                 # there is no need to save and restore environment variables if
                 # the user decided to pass the password on the command line.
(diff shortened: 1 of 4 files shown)


pmoravec:
The PR is not yet complete - it works well just for Software collections with "rh" as its provider. I haven't found a way (so far) to identify a provider for given SCL (pending on me).

So the https://github.com/sosreport/sos/pull/1154/commits/0b8aaaa7134b658a92c34b1be3d67307e93e7fb4#diff-ee17c474955d33a9ccffd4fd96ab0773R1075 is planned to be updated.

Apart of that, the PR should be OK.

pmoravec:
Provider is now properly loaded, it is always in `/etc/scl/prefixes/${scl}` file, like:

    # cat /etc/scl/prefixes/tfm 
    /opt/theforeman
    # cat /etc/scl/prefixes/rh-postgresql95 
    /opt/rh
    #

bmr-cymru:
```python
    def convert_cmd_scl(self, scl, cmd):
        """wrapping command in "scl enable" call and adds proper PATH
        """
        # have some default prefix to PATH
        prefix = '/opt/rh'
```

Really, I think the selection of the default path prefix for SCL commands should be a callback into the policy: `"/opt/rh"` is decidedly Red Hat specific, but since only Red Hat distributions are currently using the SCLPlugin mechanism I am OK with that being fixed in a follow up patch.

So I'd like the above line to eventually be:

```python
        prefix = self.policy().default_scl_prefix()
```

pmoravec:
No problem with changing that.

sos/plugins/postgresql.py
```diff
@@ -47,34 +48,34 @@ class PostgreSQL(Plugin):
         ('dbport', 'database server port number', '', '5432')
     ]
 
-    def pg_dump(self, pg_dump_command="pg_dump", filename="sos_pgdump.tar"):
+    def pg_dump(self, scl=None):
         if self.get_option("dbname"):
             if self.get_option("password") or "PGPASSWORD" in os.environ:
                 self.tmp_dir = tempfile.mkdtemp()
```
bmr-cymru: Why are we using `tempfile` directly here and not simply directing the output into the desired path in the report archive?

It makes no sense to generate a file by running a command in `setup()` and then collect it: it should just be written to the report in the first place.

This is an anti-pattern we've been eliminating for years, first in Gluster and the various RHEV plugins, and now here... :-/

I realise this wasn't introduced in these patches but it needs to go, and soon (along with the `postproc()` cleanup).

sos/plugins/postgresql.py
```diff
@@ -47,34 +48,34 @@ class PostgreSQL(Plugin):
         ('dbport', 'database server port number', '', '5432')
     ]
 
-    def pg_dump(self, pg_dump_command="pg_dump", filename="sos_pgdump.tar"):
+    def pg_dump(self, scl=None):
```
bmr-cymru: Suggest renaming this method to `do_pg_dump()` or similar: it's annoying trying to discriminate the method name from the command name in this file.

sos/plugins/postgresql.py
```diff
@@ -35,6 +35,7 @@ class PostgreSQL(Plugin):
     packages = ('postgresql',)
 
     tmp_dir = None
+    filename = 'sos_pgdump.tar'
```
bmr-cymru: How can this work if there is a system copy of Postgres as well as an SCL version (or multiple SCL versions)?

I don't see how we can add this capability and not change the path of the data: I don't see anything in the plugin that prevents the SCL `self.pg_dump()` call from overwriting the work of the call in the main superclass `setup()`? I also don't really know what that is going to do to the collected file records: we are collecting a file twice and it is not the same file!

Also, doesn't that mean that we have these dumps under some bizarre tmp-named path in `/tmp`? Rather than in the `sos_commands/` tree where they belong?

Although the SCL work needs to land soon it seems that this plugin needs major work anyway...

Issue #1155, opened 2017-11-28
Regresion from [sosreport] add per-plugin package verification lists


BryanQuigley:
All packages =  lines don't appear to work in Ubuntu (and likely debian)

This results in less plugins running and none running based on installed packages.  Two affected plugins are cups (which is what I bisected) and acpid. 

On Trusty it goes from 59 to 50 plugins, on Artful from 63 to 53.

- https://github.com/sosreport/sos/commit/82e4ee224706b5eb377b35f88c852dd79c7cb25d
Found during SRU testing - https://bugs.launchpad.net/ubuntu/+source/sosreport/+bug/1734983

BryanQuigley:
Actually I'm not sure this works for RedHat either:
PackageManager has the order defined as:
 chroot=None, query_command=None, 
 verify_command=None, verify_filter=None

Redhat policy calls it 
query_command=self._rpmq_cmd, verify_command=self._rpmv_cmd,                                           verify_filter=self._rpmv_filter, chroot=sysroot

and debian and suse just seem to expect the query command to be first.

BryanQuigley:
With this fix 61 plugins run out of 63 expected.  of the two different than expected
dmraid - I don't have installed, package= was added
boot - package names were added, but none patch debian/ubuntu

bmr-cymru:
Order is not significant for kwargs: it's mixing kwarg and non-kwarg call styles that causes the problem: the Debian policy is currently passing everything positionally which is incorrect for a method call that only uses keyword arguments.

That said: I prefer when we keep declaration/call ordering consistent because it makes the code easier to read and follow, so as well as reverting the argument ordering on the PM initialiser we need to ensure the Debian policy uses the correct kwarg names in the call.

slashdd:
:+1: +1 on fixing the ordering.

Another thing that can modifiy is `sos/policies/debian.py` to be align with RHEL policies.

```
diff --git a/sos/policies/debian.py b/sos/policies/debian.py
index dc7f2e6..258b841 100644
--- a/sos/policies/debian.py
+++ b/sos/policies/debian.py
@@ -10,8 +10,9 @@ class DebianPolicy(LinuxPolicy):
     vendor_url = "http://www.debian.org/"
     report_name = ""
     ticket_number = ""
-    package_manager = PackageManager(
-        "dpkg-query -W -f='${Package}|${Version}\\n'")
+    _debq_cmd = "dpkg-query -W -f='${Package}|${Version}\\n'"
[...]

PR #1156, opened 2017-11-29
[policies] Reset to expected variable order for packagemanager

BryanQuigley:
Commit 82e4ee2 changed the order of variables in packagemanager
to what none of the policies expected.  This sets it back so
query is always first.  Red Hat policy is the only one that
uses more than one variable for it.

Closes: #1155 

---

### DIFF
--- a/sos/policies/__init__.py
+++ b/sos/policies/__init__.py
@@ -59,8 +59,8 @@ class PackageManager(object):
     verify_filter = None
     chroot = None
 
-    def __init__(self, chroot=None, query_command=None,
-                 verify_command=None, verify_filter=None):
+    def __init__(self,  query_command=None, verify_command=None,
+                 verify_filter=None, chroot=None,):
         self.packages = {}
 
         self.query_command = query_command if query_command else None


bmr-cymru:
Yes: that's a better fix.

I'll look over this some more though: reading the verify patch back there's a load of completely unnecessary "x = foo if foo else None.." that appears completely redundant.

slashdd:
Dear maintainer,

Do we have an idea as of when this patch will be merge upstream ?

We would like to release a new sosreport version in Ubuntu beginning of Jan 2018 and this bug is a blocker for us at the moment.

Regards,
Eric

bmr-cymru:
@slashdd we don't have a release planned until April 2018 - you can find our release schedules in the [GitHub milestones](https://github.com/sosreport/sos/milestones).

I've approve this commit, but before it's merged I want us to also address the bug in the Debian policy that caused it to be a problem in the first place.

If you're happy to pick those patches up from Git and include them in your update, that's fine. Otherwise we can speak to the other Ubuntu and Debian contributors and see about making a point release to include this.

slashdd:
@bmr-cymru 

Are you referring to what I have posted in PR #1155 ?

```
diff --git a/sos/policies/debian.py b/sos/policies/debian.py
index dc7f2e6..258b841 100644
--- a/sos/policies/debian.py
+++ b/sos/policies/debian.py
@@ -10,8 +10,9 @@ class DebianPolicy(LinuxPolicy):
     vendor_url = "http://www.debian.org/"
     report_name = ""
     ticket_number = ""
-    package_manager = PackageManager(
-        "dpkg-query -W -f='${Package}|${Version}\\n'")
+    _debq_cmd = "dpkg-query -W -f='${Package}|${Version}\\n'"
+    _debv_cmd = "dpkg --verify" 
+    _debv_filter = ""
     valid_subclasses = [DebianPlugin]
     PATH = "/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games" \
            + ":/usr/local/sbin:/usr/local/bin"
@@ -20,8 +21,11 @@ class DebianPolicy(LinuxPolicy):
         super(DebianPolicy, self).__init__(sysroot=sysroot)
         self.report_name = ""
         self.ticket_number = ""
-        self.package_manager = PackageManager(
-            "dpkg-query -W -f='${Package}|${Version}\\n'")
+        self.package_manager = PackageManager(query_command=self._debq_cmd,
+                                              verify_command=self._debv_cmd,
+                                              verify_filter=self._debv_filter,
+                                              chroot=sysroot)
+
         self.valid_subclasses = [DebianPlugin]
```

bmr-cymru:
Not directly but those changes cover what I'd like to see: thanks.

It's not really about aligning to the RHEL policy though, it's just about using the Python interface in the way it is written: Python does not impose any kind of ordering on keyword args, but if the keywords are missing it will still attempt to honour the call by mapping positional call arguments to corresponding keywords in the declaration.

This behaviour can be unhelpful since at times it masks buggy use of an interface as is the case here.

in this case the Debian policy invokes the `PackageManager` call as though it only uses positional arguments. This is undoubtedly wrong and a bug and needs to be fixed.

The change requested here is actually purely cosmetic: it's just about reverting to the historic argument ordering for this method, which I agree is a positive change since there was never any good reason to have messed around with it, but that is entirely separate from the Debian policy's historical misuse of this interface.

slashdd:
@bmr-cymru  
Ok thanks, I'll then do a PR with the above later today.

bmr-cymru:
Sounds good! 👍

slashdd:
Done via #1162

BryanQuigley:
Closing this as #1162 is the better way to fix this.

PR #1158, opened 2017-11-30
[postgresql] use a single name for the db dump file

dougsland:
Currently, there are two names for the file that contain
the postgresql db dump. The original one, which is sos_pgdump.tar
and the new one sos_scl_pgdump.tar based on postgresql95.
However, changing the name of file that contain the pgdump
because of new technologies break the tools that are
based on sos tool. This patch make sure we use the same name.

---

### DIFF
--- a/sos/plugins/postgresql.py
+++ b/sos/plugins/postgresql.py
@@ -141,8 +141,7 @@ class RedHatPostgreSQL(PostgreSQL, SCLPlugin):
 
         if scl in self.scls_matched:
             self.pg_dump(
-                pg_dump_command="scl enable rh-postgresql95 -- pg_dump",
-                filename="sos_scl_pgdump.tar"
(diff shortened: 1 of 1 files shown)


pmoravec:
This is already included in PR 1154, in particular in https://github.com/sosreport/sos/pull/1154/files#diff-93be9976582cadc41ed1cec376205196R55 .

dougsland:
Duplicate of #1154

dougsland:
Looking forward to see PR1154 merged asap. Our tool still broken :(

bmr-cymru:
I doubt that is going to change for the reasons I mention in the review comments: unfortunately I think your tool will need to adapt to this. We can keep the file name the same, but only if we put each dump in a subdirectory named for the SCL version. We can also keep the *main* Postgres install's file name the same, but I don't see a way that we can support both without any user visible change at all.

PR #1162, opened 2017-12-06
[sosreport] Fix debian policy PackageManager misusage.

slashdd:
Python does not impose any kind of ordering on keyword args,
but if the keywords are missing it will still attempt to honour
the call by mapping positional call arguments to corresponding
keywords in the declaration.

In this case the Debian policy invokes the PackageManager call
as though it only uses positional arguments. This is undoubtedly
wrong.

---


bmr-cymru:
Nope: ack from me. I'll get this included asap (iirc I was starting on #1120 last night when I stopped).

PR #1163, opened 2017-12-11
 [ipa] use correct PKI directories for tomcat version

tscherf:
The PKI subsystem uses different folders in IPA v3 and v4 for the NSS DB.
Therefore, different paths strings need to be passed to certutil, based on
the installed IPA version.

### DIFF
--- a/sos/plugins/ipa.py
+++ b/sos/plugins/ipa.py
@@ -79,6 +79,9 @@ class Ipa(Plugin, RedHatPlugin):
         self.pki_tomcat_dir_v4 = "/var/lib/pki/pki-tomcat"
         self.pki_tomcat_dir_v3 = "/var/lib/pki-ca"
 
+        self.pki_tomcat_conf_dir_v4 = "/etc/pki/pki-tomcat/ca"
+        self.pki_tomcat_conf_dir_v3 = "/etc/pki-ca"
+
         if self.ipa_server_installed():
             self._log_debug("IPA server install detected")
 
@@ -104,7 +107,6 @@ class Ipa(Plugin, RedHatPlugin):
             "/etc/dirsrv/slapd-*/schema/99user.ldif",
             "/etc/hosts",
             "/etc/named.*",
-            "/etc/pki-ca/CS.cfg",
             "/etc/ipa/ca.crt",
             "/etc/ipa/default.conf",
             "/var/lib/certmonger/requests/[0-9]*",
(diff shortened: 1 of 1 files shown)


sos/plugins/ipa.py
```diff
@@ -120,14 +120,24 @@ def setup(self):
         self.add_forbidden_path("/etc/dirsrv/slapd-*/pwdfile.txt")
         self.add_forbidden_path("/etc/named.keytab")
 
+        ### Make sure to lookup the right NSS DB for PKI certs
+        if ipa_version == "v3":
+            self.add_cmd_output([
+                "certutil -L -d %s/alias" % self.pki_tomcat_dir_v3
+            ])
```
bmr-cymru: No need to pass a list for a single command: just `self.add_cmd_output("command string")`.

sos/plugins/ipa.py
```diff
@@ -120,14 +120,24 @@ def setup(self):
...
+        elif ipa_version == "v4":
+            self.add_cmd_output([
+                "certutil -L -d %s/alias" % self.pki_tomcat_dir_v4
```
bmr-cymru: Ditto.

bmr-cymru: Considering that these two blocks are so similar though, why not just assign `pki_tomcat_dir` inside the conditionals, and then add the command once: this means there is only one copy of the (identical) command string to maintain and it prevents any sync problems when someone edits the file in future.

bmr-cymru: You could also then use a conditional assignment and get rid of the blocks entirely:

```
        pki_tomcat_dir = self.pki_tomcat_dir_v4 if ipa_version == "v4" else self.pki_tomcat_dir_v3
        self.add_cmd_output("certutil -L -d %s/alias" % self.pki_tomcat_dir)
```

bmr-cymru: In fact (although it's not code changed in this patch): there is no good reason for either of the `pki_tomcat_dir` variables to be instance members of the plugin class. They are constants. They can be set at file scope and then referenced throughout the plugin without any need for the `self.` prefix (which causes line lengths to expand).

tscherf: ack. makes sense. I changed the patch and also included a change for the configuration files. The variables pki_tomcat_dir and pki_tomcat_conf_dir are still instance members of the plugin class though. Maybe this can be changed with a later commit. I think the whole plugin needs some more work.

tscherf:
I have more changes I'd like to get merged into the ipa module. But I'd like to base them on an up2date version of the module. Do you want me to perform any more updates to the patch in this PR before it can be merged?

pmoravec:
I think any approach is OK wrt. more commits or incrementally. IMHO the current commit should be separate as it changes one logical/coherent area, and other changes should be done in further commits on top of this one.

There are few pep8 issues (see https://travis-ci.org/sosreport/sos/jobs/314767272 lines 1499-1501), apart of that +1 to the commit / PR.

tscherf:
I fixed the PEP8 warnings with the latest commit.

pmoravec:
+1

tscherf:
bump for review

Issue #1169, opened 2017-12-19
[utilities] collecting binary data removes "non-UTF-8 bytes"


pmoravec:
Collecting literally any command output assumes the data are a valid UTF-8 string and the extra bytes are silently ignored. Usually, this is OK.

But sosreport collects also tarballs, at least on 2 places (`postgresql` and `lvm2` plugins) where ignoring the bytes causes sosreport stores an invalid file (i.e. stores tarball that cant restore its content).

There can be workaround in storing the binary output file somewhere and call `add_copy_spec`, but ideally `add_cmd_output` should be extended by some option like "treat_binary" to prevent ignoring the "non-UTF-8 bytes".

pmoravec:
Workaround for `postgresql`:

    diff --git a/sos/plugins/postgresql.py b/sos/plugins/postgresql.py
    index 9ba696be..071b24e4 100644
    --- a/sos/plugins/postgresql.py
    +++ b/sos/plugins/postgresql.py
    @@ -20,6 +20,7 @@
     
     import os
     import tempfile
    +from shutil import rmtree
     
     from sos.plugins import (Plugin, RedHatPlugin, UbuntuPlugin, DebianPlugin,
                              SCLPlugin)
    @@ -45,6 +46,8 @@ class PostgreSQL(Plugin):
             ('dbport', 'database server port number', '', '5432')
         ]
     
    +    tmpdirs = []
    +
         def do_pg_dump(self, scl=None, filename="pgdump.tar"):
             if self.get_option("dbname"):
                 if self.get_option("password") or "PGPASSWORD" in os.environ:
    @@ -54,22 +57,38 @@ class PostgreSQL(Plugin):
                     if self.get_option("password") is not False:
                         os.environ["PGPASSWORD"] = str(self.get_option("password"))
     
    +                # workaround to add_cmd_output for non UTF-8 binary output:
    +                # create the tarball file and copy it instead of collecting
    +                # its string output. Need to remove temp.dir(s) afterwards
    +                tmp_dir = tempfile.mkdtemp()
    +                self.tmpdirs.append(tmp_dir)
    +                dest_file = os.path.join(tmp_dir, filename)
    +
                     if self.get_option("dbhost"):
    -                    cmd = "pg_dump -U %s -h %s -p %s -w -F t %s" % (
    +                    cmd = "pg_dump -U %s -h %s -p %s -w -f %s -F t %s" % (
                             self.get_option("username"),
                             self.get_option("dbhost"),
[...]

PR #1170, opened 2017-12-20
[plugins] allow add_cmd_output to collect binary output

pmoravec:
If a command output is a true binary data, allow add_cmd_output to
collect the raw content and dont try to decode it as UTF-8.

Resolves: #1169

---

### DIFF
--- a/sos/archive.py
+++ b/sos/archive.py
@@ -85,6 +85,9 @@ class Archive(object):
     def add_string(self, content, dest):
         raise NotImplementedError
 
+    def add_binary(self, content, dest):
+        raise NotImplementedError
+
     def add_link(self, source, link_name):
         raise NotImplementedError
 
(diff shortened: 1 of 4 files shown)


sos/archive.py
```diff
@@ -82,7 +82,7 @@ def log_debug(self, msg):
     def add_file(self, src, dest=None):
         raise NotImplementedError
 
-    def add_string(self, content, dest):
+    def add_string(self, content, dest, treat_binary):
         raise NotImplementedError
```
bmr-cymru: `binary` alone seems sufficient; kwargs already take up enough line width.

sos/archive.py
```diff
@@ -198,12 +198,14 @@ def add_file(self, src, dest=None):
         self.log_debug("added %s to FileCacheArchive '%s'" %
                        (file_name, self._archive_root))
 
-    def add_string(self, content, dest):
+    def add_string(self, content, dest, treat_binary=False):
```
bmr-cymru: It's not a string if it's binary, is it?

pmoravec:
PR updated:
- `treat_binary` shortened
- `add_binary` method added
- adding new flag to `executed_commands`
- methods `do_cmd_[command|private]_sub` prevented from applying regexp substitution on binary data

PR #1174, opened 2017-12-22
[gluster] Optionally enable dump

psachin:
Currently a dump is collected every time sosreport is run. Dump is not
required by Gluster support team and also consumes space in
`/var/run/gluster/` unnecessarily.

This patch adds an option to enable dump when required using
```
-k gluster.dump
```

Command
```
sosreport -o gluster -k gluster.dump
```

Dump will is disabled by default.

(Few `pylint` changes are also included with this patch which are not
related to enhancement. Let me know if you want me to remove those.)

---

### DIFF
--- a/sos/plugins/gluster.py
+++ b/sos/plugins/gluster.py
@@ -30,6 +30,8 @@ class Gluster(Plugin, RedHatPlugin):
     packages = ["glusterfs", "glusterfs-core"]
     files = ["/etc/glusterd", "/var/lib/glusterd"]
 
+    option_list = [("dump", "enable glusterdump support", "slow", False)]
+
     def get_volume_names(self, volume_file):
         """Return a dictionary for which key are volume names according to the
         output of gluster volume info stored in volume_file.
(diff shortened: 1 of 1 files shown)


sos/plugins/gluster.py
```diff
@@ -111,32 +113,31 @@ def setup(self):
...
-                                    "gluster volume heal %s info" % volname,
-                                    "gluster volume heal %s info split-brain"
-                                    % volname,
-                                    "gluster snapshot list %s" % volname,
-                                    "gluster volume quota %s list" % volname,
-                                    "gluster volume rebalance %s status"
-                                    % volname,
-                                    "gluster snapshot info %s" % volname,
-                                    "gluster snapshot status %s" % volname])
+                self.add_cmd_output(
```
TurboTurtle: Minor styling correction here. Put the starting bracket in the same line as `add_cmd_output` and the closing bracket and paren should be on a new line with the same indention as the call.

E.G.:
```
self.add_cmd_output([
    ...list...
])
```

psachin:
Hi @TurboTurtle 

I fixed the indentation to match with `add_copy_spec()` above and to also pass `flake8` test. Let me know if that's OK.

TurboTurtle:
LGTM

psachin:
@bmr-cymru Can you please review as well?

bmr-cymru:
Merged, thanks!

Issue #469, opened 2015-01-20
[sosreport] tarball names make my eyes bleed


bmr-cymru:
```
sosreport-bmr-rhel7-vm1-20150120185311.tar.xz
sosreport-bmr-rhel7-vm1-20150120185466.tar.xz
sosreport-bmr-rhel7-vm1-20150120190021.tar.xz
```

I'd meant to file this for review for 3.2 but it missed the boat. We've changed the tar/directory name format a few times over the years but I still find the long numbers awkward and clumsy to deal with (esp. at the terminal when working with several reports from the same host at slightly different times).

One improvement would be to just add dashes for the date components:

```
sosreport-bmr-rhel7-vm1-2015-01-20-185466.tar.xz
```

But I think we can do better; either way, if we're going to make another change it should be something we're going to stick with for a reasonable length of time.

battlemidget:
Maybe break it up under a directory labeled like 'sos-2015-01-20/sosreport-bmr-rhel7.tar.xz' -- just throwing out an idea, not really sure the best way to handle this

TurboTurtle:
I'm not alone with the bleeding eyes it seems!

I was thinking that reducing the timestamp and adding a short 3-character random string would do well. It would still allow for organizing/tab-completing by date for the most part, and would make referencing a specific sosreport easier between multiple parties - even without referencing the timestamp.

For example `sosreport-jhunsake.1234567-2017-01-24-iia.tar.xz` could be referenced simply as "sosreport iia" or somesuch between parties as the chance for string collisions is extremely low. 

I have something whipped up for this but held back on the PR for now.

https://github.com/TurboTurtle/sos/commit/5b7ccc2f27df939105e9e31927987830d7c87151

bmr-cymru:
I'd be inclined to go for 7 chars for the hash/uniquifier - 3 chars doesn't give a whole lot of space for environments using large-scale automation.

More generally though (and the reason I didn't push too hard on this, and kinda parked it..) is that it needs coordination with a whole bunch of stakeholders at Red Hat, Canonical and elsewhere.

If you're happy to work on this, and help come up with a plan for when we are going to introduce this - or (I literally just thought of this) - a policy/config file tunable to switch between "legacy naming" and "nice naming" (or whatever.. :) ), that would be really awesome!

TurboTurtle:
I like the idea of a policy/tunable, and sure I can run with this.

[...]

PR #1175, opened 2017-12-24
[sosreport] Allow policies to change report name generation

TurboTurtle:
Previously, sosreport tarball names consisted of a very long date
string that was not pleasant to read, especially when dealing with
multiple sosreports for an ongoing issue.

This allows policies to determine how the report name should be
formatted by adding a 'name_scheme' attribute to the policies. By
default, this is set to 'legacy' and will result in names that are
unchanged - E.G this will give tarball names like
'sosreport-tux.123456-20170120185433'.

Additionally, there is a new 'friendly' name scheme that results in
tarball names like 'sosreport-tux-123456-2017-12-24-ezcfcop.tar.xz'.

A policy may also set name_scheme to 'custom' and provide it's own
pattern by setting the 'name_scheme_string' to a format() parsable
string.

This also sets the Red Hat policy to use the new 'friendly' pattern.

Resolves: #469

---

### DIFF
--- a/sos/policies/__init__.py
+++ b/sos/policies/__init__.py
@@ -6,6 +6,8 @@ import platform
 import time
 import fnmatch
 import tempfile
+import random
+import string
 from os import environ
 
 from sos.utilities import (ImporterHelper,
@@ -195,6 +197,7 @@ No changes will be made to system configuration.
     vendor_text = ""
     PATH = ""
     default_scl_prefix = ""
+    name_pattern = 'legacy'
 
     _in_container = False
     _host_sysroot = '/'
@@ -206,7 +209,6 @@ No changes will be made to system configuration.
         tests to construct PATH must call self.set_exec_path() after
         modifying PATH in their own initializer."""
         self._parse_uname()
-        self.report_name = self.hostname
         self.case_id = None
         self.package_manager = PackageManager()
         self._valid_subclasses = []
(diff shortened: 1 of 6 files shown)


sos/policies/redhat.py
```diff
@@ -45,6 +45,7 @@ class RedHatPolicy(LinuxPolicy):
     _in_container = False
     _host_sysroot = '/'
     default_scl_prefix = '/opt/rh'
+    name_scheme = 'friendly'
```
pmoravec: Is there a reason to have non-default `name_scheme` for one specific policy? I understand your intention, but I dont understand what is specific on the scheme for different policies.

(maybe have this as a user-configured option? or rather directly have sosreport option --report-namescheme="-{name}-{case}-%Y-%m-%d-{rand}" that will be appended after "`sosreport`" string?)

sos/policies/__init__.py
```diff
@@ -261,11 +265,40 @@ def get_archive_name(self):
...
+        rand = ''.join(random.choice(string.lowercase) for x in range(7))
+
+        nstr = "sosreport{name}{case}"
+
+        if self.name_scheme == 'legacy':
+            case = ('.' + case if case else '')
+            nstr += '-%Y%m%d%H%M%S'
+        if self.name_scheme == 'friendly':
+            case = ('-' + case if case else '')
+            nstr += '-%Y-%m-%d-{rand}'
```
pmoravec: From user experience, I prefer having also time (not only date) in the final report name. It is better for identifying and sorting reports from the same machine generated several times during the same day (well this rather happens during my testing than during a support experience).

But that is my personal point of view only.

pmoravec:
In general, I like it. Some comments in the commit, IMHO the sosreport (cmdline) option would be nice here.

sos/policies/redhat.py
TurboTurtle: I am wanting to change the RH naming scheme because that's where I have a pulse with people using sosreports on a daily basis. I can't say the same for any other OSes, so I left those alone.

I think this should be policy driven because it's unlikely (in my opinion) that a user would actively change the pattern via individual runs of sos because that would be a bit clunky.

sos/policies/__init__.py
TurboTurtle: From #469 I dropped the time from the sosreport in favor of the random 7-character string as I believe it is easier to read at a glance than a timestamp. Even a further formatted date string is a bit harsh on the eyes in my opinion. The 7-character string can be used as both a global identifier when dealing with a mass of sosreports, and an identifier for multiple sosreports from the same day.

pmoravec: Both makes sense; "harsh on the eyes" can be minimized using `-%Y-%m-%d_%H-%M-%S` format, but I understand it is subjective.

iranzo: I prefer as pmoravec++ suggested, but I'll also suggest to add hostname in the file, for cases reported against OSP, where you can get 3 controllers + several computes on the same case it will make a lot easier to differentiate them

pmoravec: hostname definitely should be part of the name (how I could forget it?). There are numerous situations where one needs to work with sosreports from several systems inter-communicating and having a problem, and we should be able to distinguish among the reports easily.

TurboTurtle:
Made some changes:

- Simplified the toggle between 'legacy', 'friendly' and 'custom' name patterns
- Removed the user prompt for a name and always set the short hostname now
- Added a `--label` option to allow a user-provided string to be appended after the hostname (and `--name` maps over to this now)

Issue #408, opened 2014-09-24
"date" output should (also) be in UTC?


portante:
Seems like the default "date" command output will (much of the time) include an abbreviated timezone string, but that string is a bit ambiguous for some time zones (see http://stackoverflow.com/questions/1703546/parsing-date-time-string-with-timezone-abbreviated-name-in-python/4766400#4766400).

If we added `date --utc` output as well, consumers of reports could more easily reason about the time a sosreport was collected compared to other timestamps.

For example, what timezone does "IST" represent? While we have /etc/localtime information for some systems, it not be always be available.

TurboTurtle:
Closing per 971b958

Issue #314, opened 2014-07-01
[general] logs & data collection review


bmr-cymru:
General is a bit of a catch-all. Clean it up and remove archaea like:

```
[plugin:general] added copyspec '/var/log/pm/suspend.log'
[plugin:general] added copyspec '/var/log/up2date'
```

TurboTurtle:
Closing per 971b958

PR #1177, opened 2017-12-27
[general|host|date] Break up general into other plugins

TurboTurtle:
Removes the old archaic collection points noted in #314 (the last RHEL distro to use `up2date` was RHEL 4 as far as I know) and adds `date --utc` as requested in #408.

Additionally, merge the `system` and `lsbrelease` plugins into general. The `system` plugin only served to capture `/proc/sys/` and the `lsbrelease` plugin only served to run `lsb_release` and had a duplicate call to copy `/etc/lsb-release` - in either case these seem very "general" to me since they aren't distro or (realistically) component specific. 

---


bmr-cymru:
Nack to the merging: this is going in the wrong direction. The `general` plugin is an ancient relic that's always been a mish-mash of unrelated bits - we've made various efforts to split things out of it and to improve granularity and cohesiveness of the plugin set.

What do you see as the benefit to users or analysts of having these in a single plugin?

TurboTurtle:
When I was looking at 314, the thought crossed my mind about what the functional difference between "general" and "core" system information actually was in the context of sos - semantic to be sure but that's where it came from.  

Part of my reasoning for merging system was that it seemed a bit odd to have a plugin whose only purpose is to collect `/proc/sys`, while another (general) collects `/proc/stat`, with neither of these being the process plugin, but still having the general plugin collect system and component independent information (which both of those are). For lsbrelease, general already captured `/etc/lsb-release` (at least for Debian systems), and moving that to a collection of all things `/etc/*-release` made the call in lsbrelease redundant. For the output of the `lsb_release` it again just seemed "generic" on it's own.

I did see about moving stuff _out_ of general and into other plugins (for example, both `/proc/` collections into process), but I found it harder to find a new home for the other items general collects. 

In the end I didn't see this so much as about end users or analysts, but moreso about consolidating similar items for easier upkeep.

I'm happy to drop the merges from the PR if you want, or take another stab at moving stuff out of general if that's what you meant by going in the wrong direction.

TurboTurtle:
@bmr-cymru per our IRC chat, I've updated the PR to break the general plugin out into others. Moved some bits into `process` and `system`, while creating the new `date` plugin to handle non-ntp/chrony time bits (couldn't call it time due to a conflict with the standard library) and the new `host` plugin to handle host identification.

bmr-cymru:
:+1:

PR #1178, opened 2017-12-29
[openstack_cinder] check for api service running via cinder_wsgi

stuggi:
With OSP11 cinder api changed to run via https wsgi. To check for
running cinder-manage command we also need to take this situation.
The change checks for cinder_wsgi process.

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -1026,6 +1026,17 @@ class Plugin(object):
         else:
             return html
 
+    def check_process_by_name(self, pr):
+        """Checks if a named process is listed in ps -ef output."""
+        ps = self.get_command_output("ps -ef")
+        status = False
+        if ps['status'] == 0:
+            for line in ps['output'].splitlines():
+                if pr in line:
+                    status = True
+                    break
+        return status
+
 
 class RedHatPlugin(object):
     """Tagging class for Red Hat's Linux distributions"""
--- a/sos/plugins/openstack_cinder.py
+++ b/sos/plugins/openstack_cinder.py
@@ -31,26 +31,29 @@ class OpenStackCinder(Plugin):
 
     def setup(self):
 
-        # collect commands output only if the openstack-cinder-api service
-        # is running
-        service_status = self.get_command_output(
-            "systemctl status openstack-cinder-api.service"
-        )
+        # check if either standalone (cinder-api) or httpd wsgi (cinder_wsgi)
+        # is up and running
+        cinder_process = ["cinder_wsgi", "cinder-api"]
+        in_ps = False
+        for process in cinder_process:
+            in_ps = self.check_process_by_name(process)
+            if in_ps:
+                break
 
         container_status = self.get_command_output("docker ps")
         in_container = False
+        cinder_config = ""
         if container_status['status'] == 0:
             for line in container_status['output'].splitlines():
                 if line.endswith("cinder_api"):
                     in_container = True
+                    # if containerized we need to pass the config to the cont.
+                    cinder_config = "--config-dir " + self.var_puppet_gen + \
+                                    "/etc/cinder/"
+                    break
 
-        if (service_status['status'] == 0) or in_container:
-            cinder_config = ""
-            # if containerized we need to pass the config to the cont.
-            if in_container:
-                cinder_config = "--config-dir " + self.var_puppet_gen + \
-                                "/etc/cinder/"
-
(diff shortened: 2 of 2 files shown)


sos/plugins/openstack_cinder.py
```diff
@@ -37,14 +37,22 @@ def setup(self):
...
+                if "cinder_wsgi" in line:
+                    in_ps = True
+                    break
+
         container_status = self.get_command_output("docker ps")
         in_container = False
         if container_status['status'] == 0:
             for line in container_status['output'].splitlines():
                 if line.endswith("cinder_api"):
                     in_container = True
```
pmoravec: Patch sounds OK, maybe trivial enhancement outside of its scope would be calling "break" here (like on line 46)?

stuggi: yes makes sense. I submitted a change with the additional "break"

sos/plugins/openstack_cinder.py
```diff
@@ -37,14 +37,23 @@ def setup(self):
...
+                    break
 
-        if (service_status['status'] == 0) or in_container:
+        if (service_status['status'] == 0) or in_container or in_ps:
```
bmr-cymru: Do we really need three separate conditions to enable this? And why is there a test for `in_container` inside the block? This does not seem all that readable or optimal right now.

Why not place the set up of `cinder_config` outside this block, simplifying it and placing the `cinder-manage` command collection closer to the condition that enables it?

I suspect that would also allow the condition to be simplified.

sos/plugins/openstack_cinder.py
```diff
@@ -37,14 +37,23 @@ def setup(self):
             "systemctl status openstack-cinder-api.service"
         )
 
+        ps_status = self.get_command_output("ps -ef")
```
bmr-cymru: I'm not really keen on this in plugins. The model we talked about some time ago (but which still is not implemented) is for a set of "core" plugins to collect this type of information and other plugins can then use it in their work. That isn't ready today, and is a very considerable piece of work, but I would prefer it if we thought about how we want things to look now, and carried out changes like this in a way that aligns to that future direction.

What I'd like  to see is a new `Plugin` method that calls `ps` and perform the search. Then if/when we ever move to the rich plugin model that method simply becomes a way to access the ps data cached by the process plugin (or whatever).

Cramming it into individual plugins invites proliferation and makes it more likely the next plugin author to need something like this will invent their own wheel again...

sos/plugins/openstack_cinder.py
stuggi: I reviewed that part and removed one condition by checking the ps for either cinder-api or cinder_wsgi. with this we can remove one of the conditions. 

Also moved the cinder_config part in when we check for running cinder in a container env.

sos/plugins/openstack_cinder.py
stuggi: I have introduced a new method check_process_by_name in my last update within this PR. Let me know if I should do it in a separate PR as it is a general enhancement or if it is ok within this PR.

sos/plugins/__init__.py
```diff
@@ -1026,6 +1026,17 @@ def report(self):
         else:
             return html
 
+    def check_process_by_name(self, pr):
```
bmr-cymru: This adds an entire new interface to the `Plugin` class: it needs to be in a separate commit with an adequate explanation in the commit message.

sos/plugins/__init__.py
```diff
@@ -1026,6 +1026,17 @@ def report(self):
...
+        """Checks if a named process is listed in ps -ef output."""
```
bmr-cymru: That's not actually what this method does: it returns `True` if the given `pr` string is a *substring* of any running process's `ps` "full format" listing.. i.e. command name, arguments, etc. That's probably **not** what is wanted for this interface.

I'm dubious that the `ps` command is even the best choice of tool here. We could walk the process tree ourselves (not that costly when just checking names), or use a more focused tool like `pgrep`.

sos/plugins/__init__.py
bmr-cymru: The interface itself is fine, although I would suggest using something like `process`, `name`, or `command` instead of `pr` - there's only a single method argument here after self, so we're not getting short on space and API argument names should be as descriptive as possible where space allows.

stuggi: submitted https://github.com/sosreport/sos/pull/1186 for the new method

Issue #763, opened 2016-02-10
[plugins] make Plugin.policy an object instead of a function


bmr-cymru:
Having to call a function to retrieve the active policy in plugins is clumsy, awkward, and unnecessary. It also seems to put people off from using the functionality it offers since you have to go through ugly machinations like:

```python
    pkg = self.policy().package_manager.all_pkgs()[pkgname]
    if int(pkg['version']) > 1:
        opt = "-T"
    else
        opt = "-W"
```

The call is just pulling a value out of the commons and returning it:

```python
    def policy(self):
        return self.commons["policy"]
```

So we can just set `self.policy` in the constructor and be done with it.

The bigger task is to fix up the (few) plugins that use this:

```
 $ grep 'policy()' sos/plugins/*.py
sos/plugins/__init__.py:        return self.policy().pkg_by_name(package_name) is not None
sos/plugins/iprconfig.py:        arch = self.policy().get_arch()
sos/plugins/kernel.py:            "/lib/modules/%s/modules.dep" % self.policy().kernel_version(),
sos/plugins/nfsserver.py:        default_runlevel = self.policy().default_runlevel()
sos/plugins/nfsserver.py:        nfs_runlevels = self.policy().runlevel_by_service("nfs")
sos/plugins/powerpc.py:        return "ppc64" in self.policy().get_arch()
sos/plugins/processor.py:        if '86' in self.policy().get_arch():
sos/plugins/rpm.py:                    self.policy().package_manager.all_pkgs_by_name_regex
sos/plugins/s390.py:        return (self.policy().get_arch() == "s390")
sos/plugins/sunrpc.py:        if self.policy().default_runlevel() in \
sos/plugins/sunrpc.py:                self.policy().runlevel_by_service(self.service):
```

Amitgb14:
Hi @bmr-cymru, Are you working on this issue?

bmr-cymru:
Yes

PR #1179, opened 2018-01-03
[Plugin] Make Plugin.policy an object instead of a function

TurboTurtle:
Moves Plugin.policy to be an object for plugins instead of a function,
making it easier to leverage the active policy within sos plugins.

Also updates existing plugins that make use of `self.policy`.

Resolves: #763 

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -142,6 +142,7 @@ class Plugin(object):
         self.copy_strings = []
         self.collect_cmds = []
         self.sysroot = commons['sysroot']
+        self.policy = commons['policy']
 
         self.soslog = self.commons['soslog'] if 'soslog' in self.commons \
             else logging.getLogger('sos')
@@ -176,9 +177,6 @@ class Plugin(object):
     def _log_debug(self, msg):
         self.soslog.debug(self._format_msg(msg))
 
-    def policy(self):
-        return self.commons["policy"]
-
     def join_sysroot(self, path):
         if path[0] == os.sep:
             path = path[1:]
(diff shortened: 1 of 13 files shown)


TurboTurtle:
If I'm reading the Travis log et al correctly, it looks like the tests that are failing are ones that pass in their own `commons` to the `MockPlugin`, thus avoiding `sos.policies.load()`.

This should be correctable by:

```
@@ -32,7 +32,7 @@ import errno
 # PYCOMPAT
 import six
 from six.moves import zip, filter
-
+from sos.policies import load
 
 def _to_u(s):
     if not isinstance(s, six.text_type):
@@ -142,7 +142,7 @@ class Plugin(object):
         self.copy_strings = []
         self.collect_cmds = []
         self.sysroot = commons['sysroot']
-        self.policy = commons['policy']
+        self.policy = commons['policy'] if 'policy' in commons else load()
 
         self.soslog = self.commons['soslog'] if 'soslog' in self.commons \
             else logging.getLogger('sos')

```

But I'm not sure if this is the "correct" way to handle this for the tests (or if `load()` should include a `sysroot` since we'd be loading the `GenericPolicy` at that point).

bmr-cymru:
Having a `sysroot` shouldn't matter since it's always defaulted to `None` anyway.

I'm not sure why there's a need to `load()` though? We have never used "real" policies in these tests before, so I'd expect us to just update the mocking for the new style.

Will take a look now.

bmr-cymru:
This is the kind of pattern to use:

```diff
diff --git a/tests/option_tests.py b/tests/option_tests.py
index e8a26e2d..c2a3e504 100644
--- a/tests/option_tests.py
+++ b/tests/option_tests.py
@@ -3,12 +3,14 @@
 import unittest
 
 from sos.plugins import Plugin
+from sos.policies import LinuxPolicy
 
 class GlobalOptionTest(unittest.TestCase):
 
     def setUp(self):
         self.commons = {
             'sysroot': '/',
+            'policy': LinuxPolicy(),
             'global_plugin_options': {
                 'test_option': 'foobar',
                 'baz': None,
```

TurboTurtle:
Not sure what's going on with TravisCI. It's marked here as pending, on Travis the builds are marked as failed but the logs show they passed.

bmr-cymru:
Seems like it's been having some comms issues lately. I've resubmitted that job.

PR #1180, opened 2018-01-07
[nis] enable plugin also when package or config file present

pmoravec:
Enable nis plugin also when ypserv package is installed or
if /etc/ypserv.conf exists.

Resolves: #1180

---

### DIFF
--- a/sos/plugins/nis.py
+++ b/sos/plugins/nis.py
@@ -22,7 +22,9 @@ class Nis(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
     plugin_name = 'nis'
     profiles = ('identity', 'services')
 
-    files = ('/var/yp',)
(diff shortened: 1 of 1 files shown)


sos/plugins/nis.py
```diff
@@ -22,7 +22,9 @@ class Nis(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
     plugin_name = 'nis'
     profiles = ('identity', 'services')
 
-    files = ('/var/yp',)
+    files = ('/var/yp', '/etc/ypserv.conf')
+
+    packages = 'ypserv'
```
bmr-cymru: The `packages` member must be an iterable type...

I never bothered to make this change since every NIS pacakge I've ever seen has owned `/var/yp`. Since other distros could, potentially, not do that I don't really have an objection to it but the patch needs to be respun to fix the type.

PR #1188, opened 2018-01-15
[buildah] Add new plugin

TurboTurtle:
Adds a plugin for Buildah, a new container and image builder

Complete with the Bostonian accent we talked about @bmr-cymru 

---


TurboTurtle:
Realized I forgot the vim tags. Added, re-pushed.

Issue #1189, opened 2018-01-16
[bug] "kubernetes.podlogs=on" can't get logs of pods


alicealvy:
# sosreport -k kubernetes.podlogs=on

the result  can't get logs of pods, please look the link  https://github.com/sosreport/sos/blob/master/sos/plugins/kubernetes.py#L116-L122
something is wrong.

pmoravec:
That code is evidently wrongly block-aligned:

- it is executed under `if self.get_option('all'):` only (is this intentional?)
- it refers to variable `knsp` that is set inside a loop `for n in knsps:` while this wrong code is executed *outside* that loop. So this code gets assigned `knsp` from the latest loop execution (or an exception is raised if loop isnt executed at all)

I dont know when "podlogs" should be collected, under what scope. @TurboTurtle , can you have a look, please?

TurboTurtle:
Ah, been a while but I remember what's going on here. The `kubectl logs` command only works for the current namespace that `kubectl` is executing in unless it is explicitly set like `kubectl --namespace=test logs mypod`, so we have no way to get pod logs without the namespaces explicitly set.

So realistically what we should be doing here is re-arranging the block so that we get the list of namespaces regardless, and then separately check for the `all` or `podlogs` options.

I can work something out shortly for this.

TurboTurtle:
There was another problem with pod logs collection I found, also resolved with the above PR.

PR #1190, opened 2018-01-16
[kubernetes] Fix collection of pod logs

TurboTurtle:
Corrects a few problems with the collection of pod logs via kubectl.
First, no longer require the 'all' plugin option, and make pod
collection a check of its own. Second, fix a malformed kubectl command
to actually collect the logs from a given pod.

Additionally, no longer collect duplicate --all-namespaces output if we
collect this output separately for each namespace.

Resolves: #1189

---


pmoravec:
+1 makes sense to me.

PR #1191, opened 2018-01-17
[rabbitmq] Log collection fixes when rabbitmq runs in a container

mbaldessari:
When rabbitmq is running inside a container all the rabbitmqctl
commands will fail with some obscure (welsh?) error messages like:

$ rabbitmqctl cluster_status
erno" ienrirto rt elromgigneart ipnrge sienn td)o _booetr"r,o{r: b<a0d.ar2g.,0[>{
l_prim_loader,check_file_result,3,[]},{init,get_boot,1,[]},{init,get_boot,2,[]},{init,do_boot,3,[]}]}}
init terminating in do_boot ()

When run inside containers we need to run any rabbitmqctl command inside
the container, so that output is properly collected:
$ docker exec -t rabbitmq-bundle-docker-0 rabbitmqctl cluster_status
Cluster status of node 'rabbit@controller-0' ...
[{nodes,[{disc,['rabbit@controller-0','rabbit@controller-1',
                'rabbit@controller-2']}]},
 {running_nodes,['rabbit@controller-2','rabbit@controller-1',
                 'rabbit@controller-0']},
 {cluster_name,<<"rabbit@controller-0.localdomain">>},
 {partitions,[]},
 {alarms,[{'rabbit@controller-2',[]},
          {'rabbit@controller-1',[]},
          {'rabbit@controller-0',[]}]}]

While we're at it we also collect the logs of any rabbitmq container.
This is particularly useful in containerized openstack deployments
where short-lived containers are used to set up the cloud.

Since the info output by 'rabbitmqctl cluster_status' and 'rabbitmqctl
cluster_status' is already contained in 'rabbitmqctl report', we just
remove the former two commands.

First reported via RHBZ#1525620

Signed-off-by: Michele Baldessari <michele@acksyn.org>
Signed-off-by: John Eckersberg <jeckersb@redhat.com>

### DIFF
--- a/sos/plugins/rabbitmq.py
+++ b/sos/plugins/rabbitmq.py
@@ -28,9 +28,26 @@ class RabbitMQ(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
     packages = ('rabbitmq-server',)
 
     def setup(self):
-        self.add_cmd_output("rabbitmqctl report")
-        self.add_cmd_output("rabbitmqctl cluster_status")
-        self.add_cmd_output("rabbitmqctl list_policies")
+        container_status = self.get_command_output(
+            "docker ps -a --format='{{ .Names }}'")
+
+        in_container = False
+        container_names = []
+        if container_status['status'] == 0:
+            for line in container_status['output'].splitlines():
+                if line.startswith("rabbitmq"):
+                    in_container = True
+                    container_names.append(line)
+
+        if in_container:
+            for container in container_names:
+                self.add_cmd_output('docker logs {0}'.format(container))
+                self.add_cmd_output(
+                    'docker exec -t {0} rabbitmqctl report'
+                    .format(container)
+                )
+        else:
+            self.add_cmd_output("rabbitmqctl report")
 
         self.add_copy_spec([
             "/etc/rabbitmq/*",


mbaldessari:
Meh pushed an older version by mistake. Updating

bmr-cymru:
```
erno" ienrirto rt elromgigneart ipnrge sienn td)o _booetr"r,o{r: b<a0d.ar2g.
```

As a Welshman, I can assure you that that is most definitely not Welsh: Ych a fi!

jeckersb:
👍

sos/plugins/rabbitmq.py
```diff
@@ -28,9 +28,25 @@ class RabbitMQ(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
...
+                if line.startswith("rabbitmq"):
+                    in_container = True
+                    container_names.append(line)
+
+        if in_container:
+            for container in container_names:
+                self.add_cmd_output('docker logs {0}'.format(container))
+                self.add_cmd_output(
+                    'docker exec -t {0} /bin/sh -c "rabbitmqctl report"'
+                    .format(container))
```
bmr-cymru: The closing paren of the `self.add_cmd_output()` call should be on the following line, indented to the same level as the call.

sos/plugins/rabbitmq.py
```diff
@@ -28,9 +28,25 @@ class RabbitMQ(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
...
+            for line in container_status['output'].splitlines():
+                if line.startswith("rabbitmq"):
+                    in_container = True
+                    container_names.append(line)
+
+        if in_container:
+            for container in container_names:
+                self.add_cmd_output('docker logs {0}'.format(container))
+                self.add_cmd_output(
+                    'docker exec -t {0} /bin/sh -c "rabbitmqctl report"'
```
bmr-cymru: Is a shell really needed here? Np if it is, just checking.

mbaldessari: Nope not strictly needed. I added it because I thought it could be useful in case commands needed to become more convoluted, but that's not the case. Will remove

sos/plugins/rabbitmq.py
mbaldessari: Ack, thanks

mbaldessari:
Ok tested the last version on both containerized and non-containerized clouds and collection seems correct

bmr-cymru:
Thanks! We can always add the shell back in later if we need it - we just try to avoid them unless required.

I'll get this merged in the next batch.

mbaldessari:
Diolch ;)

bmr-cymru:
Croeso!

bmr-cymru:
Merged in commit 2780f6d6ec3a4db72ed5a00aea15ac750394314c.

PR #1205, opened 2018-02-05
[vdsm] add vdsm plugin

igoihman:
Signed-off-by: Irit Goihman <igoihman@redhat.com>

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -1389,6 +1389,22 @@ class Plugin(object):
             return False
         return status
 
+    def get_process_pids(self, process):
+        """Returns PIDs of all processes with process name.
+        If the process doesn't exist, returns an empty list"""
+        pids = []
+        cmd_line_glob = "/proc/[0-9]*/cmdline"
+        cmd_line_paths = glob.glob(cmd_line_glob)
+        for path in cmd_line_paths:
+            try:
+                with open(path, 'r') as f:
+                    cmd_line = f.read().strip()
+                    if process in cmd_line:
+                        pids.append(path.split("/")[2])
+            except IOError as e:
+                continue
+        return pids
+
 
 class RedHatPlugin(object):
     """Tagging class for Red Hat's Linux distributions"""
--- a/sos/plugins/vdsm.py
+++ b/sos/plugins/vdsm.py
@@ -22,6 +22,10 @@ import re
 # there is a stuck lvm command.
 # locking_type=0
 #
+# To prevent modifications to volume group metadata (for e.g. due to a
+# automatically detected inconsistency), metadata_read_only is set to 1.
+# metadata_read_only=1
+#
 # use_lvmetad is set to 0 in order not to show cached, old lvm metadata.
 # use_lvmetad=0
 #
@@ -31,6 +35,7 @@ import re
 LVM_CONFIG = """
 global {
     locking_type=0
+    metadata_read_only=1
     use_lvmetad=0
 }
 devices {
--- a/sos/plugins/vdsm.py
+++ b/sos/plugins/vdsm.py
@@ -65,6 +65,7 @@ class Vdsm(Plugin, RedHatPlugin):
         self.add_forbidden_path('/etc/pki/libvirt/private/*')
 
         self.add_cmd_output('service vdsmd status')
+        self.add_cmd_output('service supervdsmd status')
 
         self.add_copy_spec([
             '/tmp/vds_installer*',
--- a/sos/plugins/vdsm.py
+++ b/sos/plugins/vdsm.py
@@ -45,7 +45,7 @@ LVM_CONFIG = re.sub(r"\s+", " ", LVM_CONFIG).strip()
 
 
 class Vdsm(Plugin, RedHatPlugin):
-    """vdsm Plugin"""
+    """vdsm plugin"""
 
     packages = (
         'vdsm',
--- a/sos/plugins/vdsm.py
+++ b/sos/plugins/vdsm.py
@@ -1,0 +1,146 @@
+# Copyright (C) 2018 Red Hat, Inc.
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.plugins import Plugin, RedHatPlugin
+
+import glob
+import json
+import re
+
+
+# This configuration is based on vdsm.storage.lvm.LVM_CONF_TEMPLATE.
+#
+# locking_type is set to 0 in order to match lvm sos commands. With this
+# configuration we don't take any locks, so we will never block because
+# there is a stuck lvm command.
+# locking_type=0
+#
+# use_lvmetad is set to 0 in order not to show cached, old lvm metadata.
+# use_lvmetad=0
+#
+# preferred_names and filter config values are set to capture Vdsm devices.
+# preferred_names=[ '^/dev/mapper/' ]
+# filter=[ 'a|^/dev/mapper/.*|', 'r|.*|' ]
+LVM_CONFIG = """
+global {
+    locking_type=0
+    use_lvmetad=0
+}
+devices {
+    preferred_names=["^/dev/mapper/"]
+    ignore_suspended_devices=1
+    write_cache_state=0
+    disable_after_error_count=3
+    filter=["a|^/dev/mapper/.*|", "r|.*|"]
+}
+"""
+LVM_CONFIG = re.sub(r"\s+", " ", LVM_CONFIG).strip()
+
+
+class Vdsm(Plugin, RedHatPlugin):
+    """vdsm Plugin"""
+
+    packages = (
+        'vdsm',
+        'vdsm-client',
+    )
+
+    plugin_name = 'vdsm'
+
+    def setup(self):
+        self.add_forbidden_path('/etc/pki/vdsm/keys/*')
+        self.add_forbidden_path('/etc/pki/vdsm/libvirt-spice/*-key.*')
+        self.add_forbidden_path('/etc/pki/libvirt/private/*')
+
+        self.add_cmd_output('service vdsmd status')
+
+        self.add_copy_spec([
+            '/tmp/vds_installer*',
+            '/tmp/vds_bootstrap*',
+            '/etc/vdsm/*'
+        ])
+
+        limit = self.get_option('log_size')
+
+        self.add_copy_spec('/var/log/vdsm/*', sizelimit=limit)
+
+        self._add_vdsm_forbidden_paths()
+        self.add_copy_spec([
+            '/var/run/vdsm/*',
+            '/usr/libexec/vdsm/hooks',
+            '/var/lib/vdsm'
+        ])
+
+        qemu_pids = self.get_process_pids('qemu-kvm')
+        if qemu_pids:
+            files = ["cmdline", "status", "mountstats"]
+            self.add_copy_spec([
+                "/proc/%s/%s" % (pid, name)
+                for pid in qemu_pids
+                for name in files
+            ])
+        self.add_cmd_output([
+            "ls -ldZ /etc/vdsm",
+            "su vdsm -s sh -c 'tree -l /rhev/data-center'",
+            "su vdsm -s sh -c 'ls -lR /rhev/data-center'"
+        ])
+        self.add_cmd_output([
+            "lvm vgs -v -o +tags --config \'%s\'" % LVM_CONFIG,
+            "lvm lvs -v -o +tags --config \'%s\'" % LVM_CONFIG,
+            "lvm pvs -v -o +all --config \'%s\'" % LVM_CONFIG
+        ])
+
+        self.add_cmd_output([
+            'vdsm-client Host getCapabilities',
+            'vdsm-client Host getStats',
+            'vdsm-client Host getAllVmStats',
+            'vdsm-client Host getVMFullList',
+            'vdsm-client Host getDeviceList',
+            'vdsm-client Host hostdevListByCaps',
+            'vdsm-client Host getAllTasksInfo',
+            'vdsm-client Host getAllTasksStatuses'
+        ])
+
+        try:
+            res = self.call_ext_prog(
+                'vdsm-client Host getConnectedStoragePools'
+            )
+            if res['status'] == 0:
+                pools = json.loads(res['output'])
+                for pool in pools:
+                    self.add_cmd_output(
+                        'vdsm-client StoragePool getSpmStatus'
+                        ' storagepoolID={}'.format(pool)
+                    )
+        except ValueError as e:
+            self._log_error(
+                'vdsm-client Host getConnectedStoragePools: %s' % (e)
+            )
+
+        try:
+            res = self.call_ext_prog('vdsm-client Host getStorageDomains')
+            if res['status'] == 0:
+                sd_uuids = json.loads(res['output'])
+                dump_volume_chains_cmd = 'vdsm-tool dump-volume-chains %s'
+                self.add_cmd_output([
+                    dump_volume_chains_cmd % uuid for uuid in sd_uuids
+                ])
+        except ValueError as e:
+            self._log_error(
+                'vdsm-client Host getStorageDomains: %s' % (e)
+            )
+
+    def _add_vdsm_forbidden_paths(self):
+        """Add confidential sysprep vfds under /var/run/vdsm to
+         forbidden paths """
+
+        for file_path in glob.glob("/var/run/vdsm/*"):
+            if file_path.endswith(('.vfd', '/isoUploader', '/storage')):
+                self.add_forbidden_path(file_path)
--- a/sos/plugins/vdsm.py
+++ b/sos/plugins/vdsm.py
@@ -72,9 +72,7 @@ class Vdsm(Plugin, RedHatPlugin):
             '/etc/vdsm/*'
         ])
 
-        limit = self.get_option('log_size')
-
-        self.add_copy_spec('/var/log/vdsm/*', sizelimit=limit)
+        self.add_copy_spec('/var/log/vdsm/*')
 
         self._add_vdsm_forbidden_paths()
         self.add_copy_spec([
--- a/sos/plugins/vdsm.py
+++ b/sos/plugins/vdsm.py
@@ -43,7 +43,7 @@ devices {
     ignore_suspended_devices=1
     write_cache_state=0
     disable_after_error_count=3
-    filter=["a|^/dev/mapper/.*|", "r|.*|"]
+    filter=["a|^/dev/disk/by-id/dm-uuid-mpath-|", "r|.+|"]
 }
 """
 LVM_CONFIG = re.sub(r"\s+", " ", LVM_CONFIG).strip()


bronhaim:
Hi, @bmr-cymru Can you please take a look ? we want to include this plugin in next sos release

Thanks

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,153 @@
...
+        if logsize is not None:
+            self.add_copy_spec('/var/log/vdsm/*', logsize)
+        else:
+            self.add_copy_spec('/var/log/vdsm/*')
+
+        self._add_vdsm_run_dir()
+        self.add_copy_spec('/usr/libexec/vdsm/hooks')
+        self.add_copy_spec('/var/lib/vdsm')
+        self.add_copy_spec('/var/log/ovirt.log')
+        self.add_copy_spec('/var/log/sanlock.log')
```
TurboTurtle: We can collapse all these `add_copy_spec()` calls into a single call that takes a list of the files.

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,153 @@
...
+        self.add_cmd_output("/usr/bin/abrt-cli list")
+
+        self.add_cmd_output('vdsm-client Host getCapabilities')
+        self.add_cmd_output('vdsm-client Host getStats')
+        self.add_cmd_output('vdsm-client Host getAllVmStats')
+        self.add_cmd_output('vdsm-client Host getVMList')
+        self.add_cmd_output('vdsm-client Host getDeviceList')
+        self.add_cmd_output('vdsm-client Host hostdevListByCaps')
+        self.add_cmd_output('vdsm-client Host getAllTasksInfo')
+        self.add_cmd_output('vdsm-client Host getAllTasksStatuses')
```
TurboTurtle: Same here, or we could have the list contain subcmds and iterate through those, e.g.:

```python
subcmds= [
    'getStats',
    [...]
]

for subcmd in subcmds:
    self.add_cmd_output('vdsm-client Host %s' % subcmd)
```

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,153 @@
...
+    """vdsm Plugin"""
+
+    packages = (
+        'vdsm',
+        'vdsm-client',
+    )
+
+    plugin_name = 'vdsm'
+
+    optionList = [("logsize", 'max size (MiB) to collect per log file', '',
```
TurboTurtle: Is this expected to be set separately/differently from the global `--log-size` parameter?

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,153 @@
...
+            "/sbin/lvm lvs -v -o +tags --config \'%s\'" % LVM_CONFIG)
+        self.add_cmd_output(
+            "/sbin/lvm pvs -v -o +all --config \'%s\'" % LVM_CONFIG)
+
+        self.add_cmd_output("/sbin/fdisk -l")
+        self.add_cmd_output("/usr/bin/iostat")
+        self.add_cmd_output("/sbin/iscsiadm -m node")
+        self.add_cmd_output("/sbin/iscsiadm -m session")
+        self.add_cmd_output("/usr/sbin/nodectl info")
+        self.add_cmd_output("/usr/bin/abrt-cli list")
```
TurboTurtle: Same here, `add_cmd_output()` can take a list instead of making multiple calls.

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,153 @@
...
+                self.add_copy_spec("/proc/%s/cmdline" % pid)
+                self.add_copy_spec("/proc/%s/status" % pid)
+                self.add_copy_spec("/proc/%s/mountstats" % pid)
+        self.add_cmd_output("/bin/ls -ldZ /etc/vdsm")
+        self.add_cmd_output(
+            "/bin/su vdsm -s /bin/sh -c '/usr/bin/tree -l /rhev/data-center'")
+        self.add_cmd_output(
+            "/bin/su vdsm -s /bin/sh -c '/bin/ls -lR /rhev/data-center'")
+        self.add_cmd_output(
+            "/sbin/lvm vgs -v -o +tags --config \'%s\'" % LVM_CONFIG)
```
TurboTurtle: For all the LVM commands that use `LVM_CONFIG`, I think we would want to use the `suggest_filename` parameter to not have quite so harsh of a file name in the archive. E.G. just having `sbin_lvm_vgs_v_o_tags`.

igoihman:
@bmr-cymru  please review latest changes according to your requests. Thanks!

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,153 @@
...
+        self.add_copy_spec(['/tmp/vds_installer*',
+                            '/tmp/vds_bootstrap*',
+                            '/etc/vdsm/*'])
+
+        limit = self.get_option('log_size')
+
+        if limit is not None:
+            self.add_copy_spec('/var/log/vdsm/*', sizelimit=limit)
+        else:
+            self.add_copy_spec('/var/log/vdsm/*')
```
pmoravec: These four lines above have the same effect as:

`self.add_copy_spec('/var/log/vdsm/*', sizelimit=limit)`

since `add_copy_spec` method has default `sizelimit=None`. See https://github.com/sosreport/sos/blob/master/sos/plugins/__init__.py#L524

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,153 @@
...
+                    'vdsm-tool dump-volume-chains %s' % sd_uuid)
+
+    def _add_vdsm_run_dir(self):
+        """Add everything under /var/run/vdsm except possibly confidential
+        sysprep vfds and sockets"""
+
+        import glob
+
+        for f in glob.glob(
+                "/var/run/vdsm/*"):
```
pmoravec: Can't there be potentially millions of files in that dir, where walkthrough like this would take ages? (we had that experience with `jars` plugin..)

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,153 @@
...
+                            suggest_filename="sbin_lvm_lvs_v_o_tags")
+        self.add_cmd_output("/sbin/lvm pvs -v -o +all --config \'%s\'"
+                            % LVM_CONFIG,
+                            suggest_filename="sbin_lvm_pvs_v_o_all")
+        self.add_cmd_output(["/sbin/fdisk -l",
+                             "/usr/bin/iostat",
+                             "/sbin/iscsiadm -m node",
+                             "/sbin/iscsiadm -m session",
+                             "/usr/sbin/nodectl info",
+                             "/usr/bin/abrt-cli list"])
```
pmoravec: The `abrt-cli list` is already collected by `abrt` plugin - is there a reason to collect it also here?

sos/plugins/vdsm.py
bronhaim: vdsm creates files under /var/run/vdsm and we don't maintain temporary cache or links which can lead to millions of files. the amount of files is more or less constant

sos/plugins/vdsm.py
bronhaim: can we add mandatory requirement for sos abrt plugin with this plugin? if yes I will omit it from the output

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,153 @@
...
+        if limit is not None:
+            self.add_copy_spec('/var/log/vdsm/*', sizelimit=limit)
+        else:
+            self.add_copy_spec('/var/log/vdsm/*')
+
+        self._add_vdsm_run_dir()
+        self.add_copy_spec(['/usr/libexec/vdsm/hooks',
+                            '/var/lib/vdsm',
+                            '/var/log/ovirt.log',
+                            '/var/log/sanlock.log'])
```
bmr-cymru: Nack: this is already collected by the sanlock plugin.

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,153 @@
...
+
+        if limit is not None:
+            self.add_copy_spec('/var/log/vdsm/*', sizelimit=limit)
+        else:
+            self.add_copy_spec('/var/log/vdsm/*')
+
+        self._add_vdsm_run_dir()
+        self.add_copy_spec(['/usr/libexec/vdsm/hooks',
+                            '/var/lib/vdsm',
+                            '/var/log/ovirt.log',
```
bmr-cymru: What's in this file? We don't collect a file of this name in the `ovirt` plugin, so this looks a little strange here.

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,153 @@
...
+                            '/var/log/sanlock.log'])
+
+        res = self.call_ext_prog('/usr/bin/pgrep qemu-kvm')
```
bmr-cymru: We have a method to test whether a process is running. Adding one to return also the PID(s) should be straightforward and is preferable to shelling out to `pgrep`.

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,153 @@
...
+        os.environ["LVM_SYSTEM_DIR"] = "/var/run/vdsm/lvm"
+        self.add_forbidden_path('/etc/pki/vdsm/keys/*')
+        self.add_forbidden_path('/etc/pki/vdsm/libvirt-spice/*-key.*')
+        self.add_forbidden_path('/etc/pki/libvirt/private/*')
+
+        self.add_cmd_output('service vdsmd status')
+
+        self.add_copy_spec(['/tmp/vds_installer*',
+                            '/tmp/vds_bootstrap*',
+                            '/etc/vdsm/*'])
```
bmr-cymru: All of these multi-line list-in-method-args calls need to use the sos house style:

```python
   self.some_method([
        "foo",
        "bar
    ])
```

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,153 @@
...
+            for line in res['output']:
+                pid = line.strip()
+                self.add_copy_spec(["/proc/%s/cmdline" % pid,
+                                    "/proc/%s/status" % pid,
+                                    "/proc/%s/mountstats" % pid])
+        self.add_cmd_output(["/bin/ls -ldZ /etc/vdsm",
+                            "/bin/su vdsm -s /bin/sh -c\
+                             '/usr/bin/tree -l /rhev/data-center'",
+                             "/bin/su vdsm -s /bin/sh -c\
+                              '/bin/ls -lR /rhev/data-center'"])
```
bmr-cymru: Why do any of these commands need to use a shell?

bmr-cymru: also, please do not hard code paths in `sos` plugins unless it is absolutely necessary (i.e. a binary that is *never* expected to be in PATH, or in a standard `bin/` directory). We use a policy mechanism to control `PATH` to account for distribution variations and hard coding these defeats that.

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,153 @@
...
+        self.add_cmd_output("/sbin/lvm vgs -v -o +tags --config \'%s\'"
+                            % LVM_CONFIG,
+                            suggest_filename="sbin_lvm_vgs_v_o_tags")
+        self.add_cmd_output("/sbin/lvm lvs -v -o +tags --config \'%s\'"
+                            % LVM_CONFIG,
+                            suggest_filename="sbin_lvm_lvs_v_o_tags")
+        self.add_cmd_output("/sbin/lvm pvs -v -o +all --config \'%s\'"
+                            % LVM_CONFIG,
+                            suggest_filename="sbin_lvm_pvs_v_o_all")
```
bmr-cymru: I would be a lot happier if we could find a way to include this in the standard `lvm` plugin. Also, please do not use `suggest_filename` unless absolutely necessary: it is intended only to allow overriding an automatically generated file name when it would either be excessively long for readability purposes, or would exceed a path length limit and so lose uniqueness.

sos/plugins/vdsm.py
bmr-cymru: Nack to all of this. These are collected in other plugins and this is not the approach we use in sos: each plugin gathers only the information relevant to the component it supports. This is the only way to make things manageable with hundreds of plugins and not end up in a mess of duplicate and confusing data placement.

We can possibly make an exception for the LVM data for `vdsm`, since it uses it in a very particular way, and we may not be able to immediately find a clean way to integrate this into the `lvm` plugin, but for these generic commands it's a very strong nack.

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,153 @@
...
+                             "/usr/bin/abrt-cli list"])
+
+        self.add_cmd_output(['vdsm-client Host getCapabilities',
+                             'vdsm-client Host getStats',
+                             'vdsm-client Host getAllVmStats',
+                             'vdsm-client Host getVMList',
+                             'vdsm-client Host getDeviceList',
+                             'vdsm-client Host hostdevListByCaps',
+                             'vdsm-client Host getAllTasksInfo',
+                             'vdsm-client Host getAllTasksStatuses'])
```
bmr-cymru: Ack (subject to style)

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,153 @@
...
+                self.add_cmd_output(
+                    'vdsm-client StoragePool getSpmStatus'
+                    ' storagepoolID={}'.format(pool))
+
+        res = self.call_ext_prog('vdsm-client Host getStorageDomains')
+        if res['status'] == 0:
+            sd_uuids = ast.literal_eval(res['output'])
+            for sd_uuid in sd_uuids:
+                self.add_cmd_output(
+                    'vdsm-tool dump-volume-chains %s' % sd_uuid)
```
bmr-cymru: Ack (subject to style)

sos/plugins/vdsm.py
bmr-cymru: In that case, why not blacklist the entries that you don't want to collect, and get everything else?

This is sort of duplicating the `add_forbidden_path()` method in an ad-hoc way.

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,153 @@
...
+        """Add everything under /var/run/vdsm except possibly confidential
+        sysprep vfds and sockets"""
+
+        import glob
+
+        for f in glob.glob(
+                "/var/run/vdsm/*"):
+            if not f.endswith('.vfd') and not f.endswith('/isoUploader') \
+                    and not f.endswith('/storage') \
+                    and not stat.S_ISSOCK(os.stat(f).st_mode):
```
bmr-cymru: Sockets are not a problem (they are properly duplicated by the `Archive` classes`), and can in some cases provide useful diagnostic detail (permissions).

bmr-cymru:
Strictly speaking, it was a week ago... :-)

However, we've taken a decision to delay the release for now - we've been trying to move to a calendar based release model for the last few cycles, and that's been going well, however, right now we have  a large backlog of RHEL-relevant PRs, and no other distribution is currently waiting to package a new release.

Given that, we've decided to try to get as many as possible of the remaining requests in given that we have a little more time yet before we start to run into distribution deadlines.

I can help with some of the changes in this plugin, so let me know and we can hopefully work something out.

sos/plugins/vdsm.py
igoihman: The commands need to be executed as user vdsm.

sos/plugins/vdsm.py
igoihman: I was asked in the previous review to use `suggest_filename` because the generated file name is too long.

sos/plugins/vdsm.py
igoihman: can you point me to this method?

sos/plugins/vdsm.py
bmr-cymru: Well, sure, but why do they need an absolute path? They should run in the `PATH` context of the `su`'d user.

bmr-cymru: And why do they need a shell? That has nothing to do with switching user (unless you are relying on profile and rc expansion to establish an environment).

sos/plugins/vdsm.py
igoihman: done

sos/plugins/vdsm.py
bmr-cymru: By whom? I can't find it anywhere in the history for this one - it is something a lot of new contributors do, because they think it is more "friendly" - we reserve its use for *very* long command lines - things that could even blow the filename length limit for example.

This includes RPM commands with many dozens or hundreds of packages on the command line, as well as the `modinfo` calls that run for every kernel module on the system.

The reason it's harmful to overuse it is that it prevents the actual command run from being mechanically recovered from the file name, and it prevents something trying to comprehend the archive from doing the reverse: determining the output file name from the command run.

bmr-cymru: If it was me then I apologise - but I'd be surprised as these command lines are really not that long at all...

sos/plugins/vdsm.py
bmr-cymru: `Plugin.check_process_by_name`

```python
        """Checks if a named process is found in /proc/[0-9]*/cmdline.
        Returns either True or False."""
```

bmr-cymru:
Does VDSM use `LVM_SYSTEM_DIR` at all, or is there a valid `lvm.conf` file on the system that would have the same content as we are injecting via `LVM_CONFIG`?

That's a potential way to deal with some of the command line problems.

sos/plugins/vdsm.py
igoihman: I see it returns only true or false and not the PID. We want to collect some data based on that PID.

bmr-cymru: > We have a method to test whether a process is running.

That is the method above.

> Adding one to return also the PID(s) should be straightforward and is preferable to shelling out to pgrep.

Is the one that isn't written yet, but as I said, would be easy to add since that is what you want.

sos/plugins/vdsm.py
bmr-cymru: Like I said, it is defined by the policy: you'll find that in the `sos.policy` module, specifically for Red Hat distributions in the file `sos/policies/redhat.py`. There are separate policies for RHEL, Fedora and other variants. Each of these may define its own path - this mechanism was introduced for two reasons:

  * to support UsrMove in Fedora and later RHEL
  * to isolate `sosreport` from broken `PATH` settings in the user environment

The general `RedHatPolicy` (parent class of the concrete product/distro classes) defines it like this:

```python
        # handle PATH for UsrMove
        if 'filesystem' not in pkgs:
            print("Could not find 'filesystem' package: "
                  "assuming PATH settings")
            usrmove = True
        else:
            filesys_version = pkgs['filesystem']['version']
            usrmove = True if filesys_version[0] == '3' else False

        if usrmove:
            self.PATH = "/usr/sbin:/usr/bin:/root/bin"
        else:
            self.PATH = "/sbin:/bin:/usr/sbin:/usr/bin:/root/bin"
        self.PATH += os.pathsep + "/usr/local/bin"
        self.PATH += os.pathsep + "/usr/local/sbin"
        self.set_exec_path()
```

You should find that this closely matches the ordering defined by the system. In my case, on Fedora, my system defined `PATH` includes `/usr/local/bin:/usr/local/sbin:/usr/bin:/usr/sbin`.

> when I run it 'sh' command on my computer it tries to run /usr/bin/sh instead of /bin/sh. I haven't made changes to PATH variable.

I don't know what your computer is running so I don't know what's correct, but this behaviour is consistent with recent Fedora and RHEL: `/bin` is just a symlink anyway, so what's the point of searching it?

```
$ ls -ld /bin
lrwxrwxrwx. 1 root root 7 Feb 10  2017 /bin -> usr/bin
```

Currently `sos` does not handle loading paths dynamically from `/etc/profile.d` but it is something we may add in future based on need (it's ugly since the only way to do it correctly is to invoke a shell and then ask it how `PATH` is defined - there's no way we will attempt to parse the shell drop-ins ourselves).

sos/plugins/vdsm.py
igoihman: done, it's a leftover from an old version

sos/plugins/vdsm.py
igoihman: done

sos/plugins/vdsm.py
igoihman: No problem, I will remove it.

sos/plugins/vdsm.py
igoihman: OK, removed

sos/plugins/vdsm.py
igoihman: done, please see the second review

sos/plugins/vdsm.py
igoihman: OK, I guess the machine I used for testing wasn't configured well, I removed all the paths.

sos/plugins/vdsm.py
igoihman: done

igoihman:
@bmr-cymru please review the updated version, thanks!

igoihman:
@bmr-cymru  @TurboTurtle Can you review this pull request?

bmr-cymru:
@igoihman sorry for the delay, had a lot on with release deadline stuff lately. I will get a review added here this week.

igoihman:
@bmr-cymru Thanks!

sos/plugins/__init__.py
```diff
@@ -1123,6 +1123,21 @@ def check_process_by_name(self, process):
...
+        cmd_line_paths = glob.glob(cmd_line_glob)
+        for path in cmd_line_paths:
+            try:
+                with open(path, 'r') as f:
+                    cmd_line = f.read().strip()
+                    if process in cmd_line:
+                        return path.split("/")[2]
+            except IOError as e:
+                continue
+        return -1
```
bmr-cymru: I think it would be better to use `None` as the return type here. In some ways, in Python style, an exception would be preferred (rather than signalling the error with a return value), but I think that is overkill here - it would force us to wrap all those calls in `try`/`except` blocks.

The problem with `-1` is twofold: the test looks less obvious (`if foo != -1` vs. `if not foo` of `if foo is not None`). The second form of the `None` comparison is only needed if we want to distinguish from `0` - I'm inclined to think that's not necessary since `pid 0` is the swapper thread, and we don't really want to do anything with that special value. The other issue with `-1` is that it is a special value recognised by many process tools as well as syscalls: in the signal sending syscalls, as well as the `kill` program, `-1` means "all processes with a PID larger than 1 are signaled". I know that isn't going to cause trouble for your use but it could become relevant in other situations.

sos/plugins/__init__.py
```diff
@@ -1123,6 +1123,21 @@ def check_process_by_name(self, process):
             return False
         return status
 
+    def get_process_pid(self, process):
+        """Returns a process PID if it exists.
+        If th process doesn't exist, returns -1"""
```
bmr-cymru: `s/th/the`

sos/plugins/__init__.py
```diff
@@ -1123,6 +1123,21 @@ def check_process_by_name(self, process):
...
+        cmd_line_glob = "/proc/[0-9]*/cmdline"
+        cmd_line_paths = glob.glob(cmd_line_glob)
+        for path in cmd_line_paths:
```
bmr-cymru: It's worth trying this out on a system with a very large number of threads running. Sometimes we've found external tools are quicker, but it's not a definite call (I had a feeling we had discussed this before, but I can't find that conversation now or what the conclusion was).

Anyway, it's not a big deal - many tools show a slowdown on those sorts of heavily loaded systems anyway.

At some point it may be useful to allow exact/inexact matching (i.e. an exact command line vs. any instance of a program), but it's not needed today so there's no need to implement it.

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,153 @@
...
+    locking_type=0
+    use_lvmetad=0
+}
+devices {
+    preferred_names=[ "^/dev/mapper/" ]
+    ignore_suspended_devices=1
+    write_cache_state=0
+    disable_after_error_count=3
+    filter=[ "a|^/dev/mapper/.*|", "r|.*|" ]
+}
```
bmr-cymru: Need to run this through the `pycodestyle` command - pep8 doesn't like space around `[]` and `()` members:

```
$ echo 'foo = [ "bar" ]' | pycodestyle -
stdin:1:8: E201 whitespace after '['
stdin:1:14: E202 whitespace before ']'
```

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,153 @@
...
+
+    packages = (
+        'vdsm',
+        'vdsm-client',
+    )
+
+    plugin_name = 'vdsm'
+
+    def setup(self):
+        os.environ["LVM_SYSTEM_DIR"] = "/var/run/vdsm/lvm"
```
bmr-cymru: It may be necessary to remove this from `os.environ` at the end of `setup()`: today plugins can run in parallel, and in any order, which means that if the LVM2 plugin ran after `vdsm` it would pick up this environment setting.

Assuming that's not wanted then I think it's the best option (even if VDSM deployments don't use the `lvm2` plugin, we should try not to interfere). If that is the case then we may need to think about a better way to coordinate them.

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,153 @@
...
+
+        self.add_copy_spec([
+            '/tmp/vds_installer*',
+            '/tmp/vds_bootstrap*',
+            '/etc/vdsm/*'
+        ])
+
+        limit = self.get_option('log_size')
+
+        self.add_copy_spec('/var/log/vdsm/*', sizelimit=limit)
```
bmr-cymru: Optional to also support the `all_logs` plugin option here (which would just grab "current" logs if not set). If these are always wanted for VDSM though that's fine.

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,153 @@
...
+        self._add_vdsm_forbidden_paths()
+        self.add_copy_spec([
+            '/var/run/vdsm/*',
+            '/usr/libexec/vdsm/hooks',
+            '/var/lib/vdsm'
+        ])
+
+        qemu_pid = self.get_process_pid('qemu-kvm')
+        if not qemu_pid == -1:
+                self.add_copy_spec(["/proc/%s/cmdline" % qemu_pid,
```
bmr-cymru: List here should be in this form:

```python
    self.add_copy_spec([
        "/proc/%s/cmdline" % qemu_pid,
        "/proc/%s/status" % qemu_pid,
        "/proc/%s/mountstats" % qemu_pid
    ])
```

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,153 @@
...
+        self.add_cmd_output(
+            "lvm vgs -v -o +tags --config \'%s\'" % LVM_CONFIG
+        )
+        self.add_cmd_output(
+            "lvm lvs -v -o +tags --config \'%s\'" % LVM_CONFIG
+        )
+        self.add_cmd_output(
+            "lvm pvs -v -o +all --config \'%s\'" % LVM_CONFIG
+        )
+        self.add_cmd_output([
```
bmr-cymru: Thse can be a single call:

```python
    self.add_cmd_output([
        "lvm vgs -v -o +tags --config \'%s\'" % LVM_CONFIG,
         "..."
    ])

File: sos/plugins/vdsm.py [hunk 26]
```diff
@@ -0,0 +1,153 @@
...
+            "fdisk -l",
+            "iostat"
+        ])
```
bmr-cymru: These are already collected in other plugins: unless there's a reason they must be disabled on VDSM it would be preferable to not duplicate that.

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,153 @@
...
+            'vdsm-client Host getDeviceList',
+            'vdsm-client Host hostdevListByCaps',
+            'vdsm-client Host getAllTasksInfo',
+            'vdsm-client Host getAllTasksStatuses'
+        ])
+
+        res = self.call_ext_prog('vdsm-client Host getConnectedStoragePools')
+        if res['status'] == 0:
+            pools = ast.literal_eval(res['output'])
+            for pool in pools:
```
bmr-cymru: Does this need any exception handling for parse failures on the `vdsm-client` data?

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,153 @@
...
+                self.add_cmd_output(
+                    'vdsm-client StoragePool getSpmStatus'
+                    ' storagepoolID={}'.format(pool)
+                )
+
+        res = self.call_ext_prog('vdsm-client Host getStorageDomains')
+        if res['status'] == 0:
+            sd_uuids = ast.literal_eval(res['output'])
```
bmr-cymru: Ditto.

sos/plugins/vdsm.py
bmr-cymru: (it seems to throw at least `ValueError` for a malformed string).

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,153 @@
...
+
+            for sd_uuid in sd_uuids:
+                self.add_cmd_output(
+                    'vdsm-tool dump-volume-chains %s' % sd_uuid
+                )
+
```
bmr-cymru: All of these `for x in xes: self.add_cmd_output()` can be rewritten as list comprehensions, usually making them shorter and more readable, e.g.:

```python
    dump_volume_chains_cmd = 'vdsm-tool dump-volume-chains %s'
    self.add_cmd_output([
        dump_volume_chains_cmd % uuid for uuid in sd_uuids
    ])
```

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,153 @@
...
+    def _add_vdsm_forbidden_paths(self):
+        """Add confidential sysprep vfds under /var/run/vdsm to
+         forbidden paths """
+
+        import glob
```
bmr-cymru: Imports need to be at the top of the file (pep8).

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,153 @@
...
+
+        for file_path in glob.glob("/var/run/vdsm/*"):
+            if file_path.endswith('.vfd') or \
+                    file_path.endswith('/isoUploader') or \
+                    file_path.endswith('/storage'):
+                self.add_forbidden_path(file_path)
```
bmr-cymru: This can probably be generalised a bit to have a list of suffixes, for this short set it's not too bad so no need to change for now, but if more appear in the future it's worth considering.

sos/plugins/vdsm.py
igoihman: pycodestyle didn't raise any error.

bmr-cymru: Ah, right - it's inside a string, of course. I don't really care how you format external data in strings, beyond their own requirements (LVM2 config snippets in this case), although if it's not a big deal for you then it might help avoid other readers having a double take here (looking at the diff it's less clear that it's a string than seeing it in context).

bmr-cymru: Ah, right - it's inside a string, of course. I don't really care how you format external data in strings, beyond their own requirements (LVM2 config snippets in this case), although if it's not a big deal for you then it might help avoid other readers having a double take here (looking at the diff it's less clear that it's a string than seeing it in context).

bmr-cymru: Ah, right - it's inside a string, of course. I don't really care how you format external data in strings, beyond their own requirements (LVM2 config snippets in this case), although if it's not a big deal for you then it might help avoid other readers having a double take here (looking at the diff it's less clear that it's a string than seeing it in context).

bmr-cymru: Ah, right - it's inside a string, of course. I don't really care how you format external data in strings, beyond their own requirements (LVM2 config snippets in this case), although if it's not a big deal for you then it might help avoid other readers having a double take here (looking at the diff it's less clear that it's a string than seeing it in context).

bmr-cymru: Ah, right - it's inside a string, of course. I don't really care how you format external data in strings, beyond their own requirements (LVM2 config snippets in this case), although if it's not a big deal for you then it might help avoid other readers having a double take here (looking at the diff it's less clear that it's a string than seeing it in context).

bmr-cymru: Ah, right - it's inside a string, of course. I don't really care how you format external data in strings, beyond their own requirements (LVM2 config snippets in this case), although if it's not a big deal for you then it might help avoid other readers having a double take here (looking at the diff it's less clear that it's a string than seeing it in context).

bmr-cymru: Ah, right - it's inside a string, of course. I don't really care how you format external data in strings, beyond their own requirements (LVM2 config snippets in this case), although if it's not a big deal for you then it might help avoid other readers having a double take here (looking at the diff it's less clear that it's a string than seeing it in context).

bmr-cymru: Ah, right - it's inside a string, of course. I don't really care how you format external data in strings, beyond their own requirements (LVM2 config snippets in this case), although if it's not a big deal for you then it might help avoid other readers having a double take here (looking at the diff it's less clear that it's a string than seeing it in context).

bmr-cymru: Ah, right - it's inside a string, of course. I don't really care how you format external data in strings, beyond their own requirements (LVM2 config snippets in this case), although if it's not a big deal for you then it might help avoid other readers having a double take here (looking at the diff it's less clear that it's a string than seeing it in context).

igoihman: I removed the spaces

sos/plugins/__init__.py
igoihman: OK, leaving this as is.

igoihman:
ping

sos/plugins/vdsm.py
igoihman: :)

igoihman:
@bmr-cymru could you review? Thanks

mwperina:
Bryn, any progress?
@bmr-cymru

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,151 @@
...
+global {
+    locking_type=0
+    use_lvmetad=0
+}
+devices {
+    preferred_names=["^/dev/mapper/"]
+    ignore_suspended_devices=1
+    write_cache_state=0
+    disable_after_error_count=3
+    filter=["a|^/dev/mapper/.*|", "r|.*|"]
```
nirs: This matches also lvs under /dev/mapper/vg-lv.
I think the best filter we can use is:

    filter = ["a|^/dev/disk/by-id/dm-uuid-mpath-|", "r|.+|"]

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,151 @@
...
+    packages = (
+        'vdsm',
+        'vdsm-client',
+    )
+
+    plugin_name = 'vdsm'
+
+    def setup(self):
+        lvm_system_dir = os.environ["LVM_SYSTEM_DIR"]
+        os.environ["LVM_SYSTEM_DIR"] = "/var/run/vdsm/lvm"
```
nirs: We don't need this, our LVM_CONFIG replace /run/vdsm/lvm/lvm.conf.

I'm removing it now from vdsm plugin:
https://gerrit.ovirt.org/c/96775

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,151 @@
...
+        self.add_forbidden_path('/etc/pki/vdsm/keys/*')
+        self.add_forbidden_path('/etc/pki/vdsm/libvirt-spice/*-key.*')
+        self.add_forbidden_path('/etc/pki/libvirt/private/*')
+
+        self.add_cmd_output('service vdsmd status')
```
nirs: What about supervdsmd?

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,151 @@
...
+            '/usr/libexec/vdsm/hooks',
+            '/var/lib/vdsm'
+        ])
+
+        qemu_pid = self.get_process_pid('qemu-kvm')
+        if qemu_pid:
+            self.add_copy_spec([
+                "/proc/%s/cmdline" % qemu_pid,
+                "/proc/%s/status" % qemu_pid,
+                "/proc/%s/mountstats" % qemu_pid
```
nirs: We can have many qemu-kvm processes. What is the purpose of this?

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,151 @@
...
+            'vdsm-client Host getAllTasksInfo',
+            'vdsm-client Host getAllTasksStatuses'
+        ])
+
+        try:
+            res = self.call_ext_prog(
+                'vdsm-client Host getConnectedStoragePools'
+            )
+            if res['status'] == 0:
+                pools = ast.literal_eval(res['output'])
```
nirs: This is wrong. The output of vdsm-client commands is json, not python dict representation,
so it should be parsed with json.loads().

bmr-cymru: Thanks. I thought some of these looked suspicious (and also missing exception handling).

sos/plugins/vdsm.py
nirs: The original code handled all qemu-kvm instances.

        p = subprocess.Popen(['/usr/bin/pgrep', 'qemu-kvm'],                                                                                                                 
                             stdout=subprocess.PIPE)
        out, err = p.communicate()
        for line in out.splitlines():
            pid = line.strip()
            self.addCopySpec("/proc/%s/cmdline" % pid)
            self.addCopySpec("/proc/%s/status" % pid)
            self.addCopySpec("/proc/%s/mountstats" % pid)

We becnhmarked collecting of process pids and found that pgrep is faster then writing your own version,
and for this case the pid per line is very easy to consume.
See https://gerrit.ovirt.org/c/69331/

sos/plugins/vdsm.py
```diff
@@ -0,0 +1,151 @@
...
+    write_cache_state=0
+    disable_after_error_count=3
+    filter=["a|^/dev/mapper/.*|", "r|.*|"]
+}
+"""
+LVM_CONFIG = re.sub(r"\s+", " ", LVM_CONFIG).strip()
+
+
+class Vdsm(Plugin, RedHatPlugin):
+    """vdsm Plugin"""
```
nirs: plugin?

sos/plugins/__init__.py
nirs: This finds only the first process, but we need all of them.

sos/plugins/vdsm.py
bmr-cymru: ack

bmr-cymru:
OK: first of all, I want to apologise again for the long delays in getting this work fully reviewed and included in master. We've all ended up with more on our plates over the last couple of cycles than expected, and this review (mostly just down to its size, and the fact that there is already a working solution in products), has always been the victim that got knocked off the end.

We will release sos-3.7 today, and I was hoping to be able to include this, but I think there are still a couple of minor items to address. Some of these I can take on, but I will need input from @igoihman or others for the rest. In spite of the fact that this will miss the 3.7 cut off, I'd be happy to make a 3.7.1 release in the near future to include this work if that would be helpful.

Having re-reviewed the current state today, the remaining items are:

* Switch to `metadata_read_only` in LVM config (fixed in my tree)
* Remove explicit size limits (now automated - fixed in my tree)
* Minor typos/text style (fixed in my tree)
* Replace Python "pgrep" implementation with callout (trivial)
* Record exceptions in getConnectedStoragePools in log, not as file (trivial)
* getConnectedStoragePools parsing (switch to JSON)

It's only the last of these that I need some input for - can we simply parse the JSON with `json.loads()` and then iterate over the elements of the result? Is `ast` actually needed in the plugin?

Once we clean up those last few bits the plugin will be ready to go in. Thanks again to everyone who has worked on this, or provided review or comments.

bmr-cymru:
I've pushed a branch `bmr-vdsm` with my current changes - please feel free to comment/fork.

nirs:
An important fix is to collect all qemu logs like old vdsm plugin. In this PR and also in your bmr-vdsm tree, we collect only the first qemu process logs.

I'm not sure @igoihman will have time for this, but maybe @msobczyk can complete this?

igoihman:
@nirs I will complete it next week. Thanks for the review!

bmr-cymru:
@nirs that's covered by this point, right?

* Replace Python "pgrep" implementation with callout (trivial)

So we'd have something like:
```python
        qemu_pids = self.get_process_pids('qemu-kvm')
        if qemu_pids:
            self.add_copy_spec(["/proc/%s/cmdline" % pid for pid in qemu_pids])
            self.add_copy_spec(["/proc/%s/status" % pid for pid in qemu_pids])
            self.add_copy_spec(["/proc/%s/mountstats" % pid for pid in qemu_pids])
```

nirs:
@bmr-cymru this should work, but I think it makes more sense to copy /proc/pid/cmdline, /proc/pid/status, and /proc/pid/mountstats before we move to the next pid. This increase the chance that we have complete info per vm, in case a vm is shut down in the middle of the copy.

I hope that missing files (because a vm was shut down in the middle) are handled correctly by sos.

So maybe something like:

```python
    files = ["cmdline", "status", "mountstats"]
    self.add_copy_spec([
        "/proc/%s/%s" % (pid, name)
        for pid in qemu_pids
        for name in files
    ])
```
```
>>> files = ["cmdline", "status", "mountstats"]
>>> qemu_pids = [1, 2, 3]
>>> ["/proc/%s/%s" % (pid, name) for pid in qemu_pids for name in files]
['/proc/1/cmdline', '/proc/1/status', '/proc/1/mountstats', '/proc/2/cmdline', '/proc/2/status', '/proc/2/mountstats', '/proc/3/cmdline', '/proc/3/status', '/proc/3/mountstats']
```

bmr-cymru:
Yes, I like that - thanks!

bmr-cymru:
> I  hope that missing files (because a vm was shut down in the middle) are handled correctly by sos.

Yes - anything missing will be present in the logs, but will not cause any error. There is a brief time delay bewteen `setup()` and `collect()`, during which time PIDs could disappear, but this is relatively short (esp. today with threaded execution), since `setup()` is mostly just list munging and a few fairly cheap commands.

bmr-cymru:
This pull request **introduces 3 alerts** when merging 1dce093efe9c88f49d6aa572bd6a21ab0dc30f01 into 9b1f9472e7b0b8a993f635c8d4c757b59c46ed0f - [view on LGTM.com](https://lgtm.com/projects/g/sosreport/sos/rev/pr-a85a07dea2759f7dea84c29d535e77a09642a4d2)

**new alerts:**

* 2 for Unsupported format character
* 1 for Unused local variable

---

bmr-cymru:
This pull request **introduces 1 alert** when merging ab2d29177801eb9acbbdb39f216fd8f67ab49afd into 9b1f9472e7b0b8a993f635c8d4c757b59c46ed0f - [view on LGTM.com](https://lgtm.com/projects/g/sosreport/sos/rev/pr-9af238707edfa757243953fc6238f58083733dc5)

**new alerts:**

* 1 for Unused import

---

igoihman:
@nirs @bmr-cymru I addresses all remaining items, including capturing all `qemu-kvm` pids.
@bmr-cymru I copied the commits from your branch, thanks for helping with this effort!

PR #1206, opened 2018-02-06
[yum] Get a list of available plugins.

dmach:
Signed-off-by: Daniel Mach <dmach@redhat.com>

---

### DIFF
--- a/sos/plugins/yum.py
+++ b/sos/plugins/yum.py
@@ -44,6 +44,13 @@ class Yum(Plugin, RedHatPlugin):
         # Get a list of channels the machine is subscribed to.
         self.add_cmd_output("yum -C repolist")
 
+        # Get a list of available plugins
+        cmd = """bash -c '
+            for i in /usr/lib/yum-plugins/*.py; do
+                echo "$i ($(rpm -qf $i))";
+            done'"""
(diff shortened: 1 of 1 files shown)


bmr-cymru:
ack to the effort: we should probably do this for `dnf` too.

sos/plugins/yum.py
```diff
@@ -44,6 +44,13 @@ def setup(self):
         # Get a list of channels the machine is subscribed to.
         self.add_cmd_output("yum -C repolist")
 
+        # Get a list of available plugins
+        cmd = """bash -c '
+            for i in /usr/lib/yum-plugins/*.py; do
+                echo "$i ($(rpm -qf $i))";
+            done'"""
```
bmr-cymru: Please don't put shell code in `sos`: there's absolutely no reason to not do this in python. It may also be better to simply collect the whole of `/etc/yum/pluginconf.d` since each plugin is required to have a configuration file there which will also reveal the enable states.

bmr-cymru:
This is still a useful idea, but since there's been no response to the review comments in two months I'll rewrite it myself for 3.6.

bmr-cymru:
Fixed in commit 26c45f41c527f1193aa4a12814d7d07f5505fe07

Issue #605, opened 2015-07-21
Review SELinux commands in enforcing/permissive/disabled modes.


bmr-cymru:
We had a bug report against RHEL recently where a user noticed setfiles messages leaking through from fixfiles when run with SELinux disabled:

```
# sosreport -vvv --batch --alloptions &> /tmp/sosreport.$(date +%Y%m%d-%H%M%S).log
usage:  /sbin/setfiles [-dnpqvW] [-o filename] [-r alt_root_path ] spec_file pathname...
usage:  /sbin/setfiles -c policyfile spec_file
usage:  /sbin/setfiles -s [-dnpqvW] [-o filename ] spec_file
```

Although this case appears to be a fixfiles bug we have never audited the commands issued by the SELinux module to ensure they behave correctly / predictably in different SELinux modes.

Only enabling the plugin on systems where SELinux is currently enabled is not an option since that causes a loss of diagnostics on hosts where SELinux has been temporarily disabled.

If there are commands that /must/ only run when SELinux is on it's trivial to gate these on the state of SELinux via the Python SELinux bindings or via /sys.

TurboTurtle:
Closing per f72aaa6

PR #1209, opened 2018-02-07
[selinux] Use semanage and only run commands if selinux is not disabled

TurboTurtle:
Resolves: #605 

This has the selinux plugin run in either `enforcing` or `permissive` mode, and skipping the bulk of selinux command output if it is `disabled`. However, even if it is disabled, still collect config information. If selinux is disabled, the output from the selinux commands is either useless or simply empty, hence why we skip if it is `disabled`.

---

### DIFF
--- a/sos/plugins/selinux.py
+++ b/sos/plugins/selinux.py
@@ -23,30 +23,42 @@ class SELinux(Plugin, RedHatPlugin):
     profiles = ('system', 'security', 'openshift')
 
     option_list = [("fixfiles", 'Print incorrect file context labels',
-                    'slow', False),
-                   ("list", 'List objects and their context', 'slow', False)]
+                    'slow', False)]
     packages = ('libselinux',)
 
     def setup(self):
-        self.add_copy_spec("/etc/selinux")
-        self.add_cmd_output([
-            "sestatus -b",
-            "semodule -l",
-            "selinuxdefcon root",
-            "selinuxconlist root",
-            "selinuxexeccon /bin/passwd",
-            "semanage -o -",
-            "ps axuZww"
+        self.add_copy_spec([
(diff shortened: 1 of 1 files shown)


bmr-cymru:
Ack, lgtm - my only concern was that we'd still collect config data, in case the user has a problem enabling SELinux.

sos/plugins/selinux.py
```diff
@@ -23,30 +23,42 @@ class SELinux(Plugin, RedHatPlugin):
...
+        self.add_cmd_output('sestatus')
+
+        state = self.get_command_output('getenforce')['output']
+        if state is not 'Disabled':
             self.add_cmd_output([
-                "semanage fcontext -l",
-                "semanage user -l",
-                "semanage login -l",
-                "semanage port -l"
+                'ps auxZww',
```
bmr-cymru: I struggle with this: on the one hand, process data belongs in the process plugin. On the other, this should only be enabled when SELinux is, and there's still some variation in `ps` commands supporting the option at all.

I think we can only clean this up when we have some level of data exposure between plugins; if the process plugin can get a "SELinux is enabled" predicate then we can move this over, but I think for now it has to go here.

sos/plugins/selinux.py
```diff
@@ -23,30 +23,42 @@ class SELinux(Plugin, RedHatPlugin):
...
+                'user',
+                'port',
+                'login',
+                'node',
+                'interface',
+                'module'
+            ]
+
+            for subcmd in subcmds:
+                self.add_cmd_output("semanage %s -l" % subcmd)
```
bmr-cymru: `add_cmd_output(["semanage %s -l" % sc for sc in subcmds])`?

PR #1211, opened 2018-02-10
[kernel] Add collection of tracing information

TurboTurtle:
Adds collection of /sys/kernel/debug/tracing and /sys/kernel/debug/sunrpc information to the kernel plugin.

Skips items like `trace_pipe` and `per_cpu/*` that would cause either sos to hang or a potential load spike.

---

### DIFF
--- a/sos/plugins/kernel.py
+++ b/sos/plugins/kernel.py
@@ -57,6 +57,18 @@ class Kernel(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
         ])
 
         clocksource_path = "/sys/devices/system/clocksource/clocksource0/"
+
+        forbids = [
+            'trace_pipe',
+            'README',
+            'trace_stat/*',
+            'per_cpu/*',
+            'events/*'
+        ]
+
+        for forbid in forbids:
+            self.add_forbidden_path('/sys/kernel/debug/tracing/%s' % forbid)
+
         self.add_copy_spec([
             "/proc/modules",
             "/proc/sys/kernel/random/boot_id",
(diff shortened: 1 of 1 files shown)


sos/plugins/kernel.py
```diff
@@ -65,6 +77,8 @@ def setup(self):
             "/sys/module/*/refcnt",
             "/sys/module/*/taint",
             "/sys/firmware/acpi/*",
+            "/sys/kernel/debug/tracing/*",
+            "/sys/kernel/debug/sunrpc/*",
```
bmr-cymru: Wouldn't this be better in the `sunrpc` plugin?

bmr-cymru: (which may need some work on its `check_enabled()`...)

TurboTurtle: Yup, you're right. I totally forgot that was a thing. I'll remove this bit later today.

bmr-cymru: Cool 👍 

I think we all did a bit.. It doesn't seem to have had much lovin' in a while...

TurboTurtle:
Updated to remove the sunrpc bit.

sos/plugins/kernel.py
```diff
@@ -57,6 +57,18 @@ def setup(self):
...
+        forbids = [
+            'trace_pipe',
+            'README',
+            'trace_stat/*',
+            'per_cpu/*',
+            'events/*'
+        ]
+
+        for forbid in forbids:
+            self.add_forbidden_path('/sys/kernel/debug/tracing/%s' % forbid)
```
bmr-cymru: We really should make this accept a singleton or a list like everything else...

PR #1212, opened 2018-02-10
[filesys] Clean lsof output, capture lslocks and mountinfo for pids

TurboTurtle:
Cleans the output of lsof, removing errors and warnings that are
expected due to the lsof command that is being run. Other errors that
lsof might produce to stderr are still preserved.

Additionally, capture lslocks output and /proc/pid/mountinfo for each
pid.

---

### DIFF
--- a/sos/plugins/filesys.py
+++ b/sos/plugins/filesys.py
@@ -33,13 +33,15 @@ class Filesys(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
             "/proc/self/mounts",
             "/proc/self/mountinfo",
             "/proc/self/mountstats",
+            "/proc/[0-9]*/mountinfo",
             "/proc/mounts"
         ])
         self.add_cmd_output("mount -l", root_symlink="mount")
         self.add_cmd_output("df -al -x autofs", root_symlink="df")
         self.add_cmd_output([
             "df -ali -x autofs",
-            "findmnt"
+            "findmnt",
+            "lslocks"
         ])
 
         if self.get_option('lsof'):
@@ -59,5 +61,12 @@ class Filesys(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
             r"(password=)[^\s]*",
             r"\1********"
         )
+
+        # remove expected errors from lsof due to command formatting, but still
+        # keep stderr so other errors are reported
+        regex = (r"(lsof: (avoiding (.*?)|WARNING: can't stat\(\) (.*?))|"
+                 "Output information may be incomplete.)\n")
+
+        self.do_cmd_output_sub("lsof", regex, '')
 
 # vim: set et ts=4 sw=4 :


sos/plugins/filesys.py
```diff
@@ -33,13 +33,15 @@ def setup(self):
             "/proc/self/mounts",
             "/proc/self/mountinfo",
             "/proc/self/mountstats",
+            "/proc/[0-9]*/mountinfo",
```
bmr-cymru: This is potentially expensive on systems with many threads. Can we get some performance figures for extreme cases to evaluate this?

sos/plugins/filesys.py
```diff
@@ -60,4 +62,11 @@ def postproc(self):
             r"\1********"
         )
 
+        # remove expected errors from lsof due to command formatting, but still
+        # keep stderr so other errors are reported
+        regex = (r"(lsof: (avoiding (.*?)|WARNING: can't stat\(\) (.*?))|"
+                 "Output information may be incomplete.)\n")
+
+        self.do_cmd_output_sub("lsof", regex, '')
+
```
bmr-cymru: Does this not appear on `stderr`? We can throw that away, unless the intent is to preserve other error messages.

bmr-cymru: Ack per the comment.

sos/plugins/filesys.py
```diff
@@ -33,13 +33,15 @@ def setup(self):
...
+            "/proc/[0-9]*/mountinfo",
             "/proc/mounts"
         ])
         self.add_cmd_output("mount -l", root_symlink="mount")
         self.add_cmd_output("df -al -x autofs", root_symlink="df")
         self.add_cmd_output([
             "df -ali -x autofs",
-            "findmnt"
+            "findmnt",
+            "lslocks"
```
bmr-cymru: Ack

sos/plugins/filesys.py
TurboTurtle: These are the most heavily loaded test systems I have access to:

RHEL 7 box running 50 kubernetes pods w/ 4770 threads, without this change, filesys plugin only:

```
real	0m3.168s
user	0m1.278s
sys	0m0.210s
```

Same box, with the change:

```
real	0m4.007s
user	0m2.004s
sys	0m0.743s
```

And from a RHV hypervisor running 30 VMs w/ 1200 threads, unpatched:

```
real	0m0.933s
user	0m0.584s
sys	0m0.091s
```

Same system, with this patch:

```
real	0m1.212s
user	0m0.903s
sys	0m0.319s
```

bmr-cymru: Needs a bit more than that really - customer systems that have proved problematic have had > 40K threads. My laptop tops out at around 10k, but with 8k running I see:

8K threads
```
# time sosreport -v --batch -o filesys
...

real	0m19.272s
user	0m9.917s
sys	0m8.136s
```

Vs.:

400 threads
```
# time sosreport -v --batch -o filesys
...

real	0m2.800s
user	0m2.473s
sys	0m0.421s
```

This isn't the end of the world for sure but I'd still like to see some more numbers from large memory systems with very large numbers of threads running.

To set the threads up I use:

```
mkthreads() { for i in $(seq 1 $1); do echo sleep 300; done }
```
```

Comment (TurboTurtle) [reply sos/plugins/filesys.py]: Oh, I see what you mean now.

Did some digging and got my hands on a bigger test box that topped out at ~32k threads:

```
real	0m49.234s
user	0m27.526s
sys	0m16.285s
```

bmr-cymru: That still ain't so bad... The last similar patch (to add a whole bunch of files from `/proc/<pid>` floundered on these types of hosts - tens of minutes kinda thing...

PR #1217, opened 2018-02-16
[rpm] Collect package-cleanup and add option for rpm DB

TurboTurtle:
Adds collection of package-cleanup output if the package-cleanup command
is available on the system.

Adds the rpmdb option, disabled by default, to collect the contents of
/var/lib/rpm to diagnose RPM DB problems. Users should note that use of
this option will typically significantly increase the size of the final
sosreport archive.

---

### DIFF
--- a/sos/plugins/rpm.py
+++ b/sos/plugins/rpm.py
@@ -13,6 +13,7 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
 from sos.plugins import Plugin, RedHatPlugin
+from sos.utilities import is_executable
 
 
 class Rpm(Plugin, RedHatPlugin):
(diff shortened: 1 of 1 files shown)


bmr-cymru:
I'm struggling a bit with some of the changes in this patch - I think they need to be split up, or we need to reconsider the division of labour between the `yum` and `rpm` plugins.

We're collecting the `package-cleanup` output in the `rpm` plugin:

```python
        if is_executable('package-cleanup'):
            self.add_cmd_output([
                'package-cleanup --problems',
                'package-cleanup --dupes'
            ])
```

(and also gating it on an `is_executable()` test - this should never normally be required in `sos`, since command execution is expected to be speculative).

Formally this is a better fit to the `yum` plugin, since the command is provided by the `yum-utils` package; for a start, we have a conflict to deal with between RHEL and recent Fedora: in f26 the `package-cleanup` command is provided by *both* `yum-utils` and `dnf-utils`, with a conflict between the two preventing them from being simultaneously installed:

```
# dnf -y install dnf-utils
Last metadata expiration check: 2:29:47 ago on Wed 09 May 2018 08:58:09 BST.
Error: 
 Problem: problem with installed package yum-utils-1.1.31-512.fc26.noarch
  - package dnf-utils-2.1.1-1.fc26.noarch conflicts with yum-utils < 1.1.31-513 provided by yum-utils-1.1.31-512.fc26.noarch
  - conflicting requests
  - package dnf-utils-2.1.5-4.fc26.noarch conflicts with yum-utils provided by yum-utils-1.1.31-512.fc26.noarch
(try to add '--allowerasing' to command line to replace conflicting packages or '--skip-broken' to skip uninstallable packages)
```

I am inclined to think that we either need to create a new "package cleanup", or "package tools" plugin (don't really like this), or we put the command into both the `yum` and `dnf` plugins with the understanding that only one will execute, and that the command will be speculatively collected if present. I don't like this duplication much either, but for now it feels like it may be our best option.

The RPMDB collection looks fine, although I'm wondering what the `lsof` invocation buys us beyond what we're already collecting (optionally in `filesys` and by default in `process`).

bmr-cymru:
The other option to avoid duplication is to merge the `dnf` and `yum` plugins; there's some justification for that, but I've not looked at the details involved and it seems like it may be tending towards a foolish consistency...

PR #1218, opened 2018-02-16
[ntp] Collect hwclock and ntpq -c output

TurboTurtle:
Adds collection of hwclock, and adds -n to the ntpq -p command
collected.

Collects `ntpq -c as` output and will also iterate over all association
ids listed to collect `ntpq -c 'rv $id'`.

---

### DIFF
--- a/sos/plugins/ntp.py
+++ b/sos/plugins/ntp.py
@@ -32,8 +32,14 @@ class Ntp(Plugin):
         ])
         self.add_cmd_output([
             "ntptime",
-            "ntpq -p"
+            "ntpq -pn",
+            "ntpq -c as"
         ])
+
+        ids = self.get_command_output('ntpq -c as')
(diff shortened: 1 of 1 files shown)


bmr-cymru:
What's the point of collecting `hwclock` twice? (in this patch, and in the `date` plugin since you added it last year ;-) ).

I can trim this out as I merge, but is the thinking that these plugins will not always be run? They're both default on (although `date` could use adding to a profile) so that seems a stretch.

TurboTurtle:
Nope, that was an oversight of me not pulling it out here.

Originally, I was adding it here since there wasn't a better place for it. But splitting up the `general` plugin changed that, and I just forgot to cycle back on this one.

TurboTurtle:
Removed the duplicate `hwclock` collection.

PR #1219, opened 2018-02-19
[postfix|sendmail] Add mailq output

TurboTurtle:
Adds collection of mailq output to the postfix and sendmail plugins.
Sendmail also collects 'mailq -Ac'.

Additionally, cleanup redundant checks and add_copy_spec in sendmail
plugin.

---


bmr-cymru:
Ack: duplicating `mailq` is slightly annoying, but it's better than introducing a new `mail` plugin just for this one command.

PR #1221, opened 2018-02-21
[networking] Minor update to add sysfs netdev stats and complete /etc/NetworkManager/ contents

ptalbert:
---


TurboTurtle:
LGTM, my only question would be are there areas under `/etc/NetworkManager` where we'd need to obfuscate plaintext passwords?

ptalbert:
The networking plugin's postproc function already takes care of obfuscating keys under /etc/NetworkManager/system-connections/.

sos/plugins/networking.py
```diff
@@ -131,8 +131,7 @@ def setup(self):
             "/etc/host*",
             "/etc/resolv.conf",
             "/etc/network*",
-            "/etc/NetworkManager/NetworkManager.conf",
-            "/etc/NetworkManager/system-connections",
+            "/etc/NetworkManager/",
```
bmr-cymru: What about potential secrets stored in this location? The documentation for `system-connections` explicitly mentions that these files may contain sensitive data:

```
       keyfile
           The keyfile plugin is the generic plugin that supports all the connection types and capabilities
           that NetworkManager has. It writes files out in an .ini-style format in
           /etc/NetworkManager/system-connections.

           The stored connection file may contain passwords and private keys, so it will be made
           readable only to root, and the plugin will ignore files that are readable or writeable by any user or group
           other than root.
```

I realise we've been sucking up `system-connections` without this for some time but if we are going to increase the set of data collected (inc. the `VPN/` subdirectory) we really need to fix this at the same time.

bmr-cymru: Ah, gotcha, yes we already have rules for system-connections. It's not entirely clear whether there are other locations that could contain secrets (only the keyfile plugin mentions it), but if we are already processing the documented locations for this tree then I'm fine with the change.

ptalbert: I am peeking at the `nm-settings` man page and from what I can see, there are potentially other "secret" values that could be in the `system-connections` files.

While we are at it, sosreport already records the entire `/etc/sysconfig/network-scripts/` directory. The `ifcfg` files in here could have secrets, as do any `keys-<connection>` files. Should we be masking out any secret lines in those files as well? Or is this functionality that a subsequent tool such as soscleaner should be handling, in which case, don't bother masking system-connections?

bmr-cymru: Better if we can get them all strimmed out in sos itself - although I wouldn't hold this PR up for that work as it's separate cleanup of a long standing omission.

bmr-cymru: Just returning to this subthread: I'll add a follow up patch when I merge this to add `/etc/sysconfig/network-scripts/key-*` to the forbidden files lists, but was there anything else specific we're aware of for now to fix up?

We're probably still a few weeks at least away from releasing 3.6 but I'd like to get as many of these addressed before we do.

PR #1224, opened 2018-02-23
[os_net_config] Capture the os-net-config run time data

krsacme:
For DPDK OpenStack deployments with TripleO, os-net-config
binds the DPDK PMD driver to the interface and writes the
details in to the directory /var/lib/os-net-config. Adding
this directory for the report collection.

Signed-off-by: Saravanan KR <skramaja@redhat.com>

---

### DIFF
--- a/sos/plugins/os_net_config.py
+++ b/sos/plugins/os_net_config.py
@@ -24,6 +24,7 @@ class OsNetConfig(Plugin, RedHatPlugin, UbuntuPlugin, DebianPlugin):
 
     def setup(self):
         self.add_copy_spec("/etc/os-net-config")
+        self.add_copy_spec("/var/lib/os-net-config")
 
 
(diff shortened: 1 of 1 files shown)


bmr-cymru:
Please amend your commit to use the correct `[os_net_config]` tag as the prefix to the subject line.

Other than that the change seems fine, but are there any possible secrets, or large quantities of data (e.g. logs) present at this location?

krsacme:
> are there any possible secrets, or large quantities of data (e.g. logs) present at this location?

@bmr-cymru  Thanks for the review. There are no secrets in this folder. And it will have 1 yaml for for DPDK and 2 yaml files for SR-IOV with all the required mapping information. And the content will be less that few 100 bytes.

bmr-cymru:
:+1:

PR #1227, opened 2018-02-25
[ovirt-provider-ovn] Introducing a plugin for ovirt-provider-ovn.

leongold:
---

### DIFF
--- a/sos/plugins/ovirt_provider_ovn.py
+++ b/sos/plugins/ovirt_provider_ovn.py
@@ -1,0 +1,40 @@
+# Copyright (C) 2018 Red Hat, Inc.,
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+from sos.plugins import Plugin, RedHatPlugin
+
+
+class OvirtProviderOvn(Plugin, RedHatPlugin):
+    """oVirt OVN Provider
+    """
+
+    provider_conf = '/etc/ovirt-provider-ovn/ovirt-provider-ovn.conf'
+    packages = ('ovirt-provider-ovn',)
+    plugin_name = 'ovirt_provider_ovn'
+    profiles = ('virt',)
+
+    def setup(self):
+        self.add_copy_spec(self.provider_file)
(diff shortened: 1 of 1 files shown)


sos/plugins/ovirt_provider_ovn.py
```diff
@@ -0,0 +1,33 @@
...
+class OvirtProviderOvn(Plugin, RedHatPlugin):
+    """oVirt OVN Provider
+    """
+
+    packages = ('ovirt-provider-ovn',)
+    plugin_name = 'ovirt_provider_ovn'
+    profiles = ('virt',)
+
+    def setup(self):
+        self.add_copy_spec(['/etc/ovirt-provider-ovn/ovirt-provider-ovn.conf'])
```
bmr-cymru: There's no need for a list here: `Plugin.add_copy_spec()` accepts either a single string argument, or a list of strings.

sos/plugins/ovirt_provider_ovn.py
```diff
@@ -0,0 +1,33 @@
...
+
+        spec = '/var/log/ovirt-provider-ovn.log'
+        if self.get_option('all_logs'):
+            spec += '*'
+        self.add_copy_spec([spec])
```
bmr-cymru: Please implement log size limiting using the global `--log-size` option, or add a comment explaining why log size limiting is not useful for these logs (e.g. they are guaranteed to be a fixed or small size).

See for the `openvswitch` plugin for an example of how to do this in combination with `--all-logs`.

sos/plugins/ovirt_provider_ovn.py
```diff
@@ -25,9 +25,10 @@ class OvirtProviderOvn(Plugin, RedHatPlugin):
     profiles = ('virt',)
 
     def setup(self):
-        self.add_copy_spec(['/etc/ovirt-provider-ovn/ovirt-provider-ovn.conf'])
+        self.add_copy_spec('/etc/ovirt-provider-ovn/ovirt-provider-ovn.conf')
 
+        self.limit = self.get_option('log_size')
```
bmr-cymru: There's no need to make this a member of the `OvirtProviderOvn` class: it is assigned and referenced in only a single method.

The pattern in the `openvswitch` plugin is perfectly correct.

leongold:
I guess we don't even need a local variable for sizelimit.

bmr-cymru:
Right: for this case, since we're only using it for a single call you can just grab it straight from `get_option()`. There are a few examples in the tree still that get this wrong: we're slowly cleaning those up as we go.

leongold:
Hey Bryn, anything else I'm missing?

bmr-cymru:
Looks good, thanks for making the changes. The only other thing that could come up would be if there could be any sensitive data (keys, passwords or other secrets) in the configuration locations collected - if not then this is ready for merge.

leongold:
Thanks Bryn, this can be merged

pmoravec:
Reading https://github.com/oVirt/ovirt-provider-ovn/blob/master/README.adoc , shouldnt we obfuscate value of e.g. `ovirt-sso-client-secret` ?

Otherwise +1.

pmoravec:
+1

leongold:
ping

dougsland:
+1

Issue #1228, opened 2018-02-27
Collect the output of 'systemd-analyze blame'


jcastill:
The output of this command is very useful when diagnosing slow boot issues, and we need more detailed information that the current command 'systemd-analyze' cannot provide.

PR #1229, opened 2018-02-27
[systemd] Add 'systemd-analyze blame' to systemd plugin

jcastill:
This patch adds the output of 'systemd-analyze blame', which is
very useful when we try to analyze slow boots, providing more
fine-grained information than the one we currently have with
'systemd-analyze' in the sosreport.

Solves #1228.

---

- [x Is the subject and message clear and concise?


pmoravec:
+1

PR #1234, opened 2018-03-02
[openvswitch] verbose tunnel port output

teknoraver:
Run the command "ovs-appctl tnl/ports/show" with verbose output.
Leave the non verbose output as well, as the output differs.

Signed-off-by: Matteo Croce <mcroce@redhat.com>

---
Please place an 'X' inside each '[X]' to confirm you adhere to our [Contributor Guidelines](https://github.com/sosreport/sos/wiki/Contribution-Guidelines)

### DIFF
--- a/sos/plugins/openvswitch.py
+++ b/sos/plugins/openvswitch.py
@@ -71,7 +71,7 @@ class OpenVSwitch(Plugin):
             # Capture tnl arp table"
             "ovs-appctl tnl/arp/show",
             # Capture a list of listening ports"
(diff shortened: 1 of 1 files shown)


pmoravec:
Isnt output/info from "ovs-appctl tnl/ports/show" included in the verbosed version? I guess not so, then adding (instead of replacing) the verbose version makes sense.

teknoraver:
@pmoravec yes the non verbose output is included in the verbose too, but there could be some scripts relying on it

teknoraver:
rebased, any other option?

PR #1235, opened 2018-03-02
[archive] Change log level for missing compression tools

TurboTurtle:
Changes the log level from error to info when sos is checking for valid
compression tools (xz, bzip2, gzip) and finds that one does not exist.

The fact that one of the compression methods isn't available is not
necessarily an error depending on where sos is being run, for example in
a container that provides xz but not gzip. In this case an error is
printed saying gzip is not available, even if sos would never have used
gzip due to the presence of xz, which could cause end user confusion.

If no supported compression methods are available, then that is still
properly reported as an error.

---

### DIFF
--- a/sos/archive.py
+++ b/sos/archive.py
@@ -428,7 +428,7 @@ class TarFileArchive(FileCacheArchive):
             if is_executable(method):
                 methods.append(method)
             else:
-                self.log_error("\"%s\" command not found." % method)
+                self.log_info("\"%s\" compression method unavailable" % method)
         if self.method in methods:
             methods = [self.method]
 


bmr-cymru:
Ack to the re-levelling, but I also kinda think the message is wrong:

`"foo" command not found`

Well, we didn't want to run it anyway...

How about something like "not using compression method "foo": no command found"?

bmr-cymru:
(or command not found, or whatever...)

TurboTurtle:
Makes sense. Updated to say `"foo" compression method unavailable`.

bmr-cymru:
:+1:

PR #1237, opened 2018-03-03
[chrony] collect chronyc -n -c sources

pmoravec:
The options generate output better for further parsing

Resolves: #1237

---


pmoravec:
It can be debatable if this command shall replace the currently collected `chronyc sources` one. The difference is in IP->hostname translation and CSV format. Current is better for humans, 2nd for further parsing:


    # chronyc sources
    210 Number of sources = 4
    MS Name/IP address         Stratum Poll Reach LastRx Last sample               
    ===============================================================================
    ^+ blue.1e400.net                3  10   337  1021  +3535us[+3315us] +/-  128ms
    ^* horp-bsd01.horp.io            2  10   377   915   -620us[ -841us] +/-  127ms
    ^+ mirror1.sjc02.svwh.net        2  10   377   648   +702us[ +702us] +/-  123ms
    ^+ resolver1.skyfiberintern>     2  10   377     9  -1498us[-1498us] +/-  125ms
    # chronyc -n -c sources
    ^,+,66.228.42.59,3,10,277,32,0.005840624,0.005840624,0.118138827
    ^,*,192.111.144.114,2,10,377,966,-0.000620219,-0.000841377,0.127188414
    ^,+,72.5.72.15,2,10,377,699,0.000701747,0.000701747,0.122738905
    ^,+,162.210.110.4,2,10,377,60,-0.001497601,-0.001497601,0.125288486
    #

bmr-cymru:
I'm not sure I see a need for having both, but there are two separate changes here to consider:

* -n, which has collection side benefits if name resolution is broken (we use this type of option other plaes)
- -c, for CSV output, which is more debatable: it's possible to convert between either format, so if anything, i'd be inclined to not change that and to suggest anyone who wants it does a `s/[[:space:]]/,/` type conversion.

bmr-cymru:
Is there a specific motivation for CSV formatting?

pmoravec:
Makes sense, `-n` is valuable in case of hostname resolving problems, while different formatting is not required - dropped the `-c` from the command.

bmr-cymru:
I think we might as well delete the `chronyc sources`: it's not gaining us anything over the `-n` output, except for hostnames (which may wind up in time outs).

PR #1240, opened 2018-03-05
[docs] Update Debian and Ubuntu command

mikelolasagasti:
Since Debian Jessie (8) and Ubuntu 14.04 `apt` command is available, no need to use `apt-get`.

Debian `stable` and `old-stable` have sosreport available, no need to mention `Sid`.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,17 +91,17 @@ Fedora/RHEL users install via yum:
 yum install sos
 ```
 
-Debian(Sid) users install via apt:
+Debian users install via apt:
 
 ```
-apt-get install sosreport
+apt install sosreport
 ```
 
 
-Ubuntu(14.04 LTS and above) users install via apt:
+Ubuntu (14.04 LTS and above) users install via apt:
 
 ```
-sudo apt-get install sosreport
+sudo apt install sosreport
 ```
 
  [0]: https://github.com/sosreport/sos/wiki/Contribution-Guidelines


bmr-cymru:
Thanks for the patch - I'd like to get an ack from one of our Debian / Ubuntu contributors as I'm less familiar with those distros.

bmr-cymru:
>  "apt" is intended as an end user interface and enables some options better suited for interactive usage.

We're not using them interactively: are any of these options likely to result in breakage? E.g. blocking for terminal IO etc.?

slashdd:
@bmr-cymru 
APT is widely used through Ubuntu since 2014 (Trusty).
I'm :+1 in approving this documentation change.

Additionally, all new or updated Canonical/Ubuntu related project documentation refer to "apt" now.

Example 
https://maas.io/install
https://linuxcontainers.org/lxd/getting-started-cli/

slashdd:
@mikelolasagasti 

Why not adding sudo for Debian, just like you did for Ubuntu ?

```
test@debian:/$ apt install cowsay
E: Could not open lock file /var/lib/dpkg/lock - open (13: Permission denied)
E: Unable to lock the administration directory (/var/lib/dpkg/), are you root?
```

slashdd:
@mikelolasagasti 
I forgot for a second that sudo is not installed by default in Debian as oppose to Ubuntu. That would explain why sudo isn't mention. Would it be worth mentioning it requires root privileges to run apt install in Debian or keep it as is ?

bmr-cymru:
I'm fine with whatever is "natural" for the distro in question: if Debian docs don't illustrate `sudo` then we shouldn't. Likewise, if Ubuntu does then we do etc.

The idea is we should be giving examples that make sense for the distro we're giving them for, without going to the extent of bending over backwards to explain that root privileges might be needed for package installation.

slashdd:
sound good !

Issue #1243, opened 2018-03-06
oVirt hosted-engine plugin collects by default only latest setup log


didib:
It should be useful, and not too heavy, to collect all setup logs.

Since oVirt 4.2.1, this includes separate ansible logs.

See also: https://bugzilla.redhat.com/show_bug.cgi?id=1542849

PR #1244, opened 2018-03-06
[ovirt_hosted_engine] Collect all setup logs

didib:
In oVirt 4.2.1 and later, setup also runs several ansible playbooks,
each logging to its own file. Collect everything.

There should not be many files usually, as if all goes well users run
setup only once. Otherwise, it is useful for debugging to get older logs
as well.

Resolves: #1243

---

### DIFF
--- a/sos/plugins/ovirt_hosted_engine.py
+++ b/sos/plugins/ovirt_hosted_engine.py
@@ -32,7 +32,6 @@ class OvirtHostedEngine(Plugin, RedHatPlugin):
     plugin_name = 'ovirt_hosted_engine'
     profiles = ('virt',)
 
-    SETUP_LOG_GLOB = '/var/log/ovirt-hosted-engine-setup/*.log'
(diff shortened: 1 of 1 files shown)


sos/plugins/ovirt_hosted_engine.py
```diff
@@ -59,10 +59,6 @@ def setup(self):
         all_setup_logs = glob.glob(self.SETUP_LOG_GLOB)
         all_setup_logs.sort(reverse=True)
```
bmr-cymru: What is the point of making this list, sorting it, and then ... doing nothing with it apart from testing it has non-zero length?

This just looks like left-over garbage.

didib:
Agreed. Removed all relevant redundant code.

didib:
Sorry, tried and failed to find how to mark the "Changes requested" as "Done". They are done, though.

didib:
@bmr-cymru , anything else?

sandrobonazzola:
@bmr-cymru anything else to be changed?

PR #904, opened 2016-12-20
[libvirt] Include '/etc/libvirt/qemu.conf'

chaitshastri:
Parameter such as "security_driver"(included in this conf
file) is responsible to enable SElinux on qemu instances.
One such case was observed when live migration failed as
the driver was set to "none" which was hard to debug as
this file was not included in soseport.

Closes-Bug: #903

PR #1245, opened 2018-03-07
[libvirt] Add file /etc/libvirt/qemu.conf

zerodayz:
---

### DIFF
--- a/sos/plugins/libvirt.py
+++ b/sos/plugins/libvirt.py
@@ -40,6 +40,7 @@ class Libvirt(Plugin, RedHatPlugin, UbuntuPlugin, DebianPlugin):
             "/etc/libvirt/lxc.conf",
             "/etc/libvirt/nwfilter/*.xml",
             "/etc/libvirt/qemu/*.xml",
+            "/etc/libvirt/qemu.conf",
             "/var/run/libvirt/",
             "/etc/libvirt/qemu/networks/*.xml",
             "/etc/libvirt/qemu/networks/autostart/*.xml",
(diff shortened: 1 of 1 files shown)


pmoravec:
Good catch.

Could you please add that file to password obfuscation in `postproc` as well? (I think you should update the loop variable from `xmlfile` to something more general)

sos/plugins/libvirt.py
```diff
@@ -69,12 +70,17 @@ def setup(self):
...
-                self.do_file_sub(
-                    xmlfile,
-                    r"(\s*passwd=\s*')([^']*)('.*)",
-                    r"\1******\3"
-                )
+        # get files
+        config_files = [glob.glob('/etc/libvirt/qemu/*.xml'),
+                        glob.glob('/var/run/libvirt/qemu/*.xml'),
+                        glob.glob('/etc/libvirt/*.conf')]
+
```
bmr-cymru: I'm not keen on this: `config_files` implies a list of config files, but it's not: it's a list of lists or config  files, and then it gets flattened out in the weird looking `y for x in c for y in x`.

Why the extra complexity? Why not just assign `config_files` from the sum (append) of the three glob lists?

zerodayz:
Is there anything else needed @bmr-cymru

psachin:
@zerodayz I think https://github.com/sosreport/sos/pull/904/files is related and already ack'd by @battlemidget & @BryanQuigley

zerodayz:
@psachin that one is wrong imo you are postprocessing just one file.

bmr-cymru:
@zerodayz you're correct regarding the postprocessing, however, the commit message for your change is misleading and it really should be two separate patches:

 * One is a fix for the already-collected `/etc/libvirt/qemu/*.xml` postprocessing but this is not mentioned
 * The second adds the file `qemu.conf`

If you split these up (`git reset`/`git commit --amend`) and update the branch with `git push --force`, I'll get these merged and close the other PR.

I'm also not that keen on all the globbing and list mangling in the `postproc()` method: why not just use the `do_path_regex_sub()` method. It's designed for this type of use, supports regex syntax, and operates on the plugin's collected files list (so it doesn't go back and make more directory lookups for the globbing).

The two `.xml` rules can be done in a single call and the `.conf` one can share the same regex/replace strings via local variables.

bmr-cymru:
I'm going to delete the 2nd hunk of this patch and re-write it:

```diff
@@ -70,10 +71,16 @@ class Libvirt(Plugin, RedHatPlugin, UbuntuPlugin, DebianPlugin):
                 self.add_copy_spec("/proc/%s/%s" % (pid, pf))
 
     def postproc(self):
-        for loc in ["/etc/", "/var/run/"]:
-            for xmlfile in glob.glob(loc + "libvirt/qemu/*.xml"):
+        # get files
+        libvirt_lists = [glob.glob('/etc/libvirt/qemu/*.xml'),
+                         glob.glob('/var/run/libvirt/qemu/*.xml'),
+                         glob.glob('/etc/libvirt/*.conf')]
+
+        # parse files
+        for libvirt_list in libvirt_lists:
+            for libvirt_file in libvirt_list:
                 self.do_file_sub(
-                    xmlfile,
+                    libvirt_file,
                     r"(\s*passwd=\s*')([^']*)('.*)",
                     r"\1******\3"
                 )
```

There is no need to iterate over a list of globs when there is already a `do_path_regex_sub()` method.

PR #1246, opened 2018-03-08
[openstack_*] obfuscate just passwords from connection URIs

pmoravec:
do not obfuscate whole connection URI but just the password there

currently, for a config line:

    connection=mysql+pymysql://nova_api:tYscAjtg6FC4u8rDxDPnXAkkp@1.2.3.4/nova_api?read_default_file=/etc/my.cnf.d/tripleo.cnf&read_default_group=tripleo

sosreport obfuscates to:

    connection=*********

while it should just do:

    connection=mysql+pymysql://nova_api:*********@1.2.3.4/nova_api?read_default_file=/etc/my.cnf.d/tripleo.cnf&read_default_group=tripleo

Resolves: #1246

---

### DIFF
--- a/sos/plugins/openstack_ceilometer.py
+++ b/sos/plugins/openstack_ceilometer.py
@@ -49,20 +49,30 @@ class OpenStackCeilometer(Plugin):
         if self.get_option("verify"):
             self.add_cmd_output("rpm -V %s" % ' '.join(self.packages))
 
+    def apply_regex_sub(self, regexp, subst):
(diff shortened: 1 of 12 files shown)


pmoravec:
Two different regexp need to be applied.

Looking to the pattern of code, can't we unify the copy&paste commands there in some way? (I dont see an easy way, though - maybe extend `plugin` class for these 2 types of regexps?).

Kindly cc @lyarwood for review of changes / testing them (I checked for copy&paste errors several times, but one can't be enough careful in the >500lines patch).

Issue #1251, opened 2018-03-20
[abrt] Collect abrt configuration files


jcastill:
Configuration files in /etc/abrt/ are very useful when investigating why an application core couldn't be created, but unfortunately these files are not collected in sosreport.

PR #1252, opened 2018-03-20
[abrt] Add config files to sosreport

jcastill:
This patch adds the configuration files /etc/abrt/abrt.conf and 
/etc/abrt/abrt-action-save-package-data.conf that were not collected
 in the abrt plugin.

Resolves: #1251

---

### DIFF
--- a/sos/plugins/abrt.py
+++ b/sos/plugins/abrt.py
@@ -41,4 +41,9 @@ class Abrt(Plugin, RedHatPlugin):
         if self.get_option('detailed') and list_file:
             self.info_detailed(list_file)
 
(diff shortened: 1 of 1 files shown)


pmoravec:
+1 for the patch. Doesn't it make sense to collect also `/etc/abrt/plugins`?

jcastill:
I think it makes a lot of sense to collect the plugins as well. Shall I re-work the patch to add this?

bmr-cymru:
:+1:

jcastill:
Apologies guys, I got a trailing whitespace that I missed (basically forgot to run pep8), but sending it again.

PR #643, opened 2015-09-15
[NetworkManager] move NetworkManager-specific content to a new plugin.

robbmanes:
As NetworkManager is being further developed as a stand-alone service,
it is likely that new output will be necessary.  Therefore, further
cluttering the existing networking plugin may be inadvisable and the
two should be separated.

This commit simply removes any NetworkManager-specific output from the
networking plugin and moves it to it's own NetworkManager plugin.  From
there expansion on to the service can be done easily without fear of
disruption of the underlying networking plugin.

Signed-off-by: Robb Manes rmanes@redhat.com

PR #1255, opened 2018-03-30
[networking] Move NetworkManager content to a new plugin

robbmanes:
As NetworkManager is non-essential to configuring a network on most
distributions, and it has its own concepts of how to handle connections
different than the traditional means of ifconfig and ip, move anything
NetworkManager-related to a separate plugin.

Signed off by: Robb Manes <robb.manes@gmail.com>

---

### DIFF
--- a/sos/plugins/networking.py
+++ b/sos/plugins/networking.py
@@ -131,8 +131,6 @@ class Networking(Plugin):
             "/etc/host*",
             "/etc/resolv.conf",
             "/etc/network*",
-            "/etc/NetworkManager/NetworkManager.conf",
-            "/etc/NetworkManager/system-connections",
             "/etc/nftables",
             "/etc/sysconfig/nftables.conf",
             "/etc/nftables.conf",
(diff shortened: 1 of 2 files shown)


robbmanes:
As per the discussion in [PR643](https://github.com/sosreport/sos/pull/643) I hold the opinion of treating NetworkManager as a service and collecting any configuration independent of the network plugin.  Particularly on devices like laptops or teleco systems with many different connections in NetworkManager and hundreds of interfaces, virtual and non-virtual, I think this will cause less clutter in the `networking` directory and create a clear distinction between what is configured on the system as reported by other tooling like `ifconfig` and `ip` versus the configuration NetworkManager currently has, which may vary.  I'm open to suggestions for improvement of this as well.

TurboTurtle:
I like the idea of splitting NM out, +1 on the move.

My only question is, are there distributions that are not shipping NM (or at least, not by default) that the new plugin shouldn't be enabled on (`packages = ('NetworkManager',)`)?

robbmanes:
I actually didn't know about the `package` value.  Agreed ideally we can skip running this entirely if the package isn't present.  Added it to my PR after I verified the package name across RPM and DEB distros is uniformly named `NetworkManager`, thanks for the advice.

PR #1259, opened 2018-04-02
[ovn] Add new plugins for Open Virtual Network

putnopvut:
OVN is a sub-project of Openvswitch used to define logical networks for
a cluster of OVS instances. The two plugins defined here are
"ovn-central", which runs on a single server, and "ovn-host" which runs
on each of the hypervisors running OVS.

These plugins gather runtime information about the configured virtual
networks.

Signed-off-by: Mark Michelson <mmichels@redhat.com>

---

### DIFF
--- a/sos/plugins/ovn_central.py
+++ b/sos/plugins/ovn_central.py
@@ -1,0 +1,105 @@
+# Copyright (C) 2018 Mark Michelson <mmichels@redhat.com>
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+import json
+import os
+import six
+
+
+class OVNCentral(Plugin):
+    """ OVN Northd
+    """
+    plugin_name = "ovn_central"
+    profiles = ('network', 'virt')
+
+    def add_database_output(self, filename, cmds, ovn_cmd, skip=[]):
+        try:
+            with open(filename, 'r') as f:
+                try:
+                    db = json.load(f)
+                except:
+                    # If json can't be parsed, then exit early
+                    self._log_error("Cannot parse JSON file %s" % filename)
+                    return
+                try:
+                    for table in six.iterkeys(db['tables']):
+                        if table not in skip:
+                            cmds.append('%s list %s' % (ovn_cmd, table))
+                except AttributeError:
+                    self._log_error("DB schema %s has no 'tables' key" %
+                                    filename)
+                    return
+        except IOError as ex:
+            self._log_error("Could not open DB schema file %s: %s" % (filename,
+                                                                      ex))
+            return
+
+    def setup(self):
+        ovs_rundir = os.environ.get('OVS_RUNDIR')
+        for pidfile in ['ovnnb_db.pid', 'ovnsb_db.pid', 'ovn-northd.pid']:
+            self.add_copy_spec([
+                os.path.join('/var/lib/openvswitch/ovn', pidfile),
+                os.path.join('/usr/local/var/run/openvswitch', pidfile),
+                os.path.join('/var/run/openvswitch/', pidfile),
+                os.path.join('/run/openvswitch/', pidfile),
+            ])
+
+            if ovs_rundir:
+                self.add_copy_spec(os.path.join(ovs_rundir, pidfile))
+
+        # Some user-friendly versions of DB output
+        cmds = [
+            'ovn-sbctl lflow-list',
+            'ovn-nbctl get-ssl',
+            'ovn-nbctl get-connection',
+            'ovn-sbctl get-ssl',
+            'ovn-sbctl get-connection',
+        ]
+
+        schema_dir = '/usr/share/openvswitch'
+
+        self.add_database_output(os.path.join(schema_dir, 'ovn-nb.ovsschema'),
+                                 cmds, 'ovn-nbctl')
+        self.add_database_output(os.path.join(schema_dir, 'ovn-sb.ovsschema'),
+                                 cmds, 'ovn-sbctl', ['Logical_Flow'])
+
+        self.add_cmd_output(cmds)
+
+        self.add_copy_spec("/etc/sysconfig/ovn-northd")
+
+        ovs_dbdir = os.environ.get('OVS_DBDIR')
+        for dbfile in ['ovnnb_db.db', 'ovnsb_db.db']:
+            self.add_copy_spec([
+                os.path.join('/var/lib/openvswitch/ovn', dbfile),
+                os.path.join('/usr/local/etc/openvswitch', dbfile),
+                os.path.join('/etc/openvswitch', dbfile),
+                os.path.join('/var/lib/openvswitch', dbfile),
+            ])
+            if ovs_dbdir:
+                self.add_copy_spec(os.path.join(ovs_dbdir, dbfile))
+
+        self.add_journal(units="ovn-northd")
+
+
+class RedHatOVNCentral(OVNCentral, RedHatPlugin):
+
+    packages = ('openvswitch-ovn-central', )
+
+
+class DebianOVNCentral(OVNCentral, DebianPlugin, UbuntuPlugin):
+
+    packages = ('ovn-central', )
--- a/sos/plugins/ovn_host.py
+++ b/sos/plugins/ovn_host.py
@@ -1,0 +1,57 @@
+# Copyright (C) 2018 Mark Michelson <mmichels@redhat.com>
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+import os
+from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+
+
+class OVNHost(Plugin):
+    """ OVN Controller
+    """
+    plugin_name = "ovn_host"
+    profiles = ('network', 'virt')
+
+    def setup(self):
+        pidfile = 'ovn-controller.pid'
+        pid_paths = [
+                '/var/lib/openvswitch/ovn',
+                '/usr/local/var/run/openvswitch',
+                '/var/run/openvswitch',
+                '/run/openvswitch'
+        ]
+        if os.environ.get('OVS_RUNDIR'):
+            pid_paths.append(os.environ.get('OVS_RUNDIR'))
+        self.add_copy_spec([os.path.join(pp, pidfile) for pp in pid_paths])
+
+        self.add_copy_spec('/etc/sysconfig/ovn-controller')
+
+        self.add_cmd_output([
+            'ovs-ofctl -O OpenFlow13 dump-flows br-int',
+            'ovs-vsctl list-br',
+            'ovs-vsctl list OpenVswitch',
+        ])
+
+        self.add_journal(units="ovn-controller")
+
+
+class RedHatOVNHost(OVNHost, RedHatPlugin):
+
+    packages = ('openvswitch-ovn-host', )
+
+
+class DebianOVNHost(OVNHost, DebianPlugin, UbuntuPlugin):
+
+    packages = ('ovn-host', )
--- a/sos/plugins/openvswitch.py
+++ b/sos/plugins/openvswitch.py
@@ -28,12 +28,20 @@ class OpenVSwitch(Plugin):
         all_logs = self.get_option("all_logs")
         limit = self.get_option("log_size")
 
+        log_dirs = [
+            '/var/log/containers/openvswitch/',
+            '/var/log/openvswitch/',
+            '/usr/local/var/log/openvswitch/',
+        ]
+
+        if os.environ.get('OVS_LOGDIR'):
+            log_dirs.append(os.environ.get('OVS_LOGDIR'))
+
         if not all_logs:
-            self.add_copy_spec("/var/log/openvswitch/*.log",
-                               sizelimit=limit)
+            self.add_copy_spec([os.path.join(log_dir, '*.log') for log_dir in
+                               log_dirs], sizelimit=limit)
         else:
-            self.add_copy_spec("/var/log/openvswitch/",
-                               sizelimit=limit)
+            self.add_copy_spec(log_dirs, sizelimit=limit)
 
         self.add_copy_spec([
             "/var/run/openvswitch/ovsdb-server.pid",


sos/plugins/ovn_central.py
```diff
@@ -0,0 +1,85 @@
...
+                # information
+                if table != 'Logical_Flow':
+                    cmds.append('ovn-sbctl list %s' % table)
+
+        # Some user-friendly output of DB output
+        cmds.append('ovn-sbctl lflow-list')
+        cmds.append('ovn-nbctl get-ssl')
+        cmds.append('ovn-nbctl get-connection')
+        cmds.append('ovn-sbctl get-ssl')
+        cmds.append('ovn-sbctl get-connection')
```
bmr-cymru: Any reason for making an `append()` call for these static items, rather than simply initialising the list to contain them in the first place? That's more in keeping with existing style in sos, as well as avoiding some (admittedly trivial) runtime cost. (note that the order of the commands in the list passed to `add_cmd_output()` has no significance).

sos/plugins/ovn_host.py
```diff
@@ -0,0 +1,56 @@
...
+        else:
+            self.add_copy_spec("/var/log/openvswitch/",
+                               sizelimit=limit)
+
+        self.add_copy_spec('/var/run/openvswitch/ovn-controller.pid')
+
+        self.add_cmd_output('ovs-ofctl -O OpenFlow13 dump-flows br-int')
+        self.add_cmd_output('ovs-vsctl list-br')
+        self.add_cmd_output('ovs-vsctl list OpenVswitch')
+
```
bmr-cymru: Why not use a single `add_cmd_output()` call here? Since there are no differences in kwargs passed to the method you can just use an in-place list definition here.

Similarly the two `add_copy_spec()` calls around these can be merged into a single call.

sos/plugins/ovn_central.py
putnopvut: Thanks, I'll make this change.

sos/plugins/ovn_host.py
putnopvut: I'll make these changes as well. Thanks for the quick feedback on my patch.

bmr-cymru:
Thanks! :+1:

pmoravec:
ACK

sos/plugins/ovn_central.py
```diff
@@ -25,29 +25,47 @@ class OVNCentral(Plugin):
...
+        # If the user specified a plugin option, definitely
+        # use that value.
+        our_dir = self.get_option(plugin_var)
+        if our_dir:
+            return our_dir
+
+        for candidate in candidates:
+            if candidate and os.path.exists(candidate):
+                return candidate
+
```
pmoravec: Missing return value if out_dir = '' and no candidate matches - then None is returned, what raises exception in `os.path.join(None,...)`

sos/plugins/ovn_central.py
```diff
@@ -0,0 +1,122 @@
...
+    def get_dir(self, plugin_var, *candidates):
+        # If the user specified a plugin option, definitely
+        # use that value.
+        our_dir = self.get_option(plugin_var)
+        if our_dir:
+            return our_dir
+
+        for candidate in candidates:
+            if candidate and os.path.exists(candidate):
+                return candidate
```
pmoravec: Does this mean just the very first candidate log directory will be collected? Dont you want to collect all of them? (and why it could be an issue to collect all of them? why dont simply call:

    self.add_copy_spec([log_dir,
                                 os.environ.get('OVS_LOGDIR'),  # works well even if the var isnt set and the cmd returns None
                                 '/var/log/containers/openvswitch',
                                 '/var/log/openvswitch',
                                 '/usr/local/var/log/openvswitch'
                                ])

sos/plugins/ovn_host.py
```diff
@@ -0,0 +1,87 @@
...
+    """ OVN Controller
+    """
+    plugin_name = "ovn_host"
+    profiles = ('network', 'virt')
+    option_list = [
+        ('log_dir', 'Directory where OVN logs are found', '', ''),
+        ('runtime_dir', 'Directory where OVN runtime files are found', '', ''),
+    ]
+
+    def get_dir(self, plugin_var, *candidates):
```
pmoravec: same comments from `ovn_central` plugin apply here.

pmoravec:
See https://github.com/pmoravec/sos/tree/sos-pmoravec-ovn-central-host / https://github.com/pmoravec/sos/commit/c4fcbdf3580050400eeaa1e265ec659bc89123cf . (also worth before merging stuff like this to squash commits into one / one per plugin)

putnopvut:
@bmr-cymru Just to be clear, are you nacking the commits I added to the pull request or are you nacking the commits that @pmoravec linked to? Or both?

putnopvut:
I've created a new version of this changeset that gets rid of plugin variables and incorporates @pmoravec's changes, along with a few more improvements. I want to take @pmoravec's advice of squashing the commits down to a single one. However, since I have already pushed this set of changes to my repo, I think it means I'll have to create a separate remote branch, and also a new pull request. I'm not sure what the etiquette is with these sorts of things on github, so is it all right that I'll be creating a separate pull request for the same thing, or is there a better way of doing this that I'm unaware of? Thanks!

pmoravec:
Thanks for the concern. It is fully OK to overwrite git history of your local branch and push it (with --force option) here: the PR will be automatically updated.

bmr-cymru:
Yep: no need for a new branch/PR. Just perform the edits on your local branch, and then force update the remote HEAD with `git push --force`. This overwrites the remote branch with your new changes (but you can still fish the old one out if it's ever needed via `git reflog`).

numansiddique:
Any idea when this pull request be accepted ? We need this for https://bugzilla.redhat.com/show_bug.cgi?id=1560845 and https://bugzilla.redhat.com/show_bug.cgi?id=1560849

Thanks
Numan

pmoravec:
Seems OK for me, though few minor improvements could be made, like:

- remove `self.add_cmd_output(cmds)` and replace `cmds = ..` & `cmds.append(..)` by directly `self.add_cmd_output(..)` - as `add_cmd_output` can be called any number of times in a plug-in, both for a single command and/or for a tuple of them

- the `try` .. `except` block might be split to two - if the 1st file does not exist but 2nd does, let collect commands outputs triggered from the 2nd file

- is the `json.load` robust enough in case invalid json format provided? Or wouldnt an uncaught exception be raised (so it is worth adding that exception type to the `except` statement)?

- `os.path.join('/var/lib/openvswitch/ovn', 'ovn-controller.pid'),` and similar - why dont provide simply `/var/lib/openvswitch/ovn/ovn-controller.pid` there? Or if the pidfile shall be dealt as a possible future option (or iterable variable), let call e.g.:

```
    pidfile = 'ovn-controller.pid'
    ..
    os.path.join('/var/lib/openvswitch/ovn', pidfile
```

bmr-cymru:
i'm not really bothered by the command list building - it's a legitimate way to use the interface & I don't see any benefit of using multiple `add_cmd_output()` calls (if anything, the fewer calls the better as it allows us more opportunity in the future to make global decisions about commands).

The exception handling does need a re-think: both to split it up for the two distinct exception paths, and also to handle JSON exceptions better: it looks at the moment like the first missing key will abort further processing of that file, which is probably not what is wanted. Hardening for other possible exceptions may also be worthwhile (I haven't checked what can be thrown yet).

The list passed to `add_copy_spec()` for the pidfiles does look a bit odd: since we're consing the same file name onto every entry it can be rewritten as a list comprehension:

```python
    pidfile = 'ovn-controller.pid'
    pid_paths = [
            '/var/lib/openvswitch/ovn',
            '/usr/local/var/run/openvswitch',
            '/var/run/openvswitch',
            '/run/openvswitch'
    ]
    if os.environ.get('OVS_RUNDIR'):
        pid_paths.append(os.environ.get('OVS_RUNDIR')
    self.add_copy_spec([join(pp, pidfile) for pp in pid_paths])
```

bmr-cymru:
I'll try to make a more thorough review of this in the next day or two, but a few initial comments:

* The proliferation of openvswitch log collection is not acceptable. There is an existing plugin designed to collect this.

* The path options are still problematic (even if collection is guarded by some `*.log` globs and size limiting). There should be no need for this on a system using packaged software.

* The plugin needs to have at least _some_ triggers. File, package, command, and kernel module are all acceptable but this plugin should not be "default on" since it is not applicable to the vast majority of systems.

It's not clear from anything in the comments or commit messages to indicate the real relationship between OVN and the existing openvswitch support in `sos`. They apparently have some shared state directories in `/var`: if they are that closely integrated then it may be that some of these changes should be merged into that plugin.

putnopvut:
Hi Bryn, thanks for the feedback. Unfortunately, I'm not really sure I fully understand the critiques on your three bullet points.

- I'm not sure what you mean by "proliferation of openvswitch log collection". When you say there is an existing plugin, are you referring to the openvswitch plugin?
- I removed the path options in the latest change to the pull request. You can no longer specify "-k <some_path>" when running the OVN plugins. Or were you referring to something else?
- In sosreport terminology, I'm not familiar with what a "trigger" is or how it is defined. I created the Red Hat and Debian variant plugin classes where a package name is specified. I used the openvswitch plugin as a guide here. If I've not done it properly, can you point me to something that does it properly?

As for merging with the existing openvswitch plugin, I can understand that. Sourcewise, OVN is a sub-project in the same git repo as openvswitch and relies on the same common libraries. It uses some of the same paths that openvswitch uses (e.g. logging and runtime information). OVN packages are provided separately from openvswitch, but those OVN packages are dependent on openvswitch. So any system that has OVN also has openvswitch. I'm not sure what the general rules are with sosreport. Is it typically that a plugin corresponds directly with a package? If so, then that's what I've got going right now. Or can a plugin more loosely account for a collection of related packages? If so, can you point me to an example of one that works this way? One issue that still arises is that I think OpenStack doesn't get much use out of file collection in the current openvswitch plugin because its hardcoded paths are not the same as what OpenStack uses in its installations.  Path listing similar to what is used in the current OVN plugins (or the creation of a RedHatOpenStackPlugin class as was suggested earlier) would be necessary still.

pmoravec:
**Re plugin trigger** - each plugin is by default enabled if either:

- any package from `packages` list is installed, or
- any file from `files` list is present

See e.g. https://github.com/sosreport/sos/blob/master/sos/plugins/abrt.py#L26-L27 .

Anyway, I see you have enabled the plugins via child classes `RedHatOVNCentral` and `DebianOVNCentral` and these triggers seem OK to me - if (and only if) the package from such child class is present, the plugin is enabled by default. Verified also on a test system. So this seems OK to me, no @bmr-cymru ?


**Re proliferation of openvswitch logs** - yes, these logs are suggested to be collected twice, by two different plugins. Whereas sosreport's approach is to have independent plugins, each responsible for collecting a unique data. But per my understanding:

- `openvswitch-ovn-central` package for RHEL requires `openvswitch` while `ovn-central` for Debian requires `openvswitch-switch` package to be installed
- therefore, by default, `ovn_central` plugin will be enabled *only* when `openvswitch` plugin will be enabled, and this plugin already collects (all?) openvswitch logs
- same applies for `ovn_host` plugin that will be (by default) enabled only if `openvswitch` plugin is enabled

So for these two plugins, you (as a future "user" of them) dont need to bother with collecting the openvswitch logs, since `openvswitch` plugin will collect them.

Anyway it is fully legitimate to collect some plugin specific logs, if there are any. But as I see both plugins collect `/var/log/containers/openvswitch/`, `/var/log/openvswitch/` and `/usr/local/var/log/openvswitch/`, shall not rather the "parent" `openvswitch` plugin collect all of these?

I would suggest updating `openvswitch` plugin by the code https://github.com/sosreport/sos/pull/1259/files#diff-7e5b784f4c6844be676647d273762167R53-R63, does it make sense?

(or doesnt it make sense to have all OVN functionality in one plugin that is triggered by either package? sosreport is OK in calling nonexisting commands and collecting not-present files..)

**Re collecting arbitrary path** - the concern is that `OVS_LOGDIR` env.variable can be (theoretically, by a typo or accident) set up incorrectly. For example to `/` (such errors really can happen). Then the plugin will collect **everything**, esp. when called with `--all-logs` option.

bmr-cymru:
> * any package from packages list is installed, or
> * any file from files list is present

Missing two others:

* An executable command in the commands list is found in PATH
* A kernel module named in kernel_mods is found to be loaded

bmr-cymru:
For the openvswitch logs, the problem is this:

```
$ git grep 'var.*log.*openvswitch'
sos/plugins/openvswitch.py:            self.add_copy_spec("/var/log/openvswitch/*.log",
sos/plugins/openvswitch.py:            self.add_copy_spec("/var/log/openvswitch/",
sos/plugins/ovn_central.py:            '/var/log/containers/openvswitch/',
sos/plugins/ovn_central.py:            '/var/log/openvswitch/',
sos/plugins/ovn_central.py:            '/usr/local/var/log/openvswitch/',
sos/plugins/ovn_host.py:            '/var/log/containers/openvswitch/',
sos/plugins/ovn_host.py:            '/var/log/openvswitch/',
sos/plugins/ovn_host.py:            '/usr/local/var/log/openvswitch/',
```

So we have the same set of paths now being collected in three different plugins - as @pmoravec explained, the `openvswitch` plugin already collects these files. Either the OVN plugin(s) should become a part of the `openvswitch` plugin if they are so intimately related that they cannot be separated, or the OVN bits should collect only the files that differentiate the two.

bmr-cymru:
I don't see the path options now - most likely I had an old version of the diff open at the time.

bmr-cymru:
> Re collecting arbitrary path - the concern is that OVS_LOGDIR env.variable can be (theoretically, by a typo or accident) set up incorrectly.

I'm not so worried by environment variables - since they are defined by something outside of sos. The problem specifically with plugin options is that we don't have a lot of space to explain them or describe how to use them. With environment options defined by other projects that problem's no longer on our heads.

bmr-cymru:
For triggers it's as @pmoravec mentioned - paths, package names, executable command names, or kernel modules that wen present should cause this plugin to run. This means that the plugin only runs when it should, and avoids needlessly burning cycles checking for files and commands that are not there.

We prefer to have at least one trigger for the vast majority of plugins - with over 250 in the tree it's really best that they only run when they have a job to do.

It can be tricky getting triggers and the division of labour right between sos plugins, especially for components like OVN that are closely related to other infrastructure pieces but I'm sure we can find a solution here that will work well.

bmr-cymru:
Eh, nevermind - I obviously had an outdated version of the commit open in a tab yesterday. The triggers are there, and are implemented correctly:

```python
class RedHatOVNCentral(OVNCentral, RedHatPlugin):

    packages = ('openvswitch-ovn-central', )


class DebianOVNCentral(OVNCentral, DebianPlugin, UbuntuPlugin):

packages = ('ovn-central', )
```

Etc.

So really all we need to do now is sort out who is responsible for the main `openvswitch` logs: since they're already collected by the `openvswitch` plugin, and since that should be enabled wherever these new plugins are, I think we can just remove it from the pull & merge the rest.

putnopvut:
Thanks for all your feedback. I'll update the pull request as soon as possible.

putnopvut:
I've addressed the comments. Due to the changes requested in the openvswitch plugin, it felt natural to push this as two separate commits.

pmoravec:
Sounds OK to me. Just see formatting / pep8 failures e.g. in https://travis-ci.org/sosreport/sos/jobs/380413370

bmr-cymru:
> it felt natural to push this as two separate commits.

Yep. That's the right approach if there isn't a strong logical connection (like say, you're moving collection from one plugin into another).

The `openvswitch` changes look fine - I don't see any problem using the OVS environment variables there since as you explained the projects are so closely entwined.

bmr-cymru:
```python
            self.add_copy_spec([os.path.join(log_dir, '*.log') for log_dir in
                               log_dirs], sizelimit=limit)
            self.add_copy_spec([os.path.join(log_dir, '*.log') for log_dir in
                               log_dirs], sizelimit=limit)
```

Missing an import for `os` here - I'll fix this up as I merge (and try to trim the list comprehensions to fit on a single line).

bmr-cymru:
Can we get a list of package names for the supported distributions for these components?

I missed this during the reviews, but we really need to add these or they will run in every environment.

Also, if you can suggest the same for openvswitch that would be very useful.

putnopvut:
I'm a bit confused. I've already specified package names in the RedHatPlugin, DebianPlugin, and UbuntuPlugin classes in the pull request.

But just to be clear, for the ovn_central plugin, the Red Hat package is openvswitch-ovn-central. The Debian/Ubuntu package is ovn-central. For the ovn_host plugin, the Red Hat package is called openvswitch-ovn-host. The Debian/Ubuntu package is ovn-host.

Similarly, openvswitch already specifies package names in the same way. The Red Hat packages are openvswitch and openvswitch-dpdk. The Debian/Ubuntu package is openvswitch-switch

bmr-cymru:
Nevermind - just me glancing at the top of the files & not seeing them in the base class. It's still relatively rare to have differentiated tagging classes, so it's easy to overlook (and I've been working through ~130 pulls in the last few weeks - it's easy for things to fall out of working memory ;-) ).

PR #1262, opened 2018-04-03
[infiniband] collect info from all active ports of all IB devices

pmoravec:
currently just the very first active port is checked

Resolves: #1262

---

### DIFF
--- a/sos/plugins/infiniband.py
+++ b/sos/plugins/infiniband.py
@@ -14,6 +14,7 @@
 # with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
+import os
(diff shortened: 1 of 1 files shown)


pmoravec:
Thanks Hongang Li for initial version of the patch.

bmr-cymru:
I'm not really in favour of this approach if the root cause is a deficiency in the tools that we are calling.

If this is confusing to users of `sos`, how can it not be confusing for users who are running these tools themselves?

pmoravec:
I dont know if/how confusing the behaviour is, or if it is worth fallback behaviour when not providing options. See e.g. https://linux.die.net/man/8/infiniband-diags and "Multiple port/Multiple CA support:" paragraph.

It makes sense to collect that info from all active ports - it is up to discussion if `sos` or infiniband tools shall provide the "get data from all ports" feature.

Issue #1266, opened 2018-04-10
bug in _collect_strings causes error trying to str.decode()


ddstreet:
In add_string_as_file and _collect_strings, the first line of each string
is printed to debug log, however there is a bug in _collect_strings that
tries to decode the first line of '...' instead of the actual string;
this causes a error like:

  File "/usr/share/sosreport/sos/sosreport.py", line 1300, in collect
    plug.collect()
  File "/usr/share/sosreport/sos/plugins/__init__.py", line 877, in collect
    self._collect_strings()
  File "/usr/share/sosreport/sos/plugins/__init__.py", line 860, in _collect_strings
    (content.splitlines()[0]).decode('utf8', 'ignore'))
AttributeError: 'str' object has no attribute 'decode'

TurboTurtle:
Closing per 476bd7f

PR #1267, opened 2018-04-10
[Plugin] Fix string decoding for debug log output

ddstreet:
In add_string_as_file and _collect_strings, the first line of each string
is printed to debug log, however there is a bug in _collect_strings that
tries to decode the first line of '...' instead of the actual string;
this causes a error like:

  File "/usr/share/sosreport/sos/sosreport.py", line 1300, in collect
    plug.collect()
  File "/usr/share/sosreport/sos/plugins/__init__.py", line 877, in collect
    self._collect_strings()
  File "/usr/share/sosreport/sos/plugins/__init__.py", line 860, in _collect_strings
    (content.splitlines()[0]).decode('utf8', 'ignore'))
AttributeError: 'str' object has no attribute 'decode'

This simplifies the (string or bytearray)-to-first-line-string of both
functions, which also fixes the bug using the wrong variable to decode.

Fixes: 7ba2fb47d05b608f3863fda9271894e38b43937a
Signed-off-by: Dan Streetman <ddstreet@canonical.com>

---

Note - commit msg is line-split and wrapped at 72 chars, except for the python backtrace.  I edited the checklist to indicate all X's, since I assume the python backtrace is exempted from the 72-char line limit

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -709,12 +709,10 @@ class Plugin(object):
     def add_string_as_file(self, content, filename):
         """Add a string to the archive as a file named `filename`"""
         self.copy_strings.append((content, filename))
-        if isinstance(content, six.string_types):
-            content = "..." + content.splitlines()[0]
-        else:
-            content = ("..." +
-                       (content.splitlines()[0]).decode('utf8', 'ignore'))
-        self._log_debug("added string '%s' as '%s'" % (content, filename))
+        content = content.splitlines()[0]
+        if not isinstance(content, six.string_types):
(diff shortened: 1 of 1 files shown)


ddstreet:
this fixes #1266

slashdd:
Commit bf8433945aeced7c0d73686ec83d826cc33bbd28 fixes the situation for me on my reproducer.

bmr-cymru:
:+1: thanks!

pmoravec:
+1 good fix.

bmr-cymru:
Merged, thanks

PR #1268, opened 2018-04-13
[kernel] Disable gathering /proc/timer* statistics by default

rmetrich:
Disable gathering `/proc/timer*` statistics by default, a new option 'kernel.with-timer' enables gathering these.

If `/proc/timer_list` is huge, then kernel will experience issues with processing all the timers since it needs to spin in a tight loop inside the kernel.

We have tried to fix it from kernel side, added `touch_nmi_watchdog()` to silence softlockups, `cond_resched()` to fix RCU stall issue but with such huge number of timers the RHEL7 kernel is still hangs.
It can reproduced somehow on upstream kernel (however, there will be workqueue lockups).

We came to conclusion that reading `/proc/timer_list` should be disabled in sosreport.
Since `/proc/timer_stats` is tight to `/proc/timer_list`, both are disabled at the same time.

---

### DIFF
--- a/sos/plugins/kernel.py
+++ b/sos/plugins/kernel.py
@@ -27,6 +27,10 @@ class Kernel(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
 
     sys_module = '/sys/module'
 
+    option_list = [
+        ("with-timer", "gather /proc/timer* statistics", "slow", False)
+    ]
+
     def setup(self):
         # compat
         self.add_cmd_output("uname -a", root_symlink="uname")
@@ -83,7 +87,6 @@ class Kernel(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
             "/proc/driver",
             "/proc/sys/kernel/tainted",
             "/proc/softirqs",
-            "/proc/timer*",
             "/proc/lock*",
             "/proc/misc",
             "/var/log/dmesg",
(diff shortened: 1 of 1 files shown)


bmr-cymru:
Is anything exposed that could guide this collection automatically? I.e. is there a way to obtain a count of the timers without walking the list and running into the contention problems?

bmr-cymru:
(It's very unfortunate to add a plugin option to work around a kernel bug: once it's in there, it's very hard to remove, or to ever make default again: we'd need to implement cross-distribution kernel version checks once known fixed versions are available).

sos/plugins/kernel.py
```diff
@@ -92,4 +95,11 @@ def setup(self):
             clocksource_path + "current_clocksource"
         ])
 
+        if self.get_option("with-timer"):
+            # This can be very slow, depending on the number of timers,
+            # and may also cause softlockups
+            self.add_copy_spec([
+                "/proc/timer*"
+            ])
```
bmr-cymru: Please don't use a list for a single item or break the call over multiple lines: this can simply be written:

```python
        self.add_copy_spec("/proc/timer*")
```

rmetrich:
Currently asking my kernel colleagues for how to get the timers count (if any).

bmr-cymru:
It doesn't look good: there's nothing obvious in `/sys` or `/proc` at all, and obviously no syscall to obtain this.

rmetrich:
From my understanding, nobody (kernel guys) cares about /proc/timer* stuff.

bmr-cymru:
Thanks for respinning. In my ten years in the support engineering group I don't recall ever needing to look at this file. If timers came up it was typically a crash (often in a 3rd party module), and debugging proceded via vmcore analysis.

I think given the fact this also affects upstream and there doesn't seem to be much hope of a solution in the near term this is the best we can do for now.

Thanks for respinning the patch: I'll get it merged before we release 3.6 (next week or two).

PR #1269, opened 2018-04-16
dmesg: print logs with human-readable timestamps

pkalever:
Signed-off-by: Prasanna Kumar Kalever <prasanna.kalever@redhat.com>

### DIFF
--- a/sos/plugins/kernel.py
+++ b/sos/plugins/kernel.py
@@ -51,7 +51,7 @@ class Kernel(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
             extra_mod_paths.extend(glob.glob(pattern))
 
         self.add_cmd_output([
-            "dmesg",
+            "dmesg -T",
             "sysctl -a",
             "dkms status",
             "find %s -ls" % " ".join(extra_mod_paths)


lxbsz:
It looks fine for me. And this is really very useful for us in some cases.

bmr-cymru:
Please follow the [Contributor Guidelines](https://github.com/sosreport/sos/wiki/Contribution-Guidelines) for patch and commit message formatting: your commit subject should have a `[<component>]` prefix inside square brackets, not the file or command being collected.

More generally I'm not sure I see a great benefit to this. Where logs are already (or by default) formatted with human readable dates we collect those, but we typically favour the basic representation when it is available - it is easier to convert from that format to human readable than vice-versa.

bmr-cymru:
> this is really very useful for us in some cases.

How come? The raw timestamps are there: it's a trivial act of scripting to format those into a given date format and timezone.

bmr-cymru:
It's just a case of filtering the data through something that performs the date conversion: classically it'd probably be an `awk` script modifying field zero by passing it through the `date` program. These days Python or Perl would be a good option.

You need to take the `btime` field from `/proc/stat` (collected by the `general` plugin), and add this to each time value reported in the `dmesg` log, and then pass the result to `date --date` to perform the conversion into whatever format and timezone you would like.

pmoravec:
Not a kernel guy, but from manpages

**Be aware that the timestamp could be inaccurate!**  The **time** source used for the logs is **not updated after** system **SUSPEND/RESUME**.

So if I pause and resume a VM, won't I get wrong timestamps in `dmesg -T` output?

bmr-cymru:
Hmm: that is a big reason that I do not like this change. I think here we need to leave the raw values alone: they may require more interpretation by the analyst, but they are not going to introduce uncertainties in the data.

It's a shame that `dmesg` itself cannot post-process these files to reformat the timestamp. That would be far preferable to performing it at collection time (but is still fairly easy to script around, with the caveat that this conversion also suffers from inaccuracy around suspend/resume).

PR #1271, opened 2018-04-18
Collect install-undercloud.log for a containerized undercloud

EmilienM:
When containerizing the undercloud, we're moving install-undercloud.log
into $HOME which makes it more accessible for the users, while we're
removing instack(-undercloud) bits over time.

This patch add $HOME/install-undercloud.log file and document why we're
doing this. Note that we still collect old file for previous releases.


EmilienM:
I'm doing it on one single PR: https://github.com/sosreport/sos/pull/1279

PR #1274, opened 2018-04-23
[azure] Added /var/log/azure directory to azure plugin

davemulford:
---


pmoravec:
See https://github.com/sosreport/sos/blob/master/sos/plugins/__init__.py#L536 - you can specify the limit via `size_limit` option, like:

    self.add_copy_spec("/var/log/azure/", sizelimit=10)

(it also makes sense to use global `all_logs` or `log_size` (default 10MB) options like https://github.com/sosreport/sos/blob/master/sos/plugins/cups.py#L28-L34 does)

bmr-cymru:
Please also follow the [contributor guidelines](https://github.com/sosreport/sos/wiki/Contribution-Guidelines) for patch formatting: the subject needs to be in `[<component>] change description` format, and all contributions require a DCO ("Signed-off-by") in the commit body. For this it'd be something like `[azure] add log files from /var/log/azure` (once the `--log-size` and `--all-logs` options are hooked up).

sos/plugins/azure.py
```diff
@@ -33,10 +33,11 @@ def setup(self):
             "/etc/default/kv-kvp-daemon-init",
             "/etc/waagent.conf",
             "/sys/module/hv_netvsc/parameters/ring_size",
-            "/sys/module/hv_storvsc/parameters/storvsc_ringbuffer_size",
-            "/var/log/azure*"
+            "/sys/module/hv_storvsc/parameters/storvsc_ringbuffer_size"
         ])
 
+        self.add_copy_spec("/var/log/azure*", sizelimit=10)
```
bmr-cymru: This will not work if `/var/log/azure*` expands to a directory (`/var/log/azure/`). You need to use a glob pattern that will match a list of log files (e.g. `*.log` is common) to have the size limit effectively applied. There's more information on this and the other global log limiting options in the GitHub wiki.

bmr-cymru:
Also need to squash the two commits together before we can merge: you can use `git rebase -i` with the `squash` command, or `git reset` / `git commit --amend` and then `git push --force`.

sos/plugins/azure.py
davemulford: The `/var/log/azure*` glob does expand to a directory that will contain other directories and/or files. The Azure extensions provided perform their own logging under this directory, and don't always follow a common filename pattern.

I think deleting this PR is probably the best option here, but let me know your thoughts.

bmr-cymru: We can still limit them (assuming they are in a flat directory structure - i.e. no more subdirectories beneath `/var/log/azure`) - the glob just needs to use `/var/log/azure/*` in that case (all the files within the directory).

The lack of any pattern or structure to the names does mean that we cannot properly implement `--all-logs` for this plugin, since there is no way for us to tell "current" logs apart: the option is mainly used with traditional UNIX-style log rotation, where "old" logs have an appended counter or date string.

What will happen today in the event that the limit is reached is that we will collect logs in most recently written order up until the limit is reached: this is still useful in many cases since the most recently written data is often of most interest, so even though we can't implement everything for Azure I still think it's worth going ahead and adding this.

davemulford: Thanks for that explanation! That explains why I wasn't seeing the `sizelimit` working when testing the plugin.

The /var/log/azure directory will not be a flat structure, unfortunately. If it is helpful, I can create a test virtual machine and install some plugins to show you what I'm working with.

bmr-cymru: Thanks - that'd definitely be useful. The other (more complex, but sometimes more effective) approach is to see if there is something in the Azure environment that can be interrogated to get a list of log files, and that we can use to prioritise or select them for collection. It's uncommon today for plain logs but we do have a couple of other bits of data collection that work this way (e.g. parse a list of things to collect from some command output, and then generate a list of files to collect from that).

bmr-cymru: Side note: we're just setting up a meeting with some of the core folks to address outstanding issues in the content size limit handling in sos. Some of this may land in time for the next release - the main work at the moment is introducing size limiting for command output, and making sure that command and file size limits behave consistently.

As part of this work I'd like to see if we can tackle some of the long-standing limitations in the general `add_copy_spec()` size limiting: some of those changes may help with this particular use case.

davemulford: Sorry, just now getting time to come back to this. Below is an example of the Azure Recovery Services extension logging. The directory depth is indeterminate, and log files don't follow normal filename conventions (e.g., `*.log`).

~~~
$ sudo find /var/log/azure -type f
/var/log/azure/Microsoft.Azure.RecoveryServices.SiteRecovery.Linux/1.0.0.9100/SiteRecoveryExtension
/var/log/azure/Microsoft.Azure.RecoveryServices.SiteRecovery.LinuxRHEL7/1.0.0.9100/SiteRecoveryExtension
~~~

As a temporary workaround, would it be possible to perform a `find /var/log/azure -type f` as shown above, then loop through all the results and add them individually via `add_copy_spec()`?

The work you talk about with `add_copy_spec()` and size limiting does sound promising, so if you think it's worth waiting on that, I certainly can.

davemulford:
I just squashed the commits on what I feel is a decent workaround for what I was trying to accomplish in the beginning. Thanks for all the help and guidance!

Is there a way to request another code review?

sos/plugins/azure.py
```diff
@@ -16,6 +16,7 @@
 
 from sos.plugins import Plugin, UbuntuPlugin, RedHatPlugin
 import os
+import sys
```
bmr-cymru: You're importing `sys`, but you don't use it anywhere...

bmr-cymru:
@davemulford - you did :-) Just took a little while as we had a large backlog outstanding.

bmr-cymru:
Think you need to do `git fetch master`,`git rebase master` and a `git push --force` there bud ;-)

davemulford:
Thanks! I thought I had done it properly, but I'm just breaking things at this point.

bmr-cymru:
Hi @davemulford you seem to have a bit of a problem with this branch - you're re-pushing a branch that contains ~107 commits that are already in master, plus your changes, plus a merge commit (to join the history of master and your branch).

What you need to do here is to rebase your branch to that all of the already upstream commits appear first, and then just your changes remain at the end.

This should in most cases be automatic if you just run `git rebase master` (so long as there are no conflicts between the base branch and your changes). You can then re-push the commits with `git push --force`.

It looks like the push you did a few moments ago just had the same 108 commits as the previous one.

This:
```
Merge branch 'master' of https://github.com/sosreport/sos d0f9a03
```

Needs to go from the history, and the branch should just look like this:

```
[azure] Added /var/log/azure directory to sosreport 2179ebe
```

bmr-cymru:
Crossing messages :)

If you don't get it fixed up give me a shout next week - we can get together on Freenode IRC or discuss via email if that's easier.

bmr-cymru:
Your commits are all there so nothings lost - we can get it sorted out

davemulford:
Thanks for your understanding and kindness. I think I understand what's needed and will try this out today. If not, I'll get in touch next week. Thanks again!!

bmr-cymru:
Sorry, I'm way behind on the pull request backlog, so this time I will sort this out myself - in future, please do not merge from sos master into your development branch. This creates a huge mess of conflicting commits that the maintainer has to untangle to be able to merge your changes:

```
$ git rebase master
First, rewinding head to replay your work on top of it...
Applying: [azure] Added /var/log/azure directory to sosreport
Using index info to reconstruct a base tree...
M	sos/plugins/azure.py
Falling back to patching base and 3-way merge...
Auto-merging sos/plugins/azure.py
Applying: [yum] fix yum plugin data collection
Using index info to reconstruct a base tree...
M	sos/plugins/yum.py
Falling back to patching base and 3-way merge...
No changes -- Patch already applied.
Applying: [Plugin] allow lists to be passed to add_forbidden_path()
Using index info to reconstruct a base tree...
M	sos/plugins/__init__.py
Falling back to patching base and 3-way merge...
Auto-merging sos/plugins/__init__.py
CONFLICT (content): Merge conflict in sos/plugins/__init__.py
error: Failed to merge in the changes.
```

... lather, rinse, repeat for every patch that appears in the commit list above.

Instead of merging, you want to update your master branch (`git checkout master; git pull`, or use fetch and merge manually), and then `git rebase` your branch on top of it. That will give you just your commits on top of a clean upstream master and makes the merge operation much more straightforward.

PR #1277, opened 2018-04-28
[cs] Updates for CS 9

83problems:
Updated header with new format.

pki-base is a common package installed with any other pki- related
package.

Removed dirsrv entries since ds plugin handles those.

Created for loop to get CA names from /var/lib/pki/ since any name
can be used besides the default. Replaced the * glob.

Sosreport will ignore password.conf and key3.db for CS 9.

Certutil will print certificates in nssdb.

Grabs more log files, spawn and destroy logs, and all configuration
files.

Obfuscates passwords in CA's server.xml and tomcat-users.xml.

Signed-off-by: Brian Gribble <bgribble@redhat.com>

---

### DIFF
--- a/sos/plugins/cs.py
+++ b/sos/plugins/cs.py
@@ -1,60 +1,55 @@
-# Copyright (C) 2007-2010 Red Hat, Inc., Kent Lamb <klamb@redhat.com>
-#                                        Marc Sauton <msauton@redhat.com>
-#                                        Pierre Carrier <pcarrier@redhat.com>
-
-# This program is free software; you can redistribute it and/or modify
-# it under the terms of the GNU General Public License as published by
-# the Free Software Foundation; either version 2 of the License, or
-# (at your option) any later version.
-
-# This program is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU General Public License for more details.
-
-# You should have received a copy of the GNU General Public License along
-# with this program; if not, write to the Free Software Foundation, Inc.,
-# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+# Copyright (C) 2018 Brian Gribble <bgribble@redhat.com>, Red Hat.
+# All rights reserved.
+#
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions
+# of the version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further
+# information.
 
 from sos.plugins import Plugin, RedHatPlugin
 from os.path import exists
 from glob import glob
+import os
 
 
 class CertificateSystem(Plugin, RedHatPlugin):
-    """Certificate System and Dogtag
+    """ Certificate System and Dogtag
     """
 
     plugin_name = 'cs'
     profiles = ('identity', 'security')
 
     packages = (
-        "redhat-cs",
-        "rhpki-common",
-        "pki-common",
-        "redhat-pki",
-        "dogtag-pki",
-        "pki-base"
+        'redhat-cs',
+        'rhpki-common',
+        'pki-common',
+        'pki-base'
     )
 
     files = (
-        "/opt/redhat-cs",
-        "/usr/share/java/rhpki",
-        "/usr/share/java/pki"
+        '/opt/redhat-cs',
+        '/usr/share/java/rhpki',
+        '/usr/share/java/pki'
     )
 
     def checkversion(self):
-        if self.is_installed("redhat-cs") or exists("/opt/redhat-cs"):
+        """ Check if Certificate System 7, 8, or 9 is installed.
+        """
+
+        if self.is_installed('redhat-cs') or exists('/opt/redhat-cs'):
             return 71
-        elif self.is_installed("rhpki-common") or \
-                len(glob("/var/lib/rhpki-*")):
+        elif self.is_installed('rhpki-common') or \
+                len(glob('/var/lib/rhpki-*')):
             return 73
-        # 8 should cover dogtag
-        elif self.is_installed("pki-common"):
+        # 8 should cover Dogtag.
+        elif self.is_installed('pki-common'):
             return 8
-        elif self.is_installed("redhat-pki") or \
-                self.is_installed("dogtag-pki") or \
-                self.is_installed("pki-base"):
+        # pki-base is the common package for 9.
+        elif self.is_installed('pki-base'):
             return 9
         return False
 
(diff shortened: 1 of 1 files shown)


83problems:
I have attached an sosreport from the update to the cs plugin.
[sosreport-cs.tar.gz](https://github.com/sosreport/sos/files/1957986/sosreport-cs.tar.gz)

sos/plugins/cs.py
```diff
@@ -1,128 +1,140 @@
-# Copyright (C) 2007-2010 Red Hat, Inc., Kent Lamb <klamb@redhat.com>
-#                                        Marc Sauton <msauton@redhat.com>
-#                                        Pierre Carrier <pcarrier@redhat.com>
```
bmr-cymru: You can't just nuke everyone else's copyright. I get that there is substantive churn in this patch, but a lot of it is just e.g. jiggling quote styles. Please add your own declaration of copyright if you wish to but leave the previous authors intact (unless you've been in contact with them and they've asked for their details to be removed).

sos/plugins/cs.py
```diff
@@ -1,128 +1,140 @@
...
-        if self.is_installed("redhat-cs") or exists("/opt/redhat-cs"):
+        """ Check if Certificate System 7, 8, or 9 is installed.
+        """
+
+        if self.is_installed('redhat-cs') or exists('/opt/redhat-cs'):
             return 71
-        elif self.is_installed("rhpki-common") or \
-                len(glob("/var/lib/rhpki-*")):
+        elif self.is_installed('rhpki-common') or \
+                len(glob('/var/lib/rhpki-*')):
```
bmr-cymru: Is there really any need to change all this quoting? It makes all your other changes much harder to spot and adds needless churn to this patch: the correct way to make this kind of change is to do it in a separate patch, either as cleanup before the new change, or to make things consistent afterwards. That allows reviewers to focus on the actual content change rather than line after line of quote twiddling.

sos/plugins/cs.py
```diff
@@ -1,128 +1,140 @@
...
+                '/var/log/pki-*/ra-debug.log',
+                '/var/log/pki-*/transactions',
+                '/var/log/pki-*/selftests.log',
+                '/var/log/pki-*/system'
             ])
         if csversion == 9:
-            # Get logs and configs for each subsystem if installed
-            for subsystem in ('ca', 'kra', 'ocsp', 'tks', 'tps'):
+            for dirs in os.listdir('/var/lib/pki'):
+
```
bmr-cymru: No need for a blank line.

sos/plugins/cs.py
```diff
@@ -1,128 +1,140 @@
...
+                # Files containing sensitive information.
+                self.add_forbidden_path('/etc/pki/' + dirs
+                                        + '/password.conf')
+                self.add_forbidden_path('/etc/pki/' + dirs
+                                        + '/alias/key3.db')
+
```
bmr-cymru: You can add these both in a single call to `self.add_forbidden_path(["list", "of", "paths"])`.

sos/plugins/cs.py
```diff
@@ -1,128 +1,140 @@
...
-                "/var/log/pki-*/system"
+                '/var/lib/pki-*/conf/*cfg*',
+                '/var/log/pki-*/debug',
+                '/var/log/pki-*/catalina.*',
+                '/var/log/pki-*/localhost.*',
+                '/var/log/pki-*/manager.*',
+                '/var/log/pki-*/ra-debug.log',
+                '/var/log/pki-*/transactions',
+                '/var/log/pki-*/selftests.log',
+                '/var/log/pki-*/system'
```
bmr-cymru: There's a lot of log data mixed up with configuration here: generally in `sos` we try to separate those, so that for e.g. size limits can be applied to log data. If we are going to update this plugin for the next release it would be good to include that.

sos/plugins/cs.py
```diff
@@ -1,128 +1,140 @@
...
+                                        + '/alias/key3.db')
+
+                # Get certificates from CA.
+                try:
+                    certpath = os.path.join('/var/lib/pki', dirs)
+                    self.add_cmd_output('certutil -L -d %s/alias'
+                                        % certpath)
+                except:
+                    self._log_warn('Could not list /var/lib/pki')
+
```
bmr-cymru: This doesn't work. If you catch an exception in `self.add_cmd_output()` then that is an error *adding* and not *running* the command. It's also extremely unlikely to happen since all we really do in that method is string and list manipulation.

If the intent is to log a warning when the command fails, then there's no need: the command execution code in the `collect()` phase already does that.

sos/plugins/cs.py
```diff
@@ -1,128 +1,140 @@
...
                 self.add_copy_spec([
-                    "/var/lib/pki/*/" + subsystem + "/conf/CS.cfg",
-                    "/var/lib/pki/*/logs/" + subsystem + "/system",
-                    "/var/lib/pki/*/logs/" + subsystem + "/transactions",
-                    "/var/lib/pki/*/logs/" + subsystem + "/debug",
-                    "/var/lib/pki/*/logs/" + subsystem + "/selftests.log"
+                    '/var/lib/pki/' + dirs,
+                    '/etc/pki/' + dirs,
+                    '/var/log/pki/' + dirs,
+                    '/var/log/pki/pki-*-spawn.*'
```
bmr-cymru: Ditto for logs & config mixed here.

sos/plugins/cs.py
```diff
@@ -1,128 +1,140 @@
...
+                self.do_path_regex_sub(
+                    r'\/etc\/pki\/' + dirs + '\/server.xml',
+                    r'%s=(\S*)' % attr,
+                    r'%s="********"' % attr
+                )
+            self.do_path_regex_sub(
+                r'\/etc\/pki\/' + dirs + '\/tomcat-users.xml',
+                r'password=(\S*)',
+                r'password="********"'
+            )
```
bmr-cymru: Ack.

83problems:
Thanks for the suggestions.  I'm thinking of closing this PR and resubmitting it with less changes to have more discussions with each one.  Thoughts?

bmr-cymru:
Thanks - you can just make the changes in your current branch (using `git commit --amend` or `git rebase -i`), and then force-update the branch with `git push --force`. That will keep this PR active and avoids the need to split of the discussion into a new one.

Things like quote cleanup are fine (we have a tonne of inconsistency in things like this), but they are much easier to handle in a separate patch, so that functional changes stand out from the noise.

83problems:
I'm having a difficult time reproducing the pycodestyle error from Travis. I understand why it's complaining but it's not complaining in CentOS or Fedora. I'll keep trying to work around the error.

sos/plugins/cs.py
```diff
@@ -60,69 +60,166 @@ def checkversion(self):
...
+                # Grab logs for each subsystem.
+                for subsystem in ("ca", "kra", "ocsp", "tks", "tps"):
+                    self.add_copy_spec([
+                        "/var/log/pki/" + dirs + "/catalina.*",
+                        "/var/log/pki/" + dirs + "/host-manager.*",
+                        "/var/log/pki/" + dirs + "/localhost.*",
+                        "/var/log/pki/" + dirs + "/manager.*",
+                        "/var/log/pki/" + dirs + "/" + subsystem + "/debug",
+                        "/var/log/pki/" + dirs + "/"
+                                        + subsystem + "/selftests",
```
pmoravec: Put the plus sign at the end of previous line, like:

                        "/var/log/pki/" + dirs + "/" +
                                          subsystem + "/selftests",

to prevent the "W503 line break before binary operator" pep error.

bmr-cymru:
No merge, duplicate, or fixup commits in merge branches please: you'll need to rebase this with `git rebase -i` and tidy up the history before it can go to master (update the branch with `git push --force` when you are done).

bmr-cymru:
You also have some Travis CI failures that need attention.

bmr-cymru:
Most likely it is a change in a newer upstream version than you have installed. Try running the script from a recent clone of the git repository:

```
    https://github.com/jcrocholl/pep8.git
```

bmr-cymru:
Still need to tidy up those commits, then I can take a look at them...

bmr-cymru:
No response since last September, and branch is not in a mergeable state: closing. Please re-open if you intend to carry on with this work.

PR #1281, opened 2018-05-02
[systemd] collect unit paths

rmetrich:

---

This PR proposes to collect unit file names found in standard systemd paths. It will enable to find out unwanted customizations, such as creating a symlink to some outside place.


pmoravec:
+1

rmetrich:
Damned, sorry for the typo (didn't test the "user" part).
We may indeed collect these, don't know what is better.

PR #1282, opened 2018-05-02
[libraries] collect standard library paths

rmetrich:

---

This PR proposes to collect the list of libraries found in lib and usr/lib (respectively lib64 and usr/lib64).
This may be useful to detect non-standard libraries pointing to some outside place.

### DIFF
--- a/sos/plugins/libraries.py
+++ b/sos/plugins/libraries.py
@@ -26,4 +26,16 @@ class Libraries(Plugin, RedHatPlugin, UbuntuPlugin):
             self.add_cmd_output("ldconfig -v -N -X")
         self.add_cmd_output("ldconfig -p -N -X")
 
+        # Collect library directories from ldconfig's cache
+        cmd = self.get_command_output("ldconfig -p -N -X")
+        dirs = set()
+        if cmd['status'] == 0:
+            for l in cmd['output'].splitlines():
+                s = l.split(" => ", 2)
+                if len(s) != 2:
+                    continue
+                dirs.add(s[1].rsplit('/', 1)[0])
+        self.add_cmd_output("ls -lanH %s" % " ".join(dirs),
+                            suggest_filename="ld_so_cache")
+
 # vim: set et ts=4 sw=4 :


bmr-cymru:
It would be fine to include a change like this, as well as the systemd one as separate commits on the same branch - it cuts down on duplicate PRs and manual labour when merging related changes.

bmr-cymru:
I'm not massively fond of this as there are numerous other places where DSOs may live, and we're somewhat blinded to analyse it without knowing LD_LIBRARY_PATH and other settings.

Before adding this I think I'd rather we added things like `ldconfig -p` and `ldconfig -NXv` output (which might also be parsed, along with `/etc/ld.so.conf*`) to generate a better list of libraries to examine.

rmetrich:
we have `ldconfig` already and `/etc/ld.so.conf*` content.
I understand the list is not exhaustive, but should help screening 3rd party libs overloading system ones.

bmr-cymru:
We also should be storing LD_LIBRARY_PATH and related environment variables - right now they are only captured in the `procenv` plugin, which is not installed on many Red Hat systems.

bmr-cymru:
> we have ldconfig already and /etc/ld.so.conf* content.

We capture them, but we  don't do anything with them: since they have a more global view of system DSO configuration than the four hard-coded paths in this patch they give a better level of insight across the system, and can be parsed to drive things like directory listings, or other data collection.

rmetrich:
OK. From `ldconfig_-p_-N_-X` we could do some "ls" on the right part to get the effective list.
But I guess this has to be done outside of setup() and I don't know how to do that.

rmetrich:
Now collect all libs from `ldconfig` output.
Filename is "hacked" since it would be too long

rmetrich:
Alright. Now getting the directories only and doing a `ls` on these, after following symlinks (using `-H`) otherwise `/usr/lib64` libs do not show up, since `/lib64` is a symlink to `/usr/lib64` and libraries in ldconfig are usually referenced as `/lib64/path.so`

bmr-cymru:
I've made some changes to the patch to make better use of `sos` APIs and to not call `ldconfig` repeatedly with the same arguments:

  * Use `self.get_command_output_now()` to immediately collect `ldconfig -p -N -x` output
  * Enclose file IO loop in `with` clause
  * Do not call `add_cmd_output()` with an empty `dirs` list (unlikely but possible if `ldconfig` is broken).

rmetrich:
Thanks!

bmr-cymru:
Welcome - thanks for the PR!

PR #1285, opened 2018-05-03
[systemd] collect additional useful information

rmetrich:

---

- Including `/run/systemd` directory enables to debug generated units, which is useful when customization are made by the customer, or when session leaks occur.
- `systemd-analyze dump` and `systemd-analyze plot` enable to understand the dynamic of the boot (`dump` has the raw data, `plot` is nice in a browser)

### DIFF
--- a/sos/plugins/systemd.py
+++ b/sos/plugins/systemd.py
@@ -45,10 +45,14 @@ class Systemd(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
             "systemd-delta",
             "systemd-analyze",
             "systemd-analyze blame",
+            "systemd-analyze dump",
             "journalctl --list-boots",
             "ls -lR /lib/systemd",
             "timedatectl"
         ])
+
+        self.add_cmd_output("systemd-analyze plot",
+                            suggest_filename="systemd-analyze_plot.svg")
 
         if self.get_option("verify"):
             self.add_cmd_output("journalctl --verify")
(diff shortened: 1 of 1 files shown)


BryanQuigley:
systemd-analyze dump outputs a 1.2 MB file on machine that I'm still uncertain of how useful it would be (or if it can replace one of the other commands)
systemd-analyze plot seems immediately useful and is < 200 kb to convey the same information in a nicer to read why.  Can we have that saved as an .svg for easier recognizing/opening.

/run/systemd also captures another copy of parts of the systemd journal - might be wise to be more specific what you find useful in /run/systemd

Also, your commit message is blank - looks like it got moved to the wrong place somehow.

BryanQuigley:
Oh, and it looks like you need to rebase.

rmetrich:
`systemd-analyze dump` contains the whole systemd tree, with dependencies, ordering, aliases, and start/stop times. It is well compressed and very useful when there are ordering cycles due to custom units.
I removed some of the `/run/systemd` to limit to useful data.

bmr-cymru:
Yeah: I tested that on a few systems here. It runs very quickly, and although it produces a few MiB of data, it compressed down to around 50-60K in most of my tests. I don't see any reason to not include it.

bmr-cymru:
:+1: that's inline with what I'm seeing on Fedora.

bmr-cymru:
Merged - thanks

PR #1286, opened 2018-05-03
[x11] Add .old files to check previous executions

mikelolasagasti:

---

### DIFF
--- a/sos/plugins/x11.py
+++ b/sos/plugins/x11.py
@@ -28,7 +28,9 @@ class X11(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
         self.add_copy_spec([
             "/etc/X11",
             "/var/log/Xorg.*.log",
+            "/var/log/Xorg.*.log.old",
             "/var/log/XFree86.*.log",
+            "/var/log/XFree86.*.log.old",
         ])
 
         self.add_forbidden_path([


bmr-cymru:
But the point of `--all-logs` is that it includes rotated log files (and non-default locations)...

```
       --all-logs
              Tell plugins to collect all possible log data ignoring any size limits and
              including logs in non-default locations. This option may significantly
              increase the size of reports.
```

Without `--all-logs` we should collect X11 logs for every X display (.0 for GDM, .1 for user etc.), but we should _not_ collect any `*.old` files unless `--all-logs` is given.

bmr-cymru:
> The way I've implemented it in the past (in logs and apport) is that we should collect the current log and one previous one by default (to match current and last). Then everything else should be collected by --all-logs - usually at the point the logs on Ubuntu are compressed so specifying a sizelimit become useless.

That was never really the intent... From the original commit message for the option:

```
commit 41e3d50b7c81cd23ee651e81eb0db5fb1aadd693
Date:   Tue Jul 8 12:42:00 2014 +0100

    [plugin] implement global --all-logs option
    
    Implement a global '--all-logs' option that causes all active
    plugins to attempt to collect all logs present regardles of
    rotation or size limiting. This replaces the numerous per-plugin
    'all_logs' options and avoids the need for clumsy command lines
    like:
    
     -k audit.all_logs,cron.all_logs,logs.all_logs,pcp.all_logs,...
    
    The audit, cron, cups, libvirt, logs, mysql, and pcp plugins have
    been converted to use the new option.
    
    Fixes #305.
    
```

So the original design was to *not* collect any rotated logs without `--all-logs`.

We have not been as consistent in applying this in reviews as I would have liked, but I don't see a good reason now abandon the original intent: if anything, I'd like to fix existing plugins that don't follow this model, as well as to improve the man pages and tool docs to make this absolutely clear to users (incidentally, we just moved the verbose journal boot logs behind `--all-logs`: I wouldn't have an issue extending this to `last` logs as well).

One place we are definitely inconsistent in today is in applying the size limit with `--all-logs`; some plugins still size limit with the option, some don't. I wouldn't be surprised if there's at least one plugin that does both in different places.

So really, the most important thing here is that we set out our intent in code, comments, help text, and documentation, and then stick to it.

I'm open to a discussion of what the defaults and expected behaviour are but we really must get away from the situation where each plugin author makes his or her own decisions: it's unfair to our users if nothing else.

BryanQuigley:
(incidentally, we just moved the verbose journal boot logs behind --all-logs: I wouldn't have an issue extending this to last logs as well).
That's what makes sense to me.  That's what these .old files are - Xorg logs IMU are predictable for this purpose.

>One place we are definitely inconsistent in today is in applying the size limit with --all-logs; some plugins still size limit with the option, some don't. I wouldn't be surprised if there's at least one plugin that does both in different places.
And I fully agree that is fully defined (so that plugin would be wrong and should change).

>I'm open to a discussion of what the defaults and expected behaviour are but we really must get away from the situation where each plugin author makes his or her own decisions: it's unfair to our users if nothing else.
I agree we should be consistent - and I think the current and last log (which depending on the application/method may be done using log rotation) is the right choice.  In the general sense this will line up with the current and last boot - but sometimes current and last run of a service, etc.

If we don't collect the previous boot logs (which on Ubuntu are usually rotated to kern.log.1) we're going to miss out on a large percentage of cases where the issue occurred on the previous boot.  It is also possible that the actual previous boot is now in kern.log.5.gz, but I think we should just try to collect 1 previous log if it's not predictable in that way (so current=kern.log last=kern.log.1).

PR #1288, opened 2018-05-09
[kernel] Avoid capturing blanket 'find -ls'

TurboTurtle:
Previously, if sos was run in an environment where '/lib/modules/*' did
not exist, the kernel plugin would pass an empty extra_mod_paths list to
'find -ls', which would in turn cause sos to execute 'find -ls' on the
entire filesystem.

While this situation may seem rare, this is a possiblity whenever sos is
run in a container that does not have the proper environment variable
set that indicates to the active policy that sos is in fact in  a
container. On top of generating a sosreport with little to useful
information in this case, the generated sosreport could potentially be
very large due to the collection of a blanket 'find -ls'.

This prevents this from occuring, by not running 'find -ls' if the
extra_mod_paths list is empty.

---

### DIFF
--- a/sos/plugins/kernel.py
+++ b/sos/plugins/kernel.py
@@ -54,11 +54,13 @@ class Kernel(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
         for pattern in extra_mod_patterns:
             extra_mod_paths.extend(glob.glob(pattern))
 
+        if extra_mod_paths:
+            self.add_cmd_output("find %s -ls" % " ".join(extra_mod_paths))
+
         self.add_cmd_output([
             "dmesg",
             "sysctl -a",
-            "dkms status",
-            "find %s -ls" % " ".join(extra_mod_paths)
+            "dkms status"
         ])
 
         clocksource_path = "/sys/devices/system/clocksource/clocksource0/"


TurboTurtle:
Just had this happen on a support case with Atomic Host as the OS. In this case, a container that wasn't configured properly (no `container` or `host_sysroot` env vars set) was used to run sos. The container didn't have anything under `lib/modules/` and the resulting `find -ls` was nearly a GB by itself. :\

bmr-cymru:
>While this situation may seem rare, this is a possiblity whenever sos is
run in a container that does not have the proper environment variable
set that indicates to the active policy that sos is in fact in  a
container. On top of generating a sosreport with little to useful
information in this case, the generated sosreport could potentially be
very large due to the collection of a blanket 'find -ls'.

While I'm concerned by this behaviour, I'm also worried about the circumstances in which you report it happening:

> sos is run in a container that does not have the proper environment variable set

What are these situations? Is this users manually running in a container, not setting command line options, or container runtimes that are (still...) failing to set the agreed upon variables to allow detection.

We need to understand this, and not paper over the cracks until we have done so: if sos is being used in a container without proper settings then it is very unlikely to behave as expected, and may produce output that is misleading, confusing, or just plain wrong.

bmr-cymru:
> Just had this happen on a support case with Atomic Host as the OS. In this case, a container that wasn't configured properly (no container or host_sysroot env vars set) was used to run sos. The container didn't have anything under lib/modules/ and the resulting find -ls was nearly a GB by itself. :\

We need to know how this happened: what release/version, what instructions the user followed, where they got them from, and how they ran the command.

TurboTurtle:
In this case, the end user was using the `rhel-tools` container image which used to have sos but no longer does (that's `support-tools` now). All of RH's current docs reflect this change to use support-tools instead of rhel-tools, but it appears the user pulled the `rhel-tools:latest` image on Atomic Host (7.5) and manually installed sos then ran it from there. 

I'm not aware of any instructions in any RH-family distro to do this. I think it was just a user not aware that `support-tools` exists and is the replacement for `rhel-tools` when it comes to running sos on Atomic.

bmr-cymru:
Well, that really does sound like a problem then: that's how it is supposed to work. The Atomic runtime is supposed to set these variables for every container, we pick them up and react. I'd be curious as to what is now  being done differently in the `rhel-tools` SPC vs. the `support-tools` build.

PR #1289, opened 2018-05-09
[storage] Add systool output for fc classes

jmagrini:
In the land of product support, fc class output is very useful for troubleshooting
fcp issues and path failures.  This plugin captures output of several fc classes:
  - fc_host
  - fc_remote_ports

Signed-off-by: Jon Magrini <jmagrini@redhat.com>

### DIFF
--- a/sos/plugins/systool.py
+++ b/sos/plugins/systool.py
@@ -1,0 +1,43 @@
+# systool.py
+# Copyright (C) 2007-2014 Red Hat, Inc., Jon Magrini <jmagrini@redhat.com>
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
+
+import os
+import os.path
+from sos.plugins import Plugin, RedHatPlugin, os
+
+class sysfsutils(Plugin, RedHatPlugin):
+    """View system device information by bus, class, and topology
+    """
+
+    plugin_name = 'sysfsutils'
+    profiles = 'storage'
+    packages = 'sysfsutils'
+    kernel_mods = ('scsi_transport_fc', 'scsi_tgt')
+
+    def get_systool_output(self):
+        """ systool fc specific information - commands
+        """
+        self.add_cmd_output([
+            "/usr/bin/systool -c fc_host -v",
+            "/usr/bin/systool -c fc_remote_ports -v"
+        ])
+
+    def setup(self):
+        # If sysfsutils is installed collect data
+        if os.path.isfile("/usr/bin/systool"):
+            self.add_custom_text("sysfsutils is installed.<br>")
+            self.get_systool_output()


sos/plugins/systool.py
```diff
@@ -0,0 +1,43 @@
...
+        """ systool fc specific information - commands
+        """
+        self.add_cmd_output([
+            "/usr/bin/systool -c fc_host -v",
+            "/usr/bin/systool -c fc_remote_ports -v"
+        ])
+
+    def setup(self):
+        # If sysfsutils is installed collect data
+        if os.path.isfile("/usr/bin/systool"):
```
TurboTurtle: Is this ever expected to not be present if the `sysfsutils` package is installed? According to a `yum provides`, that's where it comes from.

If it is,  you can use `commands = ('systool', )` to only activate the plugin if that command is present much like how `packages = ('sysfsutils', )` works. In either case, the use of `os.path.isfile` is unnecessary.

sos/plugins/systool.py
```diff
@@ -0,0 +1,43 @@
...
+import os.path
+from sos.plugins import Plugin, RedHatPlugin, os
+
+class sysfsutils(Plugin, RedHatPlugin):
+    """View system device information by bus, class, and topology
+    """
+
+    plugin_name = 'sysfsutils'
+    profiles = 'storage'
+    packages = 'sysfsutils'
```
TurboTurtle: Stylistically `packages` is usually a tuple, e.g. `packages = ('sysfsutils', )`. A string works here because we convert strings to a list internally, however only a few plugins use a list here and no other plugins use a string currently (anymore?).

sos/plugins/systool.py
```diff
@@ -0,0 +1,43 @@
...
+        self.add_cmd_output([
+            "/usr/bin/systool -c fc_host -v",
+            "/usr/bin/systool -c fc_remote_ports -v"
+        ])
+
+    def setup(self):
+        # If sysfsutils is installed collect data
+        if os.path.isfile("/usr/bin/systool"):
+            self.add_custom_text("sysfsutils is installed.<br>")
+            self.get_systool_output()
```
TurboTurtle: Why not pull `get_systool_output()` directly into `setup()`, especially if `systool` is not expected to ever not be present when `sysfsutils` is installed and we're not doing anything else in terms of collection?

sos/plugins/systool.py
```diff
@@ -0,0 +1,43 @@
...
+        """
+        self.add_cmd_output([
+            "/usr/bin/systool -c fc_host -v",
+            "/usr/bin/systool -c fc_remote_ports -v"
+        ])
+
+    def setup(self):
+        # If sysfsutils is installed collect data
+        if os.path.isfile("/usr/bin/systool"):
+            self.add_custom_text("sysfsutils is installed.<br>")
```
TurboTurtle: Isn't this implied by the fact the plugin was run based on the presence of the `sysfsutils` package?

sos/plugins/systool.py
```diff
@@ -0,0 +1,43 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
+
+import os
+import os.path
+from sos.plugins import Plugin, RedHatPlugin, os
```
TurboTurtle: Redundant imports of `os`. See notes below about plugin enablement and `/usr/bin/systool`.

sos/plugins/systool.py
```diff
@@ -0,0 +1,43 @@
+# systool.py
+# Copyright (C) 2007-2014 Red Hat, Inc., Jon Magrini <jmagrini@redhat.com>
```
TurboTurtle: 2007-2014? Since this is a new plugin, date should be current year.

bmr-cymru:
Is `systool` now un-deprecated? We stripped this stuff out years ago because the tool and corresponding library were no longer maintained upstream.

sos/plugins/systool.py
bmr-cymru: Plugin class names (other than derived classes with a tagging class prefix, e.g., `RedHatFooPlugin`) must match the filename that contains the. It's either `systool` or `sysfsutils`, but not both.

sos/plugins/systool.py
bmr-cymru: Please also do not import Python modules via sos modules that also happen to include them. This makes for confusing module relationships.

sos/plugins/systool.py
```diff
@@ -0,0 +1,43 @@
...
+    profiles = 'storage'
+    packages = 'sysfsutils'
+    kernel_mods = ('scsi_transport_fc', 'scsi_tgt')
+
+    def get_systool_output(self):
+        """ systool fc specific information - commands
+        """
+        self.add_cmd_output([
+            "/usr/bin/systool -c fc_host -v",
+            "/usr/bin/systool -c fc_remote_ports -v"
```
bmr-cymru: Nack. Absolute paths have not been allowed in `sos` (outside extremely specific cases where a binary is never in `PATH`) for many years.

Just call the program by its executable name and leave the policy mechanism to figure out where to find it.

sos/plugins/systool.py
bmr-cymru: Nack. Never test for the presence of files or executables before collecting/running them (as @TurboTurtle mentions, the plugin enablement infrastructure does that for you when required). It's redundant and makes the plugin code unnecessarily indented and harder to follow.

sos/plugins/systool.py
bmr-cymru: This isn't really adding anything. That data is already in the report via both the package manager data and the fact that the plugin was enabled.

bmr-cymru:
No upstream changes at sysfsutils since 2006:

http://linux-diag.sourceforge.net/Sysfsutils.html

jtpittman195:
All, just to note, apart from the style issues and concerns of upstream changes to sysfsutils in the patch, we ask for systool output, literally, every week, and probably more realistically multiple times a week.  Until a better tool comes along to parse the fc sysfs data, we'll continue to have to do so.  Digging through sysfs is awful, time-consuming, and requires knowledge of fc code structure.  In any case, not including these commands in some form is creating a great deal of work.  Is there any solution you guys can see here?  Any way to get these commands included?  Maybe under an already established plugin?

bmr-cymru:
It's really just a pretty-printer: you can also use `udevadm` to view sysfs attributes in pretty notation, e.g.:

```
# udevadm info -a /sys/class/scsi_host/*

Udevadm info starts with the device specified by the devpath and then
walks up the chain of parent devices. It prints for every device
found, all possible attributes in the udev rules key format.
A rule to match, can be composed by the attributes of the device
and the attributes from one single parent device.

  looking at device '/devices/pci0000:00/0000:00:01.1/ata1/host0/scsi_host/host0':
    KERNEL=="host0"
    SUBSYSTEM=="scsi_host"
    DRIVER==""
    ATTR{unchecked_isa_dma}=="0"
    ATTR{state}=="running"
    ATTR{cmd_per_lun}=="1"
    ATTR{sg_tablesize}=="128"
    ATTR{active_mode}=="Initiator"
    ATTR{sg_prot_tablesize}=="0"
    ATTR{unique_id}=="1"
    ATTR{supported_mode}=="Initiator"
    ATTR{proc_name}=="ata_piix"
    ATTR{prot_capabilities}=="0"
    ATTR{can_queue}=="1"
    ATTR{prot_guard_type}=="0"
    ATTR{use_blk_mq}=="0"
    ATTR{eh_deadline}=="off"
    ATTR{host_busy}=="0"

  looking at parent device '/devices/pci0000:00/0000:00:01.1/ata1/host0':
    KERNELS=="host0"
    SUBSYSTEMS=="scsi"
    DRIVERS==""

  looking at parent device '/devices/pci0000:00/0000:00:01.1/ata1':
    KERNELS=="ata1"
    SUBSYSTEMS==""
    DRIVERS==""

  looking at parent device '/devices/pci0000:00/0000:00:01.1':
    KERNELS=="0000:00:01.1"
    SUBSYSTEMS=="pci"
    DRIVERS=="ata_piix"
    ATTRS{irq}=="0"
    ATTRS{subsystem_vendor}=="0x1af4"
    ATTRS{broken_parity_status}=="0"
    ATTRS{class}=="0x010180"
    ATTRS{driver_override}=="(null)"
    ATTRS{consistent_dma_mask_bits}=="32"
    ATTRS{dma_mask_bits}=="32"
    ATTRS{local_cpus}=="3"
    ATTRS{device}=="0x7010"
    ATTRS{enable}=="1"
    ATTRS{msi_bus}==""
    ATTRS{local_cpulist}=="0-1"
    ATTRS{vendor}=="0x8086"
    ATTRS{resource1}=="P"
    ATTRS{resource3}==""
    ATTRS{subsystem_device}=="0x1100"
    ATTRS{numa_node}=="-1"
    ATTRS{d3cold_allowed}=="0"

  looking at parent device '/devices/pci0000:00':
    KERNELS=="pci0000:00"
    SUBSYSTEMS==""
    DRIVERS==""

```

I'm surprised that just formatting this causes such a burden. When I worked in SEG we just used ad-hoc shell and awk scripts to comprehend the raw data and this worked better for our needs at the time.

I'm strongly opposed to re-introducing `systool` to `sos`, since we removed it a number of years ago, the tool is deprecated, is unmaintained upstream (and has been since 2006!), is hated by the kernel developers and maintainers, and is disappearing from distributions already.

Note that it's not true that "code knowledge" is needed to interpret the sys data: `systool` is just doing simple mechanical transformation and formatting. A simple shell script can perform the same job.

There are two options that I can currently think of that would be acceptable to the `sos` team from a maintenance and supportability standpoint:

* We collect `udevadm info` output for the relevant sysfs paths. This is supported by the `udev` package for RHEL7 onwards (haven't checked other distro families).

* We implement our own pretty-printer for this data, based on the existing `systool` notation

Both of these would be a better result for our developers and users - they do not rely on deprecated/unmaintained or optional packages and would ensure that the data is collected first time, every time. I'm reasonably happy to go ahead with either at this stage, although there's an obvious preference for `udevadm` since it is both generic and less work to implement.

bmr-cymru:
I'm going to close this request out now, as we have the `add_udev_info()` support for both fibrechannel classes and regular block devices in master now.

Let us know if there is anything else we can add for this.

PR #1290, opened 2018-05-11
[fibrechannel] Add new plugin

TurboTurtle:
Adds a new plugin for fibrechannel devices.

---

### DIFF
--- a/sos/plugins/fibrechannel.py
+++ b/sos/plugins/fibrechannel.py
@@ -1,0 +1,43 @@
+# Copyright (C) 2018 Red Hat, Inc. Jake Hunsaker <jhunsake@redhat.com>
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+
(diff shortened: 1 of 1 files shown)


sos/plugins/__init__.py
```diff
@@ -860,6 +860,26 @@ def add_journal(self, units=None, boot=None, since=None, until=None,
         self._log_debug("collecting journal: %s" % journal_cmd)
         self._add_cmd_output(journal_cmd, None, None, timeout)
 
+    def add_udev_info(self, device, attrs=False):
```
bmr-cymru: This is a big new interface in the `Plugin` class: it deserves its own commit, and description.

bmr-cymru: It might be good to also support `-p` so that we can directly replace existing `add_cmd_output()` calls for `udevadm` with this method. This also needs some thought as to how it's going to function in a container - is `/sys` mounted where we'd expect?

TurboTurtle:
Ack. I split the `add_udev_info()` addition into a new PR, #1293.

As far as the RH support team has told me, there isn't anything else they need to see in a sosreport beyond what's already captured (lspci, etc...) once this plugin exists.

PR #1291, opened 2018-05-12
[omnipath_*] Add two plugins

pmoravec:
plugins added:
- omnipath_manager for OPA fabric subnet manager
- omnipath_client for management/head node of the cluster

Resolves: #1291

---

### DIFF
--- a/sos/plugins/omnipath_client.py
+++ b/sos/plugins/omnipath_client.py
@@ -1,0 +1,51 @@
+# Copyright (C) 2018 Red Hat, Inc., Pavel Moravec <pmoravec@redhat.com>
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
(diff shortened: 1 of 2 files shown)


sos/plugins/omnipath_client.py
```diff
@@ -0,0 +1,52 @@
...
+
+        # opacapture generates a tarball of given name we should collect;
+        # rather than storing it somewhere under /var/tmp and copying it via
+        # add_copy_spec, add it directly to sos_commands/<plugin> dir by
+        # changing dir there temporarily and collecting the cmd now
+        cwd = os.getcwd()
+        os.chdir(self.get_cmd_output_path())
+        self.get_cmd_output_now("/usr/sbin/opacapture my_opacapture.tgz",
+                                suggest_filename="opacapture")
+        os.chdir(cwd)
```
bmr-cymru: Are you really saying that this command is too dumb to simply accept a path on the command line, rather than a file name?

That is .... stunning.

bmr-cymru: But this gets a thorough "NACK" from me: plugins must not alter global state. What happens if an exception fires between the two calls to `os.chdir()`? It isn't even wrapped in any try/finally!!

sos/plugins/omnipath_client.py
```diff
@@ -0,0 +1,52 @@
+# This program is free software; you can redistribute it and/or modify
```
bmr-cymru: No copyright statement?

sos/plugins/omnipath_manager.py
```diff
@@ -0,0 +1,50 @@
...
+
+    plugin_name = 'omnipath_manager'
+    profiles = ('hardware',)
+
+    packages = ('opa-fm')
+
+    def setup(self):
+
+        self.add_cmd_output([
+            "/usr/lib/opa-fm/bin/config_check -v -d -s",
```
bmr-cymru: Why are these all hidden away in this location? How is the administrator expected to run them if they are required? Are they added to PATH by some script in `/etc/profile.d`?

sos/plugins/omnipath_client.py
pmoravec: Added.

PR #1293, opened 2018-05-14
[Plugins] Add new add_udev_info() method

TurboTurtle:
Adds a new collection method add_udev_info() which will collect 'udevadm
info' output for a given device or list of devices.

An optional 'attrs' boolean will, if True, have udevadm perform an
attribute-walk of the device's sysfs attributes.

This method can take either device names such as '/dev/sda' or sysfs
paths such as '/sys/class/scsi_host/host0' and these styles may be
freely interchanged in a list given to a single call.

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -860,6 +860,26 @@ class Plugin(object):
         self._log_debug("collecting journal: %s" % journal_cmd)
         self._add_cmd_output(journal_cmd, None, None, timeout)
 
+    def add_udev_info(self, device, attrs=False):
+        """Collect udevadm info output for a given device
+
+        Arguments:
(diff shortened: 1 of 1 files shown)


TurboTurtle:
From #1290 

>It might be good to also support -p so that we can directly replace existing add_cmd_output() calls for udevadm with this method. This also needs some thought as to how it's going to function in a container - is /sys mounted where we'd expect?

The `-p` doesn't actually gain us much, it just allows us to not specify '/sys' in the device path. From the man page:

>       -p, --path=DEVPATH
>           The /sys path of the device to query, e.g.  [/sys]/class/block/sda. Note that this option usually is >not very useful, since udev can guess the type of the argument, so udevadm
>          --devpath=/class/block/sda is equivalent to udevadm /sys/class/block/sda.

PR #1248, opened 2018-03-16
[docs] Fixed errors while generating docs.

psachin:
I see below errors when generating docs using `make`. Decided to
cleanup a bit. The docstring look better than previous and no more
errors when generating docs.
```
/home/psachin/github/sos/sos/plugins/__init__.py:docstring of sos.plugins.Plugin.add_journal:5: ERROR: Unexpected indentation.
/home/psachin/github/sos/sos/plugins/__init__.py:docstring of sos.plugins.Plugin.add_journal:6: WARNING: Block quote ends without a blank line; unexpected unindent.
/home/psachin/github/sos/sos/plugins/__init__.py:docstring of sos.plugins.Plugin.add_journal:13: WARNING: Definition list ends without a blank line; unexpected unindent.
/home/psachin/github/sos/sos/plugins/__init__.py:docstring of sos.plugins.Plugin.add_journal:15: ERROR: Unexpected indentation.
/home/psachin/github/sos/sos/plugins/__init__.py:docstring of sos.plugins.Plugin.add_journal:16: WARNING: Block quote ends without a blank line; unexpected unindent.
/home/psachin/github/sos/sos/plugins/__init__.py:docstring of sos.plugins.Plugin.add_journal:18: WARNING: Definition list ends without a blank line; unexpected unindent.
```

Let me know if anything else needs cleanup in `sos/sos/plugins/__init__.py`

Signed-off-by: Sachin Patil <psachin@redhat.com>

---
Please place an 'X' inside each '[]' to confirm you adhere to our [Contributor Guidelines](https://github.com/sosreport/sos/wiki/Contribution-Guidelines)

- [X] Is the commit message split over multiple lines and hard-wrapped at 72 characters?
- [X] Is the subject and message clear and concise?
- [X] Does the subject start with **[plugin_name]** if submitting a plugin patch or a **[section_name]** if part of the core sosreport code?
- [X] Does the commit contain a **Signed-off-by: First Lastname <email@example.com>**?

PR #1305, opened 2018-05-22
[watchdog] Add watchdog plugin

nrwahl2:
This new plugin collects information related to the watchdog service,
including configuration, custom executables for test-binary and
repair-binary, and stdout/stderr log files.

This will be helpful to support teams seeking to understand why an
unexpected reboot occurred or, conversely, why a reboot did not
occur as expected.

Signed-off-by: Reid Wahl <nwahl@redhat.com>

---

### DIFF
--- a/sos/plugins/watchdog.py
+++ b/sos/plugins/watchdog.py
@@ -1,0 +1,91 @@
+# Copyright (C) 2018 Red Hat, Inc., Reid Wahl <nwahl@redhat.com>
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
+
+from sos.plugins import Plugin, RedHatPlugin
+
+from glob import glob
+import os
+
+
+class Watchdog(Plugin, RedHatPlugin):
+    '''Watchdog information.'''
+    plugin_name = 'watchdog'
+    profiles = ('system',)
+    packages = ('watchdog',)
+    option_list = [
+        ('conf_file', 'watchdog config file', 'fast', '/etc/watchdog.conf'),
+    ]
+
+    def get_log_dir(self, conf_file):
+        '''Get watchdog log directory.
+
+        Get watchdog log directory path configured in ``conf_file``.
+
+        :returns: The watchdog log directory path.
+        :returntype: str.
+        :raises: IOError if ``conf_file`` is not readable.
+        '''
+        log_dir = None
+
+        with open(conf_file, 'r') as conf_f:
+            for line in conf_f:
+                line = line.split('#')[0].strip()
+
+                try:
+                    (key, value) = line.split('=', 1)
+                    if key.strip() == 'log-dir':
+                        log_dir = value.strip()
+                except ValueError:
+                    pass
+
+        return log_dir
+
+    def setup(self):
+        '''Collect watchdog information.
+
+        Collect configuration files, custom executables for test-binary
+        and repair-binary, and stdout/stderr logs.
+        '''
+        conf_file = self.get_option('conf_file')
+        log_dir = '/var/log/watchdog'
+
+        # Get service configuration and sysconfig files
+        self.add_copy_spec([
+            conf_file,
+            '/etc/sysconfig/watchdog',
+        ])
+
+        # Get custom executables
+        self.add_copy_spec([
+            '/etc/watchdog.d',
+            '/usr/libexec/watchdog/scripts',
+        ])
+
+        # Get logs
+        try:
+            res = self.get_log_dir(conf_file)
+            if res:
+                log_dir = res
+        except IOError as ex:
+            self._log_warn("Could not read %s: %s" % (conf_file, ex))
+
+        if self.get_option('all_logs'):
+            log_files = glob(os.path.join(log_dir, '*'))
+        else:
+            log_files = (glob(os.path.join(log_dir, '*.stdout')) +
+                         glob(os.path.join(log_dir, '*.stderr')))
+
+        self.add_copy_spec(log_files, sizelimit=self.get_option('log_size'))


sos/plugins/watchdog.py
```diff
@@ -0,0 +1,90 @@
...
+        '''Initialize Watchdog class.'''
+        Plugin.__init__(self, commons)
+        self.limit = self.get_option('log_size')
+        self.watchdog_conf = '/etc/watchdog.conf'
+        self.conf_vars = dict()
+        self.log_dir = None
+
+    def parse_watchdog_conf(self):
+        '''Parse variables from watchdog config file.'''
+        comment_pattern = re.compile(r'#.*')
```
nrwahl2: Somewhat crude way to remove comments at end of line rather than simply skipping lines that begin with a comment. This does not guard against removing a literal `#` character in a quoted string. However, to my knowledge there is no reason a literal `#` would be used in a value in `/etc/watchdog.conf`. So I chose not to make the comment parsing any more complicated than this.

sos/plugins/watchdog.py
```diff
@@ -0,0 +1,90 @@
...
+        eq_pattern = re.compile(r'\s*=\s*')
+
+        with open(self.watchdog_conf, 'r') as conf_file:
+            for line in conf_file:
+                line = comment_pattern.sub('', line)
+                line = line.strip()
+
+                try:
+                    (key, value) = eq_pattern.split(line, 1)
+                    self.conf_vars[key] = value
```
pmoravec: Why to store all configuration in `conf_vars`, if just the only `log-dir` is used further? Isn't it easier to grep for files with pattern like `[ ]*log-dir[ ]*=[ ]*$1` (not correct regexp) and store the value?

pmoravec:
+1 to the plugin, my comment is rather optional.

sos/plugins/watchdog.py
nrwahl2: That is a good point. I normally try to take a general, extensible approach to parsing log files, as I've done here. However, I don't foresee any other parameters in `/etc/watchdog.conf` being useful now or in the future in this context.

I can put together an alternate version using only `log-dir` in the morning (it's late here now) if you want, and we can decide which one looks cleaner. Readability is going to trump computational/memory efficiency here IMO, with this being such a small file to parse.

sos/plugins/watchdog.py
```diff
@@ -0,0 +1,90 @@
...
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
+'''Collect info from watchdog to add to sosreport.
+
+Classes:
+Watchdog -- collects watchdog configs, custom executables, and logs
+'''
```
bmr-cymru: We don't really do plugin file level docstrings at the moment. They don't really add a whole lot beyond what the existing plugin metadata can display.

There's currently an effort to improve the `sos` docstrings in general that @psachin is working on (see #1248), if you'd be interested in getting involved in that.

sos/plugins/watchdog.py
```diff
@@ -0,0 +1,90 @@
...
+
+import re
+
+from sos.plugins import Plugin, RedHatPlugin
+
```
bmr-cymru: It's different from typical python conventions but we normally place the `sos` module imports at the top, followed by any standard library or other import statements.

sos/plugins/watchdog.py
```diff
@@ -0,0 +1,90 @@
...
+    Methods:
+    parse_watchdog_conf -- parse variables from config file
+    setup -- assemble watchdog info for sosreport
+
+    Instance variables:
+    limit -- maximum log size
+    watchdog_conf -- watchdog service configuration file
+    conf_vars -- variables from watchdog config file
+    log_dir -- directory containing watchdog stderr and stdout logs
+    '''
```
bmr-cymru: We use the Sphinx package for generating documentation, and Sphinx markup in docstring comments: class members should be documented by a preceding `#:` comment, and method documentation belongs in the method docstring itself.

See [Sphinx tutorials](https://thomas-cokelaer.info/tutorials/sphinx/docstring_python.html) or other projects (like [boom](https://github.com/bmr-cymru/boom/blob/master/boom/bootloader.py)) for examples.

sos/plugins/watchdog.py
```diff
@@ -0,0 +1,90 @@
...
+    packages = ('watchdog',)
+
+    def __init__(self, commons):
+        '''Initialize Watchdog class.'''
+        Plugin.__init__(self, commons)
+        self.limit = self.get_option('log_size')
+        self.watchdog_conf = '/etc/watchdog.conf'
+        self.conf_vars = dict()
+        self.log_dir = None
+
```
bmr-cymru: Nack, please don't override `Plugin.__init__()` just to set class members (there also seems little point in stashing `self.get_option("log_size")` in a member, just to access it in exactly one place in `setup()`. Either use a local (reasonable if it is to be used more than once), or just make the `get_option()` call in-line.

sos/plugins/watchdog.py
bmr-cymru: As @pmoravec said, there's not much point in fully parsing this file if the results will never be used. The code to simply extract the `log-dir` key and its value is more compact and easier to read.

There's also no need to deploy a regex for comment removal - since the comment is started by `#` you can simply split on that character and discard everything else.

bmr-cymru: You're collecting the file in any case, so the dictionary really is not adding any value at this point.

nrwahl2: > There's also no need to deploy a regex for comment removal.
Yep, duh... I think I had something more complex initially and dropped it to that.

nrwahl2:
@bmr-cymru I appreciate your feedback. I'm working on fixing these issues.

Many are stylistic. Is there a style guide? I tried to model after certain existing plugins (where there is often variation in style), and then to align the result with PEP-8, PEP-257, and anything remaining pylint suggested at the end of that (e.g., overriding the `__init__` method to define attributes). I also used the "How to Write a Plugin" guide.

I'll be glad to use the project's preferred style going forward.

nrwahl2:
@bmr-cymru : Here's another shot at it. We might not be 100% in alignment still. This ought to be closer.

sos/plugins/watchdog.py
```diff
@@ -0,0 +1,84 @@
...
+        Get watchdog log directory path configured in ``conf_file``.
+
+        :returns: The watchdog log directory path.
+        :returntype: str.
+        :raises: IOError if ``conf_file`` is not readable.
+        '''
+        log_dir = None
+        log_dir_pattern = re.compile(r'log-dir\s*=\s*(.*)')
+
+        with open(self.get_option('conf_file'), 'r') as conf_file:
```
bmr-cymru: Why get this twice? If you're implementing a method, then why not simply pass the `watchdog_conf` down from the parent method? (it would also make the reference to ``conf_file`` in the docstring more clear: it looks odd, as it's neither a class member, nor a function argument, but an sos plugin option name).

As a general style thing, it's also good to always use the same name for a variable representing the same thing, unless there's a good reason not to: i.e. I'd suggest either `conf_file` in both places, or `watchdog_conf` (not much to chose between them aside form the shorter length of `conf_file`).

sos/plugins/watchdog.py
```diff
@@ -0,0 +1,84 @@
...
+            for line in conf_file:
+                line = line.split('#')[0].strip()
+
+                try:
+                    match = log_dir_pattern.match(line)
```
bmr-cymru: I'm still not really convinced that this warrants a regex, but hopefully we might see some progress on a general config parsing facility soon which would make this moot.

sos/plugins/watchdog.py
```diff
@@ -0,0 +1,84 @@
...
+
+        # Get logs
+        try:
+            res = self.get_log_dir()
+            if res:
+                log_dir = res
+        except IOError as ex:
+            self._log_warn("Could not parse %s: %s" % (watchdog_conf, ex))
+
+        self.add_copy_spec(log_dir, sizelimit=self.get_option('log_size'))
```
bmr-cymru: If you want size limiting to work you need to pass a glob that expands to a list of files: size limiting does not apply to recursive directory collection currently. Generally this means something like appending "*" for the `--all-logs` case, and e.g. `*.log` for the non `--all-logs` side.

sos/plugins/watchdog.py
```diff
@@ -0,0 +1,84 @@
...
+            '/usr/libexec/watchdog/scripts',
+        ])
+
+        # Get logs
+        try:
+            res = self.get_log_dir()
+            if res:
+                log_dir = res
+        except IOError as ex:
+            self._log_warn("Could not parse %s: %s" % (watchdog_conf, ex))
```
bmr-cymru: "Could not read" may be a better message here, since it matches the type of exception you are catching ("Could not parse" could be seen to suggest a syntactic problem with the file content).

sos/plugins/watchdog.py
nrwahl2: Honestly, that's how I had implemented it at first in the revision. And then I figured you would say it was unnecessary/clunky to pass the conf file as an argument to `get_log_dir`, so I switched to this. The coin landed on the wrong side.

Duly noted.

bmr-cymru: Sorry you had to re-do it again - generally, if I don't say anything then something is OK ;-)

sos/plugins/watchdog.py
nrwahl2: What do you think of this? Still need to be able to strip whitespace due to the format of the file. `ValueError` for no `=` separator.

```
                try:
                    (key, value) = line.split('=', 1)
                    if key.strip() == 'log-dir':
                        log_dir = value.strip()
                except ValueError:
                    pass
```

Can then axe the `re` import since this was the only application of it.

bmr-cymru: That's the kind of parsing I tend to favour, but we're really getting into personal preference at this point - there are plenty of plugins that do use `re` for this sort of task (and even more trivial tasks..).

I wouldn't hold a plugin up just for a minor change like this and in any case, we're actively looking at adding some kind of general config handling mechanism now anyway so even if it does go in using `re` to start it may not stay in that form for long.

nrwahl2: I like it better too now that I'm looking at it, for such simple parsing. Much easier to see what's actually going on.

sos/plugins/watchdog.py
nrwahl2: I would rather you deeply critique everything that comes in, than ack a bunch of stuff that works but looks sloppy or doesn't fit with the project's style.

TurboTurtle:
Looks like the `pep8 sos` failure is because:

- PEP8 *used* to enforce the opposite of what is written now
- It (the PEP) was updated after pep8 (the command) was renamed to `pycodestyle` (https://github.com/PyCQA/pycodestyle/issues/466)

So `pep8` never got updated, whereas `pycodestyle` has been, and will not report this as a problem (and seems to accept the previous style as well by default).

@bmr-cymru I think we should look at updating the checks to using `pycodestyle` after the 3.6 release, and for this PR use the previous guideline on operator splits.

nrwahl2:
@TurboTurtle Either that or ditch `+` in favor of `list.extend()` or something like it.

bmr-cymru:
I find this kind of pep8 stuff really annoying: having a "hanging" operator used to be the enforced style (totally contrary to e.g. common C convention, where you put the operator at the beginning of the continuation line where it is more visually prominent).

Anyway: we can find something that will work, whether that's your current patch or something that manages to keep both of them happy.

It's a holiday in the UK today, and I'm on vacation for the rest of the week, but I'll be back reviewing and merging `sos` stuff for an hour or two each day.

bmr-cymru:
Is there any reason to _not_ switch to `pycodestyle` before 3.6? If we get clean runs on it now then we might as well make the switch, and not carry off-convention changes in the tree.

One problem could be packagers running something like our git hook script during package build in environments where `pycodestyle` is not available: I'm not sure that's a big problem though, as our own test targets (`make` and `distutils`) only run the `nose` tests and not `pep8`.

TurboTurtle:
Not particularly, I was just thinking that it would be slightly awkward to change it in the middle of trying to close out a release. With the current master, we'd need to make 81 fixups just by changing out to `pycodestyle` - which while not terrible, seems extraneous to the 3.6 release to me.

bmr-cymru:
```
$ pycodestyle sos
sos/__init__.py:31:1: E305 expected 2 blank lines after class or function definition, found 1
sos/archive.py:407:9: E722 do not use bare except'
sos/sosreport.py:1025:17: E722 do not use bare except'
sos/sosreport.py:1034:25: E722 do not use bare except'
sos/sosreport.py:1040:17: E722 do not use bare except'
sos/sosreport.py:1243:13: E722 do not use bare except'
sos/sosreport.py:1289:13: E722 do not use bare except'
sos/sosreport.py:1299:17: E722 do not use bare except'
sos/sosreport.py:1413:13: E722 do not use bare except'
sos/sosreport.py:1434:13: E722 do not use bare except'
sos/sosreport.py:1479:13: E722 do not use bare except'
sos/utilities.py:43:9: E722 do not use bare except'
sos/policies/__init__.py:472:9: E722 do not use bare except'
sos/policies/__init__.py:489:9: E722 do not use bare except'
sos/policies/debian.py:44:9: E722 do not use bare except'
sos/policies/ibmkvm.py:42:9: E722 do not use bare except'
sos/policies/ibmkvm.py:67:9: E722 do not use bare except'
sos/policies/redhat.py:25:1: E722 do not use bare except'
sos/policies/redhat.py:124:13: E722 do not use bare except'
sos/policies/redhat.py:140:1: E305 expected 2 blank lines after class or function definition, found 1
sos/policies/redhat.py:211:9: E722 do not use bare except'
sos/policies/redhat.py:221:9: E722 do not use bare except'
sos/policies/redhat.py:260:9: E722 do not use bare except'
sos/policies/suse.py:67:13: E722 do not use bare except'
sos/policies/ubuntu.py:23:9: E722 do not use bare except'
sos/policies/ubuntu.py:36:9: E722 do not use bare except'
sos/plugins/__init__.py:943:9: E722 do not use bare except'
sos/plugins/dnf.py:31:13: E722 do not use bare except'
sos/plugins/ds.py:51:9: E722 do not use bare except'
sos/plugins/etcd.py:58:9: E722 do not use bare except'
sos/plugins/etcd.py:66:13: E722 do not use bare except'
sos/plugins/gluster.py:46:9: E722 do not use bare except'
sos/plugins/gluster.py:74:9: E722 do not use bare except'
sos/plugins/haproxy.py:52:9: E722 do not use bare except'
sos/plugins/infiniband.py:63:17: E722 do not use bare except'
sos/plugins/infiniband.py:72:17: E722 do not use bare except'
sos/plugins/logs.py:96:13: E722 do not use bare except'
sos/plugins/navicli.py:69:13: E722 do not use bare except'
sos/plugins/networking.py:35:9: E722 do not use bare except'
sos/plugins/networking.py:65:9: E722 do not use bare except'
sos/plugins/networking.py:155:9: E722 do not use bare except'
sos/plugins/networking.py:162:9: E722 do not use bare except'
sos/plugins/networking.py:276:9: E722 do not use bare except'
sos/plugins/nfsserver.py:34:9: E722 do not use bare except'
sos/plugins/ovn_central.py:28:17: E722 do not use bare except'
sos/plugins/pcp.py:51:9: E722 do not use bare except'
sos/plugins/pcp.py:60:13: E722 do not use bare except'
sos/plugins/pcp.py:67:9: E722 do not use bare except'
sos/plugins/powerpc.py:32:9: E722 do not use bare except'
sos/plugins/sapnw.py:16:5: E722 do not use bare except'
sos/plugins/sar.py:46:9: E722 do not use bare except'
```

Oh good lord... ISWYM.

I assumed the two were otherwise in sync, bar this one wart. Whose terrible idea was this mess?? :-)

bmr-cymru:
Tested log file extraction with and w/out `--all-logs`:

```
  Starting 1/1  : watchdog        [Running: watchdog]
[plugin:watchdog] collecting path '/etc/sysconfig/watchdog'
[plugin:watchdog] collecting path '/var/log/notwatchdog/qux.stdout'
[plugin:watchdog] collecting path '/usr/libexec/watchdog/scripts'
[plugin:watchdog] collecting path '/etc/watchdog.d'
[plugin:watchdog] collecting path '/etc/watchdog.conf'
  Finished running plugins
```

```
  Starting 1/1  : watchdog        [Running: watchdog]
[plugin:watchdog] collecting path '/etc/watchdog.d'
[plugin:watchdog] collecting path '/usr/libexec/watchdog/scripts'
[plugin:watchdog] collecting path '/etc/watchdog.conf'
[plugin:watchdog] collecting path '/etc/sysconfig/watchdog'
[plugin:watchdog] collecting path '/var/log/notwatchdog/qux.stdout'
[plugin:watchdog] collecting path '/var/log/notwatchdog/qux'
  Finished running plugins
```

And default (empty log dir):

```
  Starting 1/1  : watchdog        [Running: watchdog]
[plugin:watchdog] collecting path '/etc/sysconfig/watchdog'
[plugin:watchdog] collecting path '/usr/libexec/watchdog/scripts'
[plugin:watchdog] collecting path '/etc/watchdog.d'
[plugin:watchdog] collecting path '/etc/watchdog.conf'
  Finished running plugins
```

PR #1311, opened 2018-05-24
[origin] Collect *.conf under /etc/origin/node

nak3:
Since OCP 3.6+, there are some DNS config under /etc/origin/node/ such
as `node-dnsmasq.conf` and `resolv.conf`. Although current sosreport
does not gather them, these are important for troubleshooting of DNS
issues.

This patch updates origin.py to collect *.conf under
/etc/origin/node/.

---


nak3:
cc @codificat

sos/plugins/origin.py
```diff
@@ -150,6 +150,7 @@ def setup(self):
             self.add_copy_spec([
                 self.node_cfg,
                 os.path.join(self.node_base_dir, "*.crt"),
+                os.path.join(self.node_base_dir, "*.conf"),
```
codificat: Being a bit picky: the use of wildcards is usually discouraged, and in this case maybe we could reference the files with their full names? `{resolv,node-dnsmasq}.conf`

The existing `"*.crt"` wildcard is motivated by the fact that the cert file names are not fixed: they depend on the node's respective k8s object name.

Should we reference the files by full name instead?

nak3:
Thank you @codificat . Updated. (As it is just one line update, I already squashed.)

PR #1312, opened 2018-05-24
[logs] Revert "[logs] implement --all-logs for Red Hat messages* and secure*"

k-hagio:
```
commit 5d25dd31e731cc0d2d09a1b242803373937b4863 ("[logs] implement
--all-logs for Red Hat messages* and secure*") dropped rotated
/var/log/messages files by default.  This will make support enginners
ask their customers to collect them separately again and again because
these are the most important and the most freqently referenced files
for support engineers.  So it should be limited only with sizelimit
by default.

Also, if --all-logs is specified and the variable "messages" ends with
"*", os.path.exists(messages) always returns False, and it collects
journal unexpectedly.  So "*" should not be added to the variable
"messages" itself.

Signed-off-by: Kazuhito Hagio <k-hagio@ab.jp.nec.com>
```
I strongly recommend reverting commit 5d25dd3.
Is there any strong reason to drop rotated messages by default?

---

### DIFF
--- a/sos/plugins/logs.py
+++ b/sos/plugins/logs.py
@@ -79,14 +79,8 @@ class RedHatLogs(Logs, RedHatPlugin):
     def setup(self):
         super(RedHatLogs, self).setup()
         messages = "/var/log/messages"
-        secure = "/var/log/secure"
-
-        if self.get_option("all_logs"):
-            messages += "*"
-            secure += "*"
-
-        self.add_copy_spec(secure, sizelimit=self.limit)
-        self.add_copy_spec(messages, sizelimit=self.limit)
+        self.add_copy_spec("/var/log/secure*", sizelimit=self.limit)
+        self.add_copy_spec(messages + "*", sizelimit=self.limit)
 
         # collect three days worth of logs by default if the system is
         # configured to use the journal and not /var/log/messages


bmr-cymru:
> Is there any strong reason to drop rotated messages by default?

That's what `--all-logs` means: collect all log files found, including rotated logs.

There's currently a discussion ongoing about reworking limits and selection overall, and you're welcome to join in with that, but since this is the current definition of `--all-logs` the previous behaviour of this one plugin amounts to a bug: if anything we could consider an exception allowing it to collect only the _previous_ log file, on the understanding that this will be fixed when we address the other size and selection limitations, but it is confusing for users to have global options that do one thing with one plugin, and something else with another.

k-hagio:
@bmr-cymru
OK, thank you for the comment. I didn't realize that only `--all-logs` includes rotated logs. I will withdraw this patch.

> There's currently a discussion ongoing about reworking limits and selection overall, and you're welcome to join in with that,

Is there an issue or somewhere discussing about this?

There is the fact that users are used to run sosreport with no option and they think that rotated messages are included in sos archive.  So I think that this is a critical change at least to OS support people and it will increase their and customers' load considerably...

bmr-cymru:
> Is there an issue or somewhere discussing about this?

No - right now it's been taking place via IRC and email. I'll try to write up the current status and file it as an issue, so that other contributors can get involved.

> So I think that this is a critical change at least to OS support people and it will increase their and customers' load considerably...

I'm happy to make an exception for these files, at least until we have a better way of limiting the scope of files collected: rotated logs are useful, but they do tend to inflate the report size and runtime.

bmr-cymru:
If we had a reliable way of collecting the previous file only that could be a good compromise; it's a shame in a way that on rhel7 the default rotate stamp is a date rather than an integer.

bmr-cymru:
That's a possibility - I think we should discuss it along with the other current ideas around improving size limiting (e.g. the command size limits now in PR).

We're discussing this internally at Red Hat at the moment: we can either use `sos-devel@redhat.com` for mail discussion or we can get on a voice/video call if that suits people better.

bmr-cymru:
Closed in commit 5b5f41585e87fd9ea93092a07c69e4b05aaff96d

pmoravec:
+1 for the `messages[:-1]`, now the "if" condition returns True on any systemd machine.

bmr-cymru:
Ugh. Right. I'd forgotten about the need for that - fix pushed.

k-hagio:
Thank you for the quick response! I appreciate it.

Issue #1306, opened 2018-05-22
LXD needs a packages list


bmr-cymru:
```python
class LXD(Plugin, UbuntuPlugin):
    """LXD is a containers hypervisor.
    """
    plugin_name = 'lxd'
    profiles = ('container',)

```

Result: ENOPACKAGES.

slashdd:
Will update this PR shortly since LXD with recent release of Ubuntu can be installed from standard deb package (dpkg, apt, ... and family) or as a snap. I'll take benefit of this request to adapt the plugin for snap as well.

PR #1313, opened 2018-05-25
[lxd] Add check for the presence of lxd package

slashdd:
Closes: #1306

---

### DIFF
--- a/sos/plugins/lxd.py
+++ b/sos/plugins/lxd.py
@@ -16,6 +16,7 @@ class LXD(Plugin, UbuntuPlugin):
     """
     plugin_name = 'lxd'
     profiles = ('container',)
+    packages = ('lxd',)
 
     def setup(self):
         self.add_copy_spec([


bmr-cymru:
Thanks!

slashdd:
Note that lxd is also available as a snap in recent Ubuntu release. `packages = ('lxd',)` will only cover the case of lxd being installed the deb way. What if we add `files = ('/snap/bin/lxd',)` so that the plugin detect both scenario (deb package and snap)

bmr-cymru:
Should we think about adding support for triggering on snaps? Either as a new `snaps` member to `Plugin`, or as an extension to the Ubuntu policies to allow these to be recognised as packages (i.e. `PackageManager` will merge the two sources of package data).

slashdd:
Yeah, integrating into the current Ubuntu policies would make sense to me. I'll report a new issue about that topic and will leave that PR as is for now.

Issue #1304, opened 2018-05-21
[Plugin] don't collect rpm -V if verify list is empty


bmr-cymru:
```
[plugin:ovn_host] collecting output of 'journalctl --no-pager  --unit ovn-controller'
[plugin:ovn_host] collecting output of 'rpm -V '
```

Pointless make-work.

PR #1314, opened 2018-05-25
[Plugin] Do not run verify on empty package list

TurboTurtle:
In some instances, the regex match in all_pkgs_by_name_regex() was
returning an empty list when no match was made. This in turn caused sos
to run an empty verify command, e.g. 'rpm -V '.

By filtering the list of packages to verify, we can remove these empty
lists and avoid the empty verify command.

Closes: #1304

---

### DIFF
--- a/sos/policies/__init__.py
+++ b/sos/policies/__init__.py
@@ -161,7 +161,7 @@ class PackageManager(object):
             return None
 
         by_regex = self.all_pkgs_by_name_regex
-        verify_list = map(by_regex, packages)
+        verify_list = filter(None, map(by_regex, packages))
 
         # No packages after regex match?
         if not verify_list:


TurboTurtle:
This one was rather odd. Even though there was no match, `re.match(pkg)` was returning an empty list in some cases. For example, on my system which does not have the `ypserv` package, the `nis` plugin would return `[ [], [], [], [], [], [] ]` (five because nis isn't using a tuple for `packages` - will address that in a different commit). When I changed `nis` to using a tuple, I was still seeing a return of `[[]]`

Even stranger, if you print the matches from `all_pkgs_by_name_regex()`, you get this:
```
 Setting up archive ...
 Setting up plugins ...
<generator object <genexpr> at 0x7f5d783e3c80>
<generator object <genexpr> at 0x7f5d783e3c80>
<generator object <genexpr> at 0x7f5d783e3c80>
<generator object <genexpr> at 0x7f5d783e3cd0>
<generator object <genexpr> at 0x7f5d783e3d20>
<generator object <genexpr> at 0x7f5d783e3d70>

  Running 1/1: nis...        
```
I haven't found anything upstream similar to this behavior.

Filtering the `verify_list` returned by `all_pkgs_by_name_regex()` seems to me like the most straight-forward way to handle this oddity.

PR #1320, opened 2018-05-25
[sosreport] Add mechanism to encrypt final archive

TurboTurtle:
Adds an option to encrypt the resulting archive that sos generates.
There are two methods for doing so:

	--encrypt	Uses a key-pair for asymmetric encryption
	--encrypt-pass  Uses a password for symmetric encryption

For key-pair encryption, the key-to-be-used must be imported into the
root user's keyring, as gpg does not allow for the use of keyfiles.

If the encryption process fails, sos will not abort as the unencrypted
archive will have already been created. The assumption being that the
archive is still of use and/or the user has another means of encrypting
it.

---

### DIFF
--- a/man/en/sosreport.1
+++ b/man/en/sosreport.1
@@ -22,6 +22,8 @@ sosreport \- Collect and package diagnostic and support data
           [--log-size]\fR
           [--all-logs]\fR
           [-z|--compression-type method]\fR
+          [--encrypt-key KEY]\fR
+          [--encrypt-pass PASS]\fR
           [--experimental]\fR
           [-h|--help]\fR
 
@@ -120,6 +122,32 @@ increase the size of reports.
 .B \-z, \--compression-type METHOD
 Override the default compression type specified by the active policy.
 .TP
+.B \--encrypt-key KEY
+Encrypts the resulting archive that sosreport produces using GPG. KEY must be
+an existing key in the user's keyring as GPG does not allow for keyfiles.
+KEY can be any value accepted by gpg's 'recipient' option.
+
+Note that the user running sosreport must match the user owning the keyring
+from which keys will be obtained. In particular this means that if sudo is
+used to run sosreport, the keyring must also be set up using sudo
+(or direct shell access to the account).
+
+Users should be aware that encrypting the final archive will result in sos
+using double the amount of temporary disk space - the encrypted archive must be
+written as a separate, rather than replacement, file within the temp directory
+that sos writes the archive to. However, since the encrypted archive will be
+the same size as the original archive, there is no additional space consumption
+once the temporary directory is removed at the end of execution.
+
+This means that only the encrypted archive is present on disk after sos
+finishes running.
+
+If encryption fails for any reason, the original unencrypted archive is
+preserved instead.
+.TP
+.B \--encrypt-pass PASS
+The same as \--encrypt-key, but use the provided PASS for symmetric encryption
+rather than key-pair encryption.
 .TP
 .B \--batch
 Generate archive without prompting for interactive input.
--- a/sos/sosreport.py
+++ b/sos/sosreport.py
@@ -316,6 +316,13 @@ def _parse_args(args):
     preset_grp.add_argument("--del-preset", type=str, action="store",
                             help="Delete the named command line preset")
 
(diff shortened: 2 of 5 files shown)


sos/sosreport.py
```diff
@@ -740,13 +771,25 @@ def get_temp_file(self):
         return self.tempfile_util.new()
 
     def _set_archive(self):
+        if self.opts.encrypt_key and self.opts.encrypt_pass:
+            self.soslog.error("Password and key-pair encryption requested at "
+                              "the same time. Choose either but not both.")
+            raise SystemExit()
```
bmr-cymru: Could use a mutex group for these, rather than post-hoc validation? I just did this for presets, but I'm not married to it (makes the parser block a bit uglier, but means we don't have to check ourselves).

sos/sosreport.py
```diff
@@ -541,6 +543,28 @@ def compression_type(self, value):
...
+    @property
+    def encrypt_key(self):
+        if self._options is not None:
+            return self._options.encrypt_key
+        return self._encrypt_key
+
+    @encrypt_key.setter
+    def encrypt_key(self, value):
+        self._check_options_initialized()
+        self._encrypt_key = value
```
bmr-cymru: Sorry you had to write all this... It will go away with the presets patches :)

man/en/sosreport.1
```diff
@@ -119,6 +121,27 @@ increase the size of reports.
 .B \-z, \--compression-type METHOD
 Override the default compression type specified by the active policy.
 .TP
+.B \--encrypt KEY
+Encrypts the resulting archive that sosreport produces using GPG. KEY must be
+an existing key in the root user's keyring as GPG does not allow for keyfiles.
```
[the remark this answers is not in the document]
TurboTurtle: I had that originally, then thought that some users might get a bit frustrated if they ran `sudo sosreport` only to find out that the key they imported into *their* keyring won't work when using `sudo` (and they might have restrictive sudo configs)

sos/sosreport.py
TurboTurtle: Forgot all about that. Will update.

sos/sosreport.py
TurboTurtle: Yay! \o/

man/en/sosreport.1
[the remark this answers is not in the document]
bmr-cymru: I see what you mean, although in a way "that's how sudo works", but I don't think there's any harm to adding a follow-on paragraph noting this. Something like:

>"Note that the user running sosreport must match the user owning the keyring from which keys will be obtained. In particular this means that if sudo is used to run sosreport, they keyring must also be set up using sudo (or direct shell access to the account)."

Wording can be improved but I think we can handle this through docs. I'm not keen on it, but if that's felt to not be enough then we could implement a runtime check or warning "Could not encrypt: has <key> missing from <user> account keyring".

TurboTurtle:
Rebased to master following the SoSOptions changes. Moved `--encrypt` to `--encrypt-key` to be more consistent with `--encrypt-pass`, and added the note regarding keyrings and sudo.

The remaining outstanding bit is the passphrase in the gpg command we run. There's a way to pass it via `--passphrase-fd` and instead piping it into the command we shell-out to, but this would still be accessible via `/proc/` on the host, so I don't think it's really any better. Plus, the GPG mailing list seems to suggest against it if at all possible.

TurboTurtle:
Ok, after a little back and forth I believe I've updated this to address the concerns from the security review:

1) Prevent the encryption passphrase from leaking in process listings using passphrase-fd0
2) Prevent a passphrase like 'redhat -o /etc/THISISBAD' from overriding gpg options (in this case -o)

To do this we're using passphrase-fd 0 to read from STDIN via echo. This does however require the invocation of a bash shell as otherwise non-root users running sosreport with sudo would be unable to encrypt using a passphrase (as the command would literally just run an echo of the rest of the command). To prevent the passphrase from leaking in process listings, we set the passphrase to a sos_gpg env var passed to the `sos_get_command_output()` call.

I'm not thrilled about the bash invocation, but I am unable to find a better (or even working) approach when using passphrase-fd 0 .

Issue #1208, opened 2018-02-06
Run --verify by default to be 'on' and move current argument to --no-verify


iranzo:
Run verify by default so that we've detailed plugin data and rpm package modification for detected plugins and include an option to disable it if not wanted.

bmr-cymru:
Like I said on IRC, we will need performance numbers to evaluate the suitability of this change. You can help the PR to progress by preparing some: we'd need the number of packages installed (a package list would also help, esp. for any found corner cases), and the runtimes for `sos` with and without `--verify`.

iranzo:
@bmr-cymru do you think we could also include a timestamp inside sosreport so when generated, it wil include it?

We do something like that in citellus generated json for each plugin plus the total time (I will raise another issue if you think it's ok, as this could benefit other calculations in the future)

bmr-cymru:
We've had built-in profiling for plugins and command execution for literally years now... Just search for messages like:

```
[plugin:sar] collected output of 'sadf' in 0.007418632507324219
...
[plugin:psacct] collected plugin 'psacct' in 0.0055577754974365234
```

iranzo:
On collab-shell with 3.4-13.el7
time sosreport --batch
~~~
real	1m18.853s
user	0m15.281s
sys	0m5.649s
~~~

time sosreport --batch --verify
~~~
real	0m48.590s
user	0m19.361s
sys	0m4.898s
~~~

I guess it's because of caching of tiles by the FS

iranzo:
On undercloud-0, a VM inside openstack01 baremetal host

time sosreport --batch
~~~
real	10m21.760s
user	3m29.253s
sys	1m26.285s

~~~

time sosreport --batch --verify
~~~

real	11m16.479s
user	6m32.532s
sys	1m13.579s

~~~

iranzo:
Probably could be closed by: https://github.com/sosreport/sos/pull/1323

BryanQuigley:
Closing per last comment

Issue #1200, opened 2018-01-31
Add required files to detect pip installs on system


iranzo:
As we don't have rpm -V $(python-files) on the sosreport we cannot tell if a file was installed outside of the rpm package management

Alternativelly we could use:

~~~
It's not a perfect solution, but if installed-files.txt exists in the .egg-info for a library, it was most likely installed by pip. Won't catch easy_install or things installed via wheels, but it would be one more datapoint?
~~~

As suggested by cjones, but we will need to collect or report that presence of files in the sosreport

bmr-cymru:
Really?
```
$ rpm -qf /usr/lib64/python2.7/site-packages/cffi-1.9.1-py2.7.egg-info/installed-files.txt
python2-cffi-1.9.1-2.fc26.x86_64
```

bmr-cymru:
> As we don't have rpm -V $(python-files) on the sosreport we cannot tell if a file was installed outside of the rpm 

Why not add a verify list to the python plugin, and ask your users to run `sosreport --verify`? It's what it's for ;)

But anyway: this still won't do what you asked for. It will show you *packaged* files, but it's not going to report anything for *unpackaged* files at all. You'd need to take a global file list and subtract everything from the RPM data (and there are more efficient RPM options than --verify for this).

bmr-cymru:
Surely if a module is installed and is not in `installed-rpms.txt` then you know that it is either PIP or a local installation, isn't that sufficient?

Failing that then either `pip freeze` or `python -c 'help("modules")'` followed by comparison to the RPM data.

iranzo:
Will this be covered too with https://github.com/sosreport/sos/pull/1323 ?

bmr-cymru:
Not really: python packages still need to be on a verify list to be checked (unless `rpm.rpmva` is set).

I think the work @pierg75 was doing in #1066 is probably more of a solution for this.

TurboTurtle:
Between the `unpackaged` plugin, and the option to generate hashes for pyfiles in the `python` plugin (which can serve as a measure of validation), I'm considering this covered, and closing.

PR #1323, opened 2018-05-27
[sosreport] add command line presets

bmr-cymru:
This isn't completely ready to be merged as yet, but it's a big series and starting to come together: it should be in fairly solid testable shape and I wanted to start getting feedback on the general idea as well as the code.

This branch adds a command line preset facility to `sos`: a preset is a named collection of command line argument values that can be applied either by user action (`--preset=`), or by auto-detection through the system policy. This is intended to save support users and customers time, by quickly recalling commonly-used configurations that will collect the right data for a given situation.

This enables different products, or different types of case within a product, to use separate command line defaults: for e.g. presets can enable options like `--all-logs` that are normally disabled by default, enable or disable plugins, and adjust the temporary directory and other options. These can be shipped as product defaults, or be added by the user later.

Values entered on the command line will take precedence over preset values for all atomic valued arguments. Arguments accepting a list (`--onlyplugins`, `--profiles`, etc.) are merged with the user-specified values (so it is possible to extend a list of preset-defined plugins or profiles etc.).

Presets are displayed by running `sosreport --list-presets`:
```
[root@bmr-rhel7-vm1 tmp]# sosreport --list-presets

sosreport (version 3.5)

The following presets are available:

         name: rhosp
  description: Red Hat OpenStack Platform
         note: This preset may increase report size

         name: satellite
  description: Red Hat Satellite
         note: This preset may increase report size and run time

         name: rhel
  description: Red Hat Enterprise Linux

         name: logs-only
  description: 

         name: rhv
  description: Red Hat Virtualization
         note: This preset may increase report run time

```

With `--verbose` the preset arguments are displayed:

```
[root@bmr-rhel7-vm1 tmp]# sosreport --list-presets -v

sosreport (version 3.5)

The following presets are available:

         name: rhosp
  description: Red Hat OpenStack Platform
         note: This preset may increase report size
      options: --all-logs

         name: satellite
  description: Red Hat Satellite
         note: This preset may increase report size and run time
      options: --all-logs --verify

         name: rhel
  description: Red Hat Enterprise Linux
      options:

         name: logs-only
  description: 
      options: --all-logs --batch --onlyplugins=logs --verbosity=3

         name: rhv
  description: Red Hat Virtualization
         note: This preset may increase report run time
      options: --all-logs --verify

```

Presets can be defined by both policies and the user: policy presets are defined in the respective policy class source and are registered with the policy core at runtime. User defined presets are added and removed with the `--add-preset` and `--del-preset` options and are stored as JSON files (one preset per file, although the format can represent any number) by default at `/var/lib/sos/presets` (note: the build scripts have not yet been updated to create this directory automatically: it must be created manually before using presets). Preset files can also be copied into the directory by hand (e.g. from product support downloads).

To add a preset, type the `sosreport` command line as normal, but add `--add-preset <name>`. Presets are removed with `sosreport --del-preset <name>`. Built-in presets cannot be deleted by the user.

```
[root@bmr-rhel7-vm1 tmp]# sosreport -vvv --batch --debug --profile system --add-preset qux
set sysroot to '/' (default)

sosreport (version 3.5)

Added preset 'qux' with options -vvv --batch --debug --profile system

[root@bmr-rhel7-vm1 tmp]# sosreport --del-preset qux

sosreport (version 3.5)

Deleted preset 'qux'

# Adding a description and note
[root@bmr-rhel7-vm1 tmp]# sosreport -vvv --batch --debug --profile system --add-preset qux --desc "Debugging preset" --note "Traps to pdb"
set sysroot to '/' (default)

sosreport (version 3.5)

Added preset 'qux' with options -vvv --batch --debug --profile system --desc Debugging preset --note Traps to pdb

[root@bmr-rhel7-vm1 tmp]# sosreport --list-presets -v

sosreport (version 3.5)

The following presets are available:

[...]
         name: qux
  description: Debugging preset
         note: Traps to pdb
      options: --batch --debug=1 --profiles=system --verbosity=3
[...]

```
Policies that have not opted-in to the presets framework will continue to use the same command line defaults as before and these are now defined in a single location in the source.

---

### DIFF
--- a/sos/__init__.py
+++ b/sos/__init__.py
@@ -45,9 +45,9 @@ _sos = _default
 _arg_names = [
     'all_logs', 'batch', 'build', 'case_id', 'chroot', 'compression_type',
     'config_file', 'debug', 'enableplugins', 'experimental', 'label',
-    'list_plugins', 'list_profiles', 'log_size', 'noplugins', 'noreport',
-    'onlyplugins', 'plugopts', 'preset', 'profiles', 'quiet', 'sysroot',
-    'tmp_dir', 'usealloptions', 'verbosity', 'verify'
+    'list_plugins', 'list_presets', 'list_profiles', 'log_size', 'noplugins',
+    'noreport', 'onlyplugins', 'plugopts', 'preset', 'profiles', 'quiet',
+    'sysroot', 'tmp_dir', 'usealloptions', 'verbosity', 'verify'
 ]
 
 #: Arguments with non-zero default values
@@ -72,6 +72,7 @@ class SoSOptions(object):
     experimental = False
     label = ""
     list_plugins = False
+    list_presets = False
     list_profiles = False
     log_size = _arg_defaults["log_size"]
     noplugins = []
--- a/sos/__init__.py
+++ b/sos/__init__.py
@@ -16,7 +16,6 @@ gettext to internationalize messages.
 """
 
 import gettext
-from collections import deque
 
 __version__ = "3.5"
 
@@ -171,5 +170,20 @@ class SoSOptions(object):
             else:
                 if replace or not getattr(self, arg):
                     self._copy_opt(arg, src)
+
+    def dict(self):
+        """Return this ``SoSOptions`` option values as a dictionary of
+            argument name to value mappings.
+
+            :returns: a name:value dictionary of option values.
+        """
+        odict = {}
+        for arg in _arg_names:
+            value = getattr(self, arg)
+            # Do not attempt to store --add-preset <name> in presets
+            if arg == 'add_preset':
+                value = None
+            odict[arg] = value
+        return odict
 
 # vim: set et ts=4 sw=4 :
--- a/sos/__init__.py
+++ b/sos/__init__.py
@@ -16,6 +16,7 @@ gettext to internationalize messages.
 """
 
 import gettext
+import six
 
 __version__ = "3.5"
 
@@ -123,19 +124,23 @@ class SoSOptions(object):
             :param suffix: arbitrary suffix string
             :param literal: print values as Python literals
         """
-        fmt = prefix
-        arg_fmt = "=%s" if not quote else "='%s'"
+        args = prefix
+        arg_fmt = "=%s"
         for arg in _arg_names:
-            fmt += arg + arg_fmt + sep
-        fmt.strip(sep)
-        fmt += suffix
+            args += arg + arg_fmt + sep
+        args.strip(sep)
 
+        vals = [getattr(self, arg) for arg in _arg_names]
         if not quote:
             # Convert Python source notation for sequences into plain strings
-            args = [getattr(self, arg) for arg in _arg_names]
-            args = [",".join(a) if _is_seq(a) else a for a in args]
+            vals = [",".join(v) if _is_seq(v) else v for v in vals]
+        else:
+            def is_string(val):
+                return isinstance(val, six.string_types)
+            # Only quote strings if quote=False
+            vals = ["'%s'" % v if is_string(v) else v for v in vals]
 
-        return fmt % tuple(args)
+        return (args % tuple(vals)).strip(sep) + suffix
 
     def __str__(self):
         return self.__str()
--- a/sos/__init__.py
+++ b/sos/__init__.py
@@ -55,6 +55,7 @@ _arg_defaults = {
     "log_size": 10,
     "chroot": "auto",
     "compression_type": "auto",
+    "preset": "auto"
 }
 
 
--- a/sos/__init__.py
+++ b/sos/__init__.py
@@ -59,6 +59,13 @@ _arg_defaults = {
 }
 
 
+def _is_seq(val):
+    """Return true if val is an instance of a known sequence type.
+    """
+    val_type = type(val)
+    return val_type is list or val_type is tuple
+
+
 class SoSOptions(object):
     add_preset = ""
     alloptions = False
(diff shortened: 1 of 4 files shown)


TurboTurtle:
I really like this, and being able to define plugin options for presets is going to be really powerful for us.

My only question is - are policies the right place to define these? Or should (at least builtins) be defined like how we do plugins? My thought for this is since presets aren't going to be limited strictly to products, we could have presets like `cantboot` or `panicing` or similar, that could be applicable to any distribution. Having preset defined in their own space could make maintaining these easier.

bmr-cymru:
I guess implementation-wise my approach for those types of case would be to add them to `LinuxPolicy`, or even `Policy` (since we don't really support any non-*nix like platforms any more): the default strategy for policy subclasses is to add their presets using `dict.update()`, so they are merged with the presets of all parent classes (so you can have presets for "all linux distros", "all Red Hat distros", "all RHEL", etc.). I should probably abstract this out as a `Policy.register_presets()` method to make this more explicit, but I did want to have the profiles in charge of it, rather than making it automagic - that way profiles can completely disable presets if they want to, or edit/clear the list before starting.

In terms of how it came to be this way, initially I was calling the feature "products", so it was natural to make the association with the policy classes since products are defined by the same bodies as the distro policies. I get what you're saying, and if anything it might be worth moving `add_preset()` and `del_preset()` out of the Policy class itself - either into a function or another module, as these really are a bit orthogonal to the built-in presets.

I like the fact that the preset definitions in the source are all in one place - this is one downside of the profile implementation imho: it was easy to write, and it's easy for plugin authors, but I end up either using `sosreport` or running complicated little pipelines to get an overall view of them. This ended up a bit of a theme in the whole series: like putting all the non-null command line defaults in one dictionary.

bmr-cymru:
Hmm, the one reason to have `add_preset()` and `del_preset()` in the policies class at all is that the policy should probably have control over the `/var/lib/sos/presets` path (unless it's a build time setting).

bmr-cymru:
I've rebased the branch, merging some fixups into their parent commits, and started shuffling some of the general changes out of the branch and into master. I'll carry on with this over this week while I'm doing the other reviews and see if I can make the PR a bit easier to review (the changes to `SoSOptions` and then move to `sos` is unfortunate, but I'm not sure it's worth dealing with the conflicts to eliminate).

bmr-cymru:
No, that's not quite right: there are two distinct types of preset:

  * Product/profile presets (stored in the source code of the relevant policy)
  * User presets (stored by default in `/var/lib/sos/presets`

Product presets are under the control of policy authors and can exist at any level: `Policy`, `LinuxPolicy`, `RedHatPolicy`, `RHELPolicy`, `FedoraPolicy`, `RedHatAtomicPolicy`, etc. The presets of lower levels of the class hierarchy are merged with those of the upper levels by default (so `RHELPolicy` has all of the `Policy`, `LinuxPolicy`, `RedHatPolicy` presets).

In source they look like this:

```python
rh_presets = {
    "": PresetDefaults(name="rhel"),
    "rhel7": PresetDefaults(name="rhel7"),
    "rhosp": PresetDefaults(name="rhosp", opts=SoSOptions(all_logs=True)),
    "rhv": PresetDefaults(name="rhv",
                          opts=SoSOptions(all_logs=True, verify=True))
}
```

(a dict of `name:PresetDefaults()` mappings).

User presets are for end-users to store their own preferences for later recall (or for a support provider to send over a ready-made preset file to drop in `/var`. The only connection they have to profiles is that the profile controls the path in `/var` where profiles are stored, so for now it also has the load and store methods. Only user presets are stored on-disk in JSON notation:

```json
{
    "chroot-never": {
        "note": "",
        "args": {
            "list_profiles": false, "log_size": 10, "quiet": false, "tmp_dir": "",
            "enableplugins": [], "compression_type": "auto", "list_presets": false,
            "chroot": "never", "plugopts": [], "verify": false, "all_logs": false, "label": "",
            "note": null, "case_id": "", "build": false, "sysroot": null, "onlyplugins": [],
            "add_preset": null, "config_file": "", "list_plugins": false, "noreport": false,
            "alloptions": false, "desc": null, "del_preset": null, "verbosity": 1,
            "batch": false, "profiles": [], "preset": "", "experimental": false,
            "debug": false, "noplugins": []
        },
        "desc": ""
    }
}
```

> Will sosreport detect multiple presets of the same name in /var/lib/sos/presets

No, because that's impossible ;-) "preset_name" -> "file_name".

But yes, sos will check for that and raise an error:

```
# sosreport -vvv --debug --add-preset vvverbose-debug
set sysroot to '/' (default)

sosreport (version 3.5)

A preset named 'vvverbose-debug' already exists
```

> The 2nd point isn't covered in either way, am I right? I would ideally see a product deploying preset and also enabling it by default

Yes, it is:

```
commit e366e499984c9928104a18335554f6c6565f6d17
Date:   Sat May 26 20:35:33 2018 +0100

    [sos] add automatic preset detection infrastructure
    
    Add a new special value for the --preset argument, 'auto', which
    enables automatic detection of the preset identifier string by
    the policy class.
    
    A new method, probe_preset() is added which should be implemented
    by derived classes of Policy to return a ProductDefaults() object
    initialised appropriately for the running preset.
    
```

The `auto` value is default and enables profile driven auto-detection. Auto detection for profiles that wish to opt-in is enabled by adding a `probe_preset()` method to the relevant policy class, e.g.:

```python
    def probe_preset(self):
        # Package based checks
        if self.pkg_by_name("satellite-common") is not None:
            return self.find_preset(SATELLITE)

        # Vanilla RHEL is default
        return self.find_preset(RHEL)

```

These are just dummies to have something for testing: obviously we need to come up with more checks & their corresponding values.

> by extending presets by default option with the same impact like calling sosreport 

Way ahead of you dude ;-)

> I wanted to try testing it / play with the feature, but applying the patch to current upstream, I always get:
> 
> ```
> # ./sosreport --batch --debug --profile system --add-preset qux2
> 
> sosreport (version 3.5)
> 
> Could not add preset: Cannot write built-in preset
> #
> ```

Uggh. I'm sorry about that: I broke that in yesterday's rebase and hadn't notice. I'll get a fix pushed asap.

iranzo:
Should this also include Close for #1208  and #1200 ?

bmr-cymru:
Merged

PR #1326, opened 2018-05-31
[mssql] Add new plugin

tanaka-takayoshi:
Adds a new plugin for Microsoft SQL Server on Linux.

Collect log files.
Collect journal logs.
Collect keytab file info if AD auth is configured.

Signed-off-by: Takayoshi Tanaka <tatanaka@redhat.com>

---

### DIFF
--- a/sos/plugins/mssql.py
+++ b/sos/plugins/mssql.py
@@ -1,0 +1,98 @@
+# Copyright (C) 2018 Red Hat, K.K., Takayoshi Tanaka <tatanaka@redhat.com>
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+from sos.plugins import Plugin, RedHatPlugin
+import os
+
+
+class MsSQL(Plugin, RedHatPlugin):
+    """Microsoft SQL Server on Linux
+    """
+
+    plugin_name = "mssql"
+    profiles = ('services',)
+    packages = ('mssql-server',)
+
+    option_list = [
+        ('mssql_conf', 'SQL Server configuration file.', '',
+         '/var/opt/mssql/mssql.conf')
+    ]
+
+    def setup(self):
+        mssql_conf = self.get_option('mssql_conf')
+
+        # Pick error log file from mssql_conf.
+        # Expecting the following format
+        # ```
+        # [filelocation]
+        # errorlogfile = /var/opt/mssql/log
+        # [sqlagent]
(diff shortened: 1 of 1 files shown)


sos/plugins/mssql.py
```diff
@@ -0,0 +1,94 @@
...
+        # errorlogfile = /var/opt/mssql/log
+        # [sqlagent]
+        # errorlogfile = /var/opt/mssql/log/sqlagentstartup.log
+        # [network]
+        # kerberoskeytabfile = /var/opt/mssql/secrets/mssql.keytab
+        # ```
+        # Default values are described in the document:
+        # https://docs.microsoft.com/en-us/sql/linux
+        # /sql-server-linux-configure-mssql-conf
+        # ?view=sql-server-linux-2017#mssqlconf-format
```
bmr-cymru: Let's just drop this URL; interested parties can easily find it if they need it. It's just a plan INI file, so we can simply document the sections and keys that we care about.

sos/plugins/mssql.py
```diff
@@ -0,0 +1,94 @@
...
+                self.add_cmd_output('ls -l %s' % kerberoskeytabfile)
+                self.add_cmd_output('klist -e -k %s' % kerberoskeytabfile)
+            else:
+                self._log_error(keytab_err)
+
+        # Expecting mssql_conf doesn't includeno sensitive information.
+        self.add_copy_spec([
+            mssql_conf,
+            errorlogfile + '/*',
+            sqlagent_errorlogfile
```
bmr-cymru: This should only add all the log files if `--all-logs` is given. Please also implement size limiting for the log files only (no need for configuration). Running `git grep 'get_option.*log_size'` will give you plenty of examples to look at.

sos/plugins/mssql.py
```diff
@@ -0,0 +1,94 @@
...
+                    if section == '[filelocation]':
+                        errorlogfile = words[1].strip()
+                    elif section == '[sqlagent]':
+                        sqlagent_errorlogfile = words[1].strip()
+                elif words[0].strip() == 'kerberoskeytabfile':
+                    if section == '[network]':
+                        kerberoskeytabfile = words[1].strip()
+        except IOError as ex:
+            self._log_error('Could not open conf file %s: %s' % (mssql_conf,
+                                                                 ex))
```
bmr-cymru: This is a little more complex than we'd generally like for plugin config reading code, but the alternative would be bringing in something like `ConfigParser` which isn't ideal right now: with any luck we will be able to replace this with a generic config file reading facility in the near future.

sos/plugins/mssql.py
```diff
@@ -0,0 +1,94 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+import os
+from sos.plugins import Plugin, RedHatPlugin
```
bmr-cymru: Generally in `sos` we import the `sos` modules first and then general Python modules. This is not the usual Python convention but it's something we've been doing for a long time - sticking to it means that if we do decide to change in the future it is easier to apply a scripted change to all source files at once.

sos/plugins/mssql.py
tanaka-takayoshi: I pushed a new commit as per comment. However, should I fix this code right now?

bmr-cymru: It is quite complex but I think it is OK for now: we do not yet have the generic config reading work done, so there is no better solution for now.

PR #1327, opened 2018-06-04
[filesys] Add permission and selinux check for /tmp

TurboTurtle:
Adds a check for the /tmp directory for permissions and selinux
context. If either of these are incorrect, the system in question can
exhibit strange behavior and otherwise be difficult for support teams to
detect.

On systems where SELinux is disabled, the permission check is still
relevant and the command still produces useful output with the -Z
option.

---

### DIFF
--- a/sos/plugins/filesys.py
+++ b/sos/plugins/filesys.py
@@ -9,7 +9,7 @@
 from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
 
 
-class Filesys(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
+class Filesys(Plugin, DebianPlugin, UbuntuPlugin):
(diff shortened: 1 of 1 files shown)


sos/plugins/filesys.py
```diff
@@ -38,7 +38,8 @@ def setup(self):
         self.add_cmd_output([
             "df -ali -x autofs",
             "findmnt",
-            "lslocks"
+            "lslocks",
+            "ls -ltradZ /tmp"
```
bmr-cymru: What about systems where SELinux is disabled at *build* time? Presumably we'll get a file with an error in it - not the end of the world, but it's always good to avoid leaving those lying around.

Part of me is inclined to say that we should move this into a new `RedHatFilesys` class: we know that all Red Hat builds all enable the `-Z` switch but other distributions may not.

PR #1329, opened 2018-06-05
 [sosreport] Handle cases when creating archive fails

pmoravec:
---


pmoravec:
Reproducer scenario being fixed: concurrently removing `tmp-dir` when `sosreport` is about to call `archive = self.archive.finalize(..)`.

The two auxiliary commits fix other consequences of `tmp-dir` being gone and improve user experience in getting more information what failed. (anyway I dont stick to the logging change in exception handling).

PR #1330, opened 2018-06-05
[Ubuntu] Add ubuntu-advantage status and change u-s-s to better default

BryanQuigley:
Add ubuntu-advantage status which is installed by default on Ubuntu and
shows when various UA items are enabled - esm, fips, and livepatch.

Switch ubuntu-support-status to always --show-all. It takes exactly
the same amount of time to run (in both cases it's relatively slow at
3 seconds), and only adds 30k uncompressed/12k compressed.

---


BryanQuigley:
@battlemidget as the original author, do these changes make sense to you?

Issue #1333, opened 2018-06-07
New threads functionaility doesn't work if --config-file unspecified


BryanQuigley:
If I have a file /etc/sos.conf
$ sudo ./sosreport
....
Traceback (most recent call last):
  File "./sosreport", line 19, in <module>
    main(sys.argv[1:])
  File "/home/bryan/git/sos/sos/sosreport.py", line 1695, in main
    sos.execute()
  File "/home/bryan/git/sos/sos/sosreport.py", line 1670, in execute
    self.collect()
  File "/home/bryan/git/sos/sos/sosreport.py", line 1285, in collect
    self.plugpool = ThreadPoolExecutor(self.opts.threads)
  File "/home/bryan/git/sos/sos/sosreport.py", line 549, in threads
    return self._threads
AttributeError: 'SoSOptions' object has no attribute '_threads'

Whereas if I specify it sudo ./sosreport --config-file /etc/sos.conf
....
It works fine.

BryanQuigley:
I did try just adding
    _threads = 4
to class SoSOptions(object) in sosreport.py

which fixes it, but somehow messes up formatting output (all plugins on single line)...

TurboTurtle:
Huh. So this is a trivial fix:

```
diff --git a/sos/sosreport.py b/sos/sosreport.py
index 75e33e83..71984d32 100644
--- a/sos/sosreport.py
+++ b/sos/sosreport.py
@@ -225,6 +225,7 @@ class SoSOptions(object):
     _sysroot = None
     _chroot = 'auto'
     _compression_type = 'auto'
+    _threads = 4
 
     _options = None
```

But what I'm surprised at is I only see this behavior when I run from master. If I run from the feature branch the PR was made from, sosreport runs normally, even though as far as I can tell no other recent commits touched anything that the threads commit touched.

```
$ git branch
  atomic-info
  atomic-rpm
  block-udev
  buildah
* concurrent
[...]
$ sudo ./sosreport --batch

sosreport (version 3.5)

 Running plugins. Please wait ...

  Starting 1/98 : abrt            [Running: abrt]
  Starting 2/98 : alternatives    [Running: abrt alternatives]
  Starting 3/98 : anaconda        [Running: abrt alternatives anaconda]
  Starting 4/98 : anacron         [Running: abrt alternatives anaconda anacron]
  Starting 5/98 : ansible         [Running: abrt alternatives anaconda ansible]
[...]

PR #1336, opened 2018-06-08
[sosreport] Fix threads execution with no options and output formatting

TurboTurtle:
Relevant to #1333 - this fixes running with threads with a bare `sosreport` command, and also fixes the mangled output for plugin status. 

Even with the `SoSOption()` changes coming with presets, the string formatting changes here would be needed to not mangle the output.

Resolves: #1333 

---

### DIFF
--- a/sos/sosreport.py
+++ b/sos/sosreport.py
@@ -225,6 +225,7 @@ class SoSOptions(object):
     _sysroot = None
     _chroot = 'auto'
     _compression_type = 'auto'
+    _threads = 4
 
     _options = None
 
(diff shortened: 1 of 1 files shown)


TurboTurtle:
For reference, `ljust(90)` is used because due to the length of some plugin names, the output line can be upwards of 86 characters from my tests. If we have a line longer than the `ljust()` setting, we end up with new lines being printed until an output line gets back to be at or below that setting.

This can lead to confusing output where for example plugins 60-63 (or whatever) will be printed on individual lines, with the line immediately following it being able to be refreshed, thus at the end showing "Finishing plugins" with an entirely separate list of running plugins, which could lead to confusion as to if sos actually ran all the loaded plugins or not.

Issue #1196, opened 2018-01-29
Add usage of Threads to "xz"-command on archive-routine


knumskull:
On systems, with big amount of logfiles, running `sosreport` will take a lot of time.
A lot of time is lost during the archiving-process. This could be improved by using threads for `xz`-command easily.

A possible way could be to use 20% of the available cores for the archiving process.

~~~
# diff ~/archive.py /usr/lib/python2.7/site-packages/sos/archive.py
28,29d27
< import multiprocessing
< import math
437,439d434
<             # use multiple threads if using xz
<             if cmd == "xz":
<                 cmd = "%s -T%d" % (cmd,math.ceil(multiprocessing.cpu_count()/5))
~~~

TurboTurtle:
Looks like `xz-5.2` is what added threaded compression - which looks like it was in F24 and RHEL 7.3. I'm not sure what Debian/Ubuntu releases it ships in.

Thinking we should do this at the same time we address running plugins in parallel, @bmr-cymru @pmoravec

pmoravec:
Running plugins in parallel and this improvement are imho independent on each other but both aiming the same direction - they can just share sosreport's setting of # of threads (that would make sense imho).

Why use math.ceil, if dividing returns integer value every time? What about:

- passing new sosreport option `threads` via `archive = self.archive.finalize(..)` call
- updating relevant code to:


```
    if cmd == "xz":
        if threads == None:   # or 0, simply cmdline option not set
            threads = multiprocessing.cpu_count()/5
            if threads == 0:
                threads = 1  # so we dont need to use math.ceil and import math
        cmd = "%s -T%d" % (cmd, threads)
```

knumskull:
Yes, makes much more sense. Forgot about python is doing the right math here.

TurboTurtle:
We'd also want to check to xz version - if a user has an old xz package without this check, we'll fail to create the archive every time.

pmoravec:
Depends on #1199 , once that is merged, we can propose patch here like:

    --- /usr/lib/python2.7/site-packages/sos/sosreport.py.orig	2018-03-03 14:50:14.271792916 +0100
    +++ /usr/lib/python2.7/site-packages/sos/sosreport.py	2018-03-03 14:54:44.338248720 +0100
    @@ -769,9 +769,9 @@ class SoSReport(object):
                                         self.policy.get_archive_name())
             if self.opts.compression_type == 'auto':
                 auto_archive = self.policy.get_preferred_archive()
[...]

PR #1338, opened 2018-06-11
[archive] use threads for xz compression

pmoravec:
Resolves: #1196

---

### DIFF
--- a/sos/archive.py
+++ b/sos/archive.py
@@ -133,9 +133,11 @@ class FileCacheArchive(Archive):
     _archive_root = ""
     _archive_name = ""
 
-    def __init__(self, name, tmpdir):
+    def __init__(self, name, tmpdir, policy, threads):
         self._name = name
         self._tmp_dir = tmpdir
+        self._policy = policy
+        self._threads = threads
         self._archive_root = os.path.join(tmpdir, name)
         os.makedirs(self._archive_root, 0o700)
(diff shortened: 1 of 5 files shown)


sos/archive.py
```diff
@@ -133,9 +133,10 @@ class FileCacheArchive(Archive):
     _archive_root = ""
     _archive_name = ""
 
-    def __init__(self, name, tmpdir):
+    def __init__(self, name, tmpdir, threads):
```
bmr-cymru: You can't just stuff this into `FileCacheArchive`; it's part of a class hierarchy that defines interfaces: they all need the same initialiser arguments.

pmoravec:
Finally fixed travis failures, now the PR should be OK.

sos/sosreport.py
```diff
@@ -757,11 +757,22 @@ def get_temp_file(self):
     def _set_archive(self):
         archive_name = os.path.join(self.tmpdir,
                                     self.policy.get_archive_name())
+        # determine number of threads to use for compressing - applicable
+        # only for xz and of version 5.2 or higher
+        try:
+            xz_version = self.policy.package_manager\
+                         .all_pkgs()['xz']["version"]
```
BryanQuigley: xz-utils is the package name on Ubuntu/Debian.

pmoravec:
Updated PR according to the archive/policy discussion.

PR #1341, opened 2018-06-11
[fibrechannel] test if dir exists before listing it

pmoravec:
Resolves: #1341

---

### DIFF
--- a/sos/plugins/fibrechannel.py
+++ b/sos/plugins/fibrechannel.py
@@ -29,7 +29,8 @@ class Fibrechannel(Plugin, RedHatPlugin):
         ]
 
         for loc in dirs:
-            devs.extend([loc + device for device in os.listdir(loc)])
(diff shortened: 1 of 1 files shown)


pmoravec:
To prevent below uncaught plugin exception when enabling the plugin on not-FC-system:

    $ cat fibrechannel-plugin-errors.txt
      File "/root/sos-master/sos/sosreport.py", line 1245, in setup
        plug.setup()
      File "/root/sos-master/sos/plugins/fibrechannel.py", line 32, in setup
        devs.extend([loc + device for device in os.listdir(loc)])
    OSError: [Errno 2] No such file or directory: '/sys/class/fc_host/'
    $

pmoravec:
This still raises the same exception :(

bmr-cymru:
Closed in commit f57e6df8b59fe81024cac58bf81bcc536e6dafc5

PR #1344, opened 2018-06-13
[global] Transition to pycodestyle

TurboTurtle:
The `pep8` tool has been renamed to `pycodestyle` to avoid confusion
with pep8 the document. pep8 (the tool) will no longer be updated and as
such we should transition to the new pycodestyle.

This commit fixes a number of PEP8 issues with the new pycodestyle
linter that has tests for new additions to PEP8 that the pep8 tool did
not include. Most of these are exception handling fixes like:

	sos/plugins/logs.py:96:13: E722 do not use bare 'except'

Which are relatively straight forward to address, by instead catching
just the applicable exceptions, e.g. IOError, ValueError, etc... In
cases where there were many possible exceptions or where the "bare
'except'" was a final catch-all, we now use 'except Exception' to catch
anything from the base exception class.

Another frequent correction was for escape sequencing, such as:

	sos/plugins/logs.py:48:15: W605 invalid escape sequence '\S'

The solution for this is to mark these regex strings as raw strings,
which is preferred by `re` as future releases of that module may/will
become more strict in its parsing of regex strings.

---


TurboTurtle:
I imagine there will be a few more corrections needed as we close out 3.6, so I am opening this now to get the discussion started rather than proposing this for 3.6 as well. I'll keep this updated and rebased as needed until then.

Issue #480, opened 2015-01-28
[plugins] investigate exception propagation from _copy_dir()


bmr-cymru:
Commit a668851 works around a problem that causes sos to abruptly terminate in Docker container environments:

```
  >>> import os
  >>> os.listdir("/host/proc/sys/fs/binfmt_misc/")
  Traceback (most recent call last):
    File "<stdin>", line 1, in <module>
  OSError: [Errno 2] No such file or directory: '/host/proc/sys/fs/binfmt_misc/'
```

Handling the OSError in `_copy_dir()` itself gives the desired behaviour but it's unclear why this wasn't already caught (and either re-raised or logged but allowing the rest of the sos run to continue) by the existing high-level exception handling.

bmr-cymru:
Where is the original backtrace? That would tell us why it was not caught.

BryanQuigley:
cat sos_logs/kernel-plugin-errors.txt 
Traceback (most recent call last):
  File "/root/sos/sos/sosreport.py", line 1320, in collect_plugin
    plug.collect()
  File "/root/sos/sos/plugins/__init__.py", line 940, in collect
    self._collect_copy_specs()
  File "/root/sos/sos/plugins/__init__.py", line 898, in _collect_copy_specs
    self._do_copy_path(path)
  File "/root/sos/sos/plugins/__init__.py", line 435, in _do_copy_path
    self._copy_dir(srcpath)
  File "/root/sos/sos/plugins/__init__.py", line 387, in _copy_dir
    raise e
  File "/root/sos/sos/plugins/__init__.py", line 378, in _copy_dir
    for afile in os.listdir(srcpath):
PermissionError: [Errno 13] Permission denied: '/sys/fs/pstore'

bmr-cymru:
Thanks!

bmr-cymru:
Two possible things here - am in a meeting right now but I'll try to update later.

BryanQuigley:
I'm currently looking at just moving that bit to walk - https://github.com/BryanQuigley/sos/commit/82633f51adf8412e7cb4907999390720247c254e

seems to work in my testing so far..  need to add reporting for errors..  It may actually be possible to move all of copy_path to walk but testing that was complicated...

BryanQuigley:
@slashdd in 3.7 testing I refound this for LXD containers.   Don't see why it moved from kernel->system though (was system in 3.6 afaict)

sos_logs/system-plugin-errors.txt
Traceback (most recent call last):
[...]

PR #1348, opened 2018-06-14
[plugins] Check read access before trying to list a dir

BryanQuigley:
I found a permission error when trying to copy /sys/fs/pstore when
running on LXD containers.  Adding a permission check first allows
it to skip folders it can't list.

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -431,7 +431,7 @@ class Plugin(object):
             self._copy_symlink(srcpath)
             return
         else:
-            if stat.S_ISDIR(st.st_mode):
+            if stat.S_ISDIR(st.st_mode) and os.access(srcpath, os.R_OK):
                 self._copy_dir(srcpath)
                 return
 


BryanQuigley:
This does seem similar to https://github.com/sosreport/sos/issues/480 and I'm fine redoing it as exception handling instead.  Going to look into see if we can handle the general case better too..

BryanQuigley:
Looking into changing _do_copy_path by using os.walk.. instead of this.

bmr-cymru:
Can we get the full original backtrace please?

sos/plugins/__init__.py
```diff
@@ -431,7 +431,7 @@ def _do_copy_path(self, srcpath, dest=None):
             self._copy_symlink(srcpath)
             return
         else:
-            if stat.S_ISDIR(st.st_mode):
+            if stat.S_ISDIR(st.st_mode) and os.access(srcpath, os.R_OK):
```
bmr-cymru: This doesn't feel entirely right. What is _actually_ blocking the access here? Some part of the container compartmentalisation? On Linux root has always been able to bust through file system permissions and we rely on that to a fairly heavy extent - this change may have a side effect of causing us to not collect other data that we _can_ read simply because the permissions are either restrictive for some reason or broken. It is a **major** advantage for a tool like sos to be able to do this and to properly collect files even if the administrator has done something like `chmod 0000`.

bmr-cymru: OK: I did some testing & discussed this with the Red Hat `sos` folks on a call today. Luckily it seems that `os.access()` uses `access(2)`, and that that appears to give the correct result, even if the root user is relying on being able to bust through `000` file modes.

I'll merge the patch today and we can keep a close eye for any problems in the time remaining to 3.6 (end of the week).

PR #1349, opened 2018-06-15
[sosreport] call traceback.print_exc without arguments

pmoravec:
to prevent uncaught exception on py3

Resolves: #1349

---

### DIFF
--- a/sos/sosreport.py
+++ b/sos/sosreport.py
@@ -1228,10 +1228,9 @@ class SoSReport(object):
             else:
                 raise e
         except Exception as e:
-            import traceback
             self.ui_log.error("")
             self.ui_log.error(" Unexpected exception setting up archive:")
-            traceback.print_exc(e)
+            traceback.print_exc()
             self.ui_log.error(e)
         self._exit(1)
 


pmoravec:
Reproducer (artificial, now): revert back https://github.com/sosreport/sos/commit/68dbbf97bc88e11f067753cd6496842633c3410e :) and run sosreport under python3:

    # sosreport --batch
    ..
     Setting up archive ...
    
     Unexpected exception setting up archive:
      File "/usr/lib/python3.6/site-packages/sos/sosreport.py", line 1208, in prework
        self._set_archive()
      File "/usr/lib/python3.6/site-packages/sos/sosreport.py", line 750, in _set_archive
        self.policy.get_archive_name())
      File "/usr/lib/python3.6/site-packages/sos/policies/__init__.py", line 297, in get_archive_name
        rand = ''.join(random.choice(string.lowercase) for x in range(7))
      File "/usr/lib/python3.6/site-packages/sos/policies/__init__.py", line 297, in <genexpr>
        rand = ''.join(random.choice(string.lowercase) for x in range(7))
    AttributeError: module 'string' has no attribute 'lowercase'
    
    During handling of the above exception, another exception occurred:
    
      File "/usr/sbin/sosreport", line 25, in <module>
        main(sys.argv[1:])
      File "/usr/lib/python3.6/site-packages/sos/sosreport.py", line 1625, in main
        sos.execute()
      File "/usr/lib/python3.6/site-packages/sos/sosreport.py", line 1598, in execute
        self.prework()
      File "/usr/lib/python3.6/site-packages/sos/sosreport.py", line 1226, in prework
        traceback.print_exc(e)
      File "/usr/lib64/python3.6/traceback.py", line 159, in print_exc
        print_exception(*sys.exc_info(), limit=limit, file=file, chain=chain)
      File "/usr/lib64/python3.6/traceback.py", line 100, in print_exception
        type(value), value, tb, limit=limit).format(chain=chain):
      File "/usr/lib64/python3.6/traceback.py", line 497, in __init__
        capture_locals=capture_locals)
      File "/usr/lib64/python3.6/traceback.py", line 332, in extract
        if limit >= 0:
    TypeError: '>=' not supported between instances of 'AttributeError' and 'int'


The first traceback is due to the fixed #1297, but any such error raises the 2nd traceback caused by improper calling `traceback.print_exc()` that shouldnt take the exception as argument.

(also "import traceback" is redundant as it is already imported at the beginning of file)

bmr-cymru:
More to the point: why are we printing a full exception in non `--debug` mode _at all_ (we should just say that it happened, and print the `str()` of it - we only inflict backtrace spew on the user if they asked for it with `--debug`), and while we are there, why is it using a completely different call from the one that's used on line 794 in the same file (`SoSReport._exception()`)?

bmr-cymru:
(in fact, the more I look at the `sosreport` exception handling, the more it looks quite bogus... It's very old, and some of it dates back to before `2.2`).

PR #1355, opened 2018-06-20
[sosreport] Correct verbosity level when -v is not specified

TurboTurtle:
Fixes an issue where verbosity would be set to 'None' rather than 0 in
the event that a sosreport was run with options, when those options did
not include -v at all. This meant that if any options were specified, a
newline would be printed for each started plugin when instead we should
still have been refreshing a single line.

Related to this, the previous newline behavior was beneficial when
running with --batch, so that behavior is maintained if --batch is
specified.

---


TurboTurtle:
This, together with #1354, should resolve the remaining concerns for #1333

PR #1356, opened 2018-06-20
[rpm] Fix awk string escaping

TurboTurtle:
With the transition to pycodestyle, we needed to make the filter_cmd
string that defined the awk string a raw string. This in turn broke the
escaping and would result in the following for the installed-rpms file:

    awk: cmd. line:1: {printf \%-59s
    awk: cmd. line:1:         ^ syntax error

This fixes the escaping so that pycodestyle still passes the check on
the string, and that awk still accepts the string and we get the proper
output in the installed-rpms file.

---


bmr-cymru:
Merged the same patch this morning before I'd seen your PR - thanks ;)

PR #1357, opened 2018-06-21
[plugin] Use log_size for limiting size of journal

TurboTurtle:
Corrects the _add_cmd_output() call from add_journal() to use the
log_size determined by comparing sos options, sizelimit, and the default
journal size instead of always using sizelimit.

---


TurboTurtle:
Handled in 6c22ae8

PR #1360, opened 2018-06-23
[plugins] Remove unnecessary sizelimits

TurboTurtle:
Since setting sizelimit to be equal to the log_size option unless
explicitly set otherwise, there is no longer a need for plugins to do
something like:

	self.limit = self.get_option('log_size')
	self.add_copy_spec('/foo/bar', sizelimit=self.limit)

As such, update plugins that were just using the log_size setting to no
longer set this sizelimit value unnecessarily.

---

### DIFF
--- a/sos.spec
+++ b/sos.spec
@@ -2,7 +2,7 @@
 
 Summary: A set of tools to gather troubleshooting information from a system
 Name: sos
-Version: 3.5
+Version: 3.6
 Release: 1%{?dist}
 Group: Applications/System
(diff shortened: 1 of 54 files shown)


TurboTurtle:
Also made spot style fix-ups where appropriate (such as single-item lists -> strings) and removed a duplicate `add_copy_spec()` from the OVS plugin.

bmr-cymru:
Merged in commit 967c015518c1aa135aa6007329972f031ffe12fc

Issue #216, opened 2013-11-07
Fix rhel_version and rename to dist_version()


bmr-cymru:
The Red Hat policy class historically provided rhelVersion() to determine the version of the distribution on which a plug-in was running.

This has been carried over to rhel_version() in the revised policy class but no longer functions due to changes in the underlying package manager abstraction.

This is generally useful infrastructure that should be available to plug-ins on any supported platform: fix the method and rename it dist_version().

PR #1361, opened 2018-06-23
[redhat] Fix RHELPolicy.dist_version()

TurboTurtle:
The RHELPolicy dist_version() was always returning False due to an
invalid call to all_pkgs_by_name_regex(). Correct this call and use the
discovered redhat-release package to determine version.

Resolves: #216 

---


bmr-cymru:
Merged in commit 57d515a2fdcc0b5ed9536d4ab622dc291cfab3c7

PR #1362, opened 2018-06-23
[utilities|i18n] Correct locale output

TurboTurtle:
First, allows any command to filter out unwanted environment variables from the environment used by `sos_get_command_output()`.

Using this, removes the `LC_ALL` environment variable from the capture of `locale` which was causing erroneous output due to that variable being set to get consistent locale output from all other commands captured by sos.

---

### DIFF
--- a/sos.spec
+++ b/sos.spec
@@ -2,7 +2,7 @@
 
 Summary: A set of tools to gather troubleshooting information from a system
 Name: sos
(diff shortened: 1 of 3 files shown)


bmr-cymru:
Merged in commit 97454f9a3e3123ee5a1054de2c8032edc2a965fd

PR #1363, opened 2018-06-24
[kernel] handle case when bpftool installed but not implemented

pmoravec:
on older kernels, bpftool can return error

"can't get next program: Function not implemented"

that lacks "id" we search for in the dict.

Resolves: #1363

---

### DIFF
--- a/sos.spec
+++ b/sos.spec
@@ -2,7 +2,7 @@
 
 Summary: A set of tools to gather troubleshooting information from a system
 Name: sos
-Version: 3.5
+Version: 3.6
 Release: 1%{?dist}
 Group: Applications/System
 Source0: http://people.redhat.com/breeves/sos/releases/sos-%{version}.tar.gz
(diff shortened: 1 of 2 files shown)


pmoravec:
This is to prevent backtraces like:

      File "/usr/lib/python2.7/site-packages/sos/sosreport.py", line 1009, in setup
        plug.setup()
      File "/usr/lib/python2.7/site-packages/sos/plugins/kernel.py", line 138, in setup
        for prog_id in self.get_bpftool_prog_ids(prog_file):
      File "/usr/lib/python2.7/site-packages/sos/plugins/kernel.py", line 37, in get_bpftool_prog_ids
        out.append(prog_data[item]["id"])
    KeyError: 'id'

when `bpftool prog -j` returns

    [{"error":"can't get next program: Function not implemented"}]

as running on kernel that does not support eBPF.

PR #1366, opened 2018-06-25
[rhv-log-collector-analyzer] Add new plugin for rhv

dougsland:
collect:

 - Generated HTML file from --live stored in temporary dir.

Example from sosreport -o rhv_log_collector_analyzer:
  sos_commands/rhv_log_collector_analyzer/rhv-log-collector-analyzer_--json
  tmp/rhv-log-collector-analyer-report.html

---

### DIFF
--- a/sos/plugins/rhv_log_collector_analyzer.py
+++ b/sos/plugins/rhv_log_collector_analyzer.py
@@ -1,0 +1,55 @@
+# Copyright (C) 2018 Red Hat, Inc.
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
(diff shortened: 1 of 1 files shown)


sos/plugins/rhv_log_collector_analyzer.py
```diff
@@ -0,0 +1,55 @@
...
+
+    report = "{tempdir}/rhv-log-collector-analyer-report.html".format(
+        tempdir=tempfile.gettempdir())
+
+    def setup(self):
+
+        self.get_command_output(
+            "rhv-log-collector-analyzer"
+            " --live"
+            " --html={report}".format(report=self.report)
```
pmoravec: I recommend using `runat=self.get_cmd_output_path()` to avoid the self-managing of `self.report` file.

PR #1367, opened 2018-06-25
[rhv-log-collector-analyzer] Add new plugin for RHV

dougsland:
collect:

 - Generated HTML file from --live stored in temporary dir.

Example from sosreport -o rhv_log_collector_analyzer:
  rhv_log_collector_analyzer/rhv-log-collector-analyzer_--json
  tmp/rhv-log-collector-analyer-report.html

---

- []x Is the subject and message clear and concise?

### DIFF
--- a/sos/plugins/rhv_log_collector_analyzer.py
+++ b/sos/plugins/rhv_log_collector_analyzer.py
@@ -1,0 +1,55 @@
+# Copyright (C) 2018 Red Hat, Inc.
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+import os
+import tempfile
+
+from sos.plugins import Plugin, RedHatPlugin
+
+
+class RhvLogCollectorAnalyzer(Plugin, RedHatPlugin):
+    """RHV Log Collector Analyzer"""
+
+    packages = (
+        'rhv-log-collector-analyzer',
+    )
+
+    plugin_name = 'rhv_log_collector_analyzer'
+    profiles = ('virt',)
+
+    report = "{tempdir}/rhv-log-collector-analyer-report.html".format(
+        tempdir=tempfile.gettempdir())
+
+    def setup(self):
+
+        self.get_command_output(
+            "rhv-log-collector-analyzer"
+            " --live"
+            " --html={report}".format(report=self.report)
+        )
+
+        if os.path.exists(self.report):
+            self.add_copy_spec("{report}".format(report=self.report))
+
+        self.add_cmd_output(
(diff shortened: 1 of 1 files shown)


sos/plugins/rhv_log_collector_analyzer.py
```diff
@@ -0,0 +1,55 @@
...
+    def setup(self):
+
+        self.get_command_output(
+            "rhv-log-collector-analyzer"
+            " --live"
+            " --html={report}".format(report=self.report)
+        )
+
+        if os.path.exists(self.report):
+            self.add_copy_spec("{report}".format(report=self.report))
```
pmoravec: Copying from previous closed PR: I recommend using `runat=self.get_cmd_output_path()` to avoid the self-managing of `self.report` file.

sos/plugins/rhv_log_collector_analyzer.py
```diff
@@ -0,0 +1,55 @@
...
+
+from sos.plugins import Plugin, RedHatPlugin
+
+
+class RhvLogCollectorAnalyzer(Plugin, RedHatPlugin):
+    """RHV Log Collector Analyzer"""
+
+    packages = (
+        'rhv-log-collector-analyzer',
+    )
```
bmr-cymru: This doesn't need to be split over three lines.

sos/plugins/rhv_log_collector_analyzer.py
```diff
@@ -0,0 +1,55 @@
...
+
+    plugin_name = 'rhv_log_collector_analyzer'
+    profiles = ('virt',)
+
+    report = "{tempdir}/rhv-log-collector-analyer-report.html".format(
+        tempdir=tempfile.gettempdir())
+
```
bmr-cymru: Absolute NACK: plugins must never do their own thing with temporary files and directories. You need to use the facilities that the `Plugin` and `Archive` classes provide you.

sos/plugins/rhv_log_collector_analyzer.py
```diff
@@ -0,0 +1,55 @@
...
+    def setup(self):
+
+        self.get_command_output(
+            "rhv-log-collector-analyzer"
+            " --live"
+            " --html={report}".format(report=self.report)
+        )
```
bmr-cymru: How is this going to work? You are _requesting a command to be run_ by `self.get_command_output()`, but all this does is to add it to a list to be collected during the `collect()` phase of the run. You need to use the `_now()` variant if you want to try to do something with the data in the same method.

sos/plugins/rhv_log_collector_analyzer.py
bmr-cymru: NACK: You need to find a way to get this written directly into the archive directory structure, either by having the command write it to `stdout`, or by using some `--output-file` option.

sos/plugins/rhv_log_collector_analyzer.py
```diff
@@ -0,0 +1,55 @@
...
+            self.add_copy_spec("{report}".format(report=self.report))
+
+        self.add_cmd_output(
+            "rhv-log-collector-analyzer"
+            " --json"
+        )
+
+    def postproc(self):
+        if os.path.exists(self.report):
+            os.unlink("{report}".format(report=self.report))
```
bmr-cymru: The `Plugin.postproc()` hook is for post-processing data (e.g. removing passwords and other secrets). It should not be used for cleaning up temporary files, firstly because it is unreliable: an exception elsewhere may abort cleanup processing, which makes us breach the guarantees made in the `sosreport` preamble text, and also because the `Archive` will allow you to store those files directly into its path structure - so there's no need to create and cleanup files outside of these directories.

dougsland:
Hello, thanks for reviews, going to address all points (less the name with multiple dashes as it's already build in brew)

PR #1368, opened 2018-06-26
[kubernetes] Add support for Ubuntu

slashdd:
There are multiples ways to run a Kubernetes cluster with Ubuntu.
This commit is based on the "Official Ubuntu Guides", The Canonical
Distribution of Kubernetes deployed using "conjure-up".

---

### DIFF
--- a/sos/plugins/kubernetes.py
+++ b/sos/plugins/kubernetes.py
@@ -9,11 +9,11 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.plugins import Plugin, RedHatPlugin
+from sos.plugins import Plugin, RedHatPlugin, UbuntuPlugin
 from os import path
 
 
-class kubernetes(Plugin, RedHatPlugin):
+class kubernetes(Plugin, RedHatPlugin, UbuntuPlugin):
 
     """Kubernetes plugin
     """
@@ -22,7 +22,7 @@ class kubernetes(Plugin, RedHatPlugin):
     # atomic-openshift-master package to provide kubernetes
     packages = ('kubernetes', 'kubernetes-master', 'atomic-openshift-master')
     profiles = ('container',)
-    files = ("/etc/origin/master/master-config.yaml",)
+    files = ("/etc/origin/master/master-config.yaml", "/root/cdk/server.key")
 
     option_list = [
         ("all", "also collect all namespaces output separately",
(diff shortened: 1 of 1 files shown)


slashdd:
Related documents:
https://kubernetes.io/docs/getting-started-guides/ubuntu/#official-ubuntu-guides
https://www.ubuntu.com/kubernetes

sos/plugins/kubernetes.py
```diff
@@ -9,13 +9,29 @@
...
-    """Kubernetes plugin
+    """Ubuntu Kubernetes plugin
+    """
+
+    files = ("/snap/bin/kubectl",)
+
+    def setup(self):
+        self.add_cmd_output([
+            "snap run kubectl get nodes -o wide",
+            "snap run kubectl cluster-info",
```
TurboTurtle: This is really the only new command being captured by `Ubuntukubernetes`, the other bits are all already captured. Is there any reason we couldn't add `cluster-info` to the existing collection and just set `kube_cmd` to `snap run kubectl` for Ubuntu while keeping it `kubectl` elsewhere?

slashdd: Sure, I'll have a look at it and re-submit a PR later.

sos/plugins/kubernetes.py
```diff
@@ -35,7 +35,22 @@ class kubernetes(Plugin, RedHatPlugin):
...
         if any([
             path.exists("/var/run/kubernetes/apiserver.key"),
-            path.exists("/etc/origin/master/master-config.yaml")
+            path.exists("/etc/origin/master/master-config.yaml"),
+            path.exists("/root/cdk/server.key")
+        ]):
+            return True
+        return False
+
+    def check_is_ubuntu(self):
```
TurboTurtle: The Plugin class already has this kind of OS-detection functionality built-in with the tagging mechanic.

This means we can just move `kube_cmd` to a class attr, and then use something like this for Ubuntu:

```python
class UbuntuKubernetes(kubernetes, UbuntuPlugin):

    files = ("/root/cdk/server.key", )

    def setup():
        self.kube_cmd = "snap run kubectl "
        super(UbuntuKubernetes, self).setup()
```

After that we'd just need to either transform `self.kube_cmd` back into `kube_cmd` in `Kubernetes.setup()` or modify all references of kube_cmd to be `self.kube_cmd` (probably best).

bmr-cymru:
No response to review comments since last July: if this functionality is still required, please re-open this pull request.

PR #1372, opened 2018-07-02
[pacemaker] Collect stonith and quorum status

TurboTurtle:
Adds collection of stonith and quorum states within the cluster to the
pacemaker plugin. If either of these are not used in a given cluster,
these commands simply report that the resource is not in use so this is
safe to run on all clusters.

---


TurboTurtle:
Merged in 1385628

Closing.

PR #1379, opened 2018-07-08
jtpittman195 multipath commands update

jtpittman195:
Request merge:
[multipath] use -ll for path checker and prio inclusion
[multipath] add multipath -t output

### DIFF
--- a/sos/plugins/multipath.py
+++ b/sos/plugins/multipath.py
@@ -24,6 +24,7 @@ class Multipath(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
         self.add_cmd_output([
             "multipath -ll",
             "multipath -v4 -ll",
+            "multipath -t",
             "multipathd show config"
         ])
 
--- a/sos/plugins/multipath.py
+++ b/sos/plugins/multipath.py
@@ -22,7 +22,7 @@ class Multipath(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
             "/etc/multipath.conf"
         ])
         self.add_cmd_output([
-            "multipath -l",
+            "multipath -ll",
             "multipath -v4 -ll",
             "multipathd show config"
         ])


sos/plugins/multipath.py
```diff
@@ -22,8 +22,9 @@ def setup(self):
             "/etc/multipath.conf"
         ])
         self.add_cmd_output([
-            "multipath -l",
+            "multipath -ll",
```
TurboTurtle: We already collect this (`-ll`) on the very next line, with increased verbosity. 

Also, why remove `-l` here? Isn't there value in capturing both of these?

jtpittman195: Right now, only the -l is gathered, I added the -ll.  There's no value that I'm aware of in leaving -l; everything that is present in -l will be in -ll, and the verbosity does not increase noticeably.

TurboTurtle: [That is what we already do](https://github.com/sosreport/sos/blob/master/sos/plugins/multipath.py#L25-L26) today. `-l`, and `-v4 -ll`.

Your patch would replace that behavior with `-ll` and `-v4 -ll`. Meaning the same exact command, but again with much less verbosity.

We already collect `-ll` - why do we need to collect it again at the lower verbosity setting?

jtpittman195: Yep, that's right; that's exactly what we need, the -ll and then the -v4 -ll.  I edited my last note when I realized what you were asking.  Did not see you had already commented.  Will paste here:
///paste
Right now, only 'multipath -l' and ' multipath -v4 -ll' is gathered. The 'multipath -ll' does what the 'multipath -l' is supposed to... give a standard, easily readable, output that is not so verbose that you have to grep through to get anything done. The 'multipath -l' has a problem in that it leaves out important info for troubleshooting that the 'multipath -ll' does not. There's no value that I'm aware of in leaving -l if we can switch to -ll; everything that is present in -l will be in -ll, along w/ the important prio stuff, and the verbosity does not increase noticeably.

TurboTurtle:
After an IRC discussion, the move is to include a bare `-ll` since the `-v4` variant can be troublesome to work through. Since we aren't losing any information here, and the fact that we do this in a few other plugins already, I think this is an okay change to make.

PR #1380, opened 2018-07-10
[rhv-log-collector-analyzer] Add new plugin for RHV

dougsland:
collect:

- Generated HTML file from --live

---

### DIFF
--- a/sos/plugins/rhv_analyzer.py
+++ b/sos/plugins/rhv_analyzer.py
@@ -1,0 +1,40 @@
+# Copyright (C) 2018 Red Hat, Inc.
+#
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.plugins import Plugin, RedHatPlugin
+
+
+class RhvLogCollectorAnalyzer(Plugin, RedHatPlugin):
+    """RHV Log Collector Analyzer"""
+
+    packages = ('rhv-log-collector-analyzer',)
+
+    plugin_name = 'RhvLogCollectorAnalyzer'
+    profiles = ('virt',)
+
+    def setup(self):
+        tool_name = 'rhv-log-collector-analyzer'
+        report = "{dircmd}/analyzer-report.html".format(
+            dircmd=self.get_cmd_output_path()
+        )
+
+        self.add_cmd_output(
+            "{tool_name}"
+            " --live"
+            " --html={report}".format(
+                report=report, tool_name=tool_name)
+        )
+
+        self.add_cmd_output(
+            "{tool_name}"
+            " --json".format(tool_name=tool_name)
+        )
+
+# vim: expandtab tabstop=4 shiftwidth=4


sos/plugins/rhv_log_collector_analyzer.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def setup(self):
+        report = "{dircmd}/rhv-log-collector-analyer-report.html".format(
+            dircmd=self.get_cmd_output_path()
+        )
+
+        self.get_cmd_output_now(
+            "rhv-log-collector-analyzer"
+            " --live"
+            " --html={report}".format(report=report)
```
TurboTurtle: This generates two files. 
One, the HTML report we want, and a second that is the command name - e.g. `rhv-log-collector-analyzer_--live_--html_.var.tmp.sos.crx5Wy.sosreport-localhost.localdomain.localdomain-20180710134357.sos_commands.rhv_log_collector_analyzer.rhv-log-collector-analyer-report.html` which contains the stdout of this command. Is this output ever useful on its own?

sos/plugins/rhv_log_collector_analyzer.py
```diff
@@ -0,0 +1,42 @@
...
+    profiles = ('virt',)
+
+    def setup(self):
+        report = "{dircmd}/rhv-log-collector-analyer-report.html".format(
+            dircmd=self.get_cmd_output_path()
+        )
+
+        self.get_cmd_output_now(
+            "rhv-log-collector-analyzer"
+            " --live"
```
TurboTurtle: Should we gate this on the presence of the ovirt-engine package as well? Are there any situations where `rhv-log-collector-analyzer` might be run on a non-ovirt-manager installation?

dougsland: When using --live requires Engine but it is possible to run with remote Engine.

sos/plugins/rhv_log_collector_analyzer.py
dougsland: Yes, better keep, if for some reason an error occur we could get the issue.

sos/plugins/rhv_log_collector_analyzer.py
```diff
@@ -0,0 +1,42 @@
...
+            " --html={report}".format(report=report)
+        )
+
+        self.get_cmd_output_now(
```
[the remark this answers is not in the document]
dougsland: okay.

sos/plugins/rhv_log_collector_analyzer.py
```diff
@@ -0,0 +1,42 @@
+# Copyright (C) 2018 Red Hat, Inc.
+
+# This program is free software; you can redistribute it and/or modify
```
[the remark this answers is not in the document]
dougsland: okay

sos/plugins/rhv_log_collector_analyzer.py
```diff
@@ -0,0 +1,42 @@
...
+class RhvLogCollectorAnalyzer(Plugin, RedHatPlugin):
+    """RHV Log Collector Analyzer"""
+
+    packages = ('rhv-log-collector-analyzer',)
+
+    plugin_name = 'rhv_log_collector_analyzer'
+    profiles = ('virt',)
+
+    def setup(self):
+        report = "{dircmd}/rhv-log-collector-analyer-report.html".format(
```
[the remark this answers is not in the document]
dougsland: okay

sos/plugins/rhv_log_collector_analyzer.py
dougsland: @dankenigsberg yes, I think for now we should add the requirement for now.

dougsland: Thinking twice here, I wouldn't add such requirement static in the plugin because:
- we have such verification in our spec [1]
- rhv-log-collector-analyzer --live soon or later will analyze hosts as well

[1]
'''
%if 0%{?rhel} >= 7 || 0%{?fedora}
Requires:       ovirt-engine
%else
Requires:       rhevm-setup
%endif
''''

bmr-cymru: If it just comes down to using or not using `--live` then just check for the package presence (or some other test) and add or don't add it as appropriate.

bmr-cymru:
Please squash the fixup commit into the main one and update with `git push --force` - we don't allow fixups in master, and it makes it easier to review.

sos/plugins/rhv_log_collector_analyzer.py
```diff
@@ -35,7 +29,7 @@ def setup(self):
             " --html={report}".format(report=report)
         )
 
-        self.get_cmd_output_now(
+        self.add_cmd_output(
```
bmr-cymru: This still needs to be done for the `rhv-log-collector-analyzer --live --html...` version. You're not even keeping the return value so it's entirely useless to use `get_cmd_output_now()` here; the directory returned by `get_cmd_output_path()` will still be valid for async collection.

sos/plugins/rhv_log_collector_analyzer.py
```diff
@@ -0,0 +1,42 @@
...
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+from sos.plugins import Plugin, RedHatPlugin
+
+
+class RhvLogCollectorAnalyzer(Plugin, RedHatPlugin):
+    """RHV Log Collector Analyzer"""
+
+    packages = ('rhv-log-collector-analyzer',)
+
+    plugin_name = 'rhv_log_collector_analyzer'
```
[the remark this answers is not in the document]
TurboTurtle: +1 for `rhv_analyzer`, it's short and easily correlated.

TurboTurtle:
This revision looks good to me now.

sos/plugins/rhv_analyzer.py
```diff
@@ -0,0 +1,40 @@
...
+
+from sos.plugins import Plugin, RedHatPlugin
+
+
+class RhvLogCollectorAnalyzer(Plugin, RedHatPlugin):
+    """RHV Log Collector Analyzer"""
+
+    packages = ('rhv-log-collector-analyzer',)
+
+    plugin_name = 'rhv_analyzer'
```
bmr-cymru: Your plugin name needs to match the lower-case class name: either "RhvLogCollectorAnalyzer" and "rhvlogcollectoranalyzer", or "Rhv_Analyzer" and "rhv_analyzer".

bmr-cymru:
Are you sure that's the name you want to go with? I thought we discussed this earlier, and decided that "rhv_analyzer" was a better idea for users who may need to enter it on the command line?

Ultimately. it's your plugin and your choice, but that is not a very friendly piece of text to type...

bmr-cymru:
The value of `plugin_name` also needs to be in lower case, like I said yesteday...

bmr-cymru:
OK, I am just going to change everything to match the current file name (`rhv_analyzer`) and merge it as that.

dankenigsberg:
@bmr-cymru thanks.
Would you now agree to take it downstream, to el7.6?

bmr-cymru:
@dankenigsberg that's not my call at all - I only maintain upstream. I wouldn't have any objection to it, but it's really down to the @pmoravec, the RHEL maintainer and the Red Hat processes.

A bug has been files (bz#1600158 - you seem to be on the CC list), but it has no acks and was filed _way_ too late to make the original development cutoff (11th July!). It is still possible for it to get into the release but someone needs to go about getting the requisite acks and exception status for it if that is going to happen.

PR #1381, opened 2018-07-11
[ovirt_node] New plugin for oVirt Node

nijinashok:
oVirt Node is a small scaled down version used for hosting virtual
machines. The plugin collects node specific information like
upgrade log, the layer structure etc.

Signed-off-by: Nijin Ashok <nashok@redhat.com>

---


TurboTurtle:
LGTM, my only suggestion is to add a Copyright line above the license header.

nijinashok:
Thanks Jake. I have added "Copyright" info.

PR #1383, opened 2018-07-12
[kube|etcd|origin] Update plugins for static pod deployments

### LINKED ISSUE #1165: [RFE] origin (OpenShift) plugin to collect statistics information
https://docs.openshift.com/container-platform/latest/admin_guide/monitoring_images.html#monitoring-images-view-images

TurboTurtle:
Per [Origin 3.10 release notes](https://github.com/openshift/openshift-docs/issues/8651#issuecomment-401484357) the static pod configuration for Origin/OCP changes the way we need to collect etcd information and potentially how the kubernetes plugin should be enabled.

Also, correct the kubernetes syntax for specifying a config file override.

Finally, resolve #1165 by including `oc adm top` output.

---

### DIFF
--- a/sos/plugins/etcd.py
+++ b/sos/plugins/etcd.py
@@ -10,6 +10,7 @@
 # See the LICENSE file in the source distribution for further information.
 
 from sos.plugins import Plugin, RedHatPlugin
+from os import path
 
 
 class etcd(Plugin, RedHatPlugin):
@@ -19,10 +20,14 @@ class etcd(Plugin, RedHatPlugin):
     plugin_name = 'etcd'
     packages = ('etcd',)
     profiles = ('container', 'system', 'services', 'cluster')
-
-    cmd = 'etcdctl'
+    files = ('/etc/origin/node/pods/etcd.yaml',)
 
     def setup(self):
+        if path.exists('/etc/origin/node/pods/etcd.yaml'):
+            etcd_cmd = 'master-exec etcd etcd etcdctl'
+        else:
+            etcd_cmd = 'etcdctl'
+
         etcd_url = self.get_etcd_url()
 
         self.add_forbidden_path('/etc/etcd/ca')
@@ -35,7 +40,7 @@ class etcd(Plugin, RedHatPlugin):
            'ls --recursive',
         ]
 
-        self.add_cmd_output(['%s %s' % (self.cmd, sub) for sub in subcmds])
+        self.add_cmd_output(['%s %s' % (etcd_cmd, sub) for sub in subcmds])
 
         urls = [
             '/v2/stats/leader',
--- a/sos/plugins/kubernetes.py
+++ b/sos/plugins/kubernetes.py
@@ -18,11 +18,16 @@ class kubernetes(Plugin, RedHatPlugin):
     """Kubernetes plugin
     """
 
-    # Red Hat Atomic Platform and OpenShift Enterprise use the
-    # atomic-openshift-master package to provide kubernetes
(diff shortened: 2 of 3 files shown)


sos/plugins/etcd.py
```diff
@@ -19,10 +20,14 @@ class etcd(Plugin, RedHatPlugin):
...
     profiles = ('container', 'system', 'services', 'cluster')
-
-    cmd = 'etcdctl'
+    files = ('/etc/origin/node/pods/etcd.yaml',)
 
     def setup(self):
+        if path.exists('/etc/origin/node/pods/etcd.yaml'):
+            self.cmd = 'master-exec etcd etcd etcdctl'
+        else:
+            self.cmd = 'etcdctl'
```
sferich888: In theory this should just work; however, some RHEL versions use etcdv2 and others use etcdv3, and this command works using an env variable to determine which subcommand (etcdctl2 or etcdctl3) you want to use. 

I wonder if/how we could better determine what db schema etcd was using (outside of these commands) so we know what command to properly call? Without this, we might need to try calling both to see which one works?

TurboTurtle: I don't think we need to version-control the command we use since the `etcd` package drops `/usr/bin/etcdctl` regardless of which version that's installed. The package is responsible for this, so sos shouldn't attempt to circumvent it even if it is in an attempt to ensure versioning.

sos/plugins/kubernetes.py
```diff
@@ -33,12 +37,7 @@ class kubernetes(Plugin, RedHatPlugin):
     ]
 
     def check_is_master(self):
-        if any([
-            path.exists("/var/run/kubernetes/apiserver.key"),
-            path.exists("/etc/origin/master/master-config.yaml")
-        ]):
-            return True
-        return False
+        return any([path.exists(f) for f in self.files])
```
bmr-cymru: The logic here looks a bit weird to me: we have a plugin files list, which, if any of the files exist causes the plugin to run (the plugin will also run if those files do not exist, but one or more package matches are present).

Then we have a separate `check_is_master()` method, that *used* to check two specific "master" paths (the API server key, and the master-config YML file).

Now it just does the same as the default `check_enabled()` does for the `files` list: it returns `True` if any of the `Kubernetes.files` paths exist - so if the plugin is running (at least from a file trigger), then `check_is_master()` is always `True`.

Is there something subtle I've missed here? If the logic is correct then we should be able to dispense with all the call to `check_is_master()`; if it's not then it seems like there's some false sharing between the plugin `files` list and the list of files that `check_is_master()` should be operating on.

sos/plugins/etcd.py
bmr-cymru: The `self.cmd` member seems to be used in just one place in exactly one method (`setup()`): let's ditch it and just use a local variable instead? (`etcd_cmd` would be a bit more readable and no great burden since we lose the `self.` prefix for e.g.).

sos/plugins/kubernetes.py
TurboTurtle: Right, the `files` list is strictly locations that a kubernetes master would have that a worker node would not have - there are no files (at least that I'm aware of) that a node would have that a master wouldn't have. For worker node collection we're relying on just the package check.

The reason for this is that we still want to collect the `kube*` journals but literally everything else in the plugin would not be available on the worker nodes.

sos/plugins/etcd.py
TurboTurtle: ack.

PR #1385, opened 2018-07-23
[krb5] Add more files to krb5 plugin

frozencemetery:
Add files for KDC configuration and logging, including kadmin files
and a klist of the K/M stash.

Gather any additional configuration snippet files in
/etc/krb5.conf.d (which is configured by default on Fedora- and
RHEL-like systems, and hopefully on Debian systems in the future).
The sssd plugin already takes care of
/var/lib/sss/pubconf/krb5.include.d/, so don't include that.

Signed-off-by: Robbie Harwood <rharwood@redhat.com>

---

### DIFF
--- a/sos/plugins/krb5.py
+++ b/sos/plugins/krb5.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2013 Red Hat, Inc., Bryn M. Reeves <bmr@redhat.com>
+# Copyright (C) 2013,2018 Red Hat, Inc., Bryn M. Reeves <bmr@redhat.com>
 
 # This file is part of the sos project: https://github.com/sosreport/sos
 #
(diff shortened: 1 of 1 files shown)


sos/plugins/krb5.py
```diff
@@ -19,8 +19,34 @@ class Krb5(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
     packages = ('krb5-libs', 'krb5-user')
 
     def setup(self):
-        self.add_copy_spec("/etc/krb5.conf")
+        self.add_copy_spec([
+            "/etc/krb5.conf",
+            "/etc/krb5.conf.d/*",
+            "/var/log/krb5kdc.log",
+            "/var/log/kadmind.log",
```
TurboTurtle: So, syntactically the trailing comma on multiline lists is okay, but stylistically we don't do this anywhere else in sos (at least not that I'm aware of).

I'm inclined to say for consistency sake, let's remove the trailing commas. @bmr-cymru thoughts?

sos/plugins/krb5.py
```diff
@@ -19,8 +19,34 @@ class Krb5(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
...
 
+class RedHatKrb5(Krb5, RedHatPlugin):
+
+    def setup(self):
+        super(RedHatKrb5, self).setup()
+        self.add_copy_spec([
+            "/var/kerberos/krb5kdc/kadm5.acl",
+            "/var/kerberos/krb5kdc/kdc.conf",
+        ])
+        self.add_cmd_output("klist -ekt /var/kerberos/krb5kdc/.k5*")
```
TurboTurtle: Since the only difference between `RedHatKrb5` and `DebianKrb5` is the directory in which these files are located, I think it would be better to move these collections to the base `Krb5.setup()`, and have the tagging classes set the directory instead of specifying them individually for each distro.  E.G. something like:

```python 
class Krb5(Plugin):

    kdir = "/whatever/upstream/default/is"

    def setup(self):
    [...]
    kfiles = ["kadm5.acl", "kdc.conf"]
    self.add_copy_spec([
        "%s/%s" % (self.kdir, kfile) for kfile in kfiles
    ])

class RedHatKrb5(Krb5, RedHatPlugin):

    def setup(self):
        self.kdir = "/var/kerberos/krb5kdc"
        super(RedHatKrb5, self).setup()
```

This would make adding both more distro support and more files to collect easier down the road.

sos/plugins/krb5.py
bmr-cymru: There are odd examples of it. I don't have a huge problem either way (to be honest, I agree with the argument that it reduces the chance of continuation/concatenation bugs, but it just looks so ... ugly :-/ ).

bmr-cymru: I'd love something that could semantically validate lists and warn on missing commas - we have some cases of genuine concatenation but they are rare and easy to filter.

sos/plugins/krb5.py
bmr-cymru: Ack

bmr-cymru: (ditto for the `klist`s)

frozencemetery: This makes sense.  It's a bit more complicated than that because neither actually use the upstream default (/var/krb5kdc), but I'll see what I can do.

bmr-cymru: > This makes sense. It's a bit more complicated than that because neither actually use the upstream default (/var/krb5kdc), but I'll see what I can do.

Yeah: that's a bit of a weird case. One way to do it that reflects reality at least is to make the upstream default the value in the base class (which is effectively an abstract `Plugin`), and then override it in both children... so the upstream one is never actually really used... It's more documentation than anything at that point, but it's one option.

bmr-cymru: (the alternative is to pick one of the real ones for the base class and just add a comment explaining it).

frozencemetery:
Thanks for suggestions; updated!

PR #1386, opened 2018-07-23
[plugins] Add plugin for gssproxy

frozencemetery:
gssproxy stores its configuration in /etc/gssproxy.  Also capture the
mech configuration so that we can tell if gssproxy is enabled and any
other GSS mechs in use.

Signed-off-by: Robbie Harwood <rharwood@redhat.com>

---

### DIFF
--- a/sos/plugins/gssproxy.py
+++ b/sos/plugins/gssproxy.py
@@ -1,0 +1,26 @@
+# Copyright (C) 2018 Red Hat, Inc., Robbie Harwood <rharwood@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.plugins import Plugin
+
+
+class GSSProxy(Plugin):
+    """GSSAPI Proxy
+    """
+
+    plugin_name = "gssproxy"
+    profiles = ('services', 'security', 'identity')
+    packages = ('gssproxy',)
+
+    def setup(self):
+        self.add_copy_spec([
+            "/etc/gssproxy/*.conf",
+            "/etc/gss/mech.d/*"
+        ])


TurboTurtle:
LGTM, my only question would be if there is any sensitive data in the `*.conf` files that we might need to obfuscate?

sos/plugins/gssproxy.py
```diff
@@ -0,0 +1,24 @@
...
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+
```
bmr-cymru: Looks like you're only using `Plugin`?

sos/plugins/gssproxy.py
```diff
@@ -0,0 +1,24 @@
...
+class GSSProxy(Plugin):
+    """GSSAPI Proxy
+    """
+
+    plugin_name = "gssproxy"
+    profiles = ('services', 'security', 'identity')
+    packages = ('gssproxy',)
+
+    def setup(self):
+        self.add_copy_spec("/etc/gssproxy/*.conf")
```
bmr-cymru: Let's use a list here:

```python
        self.add_copy_spec([
            "/etc/gssproxy/*.conf",
            "/etc/gss/mech.d/*"
        ])
```

frozencemetery:
Thanks, updated!

Good question about secrets.  gssproxy doesn't have any secrets in configuration files.  The most sensitive things that could be exposed are usernames.  Full configuration is in [gssproxy.conf(5)](https://manpages.debian.org/stretch/gssproxy/gssproxy.conf.5.en.html).  (gssproxy uses system keytabs for secret storage, so its secrets are in /etc/krb5.keytab, /var/lib/gssproxy/clients, and such, and presumably wouldn't end in .conf.)

bmr-cymru:
Your plugin should include tagging classes for the distributions it's supposed to run on: normally you only have a "naked" Plugin class (`class SomePlug(Plugin):`) if it also has one or more derived classes in the same file that have different sets of tagging classes present.

We've also been requiring VIM modelines in all source files for a while now.

I'll add these as I merge: no need to re-push the branch.

Issue #83, opened 2012-12-12
Add a 'services' member to Plugin?


bmr-cymru:
We already have "files" and "packages" but some modules implement runlevel checks to gather service information, e.g. sunrpc:

```
    def checkenabled(self):
       if self.policy().runlevelDefault() in self.policy().runlevelByService("rpcbind"):
          return True
       return False
```

This could be made more generic and allow easier maintenance by adding a "services" list and checking it in the default checkenabled() implementation.

That way we could avoid duplicate code between Red Hat and Debian which use different names for this service.

PR #1387, opened 2018-07-26
[Policies|Plugins] Add services member

TurboTurtle:
Adds a services member to facilitate plugin enablement. This is tied to
a new InitSystem class that gets attached to policies. The InitSystem
class is used to determine services that are present on the system and
what those service statuses currently are (e.g. enabled/disable).

Plugins can now specify a set of services to enable the plugin on if
that service exists on the system, similar to the file, command, and
package checks.

Additionally, the Plugin class now has methods to check on service
states, and make decisions based off of. For example:

	def setup(self):
	    if self.is_service('foobar'):
	        self.add_cmd_output('barfoo')

Currently, only systemd has actual functionality for this. The base
InitSystem inherited by policies by default will always return False for
service checks, thus resulting in the same behavior as before this
change.

The Red Hat family of distributions has been set to systemd, as all
current versions of those distributions use systemd.

Resolves: #83

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -123,6 +123,7 @@ class Plugin(object):
     files = ()
     commands = ()
     kernel_mods = ()
+    services = ()
     archive = None
     profiles = ()
     sysroot = '/'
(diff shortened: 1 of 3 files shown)


TurboTurtle:
I do have one concern - the default `return False` behavior for the base `InitSystem`. 

This means nothing changes for plugins that use the new services enablement check on distributions that don't have an InitSystem set in their policy - however, it also prevents us from doing blanket stuff like this in the docker plugin:

```python
        if self.get_service_status('docker')['status'] == 'active':
            for subcmd in subcmds:
                self.add_cmd_output("docker %s" % subcmd)
```

bmr-cymru:
Can you reasonably have them use a mock `InitSystem` that always returns `True` here?

Longer term, all of these types of boolean tests should be subsumed into predicates anyway; then you can have rules like "if policy lacks proper InitSystem all service predicates evaluate True".

PR #1390, opened 2018-07-28
[block] collect luksDump for all encrypted devices

pmoravec:
Call "cryptsetup luksDump /dev/sd*" for all encrypted devices

Resolves: #1390

---

### DIFF
--- a/sos/plugins/block.py
+++ b/sos/plugins/block.py
@@ -19,6 +19,22 @@ class Block(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
     verify_packages = ('util-linux',)
     files = ('/sys/block',)
 
+    def get_luks_devices(self, lsblk_file):
+        out = []
+        try:
+            lsblk_out = open(lsblk_file).read()
+        except IOError:
+            return out
+        for line in lsblk_out.splitlines():
+            # find in output lines like
+            # |-sda2    crypto_LUKS    <uuid>
+            # and separate device name - it will be 1st string on the line
+            # after first '-'
+            if 'crypto_LUKS' in line:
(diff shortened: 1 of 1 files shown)


pmoravec:
I am not sure if `block` plugin is the best place to put the code into, but havent found a better one and creating an extra plugin just for this small enhancement seems a bit overkill for me.

sos/plugins/block.py
```diff
@@ -51,4 +51,15 @@ def setup(self):
                     "fdisk -l %s" % disk_path
                 ])
 
+        lsblk_result = self.call_ext_prog("lsblk -P -o +FSTYPE -p")
```
bmr-cymru: Perhaps we should also add `lsblk -f``... I don't think we're capturing FSTYPE anywhere.

bmr-cymru: (we could also do it here - the code isn't really using `-P` name=value parsing at all, so a simple `in` test of the line would work fine. Then we can store the result in the report for analysis use as well).

pmoravec:
Rebased with:

- calling `lsblk -f -a`
- parsing it output and calling `cryptsetup luksDump` for encrypted devices

Since `lsblk` is newly collected while also utilized for generating list of encrypted devices, I suggest to extend `block` plugin.

pmoravec:
`dmsetup` does not collect inactive devices :( So currently it seems to me that `lsblk` is the best option to detect all LUKS devices, and the code should be added either to:

- `block` plugin due to `lsblk` - but the encrypted devices dont have too much common with block devices (this is the state of the PR now)
- to a separate plugin that will collect (or just call?) a command that should belong to `block` plugin

No strong opinion from me here.

Issue #403, opened 2014-09-18
[megacli] command invocation is broken


bmr-cymru:
Need to use the full path (or temporarily modify our PATH) to invoke MegaCLI:

```
>             "/opt/MegaRAID/MegaCli/MegaCli64 LDPDInfo -aALL",
>             "/opt/MegaRAID/MegaCli/MegaCli64 -AdpAllInfo -aALL",
>             "/opt/MegaRAID/MegaCli/MegaCli64 -AdpBbuCmd -GetBbuStatus -aALL",
>             "/opt/MegaRAID/MegaCli/MegaCli64 -ShowSummary -aALL"
```

Also the `-ShowSummary` argument is `-aALL` not `-a0` as currently implemented.

bmr-cymru:
You may want to check a few things for this - looking at the latest downloads on the Broadcom site (they acquired LSI a while back), I don't see any downloads for MegaCLI any more - only `storcli` and `MegaRAID Storage Manager`.

One problem we had with this was that the packages installed to different locations (and there was also a tarball option which you could unpack wherever you liked).

If MegaCLI is still relevant then it might mean we need to add a way to search for things in the enviornment-defined `PATH` if they don't exist in our fixed `PATH`.

PR #1392, opened 2018-08-01
[megacli] Correct MegaCli commands and enablement

TurboTurtle:
The MegaCli64 command needs to be called using the full path to the
binary. Also, correct the -ShowSummary adapter argument.

Additionally, use the builtin 'files' check to enable the plugin only if
the binary exists.

Resolves: #403

---

### DIFF
--- a/sos/plugins/megacli.py
+++ b/sos/plugins/megacli.py
@@ -9,8 +9,6 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-import os
-import os.path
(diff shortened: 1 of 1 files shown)


TurboTurtle:
In 403 it was mentioned we either need the full path or change our path temporarily. I prefer using the full path as it is more explicit, especially since this is a vendor-supplied binary (I'm not aware of any distros that ship it themselves) that defaults to installing itself in /opt.

PR #1393, opened 2018-08-06
[plugin_hardware] Add persistent memory information

mizumm:
Add the result of ndctl command to get persistent memory information.

Signed-off-by: Masayoshi Mizuma msys.mizuma@gmail.com
---

### DIFF
--- a/sos/plugins/hardware.py
+++ b/sos/plugins/hardware.py
@@ -28,6 +28,7 @@ class Hardware(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
         ])
 
         self.add_cmd_output("dmidecode", root_symlink="dmidecode")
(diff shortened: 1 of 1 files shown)


sscargal:
I believe Persistent Memory information is best collected using a new 'pmem' plugin as there's more data that would be useful to obtain than just the ndctl command proposed here.  I'll work on the proposed plugin.

sscargal:
I've created a new 'pmem' plugin to collect data from ndctl, ixpdimm-cli, and ipmctl.  See pull request #1398

bmr-cymru:
There is no need to burn through PR numbers because a pull has changed direction: just update/amend the branch, `git push --force`, and update the PR fields as appropriate. This keeps all the conversation in one place, and ensures that things do not get lost.

PR #1394, opened 2018-08-09
[process] make lsof execution optional

pmoravec:
Make calling of lsof command optional (but enabled by default).

There are systems with millions of open files where lsof takes many minutes to execute while its output is usually not important. The option allows skipping such call while default behaviour remains unchanged.

Resolves: #1394

---


pmoravec:
PR updated, also with removal of `or self.get_option("all_logs")` from `lsof-threads` option per https://github.com/sosreport/sos/issues/1395#issuecomment-411736167 .

pmoravec:
Branch updated to include also (small) extension from #1395 that directly follows upon this change.

PR #1406, opened 2018-08-29
[cri-o] Add support for gathering information on cri-o containers

rhatdan:
Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>

---

### DIFF
--- a/sos/plugins/crio.py
+++ b/sos/plugins/crio.py
@@ -1,0 +1,74 @@
+# Copyright (C) 2018 Red Hat, Inc. Daniel Walsh <dwalsh@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.plugins import Plugin, RedHatPlugin, UbuntuPlugin
+
+
+class CRIO(Plugin):
+
+    """CRI-O containers
+    """
+
+    plugin_name = 'crio'
+    profiles = ('container',)
+    packages = ('cri-o', "cri-tools")
+
+    option_list = [
+        ("all", "enable capture for all containers, even containers "
+            "that have terminated", 'fast', False),
+        ("logs", "capture logs for running containers",
+            'fast', False),
+    ]
+
+    def setup(self):
+        self.add_copy_spec([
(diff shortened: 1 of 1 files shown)


rhatdan:
@mrunalp @runcom PTAL

runcom:
:+1: awesome

sos/plugins/cri-o.py
```diff
@@ -0,0 +1,99 @@
+# Copyright (C) 2018 Red Hat, Inc. Daniel Walsh <dwalsh@redhat.com>
```
bmr-cymru: We'll need a different filename - `cri_o` most likely - Python modules are not allowed hyphens in the name.

bmr-cymru: Or `crio`?

sos/plugins/cri-o.py
```diff
@@ -0,0 +1,99 @@
...
+                for container in insp:
+                    self.add_cmd_output("crictl logs -t %s" % container)
+
+
+class RedHatCri-O(Cri-O, RedHatPlugin):
+
+    packages = ('cri-o', "cri-tools")
+
+    def setup(self):
+        super(RedHatCri-O, self).setup()
```
bmr-cymru: We generally strip non-alpha characters when forming the class name, so `cri_o` -> `CriO`, `kata_containers` -> `KataContainers`.

sos/plugins/cri-o.py
```diff
@@ -0,0 +1,99 @@
...
+
+        self.add_copy_spec([
+            "/etc/containers/"
+        ])
+
+
+class UbuntuCri-O(Cri-O, UbuntuPlugin):
+
+    packages = ('cri-o', "cri-tools")
```
bmr-cymru: Since the `packages` lists are identical for the Red Hat and Ubuntu plugins there's no need to use separate subclasses here - we can just set those values in the main `CriO` class, and have that class inherit from `Plugin, RedHatPlugin, UbuntuPlugin`.

sos/plugins/cri-o.py
rhatdan: Changed to crio.py

sos/plugins/cri-o.py
rhatdan: Ok Changed to CRIO

sos/plugins/cri-o.py
rhatdan: Fixed.

bmr-cymru:
Merged in commit 64d7cccf48aee9a07e2e4c5237034638cae30391

PR #1407, opened 2018-08-29
[plugins] Add support for gathering information on podman containers

rhatdan:
Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>

---


rhatdan:
@mheon @baude PTAL

rhatdan:
@EmilienM @vrothberg PTAL

vrothberg:
We have something similar at SUSE, but we're also inspecting images which came in handy quite a few times.

LGTM (with limited knowledge about the sos tooling)

sos/plugins/podman.py
```diff
@@ -0,0 +1,101 @@
...
+            'pod ps',
+            'pod ps -a',
+            'ps',
+            'ps -a',
+            'stats --no-stream',
+            'version',
+        ]
+
+        for subcmd in subcmds:
+            self.add_cmd_output("podman %s" % subcmd)
```
bmr-cymru: We can write that a little more compactly as:

```python
    self.add_cmd_output(["podman %s" % s for s in subcmds])
```

That said, this pattern is appearing in four plugins now (plus this PR). I'm inclined to think we should add a `add_cmd_with_subcmds()` method (or a less horrible name):

```python
    subcmds = [...]
    self.add_cmd_with_subcmds("podman %s", subcmds)
```

We're discussing a change at the moment to eliminate more explicit iteration and conditions in plugins (the eventual aim is that most will be purely declarative - just member variables and little or no executable code), and this kind of interface would fit into that.

sos/plugins/podman.py
```diff
@@ -0,0 +1,101 @@
...
+        img_cmd = 'podman images -q'
+        insp = set()
+
+        for icmd in [ps_cmd, img_cmd]:
+            result = self.get_command_output(icmd)
+            if result['status'] == 0:
+                for con in result['output'].splitlines():
+                    insp.add(con)
+
+        insp = list(insp)
```
bmr-cymru: Shouldn't be needed if we're just going to iterate over the set's members (we should also fix the one in `docker`).

sos/plugins/podman.py
```diff
@@ -0,0 +1,101 @@
...
+        if insp:
+            for container in insp:
+                self.add_cmd_output("podman inspect %s" % container)
+
+            if self.get_option('logs'):
```
bmr-cymru: Why not put this test under the first loop, since we're already iterating over all of `insp`? (I see this also came from`docker`...)

sos/plugins/podman.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+class RedHatPodman(Podman, RedHatPlugin):
+
+    packages = ('podman')
+
+    def setup(self):
+        super(RedHatPodman, self).setup()
+
+        self.add_copy_spec([
```
bmr-cymru: No need to wrap one path in a list - the method also accepts a single string (also from `docker`).

sos/plugins/podman.py
bmr-cymru: Thanks - I'll think about the subcommand interface some more & talk to the rest of the folks about it: we'll update the plugin to use it once it lands.

bmr-cymru:
Thanks - looks good now. Travis is flagging one whitespace error, but I'll fix that up as I merge.

I'll try to get the changes for Docker done this afternoon, so we have everything in-sync; there's a QE failure that I'm looking into right now that's a bit hot, but hopefully I'll get to this later.

PR #1411, opened 2018-08-31
[openstack_nova] remove too restrictive check_enabled

pmoravec:
Enable the plugin just based on package presence.

Resolves: #1411

---

### DIFF
--- a/sos/plugins/openstack_aodh.py
+++ b/sos/plugins/openstack_aodh.py
@@ -18,13 +18,7 @@ class OpenStackAodh(Plugin, RedHatPlugin):
     plugin_name = "openstack_aodh"
     profiles = ('openstack', 'openstack_controller')
 
-    packages = (
-        'openstack-aodh-api',
-        'openstack-aodh-listener',
-        'openstack-aodh-notifier',
-        'openstack-aodh-evaluator,'
-        'openstack-aodh-common'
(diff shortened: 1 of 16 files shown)


bmr-cymru:
It's a bit more than this unfortunately - I think we'll need to consider a workaround for Red Hat for now (checking for `openstack-selinux` seems the best suggestion), and then consider some better mechanism for detecting these containerised installations.

bmr-cymru:
In so far as the `check_enabled()` bugs go, the patch is fine - although it should also delete the bogus `nova` member from each class.

The problem is that on containerised deployments it will still evaulate `False` since those packages are not in the host package manager's list.

pmoravec:
OK, all openstack_* plugins updated such that each (and only) RedHatPlugin is enabled by `openstack-selinux` package only. Verifying packages is preserved against the same original sets of packages.

bmr-cymru:
I am also wondering if it's even possible for us to verify these leaf packages in the plugin with our current infrastructure? Wouldn't these require us to run the `rpm` commands in the container namespace (or against the container's RPMDB if it is accessible from the host fs)?

pmoravec:
OK, minimalistic version of the patch - `--verify` option will verify only `openstack-selinux` package, then.

PR #1415, opened 2018-09-05
[ovirt] Don't collect engine heap dump

nijinashok:
Currently the java heap dump of engine process is collected from
"/var/log/ovirt-engine/dump". The size of these dumps are very
large which will increase the size of collected sosreport
significantly. The dumps are not required for most of the cases
and hence ignore this directory and collect the 'ls' output of
the dump directory so that the support can collect it separately
if required.

Signed-off-by: Nijin Ashok <nashok@redhat.com>

---


bmr-cymru:
> The dumps are not required for most of the cases

So why not make it optional for the (presumably) some cases where it is required?

Just add a new plugin option, say `ovirt.heapdump` and users can enable it when needed (or when instructed to by a support agent).

PR #1416, opened 2018-09-05
[plugins] fix 6db459e for SCL services

pmoravec:
Calling _files_pkgs_or_cmds_present for SCLs lacks "services"
argument that was added in 6db459e commit.

Resolves: #1416

---


pmoravec:
Current upstream has a regression due to 6db459e commit on systems with SCLs (i.e. when `postgresql` installed from SCL):

    plugin postgresql does not install, skipping: _files_pkgs_or_cmds_present() takes exactly 5 arguments (4 given)

PR #1419, opened 2018-09-06
[policies] add Container-Optimized OS support

xueweiz:
---

### DIFF
--- a/sos/policies/cos.py
+++ b/sos/policies/cos.py
@@ -1,0 +1,36 @@
+# Copyright (C) Red Hat, Inc. 2020
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import CosPlugin
+from sos.policies import LinuxPolicy
+
+
+class CosPolicy(LinuxPolicy):
+    distro = "Container-Optimized OS"
+    vendor = "Google Cloud Platform"
+    vendor_url = "https://cloud.google.com/container-optimized-os/"
+    valid_subclasses = [CosPlugin]
+    PATH = "/usr/local/bin:/usr/local/sbin:/usr/bin:/usr/sbin:/bin:/sbin"
+
+    @classmethod
+    def check(cls, remote=''):
+
+        if remote:
+            return cls.distro in remote
+
+        try:
+            with open('/etc/os-release', 'r') as fp:
+                os_release = dict(line.strip().split('=') for line in fp)
+                id = os_release.get('ID')
+                return id == 'cos'
+        except IOError:
+            return False
+
+# vim: set et ts=4 sw=4 :
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -1995,6 +1995,11 @@ class SuSEPlugin(object):
     pass
 
 
+class CosPlugin(object):
+    """Tagging class for Container-Optimized OS"""
+    pass
+
+
 class IndependentPlugin(object):
     """Tagging class for plugins that can run on any platform"""
     pass
--- a/sos/report/plugins/cgroups.py
+++ b/sos/report/plugins/cgroups.py
@@ -6,10 +6,11 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.report.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+from sos.report.plugins import (Plugin, RedHatPlugin, DebianPlugin,
+                                UbuntuPlugin, CosPlugin)
 
 
-class Cgroups(Plugin, DebianPlugin, UbuntuPlugin):
+class Cgroups(Plugin, DebianPlugin, UbuntuPlugin, CosPlugin):
 
     short_desc = 'Control groups subsystem'
 
--- a/sos/report/plugins/devices.py
+++ b/sos/report/plugins/devices.py
@@ -6,10 +6,11 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.report.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+from sos.report.plugins import (Plugin, RedHatPlugin, DebianPlugin,
+                                UbuntuPlugin, CosPlugin)
 
 
-class Devices(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
+class Devices(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin, CosPlugin):
 
     short_desc = 'devices specific commands'
 
--- a/sos/report/plugins/docker.py
+++ b/sos/report/plugins/docker.py
@@ -8,10 +8,11 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.report.plugins import Plugin, RedHatPlugin, UbuntuPlugin, SoSPredicate
+from sos.report.plugins import (Plugin, RedHatPlugin, UbuntuPlugin,
+                                SoSPredicate, CosPlugin)
 
 
-class Docker(Plugin):
+class Docker(Plugin, CosPlugin):
 
     short_desc = 'Docker containers'
     plugin_name = 'docker'
--- a/sos/report/plugins/filesys.py
+++ b/sos/report/plugins/filesys.py
@@ -6,10 +6,11 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.report.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+from sos.report.plugins import (Plugin, RedHatPlugin, DebianPlugin,
+                                UbuntuPlugin, CosPlugin)
 
 
-class Filesys(Plugin, DebianPlugin, UbuntuPlugin):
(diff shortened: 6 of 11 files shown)


xueweiz:
@TurboTurtle Hi Jake, we are trying to use sosreport to support COS's customer issues. Could you help take a look of the patches in this PR? Thanks! :)

bmr-cymru:
Please put all of your commits in the format described in the [Contributor Guidelines](https://github.com/sosreport/sos/wiki/Contribution-Guidelines):

    ```
        [subsystem] brief description of change
    ```

Where `subsystem` is the plugin, or Python module(s) that have changed and the brief description summarises the content of the patch.

sos/policies/cos.py
```diff
@@ -0,0 +1,28 @@
+from __future__ import with_statement
```
bmr-cymru: You shouldnt need this on any supported Python runtime.

sos/policies/cos.py
```diff
@@ -0,0 +1,28 @@
...
+        try:
+            with open('/etc/os-release', 'r') as fp:
+                lines = fp.readlines()
+                for line in lines:
+                    if line.startswith("ID="):
+                        return line.split("=")[1].strip() == "cos"
+            return False
+        except IOError:
+            return False
+
```
bmr-cymru: You seem to have some missing bits here: you're importing `PackageManager`, but then not using it anywhere. This means that package name triggers and package verification for plugins won't work with this policy. Normally this is initialised inside the policy `__init__()` method so that policy -> package manager callbacks from plugin code work correctly.

Setting a `name_pattern` will also allow you to control default report naming templates.

sos/plugins/docker.py
```diff
@@ -8,10 +8,10 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.plugins import Plugin, RedHatPlugin, UbuntuPlugin
+from sos.plugins import Plugin, RedHatPlugin, UbuntuPlugin, CosPlugin
 
 
-class Docker(Plugin):
+class Docker(Plugin, CosPlugin):
```
bmr-cymru: This should be in a separate class (as the other examples in this file). The base `Docker` class is really abstract since the package can be deployed in different ways on different distributions (see the differing file collection lists in Red Hat vs. Ubuntu).

This is also where you get to define package triggers for the plugin.

sos/plugins/logs.py
```diff
@@ -8,10 +8,11 @@
 
 import os
 import glob
-from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+from sos.plugins import (Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin,
+    CosPlugin)
 
 
-class Logs(Plugin):
+class Logs(Plugin, CosPlugin):
```
bmr-cymru: Ditto here: it's hard to believe that you don't need any of the log path adjustments that the other derived classes in this file use.

sos/policies/cos.py
xueweiz: Thanks!

sos/policies/cos.py
xueweiz: Good catch. Yeah I should not import `PackageManager`, because COS actually do not have a package manager, similar to what CoreOS does.

sos/plugins/docker.py
xueweiz: I see. Thanks!

sos/plugins/logs.py
xueweiz: Oh sorry for the confusion. I changed this into a separate class in the following commit.

xueweiz:
> It's a very minimal set of plugins to enable. Most of these either don't have package triggers, or have both package and path triggers, so they will actually run, but it would give a very limited set of data compared to what users typically expect of sosreport.

I agree. But this is because the special nature of COS. This operating is very locked down, and there is no package manager, a large set of directories are mounted as read-only and no-exec. Basically the operating system is only used to run docker containers. The OS itself is very minimum, and very locked down. Hence the majority of the plugins do not apply here.

The most useful thing we care about, is kernel, systemd-journal, and docker logs. Most of the other things is not very important for COS's use case.

> Never collect the whole of /var/log by default.
I see. Thanks for the suggestion!

xueweiz:
Hi @bmr-cymru , could you help take a look of the patch again? :P Thanks! 
(I just realized that I probably shouldn't force-push the changes. Sorry for the inconvenience!)

bmr-cymru:
> Hi @bmr-cymru , could you help take a look of the patch again? :P Thanks!

Time is short at the moment, but I will get to this eventually.

> (I just realized that I probably shouldn't force-push the changes. Sorry for the inconvenience!)

No that's the correct thing to do: if you have changes to update a PR then always `git push --force` - closing and filing a new one just burns through PR numbers and fragments the conversation.

TurboTurtle:
Cycling back around on this. @xueweiz are you still interested in bringing COS support to sos?

At a glance there are only a couple of items at the moment:

- Even if there is not a package manager, I think a COS policy should still implement one that simply returns `True` for any checks so we don't inadvertently block plugins from running that should be run given a `CosPlugin` tag. Will need to test if this is the best approach or not, just thinking out loud at the moment.

- Rebase against current master
- Related to the rebase, the `docker` plugin now collects `/etc/docker/daemon.json` so there's no need for an override there any more. Also, please take a look at the `logs` plugin as it was recently overhauled and we may not need an override there either.
- Ideally, enable an `InitSystem` abstraction for the COS policy (E.G. https://github.com/sosreport/sos/blob/master/sos/policies/redhat.py#L39). `SoSPredicate`s that gate command/file collection make use of this for service checks. This will also be used when we implement a container runtime abstraction (in the event that COS will/does support runtimes other than docker) as per #1873

TurboTurtle:
> 
> * Even if there is not a package manager, I think a COS policy should still implement one that simply returns `True` for any checks so we don't inadvertently block plugins from running that should be run given a `CosPlugin` tag. Will need to test if this is the best approach or not, just thinking out loud at the moment.


Nope, I was wrong here. I was forgetting the enablement checks on packages is _entirely_ in `Plugin`.

TurboTurtle:
One last check-in attempt. @xueweiz are you still wanting to bring COS support into sos?

xueweiz:
Hi @TurboTurtle ! Thanks for the explanation! Sorry I missed the update email. Yes I still want to bring COS support into sosreport.

I will make the suggested changes in a few days and update this PR :) Thanks!

xueweiz:
Hi @TurboTurtle , thanks so much for the pointers above! And I'm very sorry for the delay on the response. I just followed your suggestions above and updated this PR:

> Even if there is not a package manager, I think a COS policy should still implement one that simply returns True for any checks so we don't inadvertently block plugins from running that should be run given a CosPlugin tag.

I actually didn't quite get this part. I thought we can hand pick which plugins to enable by inheriting the selected plugins on `CosPlugin`. Isn't that enough? How would returning `True` from package manager changing that logic?

> Related to the rebase, the docker plugin now collects /etc/docker/daemon.json so there's no need for an override there any more.

You are right. We can remove this override now :) Thanks!

> Also, please take a look at the logs plugin as it was recently overhauled and we may not need an override there either.

You are right. Now the logs plugin is already copying the journal logs binary [here](https://github.com/sosreport/sos/blob/ab443738f139cbee114aa96dd8d3b89089c7f927/sos/report/plugins/logs.py#L66). So there is no need for the override.

> Ideally, enable an InitSystem abstraction for the COS policy (E.G. https://github.com/sosreport/sos/blob/master/sos/policies/redhat.py#L39). SoSPredicates that gate command/file collection make use of this for service checks. This will also be used when we implement a container runtime abstraction (in the event that COS will/does support runtimes other than docker) as per #1873

I don't think this is necessary. Please let me know if you think otherwise.

I looked at `LinuxPolicy`'s code, seems it will automatically assign `SystemdInit()` when detecting the `/run/systemd/system/` directory. This logic will trigger on COS, and will assign `SystemdInit()` to the `CosPolicy`. I think that should be good enough.

TurboTurtle:
> I actually didn't quite get this part. I thought we can hand pick which plugins to enable by inheriting the selected plugins on `CosPlugin`. Isn't that enough? How would returning `True` from package manager changing that logic?
>
I hadn't realized that none of the Plugins that you added `CosPlugin` to had package checks. So for now, this is fine - but should COS want to enabled additional plugins in the future, if those plugins have any `packages` tuple defined, then the enablement check makes a call to the policy's `PackageManager`, which for COS would result in an exception when none is defined.
 

> I don't think this is necessary. Please let me know if you think otherwise.
> 
> I looked at `LinuxPolicy`'s code, seems it will automatically assign `SystemdInit()` when detecting the `/run/systemd/system/` directory. This logic will trigger on COS, and will assign `SystemdInit()` to the `CosPolicy`. I think that should be good enough.

Yup, that was a more recent change to auto load systemd - no need to do anything on your end.

xueweiz:
@TurboTurtle Hi Jake, I'm seeing travis failing with error `mkdir: cannot create directory ‘/var/tmp/sosreport_test/’: Disk quota exceeded` in the `./tests/simple.sh`.

Have you seen this before? I wonder whether it indicates the test VM was full, or sosreport is suddenly collecting a LOT more data now (which I doubt).

And could you help re-run the tests? Maybe this is just a flake. Thanks!

BryanQuigley:
@xueweiz it appears that was a transient failure in the ppc build

xueweiz:
Hi @TurboTurtle  thanks for the review and the suggestions!

I've fixed the style error. I think we can defer the implementation of `PackageManager` when we know that we really needs it. WDYT? Could you help review this PR again?

Thanks!

PR #1422, opened 2018-09-08
[juju] catch exceptions when "juju status" command fails

pmoravec:
---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -742,9 +742,10 @@ class Plugin(object):
     def add_string_as_file(self, content, filename):
         """Add a string to the archive as a file named `filename`"""
         self.copy_strings.append((content, filename))
-        content = content.splitlines()[0]
-        if not isinstance(content, six.string_types):
-            content = content.decode('utf8', 'ignore')
+        if content:
+            content = content.splitlines()[0]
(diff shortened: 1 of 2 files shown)


pmoravec:
Reproducer of underlying problem: manually enable `juju` plugin on a system without `juju` installed. Then we try to collect empty string in `add_string_as_file` method (first commit), and parse the empty response as a JSON object.

There is some code duplication in `add_string_as_file` and `_collect_strings` wrt. parsing content to be logged - this can be put into one common routine. Anyway I feel this code unification would rather make the code less readable?

pmoravec:
This needs enhancement for:

    # cat /var/tmp/test-symlink-fix-new/sosreport-pmoravec-f28-2018-09-12-zajlhgo/sos_logs/unpackaged-plugin-errors.txt | sed "s/^/    /g"
      File "/root/pmoravec-sos-test/sos-master/sos/sosreport.py", line 1116, in collect_plugin
        plug.collect()
      File "/root/pmoravec-sos-test/sos-master/sos/plugins/__init__.py", line 966, in collect
        self._collect_strings()
      File "/root/pmoravec-sos-test/sos-master/sos/plugins/__init__.py", line 947, in _collect_strings
        content = string.splitlines()[0]
    IndexError: list index out of range

PR #1426, opened 2018-09-12
[insights] update config and log file

ahumbe:
Signed-off-by: Ashish Humbe <ahumbe@redhat.com>

---

### DIFF
--- a/sos/plugins/insights.py
+++ b/sos/plugins/insights.py
@@ -15,7 +15,7 @@ class RedHatInsights(Plugin, RedHatPlugin):
     plugin_name = 'insights'
     packages = ['redhat-access-insights', 'insights-client']
     profiles = ('system', 'sysmgmt')
-    conf_file = ('/etc/insights-client/insights-client.conf', \
+    conf_file = ('/etc/insights-client/insights-client.conf',
                  '/etc/redhat-access-insights/redhat-access-insights.conf')
 
     def setup(self):
--- a/sos/plugins/insights.py
+++ b/sos/plugins/insights.py
@@ -13,16 +13,16 @@ class RedHatInsights(Plugin, RedHatPlugin):
     '''Collect config and log for Red Hat Insights
     '''
     plugin_name = 'insights'
-    packages = ['redhat-access-insights']
+    packages = ['redhat-access-insights', 'insights-client']
     profiles = ('system', 'sysmgmt')
-    conf_file = '/etc/redhat-access-insights/redhat-access-insights.conf'
+    conf_file = ('/etc/insights-client/insights-client.conf', '/etc/redhat-access-insights/redhat-access-insights.conf')
 
     def setup(self):
         self.add_copy_spec(self.conf_file)
         if self.get_option("all_logs"):
-            self.add_copy_spec("/var/log/redhat-access-insights/*.log*")
+            self.add_copy_spec("/var/log/insights-client/*.log*")
         else:
-            self.add_copy_spec("/var/log/redhat-access-insights/*.log")
+            self.add_copy_spec("/var/log/insights-client/insights-client.log")
 
     def postproc(self):
         self.do_file_sub(
--- a/sos/plugins/insights.py
+++ b/sos/plugins/insights.py
@@ -13,16 +13,10 @@ class RedHatInsights(Plugin, RedHatPlugin):
     '''Collect config and log for Red Hat Insights
     '''
     plugin_name = 'insights'
-<<<<<<< HEAD
     packages = ['redhat-access-insights', 'insights-client']
     profiles = ('system', 'sysmgmt')
     conf_file = ('/etc/insights-client/insights-client.conf', \
                  '/etc/redhat-access-insights/redhat-access-insights.conf')
-=======
-    packages = ['insights-client']
-    profiles = ('system', 'sysmgmt')
-    conf_file = '/etc/insights-client/insights-client.conf'
->>>>>>> 1f5570a5afa3aa0dfc6590e275ba83d3321e3a13
 
     def setup(self):
         self.add_copy_spec(self.conf_file)
@@ -30,12 +24,8 @@ class RedHatInsights(Plugin, RedHatPlugin):
             self.add_copy_spec("/var/log/insights-client/*.log*")
             self.add_copy_spec("/var/log/redhat-access-insights/*.log*")
         else:
-<<<<<<< HEAD
             self.add_copy_spec("/var/log/insights-client/insights-client.log")
             self.add_copy_spec("/var/log/redhat-access-insights/*.log")
-=======
-            self.add_copy_spec("/var/log/insights-client/*.log")
->>>>>>> 1f5570a5afa3aa0dfc6590e275ba83d3321e3a13
 
     def postproc(self):
         self.do_file_sub(
--- a/sos/archive.py
+++ b/sos/archive.py
@@ -428,7 +428,7 @@ class FileCacheArchive(Archive):
                 source = os.path.relpath(source)
                 self.log_debug("Adding link %s -> %s for link follow up" %
                                (link_name, source))
-                self.add_link(source, link_path)
+                self.add_link(source, link_name)
             elif os.path.isdir(host_source):
                 self.log_debug("Adding dir %s for link follow up" % source)
                 self.add_dir(host_source)


sos/plugins/insights.py
```diff
@@ -13,16 +13,16 @@ class RedHatInsights(Plugin, RedHatPlugin):
     '''Collect config and log for Red Hat Insights
     '''
     plugin_name = 'insights'
-    packages = ['redhat-access-insights']
+    packages = ['insights-client']
```
bmr-cymru: Can you be **sure** that all systems that will run the updated plugin will also have the updated Insights package? If not, then we should leave both possible names in the plugin for now, until the old name is entirely obsolete.

sos/plugins/insights.py
```diff
@@ -13,16 +13,16 @@ class RedHatInsights(Plugin, RedHatPlugin):
...
     profiles = ('system', 'sysmgmt')
-    conf_file = '/etc/redhat-access-insights/redhat-access-insights.conf'
+    conf_file = '/etc/insights-client/insights-client.conf'
```
bmr-cymru: Same comment here: if this applies to e.g. two different major versions of Red Hat Enterprise Linux then both paths need to be handled since the plugin could be used on either.

sos/plugins/insights.py
```diff
@@ -13,16 +13,16 @@ class RedHatInsights(Plugin, RedHatPlugin):
...
 
     def setup(self):
         self.add_copy_spec(self.conf_file)
         if self.get_option("all_logs"):
-            self.add_copy_spec("/var/log/redhat-access-insights/*.log*")
+            self.add_copy_spec("/var/log/insights-client/*.log*")
         else:
-            self.add_copy_spec("/var/log/redhat-access-insights/*.log")
+            self.add_copy_spec("/var/log/insights-client/*.log")
```
bmr-cymru: Likewise (and since these are just log files passed to `add_copy_spec()` you can just add them all: the plugin will add those that exist and ignore the rest.

sos/plugins/insights.py
ahumbe: The new insights-client package obsolete the older package redhat-access-insights so ideally every client should have new/updated package installed. 

# rpm -q insights-client --obsoletes 
redhat-access-insights <= 1.0.13-2.el7_5
redhat-access-proactive <= 0.3.3-0.el7_5

In this case, should we keep both package names?

sos/plugins/insights.py
ahumbe: With the updated package old config file is kept as soft link pointing to new config file. 

# ll /etc/redhat-access-insights/redhat-access-insights.conf 
lrwxrwxrwx. 1 root root 41 Sep  5 21:45 /etc/redhat-access-insights/redhat-access-insights.conf -> /etc/insights-client/insights-client.conf

so I thought to collect the new file.

sos/plugins/insights.py
ahumbe: The old log file path does not exist. 

Should I remove the "else" part to collect all insight logs?

bmr-cymru: This:

```python
            self.add_copy_spec("/var/log/insights-client/*.log*")
        else:
self.add_copy_spec("/var/log/insights-client/*.log")
```

Is incorrect - the two branches collect the exact same data. One should collect exactly one log file (the current), e.g. `insights.log` if that's the file name used, and the other collects "all" logs, using the `*.log` glob (this bug was already present in the plugin before your changes).

bmr-cymru:
> As per earlier commit: 967c015, I see the size option was removed. So can we ignore it?

Oh yes - keep forgetting that that's now merged. You're right - you only need to use an explicit size parameter now if for some reason the default is not suitable. The command line `--log-size` is automatically applied for calls that do not specify an explicit limit.

sos/plugins/insights.py
```diff
@@ -13,16 +13,16 @@ class RedHatInsights(Plugin, RedHatPlugin):
     '''Collect config and log for Red Hat Insights
     '''
     plugin_name = 'insights'
-    packages = ['redhat-access-insights']
+    packages = ['redhat-access-insights', 'insights-client']
     profiles = ('system', 'sysmgmt')
-    conf_file = '/etc/redhat-access-insights/redhat-access-insights.conf'
+    conf_file = ('/etc/insights-client/insights-client.conf', '/etc/redhat-access-insights/redhat-access-insights.conf')
```
bmr-cymru: This line looks waaaaaaaayy too long. Take a look at the Travic.CI links in the pull request and you can see the full results.

You may want to install `pycodestyle` locally, and run it from the root of your git tree as:

```
$ pycodestyle sos
```

To save time and get any errors immediately without having to wait to push.

sos/plugins/insights.py
```diff
@@ -13,16 +13,16 @@ class RedHatInsights(Plugin, RedHatPlugin):
...
 
     def setup(self):
         self.add_copy_spec(self.conf_file)
         if self.get_option("all_logs"):
-            self.add_copy_spec("/var/log/redhat-access-insights/*.log*")
+            self.add_copy_spec("/var/log/insights-client/*.log*")
         else:
-            self.add_copy_spec("/var/log/redhat-access-insights/*.log")
+            self.add_copy_spec("/var/log/insights-client/insights-client.log")
```
bmr-cymru: I thought we were going to keep both logs for now, too?

If there's a point when we think that the old package is safely deprecated (e.g. it was only available on RHEL7, never 8, and 7 is no longer receiving new sos releases) then we can remove this at that time.

sos/plugins/insights.py
```diff
@@ -13,15 +13,18 @@ class RedHatInsights(Plugin, RedHatPlugin):
...
+                 '/etc/redhat-access-insights/redhat-access-insights.conf')
 
     def setup(self):
         self.add_copy_spec(self.conf_file)
         if self.get_option("all_logs"):
+            self.add_copy_spec("/var/log/insights-client/*.log*")
             self.add_copy_spec("/var/log/redhat-access-insights/*.log*")
         else:
+            self.add_copy_spec("/var/log/insights-client/insights-client.log")
             self.add_copy_spec("/var/log/redhat-access-insights/*.log")
```
ahumbe: I had to keep the log file name as "/var/log/redhat-access-insights/*.log" because it was exceeding char limit and there is only one log file with extension ".log " old log files are named as *.log.1 , *.log.2 , *.log.3 ...etc

bmr-cymru: We don't allow merge commits in `master` - please rebase your branch to only have the necessary commits for your changes and `git push --force`.

sos/plugins/insights.py
```diff
@@ -13,16 +13,29 @@ class RedHatInsights(Plugin, RedHatPlugin):
     '''Collect config and log for Red Hat Insights
     '''
     plugin_name = 'insights'
+<<<<<<< HEAD
```
bmr-cymru: Are you sure you want this line in your commit??

bmr-cymru: You need to `git rebase -i`, and then use squash or fixup commands to combine all of these commits into a single commit that applies to `master`:

```
commit 9e5d619e89b67bdf412e96715893a6139904f1ab (HEAD -> bmr-tmp)
Author: Ashish Humbe <ahumbe@redhat.com>
Date:   Fri Oct 5 14:00:28 2018 +0530

    [insights] update config and log file

commit 6c134929569d892d01ce68ae02cb1810eee4f681
Author: Ashish Humbe <ahumbe@redhat.com>
Date:   Fri Oct 5 13:21:32 2018 +0530

    [insights] update config and log file

commit 2ca7cbcd4c1b0f617e6ff2200a0281e172e9eba7
Merge: 939d32eb 1f5570a5
Author: Ashish Humbe <ahumbe@redhat.com>
Date:   Fri Oct 5 13:13:29 2018 +0530

    Merge branch 'insights_log' of https://github.com/ahumbe/sos into insights_log

commit 939d32eb9c23ab6aaaa62ec34dc8d0e7449388f4
Author: Ashish Humbe <ahumbe@redhat.com>
Date:   Wed Sep 12 15:41:13 2018 +0530

    [insights] update config and log file

commit 1f5570a5afa3aa0dfc6590e275ba83d3321e3a13
Author: Ashish Humbe <ahumbe@redhat.com>
Date:   Wed Sep 12 15:41:13 2018 +0530

    [insights] update config and log file

```

bmr-cymru:
We will be closing `sos-3.7` very soon: there have been no updates or replies to review comments in this PR for six months. Please update the branch immediately if you wish to see this included in the next release.

bmr-cymru:
This pull request is a bit of a mess. There are four commits all with the same subject, and one merge commit. PLEASE, when we ask that you rebase, and put your patches into order, it's not just a vanity request - this costs serious time and effort for the maintainer who is trying to understand which changes are deliberate and needed, and which were accidents or mistakes when putting the branch together.

On average a reviewed PR takes me about a minute to merge. This has taken half an hour so far, and after yet more rebase conflicts I am just going to ditch the whole thing and re-commit it from the latest version.

Please review the results once they land in master to ensure that they are as intended.

PR #1429, opened 2018-09-13
[openvswitch] Add FDB bridge statistics

chaudron:
Add the ovs-appctl fdb/stats command for each bridge, and do two
iterations. This so the output can be used to get an idea how fast
the counters increment.

Signed-off-by: Eelco Chaudron <echaudron@gmail.com>

---

### DIFF
--- a/sos/plugins/openvswitch.py
+++ b/sos/plugins/openvswitch.py
@@ -43,6 +43,16 @@ class OpenVSwitch(Plugin):
             "/var/run/openvswitch/ovs-vswitchd.pid"
         ])
 
+        # Gather fdb statistical output for each OVS bridge on the host.
+        br_list_result = self.call_ext_prog("ovs-vsctl list-br")
+        if br_list_result['status'] == 0:
+            for br in br_list_result['output'].splitlines():
+                # FDB statistics command, first iteration do this early on
+                # so we get some natural delay between the two iterations.
+                self.add_cmd_output(
+                    ["sh -c \"ovs-appctl fdb/stats-show %s && date +%%T.%%3N\"" % br],
+                    suggest_filename="ovs-appctl fdb/stats-show %s  seq1" %br)
+
         self.add_cmd_output([
             # The '-s' option enables dumping of packet counters on the
             # ports.
@@ -137,6 +147,11 @@ class OpenVSwitch(Plugin):
                             "ovs-ofctl -O %s dump-flows %s" % (flow, br),
                             "ovs-ofctl -O %s dump-ports-desc %s" % (flow, br)
                         ])
+
+                # FDB statistics command, second iteration
+                self.add_cmd_output(
+                    ["sh -c \"ovs-appctl fdb/stats-show %s && date +%%T.%%3N\"" % br],
+                    suggest_filename="ovs-appctl fdb/stats-show %s  seq2" %br)
 
         # Gather info on the DPDK mempools associated with each DPDK port
         br_list_result = self.call_ext_prog("ovs-vsctl -t 5 list-br")


bmr-cymru:
Please just re-open that request, and keep using this pull request. Filing a new one for every change is disruptive, because it fractures the conversation and means you have to potentially look in several places.

Fix your branch locally (use `git rebase -i` or `git commit --amend`), and then update the current branch and PR using `git push --force`.

If you're testing code style locally, note that we now use the `pycodestyle` command rather than the deprecated `pep8`. The former catches some additional cases that `pep8` does not, so it's worth using to avoid missing errors before pushing.

sos/plugins/openvswitch.py
```diff
@@ -43,6 +43,16 @@ def setup(self):
...
 
+        # Gather fdb statistical output for each OVS bridge on the host.
+        br_list_result = self.call_ext_prog("ovs-vsctl list-br")
+        if br_list_result['status'] == 0:
+            for br in br_list_result['output'].splitlines():
+                # FDB statistics command, first iteration do this early on
+                # so we get some natural delay between the two iterations.
+                self.add_cmd_output(
+                    ["sh -c \"ovs-appctl fdb/stats-show %s && date +%%T.%%3N\"" % br],
+                    suggest_filename="ovs-appctl fdb/stats-show %s  seq1" %br)
```
bmr-cymru: Nack. We don't use the shell in `sos` unless it is completely unavoidable. There is no need to datestamp any of the data collected by sos, since the inode timestamps and logs already provide this information in multiple forms.

This also means you can get rid of `suggest_filename` - this is really intended to be used only when the automatically generated name is unwieldy, or exceeds file system name length limits.

sos/plugins/openvswitch.py
```diff
@@ -43,6 +43,16 @@ def setup(self):
             "/var/run/openvswitch/ovs-vswitchd.pid"
         ])
 
+        # Gather fdb statistical output for each OVS bridge on the host.
+        br_list_result = self.call_ext_prog("ovs-vsctl list-br")
```
bmr-cymru: `list_br_result` may be a better name here since it matches the syntax of the `ovs-vsctl` command, and would make these lines more readable.

sos/plugins/openvswitch.py
bmr-cymru: There's no need to wrap a single command in a list to pass to `add_cmd_output()` - if you look at the docs, or the source, you'll see that it accepts either a single string, or a list of strings.

sos/plugins/openvswitch.py
```diff
@@ -138,6 +148,11 @@ def setup(self):
                             "ovs-ofctl -O %s dump-ports-desc %s" % (flow, br)
                         ])
 
+                # FDB statistics command, second iteration
+                self.add_cmd_output(
+                    ["sh -c \"ovs-appctl fdb/stats-show %s && date +%%T.%%3N\"" % br],
+                    suggest_filename="ovs-appctl fdb/stats-show %s  seq2" %br)
+
```
bmr-cymru: Same here: no shell or date please.

sos/plugins/openvswitch.py
chaudron: I'm using the suggest_filename here because the same command is executed twice. I will change it and only use it the second run.

sos/plugins/openvswitch.py
chaudron: Will remove the shell part

sos/plugins/openvswitch.py
chaudron: Changed

chaudron:
It does not allow me to re-open it due to the suggested forced push above. Guess I had to re-open it before :( I'll send a new push request for now.

bmr-cymru:
It's not that - there's something odd about how this PR was closed. Normally an org. admin can re-open any pull, but since you first closed this one the "Reopen and comment" button is greyed out for me.

bmr-cymru:
Weird. We can live with another PR for this one though, but it would be good to understand how it came to be "locked" like this (I just checked and the conversation lock has not been enabled).

PR #1431, opened 2018-09-13
[openvswitch] Add FDB bridge statistics

chaudron:
Add the ovs-appctl fdb/stats command for each bridge, and do two
iterations. This so the output can be used to get an idea how fast
the counters increment.

Signed-off-by: Eelco Chaudron <echaudron@gmail.com>

---

### DIFF
--- a/sos/plugins/openvswitch.py
+++ b/sos/plugins/openvswitch.py
@@ -106,7 +106,8 @@ class OpenVSwitch(Plugin):
                     "ovs-ofctl dump-ports %s" % br,
                     "ovs-ofctl queue-get-config %s" % br,
                     "ovs-ofctl queue-stats %s" % br,
-                    "ovs-ofctl show %s" % br
+                    "ovs-ofctl show %s" % br,
+                    "ovs-appctl fdb/stats-show %s" % br
                 ])
 
                 # Flow protocols currently supported


chaudron:
I'm replying to this post in the closed pull request: https://github.com/sosreport/sos/pull/1429#discussion_r217403923

> That's a bit obscure - an unusual use like that always warrants a comment, so that it's clear to another reader (and so that someone doesn't delete the 'duplicate' version).
> 
> That said, I have some reservations about this - typically, interval and rate conversion is handled by statistics sources, rather than by collection tools like `sos`. We don't really have infrastructure for doing this at the moment, and the result looks a bit confusing. It would be much better if there was a single command that could be run that would provide either the two lots of data, or better yet, to handle the rate conversion for us as well.
> 
> The attempt to provide some elapsed time is also troublesome here; you cannot assume that because a command was the first to be added it will always be the first to be executed. Calling `add_cmd_output()` by itself only adds the command strings to a list. Collection takes place later in a thread pool (one thread per plugin). Right now we do not do any re-sorting of the commands, but there are no guarantees of this and it is not behaviour that I would encourage anybody to rely on (since it's going to break if we make more changes to the parallel executor, or to how the command lists are handled).
> 
> Aside from this being unpredictable I think it's unlikely that the interval that will elapse is going to be long enough to really accumulate any "interesting" statistics. Again, another common pattern for components that generate these is to store one or more sets somewhere in the file system (`/var/lib` typically). It would make perfect sense for `sos` to collect these.

Currently, OVS does not support automatic storing of counter snapshots or displaying pps stats over time. I also do not think those are needed in a normal environment, as a simple grep/sed/sleep loop will do it.

An interval of .5 seconds would be enough, but as you mention if the process gets optimized this can not be guaranteed.

> I'm not going to flat out nack collecting this statistics data for now, but we need to find a better mechanism for it before we can include this.

A simple solution would be to use the sh -c "" and at the sleep, but I assume you do not want the sleep in the command execution.

Assuming the above I'll re-submit the patch removing the second iteration, and based on the counters we can ask customers to get us two sos reports. Once sos repostings gets some infrastructure to support this I can add it,

bmr-cymru:
> Currently, OVS does not support automatic storing of counter snapshots or displaying pps stats over time. I 
> also do not think those are needed in a normal environment, as a simple grep/sed/sleep loop will do it.

To be honest, the more I think about this, the more I think it is not `sos`'s job to provide interval sampling and rate conversion for different component's statistics counters. There's enough complexity maintaining the core functionality.

It would be trivial for the OpenVSwitch folks to ship a simple shell or Python script to collect this, and to carry out whatever conversion is needed. Putting it in `sos` places a greater burden on both sides since there's now a need to coordinate changes to counter sets etc. between two different repositories. If it's important to support the component then it seems odd to not want to include something like this. Raw counter values, like `/proc/diskstats`, or the `device-mapper` statistics facility are not of much use to end users on their own.

Happy to collect a single command or snapshot of the data for this, but if the project needs something more advanced then it should provide it upstream itself.

> A simple solution would be to use the sh -c "" and at the sleep, but I assume you do not want the sleep in 
> the command execution.

That's not a solution really, it's a hack, and as I explained before we do not permit the use of the shell in `sos` plugins for very good and longstanding reasons: for one thing, it drives a truck through our ability to ensure that the called process does only what we expect.

chaudron:
Update patch to only do a single show bridge, with simplifies the patch to a single line add...

chaudron:
Anything else needed to push this to master?

bmr-cymru:
Time mostly - we are all tied up at the moment with deadlines and release work for other projects. Once that's out of the way we'll get back to merging PRs.

PR #1434, opened 2018-09-14
[plugins] Add new plugin cifs

pierg75:
This collects few configuration files and cifs data.

Signed-off-by: Pierguido Lambri <plambri@redhat.com>

---

### DIFF
--- a/sos/plugins/cifs.py
+++ b/sos/plugins/cifs.py
@@ -1,0 +1,32 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+
+
+class Cifs(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
+    """SMB file system information
+    """
+    plugin_name = 'cifs'
+    profiles = ('storage', 'network', 'cifs')
+    packages = ['cifs-utils']
+
+    def setup(self):
+        self.add_copy_spec([
+            "/etc/request-key.d/cifs.spnego.conf",
+            "/etc/request-key.d/cifs.idmap.conf",
+            "/proc/keys",
+            "/proc/fs/cifs/LinuxExtensionsEnabled",
+            "/proc/fs/cifs/SecurityFlags",
+            "/proc/fs/cifs/Stats",
+            "/proc/fs/cifs/DebugData",
+        ])
+        return
+
+
+# vim: set et ts=4 sw=4 :


bmr-cymru:
`[cifs] add new plugin`, please ;-)

sos/plugins/cifs.py
```diff
@@ -0,0 +1,32 @@
...
+from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+
+
+class Cifs(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
+    """SMB file system information
+    """
+    plugin_name = 'cifs'
+    profiles = ('storage', 'network', 'cifs')
+    packages = ['cifs-utils']
+
```
bmr-cymru: Might as well be a tuple here. It's never going to benefit from any of the list features.

sos/plugins/cifs.py
```diff
@@ -0,0 +1,32 @@
...
+    def setup(self):
+        self.add_copy_spec([
+            "/etc/request-key.d/cifs.spnego.conf",
+            "/etc/request-key.d/cifs.idmap.conf",
+            "/proc/keys",
+            "/proc/fs/cifs/LinuxExtensionsEnabled",
+            "/proc/fs/cifs/SecurityFlags",
+            "/proc/fs/cifs/Stats",
+            "/proc/fs/cifs/DebugData",
+        ])
```
bmr-cymru: The `/proc` suff duplicates data already collected by the `filesys` plugin (which should be considered "lower level" than any particular file system plugin) - we already NACKed a proposed `cifs` plugin a while ago because that was the only thing that it collected. We should collect `/proc/keys`, but I do not think that this plugin is the right place for it.

And don't we want any of the `/etc/` stuff the CIFS packages provide?

```
$ ls /etc/cifs-utils/
idmap-plugin
$ rpm -qf /etc/cifs-utils/
cifs-utils-6.7-5.fc26.x86_64
sssd-client-1.16.1-4.fc26.x86_64
```

I'm pretty sure there must be other command output, config files etc. that would be useful for this file system.

pierg75: Sorry I thought this time I did my homework :(

About  `idmap-plugin -> /etc/alternatives/cifs-idmap-plugin`, it's a link to `/etc/alternatives/cifs-idmap-plugin -> /usr/lib64/cifs-utils/cifs_idmap_sss.so` which is provided by the `sssd-client` package (and it's not really interesting).

About `/proc/keys` actually this idea started because I had to ask it all the time and I thought that we don't collect much about cifs, so I gave it a go.
I've searched through the code for the keyword `cifs` and obviously forgot to check `proc/fs`, but I just ran a sos on a RHEL7 machine and nothing was captured from `proc/fs`:

```
# ls -l /proc/fs/
total 0
dr-xr-xr-x. 2 root root 0 Sep 17 16:51 cifs
dr-xr-xr-x. 2 root root 2 Sep 17 16:51 nfsd
dr-xr-xr-x. 2 root root 0 Sep 17 16:51 xfs

# pwd
/var/tmp/sosreport-rhel7-test2-win-20180917164524

# ls proc/fs
ls: cannot access proc/fs: No such file or directory
```

Do I miss something?
It seems to work on the upstream version on Fedora though.

About other commands, I don't really think there's a lot of cifs commands. Most of the executables installed by `cifs-utils` are the upcall and idmap utilities and the mount helper (`mount.cifs`)

PR #1436, opened 2018-09-20
[lustre] Collect ring buffer dumps on all_logs

utopiabound:
Collect emergancy dumps of the lustre ring buffer if all_logs is set.

---

### DIFF
--- a/sos/plugins/lustre.py
+++ b/sos/plugins/lustre.py
@@ -51,4 +51,8 @@ class Lustre(Plugin, RedHatPlugin):
         self.get_params("quota", ["osd-*.*.quota_slave." +
                                   "{info,limit_*,acct_*}"])
 
+        # Grab emergency ring buffer dumps
+        if self.get_option("all_logs"):
+            self.add_copy_spec("/tmp/lustre-log.*")
+
 # vim: set et ts=4 sw=4 :


sos/plugins/lustre.py
```diff
@@ -51,4 +51,10 @@ def setup(self):
         self.get_params("quota", ["osd-*.*.quota_slave." +
                                   "{info,limit_*,acct_*}"])
 
+        # Grab emergency ring buffer dumps
+        if self.get_option("all_logs"):
+            self.add_copy_spec([
+                "/tmp/lustre-log.*"
+            ])
+
```
bmr-cymru: No need to wrap a single path in a list; `add_copy_spec()` accepts either a string, or an iterable of strings.

sos/plugins/lustre.py
```diff
@@ -51,4 +51,10 @@ def setup(self):
         self.get_params("quota", ["osd-*.*.quota_slave." +
                                   "{info,limit_*,acct_*}"])
 
+        # Grab emergency ring buffer dumps
+        if self.get_option("all_logs"):
+            self.add_copy_spec([
+                "/tmp/lustre-log.*"
```
bmr-cymru: What causes this file to be written? Is it present at all times, or is some action required to cause it to be present?

bmr-cymru:
OK; that's not a problem (all file collection is speculative anyway). There's a similar (but awful) mechanism in Gluster, which requires the sending of mysterious signals (OK, SIGUSR1) and other strange rituals.

PR #1437, opened 2018-09-24
[unpackaged] compare realpaths of files

pmoravec:
To compare files in $PATH with files installed from a package, we must
expand all symlinks to their realpaths. Otherwise we get false positives
like /bin/systemctl (as /bin -> /usr/bin).

Resolves: #1437

---


pmoravec:
The patch has a drawback in performance penalty. Prior the patch, `unpackaged` plugin took 5s (and reported false positives like the `/bin/systemctl`), now it runs properly but for 11s.

But I dont see a way of improving it. Both sets must be "realpath-ed" - also the `all_frpm` - as I can imagine a package containing a symlink to a real file from other package.

PR #1441, opened 2018-10-02
[policies] Default to sha256 over md5

BryanQuigley:
Previously sha256 only used if in a FIPS environment we now
default to it over md5.

From looking at dates and availability I believe python3 installations
should all have sha256 so used that as a cutoff point.  If you are
running sosreport with python2 you will stay on md5.

---

### DIFF
--- a/sos/policies/__init__.py
+++ b/sos/policies/__init__.py
@@ -9,6 +9,7 @@ import fnmatch
 import tempfile
 import random
 import string
+import sys
 from os import environ
 
 from sos.utilities import (ImporterHelper,
@@ -611,7 +612,7 @@ No changes will be made to system configuration.
     def get_preferred_hash_name(self):
         """Returns the string name of the hashlib-supported checksum algorithm
         to use"""
-        return "md5"
+        return "sha256"
 
     def display_results(self, archive, directory, checksum):
         # Display results is called from the tail of SoSReport.final_work()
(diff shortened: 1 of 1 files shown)


bmr-cymru:
Considering that we don't use this for cryptographic purposes, and that the SHA256 hashes are a lot more cumbersome to handle (+32 chars), what's the benefit of making this change for sosreport's uses?

sos/policies/__init__.py
```diff
@@ -800,17 +801,11 @@ def get_preferred_hash_name(self):
...
-            fp = open("/proc/sys/crypto/fips_enabled", "r")
-        except IOError:
-            self._preferred_hash_name = checksum
-            return checksum
-
-        fips_enabled = fp.read()
-        if fips_enabled.find("1") >= 0:
+        if sys.version_info[0] <= 2:
+            checksum = "md5"
+        else:
```
bmr-cymru: This would break RHEL7 which has a py2 runtime and an optional FIPS mode. You can't just assume that Python 2 equates to a non-FIPS environment.

BryanQuigley:
I'd like us to default more to sha256 if we still want to have a separate signature.  the other option would be to build it into Xz.  I wanted less special case things for FIPS cases, but this patch doesn't make sense as is. Sorry.

PR #1442, opened 2018-10-03
[kubernetes] improve parsing namespaces

pmoravec:
Handle (theoretical) cases when "kubectl get namespace" returns a line with
white characters only, and prevent IndexError to be raised.

Resolves: #1442

---

### DIFF
--- a/sos/plugins/kubernetes.py
+++ b/sos/plugins/kubernetes.py
@@ -69,7 +69,9 @@ class kubernetes(Plugin, RedHatPlugin):
 
         # get all namespaces in use
         kn = self.get_command_output('%s get namespaces' % kube_cmd)
-        knsps = [n.split()[0] for n in kn['output'].splitlines()[1:] if n]
+        # namespace is the 1st word on line, until the line has spaces only
+        kn_output = kn['output'].splitlines()[1:]
+        knsps = [n.split()[0] for n in kn_output if n and len(n.split())]
 
         resources = [
             'limitrange',


pmoravec:
This is rather improvement of theoretical situation when `kubectl` returns some error output like:

    Error: unknown flag: --config\n\n\nExamples:\n  # List all pods in ps output format.\n  kubectl get pods\n  \n  # List all pods in ps output format with more information (such as node name).\n  kubectl get pods -o wide\n  \n..'

Then parsing the lines '\n  \n' raises the exception.

sos/plugins/kubernetes.py
```diff
@@ -69,7 +69,9 @@ def setup(self):
 
         # get all namespaces in use
         kn = self.get_command_output('%s get namespaces' % kube_cmd)
-        knsps = [n.split()[0] for n in kn['output'].splitlines()[1:] if n]
+        # namespace is the 1st word on line, until the line has spaces only
+        knsps = [n.split()[0] for n in kn['output'].splitlines()[1:]
+                 if n and len(n.split())]
```
bmr-cymru: I wonder if this might be more readable as:

```python
    kn_output = kn['output'].splitlines()[1:]
    knsps = [n.split()[0] for n in kn_output if n and len(n.split())]
```

70 chars on the long line... I don't know, but that hanging `if` really confused my eyes for a moment on the first look.

pmoravec: Good point for more code clarity there, updating the PR..

PR #1444, opened 2018-10-04
[pacemaker] avoid unnecessary collection

kgaillot:
---

### DIFF
--- a/sos/plugins/pacemaker.py
+++ b/sos/plugins/pacemaker.py
@@ -66,9 +66,6 @@ class Pacemaker(Plugin):
 
     def setup(self):
         self.add_copy_spec([
-            # Pacemaker cluster configuration file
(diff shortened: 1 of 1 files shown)


kgaillot:
You're right. Red Hat distros ship pcs only, and Debian distros typically offer both crm shell and pcs (users typically pick one). I left the pcs handling separate for future-proofing, both because other distros (e.g. OpenSuSE) ship only crm shell and might one day get added to sos, and also because other front-ends are or may become available that may become popular enough to add as another option. But as you say, a few command-not-found messages aren't a big deal anyway.

bmr-cymru:
Makes sense - we do now have a policy for SuSE, so porting the Pacemaker plugin should be pretty straightforward following the existing pattern.

Issue #1264, opened 2018-04-04
[policies] add policies for missing RH rebuild distros


bmr-cymru:
Currently we report as RHEL/Red Hat on distributions like CentOS, Scientific Linux etc. Fix this by adding policy subclasses for these builds and overriding the appropriate properties from the base RHEL class.

mangelajo:
Oh, yes, it's handled in packaging here: 

https://git.centos.org/blob/rpms!sos.git/c7/SOURCES!sos-3.4-centos-branding.patch

But it's really painful to discover this through a 2h debugging session when your devel env of sos (under CentOS) does not recognize your system, and returns an empty list of plugins after filtering...

bmr-cymru:
I reached out to the CentOS folks some years ago (both before and after the Red Hat join up), and there was not much interest in working on a CentOS policy at the time, and it was also cumbersome for me to get maintainer access to the sos SRPM in order to fix it myself.

I have not enquired in a long time, so maybe now is a good time to get back to this. No distro should be patching policy files downstream, unless it is to address actual bugs in released `sos` versions.

bmr-cymru:
Right: that patch is exactly what Oracle were doing a few years back. It's not the best approach. Better to either add to the `redhat` file, or make a new one and `import sos.policied.redhat`. Then you either add the `RedHatPlugin` tagging class to the new policy, or better yet, subclass the existing ones and modify the instance variables to suit. This is easier and a lot more maintainable.

mangelajo:
@bmr-cymru there you have it, I hope the patch looks good :)

TurboTurtle:
Closing this as the CentOS policy is now defined in master alongside other Red Hat family distros, see #1447

PR #1447, opened 2018-10-10
[policies] Detect CentOS distribution without D/S patches

mangelajo:
This partially fixes issue #1264 by handling the CentOS and
CentOS Atomic hosts, partially based on the CentOS packaging
patch [1]. Without this patch, if you try to develop sosreports
on a CentOS host, you will get an empty list of plugins.

[1] https://git.centos.org/blob/rpms!sos.git/c7/SOURCES!sos-3.4-centos-branding.patch

Signed-off-by: Miguel Angel Ajo <majopela@redhat.com>

---

### DIFF
--- a/sos/policies/redhat.py
+++ b/sos/policies/redhat.py
@@ -279,7 +279,7 @@ No changes will be made to system configuration.
                 if line.startswith("NAME"):
                     (name, value) = line.split("=")
                     value = value.strip("\"'")
-                    if value.startswith(RHEL_RELEASE_STR):
+                    if value.startswith(cls.distro):
                         return True
         return False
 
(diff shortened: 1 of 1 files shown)


bmr-cymru:
Ack to the approach and a quick look at the patch, but before we could include this in a new release can you make sure that someone in CentOS is going to be aware that a new version is coming which they will need to adapt their packaging for? We try not to drop surprised like that on distributions without at least a headsup :-)

mangelajo:
I used twitter to the patch author, https://twitter.com/mangel_ajo/status/1050297348486832130 .

mangelajo:
Johnny didn't see my twitter message, I sent an email, @bmr-cymru you're in copy of the message.  Cheers :)

PR #1451, opened 2018-10-15
[networking] Collect NUMA Node of each NIC

superjamie:
It is often useful to know the NUMA locality of each network device.
Collect /sys/class/net/*/device/numa_node to add this information.

Signed-off-by: Jamie Bainbridge <jamie.bainbridge@gmail.com>

---

### DIFF
--- a/sos/plugins/networking.py
+++ b/sos/plugins/networking.py
@@ -129,6 +129,7 @@ class Networking(Plugin):
             "/etc/sysconfig/nftables.conf",
             "/etc/nftables.conf",
             "/etc/dnsmasq*",
+            "/sys/class/net/*/device/numa_node",
             "/sys/class/net/*/flags",
             "/sys/class/net/*/statistics/",
             "/etc/iproute2"


pmoravec:
ACK, just for coherent / code readability reasons, it makes a bit more sense to enumerate the `/sys/class/net/*/device/numa_node` around other `/sys/class/net/*`paths collected (i.e. add the line before `/etc/iproute2` and not after).

superjamie:
Cool thanks. I modified the patch as suggested.

pmoravec:
+1 thanks.

bmr-cymru:
It's worth having a thorough check through a test report with this change - we've made a bunch of changes that affect `/sys` collection recently, and those have been tested heavily against `/sys/devices`, and `/sys/block` - I do not envisage any problems with the extra `/sys/net` bits, but it's good to rule that out early on.

bmr-cymru:
The main thing to look for is symlink correctness:

```
# find -L /path/to//sosreport -lname "*" 2>&1 | grep -v 'File system loop'
```

And see if any broken links pop up.

Issue #1452, opened 2018-10-15
[filesys] incorrect interface when using dumpe2fs 1.42.13


xueweiz:
I noticed that the output of filesys plugin on Ubuntu 16.04.5 has the file "/sos_commands/filesys/dumpe2fs_.dev.sda1_." with below content:
> dumpe2fs 1.42.13 (17-May-2015)
> Usage: dumpe2fs [-bfhixV] [-o superblock=<num>] [-o blocksize=<num>] device

This means there was some bug preventing me from getting dumpe2fs results.

Seems that when sosreport running command `dumpe2fs -h /dev/sda1 /`, the interface is wrong. See below examples when I tried to use dumpe2fs with different parameters, which shows the correct interface should be `dumpe2fs -h /dev/sda1`.

> xueweiz@ubuntu:\~$ grep ext4 /proc/mounts
> /dev/sda1 / ext4 rw,relatime,data=ordered 0 0

> xueweiz@ubuntu:\~$ sudo dumpe2fs -h /dev/sda1 /
> dumpe2fs 1.42.13 (17-May-2015)
> Usage: dumpe2fs [-bfhixV] [-o superblock=<num>] [-o blocksize=<num>] device

> xueweiz@ubuntu:\~$ sudo dumpe2fs -h /dev/sda1
> dumpe2fs 1.42.13 (17-May-2015)
> Filesystem volume name:   cloudimg-rootfs
> Last mounted on:          /
> Filesystem UUID:          86a66005-05be-4164-9f67-92db034d8e6b
> Filesystem magic number:  0xEF53

Then I tried the same thing on Redhat 9, and the bug did not reproduce. Seems the reason is Redhat 9 is using dumpe2fs 1.42.9, where both interfaces (`dumpe2fs -h /dev/sda1 /` and `dumpe2fs -h /dev/sda1`) are tolerated:

> [xueweiz@redhat2 ~]$ grep ext4 /proc/mounts
> /dev/sdb /mnt/disks/2 ext4 rw,seclabel,relatime,discard,data=ordered 0 0

> [xueweiz@redhat2 ~]$ sudo dumpe2fs -h /dev/sdb /mnt/disks/2
> dumpe2fs 1.42.9 (28-Dec-2013)
> Filesystem volume name:   <none>
> Last mounted on:          <not available>
> Filesystem UUID:          df40ff91-df45-4fb1-bac7-7c06d9971a7b
> Filesystem magic number:  0xEF53
[...]

PR #1453, opened 2018-10-15
[filesys] Use the correct interface when calling dumpe2fs

xueweiz:
dumpe2fs takes one parameter indicating the device to be dumped.

Resolves: #1452

Signed-off-by: Xuewei Zhang <xueweiz@google.com>

---


xueweiz:
Thanks for the suggestion! Just updated it.

xueweiz:
@bmr-cymru Hi Bryn, could you help take a look of the patch again? Thanks!

PR #1456, opened 2018-10-18
[logging] Always cleanup temp directory

TurboTurtle:
When running `sosreport -l`, or the more specific --list-* options, sos
still creates a temp directory and log files, however we were not
cleaning those up - so over time we could have a multitude of sos temp
directories littering the filesystem.

Now always cleanup the temp directory after execution, unless the
--build option is specified by the user.

---

### DIFF
--- a/sos/sosreport.py
+++ b/sos/sosreport.py
@@ -1497,5 +1497,6 @@ def main(args):
     """The main entry point"""
     sos = SoSReport(args)
     sos.execute()
+    sos._cleanup()
 
 # vim: set et ts=4 sw=4 :


bmr-cymru:
Do we even want to keep it in the `--build` case? I'm not sure I see a use (since we'll have already copied all the data into the output report path).

TurboTurtle:
I was thinking the same thing originally, but the --help on --build says "preserve the temporary directory", which is what made add that check.

bmr-cymru:
The message is the bug. What it should say (and has always meant) is something like "do not create a compressed archive of the final report directory".

bmr-cymru:
(in the past, when we didn't care about racey tmpfiles use they were the same thing - the "temporary directory" _was_ `/var/tmp/sosreport-blahblahblah/`. That's not been the case for a few years since we push them all down into `sos.XXX` in order to close that vulnerability).

TurboTurtle:
Updated to not reference the `--build` option. As we talked about on IRC, with debug mode I think saying "this is debug mode so you get all the pieces" is the right approach. In my test I still had the sos.XXXX directory hanging around post-trap.

PR #1458, opened 2018-10-20
[networking] call ethtool for proper vlan device names

pmoravec:
Dont rely on device names from "ip -o link" output where the names
contain also physical device name. Rather get the list of device names
from listing /sys/class/net/ directory.

Resolves: #1458

---

### DIFF
--- a/sos/plugins/networking.py
+++ b/sos/plugins/networking.py
@@ -43,18 +43,6 @@ class Networking(Plugin):
             out.append(br_name)
         return out
 
-    def get_eth_interfaces(self, ip_link_out):
-        """Return a dictionary for which keys are ethernet interface
-        names taken from the output of "ip -o link".
-        """
-        out = {}
-        for line in ip_link_out.splitlines():
-            match = re.match('.*link/ether', line)
-            if match:
-                iface = match.string.split(':')[1].lstrip()
-                out[iface] = True
-        return out
-
     def get_ip_netns(self, ip_netns_file):
         """Returns a list for which items are namespaces in the output of
         ip netns stored in the ip_netns_file.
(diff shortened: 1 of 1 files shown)


pmoravec:
Kudos to Jonathan Maxwell from Red Hat who wrote the patch.

sos/plugins/networking.py
```diff
@@ -251,8 +227,15 @@ def setup(self):
             # Devices that exist in a namespace use less ethtool
             # parameters. Run this per namespace.
             for namespace in self.get_ip_netns(ip_netns_file):
-                for eth in self.get_netns_devs(namespace):
-                    ns_cmd_prefix = cmd_prefix + namespace + " "
+                ns_cmd_prefix = cmd_prefix + namespace + " "
+                netns_netdev_list = self.call_ext_prog(ns_cmd_prefix +
+                                                       "ls -1 /sys/class/net/")
+                for eth in netns_netdev_list['output'].splitlines():
```
bmr-cymru: Is there a reason not to use `os.listdir()` here? It is typically cleaner than shelling out to `ls`...

sos/plugins/networking.py
```diff
@@ -251,8 +227,15 @@ def setup(self):
             # Devices that exist in a namespace use less ethtool
             # parameters. Run this per namespace.
             for namespace in self.get_ip_netns(ip_netns_file):
-                for eth in self.get_netns_devs(namespace):
-                    ns_cmd_prefix = cmd_prefix + namespace + " "
+                ns_cmd_prefix = cmd_prefix + namespace + " "
+                netns_netdev_list = self.call_ext_prog(ns_cmd_prefix +
+                                                       "ls -1 /sys/class/net/")
```
bmr-cymru: Is there a reason not to use `os.listdir()` here? It is typically cleaner than shelling out to `ls`...

bmr-cymru: Ah, OK. The ns prefix. I wonder if there isn't a better way to do that, it's kinda fugly...

pmoravec: Yeah, haven't found a direct pythonic way outside calling shell. I can add a comment if it clarifies the code..?

PR #1459, opened 2018-10-22
[openstack_novajoin] Novajoin logs and conf

valleedelisle:

---

### DIFF
--- a/sos/plugins/openstack_novajoin.py
+++ b/sos/plugins/openstack_novajoin.py
@@ -1,0 +1,40 @@
+# Copyright (C) 2018 Red Hat, Inc., David Vallee Delisle <dvd@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.plugins import Plugin, RedHatPlugin
+
+
+class OpenStackNovajoin(Plugin):
+    """OpenStack Novajoin
+    """
+    plugin_name = "openstack_novajoin"
+    profiles = ('openstack', 'openstack_undercloud')
+
+    def setup(self):
+        self.add_copy_spec("/etc/novajoin/")
+        if self.get_option("all_logs"):
+            self.add_copy_spec("/var/log/novajoin/")
+        else:
+            self.add_copy_spec("/var/log/novajoin/*.log")
+
+    def postproc(self):
+        regexp = (r"(?i)password=(.*)")
+        self.do_file_sub("/etc/novajoin/join.conf", regexp,
+                         r"password=*********")
+
+
+class RedHatNovajoin(OpenStackNovajoin, RedHatPlugin):
+
+    packages = ('python-novajoin',)
+
+    def setup(self):
+        super(RedHatNovajoin, self).setup()
+
+# vim: set et ts=4 sw=4 :


sos/plugins/openstack_novajoin.py
```diff
@@ -0,0 +1,47 @@
...
+
+class OpenStackNovajoin(Plugin):
+    """OpenStack Novajoin
+    """
+    plugin_name = "openstack_novajoin"
+    profiles = ('openstack', 'openstack_undercloud')
+
+    def setup(self):
+        if self.get_option("all_logs"):
+            self.add_copy_spec([
```
TurboTurtle: So we seem to have dropped `/etc/novajoin/` in this revision. Was that intentional?

We could collapse this into a single call using something like:

```python
nova_dir = "/var/log/novajoin/"
    nova_dir += "*.log"

self.add_copy_spec([
    nova_dir,
    "/etc/novajoin/"
])
```

sos/plugins/openstack_novajoin.py
```diff
@@ -0,0 +1,47 @@
...
+                  "(_KEY)*=)(.*)")
+        self.do_file_sub("/var/log/novajoin/*.log",
+                         regexp, r"\g<var>*********")
+        self.do_file_sub("/etc/novajoin/novajoin.conf", regexp,
+                         r"\g<var>*********")
+
+
+class RedHatNovajoin(OpenStackNovajoin, RedHatPlugin):
+
+    packages = ('python-novajoin',)
```
TurboTurtle: Is this the only package to enable on?

sos/plugins/openstack_novajoin.py
```diff
@@ -0,0 +1,47 @@
...
+    """OpenStack Novajoin
+    """
+    plugin_name = "openstack_novajoin"
+    profiles = ('openstack', 'openstack_undercloud')
+
+    def setup(self):
+        if self.get_option("all_logs"):
+            self.add_copy_spec([
+                "/var/log/novajoin/"
+            ])
```
bmr-cymru: Both `add_copy_spec()` and `add_cmd_output()` accept either a single string, or a list of strings - please don't wrap single items in `[...]` as it causes things to spill over multiple lines unnecessarily.

sos/plugins/openstack_novajoin.py
```diff
@@ -0,0 +1,47 @@
...
+        else:
+            self.add_copy_spec([
+                "/var/log/novajoin/*.log"
+            ])
```
bmr-cymru: Ditto.

sos/plugins/openstack_novajoin.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def postproc(self):
+        regexp = (r"(?i)(?P<var>(UNDERCLOUD|OS)_[^\s|^=]*[_]*(PASSWORD|SECRET)"
+                  "(_KEY)*=)(.*)")
+        self.do_file_sub("/var/log/novajoin/*.log",
+                         regexp, r"\g<var>*********")
+        self.do_file_sub("/etc/novajoin/novajoin.conf", regexp,
```
bmr-cymru: With large log files (e.g. `--all-logs --log-size 0`) this may become expensive. Does this component really write passwords and other secrets to its logs?

sos/plugins/openstack_novajoin.py
```diff
@@ -0,0 +1,47 @@
...
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.plugins import Plugin, RedHatPlugin
+import os
```
bmr-cymru: You don't seem to be using this anywhere - if it's not needed, best to just drop it.

sos/plugins/openstack_novajoin.py
valleedelisle: I removed the search in the logs because I didn't see anything that looks like a password, but there's one in novajoin.conf. I don't think the log folder will ever grow much. It's not a very verbose service from what I can see.

PR #1461, opened 2018-10-23
[sosreport] Use 'append' mode in _log_plugin_exception()

ins-pirat-ion:
[archive]

Add mode (e.g. to allow append) support to add_string() function.

[sosreport]

Use 'append' mode in _log_plugin_exception() so previous
tracebacks are not overwritten when more of them occur
in one module.

Signed-off-by: Filip Krska <fkrska@redhat.com>

---

### DIFF
--- a/sos/archive.py
+++ b/sos/archive.py
@@ -85,7 +85,7 @@ class Archive(object):
     def add_file(self, src, dest=None):
         raise NotImplementedError
 
-    def add_string(self, content, dest):
+    def add_string(self, content, dest, mode='w'):
         raise NotImplementedError
 
     def add_binary(self, content, dest):
@@ -366,7 +366,7 @@ class FileCacheArchive(Archive):
             self.log_debug("added %s to FileCacheArchive '%s'" %
                            (file_name, self._archive_root))
 
-    def add_string(self, content, dest):
+    def add_string(self, content, dest, mode='w'):
         with self._path_lock:
             src = dest
 
@@ -376,7 +376,7 @@ class FileCacheArchive(Archive):
             # on file content.
             dest = self._check_path(dest, P_FILE, force=True)
 
-            f = codecs.open(dest, 'w', encoding='utf-8')
+            f = codecs.open(dest, mode, encoding='utf-8')
             if isinstance(content, bytes):
                 content = content.decode('utf8', 'ignore')
             f.write(content)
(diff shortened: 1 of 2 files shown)


ins-pirat-ion:
Can be reproduced e.g. with braking unpackaged plugin:


--- sos/plugins/unpackaged.py.orig	2018-06-25 06:42:14.000000000 -0400
+++ sos/plugins/unpackaged.py	2018-10-23 10:50:20.588818766 -0400
@@ -51,9 +51,6 @@
                             path = os.path.abspath(os.readlink(path))
                     except Exception:
                         continue
-                    file_list.append(path)
-                for name in dirs:
-                    file_list.append(os.path.join(root, name))
 
             return file_list
 
@@ -75,7 +72,7 @@
         for d in get_env_path_list():
             all_fsystem += all_files_system(d)
         not_packaged = [x for x in all_fsystem if x not in all_frpm]
-        not_packaged_expanded = format_output(not_packaged)
+        not_packaged_expanded = ""
         self.add_string_as_file('\n'.join(not_packaged_expanded), 'unpackaged')
 
^^ throws 2 exceptions:

 Setting up plugins ...
caught exception in plugin method "unpackaged.setup()"
writing traceback to sos_logs/unpackaged-plugin-errors.txt

  Starting 1/1   unpackaged      [Running: unpackaged]
caught exception in plugin method "unpackaged.collect()"
writing traceback to sos_logs/unpackaged-plugin-errors.txt

Before https://github.com/sosreport/sos/commit/b96bdab03f06408e162b1733b20e8ba9fbf8e012 only first one is written in sos_logs/unpackaged-plugin-errors.txt, after b96bdab only the last one.

ins-pirat-ion:
> Wait a second... can you explain what happened here that prompted this?

I noticed those 2 exceptions (and only one traceback logged) when investigating https://bugzilla.redhat.com/show_bug.cgi?id=1632251 (just the real world reproducer was more complicated there...)

> The only currently valid case of this that I can think of (and I'd need to check `sosreport.py` and `plugins/__init__.py` to be certain), is an exception during more than one of the plugins' `setup()`, `collect()` and `postproc()` methods - that is a legitimate case that we should fix, as I suspect as you point out that today it will overwrite the earlier exception with the later. I would prefer though to remove the ambiguity here by including the method name in the exception file name: something like `unpackaged-plugin-setup-errors.txt`, `unpackaged-plugin-postproc-errors.txt` etc. rather than just writing them into the same file.

I agree, ^^ would be more coherent approach, thanks.

PR #1462, opened 2018-10-25
[cnv] new sos plugin for Container-native virtualization

vakhanna:
Addition of a new sos plugin. CNV plugin collects information related to
Container-native Virtualization add-on to OpenShift Container Platform

Signed-off-by: Varun Khanna <vkhanna@redhat.com>

---

### DIFF
--- a/sos/plugins/cnv.py
+++ b/sos/plugins/cnv.py
@@ -1,0 +1,234 @@
+# Copyright (C) 2018 Red Hat, Inc., Varun Khanna <vkhanna@redhat.com>
+#
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+#
+# This plugin collects information related to Container-native Virtualization
+# add-on to OpenShift Container Platform.
+
+from sos.plugins import Plugin, RedHatPlugin
+import os.path
+
+
+class CNV(Plugin, RedHatPlugin):
+
+    ''' Container-native Virtualization '''
+
+    plugin_name = 'cnv'
+    profiles = ('openshift',)
+
+    master_base_dir = "/etc/origin/master"
+    master_cfg = os.path.join(master_base_dir, "master-config.yaml")
+
+    # This plugin currently will also run on OCP with no cnv add-on.
+    # In future, add a check to run the plugin only when cnv is installed
+    files = (master_cfg,)
+
+    option_list = [
+        ("vm", "collect VM related data, usage: cnv.vm=<vmi name>", '', ""),
+    ]
+
+    def get_resc_ns(self, resc):
+        out = []
+        try:
+            resc_ns_out = open(resc).read()
+        except IOError:
+            return out
+        for line in resc_ns_out.splitlines():
+            if line.startswith("NAMESPACE") \
+                    or line.startswith("No resources found") \
+                    or line.startswith("Error") \
+                    or line.startswith("error"):
+                pass
+            else:
+                out.append(line.split()[1] + " -n " + line.split()[0])
+        return out
+
+    def get_vm_ns(self, vms, option):
+        out = []
+        try:
+            vms_ns_out = open(vms).read()
+        except IOError:
+            return out
+        for line in vms_ns_out.splitlines():
+            if line.startswith("NAMESPACE") \
+                    or line.startswith("No resources found") \
+                    or line.startswith("Error") \
+                    or line.startswith("error"):
+                pass
+            else:
+                if line.split()[1] == option:
+                    out.append(line.split()[1] + " -n " + line.split()[0])
+        return out
+
+    def get_node_info(self, pods):
+        out = ""
+        try:
+            node_out = open(pods).read()
+        except IOError:
+            return out
+        for line in node_out.splitlines():
+            if line.startswith("NAMESPACE") \
+                    or line.startswith("No resources found") \
+                    or line.startswith("Error") \
+                    or line.startswith("error"):
+                pass
+            else:
+                out = line.split()[7]
+        return out
+
+    def setup(self):
+        oc_admin_cfg = os.path.join(self.master_base_dir, "admin.kubeconfig")
+        oc_cmd = "%s --config=%s" % ("oc", oc_admin_cfg)
+
+        all_dep = [
+            "%s get deployment -l kubevirt.io "
+            "--all-namespaces -o wide" % oc_cmd,
+        ]
+        all_svc = [
+            "%s get service -l kubevirt.io "
+            "--all-namespaces -o wide" % oc_cmd,
+            "%s get service -l cdi.kubevirt.io "
+            "--all-namespaces -o wide" % oc_cmd,
+        ]
+        all_pods = [
+            "%s get pods -l kubevirt.io=virt-controller "
+            "--all-namespaces -o wide" % oc_cmd,
+            "%s get pods -l kubevirt.io=virt-api "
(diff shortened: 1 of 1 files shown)


sos/plugins/cnv.py
```diff
@@ -0,0 +1,234 @@
...
+            "%s get pods -l cdi.kubevirt.io "
+            "--all-namespaces -o wide" % oc_cmd,
+            "%s get pods -l app=containerized-data-importer "
+            "--all-namespaces -o wide" % oc_cmd,
+        ]
+        all_stor = [
+            "%s get pvc --all-namespaces -o wide" % oc_cmd,
+        ]
+
+        subcmds = [
```
TurboTurtle: Most of these are either already collected by the kubernetes plugin, or are proposed to be added to it in #1424

sos/plugins/cnv.py
```diff
@@ -0,0 +1,234 @@
...
+        ]
+        subcmds_svc = [
+            "describe svc",
+        ]
+        subcmds_pods = [
+            "describe pod",
+            "logs",
+        ]
+
+        self.add_cmd_output(["%s version" % oc_cmd])
```
TurboTurtle: I would assume this should go in the origin plugin instead.

sos/plugins/cnv.py
```diff
@@ -0,0 +1,234 @@
...
+        self.add_cmd_output(["virtctl --kubeconfig %s version" % oc_admin_cfg])
+        for s in subcmds:
+            self.add_cmd_output(["%s " % oc_cmd + "get " + s])
+        for dep in all_dep:
```
TurboTurtle: These are generic enough kubernetes resources that they should be captured in the kubernetes plugin - #1424 again.

sos/plugins/cnv.py
```diff
@@ -0,0 +1,234 @@
...
+            for d in self.get_resc_ns(self.get_cmd_output_now(dep)):
+                for s in subcmds_dep:
+                    self.add_cmd_output(["%s " % oc_cmd + s + " " + d])
+        for svc in all_svc:
```
TurboTurtle: Same here - they're generic kubernetes resources, and in my opinion should be folded in therre.

sos/plugins/cnv.py
```diff
@@ -0,0 +1,234 @@
...
+            for sv in self.get_resc_ns(self.get_cmd_output_now(svc)):
+                for s in subcmds_svc:
+                    self.add_cmd_output(["%s " % oc_cmd + s + " " + sv])
+        for pod in all_pods:
```
TurboTurtle: Ditto.

sos/plugins/cnv.py
```diff
@@ -0,0 +1,234 @@
...
+            for p in self.get_resc_ns(self.get_cmd_output_now(pod)):
+                for s in subcmds_pods:
+                    self.add_cmd_output(["%s " % oc_cmd + s + " " + p])
+        for stor in all_stor:
+            for pvc in self.get_resc_ns(self.get_cmd_output_now(stor)):
+                self.add_cmd_output(["%s get pvc -o yaml " % oc_cmd + pvc])
```
TurboTurtle: We already get pvcs in the kubernetes plugin.

sos/plugins/cnv.py
```diff
@@ -0,0 +1,234 @@
...
+        self.add_cmd_output("%s get vm --all-namespaces" % oc_cmd)
+        vmis = self.get_cmd_output_now("%s get vmi --all-namespaces" % oc_cmd)
+
+        if self.get_option("vm"):
+            options = self.get_option("vm")
+            if isinstance(options, bool):
+                return
```
TurboTurtle: So just to be clear,  unless a user explicitly runs `sosreport -k cnv.vm=foobar` we don't want to do _any_ vm/vmi related collection?

sos/plugins/cnv.py
```diff
@@ -0,0 +1,234 @@
...
+
+            subcmds_vm = [
+                "describe vm",
+                "describe vmi",
+                "get vm -o yaml",
+                "get vmi -o yaml",
```
TurboTurtle: I think these could (should) be added to the origin plugin, or the kubernetes plugin if these kinds of resources are available on the base project.

sos/plugins/cnv.py
```diff
@@ -0,0 +1,234 @@
...
+            ]
+            node_pods = [
```
TurboTurtle: Likewise, I think these should be added to either the kubernetes or origin plugins.

sos/plugins/cnv.py
```diff
@@ -0,0 +1,234 @@
...
+                            if vm.split()[0] in l:
+                                vmi_l = "-l " + l + " --all-namespaces -o wide"
+                p_ln = self.get_cmd_output_now("%s get pods " % oc_cmd + vmi_l)
+                for pod in self.get_resc_ns(p_ln):
+                    for s in subcmds_pods:
+                        self.add_cmd_output(["%s " % oc_cmd + s + " " + pod])
+                    exc = "%s " % oc_cmd + "exec " + pod + " -- virsh dumpxml "
+                    dmn = vm.split()[2] + "_" + vm.split()[0]
+                    self.add_cmd_output([exc + dmn])
+                node = self.get_node_info(p_ln)
```
TurboTurtle: This all seems a bit heavy for vm pod names, especially considering the kubernetes plugin already handles the collection of items that we're getting the pod names for later on around line 215.

Surely there's a more direct way of identifying vm pods?

sos/plugins/cnv.py
```diff
@@ -0,0 +1,234 @@
...
+                                    or line.startswith("Error") \
+                                    or line.startswith("error"):
+                                pass
+                            else:
+                                if line.split()[7] == node:
+                                    p = line.split()[1]
+                                    ns = " -n " + line.split()[0]
+                                    vm_pods.append(p + ns)
+                for pod in vm_pods:
+                    for s in subcmds_pods:
```
TurboTurtle: The kubernetes plugin already handles describe pods and pod logs via the `kubernetes.podlogs` option (disabled by default because it can get to be an absurd amount of output and we can't reliably obfuscate it all)

sos/plugins/cnv.py
```diff
@@ -0,0 +1,234 @@
...
+                        self.add_cmd_output(["%s " % oc_cmd + s + " " + pod])
+                    if pod.startswith("sdn"):
```
TurboTurtle: Again I think this is better served being in the origin plugin.

sos/plugins/cnv.py
```diff
@@ -0,0 +1,234 @@
...
+                    node_pod_ext = self.call_ext_prog(pod)
+                    if node_pod_ext['status'] == 0:
+                        for line in node_pod_ext['output'].splitlines():
+                            if line.startswith("NAMESPACE") \
+                                    or line.startswith("No resources found") \
+                                    or line.startswith("Error") \
+                                    or line.startswith("error"):
+                                pass
+                            else:
+                                if line.split()[7] == node:
```
TurboTurtle: Ok, now I'm a bit confused at this point. We seem to running multiple `get pods` commands for those with specific cnv-related labels, and trying to extract a pod (yes? node?) name and see if it matches the node name we got from a  generic `get pods` that is looking for specific labels? 

Can you provide an example of what exactly we're comparing against? Because right now I don't follow why we need to jump through all these hoops instead of just being able to say "get all the pod names that are cnv resources with XYZ label" like we're able to do with other kube/origin resources.

sos/plugins/cnv.py
TurboTurtle: Update: there is actually already a version collection in the kubernetes plugin. Does the `oc` command for version just wrap kubectl or does it give different/additional output? If the former there's no need to duplicate it, and if it's the latter it should still go into the origin plugin.

PR #1463, opened 2018-10-26
[utilities] Remove the __path__ attribute comparison with empty string

sourabhjains:
This patch removes an unnecessary comparison on package's __path__
attribute with an empty string in get_modules method.

When we pass an empty path to _find_plugins_from_list function it
returns None. The NoneType is not an iterable entity, so this
will leads to runtime error when we extend the plugins list with
None value.

Signed-off-by: Sourabh Jain <sourabhjain@linux.ibm.com>

---

### DIFF
--- a/sos/utilities.py
+++ b/sos/utilities.py
@@ -285,7 +285,7 @@ class ImporterHelper(object):
         package. """
         plugins = []
         for path in self.package.__path__:
-            if os.path.isdir(path) or path == '':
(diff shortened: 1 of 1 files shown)


sourabhjains:
Thanks, Bryn

- Sourabh Jain

sourabhjains:
Hello Bryn,

The pull request is still in open state.
Do you need any modification?

-Sourabh Jain

PR #1464, opened 2018-10-27
[openswan] Forbid collection of secrets and certdb

robbmanes:
When collecting IPSec data, currently nothing prevents the collection of
keys or the cert.db files.  Ensure we don't collect this information as
it is private.

Thanks to Robert Bost (@bostrt) for discovering this and Matt Rogers
(@mrogers950) for double checking what we should ignore.

---

### DIFF
--- a/sos/plugins/openswan.py
+++ b/sos/plugins/openswan.py
@@ -42,6 +42,10 @@ class Openswan(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
         if self.get_option("ipsec-barf"):
             self.add_cmd_output("ipsec barf")
 
-        self.add_forbidden_path("/etc/ipsec.d/key[3-4].db")
+        self.add_forbidden_path([
+            '/etc/ipsec.secrets',
+            '/etc/ipsec.secrets.d/*',
+            '/etc/ipsec.d/*.db',
+            '/etc/ipsec.d/*.secrets'])
 
 # vim: set et ts=4 sw=4 :


sos/plugins/openswan.py
```diff
@@ -42,6 +42,8 @@ def setup(self):
         if self.get_option("ipsec-barf"):
             self.add_cmd_output("ipsec barf")
 
-        self.add_forbidden_path("/etc/ipsec.d/key[3-4].db")
+        self.add_forbidden_path('/etc/ipsec.secrets')
+        self.add_forbidden_path('/etc/ipsec.d/*.db')
+        self.add_forbidden_path('/etc/ipsec.d/*.secrets')
```
mrogers950: You may also want to forbid /etc/ipsec.secrets.d/* for older installations.

robbmanes: Done, thanks!

bmr-cymru: These days the `add_forbidden_path()` interface accepts lists as well as strings (like `add_copy_spec()` and `add_cmd_output()`, so you can replace these with a single call:

```python
        self.add_forbidden_path([
            '/etc/ipsec.secrets',
            '/etc/ipsec.d/*.db',
            '/etc/ipsec.d/*.secrets'
        ])
```

sos/plugins/openswan.py
```diff
@@ -42,6 +42,9 @@ def setup(self):
         if self.get_option("ipsec-barf"):
             self.add_cmd_output("ipsec barf")
 
-        self.add_forbidden_path("/etc/ipsec.d/key[3-4].db")
```
bmr-cymru: Is this definitely no longer needed for any shipping versions? Happy to get rid of it if so, just want to be sure.

robbmanes: I personally don't know, but I assumed `self.add_forbidden_path('/etc/ipsec.d/*.db')` would take care of it, as we really shouldn't capture _any_ `.db` file.  Does the wildcard work here for us?

bmr-cymru: Eh.. never mind, I apparently had glob-blindness yesterday ;-)

robbmanes: :+1: No worries!

sos/plugins/openswan.py
robbmanes: :+1: TIL; done, thanks!

bmr-cymru: Thanks!

sos/plugins/openswan.py
```diff
@@ -42,6 +42,9 @@ def setup(self):
...
+        self.add_forbidden_path('/etc/ipsec.secrets',
```
TurboTurtle: Missing the `[]` in this call...:

```
  File "/home/jake/projects/sos/sos/sosreport.py", line 1033, in setup
    plug.setup()
  File "/home/jake/projects/sos/sos/plugins/openswan.py", line 42, in setup
    '/etc/ipsec.d/*.secrets')
TypeError: add_forbidden_path() takes exactly 2 arguments (5 given)
```

bmr-cymru: Aw jeez.. I have new glasses this week. I swear they have made me blind - since you need to respin to list-ise the members anyway, it would be better to put this in the `sos` list call style:

```python
        self.add_forbidden_path([
            "first_path",
            "second_path",
            "last_path"
        ])
```

robbmanes: > Missing the `[]` in this call...:
Sorry, tests didn't catch this.  Updated.

> Aw jeez.. I have new glasses this week. I swear they have made me blind - since you need to respin to list-ise the members anyway, it would be better to put this in the `sos` list call style:
Should be updated appropriately, let me know if I missed anything.  Thanks!

bmr-cymru: Thanks! And sorry for the fuzzy reviews this week :-)

bmr-cymru:
Merged in commit cbedcff40c44713c82c2e676d6dfcc8e502d3566

PR #1465, opened 2018-10-27
[ipsec] Remove IPSec plugin in favor of OpenSwan

robbmanes:
Red Hat distributions have not used the ipsec-tools and racoon packages
since RHEL5, and all of this information is collected in the OpenSwan
module via the same paths.

Similarly, Debian distributions have not provided this since wheezy, and
the same data is collected in OpenSwan.

---


robbmanes:
I very much would appreciate ack from @mrogers950 for removal of this plugin as well as a Debian maintainer to be absolutely certain this isn't something still used.  My research suggests no for Debian distributions but I am definitely not an expert personally.

mrogers950:
ipsec-tools is ancient and at least not supported by Red Hat. But it would be up to the sos maintainers if they want to keep legacy plugins around or support ones this old. I am personally ok with it being removed.

bmr-cymru:
We don't remove things that still have users - that includes distros that are shipping the package in new versions, as well as distros that are supporting an old distro version but that still receives updates to the `sos` package - for that reason we need a sign off from all the downstreams that could use it.

In this specific case though, that's only distributions that enable the `RedHatPlugin` tagging class:

```python
class ipsec(Plugin, RedHatPlugin):
    """ipsec related information
    """
```

So Debian et al aren't really relevant here: just Fedora, RHEL/Atomic/etc., CentOS, the two IBM policies, and the SuSE policy (that implemented `RedHatPlugin` as a bridge to enabling more SuSE-native plugins.

My guess is this would be fine for all of those but we should check before actually deleting it.

PR #1469, opened 2018-10-31
[kernel] Stop collecting char devices, instead move common to plugin

BryanQuigley:
Instead of creating char devices which is usually just
/dev/null symlinked from systemd- we can just check that a
couple key ones exist.

Creating these character devices makes it harder to manage
(delete) extracted sosreports without more permissions.

I can't envision the exact attack/problem but I think it's better
practice if sosreport doesn't create these special devices on
collecting systems.

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -68,11 +68,10 @@ def _path_in_path_list(path, path_list):
 
 def _node_type(st):
     """ return a string indicating the type of special node represented by
-    the stat buffer st (block, character, fifo, socket).
+    the stat buffer st (block, fifo, socket).
     """
     _types = [
         (stat.S_ISBLK, "block device"),
-        (stat.S_ISCHR, "character device"),
         (stat.S_ISFIFO, "named pipe"),
         (stat.S_ISSOCK, "socket")
     ]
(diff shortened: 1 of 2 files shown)


bmr-cymru:
It's a bit of a big hammer, and if we're removing it we should be at least grabbing a listing of the device permissions - file doesn't display those.

BryanQuigley:
We already capture the permissions in block.py.  
          "ls -lanR /dev",

Although I would be happy to move that to kernel if that makes more sense...
(and maybe drop -n and add -h)

bmr-cymru:
This pull request **introduces 1 alert** when merging 877c57e70f1742e2063ad39e500198aa45f25c89 into 22712dd4bf3ab25b68748cff955bb41fc4284ae1 - [view on LGTM.com](https://lgtm.com/projects/g/sosreport/sos/rev/pr-397e128d2490744ad57e375be9c73d1e09472a06)

**new alerts:**

* 1 for Implicit string concatenation in a list

---

BryanQuigley:
Oops, fixed

bmr-cymru:
> We already capture the permissions in block.py.
> "ls -lanR /dev",

True. My main objection here is the big hammer approach of crippling this at the archive level - I don't think that's necessary, and I haven't got any good reason for making this impossible. I think it's sufficient to remove the current collection, document it, and just not pick those specific entries up.

BryanQuigley:
Oh, ok.  When I started I didn't realize /dev/null was the only char device collected.. It was one line to drop it in #1595

PR #1471, opened 2018-11-05
[plugins] Decrease log verbosity when do_file_sub failed on missing file

pmoravec:
In case do_file_sub fails to apply a substitution on a non-existing file,
don't log it as an error log to stdout.

Resolves: #1471

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -322,8 +322,14 @@ class Plugin(object):
             else:
                 replacements = 0
         except Exception as e:
-            msg = "regex substitution failed for '%s' with: '%s'"
-            self._log_error(msg % (path, e))
+            # if trying to regexp a nonexisting file, dont log it as an
+            # error to stdout
+            if e.errno == errno.ENOENT:
+                msg = "file '%s' not collected, substitution skipped"
+                self._log_debug(msg % path)
+            else:
+                msg = "regex substitution failed for '%s' with: '%s'"
+                self._log_error(msg % (path, e))
             replacements = 0
         return replacements
 


sos/plugins/__init__.py
```diff
@@ -323,7 +323,12 @@ def do_file_sub(self, srcpath, regexp, subst):
                 replacements = 0
         except Exception as e:
             msg = "regex substitution failed for '%s' with: '%s'"
-            self._log_error(msg % (path, e))
+            # if trying to regexp a nonexisting file, dont log it as an
+            # error to stdout
+            if isinstance(e, FileNotFoundError):
```
bmr-cymru: Could it be better to test for `e.errno == errno.ENOENT`? That way we would still do the right thing if the class isn't actually an instance of `FileNotFound`. We're in standard library modules here, so perhaps we can already conclude it's guaranteed to only be this class but possibly worth checking (we use `errno` in a number of other places for this discrimination).

pmoravec: Sure, PR rebased.

sos/plugins/__init__.py
```diff
@@ -323,7 +323,12 @@ def do_file_sub(self, srcpath, regexp, subst):
                 replacements = 0
         except Exception as e:
             msg = "regex substitution failed for '%s' with: '%s'"
-            self._log_error(msg % (path, e))
+            # if trying to regexp a nonexisting file, dont log it as an
+            # error to stdout
+            if e.errno == errno.ENOENT:
```
bmr-cymru: I'd prefer the comment to say something like "Processing a file that was not collected is not an error" - I am not even sure we need an info log in that event. We already know it wasn't collected.

pmoravec: OK, verbosity decreased to debug and log statement rephrased.

bmr-cymru:
This pull request **introduces 1 alert** when merging 181a38720368b17834108e7e72f8d4c7f7adafdf into 031ff485afd888a5ecf9297bde2c2659cf3e1ec5 - [view on LGTM.com](https://lgtm.com/projects/g/sosreport/sos/rev/pr-f4b9fe040719936572fb1877991b05db1a66d2d2)

**new alerts:**

* 1 for Variable defined multiple times

---

PR #1472, opened 2018-11-05
[networking] Replace "brctl: by "bridge" commands

pmoravec:
As bridge-utils containing brctl command are being deprecated,
sosreport should call bridge command instead.

Although the mapping of the commands is not 1:1, the data collected
(together with few "ip .." commands) will remain the same.

Resolves: #1472

---

### DIFF
--- a/sos/plugins/networking.py
+++ b/sos/plugins/networking.py
@@ -25,24 +25,6 @@ class Networking(Plugin):
     # switch to enable netstat "wide" (non-truncated) output mode
     ns_wide = "-W"
 
-    def get_bridge_name(self, brctl_file):
-        """Return a list for which items are bridge name according to the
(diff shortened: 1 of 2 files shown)


pmoravec:
Two comments for possible concerns:

- `bridge` commands being called do not seem to load `bridge` kernel module, contrary to `brctl` command. So there should not be a need for `if self.is_module_loaded("bridge"):` test.
- The patch collects forwarding DB entries per all bridges once, while originally `brctl showmacs <bridge>` collected it per device. As the one output file contains all such outputs per device, I dont see a big benefit in calling the command per device.

bmr-cymru:
> I dont see a big benefit in calling the command per device.

Agreed. Since the format of the data is going to change anyway there seems little point in slavishly preserving aspects of the old behaviour.

PR #1473, opened 2018-11-06
[podman] allow the plugin for RedHatPlugin and UbuntuPlugin

pmoravec:
Until Podman inherits RedHatPlugin and/or UbuntuPlugin, the plugin
can not be executed on underlying distros.

Further, remove one redundant test as "for container in insp" will
work properly also for empty "insp".

Resolves: #1473

---

### DIFF
--- a/sos/plugins/podman.py
+++ b/sos/plugins/podman.py
@@ -11,7 +11,7 @@
 from sos.plugins import Plugin, RedHatPlugin, UbuntuPlugin
 
 
-class Podman(Plugin):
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

pmoravec:
FYI it is possible that the original plugin worked well at the time of writing the PR and sosreport has some regression else-where. I just tested it now on current upstream with this result (`sosreport -e podman` fails as the plugin isnt found, `sosreport -l` does not list it).

PR #1474, opened 2018-11-09
[etcd] Include journal

juanluisvaladas:
---


juanluisvaladas:
Friendly bump, if I understand correctly the contributing process, we need a second reviewer for this case, could I get someone else to review it? Thanks

bmr-cymru:
Nope. Not needed for something so trivial. We only require multiple acks from core reviewers when things affect more than one downstream, or where there are conflicts among different interest groups.

Issue #1475, opened 2018-11-09
[juju] juju plugin doesn't detect juju if installed via snap


slashdd:
juju plugin doesn't detect if juju is installed via snap. It only detect the files which come with conventional .deb package

```
class Juju(Plugin, UbuntuPlugin):
    """ Juju orchestration tool
    """

    plugin_name = 'juju'
    profiles = ('virt', 'sysmgmt')
    files = ('/usr/bin/juju', '/usr/bin/juju-run')
```
The juju binary of snap juju is found in : `/snap/bin/juju`

Eventually, it would be great to attempt to add 'snap' as in the package manager just like rpm, deb but for now having a check for `/snap/bin/juju` will suffice.

slashdd:
Proposal fix: 
```
    plugin_name = 'juju'
    profiles = ('virt', 'sysmgmt')
    files = ('/usr/bin/juju', '/usr/bin/juju-run', '/snap/bin/juju')
```

bmr-cymru:
We don't have a mechanism for this right now (nor for related things like Flatpak). I would rather see proposals for addressing that than making lots of plugin-specific changes that would need to be undone or modified when we do have that support.

Anyway, so long as the snap binary would be in `PATH` there is no need for these hard-coded paths. Just add a `Juju.commands` member that includes the `juju` executable name. The plugin will then be enabled whenever that command was present. Currently this may need policy modification depending on how a policy initialises the `PATH` environment - for Red Hat, it would need to be added to the fixed path list. Again, a better mechanism for doing this would be welcome, e.g. a generic policy method to probe for these types of package manager and extend `PATH` appropriately, but in the meantime this approach would keep the scope of the changes to more of a minimum.

BryanQuigley:
FYI, Adding /snap/bin to path for Debian/Ubuntu in #1488

PR #1476, opened 2018-11-09
[juju] Add detection when juju is installed as a snap.

slashdd:
Fixes Issue #1475

Conventional DEB package extract juju binary in /usr/bin/juju,
but SNAP extract juju binary in /snap/bin/juju.

---


bmr-cymru:
I'd rather we explore a more generic method for this than adding fixed paths to potentially many plugins. See my comments in #1475 for some thoughts on how that could work.

PR #1477, opened 2018-11-10
[blueprints] New plugin for lorax-composer

pmoravec:
lorax-composer is an API server for building disk images using
Blueprints. The plugin collects composer config and logs and few
composer-cli command outputs.

Resolves: #1477

---

### DIFF
--- a/sos/plugins/composer.py
+++ b/sos/plugins/composer.py
@@ -1,0 +1,39 @@
+from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+
+
+class Composer(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
+    """Lorax Composer
+    """
+
+    plugin_name = 'composer'
+    profiles = ('sysmgmt', 'virt', )
+
+    packages = ('composer-cli',)
+
+    def _get_blueprints(self):
+        blueprints = []
+        bp_result = self.get_command_output("composer-cli blueprints list")
+        if bp_result['status'] != 0:
+            return blueprints
+        for line in bp_result['output'].splitlines():
+            blueprints.append(line)
+        return blueprints
+
+    def setup(self):
+        self.add_copy_spec([
+            "/etc/lorax/composer.conf",
+            "/var/log/lorax-composer/composer.log"
+            "/var/log/lorax-composer/dnf.log"
+            "/var/log/lorax-composer/program.log"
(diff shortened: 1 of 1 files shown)


pmoravec:
Thanks Chris Williams from Red Hat for the initial patch proposal.

sos/plugins/blueprints.py
```diff
@@ -0,0 +1,41 @@
...
+            "/var/log/lorax-composer/composer.log"
+            "/var/log/lorax-composer/dnf.log"
+            "/var/log/lorax-composer/program.log"
+            "/var/log/lorax-composer/server.log"
+        ])
+        blueprints = self._get_blueprints()
+        for blueprint in blueprints:
+            self.add_cmd_output([
+                "composer-cli blueprints show %s" % blueprint
+            ])
```
bmr-cymru: No list for a single command.

sos/plugins/blueprints.py
```diff
@@ -0,0 +1,41 @@
+from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+
+
+class Blueprints(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
+    """Composer Blueprints
```
bmr-cymru: It seems a little weird that nothing in here is called "blueprints", apart from the plugin and the sub-command we are running?

Why not "composer", "lorax", or some variant on that? Would we ever want to add any more non-blueprint commands here? Blueprint is a pretty generic name and it would be much easier to justify if the package names or upstream project included it.

sos/plugins/blueprints.py
```diff
@@ -0,0 +1,41 @@
...
+    """
+
+    plugin_name = 'blueprints'
+    profiles = ('network', 'hardware', )
```
bmr-cymru: network? hardware? I thought this was a thing for composing virt images?

sos/plugins/blueprints.py
```diff
@@ -0,0 +1,41 @@
...
+
+    packages = ('composer-cli',)
```
bmr-cymru:
```python
        class ComposerCli(Plugin):
```

? ;-)

pmoravec:
Thanks for comments, incorporated.

bmr-cymru:
I've added another commit on top to avoid running the `blueprints list` command twice, but I'm not able to test locally as I don't have those components installed.

PR #1481, opened 2018-11-15
[openstack_instack] Support for multiple stacks

jtaleric:
The OpenStack undercloud can support greater than 1 stack. 
This patch will iterate through all the possible stacks deployed
on the undercloud.

Signed-off-by: Joe Talerico <rook@redhat.com>

---

### DIFF
--- a/sos/plugins/openstack_instack.py
+++ b/sos/plugins/openstack_instack.py
@@ -11,6 +11,7 @@
 
 from sos.plugins import Plugin, RedHatPlugin
 import os
+import re
 
 
 NON_CONTAINERIZED_DEPLOY = [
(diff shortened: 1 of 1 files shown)


bmr-cymru:
Please also follow the [contributor guidelines](https://github.com/sosreport/sos/wiki/Contribution-Guidelines) for patch formatting: the subject needs to be in `[<component>] change description` format, and all contributions require a DCO ("Signed-off-by") in the commit body.

jtaleric:
@bmr-cymru Thanks - Let me know if there is anything else missing.

jtaleric:
@bmr-cymru Any further feedback?

sos/plugins/openstack_instack.py
```diff
@@ -62,29 +63,34 @@ def setup(self):
                                 "the environment file for the user intended "
                                 "to connect to the OpenStack environment.")
         else:
+            # capture all the possible stack uuids
+            get_stacks = "openstack stack list"
+            stacks = self.call_ext_prog(get_stacks, timeout=600)['output']
```
TurboTurtle: If this is set to `timeout=600` and can be expected to ever take that long, we'll need to adjust the plugin's timeout. Otherwise, the timeout of 600 seconds is longer than the default timeout for the whole plugin (which is 300).

sos/plugins/openstack_instack.py
```diff
@@ -62,29 +63,34 @@ def setup(self):
...
-                            timeout=600)
+            for sid in stack_ids:
+                self.add_cmd_output("openstack stack show %s" % sid[0])
+                self.add_cmd_output(
+                        "openstack stack resource list -n 10 %s" % sid[0],
+                        timeout=600)
+
+                # get details on failed deployments
+                cmd = "openstack stack resource list -f value -n 5 %s" % sid[0]
+                deployments = self.call_ext_prog(cmd, timeout=600)['output']
```
TurboTurtle: Same here, with the timeout.

jtaleric:
@TurboTurtle I don't mind changing it to 300sec, I didn't know if there was an original reason for the 600...

TurboTurtle:
That should be fine - we just need to adjust the timeout in the right spot. If that command isn't known to need up to 10 minutes to run, then we should be alright adjusting the command timeout down.

jtaleric:
@TurboTurtle alright, let me know if there are any other suggestions! Thanks!

bmr-cymru:
There are a few style fixes needed here - calling `add_cmd_output()` with a list rather than multiple times in succession, and the indentation of the innermost block in the `sid in stack_ids` loop is doubled (and can be rewritten to reduce the indentation), but I'll go ahead and merge this today with those changes.

bmr-cymru:
Merged with two extra commits - please test if you have the opportunity.

PR #1482, opened 2018-11-16
(WIP) [services] Make it work in debian and condense differences

BryanQuigley:
WIP because I'm unsure which is the correct way to write a plugin that only needs changes for one distro.

Adds other distros to initial class
azure.py, cgroups.py

Creates "empty" classes:
corosync.py

In my testing, doing it the initial class way ends up creating a chkconfig file (so I guess the RH plugin runs?)

[services] Make it work in debian and condense differences
The debian/ubuntu plugin's commands startin with /sbin/ have not been
run in any supported releases (and they have not been missed).

Systemd systems capture an equivalent of "service --status-all"
(although this will double it up) I don't believe it needs to be
begind a flag

---

### DIFF
--- a/sos/plugins/services.py
+++ b/sos/plugins/services.py
@@ -9,26 +9,23 @@
 from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
 
 
-class Services(Plugin):
+class Services(Plugin, DebianPlugin, UbuntuPlugin):
     """System services
     """
 
     plugin_name = "services"
     profiles = ('system', 'boot')
 
-    option_list = [("servicestatus", "get a status of all running services",
-                    "slow", False)]
-
     def setup(self):
         self.add_copy_spec([
             "/etc/inittab",
-            "/etc/rc.d"
+            "/etc/rc.d",
+            "/etc/rc*.d"
         ])
-        if self.get_option('servicestatus'):
-            self.add_cmd_output("/sbin/service --status-all")
         self.add_cmd_output([
             "/sbin/runlevel",
-            "ls /var/lock/subsys"
+            "ls /var/lock/subsys",
+            "service --status-all"
         ])
 
 
@@ -38,16 +35,5 @@ class RedHatServices(Services, RedHatPlugin):
         super(RedHatServices, self).setup()
         self.add_cmd_output("/sbin/chkconfig --list", root_symlink="chkconfig")
 
-
-class DebianServices(Services, DebianPlugin, UbuntuPlugin):
-
-    def setup(self):
-        super(DebianServices, self).setup()
-        self.add_copy_spec("/etc/rc*.d")
-
-        self.add_cmd_output("/sbin/initctl show-config",
-                            root_symlink="initctl")
-        if self.get_option('servicestatus'):
-            self.add_cmd_output("/sbin/initctl list")
 
 # vim: set et ts=4 sw=4 :


sos/plugins/services.py
```diff
@@ -9,26 +9,23 @@
...
+class Services(Plugin, DebianPlugin, UbuntuPlugin):
     """System services
     """
 
     plugin_name = "services"
     profiles = ('system', 'boot')
 
-    option_list = [("servicestatus", "get a status of all running services",
-                    "slow", False)]
-
```
bmr-cymru: Didn's see any rationale for dropping the option here? In some ways, it's moot, since most distros have moved to Systemd anyway and use that plugin for this information but we shouldn't remove options without good reason (they break automation scripts that assume the option exists).

sos/plugins/services.py
```diff
@@ -9,26 +9,23 @@
...
     def setup(self):
         self.add_copy_spec([
             "/etc/inittab",
-            "/etc/rc.d"
+            "/etc/rc.d",
+            "/etc/rc*.d"
```
bmr-cymru: What's the structure on Debian? On Red Hat SysV we have `/etc/rc.d/`, and all the `/etc/rcX.d` are simply symlinks into the correspondingly named directory in `rc.d`.

```
# ll /etc/rc*.d -d
lrwxrwxrwx.  1 root root   10 May  6  2016 /etc/rc0.d -> rc.d/rc0.d
lrwxrwxrwx.  1 root root   10 May  6  2016 /etc/rc1.d -> rc.d/rc1.d
lrwxrwxrwx.  1 root root   10 May  6  2016 /etc/rc2.d -> rc.d/rc2.d
lrwxrwxrwx.  1 root root   10 May  6  2016 /etc/rc3.d -> rc.d/rc3.d
lrwxrwxrwx.  1 root root   10 May  6  2016 /etc/rc4.d -> rc.d/rc4.d
lrwxrwxrwx.  1 root root   10 May  6  2016 /etc/rc5.d -> rc.d/rc5.d
lrwxrwxrwx.  1 root root   10 May  6  2016 /etc/rc6.d -> rc.d/rc6.d
drwxr-xr-x. 10 root root 4096 Apr 28  2017 /etc/rc.d
```

sos/plugins/services.py
```diff
@@ -39,15 +36,4 @@ def setup(self):
...
-
-    def setup(self):
-        super(DebianServices, self).setup()
-        self.add_copy_spec("/etc/rc*.d")
-
-        self.add_cmd_output("/sbin/initctl show-config",
-                            root_symlink="initctl")
-        if self.get_option('servicestatus'):
-            self.add_cmd_output("/sbin/initctl list")
-
```
bmr-cymru: What was actually broken for Debian here? It's not clear from the patch description.

bmr-cymru:
If you need changes in one distro, and the main class is already appropriate for **all** distros then you just need to create a new derived class for the one distro with changes.

If the main class is **not** already suitable for all classes, then you need to create two or more subclasses, and move anything inappropriate from the parent class to the corresponding subclass.

sos/plugins/services.py
BryanQuigley: Also symlinks, but not the directories;
$ ll /etc/rc*.d
/etc/rc0.d:
total 16
drwxr-xr-x   2 root root  4096 Nov 23 13:50 ./
drwxr-xr-x 149 root root 12288 Nov 22 09:55 ../
lrwxrwxrwx   1 root root    20 Apr 14  2018 K01alsa-utils -> ../init.d/alsa-utils*
lrwxrwxrwx   1 root root    13 Apr 19  2018 K01atd -> ../init.d/atd*
lrwxrwxrwx   1 root root    22 Apr 14  2018 K01avahi-daemon -> ../init.d/avahi-daemon*

I figure there's no harm in making sure whatever symliks should be there are there on all distros?

sos/plugins/services.py
BryanQuigley: Oops, nothing - made an assumption based on the /sbin/service 
- having said that this seems like it duplicates what is in upstart.py

sos/plugins/services.py
BryanQuigley: We default to collecting the equivalent for systems without putting it behind a flag.

BryanQuigley:
Thanks for the comments!  Started from scatch in #1484 just making the smaller changes I really wanted. 

Wasn't thinking dropping the option would break anything, so it can stay.
rc0*.d doesn't seem particularly useful on Fedora/RH so leaving that off as well

PR #1485, opened 2018-11-27
[policies] Fix default xz version

pieqq:
Hello! I've ran into an issue (described in the commit message below) and am proposing a patch for it. It's my first contribution to sosreport, so I hope I'm doing it right. Please let me know if anything needs to be amended or adjusted.

Thanks!

---
While running some tests using sosreport in a snap, I ran into the
following issue:

[archive:TarFileArchive] An error occurred compressing the archive:
unorderable types: int() >= str()

This is because the `xz_version` variable might be set to a list
containing one integer, but it was later compared to a list of strings,
raising an exception.

---

### DIFF
--- a/sos/policies/__init__.py
+++ b/sos/policies/__init__.py
@@ -537,7 +537,7 @@ No changes will be made to system configuration.
                 xz_version = self.package_manager\
                                  .all_pkgs()[xz_package]["version"]
             except Exception as e:
-                xz_version = [0]  # deal like xz version is really old
+                xz_version = [u'0']  # deal like xz version is really old
             if xz_version >= [u'5', u'2']:
                 cmd = "%s -T%d" % (cmd, threads)
         return cmd


bmr-cymru:
The patch looks fine for now - thanks. One question: do you know what the exception was that triggered execution of the broken block?

We're having a few discussions around how to support `sos` in snaps at the moment, and this might be useful information (one issue is the handling of the package manager data in this environment).

pieqq:
I'm working on [a little tool](https://launchpad.net/qabro) packaged as a snap that uses `sosreport` to generate a report and attach it to a Launchpad issue.

Yesterday, while generating a new snap, I discovered that v3.6 of `sosreport` had been pushed in the Ubuntu repositories (until now it had been v3.5.1). When I did a test run, I saw the issue mentioned above, and after searching in sosreport source code I found the culprit.

I didn't investigate deeper because I found out `sosreport` has a `-z` parameter to choose the desired compression format. I tested it with `-z gzip` and it worked as expected, so I've just modified my own code to generate a `.tar.gz` archive. In particular, I didn't check why `sosreport` decided to go with `xz` compression system even though `xz` is not available inside my snap. (That said, a simple solution might be to simply add `xz-utils` to my snap...)

PR #1486, opened 2018-11-28
[firewalld] Fetch "direct" rule details as well as overall state and …

ptalbert:
…logging setting

Direct Rules are an increasingly common way to control the firewall, but
current sos plugin output does not get these.

Update the plugin to grab firewall-cmd output about Direct Rules, and
also pick up the 'state' output and newer logging settings.

---

### DIFF
--- a/sos/plugins/firewalld.py
+++ b/sos/plugins/firewalld.py
@@ -36,7 +36,15 @@ class FirewallD(Plugin, RedHatPlugin):
         # docker containers.
         self.add_cmd_output([
             "firewall-cmd --list-all-zones",
-            "firewall-cmd --permanent --list-all-zones"
+            "firewall-cmd --direct --get-all-chains",
+            "firewall-cmd --direct --get-all-rules",
+            "firewall-cmd --direct --get-all-passthroughs",
+            "firewall-cmd --permanent --list-all-zones",
+            "firewall-cmd --permanent --direct --get-all-chains",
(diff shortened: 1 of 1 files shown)


sos/plugins/firewalld.py
```diff
@@ -36,7 +36,15 @@ def setup(self):
...
             "firewall-cmd --list-all-zones",
-            "firewall-cmd --permanent --list-all-zones"
+            "firewall-cmd --direct --get-all-chains",
+            "firewall-cmd --direct --get-all-rules",
+            "firewall-cmd --direct --get-all-passthroughs",
+            "firewall-cmd --permanent --list-all-zones",
+            "firewall-cmd --permanent --direct --get-all-chains",
+            "firewall-cmd --permanent --direct --get-all-rules",
+            "firewall-cmd --permanent --direct --get-all-passthroughs",
+	    "firewall-cmd --state",
```
TurboTurtle: Tab alignment needs fixing and is why the Travis tests failed.

ptalbert:
```suggestion
            "firewall-cmd --state",
```

sos/plugins/firewalld.py
```diff
@@ -36,7 +36,15 @@ def setup(self):
...
+	    "firewall-cmd --get-log-denied"
```
ptalbert:
```suggestion
            "firewall-cmd --get-log-denied"
```

PR #1487, opened 2018-11-28
[podman|docker] Add postprocessing for container inspect output

TurboTurtle:
Adds a postproc for the podman and docker plugins to attempt to
obfuscate sensitive keys in 'inspect' output for those runtimes.
Previously, these keys were being captured in plaintext which could lead
to passwords or similar being leaked when sysadmins configure containers
with environment variables (or similar) that contain this data.

Specifically, we match against 'key=value' pairs as that is how the
container runtimes accept and print these pairs, like so:

            "Env": [
                "mypassword=supersecret",
                "container=oci"
            ],

By comparison, the inspect outputs now read like the following when a
potentially sensitive key is found:

            "Env": [
                "mypassword=********",
                "container=oci"
            ],

---

### DIFF
--- a/sos/plugins/docker.py
+++ b/sos/plugins/docker.py
@@ -83,6 +83,23 @@ class Docker(Plugin):
                 if self.get_option('logs'):
                     self.add_cmd_output("docker logs -t %s" % container)
 
+    def postproc(self):
+        # Attempts to match key=value pairs inside container inspect output
+        # for potentially sensitive items like env vars that contain passwords.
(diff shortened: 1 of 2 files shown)


TurboTurtle:
A couple notes:

1. Both `podman` and `docker` format their respective `inspect $thing` output the same way, and both accept env vars and such in the same manner, so the postproc works for both of them

2. I am injecting the ending `"` character in the substitution so that we don't break anything that may be scanning the `inspect` output files.

PR #1488, opened 2018-11-28
[debian/lxd] Add /snap/bin and have lxd detection use it

BryanQuigley:
Adds /snap/bin to Debian policy path
Have lxd plugin use it for detecting when it should run

---

### DIFF
--- a/sos/plugins/lxd.py
+++ b/sos/plugins/lxd.py
@@ -17,6 +17,7 @@ class LXD(Plugin, UbuntuPlugin):
     plugin_name = 'lxd'
     profiles = ('container',)
     packages = ('lxd',)
+    commands = ('lxc', 'lxd')
 
     def setup(self):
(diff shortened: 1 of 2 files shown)


bmr-cymru:
I really think we need a more coherent strategy for dealing with `snap` than just adding things piecemeal to individual plugins. This does not feel maintainable and I think we are making more work for ourselves in the future by sweeping it under the carpet.

As a start I'd prefer to have the policy change and the plugin change as separate patches. It's less confusing when looking at the history and it means you can explain the difference between `lxc` and `lxd` in the commit message (which isn't at all obvious from the current patch).

We also need to deal with the package information side of things - I am worried that if we don't do that soon, we will start using command presence in `$PATH` as a workaround and that will leave a lot of change to undo later on when we do have this in the `PackageManager` class (or elsewhere).

PR #1489, opened 2018-11-29
[subscription_manager] Add syspurpose collection

TurboTurtle:
Adds syspurpose output collection which is a new tool to assist with
subscription manager on RHEL systems.

---

### DIFF
--- a/sos/plugins/subscription_manager.py
+++ b/sos/plugins/subscription_manager.py
@@ -7,6 +7,7 @@
 # See the LICENSE file in the source distribution for further information.
 
 from sos.plugins import Plugin, RedHatPlugin
+import glob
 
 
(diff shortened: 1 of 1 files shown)


bmr-cymru:
What is the syspurpose of this? ;-)

TurboTurtle:
I've updated the commit to also include adding `rct` output instead of making another separate PR for that.

syspurpose replaces variants for RHEL 8 - so instead of "RHEL 8 Workstation" and "RHEL 8 Server", there's just "RHEL 8" with syspurpose used to set the.... purpose.

bmr-cymru:
> just "RHEL 8" with syspurpose used to set the.... purpose.

∃ purpose "pass the butter"?

PR #1490, opened 2018-11-29
Collect additional sysfs attribute data from block/<dev>/queue, scsi_host, and fc_vports

jtpittman195:
---

In order to aide in efficient troubleshooting, gather a bit more data from sysfs in relation to block devices queue directory, the scsi_host subsystem, and the fc_vports directory.  All data is gathered through the add_udev_info method.  These changes are an attempt at resolution for Red Hat bz1090485.

### DIFF
--- a/sos/plugins/block.py
+++ b/sos/plugins/block.py
@@ -51,8 +51,7 @@ class Block(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
             "/etc/blkid.tab",
             "/run/blkid/blkid.tab",
             "/proc/partitions",
-            "/proc/diskstats",
-            "/sys/block/*/queue/scheduler"
+            "/proc/diskstats"
         ])
 
         if os.path.isdir("/sys/block"):
@@ -60,8 +59,10 @@ class Block(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
                 if disk.startswith("ram"):
                     continue
                 disk_path = os.path.join('/dev/', disk)
+                queue_path = os.path.join('/sys/block/', disk, 'queue')
                 self.add_udev_info(disk_path)
                 self.add_udev_info(disk_path, attrs=True)
+                self.add_udev_info(queue_path, attrs=True)
                 self.add_cmd_output([
                     "parted -s %s unit s print" % (disk_path),
                     "fdisk -l %s" % disk_path
--- a/sos/plugins/scsi.py
+++ b/sos/plugins/scsi.py
@@ -6,6 +6,7 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
+import os
 from sos.plugins import Plugin, RedHatPlugin, UbuntuPlugin, DebianPlugin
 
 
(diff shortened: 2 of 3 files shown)


sos/plugins/scsi.py
```diff
@@ -32,4 +33,9 @@ def setup(self):
             "sg_map -x"
         ])
 
+        if os.path.isdir("/sys/class/scsi_host"):
+            for host in os.listdir("/sys/class/scsi_host"):
+                host_path = os.path.join('/sys/class/scsi_host/', host)
+                self.add_udev_info(host_path, attrs=True)
+
```
bmr-cymru: I think this would look better using a glob: that way, you can get rid of the explicit presene test, and you don't have to do any path mangling on the directory content. We can also lose the explicit list walk since the `add_udev_info()` method supports list arguments:

```python
    hosts = glob.glob("/sys/class/scsi_host/*")
    self.add_udev_unfo(hosts, attrs=True)
```

sos/plugins/block.py
```diff
@@ -51,17 +51,18 @@ def setup(self):
...
 
         if os.path.isdir("/sys/block"):
             for disk in os.listdir("/sys/block"):
                 if disk.startswith("ram"):
                     continue
                 disk_path = os.path.join('/dev/', disk)
+                queue_path = os.path.join('/sys/block/', disk, 'queue')
                 self.add_udev_info(disk_path)
                 self.add_udev_info(disk_path, attrs=True)
+                self.add_udev_info(queue_path, attrs=True)
```
bmr-cymru: This code is a bit old and crusty: looking at it, I can see where the pattern in the `scsi.py` patch originated. I think a bit of clean up here has been overdue for a while & we can get that done while incorporating these changes. From a quick look I think we can use a similar approach to my suggestion for the other commit (I'm also not really sure we need the `disk_path` udev info twice - `attrs=True` should only add extra information, but perhaps there was a historical reason that people liked the "simpler" data format).

Anyway, either way: the changes are reasonable and we will get them queued up for 3.7 one way or another, but it would be good to address some of the sprawl that has crept into the block plugin over the years as well.

TurboTurtle: Just to chime in here - yes we're grabbing `disk_path` twice because RH support engineers wanted to keep the 'simpler' output as well when we added `add_udev_info()`.

bmr-cymru: That's what I suspected; it makes my kinda sad, but hey ho ;-)

If anything the `-a` output is the more human-readable, as well as including data for the entire device chain: it includes comments, and has whitespace structure designed to make it easy-to-read. In contrast, without `-a` you just get a list of name/value pairs tagged with P/N/S/E etc. indicating the type (physical, name, symlink, env var etc.).

Unfortunately since there isn't any canonical non-file system representation for sysfs data (and since `sysfsutils` is canonically awful), there isn't much we can do.

jtpittman195:
Bryn, thanks a lot for the review and acceptance; we appreciate it.  These values will help us a great deal in complex fabrics and perf cases.  If it's ok, we'll leave the optimization edits to you and Jake as it's a bit outside of our wheelhouse.  Thanks again.

sos/plugins/block.py
jtpittman195: Bryn, Jake, just wanted to update; sorry for the wait.  Upstream, it's not looking good for running udevadm against the queue directory (https://github.com/systemd/systemd/issues/11183).  I've asked Michal for more info in the RH bz (https://bugzilla.redhat.com/show_bug.cgi?id=1657476), but I feel we'll end up having to cp the contents manually.

jtpittman195: Thanks Bryn, appreciate it. Would guys prefer I rebase/amend, close and send a v2, or just let you guys edit by hand?

PR #1491, opened 2018-11-29
[cockpit] Add new plugin

TurboTurtle:
Adds a new plugin to collect information about Cockpit, a browser-based
system configuration and management tool.

---


BryanQuigley:
Looks good - cockpit is available on Debian/Ubuntu as well.  Same package names.

TurboTurtle:
Updated to include Debian and Ubuntu.

Issue #1446, opened 2018-10-08
[sosreport,utilities] sosreport can hang on timeouted plugin


pmoravec:
In particular scenarios, sosreport can hang infinitely when waiting to shutdown a thread for timeouted plugin, with backtrace (of the only "running" process that time):

    /usr/lib64/python2.7/threading.py, line 362, in wait
    /usr/lib64/python2.7/threading.py, line 954, in join
    /usr/lib/python2.7/site-packages/concurrent/futures/thread.py, line 148, in shutdown
    /root/sos-master/sos/sosreport.py, line 1077, in collect

We must react on timeouted plugin in better way.

Particular reproducer: https://github.com/sosreport/sos/pull/1445#issuecomment-427761038 (optionally decrease plugin's timeout in https://github.com/sosreport/sos/blob/master/sos/plugins/__init__.py#L130 to lower value to see the problem faster).

pmoravec:
@TurboTurtle , any idea?

TurboTurtle:
IIRC this (only?) happens when a command we're collecting enters D-state, right?

I'm not immediately aware of a way to force-kill a thread in that state, but I want to properly scope what we're looking at here. Are there any other scenarios you've found where this could happen?

TurboTurtle:
As mentioned in PR #1492, I found that this was because `shutdown()` only cancels pending jobs, not running ones. To cancel running jobs we need to use `_threads.clear()` on the threadpool.

PR #1492, opened 2018-11-29
[sosreport] Clear nested threadpool threads on timeout

TurboTurtle:
We nest ThreadPoolExecutors to run plugins to enable whole-plugin
timeouts, with the timeout being handled by the nested threadpool.
However, calling shutdown() on the nested thread is incorrect in the
event of a plugin timeout, as it only cancels pending jobs, not
currently running jobs that the timed-out plugin is.

Instead, call _threads.clear() to forcibly cancel the running job, thus
allowing the thread to return to the higher level threadpool and allow
other plugins to run via that thread.

This still allows data captured up to the point of the timeout to be
written to the sosreport.

Resolves: #1446

---

### DIFF
--- a/sos/sosreport.py
+++ b/sos/sosreport.py
@@ -1096,7 +1096,7 @@ class SoSReport(object):
             except TimeoutError:
                 self.ui_log.error("\n Plugin %s timed out\n" % plugin[1])
                 self.running_plugs.remove(plugin[1])
-                pool.shutdown(wait=False)
+                pool._threads.clear()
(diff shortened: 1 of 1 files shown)


TurboTurtle:
Note: when testing and artificially causing timeouts, using an extremely low value  (e.g. 5 seconds) can cause exceptions to be generate when trying to write to the archive. I did not experience any such write-exceptions when the timeout value was 10 seconds or higher.

PR #1493, opened 2018-11-29
[sosreport] Allow users to set timeout for plugins

TurboTurtle:
Previously, plugin timeouts were only able to be set in the plugin
itself, and could not be controlled by users. Now, a new timeout option
has been added to allow this. If a user specifies a timeout, then that
timeout applies to all plugins being run. Setting a timeout value of 0
(zero) means that no timeout will be applied.

---

### DIFF
--- a/man/en/sosreport.1
+++ b/man/en/sosreport.1
@@ -13,6 +13,7 @@ sosreport \- Collect and package diagnostic and support data
           [--batch] [--build] [--debug]\fR
           [--label label] [--case-id id] [--ticket-number nr]
           [--threads threads]
+          [--timeout timeout]\fR
           [-s|--sysroot SYSROOT]\fR
           [-c|--chroot {auto|always|never}\fR
           [--tmp-dir directory]\fR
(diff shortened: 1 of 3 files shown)


pmoravec:
Nice; would it be possible to extend the PR to set the timeout option independently to each plugin? I.e. if I want to use default timeout for most plugins, but want to increase the plugin timeout just for one plugin that I know will take much time..? As when I try now:

    # sosreport -o java --batch --build -k java.timeout=500
    ..
    no such option "timeout" for plugin (java)
    #

bmr-cymru:
> Nice; would it be possible to extend the PR to set the timeout option independently to each plugin?

This is something of a requirement I think - we discussed this in the past, I think even as far as possible command line syntax.

We also shouldn't use `--timeout` for this unless the syntax is going to allow control of *all* timeouts: command, plugin and possibly others in the future (a whole run watchdog or file IO timeouts for e.g.). If it's just going to control plugins I think `--plugin-timeout` is better. This could either use a parsed syntax to set global and per-plugin timeouts separately, or we could use the notation @pmoravec shows and do some internal mapping to a (default) plugin option.

TurboTurtle:
I had thought about do per-plugin timeout control, but my reasoning for this approach is that it's A) unlikely that a user will run into a situation where they actually set separate timeouts - the most likely scenario is that a single timeout would be set on a single failing plugin and B) if a user is setting a timeout for a specific plugin due to that plugin timing out, then I don't see a scenario where a user is actually going to say "I'm willing to wait 10 minutes for this logs plugin to run, but I can't wait longer than usual for the other plugins" - especially now that we run concurrent plugins. 

I like the `--plugin-timeout` option name change. Also intrigued by putting timeout in the plugin options list, not sure which would be better in the end.

TurboTurtle:
Closing this in favor of #1499 which provides both the `-k` syntax and a global option.

PR #1495, opened 2018-12-01
[stratis] Add new plugin

TurboTurtle:
Adds a new plugin for the stratis storage manager

---

### DIFF
--- a/sos/plugins/stratis.py
+++ b/sos/plugins/stratis.py
@@ -1,0 +1,33 @@
+# Copyright (C) 2018 Red Hat, Inc., Jake Hunsaker <jhunsake@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.plugins import Plugin, RedHatPlugin
+
+
+class Stratis(Plugin, RedHatPlugin):
+    '''Stratis Storage'''
+
+    packages = ('stratis-cli', 'stratisd')
+    services = ('stratisd',)
+    profiles = ('storage',)
+
+    def setup(self):
+        subcmds = [
+            'pool list',
+            'filesystem list',
+            'blockdev list',
+            'daemon redundancy',
+            'daemon version'
+        ]
+
+        self.add_cmd_output(["stratis %s" % subcmd for subcmd in subcmds])
+        self.add_journal('stratisd')
+
+# vim: set et ts=4 sw=4 :


pmoravec:
Could we collect also for each pool

    stratis blockdev list <pool>
    stratis filesystem list <pool>

?

Here is the updated (and tested) code:

    class Stratis(Plugin, RedHatPlugin):
        '''Stratis Storage'''
    
        packages = ('stratis-cli', 'stratisd')
        services = ('stratisd',)
        profiles = ('storage',)
    
        def get_pool_names(self, pool_file):
            """Return a list of pool names from "stratis pool list" cmd output
               where the cmd output looks like:
               Name     Total Physical Size  Total Physical Used
               pool1                  3 GiB               60 MiB
               pool2                  3 GiB               60 MiB
            """
            out = []
            try:
                pool_out = open(pool_file).read()
            except IOError:
                return out
            for line in pool_out.splitlines()[1:]:
                out.append(line.split()[0])
            return out
    
        def setup(self):
            subcmds = [
                'filesystem list',
                'blockdev list',
                'daemon redundancy',
                'daemon version'
            ]
    
            self.add_cmd_output(["stratis %s" % subcmd for subcmd in subcmds])
            self.add_journal('stratisd')
    
            pool_file = self.get_cmd_output_now("stratis pool list")
            if pool_file:
                for pool in self.get_pool_names(pool_file):
                    self.add_cmd_output([
                        "stratis blockdev list "+pool,
                        "stratis filesystem list "+pool
                    ])
    
    # vim: set et ts=4 sw=4 :

bmr-cymru:
```
    def get_pool_names(self, pool_file):
        """Return a list of pool names from "stratis pool list" cmd output
           where the cmd output looks like:
           Name     Total Physical Size  Total Physical Used
           pool1                  3 GiB               60 MiB
           pool2                  3 GiB               60 MiB
        """
        out = []
        try:
            pool_out = open(pool_file).read()
        except IOError:
            return out
        for line in pool_out.splitlines()[1:]:
            out.append(line.split()[0])
        return out
```

What's filtering out those column headers here? Does `stratis` support a `--noheadings` type option?

bmr-cymru:
And what happens if the command returns no output?

pmoravec:
Good point.

Surprisingly both `python2` and `python3` deal with the code well, since in both:

    >>> ''.splitlines()[1:]
    []

and also the particular code I suggest returns empty list as desired.

Anyway, maybe better not in relying to this and add a condition there - leaving that on @TurboTurtle .

bmr-cymru:
No, I think it's fine. I hadn't had enough coffee to quell the paranoia this morning. ;-)

TurboTurtle:
LGTM, I'll pull in the changes shortly.

We do seem to be doing a lot of "get this command output now and iterate over it" recently. I wonder if we should put something directly into `Plugin` to handle this so we don't have to keep re-writing it in plugins.

PR #1497, opened 2018-12-04
Collect full db dump

didib:
---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -569,7 +569,8 @@ class Plugin(object):
         a single file the file will be tailed to meet sizelimit. If the first
         file in a glob is too large it will be tailed to meet the sizelimit.
         """
-        sizelimit = sizelimit or self.get_option("log_size")
+        if sizelimit is None:
+            sizelimit = self.get_option("log_size")
 
         if self.get_option('all_logs'):
             sizelimit = None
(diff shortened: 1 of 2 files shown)


sos/plugins/__init__.py
```diff
@@ -569,7 +569,7 @@ def add_copy_spec(self, copyspecs, sizelimit=None, tailit=True):
         a single file the file will be tailed to meet sizelimit. If the first
         file in a glob is too large it will be tailed to meet the sizelimit.
         """
-        sizelimit = sizelimit or self.get_option("log_size")
+        sizelimit = sizelimit if sizelimit is not None else self.get_option("log_size")
```
bmr-cymru: This isn't going to work as a single line, as it exceeds the pep8 length limit. We can either stuff `get_option("log_size")` into a local, or rework it as a simple branched `if: ... else:`.

sos/plugins/__init__.py
```diff
@@ -703,7 +703,7 @@ def add_cmd_output(self, cmds, suggest_filename=None,
             cmds = [cmds]
         if len(cmds) > 1 and (suggest_filename or root_symlink):
             self._log_warn("ambiguous filename or symlink for command list")
-        sizelimit = sizelimit or self.get_option("log_size")
+        sizelimit = sizelimit if sizelimit is not None else self.get_option("log_size")
```
bmr-cymru: Ditto.

didib:
I don't like the review process on github. Not sure how to mark that I added Signed-Off-By.

didib:
Not sure how to mark that I added Signed-Off-By.

didib:
Sorry for spamming - when added the second comment, I couldn't see the first one, so thought there might have been some problem submitting it... Did I say I do not like github? I almost always find gerrit much cleaner.

didib:
Now verified that it works. Created a dummy huge table, ran old sosreport and saw that the dump is partial, patched with current PR, ran again and saw that it's complete.

pmoravec:
ACK from me, codechange is sane, verified on my system.

PR #1499, opened 2018-12-04
[sosreport] Allow user-controllable plugin timeouts

TurboTurtle:
Allows users to specify a timeout for each plugin using the '-k
plugin.timeout=value' syntax by adding the 'timeout' option to every
plugin.

Additionally, adds the --plugin-timeout option to set a timeout for
_all_ plugins. If --plugin-timeout and a specific -k timeout option is
provided, the -k timeout option will be applied for those specific
plugins, with --plugin-timeout being applied to all others.

In either case, specifying a timeout of 0 seconds results in no timeout
being applied. In the event that an invalid timeout is set, the timeout
will be set to the default value for the plugin.

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -127,7 +127,7 @@ class Plugin(object):
     archive = None
     profiles = ()
     sysroot = '/'
-    timeout = 300
+    plugin_timeout = 300
 
     def __init__(self, commons):
         if not getattr(self, "option_list", False):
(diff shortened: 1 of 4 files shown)


sos/plugins/__init__.py
```diff
@@ -150,12 +150,40 @@ def __init__(self, commons):
...
+            elif opt_timeout is not None and own_timeout > -1:
+                _timeout = own_timeout
+            else:
+                return None
+        except ValueError:
+            return self.plugin_timeout  # Default to known safe value
+        if _timeout is not None and _timeout > -1:
+            return _timeout
+        return self.plugin_timeout
+
```
pmoravec: I thought the `elif` are wrong but I am unable to find particular counter-example.

Can't something be simplified if we allow just numerical values for the options? Like `./sosreport --log-size=a1` fails with `sosreport: error: argument --log-size: invalid int value: 'a1'`?

pmoravec: Havent realized that complexity behind, what can easily justify that code, then.

bmr-cymru: Yeah, that kinda sucks but as @TurboTurtle explained it's a consequence of how the Python bits deal with certain special values. All that said, it's really not too bad - we have far more icky blocks in the rest of the codebase, and the logic is quite easy to follow if you understand the intent of the precedence.

bmr-cymru:
Thanks for working on this - very nice feature.

bmr-cymru:
> maybe the restriction of argument values to numbers only would be beneficial?

Let's keep that discussion going in #1526

PR #1500, opened 2018-12-05
[postgresql] Use the custom dump format

didib:
Instead of tar.

'custom' should be smaller, faster, and allows more features - in
particlar, using the '-j' option to pg_restore (to run stuff in
parallel, so hopefully faster).

One drawback is that you can't use 'tar' and 'less' to open and
view it - you need pg_restore. But I think it's worth it.

Signed-off-by: Yedidyah Bar David <didi@redhat.com>

---

### DIFF
--- a/sos/plugins/postgresql.py
+++ b/sos/plugins/postgresql.py
@@ -39,7 +39,7 @@ class PostgreSQL(Plugin):
         ('dbport', 'database server port number', '', '5432')
     ]
 
-    def do_pg_dump(self, scl=None, filename="pgdump.tar"):
+    def do_pg_dump(self, scl=None, filename="pgdump.dump"):
         if self.get_option("dbname"):
             if self.get_option("password") or "PGPASSWORD" in os.environ:
                 # We're only modifying this for ourself and our children so
@@ -49,14 +49,14 @@ class PostgreSQL(Plugin):
                     os.environ["PGPASSWORD"] = str(self.get_option("password"))
 
                 if self.get_option("dbhost"):
-                    cmd = "pg_dump -U %s -h %s -p %s -w -F t %s" % (
+                    cmd = "pg_dump -U %s -h %s -p %s -w -F c %s" % (
                         self.get_option("username"),
                         self.get_option("dbhost"),
                         self.get_option("dbport"),
                         self.get_option("dbname")
                     )
                 else:
-                    cmd = "pg_dump -C -U %s -w -F t %s " % (
+                    cmd = "pg_dump -C -U %s -w -F c %s " % (
                         self.get_option("username"),
                         self.get_option("dbname")
                     )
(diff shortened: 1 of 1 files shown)


pmoravec:
+1, ack.

didib:
Now verified locally.

didib:
> This change may break our support team's utility to restore customer databases locally.

I think it should be pretty easy to fix. I can help if you want.

> 
> It will also certainly break any examination scripts or tools that try to summarize environment data from the db dump (like pysos - especially when I get around to revitalizing it) to avoid the need to restore the whole DB (e.g. giving a breakdown of existing DCs, Clusters, and Storage Domains).

Same.

I am not claiming this patch has no consequences - just that I think that it's worth it.

Please note that you can easily view/dump parts of a custom dump without first restoring to a database. E.g.:

# convert to SQL (like the 'plain' format)
pg_restore engine.dump | less

# Output a single table
pg_restore -t schema_version engine.dump | less

TurboTurtle:
> > It will also certainly break any examination scripts or tools that try to summarize environment data from the db dump (like pysos - especially when I get around to revitalizing it) to avoid the need to restore the whole DB (e.g. giving a breakdown of existing DCs, Clusters, and Storage Domains).
> 
> Same.
>
> Please note that you can easily view/dump parts of a custom dump without first restoring to a database. E.g.:
> 
>
> pg_restore engine.dump | less
> 
> pg_restore -t schema_version engine.dump | less

This assumes pg_restore is on any system where a support engineer may be handling sosreports. 

With the current tar format we can grok through the dump files like any other text file easily without relying on external utilities.

didib:
> This assumes pg_restore is on any system where a support engineer may be handling sosreports.

Indeed. Even more, it requires a compatible version. pg_restore of PG 8 will not open dumps created by PG 9.

> 
> With the current tar format we can grok through the dump files like any other text file easily without relying on external utilities.

Indeed, and I personally appreciate very much the ability to do so. The question is - is this the most important "feature", on which to base the decision?

I am closing this PR. If there is ever interest in either accepting it as-is, or specific change requests (e.g. make this optional), feel free to ping me. But if the objection is deeper, on the grounds of "when you debug a sosreport, it's more important that you have the feeling of freedom from specific tools offered only by using very common formats, ideally text-based, over any other factor (speed, features, etc.)", then I do not think I have much to add...

For the record - in oVirt's engine-backup tool, we used the plain format for some releases. At some point I did some testing with the various formats and options, mainly performance-related, but also others, and eventually decided to make the format default to 'custom', and optionally allow 'plain'. We never looked back, and I judging by user reports, I think almost eveyone uses the default, 'custom'. One "feature" we noticed and used only after making the change, is that with the plain format, and I think tar format too, using '--no-owner' or '--no-privileges' must be done during dump time, whereas with 'custom', you can make the choice at restore time. Logic seems to me to be that with the plain format (and tar, which is mostly plain but with the actual table data being split to separate files), once the dump is created, relevant tools do not want to parse it, other than with 'psql'. But with the 'custom' format, I guess these things are more easily tagged/marked/etc. in the dump, so it's safer for other tools (e.g. pg_restore) to parse them. Obviously, one might claim that the considerations for a backup/restore tool are significantly different from a debug tool.

PR #1501, opened 2018-12-05
[mysql] Enable mysql for OSP container env

stuggi:
In containerized OSP env we now do not have no mysql packages installed
outside the containers. Therefore the mysql plugin does not auto run.
This adds the openstack-selinux package to the RH package list of the
mysql plugin like on the OSP plugins.

---

### DIFF
--- a/sos/plugins/mysql.py
+++ b/sos/plugins/mysql.py
@@ -82,7 +82,8 @@ class RedHatMysql(Mysql, RedHatPlugin):
         'mysql-server',
         'mysql',
         'mariadb-server',
-        'mariadb'
+        'mariadb',
+        'openstack-selinux'
     )
 
     def setup(self):


pmoravec:
I would prefer rather more generic approach of enabling plugins for OpenStack, as the workaround applied in 20ish plugins "enable plugin mysql (or any other) whenever openstack-selinux is installed" is.. well rather workaround.

What about extending RHOSP presets options in https://github.com/sosreport/sos/blob/master/sos/policies/redhat.py#L225 to enable the 20ish plugins (openstack_* and mysql included)?

lyarwood:
> Hello @stuggi , @lyarwood and @TurboTurtle
> 
> Any idea about generic way of identifying a system is an OpenStack instance or not / sos OSP plugins (some/all) shall be run by default? E.g. if you log into a system or container, how could you detect some part of OpenStack runs on it?

From the host is container `foo` running, from within the container is package `bar` installed.

I've discussed introducing a `ContainerPlugin` base class in the past to wrap commands in the required docker or podman command lines, grab files from within containers etc but I've not had time to make a start on this.

> (so far we have 2 possible ideas of solutions:
> 
> * enable all OSP plugins whenever `openstack-selinux` package is present, like https://github.com/sosreport/sos/tree/bmr-openstack-triggers

Right that's the existing workaround used in the other plugins and the best approach for now IMHO.

> * extend RHOSP preset (enabled by `rhosp-release` package presence) to enable all osp* plugins
>   )

There's no rhosp-release package so that's a non-starter.

PR #1502, opened 2018-12-05
[peripety] Add new plugin

TurboTurtle:
Adds a new plugin for the storage event monitoring daemon peripety.

---

### DIFF
--- a/sos/plugins/peripety.py
+++ b/sos/plugins/peripety.py
@@ -1,0 +1,41 @@
+# Copyright (C) 2018 Red Hat, Inc., Jake Hunsaker <jhunsake@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.plugins import Plugin, RedHatPlugin
+from re import match
+import glob
+
+
+class Peripety(Plugin, RedHatPlugin):
+    '''Peripety Storage Event Monitor'''
+
+    packages = ('peripety',)
+    services = ('peripetyd',)
+
+    def setup(self):
+        self.add_copy_spec('/etc/peripety.conf')
+
+        forbid_reg = [
+            'vd.*',
+            'sr.*',
+            'loop.*',
+            'ram.*'
+        ]
+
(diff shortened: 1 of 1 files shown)


sos/plugins/peripety.py
```diff
@@ -0,0 +1,39 @@
...
+    packages = ('peripety',)
+    services = ('peripetyd',)
+
+    def setup(self):
+        self.add_copy_spec('/etc/peripety.conf')
+
+        forbid_reg = [
+            'vd.*',
+            'sr.*'
+        ]
```
bmr-cymru: Do we need to split this over multiple lines? Looks to be pretty short strings, or do we expect further additions in future?

sos/plugins/peripety.py
```diff
@@ -0,0 +1,39 @@
...
+
+        disks = filter(lambda x: not any(match(reg, x) for reg in forbid_reg),
+                       [d.split('/')[-1] for d in glob.glob('/sys/block/*')])
```
bmr-cymru: I'd really like to get things like the long-standing ask for disk/network etc. iterators/collections done for this. Then we can lose all these ad-hoc "walk all the disks" type things.

Does this want to be also filtering things like ram disks, loopbacks etc.?

sos/plugins/peripety.py
```diff
@@ -0,0 +1,39 @@
...
+        for disk in disks:
+            self.add_cmd_output([
+                "prpt info %s" % disk,
+                "prpt query --blk %s" % disk
+            ])
```
bmr-cymru: I wonder if this couldn't be done more compactly with a list comprehension? (didn't try).

sos/plugins/peripety.py
TurboTurtle: Yep - I imagine this list will be built out in the future.

sos/plugins/peripety.py
TurboTurtle: I tried a couple variations and couldn't get more compact than that and they resulted in two for-loops instead of one.

sos/plugins/peripety.py
TurboTurtle: I wasn't sure if we were still looking at those iterators from a previous conversation we had on IRC. If those are still on the wish list I agree that'd be a big win to get something hammered out there.

I _think_ loopbacks should be fine, I'm not sure on ram disks. Will test it out.

TurboTurtle: Tried loopbacks and ramdisks, both fail so I've added them to the forbids list.

peripety is still pretty new so I imagine these will be tweaked a decent amount going forward.

PR #1503, opened 2018-12-06
[networking] Extend ethtool command set

ptalbert:
Update the list of ethtool commands to include:

ethtool -e (EEPROM dump)
ethtool -P (permanent MAC address)
ethtool -l (channel/queue settings)
ethtool --phy-statistics
ethtool --show-priv-flags
ethtool --show-eee

All of the above are helpful in understanding the state of modern NICs.
And -P is nice to have as otherwise there is no reliable way to see the
permanent MAC of team ports.

---


bmr-cymru:
Modified for the switch to `/sys` for `eth` device discovery.

PR #1508, opened 2018-12-10
[Plugin] Add iterator for block devices, update relevant plugins, and remove add_udev_info 

TurboTurtle:
This series of commits adds a new `add_blockdev_cmd()` method to the Plugin class that allows for easy iteration of commands over block devices on a system running sosreport.

This removes the need for each plugin to independently enumerate block devices (and in turn standardizes the way we do that). The `add_blockdev_cmd()` method can take either a string or list of strings as commands, substituting the `%(dev)s` var for the block device. 

By default, this method will run against block devices (per `/sys/block/`) and prepend devices names with `/dev/`. This can be controlled by setting the `prepend_path` parameter, and the `devices` can be set to either `block` for block devices (default behavior) `fibre` for fibre devices, or handed a list of device names manually.

Additionally, a `whitelist` or `blacklist` can be applied to filter which devices are valid for commands to be run against. Both of that arguments can be either a regex string or a list of regex strings. 

Included in this are update to the `block`, `nvme`, and `fibrechannel` plugins to use this new method.

Finally, remove `add_udev_info()` as it's initial concept was easy device iteration, which this new method handles better and also provides for the previously mentioned device filters. 

I imagine this will eventually be expanded upon to include network devices, so the actual work of filtering and building the commands is handled by `_add_device_cmd()` and `add_blockdev_cmd()` simply handles setting the correct device list - much like I envision a potential `add_netdev_cmd()` will do.

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -459,6 +459,7 @@ class Plugin(object):
         self.collect_cmds = []
         self.sysroot = commons['sysroot']
         self.policy = commons['policy']
+        self.devices = commons['devices']
 
         self.soslog = self.commons['soslog'] if 'soslog' in self.commons \
             else logging.getLogger('sos')
@@ -1136,6 +1137,69 @@ class Plugin(object):
                 else:
                     # size limit not hit, copy the file
                     self._add_copy_paths([_file])
+
+    def add_blockdev_cmd(self, cmds, devices='block', timeout=300,
+                         sizelimit=None, chroot=True, runat=None, env=None,
+                         binary=False, prepend_path=None, whitelist=[],
+                         blacklist=[]):
+        """Run a command or list of commands against storage-related devices.
+
+        Any commands specified by cmd will be iterated over the list of the
+        specified devices. Commands passed to this should include a '%(dev)s'
+        variable for substitution.
+
+            devices - Either 'block', 'fibre', or a list of device paths to
+                      run against. If set to 'block' or 'fibre', the commands
+                      will be run against the matching list of discovered
+                      devices.
+            prepend_path - the leading path for block device names, e.g.
+                        '/dev/' or '/sys/block/'.
+            whitelist - limit the devices iterated over. May be a str or list
+                        of strings (e.g. regexes)
+            blacklist - exclude the devices iterated over. May be a str or list
+                        of strings (e.g. regexes)
+        """
+        if devices == 'block':
+            prepend_path = prepend_path or '/dev/'
+            devices = self.devices['block']
+        if devices == 'fibre':
+            devices = self.devices['fibre']
+        self._add_device_cmd(cmds, devices, timeout=timeout,
+                             sizelimit=sizelimit, chroot=chroot, runat=runat,
+                             env=env, binary=binary, prepend_path=prepend_path,
+                             whitelist=whitelist, blacklist=blacklist)
+
+    def _add_device_cmd(self, cmds, devices, timeout=300, sizelimit=None,
+                        chroot=True, runat=None, env=None, binary=False,
+                        prepend_path=None, whitelist=[], blacklist=[]):
+        """Run a command against all specified devices on the system.
+        """
+        if isinstance(cmds, six.string_types):
+            cmds = [cmds]
+        if isinstance(devices, six.string_types):
+            devices = [devices]
+        if isinstance(whitelist, six.string_types):
+            whitelist = [whitelist]
+        if isinstance(blacklist, six.string_types):
+            blacklist = [blacklist]
+        sizelimit = sizelimit or self.get_option('log_size')
+        for cmd in cmds:
+            for device in devices:
+                _dev_ok = True
+                if whitelist:
+                    if not any(re.match(wl, device) for wl in whitelist):
+                        _dev_ok = False
+                if blacklist:
+                    if any(re.match(blist, device) for blist in blacklist):
+                        _dev_ok = False
+                if not _dev_ok:
+                    continue
+                if prepend_path:
+                    device = os.path.join(prepend_path, device)
+                _cmd = cmd % {'dev': device}
+                self._add_cmd_output(cmd=_cmd, timeout=timeout,
+                                     sizelimit=sizelimit, chroot=chroot,
+                                     runat=runat, env=env, binary=binary)
 
     def _add_cmd_output(self, **kwargs):
         """Internal helper to add a single command to the collection list."""
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -1593,25 +1593,6 @@ class Plugin(object):
         self._add_cmd_output(cmd=journal_cmd, timeout=timeout,
                              sizelimit=log_size, pred=pred)
 
-    def add_udev_info(self, device, attrs=False):
-        """Collect udevadm info output for a given device
-
-        :param device: A string or list of strings of device names or sysfs
-                       paths. E.G. either '/sys/class/scsi_host/host0' or
-                       '/dev/sda' is valid.
-        :param attrs: If True, run udevadm with the --attribute-walk option.
-        """
-        udev_cmd = 'udevadm info'
-        if attrs:
-            udev_cmd += ' -a'
-
-        if isinstance(device, six.string_types):
-            device = [device]
-
-        for dev in device:
-            self._log_debug("collecting udev info for: %s" % dev)
-            self._add_cmd_output(cmd='%s %s' % (udev_cmd, dev))
-
     def _expand_copy_spec(self, copyspec):
         return glob.glob(copyspec)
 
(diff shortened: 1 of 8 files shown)


TurboTurtle:
Some notes:

- The tests are failing because `MockPlugin` passes in it's own commons dict, and I'm not sure that adding a devices dict to this is the right approach.

- I had `add_blockdev_cmd()` default to block devices since we only have one plugin doing anything with fibre devices - perhaps it'd be best to still enumerate the devices in the `fibrechannel` plugin and remove it from the devices dict? I don't see a need for a `add_fibre_cmd()` or similar given the single plugin use of it.

- Defaulting to `/dev/` path munging since everything that this would benefit today uses /dev/ paths, and none (that I know of anyways) accept `/sys/block/` paths.

sos/plugins/__init__.py
```diff
@@ -677,6 +678,69 @@ def check_ext_prog(self, prog):
...
+                    if not any(re.match(wl, device) for wl in whitelist):
+                        _dev_ok = False
+                if blacklist:
+                    if any(re.match(blist, device) for blist in blacklist):
+                        _dev_ok = False
+                if not _dev_ok:
+                    continue
+                if prepend_path:
+                    device = os.path.join(prepend_path, device)
+                sizelimit = sizelimit or self.get_option('log_size')
```
pmoravec: This is worth calling just once before the 2 nested `for` cycles.

sos/plugins/__init__.py
```diff
@@ -677,6 +678,69 @@ def check_ext_prog(self, prog):
...
+                _cmd = cmd % {'dev': device}
+                self._add_cmd_output(_cmd, timeout=timeout,
```
pmoravec: Maybe calling _log_debug similarly like deleted "`collecting udev info for:`" one? (though this method is more general and another debug will be printed in `_add_cmd_output` either way - the debug would have sense only if one is interested to know where the call comes from)

pmoravec:
Nice.

Few minor comments inline.

Further, does it make sense to update `ata` or `vhostmd` plugins the same way?

pmoravec:
(plus there is some small merge conflict in `sos/sosreport.py`)

pmoravec:
(this needs re-basing due to several conflicts (plus few very minor issues I spotted), otherwise LGTM.

TurboTurtle:
I've rebased this PR and moved the `sizelimit`  determination outside the loops.

I didn't add a debug message as I think the message logged by `_add_cmd_output` is sufficient.

TurboTurtle:
Finally got the last few bits squared away, sorry for the noise.

sos/sosreport.py
```diff
@@ -367,9 +367,51 @@ def __init__(self, args):
...
+        try:
+            devs = []
+            devdirs = [
+                'fc_host',
+                'fc_transport',
+                'fc_remote_ports',
+                'fc_vports'
+            ]
+            for devdir in devdirs:
+                if os.isdir(devdir):
```
pmoravec: Question from devices noob: is this test correct? To check for `./fc_host` directory presence (and others in `pwd`)?

Shall not we test `os.isdir("/sys/class/%s" % devdir)` or similar?

TurboTurtle: Good catch, looks like I inadvertently dropped that when refactoring the discovery bits. I'll update that later today.

sos/plugins/nvme.py
```diff
@@ -16,29 +15,23 @@ class Nvme(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
...
-                "smartctl --all /dev/%s" % dev,
-                "nvme list-ns /dev/%s" % dev,
-                "nvme fw-log /dev/%s" % dev,
-                "nvme list-ctrl /dev/%s" % dev,
-                "nvme id-ctrl -H /dev/%s" % dev,
-                "nvme id-ns -H /dev/%s" % dev,
-                "nvme smart-log /dev/%s" % dev,
-                "nvme error-log /dev/%s" % dev,
-                "nvme show-regs /dev/%s" % dev,
-                "nvme get-ns-id /dev/%s" % dev
```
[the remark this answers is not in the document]
TurboTurtle: `smartctl` should be there, yes. However, for `nvme get-ns-id` it looks to be effectively duplicate information from `nvme list-ns $dev` and just a single point of data from `nvme id-ns 0H $dev`, so I dropped that one - but I can certainly re-add it if desired.


```
$ sudo nvme get-ns-id /dev/nvme0n1
nvme0n1: namespace-id:1

$ sudo nvme list-ns /dev/nvme0n1
[   0]:0x1

$ sudo nvme id-ns -H /dev/nvme0n1
NVME Identify Namespace 1:
nsze    : 0x3a386030
ncap    : 0x3a386030
nuse    : 0xc8356f0
[... lots more...]
```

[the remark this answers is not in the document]
pmoravec: I dont know either command and dont request adding them. Just pointing to differences as possible regressions. But from your comment I deduce that `smartctl` is worth to collect while the later not.

sos/plugins/__init__.py
```diff
@@ -957,6 +958,69 @@ def time_filter(path):
...
+    def _add_device_cmd(self, cmds, devices, timeout=300, sizelimit=None,
+                        chroot=True, runat=None, env=None, binary=False,
+                        prepend_path=None, whitelist=[], blacklist=[]):
+        """Run a command against all specified devices on the system.
+        """
+        if isinstance(cmds, six.string_types):
+            cmds = [cmds]
+        if isinstance(devices, six.string_types):
+            devices = [devices]
+        if whitelist and isinstance(whitelist, six.string_types):
```
pmoravec: Does not the 2nd condition imply the first one? Hence isnt the first one redundant? (i.e. "if isinstance(..):" shall be sufficient & equivalent & bit faster).

The same applies to 2 lines below.

sos/plugins/block.py
```diff
@@ -39,19 +38,13 @@ def setup(self):
...
 
-        if os.path.isdir("/sys/block"):
-            for disk in os.listdir("/sys/block"):
-                if disk.startswith("ram"):
-                    continue
-                disk_path = os.path.join('/dev/', disk)
-                queue_path = os.path.join('/sys/block/', disk, 'queue')
-                self.add_udev_info(disk_path)
-                self.add_udev_info(disk_path, attrs=True)
-                self.add_udev_info(queue_path, attrs=True)
```
pmoravec: I think this command is not carried forward. At least when running sosreport without and with this PR and comparing collected files, these four are missing on my system:

    sos_commands/block/udevadm_info_-a_.sys.block.dm-0.queue
    sos_commands/block/udevadm_info_-a_.sys.block.dm-1.queue
    sos_commands/block/udevadm_info_-a_.sys.block.sda.queue
    sos_commands/block/udevadm_info_-a_.sys.block.sr0.queue

(not sure if that data is important to collect or not, just spotting difference / potential regression)

sos/plugins/__init__.py
TurboTurtle: I've removed the duplicate check in my most recent rebase.

sos/plugins/block.py
TurboTurtle: I had not moved those forward because on RHEL 8 and Fedora 28+, it does not appear to generate useful information:

RHEL:
```
# udevadm info -a /sys/block/sda/queue/
Unknown device, absolute path in /dev/ or /sys expected.
# udevadm info -a /sys/block/sda/queue
Unknown device, absolute path in /dev/ or /sys expected.
# udevadm info -a /sys/block/dm-1/queue/
Unknown device, absolute path in /dev/ or /sys expected.
```

Fedora:
```
Unknown device "/sys/block/dm-0/queue": No such device
Unknown device "/sys/block/nvme0n1/queue": No such device
Unknown device "/sys/block/sda/queue": No such device
```


But from the PR that added it, it says "very important". @jtpittman195  can you shed any light on this? Is ^ still important or do we need to change how we're calling it?

jtpittman195: Jake, sorry for the delay.  You're right, it fails on rhel8 and fedora due to a change in systemd.  Since sys/block/sdc/queue is not a device, udevadm rejects it.  The values it grabbed were ones that we used often with tuning and alignment cases, so if you're able to grab them differently via a sysfs copy or the like, that would be immensely helpful to us.  Thanks again.

TurboTurtle:
After chatting with @jtpittman195 about this, I've updated this PR to now collect the files under `/sys/block/*/queue/`, skipping the `iosched` directory therein.

@bmr-cymru PTAL
@BryanQuigley I seem to have recently lost the ability to restart Travis jobs, not sure why. But it looks like the Travis failure is due to something within Travis rather than this commit.

BryanQuigley:
Reset travis - although I think a rebase and force push might do a better job.

PR #1509, opened 2018-12-11
[redhat] Update policy to identify RHEL 8

TurboTurtle:
Now that RHEL 8 Beta is out, update the redhat policy dist_version to be
able to properly set the version.

---

### DIFF
--- a/sos/policies/redhat.py
+++ b/sos/policies/redhat.py
@@ -295,6 +295,8 @@ No changes will be made to system configuration.
                 return 6
             elif pkgname[0] == "7":
                 return 7
+            elif pkgname[0] == "8":
+                return 8
         except Exception:
             pass
         return False


TurboTurtle:
Nothing currently uses `dist_version()` as far as I can tell, but if we start using it in the future we should already have this taken care of.

bmr-cymru:
It correctly loads the `rhel` policy though, right? (it has every time I've tested it).

Much like `PackageManager` there is a lot of crap here that can be cleaned up.

TurboTurtle:
Yeah, it loads. Just saw this while looking through something else, and didn't want to let it be forgotten.

Unless it should be forgotten, in which case why don't we just remove it?

bmr-cymru:
If it's there, it's worth fixing - I'd be interested in seeing more work on this to really tidy up the policy and related classes a bit, but in the meantime things that fix something are always welcome.

PR #1511, opened 2018-12-12
[Plugin] fix get_option() on Python3 runtimes

bmr-cymru:
Fix `get_option()` to properly distinguish string and iterable types on Python3 runtimes.

---


bmr-cymru:
Obsoleted.

PR #1513, opened 2018-12-12
[tuned] Collect verify output

TurboTurtle:
Adds collection of 'tuned-adm verify' to allow for quick and easy
reference as to if tunables have been changed on a system using tuned.

---


bmr-cymru:
Is this a cheap operation? If so then let's grab it by default if it is useful. Otherwise if there's a noticable cost (I'm guessing not as presumably it's just config validation of some kind?) then let's put it behind `--verify`.

TurboTurtle:
Yeah this is just verifying (select) sysctl values against the currently loaded profile. It should be very cheap in all cases.

Issue #1504, opened 2018-12-06
Revise debian/ubuntu logs


BryanQuigley:
A TODO for myself..

Must kill, --all-logs collecting all of /var/log/, change to something like all syslog* kern*

Should try to detect the journal being persistent by looking at /var/log/journal

Try to detect if the journal has the last 7 days of logs and then not collect syslog.  Ideally collect last 7 days of syslog* kern* on default run.

PR #1519, opened 2018-12-19
[logs] Debian: Add Journal support and stop full tarball of /var/log

BryanQuigley:
WIP/Feedback wanted

These changes are only on the Debian/Ubuntu bit of logs

Persitant systemd journaling (/var/log/journal exists)
 use that instead of collecting any kern.log/syslog from /var/log.
 We also implement a log_days similar to the RH bit, but we default
 to 7 days instead of 3.
 --all-logs removes the day limit and gets it all

Rsyslog style (assumed if /var/log/journal doesn't exist
 default run: collect all 4 of syslog[.1] and kern.log[.1]
 (Same as previous behaviour).
 --all-logs: collect syslog* and kern.log*

The notable change for --all-logs is it used to just collect all of
/var/log which causes MANY issues.

Before this change using --all-logs on systems with persistant journal
we would collect duplicate log files as such:
 * Collect all of syslog*/kern*log
 * Collect the entire journal via /var/log/journal
 * Collect the last two boots of journal via running journalctl command

Even with this change we still collect the journal last/current boot
with whatever is copied by journal since log_days.

---

### DIFF
--- a/sos/plugins/logs.py
+++ b/sos/plugins/logs.py
@@ -104,21 +104,43 @@ class RedHatLogs(Logs, RedHatPlugin):
 
 class DebianLogs(Logs, DebianPlugin, UbuntuPlugin):
 
+    option_list = [
+        ("log_days", "the number of days logs to collect", "", 7)
+    ]
+
     def setup(self):
         super(DebianLogs, self).setup()
-        if not self.get_option("all_logs"):
-            self.add_copy_spec([
-                "/var/log/syslog",
-                "/var/log/syslog.1",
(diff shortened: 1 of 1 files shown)


BryanQuigley:
Posting for feedback of the wider group.

BryanQuigley:
Added auth log as well, one of most common reason for using --all-logs I've seen, and it's definitely in the same set as syslog/kern on Ubuntu.

BryanQuigley:
This would close  #1504 which I forgot I opened...

PR #1520, opened 2018-12-20
[pcp] Collect summary data for the current day

TurboTurtle:
Adds collection of a pmstat summary output using 5 minute intervals for
the current day, and makes a root symlink to it in the archive.

---


sos/plugins/pcp.py
```diff
@@ -136,6 +136,16 @@ def setup(self):
 
         # Need to get the current status of the PCP infrastructure
         self.add_cmd_output("pcp")
-
+        # Collect a summary for the current day
+        res = self.get_command_output('pcp')
```
pmoravec: Can't be merged with `self.add_cmd_output("pcp")`` few lines above into one call?

PR #1524, opened 2018-12-26
[python] Update plugin to run properly on RHEL 8

TurboTurtle:
RHEL 8 replaces the standard 'python' binary with platform-python,
following PEP394. RHEL 8 sysadmins can however install python2 and/or
python3 alongside this system installation of python and optionally also
set 'python' to point to either of these installations.

This updates the python plugin to enable on the presence of any of these
packages and also get version information for any version of python
installed, including the platform-python installation.

---

### DIFF
--- a/sos/plugins/python.py
+++ b/sos/plugins/python.py
@@ -11,7 +11,7 @@
 from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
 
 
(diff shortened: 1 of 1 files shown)


TurboTurtle:
Note that this depends on #1509 to function properly for `platform-python`.

pmoravec:
+1

PR #1526, opened 2019-01-01
[plugins] Preserve data type of a plugin option to the default one

pmoravec:
Some plugin options rely on the default data type that can be
automatically overwritten by Python (e.g. from str to int).

The commit preserves the original data type to ensure implicit data
type specified in option definition is always used.

Resolves: #1526

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -509,9 +509,14 @@ class Plugin(object):
         return (self.opt_names, self.opt_parms)
 
     def set_option(self, optionname, value):
-        '''set the named option to value.'''
+        """Set the named option to value. Ensure the original type
+           of the option value is preserved.
+        """
         for name, parms in zip(self.opt_names, self.opt_parms):
             if name == optionname:
+                defaulttype = type(parms['enabled'])
+                if defaulttype != type(value) and defaulttype != type(None):
+                    value = (defaulttype)(value)
                 parms['enabled'] = value
                 return True
         else:


pmoravec:
.. and now with tests passing as well :)

bmr-cymru:
The code looks reasonable, and simple enough, but I am wondering if there is any value in keeping the (slightly weird...) option type coercion in `sos/sosreport.py`?

This is the stuff I mean:

```python
    def _set_tunables(self):
        if self.config.has_section("tunables"):
            if not self.opts.plugopts:
                self.opts.plugopts = []

            for opt, val in self.config.items("tunables"):
                if not opt.split('.')[0] in self._get_disabled_plugins():
                    self.opts.plugopts.append(opt + "=" + val)
        if self.opts.plugopts:
            opts = {}
            for opt in self.opts.plugopts:
                # split up "general.syslogsize=5"
                try:
                    opt, val = opt.split("=")
                except ValueError:
                    val = True
                else:
                    if val.lower() in ["off", "disable", "disabled", "false"]:
                        val = False
                    else:
                        # try to convert string "val" to int()
                        try:
                            val = int(val)
                        except ValueError:
                            pass
```

Taken together I am not sure how much sense these two bits of code make. Obviously there are plugins that rely today on receiving an `int` for an argument, but maybe we can simplify the whole scheme some how?

E.g. we could say that _all_ options are strings, and add a new `get_int_option()` or `get_option(type=int)` interface? It would be cleaner and more reliable in some ways I think, plus it makes it explicit how the option is to be treated and puts it in the control of the plugin writer.

pmoravec:
I agree with that approach though that isnt feasible in 3.7 timeframe - could be this commit merged before 3.7 and I will rework the better solution subsequently?

bmr-cymru:
I've filed an issue to review this after we release 3.7. I'll take the fix proposed here for now, but I'll add a FIXME comment referencing the new issue.

PR #1527, opened 2019-01-04
[redhat] Add RHCOS policy

TurboTurtle:
Adds a policy for Red Hat CoreOS.

---

### DIFF
--- a/sos/policies/redhat.py
+++ b/sos/policies/redhat.py
@@ -379,6 +379,48 @@ organization before being passed to any third party.
         return self.find_preset(ATOMIC)
 
 
+class RedHatCoreOSPolicy(RHELPolicy):
+    distro = "Red Hat CoreOS"
+    msg = _("""\
+This command will collect diagnostic and configuration \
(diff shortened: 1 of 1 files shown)


sos/policies/redhat.py
```diff
@@ -379,6 +379,48 @@ def probe_preset(self):
...
+
+Any information provided to %(vendor)s will be treated in \
+accordance with the published support policies at:\n
+  %(vendor_url)s
+
+The generated archive may contain data considered sensitive \
+and its content should be reviewed by the originating \
+organization before being passed to any third party.
+%(vendor_text)s
+""")
```
bmr-cymru: This is the same as the RHEL `msg`, isn't it, modulo "and installed applications" - I'd rather we kept the legal disclaimer in one place, and not have to remember to update multiple copies.

There's two ways to do that - we can make it a formal member, and paste it into the message, or we can define a file level string and mung it into the message in each class.

bmr-cymru:
Closed in fa06bc0

PR #1528, opened 2019-01-04
[rhcos] Add new plugin

TurboTurtle:
Adds a new plugin for Red Hat CoreOS

---

### DIFF
--- a/sos/plugins/rhcos.py
+++ b/sos/plugins/rhcos.py
@@ -1,0 +1,30 @@
+# Copyright (C) 2019 Red Hat, Inc., Jake Hunsaker <jhunsake@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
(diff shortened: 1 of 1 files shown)


sos/plugins/rhcos.py
```diff
@@ -0,0 +1,30 @@
...
+class RHCoreOS(Plugin, RedHatPlugin):
+    """Red Hat CoreOS"""
+
+    plugin_name = 'rhcos'
+    packages = ('redhat-release-coreos', 'coreos-metadata')
+
+    def setup(self):
+        units = ['coreos-growpart', 'coreos-firstboot-complete']
+        for unit in units:
+            self.add_journal(unit)
```
bmr-cymru: Add journal does the string/list trick, like the other `add_*()` methods, so this can be just:

```python
        units = ['coreos-growpart', 'coreos-firstboot-complete']
        self.add_journal(units)
```

TurboTurtle: Same thing as in #1529 - passing a list compresses all the units into a single journalctl command, I think it'd be better to have them separately.

bmr-cymru:
Closed in fa06bc09c95c52565e29173535c7422608e9a29b

PR #1529, opened 2019-01-04
[rpmostree] Add new plugin

TurboTurtle:
Adds a new plugin for rpm-ostree, which is no longer limited to use in
Atomic Host.

---

### DIFF
--- a/sos/plugins/rpmostree.py
+++ b/sos/plugins/rpmostree.py
@@ -1,0 +1,40 @@
+# Copyright (C) 2019 Red Hat, Inc., Jake Hunsaker <jhunsake@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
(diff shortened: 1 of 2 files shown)


sos/plugins/rpmostree.py
```diff
@@ -0,0 +1,40 @@
...
+        ]
+
+        self.add_cmd_output(["rpm-ostree %s" % subcmd for subcmd in subcmds])
+
+        units = [
+            'rpm-ostreed',
+            'rpm-ostreed-automatic',
+            'rpm-ostree-bootstatus'
+        ]
+        for unit in units:
```
bmr-cymru: Ditto for this one.

TurboTurtle: Passing `add_journal()` a list puts all the units in a single `journalctl` command, which could lead to sizing problems. Also I think having each journal as a separate file makes it easier to dig through.

bmr-cymru: > Passing add_journal() a list puts all the units in a single journalctl command, which could lead to sizing problems. Also I think having each journal as a separate file makes it easier to dig through.

Well then let's change that. It doesn't seem especially helpful behaviour.

Issue #855, opened 2016-07-18
[sosreport] /etc/sos.conf to support (more/some/all) command line options


pmoravec:
Sosreport might be deployed in environments where different command line options might need to be used by default all the time. I.e. limit log size, set some plugin options, skip some plugins etc. It makes sense to store such defaults in `/etc/sos.conf`.

Shall `sos.conf` support all or specific types of parameters?

Shall cmdline options take precedence of the options in `sos.conf` (I think so)?

pmoravec:
Hm, I can add a new section (say `[general]` to `sos.conf` and load sosreport's cmdline options from there - but only _after_ they are loaded from cmdline. And I dont see a simple way to check if just parsed option from config has or has not been provided in cmdline - optparser does not have a method for that :(

So now I would end up with config options preceding cmdline options - something not intended.

pmoravec:
Since the time of opening this issue, presets were introduced.

I expect the right ordering of evaluation of tunables/settings should be:

1. apply presets
2. apply changes from config file (either `/etc/sos.conf` or the file from `--config-file=..` one), optionally overwrite those set via tunables
3. apply changes from command line, optionally overwrite those set via tunables or from config file

pmoravec:
OK, now the ordering is different: first cmdline options are loaded, then overwritten by those in `sos.conf`.

pmoravec:
There is a bit chicken-egg problem:

- the very first options to load are from presets
- to find preset, we need to know policy
- to set policy and/or preset, we need to evaluate command line options and config file
- but evaluating command line options and config file should be done _after_ evaluating presets

(similar chicken-egg problem is with evaluating non-default config file prior command line options where we need to process cmdline before the config - this can be yet resolved by some trick)

Gladly, `SoSOptions.merge` method helps here a lot:

- `self.opts` = "process command line options (to know config file or presets or sysroot)"
- `self.cfgfileopts` = "process config file from `self.opts["config_file"]`"
- `self.opts.merge(self.cfgfileopts)` (but fix the bug in `_merge_opt` to add src option to the _beginning_ of the list/tuple)
- set `policy` per `sysroot`
- `self.presetopts` = "process preset loaded from `self.opts["preset"]` and `self.opts["sysroot"]`"
- `self.opts.merge(self.presetopts)`

PR #1530, opened 2019-01-07
[sosreport] support all command line options in /etc/sos.conf

pmoravec:
- SoSOptions class changes:
  - from_file method to load load options from file
  - encapsulation of particular options into __init__ method
  - updated merge logic to replace non-default values only

- sosreport.py changes:
  - _parse_args (renamed to _get_parser) newly returns just the
    ArgumentParser object as parsing itself is done elsewhere
  - SoSOptions are built in ordering:
    - defaults from parser (with emtpy cmdline) loaded
    - cmdline options replace defaults
    - config file known even now
    - config.file options replace defaults until already set by
      cmdline (i.e. until SoSOptions marked them as nondefault)
    - presets can be known even now
    - options from given preset replace defaults, until the options
      are already updated from cmdline or config.file
  - some extra logging/formatting added
  - some extra comments added

Resolves: #855

---

### DIFF
--- a/sos/__init__.py
+++ b/sos/__init__.py
@@ -18,6 +18,13 @@ gettext to internationalize messages.
 import gettext
 import six
 
+from argparse import ArgumentParser
+
+if six.PY3:
+    from configparser import ConfigParser, ParsingError, Error
+else:
+    from ConfigParser import ConfigParser, ParsingError, Error
+
 __version__ = "3.6"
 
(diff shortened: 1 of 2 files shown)


bmr-cymru:
Thanks for working on this - really good to get this dealt with finally! I've made a quick pass through the code, and everything looks reasonable. I have some boom work to catch up on tomorrow but this is next in my list.

Issue #1533, opened 2019-01-09
lvm2 plugin use locking_type=0, may corrupt lvm on shared storage


nirs:
lvm2 plugin contains this note:

    # use locking_type 0 (no locks) when running LVM2 commands,
    # from lvm.conf:
    # Turn locking off by setting to 0 (dangerous: risks metadata
    # corruption if LVM2 commands get run concurrently).
    # None of the commands issued by sos ever modify metadata and this
    # avoids the possibility of hanging lvm commands when another process
    # or node holds a conflicting lock.

And then it use this config:

    --config 'global { locking_type= 0 }'

To run vgscan, pvscan, pvs, vgs, lvs.

According to David Teigland, *any* lvm command may attempt to recover VG metadata, so running
the lvm2 plugin on oVirt hypervisor is risky. In oVirt, same VG is accessed from multiple hosts. Only 
one host is allowed to do metadata changes, so other hosts must always use locking_type=4.

oVirt recommends setting up lvm filter hiding all shared storage from local lvm commands. A host with
such lvm is safe, but since lvm filter is not enforced yet by oVirt, we cannot rely on having proper lvm
filter.

Here is a related bug about corrupted VG metadata caused by "read-only" commands such as
"lvchange --refresh" with locking_type=1.
https://bugzilla.redhat.com/1553133

I think we should replace the config with:

    --config "global { locking_type=4 wait_for_locks=0 }"

At least when running on oVirt hypervisor, or maybe sosreport user should disable lvm2 plugin when
running on oVirt hypervisor?

vdsm plugin already collect lvm output from pvs, vgs, and lvs. We can enhance it to collect more info
collected by lvm2 plugin.

bmr-cymru:
We need resolution of the questions on the LVM side before we can do anything here: it's a bit daft that there is no way to prevent silent metadata modifications, or to use the tools in a truly read-only fashion (even when LVM locking is broken).

bmr-cymru:
On Thu, Jan 10, 2019 at 03:27:55PM -0800, Nir Soffer wrote:
> This may work better - avoiding locking issues, and getting all data in case of errors:
> 
>     --config 'global {locking_type=0  metadata_read_only = 0}'
> 
> It works for pvscan, vgscan, pvs, vgs, and lvs.

Thanks - that looks much more appealing.
 
> Regarding duplicate collection, I would happily delete lvm commands from vdsm plugin if it was possible,
[...]

PR #1538, opened 2019-01-12
[lvm2] Use metadata_read_only=1 for lvm commands

nirs:
We use locking_type=0 to ensure that we can collect data even if another
process holds a conflicting lock. This introduce a small risk since even
the read-only commands we use may attempt to recover on-disk data in
some cases. The risk is bigger when running lvm2 plugin in oVirt system,
when LVM is used on shared storage, accessible from many hosts.

Add metadata_read_only=1 to ensure that no command can attempt to do
unwanted on-disk changes.

Resolves: #1533

---

### DIFF
--- a/sos/plugins/lvm2.py
+++ b/sos/plugins/lvm2.py
@@ -37,14 +37,18 @@ class Lvm2(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
         self.add_cmd_output(cmd, chroot=self.tmp_in_sysroot())
 
     def setup(self):
-        # use locking_type 0 (no locks) when running LVM2 commands,
-        # from lvm.conf:
-        # Turn locking off by setting to 0 (dangerous: risks metadata
-        # corruption if LVM2 commands get run concurrently).
-        # None of the commands issued by sos ever modify metadata and this
-        # avoids the possibility of hanging lvm commands when another process
-        # or node holds a conflicting lock.
-        lvm_opts = '--config="global{locking_type=0}"'
+        # When running LVM2 comamnds:
+        # - use locking_type 0 (no locks) from lvm.conf: Turn locking
+        #   off by setting to 0 (dangerous: risks metadata corruption if
+        #   LVM2 commands get run concurrently).  This avoids the
+        #   possibility of hanging lvm commands when another process or
+        #   node holds a conflicting lock.
+        # - use metadata_read_only 1 (forbidd on-disk changes). Although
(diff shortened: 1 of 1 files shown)


nirs:
@bmr-cymru Can you review?

bmr-cymru:
It's on the list: generally the patch is simple enough & does what it says. I'm considering whether there are legitimate cases where we may want to be able to bypass this in order to still collect data when LVM is genuinely broken.

bmr-cymru:
Since the VDSM plugin you ship is currently out-of-tree, there's nothing stopping you doing that if you judge it to be right for your products.

After discussing this problem with Alasdair before DevConf I am more inclined to see this as a regression in LVM2 - as I thought, we had discussed this change with the team when it was first made, and we were assured that this would give the correct behaviour. That has now (very quietly) changed - I don't think we want to make a change that makes collection less reliable in the event of locking failure (or that would need an extra plugin option to "drill through" that situation) unless it is completely unavoidable.

nirs:
@bmr-cymru, this may be a regression in lvm2, but we have a bug reported with CentOS 7.4 when
lvchange --refresh corrupted vg metadta on shared storage:
https://bugzilla.redhat.com/1553133

So this is not a very new regression. I think the situation is worse in 7.6, since we see these failures
now in QE automated tests from time to time, here is one example:
https://bugzilla.redhat.com/show_bug.cgi?id=1553133#c61

According to David Teigland, this can be caused also by any lvm command, not only lvchange:
https://bugzilla.redhat.com/show_bug.cgi?id=1553133#c66

Can you explain why this change makes collection less reliable in the event of locking failure?

bmr-cymru:
> Can you explain why this change makes collection less reliable in the event of locking failure?

Because it will fail to lock, wait, time out, and exit without producing any output. It's the entire reason for the status quo.

If you're seeing this in RHV, then go ahead and make a change if you see fit - we have a bit more time upstream and are not under current release pressure, so I am not going to rush to commit a change that may have unintended consequences for our many RHEL and other users (we have no reports of the problem you describe on plain RHEL at this time).

bmr-cymru:
OK. Having discussed this with Alasdair, I do consider this to be a regression in LVM2, but considering that the upstream approach now is to deprecate (and soon remove) the `locking_mode` entirely, I think it's not useful to try to get this reverted back to the state that `sos` has conventionally expected.

There are no plans that I am aware of to remove `metadata_read_only` at this time, so we can add that, and we should not need to consider this again until the next upstream change to entirely remove the locking modes.

PR #1541, opened 2019-01-17
Linux switch extensions

kvadrage:
---

### DIFF
--- a/sos/plugins/bird.py
+++ b/sos/plugins/bird.py
@@ -1,0 +1,50 @@
+# Copyright (C) 2019 Alexander Petrovskiy <alexpe@mellanox.com>
+#
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.plugins import Plugin, RedHatPlugin, UbuntuPlugin, DebianPlugin
+
+
+class Bird(Plugin, RedHatPlugin, UbuntuPlugin, DebianPlugin):
+    """Bird routing daemon
+    """
+
+    plugin_name = 'bird'
+    profiles = ('network',)
+    packages = ('bird',)
+
+    def setup(self):
+        self.add_copy_spec([
+            "/etc/bird/*",
+            "/etc/bird.conf"
+        ])
+        self.add_cmd_output([
+            "birdc show status",
+            "birdc show memory",
+            "birdc show protocols all",
+            "birdc show interfaces",
+            "birdc show route all",
+            "birdc show symbols",
+            "birdc show bfd sessions",
+            "birdc show babel interfaces",
+            "birdc show babel neighbors",
+            "birdc show babel entries",
+            "birdc show babel routes",
+            "birdc show ospf",
+            "birdc show ospf neighbors",
+            "birdc show ospf interface",
+            "birdc show ospf topology",
+            "birdc show ospf state all",
+            "birdc show ospf lsadb",
+            "birdc show rip interfaces",
+            "birdc show rip neighbors",
+            "birdc show static"
+        ])
+
+# vim: set et ts=4 sw=4 :
--- a/sos/plugins/networking.py
+++ b/sos/plugins/networking.py
@@ -155,6 +155,54 @@ class Networking(Plugin):
             pred=SoSPredicate(self, kmods=['ip6table_filter'])
         )
 
+        # Get devlink output
+        self.add_cmd_output([
+                    "devlink dev",
+                    "devlink dev param show",
+                    "devlink dev info",
+                    "devlink port",
+                    "devlink sb show",
+                    "devlink sb pool show",
+                    "devlink sb port pool show",
+                    "devlink sb tc bind show",
+                    "devlink -s -v trap show"
+        ])
+
+        # Collect information about devlink devices and dpipe tables
+        devlink_dev = self.exec_cmd("devlink dev")
+        if devlink_dev['status'] == 0:
+            devices = []
+            dev_tables = []
+            for line_dev in devlink_dev['output'].splitlines():
+                devices.append(line_dev)
+                # Collect information about devlink dpipe tables
+                devlink_dev_tables = self.exec_cmd(
+                    "devlink dpipe table show %s" % line_dev
+                    )
+                if devlink_dev_tables['status'] != 0:
+                    continue
+                for line_tables in devlink_dev_tables['output'].splitlines():
+                    line_tables = line_tables.strip()
+                    if line_tables.startswith("name "):
+                        dev_tables.append((line_dev, line_tables.split()[1]))
+
+            # Capture per-device output from devlink
+            devlink_cmds = [
+                "devlink sb occupancy snapshot %s",
+                "devlink sb occupancy show %s",
+                "devlink -v resource show %s",
+                "devlink dpipe table show %s"
+            ]
+            self.add_cmd_output([
+                cmd % dev for cmd in devlink_cmds for dev in devices
+            ])
+
+            # Capture per-device per-table output from devlink dpipe
+            devlink_dpipe_cmd = "devlink dpipe table dump %s name %s"
+            self.add_cmd_output([
+                devlink_dpipe_cmd % (dev, table) for dev, table in dev_tables
+            ])
+
         # Get ethtool output for every device that does not exist in a
         # namespace.
         for eth in listdir("/sys/class/net/"):
--- a/sos/plugins/networking.py
+++ b/sos/plugins/networking.py
@@ -112,6 +112,7 @@ class Networking(Plugin):
         self.add_cmd_output([
             "netstat -s",
             "netstat %s -agn" % self.ns_wide,
+            "ip vrf",
             "ip route show table all",
             "ip -6 route show table all",
             "ip -4 rule",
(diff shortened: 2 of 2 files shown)


sos/plugins/networking.py
```diff
@@ -209,6 +234,25 @@ def setup(self):
...
+        if self.is_module_loaded("devlink"):
+            self.add_cmd_output("devlink dev show")
+            self.add_cmd_output("devlink port show")
+            self.add_cmd_output("devlink sb show")
+            self.add_cmd_output("devlink sb pool show")
+            self.add_cmd_output("devlink sb port pool show")
+            self.add_cmd_output("devlink sb tc bind show")
+            self.add_cmd_output("devlink region show")
+            self.add_cmd_output("devlink dev param show")
+            for dev in self.get_devlink_devs():
```
bmr-cymru: All the `add_*()` collection methods in `sos` accept either a string with a single item to collect, or a list of strings with multiple items. Please change this to make a single call to `add_cmd_output()` with each `devlink` command as an element in the list argument.

sos/plugins/networking.py
```diff
@@ -68,6 +68,31 @@ def get_netns_devs(self, namespace):
...
+        try:
+            devlink_tables_out = open(devlink_tables_file).read()
+        except IOError:
+            return out
+        for line in devlink_tables_out.splitlines():
+            # match only lines where table name is specified
+            match = re.match('\s*name\s+(\w+)', line)
+            if match:
+                out.append(match.group(1))
+        return out
```
bmr-cymru: Ack.

sos/plugins/networking.py
```diff
@@ -209,6 +234,25 @@ def setup(self):
...
+            self.add_cmd_output("devlink sb show")
+            self.add_cmd_output("devlink sb pool show")
+            self.add_cmd_output("devlink sb port pool show")
+            self.add_cmd_output("devlink sb tc bind show")
+            self.add_cmd_output("devlink region show")
+            self.add_cmd_output("devlink dev param show")
+            for dev in self.get_devlink_devs():
+                self.add_cmd_output("devlink -v resource show %s" % dev)
+                self.add_cmd_output("devlink dpipe header show %s" % dev)
+                devlink_table_file = self.get_cmd_output_now("devlink dpipe table show %s" % dev)
```
bmr-cymru: Ditto - you can just pass the entire list returned by `get_devlink_devs()` into a single call, using a list comprehension to build the commands from the format string and the device list.

sos/plugins/networking.py
```diff
@@ -209,6 +234,25 @@ def setup(self):
...
+                if devlink_table_file:
+                    for table in self.get_devlink_tables(devlink_table_file):
+                        self.add_cmd_output("devlink dpipe table dump %s name %s" % (dev, table))
+
```
bmr-cymru: Same here.

sos/plugins/bird.py
```diff
@@ -0,0 +1,36 @@
...
+        ])
+        self.add_cmd_output("birdc show status")
+        self.add_cmd_output("birdc show memory")
+        self.add_cmd_output("birdc show protocols all")
+        self.add_cmd_output("birdc show route all")
+        self.add_cmd_output("birdc symbols")
+        self.add_cmd_output("birdc show bfd sessions")
+        self.add_cmd_output("birdc show ospf")
+        self.add_cmd_output("birdc show rip")
+        self.add_cmd_output("birdc show static")
```
bmr-cymru: Please convert to list.

pmoravec:
Hello @kvadrage , could you please follow the suggestions (and rebase your PR to current upstream as a new conflict has appeared)?

kvadrage:
> Hello @kvadrage , could you please follow the suggestions (and rebase your PR to current upstream as a new conflict has appeared)?

Hello! Sorry for long delay. I've just rebased the PR and updated it. Some commits were excluded. 
Kindly waiting for your new feedback.

sos/plugins/networking.py
```diff
@@ -155,6 +155,54 @@ def setup(self):
...
+                    "devlink dev",
+                    "devlink dev param show",
+                    "devlink dev info",
+                    "devlink port",
+                    "devlink sb show",
+                    "devlink sb pool show",
+                    "devlink sb port pool show",
+                    "devlink sb tc bind show",
+                    "devlink -s -v trap show"
+                ])
```
bmr-cymru: This line is over-indented (the closing ']' should align with the 's' in `self.add_cmd_output()`).

sos/plugins/networking.py
```diff
@@ -155,6 +155,54 @@ def setup(self):
...
+                    )
+                if devlink_dev_tables['status'] != 0:
+                    continue
+                for line_tables in devlink_dev_tables['output'].splitlines():
+                    line_tables = line_tables.strip()
+                    if line_tables.startswith('name'):
+                        dev_tables.append((line_dev,
+                                            line_tables.split()[1]))
+            # Capture per-device output from devlink
+            self.add_cmd_output(["devlink sb occupancy snapshot %s" % dev
```
bmr-cymru: If you're having to split a call like this the preferred format is:

```python
    self.add_cmd_output([
        "devlink sb occupancy snapshot %s" % dev for dev in devices
    ])
```

That said, I think we can improve formatting and readability further here by using a table of commands:

```python
    devlink_cmds = [
        "devlink sb occupancy snapshot %s",
        "devlink sb occupancy show %s",
        ...
    ]

    self.add_cmd_output([cmd % dev for cmd in devlink_cmds for dev in devices])
```

E.g.:

```python
>>> cmds = ["foo %s", "bar %s", "baz %s", "qux %s"]
>>> devs = ["xuq", "zab", "rab", "oof"]
>>> [(cmd % dev) for cmd in cmds for dev in devs]
['foo xuq', 'foo zab', 'foo rab', 'foo oof', 'bar xuq', 'bar zab', 'bar rab', 'bar oof', 'baz xuq', 'baz zab', 'baz rab', 'baz oof', 'qux xuq', 'qux zab', 'qux rab', 'qux oof']
```

sos/plugins/bird.py
bmr-cymru: Still please convert to a list... ;-)

bmr-cymru:
```python
        self.add_cmd_output([
            "birdc show status",
            ...
        ])
```

sos/plugins/bird.py
```diff
@@ -0,0 +1,36 @@
...
+
```
bmr-cymru: Please add a vim mode line here:


```
# vim: set et ts=4 sw=4 :
```

sos/plugins/networking.py
kvadrage: Fixed.

sos/plugins/networking.py
kvadrage: Thank you! I rewrote this part.

sos/plugins/bird.py
kvadrage: Fixed

kvadrage:
Hi @bmr-cymru !
Can you please check my last commits?

kvadrage:
Hi!

Do you have any additional comments for my last commits?

sos/plugins/networking.py
```diff
@@ -154,6 +155,54 @@ def setup(self):
...
+        self.add_cmd_output([
+                    "devlink dev",
+                    "devlink dev param show",
+                    "devlink dev info",
+                    "devlink port",
+                    "devlink sb show",
+                    "devlink sb pool show",
+                    "devlink sb port pool show",
+                    "devlink sb tc bind show",
+                    "devlink -s -v trap show"
```
TurboTurtle: These lines are over-indented. 4-spaces indented from the calling line, E.G.:

```python
self.add_cmd_output([
    "devlink", 
    "devlink dev param ",
    "devlink dev info",
    ...
])
```

sos/plugins/networking.py
```diff
@@ -154,6 +155,54 @@ def setup(self):
...
+        devlink_dev = self.exec_cmd("devlink dev")
+        if devlink_dev['status'] == 0:
+            devices = []
+            dev_tables = []
+            for line_dev in devlink_dev['output'].splitlines():
+                devices.append(line_dev)
+                # Collect information about devlink dpipe tables
+                devlink_dev_tables = self.exec_cmd(
+                    "devlink dpipe table show %s" % line_dev
+                    )
```
TurboTurtle: Since we're calling this command later on on line 194, instead of `exec_cmd` we can use `collect_cmd_output` which will both save the output to the archive and give you the output back directly in the `output` dict key.

In this loop there's nothing I immediately see that would need to change, we'd just need to remove the `devlink dpipe table show` entry from `devlink_cmds` below.

sos/plugins/networking.py
```diff
@@ -154,6 +155,54 @@ def setup(self):
...
+                    "devlink port",
+                    "devlink sb show",
+                    "devlink sb pool show",
+                    "devlink sb port pool show",
+                    "devlink sb tc bind show",
+                    "devlink -s -v trap show"
+        ])
+
+        # Collect information about devlink devices and dpipe tables
+        devlink_dev = self.exec_cmd("devlink dev")
```
TurboTurtle: Should we save `devlink dev` output? Is it ever useful on its own? If so, convert `exec_cmd` to `collect_cmd_output` here.

TurboTurtle: And now I see we're calling it separately above. So yeah, here we can convert to `collect_cmd_output` and then remove the `devlink dev` entry from line 160.

TurboTurtle:
@kvadrage re-ping. This looks to be in good shape still if you'd like for it to be merged. It'd just need a rebase to current master and then the minor `collect_cmd_output` changes noted above.

bmr-cymru (commented):
Sorry, I hadn't noticed the review request for this PR. If this is still relevant please rebase and re-request review.

PR #1546, opened 2019-01-27
[kdump] Collect vmcore-dmesg.txt files based on path and filesystem

mikelolasagasti:
kdump path can be changed with 'path' variable in kdump.conf. kdump
plugin always checks agains /var/crash/*/vmcore-dmesg.txt but path
could be other, so path variable is parsed first and vmcore-dmesg.txt
files are added based on the path that is configured.

---

### DIFF
--- a/sos/plugins/kdump.py
+++ b/sos/plugins/kdump.py
@@ -27,12 +27,39 @@ class RedHatKDump(KDump, RedHatPlugin):
     files = ('/etc/kdump.conf',)
     packages = ('kexec-tools',)
 
+    def fstab_parse_fs(self, device):
+        with open('/etc/fstab', 'r') as fp:
+            for line in fp:
+                if line.startswith((device)):
+                    return line.split()[1].rstrip('/')
+        return ""
+
+    def read_kdump_conffile(self):
+        fs = ""
+        path = "/var/crash"
+
+        with open('/etc/kdump.conf', 'r') as fp:
+            for line in fp:
+                if line.startswith("path"):
+                    path = line.split()[1]
+                elif line.startswith(("ext2", "ext3", "ext4", "xfs")):
+                    device = line.split()[1]
+                    fs = self.fstab_parse_fs(device)
+        return fs + path
+
     def setup(self):
         self.add_copy_spec([
             "/etc/kdump.conf",
             "/etc/udev/rules.d/*kexec.rules",
             "/var/crash/*/vmcore-dmesg.txt"
         ])
+        try:
+            path = self.read_kdump_conffile()
+        except Exception:
+            # set no filesystem and default path
+            path = "/var/crash"
+
+        self.add_copy_spec("{}/*/vmcore-dmesg.txt".format(path))
 
 
 class DebianKDump(KDump, DebianPlugin, UbuntuPlugin):


sos/plugins/kdump.py
```diff
@@ -30,10 +30,20 @@ class RedHatKDump(KDump, RedHatPlugin):
...
 
+        # default path if unset
+        path = "/var/crash"
+
+        fp = open('/etc/kdump.conf', 'r')
+        for line in fp.readlines():
+            if line.startswith("path"):
+                path = line.split()[1]
+        fp.close()
+
```
bmr-cymru: You need to handle possible exceptions in this block (open, read, split).

k-hagio:
Note that the `path` value is **relative** in a dump target if specified. According to the manpage of `kdump.conf`: 
```
       path <path>
              "path" represents the file system path in which vmcore  will  be
              saved.  If a dump target is specified in kdump.conf, then "path"
              is relative to the specified dump target.
```
I've seen some customer systems that configure like this:
```
xfs /dev/mapper/vg0-lv2-var_crash
path /
```
where the xfs filesystem is mounted on `/var/crash`. In this case, the data is written to `/var/crash/xxx/vmcore-dmesg.txt`. So I think it is better to leave `/var/crash/*/vmcore-dmesg.txt` as it is, if it is merged.

bmr-cymru:
Sure, but if no target is given then it's a path into the root file system and the argument still applies - and even if a target is given, nothing stops the user from setting path to something strange.

I think these configurations are extremely rare, but it is possible for them to exist.

k-hagio:
Sorry, my expression was not good..
I don't oppose this patch. Parsing path value is good. But is there need to remove this line?
```
-            "/var/crash/*/vmcore-dmesg.txt"
```
If this line is removed, sosreport will start to miss it in some cases like I said above.

mikelolasagasti:
Thanks for reviewing @k-hagio and @bmr-cymru. I made the patch after reviewing a customer's configuration, but it's clear that is incomplete. I'll try to extend current path logic by checking nfs/fs_type device.

mikelolasagasti:
I extended the PR and now checks if conf file has set a device and will check in fstab if a mount point is available for that device and applies path on top of it.  The old default "/var/crash/*/vmcore-dmesg.txt" is not present as the new code constructs that path using default values.

sos/plugins/kdump.py
```diff
@@ -27,13 +27,71 @@ class RedHatKDump(KDump, RedHatPlugin):
...
+                if line.startswith((device)):
+                    try:
+                        self.fs = line.split()[1].rstrip('/')
+                        return True
+                    except (IndexError):
+                        return False
+        return False
+
+    def kdump_parse_path(self):
+
```
bmr-cymru: Why a blank line here?

sos/plugins/kdump.py
```diff
@@ -27,13 +27,71 @@ class RedHatKDump(KDump, RedHatPlugin):
...
+                return False
+
+            for line in lines:
+                if line.startswith((device)):
+                    try:
+                        self.fs = line.split()[1].rstrip('/')
+                        return True
+                    except (IndexError):
+                        return False
+        return False
```
bmr-cymru: Since you don't log anything for the differentiated cases there isn't much point in handling the exceptions here and then signalling success/failure with a boolean return (that's generally frowned on in Python - return a value, and use the exception mechanism to signal exceptional events).

sos/plugins/kdump.py
```diff
@@ -27,13 +27,71 @@ class RedHatKDump(KDump, RedHatPlugin):
...
+    def kdump_parse_path(self):
+
+        for line in self.kdumpconf:
+            if line.startswith("path"):
+                try:
+                    self.path = line.split()[1]
+                    return True
+                except (IndexError):
+                    return False
+
```
bmr-cymru: Why are we going through this whole file twice?

sos/plugins/kdump.py
```diff
@@ -27,13 +27,71 @@ class RedHatKDump(KDump, RedHatPlugin):
...
+    def read_kdump_conffile(self):
+        try:
+            fp = open('/etc/kdump.conf', 'r')
+            self.kdumpconf = fp.readlines()
+            fp.close()
+            return True
+
+        except IOError:
+            return False
```
bmr-cymru: Ditto for boolean returns. You can also simplify the IO by using a `with` block.

sos/plugins/kdump.py
```diff
@@ -27,13 +27,71 @@ class RedHatKDump(KDump, RedHatPlugin):
...
-            "/var/crash/*/vmcore-dmesg.txt"
+            "/etc/udev/rules.d/*kexec.rules"
         ])
 
+        if self.read_kdump_conffile():
+            if not self.kdump_parse_fs():
+                self.fs = ""
+            if not self.kdump_parse_path():
+                self.path = "/var/crash"
+        else:
```
bmr-cymru: Using member variables to pass data between methods is confusing - having the default set dozens of lines away from the decision on the value means you have to scroll up and down all the time.

A better style is to just return the values from the functions, assign them once (have the function *return* the default, if that's the case to use), and then you can put all the exception handling into this block and not have it spread out through all the methods, along with the booleans.

sos/plugins/kdump.py
```diff
@@ -27,13 +27,71 @@ class RedHatKDump(KDump, RedHatPlugin):
     files = ('/etc/kdump.conf',)
     packages = ('kexec-tools',)
 
+    def kdump_parse_fs(self):
+        device = None
+        for line in self.kdumpconf:
+            if line.startswith(("ext2", "ext3", "ext4", "nfs", "xfs")):
```
k-hagio: Personally I think we should avoid active nfs access by tools like sosreport...

k-hagio:
> The old default "/var/crash/*/vmcore-dmesg.txt" is not present as the new code constructs that path using default values.

Thanks for your effort, but I still request it to keep the default path, because
1. a target can be specified with `UUID=xxxxx` format, and
2. a target path in kdump.conf is not always same as one in /etc/fstab.
```
ex)
ext4 UUID=03138356-5e61-4ab3-b58e-27507ac41937
path /
```
It looks hard to determine a mount point exactly with small and simple code, so I think some compromise might be good here.

bmr-cymru:
What's the status here now? Is any of the current work in a shape to merge?

mikelolasagasti:
@bmr-cymru I've simplified the code, should cover all your comments. I also removed NFS check for device and kept the original path as suggested by @k-hagio. Thanks for your reviews!

pmoravec:
PR sounds good to me. We might remove explicit collection of

                "/var/crash/*/vmcore-dmesg.txt"

if we do collect

    self.add_copy_spec("{}/*/vmcore-dmesg.txt".format(path))

But I think the current approach is OK. Since:
- the probably redundant `copy_spec` record is of no harm
- it is another way of failover of collecting the default location

So I am OK with the current implementation.

k-hagio:
669fa92 looks good to me and tested OK for some cases. So fwiw, +1 from me. Thanks.

PR #1548, opened 2019-01-30
Cos 3.6 kdump

xueweiz:
---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -1253,6 +1253,11 @@ class SuSEPlugin(object):
     pass
 
 
+class CosPlugin(object):
+    """Tagging class for Container-Optimized OS distributions"""
+    pass
+
+
 class IndependentPlugin(object):
     """Tagging class for plugins that can run on any platform"""
     pass
--- a/sos/plugins/cgroups.py
+++ b/sos/plugins/cgroups.py
@@ -6,10 +6,11 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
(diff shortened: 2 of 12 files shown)


xueweiz:
This pull request is to add kdump support in COS, which is the third commit in this series.
The first two is necessary as the context, and are being reviewed at https://github.com/sosreport/sos/pull/1419

bmr-cymru:
> Hi Bryan,

It's Bryn - no 'a'.

> COS is a locked-down OS, without any PackageManager

It will still have a set of installed packages - much like e.g. Atomic with RPMOSTree (albeit by a different mechanism).

The policy should emulate this - if necessary by providing or obtaining a static list of all packages and their versions, and exposing this through the `sos.policy.PackageManager` abstraction. You don't need to shell out to an actual package manager command for this, you just need to provide the data. It's hard for me to believe that an operating system provides no way to identify the software and its versions that are included - it'd be a nightmare to audit for a start.

> I think for now, only triggering the "whitelisted" plugins on COS is good enough for now. Later I can do a through scan, and get a more complete list of plugins that works on COS. Thanks!

It's more a matter of having a policy and plugin set that fits in with the overall `sos` scheme - the current COS policy is *far* too minimal and will not deliver the experience that our users expect. It needs to improve.

bmr-cymru:
ChromiumOS is the development version of ChromeOS, so it is based on Gentoo, which uses the Portage package manager: there should be files available in the read-only root file system that will provide the information that sos needs.

Note that even after doing this, working on this OS may be challenging since most commands and higher level tools are containerised. You can look at the work that we did for Atomic to get an idea of what is involved but it is non-trivial to support this kind of system today.

PR #1561, opened 2019-02-08
[watchdog] collect wdctl /dev/watchdog* outputs

pmoravec:
Resolves: #1561

---

### DIFF
--- a/sos/plugins/watchdog.py
+++ b/sos/plugins/watchdog.py
@@ -85,4 +85,8 @@ class Watchdog(Plugin, RedHatPlugin):
 
         self.add_copy_spec(log_files)
 
+        # Get output of "wdctl <device>" for each /dev/watchdog*
+        for dev in glob('/dev/watchdog'):
+            self.add_cmd_output("wdctl %s" % dev)
+
 # vim: set et ts=4 sw=4 :


sos/plugins/watchdog.py
```diff
@@ -85,4 +85,8 @@ def setup(self):
 
         self.add_copy_spec(log_files)
 
+        # Get output of "wdctl <device>" for each /dev/watchdog*
+        for dev in glob('/dev/watchdog'):
```
kgaillot: The * is missing

Also, this will catch most cases in practice, but watchdog devices can be named arbitrarily. sbd, for example, looks for all links in /sys/class/watchdog to get major/minor numbers then searches through /dev to find them (jumping through hoops to avoid triggering SELinux policies by stat'ing links in /dev):

https://github.com/ClusterLabs/sbd/blob/master/src/sbd-common.c#L277

That may be overkill here, and /dev/watchdog* sufficient for this purpose.

bmr-cymru: If you're going to add a review after the pull is closed, better to look at the version that was actually merged:

```
commit fcdb051448bdc1e5abc4af3a4ec59e4ceae97d74
Author: Pavel Moravec <pmoravec@redhat.com>
Date:   Fri Feb 8 08:10:07 2019 +0100

    [watchdog] collect wdctl /dev/watchdog* outputs
    
    Resolves: #1561
    
diff --git a/sos/plugins/watchdog.py b/sos/plugins/watchdog.py
index a3bdef05..43d457b2 100644
--- a/sos/plugins/watchdog.py
+++ b/sos/plugins/watchdog.py
@@ -85,4 +85,8 @@ class Watchdog(Plugin, RedHatPlugin):
 
         self.add_copy_spec(log_files)
 
+        # Get output of "wdctl <device>" for each /dev/watchdog*
+        for dev in glob('/dev/watchdog*'):
+            self.add_cmd_output("wdctl %s" % dev)
+
 # vim: set et ts=4 sw=4 :
```

I add my Signed-off-by and change the commit ID if I fix up a little error like that during merge.

> That may be overkill here, and /dev/watchdog* sufficient for this purpose.

I don't think this is appropriate in sos. It's possible to rename all kinds of devices in weird and wonderful ways, and we collect the configuration that is supposed to control that - if someone goes beyond that and deliberately makes the system obtuse then that may hamper our efforts at first fault troubleshooting, but it's a reasonable compromise between making sos more complex, and slowing down runs for all of our users - there is a cost fo that kind of set up, after all.

kgaillot: Whoops, I was following links from elsewhere and didn't notice this one was closed. Looks good.

bmr-cymru: No worries.. GitHub kinda does some strange things with review links :-)

bmr-cymru: And for the other problem - one of the things we've been doing over recent years is to add more and more of `/sys` to the report - the hope being that even if we aren't collecting the watchdog named `/dev/omgsecretnotevenadogdog" watchdog, we should have some evidence that it existed... 

So thinking about your comments and what ClusterLabs is doing, it is  probably worth our collecting the watchdog class from there too...

PR #1562, opened 2019-02-08
[postgresql] Use postgres 10 scl if installed

TurboTurtle:
Updates the plugin to check for the presence of the PostgreSQL10 scl
package and if present, use that instead of the 9.5 scl.

This is primarily aimed at RHV environments as 4.3 and later use version
10.

---

### DIFF
--- a/sos/plugins/postgresql.py
+++ b/sos/plugins/postgresql.py
@@ -80,13 +80,24 @@ class PostgreSQL(Plugin):
 
 class RedHatPostgreSQL(PostgreSQL, SCLPlugin):
 
-    packages = ('postgresql', 'rh-postgresql95-postgresql-server', )
+    packages = (
+        'postgresql',
+        'rh-postgresql95-postgresql-server',
+        'rh-postgresql10-postgresql-server'
+    )
 
     def setup(self):
         super(RedHatPostgreSQL, self).setup()
 
-        scl = "rh-postgresql95"
         pghome = self.get_option("pghome")
+
+        scl = None
+        for pkg in self.packages[1:]:
+            # The scl name, package name, and service name all differ slightly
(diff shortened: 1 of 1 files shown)


sos/plugins/postgresql.py
```diff
@@ -80,12 +80,19 @@ def setup(self):
...
+    )
 
     def setup(self):
         super(RedHatPostgreSQL, self).setup()
 
-        scl = "rh-postgresql95"
+        if self.is_installed('rh-postgresql10-postgresql-server'):
+            scl = 'rh-postgresql10'
+        else:
+            scl = 'rh-postgresql95'
```
bmr-cymru: Not hugely keen on this - can't we extend the `SCLPlugin` mechanism so that when a named instance (e.g. `rh-postgresql10-postgresql-server`) is present, that this information is exposed to the plugin somehow? That way we don't need to open-code version variation in every plugin that uses this.

pmoravec: The plugin already has that method - the test can be safely replaced by

    if 'rh-postgresql10' in self._get_scls():

(but for future with more postgres versions in SCLs, we would have to come with smarter solution, like "give me highest/newest installed SCL version of postgres")

TurboTurtle: Actually since #1567 was merged, we can do this based on the *active* scl, which is the desired end result anyways. I'll update this shortly.

TurboTurtle:
I've tested the latest revision against RHV 4.2 and 4.3, which make use of the differing SCL versions (while 4.3 will have both installed) and the `postgresql` plugin now collects a `pgdump` using the correct SCL.

bmr-cymru:
Thanks!

Issue #1550, opened 2019-01-30
Have sosreport collect /etc/environment.


alexmcan:

BryanQuigley:
or systemctl show-environment, any idea which is more useful?

alexmcan:
systemctl show-environment doesn't show what's in in /etc/environment

`ubuntu@ldscert:~$ cat /etc/environment
PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games"
no_proxy=127.0.0.1,localhost
NO_PROXY=127.0.0.1,localhost

ubuntu@ldscert:~$ systemctl show-environment
LANG=en_US.UTF-8
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`

PR #1564, opened 2019-02-14
[pam] collect /etc/environment

BryanQuigley:
Resolves: #1550

---

### DIFF
--- a/sos/plugins/system.py
+++ b/sos/plugins/system.py
@@ -22,6 +22,7 @@ class System(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
             "/proc/sys",
             "/etc/sysconfig",
             "/etc/default",
+            "/etc/environment",
         ])
 
         # FIXME: provide a a long-term solution for #1299


bmr-cymru:
Why `pam`? That seems a pretty arbitrary place for this. I would have thought that `system` were the more natural fit, given that this is system-wide environment settings (and the packaging history of the file on various distros - it's not "owned" by packages that provide binaries typically, e.g. on Red Hat it is part of the (ancient) historical `setup` RPM package).

BryanQuigley:
I was kindof surprised myself by pam - I just found it mentioned in the man page - https://manpages.ubuntu.com/manpages/bionic/man8/pam_env.8.html

I tried it on both Ubuntu and Fedora cause I wasn't quite sure.  Happy to change it to system, but figured I should default to where man pages says it's from..

bmr-cymru:
That just says that `/etc/environment` is the default environment file that `pam_env` will read - not that it owns it, or that it "belongs" to the PAM project.

Issue #1457, opened 2018-10-20
add systemd-resolve/resolvectl commands


BryanQuigley:
systemd-resolve --statistics
systemd-resolve --status

for bonus points test against some domain.. like example.com

I plan to build up a list of commands I'd like to see as I debug a DNSSEC issue, and then I'll make it into a pull request.

PR #1565, opened 2019-02-14
[systemd] Add many other lists, statuses, *ctls

BryanQuigley:
systemctl
 list-jobs - any currently runnign
 list-dependencies - nice dependency tree
 list-timers --all - what timers are setup/active
 list-machines - for systemd container orcestration

systemd-resolve
--status - what link is using what dns server, etc
--statistics  -  how cache/DNSSEC/etc are doing

busctl
 list - shows what's on the dbus (but likely future proof if new bus)
 status - shows systemds status on the bus

bootctl (primarily for the systemd built-in boot)
 list - boot loader entries if using the spec
 status - brief info about the system boot

I'm leaving other *ctls alone like locatectl and networkctl.
But we likely need to figure out how to handle them as well.

Resolves: #1457

---


bmr-cymru:
I don't think all of this belongs in the main `systemd` plugin: jobs, deps, and timers certainly. Containers possibly but needs a bit of thought if there isn't a better location.

For the ancillary stuff - especially where it is optional and not often used - I think it belongs elsewhere. This is especially the case for `systemd-boot` since it is not widely used by any of our major distros (or wasn't, last I checked).

For `busctl`, we already have a `dbus` plugin - I think this is a better fit, even if we were to consider renaming it to something more generic that acknowledges that different system busses may exist.

BryanQuigley:
Dropped busctl/bootctl.  

list-machines output also reports on the host so I think it makes sense:
 sudo systemctl list-machines
  NAME           STATE    FAILED JOBS
● desktop (host) degraded      2    0

1 machines listed.

systemd-resolve is still a bit of a gray area, but I think it makes more sense here as opposed to making a new plugin that also triggers when systemd package is installed.  

I was surprised by how useful bootctl status output was even if you aren't using it:
sudo bootctl status
systemd-boot not installed in ESP.
System:
     Firmware: n/a (n/a)
  Secure Boot: disabled
   Setup Mode: setup

Current Boot Loader:
      Product: n/a
     Features: ✗ Boot counting
               ✗ Menu timeout control
               ✗ One-shot menu timeout control
               ✗ Default entry control
               ✗ One-shot entry control
          ESP: n/a
         File: └─n/a

Available Boot Loaders on ESP:
          ESP: /boot/efi (/dev/disk/by-partuuid/acb3bed4-d687-4160-aef6-ec4741b0d206)
         File: └─/EFI/BOOT/BOOTX64.EFI

Boot Loaders Listed in EFI Variables:
        Title: ubuntu
           ID: 0x0000
       Status: active, boot-order
    Partition: /dev/disk/by-partuuid/acb3bed4-d687-4160-aef6-ec4741b0d206
         File: └─/EFI/UBUNTU/SHIMX64.EFI

0 entries, no entry could be determined as default.

This doesn't make sense to me to add to boot.py, but a bunch of plugins that all depend on just the systemd package seems odd as well.

PR #1568, opened 2019-02-15
[services] Don't collect chkconfig on RHEL 7 or newer

TurboTurtle:
RHEL 7 and newer do not use chkconfig any more, so we should not pollute
the archive root with a chkconfig call on those systems.

---

### DIFF
--- a/sos/plugins/services.py
+++ b/sos/plugins/services.py
@@ -36,7 +36,9 @@ class RedHatServices(Services, RedHatPlugin):
 
     def setup(self):
         super(RedHatServices, self).setup()
-        self.add_cmd_output("/sbin/chkconfig --list", root_symlink="chkconfig")
+        if self.policy.dist_version() <= 6:
+            self.add_cmd_output("/sbin/chkconfig --list",
+                                root_symlink="chkconfig")
 
 
 class DebianServices(Services, DebianPlugin, UbuntuPlugin):


TurboTurtle:
For the inevitable "why not just remove this entirely?" question. I remember this being brought up before, and there was some resistance to pulling it out entirely (though I can't remember why), so with this I'm splitting the difference and still having it in the plugin, but skipping over it for recent RHEL releases.

k-hagio:
There are some 3rd-party services using SysV format even on RHEL7 or newer, so `chkconfig --list` is still helpful to get an overview of setting of such services. If we mind the root mess, can we call it without `root_symlink` on RHEL7 or newer?

TurboTurtle:
Any specific examples? The Init System Wars (tm) pretty efficiently killed any SysV presence on systemd-based distros and even went so far as to forcibly redirect commands like `service` to systemd. The init scripts might still function, but they're certainly not given serious thought anymore - and at least on RHEL it's not something that's supported.

bmr-cymru:
Why is the root symlink so offensive? I don't really get that...

>  The init scripts might still function, but they're certainly not given serious thought anymore 

Then they may also mis-function, and that's a reason to capture them for diagnosis. This is cheap, long-standing, and for me doesn't really come close to meeting the bar for adding policy-dependent version checks in a plugin.

If there is some harm I am not getting, please explain it slowly using simple words ;-)

TurboTurtle:
Closing this after IRC discussion.

PR #1569, opened 2019-02-15
[docker] Don't run docker commands when the service is not running

TurboTurtle:
Prevents the docker plugin from running docker commands when the service
is not running. This will prevent us from collecting the same 'docker is
not running' error message over and over again from the docker commands
that would otherwise get run.

Note this depends on #1567 

---

### DIFF
--- a/sos/plugins/docker.py
+++ b/sos/plugins/docker.py
@@ -33,6 +33,14 @@ class Docker(Plugin):
             "/var/lib/docker/repositories-*"
         ])
 
+        self.add_journal(units="docker")
+        self.add_cmd_output("ls -alhR /etc/docker")
+
+        if not self.service_is_running('docker'):
+            # if docker is not running none of the below commands will provide
+            # any useful output
+            return
+
         subcmds = [
             'events --since 24h --until 1s',
             'info',
@@ -52,9 +60,6 @@ class Docker(Plugin):
         # separately grab ps -s as this can take a *very* long time
         if self.get_option('size'):
             self.add_cmd_output('docker ps -as')
-
-        self.add_journal(units="docker")
-        self.add_cmd_output("ls -alhR /etc/docker")
 
         nets = self.get_command_output('docker network ls')
 


sos/plugins/docker.py
```diff
@@ -33,6 +33,14 @@ def setup(self):
...
         ])
 
+        self.add_journal(units="docker")
+        self.add_cmd_output("ls -alhR /etc/docker")
+
+        if not self.service_is_running('docker'):
+            # if docker is not running none of the below commands will provide
+            # any useful output
+            return
+
```
bmr-cymru: As we've discussed before, I am really not big on adding more conditionality to plugins like this. It is going to make our lives harder in the future as we try to move to a more declarative plugin format, and to enable new kinds of automatic testing, validation, dry-runs etc.

Let's have a discussion on these gating mechanisms in general first - I've already stated my preference for handling them, and I don't see anything in this current change set that could not be accommodated by that.

bmr-cymru: Ah, see - I was looking for `is_service_running()` - missed this. Maybe we should rename it, since all our other service methods follow the `is_service...()` pattern?

bmr-cymru: Umm:

```
$ grep -r service_is_running sos/
sos/plugins/docker.py:        if not self.service_is_running('docker'):
sos/plugins/juju.py:def ensure_service_is_running(service):
sos/plugins/juju.py:    @ensure_service_is_running("juju-db")
```

:-)

TurboTurtle:
```python
    def is_service(self, name):
        '''Does the service $name exist on the system?'''
        return self.policy.init_system.is_service(name)

    def service_is_enabled(self, name):
        '''Is the service $name enabled?'''
        return self.policy.init_system.is_enabled(name)

    def service_is_disabled(self, name):
        '''Is the service $name disabled?'''
        return self.policy.init_system.is_disabled(name)

    def service_is_running(self, name):
        '''Is the service $name currently running?'''
        return self.policy.init_system.is_running(name)
```

With #1567 the above is the current naming for the `InitSystem` checks. Didn't notice that `juju` was already doing it's own thing.

Should we switch to `is_service_` instead of `service_is_`? We only have `is_service()` currently following the 'is_' pattern, but that does match `is_installed()`, `is_module()`...

bmr-cymru: Oh, this depends on another PR? Nothing here indicates that -  if I pull this branch, I get a broken tree. That's kinda confusing. If you have commits that absolutely depend on other commits, then put them all on one branch, and have one PR for the whole thing (the service change isn't interesting in itself - it is just an enabling patch for this change).

bmr-cymru:
Merged in commit bc34d9b938f9cd1c53120fc8badc3dff600771d3

PR #1571, opened 2019-02-18
[composer] Collect sources info for all sources

TurboTurtle:
Adds collection of info on each source found by composer-cli. The
_get_blueprints() method has been made more generic to accommodate both
blueprints and sources.

---

### DIFF
--- a/sos/plugins/composer.py
+++ b/sos/plugins/composer.py
@@ -10,14 +10,14 @@ class Composer(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
 
     packages = ('composer-cli',)
 
-    def _get_blueprints(self):
-        blueprints = []
-        bp_file = self.get_cmd_output_now("composer-cli blueprints list")
-        if bp_file:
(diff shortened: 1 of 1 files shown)


sos/plugins/composer.py
```diff
@@ -10,14 +10,14 @@ class Composer(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
 
     packages = ('composer-cli',)
 
-    def _get_blueprints(self):
-        blueprints = []
-        bp_file = self.get_cmd_output_now("composer-cli blueprints list")
+    def _get_entries(self, cmd):
+        entries = []
+        bp_file = self.get_cmd_output_now(cmd)
```
pmoravec: Worth renaming "bp_file" to "entries_file" or similar? "bp_" prefix is confusing / refers to the previous "blueprint".

Otherwise +1 to the commit/PR.

TurboTurtle: Updated the PR. Originally thought it wasn't, but on a second read yeah it was a bit confusing.

PR #1573, opened 2019-02-21
[memory] Add lsmem collection

TurboTurtle:
Adds collection of lsmem output for all memory blocks to the memory
plugin.

---

### DIFF
--- a/sos/plugins/memory.py
+++ b/sos/plugins/memory.py
@@ -32,7 +32,8 @@ class Memory(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
         self.add_cmd_output([
             "free -m",
             "swapon --bytes --show",
-            "swapon --summary --verbose"
+            "swapon --summary --verbose",
+            "lsmem -a -o RANGE,SIZE,STATE,REMOVABLE,ZONES,NODE,BLOCK"
         ])
 
 # vim: set et ts=4 sw=4 :


bmr-cymru:
Apparently `ZONES` is not universally available - we may need to do some version checking here:

```$ lsmem --help

Usage:
 lsmem [options]

List the ranges of available memory with their online status.

Options:
 -J, --json           use JSON output format
 -P, --pairs          use key="value" output format
 -a, --all            list each individual memory block
 -b, --bytes          print SIZE in bytes rather than in human readable format
 -n, --noheadings     don't print headings
 -o, --output <list>  output columns
 -r, --raw            use raw output format
 -s, --sysroot <dir>  use the specified directory as system root
     --summary[=when] print summary information (never,always or only)

 -h, --help     display this help and exit
 -V, --version  output version information and exit

Available columns:
      RANGE  start and end address of the memory range
       SIZE  size of the memory range
      STATE  online status of the memory range
  REMOVABLE  memory is removable
      BLOCK  memory block number or blocks range
       NODE  numa node of memory

For more details see lsmem(1).
$ lsmem --version
lsmem from util-linux 2.30.2
```

TurboTurtle:
Upstream it looks like `ZONES` was added in `2.32`. It was backported to `2.23` on RHEL 7.

Another spot for predicates, but we'll probably want to split out `RedHatMemory` as well for the version handling.

PR #1577, opened 2019-02-25
[mysql] Collect galera cluster node state

jcastill:
This patch collects the galera mysql node state
stored in the following file:

/var/lib/mysql/grastate.dat

It contains details of the galera node like version,
uuid, seqno, and current value in that node of
safe_to_bootstrap. The idea is that for a galera
mysql cluster, we'll get this status file from
each node in the sosreport, and we'll have a fuller
picture of the issue.

---

### DIFF
--- a/sos/plugins/mysql.py
+++ b/sos/plugins/mysql.py
@@ -36,6 +36,7 @@ class Mysql(Plugin):
             "/var/log/mysql/mysqld.log",
             "/var/log/containers/mysql/mysqld.log",
             "/var/log/mariadb/mariadb.log",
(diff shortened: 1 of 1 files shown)


jcastill:
> So long as this "belongs" to MySQL, looks fine to me (which based on the path seems likely).
> 
I was in doubt at the beginning between making a new plugin for galera or adding it to mysql, but the comment that precedes the line I added made me think it was better to stick to this plugin.

> Are these a cluster type worth considering for sos-collector too?
I'll investigate about that. I haven't played with sos-collector yet so it may take me a bit of time to implement a patch.

PR #1579, opened 2019-02-25
[podman] Correct command collection, add volume info collection

TurboTurtle:
Updates the podman plugin to collect volume information such as 'volume
ls' and 'volume inspect' for every reported volume.

Additionally, corrects the inspect iteration to not potentially call
'podman logs' on image IDs, only container IDs.

Removes the attempt to collect a podman journal - podman does not log to
a journal unit.

Finally adds '--all' to the stats command, as podman requires this
option when not specifying specific containers.

---


TurboTurtle:
Updated the PR to include the change to use an image name in the inspect calls when appropriate, which was accidentally left off the first time.

PR #1580, opened 2019-02-26
[docker] Correct commands, collect volume info, gate system df

TurboTurtle:
Fixes the potential for 'docker logs' to be called against images and
adds volume collection for docker like #1579 does for podman.

Changes image inspections to use the image name, unless the name or tag
is 'none' in which case the image ID is used.

Gates collection of 'docker system df' behind the docker.size option as
this command can take a long time to run just like 'ps -as'.

Adds the upstream docker packaging to the packages list for the RH
family of distributions.

---


pmoravec:
+1

PR #1581, opened 2019-02-26
[plugins] collect journalctl with option --all

pmoravec:
To collect also binary and long lines.

Resolves: #1581

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -845,7 +845,7 @@ class Plugin(object):
         :param sizelimit: Limit to the size of output returned in MB.
                           Defaults to the value of --log-size.
         """
-        journal_cmd = "journalctl --no-pager "
+        journal_cmd = "journalctl --no-pager --all "
         unit_opt = " --unit %s"
         boot_opt = " --boot %s"
         since_opt = " --since %s"


bmr-cymru:
Err...

```Python
 813     def add_journal(self, units=None, boot=None, since=None, until=None,
 814                     lines=None, allfields=False, output=None, timeout=None,
 815                     identifier=None, catalog=None, sizelimit=None):
 816         """Collect journald logs from one of more units.
...
 833 
 834         :param allfields: A bool. Include all journal fields
 835                            regardless of size or non-printable
 836                            characters.
 837 
...
 876         if allfields:
 877             journal_cmd += " --all"
 878 
```

This has been under the caller's control for years - I don't think we want to force it on for everything, since it can inflate journal file sizes and make them harder to read.

What is the specific request where we are missing something?

The actual effect of `-a/--all` is:

```
       -a, --all
           Show all fields in full, even if they include unprintable characters or are very long.
```

Which does a little more than just show long lines or binary fields...

I've added a comment to the Red Hat bugzilla - for now it doesn't even specify any real units where this happens, or any practical support problems where this data would be necessary. There is just a phoney test binary that writes some binary goop to the journal.

In my limited testing this doesn't have a huge impact on journal size, but we know from experience that journal behavour varies a lot from system-to-system, so I think we need a bit more justification here before we consider a change on our side (and I would much prefer targeted changes to turn this on for units where it is useful than the splattergun approach).

pmoravec:
Mea culpa, havent realized `--all-logs` offers this capability and also the original concern is bit different. Closing this PR.

Issue #284, opened 2014-05-16
Allow filtering log files by timestamp


bmr-cymru:
Extend add_copy_spec() interface for time filtering:
- Try to filter log time stamps for specified time range
- On failure to match or other error fall back to inode mtime

pmoravec:
I would like to review this improvement (and ideally implement it (by myself)):

- should there be just a new option to `add_copy_spec()` or is it worth splitting the method to "copy spec" (mainly configs that should be collected all) and "copy logs" (that shall be more filtered based on time/size)? I was inclined to the split but now I feel it won't get much benefit other than methodological.

- two options should be added?
  - "older than": collect only logs older than given number of (days?)
  - newer than": collect only logs newer than given number of (days?)
  - I think we should keep the specification of interval very simple, error-prone, and # days is a worth compromise here (maybe `pcp` or few others require hours?)
  - we will filter files against their `mtime`

- if both `sizelimit` and `[older|newer]_than` filters applied, *both* will be applied altogether (if one wants to collect logs from past week regardless of size, use `sizelimit=0`).

pmoravec:
Closing in favour of https://github.com/sosreport/sos/pull/1747

PR #1586, opened 2019-03-10
[plugins] Allow filtering log files by mtime

pmoravec:
`add_copy_spec` method extended by maxage and minage parameters.

If maxage (or minage) is specified, only the files with mtime
not older than (or not newer than, respectively) than the given
number of hours are collected.

Resulting list of files (optionally such filtered) is still ordered
by mtime from newest to oldest (to still fulfill #1082).

Resolves: #284

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -557,10 +557,13 @@ class Plugin(object):
     def _add_copy_paths(self, copy_paths):
         self.copy_paths.update(copy_paths)
 
-    def add_copy_spec(self, copyspecs, sizelimit=None, tailit=True):
-        """Add a file or glob but limit it to sizelimit megabytes. If fname is
-        a single file the file will be tailed to meet sizelimit. If the first
-        file in a glob is too large it will be tailed to meet the sizelimit.
+    def add_copy_spec(self, copyspecs, sizelimit=None, maxage=None,
+                      tailit=True):
+        """Add a file or glob but limit it to sizelimit megabytes. Collect
+        files with mtime not older than maxage hours.
+        If fname is a single file the file will be tailed to meet sizelimit.
+        If the first file in a glob is too large it will be tailed to meet
+        the sizelimit.
         """
         if sizelimit is None:
             sizelimit = self.get_option("log_size")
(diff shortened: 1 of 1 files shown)


BryanQuigley:
Haven't tested this yet, but I love the maxage concept.

It could let us drastically simplify (and maybe merge Debian/Ubuntu <> Red Hat) for the logs plugin (and my current proposed #1519). The problem I have with the current log_days is it really only can apply to systemd systems.  This would allow that variable to apply to all (albeit imperfectly).

What's the use case for minage?  
Did you check all uses for a positional argument to tailit= ?  We have call that do positional or the new style.

pmoravec:
Use case for minage: cant imagine *now*, but theoretically we might not be interested in some very verbose huge logs by a program that got mad and started to log tons of stuff recently - and we would be interested in bit older logs only (though this will need some support from command line / plugin args, if someone will require it). Anyway if there are objections to minage, I am fine to drop it from the PR.

Havent checked wrt tailit but dont see an issue..?

BryanQuigley:
Yea if we don't have a usecase for minage now, I'd drop it.  If you are thinking you want to implement the the commandline/plugin args next though - then we can leave it obviously...

tailit, Right now it is:
add_copy_spec(self, copyspecs, sizelimit=None, tailit=True)

so someone could write it like this
self.add_copy_spec("somelog", sizelimit, True)

If anyone did the above and didn't specify tailit=True, it's going to have an issue with the order change.
add_copy_spec(self, copyspecs, sizelimit=None, maxage=None, inage=None, tailit=True)

I did a quick once over and didn't see any issues, but you really would want to check the every add_copy_spec run, or the easy way out and put the new age bits after tailit.

pmoravec:
OK, `minage` does not have usage atm so I am dropping it.

All places where `tailit` is set to nondefault `False` set it via `tailit=False`, so no regression should happen. And I think it makes sense to group size+age limit options together among the method arguments, and even then have the `tailit` option. So I am leaving the positional ordering as is. Anyway, good point, havent realized this.

bmr-cymru:
This is only doing filtering by file `mtime` - i.e. accept or reject a whole file at a time. The original issue is asking for something slightly different - to be able to collect log data for a span of time. This seem to implement the fallback proposed, but not the actual per-line timestamp filtering:

* Try to filter log time stamps for specified time range
* On failure to match or other error fall back to inode mtime

Maybe it's enough but it's not the full scope of the original request (I think it was forked from an RHBZ, but I don't have it to hand).

bmr-cymru:
The patch looks OK, but am I right that nothing is using it right now? What is the intent here - to leave it to plugin authors (in which case, having this in 3.7 won't change anything), or to plumb this through to a command line option?

I would like to see in-file filtering, but I think we can add that later by expanding the definition of `maxage` (e.g. if it's an integer, it's log days, if it's a string that parses as a date then it's an absolute start time or offset).

pmoravec:
Yep, no usage in plugins now, just the option for further changes.

I agree with the future extension as well.

pmoravec:
Closing in favour of #1678.

PR #1595, opened 2019-03-19
[systemd] avoid collecting /dev/null

BryanQuigley:
Avoid "collecting" /dev/null as part of the systemd plugin.

In my testing that was the only plugin that brought it in.

Creating these character devices makes it harder to manage
(delete) extracted sosreports without more permissions.

This supersedes #1469 which removed the ability to "collect"
char devices.  As we only have one example today, let's just
exclude that.

---

### DIFF
--- a/sos/plugins/systemd.py
+++ b/sos/plugins/systemd.py
@@ -71,5 +71,6 @@ class Systemd(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
             "/etc/modules-load.d/*.conf",
             "/etc/yum/protected.d/systemd.conf"
         ])
+        self.add_forbidden_path('/dev/null')
 
 # vim: set et ts=4 sw=4 :


pmoravec:
Won't it be better to put it to a global blacklist to automatically prevent collecting it by any plugin?

BryanQuigley:
I still can't find a global blacklist..  I thought it wasn't implemented yet.  Where am I missing it?

bmr-cymru:
Pure vapourware right now :-D

bmr-cymru:
Did you ever find what was pulling it in? Presumably there is a symlink in one of systemd's link forests that points to `/dev/null`?

BryanQuigley:
Thanks for pulling!

/dev/null symlink is one way to mask a systemd unit:
"If a unit file is empty (i.e. has the file size 0) or is symlinked to /dev/null, its configuration will not be loaded and it appears with a load state of "masked", and cannot be activated. Use this as an effective way to fully disable a unit, making it impossible to start it even manually."
https://www.freedesktop.org/software/systemd/man/systemd.unit.html

Issue #1203, opened 2018-02-03
[apache] respect --log-size and collect logrotated logs


pmoravec:
Apache plugin has two issues:

- logrotated `access` and `error` logs are not collected (even I spotted few issues where I lacked them)
- bit more controversial: the plugin does not honour `--log-size`, it either collects all logs (with no limit) or 5M (RedHat) / 15M (Debian).

I would suggest the policy+plugin hardcoded defaults are removed and plugin relies on global `--log-size` setting (with 10M default). That is compliant with all other plugins.

Alternatives to my approach:
- have current limits and ignore `--log-size` for `apache` plugin (is this plugin so special?)
- use `--log-size` value *only* if explicitly specified, otherwise fallback to 5M/15M - I explored this option, but currently it is imho impossible to detect if a cmd option was set or not
- optionally align the 5M/15M values to the same hardcoded value (is there some rationale to have them different?)

Any thoughts from @bmr-cymru (who set both the limits) or from @battlemidget (if it would be ok to decrease Ubuntu default limit from 15M to 10M)?

pmoravec:
ping @bmr-cymru and @battlemidget if you have some preferences?

If not, I would vote removing the `apache` specifics and use default `--log-size` setting (with 10M default).

PR #1601, opened 2019-03-20
[apache] honour --log-size

pmoravec:
Honour --log-size option for collecting logs and dont use the
hardcoded 5/15MB size limit.

Resolves: #1203

---

### DIFF
--- a/sos/plugins/apache.py
+++ b/sos/plugins/apache.py
@@ -51,24 +51,23 @@ class RedHatApache(Apache, RedHatPlugin):
 
         self.add_forbidden_path("/etc/httpd/conf/password.conf")
 
-        # determine how much logs to collect
-        self.limit = None if self.get_option("all_logs") else 5
-
         # collect only the current log set by default
-        self.add_copy_spec("/var/log/httpd/access_log", self.limit)
-        self.add_copy_spec("/var/log/httpd/error_log", self.limit)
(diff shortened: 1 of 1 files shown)


sos/plugins/apache.py
```diff
@@ -25,6 +25,10 @@ def setup(self):
         # collect list of installed modules
         self.add_cmd_output(["apachectl -M"])
 
+        # determine how much logs to collect
+        self.limit = (None if self.get_option("all_logs")
+                      else self.get_option("log_size"))
```
TurboTurtle: Can't we just remove `self.limit` entirely from the plugin?

`add_copy_spec()` automatically checks for `--all-logs` and has the implicit `--log-size` otherwise:

```python
    def add_copy_spec(self, copyspecs, sizelimit=None, tailit=True):
        """Add a file or glob but limit it to sizelimit megabytes. If fname is
        """
        if sizelimit is None:
            sizelimit = self.get_option("log_size")

            sizelimit = None
```

pmoravec:
Good point, my tunnel view prevented me to see that :)

Reworking the PR..

pmoravec:
We might merge the `httpd[,22,24]` subdirs into a for cycle as we collect same 4 files from each, but:
- it will be less obvious what we collect
- in case different dirs will require different files to collect, I would vote for this explicit enumeration instead.

PR #1602, opened 2019-03-20
[samba] Add some additional useful information

cryptomilk:
---

This adds some additional commands which provide information which help when debugging.


bmr-cymru:
What's the rationale for the 2nd patch? Using long options is reasonable as a bit of extra documentation, but note that we don't make any guarantee currently that the commands will be executed in the order they are listed - if you must have that guarantee, there are other methods that can ensure it (e.g. `get_cmd_output_now()`). With the current thread pools, you will get commands executed in order, but it's not generally a good idea to rely on that.

I'll take it for now, but it would be good to understand the reasoning.

cryptomilk:
I don't need a guarantee that they are executed in that order and there is not need to execute them like that.

bmr-cymru:
OK, that's fine - thanks.

PR #1603, opened 2019-03-20
[travis] robustness, actually fail on errors

BryanQuigley:
This makes sure if we get errors (anything to stderr) in the build
it gets flagged.  Today, it can be easily silently ignored.

Current build is broken by recent lgtm pull.

3.4 python also appears to be broken by something else.

---


BryanQuigley:
This is a quick mockup of other changes I think we need to capture..

bmr-cymru:
Where are you seeing build breaks? Master looks OK to me according to current results, and I'm seeing a green light on the home page.

bmr-cymru:
OK, I'm with you now. Errors that don't currently cause a break.

BryanQuigley:
fixed the python3.4 build now.  thanks for the fixes to the collect errors.  all green now and more robust

bmr-cymru:
I split this up into two patches, as the changes are not really related in any way.

PR #1607, opened 2019-03-21
[manageiq] import os.environ variable properly

pmoravec:
os.environ is a variable not a module, so "import os.environ" doesnt
work, causing plugin isnt loaded.

Resolves: #1607

---


pmoravec:
Optionally, we can also

    from os import environ
    ..
        if "APPLIANCE_PG_DATA" in environ:
            pg_dir = environ["APPLIANCE_PG_DATA"]

pmoravec:
Updated to the "`from os import environ`" way that is used elsewhere.

PR #1613, opened 2019-03-22
[Plugin] Add SoSPredicate and predicates infrastructure

bmr-cymru:
Predicates add a new facility to the `Plugin` class that allow file, command, and string collection to be overridden (skipped) based on the predicate's value, without the need for conditional code in plugins themselves.

This is useful since it allows collection gating to be controlled from both within and outside the plugin; for example, implementing a global `--dry-run` mode (a long standing feature request previously needing a great deal of conditionals) becomes relatively trivial.

The current implementation allows collection to be gated based on the presence of kernel modules, running services, or the global `--dry-run` command line argument.

Plugins may define their own predicates, and pass them to methods as appropriate, or they may specify a default predicate (or a default command predicate) that will apply automatically to future operations.

When `--dry-run` is given all default plugin predicates are initialised to evaluate `False`.

Setting up and using a predicate is simple:

```python
    # Gating all collection by presence of a kernel module:
    self.set_predicate(SoSPredicate(self, kmods=["a_kernel_module"]))
```

```python
    # Gating all command collection by presence of a kernel module:
    self.set_cmd_predicate(SoSPredicate(self, kmods=["a_kernel_module"]))
```

```python
    # Gating a single command by presence of a service
    self.add_cmd_output("acmd", pred=SoSPredicate(self, services=["a_service"]))
```
```
---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -154,7 +154,7 @@ class SoSPredicate(object):
             pvalue |= self._owner.is_module_loaded(k)
 
         for s in self._services:
-            pvalue |= self._owner.is_service(s) and self._owner.is_enabled(s)
+            pvalue |= self._owner.service_is_running(s)
 
         return pvalue and not self._dry_run
 
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -749,7 +749,7 @@ class Plugin(object):
         """
         if not self.test_predicate(pred=pred):
             self._log_info("skipped copy spec '%s' due to predicate (%s)" %
-                           (copyspecs, pred))
+                           (copyspecs, self.get_predicate(pred=pred)))
             return
 
         if sizelimit is None:
@@ -884,7 +884,7 @@ class Plugin(object):
             self._log_info("added cmd output '%s'" % cmd)
         else:
             self._log_info("skipped cmd output '%s' due to predicate (%s)" %
-                           (cmd, pred))
+                           (cmd, self.get_predicate(cmd=True, pred=pred)))
 
     def add_cmd_output(self, cmds, suggest_filename=None,
                        root_symlink=None, timeout=300, stderr=True,
@@ -956,7 +956,7 @@ class Plugin(object):
 
         if not self.test_predicate(cmd=False, pred=pred):
             self._log_info("skipped string ...'%s' due to predicate (%s)" %
-                           (summary, pred))
+                           (summary, self.get_predicate(pred=pred)))
             return
 
         self.copy_strings.append((content, filename))
@@ -973,7 +973,7 @@ class Plugin(object):
 
         if not self.test_predicate(cmd=True, pred=pred):
             self._log_info("skipped cmd output '%s' due to predicate (%s)" %
-                           (exe, pred))
+                           (exe, self.get_predicate(cmd=True, pred=pred)))
             return None
 
         result = self.get_command_output(exe, timeout=timeout, stderr=stderr,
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: sos/plugins/__init__.py [hunk 1]
```diff
@@ -601,11 +723,16 @@ def get_option_as_list(self, optionname, delimiter=",", default=None):
...
         self.copy_paths.update(copy_paths)
 
-    def add_copy_spec(self, copyspecs, sizelimit=None, tailit=True):
+    def add_copy_spec(self, copyspecs, sizelimit=None, tailit=True, pred=None):
         """Add a file or glob but limit it to sizelimit megabytes. If fname is
         a single file the file will be tailed to meet sizelimit. If the first
         file in a glob is too large it will be tailed to meet the sizelimit.
         """
+        if not self.test_predicate(pred=pred):
+            self._log_info("skipped copy spec '%s' due to predicate" %
```
pmoravec: Maybe for future debugging purposes, log also the "guilty" predicate? (i.e.:

    self._log_info("skipped copy spec '%s' due to predicate '%s'" %
                                (copyspecs, self.get_predicate(pred)))

)?

sos/plugins/__init__.py
```diff
@@ -732,13 +859,17 @@ def _add_cmd_output(self, cmd, suggest_filename=None,
...
         _logstr = "packed command tuple: " + _tuplefmt
         self._log_debug(_logstr % cmdt)
-        self.collect_cmds.append(cmdt)
-        self._log_info("added cmd output '%s'" % cmd)
+
+        if self.test_predicate(cmd=True, pred=pred):
+            self.collect_cmds.append(cmdt)
+            self._log_info("added cmd output '%s'" % cmd)
+        else:
+            self._log_info("skipped cmd output '%s' due to predicate" % cmd)
```
pmoravec: Same applies here - worth mentioning the predicate?

sos/plugins/__init__.py
```diff
@@ -794,23 +926,34 @@ def _make_command_filename(self, exe):
...
+    def add_string_as_file(self, content, filename, pred=None):
         """Add a string to the archive as a file named `filename`"""
+
+        # Generate summary string for logging
+        summary = content.splitlines()[0]
+        if not isinstance(summary, six.string_types):
+            summary = content.decode('utf8', 'ignore')
+
+        if not self.test_predicate(cmd=False, pred=pred):
+            self._log_info("skipped string ...'%s' due to predicate" % summary)
```
pmoravec: .. and here :)

sos/plugins/__init__.py
```diff
@@ -794,23 +926,34 @@ def _make_command_filename(self, exe):
...
                            chroot=True, runat=None, env=None,
-                           binary=False, sizelimit=None):
+                           binary=False, sizelimit=None, pred=None):
         """Execute a command and save the output to a file for inclusion in the
         report.
         """
         start = time()
+
+        if not self.test_predicate(cmd=True, pred=pred):
+            self._log_info("skipped cmd output '%s' due to predicate" % exe)
```
pmoravec: .. and same applies here.

sos/plugins/__init__.py
```diff
@@ -840,7 +983,7 @@ def get_cmd_output_now(self, exe, suggest_filename=None,
     def is_module_loaded(self, module_name):
         """Return whether specified moudle as module_name is loaded or not"""
         if len(grep("^" + module_name + " ", "/proc/modules")) == 0:
-            return None
+            return False
         else:
             return True
```
pmoravec: Can't we simplify the method to single:

    return (len(grep("^" + module_name + " ", "/proc/modules")) > 0)

?

pmoravec:
Nice work, I like it!

My the only two (types of) comments inline - worth explicit mentioning predicate preventing some action, and one method can be simplified.

sos/plugins/__init__.py
bmr-cymru: I think we can!

PR #1616, opened 2019-03-22
[Plugin] Fix get_predicate evaluation and remove duplicate gating

TurboTurtle:
If a predicate evaluates False, we still need to return it in
get_predicate(), so update the check on if cmd_predicate is set
accordingly.

Predicate gating was done in both _add_cmd_output() and
get_cmd_output_now(), which meant that while _add_cmd_output() would
properly block commands from being added to the collection list if the
predicate evaluated False, get_cmd_output_now() would also block
_any_ commands from being collected that were added via an
add_cmd_output() call before the cmd_predicate was set.

This removes the predicate check from get_cmd_output_now() so we can
still collect command output that was added before a cmd_predicate was
set, e.g. journal units still get collected even if the service for that
unit is not running which would normally block commands from being
collected.

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -379,7 +379,7 @@ class Plugin(object):
             If no default predicate is set and a `pred` value is passed
             it will be returned.
         """
-        if cmd and self.cmd_predicate:
+        if cmd and self.cmd_predicate is not None:
             return self.cmd_predicate
         return pred or self.predicate
 
@@ -977,11 +977,6 @@ class Plugin(object):
         report.
         """
         start = time()
-
-        if not self.test_predicate(cmd=True, pred=pred):
-            self._log_info("skipped cmd output '%s' due to predicate (%s)" %
-                           (exe, self.get_predicate(cmd=True, pred=pred)))
-            return None
 
         result = self.get_command_output(exe, timeout=timeout, stderr=stderr,
                                          chroot=chroot, runat=runat,


bmr-cymru:
I'll look at this tomorrow, but that doesn't look like the right fix - the `not None` test is there for command to default precedence, and the `g_c_o_n()` gate is there to abort commands that call it directly - that needs a rethink, obviously, but it is intended.

bmr-cymru:
OK: the first part of the patch is correct. If a command predicate was set, then it should apply if `False`.

For the second part, it's a little more complex: we *do* want to evaluate the predicate if a plugin is calling `g_c_o_n()` directly, since we want to gate those commands and prevent their execution if the predicate is `False`, but we do not want to evaluate the predicate if it has already been evaluated (and found `True`) in the case of an internal call from `_collect_cmd_output()`. We can safely assume here that if we are processing a command then its predicate evaluated `True` since otherwise it would not have been added to the command list.

I can think of two ways to adress this, I'm somewhat more inclined to the first right now:

* Split `g_c_o_n()` into `get_command_output_now()` (called by user code in plugins), which evaluates the predicate, and which calls `_get_command_output_now()` which does not handle predicates. The latter would be called from `_collect_cmd_output()` and any other internal caller where predicate hooks have already been evaluated.

* Add an `evaluated` kwarg to `get_command_output_now()` that indicates whether predicates have already been evaluated for this call. This would be set by the internal callers, but should never be used by user plugin code.

I'm leaning toward the first because it makes less of a mess of the external `Plugin` API, and avoids callers mistakenly setting `evaluated` (they'd need to actually use the private method to bypass predicates).

bmr-cymru:
Actually there's another change needed in `get_predicate()`: the intended precedence is:

 * Explicit predicate passed as `pred` keyword argument
 * Implicit command default predicate if processing a command
 * Implicit plugin default predicate

So we should always return `pred` first if it has a non-None value.

Issue #1600, opened 2019-03-20
[plugin] Allow command collection to write to a plugin subdir


TurboTurtle:
TODO on me.

We have a few plugins - namely the containerization ones - that can very easily populate `sos_commands/<plugin>/` with hundreds (or thousands) of command output files. It would be beneficial if these could be written to subdirs in the plugin directory, so that everything can be organized a bit better. For example, in the `kubernetes` plugin we could have a subdir for each resource (there are a lot, and it's growing) so that it's easier to navigate for support engineers.

My initial thought is to just add a `subdir` kwarg to `add_cmd_output()` and if it's set, create and write to that directory under `sos_commands/<plugin>/`.

One thing we should address is - should this be limited to a single level of subdir like `sos_commands/<plugin>/foo`, or should we allow arbitrary amounts of subdirs e.g. `sos_commands/<plugin>/foo/bar/foobar`?

PR #1628, opened 2019-04-01
[Plugin] Allow plugins to write output to subdirs, and update container plugins

TurboTurtle:
Updates `add_cmd_output()` and the associated methods to allow plugins to specify if command output should be saved in a subdir within the plugin directory.

Additionally, update the container plugins to make use of this as these plugins are notorious for generating hundreds of files under the plugin directory, and this organization will help support representatives parsing an sosreport.

Fixes: #1600

---


TurboTurtle:
Rebased on master after merge of #1639

pmoravec:
This would be great for my planned #1663.

pmoravec:
+1, nice enhancement, works well in #1683.

PR #1629, opened 2019-04-02
[InitSystem] Cache service status information

TurboTurtle:
With the addition of predicates, there is the very real possibility that
plugins can end up gating multiple add_cmd_output() or similar calls
around the same predicate, such as service enablement.

Previously, this meant every single check of the predicate resulted in a
new InitSystem._query_service() call was made. Now, we only call
_query_service() once and save the results, since status is unlikely to
change during the course of plugin execution.

---

### DIFF
--- a/sos/policies/__init__.py
+++ b/sos/policies/__init__.py
@@ -124,17 +124,22 @@ class InitSystem(object):
         """Returns the status for the given service name along with the output
         of the query command
         """
+        _default = {
+            'name': name,
+            'status': 'missing',
+            'output': ''
+        }
+        if name not in self.services:
+            return _default
+        if 'status' in self.services[name]:
+            # service status has been queried before, return existing info
+            return self.services[name]
         svc = self._query_service(name)
         if svc is not None:
-            return {'name': name,
-                    'status': self.parse_query(svc['output']),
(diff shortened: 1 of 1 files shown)


bmr-cymru:
Wondering whether it's worth building any invalidation into this? E.g. having `sosreport` invalidate the cache after each phase?

TurboTurtle:
I don't think so, at least not at this point. If  $component is running or not during `setup()` I think it's reasonable to assume it wouldn't change states during a "normal" `collect()` phase.

If it does, then that should become all the more obvious from the resulting archive.

sos/policies/__init__.py
```diff
@@ -124,17 +124,22 @@ def get_service_status(self, name):
...
-                    'output': svc['output']
-                    }
-        else:
-            return {'name': name,
-                    'status': 'missing',
-                    'output': ''
-                    }
+            self.services[name]['status'] = self.parse_query(svc['output'])
+            self.services[name]['output'] = svc['output']
+            return self.services[name]
```
pmoravec: This changes return value type - 'name' is not further returned in the dict. But it makes sense to drop it, as 1) nobody is using it now, and 2) there is no reason to return also input argument.

Issue #1584, opened 2019-03-07
[RFE] add cloud-init config files 


shatadru:
We collect cloud-init log files after 5a72961 however we don't collect config files.

I propose a separate plug-in for cloud-init to copy /etc/cloud/ and /var/lib/cloud/ also collect /var/log/cloud-init via this plugin

bmr-cymru:
Ack (although that's the wrong commit reference - it's actually commit 36efd32ab01df2c076c3271543d674a515094375 in master). This should never have been added to the `logs` plugin in the first place - that is only for system logs, like (r)syslog and the journal.

shatadru:
Exactly, I can send a pull request for the same.

bmr-cymru:
Thanks - can just send a PR in future, if you have a proposed fix, no need for a separate issue.

PR #1630, opened 2019-04-03
[cloud_init] Add New Plugin

TurboTurtle:
Adds a new plugin for cloud-init.

Fixes: #1584

---


sos/plugins/cloud_init.py
```diff
@@ -0,0 +1,30 @@
...
+    '''
+
+    plugin_name = 'cloud_init'
+    packages = ('cloud-init',)
+    services = ('cloud-init',)
+
+    def setup(self):
+        self.add_copy_spec([
+            '/var/lib/cloud/',
+            '/etc/cloud/',
```
pmoravec: Assuming no secrets or similar in `/etc/cloud/` or `/var/lib/cloud/`.

PR #1634, opened 2019-04-04
[ipvs] Only run ipvsadm commands if the ip_vs kernel module is loaded

ptalbert:
Running any ipvsadm command can cause the kernel to autoload the
ip_vs module. So only run ipvsadm commands if the ip_vs kernel module
is found to already be loaded on the system.

---

### DIFF
--- a/sos/plugins/ipvs.py
+++ b/sos/plugins/ipvs.py
@@ -8,7 +8,7 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.plugins import Plugin, RedHatPlugin, DebianPlugin
+from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, SoSPredicate
 
 
 class Ipvs(Plugin, RedHatPlugin, DebianPlugin):
@@ -29,6 +29,6 @@ class Ipvs(Plugin, RedHatPlugin, DebianPlugin):
             "ipvsadm -Ln --stats",
             "ipvsadm -Ln --thresholds",
             "ipvsadm -Ln --timeout"
-        ])
+        ], pred=SoSPredicate(self, kmods=['ip_vs']))
 
 # vim: set et ts=4 sw=4 :


sos/plugins/ipvs.py
```diff
@@ -21,14 +21,15 @@ class Ipvs(Plugin, RedHatPlugin, DebianPlugin):
...
-        self.add_cmd_output([
-            "ipvsadm -Ln",
-            "ipvsadm -Ln --connection",
-            "ipvsadm -Ln --persistent-conn",
-            "ipvsadm -Ln --rate",
-            "ipvsadm -Ln --stats",
-            "ipvsadm -Ln --thresholds",
-            "ipvsadm -Ln --timeout"
-        ])
+        if self.is_module_loaded('ip_vs'):
```
TurboTurtle: As of 3.7 we have a new `SoSPredicate` feature that should be used for gating like this - it allows us to still have a history of when we don't run commands when the condition isn't met.

E.G.
```python
self.add_cmd_output([
     "ipvsadm -Ln",
     "ipvsadm -Ln --connection",
     "ipvsadm -Ln --persistent-conn",
     "ipvsadm -Ln --rate",
     "ipvsadm -Ln --stats",
     "ipvsadm -Ln --thresholds",
     "ipvsadm -Ln --timeout"
], pred=SoSPredicate(self, modules=['ip_vs]'))
```

ptalbert:
> Ack to the change, just need to use a predicate for gating.

Okay. Updated the commit to use the predicate thing and tested with and without the ip_vs module loaded. Works as expected for me.

bmr-cymru:
This looks a bit strange right now... Since the predicate gates literally *everything* the plugin does, why not use a kernel module trigger to enable the entire module instead?

ptalbert:
> This looks a bit strange right now... Since the predicate gates literally _everything_ the plugin does, why not use a kernel module trigger to enable the entire module instead?

The ipvs sos plugin has nothing to do if the ipvsadm program is not available however, it should not run the ipvsadm program unless the ip_vs kernel module is already loaded.

It is pretty common to have the ip_vs kernel module loaded and to not have the ipvsadm package/program installed. The most common use of the ip_vs module these days is with the keepalived service and at least with EL, the keepalived package does not depend on the ipvsadm package so it is very common to be using kernel ip_vs (via keepalived) and to not have the ipvsadm program installed on the system.

I guess it is slightly more simple to change the Ipvs() class to remove 'packages = ('ipvsadm')', set 'kernel_mods = ('ip_vs')', and then remove the kmods predicate from add_cmd_output. But then in many environments we'll record a bunch of 'command not found' output.

Personally, I think leaving it the way it is makes the most sense but I don't mind changing it if that is what you guys prefer.

Patrick

PR #1635, opened 2019-04-04
[maas,mysql,npm,pacemaker,postgresql] fix plugopts data types

pmoravec:
With new enforcement of implicit data type for plugin options, the
plugopts must have proper data types of default values and plugins must work
with them accordingly (in tests or so).

Resolves: #1635

(currently, an attempt to set e.g. postgres password fails on type conversion)

---

### DIFF
--- a/sos/plugins/maas.py
+++ b/sos/plugins/maas.py
@@ -21,10 +21,10 @@ class Maas(Plugin, UbuntuPlugin):
 
     option_list = [
         ('profile-name',
-         'The name with which you will later refer to this remote', '', False),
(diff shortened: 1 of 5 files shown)


pmoravec:
Further to this, there is a dead code:

https://github.com/sosreport/sos/blob/master/sos/plugins/mysql.py#L59-L62

since both variables contain strings so neither can be `True`.

I am happy to update the condition but I am unsure how; currently, `-k mysql.dbpass` sets the value to `'True'` while `-a` does not change the default value `''`.

PR #1636, opened 2019-04-05
[ipvs] do not execute ipvsadm if ip_vs is not loaded.

miztake:
Signed-off-by: MIZUTA Takeshi <mizuta.takeshi@jp.fujitsu.com>

---

ipvsadm is executed by ipvs plugins.
ip_vs module is loaded automatically due to the execution of ipvsadm.
If ip_vs module is not loaded, sosreport should not execute ipvsadm.
This is the same idea as executing iptables.

### DIFF
--- a/sos/plugins/ipvs.py
+++ b/sos/plugins/ipvs.py
@@ -21,14 +21,15 @@ class Ipvs(Plugin, RedHatPlugin, DebianPlugin):
     packages = ('ipvsadm',)
 
     def setup(self):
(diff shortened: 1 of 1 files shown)


pmoravec:
I suggest using predicates we recently introduced in https://github.com/sosreport/sos/pull/1613.

miztake:
Thank you for your comment.
This issue is duplicate with #1634.
Therefore, we close this request.

PR #1637, opened 2019-04-06
[manageiq] added managiq/postgresql.conf.d/*.conf to manageiq log collection

rh-dluong:
---

### DIFF
--- a/sos/plugins/manageiq.py
+++ b/sos/plugins/manageiq.py
@@ -73,6 +73,8 @@ class ManageIQ(Plugin, RedHatPlugin):
         ])
         self.add_copy_spec("/var/log/tower.log")
 
+        self.add_copy_spec("/etc/manageiq/postgresql.conf.d/*.conf")
+
         if environ.get("APPLIANCE_PG_DATA"):
             pg_dir = environ.get("APPLIANCE_PG_DATA")
             self.add_copy_spec([


sos/plugins/manageiq.py
```diff
@@ -73,6 +73,8 @@ def setup(self):
         ])
         self.add_copy_spec("/var/log/tower.log")
 
+        self.add_copy_spec("/etc/manageiq/postgresql.conf.d/*.conf")
```
TurboTurtle: Is there nothing else in `/etc/manageiq/` that's worth collecting?

rh-dluong: Nope, it looks like it was only added to hold additional postgres confs as of the latest major release.  Thanks for the inquiry though!

bmr-cymru:
Please also follow the [contributor guidelines](https://github.com/sosreport/sos/wiki/Contribution-Guidelines) for patch formatting: the subject needs to be in `[<component>] change description` format, hard wrapped to 72 characters and with a brief and reasonable description of the change in the subject line.

E.g.:

```
[manageiq] add postresql configuration path

The most recent version of CloudForms adds postgres configuration
files at:

  /etc/managiq/postgresql.conf.d/

Include these in the set of files collected by the plugin.

Signed-off by: David Luong <dluong@redhat.com>
```

You can fix your commit with `git commit --amend`, and then re-push the branch with `git push --force`.

rh-dluong:
Sorry about that @bmr-cymru , I thought that subject was referring to the PR itself and not the commit message, will keep that in mind for the future, thanks!

sos/plugins/manageiq.py
bmr-cymru: In future, if you're adding a new item following an existing `add_copy_spec()` call, please just add it as a list - there's no need to make duplicate calls:

```python
                self.add_copy_spec([
                    "/var/log/tower.log",
                    "/etc/manageiq/postgresql.conf.d/*.conf"
                ])
```

It's more readable and makes the code easier to modify in future. I will fix this up as I merge in this case.

rh-dluong: Thanks for the pointers, I'll try and keep that in mind next time.

Issue #762, opened 2016-02-10
[sosreport] add pre-policy startup message


bmr-cymru:
Right now `sosreport` is completely silent until the policy class has been loaded and the package manager initialised. Especially on systems with a slow `/var` file system, or with many packages installed, this can mean a noticeable delay before the version banner appears:

```
# sosreport
[... you wait ... time passes ...]

sosreport (version 3.2)

[...]
```

This is because `SoSReport.print_header()` is called from `SoSReport.execute()`, which is called by the `sosreport` module's `main()` function immediately after creating the object:

``` python
def main(args):
    """The main entry point"""
    sos = SoSReport(args)
    sos.execute()
```

Since policy and package manager set up take place in the constructor the header is not written until all this work has been done.

I'm considering moving the header call into the constructor itself - before policy initialisation - right now the header function uses `ui_log.info` (i.e. it goes to `stdout`) so this can easily be changed to a simple `print()` call.

Putting this out in an issue to get some acks from the other distro maintainers - this change should only be user-visible in the case that the package manager is taking a long time to load (and it should improve the interactive experience in that case) but it is a change to a long-standing aspect of sosreport's behaviour.

Amitgb14:
```
diff --git a/sos/sosreport.py b/sos/sosreport.py
index ac0827b..a441e55 100644
--- a/sos/sosreport.py
+++ b/sos/sosreport.py
@@ -682,6 +682,8 @@ class SoSReport(object):
         self.sys_tmp = None
         self.exit_process = False
 
+        self.print_header()
+
         try:
             import signal
             signal.signal(signal.SIGTERM, self.get_exit_handler())
@@ -738,7 +740,7 @@ class SoSReport(object):
             self._exit(1)
 
     def print_header(self):
-        self.ui_log.info("\n%s\n" % _("sosreport (version %s)" %
+        sys.stdout.write("\n%s\n" % _("sosreport (version %s)" %
                                       (__version__,)))
 
     def get_commons(self):
@@ -1562,7 +1564,6 @@ class SoSReport(object):
     def execute(self):
         try:
             self.policy.set_commons(self.get_commons())
-            self.print_header()
             self.load_plugins()
             self._set_all_options()
             self._set_tunables()
```

Amitgb14:
Hi @bmr-cymru, I paste patch here, if that correct then I like to submit PR.

Amitgb14:
@bmr-cymru , Is there any suggestion?

PR #1644, opened 2019-04-10
[sosreport] Print header before policy initialization

TurboTurtle:
The version header is now printed before policy initialization as part
of the SoSReport() constructor instead of in the execute() message.

Fixes: #762

---


TurboTurtle:
It seemed like #762 had been forgotten about since we never got feedback from the other distros, and I'm just going through the low-hanging fruit right now. Since we didn't have any responses in 3 years, I'm assuming there's no objection to this.

PR #1645, opened 2019-04-10
[pulp] collect db and collection sizes

pmoravec:
Collect basic db stats and also collection sizes of mongo pulp_database

Resolves: #1645

---

### DIFF
--- a/sos/plugins/foreman.py
+++ b/sos/plugins/foreman.py
@@ -12,6 +12,8 @@
 from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin,\
                         SCLPlugin
 from pipes import quote
+from re import match
+import os
 
 
 class Foreman(Plugin):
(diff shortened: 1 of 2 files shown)


pmoravec:
I enhanced this private branch to fix:
- this issue/PR (to collect mongo DB stats in `pulp` plugin)
- #1649 (support external DBs in `pulp` and `foreman` plugins)

The reason is, the later commit for `pulp` plugin is building on the commit for this issue.

Merging the three commits / the branch will resolve both #1645 and #1649.

Currently, the #1649 has one drawback for pulp external DB: `sos.log` will contain credentials to the DB like:

    2019-04-15 16:44:40,001 INFO: [plugin:pulp] collecting output of 'bash -c 'mongo pulp_database --host ktordeur-sat65-mongo.sysmgmt.lan --port 27017 -u pulp_user -p pulp_password --eval "DBQuery.shellBatchSize=200;; db.task_status.find().sort({finish_time: -1}).pretty().shellPrint()"''

but I don't know if it is possible to obfuscate it e.g. via environment variable like `PGPASSWORD`.

PR #1651, opened 2019-04-17
[sosreport] initialize disabled plugins properly when parsing sos.conf

pmoravec:
opts.noplugins is referred when parsing "tunables" section, so
the variable must be set to empty list every time.

Resolves: #1651

---

### DIFF
--- a/sos/__init__.py
+++ b/sos/__init__.py
@@ -250,8 +250,8 @@ class SoSOptions(object):
                 optlist.extend(SoSOptions._opt_to_args(opt, val))
             opts._merge_opts(argparser.parse_args(optlist), is_default)
 
+        opts.noplugins = []
         if config.has_option("plugins", "disable"):
-            opts.noplugins = []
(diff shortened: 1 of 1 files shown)


pmoravec:
Having `/etc/sos.conf`:

    [plugins]
    #disable = rpm,selinux,dovecot
    
    [tunables]
    postgresql.dbport=54321

will raise:

      File "./sosreport", line 19, in <module>
        main(sys.argv[1:])
      File "/root/sos-master/sos/sosreport.py", line 1383, in main
        sos = SoSReport(args)
      File "/root/sos-master/sos/sosreport.py", line 281, in __init__
        self.fileopts = SoSOptions().from_file(parser, self.opts.config_file)
      File "/root/sos-master/sos/__init__.py", line 271, in from_file
         if not opt.split('.')[0] in opts.noplugins:
    TypeError: argument of type 'NoneType' is not iterable

pmoravec:
(a typo in commit message fixed)

PR #1654, opened 2019-04-18
[kubernetes] Enable Ubuntu kubectl cmd from any k8s nodes

slashdd:
[kubernetes] Enable Ubuntu kubectl cmd from any k8s nodes
    
    Enable Ubuntu kubectl command, delivered as a SNAP,
    on any k8s nodes using CDK installation.
    
    CDK stands for Charmed Distribution of Kubernetes.
    
---

### DIFF
--- a/sos/plugins/kubernetes.py
+++ b/sos/plugins/kubernetes.py
@@ -9,11 +9,11 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.plugins import Plugin, RedHatPlugin
+from sos.plugins import Plugin, RedHatPlugin, UbuntuPlugin
 from os import path
 
 
-class kubernetes(Plugin, RedHatPlugin):
+class kubernetes(Plugin, RedHatPlugin, UbuntuPlugin):
 
     """Kubernetes plugin
     """
@@ -26,7 +26,8 @@ class kubernetes(Plugin, RedHatPlugin):
     files = (
         "/var/run/kubernetes/apiserver.key",
         "/etc/origin/master/",
-        "/etc/origin/node/pods/master-config.yaml"
+        "/etc/origin/node/pods/master-config.yaml",
+        "/root/cdk/kubeproxyconfig"
     )
 
     option_list = [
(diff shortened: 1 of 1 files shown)


slashdd:
File output from a generated sosreport include this proposal change
https://pastebin.ubuntu.com/p/CdKsDPJnHh/

It works on master node and worker node as they both have kubectl and the kubeconfig.

davecore82:
Tested on CDK with kubernetes-master charm version 642 and kubernetes-worker charm version 508 and seems to work well on both. LGTM. I get kubectl outputs and it also picked up my /var/run/flannel/subnet.env flannel info.

https://pastebin.ubuntu.com/p/sBvx4nXsFT/

sos/plugins/kubernetes.py
```diff
@@ -59,10 +60,16 @@ def setup(self):
         if not self.check_is_master():
             return
 
+        # Red Hat
         kube_cmd = "kubectl "
         if path.exists('/etc/origin/master/admin.kubeconfig'):
             kube_cmd += "--kubeconfig=/etc/origin/master/admin.kubeconfig"
 
+        # Ubuntu Charmed Distribution of Kubernetes
+        if path.exists('/root/cdk/kubeproxyconfig'):
```
pmoravec: Isn't it better to set `self.kube_ctl` in explicit UbuntuKubernetes and RedHatKubernetes plugins?

bmr-cymru: I think that would be less confusing - right now there's a mixture of Ubuntu and Red Hat packaging and paths in the plugin, and the path that is chosen as the key to select the Red Hat `--kubeconfig` is not listed in the plugin files list.

slashdd:
Disregard that PR. A colleague of mine will submit an updated PR by end of week for the k8s plugin, separating Ubuntu inside its own class.

Thanks for the feedback.

bmr-cymru:
I already have a branch to do that :-)

I will merge this patch as-is, then push the amended branch for review - it needs some feedback from either side to ensure the path and package lists are still correct.

slashdd:
ok thanks

bmr-cymru:
Merged in commit 71aaa6acbf5f40e617180273cc0fee94151d2efc

bmr-cymru:
Merged in commit 71aaa6acbf5f40e617180273cc0fee94151d2efc

slashdd:
@niedbalski fyi ^

bmr-cymru:
Only question is whether the path test is still needed on Ubuntu, or if the snap path for `kube_cmd` is always valid.

slashdd:
the snap path for kube_cmd is always valid.

Issue #1653, opened 2019-04-18
[juju] juju plugins stop/start juju-db and mongod


slashdd:
The juju.py plugin need refactoring, IMHO, the plugin shouldn't try to stop/start the mongo database from the state/controller server such as juju-db, mongod.

If the process is stop/failed mode, ..., then exit or skip it. 
I don't think it's the responsibility of the juju plugin to take action on this.

Also juju-db is only found in the state-server AFAIK. Current plugin tries to stop/start juju-db on all machines where juju plugin is trigger by finding the juju binary which raise exception.

**Restart of mongod after a sosreport run:**
```
# pidof mongod
12622
# sosreport -a
# pidof mongod
1381
```

**Exception:**
```
Traceback (most recent call last):
  File "/usr/share/sosreport/sos/plugins/juju.py", line 22, in wrapped_f
    raise Exception("Cannot stop {0} service".format(service))
Exception: Cannot stop juju-db service
```

pmoravec:
Indeed; `sosreport` shouldn't restart a service and imho even start one. So whole https://github.com/sosreport/sos/blob/master/sos/plugins/juju.py#L16-L31 shall be imho removed, together with places where it is called (https://github.com/sosreport/sos/blob/master/sos/plugins/juju.py#L61). Rather, this restart command might be replaced by "`service juju-db status`" to let a user to understand why mongo exports failed (if db was down).

I will file PR for this, anyway I know nothing about `juju`. So some functional review from a domain expert would be welcomed.

PR #1655, opened 2019-04-20
[juju] dont restart juju-db service

pmoravec:
sosreport shouldnt restart a service. It can only collect it's status
to notify user in case the service is stopped.

Resolves: #1653

---

### DIFF
--- a/sos/plugins/juju.py
+++ b/sos/plugins/juju.py
@@ -13,24 +13,6 @@ from sos.plugins import Plugin, UbuntuPlugin
 from json import loads as json_loads
 
 
-def ensure_service_is_running(service):
-    def wrapper(callback):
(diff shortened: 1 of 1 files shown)


slashdd:
As an fyi. Canonical is currently working on a new juju plugin to address this bug and do a good rewrite of the plugin to support upcoming features.

dnegreira:
Hi,
We are currently working on refactoring the juju plugin and separating it in several modules, one of the refactoring is indeed dropping this weird behavior. I hope to have the plugin ready in the next few days.

dnegreira:
Hi,

PR is done here: https://github.com/sosreport/sos/pull/1670

pmoravec:
OK dropping this PR in favour of #1670. Thanks for pointing and the rework!

PR #1658, opened 2019-04-23
[apache] Add syntax test and runtime status information

slashdd:
'apachectl -t' (AKA configtest) parse the configuration files
and report information about particular syntax error.

'systemctl status <service>' show runtime status information
about apache unit.

---


slashdd:
```
ls sos_commands/apache/
apachectl_-M  apachectl_-S  apachectl_-t  systemctl_status_apache2
```
```
cat sos_commands/apache/apachectl_-t 
Syntax OK
```
```
cat sos_commands/apache/systemctl_status_apache2 
* apache2.service - The Apache HTTP Server
   Loaded: loaded (/lib/systemd/system/apache2.service; enabled; vendor preset: enabled)
  Drop-In: /lib/systemd/system/apache2.service.d
           `-apache2-systemd.conf
   Active: active (running) since Tue 2019-04-23 14:19:06 UTC; 1h 32min ago
 Main PID: 3943 (apache2)
    Tasks: 54 (limit: 4915)
   CGroup: /system.slice/apache2.service
           |-3943 /usr/sbin/apache2 -k start
           |-3945 /usr/sbin/apache2 -k start
           `-3946 /usr/sbin/apache2 -k start

Apr 23 14:19:06 bioniclxd systemd[1]: Starting The Apache HTTP Server...
Apr 23 14:19:06 bioniclxd systemd[1]: Started The Apache HTTP Server.
Apr 23 14:19:06 bioniclxd systemd[1]: apache2.service: Failed to reset devices.list: Operation not permitted
Apr 23 14:19:06 bioniclxd systemd[1]: apache2.service: Failed to reset devices.list: Operation not permitted
Apr 23 14:19:06 bioniclxd systemd[1]: apache2.service: Failed to reset devices.list: Operation not permitted
Apr 23 14:19:07 bioniclxd systemd[1]: apache2.service: Failed to reset devices.list: Operation not permitted
```

slashdd:
Test ran on Centos7
```
ls sos_commands/apache/
apachectl_-M  apachectl_-S  apachectl_-t  systemctl_status_httpd
```
```
cat sos_commands/apache/apachectl_-t
AH00558: httpd: Could not reliably determine the server's fully qualified domain name, using 127.0.1.1. Set the 'ServerName' directive globally to suppress this message
Syntax OK
```
```
cat sos_commands/apache/systemctl_status_httpd 
* httpd.service - The Apache HTTP Server
   Loaded: loaded (/usr/lib/systemd/system/httpd.service; disabled; vendor preset: disabled)
   Active: inactive (dead)
     Docs: man:httpd(8)
           man:apachectl(8)
```

sos/plugins/apache.py
```diff
@@ -59,6 +61,10 @@ def setup(self):
 
         self.add_forbidden_path("/etc/httpd/conf/password.conf")
 
+        # If systemd is present, show status information.
+        if path.isdir('/run/systemd/system'):
+            self.add_cmd_output('systemctl status httpd')
```
TurboTurtle: We're probably ok to just add the `add_cmd_output()` calls. If the command fails (or is missing), a_c_o already correctly handles that situation.

sos/plugins/apache.py
```diff
@@ -97,6 +103,10 @@ def setup(self):
             "/etc/default/apache2"
         ])
 
+        # If systemd is present, show status information.
+        if path.isdir('/run/systemd/system'):
+            self.add_cmd_output('systemctl status apache2')
```
TurboTurtle: Same here, we can just add the `systemctl` commands.

TurboTurtle:
This actually got me thinking about a wrapper for collecting this kind of output, so I've opened #1667 which could definitely help here to control for systemd vs. not systemd distros.

slashdd:
@TurboTurtle ack thanks for the input.

bmr-cymru:
I agree with @TurboTurtle here: there's no need to test for `systemd` here. That said, if you do need that kind of facility, it is strongly preferred to add it either to the plugin that needs it, or to the base `Plugin` class if it is of use to multiple plugins: "Don't Repeat Yourself".

pmoravec:
IMHO `apachectl -t` is rather *verification* than *collection* command and sosreport should "just" collect data (that can be analysed or post-processed locally anyhow). But similarly as in `sar` data (where we generate `sar*` human-readable output as well), since the command can depend on `apache version`, it is easier to run it directly on the system. So LGTM.

PR #1659, opened 2019-04-23
[lxd] adapting the plugin to support lxd 3.0.X series

slashdd:
Current plugin was outdated.

---

### DIFF
--- a/sos/plugins/lxd.py
+++ b/sos/plugins/lxd.py
@@ -18,19 +18,30 @@ class LXD(Plugin, UbuntuPlugin):
     profiles = ('container',)
     packages = ('lxd',)
 
+    # Version 2.0.X:
+    # - /etc/default/lxd-bridge
+    # - /var/lib/lxd/lxd.db
+    #
+    # Version 3.0.X:
+    # - /var/lib/lxd/database/local.db
+    # - /var/lib/lxd/database/global/*
+    # - lxd-bridge no longer exist.
+    #
     def setup(self):
         self.add_copy_spec([
+            "/etc/default/lxd-bridge",
             "/var/lib/lxd/lxd.db",
-            "/etc/default/lxc-bridge",
+            "/var/lib/lxd/database/local.db",
+            "/var/lib/lxd/database/global/*",
+            "/var/log/lxd/*"
         ])
 
-        self.add_copy_spec("/var/log/lxd*")
-
-        # List of containers available on the machine
         self.add_cmd_output([
-            "lxc list",
-            "lxc profile list",
             "lxc image list",
+            "lxc list",
+            "lxc network list",
+            "lxc profile list",
+            "lxc storage list"
         ])
 
         self.add_cmd_output([


stgraber:
I'm the upstream maintainer of LXD and this seems like it should work fine for LXD 3.0.x and higher.

slashdd:
At ubuntu perspective, only xenial still uses 2.0.X (but can have 3.0.X via xenial-backports). 
Starting Bionic/18.04LTS and late, only lxd 3.0.X is available.

At package level, I will keep the 2.0.X support in Xenial package and will update the other package bionic and late to support 3.0.X

```
 lxd | 2.0.0-0ubuntu4          | xenial                    | source, amd64, arm64, armhf, i386, powerpc, ppc64el, s390x
 lxd | 2.0.2-0ubuntu1~16.04.1  | xenial-security           | source, amd64, arm64, armhf, i386, powerpc, ppc64el, s390x
 lxd | 2.0.11-0ubuntu1~14.04.4 | trusty-backports/universe | source, amd64, arm64, armhf, i386, ppc64el
 lxd | 2.0.11-0ubuntu1~16.04.4 | xenial-updates            | source, amd64, arm64, armhf, i386, powerpc, ppc64el, s390x
 lxd | 3.0.0-0ubuntu4          | bionic                    | source, amd64, arm64, armhf, i386, ppc64el, s390x
 lxd | 3.0.3-0ubuntu1~16.04.1  | xenial-backports          | source, amd64, arm64, armhf, i386, ppc64el, s390x
 lxd | 3.0.3-0ubuntu1~18.04.1  | bionic-updates            | source, amd64, arm64, armhf, i386, ppc64el, s390x
 lxd | 1:0.4                   | cosmic                    | source, all
 lxd | 1:0.4.2                 | cosmic-proposed           | source, all
 lxd | 1:0.7                   | disco                     | source, all
 lxd | 1:0.7                   | eoan                      | source, all
```

BryanQuigley:
It's just a few extra copy lines, why not include both?   It fails really gracefully if nothing is there.

slashdd:
@BryanQuigley Yeah we could, the plugin prior to this proposal was a bit offset for 2.0.X version, but this can be addressed easily. It is `/etc/default/lxd-bridge` and not `/etc/default/lxc-bridge`.

I re-submit to have both 2.0.X and 3.0.X and yes, you are right it won't cause any harm if the files aren't found.

slashdd:
@BryanQuigley Done, sosreport collect 3.0.X and remain backward compatible with 2.0.X, as if it fails, it will gracefully if file not present.

bmr-cymru:
Right: @BryanQuigley is correct. We cannot remove the old versions until we are sure that all supported distros have moved on to the 3.x series.

slashdd:
@bmr-cymru sound good. My most recent push support both 2.0.X and 3.0.X.

bmr-cymru:
The current version of the patch has deleted a couple of items that seem to belong to version 2:

```
$ git show | grep '^- '
-            "/etc/default/lxc-bridge",
-        self.add_copy_spec("/var/log/lxd*")
-        # List of containers available on the machine
-            "lxc image list",
```

I get that `lxc-bridge` is gone in 3.0.0, but don't we want to keep that and `lxc image list` for the older versions?

slashdd:
@bmr-cymru 

`lxc image list` has been preserved in eab9d10.
`/etc/default/lxc-bridge`, has been remove on purpose because it is now outdated in both 2.X and 3.X.

Confirmation:
```
$ lsb_release -cs
xenial

$ dpkg -l | grep -i lxd
ii  lxd                              2.0.11-0ubuntu1~16.04.4                    amd64        Container hypervisor based on LXC - daemon
ii  lxd-client                       2.0.11-0ubuntu1~16.04.4                    amd64        Container hypervisor based on LXC - client

$ ls /etc/default/lx*bridge
/etc/default/lxd-bridge
```

bmr-cymru:
Ah OK, it just got shuffled around - thanks for confirming.

Issue #1656, opened 2019-04-23
3.7 minor error from lstopo.py


BryanQuigley:
Warning on run:
plugin lstopo does not install, skipping: No module named 'distutils.spawn'

Issue found on Ubuntu 18.04 LTS in an LXD container.  (Not an issue on 16.04)

@slashdd 3.7 PPA testing (but I believe this is upstream..)

slashdd:
I haven't read the kvm plugin yet, but if the plugin is meant to run on both hypervisor and guest. Then we could trigger the plugin to run based on some detection pattern. For the hypervisor if `/usr/bin/kvm` exist and KVM guest using `systemd-detect-virt` if return qemu.

slashdd:
That problem rings a bell to me. Looking back at what I did in 3.7 PPA, I noticed I have removed a debian/patches/avoid.distutils.diff which prevent using distutils during runtime.

I don't know the background behind this Ubuntu fix, which btw doesn't seems to have been upstream'd, more investigation in that area will be needed, but I don't see why I would keep that fix if it differs from upstream, if I can only add a Depends: python3-distutils at runtime to make the upstream code working.
 
```
Index: sosreport/sos/plugins/lstopo.py
===================================================================
--- sosreport.orig/sos/plugins/lstopo.py
+++ sosreport/sos/plugins/lstopo.py
@@ -7,7 +7,7 @@
 # See the LICENSE file in the source distribution for further information.
 
 from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
-from distutils.spawn import find_executable
+from shutil import which
 
 
 class Lstopo(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
@@ -22,7 +22,7 @@ class Lstopo(Plugin, RedHatPlugin, Debia
         # binary depends on particular package, both require hwloc-libs one
         # hwloc-gui provides lstopo command
         # hwloc provides lstopo-no-graphics command
-        if find_executable("lstopo"):
+        if which("lstopo"):
             cmd = "lstopo"
         else:
             cmd = "lstopo-no-graphics"
```

PR #1660, opened 2019-04-23
[lstopo] Fix to not depend on distutils

BryanQuigley:
This adds support for Ubuntu and drops the depends
on distutils replacing with the sos utility.

I believe the functionality should be the same, and
appears to work on Ubuntu.

Closes: #1656

---


slashdd:
LGMT. This will indeed avoid to add 'Depends' for python3-distutils at runtime.

pmoravec:
Closed this in 46fc58c .

PR #1661, opened 2019-04-25
[openstack_instack] add ansible.log

EmilienM:
Collect /var/lib/mistral/config-download-latest/ansible.log which is an
important log to be able to debug issues with Ansible playbooks.

/var/lib/mistral/config-download-latest is a directory that exists
anyway on the undercloud and is the place where the ansible logs is
stored.

Note: we don't want to collect the whole /var/lib/mistral because it
contains sensitive informations like username/passwords/endpoints.

rhbz#1702806
rhbz#1702802

Signed-off-by: Emilien Macchi <emilien@redhat.com>

### DIFF
--- a/sos/plugins/openstack_instack.py
+++ b/sos/plugins/openstack_instack.py
@@ -22,7 +22,8 @@ NON_CONTAINERIZED_DEPLOY = [
 CONTAINERIZED_DEPLOY = [
         '/var/log/heat-launcher/',
         '/home/stack/install-undercloud.log',
-        '/home/stack/undercloud-install-*.tar.bzip2'
+        '/home/stack/undercloud-install-*.tar.bzip2',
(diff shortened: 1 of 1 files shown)


EmilienM:
Pull-request updated, HEAD is now 4d3556cc6b33376729fb740a55c9e403ee087c89

EmilienM:
Pull-request updated, HEAD is now a6ae6b162e5d8e7d5a15ceb9d3f9c66744e79196

pmoravec:
+1.

Theoretically, and independently on the requested change, the constants `NON_CONTAINERIZED_DEPLOY` and `CONTAINERIZED_DEPLOY` could be merged into one as they are referred together on one place only. Until there are some plans to collect those data separately in future, of course.

But as I wrote, that is an independent issue / code improvement/simplification.

PR #1664, opened 2019-04-29
[gluster] Add volume, state information and dump glusterd processes

TurboTurtle:
Adds collection of 'gluster volume get $vol all' output for all volume
names as well as the generate dumpfile from 'gluster get-state' when the
'dump' option is used. Note that this file will be under
/var/run/gluster in the archive.
   
Adds 'glusterd' to the list of processes sent signal USR1 when the
'dump' option is used. Removed the prep work previously done to redirect
where the statedump files are written to, as it has been reported that
those attempts do not function as intended. Statedumps are now collected
from /var/run/gluster.
    
Finally, some re-organization to ease future maintenance and updating
the size-limiting logic to use the implicit limiting in add_copy_spec().
    
---

### DIFF
--- a/sos/plugins/gluster.py
+++ b/sos/plugins/gluster.py
@@ -20,7 +20,7 @@ class Gluster(Plugin, RedHatPlugin):
     plugin_name = 'gluster'
     profiles = ('storage', 'virt')
 
-    statedump_dir = '/tmp/glusterfs-statedumps'
+    statedump_dir = '/var/run/gluster'
     packages = ["glusterfs", "glusterfs-core"]
     files = ["/etc/glusterd", "/var/lib/glusterd"]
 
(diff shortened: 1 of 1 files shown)


sos/plugins/gluster.py
```diff
@@ -91,38 +84,40 @@ def setup(self):
...
-                self.add_copy_spec('/tmp/glusterdump.options')
-                self.add_copy_spec(self.statedump_dir)
+            if os.path.exists(self.statedump_dir):
+                if self.check_ext_prog(
+                        "killall -USR1 glusterfs glusterfsd glusterd"):
+                    # let all the processes catch the signal and create
+                    # statedump file entries.
+                    time.sleep(1)
+                    self.wait_for_statedump(self.statedump_dir)
+                    self.add_copy_spec(self.statedump_dir)
```
pmoravec: <gluster_noob_question>
How big the statedumps could be? Shall not we concern about sizelimit?
</gluster_noob_question>

TurboTurtle: On a test environment here that supports a RHHI-V (RHV + gluster) deployment, the statedumps are all pretty small - under 100K. I don't think we need to worry about limiting here, at least not yet anyways.

TurboTurtle:
Updated the PR to include one extra command - `gluster volume status <volume> clients` for every volume

PR #1666, opened 2019-05-02
[frr] FRR plugin

ins-pirat-ion:
Fedora is going to ship FRR alternative to Quagga. Path to conf
is /etc/frr/.

Signed-off-by: Filip Krska <fkrska@redhat.com>

---

### DIFF
--- a/sos/plugins/frr.py
+++ b/sos/plugins/frr.py
@@ -1,0 +1,25 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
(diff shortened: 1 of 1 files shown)


ins-pirat-ion:
Can be tested with

https://copr.fedorainfracloud.org/coprs/mruprich/frr/

pmoravec:
+1

sos/plugins/frr.py
```diff
@@ -0,0 +1,27 @@
+# Copyright (C) 2007 Ranjith Rajaram <rrajaram@redhat.com>
```
TurboTurtle: Minor nit, copyright line needs to be updated/corrected.

I guess this is a c'n'p error?

ins-pirat-ion: @TurboTurtle thanks for heads-up. I've skipped the personal (c) lines now (e.g. like it is in acpid.py). I'd rather not claim copy monopoly on s/quagga/frr/g job ;). And hopefully it's better for eventual future c'n'p jobs ;).

PR #1667, opened 2019-05-02
[Plugin] Add method to collect service status based on InitSystem

TurboTurtle:
Adds a method to Plugin() to allow plugins to capture service status
information based on a policy-defined InitSystem() to use. This will
allow plugins to not need to account for host installation when trying
to collect service information, such as from 'systemctl status'
commands.

---


bmr-cymru:
Looks good: I've added patches to convert existing plugins that were using either `systemctl status $service` or `service $service status` to use the new API.

PR #1668, opened 2019-05-02
[openstack_containerlog] Fetch all container logs

cjeanner:
Until now, we didn't have a central plugin for collecting openstack container
logs. Each service was managing its own things, and this lead to missing
log files in the past.

Ensuring we fetch the content of /var/log/containers on its own ensures
we won't forget to add possible new directories, like the "stdouts" we
recently added in order to get the container STDOUT written on the disk

Therefore, this commit does 2 main things:
- cleanup all the /var/logs/containers custom copy
- create a central plugin that takes care of this location on its own

---

### DIFF
--- a/sos/plugins/container_log.py
+++ b/sos/plugins/container_log.py
@@ -1,0 +1,33 @@
+# Copyright (C) 2019 Red Hat, Inc., Cedric Jeanneret <cjeanner@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+import os
+
+
+class ContainerLog(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
+    '''All container logs
+    '''
+    plugin_name = 'container_log'
+    files = ('/var/log/containers',)
+
+    def setup(self):
+        if self.get_option('all_logs'):
+            self.add_copy_spec([
+                '/var/log/containers/',
+            ])
+        else:
+            self._subdirs('/var/log/containers')
+
+    def _subdirs(self, root='/var/log/containers'):
+        '''Get subdirs of passed root path
+        '''
+        for dirName, _, _ in os.walk(root):
+            self.add_copy_spec(os.path.join(dirName, '*.log'))
--- a/sos/plugins/gnocchi.py
+++ b/sos/plugins/gnocchi.py
@@ -43,15 +43,11 @@ class GnocchiPlugin(Plugin, RedHatPlugin):
             self.add_copy_spec([
                 "/var/log/gnocchi/*",
                 "/var/log/httpd/gnocchi*",
-                "/var/log/containers/gnocchi/*",
-                "/var/log/containers/httpd/gnocchi-api/*"
             ])
         else:
             self.add_copy_spec([
                 "/var/log/gnocchi/*.log",
                 "/var/log/httpd/gnocchi*.log",
-                "/var/log/containers/gnocchi/*.log",
-                "/var/log/containers/httpd/gnocchi-api/*log"
             ])
 
         vars_all = [p in os.environ for p in [
--- a/sos/plugins/mongodb.py
+++ b/sos/plugins/mongodb.py
@@ -32,7 +32,6 @@ class MongoDb(Plugin, DebianPlugin, UbuntuPlugin):
             self.var_puppet_gen + "/etc/",
             self.var_puppet_gen + "/etc/systemd/system/mongod.service.d/",
             "/var/log/mongodb/mongodb.log",
-            "/var/log/containers/mongodb/mongodb.log",
             "/var/lib/mongodb/mongodb.log*"
         ])
         self.add_cmd_output("du -sh /var/lib/mongodb/")
--- a/sos/plugins/mysql.py
+++ b/sos/plugins/mysql.py
@@ -34,7 +34,6 @@ class Mysql(Plugin):
             # Required for MariaDB under pacemaker (MariaDB-Galera)
             "/var/log/mysqld.log",
             "/var/log/mysql/mysqld.log",
-            "/var/log/containers/mysql/mysqld.log",
             "/var/log/mariadb/mariadb.log",
             "/var/lib/mysql/grastate.dat"
         ])
@@ -42,7 +41,6 @@ class Mysql(Plugin):
         if self.get_option("all_logs"):
             self.add_copy_spec([
                 "/var/log/mysql*",
-                "/var/log/containers/mysql*",
                 "/var/log/mariadb*"
             ])
 
--- a/sos/plugins/opendaylight.py
+++ b/sos/plugins/opendaylight.py
@@ -28,26 +28,20 @@ class OpenDaylight(Plugin, RedHatPlugin):
 
         if self.get_option("all_logs"):
 
-            # /var/log/containers/opendaylight/ path is specific to ODL
             # Oxygen-SR3 and earlier versions, and may be removed in a future
             # update.
 
             self.add_copy_spec([
                 "/opt/opendaylight/data/log/",
-                "/var/log/containers/opendaylight/",
-                "/var/log/containers/opendaylight/karaf/logs/",
             ])
 
         else:
 
-            # /var/log/containers/opendaylight/ path is specific to ODL
             # Oxygen-SR3 and earlier versions, and may be removed in a future
             # update.
 
             self.add_copy_spec([
                 "/opt/opendaylight/data/log/*.log*",
-                "/var/log/containers/opendaylight/*.log*",
-                "/var/log/containers/opendaylight/karaf/logs/*.log*",
             ])
 
         self.add_cmd_output("docker logs opendaylight_api")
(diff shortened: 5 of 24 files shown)


sos/plugins/openstack_containerlogs.py
```diff
@@ -0,0 +1,37 @@
...
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+import os
+
+
+class OpenStackCinder(Plugin):
```
TurboTurtle: I'm assuming `OpenStackCinder` is a copy-paste error and should be named something else.

But also, you need to subclass the distro plugin classes as well, E.G.:

```python
class OpenStackContainers(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
```

sos/plugins/openstack_containerlogs.py
```diff
@@ -0,0 +1,37 @@
...
+        else:
+            self.add_copy_spec(self._subdirs('/var/log/containers'))
+
+    def _subdirs(self, root='/var/log/containers'):
+        '''Get subdirs of passed root path
+        '''
+        dir_list = []
+        for dirName, subdirs, fname in os.walk(root):
+            if len(fname) != 0:
+                dir_list.append(os.path.join(dirName, '*.log'))
```
TurboTurtle: Is there really nothing else there of worth that doesn't end in `.log`?

What else is being captured by the `--all-logs` option?

bmr-cymru:
This pull request **introduces 1 alert** when merging 351c721cea5278716918ff0d03cdc5648c3ed0d9 into 1b9c2032149488a2372d188a8ed3251e364f41cf - [view on LGTM.com](https://lgtm.com/projects/g/sosreport/sos/rev/pr-6a4c47e3f49414b53f7b665589fd3061679100b6)

**new alerts:**

* 1 for Unused import

---

sos/plugins/openstack_containerlogs.py
```diff
@@ -0,0 +1,37 @@
...
+
+from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+import os
+
+
+class OpenstackContainerLog(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
+    """OpenStack container logs
+    """
+    plugin_name = "openstack_containerlog"
+    profiles = ('openstack', 'openstack_controller', 'openstack_compute')
```
TurboTurtle: Looks like my previous review got cut off somehow. Anyways...

See my notes about other components using `/var/log/containers` - we don't necessarily want to limit this plugin's scope to OpenStack.

At the same time, this plugin _should_ have some form of enablement gating like packages or services.

cjeanner: Hello! yeah, sorry for the "hard cut", I saw the notification from the bot and pushed an update right after that.

regarding the profiles: I wasn't aware that location was also used by other products - so of course we will need to update that listing. I can already check if, among the editions I've done in this commit, some plugins weren't related to openstack. To be honest I didn't think about that :/.
That would be a good start.

sos/plugins/openstack_containerlogs.py
cjeanner: hmm good question. There might be some "useless" files in some directories. In fact I took that "all_log" thingy because it was present and used in some of the openstack related plugins.

sos/plugins/openstack_containerlogs.py
cjeanner: hmm. apparently, there's also the "services" group for MongoDB (as a first quick pass).

We might actually "just" check if the /var/log/containers directory exists and, if so, enable it... Would that kind of solution be OK? Is there already a plugin using this capability?

TurboTurtle: You can use the `files` class member for `Plugin`, yeah. It ends up just doing an `os.path.exists()` on anything specified.

cjeanner: Perfect, applied in last version with your comment :). Lemme know if I missed something, but.. Should be fine. Hopefully.

Thank you for your time!

cjeanner:
@TurboTurtle hello! care to have a new look with the latest version? Your last change request should have been addressed.
Thank you!

cjeanner:
Fine for me! as long as it doesn't fall in some black corner.. We'd really need to be able to fetch that content for some customer cases, and without having to add new directories in that location :)

cjeanner:
Hello there - can we revive this PR? Would be really helpful when we have some support case :).

pmoravec:
+1 to Jake's comment about simplified maintainability but breaking the divide-and-collect approach "plugin for X collects all data for X (and nobody else does)". No strong preferences here..

Also not sure if `os.walk` can't be time consuming - can't `/var/log/containers` have rich (sub)directory structure (I dont think so but better ask and be assured than dealing with issues later on)?

cjeanner:
@pmoravec hm, right for the "divide and collect". Unless we consider "containers are one service" :).
At least, that would make things really, really easier, especially since, apparently, openshift does use the same directory - one change, many pleased people (hopefully ;)).

Regarding ```os.walk``` and time consumption, I think it's OK, especially for OpenStack since we have rules keeping 14 days of logs, not more. There aren't that many recursions as well (max 2 levels), meaning os.walk should be pretty fast.
For OpenShift, if it does use that same location, we should be in the same kind of use anyway.

The /var/log/containers structure isn't the same on all nodes since it depends on the services being actually deployed - this is the main motivation of the present change: avoid having to re-re-re-re-edit sosreport plugins when we either add or remove a service.

Does it help sorting things out? Feel free to ask for more details if needed.

cjeanner:
Hello there,

Care to merge or ask for changes for this PR? Would really, really love being able to get the whole openstack containers logs in one shot, even for new services...

Thank you!

cjeanner:
hello @TurboTurtle @pmoravec - any chance to get that PR merged? Would be really, really helpful in order to avoid future PR affecting the /var/logs/containers path..

Thank you!

pmoravec:
I have mixed feelings like described before. But I agree that containers *is* also a separate technology used by layered products, where simple plugin approach "divide and conquer" will be problematic. And in fact we already have similar approach implemented: e.g. `foreman` uses Apache `httpd` but `foreman` plugin relies on `apache` plugin to collect `httpd` logs and configs.

If there shall be container-related plugin, doesn't it make sense to also collect some configs or commands outputs related to *generic* containers (not docker-specific or so)? Is there something like that?

cjeanner:
hello @pmoravec - sorry for "late" answer. Weekend, blah.
So, it's not really possible to get "generic" container things:
- docker is a daemon with specific config location and CLI
- podman has no daemon, has other config files (not even really used in fact), and another CLI
- if we consider the LXC techno, it's another thing, different config, CLI and so on
- systemd containers? yet another thing :/

I tend to think "container config/specs" should probably be in dedicated modules since they are really different, and would rely on multiple tests in order to ensure this or that tech is used.

WDYT?

I'm checking in order to sort the conflict out, have to talk with Pavel about his change in the openstack_nova plugin....

sos/plugins/container_log.py
```diff
@@ -0,0 +1,36 @@
...
+    plugin_name = 'container_log'
+    files = ('/var/log/containers',)
+
+    def setup(self):
+        if self.get_option('all_logs'):
+            self.add_copy_spec([
+                '/var/log/containers/',
+            ])
+        else:
+            self.add_copy_spec(self._subdirs('/var/log/containers'))
```
pmoravec: Checking the merge conflict in `nova`:

1) it will be OK to replace `for novadir in novadirs:` loop by simple replacing `novadir` by `/var/log/nova/` in the loop

2) Calling single `add_copy_spec` means sizelimit is applied to *all* arguments altogether. Could you please replace it by something like:

    for _dir in self._subdirs('/var/log/containers'):
        self.add_copy_spec(_dir)

(or maybe have the `add_copy_spec` directly in `_subdirs` method? Or inline the method directly to `setup`? whatever shall be ok..)

cjeanner: Didn't realize that was you broking my PR ;).

Lemme know if the update suits your needs. We're on the same TZ, we can sort this quick hopefully.

bmr-cymru:
This pull request **introduces 1 alert** when merging 4014ca1e9688d0c94830e7102dae30c1386e1721 into aca6bd4f93b4ae5bbb1fe90dffa55297fd5fc36f - [view on LGTM.com](https://lgtm.com/projects/g/sosreport/sos/rev/pr-6efdbe4f4dbdcc802738d0f4f6fd9d9e089aff1f)

**new alerts:**

* 1 for Unused local variable

---

---

pmoravec:
@TurboTurtle or @bmr-cymru , as this change affects multiple plugins and their competencies, please provide ACK to have some consensus. I am OK with the current approach.

`dir_list = []` is obsolete, I would rename the `_subdirs` method to e.g. `collect_subdirs` and make `/var/log/containers` a variable to refer to. I can update all of that during merging.

sos/plugins/container_log.py
```diff
@@ -0,0 +1,34 @@
...
+                '/var/log/containers/',
+            ])
+        else:
+            self._subdirs('/var/log/containers')
+
+    def _subdirs(self, root='/var/log/containers'):
+        '''Get subdirs of passed root path
+        '''
+        for dirName, _, _ in os.walk(root):
+            if len(fname) != 0:
```
pmoravec: unknown variable `fname`, now :)

pmoravec: Anyway, this test can be skipped, `add_copy_spec` will ignore empty dirs. The test can be seen as some sort of optimisation ("stop treating empty dir as soon as possible"), but I expect most of dirs will have some content so the test would help a little only.

cjeanner: pfff... doing 10 things in //... ;).
Indeed, most of the directories will get some content at some point :).

PR #1670, opened 2019-05-06
[juju] Add support for Juju 2.x / refactoring work

### LINKED ISSUE #1653: [juju] juju plugins stop/start juju-db and mongod

dnegreira:
Adding support for Juju 2.x. This plugin needed some cleanup.
There are 3 use cases for the juju plugin: the juju client,
a juju deployed machine, and a juju controller.
We have broken the plugin into 3 parts to reflect this.

The juju plugin is now specifically for the juju client use case.
Juju is typically installed as a snap or a package, therefore, we added
/snap/bin to the path for Debian to handle this until a better method
for handling snaps is figured out.

Juju is configured as a user to communicate with the controller it is
therefore necessary to run the juju commands with sudo for them to
execute properly.

This plugin now collects as much information about the juju environment
as possible by using the juju command, this information includes the
configured clouds, models, spaces, model configs, status of each model
and configuration of each app in each model.

Closes #1653

Co-authored-by: David Negreira <david.negreira@canonical.com>
Co-authored-by: Nick Niehoff <nick.niehoff@canonical.com>
Signed-off-by: David Negreira <david.negreira@canonical.com>

---

### DIFF
--- a/sos/plugins/juju.py
+++ b/sos/plugins/juju.py
@@ -29,9 +29,8 @@ class Juju(Plugin, UbuntuPlugin):
     def collect_app_config(self, username, modelname, statusfilespath):
         statusinfo = {}
         try:
-            fp = open(statusfilespath, "r")
-            statusinfo = json_loads(fp.read())
-            fp.close()
+            with open(statusfilespath, 'r') as fd:
+                statusinfo = json_loads(fd.read())
         except Exception:
             return
         if "applications" in statusinfo:
@@ -48,9 +47,8 @@ class Juju(Plugin, UbuntuPlugin):
     def collect_model_output(self, username, modelsfilepath):
         modelsinfo = {}
         try:
-            fp = open(modelsfilepath, "r")
-            modelsinfo = json_loads(fp.read())
-            fp.close()
+            with open(modelsfilepath, 'r') as fd:
+                modelsinfo = json_loads(fd.read())
         except Exception:
             return
         if "models" in modelsinfo:
--- a/sos/plugins/juju.py
+++ b/sos/plugins/juju.py
@@ -9,26 +9,8 @@
 # See the LICENSE file in the source distribution for further information.
 
 import os
-from sos.plugins import Plugin, UbuntuPlugin
 from json import loads as json_loads
-
-
-def ensure_service_is_running(service):
-    def wrapper(callback):
-        def wrapped_f(self, *args, **kwargs):
-            try:
-                result = self.call_ext_prog("service {0} stop".format(service))
-                if result["status"] != 0:
-                    raise Exception("Cannot stop {0} service".format(service))
-                callback(self, *args, **kwargs)
-            except Exception as ex:
-                self._log_error("Cannot stop {0}, exception: {1}".format(
-                    service,
-                    ex.message))
-            finally:
-                self.call_ext_prog("service {0} start".format(service))
-        return wrapped_f
-    return wrapper
+from sos.plugins import Plugin, UbuntuPlugin
 
 
 class Juju(Plugin, UbuntuPlugin):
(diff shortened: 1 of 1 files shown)


bmr-cymru:
This pull request **introduces 1 alert** when merging aa2f45a1b2d4ad40d4e21bd1d8c963205de7bce6 into 1b9c2032149488a2372d188a8ed3251e364f41cf - [view on LGTM.com](https://lgtm.com/projects/g/sosreport/sos/rev/pr-909336f2c26a8dee593d0fcfe2f311efeeca96d4)

**new alerts:**

* 1 for Implicit string concatenation in a list

---

BryanQuigley:
Generally we try to squash the commits (so there isn't a separate commit for Fix missing comma) - see http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

What I usually do to make a change to an in-progress PR is make the change and then do a git commit -s --amend, then git push --force it up.

One other note: the plugin options removed are not considered useful in the new version of Juju and they were never heavily relied on.

Thanks for your work on this!

dnegreira:
Hi @BryanQuigley,

Squashed and committed with a force push, thanks for the comment on the plugin options as I forgot to mention it in the original comment.

sos/plugins/juju.py
```diff
@@ -40,82 +22,142 @@ class Juju(Plugin, UbuntuPlugin):
...
+            fp.close()
+        except Exception:
+            return
+        if "applications" in statusinfo:
+            for appname, appinfo in statusinfo["applications"].items():
+                # Get the application config for every application in every
+                # model
+                self.add_cmd_output(
+                    "sudo -i -u {} juju config --format yaml -m {} {}".format(
+                        username, modelname, appname
```
niedbalski: is there a reason for running this command under sudo?

sos/plugins/juju.py
```diff
@@ -40,82 +22,142 @@ class Juju(Plugin, UbuntuPlugin):
...
-    @ensure_service_is_running("juju-db")
-    def export_mongodb(self):
-        collections = (
-            "relations",
-            "environments",
-            "linkednetworks",
-            "system",
-            "settings",
+            fp = open(modelsfilepath, "r")
+            modelsinfo = json_loads(fp.read())
```
niedbalski: switch it to with 

open(modelsfilepath, 'r') as fd: 
  modelsinfo = jsn_loads(fd.read())

no close is required

sos/plugins/juju.py
```diff
@@ -40,82 +22,142 @@ class Juju(Plugin, UbuntuPlugin):
...
+                             "-n 500 --no-tail -m {}"
+                             .format(username, modelname)),
+                            ("sudo -i -u {} juju model-config "
+                             "--format yaml -m {}"
+                             .format(username, modelname)),
+                        ]
+                    )
+                    # Get the model status as json so we can dig deeper
+                    status = self.get_cmd_output_now(
+                        "sudo -i -u {} juju status -m {} --format json".format(
```
niedbalski: do we need sudo here?

slashdd: I would suggest to use 'su' instead. sosreport is already run as root user, 'su' should do the trick fine if you absolutely need to perform action from a specific user.

slashdd: vdsm plugin uses something similar already
https://github.com/sosreport/sos/blob/master/sos/plugins/vdsm.py#L95

slashdd: Also sudo seems to do nasty things when invoking juju command, it chown configuration files.
Launchpad bug: https://launchpad.net/bugs/1747357

dnegreira: @niedbalski sudo is to run the juju command as the specified user that has juju client configured and I understand that this might not be widely available.
I can swap the command from `sudo` to `su`, as it is available more widely and works well with this use case.
The other option that I would prefer not to do is to set the environment variable `JUJU_HOME` when running the juju command as root, this might chown some files inside the `JUJU_HOME`.

Let me know your thoughts.

BryanQuigley: Actually I prefer we fix this in Juju, but heh.. (we also can't magically backport that fix everywhere anyway)

I looked through the code and it seems RH specific plugins (katello.py, saphana.py, sapnw.py, vdsm.py are RH only, and foreman.py supports both) use su -c'.   I would expect more Ubuntu admins to be more comfortable seeing sudo in their logs..

dnegreira: @BryanQuigley @niedbalski want to come to an agreement here?
I have a slight preference of `su` over `sudo` as it is more available on distros than `sudo` by default.

Would like to move this forward either way.

battlemidget: Let's be consistent, precedent stands to use `su -` so let's stick to that.

slashdd:
Just to avoid more confusion, here's a potential 3rd options. After a discussion with the ubuntu security team on that subject to have their opinion, they were noy super keen of both 'sudo' and 'su' and suggested to drop privileges if possible instead.

sos/plugins/juju.py
slashdd: Just to add more confusion, here's a potential 3rd options. After a discussion with the ubuntu security team on that subject to have their opinion, they were not super keen of both 'sudo' and 'su' and suggested to look to possibly drop privileges if possible instead.

battlemidget: Looks like we could potentially use setpriv and is widely supported

slashdd: The outcome of the discussion with the Ubuntu security is to use `runuser -P`
https://github.com/karelzak/util-linux/blob/master/login-utils/runuser.1#L104-L115

pmoravec:
Approved until the `sudo` shall be replaced by `runuser -P` (per https://github.com/sosreport/sos/pull/1670/files#r283109272).

dnegreira:
Just an update that this is not dead and I am working on it, still trimming some issues around runuser.

BryanQuigley:
I'm thinking it might make sense to put a really simple juju plugin for the short term (that doesn't have any runuser/sudo) and focus on machine/controller.  Would that simple one be able to collect anything of note?

It might be possible to revisit this when we can run sosreport as a non-root user.  (A quick hack seems we might be closer to this then I would think - just set "requires_root = False".  and edit sos/utilities.py def _child_prep_fn(): to just immediately return.  If that works, then we would just fail to run this if you run it as root...)

pmoravec:
> It might be possible to revisit this when we can run sosreport as a non-root user. (A quick hack seems we might be closer to this then I would think - just set "requires_root = False". and edit sos/utilities.py def _child_prep_fn(): to just immediately return. If that works, then we would just fail to run this if you run it as root...)

AFAIK that hack isn't sufficient and some bigger or trickier code fix prevents that (there should be an issue elaborating on this more).

FYI we (in Red Hat) have been discussing "non-root execution of sosreport" feature a bit recently, and we identified no interest in the feature. Since then, this is the second maybe-possible use case of the feature.

pmoravec:
My understanding from https://github.com/sosreport/sos/pull/1670/files#r283109272 is, that `sudo` shall be replaced by `runuser -P`. Otherwise, just from code review perspective, ACK to the PR.

dnegreira:
We decided to simplify the plugin refactoring and have opened a new PR: https://github.com/sosreport/sos/pull/1935

PR #1671, opened 2019-05-06
[juju_controller] New plugin for juju controller

dnegreira:
The juju_controller plugin is for the juju controller use case.

A juju controller is a juju machine so the machine plugin should handle
the collection of logs from that aspect.

The only identifying piece of information that this is a juju
controller is the /var/lib/juju/db directory and in reality juju is
just a mongodb server.

This plugin instruments some mongo queries to get the status
and data of the juju database.

Co-authored-by: David Negreira <david.negreira@canonical.com>
Co-authored-by: Nick Niehoff <nick.niehoff@canonical.com>
Signed-off-by: David Negreira <david.negreira@canonical.com>

---

### DIFF
--- a/sos/plugins/juju_controller.py
+++ b/sos/plugins/juju_controller.py
@@ -1,0 +1,162 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+# Copyright (C) 2019 Nick Niehoff <nick.niehoff@canonical.com>
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+import re
+import os
+from sos.plugins import Plugin, UbuntuPlugin
+
+
+# A Juju Controller is a juju machine so this plugin assumes the juju_machine
+# plugin will be executed to collect that information
+class JujuController(Plugin, UbuntuPlugin):
+    """ Juju orchestration tool - Controller
+    """
+
+    plugin_name = 'juju_controller'
+    profiles = ('virt', 'sysmgmt')
+
+    # Use files instead of packages here because there is no package installed
+    #  on a juju controller that uniquely identifies it:
+    files = '/var/lib/juju/db'
+
+    option_list = [
+        ('export-mongodb',
+         'Export mongodb collections as json files', '', False,)
+    ]
+
+    def export_mongodb(self, juju_machine_id, dbpass):
+        # We are gonna grab the output of the mongo command
+        # and do some regexp and replace()/split() it to a list.
+        # Usually when connecting with ssl on localhost, the cli send some
+        # output which unfortunately ends up in stdin instead of stderr
+        # and we need to do this extra work to parse it.
+        collections_query = ("mongo 127.0.0.1:37017/juju "
+                             + "--authenticationDatabase admin "
+                             + "--ssl --sslAllowInvalidCertificates --quiet "
+                             + "--username {} --password {} ".format(
+                               juju_machine_id, dbpass)
+                             + "--eval 'db.getCollectionNames();'")
+
+        _collections = self.get_command_output(collections_query)
+
+        if _collections['status'] == 0:
+            collections_output = _collections['output']
+        else:
+            collections_output = ''
+
+        collections_regexp = re.compile(r'(?ms)^\[(.*)\]')
+        collections = collections_regexp.findall(collections_output)[0]
+        subs = [
+            ('\"', ''),
+            (',', ''),
+        ]
+        # In order to be able to properly query the mongodb collection
+        # tables we need to cleanup the output from the collections_query
+        # as those return the " and , symbols, which we cannot cleanly use
+        # those to cleanly query after
+        for search, replace in subs:
+            collections = collections.replace(search, replace)
+        collections = collections.split()
+
+        # We don't collect txns.log, statuseshistory, metrics, txns due to size
+        collections_ignore = ['txns.log', 'statuseshistory', 'metrics', 'txns']
+        for item in collections_ignore:
+            if item in collections:
+                collections.remove(item)
+
+        for collection in collections:
+            filename = "mongoexport_collection_{}.json".format(collection)
+            # This will execute a mongoexport command, note it will add the
+            # password to the sos log, however we remove it from the filename
+            mongoexport_cmd = (
+                "mongoexport --host 127.0.0.1 --port 37017 --db juju "
+                + "--authenticationDatabase admin --ssl "
+                + "--sslAllowInvalidCertificates --username '{}' --password "
+                + "'{}' --collection {} --jsonArray"
+            )
+            self.add_cmd_output(
+                mongoexport_cmd.format(juju_machine_id, dbpass, collection),
+                suggest_filename=filename,
+            )
+
+    def get_juju_machine_id(self):
+        # get the juju machine name machine-X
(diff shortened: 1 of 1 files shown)


sos/plugins/juju_controller.py
```diff
@@ -0,0 +1,157 @@
...
+        # Usually when connecting with ssl on localhost, the cli send some
+        # output which unfortunately ends up in stdin instead of stderr
+        # and we need to do this extra work to parse it.
+        collections_query = ("mongo 127.0.0.1:37017/juju "
+                             + "--authenticationDatabase admin "
+                             + "--ssl --sslAllowInvalidCertificates --quiet "
+                             + "--username %s --password %s "
+                             % (juju_machine_id, dbpass)
+                             + "--eval 'db.getCollectionNames();'")
+        collections_output = os.popen(collections_query).read()
```
slashdd: I would rather go with subprocess.popen()
- AFAIK os.popen() is considered deprecated.
- os.popen() is starting a process with a shell, possible injection detected, security issue.

Reference:
https://docs.openstack.org/bandit/latest/plugins/b605_start_process_with_a_shell.html
https://docs.python.org/3/library/subprocess.html

TurboTurtle: Nack to `popen()` calls directly in the plugin. Use the existing `get_command_output()` in the `Plugin` class. E.G.

```python
_collections = self.get_command_output(collections_query)
if _collections['status'] == 0:
    collections_output = _collections['output']
else:
    collections_output = ''
```

sos/plugins/juju_controller.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def setup(self):
+
+        self.add_copy_spec(
+            [
+                "/etc/default/mongodb",
+                "/var/lib/juju/bootstrap-params",
+                "/lib/systemd/system/juju-db/juju-db.service",
+            ]
+        )
```
TurboTurtle: Stylistic nitpick: Keep the brackets on the same line.

```python
self.add_copy_spec([
    'foo',
    'bar'
])
```

sos/plugins/juju_controller.py
```diff
@@ -0,0 +1,157 @@
...
+    files = "/var/lib/juju/db"
+
+    option_list = [
+        (
+            "export-mongodb",
+            "Export mongodb collections as json files",
+            "",
+            False,
+        )
+    ]
```
TurboTurtle: Stylistic nitpick: no need to space this out so much, it'd be easier to read at a glance if it were all together.

```python
option_list = [
    ('export-mongodb', 'Export mongodb collections as json', 'fast', False)
]
```

sos/plugins/juju_controller.py
```diff
@@ -0,0 +1,157 @@
...
+    def get_juju_machine_id(self):
+        # get the juju machine name machine-X
+        for dirpath, dirnames, filenames in os.walk("/var/lib/juju/agents"):
+            for dirname in dirnames:
+                match = re.match(r"(machine-\d+)", dirname)
+                if match:
+                    return match.group(1)
+        return None
+
+    def get_jujudb_pass(self, juju_machine_id):
```
TurboTurtle: We should have the user provide a password to us, rather than trying to dig it up ourselves. Like the `postgresql` plugin, in order for us to get a db dump the user *must* supply the password via `-k postgresql.password=foobar`

sos/plugins/juju_controller.py
pmoravec: As a practical example to copy&paste&modify:

https://github.com/sosreport/sos/blob/master/sos/plugins/pulp.py#L72-L75
https://github.com/sosreport/sos/blob/master/sos/plugins/pulp.py#L111
https://github.com/sosreport/sos/blob/master/sos/plugins/pulp.py#L116-L121

sos/plugins/juju_controller.py
dnegreira: The reason we fetch the password is due to the fact that the password is always automatically generated on installation and the user never sets anything here and doesn't need to know it as a general rule.
We would prefer to have the password being fetched automatically due to the fact above.
Let me know if this is enough reasoning to leave it automated, if not we can always create the parameter and instruct the users how to fetch/set the password.

pmoravec: IMHO this is still safe (up to two concerns). The password is used just internally and not leaked anywhere else than:

- in process listing during the `mongo` command execution; so any unprivileged user can get the password (is there a way like env.variable to hide it? I am afraid not..)
- in output file generated by sosreport, the password will be stored and provided to any support representative - until `suggest_filename` is used (`get_option` has it, `export_mongodb` will be updated and then it should use `suggest_filename` as well)

sos/plugins/juju_controller.py
```diff
@@ -0,0 +1,157 @@
...
+                             + "--eval 'db.getCollectionNames();'")
+        collections_output = os.popen(collections_query).read()
+        collections_regexp = re.compile(r"(?ms)^\[(.*)\]")
+        collections = collections_regexp.findall(collections_output)[0]
+        subs = [
+            ('\"', ''),
+            (',', ''),
+        ]
+        for search, replace in subs:
+            collections = collections.replace(search, replace)
```
pmoravec: it would be nice to explain in a comment why the `\"` and `,` are being removed (to get a pure list of collection names, right? I recall some mongo output change between some versions..)

sos/plugins/juju_controller.py
dnegreira: Fixed on 9c909bdb8632d8bc15671235eb125d4268e0218b - Had to put on two lines or it would exceed 79 chars

sos/plugins/juju_controller.py
dnegreira: Fixed on 9c909bdb8632d8bc15671235eb125d4268e0218b

sos/plugins/juju_controller.py
dnegreira: Hi,

For your 1st concern, there is a way to pass the password parameter empty and wait for the mongodb prompt to ask for it, I can see if there is a way to improve that part but it will be a little bit more complex.

For your second concern, I'm sorry but I'm not understanding what do you mean, we are using suggest_filename on both add_cmd_output.

sos/plugins/juju_controller.py
dnegreira: Fixed on 9c909bd

sos/plugins/juju_controller.py
```diff
@@ -0,0 +1,162 @@
...
+
+        for collection in collections:
+            filename = "mongoexport_collection_{}.json".format(collection)
+            # This will execute a mongoexport command, note it will add the
+            # password to the sos log, however we remove it from the filename
+            mongoexport_cmd = (
+                "mongoexport --host 127.0.0.1 --port 37017 --db juju "
+                + "--authenticationDatabase admin --ssl "
+                + "--sslAllowInvalidCertificates --username '{}' --password "
+                + "'{}' --collection {} --jsonArray"
```
bmr-cymru: We can't pass passwords like this - as well as appearing in the logs they will be visible to other users on the system in process listings.

For `mongoexport` it looks like the password can be [provided on stdin](https://stackoverflow.com/questions/32152666/automated-mongodump-how-to-hide-password-from-ps-output) - there isn't currently a way for plugins to pass input to external processes but if this is the only option for mongo then we may be able to make this available. An environment variable would be much better however if this is supported.

dnegreira:
We decided to simplify the plugin refactoring and have opened a new PR: https://github.com/sosreport/sos/pull/1935

PR #1551, opened 2019-01-30
WIP [openstack] Extract Placement plugin from Nova

lyarwood:
The OpenStack Placement service is being extracted from Nova [1] during
the Stein development cycle [2]. This change extracts the required
plugin logic from the original Nova plugin into a new Placement plugin
ahead of this extraction.

[1] https://docs.openstack.org/placement/latest/
[2] https://releases.openstack.org/stein/schedule.html

Signed-off-by: Lee Yarwood <lyarwood@redhat.com>

---
Please place an 'X' inside each '[]' to confirm you adhere to our [Contributor Guidelines](https://github.com/sosreport/sos/wiki/Contribution-Guidelines)

- [X] Is the commit message split over multiple lines and hard-wrapped at 72 characters?
- [X] Is the subject and message clear and concise?
- [X] Does the subject start with **[plugin_name]** if submitting a plugin patch or a **[section_name]** if part of the core sosreport code?
- [X] Does the commit contain a **Signed-off-by: First Lastname <email@example.com>**?

PR #1676, opened 2019-05-09
[openstack_placement] Extract Placement plugin from Nova

piotrkpc:
The Placement service is being extracted from Nova [1] during
the Stein development cycle [2]. This change extracts the required
plugin logic from the original Nova plugin into a new Placement plugin
ahead of this extraction.

[1] https://docs.openstack.org/placement/latest/
[2] https://releases.openstack.org/stein/schedule.html

Co-Authored-by: Piotr Kopec <pkopec@redhat.com>

Signed-off-by: Lee Yarwood <lyarwood@redhat.com>

---

### DIFF
--- a/sos/plugins/openstack_placement.py
+++ b/sos/plugins/openstack_placement.py
@@ -1,0 +1,112 @@
+# Copyright (C) 2019 Red Hat, Inc., Lee Yarwood <lyarwood@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+
+
+class OpenStackPlacement(Plugin):
+    """OpenStack Placement
+    """
+    plugin_name = "openstack_placement"
+    profiles = ('openstack', 'openstack_controller')
+
+    var_puppet_gen = "/var/lib/config-data/puppet-generated/placement"
+
+    def setup(self):
+
+        # collect commands output only if the openstack-placement-api service
+        # is running
+
+        in_container = self.running_in_container()
+
+        if self.service_is_running('openstack-placement-api') or in_container:
+            placement_config = ""
+            # if containerized we need to pass the config to the cont.
+            if in_container:
+                placement_config = "--config-dir " + self.var_puppet_gen + \
+                                "/etc/placement/"
+            self.add_cmd_output(
+                "placement-manage " + placement_config + " db version",
+                suggest_filename="placement-manage_db_version"
+            )
+
+        if self.get_option("all_logs"):
+            self.add_copy_spec([
+                "/var/log/placement/",
+                "/var/log/containers/placement/",
+                "/var/log/containers/httpd/placement-api/"
+            ])
+        else:
+            self.add_copy_spec([
+                "/var/log/placement/*.log",
+                "/var/log/containers/placement/*.log",
+                "/var/log/containers/httpd/placement-api/*log",
+            ])
+
+        self.add_copy_spec([
+            "/etc/placement/",
+            self.var_puppet_gen + "/etc/placement/",
+            self.var_puppet_gen + "/etc/my.cnf.d/tripleo.cnf",
+            self.var_puppet_gen + "/etc/httpd/conf/",
+            self.var_puppet_gen + "/etc/httpd/conf.d/",
+            self.var_puppet_gen + "/etc/httpd/conf.modules.d/*.conf",
+        ])
+
+    def running_in_container(self):
+        for runtime in ["docker", "podman"]:
+            container_status = self.get_command_output(runtime + " ps")
+            if container_status['status'] == 0:
+                for line in container_status['output'].splitlines():
+                    if line.endswith("placement_api"):
+                        return True
+        return False
+
+    def apply_regex_sub(self, regexp, subst):
+        self.do_path_regex_sub("/etc/placement/*", regexp, subst)
+        self.do_path_regex_sub(
+            self.var_puppet_gen + "/etc/placement/*",
+            regexp, subst
+        )
+
+    def postproc(self):
+        protect_keys = ["password"]
+        connection_keys = ["database_connection", "slave_connection"]
+
+        self.apply_regex_sub(
+            r"((?m)^\s*(%s)\s*=\s*)(.*)" % "|".join(protect_keys),
+            r"\1*********"
+        )
+        self.apply_regex_sub(
+            r"((?m)^\s*(%s)\s*=\s*(.*)://(\w*):)(.*)(@(.*))" %
+            "|".join(connection_keys),
+            r"\1*********\6"
+        )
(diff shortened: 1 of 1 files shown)


piotrkpc:
I've picked up the work started by @lyarwood in [PR#1551](https://github.com/sosreport/sos/pull/1551). The extraction is done so any feedback will be appreciated.

sos/plugins/openstack_placement.py
```diff
@@ -0,0 +1,124 @@
...
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+
+
+class OpenStackPlacement(Plugin):
+    """OpenStack Placement
+    """
+    plugin_name = "openstack_placement"
```
bmr-cymru: How is this going to work for older distributions that still ship the pre-extraction package layout? Remember that sos needs to support all possible shipping versions until the old ones are completely removed. For the time being, unless you are absolutely certain the old ones are gone from the field, there may be a need to support both.

sos/plugins/openstack_placement.py
```diff
@@ -0,0 +1,124 @@
...
+
+        # collect commands output only if the openstack-placement-api service
+        # is running
+        service_status = self.get_command_output(
+            "systemctl status openstack-placement-api.service"
+        )
+
+        in_container = self.running_in_container()
+
+        if (service_status['status'] == 0) or in_container:
```
TurboTurtle: Same as above. In fact you could remove `service_status` entirely here and replace it with something like

```python
if self.service_is_running('openstack-placement-api') or in_container:
```

sos/plugins/openstack_placement.py
```diff
@@ -0,0 +1,124 @@
...
+            self.add_cmd_output(
+                "placement-manage " + placement_config + " db version",
+                suggest_filename="placement-manage_db_version"
+            )
+
+        if self.get_option("all_logs"):
+            self.add_copy_spec([
+                "/var/log/placement/",
+                "/var/log/containers/placement/",
+                "/var/log/containers/httpd/placement-api/"
```
TurboTurtle: The capturing of `/var/log/containers/` in this plugin conflicts with the proposal in #1668, to move collection of that directory to a separate plugin since so many components use it.

Jury is still out on if we want to go that way, just making a note of it here.

sos/plugins/openstack_placement.py
```diff
@@ -0,0 +1,124 @@
...
+            "/etc/placement/",
+            self.var_puppet_gen + "/etc/placement/",
+            self.var_puppet_gen + "/etc/my.cnf.d/tripleo.cnf",
+            self.var_puppet_gen + "/etc/httpd/conf/",
+            self.var_puppet_gen + "/etc/httpd/conf.d/",
+            self.var_puppet_gen + "/etc/httpd/conf.modules.d/*.conf",
+        ])
+
+        if self.get_option("verify"):
+            self.add_cmd_output("rpm -V %s" % ' '.join(self.packages))
```
TurboTurtle: No need for this - sos already makes this call if `--verify` is used. Also, this would fail when `DebianPlacement` is loaded.

sos/plugins/openstack_placement.py
```diff
@@ -0,0 +1,124 @@
...
+            "|".join(connection_keys),
+            r"\1*********\6"
+        )
+
+
+class DebianPlacement(OpenStackPlacement, DebianPlugin, UbuntuPlugin):
+
+    packages = (
+        'placement'
+    )
```
TurboTurtle: No need to break this over multiple lines here.

sos/plugins/openstack_placement.py
```diff
@@ -0,0 +1,124 @@
...
+class RedHatPlacement(OpenStackPlacement, RedHatPlugin):
+
+    packages = ('openstack-selinux',)
+
+    def setup(self):
+        super(RedHatPlacement, self).setup()
+        if self.get_option("all_logs"):
+            self.add_copy_spec([
+                "/var/log/httpd/placement*",
+            ])
```
TurboTurtle: Since this is a single location, no need to make this a list.

sos/plugins/openstack_placement.py
```diff
@@ -0,0 +1,124 @@
...
+        else:
+            self.add_copy_spec([
+                "/var/log/httpd/placement*.log",
```
TurboTurtle: Same here, `add_copy_spec()` can take a string as well as a list.

sos/plugins/openstack_placement.py
piotrkpc: This still works with earlier releases because I've not removed any code from nova plugin. I've tested it against Queens (OSP13) and it seams working well, it collects all placement logs and info. If you can think of any test and release I should test against, please let me know.

sos/plugins/openstack_placement.py
piotrkpc: Done, thank you.

sos/plugins/openstack_placement.py
piotrkpc: Done, thank you.

sos/plugins/openstack_placement.py
piotrkpc: Done, thank you.

sos/plugins/openstack_placement.py
piotrkpc: Good tip, removed. However the same thing is all over other OpenStack plugins.

sos/plugins/openstack_placement.py
piotrkpc: Noted, I'll watch that PR and be happy to make necessary changes if it will land before this.

sos/plugins/openstack_placement.py
piotrkpc: as above, thanks.

sos/plugins/openstack_placement.py
```diff
@@ -0,0 +1,124 @@
...
+    profiles = ('openstack', 'openstack_controller',)
+
+    var_puppet_gen = "/var/lib/config-data/puppet-generated/placement"
+
+    def setup(self):
+
+        # collect commands output only if the openstack-placement-api service
+        # is running
+        service_status = self.get_command_output(
+            "systemctl status openstack-placement-api.service"
```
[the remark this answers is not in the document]
piotrkpc: Good tip. However the same thing is all over other OpenStack plugins.

sos/plugins/openstack_placement.py
```diff
@@ -0,0 +1,112 @@
...
+            self.var_puppet_gen + "/etc/httpd/conf.d/",
+            self.var_puppet_gen + "/etc/httpd/conf.modules.d/*.conf",
+        ])
+
+    def running_in_container(self):
+        for runtime in ["docker", "podman"]:
+            container_status = self.get_command_output(runtime + " ps")
+            if container_status['status'] == 0:
+                for line in container_status['output'].splitlines():
+                    if line.endswith("placement_api"):
```
pmoravec: If #1682 would be merged, this method can be simplified to something like:

    def running_in_container(self):
        for runtime in ["docker", "podman"]:
            co = {'cmd': runtime + " ps", 'output': 'placement_api'}
            if self.test_predicate(self, pred=SoSPredicate(self, cmd_output=co)):
                return True
        return False


Until then, no comment :).

Issue #1673, opened 2019-05-07
[sosreport] [rfe] grab logs since date


valleedelisle:
That would be great to have a `--since` switch when using --all-logs. When using `--all-logs`, we often see very large sosreport archive when we only need the recent events.

`--since` switch should only prevent archived (.gz) files from being added, otherwise we would be missing files that aren't updated often.

`--since` should also be passed to `journalctl`

TurboTurtle:
I like the idea, would require #1586 to get merged.

We'd need to talk about how we'd expect end users to use `--since` though. Do they supply a date? If so, what date format? Should we instead just specify a number of days back to collect from? E.G `--since 7`  (or maybe `--since-days 7` to make it a little more obvious?)

valleedelisle:
@TurboTurtle and I discussed this issue offline and he proposed that `--since` should also work without `--all-logs`. I think it's a very legit recommendation and will certainly implement/test this in a near future but I would like to see `--since` added as soon as possible. Some customers are making pressure for something like this.

valleedelisle:
And here's the merge request https://github.com/sosreport/sos/pull/1677/

bmr-cymru:
Let's keep all the discussion on this change in one place: #1677

PR #1677, opened 2019-05-14
[sosreport] [rfe] grab logs since date

valleedelisle:
---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -26,6 +26,7 @@ import errno
 # PYCOMPAT
 import six
 from six.moves import zip, filter
+from datetime import datetime
 
 # FileNotFoundError does not exist in 2.7, so map it to IOError
 try:
@@ -724,7 +725,8 @@ class Plugin(object):
         matches any of the option names is returned.
         """
 
-        global_options = ('verify', 'all_logs', 'log_size', 'plugin_timeout')
+        global_options = ('verify', 'all_logs', 'since' 'log_size',
+                          'plugin_timeout')
 
         if optionname in global_options:
             return getattr(self.commons['cmdlineopts'], optionname)
@@ -806,13 +808,27 @@ class Plugin(object):
                     self._log_debug("skipping forbidden path '%s'" % _file)
                     continue
                 try:
-                    current_size += os.stat(_file)[stat.ST_SIZE]
+                    filestat = os.stat(_file)
                 except OSError:
                     self._log_info("failed to stat '%s'" % _file)
+                    continue
+                current_size += filestat[stat.ST_SIZE]
+                filetime = datetime.fromtimestamp(filestat[stat.ST_MTIME])
                 if sizelimit and current_size > sizelimit:
                     limit_reached = True
                     break
-                self._add_copy_paths([_file])
+
+                if _file.endswith('.gz') or re.search(r'-\d{8}$', _file):
+                    fileisarchive = True
+                else:
+                    fileisarchive = False
+
+                if (self.get_option('since') is None or
+                    fileisarchive is False or
+                    (fileisarchive is True and
+                     self.get_option('since') is not None and
+                     self.get_option('since') <= filetime)):
+                    self._add_copy_paths([_file])
 
             if limit_reached and tailit and not _file_is_compressed(_file):
                 file_name = _file
@@ -1081,7 +1097,7 @@ class Plugin(object):
         journal_cmd = "journalctl --no-pager "
         unit_opt = " --unit %s"
         boot_opt = " --boot %s"
-        since_opt = " --since %s"
+        since_opt = " --since '%s'"
         until_opt = " --until %s"
         lines_opt = " --lines %s"
         output_opt = " --output %s"
--- a/sos/plugins/logs.py
+++ b/sos/plugins/logs.py
@@ -52,7 +52,8 @@ class Logs(Plugin):
                 self.add_copy_spec(i)
 
         if self.get_option('all_logs'):
-            self.add_journal(boot="this", allfields=True, output="verbose")
+            self.add_journal(boot="this", since=self.get_option("since"), 
+                             allfields=True, output="verbose")
             self.add_journal(boot="last", allfields=True, output="verbose")
 
     def postproc(self):
@@ -96,7 +97,7 @@ class RedHatLogs(Logs, RedHatPlugin):
             except ValueError:
                 days = 3
             if self.get_option("all_logs"):
-                since = ""
+                since = self.get_options("since")
             else:
                 since = "-%ddays" % days
             self.add_journal(since=since)
@@ -122,7 +123,7 @@ class DebianLogs(Logs, DebianPlugin, UbuntuPlugin):
 
         if journal and self.is_installed("systemd"):
             if self.get_option("all_logs"):
-                since = ""
+                since = self.get_options("since")
             else:
                 since = "-%ddays" % days
             self.add_journal(since=since)
--- a/sos/sosreport.py
+++ b/sos/sosreport.py
@@ -21,6 +21,7 @@ import os
 import errno
 import logging
 
+from datetime import datetime
 from argparse import ArgumentParser, Action
 from sos.plugins import import_plugin
 from sos.utilities import ImporterHelper, SoSTimeoutError
@@ -126,6 +127,11 @@ def _get_parser():
                         dest="all_logs", default=False,
                         help="collect all available logs regardless "
                              "of size")
+    parser.add_argument("--since", action="store",
+                        dest="since", default=None,
+                        type=lambda d: datetime.strptime(d, '%Y%m%d%H%M%S'),
+                        help="Use in conjunction with --all-logs, only"
+                             " grab files since date. Format: YYYYMMDDHHMMSS")
     parser.add_argument("--batch", action="store_true",
                         dest="batch", default=False,
                         help="batch mode - do not prompt interactively")
@@ -296,6 +302,12 @@ class SoSReport(object):
                 sys.stderr.write("Unknown preset: '%s'\n" % self.opts.preset)
                 self.preset = self.policy.probe_preset()
                 self.opts.list_presets = True
+
+        # --since depends on --all-logs
+        if self.opts.since is not None and self.opts.all_logs is False:
+            sys.stderr.write("--since can only be used with --all-logs\n")
+            self._exit(1)
+
         # --preset=auto
         if not self.preset:
             self.preset = self.policy.probe_preset()
--- a/sos/__init__.py
+++ b/sos/__init__.py
@@ -54,9 +54,9 @@ _arg_names = [
     'chroot', 'compression_type', 'config_file', 'desc', 'debug', 'del_preset',
     'dry_run', 'enableplugins', 'encrypt_key', 'encrypt_pass', 'experimental',
     'label', 'list_plugins', 'list_presets', 'list_profiles', 'log_size',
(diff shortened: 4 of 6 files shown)


sos/sosreport.py
```diff
@@ -302,6 +308,12 @@ def __init__(self, args):
                 sys.stderr.write("Unknown preset: '%s'\n" % self.opts.preset)
                 self.preset = self.policy.probe_preset()
                 self.opts.list_presets = True
+
+        # --since depends on --all-logs
+        if self.opts.since is not None and self.opts.all_logs is False:
+            sys.stderr.write("--since can only be used with --all-logs\n")
+            self._exit(1)
```
TurboTurtle: This is a nack from me until `--since` no longer depends on `--all-logs`. Way to common a use case for this restriction.

sos/sosreport.py
```diff
@@ -126,6 +127,11 @@ def _get_parser():
                         dest="all_logs", default=False,
                         help="collect all available logs regardless "
                              "of size")
+    parser.add_argument("--since", action="store",
+                        dest="since", default=None,
+                        type=lambda d: datetime.strptime(d, '%Y%m%d%H%M%S'),
+                        help="Use in conjunction with --all-logs, only"
+                             " grab files since date. Format: YYYYMMDDHHMMSS")
```
TurboTurtle: I'm not a fan of specifying a timestamp down to the second. I think this should be day-sensitive at most, and I would personally prefer only requiring the user to specifying a number of days ago to collect.

sos/plugins/logs.py
```diff
@@ -52,7 +52,8 @@ def setup(self):
                 self.add_copy_spec(i)
 
         if self.get_option('all_logs'):
-            self.add_journal(boot="this", allfields=True, output="verbose")
+            print("Getting journal --since %s" % (self.get_option("since")))
```
TurboTurtle: Can't have plugin's calling `print`. Use `_log_info()` if you need to log anything, but I don't think this needs an explicit message.

sos/plugins/logs.py
```diff
@@ -96,7 +97,7 @@ def setup(self):
             except ValueError:
                 days = 3
             if self.get_option("all_logs"):
-                since = ""
+                since = self.get_options("since")
```
TurboTurtle: As my other comment notes - I don't see why this needs to be coupled with `--all-logs`.

sos/plugins/logs.py
```diff
@@ -122,7 +123,7 @@ def setup(self):
 
         if journal and self.is_installed("systemd"):
             if self.get_option("all_logs"):
-                since = ""
+                since = self.get_options("since")
```
TurboTurtle: Same here.

bmr-cymru:
There are quite a few issues here: please follow the [contributor guidelines](https://github.com/sosreport/sos/wiki/Contribution-Guidelines) for branch formatting: in particular, we don't allow merge commits in master, so you'll need to rebase on an up-to-date upstream branch before we could merge this.

For passing the option, since this is handled only in the base `Plugin` class, you don't need to use a global option - those are only needed when individual plugins need to evaluate the option value via the `get_option()` method. Since this is all wired up behind the scenes in private `Plugin` code, it can just access the command line arguments directly (there's some inconsistency here you may notice with existing options, like `--log-size` - this is due to the fact that originally individual plugins had to implement this logic, whereas today it has been moved behind the generic `Plugin.add_copy_spec()` interface - it's just that nobody cleaned up the options handling yet).

Ack to the comments from @TurboTurtle about unhooking this from `--all-logs` - that seems very confusing to me since it literally means "collect *all* logs, including rotated and non-default locations", whereas `--since` is explicitly limiting the scope by time.

The reason we've not done something like this before was that it was not felt to be that useful unless it could filter within individual log files - e.g. the case where a server has a 2GiB `/var/log/messages` due to some config error, runaway logging, or debug output, and we only want to grab the last hour, or day or whatever. Since we're gating the collection at the file level for log files (unlike journal logs which have always had the ability to filter by the time of individual messages), I'm not sure how helpful this will be in practice?

Problem is of course every distro (and potentially every site) may have different configurations for the timestamp values in the on-disk log files. I've so far failed to find either a good heuristic for handling this, or a library that would deal with it for us.

valleedelisle:
I'm sorry about the merge mess-up I'll be fixing this soon (that's the reason why I deleted the branch, maybe there was a better way to do this). 

I'll also remove it from the global options, I'll unhook from `--all-logs` but I think it's important to take `--since` into account even when specifying `--all-logs`. We have use cases for this. I've seen sosreports of 10G+ with `--all-logs`, this is not easily workable. Maybe there's some improvement that can be done on the servers themselves with log rotation, but the point here is that we still need to have these sosreports with `--all-logs` to troubleshoot the important issue.

If we don't specify `--since`, the behavior of `--all-logs` shouldn't change at all. I'll make sure to mention that specifying `--since` effectively changes the behavior on `--all-logs`.

My initial idea was to use the log parsing classes of [os-log-merger](https://github.com/openstack/os-log-merger/blob/master/oslogmerger/oslogmerger.py). They did a very good job at parsing timestamps for all kind of logs. The only concern with this approach is the overhead time it adds to the sosreport build time. Maybe at some point we could evaluate the option to use something like this. It could address non-rotated logs, or very large log files.

Also, in my first tests, I didn't filter only archives, and `--since` was escaping too many important files. It's the reason why I only escape archives now.

I'll submit a new merge request soon. 

Thanks for the feedback, I appreciate it.

valleedelisle:
I've submitted a new merge request here: https://github.com/sosreport/sos/pull/1678 and I've opened an issue where we can discuss if you want https://github.com/sosreport/sos/issues/1673

bmr-cymru:
> If we don't specify --since, the behavior of --all-logs shouldn't change

The point is, we don't _have_ to keep `--all-logs` at all - it's been around a long time, and would cause a bit of adjustment for users, but I think it's very confusing to have an option that says "all" and another one that modifies the scope of "all" to be something less than "all". Our UI has been OK up to now, but it's never perfect - if a new option starts to make it confusing, or to make less sense, then we need to go back and revise it.

Please don't file new PRs - just keep using the same one, or the discussion gets fragmented and hard to follow. Any problems on your branch can be fixed with `git commit --amend` or `git rebase -i`, and the PR branch updated with `git push --force`.

valleedelisle:
I'm sorry Bryn, I don't understand how people would need to get adjusted to `--all-logs`, since the behavior doesn't change at all *unless* they use that new switch which is documented in the `--help`. I just know, for sure, that it was requested multiple times here in support that we find a way to alleviate the weight of `--all-logs`. This was requested by some of the people that uses `sosreport` and I don't believe it affects the people that don't need this behavior. If you have a better way to do this, I'm all open about this.

tacoelite:
Hey, team.  I just wanted to offer a quick point of clarification/question here.  As I understand it:

``--all-logs`` is file selection criteria, and 
``--since`` is date/time selection criteria

... so you could use them in combination with one another or exclusively.  _The behavior of ``--all-logs`` would not change._  For example if a user ran with the args ``--all-logs --since X`` then sosreport would pull all log messages since time X from all files.  The grab-all-files behavior of ``--all-logs`` would not change, sosreport would still get the expanded lot of files, it would just filter that expanded lot according to the date/time criteria.  All of this seems pretty intuitive to me.

But perhaps there's something I'm missing.  Is there a nuance or implication of either argument that I've glossed over?

bmr-cymru:
> or exclusively

Exactly. Which is why linking the two together doesn't make sense: it's completely valid to want to obtain messages from the current log files since a certain time, but this isn't supported by this patch.

> But perhaps there's something I'm missing. Is there a nuance or implication of either argument that I've glossed over?

The current implementation only works at the granularity of whole files - so it's effectively useless without `--all-logs` (other than for `journalctl` data which supports its own `--since` option).

sos/sosreport.py
bmr-cymru: This needs to be much more flexible. It's not reasonable to expect users to always type seconds granularity when it's trivial string munging to expand a shorter date string.

sos/plugins/__init__.py
```diff
@@ -806,13 +808,27 @@ def getmtime(path):
...
+                if _file.endswith('.gz') or re.search(r'-\d{8}$', _file):
+                    fileisarchive = True
+                else:
+                    fileisarchive = False
+
+                if (self.get_option('since') is None or
+                    fileisarchive is False or
+                    (fileisarchive is True and
+                     self.get_option('since') is not None and
+                     self.get_option('since') <= filetime)):
```
bmr-cymru: This needs tidying up: if you're calling a method three times in one conditional, it's a sign it should have a variable in this scope. I think the logic would be more clear if you break it out into several lines: set variables, and then use those in the `if` statement if necessary. It's not obvious to a reader what behaviour you are aiming for here.

sos/plugins/__init__.py
```diff
@@ -1081,7 +1097,7 @@ def add_journal(self, units=None, boot=None, since=None, until=None,
         journal_cmd = "journalctl --no-pager "
         unit_opt = " --unit %s"
         boot_opt = " --boot %s"
-        since_opt = " --since %s"
+        since_opt = " --since '%s'"
```
bmr-cymru: FIxes should be in their own patch, not buried in a large feature patch.

sos/plugins/__init__.py
```diff
@@ -806,13 +808,27 @@ def getmtime(path):
...
+                    continue
+                current_size += filestat[stat.ST_SIZE]
+                filetime = datetime.fromtimestamp(filestat[stat.ST_MTIME])
                 if sizelimit and current_size > sizelimit:
                     limit_reached = True
                     break
-                self._add_copy_paths([_file])
+
+                if _file.endswith('.gz') or re.search(r'-\d{8}$', _file):
+                    fileisarchive = True
```
bmr-cymru: A `.gz` isn't the only possible log compression: this should support other formats, or hook into `Policy` to determine the correct value for the system, or a given log file.

bmr-cymru:
> I'm sorry Bryn, I don't understand how people would need to get adjusted to --all-logs, since the behavior doesn't change at all unless they use that new switch which is documented in the --help.

You'd be surprised. People will find the new option, make assumptions about it, and try to use it: right now it's limited, and that will confuse them. People have complained in the past because using `--profile` excludes certain plugins, for e.g.

I'm not objecting to the functionality, but I do need it to be as clear, and un-confusing as possible: right now, as the patch stands, it **is** confusing. The usage message just says "Use this with --all-logs". **Why?** Why can't I use it with the defaults? What are the limitations preventing that? Why do I get different behaviours when collecting journals vs. syslog data? These are exactly the kinds of questions that users will ask when faced with this change. Believe me: we have been doing this for a while.

It's OK to give users a feature that has somewhat limited functionality sometimes, but it's not OK to expect them to read your source code to know what the limitations are and why they exist.

At a minimum, if the current version is going in, this needs to be documented in the man page, to explain the difference in behaviour between journal and log files, to have clear information in the usage message, and to produce a helpful error message if the options conflict. Alternately, you could extend the patch set to address file date filtering (extending `Policy` to provide date patterns based on distro and configuration would be one place to start), and remove the requirement for `--all-logs` completely.

sos/plugins/__init__.py
valleedelisle: Fixed in latest release

sos/plugins/__init__.py
valleedelisle: I'll do that next time I guess?

sos/plugins/__init__.py
valleedelisle: I've included all the most common compression formats in a compiled regex.

PR #1678, opened 2019-05-14
[sosreport] [rfe] grab logs since date

valleedelisle:
Adds a --since switch that takes a date as argument.  This switch
will skip the archive files with a mtime older than the date.

Also, --since affects journalctl execution for --all.

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -26,6 +26,7 @@ import errno
 # PYCOMPAT
 import six
 from six.moves import zip, filter
+from datetime import datetime
 
 # FileNotFoundError does not exist in 2.7, so map it to IOError
 try:
@@ -747,7 +748,8 @@ class Plugin(object):
         matches any of the option names is returned.
         """
 
-        global_options = ('verify', 'all_logs', 'log_size', 'plugin_timeout')
+        global_options = ('verify', 'all_logs', 'log_size', 'since',
+                          'plugin_timeout')
 
         if optionname in global_options:
             return getattr(self.commons['cmdlineopts'], optionname)
@@ -774,11 +776,20 @@ class Plugin(object):
     def _add_copy_paths(self, copy_paths):
         self.copy_paths.update(copy_paths)
 
-    def add_copy_spec(self, copyspecs, sizelimit=None, tailit=True, pred=None):
-        """Add a file or glob but limit it to sizelimit megabytes. If fname is
-        a single file the file will be tailed to meet sizelimit. If the first
-        file in a glob is too large it will be tailed to meet the sizelimit.
+    def add_copy_spec(self, copyspecs, sizelimit=None, maxage=None,
+                      tailit=True, pred=None):
+        """Add a file or glob but limit it to sizelimit megabytes. Collect
+        files with mtime not older than maxage hours.
+        If fname is a single file the file will be tailed to meet sizelimit.
+        If the first file in a glob is too large it will be tailed to meet
+        the sizelimit.
         """
+        since = None
+        if self.get_option('since'):
+            since = self.get_option('since')
+
+        logarchive_pattern = re.compile(r'.*((\.(zip|gz|bz2|xz))|[-.][\d]+)$')
+
         if not self.test_predicate(pred=pred):
             self._log_info("skipped copy spec '%s' due to predicate (%s)" %
                            (copyspecs, self.get_predicate(pred=pred)))
@@ -819,6 +830,21 @@ class Plugin(object):
                 except OSError:
                     return 0
 
+            def time_filter(path):
+                """ When --since is passed, or maxage is coming from the
+                plugin, we need to filter out older files """
+
+                if logarchive_pattern.search(path) is None:
+                    return True
+                filetime = datetime.fromtimestamp(getmtime(path))
+                if ((since and filetime < since) or
+                   (maxage and (time()-filetime < maxage*3600))):
+                    return False
+                return True
+
+            if since or maxage:
+                files = list(filter(lambda f: time_filter(f), files))
+
             files.sort(key=getmtime, reverse=True)
             current_size = 0
             limit_reached = False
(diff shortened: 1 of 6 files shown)


bmr-cymru:
Let's keep all the discussion on this change in one place.

bmr-cymru:
@valleedelisle it looks like you deleted your repository for the previous PR - please don't do that for sos pull requests. It fragments the discussion needlessly, and makes it difficult to track down the original version of changes in the future if that is required. There's no need to burn through PRs just because a branch had some problems - just modify it on your end with `git commit --amend`, or `git rebase -i`, and then update the PR branch with `git push --force`.

bmr-cymru:
Since you've nuked the branch for #1677 I am not able to re-open it, so let's use this to continue the discussion. People can refer back to the original for context if needed.

sos/plugins/__init__.py
```diff
@@ -829,13 +839,31 @@ def getmtime(path):
...
+                    fileisarchive = False
+
+                fileisnew = True
+
+                if maxage:
+                    now = time()
+                    if (now-filetime > maxage*3600):
+                        fileisnew = False
+
+                if ((maxage is None or (maxage and fileisnew is True)) and
```
pmoravec: Maybe worth describing the complex condition in a comment?

sos/plugins/__init__.py
```diff
@@ -829,13 +839,31 @@ def getmtime(path):
                     self._log_debug("skipping forbidden path '%s'" % _file)
                     continue
                 try:
-                    current_size += os.stat(_file)[stat.ST_SIZE]
+                    filestat = os.stat(_file)
                 except OSError:
                     self._log_info("failed to stat '%s'" % _file)
+                    continue
+                current_size += filestat[stat.ST_SIZE]
+                filetime = datetime.fromtimestamp(filestat[stat.ST_MTIME])
```
pmoravec: Can't this be merged with `files.sort(key=getmtime, reverse=True)` called some lines above? Or maybe before this? Such that:

- we first cut off files that are too old to `--since` option
- then we sort the remaining ones per the same `mtime` parameter

Maybe we can also prevent the double-calling of `stat` to get `mtime`, then.

pmoravec:
I like this PR more then mine #1586 that I will close. Anyway I think my latest comment can be easily applied by following the patch from my PR.

@valleedelisle , do you see the suggestion reasonable and would you be able to apply it in very few weeks?

valleedelisle:
@pmoravec This is very reasonable and great idea. I'm working on applying these recommendations and, hopefully, will have something to review by the end of the day. Thanks for the constructive and technical feedback.

sos/plugins/__init__.py
valleedelisle: I've moved everything before the loop and run a filter on the files list instead of filtering while looping

pmoravec:
Could you please squash the two commits into one (via barch rebase and force push)? Since the newer is just fixing the older. If the newer would be an independent change, an extra commit makes sense.

Anyway, testing the PR I found out one more generic problem: reproducer:

    sosreport --since 20190808 --batch --build -o auditd

will collect all `/var/log/audit/*log` files, despite some of them have `mtime` days ago. The reason is:
- `add_copy_spec` is called to a directory (`/var/log/audit`)
- expanding `copyspec` at https://github.com/sosreport/sos/blob/master/sos/plugins/__init__.py#L812 just calls `glob.glob`, to expand `*` but not to traverse a dir or follow symlink
- so `since` (and also `sizelimit!) is checked against `/var/log/audit` _directory_

And since directory `mtime` is usually `mtime` of the newest file, `/var/log/audit` is added to `copy_paths` at https://github.com/sosreport/sos/blob/master/sos/plugins/__init__.py#L841.

Ad "and also `sizelimit(!)`: this means, `add_copy_spec` called with a low size limit on a directory will collect whole directory content - since `stat` of directory returns amount of disk space required to store the directory itself - without its content..

@bmr-cymru , is it worth replacing https://github.com/sosreport/sos/blob/master/sos/plugins/__init__.py#L1198 by `os.walk`? Or shouldn't we move the `sizelimit` and `since` filter to _do_copy_path (that sounds wierd)?


(but this issue is independent on the `--since` option, though it would be great to fix it altogether, I think)

sos/plugins/__init__.py
```diff
@@ -819,6 +830,21 @@ def getmtime(path):
                 except OSError:
                     return 0
 
+            def time_filter(path):
+                """ When --since is passed, or maxage is coming from the
+                plugin, we need to filter out older files """
+
+                if archive_pattern.search(path) is None:
```
pmoravec: What is the purpose of `archive_pattern`? Current implementation means that `--since` is applied *only* on files matching the pattern. So if I edit auditd plugin to collect `/var/log/audit/*`, then `sosreport --since 20190808 --all-logs --batch --build` will collect also `/var/log/audit/audit.log.4` with mtime 4 days in past.

I agree we shall apply the `--since` to (logrotated) log files only - not to current log file (well this could be arguable) and definitely `--since` cant be applied to config files.

Or maybe `archive_pattern` just needs adding something like `log.[1-9]`? E.g.:

    archive_pattern = re.compile('.*(.zip|.gz|.bz2|.xz|[-.][0-9]+)$')

Or even better:

    archive_pattern = re.compile('.*.log.+$')

to cover *any* logrotated log file? (this can be too generic and we can skip collecting some config file like `/etc/foo/bar.logging.conf` - I prefer the prior variant)

pmoravec: Improved suggestion:

    archive_pattern = re.compile('.*(\.zip|\.gz|\.bz2|\.xz|[\.-]\d+)$')

(`.` to `\.` and `[0-9]` to `\d`)

valleedelisle: The purpose of the `archive_pattern` is that, even if we `--since`, we want to include more static log files (`dmesg` for example). We also don't want to affect `/etc`.

If we `'.*.log.+$'`, some logs like `(httpd/(access|error)_log|messages)-XXXXXXXX` won't be filtered. 

I think the most bulletproof approach would be parsing the `logrotate`'s config, but that would add some overhead.

I like the idea of `[-.][0-9]+)$`, that should fix it for `audit` if we change the `copyspecs` passed to `add_copy_spec`.

pmoravec:
(sorry for not keeping @valleedelisle as the author of the commit; wrong merge strategy used that time)

PR #1680, opened 2019-05-17
[katello] support both locations of qpid SSL certs

pmoravec:
Newer katello versions deploy certs for qpid to
/etc/pki/pulp/qpid/client.crt certs instead of
/etc/pki/katello/qpid_client_striped.crt .

Sosreport should use either of the location that exists, to successfully
run few qpid-stat commands.

Resolves: #1680

---

### DIFF
--- a/sos/plugins/katello.py
+++ b/sos/plugins/katello.py
@@ -10,6 +10,7 @@
 
 from sos.plugins import Plugin, RedHatPlugin
 from pipes import quote
+import os.path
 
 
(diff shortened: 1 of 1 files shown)


pmoravec:
We can't rely on `katello` version to detect proper cert file, since upgraded katello still uses old location (as far as I checked). So the dummy "use newer file, if it exists, or older one otherwise" approach seems to be the best.

PR #1682, opened 2019-05-20
[grub2] call grub2-config with --no-grubenv-update when appropriate

pmoravec:
On some newer grub2 versions, grub2-config removes extra args in
$kernel_opts until --no-grubenv-update option is used.

Test if the option is present in grub2-config --help and if so, use it.

Resolves: #1682

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -122,6 +122,9 @@ class SoSPredicate(object):
     #: Package presence list
     packages = []
 
+    # Command output inclusion pairs {'cmd': 'foo --help', 'output': 'bar'}
+    cmd_outputs = []
+
     def __str(self, quote=False, prefix="", suffix=""):
         """Return a string representation of this SoSPredicate with
             optional prefix, suffix and value quoting.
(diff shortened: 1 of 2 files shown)


pmoravec:
The test is a worth candidate for a predicate. But there are two competitive options:

- have a predicate "command run with argument X (i.e. `--help` or `-h`) returns output containing Y
  - then we will have to have also `invert` option in predicate to allow tests like "`grub2-mkconfig --help` does *not* contain `--no-grubenv-update`"
  - in our specific use case, we would have to run the predicate (and hence `grub2-mkconfig --help` command) twice, to evaluate individual two test cases

- or, have a more generic predicate "a command X (i.e. `grub2-mkconfig --help`) returns output containing Y"
  - then we can also use `self.test_predicate(pred=SosPredicate(cmdOutputContains({"grub2-mkconfig --help", "--no-grubenv-update"}))` to simply determine if the argument should be used or not

I prefer the second, more generic option. The code would be also better to understand (one would see command and its option altogether), "just" I can't find out concise enough predicate type name (`kmod`, `services` and.. `cmdOutputContains`?).

Some preferences, @TurboTurtle or @bmr-cymru ?

pmoravec:
2nd option used. Predicates enhanced by single command output check, not a *list* of commands - so far I dont see much usage of gating something against a group of commands, and the current enhancement is sufficient for this issue/PR.

pmoravec:
rebased the PR and enhance for a list of commands, not only one.

pmoravec:
Rebased the PR to include the services predicate merged meantime into upstream.

PR #1684, opened 2019-05-22
[plugins] fix mangled command filenames collisions

pmoravec:
- Command filenames collisions must be tested tested against absolute path, not
  relative. Otherwise false positive results about no collisions are returned.
    
- Filenames must have shorter lenghts in case collision happens. Since that appends
  "_2" to the name that still must constitute a filename of supported length.
    
Resolves: #1684

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -959,20 +959,27 @@ class Plugin(object):
         if subdir:
             # only allow a single level of subdir to be created
             plugin_dir += "/%s" % subdir.split('/')[0]
-        outfn = os.path.join(self.commons['cmddir'], plugin_dir,
-                             self._mangle_command(exe))
+        outdir = os.path.join(self.commons['cmddir'], plugin_dir)
+        outfn = self._mangle_command(exe)
 
         # check for collisions
-        if os.path.exists(outfn):
-            inc = 2
+        if os.path.exists(os.path.join(self.archive.get_tmp_dir(),
+                                       outdir, outfn)):
+            inc = 1
+            name_max = self.archive.name_max()
             while True:
-                newfn = "%s_%d" % (outfn, inc)
-                if not os.path.exists(newfn):
+                suffix = ".%d" % inc
+                newfn = outfn
+                if name_max < len(newfn)+len(suffix):
+                    newfn = newfn[:(name_max-len(newfn)-len(suffix))]
+                newfn = newfn + suffix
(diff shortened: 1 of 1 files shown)


pmoravec:
Artificial reproducer: add to `sos/plugins/kernel.py`:

    @@ -75,7 +75,6 @@ class Kernel(Plugin, RedHatPlugin, Debia
     
             if extra_mod_paths:
                 self.add_cmd_output("find %s -ls" % " ".join(extra_mod_paths))
    +            self.add_cmd_output("find %s -ls" % " ".join(extra_mod_paths+["/tmp"]))
 
to collect one extra `find` command on many module paths including `/tmp`.

Nowadays, *just* the `find .. /tmp` is collected because `os.path.exists(outfn)` tests *relative* path instead of absolute (and that file usually does not exist in $PWD). So we need to:

1) test absolute path
2) shorten filename to supper "_2" suffixes

sos/plugins/__init__.py
```diff
@@ -949,7 +949,8 @@ def file_grep(self, regexp, *fnames):
         return grep(regexp, *fnames)
 
     def _mangle_command(self, exe):
-        name_max = self.archive.name_max()
+        # shorten the name_max such that we can append "_2" in case of conflict
+        name_max = self.archive.name_max() - 5
```
bmr-cymru: Why five?

sos/plugins/__init__.py
```diff
@@ -963,11 +964,12 @@ def _make_command_filename(self, exe, subdir=None):
                              self._mangle_command(exe))
 
         # check for collisions
-        if os.path.exists(outfn):
+        if os.path.exists(os.path.join(self.archive.get_tmp_dir(), outfn)):
```
bmr-cymru: Why does this even count from '2'? It's in the original, but why keep it broken? A more conventional approach would be like:

```
collision
collision.1
collision.2
```

Etc. Or in the event that we are running up against the limit:

```
alongcollision
alongcollisi.1
alongcollisi.2
...
```

sos/plugins/__init__.py
pmoravec: Any value bigger than1 would fit - I just wanted to have enough free space for _10, _100 etc., just in case. But this is redundant change, now.

sos/plugins/__init__.py
pmoravec: OK will update the PR accordingly; then the " - 5" won't be necessary.

pmoravec:
PR updated.

Issue #1688, opened 2019-06-04
[kernel] take a lot of time to collect trace file


miztake:
sosreport now collects /sys/kernel/debug/tracing/trace.
However, in an environment where many traces are collected,
it takes a lot of time to collect sosreport.
For example, in some environments where "softirq" is high,
it may take over 2 hours for sosreport to complete.

We will need to add a timeout to the collection of /sys/kernel/debug/tracing/trace.

PR #1689, opened 2019-06-04
[kernel] Add timeout when collecting trace file

miztake:
Updates the plugin to add timeout when collecting trace file.
Collecting a whole trace file may take a lot of time.

Resolves: #1688

---

### DIFF
--- a/sos/plugins/kernel.py
+++ b/sos/plugins/kernel.py
@@ -26,6 +26,8 @@ class Kernel(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
         ("with-timer", "gather /proc/timer* statistics", "slow", False)
     ]
 
+    trace_timeout = 60
+
     def get_bpftool_prog_ids(self, prog_file):
         out = []
         try:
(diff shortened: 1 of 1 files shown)


sos/plugins/kernel.py
```diff
@@ -133,6 +136,12 @@ def setup(self):
             clocksource_path + "current_clocksource"
         ])
 
+        if self.get_option('all_logs'):
+            self.trace_timeout=0
+
+        self.add_cmd_output("cat /sys/kernel/debug/tracing/trace",
```
TurboTurtle: Nack on calling cat to collect this. Also, the whole-plugin timeout should be killing this collection well before the 2-hour mark unless you're running sosreport with `--plugin-timeout=0` or similar.

Personally I think we should just add this path as a forbidden path wholesale as I don't think we've ever purposely intended to capture this path. However, if it's useful it should be gated to either add the path as forbidden or not, not add it and then circumvent the mechanism we just leveraged.

If this is actually useful, we could gate it behind a plugin option, e.g. `-k kernel.trace=on`. This should be 'off' by default which should trigger adding it as a forbidden path. If the option is set to 'on' then we maintain the behavior today.

miztake: Thank you for your comment.
I'll consider how to switch the collection of path using new plugin option.
I close this PR and create a new one.

pmoravec:
(technically, no need to close a PR. It is sufficient to modify the commit(s) in your branch and forcefully push to github. The PR will be automatically updated to point to the new content - this should hold generally for any github project / PR. But a new PR is also sound option)

miztake:
> (technically, no need to close a PR. It is sufficient to modify the commit(s) in your branch and forcefully push to github. The PR will be automatically updated to point to the new content - this should hold generally for any github project / PR. But a new PR is also sound option)

Thank you for the advice.
From now on I'll use the method you were guided.

PR #1690, opened 2019-06-04
[Plugin] Add method to scrub certs/keys from collected files

TurboTurtle:
This adds a new do_file_private_sub() method to the Plugin class to
allow for easy, and standardized scrubbing of private data such as
certificates or key pairs from files collected by sos, in much the same
way that do_cmd_private_sub() provides that functionality for command
output.

This method only takes a path or path regex representing the
filename/pattern to match against. Matches are replaced with
"-----SCRUBBED $thing" where $thing is the specific type of content
being replaced as provided in the collected file content. For example, a
private key would be replaced with "-----SCRUBBED RSA PRIVATE KEY-----".

This is done so that support representatives can still know the type of
information that was collected, without knowing the actual sensitive
content.

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -502,6 +502,35 @@ class Plugin(object):
             replacements = None
         return replacements
 
+    def do_file_private_sub(self, pathregex):
+        '''Scrub certificate/key/etc information from files collected by sos.
+
+        Files matching the provided pathregex are searched for content that
+        resembles certificate, ssh keys, or similar information. Any matches
+        are replaced with "-----SCRUBBED $thing" where $thing is the specific
+        type of content being replaced, e.g. "-----SCRUBBED RSA PRIVATE KEY" so
+        that support representatives can at least be informed of what type of
+        content it was originally.
+
+        Positional arguments:
+            :param pathregex: A string or regex of a filename to match against
+        '''
+        certmatch = re.compile("-*BEGIN.*?-*END", re.DOTALL)
+        match = re.compile(pathregex).match
+        file_list = [f for f in self.copied_files if match(f['srcpath'])]
+        for i in file_list:
+            path = i['dstpath']
+            if not path:
+                continue
+            self._log_debug("Scrubbing %s for certificate content" % path)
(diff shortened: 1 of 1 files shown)


sos/plugins/__init__.py
```diff
@@ -502,6 +502,35 @@ def do_cmd_output_sub(self, cmd, regexp, subst):
...
+        resembles certificate, ssh keys, or similar information. Any matches
+        are replaced with "-----SCRUBBED $thing" where $thing is the specific
+        type of content being replaced, e.g. "-----SCRUBBED RSA PRIVATE KEY" so
+        that support representatives can at least be informed of what type of
+        content it was originally.
+
+        Positional arguments:
+            :param pathregex: A string or regex of a filename to match against
+        '''
+        certmatch = re.compile("-*BEGIN.*?-*END", re.DOTALL)
```
pmoravec: cf. https://github.com/sosreport/sos/blob/master/sos/plugins/__init__.py#L497 - isn't it worth having standardised r.e. for this? (i.e. have the r.e. string stored in a constant, referenced from both places?)

sos/plugins/__init__.py
```diff
@@ -502,6 +502,35 @@ def do_cmd_output_sub(self, cmd, regexp, subst):
...
+        for i in file_list:
+            path = i['dstpath']
+            if not path:
+                continue
+            self._log_debug("Scrubbing %s for certificate content" % path)
+            readable = self.archive.open_file(path)
+            content = readable.read()
+            if not isinstance(content, six.string_types):
+                content = content.decode('utf-8', 'ignore')
+            result, replacements = certmatch.subn('-----SCRUBBED', content)
```
pmoravec: Maybe also here - is it worth having a standardized / constant string for '-----SCRUBBED' ? (probably not such important like *what* to scrub).

pmoravec:
Ack in general. Just the new method basically executes `do_file_sub` method for a list of collected files - isn't it worth `do_file_private_sub()` doing just an iteration over `pathregex` and call `do_file_sub()`? To prevent code duplication.

Issue #1299, opened 2018-05-18
[kernel,system] warnings on stdout about Permission denied in python3


pmoravec:
Running sosreport under python3, I see:

      Running 1/2: kernel...
    [archive:TarFileArchive] Unable to add '/var/tmp/sos.d0dqtohj/sosreport-hostname-2018-05-18-sboruce/sys/kernel/debug/tracing/trace_marker_raw' to FileCacheArchive: [Errno 13] Permission denied: '/var/tmp/sos.d0dqtohj/sosreport-hostname-2018-05-18-sboruce/sys/kernel/debug/tracing/trace_marker_raw'
    [archive:TarFileArchive] Unable to add '/var/tmp/sos.d0dqtohj/sosreport-hostname-2018-05-18-sboruce/sys/kernel/debug/tracing/free_buffer' to FileCacheArchive: [Errno 13] Permission denied: '/var/tmp/sos.d0dqtohj/sosreport-hostname-2018-05-18-sboruce/sys/kernel/debug/tracing/free_buffer'
    [archive:TarFileArchive] Unable to add '/var/tmp/sos.d0dqtohj/sosreport-hostname-2018-05-18-sboruce/sys/kernel/debug/tracing/trace_marker' to FileCacheArchive: [Errno 13] Permission denied: '/var/tmp/sos.d0dqtohj/sosreport-hostname-2018-05-18-sboruce/sys/kernel/debug/tracing/trace_marker'
      Running 2/2: system...    
    [archive:TarFileArchive] Unable to add '/var/tmp/sos.d0dqtohj/sosreport-hostname-2018-05-18-sboruce/proc/sys/net/ipv4/route/flush' to FileCacheArchive: [Errno 13] Permission denied: '/var/tmp/sos.d0dqtohj/sosreport-hostname-2018-05-18-sboruce/proc/sys/net/ipv4/route/flush'
    [archive:TarFileArchive] Unable to add '/var/tmp/sos.d0dqtohj/sosreport-hostname-2018-05-18-sboruce/proc/sys/net/ipv6/route/flush' to FileCacheArchive: [Errno 13] Permission denied: '/var/tmp/sos.d0dqtohj/sosreport-hostname-2018-05-18-sboruce/proc/sys/net/ipv6/route/flush'

python2 does not exhibit those errors, because py3 newly raises the error when `shutil.copystat` is called to a file without read permissions.

And really:

[...]

PR #1691, opened 2019-06-05
[kernel] Don't collect trace file by default

miztake:
Updates the plugin to don't collect trace file by default. Collecting
trace file may take a lot of time, so trace file is not collected
by default, and use the new plug-in option when collecting.

Resolves: #1688

---

### DIFF
--- a/sos/plugins/nvme.py
+++ b/sos/plugins/nvme.py
@@ -21,6 +21,7 @@ class Nvme(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
         return [dev for dev in sys_block if dev.startswith('nvme')]
 
     def setup(self):
+        self.add_copy_spec("/etc/nvme/discovery.conf")
         self.add_cmd_output([
             "nvme list",
             "nvme list-subsys",
@@ -39,6 +40,5 @@ class Nvme(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
                 "nvme show-regs /dev/%s" % dev,
                 "nvme get-ns-id /dev/%s" % dev
             ])
-            self.add_copy_spec("/etc/nvme/discovery.conf")
 
 # vim: set et ts=4 sw=4 :


sos/plugins/kernel.py
```diff
@@ -138,6 +139,9 @@ def setup(self):
             # and may also cause softlockups
             self.add_copy_spec("/proc/timer*")
 
+        if not self.get_option("trace"):
+            self.add_forbidden_path("/sys/kernel/debug/tracing/trace")
```
d-hatayama: How about just as "with-timer"? I think it's more consistent.

    if self.get_option("trace"):
        self.add_copy_spec("/sys/kernel/debug/tracing/trace")

TurboTurtle: Personally I prefer it as a separate `kernel.trace`. The existing `with-timer` option is for different information and  it doesn't necessarily make sense to have both gated behind that option.

d-hatayama: I didn't mean to change the option from kernel.trace to kernel.with-timer. I meant to make the coding consistent with what is done for with-timer where add_copy_spec() is called only when necessary.

But there is an overlook to me that the pattern /sys/kernel/debug/tracing/* covers multiple files and it seems certainly verbose to write add_copy_spec() each file one by one without add_forbidden_path().

I then found add_forbidden_path() has already been used in kernel.py for tracing information:

        # FIXME: provide a a long-term solution for #1299
        self.add_forbidden_path([
            '/sys/kernel/debug/tracing/trace_pipe',
            '/sys/kernel/debug/tracing/README',
            '/sys/kernel/debug/tracing/trace_stat/*',
            '/sys/kernel/debug/tracing/per_cpu/*',
            '/sys/kernel/debug/tracing/events/*',
            '/sys/kernel/debug/tracing/free_buffer',
            '/sys/kernel/debug/tracing/trace_marker',
            '/sys/kernel/debug/tracing/trace_marker_raw',
            '/sys/kernel/debug/tracing/instances/*/per_cpu/*/snapshot_raw',
            '/sys/kernel/debug/tracing/instances/*/per_cpu/*/trace_pipe*',
            '/sys/kernel/debug/tracing/instances/*/trace_pipe'
        ])

so I think Mizuta-san's first patch is consistent enough this way.

miztake:
Hi, all.
Do you have any other comments on this PR?
Is this PR mergeable?

pmoravec:
Sounds good to me.

miztake:
Dear maintainer,

Thank you for merging this PR.

In order for this PR to work correctly, #1695 must also be modified at the same time.
So, please merge #1695 together.

pmoravec:
Hello,
I can't see where the PR was merged. d920a92 does not seem to be in git log (it's parent fc41da4 is followed by fb6363c and I am not aware of anyone with commit rights who did commit for this PR). Also this PR has no commits at this time, what is strange.

Could you please file (probably a new) PR for this issue, with proper branch on your forked repo (ideally rebased to current upstream to prevent merge conflicts)? Even better, with 2 commits: first to fix #1695, second to fix this issue? Or you can amend the PR #1695 by adding one new commit to fix this issue within that PR.

Thanks.

PR #1693, opened 2019-06-05
[openvswitch] Collect ovs threads list in ps and PMD thread details in procfs

links84:
Get detailed ps output of all ovs-vswitchd threads including last cpu known to run on, priority etc.
Also copy /proc/<PMD thread id>/{environ,cgroup,maps,numa_maps,limits,status} to indicate configured cpu affinity.
These information are critical for NFV openstack deployments.

---

### DIFF
--- a/sos/plugins/openvswitch.py
+++ b/sos/plugins/openvswitch.py
@@ -12,7 +12,7 @@ from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
 
 from os.path import join as path_join
 from os import environ
-
+import re
 
 class OpenVSwitch(Plugin):
     """ OpenVSwitch networking
@@ -42,6 +42,16 @@ class OpenVSwitch(Plugin):
             "/var/run/openvswitch/ovsdb-server.pid",
             "/var/run/openvswitch/ovs-vswitchd.pid"
         ])
+        # Collect ovs-vswitchd's pmd thread details in procfs.
+        pid = open("/var/run/openvswitch/ovs-vswitchd.pid").read().splitlines()[0]
+        # List TID & thread command name.
+        ps_tidlist_cmd = "ps -p %s -Lo tid=,comm= " % pid
+        tid_list = filter( re.compile(r'pmd').search ,  self.call_ext_prog(ps_tidlist_cmd)['output'].splitlines() )
+        # filter out list elements with string 'pmd' & if not empty, capture process details.
+        if tid_list:
+            for tid in tid_list:
+                for pf in ["environ", "cgroup", "maps", "numa_maps", "limits","status"]:
+                    self.add_copy_spec("/proc/%s/%s" % (tid.split()[0], pf))
 
         self.add_cmd_output([
             # The '-s' option enables dumping of packet counters on the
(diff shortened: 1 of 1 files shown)


links84:
Hey folks,
Can someone review this patch & help in approval if it looks good?
Thanks.

sos/plugins/openvswitch.py
```diff
@@ -42,6 +43,21 @@ def setup(self):
             "/var/run/openvswitch/ovsdb-server.pid",
             "/var/run/openvswitch/ovs-vswitchd.pid"
         ])
+        # Collect ovs-vswitchd's pmd thread details in procfs.
+        pid = open("/var/run/openvswitch/ovs-vswitchd.pid").read()\
+            .splitlines()[0]
```
pmoravec: Worth having in try-catch block in case the file cant be opened or read - see e.g. https://github.com/sosreport/sos/blob/master/sos/plugins/pulp.py#L38-L50 for example.

sos/plugins/openvswitch.py
```diff
@@ -42,6 +43,21 @@ def setup(self):
...
+        # List TID & thread command name.
+        ps_tidlist_cmd = "ps -p %s -Lo tid=,comm= " % pid
+        tid_list = filter(
+                re.compile(r'pmd').search,
+                self.call_ext_prog(ps_tidlist_cmd)['output'].splitlines())
```
pmoravec: Same here: what if "ps" would fail (unprobable) or have no output (e.g. the process with given PID does not (further) exist)?

sos/plugins/openvswitch.py
```diff
@@ -42,6 +43,21 @@ def setup(self):
             "/var/run/openvswitch/ovsdb-server.pid",
             "/var/run/openvswitch/ovs-vswitchd.pid"
         ])
+        # Collect ovs-vswitchd's pmd thread details in procfs.
+        pid = open("/var/run/openvswitch/ovs-vswitchd.pid").read()\
+            .splitlines()[0]
+        # List TID & thread command name.
+        ps_tidlist_cmd = "ps -p %s -Lo tid=,comm= " % pid
+        tid_list = filter(
+                re.compile(r'pmd').search,
```
pmoravec: Maybe a comment with example of filtered output can clarify the regexp trick..

sos/plugins/openvswitch.py
```diff
@@ -42,6 +43,21 @@ def setup(self):
...
+                self.call_ext_prog(ps_tidlist_cmd)['output'].splitlines())
+        # filter out list elements with string 'pmd' & if not empty,
+        # capture process details.
+        if tid_list:
```
pmoravec: IMHO redundant test since you gate just `for tid in tid_list` cycle by this test.

pmoravec:
Few comments added, plus it is worth squashing the 2 commits into 1.

sos/plugins/openvswitch.py
links84: Thanks for your inputs, Just to clarify , do you mean a stale pid file is present but the process is not running ( killed perhaps) ?

pmoravec: Taking my comment back here - even in such situations when "ps" command fails or returns nothing, the worst what we get is empty `tid_list`. And that is handled below well.

pmoravec:
Hello, could you please apply the comments and resolve merge conflict?

Thanks in advance.

sos/plugins/openvswitch.py
bmr-cymru: If you already have a PID from a pidfile this seems like an unnecessarily complex way of obtaining the TIDs. Something like this should be adequate:

```python
    proc_files = ["environ", "cgroup", "maps", "numa_maps", "limits", "status"]
    try:
        pid = open("/var/run/openvswitch/ovs-vswitchd.pid").read().splitlines()[0]
        self.add_copy_spec(["/proc/%s/task/*/%s" % (pid, pf) for pf in proc_files])
    except:
         self._log_error("Could not read ovs-vswitch PID")
```

links84:
Apologize for the delay. Working on this now.

links84:
I lost my branch and had considerable issue to resolve the conflicts with latest commits. So i created a new pull request.
https://github.com/sosreport/sos/pull/1962
Apologize for the inconvenience.

Issue #1692, opened 2019-06-05
[plugins] Should add_forbidden_path() be a partial match?


miztake:
As a result of calling add_forbidden_path("/sys/kernel/debug/tracing/trace"),
the following files were not collected.

  - /sys/kernel/debug/tracing/trace
  - /sys/kernel/debug/tracing/trace_clock
  - /sys/kernel/debug/tracing/trace_options

_path_in_path_list() is a partial match as follows.

https://github.com/sosreport/sos/blob/fc41da43106ba3d779a19cc910a2f430faf4a030/sos/plugins/__init__.py#L65-L66

To change _path_in_path_list() to an exact match,
we should change "p in path" to "p == path".

Should add_forbidden_path() be a partial match?

PR #856, opened 2016-07-18
[openstack_horizon] fix regex in postproc method

poornimakshirsagar:
"The regex needs to handle in openstack_horizon:
currently the configurations are listed by
"/etc/openstack-dashboard/local_settings"
causing below errors:

regex substitution failed for '/etc/openstack-dashboard/local_settings.d/_11_rcue_theme.pyc'

Purposing a fix by changing the regex to match only intended configuration
files (_.conf_) in this directory.

Signed-off-by: Poornima M. Kshirsagar pkshiras@redhat.com

PR #1695, opened 2019-06-06
[plugins] Change forbidden_path from partial-match to exact-match

miztake:
forbidden_path is evaluated on partial-match.
However, it will be correct to evaluate on exact-match.

Resolves: #1692

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -63,7 +63,7 @@ def _mangle_command(command, name_max):
 
 
 def _path_in_path_list(path, path_list):
-    return any(p in path for p in path_list)
+    return any(p == path for p in path_list)
 
 
 def _node_type(st):


TurboTurtle:
We have a lot of bits like `add_forbidden_path('/some/dir/*')`, so this needs to be a regex match, rather than exact string match.

miztake:
Thank you for your comment.

regex match will be expanded by glob.glob().

https://github.com/sosreport/sos/blob/fc41da43106ba3d779a19cc910a2f430faf4a030/sos/plugins/__init__.py#L713-L716

After being expanded by glob.glob(), forbidden_path is evaluated with _path_in_path_list().
Therefore, I think that _path_in_path_list() does not have to be a partial match.

TurboTurtle:
Good catch, I was totally overlooking we did glob expansion there.

With that in mind, this change makes sense to me.

miztake:
Hi, all.
Do you have any other comments on this PR?
Is this PR mergeable?

pmoravec:
I checked all calls of `self.add_forbidden_path` if the forbidden paths can't be a prefix of a path the given plugin wants to collect. Few potential issues found:

One forbidden path sounds rather not working at all: https://github.com/sosreport/sos/blob/master/sos/plugins/openstack_horizon.py#L44 - When the `glob.glob("*.py[co]")` can return anything..? (cc @lyarwood or @poornimakshirsagar for review - I got collected all /etc/openstack-dashboard/local_settings.d/_11_rcue_theme.py* files (cf. #856)).

In `vdsm` plugin, I am unsure if https://github.com/sosreport/sos/blob/master/sos/plugins/vdsm.py#L149-L151 does not intentionally forbid files or paths like e.g. `/var/run/vdsm/<some-subpath>/storage-options/` - @igoihman , could you comment here? Currently we forbid `/var/run/vdsm/<some-subpath>/storage*` while we will change it to  exact match forbid of `/var/run/vdsm/<some-subpath>/storage` (and the same for `.vfd` and `/isoUploader`). Is that OK?

In `rear` plugin, we *skip* collecting files like `/var/log/rear/whatever.iso.log` when `/var/log/rear/whatever.iso`is present (and skipped). That sounds unintentional and this PR will collect the `/var/log/rear/whatever.iso.log` files - I guess that is OK, @TurboTurtle ?

Same for `openswan` where we *skip* collecting `/etc/ipsec.d/<whatever>.[db|secrets]<anything>` when `/etc/ipsec.d/<whatever>.[db|secrets]` is present. But I doubt this is correct behaviour and such files even exist.

In `x11` plugin, the https://github.com/sosreport/sos/blob/master/sos/plugins/x11.py#L31 causes we dont collect `/etc/X11/Xmodmap` or `/etc/X11/Xresources` that will be newly collected. I dont see any secret there, rather useful files worth to collect => Ack.

In `etcd`, we basically skip collecting anything `/etc/etcd/ca` while we will skip just `/etc/etcd/ca` - that is intentional, right @TurboTurtle ?

TurboTurtle:
@pmoravec 

For `rear`  - it is intentional to skip collecting the `.iso` files. If there are `iso.log` files I believe that is OK to capture.
For `etcd` - correct, all contents of `/etc/etcd/ca` need to be skipped.

igoihman:
adding @nirs for `vdsm` storage questions.

nirs:
@tinez should be able to help with vdsm plugin issues.

PR #1699, opened 2019-06-11
[networking] collect commands only when required kernel modules are loaded

pmoravec:
---

### DIFF
--- a/sos/__init__.py
+++ b/sos/__init__.py
@@ -50,13 +50,13 @@ _sos = _default
 
 #: Names of all arguments
 _arg_names = [
-    'add_preset', 'alloptions', 'all_logs', 'batch', 'build', 'case_id',
-    'chroot', 'compression_type', 'config_file', 'desc', 'debug', 'del_preset',
-    'dry_run', 'enableplugins', 'encrypt_key', 'encrypt_pass', 'experimental',
-    'label', 'list_plugins', 'list_presets', 'list_profiles', 'log_size',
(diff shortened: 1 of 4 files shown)


bmr-cymru:
This is mostly fine but a couple of changes I think we should make:

 * Lets expose the `SoSPredicate` members so that the plugin code can tell the user exactly what kmods are missing. This is much more helpful.
 * Use a local state variable rather than a long line with embedded `\`.
 * Tweak the message a little so it reads better:

```Python
        allow_changes = self.get_option("allow_system_changes")
        if allow_change or sself.test_predicate(self, pred=ss_pred):
            self.add_cmd_output(ss_cmd)
        else:
            kmods = "kmods=[%s], " % (",".join(ss_pred.kmods))
            self._log_warn("skipped command '%s': required kernel modules not"
                           " loaded (%s). Use '--allow-system-changes' enable" 
                           " collection" % (ss_cmd, kmods))
```
 * List formatting should be consistent with other multi-line list arguments in sos, like `add_cmd_output()` calls etc., i.e.:

```Python
   self.some_method(arg=a_long_list[
        "foo", "bar", "baz", "qux"
    ])
```

bmr-cymru:
I have a patch for this that seems to work well: there are a few other places to check and patch now to make sure the message style is consistent for different cases. I'll get back to this soon.

pmoravec:
Resolved via f84df0f

PR #1702, opened 2019-06-13
[policies] redhat policy to use hostname instead of rhn for local name

pmoravec:
As rhn is getting obsoleted and each host has its hostname, even redhat policy
should use host name as the local name.

Resolves: #1702

---


pmoravec:
Kudos to @TurboTurtle who identified this redundant calls causing uncaught exceptions on some RHEL8 systems with RHN id.

PR #1705, opened 2019-06-19
[archive] Handle checking container sysroot in _make_leading_paths

TurboTurtle:
Previously, in _make_leading_paths(), checking host file paths did not
account for non / sysroots, for situations where sos is run in a
container and the host's / is actually mounted under /host for example.
This would lead to copy errors when trying to copy symlinks.

This method now will use sysroot if one is set, thus avoiding copy
errors.

Credit to Pavel Moravec for identification of where this flaw was
originating, and the correction within _make_leading_paths().

---

### DIFF
--- a/sos/archive.py
+++ b/sos/archive.py
@@ -139,12 +139,13 @@ class FileCacheArchive(Archive):
     _archive_root = ""
     _archive_name = ""
 
-    def __init__(self, name, tmpdir, policy, threads, enc_opts):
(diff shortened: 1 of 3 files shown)


TurboTurtle:
Updated the PR to account for the possibility of `add_node()` getting a doubled `sysroot` prepended (i.e. `/host/host/foo/bar`).

Test suite now passing as well.

PR #1707, opened 2019-06-21
[Plugin] Add get_cmd_content_now method

TurboTurtle:
Adds a new get_cmd_content_now() method that functions the same as
get_cmd_output_now() but will instead return the output of the command
instead of just the filename. The returned data is the same standard
get_command_output() dict with an additional 'filename' element.

The main purpose of this new method is to alleviate the trend in plugins
using get_cmd_output_now() and then manually opening the resulting file
in the archive for further parsing.

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -1008,7 +1008,8 @@ class Plugin(object):
     def _get_cmd_output_now(self, cmd, suggest_filename=None,
                             root_symlink=False, timeout=300, stderr=True,
                             chroot=True, runat=None, env=None,
-                            binary=False, sizelimit=None, subdir=None):
+                            binary=False, sizelimit=None, subdir=None,
+                            content=False):
         """Execute a command and save the output to a file for inclusion in the
         report.
         """
@@ -1037,8 +1038,11 @@ class Plugin(object):
         # save info for later
         self.executed_commands.append({'exe': cmd, 'file': outfn_strip,
                                        'binary': 'yes' if binary else 'no'})
-
-        return os.path.join(self.archive.get_archive_path(), outfn)
+        fname = os.path.join(self.archive.get_archive_path(), outfn)
+        if content:
+            result['filename'] = fname
+            return result
+        return fname
 
     def get_cmd_output_now(self, exe, suggest_filename=None,
                            root_symlink=False, timeout=300, stderr=True,
@@ -1055,7 +1059,25 @@ class Plugin(object):
         return self._get_cmd_output_now(exe, timeout=timeout, stderr=stderr,
                                         chroot=chroot, runat=runat,
                                         env=env, binary=binary,
-                                        sizelimit=sizelimit)
+                                        sizelimit=sizelimit, content=False)
+
+    def get_cmd_content_now(self, exe, timeout=300, root_symlink=False,
+                            suggest_filename=None, stderr=True, chroot=True,
+                            runat=None, env=None, binary=False, sizelimit=None,
+                            pred=None, subdir=None):
+        """Execute a command and save the output to a file then return the
+        output of the command
+        """
+        if not self.test_predicate(cmd=True, pred=pred):
+            self._log_info("skipped cmd output '%s' due to predicate (%s)" %
+                           (exe, self.get_predicate(cmd=True, pred=pred)))
+            return None
+
+        return self._get_cmd_output_now(exe, timeout=timeout, stderr=stderr,
+                                        chroot=chroot, runat=runat, env=env,
+                                        binary=binary, sizelimit=sizelimit,
+                                        suggest_filename=suggest_filename,
+                                        subdir=subdir, content=True)
 
     def is_module_loaded(self, module_name):
         """Return whether specified moudle as module_name is loaded or not"""


pmoravec:
While I like the idea itself, the way how it is implemented is bit confusing: method `_get_cmd_output_now` returns different _types_ of values depending on its argument.

Also, to support adoption of the new method, it would be great to apply it in some plugins that could already leverage by the new method.

TurboTurtle:
Right. That's my hesitation here too - however, in reality if we split it out we'd be copying the exact same code from `get_cmd_output_now()` and the difference between the two would be 1 or 2 lines. So we'd end up having to double any future changes made there.

bmr-cymru:
Rather than changing the return type based on the `content` flag, why not just always return a dict from `_get_cmd_output_now()`, and make `get_cmd_output_now()` just return the `'filename'` key?

I.e.

```diff
diff --git a/sos/plugins/__init__.py b/sos/plugins/__init__.py
index 945dc947..a786e848 100644
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -1112,7 +1112,7 @@ class Plugin(object):
                             root_symlink=False, timeout=300, stderr=True,
                             chroot=True, runat=None, env=None,
                             binary=False, sizelimit=None, subdir=None,
-                            changes=False, content=False):
+                            changes=False):
         """Execute a command and save the output to a file for inclusion in the
         report.
         """
@@ -1145,10 +1145,8 @@ class Plugin(object):
         self.executed_commands.append({'exe': cmd, 'file': outfn_strip,
                                        'binary': 'yes' if binary else 'no'})
         fname = os.path.join(self.archive.get_archive_path(), outfn)
-        if content:
-            result['filename'] = fname
-            return result
-        return fname
+        result['filename'] = fname
+        return result
 
     def get_cmd_output_now(self, exe, suggest_filename=None,
                            root_symlink=False, timeout=300, stderr=True,
@@ -1162,12 +1160,13 @@ class Plugin(object):
                            (exe, self.get_predicate(cmd=True, pred=pred)))
             return None
 
-        return self._get_cmd_output_now(exe, suggest_filename=suggest_filename,
-                                        root_symlink=root_symlink,
-                                        timeout=timeout, stderr=stderr,
-                                        chroot=chroot, runat=runat,
-                                        env=env, binary=binary,
-                                        sizelimit=sizelimit, content=False)
+        res = self._get_cmd_output_now(exe, suggest_filename=suggest_filename,
+                                       root_symlink=root_symlink,
+                                       timeout=timeout, stderr=stderr,
+                                       chroot=chroot, runat=runat,
+                                       env=env, binary=binary,
+                                       sizelimit=sizelimit:
+        return res['filename']
 
     def get_cmd_content_now(self, exe, timeout=300, root_symlink=False,
                             suggest_filename=None, stderr=True, chroot=True,
```

So the two `get_cmd_*_now()` methods are just simple wrappers around `_get_cmd_output_now()` which always returns a dict including the file name?

bmr-cymru:
The more I think about this, the more I'm inclined to say we should re-think this interface completely: `get_cmd_output_now()` was always a bit weird, and never really fitted with the other patterns of API use. The fact it also reads data in from a process, writes it out to disk and then only returns a file path (when the data is still in memory) also leads to clunky plugin code as @TurboTurtle noted.

So rather than preserving that bad design and adding a new interface, why not get rid of `get_cmd_output_now()` and replace it with something better?

The whole `Plugin` API has evolved over many years, and is a bit cluttered - some parts are clear and easy to follow from the method names, others not so much, so this might be another thing to consider adding to the goals for sos-4.0 work (not that it all has to happen under this PR, but it's certainly related).

One example is the various ways to run an external command (with our without capturing it in the report):

 * `get_command_output()` - run a command now and return a dictionary of results. Uses `utilities.sos_get_command_output()` (used **51 times**)
 * `add_cmd_output()` - run the command later and add it to the report (used **503 times**)
 * `call_ext_prog()` - run a command now and return the status dict. Literally `get_command_output()` with less args! (used **43 times**)
 * `check_ext_prog()` - run a command now and check the exit status for success. Literally `get_command_output()` with less args and `return ['status'] == 0` (used **7 times**)
 * `_get_cmd_output_now()` - run the command now and add it to the report. Used by both `Plugin.collect()` and `get_cmd_output_now()`.
 * `get_cmd_output_now()` - run the command now and add it to the report. Predicate testing and logging wrapper around `_get_cmd_output_now()`. (used **16 times**)

Some of the inconsistencies/nits here:

* `get_command_output()` runs the command and returns an entire status dictionary (although it's often used just to obtain output). 
 * Why "cmd", "command", "prog", "exe"? We should settle on a single term or abbreviation and use it everywhere.
 * Why does `sos_get_command_output()` need the `sos_` prefix when it's already defined in `sos.utilities`?
 * The `check_` and `call_` versions are similar to Python `subprocess` module functions, although they don't have exactly the same behaviour.
 * `add_*()` methods mostly have consistent high-level meaning - something is being added to the report collection. We use `get_*()` more loosely right now: some times it's getting internal plugin properties, other times information from the environment, and still others it's actually adding something to the report content.
 * There's no real distinction between plugin code that's using `get_command_output()` directly vs. via `call_ext_prog()`: they're doing exactly the same kind of operation, and they get the same return type back. The only difference is the more restrictive argument set for `call_ext_prog()`.

I'm running through a quick survey of the current public methods in the main `Plugin` class to see if we can make some improvements to the organisation. I think we can make life easier for ourselves and plugin authors by doing some more general clean up here and having tighter rules for method naming and visibility (e.g. if `get_command_output()` or whatever it ends up being called isn't intended for use in derived plugin classes then make it `_get_command_output()`).

bmr-cymru:
Here's a rough survey of all the normal methods defined for `Plugin`: some of this is reasonable, in other cases there are obvious cleanups (e.g. the process methods already mentioned and the fact that there are two almost identical methods for grepping through file data...).

```
Report metadata
    def add_alert(self, alertstring):
    def add_custom_text(self, text):

Async collection/report content
    def add_cmd_output(self, cmds, suggest_filename=None,
    def _add_cmd_output(self, **kwargs):
    def _add_copy_paths(self, copy_paths):
    def _expand_copy_spec(self, copyspec):
    def add_copy_spec(self, copyspecs, sizelimit=None, maxage=None,
    def add_env_var(self, name):
    def add_journal(self, units=None, boot=None, since=None, until=None,
    def add_service_status(self, services, timeout=None, pred=None):
    def add_string_as_file(self, content, filename, pred=None):
    def add_udev_info(self, device, attrs=False):
                defaulttype = type(parms['enabled'])

Path blacklist
    def add_forbidden_path(self, forbidden):
    def _is_forbidden_path(self, path):

Subprocess calls
    def _get_cmd_output_now(self, cmd, suggest_filename=None,
    def get_cmd_output_now(self, exe, suggest_filename=None,
    def get_cmd_output_path(self, name=None, make=True):
    def get_command_output(self, prog, timeout=300, stderr=True,
    def call_ext_prog(self, prog, timeout=300, stderr=True,
    def check_ext_prog(self, prog):

Plugin enablement
    def check_enabled(self):
    def default_enabled(self):
    def _check_plugin_triggers(self, files, packages, commands, services):

Timeout control
    def check_timeout(self):
    def timeout(self):

Plugin execution phases
    def collect(self):
    def _collect_cmd_output(self):
    def _collect_copy_specs(self):
    def _collect_strings(self):
    def postproc(self):
    def setup(self):
    def setup_verify(self):

Plugin IO
    def _copy_dir(self, srcpath):
    def _copy_node(self, path, st):
    def _copy_symlink(self, srcpath):
    def _do_copy_path(self, srcpath, dest=None):
    def _get_dest_for_srcpath(self, srcpath):
    def _make_command_filename(self, exe, subdir=None):
    def _mangle_command(self, exe):

Postproc substitution
    def do_cmd_output_sub(self, cmd, regexp, subst):
    def do_cmd_private_sub(self, cmd):
    def do_file_sub(self, srcpath, regexp, subst):
    def do_path_regex_sub(self, pathexp, regexp, subst):

File grepping
    def do_regex_find_all(self, regex, fname):
    def file_grep(self, regexp, *fnames):

Plugin information
    def get_description(self):
    def name(cls):

Plugin options
    def get_all_options(self):
    def get_option_as_list(self, optionname, delimiter=",", default=None):
    def get_option(self, optionname, default=0):
    def set_option(self, optionname, value):

Predicates
    def get_predicate(self, cmd=False, pred=None):
    def set_cmd_predicate(self, pred):
    def set_predicate(self, pred):
    def test_predicate(self, cmd=False, pred=None):
            def time_filter(path):

Environment helpers
    def get_process_pids(self, process):
    def check_process_by_name(self, process):
    def is_installed(self, package_name):
    def is_module_loaded(self, module_name):
    def get_service_status(self, name):
    def service_is_disabled(self, name):
    def service_is_enabled(self, name):
    def service_is_running(self, name):
    def is_service(self, name):

Sysroot handling
    def join_sysroot(self, path):
    def strip_sysroot(self, path):
    def tmp_in_sysroot(self):
    def use_sysroot(self):

Plugin logging
    def _log_debug(self, msg):
    def _log_error(self, msg):
    def _log_info(self, msg):
    def log_skipped_cmd(self, pred, cmd, kmods=False, services=False,
    def _log_warn(self, msg):
    def _format_msg(self, msg):
```

TurboTurtle:
For consolidating the command collection API, I think we could trim everything down to 3 methods:

- `add_cmd_output()` - run during `plugin.collect()`, add output to report. Keep it how everyone knows and (hopefully) loves.
- `get_cmd_output_now()` -  Wrapper around `_get_cmd_output_now()` and providing Predicate testing and logging.
- `_get_cmd_output_now()` -  run the command now, add it to the report, return status, filename, and content. Allow optionally disabling writing to the report, to provide `get_command_output()` style functionality through `get_cmd_output_now()`.

We then drop `call_ext_prog()`, `check_ext_prog()`, and `get_command_output()`. Taking a look at the current calls to `get_command_output()` a number of them using the `get_cmd_output_now()` rework described above would remove a previous call for running the same command (`pcp`, the container plugins (so at least 6 more), `virsh`, `alternatives`, `ntp`, etc...). The calls to `*_ext_prog()`, could all be replaced by `get_cmd_output_now()` and gain the same benefit as replacing `get_command_output()` - reduced number of times those commands are being run.

TurboTurtle:
Closing this in favor of #1807

PR #1708, opened 2019-06-25
[python] collect pip list

wbclark:
Signed-off-by: William Bradford Clark <wclark@redhat.com>

---


wbclark:
As currently implemented this wouldn't get called on Red Hat systems since RedHatPython.setup() doesn't call super(Python, self).setup(). I will fix this tonight. Thanks @TurboTurtle for the heads up.

Issue #1710, opened 2019-06-26
[archive] symlinks outside archive directory created by _make_leading_paths


pmoravec:
When `archive._make_leading_paths` is called to a path containing a symlink to an absolute path, the method creates a symlink _outside_ the archive directory (to the absolute path of the destination of the symlink). Copying that archive to a different machine then causes broken symlink.

Reproducer:

    mv /var/lib/mistral{,.prev}
    mkdir -p /var/lib/mistral/overcloud
    ln -s /var/lib/mistral/overcloud /var/lib/mistral/config-download-latest
    echo "some ansible log content" > /var/lib/mistral/overcloud/ansible.log

Now, call:

    sosreport -o openstack_instack --batch --build

and see that `/var/tmp/sosreport-myhostname-2019-06-25-wpyemcm/var/lib/mistral/config-download-latest` is a symlink to `/var/lib/mistral/overcloud`, while the archive directory itself does not contain a copy of the logfile.

Root cause of the problem: https://github.com/sosreport/sos/blob/master/sos/archive.py#L221-L237 calls:

    target = os.readlink(src_path)
    ..
    os.symlink(target, abs_path)

This works well for symlinks with relative destination. For absolute path destination, something like:

    if os.path.isabs(target):
        target = target.strip(os.path.commonpath([target, src_path])).strip(os.sep)

must be called in between. This will make the `target` relative from absolute.

BUT this works well only when `target` and `src_path` are in common subdirectory. If e.g. `/var/lib/mistral/config-download-latest` would be a symlink to `/var/log`, we must set `target` to `../../log` and create that relative path via `self._make_leading_paths` first.

The above proposal is sufficient for the real world scenario (Openstack Instack having `/var/lib/mistral/config-download-latest: symbolic link to /var/lib/mistral/overcloud` per my understanding), but it is not ideal.

pmoravec:
.. or we can create symlink to (not-tested):

    target = os.readlink(src_path)
    ..
    target = os.path.relpath(target, start=src_path)
    os.symlink(target, abs_path)

(and somehow take care the target path is created)

PR #1711, opened 2019-06-27
[archive] convert absolute symlink targets to relative

pmoravec:
Calling _make_leading_paths for a symlink with absolute symlink target
must create the symlink relative to the source. This will prevent
creating symlinks outside sosreport build dir.

Resolves: #1710

---

### DIFF
--- a/sos/archive.py
+++ b/sos/archive.py
@@ -232,6 +232,11 @@ class FileCacheArchive(Archive):
                     dest = self._make_leading_paths(target_src, mode=mode)
                     dest = os.path.normpath(dest)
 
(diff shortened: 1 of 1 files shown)


pmoravec:
Since the patch modifies the "well known" symlink / create dir part of `archive` class, I would really appreciate a review from @bmr-cymru and @TurboTurtle .

pmoravec:
Good point (and it makes really more sense to make `target` relative to `target_dir`, directly). PR rebased.

pmoravec:
closed this in ba77701.

Issue #1715, opened 2019-07-01
[ipa] copy httpd cert from new location


tscherf:
With the 4.7.0 release FreeIPA httpd [moved from mod_nss to mod_ssl]( https://www.freeipa.org/page/Releases/4.7.0#mod_ssl).

As a result the httpd X.509 certificate is now no longer stored in NSS DB but as plain PEM text file:  /var/lib/ipa/certs/httpd.crt

The plugin needs to copy this file into the sos archive.

PR #1716, opened 2019-07-01
[ipa] copy httpd cert from new location

tscherf:
With the FreeIPA 4.7.0 release, httpd moved from mod_nss to mod_ssl. As                                                                                                                                             
a result the httpd X.509 certificate is now no longer stored in NSS DB                                                                                                                                              
but as plain PEM text file: /var/lib/ipa/certs/httpd.crt. The plugin                                                                                                                                                
needs to copy this file into the sos archive.

https://github.com/sosreport/sos/issues/1715

Signed-off-by: Thorsten Scherf <tscherf@redhat.com>

---


pmoravec:
+1

PR #1718, opened 2019-07-03
[docker] strip network ID properly from "docker network ls"

pmoravec:
"docker network ls" prints network ID in the first column.

Resolves: #1718

---

### DIFF
--- a/sos/plugins/docker.py
+++ b/sos/plugins/docker.py
@@ -61,7 +61,7 @@ class Docker(Plugin):
         nets = self.get_command_output('docker network ls')
 
         if nets['status'] == 0:
-            n = [n.split()[1] for n in nets['output'].splitlines()[1:]]
+            n = [n.split()[0] for n in nets['output'].splitlines()[1:]]
             for net in n:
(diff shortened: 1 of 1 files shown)


TurboTurtle:
I'm not sure I agree with this change.

Grabbing the second column gives us the network _name_, which is an easier reference point for support engineers and customers than the IDs.

pmoravec:
I can't find the background story for this change now and I dont insist on the change. Must be docker network _name_ non-empty and also unique? (I barely recall the rationale is that list of 2nd words of the "docker network ls" output does not necessarily be a complete&unique list of netowkr names - that's why I wanted to failover to using IDs).

Could there be such a situation? If not, I will drop the PR.

TurboTurtle:
The name parameter is required for `docker network create`, and from a quick test it seems names can't be duplicated:

```
$ sudo docker network create test
50bc1749bf4f365b5181f2cf842ab2c9e8e0af2b934ead7562a8e424321cbca0
$ sudo docker network create test
Error response from daemon: network with name test already exists

```

pmoravec:
OK, dropping the request then. Thanks for feedback.

PR #1722, opened 2019-07-05
[process] Specify 20 for wchan length

tomofumi-yoshida:
wchan column is now truncated because there are many columns.
It is impossible to distinguish the symbol name if there are many
similar symbol names. Therefore, specify wchan size.

In this commit, we choose 20 for the wchan size but there could be
better value. Let's change this if it turns out that in the future.

Example: here is an output of ps command when plymouthd process
got hanged:

    F  UID    PID PPID PRI  NI    VSZ   RSS WCHAN  STAT TTY      TIME COMMAND
    5    0    159    1  20   0  44644  1696 n_tty_ S    ?        0:04 /bin/plymouthd --attach-to-session

The prefix n_tty_ is used in tty device driver of linux kernel
as follows:

    struct tty_ldisc_ops tty_ldisc_N_TTY = {
	    .magic           = TTY_LDISC_MAGIC,
	    .name            = "n_tty",
	    .open            = n_tty_open,
	    .close           = n_tty_close,
	    .flush_buffer    = n_tty_flush_buffer,
	    .chars_in_buffer = n_tty_chars_in_buffer,
	    .read            = n_tty_read,
	    .write           = n_tty_write,
	    .ioctl           = n_tty_ioctl,
	    .set_termios     = n_tty_set_termios,
	    .poll            = n_tty_poll,
	    .receive_buf     = n_tty_receive_buf,
	    .write_wakeup    = n_tty_write_wakeup,
	    .fasync          = n_tty_fasync,
    };

As you can see, it is impossible to find the symbol name of the
kernel function where plymouthd process slept from n_tty_.

Signed-off-by: Tomofumi Yoshida <yoshida.tomo@fujitsu.com>

---

### DIFF
--- a/sos/plugins/process.py
+++ b/sos/plugins/process.py
@@ -27,7 +27,7 @@ class Process(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
         # process group and thread options
         ps_group_opts = "pid,ppid,user,group,lwp,nlwp,start_time,comm,cgroup"
         ps_sched_opts = "flags,state,uid,pid,ppid,pgid,sid,cls,pri,addr,sz,"
-        ps_sched_opts += "wchan,lstart,tty,time,cmd"
+        ps_sched_opts += "wchan:20,lstart,tty,time,cmd"
 
         self.add_copy_spec([
             "/proc/sched_debug",


pmoravec:
What are all possible options on any kernel? As I see on RHEL7:

    # ps axo flags,state,uid,pid,ppid,pgid,sid,cls,pri,addr,sz,wchan:20,lstart,tty,time,cmd | awk '{ print $12 }' | sort -u
    -
    add_hwgenerator_rand
    devtmpfsd
    do_wait
    ep_poll
    futex_wait_queue_me
    hrtimer_nanosleep
    kauditd_thread
    khugepaged
    ksm_scan_thread
    kswapd
    kthreadd
    n_tty_read
    pipe_wait
    poll_schedule_timeou
    rcu_gp_kthread
    rescuer_thread
    scsi_error_handler
    sk_wait_data
    smpboot_thread_fn
    SYSC_semtimedop
    unix_stream_read_gen
    watchdog
    WCHAN
    worker_thread
    xfsaild
    #

See some texts still truncated (but I guess unique to determine them, so the 20chars can be sufficient)

tomofumi-yoshida:
Thank you for your comment.

The example I showed is the result on RHEL6.

The optimal wchan size may not be 20, but we thought it would be possible to find the symbol name if the wchan size is 20.

tomofumi-yoshida:
I apologize for my late comment.
I reconsidered the solution for this issue and devised following three ideas.

1) Specify the best size of wchan.
2) Use nwchan instead of wchan.
3) Divide ps command and collect wchan with the maximum kernel symbol size by new one.

Because of the following reasons, I believe that the idea 3) is the best solution.

The weak points of the idea 1):

  The simplest solution is to specify the maximum kernel symbol size for wchan.
  But they are 60 characters or so.
  The output of ps command become too long and it will be unreadable.

  Moreover, some of them are different for the trailing characters insensibly as follows.
  So if I consider the optimal size of wchan, it seems to become long.

    __pci_fixup_resume_earlyasus_hides_smbus_lpc_ich6_resume_early1582
    __pci_fixup_resume_earlyasus_hides_smbus_lpc_ich6_resume_early1583
    __pci_fixup_resume_earlyasus_hides_smbus_lpc_ich6_resume_early1584

The weak points of the idea 2):

  Use nwchan instead of wchan to output the address of the kernel function where the process is sleeping.
  sosreport collects /proc/kallsyms so I can compare the symbol and the address.
  But nwchan seems to be broken because it only output the column size for 5 or 6 characters.
  It needs 16 characters on 64bit os.


Could you tell me your opinion for my ideas?
Moreover, please tell me if you have another idea.

tomofumi-yoshida:
I'm waiting for any comment.

bmr-cymru:
I think @tomofumi-yoshida is correct here - this is a bit of an unfortunate limitation in `ps`. My first thought was also to collect the raw symbol addresses and manually look them up in `kallsyms` but I found the same problem - `ps -onwchan:<WIDTH>` and `ps -onwchan=WIDE-COLUMN-HEADING` both appear to be broken: all symbols appear as `1` or `ffffff`:

```
# ps -axo pid,nwchan=WIDE-WIDE-WIDE-NWCHAN,wchan:40,cmd | egrep ' 1 |ffff' | head
    1                     1 do_epoll_wait                            /usr/lib/systemd/systemd --switched-root --system --deserialize 18
   10                     1 -                                        [rcu_sched]
   25                     1 -                                        [khungtaskd]
   30                     1 -                                        [khugepaged]
  387                     1 -                                        [scsi_eh_0]
  390                     1 -                                        [scsi_eh_1]
  531                     1 -                                        [xfsaild/dm-3]
  629                     1 do_epoll_wait                            /usr/lib/systemd/systemd-journald
  659                     1 do_epoll_wait                            /usr/lib/systemd/systemd-udevd
  697                ffffff core_sys_select                          /usr/sbin/dmeventd -f
```

This isn't just truncation - those addresses just seem completely wrong. The `core_sys_select` symbol has this value as prefix:

```
# grep core_sys_select /proc/kallsyms 
ffffffffacacc660 t compat_core_sys_select
ffffffffacacd9a0 T core_sys_select
```

But `do_epoll_wait` doesn't even feature the digit `1`!

```
# grep do_epoll_wait /proc/kallsyms
ffffffffacb03e40 t do_epoll_wait
```

This seems odd - there are quite  few posts online showing the same behaviour, but I've not been able to find anyone complaining about or explaining this behaviour. It may be worth reporting to the procps-ng project but since we can't control the version of `ps` people are using I think we have no good option but to go for the hard coded width for `wchan` for now.

bmr-cymru:
> What are all possible options on any kernel? As I see on RHEL7:

The problem is it's not fixed: these are just the names of kernel functions. The set depends on the kernel version and loaded modules (including 3rd party modules), so it's not really possible to predict. There are some drivers (as @tomofumi-yoshida points out) that have very long function names which vary only in a short suffix indicating a hardware variation.

tomofumi-yoshida:
I pushed a new patch: 17b7c3ad1f3c4b8dd8753a9eadc21ca8b5321f86
Could you review one?
However, the patch seems somehow not to be reflected on this pull request.

PR #1727, opened 2019-07-11
[kdump] collect vmcore-dmesg.txt from a customized path as well

pmoravec:
/var/crash is configurable via "path" option in /etc/kdump.conf .

sosreport should collect vmcore-dmesg.txt from the customized locations
as well.

Resolves: #1727

---

### DIFF
--- a/sos/plugins/kdump.py
+++ b/sos/plugins/kdump.py
@@ -27,11 +27,32 @@ class RedHatKDump(KDump, RedHatPlugin):
     files = ('/etc/kdump.conf',)
     packages = ('kexec-tools',)
 
+    def find_crash_dir(self, conf_file):
+        """ Get crash directory as specified in /etc/kdump.conf
+            by line:
+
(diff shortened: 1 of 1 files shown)


mikelolasagasti:
Based on @k-hagio comments in https://github.com/sosreport/sos/pull/1546/#issuecomment-458720073 this PR would be incomplete, as path is "relative to the specified dump target". 

I didn't had time to fix my PR, but if you can add FS logic in yours, I can close my PR.

pmoravec:
Thanks for pointing to that PR, wasnt aware of it.

I expect you @mikelolasagasti know the specification (how to implement relative path case) better than me so I would prefer if you can update yours PR, but I can come up with my own implementation as well - let me know your preferences.

pmoravec:
Reviewing the #1546, that superseds mine PR and is imho worth merging. So closing mine in favour of #1546.

PR #1730, opened 2019-07-17
[openstack_instack] Add container-image-prepare log

cjeanner:
During undercloud and overcloud deploy preparation steps, a task is
fetching container images and pushing them to a local registry
(usually on the undercloud directly).

Since this task might fail at some point, we need this log in order to
properly debug the issue and provide better support.

The task itself doesn't produce any other output than this log, and only
exists on the undercloud.

---


cjeanner:
@TurboTurtle Hello! this replaces PR 1729 in a more convenient way. Hopefully you'll agree with it :)

EmilienM:
:+1:

Issue #1736, opened 2019-07-25
Sosreport too weakly stops threads of timeouted plugins


pmoravec:
Reproducer: let sosreport collect too many files for some plugin (example: create rich dir.structure under `/var/lib/sesame` and run `sosreport -o qpid --plugin-timeout=10`). Then the following can happen:

- plugin times out
- the thread for the plugin still collects data and copies files under temp/working dir
- sosreport tries to create tarball that fails in `archive`'s `cleanup`
- as a result, whole temp. dir is removed _without_ keeping the tarball created - so the sosreport produces _nothing_

When using `--build` option, the above does not happen (while the plugin thread is stopped by hitting exception like https://github.com/sosreport/sos/issues/1641 - as it cant write files to just deleted temp.dir).

Few ideas how to fix this:
- few ways of catching the `archive.cleanup` exception exist - then we can get the build dir or even the tarball, BUT whole sosreport app can raise uncaught exception like below due to the timeouted plugin _still_ generating new content:

    Traceback (most recent call last):
      File "/usr/sbin/sosreport", line 19, in <module>
        main(sys.argv[1:])
      File "/usr/lib/python2.7/site-packages/sos/sosreport.py", line 1384, in main
        sos.execute()
      File "/usr/lib/python2.7/site-packages/sos/sosreport.py", line 1369, in execute
        self._cleanup()
      File "/usr/lib/python2.7/site-packages/sos/sosreport.py", line 1322, in _cleanup
        self.archive.cleanup()
      File "/usr/lib/python2.7/site-packages/sos/archive.py", line 525, in cleanup
        shutil.rmtree(self._archive_root)
      File "/usr/lib64/python2.7/shutil.py", line 247, in rmtree
        rmtree(fullname, ignore_errors, onerror)
      File "/usr/lib64/python2.7/shutil.py", line 256, in rmtree
        onerror(os.rmdir, path, sys.exc_info())
      File "/usr/lib64/python2.7/shutil.py", line 254, in rmtree
        os.rmdir(path)
    OSError: [Errno 39] Directory not empty: '/var/tmp/sos.iyv0tN/sosreport-pmoravec-rhel76-2019-07-25-xhgfjej/etc'

- so I think the right fix should be rather forceful termination of plugins' threads before calling `archive.finalize()`

pmoravec:
Idea of a fix: when main thread detects timeout, it orders the timeouted plugin to stop collecting further data, like:

[...]

PR #1739, opened 2019-07-26
[plugins] Stop plugin execution after timeout hit

pmoravec:
When a plugin timeouts, it must stop collecting further data. Otherwise
a race issues with archive.finalize() can happen.

So any data collection must be skipped if _timeout_hit is True.

Resolves: #1736

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -652,6 +652,9 @@ class Plugin(object):
         everything below it is recursively copied. A list of copied files are
         saved for use later in preparing a report.
         '''
+        if self._timeout_hit:
+            return
+
         if self._is_forbidden_path(srcpath):
             self._log_debug("skipping forbidden path '%s'" % srcpath)
             return ''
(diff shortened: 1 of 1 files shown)


sos/sosreport.py
```diff
@@ -1005,6 +1005,7 @@ def _collect_plugin(self, plugin):
                 self.ui_log.error("\n Plugin %s timed out\n" % plugin[1])
                 self.running_plugs.remove(plugin[1])
                 self.loaded_plugins[plugin[0]-1][1]._timeout_hit = True
+                self.loaded_plugins[plugin[0]-1][1].stop()
```
TurboTurtle: Should we move
```python
self.loaded_plugins[plugin[0]-1][1]._timeout_hit = True
```

into the new `stop()` method? Or in some other way merge _stopped with _timeout_hit? Perhaps `check_timeout()` should now just `return self._stopped` instead of a separate `_timeout_hit` var?

pmoravec:
Good point in reusing the existing variable.

pmoravec:
Closed this in 8ad1f59 .

PR #1745, opened 2019-08-07
[plugins] Fix ignored parameters in get_cmd_output_now()

k-hagio:
Currently the `get_cmd_output_now()` function does not pass
`suggest_filename` and `root_symlink` parameters to
`_get_cmd_output_now()` function, so the parameters passed
from plugins are ignored.  Let's fix this.

Signed-off-by: Kazuhito Hagio &lt;k-hagio@ab.jp.nec.com&gt;

---


k-hagio:
Thanks @pmoravec for the review.
What time frame can this get merged?

PR #1748, opened 2019-08-10
[sosreport] Add gluster deployment and cleanup log

gobindadas:
---

### DIFF
--- a/sos/plugins/ovirt_hosted_engine.py
+++ b/sos/plugins/ovirt_hosted_engine.py
@@ -48,6 +48,12 @@ class OvirtHostedEngine(Plugin, RedHatPlugin):
             '/var/log/ovirt-hosted-engine-ha/agent.log',
             '/var/log/ovirt-hosted-engine-ha/broker.log',
         ])
+
+        # Add gluster deployment and cleanup log
+        self.add_copy_spec([
+            '/var/log/cockpit/ovirt-dashboard'
+        ])
+
         # Add older ovirt-hosted-engine-ha log files only if requested
         if self.get_option('all_logs'):
             self.add_copy_spec(self.HA_LOG_GLOB)


gobindadas:
@sandrobonazzola Can you please review and merge?

sos/plugins/ovirt_hosted_engine.py
```diff
@@ -48,6 +48,14 @@ def setup(self):
             '/var/log/ovirt-hosted-engine-ha/agent.log',
             '/var/log/ovirt-hosted-engine-ha/broker.log',
         ])
+
+        # Add gluster deployment and cleanup log
+        self.add_copy_spec([
+            '/var/log/cockpit/ovirt-dashboard',
```
pmoravec: Collecting whole dir implies collecting the two particular logfiles from the dir - didnt you want to collect some different directory here?

gobindadas: No, The logfiles which are mentioned under '/var/log/cockpit/ovirt-dashboard' need to be included in sosreport.

TurboTurtle: Right, so you only need the first line collecting the dir. Everything in that directory will be collected, including those files.

gobindadas: @pmoravec Ok got it. I have pushed the changes.

sandrobonazzola:
> @sandrobonazzola Can you please review and merge?

I have no merge rights here. I can start telling you that you should edit the PR to confirm you adhere to contribution guidelines.

sandrobonazzola:
@bmr-cymru can you please review?

pmoravec:
Code is OK, just commit subject + message needs improvement (see Contrib. guidelines), like:

    [ovirt_hosted_engine] Add gluster deployment and cleanup log

    Resolves: #1748 

(since the future commit will resolve this upstream issue/PR, not some bugzilla.redhat.com BZ (yet) )

PR #1752, opened 2019-08-13
[foreman] Add puppetserver's conf.d dir

mikelolasagasti:

---


pmoravec:
Why `foreman` plugin? Isn't https://github.com/sosreport/sos/blob/master/sos/plugins/puppet.py#L30 sufficient? (and if not, still I think `puppet` plugin is better place - just modify this line)

Or are there cases when `puppet` plugin` isn't enabled by default and it should (in some use cases when `foreman` is installed)?

mikelolasagasti:
My fault @pmoravec as some files are in etc/ and others in sos_commands/foreman/foreman-debug/etc/ I was not looking in the correct place.

Issue #1756, opened 2019-08-17
[docker] plugin exception when service is not running


TurboTurtle:
Against the current master, the `docker` plugin will generate an exception if the service is not running:

```
Traceback (most recent call last):
  File "/home/jake/projects/sos/sos/sosreport.py", line 947, in setup
    plug.setup()
  File "/home/jake/projects/sos/sos/plugins/docker.py", line 128, in setup
    super(RedHatDocker, self).setup()
  File "/home/jake/projects/sos/sos/plugins/docker.py", line 54, in setup
    self.add_cmd_output("docker %s" % subcmd)
  File "/home/jake/projects/sos/sos/plugins/__init__.py", line 993, in add_cmd_output
    changes=changes)
  File "/home/jake/projects/sos/sos/plugins/__init__.py", line 971, in _add_cmd_output
    self.log_skipped_cmd(pred, soscmd.cmd, kmods=bool(pred.kmods),
AttributeError: 'NoneType' object has no attribute 'kmods'
```

At a quick glance it looks like the predicate from `set_cmd_predicate()` is not being applied to `add_cmd_output()` calls that don't have an explicit predicate passed to them.

PR #1757, opened 2019-08-18
[Plugin] Apply cmd predicates to add_cmd_output

TurboTurtle:
If no predicate was passed to add_cmd_output() calls, then a predicate
set by set_cmd_predicate() was not being applied to those calls, thus
leading to a 'None' predicate being evaluated and potentially causing an
exception in plugins such as docker where the bulk of command collection
is gated by the service running.

Now if no predicate is handed to add_cmd_output(), and a cmd_predicate
has been set, then the cmd_predicate is applied before being handed off
to the _add_cmd_output() helper.

Fixes: #1756

---


pmoravec:
Closed this in aee752e .

PR #1759, opened 2019-08-19
[openstack_manila] Collect db version and protect passwords

gouthampacha:
- It'd be helpful to collect the database version
   used by the deployment for debug.
- The password protection regex did not include all the
   plaintext password configuration options.
- Add support to collect package versions for the verify mode
   of report collection.

Signed-off-by: Goutham Pacha Ravi <gouthampravi@gmail.com>

---

### DIFF
--- a/sos/plugins/openstack_manila.py
+++ b/sos/plugins/openstack_manila.py
@@ -21,26 +21,12 @@ class OpenStackManila(Plugin):
     var_puppet_gen = "/var/lib/config-data/puppet-generated/manila"
 
     def setup(self):
-        manila_config_opt = "--config-dir %s/etc/manila/"
 
-        # check eventlet-based service on the baremetal
-        in_ps = self.check_process_by_name("manila-api")
-
-        # check if manila-api is running inside a container
-        # (doesn't matter if running via httpd or as a eventlet based service)
-        in_container = self.running_in_container()
-
-        if in_container:
-            manila_config = manila_config_opt % self.var_puppet_gen
-        else:
-            manila_config = manila_config_opt % ''
-
-        # gather DB version
-        if in_ps or in_container:
-            self.add_cmd_output(
-                "manila-manage " + manila_config + " db version",
-                suggest_filename="manila_db_version"
-            )
+        config_dir = "%s/etc/manila" % (
+            self.var_puppet_gen if self.running_in_container() else ''
+        )
+        manila_cmd = "manila-manage --config-dir %s db version" % config_dir
+        self.add_cmd_output(manila_cmd, suggest_filename="manila_db_version")
 
         self.add_copy_spec([
             "/etc/manila/",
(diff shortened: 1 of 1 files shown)


gouthampacha:
Hi @pmoravec @TurboTurtle - can you please review? :)

sos/plugins/openstack_manila.py
```diff
@@ -43,6 +64,18 @@ def setup(self):
                 "/var/log/containers/httpd/manila-api/*log"
             ])
 
+        if self.get_option("verify"):
+            self.add_cmd_output("rpm -V %s" % ' '.join(self.packages))
```
TurboTurtle: This is already handled in the base `Plugin` class, so it isn't needed. I know most of the other OpenStack plugins explicitly have it in there, but we have #1696 open to address that so we shouldn't add it here.

sos/plugins/openstack_manila.py
```diff
@@ -51,12 +84,8 @@ def apply_regex_sub(self, regexp, subst):
...
 
     def postproc(self):
-        protect_keys = [
-            "nova_admin_password",  "rabbit_password",  "qpid_password",
-            "password", "netapp_nas_password", "cinder_admin_password",
-            "neutron_admin_password", "service_instance_password",
-            "transport_url"
-        ]
+        protect_keys = [".*password.*", "transport_url",
+                        "hdfs_ssh_pw", "maprfs_ssh_pw"]
```
TurboTurtle: On the surface this looks good to me, but as a sanity check measure can you provide an example of this properly scrubbing keys here before we merge?

sos/plugins/openstack_manila.py
```diff
@@ -21,6 +21,27 @@ class OpenStackManila(Plugin):
...
+            manila_config = manila_config_opt % self.var_puppet_gen
+        else:
+            manila_config = manila_config_opt % ''
+
+        # gather DB version
+        if in_ps or in_container:
+            self.add_cmd_output(
+                "manila-manage " + manila_config + " db version",
+                suggest_filename="manila_db_version"
+            )
```
TurboTurtle: I think this could be condensed a bit, like:

```python
manila_dir = "%s/etc/manila" % (
    self.var_puppet_gen if self.running_in_container() else ''
)
manila_cmd = "manila-manage --config-dir %s db version" % (manila_dir)
self.add_cmd_output(manila_cmd, suggest_filename="manila_db_version")
```

`add_cmd_output` already handles command failures and non-existing commands, so we shouldn't need to have the conditional in place on running the command, just formatting it. Or the `in_ps` check since we can be reasonably sure that `manila_api` is running on the host if this plugin is enabled already.

sos/plugins/openstack_manila.py
gouthampacha: Ack, thanks! Noticed https://github.com/sosreport/sos/blob/f2e7ac47eb96f59362c0e2c0271642377562d746/sos/plugins/__init__.py#L1373-L1384 - I'll remove this.

sos/plugins/openstack_manila.py
gouthampacha: yep, thanks for the clarification. agree this is cleaner.

sos/plugins/openstack_manila.py
gouthampacha: Yep, here it is: http://paste.openstack.org/show/763919/

gouthampacha:
Thanks for your review @TurboTurtle - addressed your comments with 93a7b23

gouthampacha:
bump :)

gouthampacha:
@TurboTurtle - can you please take a look?

PR #1761, opened 2019-08-21
[vdsm]: Fix executing shell commands

mwperina:
During review process of [1] it was suggested to replace '/bin/sh' with
just 'sh', but unfortunately this change caused an error to not include
information about VDSM mount points as described in [2]. This fix
restores previous state and add full path to shell executions.

[1] https://github.com/sosreport/sos/pull/1205
[2] https://bugzilla.redhat.com/1743304

Closes: https://bugzilla.redhat.com/1743304
Signed-off-by: Martin Perina <mperina@redhat.com>

---

### DIFF
--- a/sos/plugins/vdsm.py
+++ b/sos/plugins/vdsm.py
@@ -92,8 +92,8 @@ class Vdsm(Plugin, RedHatPlugin):
             ])
         self.add_cmd_output([
             "ls -ldZ /etc/vdsm",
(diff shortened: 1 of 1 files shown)


mwperina:
@tinez

mwperina:
Bryn, could you please review and merge? We need to get this fixed as soon as possible

@bmr-cymru

PR #1762, opened 2019-08-21
[boot] add information about secure boot state (uefi)

rmetrich:


TurboTurtle:
#1631 was opened for this previously

rmetrich:
@TurboTurtle Thanks for pointing to this, I'm hence closing my PR.

Issue #1755, opened 2019-08-16
[SoSPredicate] allow service predicates to override kmods?


bmr-cymru:
Currently if any of an `SoSPredicate` objects requirements are met, the predicate evaluates true. This means that for e.g. if a predicate specifies a list of kmods, and a service, it will execute if any one of the kernel modules are present.

For some cases, e.g. where a service is missing causing a long command timeout this behaviour may be undesirable. Right now there are no uses of predicates that would suffer this problem (they all check for either kernel modules ORservices), but this could be an issue with future predicate use.

There are a couple of options:

* Change predicate evaluation so that *all* requirements must be met. This avoids the above problem but is inflexible: in some cases "any" evaluation may be useful for predicate users.

* Add additional predicate control flags, e.g.:
  - all_kmods -> all kernel modules must be present
  - all_services -> all services must be present
  - all -> both kmods and services must be present

TurboTurtle:
What if we did something like this:

```python
# require all given objects to evaluate True
my_pred = SoSPredicate(kmods=['mod1', 'mod2'], services=['svc1', 'svc2'], required='all')

# Only one needs to evaluate True (default)
my_pred = SoSPredicate(kmods=['mod1', 'mod2'], services=['svc1', 'svc2'], required='any')

# Any number of kmods, all services
my_pred = SosPredicate(kmods=['mod1', 'mod2'], services=['svc1', 'svc2'],
                       required_kmods='any', required_services='all')
```

By having the various `required` args default to `any`, we maintain the default/current behavior and provide a good deal of flexibility. Another option could be something like:

```python
my_pred = SoSPredicate(kmods=['mod1', 'mod2'], services=['svc1', 'svc2'],
                       required={'kmods': 'any', 'services': 'all'})
```

which at first glance would be a bit easier to extend in the future if we need to add more things for predicates to evaluate.

bmr-cymru:
Yeah that looks like a good approach.

PR #1764, opened 2019-08-21
[Plugin] Allow predicates to match any or all provided elements

TurboTurtle:
Expands predicate evaluation to be able to evaluate True or False
depending on if any of the passed items are present (previous behavior)
or if _all_ the passed items need to be present. This is handled by
optionally passing a dict via the 'required' argument to SoSPredicate.
The keys in this dict are items supported by SoSPredicate (currently
'kmods' and 'services'), with values being either 'any' or 'all'.

Both kmods and services can be passed to the same predicate to allow
fine-grained control over if a command should be collected.

The default behavior remains the same, meaning that both 'kmods' and
'services' are set to 'any' for the required dict. For example,
creating a predicate like the following:

    SoSPredicate(self, kmods=['foo', 'bar'], services=['oof', 'rab'])

implies "required={'kmods': 'any', 'services': 'any'}" and thus as long
as either kmods _and_ either of the services exist, the predicate
evaluates True. If, for example, neither of the kmods are present, the
predicate will evaluate False.

Providing a predicate like the following however, will require _both_
kmods to be present and either of the services to evaluate True. If
either kmods are not present, and --allow-system-changes is not set by
the end user, this predicate will evaluate False:

    SoSPredicate(self, kmods=['foo', 'bar'], services=['oof', 'rab'],
                 required={'kmods': 'all'})

Further, --allow-system-changes is handled directly in predicate
evaluation for kmods, meaning individual plugins (like networking) do
not need to separately check for if this flag is set or not. However,
they will still need to specify 'changes=True' to the add_cmd_ouput()
call if the intent is for the warning message that gets logged to
include a mention about using --allow-system-changes.

Fixes: #1755

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -146,34 +146,53 @@ class SoSPredicate(object):
         """
         return self.__str(quote=True, prefix="SoSPredicate(", suffix=")")
 
+    def _eval_kmods(self):
+        if not self.kmods or self._owner.get_option('allow_system_changes'):
+            return True
+
+        # Are kernel modules loaded?
+        _kmods = [self._owner.is_module_loaded(k) for k in self.kmods]
+
+        if self.required['kmods'] == 'any':
+            return any(_kmods)
+        else:
+            return all(_kmods)
+
+    def _eval_services(self):
+        if not self.services:
+            return True
+
+        _svcs = [self._owner.service_is_running(s) for s in self.services]
+
+        if self.required['services'] == 'any':
+            return any(_svcs)
+        else:
+            return all(_svcs)
+
     def __nonzero__(self):
         """Predicate evaluation hook.
         """
-        pvalue = False
-
-        # Allow loading kernel modules?
-        pvalue |= self._owner.get_option("allow_system_changes")
-
-        # Are kernel modules loaded?
-        for k in self.kmods:
-            pvalue |= self._owner.is_module_loaded(k)
-
-        for s in self.services:
-            pvalue |= self._owner.service_is_running(s)
 
         # Null predicate?
         if not any([self.kmods, self.services, self.dry_run]):
             return True
 
-        return pvalue and not self.dry_run
+        return ((self._eval_kmods() and self._eval_services()) and not
+                self.dry_run)
 
-    def __init__(self, owner, dry_run=False, kmods=[], services=[]):
+    def __init__(self, owner, dry_run=False, kmods=[], services=[],
+                 required={}):
         """Initialise a new SoSPredicate object.
         """
         self._owner = owner
         self.kmods = list(kmods)
         self.services = list(services)
         self.dry_run = dry_run | self._owner.commons['cmdlineopts'].dry_run
+        self.required = {'kmods': 'any', 'services': 'any'}
+        self.required.update({
+            k: v for k, v in required.items() if
+            required[k] != self.required[k]
+        })
 
 
 class SoSCommand(object):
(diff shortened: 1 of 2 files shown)


TurboTurtle:
This will need some thorough testing. I've done some tests with the docker and networking plugins on my end, but before this gets merged we'll want to be absolutely sure that the default behavior hasn't changed from current and that any mix-and-match use cases are properly covered.

sos/plugins/networking.py
```diff
@@ -222,18 +222,10 @@ def setup(self):
                 ])
 
                 ss_cmd = ns_cmd_prefix + "ss -peaonmi"
-                ss_pred = SoSPredicate(self, kmods=['tcp_diag', 'udp_diag',
-                                                    'inet_diag', 'unix_diag',
-                                                    'netlink_diag',
-                                                    'af_packet_diag'])
```
pmoravec: Good point to delete the duplicate code (cf. L157) - originally I thought it is an omission.

sos/plugins/__init__.py
```diff
@@ -146,34 +146,54 @@ def __repr__(self):
...
 
-    def __nonzero__(self):
-        """Predicate evaluation hook.
-        """
-        pvalue = False
-
-        # Allow loading kernel modules?
-        pvalue |= self._owner.get_option("allow_system_changes")
+    def _eval_kmods(self):
+        if not self.kmods or self._owner.get_option('allow_system_changes'):
```
pmoravec: Do we really want to hook `--allow-system-changes` option to `kmod` predicates? Can't be there a use case where:

- `--allow-system-changes` is used
- some command can be executed *only* if some kernel module is loaded - otherwise, the command can hung or misbehave either way (I just speculate here, not being aware of such command)

If we would have such a command called by sosreport, then running `sosreport --allow-system-changes` can be dangerous. Despite we would guard that command via `kmod` "properly".

I rather think that `--allow-system-changes` shall be independent on predicate evaluation (despite plugins will need a little more code). See the option name is general and expects it to be used to potentially alter the system in either way - just the current usage is that it can load some kernel modules.

sos/plugins/__init__.py
```diff
@@ -146,34 +146,54 @@ def __repr__(self):
...
+        else:
+            return all(_kmods)
 
-        for s in self.services:
-            pvalue |= self._owner.service_is_running(s)
+    def _eval_services(self):
+        if not self.services:
+            return True
+
+        if self.required['services'] == 'any':
```
pmoravec: Can't be the logic from `_eval_kmods` used here as well? I.e.:

        _services = [self._owner.service_is_running(s) for s in self.services]

        if self.required['services'] == 'any':
            return any(_services)
        else:
            return all(_services)

That removes some code duplicity and imho is more readable code ("any"/"all").

pmoravec:
I have vastly tested this on more systems and havent identified any issue - data collected exactly when expected.

sos/plugins/__init__.py
TurboTurtle: My thought here is that if a user uses `--allow-system-changes`, we (and they) don't actually care about gating commands, so it should be an auto-pass.

In the event that a command that requires a kernel module to be loaded runs when it isn't loaded _and_ it doesn't load the kernel module itself (not aware of any), my expectation would be that the command fails rather than hangs.

It would also make plugin code a bit more clunky by having the evaluation ignore the option - we'd have to manually test the predicate and then run the command without a predicate, rather than evaluation the predicate with the command request at the same time. I believe this is why even the original evaluation checked against the use of `allow-system-changes`.

sos/plugins/__init__.py
TurboTurtle: Yup, it totally should be.

TurboTurtle: Updated the PR for this.

pmoravec:
Closed this in 51fbec1 .

PR #1765, opened 2019-08-21
[openstack_instack] Collect /home/stack/.tripleo/history

assafmuller:
The 'history' file includes one line for every cmdline
invocation of TripleO. The file includes useful information
to troubleshoot TripleO based OpenStack deployments, as it
includes timestamps for operations, so the troubleshooter
will now be able to know when the last operation happened.

Signed-off-by: Assaf Muller <amuller@redhat.com>

---

### DIFF
--- a/sos/plugins/openstack_instack.py
+++ b/sos/plugins/openstack_instack.py
@@ -23,7 +23,8 @@ CONTAINERIZED_DEPLOY = [
         '/var/log/heat-launcher/',
         '/home/stack/install-undercloud.log',
         '/home/stack/undercloud-install-*.tar.bzip2',
-        '/var/lib/mistral/config-download-latest/ansible.log'
+        '/var/lib/mistral/config-download-latest/ansible.log',
+        '/home/stack/.tripleo/history',
 ]
 
 
@@ -127,6 +128,9 @@ class OpenStackInstack(Plugin):
         json_regexp = r'((?m)"(%s)": )(".*?")' % "|".join(protected_json_keys)
         self.do_file_sub("/home/stack/instackenv.json", json_regexp,
                          r"\1*********")
+        self.do_file_sub('/home/stack/.tripleo/history',
+                         r'(password=)\w+',
+                         r'\1*********')
 
 
 class RedHatRDOManager(OpenStackInstack, RedHatPlugin):


pmoravec:
Can't it contain some sensitive information that are worth to obfuscate? I.e. some password or similar?

assafmuller:
It would not collect passwords as it collects CLI invocations. It would collect IP addresses but a lot of commands do that.

pmoravec:
E.g.:

    2019-08-20 17:42:03.879724 tripleo-container-image-push registry_url=192.168.24.1:8787, pull_source=None, append_tag=, username=None, password=None, dry_run=False, multi_arch=False, image_to_push=..

There is a password there :) (well, "None" here but isnt this something we should scrub?)

assafmuller:
I genuinely hope that users don't type passwords in cleartext in their CLI :( That being said on a more practical note, the TripleO CLI logs any command and subcommand ran, there's no guarantee that some commands won't have a different parameter name for a password (maybe 'pass' instead of 'passport', or 'key', or whatever...), nor future commands not yet written... So I'm not sure how to scrub it entirely, though certainly we can scrub anything that comes after 'password=*'.

What do you think?

pmoravec:
The battle with scrubbing sensitive info is never-ending, iterative process (that sometimes finds skeletons in a closet hidden for years :( ). This should not prevent collecting the history file. We just should scrub any potential secure info we are aware at the moment.

Could you please extend the `postproc` method by a call like:

    self.do_file_sub("/home/stack/.tripleo/history", r"(((.*)(password|key))=)(.*)", r"\1*********")

Not sure what all keywords shall be there (`password` and `key` shown as an example) - it could be just the `password` if it is the only one you are aware of being used now.

(and please test it on some data that it works, since I wrote it from scratch)

assafmuller:
if password is an argument in the middle of the string:
re.compile(r"(password=)\w+").sub(r"\1*********", "2019-08-20 17:42:58.386463 tripleo-container-image-push registry_url=192.168.24.1:8787, pull_source=None, append_tag=, username=None, password=None, dry_run=False, multi_arch=False, image_to_push=docker://brew-pulp-docker01.web.prod.ext.phx2.redhat.com:8888/rhosp15/openstack-mistral-engine:latest")

there is no password argument:
re.compile(r"(password=)\w+").sub(r"\1*********", "2019-08-20 17:42:58.386463 tripleo-container-image-push registry_url=192.168.24.1:8787, pull_source=None, append_tag=, username=None, hi=bye")

password is the last argument:
re.compile(r"(password=)\w+").sub(r"\1*********", "2019-08-20 17:42:58.386463 tripleo-container-image-push registry_url=192.168.24.1:8787, pull_source=None, append_tag=, username=None, password=2232")

In all 3 scenarios, I got the expected result.
I also tried the patch locally by creating /home/stack/.tripleo/history on my laptop and populating it with:

2019-08-20 17:42:58.386463 tripleo-container-image-push registry_url=192.168.24.1:8787, pull_source=None, append_tag=, username=None, password=None, dry_run=False, multi_arch=False, image_to_push=docker://brew-pulp-docker01.web.prod.ext.phx2.redhat.com:8888/rhosp15/openstack-mistral-engine:latest

I then ran sosreport -o openstack_instack, checked the resulting sosreport, and it included the 'history' file and indeed the password was masked :)

PR #1767, opened 2019-08-22
Fix and enhancements for OVN plugins

danalsan:
[ovn_host] Add support for containerized setups
[ovn_central] Add support to containerized setups
Allow reading commands output from containers
Minor enhancements to the OVN plugin

### DIFF
--- a/sos/plugins/ovn_central.py
+++ b/sos/plugins/ovn_central.py
@@ -19,30 +19,64 @@ class OVNCentral(Plugin):
     """
     plugin_name = "ovn_central"
     profiles = ('network', 'virt')
+    _container_runtime = None
+    _container_name = None
 
-    def add_database_output(self, filename, cmds, ovn_cmd, skip=[]):
+    def get_tables_from_schema(self, filename, skip=[]):
+        if self._container_name:
+            cmd = "%s exec %s cat %s" % (
+                self._container_runtime, self._container_name, filename)
+            res = self.get_command_output(cmd)
+            if res['status'] != 0:
+                self._log_error("Could not retrieve DB schema file from "
+                                "container %s" % self._container_name)
+                return
+            try:
+                db = json.loads(res['output'])
+            except Exception:
+                self._log_error("Cannot parse JSON file %s" % filename)
+                return
+        else:
+            try:
+                with open(filename, 'r') as f:
+                    try:
+                        db = json.load(f)
+                    except Exception:
+                        self._log_error(
+                            "Cannot parse JSON file %s" % filename)
+                        return
+            except IOError as ex:
+                self._log_error(
+                    "Could not open DB schema file %s: %s" % (filename, ex))
+                return
         try:
-            with open(filename, 'r') as f:
-                try:
-                    db = json.load(f)
-                except Exception:
-                    # If json can't be parsed, then exit early
-                    self._log_error("Cannot parse JSON file %s" % filename)
-                    return
-                try:
-                    for table in six.iterkeys(db['tables']):
-                        if table not in skip:
-                            cmds.append('%s list %s' % (ovn_cmd, table))
-                except AttributeError:
-                    self._log_error("DB schema %s has no 'tables' key" %
-                                    filename)
-                    return
-        except IOError as ex:
-            self._log_error("Could not open DB schema file %s: %s" % (filename,
-                                                                      ex))
-            return
+            return [table for table in six.iterkeys(
+                db['tables']) if table not in skip]
+        except AttributeError:
+            self._log_error("DB schema %s has no 'tables' key" % filename)
+
+    def add_database_output(self, tables, cmds, ovn_cmd):
+        for table in tables:
+            cmds.append('%s list %s' % (ovn_cmd, table))
+
+    def running_in_container(self):
+        for runtime in ["podman", "docker"]:
+            container_status = self.get_command_output(runtime + " ps")
+            if container_status['status'] == 0:
+                for line in container_status['output'].splitlines():
+                    if "ovn-dbs-bundle" in line:
+                        self._container_name = line.split()[-1]
+                        self._container_runtime = runtime
+                        return True
+        return False
+
+    def check_enabled(self):
+        return (self.running_in_container() or
+                super(OVNCentral, self).check_enabled())
 
     def setup(self):
+        containerized = self.running_in_container()
+
         ovs_rundir = os.environ.get('OVS_RUNDIR')
         for pidfile in ['ovnnb_db.pid', 'ovnsb_db.pid', 'ovn-northd.pid']:
             self.add_copy_spec([
@@ -68,10 +102,20 @@ class OVNCentral(Plugin):
 
         schema_dir = '/usr/share/openvswitch'
 
-        self.add_database_output(os.path.join(schema_dir, 'ovn-nb.ovsschema'),
-                                 cmds, 'ovn-nbctl')
-        self.add_database_output(os.path.join(schema_dir, 'ovn-sb.ovsschema'),
-                                 cmds, 'ovn-sbctl', ['Logical_Flow'])
+        nb_tables = self.get_tables_from_schema(os.path.join(
+            schema_dir, 'ovn-nb.ovsschema'))
+        sb_tables = self.get_tables_from_schema(os.path.join(
+            schema_dir, 'ovn-sb.ovsschema'), ['Logical_Flow'])
+
+        self.add_database_output(nb_tables, cmds, 'ovn-nbctl')
+        self.add_database_output(sb_tables, cmds, 'ovn-sbctl')
+
+        # If OVN is containerized, we need to run the above commands inside
+        # the container.
+        if containerized:
+            cmds = ['%s exec %s %s' % (self._container_runtime,
+                                       self._container_name,
+                                       cmd) for cmd in cmds]
 
         self.add_cmd_output(cmds)
 
--- a/sos/plugins/ovn_central.py
+++ b/sos/plugins/ovn_central.py
@@ -57,6 +57,8 @@ class OVNCentral(Plugin):
 
         # Some user-friendly versions of DB output
         cmds = [
+            'ovn-nbctl show',
+            'ovn-sbctl show',
             'ovn-sbctl lflow-list',
             'ovn-nbctl get-ssl',
             'ovn-nbctl get-connection',
--- a/sos/utilities.py
+++ b/sos/utilities.py
@@ -147,7 +147,7 @@ def sos_get_command_output(command, timeout=300, stderr=False,
         else:
             expanded_args.append(arg)
     try:
-        p = Popen(expanded_args, shell=False, stdout=PIPE,
+        p = Popen(expanded_args, stdin=PIPE, shell=False, stdout=PIPE,
                   stderr=STDOUT if stderr else PIPE,
                   bufsize=-1, env=cmd_env, close_fds=True,
                   preexec_fn=_child_prep_fn)
--- a/sos/plugins/ovn_host.py
+++ b/sos/plugins/ovn_host.py
@@ -12,6 +12,15 @@ import os
 from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
 
 
+pidfile = 'ovn-controller.pid'
+pid_paths = [
+        '/var/lib/openvswitch/ovn',
+        '/usr/local/var/run/openvswitch',
+        '/var/run/openvswitch',
+        '/run/openvswitch'
+]
+
+
 class OVNHost(Plugin):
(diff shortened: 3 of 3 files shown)


sos/plugins/ovn_central.py
```diff
@@ -57,20 +76,39 @@ def setup(self):
...
+            'ovn-nbctl show',
+            'ovn-sbctl show',
             'ovn-sbctl lflow-list',
             'ovn-nbctl get-ssl',
             'ovn-nbctl get-connection',
             'ovn-sbctl get-ssl',
             'ovn-sbctl get-connection',
         ]
 
+
```
pmoravec: Please remove this extra empty line (Travis fails on it - you can run "pycodestyle" on modified code to spot it locally before committing it).

sos/plugins/ovn_central.py
```diff
@@ -57,20 +76,39 @@ def setup(self):
...
         schema_dir = '/usr/share/openvswitch'
+        if containerized:
+            for db in ['nb', 'sb']:
+                cmd = ("%s cp %s:/usr/share/openvswitch/ovn-%s.ovsschema "
+                       "/tmp/" % (self._container_runtime,
+                                  self._container_name, db))
+                if self.get_command_output(cmd)['status'] != 0:
```
pmoravec: What is the purpose to copy the files? (to get the copy command output that is usefull for some troubleshooting?)

Is it safe to expose that file to publicly available `/tmp` directory - can't it be misused by some intruder?

danalsan: > What is the purpose to copy the files? (to get the copy command output that is usefull for some troubleshooting?)
> 
> Is it safe to expose that file to publicly available `/tmp` directory - can't it be misused by some intruder?
> 
> (I think I got this; you collect the file in add_database_output and I guess it is really safe to expose them since non-containerised environment exposes them in similar way - am I right?)

Yes, you're right. On non containerized environments, this is exposed in the host filesystem while in containerized environments we want to read the files to figure out what tables to pull data from.

Re. security, these files are just the DB schema and are publicly available anyways in the OVS/OVN repositories, they do not contain any data but just the schema definition.

sos/plugins/ovn_central.py
```diff
@@ -65,12 +86,37 @@ def setup(self):
...
+            for db in ['nb', 'sb']:
+                cmdline = ("%s/usr/share/openvswitch/ovn-%s.ovsschema "
+                           "/tmp/" % (cmd, db))
+                if self.get_command_output(cmdline)['status'] != 0:
+                    self._log_error("Could not retrieve DB schema file from "
+                                    "container %s", self._container_name)
+            schema_dir = '/tmp'
+            if self._container_runtime == 'podman':
+                self.get_command_output(
+                    "podman umount %s" % self._container_name)
```
TurboTurtle: Absolute nack from me. In addition to the fact that we shouldn't be mounting and unmounting resources in a plugin, there is also no reason to be capturing the output of that mount/unmount operation.

sos/plugins/ovn_central.py
```diff
@@ -65,12 +86,37 @@ def setup(self):
         ]
 
         schema_dir = '/usr/share/openvswitch'
+        if containerized:
+            if self._container_runtime == 'podman':
+                dst = self.get_command_output(
+                    "podman mount %s" % self._container_name)['output']
+                cmd = "cp " + dst.rstrip()
```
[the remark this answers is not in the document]
danalsan: > `podman cp` exists and matches the `docker` syntax to copy files out of a container. In no way should we be mounting anything anywhere in a plugin.

I took a RHEL 8.0 base system where `podman cp` doesn't exist. Do you mean is it fair to assume that whatever version podman is now packaged in RHEL8 would support that command? If so, I'll fallback to my previous version.

[root@controller-0 ovn_central]# podman cp
Command "cp" not found.
See `podman --help`.


> 
> This whole block to construct the command can be reduced to:
> 
> ```python
> if containerized:
>     cmd = "%s cp %s:" % (self._container_runtime, self._container_name)
> ```

sos/plugins/ovn_central.py
danalsan: > Absolute nack from me. In addition to the fact that we shouldn't be mounting and unmounting resources in a plugin, there is also no reason to be capturing the output of that mount/unmount operation.

There is absolutely a reason to capture the output of the *mount* operation:

# podman mount ovn-dbs-bundle-podman-0
/var/lib/containers/storage/overlay/d1679d89311b4a4fc4909b3126ab6abf1458fd4c1968061105b788867c6da69b/merged

It gives you the location where the filesystem is (actually this is not even a mount as without executing a `podman mount` one can access the filesystem directly in the host). But, again, reading the output would be needed.

For unmouting, how would you suggest doing it if it's not through `get_command_output`?

sos/plugins/ovn_central.py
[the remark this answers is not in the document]
TurboTurtle: It looks like `podman cp` is a 1.1 feature, whereas 8.0 has 1.0 podman. I'd imagine the next rebase for podman in RHEL will update it to at least 1.1 (Fedora is on 1.4 right now for reference).

In any event, I am still completely against doing mount operations within a plugin. Maybe there needs to be some logic to gate against podman-1.0, or we just accept that it won't work for that version of podman.

danalsan:
> On 22 Aug 2019, at 22:01, Jake Hunsaker <notifications@github.com> wrote:
> 
> @TurboTurtle commented on this pull request.
> 
> In sos/plugins/ovn_central.py:
> 
> > @@ -65,12 +86,37 @@ def setup(self):
>          ]
>  
>          schema_dir = '/usr/share/openvswitch'
> +        if containerized:
> +            if self._container_runtime == 'podman':
> +                dst = self.get_command_output(
> +                    "podman mount %s" % self._container_name)['output']
> +                cmd = "cp " + dst.rstrip()
> It looks like podman cp is a 1.1 feature, whereas 8.0 has 1.0 podman. I'd imagine the next rebase for podman in RHEL will update it to at least 1.1 (Fedora is on 1.4 right now for reference).
> 
> In any event, I am still completely against doing mount operations within a plugin. Maybe there needs to be some logic to gate against podman-1.0, or we just accept that it won't work for that version of podman.
> 
As I said, the directory is accesible even without mounting but I need to find it some how; it is not a real mount and anyways it’s just mounted right after the file is fetched. 

What then about using glob if Python 3 is installed to search for the files in /var/lib and “docker cp” otherwise?

We need the ovn plugin to work on RHEL8. 
Can you suggest something that would work for you and later on when podman > 1.0 is available, we will just use “podman cp”?

sos/plugins/ovn_central.py
TurboTurtle: Ah, fair point - I was reading that as `get_cmd_output` and not `get_command_output`, my fault there.

TurboTurtle:
> As I said, the directory is accesible even without mounting but I need to find it some how; it is not a real mount and anyways it’s just mounted right after the file is fetched. What then about using glob if Python 3 is installed to search for the files in /var/lib and “docker cp” otherwise? We need the ovn plugin to work on RHEL8. Can you suggest something that would work for you and later on when podman > 1.0 is available, we will just use “podman cp”?


As I said, I am still not onboard with performing file copies on the host. `podman cp` is a last resort in my mind, if we even agree to go that route at all.

What is represented by the table names in the `ovn-(nb|sb).ovsschema` files? Do they map to some other component within OVN that can be listed with some other ovn-ctl command? I would find it odd if an end user needed to go digging through schema files in order to run valid `ovn-nbctl list $schema` on their own. 

If some kind of list command like that is present, we can simplify this substantially and making it container-compatible would be trivial.

danalsan:
@TurboTurtle I changed the code to read the file inside the container to memory. If it's not a containerized setup, the old behavior is kept. This way we don't mount or copy any files into memory and it shouldn't be a problem as the schema file shouldn't grow much (it doesn't contain any data but just the tables, columns and relationships). Let me know if this works for you.
Thanks,
daniel

bmr-cymru:
This pull request **introduces 2 alerts** when merging 8178714c550458054a3d1fa94f7be218abea1fe8 into f9e0a231495adad0e165e908dd0dde768ed62ed0 - [view on LGTM.com](https://lgtm.com/projects/g/sosreport/sos/rev/pr-4f2faa7b7d6089e72d9710375d224930ba29462c)

**new alerts:**

* 1 for Unused local variable
* 1 for Wrong number of arguments in a call

---

---

danalsan:
The travis failures are unrelated to this PR, the offending lines belong to the following commit:

^2453c238 (Jake Hunsaker 2019-06-07 13:08:12 -0400  926)                         strfile = file_name.replace(os.path.sep, ".") + ".tailed"
^2453c238 (Jake Hunsaker 2019-06-07 13:08:12 -0400  927)                         self.add_string_as_file(tail(_file, sizelimit), strfile)

I can send a separate PR to fix it.

danalsan:
> The offending commit is actually [9c421f8](https://github.com/sosreport/sos/commit/9c421f819fcb40877aa873719656b8b3b6590d06)
> 
> But we can get that cleaned up in master in a moment before we close 3.8.

ack, thanks a lot.
Let me know please if there's anything you'd like me to change from this PR.

sos/utilities.py
```diff
@@ -147,7 +147,7 @@ def _child_prep_fn():
         else:
             expanded_args.append(arg)
     try:
-        p = Popen(expanded_args, shell=False, stdout=PIPE,
+        p = Popen(expanded_args, stdin=PIPE, shell=False, stdout=PIPE,
```
pmoravec: Why this change, please?

@bmr-cymru and @TurboTurtle , any comment here? Could it have some drawbacks?

sos/plugins/ovn_central.py
```diff
@@ -19,30 +19,61 @@ class OVNCentral(Plugin):
...
+                    except Exception:
+                        self._log_error(
+                            "Cannot parse JSON file %s" % filename)
+                        return
+            except IOError as ex:
+                self._log_error(
+                    "Could not open DB schema file %s: %s" % (filename, ex))
+                return
+        return [
+            table for table in six.iterkeys(db['tables']) if table not in skip]
```
pmoravec: Theoretically, the JSON file might lack 'tables' (can we 100% guarantee the json file cant be corrupted or empty?). If this would happen, an uncaught exception would be raised.

sos/plugins/ovn_central.py
```diff
@@ -66,10 +99,20 @@ def setup(self):
...
-                                 cmds, 'ovn-nbctl')
-        self.add_database_output(os.path.join(schema_dir, 'ovn-sb.ovsschema'),
-                                 cmds, 'ovn-sbctl', ['Logical_Flow'])
+        nb_tables = self.get_tables_from_schema(os.path.join(
+            schema_dir, 'ovn-nb.ovsschema'))
+        sb_tables = self.get_tables_from_schema(os.path.join(
+            schema_dir, 'ovn-sb.ovsschema'), ['Logical_Flow'])
+
+        self.add_database_output(nb_tables, cmds, 'ovn-nbctl')
+        self.add_database_output(sb_tables, cmds, 'ovn-sbctl')
```
pmoravec: Will be "cmds" variable really updated in the `cmds.append('%s list %s' % (ovn_cmd, table))` call inside `add_database_output`? As:

    >>> def plus(a,b,c):
    ...   c = a+b
    ... 
    >>> s1=3
    >>> s2=5
    >>> s=6
    >>> plus(s1,s2,s)
    >>> s
    6

(this could be over-worked by using "self.cmds" in `setup` and `add_database_output` and not passing "cmds" as its argument)

sos/utilities.py
danalsan: Thanks a lot for taking a look.
Without that change, podman hangs for me [0] and with strace I could see that the process is receiving SIGTTIN all the time.

[0] https://bugzilla.redhat.com/show_bug.cgi?id=1732525

sos/plugins/ovn_central.py
danalsan: Good point, plus it was handled before my patch. Will fix.

danalsan: Good point, plus it was handled before my patch. Will fix.
Thanks!

sos/plugins/ovn_central.py
danalsan: As it is a list, we can be sure that the cmds variable will be updated. (My patch is not changing the behavior though, it was like that before). For example:

def _bb(param):
    param.append(3)

aa = [1, 2]
_bb(aa)
print(aa)

$ python3.7 test.py
[1, 2, 3]

$ python2.7 test.py
[1, 2, 3]

pmoravec: Ah right, forgot it is a list..

Issue #1743, opened 2019-08-01
[last] soscleaner would benefit to have a better user filtering when obfuscating them


slashdd:
soscleaner is a project to perform sanitation against an existing sosreport.
The user obfuscation inside soscleaner is currently relying on `sos_commands/last/lastlog` from the sosreport it is run against but this generate too much false positives as soscleaner try to obfuscate everything found in there.

https://github.com/soscleaner/soscleaner/blob/master/soscleaner/soscleaner.py#L97

Current workaround is to maintain a ignored_user list
https://github.com/soscleaner/soscleaner/blob/master/soscleaner/soscleaner.py#L512-L563

which works well but is not ideal.

Would it be reasonable to incorporate in last plugin 

```
         self.add_cmd_output([
+          "lastlog -u 1000-60000"
```

so that soscleaner can rely on this file instead of lastlog. The goal is not to obfuscate all users such as well publicly know ('bin', 'sys', 'ubuntu', 'reboot', ....) but try to limit them to users created by users, this will benefit soscleaner user obfuscation a lot and limit the amount of false positives and avoid us to maintain a list of Linux base user such as https://salsa.debian.org/debian/base-passwd/blob/master/passwd.master and more.

For instance user 'bin' can be a user, part of a package name, a path, a file, ...

sosreport already provide a complete lastlog output, what I would like to see is in addition to that having a smaller list to make the obfuscation easier for soscleaner benefit.

slashdd:
Looking in `/etc/login.defs` on Ubuntu Bionic
```
UID_MIN			 1000
UID_MAX	         60000
```

slashdd:
@TurboTurtle @pmoravec 
Would you be okay with that proposal change in order to help user obfuscation inside SOSCleaner ?

pmoravec:
Do I get it right that current behaviour makes problems to `soscleaner` that wants to obfuscate usernames (for (all?) users with UID>1000) but `lastlog` contains a mix of system and regular users?

How the #1770 can help then, if full `lastlog` (mixing users to scrub and to show) will still be collected? Shall not we modify `sos` here to collect `lastlog -u 0-999` and `lastlog -u 1000-60000`?

slashdd:
[...]

PR #1770, opened 2019-08-30
[last] run lastlog to limit listing user to the UID_MIN-UID_MAX range.

slashdd:
Currently SOSCleaner relies on lastlog output.
This would be useful for SOSCleaner to sanitize users created
by users by making sure they are in the UID_MIN-UID_MAX range as per /etc/login.defs.

This will highly help avoiding false positive user obfuscation in SOSCleaner.

Fix: #1743

---


slashdd:
@jduncan-rva fyi

Issue #1771, opened 2019-09-03
Grafana support on Ubuntu


cjohnston1158:
Please expand the Grafana plugin to support Ubuntu. This should work with both snaps or debs.

pmoravec:
Same stuff from the same location etc. to be collected? Same package to trigger the plugin?

(the best to test it is to modify https://github.com/sosreport/sos/blob/master/sos/plugins/grafana.py#L14 by adding "`, DebianPlugin, UbuntuPlugin)`"

BryanQuigley:
Is the package name for RH/Fedora correct at "graphana"?   I couldn't find it under that name - https://pkgs.org/download/graphana.  They seem to have grafana thought

slashdd:
I found this:

https://access.redhat.com/documentation/en-us/red_hat_openstack_platform/9/html/red_hat_openstack_platform_operational_tools/installing_the_performance_monitoring_suite

```
Once that is done, install the Graphite and Grafana software by running the following command:

# yum install python-carbon graphite-web grafana httpd
```

pmoravec:
Indeed, RedHat/Fedora/EPEL has `grafana` package, not `graphana`.

PR #1772, opened 2019-09-03
[grafana] Updates Grafana plugin to work on Debian and Ubuntu.

cjohnston1158:
Closes: #1771

Signed-off-by: Chris Johnston <chris.johnston@canonical.com>

---

### DIFF
--- a/sos/plugins/grafana.py
+++ b/sos/plugins/grafana.py
@@ -8,16 +8,16 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.plugins import Plugin, RedHatPlugin
+from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
 
 
-class Grafana(Plugin, RedHatPlugin):
+class Grafana(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
     """Fetch Grafana configuration, logs and CLI output
     """
     plugin_name = "grafana"
     profiles = ('services', 'openstack', 'openstack_controller')
 
-    packages = ('graphana',)
+    packages = ('graphana', 'grafana')
 
     def setup(self):
         if self.get_option("all_logs"):


sos/plugins/grafana.py
```diff
@@ -8,16 +8,16 @@
...
 
-class Grafana(Plugin, RedHatPlugin):
+class Grafana(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
     """Fetch Grafana configuration, logs and CLI output
     """
     plugin_name = "grafana"
     profiles = ('services', 'openstack', 'openstack_controller')
 
-    packages = ('graphana',)
+    packages = ('graphana', 'grafana')
```
slashdd: That should be triggered for Xenial as it has grafana offered as a .deb package, but late release seems to only offered grafana as a snap.

`packages =` will only work for RPM and DEB, you'll need something else to trigger the SNAP grafana.

```
$ rmadison grafana
 grafana | 2.6.0+dfsg-1 | xenial/universe | source, amd64, arm64, armhf, i386, ppc64e
```

cjohnston1158: By default the charms use upstream repos to deploy with debs.

slashdd: Ok, thanks for confirming. Would be nice to make that plugin trigger when grafana snap is installed eventually, but I don't see this as a blocker for the time being, it was more a fyi ;)

pmoravec: `grafana` is the package name on Red Hat distros and I assume on Debian/Ubuntu as well.

@psachin : where the `graphana` package name comes from, please? Or is it a typo? (shall not be here just `packages = ('grafana')`?)

cjohnston1158:
> Ack to the change, but please fix the commit message for [2e1d228](https://github.com/sosreport/sos/commit/2e1d2283b7fb09692d1559eb33a5549a2d07106f) to fit the contributor guidelines ([component] <short description>)

Done

BryanQuigley:
If we're going with the same commit message, I'd just amend them to be one commit mentioning graphana is a typo.

Otherwise, change the second one to mention that graphana was a typo, and switching all to grafana.

PR #1779, opened 2019-09-06
[manageiq] modify plugin to account for regularly large files

rh-dluong:
The following 5 files are usually over the default 25mb limit:
    ansible_tower.log
    automation.log
    evm.log
    production.log
    top_output.log

Those 5 files interfere with the *.log glob to behave as users
expect.  These 5 files were broken out to miq_main_log_files.

Signed-off by: David Luong <dluong@redhat.com>

---

### DIFF
--- a/sos/plugins/manageiq.py
+++ b/sos/plugins/manageiq.py
@@ -57,10 +57,20 @@ class ManageIQ(Plugin, RedHatPlugin):
 
     # Log files to collect from miq_dir/log/
     miq_log_dir = os.path.join(miq_dir, "log")
+
+    miq_main_log_files = [
+        'ansible_tower.log',
+        'top_output.log',
+        'evm.log',
+        'production.log',
+        'automation.log',
+    ]
+
     miq_log_files = [
         '*.log',
         'apache/*.log',
         '*.txt',
+        '*.yml',
     ]
 
     def setup(self):
@@ -73,6 +83,10 @@ class ManageIQ(Plugin, RedHatPlugin):
         self.add_copy_spec([
             os.path.join(self.miq_conf_dir, x) for x in self.miq_conf_files
         ])
+
+        self.add_copy_spec([
+            os.path.join(self.miq_log_dir, x) for x in self.miq_main_log_files
+        ], sizelimit=0)
 
         self.add_copy_spec([
             os.path.join(self.miq_log_dir, x) for x in self.miq_log_files


rh-dluong:
I've not yet figured out how to filter out the items found in `miq_main_log_files` from `miq_log_files`.  Can someone suggest or assist me in this?

rh-dluong:
That or either mimic what the CloudForms UI does and just collect everything in that log dir regardless of size, which would be a much simpler fix of setting that limit to 0 in the add_copy_spec line for miq_log_files

TurboTurtle:
How large are these files expected to grow to be? We're hesitant to use `sizelimit=0` by default but are open to it when justified.

Beyond that, what other file patterns exist in that directory that `*.log` globs to?

rh-dluong:
`evm.log` itself can grow to hundreds of mb on a busy system
`top_output.log` just writes what top is to that log file every minute, so that also grows a bit.

https://bugzilla.redhat.com/show_bug.cgi?id=1746941  is what I made this PR for

Currently the issue is evm.log is almost always the last log being written to when sos is ran since it's the busiest log.  But with the add_spec collecting *.log, this is a problem since this log is almost always bigger than 25mb (we have no way of predicting how large this log will be when we need to pull it).  So the glob pretty much just collects evm.log whenver sos is run without the --all-logs option being specified. And usually we need _all_ of `evm.log` regardless of size. 

So my idea is to split out the files that tend to be larger in `miq_main_log_files` and we could possible limit those to have that limit be 0, and let the other logs just collect from the *.log glob from `miq_log_files` with the default size.  But I do not know how to exclude `miq_main_log_files` items form the glob (they all qualify for the glob).

But if we wanted parity with the log collector that CloudForms uses for its UI, the limit would be 0 for all logs in that dir since we just collect all current logs from that `/var/www/miq/vmdb/log/` dir.  If we went this way I could possibly open up a conversation with CloudForms eng to use sosreport instead of our custom one in the UI.

TurboTurtle:
In that case we'd probably want to/need to perform the glob directly in the plugin, pull the `miq_main_log_files` out from that list, and then call `add_copy_spec()` on the already-determined files from the glob. For example something like this (not actually tested, this is just off the top of my head and could certainly be cleaned up):

```python
_main_logs = [
        'ansible_tower.log',
        'top_output.log',
        'evm.log',
        'production.log',
        'automation.log',
    ]
miq_main_log_files = [os.path.join(self.miq_log_dir, _main_log) for _main_log in _main_logs] 
miq_logs_glob = [f for f in glob.glob(os.path.join(self.miq_log_dir, '*.log')) if f not in miq_main_log_files]

self.add_copy_spec(miq_logs_glob)
self.add_copy_spec([
            os.path.join(self.miq_log_dir, x) for x in self.miq_main_log_files
        ], sizelimit=0)
```

It's not exactly pretty, but it's not terrible either. Again, could certainly be cleaned up a bit but the above is the general idea.

rh-dluong:
Thanks @TurboTurtle , in my new commit, I've changed it around a little and now it works as expected.

bmr-cymru:
This pull request **introduces 2 alerts** when merging 583edc97550f3b27c1a24b962d0314da51c29b2c into f2c538bb7c10f442053aecd5d9c0a78024daefd0 - [view on LGTM.com](https://lgtm.com/projects/g/sosreport/sos/rev/pr-a3993031265f56df6a8c3a6c990e374cb14f9ef2)

**new alerts:**

* 2 for Non\-iterable used in for loop

---

---

rh-dluong:
Hey @TurboTurtle, one last question.

LGTM.com flagged that code as having 2 alerts when I defined miq_logs_glob and miq_main_log_files under class but not the `def setup()`:
https://lgtm.com/projects/g/sosreport/sos/rev/pr-a3993031265f56df6a8c3a6c990e374cb14f9ef2

The alerts no longer flag when I move it under the `def setup()`.  Is this a false alarm?  Because it does work as expected before I changed my PR, ie, it is able to be iterative when it is not under `def setup()`.

bmr-cymru:
I'm not hugely keen on this solution - we are trying to move away from individual plugins having complex globbing and other plugin-local code. This patch feels like it's going in the wrong direction.

I think it's better to make the semantics of the `add_copy_spec()` interface better defined, so that a plugin like this can rely on well-defined and expected behaviour: e.g. add the "main" log patterns *first*, with a `sizelimit=0` and then guarantee via the interface that if they are added later (with a different `sizelimit`) that the first addition is honoured - this should be documented as part of the method docstring. This seems like it would give consistent behaviour for existing users, and allows a plugin to specify in a fairly natural way that certain files be handled specially without having to jump through hoops and reimplement the behind-the-scenes `Plugin` plumbing.

rh-dluong:
Hey @bmr-cymru, that kind of change would be beyond the scope of this particular PR and affect sosreport as whole wouldn't it and would require a different PR if I'm understand you correctly, ie, I'm taking it as to change the `add_copy_spec() behavior as whole instead of just this plugin?

If that is the case, I have no problem to wait for this kind of change to happen and refactor my PR accordingly.

Or do you mean I can refactor how `add_copy_spec()` works here?  In which case, I'm out of my depth and will need further assistance.

TurboTurtle:
@rh-dluong yeah, the ideal scenario @bmr-cymru is describing would be done to `add_copy_spec()` directly. I can take it as an AI on myself. I don't think this should be _too_ difficult at first glance, but we'll see if I put my foot in my mouth about that ;-)

I'll try and carve out some time this week to work on this.

rh-dluong:
Hello @bmr-cymru and @TurboTurtle .  I have made the requested change to `add_copy_spec()` and will refactor this PR once #1802 is merged/modified.

bmr-cymru:
#1802 has been merged now - that should simplify this PR.

PR #1780, opened 2019-09-10
[kernel] Add the output of 'dmesg -H'

jcastill:
At the moment we have the output of 'dmesg' only in the sosreport, plus
the file /var/log/dmesg. But the output of 'dmesg' doesn't contain
timestamps in a human readable format. While the timestamps could be
inaccurate in some circumstances (see the man page of dmesg), it
can be still very helpful.

---

### DIFF
--- a/sos/plugins/kernel.py
+++ b/sos/plugins/kernel.py
@@ -78,6 +78,7 @@ class Kernel(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
 
         self.add_cmd_output([
             "dmesg",
+            "dmesg -H",
             "sysctl -a",
             "dkms status"
         ])


bmr-cymru:
We turned this down last year in #1269 - I guess I'm struggling to understand why it's so useful, or can't be created after-the-fact (and whether the documented limitations can be safely ignored).

BryanQuigley:
Can't you get "better" timestamps just by looking at Syslog or Systemd Journal?

(From a kernel perspective I recently learned that adding printk_time=1 initcall_debug to the kernel cmdline gets much better timing data for bootup - but that's far outside of sosreport scope).

jcastill:
To be honest I'm struggling to find a justification for this anymore.
On one hand, we don't always have the kernel ring buffer in other places - this pull request was born out of frustration when I had to diagnose an issue, and had no way to match certain logs. Looking at the default (unreadable to me) timestamps was an extra layer of complexity, since I had to use one of the scripts that exist online to transform the timestamps to something more human readable. 
On the other hand, there is the issue of 'suspend' potentially affecting the timestamps with 'dmesg -H'. But in all this time I haven't been able to reproduce that with virtual machines. Still, if it happens, you find yourself unable to reconcile the different timestamps. One could always then go back to the default timestamps and use the script. 
So, what do you think? Should we wait until 'dmesg -H' is fixed? Assuming it will get fixed. Or should we add 'dmesg -H' because when the servers are not suspended, is a very useful information, and we don't need an external script to get it?

BryanQuigley:
We already collect multiple copies of the logs, so my vote is we leave dmesg -H out

TurboTurtle:
I'm closing this out per the last few comments.

PR #1782, opened 2019-09-12
[ipa] collect ipa-healthcheck logs

rcritten:
If the [free]ipa-healthcheck package is installed collect any
logs it has generated.

Signed-off-by: Rob Crittenden <rcritten@redhat.com>

---


rcritten:
Is there something I need to do to get a second reviewer assigned?

Issue #1663, opened 2019-04-26
[plugins] extras plugin to collect user defined cmds output?


pmoravec:
I have an idea that will help rather downstream maintenance than to sos upstream. The user story behind this follows:

- somebody wants to collect some files or command outputs
- upstream PR raised, merged after some time
- backported to downstream, released after another bunch of time

The whole meantime, `sosreport` does not collect the required data.

One way how to minimize this time lapse and to be more flexible is to have an `extras` plugin that will:
- open a predefined file (`/etc/sos-extras-commands.conf` or similar, by default empty)
- interpret each line as a command to execute
- call `add_cmd_output` to them

To collect a file, one would call `cat /path/to/file`. This is a compromise to prevent issues like #1596 when a user wants to collect a file already being collected. At the end, `sos_commands/extras` directory will contain outputs of those commands.

Advantages:
- it will allow much more flexible reaction on *temporarily* enhancing sosreport in collecting some data (until the upstream PR reaches downstream in a while) - this helps downstream maintenance
- the file with the list of extra commands can be easily maintained / content automated / spread over machines, such that all systems one needs to collect some extra data can be easily updated to meet this requirement

Disadvantages:
- no control what is being collected / what command will be executed via `sosreport`. Which means:
  - we still might get an issue similar to #1596 but for commands (hypothetically, now)
  - sosreport could be blamed by failing on some rogue commands a user requested to collect (though it won't be `sosreport`'s mistake)
  - basically the feature opens doors for potential code injection with its security impacts - a user typo/error (or wrong will) would force `sosreport` to execute dangerous/unwanted command (on the other hand: a user who can modify that file can also run the commands directly)
- it will help in rather smaller requirements but wont help when e.g. one needs to add some logic around (iterate over some cmd output to call next commands, run commands under some plugin only etc.)

Any thoughs?

TurboTurtle:
One thought comes to mind - we have a pretty strict rule of "no shell code" in sos plugins. Should we do any inspection of commands listed in the config file? For example, should we check for obvious shell code constructs and refuse to run those?

lzap:
I like the idea, I am worried about different sos cadence too. I had similar idea in mind actually and created patch for our foreman-maintain tool to provide the same: https://github.com/theforeman/foreman_maintain/pull/254 but if this is implemented in sosreport in a generic way, there would be no need of this in Foreman.

pmoravec:
[...]

PR #1788, opened 2019-09-19
[sos_extras] A new plugin to collect user defined commands

pmoravec:
Enabled by presence of /etc/sos.extras.d/ directory, the plugin will
execute all commands from all files in the directory, in order to
collect extra user-defined data e.g. specific for a given system
or product (or not yet being collected by sosreport otherwise).

Every such file with commands must start by a header warning users to
consult with sos comunity the commands first, to prevent possibly
dangerous commands being executed.

Resolves: #1663

---

### DIFF
--- a/sos/plugins/sos_extras.py
+++ b/sos/plugins/sos_extras.py
@@ -1,0 +1,92 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.plugins import Plugin, RedHatPlugin, UbuntuPlugin, DebianPlugin
+import os
+import stat
+
+
+class SosExtras(Plugin, RedHatPlugin, UbuntuPlugin, DebianPlugin):
+    """Collect extra data defined in /etc/sos.extras.d/ .
+    """
+
+    """The plugin traverses 'extras_dir' directory and for each file there,
+    it executes commands or collects files optionally with sizelimit. Expected
+    content of a file:
+    - empty lines or those starting with '#' are ignored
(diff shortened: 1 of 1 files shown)


bmr-cymru:
As well as running commands isn't it useful to allow additional file system paths in this file to be collected? Maybe something like:

```
some_command_to_run --with --args
:/var/log/some_command.log -
:/var/log/some_command.big.log 0
```

Where lines beginning with `:` are a path to collect, with a size limit in the 2nd column?

I'm not sure I like the "magic disclaimer text" mechanism - I think it's good practice to explain what is and is not supported in this file, but I am not sure there's value in verifying that the disclaimer text is there before honouring the command list - this is a root-owned file in a path only writable by root. To some extent we have to trust that (and the presence of the text in the file we read doesn't really give any additional confidence). Perhaps it's better to add some warning text to the log / UI output when this plugin finds commands to execute in the file.

I think the current wording for the disclaimer can also be improved - right now it seems to imply that you /could/ put shell notation in this tile, but that you /shouldn't/ - in fact we just don't support shell notation at all (without an explicit `sh -c ...`).

pmoravec:
The extension to collecting files makes sense, I will implement it.

The disclaimer text: enforcing it as a head of any file with commands to execute sounds to me the only way how to ensure anyone playing with this feature *will* read the disclaimer, *before* using it. I am ok with any warning - e.g. just a warning printed to stdout whenever an extras file is processed? 

About the disclaimer text: that was just very first draft (of a content I am not fully sure of). What we shall warn about? Not using pipes or multiple commands ("cmd1; cmd2")? Not running commands on background? Ensure no command changes the system anyway (esp. removing a file or similar)?

And where the warning text should be? In manpages (not big audience), on web (and url link to it in the warning message)? Somewhere else?

pmoravec:
Updated the PR to collect files as well and without the disclaimer mandatory head.

pmoravec:
Good point wrt `--list-plugins`, fixed.

I also converted explicit `check_enabled` method to standard `files = ..` approach and rebased the PR to current upstream.

PR #1793, opened 2019-09-20
[subscription_manager] collect list --available

wbclark:
Collect `subscription-manager list --available`

---

### DIFF
--- a/sos/plugins/subscription_manager.py
+++ b/sos/plugins/subscription_manager.py
@@ -29,6 +29,8 @@ class SubscriptionManager(Plugin, RedHatPlugin):
             "/var/log/rhsm/rhsmcertd.log"])
         self.add_cmd_output([
             "subscription-manager list --installed",
+            "subscription-manager list --available",
+            "subscription-manager list --all --available",
             "subscription-manager list --consumed",
             "subscription-manager identity",
             "syspurpose show"


TurboTurtle:
I can see this being useful to collect, however there is definitely some info in this output that would need to be scrubbed. I image at a minimum, account number and pool IDs would need to be scrubbed. @pmoravec anything else?

wbclark:
@TurboTurtle the output fields collected (including account number, pool IDs, etc) are the same as for e.g. `subscription-manager list --consumed` which isn't currently being scrubbed in any way. So if this is indeed an issue it would need to be resolved for all the `subscription-manager list` output collected by this plugin.

TurboTurtle:
Good catch. Thinking more on it, maybe pool IDs don't need to be scrubbed, but I think account numbers certainly should be. I'll defer to Pavel on this one.

pmoravec:
I see account numbers as sort of login - why to obfuscate them in sosreport? This info can be valuable in some specific situations (to identify if wrong account is used to register a system). So if the info can sometimes help and it needs additional work to hide the info (and I am not aware of mis-using the info), why to scrub it?

Despite my background is in system management, I am not deep expert on this so I might missed something..?

pmoravec:
Pool IDs are definitely useful to collect (and not scrub) - esp. for Satellite / candlepin client systems, this can help to identify particular subscriptions o the RHSM server.

bmr-cymru:
Please use the `[component] brief description of change` format for patch subjects, e.g. `[subscription_manager] collect list --available`. You can use `git rebase` or `commit --amend` to modify the branch and update with `git push --force`.

bmr-cymru:
I don't think there's anything here that we'd normally remove in `sos` postprocessing (afaik `soscleaner` doesn't handle this data specially either).

wbclark:
@bmr-cymru updated commit message, thanks.

wbclark:
@bmr-cymru Hi Bryn, sorry about that! I've removed the redundant line and signed off on the commit.

PR #1794, opened 2019-09-24
[kernel] Add collection of livepatch information

kamalesh-babulal:
Kernel livepatch support is available since, kernel version v4.0.
Collect information relating to livepatch available under the path
/sys/kernel/livepatch/. Each of the subdirectories following the path
is a kernel livepatching module and hosts the information about the
current patching state of the functions to be livepatched by the
module.

Signed-off-by: Kamalesh Babulal <kamalesh@linux.ibm.com>

---


kamalesh-babulal:
@BryanQuigley Thanks for the review and approval.

Re-thinking, sosreport should also capture the kernel functions being livepatched
by every livepatch kernel module. The sysfs files exported per module are:
```
/sys/kernel/livepatch
/sys/kernel/livepatch/<patch>
/sys/kernel/livepatch/<patch>/enabled
/sys/kernel/livepatch/<patch>/transition
/sys/kernel/livepatch/<patch>/force
/sys/kernel/livepatch/<patch>/<object> [1]
/sys/kernel/livepatch/<patch>/<object>/<function,sympos> [1]
```
[1] are empty directories and `__init__.py:_copy_dir()` skips them, one idea is
do a `ls --recursive /sys/kernel/livepatch/`.  Which will capture the object (vmlinux or the module name) and the functions name.

```
--- a/sos/plugins/kernel.py
+++ b/sos/plugins/kernel.py
@@ -109,6 +109,7 @@ class Kernel(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
             "/sys/module/*/version",
             "/sys/firmware/acpi/*",
             "/sys/kernel/debug/tracing/*",
+            "/sys/kernel/livepatch/*",
             "/proc/kallsyms",
             "/proc/buddyinfo",
             "/proc/slabinfo",
@@ -152,4 +153,8 @@ class Kernel(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
         for map_id in self.get_bpftool_map_ids(map_file):
             self.add_cmd_output("bpftool map dump id %s" % map_id)
 
+        # collect functions being livepatched by each livepatch module
+        self.add_cmd_output("ls --recursive /sys/kernel/livepatch/")
+
+
 # vim: set et ts=4 sw=4 :
```
If the idea looks ok, I can push v2 with the update.

BryanQuigley:
I'm not against it, but I am curious if we should be copying empty directories.

It wasn't written to explicitly exclude empty directories, just to only copy files.

kamalesh-babulal:
> I'm not against it, but I am curious if we should be copying empty directories.

Thanks. Copying empty directories will be helpful in this case.

> It wasn't written to explicitly exclude empty directories, just to only copy files.

My knowledge is limited with sosreport code, the simplest way to achieve would be:
```
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -722,6 +722,9 @@ class Plugin(object):
             return
         else:
             if stat.S_ISDIR(st.st_mode) and os.access(srcpath, os.R_OK):
+                if not os.listdir(srcpath):
+                    self.archive.add_dir(srcpath)
+                    return
                 self._copy_dir(srcpath)
                 return
 ```

BryanQuigley:
I'm +1 that code change with testing so far - it adds about 40 empty other directories- https://pastebin.ubuntu.com/p/9KrKBQ7g2s/

Some are very useful to be able to confirm they exist and are empty (like apparmor disable).

I didn't test to see if has other implications on the generated sosreport.

kamalesh-babulal:
@BryanQuigley Thank you so much, for the review and testing the patch. I will push v2, with two patches, one for the empty directory copy and the original livepatch information collection patch.

PR #1795, opened 2019-09-24
[kubernetes] separate Red Hat and Ubuntu Kubernetes plugins

bmr-cymru:
Separate the Red Hat and Ubuntu versions of the kubernetes plugin to account for differences in packaging and configuration paths.

This preserves the logic for setting `kube_cmd` from #1654 - i.e. the default is `kubectl`, which is overridden in the Red Hat plugin to add admin.kubeconfig for `--kubeconfig`, and in the Ubuntu Plugin to add`/root/cdk/kubeproxyconfig` if that path exists.

If Kubernetes is only distributed as a snap on Ubuntu and will always use this path then we can simplify the `UbuntuPlugin` and get rid of the `setup()` method - it wasn't clear from the original change to the single Kubernetes class whether this was the case or not.

---

### DIFF
--- a/sos/plugins/kubernetes.py
+++ b/sos/plugins/kubernetes.py
@@ -15,22 +15,11 @@ from os import path
 import re
 
 
-class kubernetes(Plugin, RedHatPlugin, UbuntuPlugin):
-
+class Kubernetes(Plugin):
     """Kubernetes plugin
     """
 
-    # OpenShift Container Platform uses the atomic-openshift-master package
-    # to provide kubernetes
-    packages = ('kubernetes', 'kubernetes-master', 'atomic-openshift-master')
     profiles = ('container',)
-    # use files only for masters, rely on package list for nodes
-    files = (
-        "/var/run/kubernetes/apiserver.key",
-        "/etc/origin/master/",
-        "/etc/origin/node/pods/master-config.yaml",
-        "/root/cdk/kubeproxyconfig"
-    )
 
     option_list = [
         ("all", "also collect all namespaces output separately",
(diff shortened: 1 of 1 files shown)


bmr-cymru:
Updated the branch to remove a comment that seemed out of date from the Red Hat file list.

slashdd:
@niedbalski fyi

bmr-cymru:
... and to get rid of the Red Hat specific comments from the Ubuntu class.

sos/plugins/kubernetes.py
```diff
@@ -182,4 +163,31 @@ def postproc(self):
...
+    # OpenShift Container Platform uses the atomic-openshift-master package
+    # to provide kubernetes
+    packages = ('kubernetes', 'kubernetes-master', 'atomic-openshift-master')
+
+    files = (
+        '/etc/origin/master/admin.kubeconfig',
+        '/etc/origin/node/pods/master-config.yaml',
+    )
+
+    kube_cmd = "kubctl --kubeconfig=/etc/origin/master/admin.kubeconfig"
```
TurboTurtle: Small typo - `kubctl`

bmr-cymru: Thanks!

sos/plugins/kubernetes.py
```diff
@@ -182,4 +163,31 @@ def postproc(self):
...
+    # OpenShift Container Platform uses the atomic-openshift-master package
+    # to provide kubernetes
+    packages = ('kubernetes', 'kubernetes-master', 'atomic-openshift-master')
+
+    files = (
+        '/etc/origin/master/admin.kubeconfig',
+        '/etc/origin/node/pods/master-config.yaml',
+    )
+
+    kube_cmd = "kubectl --kubeconfig=/etc/origin/master/admin.kubeconfig"
```
TurboTurtle: This config file is only valid for OCP deployments of kubernetes. While this is the only type of deployment that RH currently supports, it is possible to run "plain" kubernetes on RHEL, in which case the `kubectl` commands will fail with this `--kubeconfig` as that file won't be present.

PR #1796, opened 2019-09-24
[snappy] check for connectivity to the snap store

slashdd:
Check for connectivity to the snap store "api.snapcraft.io".

---

### DIFF
--- a/sos/plugins/snappy.py
+++ b/sos/plugins/snappy.py
@@ -24,6 +24,7 @@ class Snappy(Plugin, UbuntuPlugin, DebianPlugin, RedHatPlugin):
             "snap --version",
             "snap changes"
         ])
+        self.add_cmd_output("snap debug connectivity", timeout=10)
         self.add_service_status("snapd")
         self.add_journal(units="snapd")
 


slashdd:
This can be useful information especially if the system is behind a firewall or an air gap environment.

slashdd:
Same command line is used inside _checkStoreConnectivity() of source package 'ubuntu-release-upgrader' (AKA do-release-upgrade) to upgrade Ubuntu to latest release)

Code:
https://git.launchpad.net/ubuntu/+source/ubuntu-release-upgrader/tree/DistUpgrade/DistUpgradeQuirks.py#n418

slashdd:
```
Usage:
  snap debug <command>
  ....
Available commands:
  ....
  => connectivity      Check network connectivity status
  ....
```

BryanQuigley:
I think this would need to be behind an option like other options that cause network lookups.

Regardless, I think it would be more useful to have a general proxy detector in the networking plugin.  Maybe curl a https site and record the cert used.  We already have tracert there.

slashdd:
I still think using 'snap debug connectivity' for snapd is the way to go, as we won't have to maintain the plugin as the snapstore API changes (version, ....) snap is still new, so changes may still occurs.

Using this approach also ensure that the check will be consistently tested during the snapd regression testing at any version upgrade. 

Especially that this connectivity check has been ask and needed by foundation team so that the foundations team can use it in the installer (ubuntu-release-upgrader) to ensure that users who upgrade behind a local mirror with no full internet connectivity can be warned/triggered.

So the command line is there to stay, despite what --help mentioned.

```
The debug command contains a selection of additional sub-commands.

Debug commands can be removed without notice and may not work on
non-development systems.
```
While the curl approach migth be useful for other cases, I don't think it would necessary apply for snapstore.

Thought ?

ddstreet:
> I think this would need to be behind an option like other options that cause network lookups.

I don't quite understand what you mean by this?

> Regardless, I think it would be more useful to have a general proxy detector in the networking plugin. Maybe curl a https site and record the cert used. We already have tracert there.

having that probably would be nice, but is entirely unrelated to having the snap tool check if it has connectivity to the snapstore, right?

slashdd:
## Testing
```
ls -altr sosreport-snapstore-20190924192544/SnapStoreConnectivity 
lrwxrwxrwx 1 root root 43 Sep 24 19:25 sosreport-snapstore-20190924192544/SnapStoreConnectivity -> sos_commands/snappy/snap_debug_connectivity
```
## With no full internet access
```
$ cat sosreport-snapstore-20190924192544/SnapStoreConnectivity 
Connectivity status:
 * api.snapcraft.io: unreachable
error: 1 servers unreachable
```
## With full internet access
```
$ cat sosreport-nuc7-20190924153059/SnapStoreConnectivity 
Connectivity status:
 * PASS
```

BryanQuigley:
For example the networking plugin can do a tracert to archive.ubuntu.com but you need to specify it as a plugin option or do all plugins.

>While the curl approach might be useful for other cases, I don't think it would necessary apply for snapstore.

Why do we care so much for the snap store in this context?  Also how does this work with enterprise app store?

ddstreet:
> For example the networking plugin can do a tracert to archive.ubuntu.com but you need to specify it as a plugin option or do all plugins.

Ok, but that has nothing to do with this...I'm not saying that it's a bad idea to do that, just that doing that, or not, is unrelated to this.

> 
> > While the curl approach might be useful for other cases, I don't think it would necessary apply for snapstore.
> 
> Why do we care so much for the snap store in this context? Also how does this work with enterprise app store?

We care if the system can reach the snap store, to install/upgrade snaps.  It (snap) can do that either with a direct connection or through an enterprise proxy or using an airgapped proxy.  Asking snap to tell us if it has a connection means we don't need to worry how the system is configured to access the snap store, it will tell us if it can connect, or not, regardless of the mechanism.

slashdd:
`do-release-upgrade` has a few deb2snap mechanism in place to install/upgrade package as snap (IIRC, such as gnome packages for ubuntu-desktop for instance). Also some package does have the ability to convert themselves (e.g. lxd) to a snap at package/release upgrade. (nothing to do with do-release-upgrade here).

Knowing if the snapstore is accessible will become very important (regardless of the way customer/users access it). What we want to know is if the snapstore accessible or not.

I also added a timeout, to make sure the command ends if its waiting for network/dns/.... for too long.

bmr-cymru:
There are a fairly large number of plugins that will attempt some kind of network IO already - I don't see any reason to put this behind an option unless it's expected to often take a long time, or has some kind of bad timeout/hang behaviour when the network is unreachable.

slashdd:
@bmr-cymru I saw a couple of random occasions during my testing where it took more time than usual for the cmdline to complete. Nothing dramatic but I preferred to be safe than sorry.

bmr-cymru:
Minor variation is fine and to be expected with network services - it's only if it routinely goes out to lunch for minutes at a time (to the point that default timeouts expire) that there'd be any concern. Setting your own 10s timeout is also fine.

ddstreet:
> There are a fairly large number of plugins that will attempt some kind of network IO already

this tests specifically if snap can reach the snapstore, because many large enterprise deployments are firewalled, so that general network IO will work fine - and some common internet endpoints may actually be reachable - but the snapstore might not be.  Alternately, enterprises might have no connectivity to common internet endpoints (e.g. no route to google.com), but may actually be configured with a snap proxy so that the snap application *is* able to reach the snap store.

And to clarify the 'snap proxy' is not a generic proxy the entire system is configured with like squid; it's this:
https://docs.ubuntu.com/snap-store-proxy/en/install

Hence, it's actually quite useful to specifically gather 'snap debug connectivity', regardless of what other networking tests are run, because this is the *only* way to reliably know if snap can, or can't, access the snap store.

BryanQuigley:
I thought that we couldn't do network activity, with the timeout set, I'm generally +1.  I'm not sure if I want this to clutter the root of sosreports though.

slashdd:
The `root_symlink` has been added thinking it would be more easy/obvious for support folks looking at a sosreport, since we got a couple of situation like that lately. Also since snap become more and more a thing as we progress, ...

Although, I don't have strong feeling about it. I'm fine with removing the symlink if overall consensus prefer me to do so.

BryanQuigley:
This just seems to be too much of a special case to promote (or again, maybe I'm still missing the significance of it). 

Out of the items we would want to promote to rootdir from snappy I'd think it should be snap_list_--all, as installed-snaps.

slashdd:
`root_symlink` removed

bmr-cymru:
> Hence, it's actually quite useful to specifically gather 'snap debug connectivity', regardless of what other networking tests are run,

That's why I said it didn't need to be behind an option - a bunch of plugins already do similar things, so there's no need to treat this any differently.

ddstreet:
> That's why I said it didn't need to be behind an option

ah sorry, clearly i didn't read your comment well enough :)

PR #1797, opened 2019-09-24
[manageiq] modify plugin to check for modified files supplied

rh-dluong:
We have noticed an uptick in customers modifying appliance supplied code from our official repositories.  We would like to run an rpm -V against the following 5 packages to check for this to expedite the process of elimination for engineering and support:
        cfme
        cfme-appliance
        cfme-gemset
        cfme-appliance-tools
        cfme-appliance-common

[Bugzilla 1755080](https://bugzilla.redhat.com/show_bug.cgi?id=1755080)

Signed-off by: David Luong dluong@redhat.com

---

### DIFF
--- a/sos/plugins/manageiq.py
+++ b/sos/plugins/manageiq.py
@@ -24,7 +24,13 @@ class ManageIQ(Plugin, RedHatPlugin):
 
     miq_dir = '/var/www/miq/vmdb'
 
-    packages = ('cfme',)
+    packages = (
+        'cfme',
+        'cfme-appliance',
+        'cfme-gemset',
+        'cfme-appliance-tools',
+        'cfme-appliance-common'
+    )
     files = (
         os.path.join(miq_dir, 'BUILD'),
         os.path.join(miq_dir, 'GUID'),
--- a/sos/policies/redhat.py
+++ b/sos/policies/redhat.py
@@ -194,6 +194,9 @@ RHOSP_DESC = "Red Hat OpenStack Platform"
 RHOCP = "ocp"
 RHOCP_DESC = "OpenShift Container Platform by Red Hat"
 
+RH_CFME = "cfme"
+RH_CFME_DESC = "Red Hat CloudForms"
+
 RH_SATELLITE = "satellite"
 RH_SATELLITE_DESC = "Red Hat Satellite"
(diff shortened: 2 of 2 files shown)


sos/plugins/manageiq.py
```diff
@@ -84,4 +84,9 @@ def setup(self):
                     os.path.join(pg_dir, 'postgresql.conf')
             ])
 
+        self.collectExtOutput([
+            "rpm -V cfme cfme-appliance cfme-gemset
```
TurboTurtle: This is what the `--verify` option is for. Add the package names to the `packages` class member and when `--verify` is used, the package manager defined by the policy will run a verify command against all packages listed.

rh-dluong: Thanks for the feedback, made the change as requested.

pmoravec:
Do you need to verify the packages on each and every run of sosreport (and `manageiq` plugin)? Then this change won't suffice. Either a sosreport `preset` can be created to automatically enable `--verify` option when a `cfme*` package is installed (seems legit), or the `rpm -V` call shall be added.

If you need to verify integrity of the packages on demand only, then ACK to the PR.

rh-dluong:
@pmoravec Yeah, we'd like to run this everytime sos is ran.  In that case do I also need to modify redhat.py under policies?

TurboTurtle:
@rh-dluong yes, presets get added in the policy for the distro(s) they should be available for.

Also, small nit: stylistically `packages` is a tuple, not list.

pmoravec:
See https://github.com/sosreport/sos/blob/master/sos/policies/redhat.py#L197 and copy&replace relevant code with/around RH_SATELLITE.

> @pmoravec Yeah, we'd like to run this everytime sos is ran. In that case do I also need to modify redhat.py under policies?

Yes. See e.g. https://github.com/sosreport/sos/blob/master/sos/policies/redhat.py#L197 and copy&replace relevant code with/around all RH_SATELLITE references.

rh-dluong:
Hello @pmoravec and @TurboTurtle , I have made the requested changes.

rh-dluong:
Sorry about that, I tried to format it correctly but was guessing at it.  Should be fixed now, and the commit message has been updated.  Thanks @TurboTurtle

PR #1801, opened 2019-09-26
Gather hashes for all py files

bissenbay:
`python` plugin is being added a new option `hashes` which finds all python files and calculates `sha256` checksums. 
It generates a json file like this:
```json
{
   "digests": {
      "filepath": "/usr/lib/python2.7/site-packages/typing.py",
      "sha256": "ce88d334fe5370179b4395c89fcc08bf2c30bc4cd286ab765b3ea7a53e0a73c9"
   }
}
```
The result of this plugin would be used by project Thoth (https://thoth-station.ninja/) for provenance checking purposes.
As part of our project, we analyze most of the ML related packages on PyPI (https://pypi.org) and gather checksums of all files inside wheel files.
Looking at this checksums, we can ensure that a user is using right python packages and files from valid sources. 

Signed-off-by: Bissenbay Dauletbayev <bissenbay@gmail.com>

---

### DIFF
--- a/sos/plugins/python.py
+++ b/sos/plugins/python.py
@@ -10,6 +10,9 @@
 
 from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
 import sos.policies
+import os
+import json
+import hashlib
 
 
 class Python(Plugin, DebianPlugin, UbuntuPlugin):
@@ -21,25 +24,69 @@ class Python(Plugin, DebianPlugin, UbuntuPlugin):
 
     packages = ('python',)
 
+    python_version = "python -V"
+
     def setup(self):
-        self.add_cmd_output("python -V", suggest_filename="python-version")
+        self.add_cmd_output(
+            self.python_version, suggest_filename="python-version"
+        )
         self.add_cmd_output("pip list")
 
 
 class RedHatPython(Python, RedHatPlugin):
 
     packages = ('python', 'python36', 'python2', 'platform-python')
+    option_list = [
+        ('hashes', "gather hashes for all python files", 'slow',
+         False)]
 
     def setup(self):
         self.add_cmd_output(['python2 -V', 'python3 -V'])
-        self.add_cmd_output("pip list")
         if isinstance(self.policy, sos.policies.redhat.RHELPolicy) and \
                 self.policy.dist_version() > 7:
-            self.add_cmd_output(
-                '/usr/libexec/platform-python -V',
-                suggest_filename='python-version'
-            )
-        else:
-            self.add_cmd_output('python -V', suggest_filename='python-version')
+            self.python_version = "/usr/libexec/platform-python -V"
+        super(RedHatPython, self).setup()
+
+        if self.get_option('hashes'):
+            digests = {
+                'digests': []
+            }
+
+            py_paths = [
+                '/usr/lib',
+                '/usr/lib64',
+                '/usr/local/lib',
+                '/usr/local/lib64'
+            ]
+
+            for py_path in py_paths:
+                for root, _, files in os.walk(py_path):
+                    for file_ in files:
+                        filepath = os.path.join(root, file_)
+                        if filepath.endswith('.py'):
+                            try:
+                                with open(filepath, 'rb') as f:
+                                    digest = hashlib.sha256()
+                                    chunk = 1024
+                                    while True:
+                                        data = f.read(chunk)
+                                        if data:
+                                            digest.update(data)
+                                        else:
+                                            break
+
+                                    digest = digest.hexdigest()
+
+                                    digests['digests'].append({
+                                        'filepath': filepath,
+                                        'sha256': digest
+                                    })
+                            except IOError:
+                                self._log_error(
+                                    "Unable to read python file at %s" %
+                                    filepath
+                                )
+
+            self.add_string_as_file(json.dumps(digests), 'digests.json')
 
 # vim: set et ts=4 sw=4 :


sos/plugins/gather_py_hashes.py
```diff
@@ -0,0 +1,39 @@
...
+
+    requires_root = False
+
+    def setup(self):
+        digests = {
+            "digests": []
+        }
+
+        for py_file in glob.glob('/usr/**/lib*/python*/**/*.py',
+                                 recursive=True):
```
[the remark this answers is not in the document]
pmoravec: Also, the `glob.glob` can run *really* long. On a quite fresh RHEL8.0, the call took 42 seconds and found 3.4k files. Assume one deploys many applications and the time is e.g. ten times bigger.

Also there might be security or privacy concerns from customers deploying custom python apps where sosreport will scan all the code and collect hashes.

sos/plugins/gather_py_hashes.py
```diff
@@ -0,0 +1,39 @@
...
+import os
+import json
+import hashlib
+import glob
+
+
+class GatherHashes(Plugin, RedHatPlugin):
+    """Gather hashes for all python files"""
+
+    plugin_name = "gather_py_hashes"
```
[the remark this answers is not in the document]
bmr-cymru: Right: my initial thought was that this should be part of the Python plugin. The scope of (especially default) collection needs more discussion.

sos/plugins/gather_py_hashes.py
```diff
@@ -0,0 +1,39 @@
...
+    requires_root = False
+
+    def setup(self):
+        digests = {
+            "digests": []
+        }
+
+        for py_file in glob.glob('/usr/**/lib*/python*/**/*.py',
+                                 recursive=True):
+            if os.path.isfile(py_file):
```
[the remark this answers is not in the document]
bmr-cymru: You could have a directory name ending in ".py", although it's unlikely in normal usage.

sos/plugins/python.py
```diff
@@ -20,6 +23,9 @@ class Python(Plugin, DebianPlugin, UbuntuPlugin):
     profiles = ('system',)
 
     packages = ('python',)
+    option_list = [
+        ('gather_py_hashes', "gather hashes for all python files", 'slow',
+         False)]
```
TurboTurtle: As written, this will add the `gather_py_hashes` option to the python plugin for all current distributions. However, it will only execute on Red Hat systems. That's not so great.

There are a few options here:
- Pull the `gather_py_hashes` logic out of `setup()` into its own method, which `setup()` then calls, but this would still require duplicate checks in each `*Python` class as currently written.
- Do the above, but have `RedHatPython` call `super(Python, self).setup()` before it does its own collections, and then we'd need to clean up duplicate writes to `python-version`.
- Move the `option_list` definition to `RedHatPython` and have this option only available to Red Hat family systems.

bissenbay: Thanks

bissenbay: @TurboTurtle Should I choose only one of these options? The easiest for me looks the 3rd option. I will go with that and squash.

bissenbay:
@TurboTurtle @bmr-cymru @pmoravec Could you guys please recheck? Thanks a lot!

sos/plugins/python.py
```diff
@@ -10,6 +10,9 @@
 
 from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
 import sos.policies
+import json
+import hashlib
+from pathlib import Path
```
pmoravec: This fails on `python2` by default. Either use a `py2` compatible library, or wait until we drop `py2` support in sosreport.

sos/plugins/python.py
```diff
@@ -42,4 +48,41 @@ def setup(self):
...
+
+                            digests['digests'].append({
+                                'filepath': str(py_file),
+                                'sha256': digest
+                            })
+                    except IOError:
+                        self._log_error(
+                            "Unable to read python file at %s" % py_file)
+
+            self.add_string_as_file(json.dumps(digests), 'digests.json')
```
pmoravec: This will store the `digest.json` inside `sos_strings/python/digests.json` - is that intentional? Usually, plugins collect data either via `add_copy_spec` (to copy directly into sosreport's directory structure) or `add_cmd_output` (to collect output under `sos_commands/python`) - those two locations (esp. the `sos_commands/python`) are the first places where one would search for `python` plugin output.

`add_tring_as_file` sounds a good usage here (you dont collect a file or a cmd output here), just the whole idea of:

- collecting filenames and hashes of *any* custom python scripts
- storing them on bit "unusual" location

this sounds like a Big Brother approach to me.

Btw. isn't it sufficient to e.g.:
- run something like `rpm -V $(rpm -qa | grep python)` to verify no changes in installed python packages (which basically should cover all files you are interested in, am I right?)
- files not reported by the `rpm -V` match original files in the packages (so you can get their hashes from the packages themselves, if needed)
- for files reported by the `rpm -V`, we can run the checksums

This approach sounds less intrusive as it skips custom applications, while it - at the end - provides sufficient information for your purposes.

bissenbay: @pmoravec I could not find a way how I can store the generated .json file into sosreport's directory. Of course, I would like to store it there, however, `add_string_as_file` was the only way I could find to store my .json file.

bissenbay:
@pmoravec @bmr-cymru @TurboTurtle I have some changes according to your feedback. This time I'm using `add_copy_spec` over `add_string_as_file`. The thing is that `add_copy_spec` usually is used to copy a file to the generated `.tar` file. As my `digests` file is an object in memory, I first create a temporary folder and create there a file `digests.json` and write that object in memory to that to the file.
Finally, I copy that file. But, as it is on being created on the temporary folder, that path will persist in the generated `.tar` file as well. So, the path would something like `/tmp/blablabla/digests.json`.

bmr-cymru:
`add_string_as_file()` is the right method to use here (similar to the data gathered by e.g. `unpackaged`). Using a temporary file just to collect it with `add_copy_specs()` is unnecessarily cumbersome.

sos/plugins/python.py
bmr-cymru: This isn't that different from what the `unpackaged` plugin does, and using `add_string_as_file()` is the correct way to store in-memory string data generated by a plugin in the report. It's much less common than `add_copy_spec()`, but that method should only be used to collect actual file system locations for inclusion in the report.

bissenbay:
> `add_string_as_file()` is the right method to use here (similar to the data gathered by e.g. `unpackaged`). Using a temporary file just to collect it with `add_copy_specs()` is unnecessarily cumbersome.

@bmr-cymru Thanks for the feedback. I have returned `add_string_as_file` back. PTAL @TurboTurtle @pmoravec . Thanks!

bissenbay:
@pmoravec @bmr-cymru @TurboTurtle would be like to review? Thanks!

bissenbay:
@pmoravec @bmr-cymru @TurboTurtle Is my code ready to get merged? Thanks!

bmr-cymru:
Couple of points with the current version:

  * The patch is missing a Signed-off-by and the subject needs to be have the `[python]` tag at the start.

  * I think we can call the option just "hashes" - it's easier to type on the command line and there's enough context with the option help message to explain to users what it does.

  * Right now this is only available on the Red Hat version of the plugin. It should be possible to make this available on all distributions with a bit of work: e.g. the base class should set a member named `python_version = "python -V"`, and the Red Hat `setup()` method should modify this to `platform-python` if required, before calling `super(RedHatPython, self)setup()` to do the collection.

sos/plugins/python.py
```diff
@@ -33,9 +35,11 @@ class RedHatPython(Python, RedHatPlugin):
 
     packages = ('python', 'python36', 'python2', 'platform-python')
     option_list = [
-        ('gather_py_hashes', "gather hashes for all python files", 'slow',
+        ('hashes', "gather hashes for all python files", 'slow',
          False)]
 
+    python_version = "platform-python"
```
bissenbay: @bmr-cymru you told to modify it to `platform-python`, do I correctly understand?

sos/plugins/python.py
```diff
@@ -24,6 +24,8 @@ class Python(Plugin, DebianPlugin, UbuntuPlugin):
 
     packages = ('python',)
 
+    python_version = "python -V"
```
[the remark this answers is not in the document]
bmr-cymru: Yes

sos/plugins/python.py
bmr-cymru: This needs to have the same logic as the current plugin (so it needs to be in the `setup()` method):

```python
        if isinstance(self.policy, sos.policies.redhat.RHELPolicy) and \
                self.policy.dist_version() > 7:
            self.python_version = "/usr/libexec/platform-python -V"
        super(RedHatPython, self).setup()
```

Then the main `setup()` method in the `Python` class needs to `self.add_cmd_output(self.python_version, suggest_filename='python-version')`.

That way, the `python-version` file will always use the correct command for the distro it is running on, and the `hashes` option will be available to all distributions.

sos/plugins/python.py
```diff
@@ -21,25 +24,70 @@ class Python(Plugin, DebianPlugin, UbuntuPlugin):
...
+
     def setup(self):
-        self.add_cmd_output("python -V", suggest_filename="python-version")
+        self.add_cmd_output(
+            self.python_version, suggest_filename="python-version"
```
bissenbay: I am using `self.python_version`

sos/plugins/python.py
```diff
@@ -21,25 +24,70 @@ class Python(Plugin, DebianPlugin, UbuntuPlugin):
...
         self.add_cmd_output("pip list")
         if isinstance(self.policy, sos.policies.redhat.RHELPolicy) and \
                 self.policy.dist_version() > 7:
-            self.add_cmd_output(
-                '/usr/libexec/platform-python -V',
-                suggest_filename='python-version'
-            )
-        else:
-            self.add_cmd_output('python -V', suggest_filename='python-version')
+            self.python_version = "/usr/libexec/platform-python -V"
```
bissenbay: I did according to the feedback you gave. I was not sure about lines 38-43 and deleted. Is that fine? Or should I return them back?

sos/plugins/python.py
```diff
@@ -21,25 +24,70 @@ class Python(Plugin, DebianPlugin, UbuntuPlugin):
...
+        super(RedHatPython, self).setup()
```
TurboTurtle: This is the correct approach, but now there's a need to remove the separate call to `pip list` (line 45) in `RedHatPython` since it will be collected in the base `Python.setup()` now. Otherwise we'll get duplicate files.

bissenbay: Done!

bissenbay:
@bmr-cymru @pmoravec PTAL. Thanks!

bissenbay:
@TurboTurtle @bmr-cymru @pmoravec Could you please take a look! Thanks!

goern:
so this is good to merge?! :)

TurboTurtle:
@bissenbay actually, one additional quick change is needed. The commit message needs to explain the option being added. Right now it's a blank commit message with a subject of just "Add hashes option". Checkout the [Contributor Guidelines](https://github.com/sosreport/sos/wiki/Contribution-Guidelines) for an example.

Once that's done we should be good to merge this.

TurboTurtle:
Actually, I can rewrite the commit message during merge since you have the Sign-off line in there.

bissenbay:
@goern @fridex @TurboTurtle @bmr-cymru @pmoravec Thanks!

Issue #1754, opened 2019-08-16
[Plugin] document add_cmd_output() arguments


bmr-cymru:
We currently don't have docstrings for the `add_cmd_output()` arguments:

```Python
    def add_cmd_output(self, cmds, suggest_filename=None,
                       root_symlink=None, timeout=300, stderr=True,
                       chroot=True, runat=None, env=None, binary=False,
                       sizelimit=None, pred=None, subdir=None,
                       allow_changes=False):
        """Run a program or a list of programs and collect the output"""
```

Most are quite obvious by name, but e.g. the new `changes` argument (which means that the running the command may alter system configuration, and to warn the user of this if the conditions for running the command are not met) would benefit from more explanation for plugin authors.

TurboTurtle:
Closing per #1807. The descriptions are in `_collect_cmd_output()` rather than `add_cmd_output()`, but they are present now.

PR #1807, opened 2019-10-01
[Plugin|plugins] Consolidate command collection API methods

TurboTurtle:
Removed `call_ext_prog()`, `check_ext_prog()`, and
`get_command_output()` in favor of consolidating Plugin command
collection into two public methods: `add_cmd_output()` and
`get_cmd_output_now()`.

- `add_cmd_output()` remains unchanged and is the interface to use when
plugins should collect command output during their collect() phase.

- `get_cmd_output_now()` will now return the content of a command's
execution as well as the filename within the sos archive, instead of
just returning the filename. This should help streamline plugin code
that runs a command and then iterates over that output.

The `get_cmd_output_now()` has a `save` keyword arg (default True), that
if set to False will skip writing the command output to the archive.

Also updates the plugins to no longer use `get_command_output()`,
`call_ext_prog()`, or `check_ext_prog()` - instead only using
`get_cmd_output_now()`.

For the vast majority of plugins, these calls were able to be changed
in-place with little to no further alteration of plugin code outside of
no longer making indenpendent open() calls to files created, as the
content of those files is now directly available. Exceptions are noted
below.

For the dlm, gfs2, and kvm plugins the only use of get_command_output()
was to mount /sys/kernel/debugfs which in modern distributions is no
longer needed, so those plugins have had that functionality entirely
removed instead of being ported to the new `get_cmd_output_now()`.

For the networking plugin, many of the checks are handled by
SoSPredicates, and so those command collections now use Predicates
rather than being gated by individual `get_cmd_output_now()` calls.

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -519,7 +519,7 @@ class Plugin(object):
                     self._log_warn("Cannot apply regex substitution to binary"
                                    " output: '%s'" % called['exe'])
                     continue
-                if fnmatch.fnmatch(called['exe'], globstr):
+                if fnmatch.fnmatch(called['cmd'], globstr):
                     path = os.path.join(self.commons['cmddir'], called['file'])
                     self._log_debug("applying substitution to '%s'" % path)
                     readable = self.archive.open_file(path)
@@ -957,57 +957,6 @@ class Plugin(object):
                     # size limit not hit, copy the file
                     self._add_copy_paths([_file])
 
-    def get_command_output(self, prog, timeout=300, stderr=True,
-                           chroot=True, runat=None, env=None,
-                           binary=False, sizelimit=None):
-        if self._timeout_hit:
-            return
-
-        if chroot or self.commons['cmdlineopts'].chroot == 'always':
-            root = self.sysroot
-        else:
-            root = None
-
-        result = sos_get_command_output(prog, timeout=timeout, stderr=stderr,
-                                        chroot=root, chdir=runat,
-                                        env=env, binary=binary,
-                                        sizelimit=sizelimit,
-                                        poller=self.check_timeout)
-
-        if result['status'] == 124:
-            self._log_warn("command '%s' timed out after %ds"
-                           % (prog, timeout))
-
-        # command not found or not runnable
-        if result['status'] == 126 or result['status'] == 127:
-            # automatically retry chroot'ed commands in the host namespace
-            if root and root != '/':
-                if self.commons['cmdlineopts'].chroot != 'always':
-                    self._log_info("command '%s' not found in %s - "
-                                   "re-trying in host root"
-                                   % (prog.split()[0], root))
-                    return self.get_command_output(prog, timeout=timeout,
-                                                   chroot=False, runat=runat,
-                                                   env=env,
-                                                   binary=binary)
-            self._log_debug("could not run '%s': command not found" % prog)
-        return result
-
-    def call_ext_prog(self, prog, timeout=300, stderr=True,
-                      chroot=True, runat=None):
-        """Execute a command independantly of the output gathering part of
-        sosreport.
-        """
-        return self.get_command_output(prog, timeout=timeout, stderr=stderr,
-                                       chroot=chroot, runat=runat)
-
-    def check_ext_prog(self, prog):
-        """Execute a command independently of the output gathering part of
-        sosreport and check the return code. Return True for a return code of 0
-        and False otherwise.
-        """
-        return self.call_ext_prog(prog)['status'] == 0
-
     def _add_cmd_output(self, **kwargs):
         """Internal helper to add a single command to the collection list."""
         pred = kwargs.pop('pred') if 'pred' in kwargs else None
@@ -1127,23 +1076,73 @@ class Plugin(object):
         self.copy_strings.append((content, filename))
         self._log_debug("added string ...'%s' as '%s'" % (summary, filename))
 
-    def _get_cmd_output_now(self, cmd, suggest_filename=None,
-                            root_symlink=False, timeout=300, stderr=True,
-                            chroot=True, runat=None, env=None,
-                            binary=False, sizelimit=None, subdir=None,
-                            changes=False):
+    def _collect_cmd_output_now(self, cmd, suggest_filename=None,
+                                root_symlink=False, timeout=300, stderr=True,
+                                chroot=True, runat=None, env=None,
+                                binary=False, sizelimit=None, subdir=None,
+                                changes=False):
         """Execute a command and save the output to a file for inclusion in the
         report.
+
+        Positional Arguments:
+            :param cmd:                 The command to run
+
+        Keyword Arguments:
+            :param suggest_filename:    Filename to use when writing to the
+                                        archive
+            :param root_symlink:        Create a symlink in the archive root
+            :param timeout:             Time in seconds to allow a cmd to run
+            :param stderr:              Write stderr to stdout?
+            :param chroot:              Perform chroot before running cmd?
+            :param runat:               Run the command from this location,
+                                        overriding chroot
+            :param env:                 Dict of env vars to set for the cmd
+            :param binary:              Is the output in binary?
+            :param sizelimit:           Maximum size in MB of output to save
+            :param subdir:              Subdir in plugin directory to save to
+            :param changes:             Does this cmd potentially make a change
+                                        on the system?
+
+        :returns:       dict containing status, output, and filename in the
+                        archive for the executed cmd
+
         """
         if self._timeout_hit:
             return
 
+        if chroot or self.commons['cmdlineopts'].chroot == 'always':
+            root = self.sysroot
+        else:
+            root = None
+
         start = time()
 
-        result = self.get_command_output(cmd, timeout=timeout, stderr=stderr,
-                                         chroot=chroot, runat=runat,
-                                         env=env, binary=binary,
-                                         sizelimit=sizelimit)
+        result = sos_get_command_output(
+            cmd, timeout=timeout, stderr=stderr, chroot=root,
+            chdir=runat, env=env, binary=binary, sizelimit=sizelimit,
+            poller=self.check_timeout
+        )
+
+        if result['status'] == 124:
+            self._log_warn(
+                "command '%s' timed out after %ds" % (cmd, timeout)
+            )
+
+        # command not found or not runnable
+        if result['status'] == 126 or result['status'] == 127:
+            # automatically retry chroot'ed commands in the host namespace
+            if root and root != '/':
+                if self.commons['cmdlineopts'].chroot != 'always':
+                    self._log_info("command '%s' not found in %s - "
+                                   "re-trying in host root"
+                                   % (cmd.split()[0], root))
+                    result = sos_get_command_output(
+                        cmd, timeout=timeout, chroot=False, chdir=runat,
+                        env=env, binary=binary, sizelimit=sizelimit,
+                        poller=self.check_timeout
+                    )
+            self._log_debug("could not run '%s': command not found" % cmd)
+
         self._log_debug("collected output of '%s' in %s (changes=%s)"
                         % (cmd.split()[0], time() - start, changes))
 
@@ -1153,6 +1152,7 @@ class Plugin(object):
             outfn = self._make_command_filename(cmd, subdir)
 
         outfn_strip = outfn[len(self.commons['cmddir'])+1:]
+
         if binary:
             self.archive.add_binary(result['output'], outfn)
         else:
@@ -1161,29 +1161,59 @@ class Plugin(object):
             self.archive.add_link(outfn, root_symlink)
 
         # save info for later
-        self.executed_commands.append({'exe': cmd, 'file': outfn_strip,
+        self.executed_commands.append({'cmd': cmd, 'file': outfn_strip,
                                        'binary': 'yes' if binary else 'no'})
 
-        return os.path.join(self.archive.get_archive_path(), outfn)
+        result['filename'] = (
+            os.path.join(self.archive.get_archive_path(), outfn) if outfn else
+            ''
+        )
+        return result
 
-    def get_cmd_output_now(self, exe, suggest_filename=None,
+    def collect_cmd_output(self, cmd, suggest_filename=None,
                            root_symlink=False, timeout=300, stderr=True,
                            chroot=True, runat=None, env=None,
-                           binary=False, sizelimit=None, pred=None):
+                           binary=False, sizelimit=None, pred=None,
+                           subdir=None):
         """Execute a command and save the output to a file for inclusion in the
         report.
         """
         if not self.test_predicate(cmd=True, pred=pred):
             self._log_info("skipped cmd output '%s' due to predicate (%s)" %
-                           (exe, self.get_predicate(cmd=True, pred=pred)))
-            return None
+                           (cmd, self.get_predicate(cmd=True, pred=pred)))
+            return {
+                'status': None,  # don't match on if result['status'] checks
+                'output': '',
+                'filename': ''
+            }
 
-        return self._get_cmd_output_now(exe, suggest_filename=suggest_filename,
-                                        root_symlink=root_symlink,
-                                        timeout=timeout, stderr=stderr,
-                                        chroot=chroot, runat=runat,
-                                        env=env, binary=binary,
-                                        sizelimit=sizelimit)
+        return self._collect_cmd_output_now(
+            cmd, suggest_filename=suggest_filename, root_symlink=root_symlink,
+            timeout=timeout, stderr=stderr, chroot=chroot, runat=runat,
+            env=env, binary=binary, sizelimit=sizelimit, subdir=subdir
+        )
+
+    def exec_cmd(self, cmd, timeout=300, stderr=True, chroot=True, runat=None,
+                 env=None, binary=False, pred=None):
+        """Execute a command right now and return the output and status, but
+        do not save the output within the archive.
+
+        Use this method in a plugin's setup() if command output is needed to
+        build subsequent commands added to a report via add_cmd_output().
+        """
+        if not self.test_predicate(cmd=True, pred=pred):
+            return {
+                'status': None,
+                'output': ''
+            }
+
+        if chroot or self.commons['cmdlineopts'].chroot == 'always':
+            root = self.sysroot
+        else:
+            root = None
+
+        return sos_get_command_output(cmd, timeout=timeout, chroot=root,
+                                      chdir=runat, binary=binary, env=env)
 
     def is_module_loaded(self, module_name):
         """Return whether specified moudle as module_name is loaded or not"""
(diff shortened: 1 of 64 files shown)


TurboTurtle:
This is a continuation from the conversation in #1707.

Also, it looks like this should address #1754 as a bonus, as `add_cmd_output()` eventually calls `_get_cmd_output_now()` which is where the docstring has been expanded.

sos/plugins/abrt.py
```diff
@@ -24,16 +24,15 @@ class Abrt(Plugin, RedHatPlugin):
...
-            if line.startswith("Directory:"):
-                self.add_cmd_output("abrt-cli info -d '%s'" % line.split()[1])
-
     def setup(self):
         self.add_cmd_output("abrt-cli status")
-        list_file = self.get_cmd_output_now("abrt-cli list")
-        if self.get_option('detailed') and list_file:
-            self.info_detailed(list_file)
+        abrt_list = self.get_cmd_output_now("abrt-cli list")
+        if self.get_option("detailed") and abrt_list:
```
pmoravec: Technically, abrt_list (as a return value of `get_cmd_output_now`) will always be a dict, so `and abrt_list` will always pass. We have few options here:

- test for `abrt_list["status"] == 0`
- test for `abrt_list["output"]` - that is bit redundant due to the further iteration over output splitlines
- skip this test

I am inclined to skip the test.

sos/plugins/block.py
```diff
@@ -69,10 +53,12 @@ def setup(self):
                     "fdisk -l %s" % disk_path
                 ])
 
-        lsblk_file = self.get_cmd_output_now("lsblk -f -a -l")
+        lsblk = self.get_cmd_output_now("lsblk -f -a -l")
         # for LUKS devices, collect cryptsetup luksDump
-        if lsblk_file:
-            for dev in self.get_luks_devices(lsblk_file):
-                self.add_cmd_output('cryptsetup luksDump /dev/%s' % dev)
+        if lsblk['status'] == 0:
```
pmoravec: See my comment in `abrt` - ideally, we shall be coherent here (no strong prferences, either testing for zero status or skipping such a test is fine to me).

sos/plugins/crio.py
```diff
@@ -84,7 +84,7 @@ def setup(self):
 
     def _get_crio_list(self, cmd):
         ret = []
-        result = self.get_command_output(cmd)
+        result = self.get_cmd_output_now(cmd, save=False)
         if result['status'] == 0:
```
pmoravec: ditto here (test status or skip the test)

sos/plugins/dlm.py
```diff
@@ -37,36 +33,17 @@ def setup(self):
...
-    def postproc(self):
-        if self._debugfs_cleanup and os.path.ismount(self.debugfs_path):
-            r = self.call_ext_prog("umount %s" % self.debugfs_path)
-            if r["status"] != 0:
-                self._log_error("could not unmount %s" % self.debugfs_path)
-
-        return
+        dlm_tool = "dlm_tool ls"
+        result = self.get_cmd_output_now(dlm_tool)
+        if result["status"] != 0:
```
pmoravec: ditto here (test status or skip the test)

sos/plugins/dnf.py
```diff
@@ -57,11 +57,11 @@ def setup(self):
...
 
-        if self.get_option("history"):
+        if self.get_option("history") and not self.get_option("history-info"):
             self.add_cmd_output("dnf history")
 
         if self.get_option("history-info"):
-            history = self.call_ext_prog("dnf history")
+            history = self.get_cmd_output_now("dnf history")
             transactions = -1
             if history['output']:
```
pmoravec: Although it is not directly relevant to the change in API, could we drop this test? IMHO iterating over (potentially empty) splitlines is sufficient..

sos/plugins/docker.py
```diff
@@ -104,7 +103,7 @@ def setup(self):
 
     def _get_docker_list(self, cmd):
         ret = []
-        result = self.get_command_output(cmd)
+        result = self.get_cmd_output_now(cmd, save=False)
         if result['status'] == 0:
```
pmoravec: ditto here (test for status or no test)

sos/plugins/gfs2.py
```diff
@@ -36,24 +32,6 @@ def setup(self):
         ])
 
         if self.get_option("gfs2lockdump"):
-            if self._mount_debug():
```
pmoravec: Wasn't there some rationale to collect `/sys/kernel/debug/gfs2/*` _only_ under some further condition (i.e. `_mount_debug`)?

sos/plugins/gfs2.py
```diff
@@ -36,24 +32,6 @@ def setup(self):
...
-            r = self.call_ext_prog("mount -t debugfs debugfs %s"
-                                   % self.debugfs_path)
-            if r["status"] != 0:
-                self._log_error("debugfs not mounted and mount attempt failed")
-                self._debugfs_cleanup = False
-        return os.path.ismount(self.debugfs_path)
-
-    def postproc(self):
-        if self._debugfs_cleanup and os.path.ismount(self.debugfs_path):
-            r = self.call_ext_prog("umount %s" % self.debugfs_path)
```
pmoravec: Is it really ok not to umount (something we probably mounted)?

sos/plugins/gluster.py
```diff
@@ -108,14 +94,17 @@ def setup(self):
             else:
                 self.soslog.warn("Unable to generate statedumps, no such "
                                  "directory: %s" % self.statedump_dir)
-            state = self.get_command_output("gluster get-state")
+            state = self.get_cmd_output_now("gluster get-state")
             if state['status'] == 0:
```
pmoravec: ditto here

sos/plugins/gluster.py
```diff
@@ -108,14 +94,17 @@ def setup(self):
...
                 state_file = state['output'].split()[-1]
                 self.add_copy_spec(state_file)
 
-        volume_file = self.get_cmd_output_now("gluster volume info")
-        if volume_file:
-            for volname in self.get_volume_names(volume_file):
+        volume_cmd = self.get_cmd_output_now("gluster volume info")
+        if volume_cmd['status'] == 0:
```
pmoravec: ditto here

sos/plugins/kata_containers.py
```diff
@@ -37,7 +37,7 @@ def setup(self):
 
         # obtain a list of config files by asking the runtime
         cmd = 'kata-runtime --kata-show-default-config-paths'
-        configs = self.get_command_output(cmd)
+        configs = self.get_cmd_output_now(cmd)
         if configs and configs['status']:
```
pmoravec: "`configs`" is never evaluated to `False`.

sos/plugins/kvm.py
```diff
@@ -33,19 +30,7 @@ def setup(self):
             "/sys/module/kvm_amd/srcversion",
             "/sys/module/ksm/srcversion"
         ])
-        if not os.path.ismount(self.debugfs_path):
```
pmoravec: Same comments like to `gfs2` plugin - is it OK in skipping this test?

sos/plugins/openstack_glance.py
```diff
@@ -76,7 +72,10 @@ def setup(self):
 
     def running_in_container(self):
         for runtime in ["docker", "podman"]:
-            container_status = self.get_command_output(runtime + " ps")
+            container_status = self.get_cmd_output(
```
pmoravec: to call `self.get_cmd_output_now` instead? :)

sos/plugins/pcp.py
```diff
@@ -148,10 +148,8 @@ def setup(self):
             os.path.join(self.pcp_log_dir, '*/*/config*')
         ])
 
-        # Need to get the current status of the PCP infrastructure
-        self.add_cmd_output("pcp")
         # Collect a summary for the current day
-        res = self.get_command_output('pcp')
+        res = self.get_cmd_output_now('pcp')
```
pmoravec: with `save=False`?

sos/plugins/sas3ircu.py
```diff
@@ -23,11 +23,9 @@ class SAS3ircu(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
     def setup(self):
 
         # get list of adapters
-        result = self.call_ext_prog("sas3ircu list", timeout=5)
+        result = self.get_cmd_output_now("sas3ircu list", timeout=5)
```
pmoravec: with `save=False`?

sos/plugins/selinux.py
```diff
@@ -27,7 +27,7 @@ def setup(self):
         ])
         self.add_cmd_output('sestatus')
 
-        state = self.get_command_output('getenforce')['output']
+        state = self.get_cmd_output_now('getenforce')['output']
```
pmoravec: We will newly collect/store output of this command - a duplicate info resent in other cmd output(s), but it simplifies to find the info. Not sure if worth having `save=False` or not.

sos/plugins/teamd.py
```diff
@@ -21,7 +21,7 @@ class Teamd(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
 
     def _get_team_interfaces(self):
         teams = []
-        ip_result = self.get_command_output("ip -o link")
+        ip_result = self.get_cmd_output_now("ip -o link")
```
pmoravec: with `save=False`?

sos/plugins/vdsm.py
```diff
@@ -114,7 +114,7 @@ def setup(self):
         ])
 
         try:
-            res = self.call_ext_prog(
+            res = self.get_cmd_output_now(
```
pmoravec: with `save=False`?

sos/plugins/vdsm.py
```diff
@@ -130,7 +130,7 @@ def setup(self):
             )
 
         try:
-            res = self.call_ext_prog('vdsm-client Host getStorageDomains')
+            res = self.get_cmd_output_now('vdsm-client Host getStorageDomains')
```
pmoravec: with `save=False`?

sos/plugins/vhostmd.py
```diff
@@ -20,7 +20,7 @@ class vhostmd(Plugin, RedHatPlugin):
     packages = ['virt-what']
 
     def setup(self):
-        vw = self.get_command_output("virt-what")['output'].splitlines()
+        vw = self.get_cmd_output_now("virt-what")['output'].splitlines()
```
pmoravec: with `save=False`?

sos/plugins/virsh.py
```diff
@@ -43,8 +43,7 @@ def setup(self):
 
         # get network, pool and nwfilter elements
         for k in ['net', 'nwfilter', 'pool']:
-            self.add_cmd_output('%s %s-list' % (cmd, k))
-            k_list = self.get_command_output('%s %s-list' % (cmd, k))
+            k_list = self.get_cmd_output_now('%s %s-list' % (cmd, k))
```
pmoravec: with `save=False`?

sos/plugins/virsh.py
```diff
@@ -55,7 +54,7 @@ def setup(self):
 
         # cycle through the VMs/domains list, ignore 2 header lines and latest
         # empty line, and dumpxml domain name in 2nd column
-        domains_output = self.get_command_output('%s list --all' % cmd)
+        domains_output = self.get_cmd_output_now('%s list --all' % cmd)
```
pmoravec: with `save=False`?

sos/plugins/xen.py
```diff
@@ -37,7 +42,7 @@ def check_enabled(self):
         return (self.determine_xen_host() == "baremetal")
 
     def is_running_xenstored(self):
-        xs_pid = self.call_ext_prog("pidof xenstored")['output']
+        xs_pid = self.get_cmd_output_now("pidof xenstored")['output']
```
pmoravec: with `save=False`?

sos/plugins/yum.py
```diff
@@ -99,7 +99,7 @@ def setup(self):
             # RHEL6+ alternative for this whole function:
             # self.add_cmd_output("yum-debug-dump '%s'"
             # % os.path.join(self.commons['dstroot'],"yum-debug-dump"))
-            r = self.call_ext_prog("yum-debug-dump")
+            r = self.get_cmd_output_now("yum-debug-dump")
```
pmoravec: with `save=False`?

sos/plugins/zfs.py
```diff
@@ -27,7 +27,7 @@ def setup(self):
             "zpool status -vx"
         ])
 
-        zpools = self.call_ext_prog("zpool list -H -o name")
+        zpools = self.get_cmd_output_now("zpool list -H -o name")
```
pmoravec: with `save=False`? (though this might be interesting to have..? (dont know zfs..))

sos/plugins/__init__.py
```diff
@@ -1112,59 +1061,123 @@ def _get_cmd_output_now(self, cmd, suggest_filename=None,
...
-                                         sizelimit=sizelimit)
+        result = sos_get_command_output(
+            cmd, timeout=timeout, stderr=stderr, chroot=root,
+            chdir=runat, env=env, binary=binary, sizelimit=sizelimit,
+            poller=self.check_timeout
+        )
+
+        if result['status'] == 124:
+            self._log_warn(
+                "command '%s' timed out after %ds" % (prog, timeout)
```
pmoravec: copy&paste typo:  s/prog/cmd

pmoravec:
In general, I like the consolidation. Good work.

We are not coherent in calling `if something["status"] == 0:` - usually such a test is redundant (but dont harm), but in few cases it is afaik really needed. Not sure if it makes sense to unify the approach here.

Several other plugins needs to be updated as well - see their backtraces (usually due to different return value of `add_cmd_output` (not file name but dict)):

    composer-plugin-errors.txt
      File "/root/sos-master-1807/sos/sosreport.py", line 964, in setup
        plug.setup()
      File "/root/sos-master-1807/sos/plugins/composer.py", line 30, in setup
        blueprints = self._get_entries("composer-cli blueprints list")
      File "/root/sos-master-1807/sos/plugins/composer.py", line 17, in _get_entries
        with open(ent_file, "r") as ents:
    TypeError: coercing to Unicode: need string or buffer, dict found
    
    kpatch-plugin-errors.txt
      File "/root/sos-master-1807/sos/sosreport.py", line 964, in setup
        plug.setup()
      File "/root/sos-master-1807/sos/plugins/kpatch.py", line 27, in setup
        kpatches = open(kpatch_list, "r").read().splitlines()
    TypeError: coercing to Unicode: need string or buffer, dict found
    
    openstack_keystone-plugin-errors.txt
      File "/root/sos-master-1807/sos/sosreport.py", line 964, in setup
        plug.setup()
      File "/root/sos-master-1807/sos/plugins/openstack_keystone.py", line 117, in setup
        super(RedHatKeystone, self).setup()
      File "/root/sos-master-1807/sos/plugins/openstack_keystone.py", line 55, in setup
        not(os.path.isdir(self.domain_config_dir)):
      File "/usr/lib64/python2.7/genericpath.py", line 41, in isdir
        st = os.stat(s)
    TypeError: coercing to Unicode: need string or buffer, dict found
    
      File "/root/sos-master-1807/sos/sosreport.py", line 1167, in postproc
        plug.postproc()
      File "/root/sos-master-1807/sos/plugins/openstack_keystone.py", line 100, in postproc
        r"((?m)^\s*(%s)\s*=\s*)(.*)", r"\1********")
      File "/root/sos-master-1807/sos/plugins/__init__.py", line 595, in do_path_regex_sub
        pathexp = re.compile(pathexp)
      File "/usr/lib64/python2.7/re.py", line 190, in compile
        return _compile(pattern, flags)
      File "/usr/lib64/python2.7/re.py", line 229, in _compile
        p = _cache.get(cachekey)
    TypeError: unhashable type: 'dict'
    
    pcp-plugin-errors.txt
      File "/root/sos-master-1807/sos/sosreport.py", line 964, in setup
        plug.setup()
      File "/root/sos-master-1807/sos/plugins/pcp.py", line 132, in setup
        for line in open(pmlogger_ls).read().splitlines():
    TypeError: coercing to Unicode: need string or buffer, dict found
    
    sapnw-plugin-errors.txt
      File "/root/sos-master-1807/sos/sosreport.py", line 964, in setup
        plug.setup()
      File "/root/sos-master-1807/sos/plugins/sapnw.py", line 134, in setup
        self.collect_list_instances()
      File "/root/sos-master-1807/sos/plugins/sapnw.py", line 40, in collect_list_instances
        p = open(inst_out, "r").read().splitlines()
    TypeError: coercing to Unicode: need string or buffer, dict found

sos/plugins/abrt.py
TurboTurtle: Yep, oversight on my part - should be `abrt_list['status'] == 0` here.

sos/plugins/block.py
TurboTurtle: I'm definitely in the "always check status == 0" camp.

We have no other way to gate against the iterations that follow from executing against error output when the command fails for one reason or another.

sos/plugins/dnf.py
TurboTurtle: Fair point, will update

sos/plugins/openstack_glance.py
TurboTurtle: Typo, yep - thanks for catching that.

sos/plugins/pcp.py
TurboTurtle: We previously ran `add_cmd_output('pcp')`, which would be redundant with the new approach of `get_cmd_output_now()`. So  `save=False` would be a regression here.

sos/plugins/vdsm.py
TurboTurtle: Collecting the connected storage pools output would be beneficial to RH support teams

sos/plugins/vdsm.py
TurboTurtle: Same here, this is useful data for support teams

sos/plugins/virsh.py
TurboTurtle: Same as `pcp` - without `save=False` we "automatically" get the `add_cmd_output()` call from above.

sos/plugins/virsh.py
TurboTurtle: Same here

sos/plugins/xen.py
TurboTurtle: Ack

sos/plugins/yum.py
TurboTurtle: Ack

TurboTurtle:
After talking with Bryn a bit further, I've updated this PR to also include a rename of the `get_cmd_output_now()` methods to `collect_cmd_output()`, which `add_cmd_output()` still leverages.

Also, dropped the `save` kwarg in favor of a replacement `exec_cmd()` method to replace the `*_ext_prog()` methods.


Looking at the LGTM analysis, I'm not sure where the build failure is coming from.

PR #1810, opened 2019-10-02
[login] new plugin based in last

### LINKED ISSUE #1808: Including 'login.defs' in default sosreport might help for security compliance checks
login.defs is not included by default in sosreport so most of them are missing it and it's useful to check security compliance configuration

mikelolasagasti:
login.defs is required by some security compliance checks and could
be useful to collect it. Added to system plugin, as seems the most
neutral plugin to collect it.

Closes #1808

---

### DIFF
--- a/sos/plugins/login.py
+++ b/sos/plugins/login.py
@@ -9,12 +9,12 @@
 from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
 
 
-class Last(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
+class Login(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
     """login information
     """
 
-    plugin_name = 'last'
-    profiles = ('system',)
+    plugin_name = 'login'
+    profiles = ('system', 'identity')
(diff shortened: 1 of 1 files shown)


bmr-cymru:
I agree we should collect this, but I don't think the `system` plugin is the right place for it - since we cleaned that plugin up a few years ago it's intended to be just low-level common systemwide stuff: primarily, the content of `/proc/sys`,`glibc` configuration and the two common configuration paths `/etc/default` and `/etc/sysconfig`.

I think we'd want whatever plugin collects `login.defs` to be included in the `identity` profile, which `system` isn't a part of - and the other data it collects doesn't really seem to fit in that profile.

It might be best to merge this with the `last` plugin (and add that to the `identity` profile, which it should be a part of anyway), to create a new `login` plugin. The files are provided by the same package on distributions that use `shadow-utils` so there's a good fit in terms of package triggers.

mikelolasagasti:
Thanks for the review @bmr-cymru, I was not aware of the cleaning done for `system` plugin. I moved `last` to `login`, added `identity` profile and added files `logins.defs`and `default/useradd` files from `shadow-utils`.

PR #1816, opened 2019-10-04
Handle OVN new package naming

danalsan:
Complementing this commit [0], this PR aims to capture the ovn2.* packages after the split of OVN from OVS. Since version 2.10, the new package names for both openvswitch and ovn are openvswitch2.x and ovn2.x in Red Hat Fast Datapath.

[0] https://github.com/sosreport/sos/commit/255c0b08321e9d81f69798d3df16c7548bd3c0db

### DIFF
--- a/sos/plugins/ovn_central.py
+++ b/sos/plugins/ovn_central.py
@@ -139,7 +139,7 @@ class OVNCentral(Plugin):
 
 class RedHatOVNCentral(OVNCentral, RedHatPlugin):
 
-    packages = ('openvswitch-ovn-central', )
+    packages = ('openvswitch-ovn-central', 'ovn2.*-central', )
 
 
 class DebianOVNCentral(OVNCentral, DebianPlugin, UbuntuPlugin):
--- a/sos/plugins/ovn_host.py
+++ b/sos/plugins/ovn_host.py
@@ -50,7 +50,7 @@ class OVNHost(Plugin):
 
 class RedHatOVNHost(OVNHost, RedHatPlugin):
 
-    packages = ('openvswitch-ovn-host', )
+    packages = ('openvswitch-ovn-host', 'ovn2.*-host', )
 
 
 class DebianOVNHost(OVNHost, DebianPlugin, UbuntuPlugin):


bmr-cymru:
The patch for OVN Host adds trailing whitespace to the packages tuple, so there's a Travis error, but for both of these we don't currently support regex notation in the packages list (only in `verify_packages`, to allow filtering out things like -debuginfo packages) - you'll need to pick one or more "core" packages to act as the trigger (alternately we could extend this to do a pattern match but I don't think it's worth it for this case).

danalsan:
> verify_packages

Oh sorry about the trailing space (/me ashamed).
About the regex, I tried it on a RHEL 7.7 and it worked but maybe it was by accident.
Why is it different from https://github.com/sosreport/sos/commit/255c0b08321e9d81f69798d3df16c7548bd3c0db#diff-0a9522d996653ba1be2e296998f68ea1R161 ?

Thanks!
daniel

bmr-cymru:
Actually I was mistaken - although the use of patterns here isn't entirely intended (and doesn't use regex notation). The `Plugin.is_installed()` method uses `PackageManager.pkg_by_name()`, which is supposed to return a single package matching `name`, and ends up calling `PackageManager.all_pkgs_by_name()`. This method uses `fnmatch.filter()` to obtain a result, which allows shell style wildcards (glob notation).

This is something we should look at, so I'll file an issue - for simple checks in plugin triggers it's probably OK, but it can cause some confusing results when there are multiple packages installed (since we end up just arbitrarily picking the last matching package name to return), for e.g.:

```python
Python 2.7.15 (default, May  9 2018, 11:32:33) 
[GCC 7.3.1 20180130 (Red Hat 7.3.1-2)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import sos.policies
>>> sos.policies.load()
<sos.policies.redhat.FedoraPolicy object at 0x7fa37f0b0650>
>>> p = sos.policies.load()
>>> p.package_manager
<sos.policies.PackageManager object at 0x7fa37f0b0550>
>>> p.package_manager.pkg_by_name("glibc.*")
>>> p.package_manager.pkg_by_name("glibc*")
u'glibc-langpack-en'
>>> p.package_manager.pkg_by_name("glibc")
u'glibc'
>>> p.package_manager.pkg_by_name("glibc*")
u'glibc-langpack-en'
```

The "glibc.*" pattern matches nothing, since `.` is not a shell wildcard character (and Red Hat `glibc` package names don't include a dot here). Just "glibc" returns the main package as you might expect, but "glibc*" ends up returning the `glibc-langpack-en` package since:

```python
>>> p.package_manager.all_pkgs_by_name("glibc*")
[u'glibc-devel', u'glibc-all-langpacks', u'glibc', u'glibc-common', u'glibc-headers', u'glibc-langpack-en']
```

Which is a little confusing... I'd still suggest aiming for exact package name matches in the plugin `packages` list, as we may revisit this behaviour in the future, but it is allowing pattern matches for now.

danalsan:
You'd like me to change the patch to have:
ovn2.10-central, ovn2.11-central, ovn2.12-central ? Problem is that we'll need to patch the code every 6 months.
I cannot match on ovn2.* as right now the openvswitch plugin is doing as we still need to differentiate for 'host' and 'central'.

bmr-cymru:
It's unfortunate that the OVN packages include the version string in the name. It's generally something that Red Hat distros have avoided because of problems like this.

We can keep the pattern matching behaviour in the packages list check (and use it here), but we should document what it does and address the problem with `pkg_by_name()` that I mentioned earlier (although we don't need to do that in this pull request - that can happen in #1817 ).

The current branch still needs a fix for the `'ovn2.*-host' )` whitespace.

danalsan:
Hey @pmoravec are you fine with this change? Thanks a lot!

TurboTurtle:
Merged in 7eb369e

Issue #1714, opened 2019-06-30
[general] sosreport collects some logs three times


pmoravec:
The same log entry can be collected by:

- collecting `/var/log/messages*`
- calling `self.add_journal(boot="this", catalog=True)`
- calling `add_journal` by some particular plugin

That means sosreport archive can have the same data stored up to three times.

I was thinking if we can limit this in some way, but:
- we need to copy `messages*` files for legacy reasons
- collecting whole journal is meaningful and imho we shall not skip collecting journal entries we collect in different plugins separately - we would lose chronological ordering of all events
- collecting separate journal units is welcomed

The only way we can improve is collecting `/var/log/messages*` only when not collecting journal logs?

TurboTurtle:
I don't think we can skip collecting `/var/log/messages` if we collect the journal because a number of components still log to that file and don't log to the journal. Granted, a good percentage of those are probably things that sos doesn't have plugins for, but having the messages log can help point support engineers to the fact that there is an issue in, for example, a third party/in-house application rather than with the OS or one of the components that OS ships/supports.

With our size limiting in place, I don't think this is too much of an issue any more with duplicate logs. If a component is spamming a certain log file (such as messages) to the point of making it basically useless, we have the other collection points (i.e. the whole journal and unit journals with larger default sizes) to fall back on without needing user intervention. I think the place this is really an issue is when `--all-logs` is used, but at that point the user is literally telling us "give me everything and size-limiting protections be damned" - so we're giving users exactly what they're asking for; everything on the system, which includes duplicated info that lives on the host anyways.

TurboTurtle:
Closing this as I believe it is handled by recent changes to the `logs` plugin and how we collect the journal and/or `/var/log/messages`

PR #1822, opened 2019-10-08
[logs] Substancial rewrite ending RH/U/D differences

BryanQuigley:
--all-logs totally broke due to extra s in get_options, this
resulted in me trying to fix it, but it's simply grown too
complicated IMHO. We're duplicating functionality differently
in the RH and Ubuntu/Debian sections.

This is my try at a rewrite that I hope will be good for all:
Dropped the log_days option as it's confusing with since.
Use journal if persistant, otherwise don't collect it.
Stop collecting current and last boot journals in verbose.
Default to 7 days since for journal if --all-logs is unspecified.

Also implement some basic "did it error" travis checks for the above.

Resolves: #1714

---


sos/plugins/logs.py
```diff
@@ -11,26 +11,16 @@
...
-        ])
-
-        self.add_copy_spec("/var/log/boot.log")
-        self.add_journal(boot="this", catalog=True)
-        self.add_journal(boot="last", catalog=True)
-        self.add_cmd_output("journalctl --disk-usage")
-
         confs = ['/etc/syslog.conf', '/etc/rsyslog.conf']
         logs = []
+        since = "-7days"
```
TurboTurtle: From a RH support perspective, this is a non-starter if we're also dropping the other journal collections. We get cases daily that require going farther back than 7 days, not to mention the cases of "we had an issue like 10 days ago, can you look at this for an RCA now?".

sos/plugins/logs.py
```diff
@@ -50,83 +40,31 @@ def setup(self):
             if os.path.isfile(i):
                 self.add_copy_spec(i)
 
-        if self.get_option('all_logs'):
-            self.add_journal(boot="this", since=self.get_option("since"),
-                             allfields=True, output="verbose")
-            self.add_journal(boot="last", allfields=True, output="verbose")
```
TurboTurtle: RH engineering sometimes needs the verbose journals. We moved this to an optional collection in #1225 as a result of requests to not make verbose journals a default collection, but I don't agree with removing the ability to collect them entirely.

sos/plugins/logs.py
```diff
@@ -11,26 +11,16 @@
...
     def setup(self):
-        self.add_copy_spec([
-            "/etc/syslog.conf",
-            "/etc/rsyslog.conf",
-            "/etc/rsyslog.d"
-        ])
-
-        self.add_copy_spec("/var/log/boot.log")
-        self.add_journal(boot="this", catalog=True)
-        self.add_journal(boot="last", catalog=True)
```
TurboTurtle: These collections are very important to several of our support teams.

BryanQuigley:
The big reason I had to kill the this/last is it's inconsistent with --since. We either want to be able to collect logs since a date or we want it to be based on boots (and before this commit it was implemented sometimes to use --since and sometimes not).  

For both catalog and verbose - Why does this only affecting this/last boots?  That's completely irrelevant if the data you want isn't in those boots.

BryanQuigley:
Aside: I just found it is possible for us to reduce the data dupe by jouralctl -b -2 (every boot ignoring the last/this) - that still doesn't make `since` consistent unless we add it all last/this lines too - which I *really* don't like.

sos/plugins/logs.py
BryanQuigley: Before my change, this has "this" boot defined with since, but not last.  So we could end up collecting a full last boot and only a partial this boot.

TurboTurtle:
I reached out to one of the senior RH support staff who deals predominantly with can't boot situations with customers.

They've stated the this/last journal collections are "extremely helpful when it's an emergency mode sos". With that being the case, I'm unwilling to sign off on dropping the collections highlighted in the current PR.

While I like the simplification idea, it seems that we may have an impasse here with regards to the default collection ideologies. Perhaps we should still split out a separate `RedHatLogs` class, while still keeping the bulk of the plugin standardized across distros:

```python
class RedHatLogs(Logs, RedHatPlugin):

    since = ''

    def setup(self):
        super(RedHatLogs, self).setup()
        self.add_journal(boot="this", catalog=True)
        self.add_journal(boot="last", catalog=True)
```

BryanQuigley:
Sure, I'm open to doing that,   just for clarity, since = ''  basically enabled --all-logs.. I don't think you want to do that...

if all-logs do you want it to do the same catalog bits, except in with "allfields=True, output="verbose"" output?

TurboTurtle:
With `since = ''`, we get the entire journal regardless of time period, but still limited by `log_size` (or the 100mb default for journals). `--all-logs` would remove the size limiting on top of not being limited by time period. `since = ''` is what we would want.

Ideally `--all-logs` wouldn't modify the `this` or `last` boot journals, aside from removing the size limitation, but should grab `allfields=True, output="verbose"` for the 'regular' journal collection.

BryanQuigley:
Verbose is actually more than just collecting /var/log/journal (and theoretically you can use journalctl tools to parse the raw logs in much nicer ways...).   I'm not sure I want to change something that big in this rewrite, though.

sos/plugins/logs.py
pmoravec: The `since` applied just to `this` boot but not `last` is ilogical, indeed. But collecting the verbose logs makes sense. And at least on RH distros, `journalctl --boot -1 --since <yesterday>` collects nothing when boot happened 2 days ago - so if the @BryanQuigley concern is the `since` inconsistency, let add it to the `last` boot as well.

BryanQuigley: I added both catalog and verbose to a RH only bit towards the end.  Verbose to be collected separately when all-logs specified. 

@TurboTurtle mentioned that Catalog was the really useful one for last/this boots (my reading doesn't show it's useful at all, but heh).  I left those Catalog last/this out of using since because if both occurred > since you won't get any actual boot data from either of them.

sos/plugins/logs.py
```diff
@@ -143,4 +80,32 @@ def setup(self):
...
+
+    def setup(self):
+        super(RedHatLogs, self).setup()
+        self.add_journal(boot="this", catalog=True)
+        self.add_journal(boot="last", catalog=True)
+
+        since = ""
+
+        if self.get_option("since"):
+            since = self.get_option("since")
```
pmoravec: Could this filter apply not only to the `add_journal` call below, but to *any* journalctl command called? I.e. to https://github.com/sosreport/sos/blob/master/sos/plugins/__init__.py#L1369, add line:

     since = since or self.get_option("since")


Then `--since` option will apply to collecting any journal logs. (an update of `sosreport --help` shall follow, then, currently it states `--since` applies to *files* and not journals; but I think it makes sense to have unified user experience of `--since` option).

Would be that acceptable?

BryanQuigley: Yes..  Resolving this

sos/plugins/logs.py
BryanQuigley: These are included now.

sos/plugins/logs.py
pmoravec: I cant find those commands any where, in the current commit version.

sos/plugins/logs.py
```diff
@@ -50,83 +39,31 @@ def setup(self):
...
-                since = "-%ddays" % days
-            self.add_journal(since=since)
-        else:
+                 self.add_copy_spec("/var/log/journal/*")
+        else:  # If not using journal
+            # Red Hat was collecting in all cases
+            self.add_copy_spec([
+                    "/var/log/messages*",
+                    "/var/log/secure*",
+            ])
```
pmoravec: Is it OK if we collect `messages*` and `secure` *only* when not journal and systemd? (i.e. basically on legacy systems only), @TurboTurtle ?

(my personal preference is to collect these any time, but I am OK in dropping that as duplicate info to journal logs)

sos/plugins/logs.py
```diff
@@ -50,83 +39,31 @@ def setup(self):
...
-        ("log_days", "the number of days logs to collect", "", 7)
-    ]
-
-    def setup(self):
-        super(DebianLogs, self).setup()
-        days = int(self.get_option("log_days"))
-        journal = os.path.exists("/var/log/journal/")
         self.add_copy_spec([
-            "/var/log/udev",
-            "/var/log/dist-upgrade",
```
pmoravec: These two won't be further collected - is that intentional?

sos/plugins/logs.py
```diff
@@ -50,83 +39,31 @@ def setup(self):
...
+            self.add_journal()
+            self.add_journal(boot="this", catalog=True)
+            self.add_journal(boot="last", catalog=True)
             if self.get_option("all_logs"):
-                since = self.get_options("since")
-            else:
-                since = "-%ddays" % days
-            self.add_journal(since=since)
-        else:
+                 self.add_copy_spec("/var/log/journal/*")
```
pmoravec: Travis fails on a wrong indentation here :)

pmoravec:
In general, I like the unification. It would be worth mentioning in release notes that `logs.log_days` option is deprecated, but its removal in favour of --since makes sense.

sos/plugins/logs.py
TurboTurtle: IMO we should capture those every time.

pmoravec:
One subtle thing: Could we also update `--since` help description at https://github.com/sosreport/sos/blob/master/sos/sosreport.py#L142 (e.g. `s/files/logs/`)?

sos/plugins/logs.py
BryanQuigley: Will add these to general collection then.  Also expanding our syslog/kern/auth to do 1 more rotated log.

BryanQuigley: *to non-conditional collection.

sos/plugins/logs.py
BryanQuigley: Doh'!

sos/plugins/logs.py
BryanQuigley: theoretically they should be moved elsewhere, but I shouldn't just remove them - so adding back.

pmoravec:
Thanks for the changes. @TurboTurtle , is it OK to stop collecting:

            self.add_journal(boot="this",
                             allfields=True, output="verbose")
            self.add_journal(boot="last", allfields=True, output="verbose")

but collect raw `/var/log/journal/` ? (imho this makes sense).

TurboTurtle:
> Thanks for the changes. @TurboTurtle , is it OK to stop collecting:
> 
> ```
>     if self.get_option('all_logs'):
>         self.add_journal(boot="this",
>                          allfields=True, output="verbose")
>         self.add_journal(boot="last", allfields=True, output="verbose")
> ```
> 
> but collect raw `/var/log/journal/` ? (imho this makes sense).

Those are specifically requested by the RH support engineers who work on can't boot situations. I would not advise removing them.

BryanQuigley:
@TurboTurtle 
Maybe I misunderstood earlier but I thought the Catalog one was the one useful for boot debug for some reason (I still don't understand it)?  That's the only reason I left them in:
self.add_journal(boot="this", catalog=True)
self.add_journal(boot="last", catalog=True)

When selecting all journal files via --all-logs now it should allow a support engineer to generate any output they care for - for in my testing the same size (or smaller) as verbose journal output.

TurboTurtle:
Oh, you're right - I misread that!

Yeah, current proposal looks good to me.

BryanQuigley:
Any other changes needed or can we commit this?  Thanks!

bmr-cymru:
I think I am fine with the changes being proposed in this PR, but I think it would benefit everyone to separate these out a little more - at least in future pulls - generally a single commit should be one logical change.

We've tried to avoid the kind of "update a bunch of stuff" commits that you find in the tree going back many years but there's a lot going on in this patch and it makes it difficult to review and be certain what the final result is.

I think I would have been happier reviewing and signing off on this (and it would have taken less time) if each of the bullet items from the commit message had been a separate patch:

* Dropped the log_days option as it's confusing with since.
* Use journal if persistant, otherwise don't collect it.
* If all-logs collect full raw journal insteaf of verbose as it's the same size but collects more info and is easier to parse.
* Still collecting catalog but now it is limited by since. RH request.
* Collect message/secure no matter what. Rh specific.
* Collect first 3 syslog/kern/auth only if not-journal.
* Default to just stop at 100mb limit jounral limit (which is enforced outside of plugin).

I know it's more effort when putting the branch together but it can save time for multiple reviewers as we are merging the changes, and it can help massively a year or two down the line when trying to comprehend the changes in light of some regression or other future problem.

sos/sosreport.py
```diff
@@ -139,7 +139,7 @@ def _get_parser():
     parser.add_argument("--since", action="store",
                         dest="since", default=None,
                         type=_format_since,
-                        help="Escapes archived files older than date. "
+                        help="Escapes archived logs older than date. "
```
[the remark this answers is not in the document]
BryanQuigley: that a much better word to use +1.

sos/plugins/logs.py
```diff
@@ -50,91 +39,41 @@ def setup(self):
...
-            "/var/log/unattended-upgrades"
-            ])
+        ])
+
+        self.add_cmd_output("journalctl --disk-usage")
         self.add_cmd_output('ls -alRh /var/log/')
 
+        journal = os.path.exists("/var/log/journal/")
         if journal and self.is_installed("systemd"):
+            self.add_journal()
```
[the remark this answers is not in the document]
BryanQuigley: Since is now covered by the _init_ change.

BryanQuigley:
Full agree with that, this request ended up bringing in another that we could have done separately.  Next time, will try to commit 1 first and then go on.

sos/plugins/logs.py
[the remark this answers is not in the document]
bmr-cymru: Yeah, I realised that belatedly - that kind of API change with broad impact is absolutely the kind of thing that needs to be in a separate clearly labelled commit:

```
[Plugin] automatically apply since= in add_journal()
```

It's really easy to miss a subtle point like that in a large commit that is re-shuffling a lot of code between classes in one file.

I am not sure right now that this change is OK because I don't know for certain how this will interact with other options that we pass to `journalctl` - it needs to be tested so that we don't end up with any data collection regressions.

Back when I added `add_journal()` I plumbed all the options through as kwargs because we had a request from Red Hat support that they _always_ needed the current and previous boot logs - my worry here is that using `--since` will override `--boot=last` and exclude those log messages if they fall outside the date range specified with since. @TurboTurtle did a test here and it seems to confirm that `--since` does take precedence over `--boot`, meaning that you get no messages if they are excluded by the since date.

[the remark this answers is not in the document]
BryanQuigley: That's actually one of the reason I thought it needed a bigger rewrite (the bigger since change is what came from the other PR).  

The plugin as it is is quite inconsistent on this - it currently specified since for this boot.  Which means we *might* be in scenarios where using since causes us to collect a full last boot, but two incomplete copies of this boot    We can also be in the scenario where `this` boot and the main journal collection are completely equivalent (as both hit the default size limit - although technically `this` boot would likely be larger due to catalog). 

I'm happy to move the all journal since changes out of this commit/PR.  But I think it makes sense to have since apply to all boots, or else we are just knowingly disregarding the users request.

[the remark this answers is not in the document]
bmr-cymru: The only solution I can think of for this now that keeps the "auto since" behaviour is to ignore `since`/`until` in `add_journal()` if the `boot` kwarg is used - this needs to be documented in the method docstring.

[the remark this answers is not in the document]
bmr-cymru: > I think it makes sense to have since apply to all boots, or else we are just knowingly disregarding the users request.

We already do that for `messages*` etc. - we're somewhat hamstrung here, in that the request to *always* include the current and last boot came directly from support teams, so it will be seen as a regression if `--since` starts overriding that.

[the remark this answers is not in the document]
BryanQuigley: I think it can be explained easily enough, but I'm going to reopen #1848 for further discussion on since applying to all journals.  

For now, I'll have this commit/PR just do since for the normal journal and leave this/last without it.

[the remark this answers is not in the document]
bmr-cymru: I agree that we should make this more consistent - the catch is that there are specific exceptions (like always having the current/last boot in the Red Hat logs data) that we've historically agreed to that we either need to still honour, or to clearly explain when we're making changes.

sos/plugins/logs.py
```diff
@@ -50,91 +41,41 @@ def setup(self):
...
+        journal = os.path.exists("/var/log/journal/")
         if journal and self.is_installed("systemd"):
-            if self.get_option("all_logs"):
-                since = self.get_options("since")
-            else:
-                since = "-%ddays" % days
             self.add_journal(since=since)
-        else:
+            self.add_journal(boot="this", catalog=True)
+            self.add_journal(boot="last", catalog=True)
```
pmoravec: Could you apply `since` option to both these calls? As discussed above, sosreport --since "-1 day" should not collect data from previous boot 3 days ago.

BryanQuigley: I'd prefer that but @bmr-cymru wants an exception for those for RH historical reasons.  I think we've gone back and forth on the since there several times in this PR.  Let's just land it without the since and revisit it in a future PR after we see how the teams like the other changes?

PR #1824, opened 2019-10-09
[plugins] Replace /var/run paths by /run

pmoravec:
Since FHS 3.0, the system information data since boot are located in
/run. Sosreport shall collect the data directly, without the need to
follow symlink from /var/run to /run that exists just for the sake
of legacy reasons.

Resolves: #1824

---

### DIFF
--- a/sos/plugins/ceph.py
+++ b/sos/plugins/ceph.py
@@ -55,7 +55,7 @@ class Ceph(Plugin, RedHatPlugin, UbuntuPlugin):
             "/etc/ceph/",
             "/etc/calamari/",
             "/var/lib/ceph/",
-            "/var/run/ceph/"
+            "/run/ceph/"
         ])
 
         self.add_cmd_output([
--- a/sos/plugins/docker.py
+++ b/sos/plugins/docker.py
@@ -150,7 +150,7 @@ class UbuntuDocker(Docker, UbuntuPlugin):
         super(UbuntuDocker, self).setup()
         self.add_copy_spec([
             "/etc/default/docker",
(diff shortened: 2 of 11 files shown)


pmoravec:
Few further references to `/var/run` are in `ovn_host` and `ovn_central` plugins where I wasn't certain if there is a rationale for them or not (they appear in lists with both `/var/run/<something>` and `/run/<something>` that should be effectively just one item - so why it is listed twice?).

@danalsan , is there some reasoning why listing both dirs? Or can we remove the duplicate items, one pointing to the other? In particular can we delete these 2 lines:

https://github.com/sosreport/sos/blob/master/sos/plugins/ovn_host.py#L19
https://github.com/sosreport/sos/blob/master/sos/plugins/ovn_central.py#L87

?

danalsan:
> Few further references to `/var/run` are in `ovn_host` and `ovn_central` plugins where I wasn't certain if there is a rationale for them or not (they appear in lists with both `/var/run/<something>` and `/run/<something>` that should be effectively just one item - so why it is listed twice?).
> 
> @danalsan , is there some reasoning why listing both dirs? Or can we remove the duplicate items, one pointing to the other? In particular can we delete these 2 lines:
> 
> https://github.com/sosreport/sos/blob/master/sos/plugins/ovn_host.py#L19
> https://github.com/sosreport/sos/blob/master/sos/plugins/ovn_central.py#L87
> 
> ?

I think we can remove those two lines. Thanks Pavel!

pmoravec:
Done. Thanks for prompt response.

Issue #1016, opened 2017-05-25
Need to collect the ganesha log rotated files


pdvian:
The /var/log/ganesha.log and /var/log/ganesha-gfapi.log does not provide the sufficient logs if the logs are rotated just before the sosreport execution. The nfsganesha plugin should collect the sufficient logs for analysis.

PR #1825, opened 2019-10-09
[nfsganesha] Collect rotated logs when using --all-logs

TurboTurtle:
If --all-logs is specified, the nfsganesha plugin will now gather the
rotated log files from /var/log/ganesha, not just the current log file.

Fixes: #1016

---


bmr-cymru:
Merged in commit aa785860f3ea3aafb473b96d2f8b0e8fbaf13019

PR #1828, opened 2019-10-09
[lgtm] Fix several alerts highlighted by LGTM analysis

TurboTurtle:
This patchset fixes several LGTM.com alerts for unsued imports, unused variables, and overridden methods.

Removes the `log_size` variable from the `kimchi` plugin as that should have been removed in a previous sweep when we made `log_size` implicit.

Specifically disables one alert within `load_plugins()` as a false positive.

---

### DIFF
--- a/sos/archive.py
+++ b/sos/archive.py
@@ -19,9 +19,6 @@ import errno
 import stat
 from threading import Lock
 
-# required for compression callout (FIXME: move to policy?)
-from subprocess import Popen
-
 from sos.utilities import sos_get_command_output, is_executable
 
 try:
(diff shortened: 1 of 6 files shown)


bmr-cymru:
Thanks - I've merged all but the last one as-is: for that one, I think rather than disable the warning we can just re-organise the imports to avoid the problem (even though it's not that confusing in this instance the warning is correct that both 'import module' and 'from module import...' are used together here).

I've just imported `sos.plugins` at the top of file and assigned `import_plugin = sos.plugins.import_plugin` inside `load_plugins()`.

PR #1829, opened 2019-10-12
[misc] Cosmetic changes

mikelolasagasti:
- Mainly change URLs to HTTPS
- Change documentation's install instructions to match commit 7da4e70
- Change spec file's URL
- Update IBM links
- Change Ubuntu's vendor to Canonical

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ aimed at Linux distributions and other UNIX-like operating systems.
 
 This project is hosted at:
 
-  * http://github.com/sosreport/sos
+  * https://github.com/sosreport/sos
 
 For the latest version, to contribute, and for more information, please visit
 the project pages or join the mailing list.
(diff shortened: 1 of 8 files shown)


bmr-cymru:
Thanks for the patch. This should really be split into several commits since it's changing various unrelated parts of the tree: the string inside `[]` in the subject should indicate what part of sos is being updated rather than the type of change. I'll break it up as I merge.

bmr-cymru:
For the changes to the URLs in the IBM policy the commit message should either have a sign-off from IBM, or some explanation of why the URL is being changed (e.g. it's out of date / has moved etc.). For a minor change like HTTP to HTTPS it's not important, but the existing URLs are the ones that IBM asked us to use when the policies were first created - those URLs now redirect, but not to the replacement addresses suggested in this pull request. I've left those changes out for now.

PR #1831, opened 2019-10-14
[redhat] Automatically enable cantboot preset in emergency/rescue mode

TurboTurtle:
If sos is being run in emergency or rescue mode on a RH family system,
then automatically load the cantboot preset by default, instead of other
package-based presets.

---

### DIFF
--- a/sos/policies/redhat.py
+++ b/sos/policies/redhat.py
@@ -300,6 +300,10 @@ support representative.
         return False
 
     def probe_preset(self):
+        # Emergency or rescue mode?
+        for target in ["rescue", "emergency"]:
+            if self.init_system.is_running("%s.target" % target):
+                return self.find_preset(CB)
         # Package based checks
         if self.pkg_by_name("satellite-common") is not None:
             return self.find_preset(RH_SATELLITE)


bmr-cymru:
I'm not massively keen on running commands during policy loading unless it's unavoidable (like the package manager callouts) - but if it's the only mechanism available and this automation is felt to be important then it may be justifiable.

I have some reservations about using `runlevel` for this though: it has been deprecated for a long time now, and is described as obsolete in the (systemd provided) man pages. That said, it seems to get the information from `/var/run/utmp`, which is a binary file and not trivial for us to parse in sos.

The "proper" way to do this today would be to inspect the targets that `systemd` has reached: unfortunately that's not a trivial act of interpretation either:

```
# systemctl list-units --type target 
UNIT                   LOAD   ACTIVE SUB    DESCRIPTION                  
basic.target           loaded active active Basic System                 
bluetooth.target       loaded active active Bluetooth                    
cryptsetup.target      loaded active active Local Encrypted Volumes      
getty.target           loaded active active Login Prompts                
graphical.target       loaded active active Graphical Interface          
local-fs-pre.target    loaded active active Local File Systems (Pre)     
local-fs.target        loaded active active Local File Systems           
multi-user.target      loaded active active Multi-User System            
network-online.target  loaded active active Network is Online            
network-pre.target     loaded active active Network (Pre)                
network.target         loaded active active Network                      
nfs-client.target      loaded active active NFS client services          
nss-lookup.target      loaded active active Host and Network Name Lookups
nss-user-lookup.target loaded active active User and Group Name Lookups  
paths.target           loaded active active Paths                        
printer.target         loaded active active Printer                      
remote-fs-pre.target   loaded active active Remote File Systems (Pre)    
remote-fs.target       loaded active active Remote File Systems          
rpc_pipefs.target      loaded active active rpc_pipefs.target            
rpcbind.target         loaded active active RPC Port Mapper              
slices.target          loaded active active Slices                       
sockets.target         loaded active active Sockets                      
sound.target           loaded active active Sound Card                   
sshd-keygen.target     loaded active active sshd-keygen.target           
swap.target            loaded active active Swap                         
sysinit.target         loaded active active System Initialization        
timers.target          loaded active active Timers                       

LOAD   = Reflects whether the unit definition was properly loaded.
ACTIVE = The high-level unit activation state, i.e. generalization of SUB.
SUB    = The low-level unit activation state, values depend on unit type.

27 loaded units listed. Pass --all to see loaded but inactive units, too.
To show all installed unit files use 'systemctl list-unit-files'.
```

The `multi-user.target` indicates that we're in a normal environment here (it's present in what would have been runlevels 2/3/4/5 under SysV).

Alternately querying the status of the emergency or rescue targets shows if they are in use (following `systemctl rescue` on RHEL7):

```
# systemctl status rescue.target
● rescue.target - Rescue Mode
   Loaded: loaded (/usr/lib/systemd/system/rescue.target; disabled; vendor preset: disabled)
   Active: active since Tue 2019-10-15 14:53:53 BST; 1min 10s ago
     Docs: man:systemd.special(7)
# systemctl status emergency.target
● emergency.target - Emergency Mode
   Loaded: loaded (/usr/lib/systemd/system/emergency.target; static; vendor preset: disabled)
   Active: inactive (dead)
     Docs: man:systemd.special(7)
```

Worse, this seems to interact badly with the runlevel, as reported by the command - here's the output I get following a series of `systemctl` operations switching between rescue and multi-user:

```
# systemctl rescue
# runlevel
3 1
# systemctl isolate multi-user.target
# runlevel
1 3
```

I don't know if this is a bug, or if using `isolate` to return from single user is not valid (using `telinit 4` results in the same output), but this is very confusing output in the context of what we're trying to understand here.

One option would be to look for the file `/run/nologin` (formerly `/var/run/nologin`, and going back to Red Hat Linux days...). This is created whenever the system is unable to process non-root logins (e.g. during boot up, rescue/emergency mode, or administrative lock-out). It's checked by the `pam_nologin` module and used to refuse ordinary user logins. In my tests it is reliably created when moving to the rescue or emergency targets via whatever mechanism, and it is cleaned up however you move back to a multi user runlevel/target afterwards. This should also be the case during emergency boots or when the systemd target is overridden on the kernel command line.

One other wart with this on current systems is the "emergency" boot entries created on Red Hat systems these days. These are intended to allow system recovery using a "known good" kernel and initramfs image. The entries are set up at install time and provide an alternate boot environment for system recovery, *but*, they do not by themselves select any kind of rescue or emergency target - the system boots into multi-user.target as normal, with the network active and other services loaded as normal. The only indication that a rescue environment is in use is the `BOOT_IMAGE` option present by default in recent RHEL/Fedora kernel command line options strings:

```
# cat /proc/cmdline 
BOOT_IMAGE=/vmlinuz-0-rescue-523a903dd71b4166b3ba5884464bfb1f root=/dev/mapper/rhel-root ro crashkernel=auto rd.lvm.lv=rhel/root rd.lvm.lv=rhel/swap rhgb quiet
```

bmr-cymru:
(I should add that using the rescue boot image does **not** create the `/run/nologin` flag).

jcastill:
What about using 'systemctl status' directly? Or actually, 'systemctl status |grep State'.  When I test it, three things happen depending on what's the status of the system:

a) If the system went into maintenance mode directly, 'State' is "starting".
b) If the system failed to load, but offered the possibility to go into maintenance mode or use Control+D to reboot, and we chose to go into maintenance mode, 'State' is "maintenance".
c) If the system booted without problems, 'State' is "running". 

Maybe this can help to avoid using 'runlevel'? At least for RHEL 7 and 8...

TurboTurtle:
The problem with using `systemctl status` is that a single failed unit will instead force a status of `Degraded`, and that persists across different targets. So we can't use it as a reliable method for determining emergency/rescue mode.

I think the most acceptable approach here for the time being, is to query the target status like @bmr-cymru mentions above, e.g. `systemctl status emergency.target` - if it's active I think we have a sufficiently reasonable assurance that we are in fact in that mode and not some other state (degraded, or still getting a `1` from `runlevel` when you go to single user.

I'll hold on pushing an update for a short bit for anyone to comment on using `systemctl status (emergency|rescue).target` as our check (or at least, a first check).

TurboTurtle:
I've updated this PR to use the policy-defined init system to query the rescue/emergency targets and enable on those, rather than the `runlevel` command.

bmr-cymru:
Let's change that `rlvl` local to something like `target` - we're in the Red Hat policy here, so already tied to systemd terminology. Other than that I think this is good to go.

TurboTurtle:
Ack, done.

PR #1832, opened 2019-10-15
[fwupd] new plugin

mikelolasagasti:

---

### DIFF
--- a/sos/plugins/fwupd.py
+++ b/sos/plugins/fwupd.py
@@ -1,0 +1,52 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+
+
+class Fwupd(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
+    """fwupd information
+    """
+
+    plugin_name = 'fwupd'
+    profiles = ('system', )
+
+    packages = ('fwupd',)
+
+    def setup(self):
+        self.add_cmd_output([
+            "fwupdmgr get-approved-firmware",
+            "fwupdmgr get-devices --no-unreported-check",
+            "fwupdmgr get-history",
+            "fwupdmgr get-remotes",
+            # collect json format using fwupdagent
+            "/usr/libexec/fwupd/fwupdagent get-devices",
+            "/usr/libexec/fwupd/fwupdagent get-updates",
+        ])
+
+        self.add_copy_spec([
+            "/etc/fwupd",
+        ])
+
+        self.add_journal(units="fwupd")
+
+    def postproc(self):
+        self.do_path_regex_sub(
+            "/etc/fwupd/remotes.d/*",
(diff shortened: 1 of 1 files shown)


sos/plugins/fwupd.py
```diff
@@ -0,0 +1,45 @@
...
+            "fwupdmgr get-approved-firmware",
+            "fwupdmgr get-devices --no-unreported-check",
+            "fwupdmgr get-history",
+            "fwupdmgr get-remotes",
+            "fwupdmgr get-topology --no-unreported-check"
+        ])
+
+        self.add_copy_spec([
+            "/etc/fwupd",
+            "/usr/share/fwupd",
```
pmoravec: Is whole directory required? Since e.g. `/usr/share/fwupd/metainfo/org.freedesktop.fwupd.remotes.lvfs-testing.metainfo.xml` or `/usr/share/fwupd/remotes.d/vendor/firmware/README.md` and maybe others dont seem important for toubleshooting.

I dont know `fwupd` much, but guessing that something like:

    /usr/share/fwupd/remotes.d/fwupd/metadata.xml
    /usr/share/fwupd/quirks.d

will be enough to collect..?

sos/plugins/fwupd.py
```diff
@@ -0,0 +1,45 @@
...
+        ])
+
+        self.add_journal(units="fwupd")
+
+    def postproc(self):
+        self.do_file_sub(
+            "/etc/fwupd/redfish.conf",
```
pmoravec: Nice catch. I think `/etc/fwupd/remotes.d/lvfs-testing.conf` needs the same (maybe some more or even all configs? those two have "Password" present by default).

mikelolasagasti:
I contacted @hughsie about the `/usr/share/fwupd/*` and the commands and applied what he suggested: remove one command and `/usr/share/fwupd/*` from the plugin.

He also commented that `get-devices` can report in json format using another command instead of fwupdmgr. What is preferred, regular output or json formatted?

pmoravec:
> He also commented that `get-devices` can report in json format using another command instead of fwupdmgr. What is preferred, regular output or json formatted?

From sos point of view, it does not matter (until either format requires longer time to collect, is much bigger than the other or similar). The format is rather a users' choice.

Generally, I would say json is preferred since it allows automatic post-processing of sosreport data. But standard output can be more human readable, so.. This very depends on the output and users' prefferences.

sos/plugins/fwupd.py
```diff
@@ -0,0 +1,51 @@
...
+
+        self.add_copy_spec([
+            "/etc/fwupd",
+        ])
+
+        self.add_journal(units="fwupd")
+
+    def postproc(self):
+        for remotes_conf in glob("/etc/fwupd/remotes.d/*"):
+            self.do_file_sub(
```
pmoravec: Isn't it easier to call:

    self.do_path_regex_sub("/etc/fwupd/remotes.d/*", r"Password=(.*)", r"Password=********")

? Not sure now if this will work for subdirectories as well, but I dont expect `/etc/fwupd/remotes.d/foo/bar.conf` configs to exist?

hughsie: > Not sure now if this will work for subdirectories as well

There isn't expected to be be subdirs in `remotes.d`

hughsie:
> Generally, I would say json is preferred

/me also.

bmr-cymru:
The `[...]` in the `add_copy_spec()` call are redundant here but I'll remove those when I merge - no need to respin.

PR #1842, opened 2019-10-30
[Plugin] Avoid potential duplicate sysroot join

TurboTurtle:
An issue was discovered when sos was run in a container where the
sysroot when set to something other than '/' (e.g. '/host' in the Red
Hat support-tools container) was getting duplicated in checks against
defined forbidden paths.

This was because self.join_sysroot() is called in multiple spots during
file collection phases (both in `setup()` and `collect()`) before it is
additionally called in the internal `_is_forbidden_path()`. Since the
test for if `join_sysroot()` should be called will always return True in
a containerized environment (because the sysroot is not '/'), this would
cause the container's sysroot (e.g. '/host') to be appended again,
creating for example '/host/host/foo.txt'. Since this duplicated sysroot
append would cause forbidden path comparisons to fail, sos would not
only collect blacklisted sensitive files but would also hang in the
`kernel` plugin due to the fact it was now trying to collect trace_pipe
and other such files that it should otherwise skip.

This commit adds a conditional to `_is_forbidden_path()` to not call
`join_sysroot()` if the path in question already starts with the
sysroot.

Resolves: #1842

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -758,7 +758,7 @@ class Plugin(object):
         return None
 
     def _is_forbidden_path(self, path):
-        if self.use_sysroot():
+        if self.use_sysroot() and not path.startswith(self.sysroot):
             path = self.join_sysroot(path)
         return _path_in_path_list(path, self.forbidden_paths)
 


pmoravec:
Good catch of the underlying double-prepend issue and of possible fix.

`_is_forbidden_path` is called from:
- `add_copy_spec` (https://github.com/sosreport/sos/blob/master/sos/plugins/__init__.py#L985) *after* joining sysroot (https://github.com/sosreport/sos/blob/master/sos/plugins/__init__.py#L947)
- `_do_copy_path` (https://github.com/sosreport/sos/blob/master/sos/plugins/__init__.py#L780) that is called from:
  - `_collect_copy_specs` that traverses `self.copy_paths` that already has prepended the sysroot
  - in `_copy_symlink` (https://github.com/sosreport/sos/blob/master/sos/plugins/__init__.py#L734) called on *stripped* sysroot
  - in `_copy_dir` that is called *just* from https://github.com/sosreport/sos/blob/master/sos/plugins/__init__.py#L805 inside `_do_copy_path`, that is called from (and recursivelly goto few lines above)

So, can't be an alternate fix:

    diff --git a/sos/plugins/__init__.py b/sos/plugins/__init__.py
    index e75ec82e..7719cc1e 100644
    --- a/sos/plugins/__init__.py
    +++ b/sos/plugins/__init__.py
    @@ -731,7 +731,7 @@ class Plugin(object):
     
             # skip recursive copying of symlink pointing to itself.
             if (absdest != srcpath):
    -            self._do_copy_path(self.strip_sysroot(absdest))
    +            self._do_copy_path(absdest)
             else:
                 self._log_debug("link '%s' points to itself, skipping target..."
                                 % linkdest)

to ensure `_do_copy_path` contains filepaths always with sysroot?

(I will double-read my analysis tomorrow, if I havent skipped any call)

TurboTurtle:
I don't think so for two reasons:

1. I'm seeing this cause a hang in the kernel plugin on paths such as `/sys/kernel/debug/tracing/per_cpu/cpu0/trace_pipe` which is not a symlink

2. In `_is_forbidden_path` we are prepending the sysroot only to the string being compared against the `forbidden_paths` list, and not the string we actually end up using to use as the path we're collecting.

pmoravec:
1. that was my misunderstanding, the bug appears on each and every type of path *except* (expanded) symlinks (where is the only place we strip the sysroot before adding it back in `_is_forbidden_path`)

2. I dont understand here :( anyway full alternative fix would need dropping the join_sysroot, so it would be:

    diff --git a/sos/plugins/__init__.py b/sos/plugins/__init__.py
    index e75ec82e..4f1b73ce 100644
    --- a/sos/plugins/__init__.py
    +++ b/sos/plugins/__init__.py
    @@ -731,7 +731,7 @@ class Plugin(object):
     
             # skip recursive copying of symlink pointing to itself.
             if (absdest != srcpath):
    -            self._do_copy_path(self.strip_sysroot(absdest))
    +            self._do_copy_path(absdest)
             else:
                 self._log_debug("link '%s' points to itself, skipping target..."
                                 % linkdest)
    @@ -758,8 +758,6 @@ class Plugin(object):
             return None
     
         def _is_forbidden_path(self, path):
    -        if self.use_sysroot():
    -            path = self.join_sysroot(path)
             return _path_in_path_list(path, self.forbidden_paths)
     
         def _copy_node(self, path, st):

bmr-cymru:
Either of these changes addresses the immediate problem, and allows sos to complete on a container based system. I think we need to take a bit more time to look at the wider problem however, and determine whether either of these is adequate as a workaround in the meantime. While testing with the patch from @pmoravec I noticed that this also interacts with the link follow-up changes we made last year:

```
commit 322f4a517ae336cc1443f9a399a0d15d45ec48b9
Date:   Fri Sep 7 13:11:03 2018 -0400

    [archive] add link follow-up to FileCacheArchive.add_link()
    
    Creating a link may trigger further actions in the archive: if the
    link target is a regular file, we must copy that file into the
    archive, and if the target is a symbolic link, then we must create
    that link, and copy in the link target.
    
    Handle this by calling add_file() or (recursively) add_link() in
    order to create the missing pieces of the symlink chain.
    
    These operations must take place outside of the path lock since
    they do not modify the archive namespace and will call methods of
    the Archive object that will attempt to re-acquire this lock.
    
    Resolves: #1404
    
```

```
commit 919e8671a6ab9684d59525eb7f3607b3aab08ee1
Date:   Tue Sep 11 12:16:57 2018 -0400

    [archive] fix link rewriting logic in FileCacheArchive.add_link()
    
    When processing link follow up for an original symbolic link, the
    add_link() logic incorrectly used the _original_ host link name,
    rather than the to-be-created name when calculating relative path
    structures. If the prior link is at a greater or lesser level of
    directory nesting this will lead to broken relative links in the
    archive (one level too high or too low).
    
    In some cases (systemd) this behaviour was masked due to the fact
    that identically named links exist at multiple levels of the path
    hierarchy.
    
```

This fixed a situation where a path contained a symlinked directory, e.g. "/path/to/foo", where 'to' is a symlink to a directory containing "foo". Previously the archive would break these paths by turning the intermediate component into a directory.

Now the archive class traverses the path structure and uses `add_link()` etc. to properly set up parent paths. In doing this, "follow up" tasks copy the link target - this is outside the normal sysroot handling in `Plugin`, and also avoids the forbidden path mechanism.

We need to decide if the proposed workarounds are adequate for the time being - ultimately, this work needs to be reconsidered (possibly by moving more of the work into Plugin from Archive).

TurboTurtle:
Closing in favor of #1845

PR #1845, opened 2019-11-01
[Plugin, kernel] interim sysroot fixes

bmr-cymru:
Fix several aspects of path handling that affect container environments using an alternate sysroot (`-s/--sysroot`):

  * Invalid handling of symlink and directory paths, leading to `/host` paths appearing in the archive
  * Incorrect  {strip,join}_sysroot() use in `_do_copy_path()` and `_is_forbidden_path()` leading to `/host` paths leaking into the archive and mis-blacklisting of plugin paths
  * Forbidden path glob styles in the `kernel` plugin that conflict with the change to forbidden path checking in #1695 that cause forbidden paths to be incorrectly added to the archive

This addresses current and urgent issues that make sos unusable in this environment: further problems may exist (especially in the `FileCacheArchive` link follow-up handling used when intermediate path components contain symlinks), and these need to be reviewed and addressed appropriately but this set of changes represents a minimum needed to successfully collect a report from this type of system.

This supersedes pull #1842.
---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -738,10 +738,11 @@ class Plugin(object):
 
     def _copy_dir(self, srcpath):
         try:
-            for afile in os.listdir(srcpath):
+            for name in os.listdir(srcpath):
                 self._log_debug("recursively adding '%s' from '%s'"
-                                % (afile, srcpath))
-                self._do_copy_path(os.path.join(srcpath, afile), dest=None)
+                                % (name, srcpath))
+                path = os.path.join(srcpath, name)
+                self._do_copy_path(path)
         except OSError as e:
             if e.errno == errno.ELOOP:
                 msg = "Too many levels of symbolic links copying"
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -800,7 +800,7 @@ class Plugin(object):
                 if not os.listdir(srcpath):
                     self.archive.add_dir(dest)
                     return
-                self._copy_dir(dest)
+                self._copy_dir(srcpath)
                 return
 
         # handle special nodes (block, char, fifo, socket)
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -798,9 +798,9 @@ class Plugin(object):
             if stat.S_ISDIR(st.st_mode) and os.access(srcpath, os.R_OK):
                 # copy empty directory
                 if not os.listdir(srcpath):
-                    self.archive.add_dir(srcpath)
+                    self.archive.add_dir(dest)
                     return
-                self._copy_dir(srcpath)
+                self._copy_dir(dest)
                 return
 
         # handle special nodes (block, char, fifo, socket)
@@ -808,7 +808,7 @@ class Plugin(object):
             ntype = _node_type(st)
             self._log_debug("creating %s node at archive:'%s'"
                             % (ntype, dest))
-            self._copy_node(srcpath, st)
+            self._copy_node(dest, st)
             return
 
         # if we get here, it's definitely a regular file (not a symlink or dir)
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -731,7 +731,7 @@ class Plugin(object):
 
         # skip recursive copying of symlink pointing to itself.
         if (absdest != srcpath):
-            self._do_copy_path(self.strip_sysroot(absdest))
+            self._do_copy_path(absdest)
         else:
             self._log_debug("link '%s' points to itself, skipping target..."
                             % linkdest)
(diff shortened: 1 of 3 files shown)


bmr-cymru:
I'm currently looking for other problematic `add_forbidden_path()` calls - kernel was the most obvious on the system I was testing due to the blocking behaviour reading the blacklisted paths. It's possible that other plugins need to be updated to cope with the new forbidden path handling too, but I wanted to get this branch available for review ASAP.

TurboTurtle:
As of the latest commit, I'm now seeing successful runs both in and outside of a container.

LGTM.

sos/plugins/__init__.py
```diff
@@ -738,10 +738,13 @@ def _copy_symlink(self, srcpath):
...
     def _copy_dir(self, srcpath):
         try:
-            for afile in os.listdir(srcpath):
+            for path in os.listdir(srcpath):
                 self._log_debug("recursively adding '%s' from '%s'"
-                                % (afile, srcpath))
-                self._do_copy_path(os.path.join(srcpath, afile), dest=None)
+                                % (path, srcpath))
+                if os.path.isdir(path):
+                    self._copy_dir(srcpath)
```
pmoravec: This test and branch sounds wrong to me:

- `path` is always an item inside `srcpath` dir, so it rarely would be a directory itself. So this branch would (almost) never be executed. 
- even if the branch would be executed, we would recursively call the same method with the same argument

So one approach is to have here:

    if os.path.isdir(os.path.join(srcpath, path)):
        self._copy_dir(os.path.join(srcpath, path))
    else:
        self._do_copy_path(os.path.join(srcpath, path))

(optionally using a variable for the `join`). But we should simply call *just*

    self._do_copy_path(os.path.join(srcpath, path))

every time (as the code executes even now practically every time). Otherwise, if we would recursively call `_copy_dir` directly, I am afraid we would: 1) skip `_is_forbidden_path` for the subdir (try collecting `/foo` dir but forbid `/foo/bar` subdir), 2) skip adding `/foo/bar` to the archive.

So I suggest removing the 3 lines of practically dead code.

pmoravec: Hm, my suggestion reverts back @bmr-cymru's whole change in this method, that makes me less confident.

pmoravec:
LGTM. I just dont understand how the `if os.path.isdir(path):` test can work (inside a container) but havent been able to get any issue.

About the trailing globs in `kernel`: reviewing all other plugins, I spotted 3 of them that deserve the same:

    diff -rup a/sos/plugins/ceph.py b/sos/plugins/ceph.py
    --- a/sos/plugins/ceph.py	2019-11-01 20:40:21.050459696 +0100
    +++ b/sos/plugins/ceph.py	2019-11-02 10:26:03.062548399 +0100
    @@ -103,8 +103,8 @@ class Ceph(Plugin, RedHatPlugin, UbuntuP
                 "/var/lib/ceph/*keyring*",
                 "/var/lib/ceph/*/*keyring*",
                 "/var/lib/ceph/*/*/*keyring*",
    -            "/var/lib/ceph/osd/*",
    -            "/var/lib/ceph/mon/*",
    +            "/var/lib/ceph/osd",
    +            "/var/lib/ceph/mon",
                 # Excludes temporary ceph-osd mount location like
                 # /var/lib/ceph/tmp/mnt.XXXX from sos collection.
                 "/var/lib/ceph/tmp/*mnt*",
    diff -rup a/sos/plugins/libreswan.py b/sos/plugins/libreswan.py
    --- a/sos/plugins/libreswan.py	2019-11-01 20:40:21.058459707 +0100
    +++ b/sos/plugins/libreswan.py	2019-11-02 10:26:17.737566074 +0100
    @@ -49,7 +49,7 @@ class Libreswan(Plugin, RedHatPlugin, De
     
             self.add_forbidden_path([
                 '/etc/ipsec.secrets',
    -            '/etc/ipsec.secrets.d/*',
    +            '/etc/ipsec.secrets.d',
                 '/etc/ipsec.d/*.db',
                 '/etc/ipsec.d/*.secrets'
             ])
    diff -rup a/sos/plugins/vdsm.py b/sos/plugins/vdsm.py
    --- a/sos/plugins/vdsm.py	2019-11-01 20:40:21.073459727 +0100
    +++ b/sos/plugins/vdsm.py	2019-11-02 10:25:02.065474940 +0100
    @@ -60,9 +60,9 @@ class Vdsm(Plugin, RedHatPlugin):
         plugin_name = 'vdsm'
     
         def setup(self):
    -        self.add_forbidden_path('/etc/pki/vdsm/keys/*')
    +        self.add_forbidden_path('/etc/pki/vdsm/keys')
             self.add_forbidden_path('/etc/pki/vdsm/libvirt-spice/*-key.*')
    -        self.add_forbidden_path('/etc/pki/libvirt/private/*')
    +        self.add_forbidden_path('/etc/pki/libvirt/private')
     
             self.add_service_status(['vdsmd', 'supervdsmd'])
     
(these wont probably harm now - until `libreswan` or `vdsm` is present inside a container, I guess).

sos/plugins/__init__.py
bmr-cymru: That was my mistake; the important part of the change is correcting the argument passed to `_copy_dir()` from `_do_copy_path()`, not the handling of sub-directories in the method itself.

Issue #1848, opened 2019-11-05
[plugins] --since option different impact on logfiles and journal logs


pmoravec:
In `sosreport` help:

      --since SINCE         Escapes archived files older than date. This will also
                            affect --all-logs. Format: YYYYMMDD[HHMMSS]

Apart of the heuristic problems per #1847, the behaviour is quite confusing with respect to collecting (log)files and journal logs (command output). As `--since` option causes that

- older logs are NOT collected (regardless of `--all-logs` option)
- `add_journal` for a unit is still collected completely (e.g. when a plugin wants its own journal)
- `add_journal` for `this` and `last` boot is still collected completely
  - BUT when adding `--all-logs` option, a *new* verbose journalctl of this boot *with since* is collected, BUT also verbose journalctl of prev boot *without* since is collected. So we apply `--since` only to newer logs.

I see there two counter-intuitive behaviour:
- logs apply to `--since` but journal logs dont
- with `--all-logs`, `--since` is applied to current boot but not the prev one

Doesn't it make sense to unify this? E.g.:

    diff --git a/sos/plugins/__init__.py b/sos/plugins/__init__.py
    index 1a1464c1..af45d47d 100644
    --- a/sos/plugins/__init__.py
    +++ b/sos/plugins/__init__.py
    @@ -1366,6 +1366,7 @@ class Plugin(object):
     
             journal_size = 100
             all_logs = self.get_option("all_logs")
    +        since = since or self.get_option("since")
             log_size = sizelimit or self.get_option("log_size")
             log_size = max(log_size, journal_size) if not all_logs else 0
     
    diff --git a/sos/plugins/logs.py b/sos/plugins/logs.py
    index 77e2da85..538ffb62 100644
    --- a/sos/plugins/logs.py
    +++ b/sos/plugins/logs.py
    @@ -51,8 +51,7 @@ class Logs(Plugin):
                     self.add_copy_spec(i)
     
             if self.get_option('all_logs'):
    -            self.add_journal(boot="this", since=self.get_option("since"),
    -                             allfields=True, output="verbose")
    +            self.add_journal(boot="this", allfields=True, output="verbose")
                 self.add_journal(boot="last", allfields=True, output="verbose")
     
         def postproc(self):

?

pmoravec:
Closing in favour of #1822.

TurboTurtle:
Re-closing this given comment history.

PR #1849, opened 2019-11-05
[plugins, logs] Apply --since option to journal logs

pmoravec:
journalctl / add_journal should apply --since command line option
the same way like collecting logfiles / add_copy_spec does.

Resolves: #1848

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -1366,6 +1366,7 @@ class Plugin(object):
 
         journal_size = 100
         all_logs = self.get_option("all_logs")
+        since = since or self.get_option("since")
         log_size = sizelimit or self.get_option("log_size")
         log_size = max(log_size, journal_size) if not all_logs else 0
 
--- a/sos/plugins/logs.py
+++ b/sos/plugins/logs.py
@@ -51,8 +51,7 @@ class Logs(Plugin):
                 self.add_copy_spec(i)
 
         if self.get_option('all_logs'):
-            self.add_journal(boot="this", since=self.get_option("since"),
-                             allfields=True, output="verbose")
+            self.add_journal(boot="this", allfields=True, output="verbose")
             self.add_journal(boot="last", allfields=True, output="verbose")
 
     def postproc(self):


BryanQuigley:
Can we  simplify the logs plugin first, see #1822?  I think I do fix this in that.

pmoravec:
Ah, I haven't checked that PR.

That PR fixes something from this one, but I lack there:
- `add_journal` called on a unit (say, `chronyd` from that plugin) does not respect `--since` and collects also older log entries
- I still see `journalctl_--no-pager_--catalog_--boot[|_-1]` that also do not respect `--since`

I *think* if your PR moves the `since = self.get_option("since")` to `add_journal` as `since = since or self.get_option("since")`, my concerns would be resolved - going to check this now.

BryanQuigley:
>add_journal called on a unit (say, chronyd from that plugin) does not respect --since and collects also older log entries
Hmm.. do we want to have to be explicit in the plugin that we want --since applied?  I guess I can't see where that would go poorly - except with specifying boots.  

>I still see journalctl_--no-pager_--catalog_--boot[|_-1] that also do not respect --since
I'm happy to add that if it's somehow better, but I don't understand the point of specifing boots if you will override it with since.

pmoravec:
> > add_journal called on a unit (say, chronyd from that plugin) does not respect --since and collects also older log entries
> Hmm.. do we want to have to be explicit in the plugin that we want --since applied?  I guess I can't see where that would go poorly - except with specifying boots.

I dont understand here. With my proposal, `--since` (if used) will be applied to journal units. If some plugin wants to ignore this option, it can call `add_journal(since="19700101")` which supersets the `--since` cmdline option.

> 
> > I still see journalctl_--no-pager_--catalog_--boot[|_-1] that also do not respect --since
> I'm happy to add that if it's somehow better, but I don't understand the point of specifing boots if you will override it with since.

Two use cases:
- I am interested in logs since some time, regardless if it was current or previous boot
- I know something bad happened yesterday, got rebooted the system, needs RCA but dont need much historical logs - so I can use `--since` with 2 days ago. `this` and `last` boot will ensure I got logs since past 2 days, regardless it was before or after the recent reboot (assuming no more reboots happened).

BryanQuigley:
I'm convinced on moving to since to apply by default for all plugins journal - as you said a plugin can override it. Happy to approve just that bit - or incorporate it into the other PR.

pmoravec:
I am glad for the consensus. I prefer if you update your PR to have it merged altogether, but can build on top of your commit.

pmoravec:
Closing in favour of #1822 .

PR #1851, opened 2019-11-06
[virsh] add new path collection of virt-manager logs

s-morishima:
The path of virt-manager logs was changed from
"/root/.virt-manager/" to "/root/.cache/virt-manager".
New path collection is added to virsh plugin.

Signed-off-by: MORISHIMA Shigeki <s.morishima@fujitsu.com>

---

### DIFF
--- a/sos/plugins/virsh.py
+++ b/sos/plugins/virsh.py
@@ -20,10 +20,10 @@ class LibvirtClient(Plugin, RedHatPlugin, UbuntuPlugin, DebianPlugin):
 
     def setup(self):
         # virt-manager logs
-        if not self.get_option("all_logs"):
-            self.add_copy_spec("/root/.virt-manager/*", sizelimit=5)
-        else:
-            self.add_copy_spec("/root/.virt-manager/*")
+        self.add_copy_spec([
+            "/root/.cache/virt-manager/*.log",
+            "/root/.virt-manager/*.log"
+        ])
 
         cmd = 'virsh -r'
 


pmoravec:
Why not collect both dirs in either case? `sosreport` / `Plugin` class properly detects nonexisting copypath.

(plus Travis complains about wrong indentation)

s-morishima:
I'm sorry for the wrong indentation.
I will fix my patch to reflect your comments.

pmoravec:
No problem (that's why we have the automated checks) and thanks for PR and prompt answer.

sos/plugins/virsh.py
```diff
@@ -21,8 +21,10 @@ class LibvirtClient(Plugin, RedHatPlugin, UbuntuPlugin, DebianPlugin):
     def setup(self):
         # virt-manager logs
         if not self.get_option("all_logs"):
+            self.add_copy_spec("/root/.cache/virt-manager/*", sizelimit=5)
             self.add_copy_spec("/root/.virt-manager/*", sizelimit=5)
         else:
+            self.add_copy_spec("/root/.cache/virt-manager/*")
```
bmr-cymru: There are a couple of points here:

  * since you're just adding a new path, with otherwise identical arguments please use a single `add_copy_spec()` call and pass a list, i.e.:

```python
        self.add_copy_spec([
            "/root/.cache/virt-manager/*",
            "/root/.virt-manager/*"
        ], sizelimit=5)
```

  * Since you're collecting the same list of paths in both calls you could also use a local variable for the list to avoid repeating it in both `add_copy_spec()` calls.

More generally (these issues existed before your patch, but if we're changing this we should consider them):

  * The same set of logs is collected with `--all-log` and without - only the size limit is different. Normally `--all-logs` signals that rotated or other non-default log locations should be collected.

  * The plugin hard-codes a log size limit. Normally this should be left unset, so that the global `--log-size` option takes effect. If there is some special reason for a different value in this plugin it should be mentioned in the comments.

s-morishima: Thanks, I see.

s-morishima: Unfortunately, I have no special reason for uniquely setting "sizelimit=5"
in virsh plugin, and I plan to reflect the former comment only about 
the way to call add_copy_spec(). Is that ok?

bmr-cymru: If there's no reason documented for having it then let's get rid of it and simplify the plugin - if there's no additional data to collect with `--all-logs` then that's fine. We can remove that branch and just allow the automatic size limits to apply.

s-morishima: Thanks, I pushed patch v3 based on your comment.
Also, referring to other plugins (e.g. libvirt),
I change the file name of collecting virt-manager logs
from "*" to "*.log" to collect them more precisely.

PR #1854, opened 2019-11-11
[Plugin|Policy] Only call lsmod once and standardize kmod checks

TurboTurtle:
This commit makes two changes to how sos deals with kernel modules and
their state during a run of sosreport.

First, no longer call `lsmod` for every individual plugin during its
enablement check. Instead, call `Policy.lsmod()` only once during
`load_plugins()` before we begin iterating over the plugins, and save
that information as a policy class attr.

Second, have `Plugin.is_module_loaded()` check for kmod presence in the
saved policy class attr for kernel_mods, rather than checking through
`/proc/modules`. Have the plugin enablement checks now also use
`is_module_loaded()` to standardize with how `SoSPredicate`s are
checked.

Note that this change results in a significant performance increase for
sos initialization times in a RHEL 7 container.

Resolves: #1854

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -1278,11 +1278,8 @@ class Plugin(object):
                                       chdir=runat, binary=binary, env=env)
 
     def is_module_loaded(self, module_name):
-        """Return whether specified moudle as module_name is loaded or not"""
-        if len(grep("^" + module_name + " ", "/proc/modules")) == 0:
-            return False
-        else:
-            return True
+        """Return whether specified module as module_name is loaded or not"""
+        return module_name in self.policy.kernel_mods
 
     # For adding output
     def add_alert(self, alertstring):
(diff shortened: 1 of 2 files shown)


sos/sosreport.py
```diff
@@ -567,6 +567,8 @@ def load_plugins(self):
         self.profiles = set()
         using_profiles = len(self.opts.profiles)
         policy_classes = self.policy.valid_subclasses
+        # load kmods for plugin enablement checks
+        self.policy.lsmod()
```
bmr-cymru: Do we want to be driving this from `sosreport`, or should it be auto-initialised either during policy initialisation or on first use (i.e. have a `@property` method that initialises it)?

It seems a bit out of keeping with the other system data we make available through policies to have this explicitly initialised by the user (e.g. `PackageManager` and `InitSystem`).

TurboTurtle:
Updated to include that typo fix.

PR #1855, opened 2019-11-11
Changes to better integrate snap package installation types

slashdd:
---

### DIFF
--- a/sos/plugins/lxd.py
+++ b/sos/plugins/lxd.py
@@ -17,6 +17,7 @@ class LXD(Plugin, UbuntuPlugin):
     plugin_name = 'lxd'
     profiles = ('container',)
     packages = ('lxd',)
+    commands = ('lxd',)
 
     # Version 2.0.X:
     # - /etc/default/lxd-bridge
--- a/sos/policies/ubuntu.py
+++ b/sos/policies/ubuntu.py
@@ -8,6 +8,8 @@ class UbuntuPolicy(DebianPolicy):
     distro = "Ubuntu"
     vendor = "Canonical"
     vendor_url = "https://www.ubuntu.com/"
+    PATH = "/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games" \
+           + ":/usr/local/sbin:/usr/local/bin:/snap/bin"
 
     def __init__(self, sysroot=None):
         super(UbuntuPolicy, self).__init__(sysroot=sysroot)
--- a/sos/plugins/kubernetes.py
+++ b/sos/plugins/kubernetes.py
@@ -189,6 +189,6 @@ class UbuntuKubernetes(Kubernetes, UbuntuPlugin):
 
     files = ('/root/cdk/kubeproxyconfig',)
 
-    kube_cmd = "/snap/bin/kubectl --kubeconfig=/root/cdk/kubeproxyconfig"
+    kube_cmd = "kubectl --kubeconfig=/root/cdk/kubeproxyconfig"
 
 # vim: et ts=5 sw=4


sos/plugins/lxd.py
```diff
@@ -17,6 +17,7 @@ class LXD(Plugin, UbuntuPlugin):
     plugin_name = 'lxd'
     profiles = ('container',)
     packages = ('lxd',)
+    files = ('/snap/bin/lxd')
```
TurboTurtle: Since `/snap/bin` is being added as a path, I think this should actually be:

```
commands = ('lxd',)
```

bmr-cymru: If the Ubuntu policy is extended so that the `PackageManager` instance reflects the presence of snap packages then there's no need for a command or file trigger - as long as the snap package is called "lxd", or something else we can add to the packages tuple then we can just enable the plugin on that basis. It shouldn't be too hard to merge the package lists from `dpkg` and snap here.

slashdd: I agree it would be the ideal scenario. I think last time I checked a year ago and one of the challenge was that the PackageManager doesn't support detection of >= 2 packages type at the same time. It's either it will work with SNAP only or DEB only, but not both in the same time iirc. There was also other challenges due to the fact that SNAP doesn't behave like DEB nor RPM, meaning SNAP would need is own customise flavor of PackageManager.

bmr-cymru: The `PackageManager` is a really lightweight abstraction - it doesn't impose any such restrictions. The main thing it does is maintain a dictionary of package name/version/release data:

```python
    packages[name] = {'name': name, 'version': version, 'release': release}
```

There are optional hooks for `query_command`, `verify_command` and `files_command`, but implementations aren't obliged to use those - they are just a quick way of bootstrapping a new package manager type without having to write too much code (e.g. I've considered going back to using the Python RPM bindings in the Red Hat policies for performance reasons - the round trip via `rpm -qa` to text and back is quite costly on systems with many packages installed).

A combined dpkg/snap `PackageManager` would just need to override `get_pkg_list()` to obtain NVR data from both and add those to the `packages` dictionary.

bmr-cymru: (also note that `PackageManager` support isn't a prerequisite for the suggestion @TurboTurtle made - evaluation of `commands` triggers just relies on the ability to search `PATH` - see `sos.utilities.is_executable()`).

slashdd: @bmr-cymru just to make sure we are on the same page. Do you want me to explore the get_pkg_list() to obtain NVR data from both approach or use the `commands` approach @TurboTurtle suggested ?

bmr-cymru: I'm just saying that both are valid ways of triggering the plugin and are preferable to a hard coded absolute path in the plugin `files` list - implementing `PackageManager` support for snap (and doing something similar for the other auxiliary package managers in the policy classes for other distros) has some obvious advantages but is a bigger piece of work to undertake.

slashdd: I totally agree that in the long run we would want a `PackageManager` support for snap (and even for flatpak)

What do you think if for now if I change the PR for `commands`, but will consider the former in my next set of changes to submit in sosreport. That's definitely something I want to pursue for quite some time. My previous attempts weren't successful and I unfortunately didn't had much time to resume the work on it but will do in 2020.

Does that look like a good plan for you ?

slashdd: For now we want to cover snap specifically for Canonical products only : lxd, juju, canonical-livepatch and CDK (Canonical Distribution of K8s). Other than that no other plugins should pay attention to snap for now, but in the long run yes they might so the `PackageManager` approach will be ideal before we cross that bridge, but we still have time. If for now we can cover Canonical product while I'm thinking/working about the ultimate snap pkg support, it would be good.

slashdd: @BryanQuigley thoughts ?

sos/policies/ubuntu.py
```diff
@@ -8,6 +8,8 @@ class UbuntuPolicy(DebianPolicy):
     distro = "Ubuntu"
     vendor = "Canonical"
     vendor_url = "https://www.ubuntu.com/"
+    PATH = "/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games" \
+           + ":/usr/local/sbin:/usr/local/bin:/snap/bin"
```
slashdd: @TurboTurtle @bmr-cymru @pmoravec @BryanQuigley Just thinking out loud after a discussion I had with Bryan. Would it make sense to you in a near future to stop hardcoding PATH for policies and rely on system PATH environment variable instead ? and fallback to this hardcoded one if no system PATH env variable found or empty ?

slashdd: In the same time it can be misleading depending how bad the system is impacted by problem 'X'. Maybe the hardcoded PATH is a safety net to make sure the PATH will be reliable in case the system is too broken or that we can't rely on system PATH env ?

bmr-cymru: Right, that's the problem: we introduced `PATH` in policies for two reasons. To avoid the need to maintain absolute paths in plugin code and so that whatever messed up environment the user might have when running sos doesn't affect the execution of commands by the tool.

This was at the time of the "UsrMove" change in Red Hat and Fedora: prior to this we used absolute paths in many places in sos. Having a stable `PATH` defined by policies allowed us to simplify plugin and increase code sharing by just using the binary name and to ensure that we would be able to find executables even if the user's PATH was incorrectly set.

I considered adding something to obtain the path by parsing/interpreting `/etc/profile` for Red Hat distributions but that gets quite complex (and there's no guarantee that the file hasn't been modified/broken, so we would still need some kind of heuristic to validate the resulting path list).

slashdd: Gotcha, thanks.

BryanQuigley:
Merged via c1d1f459b,  2afbe9ae574446, and c1d1f459b

Issue #1846, opened 2019-11-04
[openstack_horizon] fix useless forbidden paths


bmr-cymru:
The current horizon plugin has:

```python
        self.add_forbidden_path("*.py[co]")
```

This attempts to forbid collection of Python source/object files however the forbidden paths mechanism works on actual paths - since no directory prefix is given this pattern will not match any pathnames and will not exclude any data from the report.

PR #1857, opened 2019-11-13
[openstack_horizon] fix useless forbidden paths

jcastill:
The code has currently the following line:

        self.add_forbidden_path("*.py[co]")

But add_forbidden_path() needs a directory prefix, so this won't
work. This line was introduced via commit 4d75385ad to solve
issue #856, that discussed the postproc method in
openstack_horizon. It seems that the python object and source
files are present in /etc/openstack-dashboard/local_settings.d/
used to customize the dashboard. This patch adds that directory
to the add_forbidden_path() line.

Closes: #1846

---

### DIFF
--- a/sos/plugins/openstack_horizon.py
+++ b/sos/plugins/openstack_horizon.py
@@ -41,7 +41,9 @@ class OpenStackHorizon(Plugin):
             self.var_puppet_gen + "/horizon/etc/httpd/conf.modules.d/*.conf",
             self.var_puppet_gen + "/memcached/etc/sysconfig/memcached"
         ])
-        self.add_forbidden_path("*.py[co]")
+        self.add_forbidden_path(
+            "/etc/openstack-dashboard/local_settings.d/*.py[co]"
+        )
 
     def postproc(self):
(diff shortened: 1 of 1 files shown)


sos/plugins/openstack_horizon.py
```diff
@@ -41,7 +41,8 @@ def setup(self):
             self.var_puppet_gen + "/horizon/etc/httpd/conf.modules.d/*.conf",
             self.var_puppet_gen + "/memcached/etc/sysconfig/memcached"
         ])
-        self.add_forbidden_path("*.py[co]")
+        self.add_forbidden_path(
+            "/etc/openstack-dashboard/local_settings.d/*.py[co]")
```
TurboTurtle: Minor stylistic nitpick - sos-style would be 

```python
self.add_forbidden_path(
    "/etc/openstack-dashboard/local_settings.d/*.py[co]"
)
```

jcastill:
I'll do it now. Sorry, I used autopep8 to fix the long line, and I thought that was enough.

pmoravec:
Fixed in 43d1562.

PR #1859, opened 2019-11-13
[sapnw] implement collect_cmd_output API change

pmoravec:
Since e8bb94c, collect_cmd_output returns a dict instead of filename.
Reflect the change in testing command output and accessing filename.

Resolves: #1859

---

### DIFF
--- a/sos/plugins/sapnw.py
+++ b/sos/plugins/sapnw.py
@@ -32,14 +32,13 @@ class sapnw(Plugin, RedHatPlugin):
             "/usr/sap/hostctrl/exe/saphostctrl -function ListInstances",
             suggest_filename="SAPInstances"
         )
-        if not inst_out:
+        if inst_out['status'] != 0:
             return
 
         sidsunique = set()
         # Cycle through all the instances, get 'sid', 'instance_number'
         # and 'vhost' to determine the proper profile
-        p = open(inst_out, "r").read().splitlines()
-        for inst_line in p:
+        for inst_line in inst_out['output'].splitlines():
             if "DAA" not in inst_line:
                 fields = inst_line.strip().split()
                 sid = fields[3]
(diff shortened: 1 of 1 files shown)


sos/plugins/sapnw.py
```diff
@@ -32,13 +32,13 @@ def collect_list_instances(self):
...
         )
-        if not inst_out:
+        if inst_out['status'] != 0:
             return
 
         sidsunique = set()
         # Cycle through all the instances, get 'sid', 'instance_number'
         # and 'vhost' to determine the proper profile
-        p = open(inst_out, "r").read().splitlines()
+        p = open(inst_out['filename'], "r").read().splitlines()
```
TurboTurtle: Since this is using `collect_cmd_output`, we can just use the `output` key instead of opening the file. E.G.

`for line in inst_out['output'].splitlines():`

sos/plugins/sapnw.py
```diff
@@ -99,10 +99,10 @@ def collect_list_dbs(self):
             suggest_filename="SAPDatabases"
         )
 
-        if not db_out:
+        if db_out['status'] != 0:
             return
 
-        dbl = open(db_out, "r").read().splitlines()
+        dbl = open(db_out['filename'], "r").read().splitlines()
```
TurboTurtle: Same here. `collect_cmd_output` already returns the command's output in the dict.

sos/plugins/sapnw.py
bmr-cymru: But `collect_cmd_output()` also returns an `'output'` entry in the dictionary - we should just `splitlines()` on that and use it, rather than re-read the content into memory a second time.

sos/plugins/sapnw.py
bmr-cymru: Ditto.

pmoravec:
Thanks for such obvious feedback, PR updated.

Issue #286, opened 2014-05-16
Mechanism to selectively disable post-processing


bmr-cymru:
Sometimes it's useful to preserve unmodified data in the reports, including passwords and other sensitive items (e.g. because they are required for proper diagnosis).

Allow the user to selectively disable the mechanism.

PR #1862, opened 2019-11-14
[Plugin] Allow selectively disabling postprocessing

TurboTurtle:
Adds two mechanisms by which users can choose to disable postprocessing
of collected information.

First, is a global method exposed via the `--no-postproc` option. Using
this option will skip postprocessing for all plugins.

Second, is a per-plugin option exposed via a new 'postproc' plugin
option. This is set to _True_ by default (meaning yes, perform
postprocessing), which users can set to False or off to disable
postprocessing for that plugin only; e.g. `-k podman.postproc=off`

Closes: #286
Resolves: #1862

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -329,6 +329,11 @@ class Plugin(object):
     # Default predicates
     predicate = None
     cmd_predicate = None
+    _default_plug_opts = [
+        ('timeout', 'Timeout in seconds for plugin', 'fast', -1),
+        ('postproc', 'Enable post-processing collected plugin data', 'fast',
+         True)
+    ]
 
     def __init__(self, commons):
         if not getattr(self, "option_list", False):
@@ -352,9 +357,8 @@ class Plugin(object):
         self.soslog = self.commons['soslog'] if 'soslog' in self.commons \
             else logging.getLogger('sos')
 
-        # add the 'timeout' plugin option automatically
-        self.option_list.append(('timeout', 'timeout in seconds for plugin',
-                                 'fast', -1))
+        # add the default plugin opts
+        self.option_list.extend(self._default_plug_opts)
 
         # get the option list into a dictionary
         for opt in self.option_list:
--- a/man/en/sosreport.1
+++ b/man/en/sosreport.1
@@ -10,6 +10,7 @@ sosreport \- Collect and package diagnostic and support data
           [-a|--alloptions] [-v|--verbose]\fR
           [-k plug.opt|--plugin-option plug.opt]\fR
           [--no-report] [--config-file conf]\fR
+          [--no-postproc]\fR
           [--preset preset] [--add-preset add_preset]\fR
           [--del-preset del_preset] [--desc description]\fR
           [--batch] [--build] [--debug] [--dry-run]\fR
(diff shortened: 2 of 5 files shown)


sos/plugins/__init__.py
```diff
@@ -1570,6 +1574,15 @@ def setup_verify(self):
         if verify_cmd:
             self.add_cmd_output(verify_cmd)
 
+    def _postproc(self):
```
bmr-cymru: The scoping looks funny here - normally the public (non `_`-prefixed) version of the method is the wrapper, and the private method is the wrapped one.

I think it would be less confusing to re-name the existing method to `_postproc()`, and have `sosreport` continue to call `plug.postproc()` - then depending on the option value the public method calls or doesn't call `self._postproc()` to do the actual work.

bmr-cymru:
> My only comment would be it might be nice to add these to the .travis ci for automatically making sure they continue to work error free.

It would also be good to add a unit test to verify that `Plugin` properly does/doesn't call `postproc()` based on the option values.

sos/plugins/__init__.py
TurboTurtle: The plugins doing postprocessing today all override `postproc()`, hence why I went with this approach. So if `sosreport` still calls that directly, we'd need to change the plugins to now override `_postproc()` instead, no? Seemed a bit awkward to me.

The other option could be to do this check directly in `sosreport`, e.g.:

```python
    def postproc(self):
        for plugname, plug in self.loaded_plugins:
            try:
                if plug.get_option('postproc'):
                    plug.postproc()
                else:
                    self.log_info('some skip message')
```

bmr-cymru: Yeah, I see what you mean - although I think having the `get_option()` check in sosreport looks better - we aren't super strict about public vs. private in sos (and neither is Python ;-), but it looks weird to have a "private" method called by another module, that calls a similarly-named "public" method.

TurboTurtle:
I've updated the approach to check for the `postproc` plugin option in `sosreport` rather than within `Plugin`, which matches the behavior to check for `--no-postproc` globally.

I've added a single plugin test that checks to make sure the plugin default for `postproc` remains `True`.  I've not found a way with the current unit test structure to properly test the actual checks `sosreport` does, but at least this check is _something_. All ears for any suggestions on this front (including how to test this with Travis @BryanQuigley)

BryanQuigley:
I need to propose a better way for travis going forward but just a simple:
sudo ~/virtualenv/python$TRAVIS_PYTHON_VERSION/bin/python ./sosreport --batch --no-postproc  --config-file=sos.conf

Would be a basic "check" - again doesn't check that it actually works though.

BryanQuigley:
No need to add anything more to travis.   I'll add this easily in the script in #1896 

I think all conversations are resolved?  Need any more reviews?

pmoravec:
ACK to the PR, just wouldnt it be better to have the options with equivalent (and not opposite) meaning? I mean have global `--no-postproc` (defaults to False) and plugin specific `nopostproc` (defaults to False), OR have global `--postproc` (defaults to True) and plugin specific `postproc` (defaults to True). It might be confusing to use complementary options global/plugin-wise - but that can be my subjective opinion only.

TurboTurtle:
I went with the plugin option of `postproc` since using `-k plugin.postproc=off` seems more intuitive than ` -k plugin.nopostproc=on`, I.E. "turn this off" versus "turn this on to turn this off". Similarly the global "turn it all off with this toggle" seems easier to grok to me.

PR #1865, opened 2019-11-19
[ovn_central] Execute 'podman' commands with stdin=PIPE

danalsan:
When executing commands inside a container with podman, it's
required to use a pipe for stdin or they'll hang.
This patch is accomplishing that and it might be reverted when [0]
is fixed.

[0] https://bugzilla.redhat.com/show_bug.cgi?id=1732525

Signed-off-by: Daniel Alvarez <dalvarez@redhat.com>

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -1256,7 +1256,7 @@ class Plugin(object):
         )
 
     def exec_cmd(self, cmd, timeout=300, stderr=True, chroot=True, runat=None,
-                 env=None, binary=False, pred=None):
+                 env=None, binary=False, pred=None, stdin=None):
         """Execute a command right now and return the output and status, but
         do not save the output within the archive.
 
@@ -1275,7 +1275,8 @@ class Plugin(object):
             root = None
 
         return sos_get_command_output(cmd, timeout=timeout, chroot=root,
-                                      chdir=runat, binary=binary, env=env)
+                                      chdir=runat, binary=binary, env=env,
+                                      stdin=stdin)
 
     def is_module_loaded(self, module_name):
         """Return whether specified module as module_name is loaded or not"""
--- a/sos/plugins/ovn_central.py
+++ b/sos/plugins/ovn_central.py
@@ -12,6 +12,7 @@ from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
 import json
 import os
 import six
+from subprocess import PIPE
 
 
 class OVNCentral(Plugin):
@@ -21,6 +22,12 @@ class OVNCentral(Plugin):
     profiles = ('network', 'virt')
     _container_runtime = None
     _container_name = None
+    _container_stdin = None
+
+    def exec_cmd(self, *args, **kwargs):
+        return super().get_command_output(*args,
+                                          stdin=self._container_stdin,
+                                          **kwargs)
 
     def get_tables_from_schema(self, filename, skip=[]):
         if self._container_name:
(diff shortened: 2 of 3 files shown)


danalsan:
> This is the second attempt to work around a podman bug with a hack in sos - it's hard to take this seriously when it seems that in almost six months there hasn't even been any serious analysis of the problem in podman that necessitates this workaround. As we said before, we are very reluctant to take something like this, that changes sos interfaces and the way we call external commands, without even having any understanding of the cause of the original bug in podman or why this workaround is necessary. Experience shows that comments like "we might revert this when bug XXX is fixed" tend to be wishful thinking: if we work around this now in sos, then all the pressure on the podman team to address the bug is removed. This isn't good practice for either project.

I understand your point but I'd really love you to understand mine. Especially when this second attempt that you don't seem to want to take seriously is a consequence of a broader discussion and an attempt to *isolate* the impact of the 'so called' workaround to just the ovn_central plugin.

Do I understand that you lean towards dropping this and wait until podman fixes it with no pressure on their team?

bmr-cymru:
> Especially when this second attempt that you don't seem to want to take seriously is a consequence of a broader discussion and an attempt to *isolate* the impact of the 'so called' workaround to just the ovn_central plugin.

But it completely fails to do that since it requires a global change to the `sos_get_command_output()` interface in `sos.utilities` - the other changes in `ovn_central` cannot "isolate" the change since once that global change has been made it is very difficult to unmake it. For an example of this, see this commit:

```
commit bb8f125ffc46855c7ff1bdb67532b617cc0f2f9e
Date:   Wed Oct 23 15:27:19 2019 +0100

    [archive] remove obsolete TarFile wrapper
    
    The _TarFile wrapper in sos.archive is only needed for python 2.6
    and earlier: remove it since we no longer support this version.
    
    Closes: #1834
    Resolves: #1838
    
```

We just removed that workaround  a couple of weeks ago - it had been there since 2014:

```
commit 1b15835a18c98b3adc4f7454652f2a8604bed4cf
Date:   Thu Sep 18 15:44:47 2014 +0100

    [archive] introduce py2.6 TarFile compat wrapper
    
    Python 2.6 lacks the filter parameter to TarFile.add() that the
    TarFileArchive uses to preserve permissions on archive content.
    
    Work around this by creating a wrapper class that implements the
    missing functionality (duplicated from the py2.7.5 tarfile.py)
    and use it whenever running on a python version less than 2.7.
    
```

It **IS** possible to isolate this to just the `ovn_central` plugin, but that requires the `sos_get_command_output()` function be duplicated inside the plugin, and making the modifications to `stdin=` locally in that file, where they really **cannot** affect any other plugin - that would be an appropriate downstream workaround if there is really is no other immediate solution to this problem but it is something that we would very much want to avoid taking upstream (I am sure you can see that we can only carry a small number of these "heavy" workarounds, like the `TarFileArchive` wrapper before they become a burden). I would suggest naming the function something like `__ovn_get_command_output()` in that event to avoid any confusion.

> is a consequence of a broader discussion

Where is that discussion? The linked bugzilla really does not contain much discussion or analysis of hte problem at all, and contains some very misleading statements like:

> This patch [0] fixed the issue on our side. The problem is that the 'sos' tool was not reading from stdin so it received a SIGTTIN and all the other podman commands failed.
>Perhaps podman should be more resilient but still the way we were reading the command output was not right.


That's absolutely incorrect: the change being proposed here is to have sos open a PIPE on podman's `stdin` - it's not `sos` that receives SIGTTIN, it's `podman` (since it is attempting to read from a TTY while it is in the background). If you believe that the way sos reads command output is incorrect please can you explain what that is, and how this patch fixes it?

danalsan:
All right, understood, will report downstream. Thanks for the time and explanations.

Issue #1662, opened 2019-04-25
[system] Permission denied /proc/sys/fs/binfmt_misc inside LXD container


slashdd:
Running sosreport inside LXD, does not break anything besides showing a 'permission denied' message on sosreport when trying to list /proc/sys/fs/binfmt_misc inside an LXD container from system plugin.

We noticed a behaviour change from LXD 3.0.3 to 3.12.

The denial is cause on purpose by a apparmor lxd policy:
```
./lxd/apparmor.go:  # Handle binfmt
./lxd/apparmor.go:  mount fstype=binfmt_misc -> /proc/sys/fs/binfmt_misc/,
./lxd/apparmor.go:  deny /proc/sys/fs/binfmt_misc/{,**} rwklx,
```

3.0.3 was working okay due to a bug, but new version 3.12 fix it for good, and now prevent sosreport to collect binfmt_misc as it used to only when LXD 3.12 and late is in used.

Adding `"/proc/sys/fs/binfmt_misc/*"` in the actual `add_forbidden_path` would work but will also penalize all other situation where collecting would have work just fine. Since so far this is isolated to LXD only, maybe system plugin might need its own separate `add_forbidden_path` for LXD related stuff ?

LXD bug reference:
https://github.com/lxc/lxd/issues/5688

slashdd:
or we leave it as-is because anything it's not breaking anything, it's just screaming for the permission denied but doesn't cause any harm.

slashdd:
Looking https://github.com/lxc/lxd/blob/master/lxd/apparmor.go

There might be more thing to make sure sosreport doesn't collect when inside a LXD container.

dnegreira:
Maybe we can add a verification that we are running inside an LXD container and append `/proc/sys/fs/binfmt_misc/*` to a list based on current `add_forbidden_path` list

slashdd:
Should `add_copy_spec` not copy a path if a permission denied occurred?  instead of letting a given plugin errors out ? and simply leave a WARNING|ERROR in sos.log:
`WARNING or ERROR: [plugin:system] permission denied on  '/proc/sys/fs/binfmt_misc/*'`

@BryanQuigley  thought ?

BryanQuigley:
I'd want these kinds of Access Denied errors to be shown in the output in a general sense.  We don't want to hide from the end user that we can't collect some files as root.

The other #1299 permission denied errors (the rest of the forbidden paths in system) are a case that makes more sense to just exclude.

[...]

PR #1867, opened 2019-11-19
[Plugin] Detect copy operations without the adequate access rights

slashdd:
Likely due to Linux Security Modules such as apparmor, selinux, ...

Fix: #1662

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -744,6 +744,10 @@ class Plugin(object):
                 path = os.path.join(srcpath, name)
                 self._do_copy_path(path)
         except OSError as e:
+            if e.errno == errno.EPERM or errno.EACCES:
+                msg = "Permission denied"
+                self._log_warn("_copy_dir: '%s' %s" % (srcpath, msg))
+                return
             if e.errno == errno.ELOOP:
                 msg = "Too many levels of symbolic links copying"
                 self._log_error("_copy_dir: %s '%s'" % (msg, srcpath))


slashdd:
Here's an example inside a LXD container with no access to `/proc/sys/fs/binfmt_misc`

`sos.log:2019-11-19 21:43:30,657 ERROR: [plugin:system] _copy_dir: PermissionError: Likely due to Linux Security Modules '/proc/sys/fs/binfmt_misc'`

slashdd:
After many re-consideration from my part, I have think switching from log_error to log_warn and changed/refined the msg to the following would make more sense:

`sos.log:2019-11-20 00:31:36,663 WARNING: [plugin:system] _copy_dir: /proc/sys/fs/binfmt_misc 'Permission denied likely due to Linux Security Modules (LSM) restrictions'`

slashdd:
lxd source pkg explaining why binfmt is permission denied during sosreport run inside a lxd container.

lxd/apparmor.go
```
  # deny writes in /proc/sys/fs but allow binfmt_misc to be mounted
  mount fstype=binfmt_misc -> /proc/sys/fs/binfmt_misc/,
  deny @{PROC}/sys/fs/** wklx,
```

slashdd:
@bmr-cymru @pmoravec @TurboTurtle I would like to request a review from you guys when time permit.

sos/plugins/__init__.py
```diff
@@ -744,6 +744,11 @@ def _copy_dir(self, srcpath):
                 path = os.path.join(srcpath, name)
                 self._do_copy_path(path)
         except OSError as e:
+            if e.errno == errno.EPERM or errno.EACCES:
+                msg = "Permission denied likely due to Linux Security " \
+                       + "Modules (LSM) restrictions"
+                self._log_warn("_copy_dir: '%s' %s" % (srcpath, msg))
+                return
```
pmoravec: I am OK with catching (not re-raising) these exceptions, just having 2 comments:

- afaik copying directory should not hit `EPERM` that applies to operations not permitted (user cant kill `root's` process or similar), but as some applications might mis-use this error code instead of `EACCES`, and since both are some permission errors, I am OK with having both caughts
- I would strip away the "likely due to LMS" and just mention "Permission denied:" on given `srcpath` and probably also on the error details (that can give more data than this "guess")

slashdd: Done

Issue #1863, opened 2019-11-18
[sar] limitation by --log-size does not work


miztake:
Changed to limit with --log-size instead of sa_size by the following commit.
  [[sar] collect whole sar log dir]
  https://github.com/sosreport/sos/commit/b3d9fdf568e25997eb006e575d26fb301b91414b

However, the sa file is collected even if the collection size reaches the default value of 25MB for --log-size.

It may be related to the following issues.
  https://github.com/sosreport/sos/issues/1750

PR #1870, opened 2019-11-22
[plugins] Support for directory specification in add_copy_spec()

miztake:
Previously, when a directory was specified in add_copy_spec(),
stat.ST_SIZE returned the size of the directory.
As a result, the limitation by --log-size did not work.

When a directory is specified in add_copy_spec(),
use os.walk() to find out the files under the directory.
From this, the limitation by --log-size work correctly.

Resolves: #1863

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -1425,7 +1425,14 @@ class Plugin(object):
             self._add_cmd_output(cmd='%s %s' % (udev_cmd, dev))
 
     def _expand_copy_spec(self, copyspec):
-        return glob.glob(copyspec)
+        if os.path.isdir(copyspec):
+            paths = []
+            for root, dirs, files in os.walk(copyspec):
+                for _file in files:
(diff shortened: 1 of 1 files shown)


pmoravec:
In general, I like the patch. It also works well when the copypath is a symlink to a directory (an improvement to current behaviour).

The failing unit test case is due to `add_forbidden_path` relying on `glob.glob` that resolves pathnames differently than `os.walk` in several use cases (relative path, symlinks, e.g.). So a similar change would have to be applied to

https://github.com/sosreport/sos/blob/master/sos/plugins/__init__.py#L843

I guess. BUT can't this potentially break some plugins relying on the old behaviour of `add_forbidden_path`..?

miztake:
Hi, @pmoravec,

Thank you for your comment.

I understood why Travis CI fails.
I created a patch to pass Travis CI, but I can't see the effect on other plugins.
For this, I will post a simpler patch that solves #1863.

I will close this PR.
Thanks for your cooperation.

PR #1873, opened 2019-12-03
[Policy|Plugin] Provide a Container Runtime Abstraction

TurboTurtle:
Adds a new `ContainerRuntime` abstraction class that should streamline the use/probing of a container runtime by plugins.

Plugins now have access to new methods that allow them to query for the presence of containers, images, and volumes without needing to manually specify different runtime commands to run through.

The OpenStack plugins that were doing these checks have been updated to now use the loaded `ContainerRuntime` initialized with `Policy`, as have the `docker` and `podman` plugins to get their element lists.

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -1309,27 +1309,119 @@ class Plugin(object):
 
     def exec_cmd(self, cmd, timeout=cmd_timeout, stderr=True, chroot=True,
                  runat=None, env=None, binary=False, pred=None,
-                 foreground=False):
+                 foreground=False, container=False):
         """Execute a command right now and return the output and status, but
         do not save the output within the archive.
 
         Use this method in a plugin's setup() if command output is needed to
         build subsequent commands added to a report via add_cmd_output().
         """
+        _default = {'status': None, 'output': ''}
         if not self.test_predicate(cmd=True, pred=pred):
-            return {
-                'status': None,
-                'output': ''
-            }
+            return _default
 
         if chroot or self.commons['cmdlineopts'].chroot == 'always':
             root = self.sysroot
         else:
             root = None
 
+        if container:
+            if self._get_container_runtime() is None:
+                self._log_info("Cannot run cmd '%s' in container %s: no "
+                               "runtime detected on host." % (cmd, container))
+                return _default
+            if self.container_exists(container):
+                cmd = self.fmt_container_cmd(container, cmd)
+            else:
+                self._log_info("Cannot run cmd '%s' in container %s: no such "
+                               "container is running." % (cmd, container))
+
         return sos_get_command_output(cmd, timeout=timeout, chroot=root,
                                       chdir=runat, binary=binary, env=env,
                                       foreground=foreground)
+
+    def _get_container_runtime(self, runtime=None):
+        """Based on policy and request by the plugin, return a usable
+        ContainerRuntime if one exists
+        """
+        if runtime is None:
+            if 'default' in self.policy.runtimes.keys():
+                return self.policy.runtimes['default']
+        else:
+            for pol_runtime in list(self.policy.runtimes.keys()):
+                if runtime == pol_runtime:
+                    return self.policy.runtimes[pol_runtime]
+        return None
+
+    def container_exists(self, name):
+        """If a container runtime is present, check to see if a container with
+        a given name is currently running
+        """
+        _runtime = self._get_container_runtime()
+        if _runtime is not None:
+            con = _runtime.get_container_by_name(name)
+            return con is not None
+        return False
+
+    def get_container_by_name(self, name):
+        _runtime = self._get_container_runtime()
+        if _runtime is not None:
+            return _runtime.get_container_by_name(name)
+        return None
+
+    def get_containers(self, runtime=None, get_all=False):
+        """Return a list of all container IDs from the Policy's
+        ContainerRuntime.
+
+        If `runtime` is not provided, use the Policy default. If the specified
+        `runtime is not loaded, return empty.
+        """
+        _runtime = self._get_container_runtime(runtime=runtime)
+        if _runtime is not None:
+            if get_all:
+                return _runtime.get_containers(get_all=True)
+            else:
+                return _runtime.containers
+        return []
+
+    def get_container_images(self, runtime=None):
+        """Return a list of all image names from the Policy's
+        ContainerRuntime
+
+        If `runtime` is not provided, use the Policy default. If the specified
+        `runtime is not loaded, return empty.
+        """
+        _runtime = self._get_container_runtime(runtime=runtime)
+        if _runtime is not None:
+            return _runtime.images
+        return []
+
+    def get_container_volumes(self, runtime=None):
+        """Return a list of all volume names from the Policy's
+        ContainerRuntime
+
+        If `runtime` is not provided, use the Policy default. If the specified
+        `runtime is not loaded, return empty.
+        """
+        _runtime = self._get_container_runtime(runtime=runtime)
+        if _runtime is not None:
+            return _runtime.volumes
+        return []
+
+    def get_container_logs(self, container, **kwargs):
+        """Helper to get the `logs` output for a given container
+
+        Supports passthru of add_cmd_output() options
+        """
+        _runtime = self._get_container_runtime()
+        if _runtime is not None:
+            self.add_cmd_output(_runtime.get_logs_command(container), **kwargs)
+
+    def fmt_container_cmd(self, container, cmd):
+        if self.container_exists(container):
+            _runtime = self._get_container_runtime()
+            return _runtime.fmt_container_cmd(container, cmd)
+        return cmd
 
     def is_module_loaded(self, module_name):
         """Return whether specified module as module_name is loaded or not"""
--- a/sos/policies/__init__.py
+++ b/sos/policies/__init__.py
@@ -14,6 +14,7 @@ from getpass import getpass
 from pwd import getpwuid
 from sos.utilities import (ImporterHelper,
                            import_module,
+                           is_executable,
                            shell_out,
                            sos_get_command_output)
 from sos.plugins import IndependentPlugin, ExperimentalPlugin
@@ -22,6 +23,7 @@ from sos import SoSOptions, _arg_names
 from textwrap import fill
 from six import print_
 from six.moves import input
+from pipes import quote
 
 PRESETS_PATH = "/var/lib/sos/presets"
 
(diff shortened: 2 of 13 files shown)


TurboTurtle:
Re-pushed to address the LGTM alert.

This patchset will need some better-than-normal testing across different environments before it is merge ready, since it touches a heck of a lot especially for layered technology deployments.

@danalsan would you be able to give this a check from an OVN perspective?

@battlemidget @BryanQuigley I'm fairly certain this change as written would actually break a decent amount of `docker` plugin collection on Ubuntu/Debian hosts (mainly inspect output - config collection should still be fine) mainly because the check within `DockerContainerRuntime` relies on an `InitSystem` within the `Policy` and those distros haven't hooked up to that internally yet.

I think the best solution there is to have those policies define an `InitSystem`, thoughts?

sos/policies/__init__.py
```diff
@@ -49,6 +51,107 @@ def load(cache={}, sysroot=None):
...
+    def _check_is_active(self):
+        """Check to see if the container runtime is both present AND active.
+
+        Active in this sense means that the runtime can be used to glean
+        information about the runtime itself and containers that are running.
+        """
+        return is_executable(self.binary)
+
+    def get_containers(self, get_all=False):
+        """Get a list of containers present on the system.
```
pmoravec: Maybe worth noticing here that the list contains tuples `(conatiner_id, container_name)` - it took me a while (and a "podman ps" output) to understand it from the `containers.append((ent[0], ent[-1]))` line.

sos/policies/__init__.py
```diff
@@ -49,6 +51,107 @@ def load(cache={}, sysroot=None):
...
+        if not self.active or name is None:
+            return None
+        for c in self.containers:
+            if re.match(name, c[1]):
+                return c[1]
+        return None
+
+    def get_images(self):
+        """Get a list of images present on the system
+        """
```
pmoravec: Same comment applies here: worth explaining the tuples in the list?

sos/policies/__init__.py
```diff
@@ -49,6 +51,107 @@ def load(cache={}, sysroot=None):
...
+        return containers
+
+    def get_container_by_name(self, name):
+        """Get the container ID for the container matching the provided
+        name
+        """
+        if not self.active or name is None:
+            return None
+        for c in self.containers:
+            if re.match(name, c[1]):
```
pmoravec: Will be the reg.exp. utilized or can't be just string comparison sufficient? Just for performance reasons..

sos/plugins/__init__.py
```diff
@@ -1256,27 +1257,99 @@ def collect_cmd_output(self, cmd, suggest_filename=None,
...
+    def get_container_by_name(self, name):
+        if self.container_runtime:
+            return self.container_runtime.get_container_by_name(name)
+
+    def get_containers(self, runtime=None, get_all=False):
+        """Return a list of all container IDs from the Policy's
+        ContainerRuntime.
+
+        If `runtime` is provided, return an empty list if the Policy's loaded
+        ContainerRuntime matches that runtime.
```
pmoravec: .. if the loaded ContainerRuntime does _not_ match that runtime..?

sos/plugins/__init__.py
```diff
@@ -1256,27 +1257,99 @@ def collect_cmd_output(self, cmd, suggest_filename=None,
...
+            else:
+                return [c[0] for c in self.container_runtime.containers]
+        return []
+
+    def get_container_images(self, runtime=None):
+        """Return a list of all image names from the Policy's
+        ContainerRuntime
+
+        If `runtime` is specified, return an empty list if the Policy's loaded
+        ContainerRuntime matches that runtime
```
pmoravec: ditto here

sos/plugins/__init__.py
```diff
@@ -1256,27 +1257,99 @@ def collect_cmd_output(self, cmd, suggest_filename=None,
...
+        if runtime is None or self.container_runtime.runtime_name == runtime:
+            return self.container_runtime.images
+        return []
+
+    def get_container_volumes(self, runtime=None):
+        """Return a list of all volume names from the Policy's
+        ContainerRuntime
+
+        If `runtime` is specified, return an empty list if the Policy's loaded
+        ContainerRuntime matches that runtime
```
pmoravec: ditto here

sos/policies/__init__.py
TurboTurtle: Done.

sos/policies/__init__.py
TurboTurtle: Done.

TurboTurtle:
Updated to address @pmoravec 's points. Also, now there is support for multiple runtimes on a single system. This should almost never be the case, but it could feasibly be configured and we'd still want to collect data from both. Since the `podman` and `docker` plugins independently specify their own runtimes for use, this should be unobtrusive to everything.

@danalsan were you able to test the `ovn_central` plugin or was that an eyes-only check? I don't have anything on my end with OVN deployed, so it'd be really nice to have some "real" testing on that front.

sos/plugins/__init__.py
TurboTurtle: Done.

sos/plugins/__init__.py
```diff
@@ -1256,27 +1256,109 @@ def collect_cmd_output(self, cmd, suggest_filename=None,
...
+            if self._get_container_runtime() is None:
+                self._log_info("Cannot run cmd '%s' in container %s: no "
+                               "runtime detected on host." % (cmd, container))
+                return _default
+            if self.container_exists(container):
+                cmd = self.fmt_container_cmd(container, cmd)
+            else:
+                self._log_info("Cannot run cmd '%s' in container %s: no such "
+                               "container is running." % (cmd, container))
+
```
pmoravec: Shall not we call `return _default` here? (well, current behaviour will call docker/podman cmd on non-existing container that works "well" also).

pmoravec: Hmm.. it might be valuable to let execute that command and get something like:

    Error: error getting image "3848ec06ffb6": unable to find a name and tag match for 3848ec06ffb6 in repotags: no such image

No preferences from my side here.

danalsan:
> @danalsan were you able to test the `ovn_central` plugin or was that an eyes-only check? I don't have anything on my end with OVN deployed, so it'd be really nice to have some "real" testing on that front.

It's failing for me on RHEL 8.1. Timing out on rabbit and podman plugins so perhaps we need to extend this commit https://github.com/sosreport/sos/commit/0c9a1f0cb98c5256a0ec1dec83b7c94d3bb39170 to other plugins...  I didn't catch this earlier when we tested it with @bmr-cymru 

[root@controller-0 ~]# ps -aef|grep timeout
root      864461  849771  0 09:15 pts/0    00:00:00 timeout 300s docker exec -t rabbitmq-bundle-podman-0 rabbitmqctl report
root      865838  849771  0 09:15 pts/0    00:00:00 timeout 300s podman inspect 92c3c0e2867e


Plugin podman timed out
Plugin rabbitmq timed out

Podman version:
podman-1.4.2-5.module+el8.1.0+4240+893c1ab8.x86_64


This seems unrelated to this particular PR but looks like there's another issue and ovn_central is not detected as enabled but it does in the same environment using master branch.

Hope it helps,
Daniel

PS. I won't be around in the coming days so apologies for the delays

pmoravec:
From my code review and testing on `podman` and `docker`, I can ACK it.

About the Dan's comment:
It makes sense to extend this PR to `rabbitmq` plugin as well, since it calls `docker` commands and work with containers like other updated plugins.

For the plugins timeouts:
- that is a separate issue (podman bug) that we can workaround by adding `timeout=0` to each(?) `podman` call in relevant plugins - could you @danalsan raise an issue/PR for it (or state what particular cmds needs such change)?
- why `rabbitmq` timeouts, as it does not call podman command? By hanging command `timeout 300s docker exec -t rabbitmq-bundle-podman-0 rabbitmqctl report` ? So are `docker` calls affected by the same problem like https://bugzilla.redhat.com/show_bug.cgi?id=1732525 ?
- do I interpret the `ps` output that both `docker` and `podman` are deployed on that system? My understanding from this discussion is this is not expected (if it is expected, current PR wont detect both, as `ContainerRuntime` sets `binary` to either of `docker` or `podman` but cant call both).

TurboTurtle:
If both are present, then the policy-default one will be used whenever a specific runtime isn't specified. For RHEL systems, this default is `podman`.

I'm hesitant to remove `timeout` from _all_ `podman` commands, even as a workaround to a bug within podman - and especially if we see the same behavior from `docker` commands, which shouldn't be subject to the same underlying root cause (i.e. if `docker` is also failing, then we most likely have something weird going on on the host, in my opinion).

sos/plugins/__init__.py
TurboTurtle: Hmm, fair point on executing it anyway and collecting the failure. I'll think this over a bit more.

danalsan:
> From my code review and testing on `podman` and `docker`, I can ACK it.
> 
> About the Dan's comment:
> It makes sense to extend this PR to `rabbitmq` plugin as well, since it calls `docker` commands and work with containers like other updated plugins.
> 
> For the plugins timeouts:
> 
> * that is a separate issue (podman bug) that we can workaround by adding `timeout=0` to each(?) `podman` call in relevant plugins - could you @danalsan raise an issue/PR for it (or state what particular cmds needs such change)?

So far the only ones I mentioned is the ones I'm aware of. Until EOY I won't have time to handle this time but I'll try my best to get some time on it after that.

> * why `rabbitmq` timeouts, as it does not call podman command? By hanging command `timeout 300s docker exec -t rabbitmq-bundle-podman-0 rabbitmqctl report` ? So are `docker` calls affected by the same problem like https://bugzilla.redhat.com/show_bug.cgi?id=1732525 ?

'docker' is an alias for `podman` on OpenStack deployments, this is why you see `docker` in the commandline and it's confusing as it's really running `podman`. Sorry I should've clarified this.

```
# yum whatprovides docker
podman-docker-1.0.0-2.git921f98f.module+el8.0.0+2958+4e823551.noarch : Emulate Docker CLI using podman
```


> * do I interpret the `ps` output that both `docker` and `podman` are deployed on that system? My understanding from this discussion is this is not expected (if it is expected, current PR wont detect both, as `ContainerRuntime` sets `binary` to either of `docker` or `podman` but cant call both).
ditto

TurboTurtle:
I had pushed an update to this patchset that added changes to `rabbitmq` which dropped the independent `logs` collection. After thinking this over a bit more, I don't think that is the right move - so I pushed a second update which re-added that, via an added `Plugin.get_container_logs()` method.

Also, further fixed up the docstring messages to actually make sense this time.

pmoravec:
> 'docker' is an alias for `podman` on OpenStack deployments, this is why you see `docker` in the commandline and it's confusing as it's really running `podman`. Sorry I should've clarified this.

Can't this be an issue for the https://github.com/sosreport/sos/pull/1873/files#diff-c281e660b4e6e9273cdbb534ba04dc36R973 code where we decide the container runtime implementation? But till this PR works well on OpenStack, I am fine with this.

danalsan:
> > 'docker' is an alias for `podman` on OpenStack deployments, this is why you see `docker` in the commandline and it's confusing as it's really running `podman`. Sorry I should've clarified this.
> 
> Can't this be an issue for the https://github.com/sosreport/sos/pull/1873/files#diff-c281e660b4e6e9273cdbb534ba04dc36R973 code where we decide the container runtime implementation? But till this PR works well on OpenStack, I am fine with this.

Right, this is why I flipped the order on the ovn_central plugin and checked for 'podman' first at
https://github.com/sosreport/sos/pull/1873/files#diff-7e5b784f4c6844be676647d273762167L66

bmr-cymru:
I think this is close and it passes basic smoke testing for me (rebased on current master), but I think it's too big a change to take at the last minute before 3.9 - we can finish review once the release is out and distributions that want to take on the functionality now can take a patch from master once it's merged.

TurboTurtle:
Latest update is rebasing on current master following the 3.9 tag.

PR #1875, opened 2019-12-04
[ovn_central] call podman exec without a timeout

pmoravec:
This is a workaround fix of a podman bug (see rhbz1732525) where "podman ps"
can hang when "podman exec .." is invoked in detached mode under "timeout".

Calling it without timeout works fine.

This commit can be reverted once the podman bug is fixed.

Resolves: #1875

---

### DIFF
--- a/sos/plugins/ovn_central.py
+++ b/sos/plugins/ovn_central.py
@@ -26,7 +26,8 @@ class OVNCentral(Plugin):
         if self._container_name:
             cmd = "%s exec %s cat %s" % (
                 self._container_runtime, self._container_name, filename)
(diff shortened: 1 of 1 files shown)


pmoravec:
Credit goes to @bmr-cymru .

@danalsan , could you please verify this really works (cf https://github.com/danalsan/sos/commit/61262f347a685489336646d23f5aeb73a9c55aa3)?

danalsan:
Yes, we tried this together in my setup and it works. Thanks a lot folks!

pmoravec:
FYI there would be a merge conflict between this and #1873 (trivial to resolve, though).

PR #1880, opened 2019-12-05
[pulp] Fix invalid syntax in collected json files

### LINKED ISSUE #1878: [pulp] Json syntax is broken while replacing sensitive data
Two Issues : 

* While collecting etc/pulp/server/plugins.conf.d/*json files., Sensitive data is replaced by ******* which is an invalid json value.
```
[root@sat65 sosreport-sat65-2019-12-05-rvpjhzq]# cat etc/pulp/server/plugins.conf.d/docker_importer.json 
{
    
    "proxy_host": null,
    
    
    "proxy_port": null,
    
    
    "proxy_username": null,
    
    
    "proxy_password": ********
    
}
[root@sat65 sosreport-sat65-2019-12-05-rvpjhzq]# json_reformat < etc/pulp/server/plugins.conf.d/docker_importer.json
lexical error: invalid char in json text.
                                        {          "proxy_host": null,
                     (right here) ------^
```

It should be replaced by some valid json value. Porbbaly some text with quotes like "*******" or "Filtered"

* regex also consumes the "," which also breaks the json syntax
```
cat etc/pulp/server/plugins.conf.d/yum_importer.json
{
    
    "proxy_host": null,
    
    
    "proxy_port": null,
    
    
    "proxy_username": null,
    
    
    "proxy_password": ********   <=== Regex consumed , from here
    
    
    "max_speed": null,
    
    "gpg_sign_metadata": false
}
```

Rohan21Lobo:
Plugin replaces sensitive data in etc/pulp/server/plugins.conf.d/*json
with ******* which is an invalid json value. Also it takes away a ","
and collected file is no longer a valid json. This change modifies the
regex used to get a valid json file.

Also Fixed Regex above it which was consuming any file having "conf"
even in middle of file path.

Closes #1878

Signed-off-by: Rohan Arora <roarora@redhat.com>

---

- [ X] If this commit closes an existing issue, is the line `Closes: #ISSUENUMBER` included in an independent line?

### DIFF
--- a/sos/plugins/pulp.py
+++ b/sos/plugins/pulp.py
@@ -136,10 +136,20 @@ class Pulp(Plugin, RedHatPlugin):
         return _cmd % quote(_moncmd % (_mondb, query))
 
     def postproc(self):
+
+        # Handle all ".conf" files under /etc/pulp - note that this includes
+        # files nested at several distinct directory levels. For this reason we
+        # use a regex that matches all these path components with ".*", and
+        # ensure that the path ends with ".conf".
         etcreg = r"(([a-z].*(passw|token|cred|secret).*)\:(\s))(.*)"
         repl = r"\1 ********"
-        self.do_path_regex_sub("/etc/pulp/(.*).conf", etcreg, repl)
-        jreg = r"(\s*\".*(passw|cred|token|secret).*\:)(.*)"
+        self.do_path_regex_sub(r"/etc/pulp/(.*)\.conf$", etcreg, repl)
+
+        # Now handle JSON-formatted data in the same /etc/pulp directory
+        # structure. We use a different substitution string here to preserve
+        # the file's JSON syntax.
+        jreg = r"(\s*\".*(passw|cred|token|secret).*\"\s*:\s*\")(.*)(\")"
+        repl = r"\1********\4"
         self.do_path_regex_sub("/etc/pulp(.*)(.json$)", jreg, repl)
 
 # vim: set et ts=4 sw=4 :


bmr-cymru:
Isn't the only problem here the /quoting/ of the string in the resulting JSON? We have a few other places where we post-process JSON data and I don't think we have had any complaints about those. Rather than invent a new string for obfuscation I think I would prefer if we continued to use the asterisks, just ensuring that they are quoted as a proper JSON string value.

If we are going to move to something that makes it more clear that SOS has modified the data then I think we should consider making that available via `sos.plugins` either as a constant or e.g. a function to return a "redaction" string of a certain length.

TurboTurtle:
For reference, we currently do a `******` replacement in the `docker` plugin for JSON-formatted output that simply re-adds the closing quote to preserve the formatting for the same reason this issue is trying to tackle - invalid JSON files.

https://github.com/sosreport/sos/blob/master/sos/plugins/docker.py#L126#L127

FWIW, I too _really_ don't like the `SOS_FILTERED` bare string. If we're going to break away from the **** strings we've been using for a while now, we should make it a standardized string that all plugins then use.

bmr-cymru:
> FWIW, I too really don't like the SOS_FILTERED bare string. If we're going to break away from the **** strings we've been using for a while now, we should make it a standardized string that all plugins then use.

Agreed. I think everyone gets the "********" as redaction - if we move to a new string then we need to both standardise it, and make sure the intent is equally clear while avoiding adding any new kinds of syntax problems in the files we treat this way.

Rohan21Lobo:
You are right, the real problem was the removal of quotes. I just realized my new regex in this PR already preserves the quotes. I can replace SOS_FILTERED back with *******
The only difference than previous will be earlier it used to capture any value after " : " even null which is default value in those pulp files and replace it with *s. Now it won't replace null which we should be bothered about. and will replace only string with in double quotes.
@pmoravec  @bmr-cymru  I will replace once you confirm , I already tested this BTW.

Rohan21Lobo:
Repushed with ********S
@pmoravec  @bmr-cymru  Please review

TurboTurtle:
Closed via  dd5d421

PR #1883, opened 2019-12-09
[foreman] Collect debugging info for Dynflow Sidekiq

pmoravec:
Collect sidekiq logs and services lists and statuses.

Resolves: #1883

---

### DIFF
--- a/sos/plugins/foreman.py
+++ b/sos/plugins/foreman.py
@@ -139,6 +139,13 @@ class Foreman(Plugin):
             'ping -c1 -W1 localhost'
         ])
 
+        # Dynflow Sidekiq
+        self.add_cmd_output('systemctl list-units dynflow*',
+                            suggest_filename='dynflow_units')
+        self.add_service_status('"system-dynflow\\x2dsidekiq.slice"',
+                                suggest_filename='dynflow_sidekiq_status')
+        self.add_journal(units="dynflow-sidekiq@*")
+
         # collect tables sizes, ordered
         _cmd = self.build_query_cmd(
             "SELECT table_name, pg_size_pretty(total_bytes) AS total, "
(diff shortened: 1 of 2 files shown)


pmoravec:
This is in par with `foreman-debug` stuff in https://github.com/theforeman/foreman/pull/7192 (`/etc/foreman/dynflow` is already collected via collecting `/etcforeman`).

pmoravec:
The `"system-dynflow\\x2dsidekiq.slice"` is really required, as the `systemd` unit name contains the "\x2d", not the "-".

sos/plugins/foreman.py
```diff
@@ -139,6 +139,14 @@ def setup(self):
             'ping -c1 -W1 localhost'
         ])
 
+        # Dynflow Sidekiq
+        self.add_cmd_output('systemctl list-units dynflow*',
+                            suggest_filename='dynflow_units')
+        self.add_cmd_output('systemctl status '
+                            '"system-dynflow\\x2dsidekiq.slice"',
+                            suggest_filename='dynflow_sidekiq_status')
```
TurboTurtle: The status request can be handled by the `Plugin.add_service_status()` method that hooks into the `InitSystem` abstraction. Though it does look like we'll need to plumb `**kwargs` through so `suggest_filename` can be passed properly.

pmoravec:
Good point, `add_service_status` now supports all `kwargs`. That causes a slight difference in behaviour: originally `add_service_status` called the command with `timeout=None`, now it calls with the default timeout. But I see this as an improvement rather than a regression. (removal of `pred=None` does not have an effect since that is the predicate default already).

pmoravec:
Just tested on upstream foreman, works fine.

PR #1886, opened 2019-12-10
[java] Only run java conditionally and collect more

BryanQuigley:
As this has been run unconditionaly to this point I choose to
do both commands = java, files=/usr/bin/java and packages.
There are many times Java is installed to /opt/ for say IBM's
java or Oracle's Java, or someone else's java.  To the same
 extent there are many different packages that include java
so do /usr/bin/java in case we miss some.

Add * to /etc/java*/ as some of the names can be different like
/etc/java-13-openjdk/

Remove alternatives from root symlink - I couldn't find a
reason for it to be in root, but happy to revisit.  (doesn't
work on Ubuntu)

Add java --version.  If it's in path we should at least be able
to get this.

---

### DIFF
--- a/sos/plugins/java.py
+++ b/sos/plugins/java.py
@@ -15,13 +15,16 @@ class Java(Plugin, RedHatPlugin, UbuntuPlugin, DebianPlugin):
     plugin_name = "java"
     profiles = ('webserver', 'java')
     verify_packages = ('java.*',)
+    commands = ('java',)
+    files = ('/usr/bin/java',)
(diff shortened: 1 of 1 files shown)


BryanQuigley:
Tried teston centos7 in an LXD container, but java --version didn't work.  Will investigate further.

BryanQuigley:
/me realized what the --version issue was it should have been -version (apparently Ubuntu has a modification so both work).

Reconfirmed tests on centos/8 with the change

PR #1888, opened 2019-12-11
[ceph] Add 'ceph insights' command output

badone:
This change adds the output of the recently added 'ceph insights'
command.

Signed-off-by: Brad Hubbard <bhubbard@redhat.com>

---


badone:
@pmoravec done. Seems odd to me the commit message needs to include information that is unknowable when the commit is first created, but I'm sure you have your reasons. Thanks!

bmr-cymru:
It's not unknowable - but if you want to set it before you create the PR you need to look at the current highest numbered PR/issue, and add one - we've never had a colission so far ;-)

It's something the regular contributors do, but it's no big deal for us to add it while we merge. It just helps to keep everything in sync without the need to manually close things out later.

PR #1890, opened 2019-12-11
[archive] handle OSErrors

jjansky1:
If file/folder will be present during file list creation, but will be
missing during collection OSError will throw exception

FileNotFoundError: [Errno 2] No such file or directory

Added handling of OSError into sos/archive.py and new log message with
info level to sos_logs/sos.log

---


sos/archive.py
```diff
@@ -342,12 +342,14 @@ def add_file(self, src, dest=None):
                 # path case
                 try:
                     shutil.copy(src, dest)
-                except IOError as e:
+                except (IOError, OSError) as e:
```
pmoravec: I would rather see separate "except OSError" for two reasons:

- with the current change, a fail to copy `/sys/*` or `/proc/*` wont be logged
-for other paths and `OSError`, we get two info logs, both the `caught` and `File not collected`

bmr-cymru:
Please also follow the [contributor guidelines](https://github.com/sosreport/sos/wiki/Contribution-Guidelines) for patch formatting: the subject needs to be in `[<component>] change description` format. For this patch it should be something like `[archive] handle FileNotFoundError in Archive.add_file()`.

sos/archive.py
bmr-cymru: Agreed.

sos/archive.py
```diff
@@ -342,12 +342,14 @@ def add_file(self, src, dest=None):
...
                     # Filter out IO errors on virtual file systems.
                     if src.startswith("/sys/") or src.startswith("/proc/"):
                         pass
                     else:
                         self.log_info("caught '%s' copying '%s'" % (e, src))
+                    if e.errno == 2:
```
bmr-cymru: Please use a constant from the `errno` module for this; it's already imported in `archive` so you can just replace the 2 with `errno.ENOENT`.

sos/archive.py
```diff
@@ -342,12 +342,14 @@ def add_file(self, src, dest=None):
...
+                        self.log_info("File not collected: '%s' - Reason: '%s'" % (src, e))
```
bmr-cymru: This would lead to messages like:

```
2019-12-05 14:20:50,159 INFO: [sos.archive:add_file] File not collected: '/path/not/found' - Reason: '[Errno 2] No such file or directory: '/path/not/found'
```

Which seems a bit longwinded/redundant. It's fine to just use the exception string as the message since it already includes the path.

sos/archive.py
jjansky1: IOError and OSError separated

sos/archive.py
jjansky1: errno removed as all OSErrors should be logged

sos/archive.py
jjansky1: Removed redundant message from log

PR #1894, opened 2019-12-16
[ctdb] Improve CTDB plugin

anoopcs9:
* Include main CTDB configuration and additional files
* Include list of enabled legacy event scripts

Resolves: #1894 

Signed-off-by: Anoop C S <anoopcs@redhat.com>

---

### DIFF
--- a/sos/plugins/ctdb.py
+++ b/sos/plugins/ctdb.py
@@ -20,6 +20,8 @@ class Ctdb(Plugin, DebianPlugin, UbuntuPlugin):
 
     def setup(self):
         self.add_copy_spec([
+            "/etc/ctdb/ctdb.conf",
+            "/etc/ctdb/*.options",
             "/etc/ctdb/nodes",
             "/etc/ctdb/public_addresses",
             "/etc/ctdb/static-routes",
@@ -35,7 +37,8 @@ class Ctdb(Plugin, DebianPlugin, UbuntuPlugin):
             "ctdb listnodes",
             "ctdb listvars",
             "ctdb statistics",
-            "ctdb getdbmap"
+            "ctdb getdbmap",
+            "ctdb event script list legacy"
         ])
 
 


anoopcs9:
> ACK.
> 
> It would be great to amend commit message by adding "Resolves: #1894".

Updated.

pmoravec:
Thanks!

TurboTurtle:
Code change looks OK - but are there any sensitive data points in `ctdb.conf` or the `*.options` files? Passwords of any kind, certificate contents, etc...?

anoopcs9:
> Code change looks OK - but are there any sensitive data points in `ctdb.conf` or the `*.options` files? Passwords of any kind, certificate contents, etc...?

I don't think we have anything sensitive in nature inside ctdb.conf or *.options files under /etc/ctdb/. But I will let @amitay or @martin-schwenke have a comment before we approve this change.

gd:
Yeah, also can't think of any sensitive data at all in these files.

martin-schwenke:
At the moment there is nothing sensitive in the configuration files.

In the future we have vague plans to add support for a shared secret (or similar) that will be used to authenticate connections between nodes.  We can obviously put this secret in a different file.

pmoravec:
> At the moment there is nothing sensitive in the configuration files.
> 
> In the future we have vague plans to add support for a shared secret (or similar) that will be used to authenticate connections between nodes. We can obviously put this secret in a different file.

I see few options here:

- if you know the format / name of that option already (I dont expect so), we can start obfuscating it even now
- once you start having the secret in this file (or know the format/name of the option), we can amend the plugin to obfuscate it
- or leave the secret in a different file

TurboTurtle:
LGTM, thanks for the sensitive data check.

Like @pmoravec mentions, we can always add postprocessing to the plugin for when secrets might be added to these files.

anoopcs9:
Is there any pending task that needs to be completed here?

PR #1895, opened 2019-12-17
[memory] Collect slabtop info

TurboTurtle:
Adds collection of `slabtop -o` to show the top consumers of SLAB.

Note that currently there is an issue with this command that will cause
it to hang if stdin does not receive a tty - which is exactly what
happens when commands are run via `timeout`. As such, this command is
for the moment being run with `timeout=None`, and is run in a separate
`add_cmd_output()` call at the end of the plugin.

Resolves: #1895

---

### DIFF
--- a/sos/plugins/memory.py
+++ b/sos/plugins/memory.py
@@ -36,4 +36,7 @@ class Memory(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
             "lsmem -a -o RANGE,SIZE,STATE,REMOVABLE,ZONES,NODE,BLOCK"
         ])
 
+        # slabtop -o will hang if not handed a tty via stdin
+        self.add_cmd_output("slabtop -o", foreground=True)
+
 # vim: set et ts=4 sw=4 :


BryanQuigley:
Is this helpful enough over /proc/slabinfo to warrant a possible hang?

TurboTurtle:
With just `/proc/slabinfo`, it's up to the person looking at it to manually grok through it to determine the entities with the highest usage - which can be tedious at best.

I think an actual hang when using `timeout=None` on the command is highly unlikely, and in normal use cases the command completes in less than a second. If a system is in such a state that `slabtop` hangs, I'd honestly be surprised if an sosreport run got far enough along to hit that point.

Beyond that though, the command is being called at the end of the plugin, so the other collections will have already completed and the whole plugin timeout would kick in and terminate the plugin thread that was running the hung `slabtop` command. So I think we're safe from several angles.

bmr-cymru:
OK so we now have a second case of "hangs forever without TTY" (the first being `podman exec`). It may be worth considering adding a `foreground` keyword argument here to allow the command  under timeout to be run with TTY access.

Is there a bug against `slabtop` for this problem?

TurboTurtle:
I've opened https://bugzilla.redhat.com/show_bug.cgi?id=1784585 for this on our end.

sos/plugins/memory.py
```diff
@@ -36,4 +36,8 @@ def setup(self):
             "lsmem -a -o RANGE,SIZE,STATE,REMOVABLE,ZONES,NODE,BLOCK"
         ])
 
+        # slabtop -o will hang if not handed a tty via stdin and our timeout
+        # handling removes the tty for any command run with it
+        self.add_cmd_output("slabtop -o", timeout=None)
```
pmoravec: Since #1909 is merged, is it worth changing `timeout=None` to `foreground=True`?

TurboTurtle: Yep, was either going to update this one once #1909 got merged or vice versa. Will update this PR later today.

PR #1896, opened 2019-12-17
[travis] Move to single test script and use travis job matrix

BryanQuigley:
Move all the sosreport runs to a simple bash script:
tests/simple.sh so it can be called multiple times.
This also makes it possible to run locally.

The travis job matrix let's us have 2 Ubuntu native sos
runs against installed python, while letting us to also
continue to do the old style python runs.  Ideally,
we can utilize this to do Fedora/CentOS/RHEL jobs in the
future.

Python 3.4 testing was also dropped - as it seems to not
be as well used (many test failures due to not being available).
Python 3.8 testing was added.
pycodestyle will only run once with the bionic version.

Resolves: #1896

---

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,64 @@
-language: python
-python:
-  - "2.7"
-  - "3.4"
-  - "3.5"
-  - "3.6"
-  - "3.7"
-  - "nightly"
-matrix:
-  allow_failures:
-    - python: "nightly"
+jobs:
+  include:
+    - name: "18.04 pycodestyle and native run (py3.6)"
+      os: linux
+      dist: bionic
+      language: shell
+      install: sudo apt-get update; sudo apt install pycodestyle;
+      script:
+        - "pycodestyle sos"
+        - "sudo ./tests/simple.sh"
+    - name: "16.04 native run (py3.5)"
+      os: linux
+      dist: xenial
+      language: shell
+      script: "sudo ./tests/simple.sh"
+    - name: "nosetests and travis Python 2.7"
+      os: linux
+      dist: bionic
+      language: python
+      python: "2.7"
+      install: pip install -r requirements.txt; python setup.py install;
+      script:
+        - "nosetests -v --with-cover --cover-package=sos --cover-html"
+        - "sudo ./tests/simple.sh ~/virtualenv/python$TRAVIS_PYTHON_VERSION/bin/python"
+    - name: "nosetests and travis Python 3.5"
+      os: linux
+      dist: bionic
+      language: python
+      python: "3.5"
+      install: pip install -r requirements.txt; python setup.py install;
+      script:
+        - "nosetests -v --with-cover --cover-package=sos --cover-html"
+        - "sudo ./tests/simple.sh ~/virtualenv/python$TRAVIS_PYTHON_VERSION/bin/python"
+    - name: "nosetests and travis Python 3.6"
+      os: linux
+      dist: bionic
+      language: python
+      python: "3.6"
+      install: pip install -r requirements.txt; python setup.py install;
+      script:
+        - "nosetests -v --with-cover --cover-package=sos --cover-html"
+        - "sudo ./tests/simple.sh ~/virtualenv/python$TRAVIS_PYTHON_VERSION/bin/python"
+    - name: "nosetests and travis Python 3.7"
+      os: linux
+      dist: bionic
+      language: python
+      python: "3.7"
+      install: pip install -r requirements.txt; python setup.py install;
+      script:
+        - "nosetests -v --with-cover --cover-package=sos --cover-html"
+        - "sudo ./tests/simple.sh ~/virtualenv/python$TRAVIS_PYTHON_VERSION/bin/python"
+    - name: "nosetests and travis Python 3.8"
+      os: linux
+      dist: bionic
+      language: python
+      python: "3.8"
+      install: pip install -r requirements.txt; python setup.py install;
+      script:
+        - "nosetests -v --with-cover --cover-package=sos --cover-html"
+        - "sudo ./tests/simple.sh ~/virtualenv/python$TRAVIS_PYTHON_VERSION/bin/python"
+
 notifications:
   email:
     sos-devel@redhat.com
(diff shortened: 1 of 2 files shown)


tests/simple.sh
```diff
@@ -0,0 +1,87 @@
...
+
+PYTHON=${1:-/usr/bin/python3}
+SOSPATH=${2:-./sosreport}
+
+NUMOFFAILURES=0
+
+run_expecting_sucess () {
+FAIL=false
+#Make sure clean
+rm stderr stdout
```
pmoravec: Could we have `rm -f` here, please? To prevent potential prompts on RHEL where `rm` is an alias to `'rm -i'`.

tests/simple.sh
```diff
@@ -0,0 +1,87 @@
...
+
+echo "######### DONE WITH $1 #########"
+
+if $FAIL; then 
+  return 1
+else
+  return 0
+fi
+
+}
```
pmoravec: It would be great to have indentation of the commands in the `run_expecting_success()` method - it is bit confusing to understand what is the method's code and what is the script's main body.

tests/simple.sh
```diff
@@ -0,0 +1,87 @@
...
+NUMOFFAILURES=$(($NUMOFFAILURES + $?))
+run_expecting_sucess " --batch"
+NUMOFFAILURES=$(($NUMOFFAILURES + $?))
+run_expecting_sucess " --batch -a --case-id 1010101 -c never --label TESTLABEL"
+NUMOFFAILURES=$(($NUMOFFAILURES + $?))
+run_expecting_sucess " --batch --debug --experimental --quiet -z gzip"
+NUMOFFAILURES=$(($NUMOFFAILURES + $?))
+run_expecting_sucess " --batch --no-report --no-env-vars --build --verbose -z bzip2"
+NUMOFFAILURES=$(($NUMOFFAILURES + $?))
+run_expecting_sucess " --batch -z xz -t 1 --log-size 1 -n lxd" #This fails with just lxd, why?
```
pmoravec: This fails on Red Hat distro, as `lxd` plugin is not a `RedHatPlugin` (cf. https://github.com/sosreport/sos/blob/master/sos/plugins/lxd.py#L14). Could you use some more generic plugin?

.travis.yml
```diff
@@ -1,37 +1,73 @@
...
-matrix:
-  allow_failures:
-    - python: "nightly"
+jobs:
+  include:
+    - name: "18.04 pycodestyle and native run (py3.6)"
+      os: linux
+      dist: bionic
+      language: shell
+      install: sudo apt-get update; sudo apt install pycodestyle;
```
pmoravec: A Travis newbie question: how Travis detects this job to run on Ubuntu and not Fedora/RHEL?

Also, as this works on Ubuntu but not Red Hat distros, will a Red Hat distro run `pycodestyle`? (but I expect `pycodestyle` will provide same output on either distro).

BryanQuigley: Travis only supports Ubuntu natively AFAICT for os:linux, if they do add another I expect the dist: line would be how to change it.

The Travis way run another os would be to use docker services - which I plan to investigate after this commit.

tests/simple.sh
BryanQuigley: I've opened a bug to track it #1921 .. but I did add a workaround for Travis. For whatever reason travis, the lxd plugin and --log-size don't get along.

BryanQuigley:
Sorry for the noise - debugging Travis isn't particularly fun.  As part of that I added an attempt to extract archives and cat any backtraces from them.  This helped me with the lxd/travis/--log-size issue.

tests/simple.sh
```diff
@@ -0,0 +1,132 @@
...
+    else
+      return 0
+    fi
+}
+
+# We aren't trying to catch errors here, this just let's us display the output
+archive_work() {
+    rm -f /tmp/sosreport*md5
+    mkdir /tmp/sosreport_test/
+    tar xfa /tmp/sosreport*.tar* -C /tmp/sosreport_test --strip-components=1
```
pmoravec: Will this work also for a test case with `--build`, where the `sosreport` execution output is unpacked directory ("only")?

As this procedure is called after `run_expecting_success` one, everytime, doesnt it make sense to merge the procedures (i.e. print the plugin errors after `stdout/stderr`)?

tests/simple.sh
```diff
@@ -0,0 +1,132 @@
...
+if [ -f /etc/sos.conf ]; then
+   echo "/etc/sos.conf already exists"
+else
+   echo "Creating /etc/sos.conf"
+   touch /etc/sos.conf
+fi
+
+# Runs not generating sosreports
+run_expecting_sucess " -l"
+NUMOFFAILURES=$(($NUMOFFAILURES + $?))
```
pmoravec: Until we implement some concurrency inside the script, isn't it easier the `NUMOFFAILURES` is updated directly inside the `run_expecting_success` procedure?

Likewise with the `archive_work` procedure, we will have much shorter code with individual test cases being run - easier to follow, not such error-prone when one forgets to e.g. update this variable.

pmoravec:
As a Travis newbie, I can't assess pros or cons of the PR, so just commenting the code itself. Apart of those comments: can't be the individual jobs "merged" using some template? As there is repetitive:


    - name: "nosetests and travis Python 3.5"
      os: linux
      dist: bionic
      language: python
      python: "3.5"
      install: pip install -r requirements.txt; python setup.py install;
      script:
        - "nosetests -v --with-cover --cover-package=sos --cover-html"
        - "sudo ./tests/simple.sh ~/virtualenv/python$TRAVIS_PYTHON_VERSION/bin/python"

text placed there 5 times, that differs just in python version. (But it is quite possible this can't be easily resolved using some template).

BryanQuigley:
> text placed there 5 times, that differs just in python version. (But it is quite possible this can't be easily resolved using some template).

I don't have a better solution at this time.  I'm hopeful we will replace those in the next iteration with RHEL/CentOS/Fedora tests.

tests/simple.sh
BryanQuigley: Updating NUMOFFAILURES in the procedure appears to make things much nicer, will redo and resubmit.  

Will evaluate archivework next..

PR #1899, opened 2019-12-18
[foreman] increase sizelimit for tasks export

pmoravec:
When foreman stores tens to hundreds of thousands of tasks, default
sizelimit causes the dynflow* or foreman_tasks_tasks files are truncated.

Let increase the sizelimit to 100MB.

Resolves: #1899

---

### DIFF
--- a/sos/plugins/foreman.py
+++ b/sos/plugins/foreman.py
@@ -218,12 +218,12 @@ class Foreman(Plugin):
         for table in foremandb:
             _cmd = self.build_query_cmd(foremandb[table])
             self.add_cmd_output(_cmd, suggest_filename=table, timeout=600,
-                                env=self.env)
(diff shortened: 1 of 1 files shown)


BryanQuigley:
I thought our default sizelimit was around that?   I'd actually like to explicitly document it so we are more clear.

TurboTurtle:
It's 100MB for `journalctl` output - 25MB for everything else by default.

bmr-cymru:
> I thought our default sizelimit was around that? I'd actually like to explicitly document it so we are more clear.

If we are going to document it then let's make sure it's via an automated mechanism - it should be possible for the build scripts to update the documentation based on values pulled from the code.

pmoravec:
#1900 raised for the doc issue.

PR #1906, opened 2019-12-31
[ovirt_hosted_engine] Collect rotated HA logs by default

TurboTurtle:
Updates the plugin to collect the rotated agent and broker HA logs by
default. These logs are small and can be rotated faster than other ovirt
components, leading to cases where HA logs do not go back far enough
even though we capture a lot of historical data from VDSM and other
components.

Now we will collect the rotated logs up until the `--log-size` threshold
is reached. Users may use `--all-logs` to remove this size restriction
as usual.

Related: RHBZ#1787119
Resolves: #1906

---


didib:
Looks good to me. Thanks.

PR #1902, opened 2019-12-23
[Policy] Wrap json.load() in with clause

TurboTurtle:
Wraps naked opens for `json.load()` calls in `with` statements to ensure
the file object gets closed properly.

Closes: #1322
Resolves: #1902

Signed-off-by: Jake Hunsaker <jhunsake@redhat.com>

---
Please place an 'X' inside each '[]' to confirm you adhere to our [Contributor Guidelines](https://github.com/sosreport/sos/wiki/Contribution-Guidelines)

- [x] Is the commit message split over multiple lines and hard-wrapped at 72 characters?
- [x] Is the subject and message clear and concise?
- [x] Does the subject start with **[plugin_name]** if submitting a plugin patch or a **[section_name]** if part of the core sosreport code?
- [x] Does the commit contain a **Signed-off-by: First Lastname <email@example.com>**?
- [x] If this commit closes an existing issue, is the line `Closes: #ISSUENUMBER` included in an independent line?
- [x] If this commit resolves an existing pull request, is the line `Resolves: #PRNUMBER` included in an independent line?

PR #1907, opened 2020-01-06
[ebpf] Add new plugin for eBPF commands

jcastill:
This plugin groups all bpftool commands from
plugins kernel and networking, and adds a new
one command 'bpftool cgroup tree', as requested
by rhbz#1787586.

---

### DIFF
--- a/sos/plugins/ebpf.py
+++ b/sos/plugins/ebpf.py
@@ -1,0 +1,78 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+import json
+
+
+class eBPF(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
+
+    plugin_name = 'ebpf'
+    profiles = ('system', 'kernel', 'network')
+
+    def get_bpftool_prog_ids(self, prog_json):
+        out = []
+        try:
+            prog_data = json.loads(prog_json)
+        except Exception as e:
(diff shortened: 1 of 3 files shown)


sos/plugins/ebpf.py
```diff
@@ -0,0 +1,80 @@
...
+
+class eBPF(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
+
+    plugin_name = 'ebpf'
+    profiles = ('system', 'kernel', 'network')
+
+    def get_bpftool_prog_ids(self, prog_file):
+        out = []
+        try:
+            prog_data = json.load(open(prog_file))
```
TurboTurtle: This will currently not function, as described in the second commit in #1902 - `exec_cmd()` returns json output, but here we're trying to open it the string as a file bject.

sos/plugins/ebpf.py
```diff
@@ -0,0 +1,80 @@
...
+            return out
+        for item in range(len(prog_data)):
+            if "id" in prog_data[item]:
+                out.append(prog_data[item]["id"])
+        return out
+
+    def get_bpftool_map_ids(self, map_file):
+        out = []
+        try:
+            map_data = json.load(open(map_file))
```
TurboTurtle: Same here

bmr-cymru:
Why is it desirable to have these all available in a single plugin, rather than (as now) grouped with other related commands/data collection? It's really not obvious to me from the change why we wouldn't want to keep the networking specifics in the networking plugin, for e.g.

If it's about shared infrastructure then that may suggest that those interfaces should be added to either the utilities module or the base Plugin class so that they are available in multple plugins.

TurboTurtle:
I generally agree with @jcastill on this one. We've shuffled the commands about twice (I think) already, and the tool has enough coverage that it could very well also spill into other plugins in the future. We carry `pcp`, `sar`, and a few others in separate plugins, and I agree that this tool is similar enough to warrant it - plus, it would likely be easier to maintain going forward keeping all the tool's collection in a single plugin rather than sprinkled throughout other, more generalized, plugins.

pmoravec:
I am inclined to a new plugin, expecting the bpf* stuff will grow over time and span to other domains.

PR #1909, opened 2020-01-07
[utilities] Run timeout with --foreground

TurboTurtle:
This is a two-patch set. 

The first commit makes the use of `--foreground` in our `timeout` command the default.
The second commit updates the `ovn_central` plugin to no longer use the `timeout=None` workaround.

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -1043,7 +1043,7 @@ class Plugin(object):
                        root_symlink=None, timeout=300, stderr=True,
                        chroot=True, runat=None, env=None, binary=False,
                        sizelimit=None, pred=None, subdir=None,
-                       changes=False):
+                       changes=False, foreground=False):
         """Run a program or a list of programs and collect the output"""
         if isinstance(cmds, six.string_types):
             cmds = [cmds]
@@ -1059,7 +1059,7 @@ class Plugin(object):
                                  stderr=stderr, chroot=chroot, runat=runat,
                                  env=env, binary=binary, sizelimit=sizelimit,
                                  pred=pred, subdir=subdir,
-                                 changes=changes)
+                                 changes=changes, foreground=foreground)
 
     def get_cmd_output_path(self, name=None, make=True):
         """Return a path into which this module should store collected
@@ -1149,7 +1149,7 @@ class Plugin(object):
                             root_symlink=False, timeout=300, stderr=True,
                             chroot=True, runat=None, env=None,
                             binary=False, sizelimit=None, subdir=None,
-                            changes=False):
+                            changes=False, foreground=False):
         """Execute a command and save the output to a file for inclusion in the
         report.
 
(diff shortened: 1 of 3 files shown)


pmoravec:
Can't there be a problem that:

      --foreground
                 ..
                   in this mode, children of COMMAND will not be timed out

could cause children of the invoked command will hung (though they wont block sosreport execution)?

Is it OK to leave those hung child processes after sosreport kills a timeouted command?

TurboTurtle:
So far in my testing, I've not encountered that as an issue even when using low timeouts - but it could very well be a concern that I'm simply not encountering/reproducing properly.

I *think* this is a safe thing to do, but we'll definitely want some wide testing done on this. For the moment, this does remove the need to run without a timeout for the `ovn_central` (and in the future, any runtime inspection via #1873) and the same need from the proposed #1895.

pmoravec:
I did a test:

    # cat /root/parent_sleep.sh
    /root/child_sleep.sh
    # cat /root/child_sleep.sh
    sleep 200
    date
    #

Made both scripts executable and added the `/root/parent_sleep.sh` to `add_cmd_output` with `timeout=21`.

After starting sosreport, I saw in `ps aux`:

    root     18967  0.0  0.0   8640   952 pts/0    S+   14:33   0:00 timeout --foreground 21s /root/parent_sleep.sh
    root     18969  0.0  0.1   9776  2708 pts/0    S+   14:33   0:00 /bin/sh /root/parent_sleep.sh
    root     18970  0.0  0.0   9776  1648 pts/0    S+   14:33   0:00 /bin/sh /root/parent_sleep.sh
    root     18971  0.0  0.0   4360   824 pts/0    S+   14:33   0:00 sleep 200

waited 25 seconds and then I saw:

    root     18967  0.0  0.0      0     0 pts/0    Z+   14:33   0:00 [timeout] <defunct>
    root     18970  0.0  0.0   9776  1648 pts/0    S+   14:33   0:00 /bin/sh /root/parent_sleep.sh
    root     18971  0.0  0.0   4360   824 pts/0    S+   14:33   0:00 sleep 200

and verbose output of sosreport logged:

    [plugin:extras] collecting output of '/root/parent_sleep.sh'
    [plugin:extras] command '/root/parent_sleep.sh' timed out after 21s
    [plugin:extras] collected output of '/root/parent_sleep.sh' in 200.03653192520142 (changes=False)

So the timeout did not terminated the execution after 21s, but waited until the "sleep 200" completes.

I double-checked the `--foreground` option of `timeout` is responsible for this - it does not properly terminate the child processes and rather hangs.

Gladly, at least plugin timeout applies here properly and terminates the `timeout` command properly (though the child commands still run - but without further effect to `sosreport`).

I am not sure how many commands collected by `sos` could have such hanging children, but if so, the `timeout` wont work for them.

TurboTurtle:
I've updated this change to now make `--foreground` a default addition to the `timeout` command, and instead updated the `ovn_central` plugin to use `foreground=True` rather than `timeout=None`.

All other commands will continue to run without `--foreground` as before this patch.

Thanks @pmoravec for that testing, I was just modifying the `timeout` values for my tests - so while making it the default would have been ok for sos today, we may have just run into further issues in the future.

pmoravec:
> All other commands will continue to run without `--foreground` as before this patch.

I still see default `foreground=True` on few places of the patch, and still `timeout --foreground 21s /root/parent_sleep.sh` being executed /o\

Otherwise, +1 for the approach of foreground=False by default.

PR #1910, opened 2020-01-08
[Plugin] Improve predicate failure message

TurboTurtle:
Previously, the logged warning message when a predicate evaluation fails
would be a string that contained all elements passed to the predicate,
and include an empty list for checks that were not run by the predicate.
This meant that it was not immediately clear _which_ element(s) caused
the predicate evaluation to fail.

Now, during evaluation of the predicate checks, any failures are
recorded and the new failure message is generated based on these
failures. Thus, only the missing elements are presented to the user,
allowing for easier identification of what caused a particular
collection to be skipped.

Resolves: #1910

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -174,8 +174,13 @@ class SoSPredicate(object):
         if not self.kmods or self._owner.get_option('allow_system_changes'):
             return True
 
+        _kmods = []
         # Are kernel modules loaded?
-        _kmods = [self._owner.is_module_loaded(k) for k in self.kmods]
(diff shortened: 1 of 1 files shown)


TurboTurtle:
Example output with this change:

```
[plugin:docker] skipped command 'docker version': required services missing: docker.
[plugin:docker] skipped command 'docker volume ls': required services missing: docker.
[plugin:networking] skipped command 'ss -peaonmi': required kmods missing: udp_diag, netlink_diag, af_packet_diag, tcp_diag, unix_diag. Use '--allow-system-changes' to enable collection.
```

PR #1911, opened 2020-01-08
[sosreport] Allow automatic uploading to policy or user defined locations

TurboTurtle:
This patchset allows users to have sos try to upload to an FTP location (which may be defined by `Policy`) after the archive is successfully tar'd.

Users may request this by using `--upload` if the loaded `Policy` specifies a default location and credentials. Additionally, users may request a specific location using `--upload-url`. The `--upload-user`, `--upload-directory`, and `--upload-pass` options have been added to assist with this as well.

For Red Hat systems, the policy defines the long-standing public read-only dropbox location at dropbox.redhat.com for FTP uploads if a case number is not provided. If a case number is provided, users will be prompted for Customer Portal credentials and sos will then try to upload the archive directly to the case. If a case number is provided by credentials are not, sos will fallback to uploading to the public dropbox.

If uploading fails for any reason, sos will still exit the same - as this upload attempt only occurs at the end of a successful execution.

For consideration: for basic FTP uploads, this requires the `ftplib` library to be available, and for Red Hat specifically uploading to the Customer Portal requires the `requests` library. I have not added either of these as hard dependencies for sosreport as this is optional functionality that may not be used (or may not be allowed in secure environments), and so sos does not actually 'require' these libraries to perform its main objectives.

Also included in this patchset is a fix with `policies/__init__.py` where an empty input for a case number when `--case-id` has been used on the cmdline would override the cmdline option.

---

### DIFF
--- a/man/en/sosreport.1
+++ b/man/en/sosreport.1
@@ -30,6 +30,8 @@ sosreport \- Collect and package diagnostic and support data
           [-z|--compression-type method]\fR
           [--encrypt-key KEY]\fR
           [--encrypt-pass PASS]\fR
+          [--upload] [--upload-url url] [--upload-user user]\fR
+          [--upload-directory dir] [--upload-pass pass]\fR
           [--experimental]\fR
           [-h|--help]\fR
 
@@ -257,6 +259,60 @@ Execute plugins as normal, but do not collect any file content, command
 output, or string data from the system. The resulting logs may be used
 to understand the actions that sos would have taken without the dry run
 option.
+.TP
+.B \--upload
+If specified, attempt to upload the resulting archive to a vendor defined location.
+
+This option is implied if --upload-url is used.
+
+You may be prompted for a username and password if these are not defined by the vendor
+as well. If these credentials are not provided, sos will still run and create an archive
+but will not attempt an automatic upload, instead relying on the end user to upload it
+as needed.
+
+The sosreport archive will still remain on the local filesystem even after a successful
+upload.
+
+Note that depending on the distribution sos is being run on, or the vendor policy detected during
+execution, there may be dependencies that are not strictly required by the package
+at installation time.
+
+For example, for HTTPS uploads the python-requests library must be available. If this
+library is not available, HTTPS uploads will not be attempted.
+.TP
+.B \--upload-url URL
+If a vendor does not provide a default upload location, or if you would like to upload
+the archive to a different location, specify the address here.
+
+A support protocol MUST be specified in this URL. Currently uploading is supported
+for HTTPS, SFTP, and FTP protocols.
+
+If your destination server listens on a non-standard port, specify the listening
+port in the URL.
+.TP
+.B \-\-upload-user USER
+If a vendor does not provide a default user for uploading, specify the username here.
+
+If this option is unused and upload is request, and a vendor default is not set, you
+will be prompted for one. If --batch is used and this option is omitted, no username will
+be collected and thus uploads will fail if no vendor default is set.
+.TP
+.B \-\-upload-pass PASS
+Specify the password to use for authentication with the destination server.
+
+If this option is omitted and upload is requested, you will be prompted for one.
+
+If --batch is used, this prompt will not occur, so any uploads are likely to fail unless
+this option is used.
+
+Note that this will result in the plaintext string appearing in `ps` output that may
+be collected by sos and be in the archive. If a password must be provided by you
+for uploading, it is strongly recommended to not use --batch and enter the password
+when prompted rather than using this option.
+.TP
+.B \--upload-directory DIR
+Specify a directory to upload to, if one is not specified by a vendor default location
+or if your destination server does not allow writes to '/'.
 .TP
 .B \--experimental
 Enable plugins marked as experimental. Experimental plugins may not have
--- a/sos/policies/__init__.py
+++ b/sos/policies/__init__.py
@@ -9,8 +9,9 @@ import fnmatch
 import tempfile
 import random
 import string
-from pwd import getpwuid
 
+from getpass import getpass
+from pwd import getpwuid
 from sos.utilities import (ImporterHelper,
                            import_module,
                            shell_out,
@@ -1182,7 +1183,7 @@ class LinuxPolicy(Policy):
         except socket.gaierror:
             raise Exception("unable to connect to %s" % url)
         except ftplib.error_perm as err:
-            errno = err.split()[0]
+            errno = str(err).split()[0]
             if errno == 503:
                 raise Exception("could not login as '%s'" % user)
             if errno == 550:
--- a/sos/policies/__init__.py
+++ b/sos/policies/__init__.py
@@ -24,6 +24,12 @@ from six.moves import input
 
 PRESETS_PATH = "/var/lib/sos/presets"
 
+try:
+    import requests
+    REQUESTS_LOADED = True
+except ImportError:
+    REQUESTS_LOADED = False
+
 
 def get_human_readable(size, precision=2):
     # Credit to Pavan Gupta https://stackoverflow.com/questions/5194057/
@@ -836,8 +842,21 @@ class LinuxPolicy(Policy):
     vendor = "None"
     PATH = "/bin:/sbin:/usr/bin:/usr/sbin"
     init = None
-
+    # _ prefixed class attrs are used for storing any vendor-defined defaults
+    # the non-prefixed attrs are used by the upload methods, and will be set
+    # to the cmdline/config file values, if provided. If not provided, then
+    # those attrs will be set to the _ prefixed values as a fallback.
+    # TL;DR Use _upload_* for policy default values, use upload_* when wanting
+    # to actual use the value in a method/override
+    _upload_url = None
+    _upload_directory = '/'
+    _upload_user = None
+    _upload_password = None
+    _use_https_streaming = False
     _preferred_hash_name = None
+    upload_url = None
+    upload_user = None
+    upload_password = None
 
     def __init__(self, sysroot=None):
         super(LinuxPolicy, self).__init__(sysroot=sysroot)
@@ -907,12 +926,25 @@ class LinuxPolicy(Policy):
         cmdline_opts = self.commons['cmdlineopts']
         caseid = cmdline_opts.case_id if cmdline_opts.case_id else ""
 
+        # Set the cmdline settings to the class attrs that are referenced later
+        # The policy default '_' prefixed versions of these are untouched to
+        # allow fallback
+        self.upload_url = cmdline_opts.upload_url
+        self.upload_user = cmdline_opts.upload_user
+        self.upload_directory = cmdline_opts.upload_directory
+        self.upload_password = cmdline_opts.upload_pass
+
         if not cmdline_opts.batch and not \
                 cmdline_opts.quiet:
             try:
                 self.case_id = input(_("Please enter the case id "
                                        "that you are generating this "
                                        "report for [%s]: ") % caseid)
+                # Policies will need to handle the prompts for user information
+                if cmdline_opts.upload or self.upload_url:
+                    self.prompt_for_upload_user()
+                    self.prompt_for_upload_password()
+
                 self._print()
             except KeyboardInterrupt:
                 self._print()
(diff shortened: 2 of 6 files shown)


TurboTurtle:
Whoops, left those imports in from an earlier approach for the Customer Portal uploads.

Removed and re-pushed.

BryanQuigley:
Oh, I thought we didn't want to add this back to sos (always use other tools instead), I'm happy to see it being reconsidered.

FTP is actually something we explicitly don't support with our file upload server - and I think it really makes more sense for you to default to FTPS for RH.  Why default to insecure ftp (or am I misreading)?

We support SSH's SFTP and HTTPS uploads. I'll work on this tomorrow and see which is easier to implement.  Thanks for pushing this!

TurboTurtle:
I've been made aware of a small number of RH customers wondering why they need to install redhat-support-tool for this kind of functionality. While there is a lot more to that project than uploading sosreports, there are those who install it solely for uploading. Making it native to sos would eliminate the need to install that package.

RH defaults to using the API from our customer portal, FTP is a fallback. It wouldn't be a bad idea to default to FTPS though in that situation. I hadn't considered making that a default since our public dropbox is just that - a dropbox that only supports uploading. It _does_ support FTP-ES, but doesn't require it.

For SFTP, we'd probably need to shell out for that like we do for compression and encrypting. While there are python libs for SSH, I don't think we should have sos make use of any of them in particular. About the only one that I'm aware of that is somewhat actively maintained is Paramiko - which has a penchant for vulnerabilities as it is a re-implementation of the SSH protocol and its upstream is slow moving (and also focused on other similar projects from what I can tell).

pmoravec:
Using `--upload-pass` as a command line argument, won't `sosreport` collect the password when taking a `ps` output?

Could a support of a bash variable (something like `PGPASSWORD`) be added here? Or some smart obfuscation of the password from `ps` outputs (though that can be cumbersome and a source of left-overs e.g. if we would add a new `ps`-like output)?

TurboTurtle:
> Using `--upload-pass` as a command line argument, won't `sosreport` collect the password when taking a `ps` output?
> 

Yep - that's why I worded the manpage entry the way I did. We currently carry the same concern for the encryption password, if one is used like that. However, that one is slightly less concerning since you'd already have to have decrypted the sosreport in order to see the decryption password.

> Could a support of a bash variable (something like `PGPASSWORD`) be added here? Or some smart obfuscation of the password from `ps` outputs (though that can be cumbersome and a source of left-overs e.g. if we would add a new `ps`-like output)?

A bash variable makes a lot of sense. I was mainly aiming for allowing uploading along with `--batch` by providing this option here (for when a distro-default is not used/set). But a bash variable would still allow that and avoid the entry in `ps` output.

pmoravec:
> > Using `--upload-pass` as a command line argument, won't `sosreport` collect the password when taking a `ps` output?
> 
> Yep - that's why I worded the manpage entry the way I did. We currently carry the same concern for the encryption password, if one is used like that. However, that one is slightly less concerning since you'd already have to have decrypted the sosreport in order to see the decryption password.

Ah good point and good implementation. So the bash variable will make sense mainly for `--batch` mode - still worth to have, I think, but not crucial.

sos/policies/__init__.py
```diff
@@ -900,6 +904,18 @@ def init_kernel_modules(self):
         lines = shell_out("lsmod", timeout=0).splitlines()
         self.kernel_mods = [line.split()[0].strip() for line in lines]
 
+    def get_upload_url(self):
+        """Helper function to determine if we should use the policy default
+        upload url or one provided by the user
+        """
+        return self.upload_url or self.commons['cmdlineopts'].upload_url
```
pmoravec: If upload URL will be provided both by policy and on cmdline, won't this prefer the policy one / default one? While cmdline should take precedense?

(and what if the URL will be in `/etc/sos.conf`?)

sos/policies/__init__.py
```diff
@@ -900,6 +904,18 @@ def init_kernel_modules(self):
...
+
+    def get_upload_user(self):
+        """Helper function to determine if we should use the policy default
+        upload user or one provided by the user
+        """
+        return self.upload_user or self.commons['cmdlineopts'].upload_user
```
pmoravec: Ditto here.

sos/policies/redhat.py
```diff
@@ -281,6 +287,61 @@ def check(cls):
                         return True
         return False
 
+    def get_upload_url(self):
+        if self.case_id and not self.commons['cmdlineopts'].upload_url:
+            # in this case we want to default to using the RHCP for uploads so
+            # clear the default FTP user so that sos users will be prompted
+            # for an RHCP username
```
pmoravec: I suggest rephrasing to something like "clear the default FTP use so that RHCP username must be provided via config / cmdline option or prompt". (current text confusingly ignores `--upload-user` option)

sos/policies/redhat.py
```diff
@@ -281,6 +287,61 @@ def check(cls):
...
+            self.upload_user = None
+            # this string is only used in pre_work() prompts, so this does not
+            # need to be the actual RHCP url
+            return "Red Hat Customer Portal"
+        return RH_FTP_HOST or self.commons['cmdlineopts'].upload_url
```
pmoravec: Again, should not `cmdlineopts` option (if set) have preference to `RH_FTP_HOST`?

sos/policies/redhat.py
```diff
@@ -281,6 +287,61 @@ def check(cls):
...
+
+    def get_upload_user(self):
+        # if this is anything other than dropbox, annonymous won't work
+        if self.get_upload_url() != RH_FTP_HOST:
+            return self.commons['cmdlineopts'].upload_user
```
pmoravec: Nice catch :+1:

pmoravec:
In general, I like the idea + specification + implementation a lot. Code review done, will play with some use cases if I find some issue.

sos/policies/__init__.py
TurboTurtle: Oops, yep - I had the logic backwards there, thanks.

For `/etc/sos.conf` inclusion this would still be overridden with an actual cmdline arg, and if the URL/user/whatever is specified there, it should definitely override the policy defaults imo.

sos/policies/redhat.py
TurboTurtle: Ack.

sos/policies/redhat.py
```diff
@@ -255,6 +255,9 @@ class RHELPolicy(RedHatPolicy):
 generated in %(tmpdir)s and may be provided to a %(vendor)s \
 support representative.
 """ + disclaimer_text + "%(vendor_text)s\n")
+    upload_url = 'dropbox.redhat.com'
```
pmoravec: Maybe could we point to `RH_FTP_HOST` here, to prevent constant string duplication issues?

sos/policies/redhat.py
```diff
@@ -284,6 +287,60 @@ def check(cls):
...
+        # if this is anything other than dropbox, annonymous won't work
+        if self.get_upload_url() != RH_FTP_HOST:
+            return self.commons['cmdlineopts'].upload_user
+        return self.upload_user
+
+    def upload_archive(self, archive, url, directory, user, password):
+
+        # If either are missing, we cannot upload to a case. Fall back to
+        # dropbox
+        if not self.case_id or not password:
```
pmoravec: Maybe extend it also by "or not user"? To catch (well, rare) situations `--batch --case=123 --upload-pass mysecret` without username, when we shall also fallback to dropbox.

BryanQuigley:
Awesome work.

I would prefer if the generic built-in upload method was more generic though and could be extended to other methods - I was thinking HTTPS upload would be a better default, but alas maybe that varies too.

A quick (and working) mockup of what would be awesome for Ubuntu.  Unfortunately I couldn't use the same requests methods as you.
`    def upload_archive(self, archive, url, directory, user, password):

        import os
        filename = os.path.basename(archive)
        url = "https://files.support.canonical.com/uploads/"
        user = 'ubuntu'
        password = 'ubuntu'

        print("Uploading to Canonical Support file server")

        if not self.case_id:
            print("You are uploading without a case ID, you must manually "
                  "copy the filename into a case so it can be associated")

        try:
            with open(archive, 'rb') as arc:
                import requests
                auth = (user, password)
                r = requests.put(url + filename, auth=auth, data=arc)
                if r.status_code != 201:
                    if r.status_code == 401:
                        raise Exception(
                            "Incorrect Customer Portal credentials"
                        )
                    raise Exception("POST request returned %s: %s"
                                    % (r.status_code, r.reason))
        except ImportError:
            raise Exception("Unable to upload due to missing python requests "
                            "library")`

sos/sosreport.py
```diff
@@ -1296,6 +1308,17 @@ def final_work(self):
                 except (OSError, IOError):
                     print(_("Error moving checksum file: %s" % archive_hash))
 
+        if self.opts.upload or self.opts.upload_url:
```
BryanQuigley: I'd prefer if this goes after display_results.  That way they have more info - like how big the file is.

TurboTurtle: Fair point, I'll get that moved.

TurboTurtle:
> I would prefer if the generic built-in upload method was more generic though and could be extended to other methods - I was thinking HTTPS upload would be a better default, but alas maybe that varies too.
> 

I think it does vary too much, mainly it comes down to this use of `data` vs. `files` kwarg assuming the differences remain the same between `put()` and `post()`:

>             r = requests.put(url + filename, auth=auth, data=arc)
>                                                         ^^^^
 
For `post()`, the `data` kwarg means "streaming" for the upload while `files` means "non-streaming". I hadn't tried `put()`, so I'll do some testing on that end this week. I would absolutely love to have this be a generic method in the base `Policy()`.

I suppose we could perhaps use a `upload_streaming` class attr, and have the builtin `upload_archive()` use that to determine which method to use. We'd gain a bit of standardization, and other distros wouldn't need to override quite so much should they also want this kind of functionality.

I'd need to really look over the policy-level design, as I don't want this to get too convoluted in supporting multiple methods, but I think we could default to this order: HTTPS -> SFTP -> FTP.

Spit-balling here, my initial thoughts are that policies should ideally define HTTPS as the primary target. If that fails, or the policy determines that HTTPS cannot be used (no login creds or case number in RH's example), or if `--upload-url` is provided by a user we **require** a protocol identifier in the url. Using that approach, we could have separate builtin methods for each protocol.

TurboTurtle:
Ok, I've overhauled this PR a bit.

- We now determine the upload protocol based on the url determined by both commandline options and policy definitions. If a protocol is not specified, we do not attempt any upload.

- Made the HTTPS upload functions more generic. The RedHat and Ubuntu policies now do not need to override any of the actual uploading methods to support their locations.

- Added/tested Ubuntu uploading to Canonical's HTTPS location

- We will no longer prompt for case ID/upload user/upload password if those are already provided either by policy or by command line options

- Uploading now follows `display_results()` so that users know how big of an upload they are attempting.

I have not yet implemented an SFTP upload method, though the framework is there. At the moment Red Hat does not have an SFTP location to use, and Ubuntu's HTTPS location is using well known credentials as provided by @BryanQuigley so I'm not convinced we need to add SFTP support just yet.

TurboTurtle:
Hmm. The LGTM alert for `input()` appears to only apply to python2.x. `raw_input()` doesn't exist in python3, and afaik it doesn't behave the way LGTM describes.

TurboTurtle:
Updated the manpage and --help strings to remove the references to FTP servers specifically.

BryanQuigley:
LGTM, I'll commit later today (or early tomorrow) unless you want to get more reviews.

TurboTurtle:
For something like this, I'd definitely like final signoffs by all maintainers.

So tagging @pmoravec @bmr-cymru

man/en/sosreport.1
```diff
@@ -247,6 +249,60 @@ output, or string data from the system. The resulting logs may be used
...
+.TP
+.B \--upload-url URL
+If a vendor does not provide a default upload location, or if you would like to upload
+the archive to a different location, specify the address here.
+
+A support protocol MUST be specified in this URL. Currently uploading is supported
+for HTTPS, SFTP, and FTP protocols.
+
+If your destination server listens on a non-standard port, specify the listening
+port in the URL.
```
pmoravec: Not sure if this needs to mention (and if so, if it should be here): Support of HTTP/FTP proxy fully relies on HTTP[|S]_PROXY environment variables only. Sosreport itself does not allow configuring a proxy server to upload the report via it.

Leaving on Jake if this not is worth to mention or not. I am fully OK in not explicitly mentioning it anywhere.

sos/policies/__init__.py
```diff
@@ -922,5 +948,239 @@ def pre_work(self):
...
+
+    def prompt_for_upload_password(self):
+        """Should be overridden by policies to determine if a password needs to
+        be provided for upload or not
+        """
+        if (not self.upload_password and not
+                self._upload_password) and self.upload_user:
+            msg = (
+                "Please provide the upload password for %s: "
+                % self.upload_user or self._upload_user
```
pmoravec: Due to the test 3 lines above, `self.upload_user` is evaluated to `True`, hence the `.. or self._upload_user` is redundant.

sos/policies/__init__.py
```diff
@@ -835,8 +843,15 @@ class LinuxPolicy(Policy):
...
-
+    _upload_url = None
+    _upload_directory = '/'
+    _upload_user = None
+    _upload_password = None
+    _use_https_streaming = False
     _preferred_hash_name = None
+    upload_url = None
+    upload_user = None
+    upload_password = None
```
pmoravec: For better understanding, I would love to have a comment why there are "duplicate" variables like `upload_url` / `_upload_url`.

pmoravec:
We shall comment out (in manpages?) that using `--build` and `--upload*` options is mutually exclusive. And sosreport should handle this better than now:

    sosreport --build --upload --batch
    ..
      /var/tmp/sosreport-pmoravec-rhel8-2020-01-21-zrcakeu


    Attempting upload to ftp://dropbox.redhat.com
    Upload attempt failed: expected str, bytes or os.PathLike object, not NoneType

sos/policies/__init__.py
TurboTurtle: Ack, fixed.

TurboTurtle:
Updated the PR, to include the highlights by @pmoravec 

As for `--batch` I don't think they need to be mutually exclusive, that failure behavior noted was just an oversight on my part. `--batch` should now work for vendor-default uploads that do not require users to enter authentication credentials. Tested this with the public RH dropbox, and `--batch` is working as expected.


Also, I figured out what LGTM was flagging on - specifically for `str` returns from `input()`. In python2.7, we'd need to use `raw_input()`, but in python3.x that has been renamed to just the standard `input()`. So, I've remapped `input()` to `raw_input()` when we're on python2.7 to resolve that conflict.

TurboTurtle:
Added a catch for the use of `--build` and `--upload` at the same time that @pmoravec caught. We will no longer attempt upload if `--build` is used.

Also, added an explicit exception handler to when `--upload` is used but the policy doesn't define a location.

sos/policies/redhat.py
```diff
@@ -284,6 +295,45 @@ def check(cls):
...
+            return self.commons['cmdlineopts'].upload_url
+        if (not self.case_id or not self.upload_user or not
+                self.upload_password):
+            # Cannot use the RHCP. Use anonymous dropbox
+            self.upload_user = self._upload_user
+            self.upload_directory = self._upload_directory
+            self.upload_password = None
+            return RH_FTP_HOST
+        else:
+            if self.case_id and self.upload_user and self.upload_password:
```
pmoravec: Isnt this test redundant? (opposite to "if (not self.case_id or not ..)" in else branch, so..).

sos/policies/__init__.py
```diff
@@ -923,5 +955,242 @@ def pre_work(self):
 
         return
 
+    def prompt_for_upload_user(self):
+        """Should be overridden by policies to determine if a user needs to
+        be provided or not
+        """
+        if (not self.commons['cmdlineopts'].upload_user and not
```
pmoravec: Isnt it better to test here `self.upload_user` instead of cmdlineopts param? Both shall have the same value, just everywhere else we test `self.upload_user` or `upload_password`.

TurboTurtle: Yup, this was a straggler from when I moved everything to `self.upload_*` from the cmdline opts at the beginning of the upload logic. Fixing.

sos/policies/redhat.py
TurboTurtle: It is, I think here I was initially planning some more logical checks, but ended up not needing them and didn't fully clean it up. Fixing.

TurboTurtle:
Made one final change to a new endpoint for the RH Customer Portal as we are migrating the API.

bmr-cymru:
I like the implementation. At the time we pulled the original out we were getting multiple bug reports or feature requests for secure FTP and at the time there was no usable Python implementation that was widely enough available for us to use.

If anything I think it'd be good to also allow passing of the upload parameters (at least username/password) via the environment - it'd make using the feature from scripts much easier without leaking credentials on the command line.

PR #1918, opened 2020-01-15
[openstack_tripleo] Fetch all installer related content

cjeanner:
This plugin allows to fetch the installation-related content, precisely
the configuration being launched at some point by the OpenStack
installer, known as "TripleO" or "Director".

Content of this plugin:
- paunch.log containers container configuration and startup information
- tripleo-config contains per-step configuration
- container-puppet contains per-step puppet related configuration, as well
as puppet facts cache.

---

### DIFF
--- a/sos/plugins/openstack_tripleo.py
+++ b/sos/plugins/openstack_tripleo.py
@@ -1,0 +1,47 @@
+# Copyright (C) 2020 Red Hat, Inc., Cedric Jeanneret <cjeanner@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+import re
+
+
+class OpenStackTripleO(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
+    """Fetch installation informations from OpenStack Installer
+    """
+
+    plugin_name = 'openstack_tripleo'
+    profiles = ('openstack', 'openstack_controller', 'openstack_compute')
+    packages = ('openstack-selinux',)
+
+    def setup(self):
+        # Notes: recursion is max 2 for container-puppet and tripleo-config
+        # Those directories are present on all OpenStack nodes
+        self.add_copy_spec([
+            '/var/log/paunch.log',
+            '/var/lib/container-puppet/',
+            '/var/lib/tripleo-config/',
+            '/etc/puppet/hieradata/'
+        ])
+
+    def postproc(self):
+        # Ensures we do not leak passwords from the tripleo-config and
+        # hieradata locations.
+        # Other locations don't have sensitive data.
+        secrets = r'(".*(key|password|pass|secret|database_connection))' \
+                  r'([":\s]+)(.*[^"])([",]+)'
+        rgxp = re.compile(secrets, re.IGNORECASE)
+
+        self.do_path_regex_sub('/var/lib/tripleo-config/',
+                               rgxp, r'\1\3*********\5')
+        self.do_path_regex_sub('/etc/puppet/hieradata/',
+                               rgxp, r'\1\3*********\5')
+
+
+# vim: set et ts=4 sw=4 :


sos/plugins/openstack_tripleo.py
```diff
@@ -0,0 +1,29 @@
...
+    Recursion level is max 2 for both listed directories.
+    Those files are useful in order to debug a failed install, since the
+    whole deploy is based on them. We find those directories on all the nodes.
+    '''
+    plugin_name = 'openstack_tripleo'
+
+    def setup(self):
+        self.add_copy_spec('/var/log/paunch.log')
+        self.add_copy_spec('/var/lib/container-puppet')
+        self.add_copy_spec('/var/lib/tripleo-config')
```
TurboTurtle: These can be condensed into a single `add_copy_spec()` call using a list.

Is `paunch.log` ever rotated? Do we only ever need to grab the active log and not any rotated versions if so?

sos/plugins/openstack_tripleo.py
```diff
@@ -0,0 +1,29 @@
...
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+import os
```
TurboTurtle: Unused import

sos/plugins/openstack_tripleo.py
TurboTurtle: Is there any sensitive information in this config file? Or `container-puppet` for that matter?

sos/plugins/openstack_tripleo.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class OpenStackTripleO(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
+    '''Some TripleO logs and files (OpenStack installer)
```
TurboTurtle: Minor nit: PEP257 specifies the use of triple double quotes instead of triple single quotes. Normally I wouldn't even mention this, but #1914 has been opened to correct this elsewhere, so might as well nip it early.

sos/plugins/openstack_tripleo.py
EmilienM: The directory has files with passwords.

EmilienM: not afik

cjeanner: hm, right - I think I've seen something about masking password in another plugin. Lemme check that and see if I can do something clean about that. Will deploy a thing on my side in order to inspect that.

sos/plugins/openstack_tripleo.py
cjeanner: OK! didn't know about that one, will correct!

cjeanner:
@EmilienM should we also include the hieradata content? Apparently the "puppet.py" doesn't take care of it - and since it's one of the "source of trust" for tripleo deploy, it might become handy (with the secret filtering in order to avoid sensitive data leakage).

sos/plugins/openstack_tripleo.py
```diff
@@ -0,0 +1,39 @@
...
+                '/var/log/paunch.log',
+                '/var/lib/container-puppet/',
+                '/var/lib/tripleo-config/'
+                ])
+
+    def postproc(self):
+        # Ensures we do not leak passwords from the tripleo-config location
+        # Usually, password are named as SERVICE_PASS or SERVICE_PASSWORD
+        # We do not have RSA keys in this location.
+        password_reg = r'("[a-zA-Z_]+PASS(WORD)?"):\s?"(.+)",?'
```
EmilienM: can't we do better? At least to cover "secret" and "key"

cjeanner: so those two apparently don't exist in the location, but I'll create a new method in "plugin" in order to actually mask secrets. It will be based on what we already have in oslo.utils and should be efficient.

sos/plugins/openstack_tripleo.py
```diff
@@ -0,0 +1,39 @@
...
+
+from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+
+
+class OpenStackTripleO(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
+    """Log installation informations from OpenStack Installer
+    """
+
+    plugin_name = 'openstack_tripleo'
+
```
pmoravec: When the plugin will be enabled (so far it would be *any* time)? What installed package, kernel module loaded, file present or running service should enable the plugin?

Shall not be the plugin added to a profile? Cf. https://github.com/sosreport/sos/blob/master/sos/plugins/openstack_nova.py#L24 (and https://github.com/sosreport/sos/blob/master/sos/plugins/openstack_nova.py#L208 for enablement of the plugin).

sos/plugins/openstack_tripleo.py
```diff
@@ -0,0 +1,39 @@
...
+    def setup(self):
+        # Notes: recursion is max 2 for container-puppet and tripleo-config
+        # Those directories are present on all OpenStack nodes and
+        # may contain sensitive data (password).
```
pmoravec: This implies `container-puppet` logs can also contain sensitive data, but you dont obfuscate those in `postproc`. Could you please either scrub that passwords, or re-phrase this comment? (or did I mis-interpret the text?)

sos/plugins/openstack_tripleo.py
cjeanner: good remark - will take care of that in the next update. But I will probably need to make another PR "before" this one in order to inject a new method for secret handling.

sos/plugins/openstack_tripleo.py
```diff
@@ -0,0 +1,39 @@
...
+        self.add_copy_spec([
+                '/var/log/paunch.log',
+                '/var/lib/container-puppet/',
+                '/var/lib/tripleo-config/'
+                ])
```
TurboTurtle: Stylistic nit - sos style for this is to indent one level in, and have the closing brackets on the same indention as the name of the call, E.G.:

```python
self.add_copy_spec([
    '/var/log/paunch.log',
    '/var/lib/container-puppet/'
])
```

sos/plugins/openstack_tripleo.py
cjeanner: need to rephrase it.

sos/plugins/openstack_tripleo.py
pmoravec: No need for another PR. Just modify your branch and forcefully push it to github.

sos/plugins/openstack_tripleo.py
bmr-cymru: Presumably you're referring to the stuff in `oslo.utils.strutils`? Please discuss any changes or additions to the Plugin post-processing methods before going too far with this - if it's an extensive set of changes then that may justify a separate PR to focus on just that aspect - methods in `Plugin` need to be sufficiently generic to be useful for many plugins - otherwise if they are more tailored to the needs of one specific sub-family of plugins then we may want to consider a different organisation for the changes.

cjeanner: @bmr-cymru indeed - but I've found an easier regexp that does match the current need, just have to clean the code and we're good.

sos/plugins/openstack_tripleo.py
cjeanner: oh, yeah, I was more thinking about a "generic method in order to mask secrets" (see previous comment with @bmr-cymru ) - but that one isn't useful. I'll push shortly the change adding profiles and packages.

sos/plugins/openstack_tripleo.py
cjeanner: np - will update, thanks!

xek:
Ansible host_vars and group_vars from /var/lib/mistral/*/*_vars dirs may also be useful.

cjeanner:
> Ansible host_vars and group_vars from `/var/lib/mistral/*/*_vars` dirs may also be useful.

Will have a look about the content there, and add it as well. Probably has some passwords/sensitive data..

cjeanner:
> Ansible host_vars and group_vars from `/var/lib/mistral/*/*_vars` dirs may also be useful.
@xek After a check, it looks that most of the vars are already in hieradata and other locations, directly on the overcloud nodes.
Not sure we need to repeat the same data.

I propose we go with the current patch since it will already help a lot. We might come with a follow-up if needed, but I'm pretty sure we're good with the current files.
@EmilienM any thoughts?

cjeanner:
Hello! Any ETA for the inclusion of this new plugin, @TurboTurtle @pmoravec ?

TurboTurtle:
I'll carve some time out on Monday to merge some of the approved PRs, including this one.

PR #1920, opened 2020-01-16
[ipa] Collect 'getcert list' only if service certmonger is running

jjansky1:
During collection of 'getcert list' is started certmonger service.
That is not wanted behavior, as result should be 'getcert list'
collected only if certmonger is running.

Resolves: #1920

---

### DIFF
--- a/sos/plugins/ipa.py
+++ b/sos/plugins/ipa.py
@@ -8,7 +8,7 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.plugins import Plugin, RedHatPlugin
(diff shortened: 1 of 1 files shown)


sos/plugins/ipa.py
```diff
@@ -161,6 +160,11 @@ def setup(self):
             "klist -ket /var/lib/ipa/gssproxy/http.keytab"
         ])
 
+        getcert_pred = SoSPredicate(self,
+                                    services=['certmonger'])
+
```
pmoravec: Just a nit-pick: redundant empty line.

Issue #905, opened 2016-12-21
collect s390dbf data correctly


stefan-haberland:
In the s390 plugin a call to "find /proc/s390dbf -type f" is found which does not do anything useful because normally the s390dbf is found under /sys/kernel/debug/s390dbf and beside this the names of the files are not helpful without the content.

I suggest to include the following in the s390.py self.add_copy_spec([ section:
"/sys/kernel/debug/s390dbf/*/*",

The other but less preferred solution would be to add 
"find /sys/kernel/debug/s390dbf/ -type f -print -exec cat '{}' \;",
to the  self.add_cmd_output([ section.

stefan-haberland:
Sorry but the suggestion was destroyed by the editor, it should look like:
`"/sys/kernel/debug/s390dbf/*/*",`

PR #1926, opened 2020-01-29
[s390] Correctly capture s390dbf

TurboTurtle:
Captures the content under /sys/kernel/debug/s390dbf rather than just
listing the file names.

Also included are trivial stylistic fixes.

Closes: #905
Resolves: #1926

---


pmoravec:
ACK from devel perspective. Just dont know if `/sys/kernel/debug/s390dbf` cant have a sort of tracing files we should also skip (cf. https://github.com/sosreport/sos/blob/master/sos/plugins/kernel.py#L89-L101).

TurboTurtle:
That was my concern as well, so I borrowed an s390 system from engineering and tested this plugin on it several times. No hangs seen, and added the skipping of the `flush` files from those tests. Capturing them may not do any harm, but they also don't do us any good to collect so we should just side step them anyways.

PR #1928, opened 2020-01-29
[kubernetes] Add detailed node information and use oc when available

TurboTurtle:
Two patch set that first adds collection of `describe node <node>` output for each node returned by `get nodes`. Secondly, for RH OCP installations, rather than using `kubectl` with an updated config file, use the existing `oc` binary that performs the wrapping. This will allow us to be completely sure that the information collected by sos is the same information that would be seen when an end user runs the same commands manually.

---

### DIFF
--- a/sos/plugins/kubernetes.py
+++ b/sos/plugins/kubernetes.py
@@ -89,13 +89,23 @@ class Kubernetes(Plugin):
         # these are not namespaced, must pull separately.
         global_resources = [
             'namespaces',
-            'nodes',
             'projects',
             'pvs'
         ]
         self.add_cmd_output([
             "%s get %s" % (self.kube_cmd, res) for res in global_resources
         ])
+
+        # Get detailed node information
+        nodes = self.collect_cmd_output("%s get nodes" % self.kube_cmd)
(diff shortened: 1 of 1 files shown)


sos/plugins/kubernetes.py
```diff
@@ -96,6 +96,17 @@ def setup(self):
         self.add_cmd_output([
             "%s get %s" % (self.kube_cmd, res) for res in global_resources
         ])
+
+        # Get detailed node information
+        nodes = self.exec_cmd("%s get nodes" % self.kube_cmd)
```
pmoravec: This command will be executed twice (via `add_cmd_output` few lines above as well). We can `collect_cmd_output` here and remove `nodes` from the `global_resources`, but that would leave `global_resources` incomplete..? Or some other solution to prevent twice call? (I am ok with the current implementation).

TurboTurtle: Oh, good catch. We don't currently use `global_resources` anywhere else in the Plugin, so we're good to use `collect_cmd_output()` for nodes and remove that from the list.

I'll update shortly.

Issue #1927, opened 2020-01-29
Extend gnocchi plugin to Ubuntu


cjohnston1158:
The gnocchi plugin currently only supports RedHat. Needs to be extended to Ubuntu.

PR #1929, opened 2020-01-29
[gnocchi] Add support for Debian and Ubuntu.

cjohnston1158:
Adds Gnocchi plugin support for Debian and Ubuntu.

Fixes: #1927

Signed-off-by: Chris Johnston <chris.johnston@canonical.com>

---

- [X ] If this commit closes an existing issue, is the line `Closes: #ISSUENUMBER` included in an independent line?
- [X ] If this commit resolves an existing pull request, is the line `Resolves: #PRNUMBER` included in an independent line?

### DIFF
--- a/sos/plugins/gnocchi.py
+++ b/sos/plugins/gnocchi.py
@@ -10,44 +10,31 @@
 # See the LICENSE file in the source distribution for further information.
 
 import os
-from sos.plugins import Plugin, RedHatPlugin
+from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
 
 
-class GnocchiPlugin(Plugin, RedHatPlugin):
+class Gnocchi(Plugin):
     """Gnocchi - Metric as a service"""
     plugin_name = "gnocchi"
 
     profiles = ('openstack', 'openstack_controller')
 
-    packages = (
-        'openstack-gnocchi-metricd', 'openstack-gnocchi-common',
-        'openstack-gnocchi-statsd', 'openstack-gnocchi-api',
-        'openstack-gnocchi-carbonara'
-    )
-
     requires_root = False
-
-    var_puppet_gen = "/var/lib/config-data/puppet-generated/gnocchi"
 
     def setup(self):
         self.add_copy_spec([
             "/etc/gnocchi/*",
-            self.var_puppet_gen + "/etc/gnocchi/*",
-            self.var_puppet_gen + "/etc/httpd/conf/*",
-            self.var_puppet_gen + "/etc/httpd/conf.d/*",
-            self.var_puppet_gen + "/etc/httpd/conf.modules.d/wsgi.conf",
-            self.var_puppet_gen + "/etc/my.cnf.d/tripleo.cnf"
         ])
 
         if self.get_option("all_logs"):
             self.add_copy_spec([
                 "/var/log/gnocchi/*",
-                "/var/log/httpd/gnocchi*",
+                "/var/log/{}*/gnocchi*".format(self.apachepkg)
             ])
         else:
             self.add_copy_spec([
                 "/var/log/gnocchi/*.log",
-                "/var/log/httpd/gnocchi*.log",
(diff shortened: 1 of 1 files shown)


sos/plugins/gnocchi.py
```diff
@@ -70,18 +56,64 @@ def setup(self):
...
-            self.var_puppet_gen + "/etc/gnocchi/"
-            "gnocchi.conf",
-            r"password=(.*)",
-            r"password=*****",
+
+    def postproc(self):
+        protect_keys = ["ceph_secret", "password", ]
+        self.apply_regex_sub(
+            r"((?m)^\s*(%s)\s*=\s*)(.*)" % "|".join(protect_keys),
+            r"\1*********"
```
TurboTurtle: This seems a bit awkward to me. I'd much rather we stay consistent with other plugin's where the `postproc()` call handles all the substitutions. Here I don't see the benefit to making a new method when all we're doing is consolidating the constructed regex for the same files.

sos/plugins/gnocchi.py
```diff
@@ -70,18 +56,64 @@ def setup(self):
...
+        ])
+
+        if self.get_option("all_logs"):
+            self.add_copy_spec([
+                "/var/log/httpd/gnocchi*",
+            ])
+        else:
+            self.add_copy_spec([
+                "/var/log/httpd/gnocchi*.log",
+            ])
```
TurboTurtle: `add_copy_spec()` supports both lists and strings for single file specifications. No need for a list here.

sos/plugins/gnocchi.py
```diff
@@ -70,18 +56,64 @@ def setup(self):
...
+    def setup(self):
+        super(DebianGnocchi, self).setup()
+        if self.get_option("all_logs"):
+            self.add_copy_spec([
+                "/var/log/apache2/gnocchi*",
+            ])
+        else:
+            self.add_copy_spec([
+                "/var/log/apache2/gnocchi*.log",
+            ])
```
TurboTurtle: Ditto here, no need for a single-item list.

sos/plugins/gnocchi.py
```diff
@@ -10,44 +10,32 @@
...
 
-    packages = (
-        'openstack-gnocchi-metricd', 'openstack-gnocchi-common',
-        'openstack-gnocchi-statsd', 'openstack-gnocchi-api',
-        'openstack-gnocchi-carbonara'
-    )
-
     requires_root = False
-
     var_puppet_gen = "/var/lib/config-data/puppet-generated/gnocchi"
```
TurboTurtle: If we're moving the `var_puppet_gen` file collections to the RedHat specific plugin, should we not move the definition of`self.var_puppet_gen` in the base `Gnocchi` class out as well?

Or conversely, should those file collections be updated to account for the apache package name differences as well?

TurboTurtle:
This generally looks good to me now, just one more thing to do.

Please squash the fixup commits into the original commit. If you're unsure, you can do this via `git rebase -i <commit ID before your first commit>` and in the editor marking the second two commits as 'fixup' instead of 'pick'.

BryanQuigley:
There is now an error because of moving var_puppet_gen - it's referenced in post_proc.

  File "/root/sos/sos/plugins/gnocchi.py", line 72, in postproc
    self.var_puppet_gen + "/etc/gnocchi/"
AttributeError: 'DebianGnocchi' object has no attribute 'var_puppet_gen'
  File "/root/sos/sos/plugins/gnocchi.py", line 72, in postproc
    self.var_puppet_gen + "/etc/gnocchi/"
AttributeError: 'DebianGnocchi' object has no attribute 'var_puppet_gen'

I'd also like to see the Ubuntu version activate on "python3-gnocchiclient" being installed.

Other than that +1.

TurboTurtle:
Nice catch.

We could either add a `postproc()` override to `RedHatGnocchi` that calls `super(RedHatGnocchi, self).postproc()` or just move `self.var_puppet_gen` back into the base class. I have a slight preference for the former, but either way is fine in my opinion.

Issue #1916, opened 2020-01-13
[networking] options to limit namespaces to traverse


pmoravec:
There are few use cases where `networking` plugin timeouts due to the number of networking namespaces on the system and the number of commands to collect per each such namespace. That causes further commands are not collected, in favour of namespaces-related commands that one will most probably never use. Typically:

- many hundreds to thousands of namespaces on the system, where taking 12 commands per each takes too much time altogether; while most of the data collected wont be reviewed any time (who *usually* checks each and every of thousands of namespaces details?)
- `sosreport` collects `ethtool` per namespace (https://github.com/sosreport/sos/blob/master/sos/plugins/networking.py#L242-L247) that is not required on OpenStack (per my colleagues from technical support of OSP)

Several plugin options that can help here (both option names and default values are subject to discussion, of course):
- plugin option `inspect_namespaces` with default value 100 (where 0 means "unlimited"?), that limits number of namespaces **as returned by `ip netns` command** to iterate through
- plugin option `all_namespaces` to collect all namespaces details
- plugin option `namespaces=eth*,ens*,lo*` that contains reg.exps. of namespaces to take details for (only)
- plugin option `ethtool_namespaces` to skip `ethtool` commands per namespace

My proposal:
- `inspect_namespaces` with default 100, where 0 means unlimited
- `ethtool_namespaces` with default `True` and OpenStack preset amended with `networking.ethtool_namespaces=False`

An alternative proposal:
- `all_namespaces` with default `False`, and taking details of (hardcoded) 100 namespaces
- `namespaces` to list namespaces to traverse through (additionally to default behaviour, OR *only* these)

(to make the options complete, #1585 is an alternative approach that did not raise much attention)

Any other proposal?

Cc @TurboTurtle and @bmr-cymru for kick-off, and also @slashdd for Ubuntu+Openstack feedback.

MrStupnikov:
@pmoravec I would like to ask you to not change default behavior and provide an option to users with huge amount of network namespaces to turn off any namespace-related operations if it is needed. You can put a warning message to stdout if it looks like user's environment doesn't fit default options.

If it will not work for you, then please consider increasing limit to 200.

Thanks, Alex.

pmoravec:
Increasing the limit is no problem at all (seen some proposal of 500 also, though cant that already hit the plugin timeout? dunno).

[...]

PR #1930, opened 2020-01-31
[networking] options to limit namespaces

jjansky1:
* Added posibility to define namespaces with
parameter namespaces = "". Posible regexp
like "eth* ens4".
* Added inspect_namespaces which defines how
many namespaces should be inspected, by
default unlimited
* Added ethtool_namespaces variable which
is by default true and if set to false it
will not execute ethtool commands for
namespaces
* Changed options for OpenStack environment
where ethtool_namespaces=False and
inspect_namespaces = 200

Closes: sosreport#1916

---

### DIFF
--- a/sos/plugins/networking.py
+++ b/sos/plugins/networking.py
@@ -9,6 +9,7 @@
 from sos.plugins import (Plugin, RedHatPlugin, UbuntuPlugin, DebianPlugin,
                          SoSPredicate)
 from os import listdir
+from re import match
 
 
 class Networking(Plugin):
@@ -17,10 +18,17 @@ class Networking(Plugin):
     plugin_name = "networking"
     profiles = ('network', 'hardware', 'system')
     trace_host = "www.example.com"
-    option_list = [(
+    option_list = [
         ("traceroute", "collect a traceroute to %s" % trace_host, "slow",
-         False)
-    )]
+         False),
+        ("namespace_regex", "Specific namespaces regex to be " +
+         "collected, namespaces regex should be separated by whitespace " +
+         "as for example \"eth* ens2\"", "fast", ""),
+        ("namespaces", "Number of namespaces to collect, 0 for unlimited. " +
+         "Incompatible with the namespace_regex plugin option", "slow", 0),
+        ("ethtool_namespaces", "Define if ethtool commands should be " +
+         "collected for namespaces", "slow", True)
+    ]
 
     # switch to enable netstat "wide" (non-truncated) output mode
     ns_wide = "-W"
@@ -201,13 +209,34 @@ class Networking(Plugin):
         cmd_prefix = "ip netns exec "
         if ip_netns['status'] == 0:
             out_ns = []
+            # Regex initialization outside of for loop
+            if self.get_option("namespace_regex"):
+                regex = '(?:%s)' % '|'.join(
+                        self.get_option("namespace_regex").split())
             for line in ip_netns['output'].splitlines():
                 # If there's no namespaces, no need to continue
                 if line.startswith("Object \"netns\" is unknown") \
                         or line.isspace() \
                         or line[:1].isspace():
                     continue
-                out_ns.append(line.partition(' ')[0])
+                # if namespace_regex defined, append only namespaces
+                # matching with regex
+                if self.get_option("namespace_regex"):
+                    if bool(match(regex, line)):
+                        out_ns.append(line.partition(' ')[0])
+
+                # if namespaces is defined and namespace_regex is not defined
+                # remove from out_ns namespaces with higher index than defined
+                elif self.get_option("namespaces") != 0:
+                    out_ns.append(line.partition(' ')[0])
+                    if len(out_ns) == self.get_option("namespaces"):
+                        self._log_warn("Limiting namespace iteration " +
+                                       "to first %s namespaces found"
+                                       % self.get_option("namespaces"))
+                        break
+                else:
+                    out_ns.append(line.partition(' ')[0])
+
             for namespace in out_ns:
                 ns_cmd_prefix = cmd_prefix + namespace + " "
                 self.add_cmd_output([
@@ -226,25 +255,28 @@ class Networking(Plugin):
                 # check for it
                 self.add_cmd_output(ss_cmd, pred=ss_pred)
 
-            # Devices that exist in a namespace use less ethtool
-            # parameters. Run this per namespace.
-            for namespace in out_ns:
-                ns_cmd_prefix = cmd_prefix + namespace + " "
-                netns_netdev_list = self.exec_cmd(
-                    ns_cmd_prefix + "ls -1 /sys/class/net/"
-                )
-                for eth in netns_netdev_list['output'].splitlines():
-                    # skip 'bonding_masters' file created when loading the
-                    # bonding module but the file does not correspond to
-                    # a device
-                    if eth == "bonding_masters":
-                        continue
-                    self.add_cmd_output([
-                        ns_cmd_prefix + "ethtool " + eth,
-                        ns_cmd_prefix + "ethtool -i " + eth,
-                        ns_cmd_prefix + "ethtool -k " + eth,
-                        ns_cmd_prefix + "ethtool -S " + eth
-                    ])
+            # Collect ethtool commands only when ethtool_namespaces
+            # is set to true.
+            if self.get_option("ethtool_namespaces"):
+                # Devices that exist in a namespace use less ethtool
+                # parameters. Run this per namespace.
+                for namespace in out_ns:
+                    ns_cmd_prefix = cmd_prefix + namespace + " "
+                    netns_netdev_list = self.exec_cmd(
+                        ns_cmd_prefix + "ls -1 /sys/class/net/"
+                    )
+                    for eth in netns_netdev_list['output'].splitlines():
+                        # skip 'bonding_masters' file created when loading the
+                        # bonding module but the file does not correspond to
+                        # a device
+                        if eth == "bonding_masters":
+                            continue
+                        self.add_cmd_output([
+                            ns_cmd_prefix + "ethtool " + eth,
+                            ns_cmd_prefix + "ethtool -i " + eth,
+                            ns_cmd_prefix + "ethtool -k " + eth,
+                            ns_cmd_prefix + "ethtool -S " + eth
+                        ])
 
         return
 
--- a/sos/policies/redhat.py
+++ b/sos/policies/redhat.py
@@ -174,7 +174,6 @@ ENV_HOST_SYSROOT = 'HOST'
 _opts_verify = SoSOptions(verify=True)
 _opts_all_logs = SoSOptions(all_logs=True)
 _opts_all_logs_verify = SoSOptions(all_logs=True, verify=True)
-_opts_no_lsof = SoSOptions(plugopts=['process.lsof=off'])
 _cb_profiles = ['boot', 'storage', 'system']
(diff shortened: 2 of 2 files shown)


sos/plugins/networking.py
```diff
@@ -17,10 +18,16 @@ class Networking(Plugin):
...
     profiles = ('network', 'hardware', 'system')
     trace_host = "www.example.com"
-    option_list = [(
+    option_list = [
         ("traceroute", "collect a traceroute to %s" % trace_host, "slow",
-         False)
-    )]
+         False),
+        ("namespaces", """string of specific namespaces regex to be
+          collected as \"eth* ens2\"""", "quick", ""),
```
pmoravec: Use either "fast" or "" instead of "quick" (see e.g. https://github.com/sosreport/sos/blob/master/sos/plugins/dnf.py#L23 or https://github.com/sosreport/sos/blob/master/sos/plugins/postgresql.py#L32).

Maybe better to specify explicitly that the namespaces are separated by space (which would prolong the help text, though, so maybe remove the example? not decided..)

sos/plugins/networking.py
```diff
@@ -188,9 +195,6 @@ def setup(self):
             "bridge -d vlan show"
         ])
 
-        if self.get_option("traceroute"):
-            self.add_cmd_output("/bin/traceroute -n %s" % self.trace_host)
-
```
pmoravec: Removed by mistake? :)

sos/plugins/networking.py
```diff
@@ -208,6 +212,22 @@ def setup(self):
                         or line[:1].isspace():
                     continue
                 out_ns.append(line.partition(' ')[0])
+            # if namespaces defined, remove from out_ns not matching
+            # namespaces with regex
+            if self.get_option("namespaces"):
+                regex = '(?:%s)' % '|'.join(
+                      self.get_option("namespaces").split())
+                new_ns = []
+                for ns in out_ns:
```
pmoravec: Can't be this `for ns in out_ns` cycle be merged into the previous one? I.e. why to build a list of namespaces to purge some items just afterwards?

sos/plugins/networking.py
jjansky1: changed to "fast", please check new commit

sos/plugins/networking.py
jjansky1: Removed by mistake, new commit have this option again.

sos/plugins/networking.py
jjansky1: I have implemented new options into original loop of appending ns into out_ns. Please check new commit.

sos/plugins/networking.py
```diff
@@ -207,7 +215,23 @@ def setup(self):
                         or line.isspace() \
                         or line[:1].isspace():
                     continue
-                out_ns.append(line.partition(' ')[0])
+                # if namespaces defined, remove from out_ns not matching
+                # namespaces with regex
+                if self.get_option("namespaces"):
+                    regex = '(?:%s)' % '|'.join(
+                        self.get_option("namespaces").split())
```
pmoravec: Maybe worth moving this variable initialization ahead of `for line in ..` cycle, to set it once, and not every time you loop inside the cycle?

sos/plugins/networking.py
```diff
@@ -207,7 +215,23 @@ def setup(self):
...
+                    if bool(match(regex, line)):
+                        out_ns.append(line.partition(' ')[0])
+
+                # if inspect_namespaces is defined, remove from out_ns
+                # namespaces with higher index than defined
+                elif self.get_option("inspect_namespaces") != 0:
+                    out_ns.append(line.partition(' ')[0])
+                    if len(out_ns) == self.get_option("inspect_namespaces"):
+                        break
```
pmoravec: This wont work when using both `namespaces` and `inspect_namespaces`: isnt it easier to add at the end of whole cycle `if len(out_ns) == get_option(..) then break`?

pmoravec: Ah that is intentional that the options can be used exclusively only. Taking back this, just not sure if the info about exclusive suage of the options shall not go to code or help.

sos/plugins/networking.py
jjansky1: regex initialization moved out from for loop

sos/plugins/networking.py
jjansky1: This works how was proposed in #1916 , inspect_namespaces will work only when namespaces are not defined. I have added comment into the code and added better parameter description.

sos/plugins/networking.py
```diff
@@ -17,10 +18,18 @@ class Networking(Plugin):
...
         ("traceroute", "collect a traceroute to %s" % trace_host, "slow",
-         False)
-    )]
+         False),
+        ("namespaces", "specific namespaces regex to be " +
+         "collected, namespaces should be separated by space " +
+         "as for example \"eth* ens2\"", "fast", ""),
+        ("inspect_namespaces", "define how many namespaces should be " +
+         "collected, 0 for unlimited. Works ONLY when namespaces are NOT " +
+         "defined", "slow", 0),
```
TurboTurtle: I think this could be worded differently. E.G.

`Number of namespaces to collect, 0 for unlimited. Incompatible with the namespaces plugin option`

sos/plugins/networking.py
```diff
@@ -17,10 +18,18 @@ class Networking(Plugin):
...
     trace_host = "www.example.com"
-    option_list = [(
+    option_list = [
         ("traceroute", "collect a traceroute to %s" % trace_host, "slow",
-         False)
-    )]
+         False),
+        ("namespaces", "specific namespaces regex to be " +
+         "collected, namespaces should be separated by space " +
+         "as for example \"eth* ens2\"", "fast", ""),
```
TurboTurtle: I think it'd be more user-consumable if the regexes were comma-delimited rather than space-delimited. Using commas is more consistent with how sos handles lists in options already.

sos/plugins/networking.py
```diff
@@ -17,10 +18,18 @@ class Networking(Plugin):
...
+        ("inspect_namespaces", "define how many namespaces should be " +
```
TurboTurtle: I'm not a huge fan of how the current option names map to their functions.

I think something like `namespace_regex` would make more sense here, with `namespaces` being used for the number of namespaces to iterate over.

sos/plugins/networking.py
```diff
@@ -201,13 +210,32 @@ def setup(self):
...
             for line in ip_netns['output'].splitlines():
                 # If there's no namespaces, no need to continue
                 if line.startswith("Object \"netns\" is unknown") \
                         or line.isspace() \
                         or line[:1].isspace():
                     continue
-                out_ns.append(line.partition(' ')[0])
+                # if namespaces defined, append only namespaces
+                # matching with regex
+                if self.get_option("namespaces"):
```
TurboTurtle: A log message here would be helpful, E.G. `Limiting namespace iteration to first X namespaces found`.

That way if a support engineer is wondering why they're missing namespace details, it is easier to identify.

sos/plugins/networking.py
pmoravec: Comma separated list does not work. Since in https://github.com/sosreport/sos/blob/master/sos/sosreport.py#L175-L178, the `ArgumentParser` splits the string e.g. `networking.namespaces='a*,b*,c*'` to three options `networking.namespaces=a*`, `b*` and `c*'` - it follws the same logic like `-o kernel,boot` that appends to `onlyplugins` two options.

Checking if `ArgumentParser` has a trick for this..

pmoravec: It probably has:

diff --git a/sos/sosreport.py b/sos/sosreport.py
index 2d2d1599..4a768f45 100644
--- a/sos/sosreport.py
+++ b/sos/sosreport.py
@@ -172,7 +172,7 @@ def _get_parser():
     parser.add_argument("-e", "--enable-plugins", action="extend",
                         dest="enableplugins", type=str,
                         help="enable these plugins", default=[])
-    parser.add_argument("-k", "--plugin-option", action="extend",
+    parser.add_argument("-k", "--plugin-option", action="append",
                         dest="plugopts", type=str,
                         help="plugin options in plugname.option=value "
                              "format (see -l)", default=[])

This stores whole `-k any_string_with_commas` as one item of the `plugopts` list, which allows `-k networking.namespaces=a*,b*,c*` to set the option value to `a*,b*,c*`.

HOWEVER, the side effect is each and every plugin option must have to be specificed via `-k` option. As `-k networking.traceroute=on,qpid.port=12345` now splits the options to two, while the above change does not.

sos/plugins/networking.py
jjansky1: @TurboTurtle Agree with log message, question is about log level, warn log level will show the message on screen, we can also use info level which will be only in sos_logs/sos.log. Which log level is more preferred in this case?

TurboTurtle: Hmm, I'm leaning towards warn level but I'm not committed to it. It'll help with RHOSP preset runs where the user isn't explicitly setting the value themselves, but it'd still be good to alert them that we may not be capturing all available information.

sos/plugins/networking.py
TurboTurtle: Ah, ok - I was thinking/hoping that ArgParse would be smart enough to take the quoted regex string as a whole unit, but I guess not.

Spaces it is then.

sos/plugins/networking.py
jjansky1: Warn level log was added for namespaces parameter.

sos/plugins/networking.py
jjansky1: Parameters renamed as suggested.

sos/plugins/networking.py
jjansky1: Parameter description updated as suggested.

jjansky1:
@TurboTurtle new commit provides changes as suggested, please review again. Thank you in advance.

sos/plugins/networking.py
```diff
@@ -17,10 +18,17 @@ class Networking(Plugin):
...
+    option_list = [
         ("traceroute", "collect a traceroute to %s" % trace_host, "slow",
-         False)
-    )]
+         False),
+        ("namespace_regex", "Specific namespaces regex to be " +
+         "collected, namespaces regex should be separated by whitespace " +
+         "as for example \"eth* ens2\"", "fast", ""),
+        ("namespaces", "Number of namespaces to collect, 0 for unlimited. " +
+         "Incompatible with the namespaces plugin option", "slow", 0),
```
TurboTurtle: Should read "incompatible with the namespaces_regex plugin option" following the option name changes.

sos/policies/redhat.py
```diff
@@ -191,6 +190,10 @@ def get_tmp_dir(self, opt_tmp_dir):
 
 RHOCP = "ocp"
 RHOCP_DESC = "OpenShift Container Platform by Red Hat"
+RHOSP_OPTS = SoSOptions(plugopts=[
+                             'process.lsof=off',
+                             'networking.ethtool_namespaces=False',
+                             'networking.inspect_namespaces=200'])
```
TurboTurtle: Needs to be updated to just `networking.namespaces=200`

jjansky1: Updated in latest commit.

sos/plugins/networking.py
jjansky1: Updated in latest commit.

pmoravec:
Fixed via c20bd8d . Thanks for the PR!

PR #1932, opened 2020-02-04
[sssd] call sssctl commands only when sssd service is running

pmoravec:
As sssctl timeouts otherwise.

Resolves: #1932

---

### DIFF
--- a/sos/plugins/sssd.py
+++ b/sos/plugins/sssd.py
@@ -8,7 +8,8 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+from sos.plugins import (Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin,
+                         SoSPredicate)
 
 
 class Sssd(Plugin):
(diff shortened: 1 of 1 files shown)


sos/plugins/sssd.py
```diff
@@ -28,12 +29,17 @@ def setup(self):
...
-
-        domain = self.collect_cmd_output("sssctl domain-list")
-        if domain['status'] == 0:
-            for domain_name in domain['output'].splitlines():
-                self.add_cmd_output("sssctl domain-status -o " + domain_name)
+        # call sssctl commands only when sssd service is running,
+        # otherwise the command timeouts
+        sssd_pred = SoSPredicate(self, services=["sssd"])
+        if self.test_predicate(self, pred=sssd_pred):
+            self.add_cmd_output("sssctl config-check")
```
bmr-cymru: This looks like it could be cleaner.. i.e.:

```python
        self.add_cmd_output("sssctl config-check", pred=sssd_pred)
```

sos/plugins/sssd.py
```diff
@@ -28,12 +29,17 @@ def setup(self):
...
+
+            domain = self.collect_cmd_output("sssctl domain-list")
```
bmr-cymru: Same here - `collect_cmd_output()` accepts a predicate so we could key off `domain['status'] == None` here to detect predicate failure and gate the domain stuff on that as well as `['status'] == 0`. Otherwise an early return on !predicate might look neater.

pmoravec: Havent realized this trick (what previously caused this implementation to be denied). PR updated accordingly.

sos/plugins/sssd.py
```diff
@@ -28,9 +29,14 @@ def setup(self):
             "/etc/sssd/conf.d/*.conf"
         ])
 
-        self.add_cmd_output("sssctl config-check")
+        # call sssctl commands only when sssd service is running,
+        # otherwise the command timeouts
+        sssd_pred = SoSPredicate(self, services=["sssd"])
+        self.add_cmd_output("sssctl config-check", pred=sssd_pred)
 
-        domain = self.collect_cmd_output("sssctl domain-list")
```
bmr-cymru: don't we need a `pred=sssd_pred` on this call too?

bmr-cymru: nm I'm blind today :-D

PR #1933, opened 2020-02-06
[openvswitch] Adding nuage-openvswitch to the list of packages for ovs plugin

valleedelisle:
Nokia Nuage has forked upstream OVS but basically, the same
commands should work. When a server is having nuage-ovs installed,
we don't execute the ovs plugin because it doesn't detect the
package.

---

### DIFF
--- a/sos/plugins/openvswitch.py
+++ b/sos/plugins/openvswitch.py
@@ -163,12 +163,13 @@ class OpenVSwitch(Plugin):
 
 class RedHatOpenVSwitch(OpenVSwitch, RedHatPlugin):
 
-    packages = ('openvswitch', 'openvswitch2.*', 'openvswitch-dpdk')
+    packages = ('openvswitch', 'openvswitch2.*',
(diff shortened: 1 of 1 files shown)


BryanQuigley:
I can't find packages -but Canonical is listed as a partner for them.

As the packagelist is literally the only different could we just move to one combined package list?

valleedelisle:
I'm not sure how the package would be named in the Canonical atmosphere. Can we just use `.*openvswitch.*` or there's a possible impact I'm not seeing?

BryanQuigley:
I actually don't see anything wrong with that - but we usually seem to try to be more specific.

Here is the list of in Ubuntu packages (although external providers might do more) - https://packages.ubuntu.com/search?keywords=openvswitch&searchon=names&suite=all&section=all

If we are already doing it with openvswitch2 package name - I don't see any compelling reason to not just do openvswitch.* at least.

valleedelisle:
Makes sense, I added it.

Thanks for the review.

PR #1934, opened 2020-02-11
[policies] Detect the RedHat OS variant using 'ID' field

rmetrich:
---

### DIFF
--- a/sos/policies/redhat.py
+++ b/sos/policies/redhat.py
@@ -246,6 +246,7 @@ class RHELPolicy(RedHatPolicy):
     distro = RHEL_RELEASE_STR
     vendor = "Red Hat"
     vendor_url = "https://access.redhat.com/support/"
+    id = "rhel"
     msg = _("""\
 This command will collect diagnostic and configuration \
 information from this %(distro)s system and installed \
@@ -268,6 +269,8 @@ support representative.
             release string at the beginning of the NAME field in the
             `/etc/os-release` file and returns ``True`` if it is
             found, and ``False`` otherwise.
+            In case the NAME field has been modified, check against
+            the ID field for "rhel".
 
             :returns: ``True`` if the host is running RHEL or ``False``
                       otherwise.
@@ -281,6 +284,11 @@ support representative.
                     (name, value) = line.split("=")
                     value = value.strip("\"'")
                     if value.startswith(cls.distro):
+                        return True
+                elif line.startswith("ID="):
+                    (name, value) = line.split("=")
+                    value = value.strip("\"'")
+                    if value.startswith(cls.id):
                         return True
         return False
 
@@ -326,6 +334,7 @@ class CentOsPolicy(RHELPolicy):
     distro = "CentOS"
     vendor = "CentOS"
     vendor_url = "https://www.centos.org/"
+    id = "centos"
 
 
 ATOMIC = "atomic"


rmetrich:
Hello,
Currently, RHELPolicy relies on `NAME*` field only to determine whether it's a RHEL or CentOS only.
Sometimes customers put their own `NAME` field, breaking sosreport.
An alternative to fix PR would be to rely on `/etc/redhat-release` respectively `/etc/centos-release`, similar to what is done on Fedora.
For RHEL, it would however require to check if there is no `/etc/centos-release` file since CentOS ships both `/etc/centos-release` and `/etc/redhat-release`.

sos/policies/redhat.py
```diff
@@ -412,13 +421,15 @@ class CentOsAtomicPolicy(RedHatAtomicPolicy):
     distro = "CentOS Atomic Host"
     vendor = "CentOS"
     vendor_url = "https://www.centos.org/"
+    id = "centos"
```
TurboTurtle: This would never be evaluated as this patch intends. `CentOsAtomicPolicy` subclasses `RedHatAtomicPolicy` which overrides the `check()` method of `RHELPolicy` where the changes for checking for the `id` field is being made.

I don't think we need to add an `id` check here anyways, for two reasons. 1) The current `check()` is looking for the string `Atomic` in the release file which should be sufficient. 2) Atomic Host is deprecated in favor of RHCOS, so I don't think we'll see too many rebuilds, much less those that change the release file.

sos/policies/redhat.py
```diff
@@ -412,13 +421,15 @@ class CentOsAtomicPolicy(RedHatAtomicPolicy):
...
 
 
 class FedoraPolicy(RedHatPolicy):
 
     distro = "Fedora"
     vendor = "the Fedora Project"
     vendor_url = "https://fedoraproject.org/"
+    id = "fedora"
```
TurboTurtle: Similarly, this would not be evaluated. First because `FedoraPolicy` subclasses `RedHatPolicy` and not `RHELPolicy` (intentional), and second because `FedoraPolicy` overrides `check()` and simply looks to see if `/etc/fedora-release` exists.

For Fedora, I'd say we don't ever need to check for ID, because the basic check is sufficient for all builds. If an admin is removing `/etc/fedora-release`, then there are likely major changes going on and at the very least we don't want to have sos misidentifying such a rebuild with the Fedora policy.

sos/policies/redhat.py
```diff
@@ -282,6 +285,11 @@ def check(cls):
                     value = value.strip("\"'")
                     if value.startswith(cls.distro):
                         return True
+                elif line.startswith("ID="):
+                    (name, value) = line.split("=")
+                    value = value.strip("\"'")
+                    if value.startswith(cls.id):
+                        return True
```
TurboTurtle: Ack.

rmetrich:
Hi @TurboTurtle,
what do you think of my other proposal instead: get rid off NAME (or ID) and check for `/etc/redhat-release` instead:
- For RHEL, would be existence of /etc/redhat-release but no /etc/centos-release
- For CENTOS, would be existence of /etc/centos-release

sos/policies/redhat.py
rmetrich: I agree, my intend was that it would still work if for some reason Fedora would again rely on content of `/etc/os-release` (the generic code) instead of `/etc/fedora-release`.

sos/policies/redhat.py
rmetrich: I agree, my intend was that it would still work if for some reason CentOS Atomic would again rely on content of /etc/os-release (the generic code)

TurboTurtle:
I don't think going purely on file existence for the whole RHEL family is the right approach. I like this change to handle end-user rebuild changes and I think that's a more robust stance to take.

bmr-cymru:
I think we should add something that indicates that this happened - the `/etc/os-release` file is *not* intended for local modifications because that subverts the entire purpose. While people may do that (and we must therefore cope with it), we ought to tell the user somehow that they are running a broken configuration.

That could be done in the policy, e.g. by reporting that we fell back to the ID match, or we could address it via plugins by (e.g. via package manager verification of the provider of the os-release).

BryanQuigley:
I'm doing something similar in PR #2113, thoughts welcome!

BryanQuigley:
I've hit some roadblocks in #2113 myself, but in doing the work I've determined we mostly don't care about what distro you are on - but the big thing seems to be what package manager(s) you use.

PR #1935, opened 2020-02-11
[Juju] Juju Plugin Refactoring

dnegreira:
Since the current plugin is having issues when we run sosreport
on a juju controller we have decided to refactor it and make
it simpler.

We had some discussions following three separate PRs:
https://github.com/sosreport/sos/pull/1670
https://github.com/sosreport/sos/pull/1671
https://github.com/sosreport/sos/pull/1672

We decided not to gather mongodb dumps as it will leak the
password from mongodb when running `ps`.

We decided to drop the output from juju commands as this
requires the command to be run as the user that holds the juju
credentials.

This plugin will gather logs from `/var/log/juju/*.log` and agent
configuration files from `/var/lib/juju/agents/*/agent.conf`, as
well as systemd logs related to the agents.
If sosreport is run with `--all-logs` it will gather all of the above
as well as all the files from `/var/lib/juju/` and `/var/log/juju`.

Closes: #1653

Co-authored-by: David Negreira david.negreira@canonical.com
Co-authored-by: Nick Niehoff nick.niehoff@canonical.com
Signed-off-by: David Negreira david.negreira@canonical.com

---

### DIFF
--- a/sos/plugins/juju.py
+++ b/sos/plugins/juju.py
@@ -8,27 +8,7 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-import os
 from sos.plugins import Plugin, UbuntuPlugin
-from json import loads as json_loads
-
-
-def ensure_service_is_running(service):
-    def wrapper(callback):
-        def wrapped_f(self, *args, **kwargs):
-            try:
-                result = self.call_ext_prog("service {0} stop".format(service))
-                if result["status"] != 0:
-                    raise Exception("Cannot stop {0} service".format(service))
-                callback(self, *args, **kwargs)
-            except Exception as ex:
-                self._log_error("Cannot stop {0}, exception: {1}".format(
-                    service,
-                    ex.message))
-            finally:
-                self.call_ext_prog("service {0} start".format(service))
-        return wrapped_f
-    return wrapper
 
 
 class Juju(Plugin, UbuntuPlugin):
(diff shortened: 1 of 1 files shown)


sos/plugins/juju.py
```diff
@@ -37,85 +19,59 @@ class Juju(Plugin, UbuntuPlugin):
...
-            self.add_cmd_output('ls -alRh /var/lib/juju/*')
+        # Get juju services, they are uniquely named based on the deployment
+        # so we need to parse what their names are.
+        cmd = "systemctl --no-pager --all --no-legend list-units juju\\*"
+        juju_systemctl_output = self.exec_cmd(cmd)["output"]
+        for line in iter(juju_systemctl_output.splitlines()):
+            match = re.match(r"^(juju\S+)\.", line)
+            if match:
+                self.add_journal(units=str(match.group(1)))
+                self.add_service_status(str(match.group(1)))
```
TurboTurtle: Assuming the `juju*` services are actually well-known names, this would be made a lot cleaner/simpler by using an `InitSystem` in the Ubuntu policy. E.G.

```python
class Juju(Plugin, UbuntuPlugin):
    services = ('juju', 'other_juju', 'another_juju_service')

    def setup(self):
    for service in services:
        self.add_journal(service)
```
Note that by defining the `services` class attr, you also ensure plugin enablement when any of those services are found.

And if they aren't well known, the `SystemdInit` class could be easily extended to allow searching the discovered services (and discovering them is already handled today)

In any event, `self.add_service_status` will _not_ work without an `InitSystem` being enabled by the policy.

sos/plugins/juju.py
```diff
@@ -37,85 +19,59 @@ class Juju(Plugin, UbuntuPlugin):
...
+
+        # Get agent configs for each agent.
+        agent_configs = []
+        for dirpath, dirnames, filenames in os.walk("/var/lib/juju/agents"):
+            for dirname in dirnames:
+                agent_configs.append(
+                    "/var/lib/juju/agents/" + dirname + "/agent.conf")
```
TurboTurtle: This can all be reduced to:
`self.add_copy_spec("/var/lib/juju/agents/*/agent.conf"`)

sos/plugins/juju.py
```diff
@@ -37,85 +19,59 @@ class Juju(Plugin, UbuntuPlugin):
...
+        self.add_copy_spec(agent_configs)
+
+        # Get a directory listing of /var/log/juju and /var/lib/juju
+        self.add_cmd_output(
+            ["ls -alRh /var/log/juju*", "ls -alRh /var/lib/juju*"]
+        )
```
TurboTurtle: sos-style here, please:

```python
self.add_cmd_output([
    'one_item',
    'another_item'
])

File: sos/plugins/juju.py [hunk 4]
```diff
@@ -37,85 +19,59 @@ class Juju(Plugin, UbuntuPlugin):
...
-        self.add_cmd_output([
-                "juju --version",
-                "juju -v status --format=tabular",
-        ])
-        for service in self.get_deployed_services():
-            self.add_cmd_output([
-                "juju get {}".format(service),
-                "juju get-config {}".format(service),
-                "juju get-constraints {}".format(service)
-            ])
```
TurboTurtle: Are all of these really no longer desired? Is there any command output from `juju` we should be collecting at this point, as this PR is nuking all command collect?

Note that I have absolutely no experience with juju personally so I don't have a horse in this race, but it seems odd to not collect any command output.

sos/plugins/juju.py
pmoravec: +1.

The only advantage of the current implementation is in no need to maintain the list of `juju` services (that needs some coordination/awareness/effort). But the cost is uglier code, slower to execute. Due to this argument, I would prefer the Jake's approach but I am OK also with the current.

sos/plugins/juju.py
BryanQuigley: The redesign will have this running only on machines managed by Juju, they won't even have the juju command present on them.  Yes it  does seem weird, but is fully intentional.

TurboTurtle: Ack.

sos/plugins/juju.py
dnegreira: Fixed on https://github.com/sosreport/sos/pull/1935/commits/45b69234af2f42292794acf36f5f264c4b5f5105

sos/plugins/juju.py
dnegreira: Fixed on https://github.com/sosreport/sos/pull/1935/commits/45b69234af2f42292794acf36f5f264c4b5f5105

sos/plugins/juju.py
dnegreira: Hi,
It seems that https://github.com/sosreport/sos/pull/1936 enables the policy by default and there is no need to initiate it on the plugin.

Unfortunately, there is no better way right now to find the correct juju-* services without this trick.

TurboTurtle: So the service names are not well-known? Seems strange, but we can remedy this by extending `SystemdInit` to provide a method to return service names matching a regex. I'll work on something for this today.

TurboTurtle: Opened #1943 for this. I'll merge that once we get another maintainer's ack and then this PR can be updated to use the new `get_service_names()` method.

Once that is done, this PR has my ack.

BryanQuigley: The service names are unfortunately generated - and contain a number based on the deployment..

With your PR. this just can let us make the whole bit something like:
for service in self.get_service_names("juju*"):
    self.add_journal(service)

Thanks!

dnegreira: Fixed on https://github.com/sosreport/sos/pull/1935/commits/cea2ccbbd985212b8c2c0aaea31fa14b5d267ce2
Did a test and the juju units are discovered and the journal and service status file are properly generated.

PR #1936, opened 2020-02-11
[policies] Detect systemd use instead of hardcoding it

BryanQuigley:
This just has us the builtin option to determine if we are on
systemd or not.
https://www.freedesktop.org/software/systemd/man/sd_booted.html

Resolves: #1936

---

### DIFF
--- a/sos/policies/__init__.py
+++ b/sos/policies/__init__.py
@@ -842,7 +842,8 @@ class LinuxPolicy(Policy):
     def __init__(self, sysroot=None):
         super(LinuxPolicy, self).__init__(sysroot=sysroot)
         self.init_kernel_modules()
-        if self.init == 'systemd':
+
+        if os.path.isdir("/run/systemd/system/"):
             self.init_system = SystemdInit()
         else:
             self.init_system = InitSystem()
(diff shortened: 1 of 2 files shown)


BryanQuigley:
I need to do more testing on this, but I don't see why not just use this determine if we are running systemd as init or not.

TurboTurtle:
Ack from me pending your further tests.

I originally did the manual enablement when this was all first added since I didn't want to speak for the Ubuntu/Debian distros. But this does make far more sense than manually enabling it in the individual policies.

BryanQuigley:
Testing has this change working fine - but on non-systemd systems is there any reason to not default to the service command?

TurboTurtle:
I didn't implement a `SysVInit` due to the fact I couldn't come up with a single distro that currently has a sos policy that uses it in their latest releases. For example RHEL 6 uses it, but a version of sos that includes this functionality won't be shipped on that release. I don't remember the last Fedora release to not use systemd off the top of my head - but same deal there.

I don't think there's any harm in implementing one, but I wouldn't spend any time on it personally.

BryanQuigley:
Good point - all of Ubuntu's are actually upstart.. so let's leave it as is.  Testing seems fine.

Issue #1921, opened 2020-01-18
lxd / travis issue


BryanQuigley:
This breaks the travis testing for #1896.  I haven't been able to reproduce the problem locally - perhaps some Travis specific thing?  

Why does the python callback think it's a directory, but ls/file shows it doesn't exist.  "snapshots" is not in the lxd code so it must have found it somehow from the Travis instance.

######### DONE WITH  --batch -z xz -t 1 --log-size 1 -n memory #########

Traceback (most recent call last):
  File "/home/travis/build/sosreport/sos/sos/sosreport.py", line 963, in setup
    plug.setup()
  File "/home/travis/build/sosreport/sos/sos/plugins/lxd.py", line 37, in setup
    "/var/log/lxd/*"
  File "/home/travis/build/sosreport/sos/sos/plugins/__init__.py", line 1017, in add_copy_spec
    self.add_string_as_file(tail(_file, sizelimit),
  File "/home/travis/build/sosreport/sos/sos/utilities.py", line 32, in tail
    with open(filename, "rb") as f:
IsADirectoryError: [Errno 21] Is a directory: '/var/lib/lxd/database/global/snapshots'

But trying to just the directory with travis comnands:
```echo "/var/lib/lxd/"
file /var/lib/lxd/
ls -alh /var/lib/lxd/

echo "/var/lib/lxd/database/"
file /var/lib/lxd/database/
ls -alh /var/lib/lxd/database/

echo "/var/lib/lxd/database/global"
file /var/lib/lxd/database/global
ls -alh /var/lib/lxd/database/global

echo "/var/lib/lxd/database/global/snapshots"
file /var/lib/lxd/database/global/snapshots
ls -alh /var/lib/lxd/database/global/snapshots```

Results in:
/var/lib/lxd/
/var/lib/lxd/: directory
total 8.0K
drwxr-xr-x  2 lxd  nogroup 4.0K Jan 18 20:16 .
drwxr-xr-x 49 root root    4.0K Jan 18 20:16 ..
srw-rw----  1 root lxd        0 Jan 18 20:16 unix.socket
/var/lib/lxd/database/
/var/lib/lxd/database/: cannot open `/var/lib/lxd/database/' (No such file or directory)
ls: cannot access '/var/lib/lxd/database/': No such file or directory
/var/lib/lxd/database/global
/var/lib/lxd/database/global: cannot open `/var/lib/lxd/database/global' (No such file or directory)
[...]

PR #1944, opened 2020-02-12
[lxd] Adapt the plugin for lxd snap installation

slashdd:
Nowadays, starting with Disco release, lxd deb package
is only use as a transitionning package converting
DEB to SNAP.

This fix adds the snap locations while still keeping
the plugin backward compatibility for 2.X and 3.X
DEB packages installation that are still exist,used
and supported by Canonical.

---

### DIFF
--- a/sos/plugins/lxd.py
+++ b/sos/plugins/lxd.py
@@ -34,7 +34,9 @@ class LXD(Plugin, UbuntuPlugin):
             "/var/lib/lxd/lxd.db",
             "/var/lib/lxd/database/local.db",
             "/var/lib/lxd/database/global/*",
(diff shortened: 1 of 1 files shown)


BryanQuigley:
I am not sure we should be collecting /var/lib/lxd/database/global/* or /var/snap/lxd/common/lxd/database/global/*.  See #1921 for why I've been investigating it.

What files in there are useful?  They seem like they might get big as well. Maybe a recursive ls would cover most cases?

slashdd:
I'll close this PR for now, and will re-evaluate later. There is no emergency for that to land in sosreport anytime soon.

Issue #1938, opened 2020-02-12
add_journal and add_cmd_output have different default timeout


pmoravec:
Not sure if it is a bug or a feature :) but `add_journal` calls the underlying command with `timeout=None` by default, whereas `add_cmd_output` has default timeout 300s. Is this desired?

`timeout=None` for `add_journal` is due to https://github.com/sosreport/sos/blob/master/sos/plugins/__init__.py#L1367 and https://github.com/sosreport/sos/blob/master/sos/plugins/__init__.py#L1452, while `timeout=300` for `add_cmd_output` is from https://github.com/sosreport/sos/blob/master/sos/plugins/__init__.py#L1083 .

And currently, we call `add_journal` without a `timeout` overwritten (see mainly https://github.com/sosreport/sos/blob/master/sos/plugins/logs.py#L62-L64).

On one side, we should apply a timeout to any command, by default, `journalctl` included. And having the journal timeout aligned with the default command timeout is reasonable.

On the other side, skipping collection of recent system logs just due to (too low or existing) timeout..?

TurboTurtle:
Definitely of the opinion this is a bug, and that `add_journal` should share the same timeout mechanics as `add_cmd_output`.

We don't want sos hanging forever on long-running `journalctl` commands, and the current situation a break from stated behavior ('all commands timeout...').

bmr-cymru:
Yep: bug, no doubt. Generally, our timeouts are a bit odd. For a long time the RHEL package and upstream had very different defaults, so anything that improves consistency and documentation is very welcome.

PR #1946, opened 2020-02-12
[plugins] set default cmd timeout to add_journal

pmoravec:
add_journal calls a sos command underneath that should apply same
command timeout limits like e.g. add_cmd_output method.

Further, define the command timeout value of 300 seconds just on
one place.

Resolves: #1938

---


BryanQuigley:
Committed in https://github.com/sosreport/sos/commit/dad96c5b8bf174216be5bb10f4b26cf23f374ab2

Apparently my format to have 2 Resolves on a line didn't work

Issue #1945, opened 2020-02-12
ceph plugin split for each individuals services


slashdd:
Might be good to eventually split ceph services into their own individuals plugin (osd, mgr, mon, radosgw, ...)

slashdd:
@hillpd

hillpd:
This would be a clean way to capture daemon-specific output (e.g. perf dump and the new dump_osd_network).

slashdd:
@hillpd let's talk this week about the ceph plugin split layout proposal:

ceph_osd.py
ceph_mgr.py
ceph_mds.py
ceph_mon.py
ceph_radosgw.py
and so on ....

The above copies the actual layout used for Openstack: `openstack_$OPENSTACK_COMPONENT.py`

Let's see how we can split the work together to make it happens for 4.0 (release due for this August IIRC)

slashdd:
@BryanQuigley any objections, at maintainer pov, if @hillpd starts working at splitting things up in ceph like we currently do for Openstack ?

hillpd:
@bmr-cymru I'm going to start work on splitting out the ceph plug-in per-service. The first iteration will provide in-place replacement of functionality. Then I want to add daemon-specific commands for better debug/troubleshooting. Let me know if you have any concerns with this approach.

slashdd:
A colleague of mine and @hillpd  is currently working on a patch for this particular issue. 
Expectations are to have a patch ready in a couple of weeks.

nkshirsagar:
I'm working on this issue with help from @hillpd @slashdd @pponnuvel . Shall sent out a patch for review once the changes are ready and tested.

PR #1947, opened 2020-02-12
[ceph] mimic/nautilus data collection update

hillpd:
Ceph has been focusing on administrator quality of life in the latest
releases. There are a number of shiny new tools/reports available that
would be great to have when troubleshooting ceph clusters.

This is a first pass at updating the ceph plug-in to bring it inline
with upstream changes.

This fix also addresses several inconsistencies in reports that were
captured between mon/mgr/osd services.

Further improvements are possible here, but I recommend splitting up
the ceph plug-in into service-specific modules. See Issue #1945.

Signed-off-by: Dan Hill <daniel.hill@canonical.com>

---

### DIFF
--- a/sos/plugins/ceph.py
+++ b/sos/plugins/ceph.py
@@ -32,7 +32,9 @@ class Ceph(Plugin, RedHatPlugin, UbuntuPlugin):
         'ceph-nfs@pacemaker',
         'ceph-mds@%s' % ceph_hostname,
         'ceph-mon@%s' % ceph_hostname,
-        'ceph-mgr@%s' % ceph_hostname
+        'ceph-mgr@%s' % ceph_hostname,
+        'ceph-radosgw@*',
+        'ceph-osd@*'
     )
 
     def setup(self):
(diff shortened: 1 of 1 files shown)


BryanQuigley:
Thanks Dan! Please squash Missing ',' into the other commit.

Would adding a self.add_service_status(service) make sense for line 119?

I'll try to get some more testing down tomorrow.

hillpd:
Yeah, I'm wrapping up my own testing. I will rebase and squash the commits as soon as I finish.

hillpd:
> Would adding a self.add_service_status(service) make sense for line 119?

I think that makes sense. The service status would be useful to in a sosreport captured with '-o ceph'

hillpd:
Tested this on focal + nautilus, bionic + nautilus, and xenial + luminous. The commands that were not implemented in luminous just captured error output (as expected).

I will note that the existing service listing is systemd init-specific. I have added osd + radosgw services to the list and am using systemd supported wildcards. It is an unusual configuration, but I have seen multiple radosgw unit instances on the same host.

PR #1948, opened 2020-02-13
[maas] Collect journal logs and service status infos

slashdd:

---

### DIFF
--- a/sos/plugins/maas.py
+++ b/sos/plugins/maas.py
@@ -19,6 +19,16 @@ class Maas(Plugin, UbuntuPlugin):
     profiles = ('sysmgmt',)
     packages = ('maas', 'maas-common')
 
+    services = (
+        'maas-dhcpd',
+        'maas-dhcpd6',
+        'maas-http',
+        'maas-proxy',
+        'maas-rackd',
+        'maas-regiond',
+        'maas-syslog'
+    )
+
     option_list = [
         ('profile-name',
          'The name with which you will later refer to this remote', '', ''),
@@ -55,6 +65,11 @@ class Maas(Plugin, UbuntuPlugin):
             "apt-cache policy maas-*",
             "apt-cache policy python-django-*",
         ])
+
+        self.add_service_status(self.services)
+
+        for service in self.services:
+            self.add_journal(units=service)
 
         if self.is_installed("maas-region-controller"):
             self.add_cmd_output([


sos/plugins/maas.py
```diff
@@ -56,6 +66,10 @@ def setup(self):
             "apt-cache policy python-django-*",
         ])
 
+        for service in self.services:
+            self.add_journal(units=service)
+            self.add_service_status(service)
```
pmoravec: Since both `add_journal` and `add_service_status` methods support lists as the `service` argument, you can simply call:

    self.add_journal(units=self.services)
    self.add_service_status(self.services)

Until you need a dedicated journal logfile for each individual service - then just `add_service_status` can be called to the list.

pmoravec:
The Travis build failed on `python2.7` only, on https://travis-ci.org/sosreport/sos/jobs/649774471#L2605 - the `free(): invalid pointer` does not sound like a `sosreport` bug itself, so I wont be much concerned about it..?

sos/plugins/maas.py
slashdd: done, thanks for the feedbacks.

slashdd:
We want one journal file in it's own file

slashdd:
The desire was to have each service in its own file. Sorry I intended to do this but mixed up my testing result. With that being said, the end goal is to have the following:

```
journalctl_--no-pager_--unit_maas-dhcpd
journalctl_--no-pager_--unit_maas-http   
journalctl_--no-pager_--unit_maas-rackd    
journalctl_--no-pager_--unit_maas-syslog
journalctl_--no-pager_--unit_maas-dhcpd6  
journalctl_--no-pager_--unit_maas-proxy  
journalctl_--no-pager_--unit_maas-regiond
```
I think it's more cleaner that way.

I'll re-submit the PR in a moment.

bmr-cymru:
The `add_journal()` interface is one of the few `add_*` methods where the behaviour is different depending on whether you pass a list to a single call, or make multiple calls for each entry in a list (both `add_copy_spec() and `add_cmd_output()` do the same thing either way).

I can see why you'd want to have things split out so maybe a possible refinement to this is to have an option in `add_journal()` to keep units in separate files? It's a minor cleanup buy may be useful for more than just `maas` (and it's in keeping with our usual convention of string/list munging for these calls).

PR #1951, opened 2020-02-13
[yum] Add "yum -C repolist --verbose" command output

kyle-walker:
Following the upstream DNF commit below, the "yum -C repolist" command
output will no longer include package count information:

    c4f57476 Do not load metadata for repolist commands...

This information is helpful for those searching for client-side
mismatches to upstream repositories. Such as in the case of two
repositories, where one is missing content that another includes, this
information can be crucial for determining the source of the problem.

This change includes a "yum -C repolist --verbose" command in the
resulting sosreport which includes information such as the following:

>    Repo-id            : repoid
>    Repo-name          : full reponame
>    Repo-revision      : version
>    Repo-updated       : datestamp
>    Repo-pkgs          : total repo package count
>    Repo-available-pkgs: package count
>    Repo-size          : size
>    Repo-baseurl       : url
>    Repo-expire        : datestamp
>    Repo-filename      : path

Signed-off-by: Kyle Walker <kwalker@redhat.com>

---

### DIFF
--- a/sos/plugins/yum.py
+++ b/sos/plugins/yum.py
@@ -41,6 +41,10 @@ class Yum(Plugin, RedHatPlugin):
         # Get a list of channels the machine is subscribed to.
         self.add_cmd_output("yum -C repolist")
 
+        # Get the same list, but with various statistics related to its
+        # contents such as package count.
+        self.add_cmd_output("yum -C repolist --verbose")
+
         # Get list of available plugins and their configuration files.
         if os.path.exists(YUM_PLUGIN_PATH) and os.path.isdir(YUM_PLUGIN_PATH):
             plugins = ""


sos/plugins/yum.py
```diff
@@ -41,6 +41,10 @@ def setup(self):
         # Get a list of channels the machine is subscribed to.
         self.add_cmd_output("yum -C repolist")
```
pmoravec: Isn't the non-verbose output redundant, then? This non-verbose variant has the only feature "all collected data of a repo are on one line", like:

    # yum -C repolist
    Updating Subscription Management repositories.
    Last metadata expiration check: 0:10:30 ago on Thu 13 Feb 2020 10:13:56 PM CET.
    repo id                                    repo name                      status
    RedHat_pulp_zoo_repo_pulp_zoo_repo         pulp_zoo_repo                     32
    rhel-8-for-x86_64-appstream-rpms           Red Hat Enterprise Linux 8 for 8,589
    rhel-8-for-x86_64-baseos-rpms              Red Hat Enterprise Linux 8 for 3,676
    satellite-tools-6.6-for-rhel-8-x86_64-rpms Red Hat Satellite Tools 6.6 fo    18
    #

vs. verbose output:

    # yum repolist -C --verbose
    ..
    Repo-id      : RedHat_pulp_zoo_repo_pulp_zoo_repo
    Repo-name    : pulp_zoo_repo
    Repo-revision: 1581322801
    Repo-updated : Mon 10 Feb 2020 09:20:02 AM CET
    Repo-pkgs    : 32
    Repo-size    : 77 k
    Repo-baseurl : https://pmoravec-sat66-on-rhev.gsslab.brq2.redhat.com/pulp/repos/RedHat/Library/custom/pulp_zoo_repo/pulp_zoo_repo
    Repo-expire  : 1 second(s) (last: Thu 13 Feb 2020 10:13:56 PM CET)
    Repo-filename: /etc/yum.repos.d/redhat.repo
    ..
    #

bmr-cymru:
I'm a bit confused - the commit in the pull description is for DNF, but the patch changes the `yum repolist` command in the `yum` plugin - I couldn't find an equivalent commit for the yum project itself, and the latest packages on my RHEL7 systems appear to not include this change.

Right now we do not collect repolist data for the `dnf` plugin at all so maybe that's where this change belongs?

sos/plugins/yum.py
kyle-walker: There are differences between the commands when you do not have cache on the system.

```
# rm -rf /var/cache/dnf/*
# yum -C repolist
Updating Subscription Management repositories.
repo id                                         repo name
rhel-8-for-x86_64-appstream-rpms                Red Hat Enterprise Linux 8 for x86_64 - AppStream (RPMs)
rhel-8-for-x86_64-baseos-rpms                   Red Hat Enterprise Linux 8 for x86_64 - BaseOS (RPMs)
```

Versus:

```
# rm -rf /var/cache/dnf/*
# yum -C repolist --verbose
Loaded plugins: builddep, changelog, config-manager, copr, debug, debuginfo-install, download, generate_completion_cache, needs-restarting, playground, product-id, repoclosure, repodiff, repograph, repomanage, reposync, subscription-manager, uploadprofile
Updating Subscription Management repositories.
YUM version: 4.2.17
cachedir: /var/cache/dnf
User-Agent: constructed: 'libdnf (Red Hat Enterprise Linux 8.1; generic; Linux.x86_64)'
Error: Cache-only enabled but no cache for 'rhel-8-for-x86_64-baseos-rpms'
```

We can't really just use the `--verbose` flag output in that instance since it will fail on the first repo without cache.

kyle-walker:
> I'm a bit confused - the commit in the pull description is for DNF, but the patch changes the `yum repolist` command in the `yum` plugin - I couldn't find an equivalent commit for the yum project itself, and the latest packages on my RHEL7 systems appear to not include this change.
> 
> Right now we do not collect repolist data for the `dnf` plugin at all so maybe that's where this change belongs?

The YUM project, which RHEL 7 is based on, is deprecated/obsoleted upstream. DNF is the successor to the project, and what RHEL 8 is based on. The change in DNF behaviour is the initial reason for submitting this PR, but is currently only in the upstream for the project since the change hasn't had the opportunity to be picked up in the downstream.

As to the reason for including it in yum; you can use either the `yum` or `dnf` cli endpoints to interact with the DNF stack for backwards-compatibility reasons.

```
# ls -l $(which yum) $(which dnf)
lrwxrwxrwx. 1 root root 5 Jan 14 02:15 /usr/bin/dnf -> dnf-3
lrwxrwxrwx. 1 root root 5 Jan 14 02:15 /usr/bin/yum -> dnf-3
```

Essentially, if you leave commands/plugins in place that are applicable to yum **and** dnf in the yum plugin, then you get full coverage. Since the `yum -C repolist --verbose` command output works and is useful on both stacks, I included it in the yum plugin.

bmr-cymru:
> The YUM project, which RHEL 7 is based on, is deprecated/obsoleted upstream. DNF is the successor to the project, and what RHEL 8 is based on.

Sure, but upstream sos is concerned with more than just RHEL7/8 - the information that's needed is currently spread between the yum and dnf plugins, but depending on what you're actually running there is a bit of a mish-mash of data collected:

* RHEL7 and earlier collects only the yum plugin.
* RHEL8 collects both yum and dnf
* Fedora 22 and later only collect dnf

We need to make sure that the information we collect is as consistent as the versions each distro provides allow, and that we don't rely on anything that may only be available in certain versions.

bmr-cymru:
To be clear: this isn't a problem with this specific change per-se, rather it's revealed that these two plugins have taken an unfortunate direction since the yum/dnf split. Originally, we separated the two out since we expected from the work in Fedora that DNF would take over from yum and replace it completely. With RHEL8 shipping both (even if yum is just a shim to dnf)  that hasn't turned out to be the case in practice.

It means we are either needlessly duplicating collection on some distros (RHEL8), and missing it entirely on others (Fedora).

I'm open to suggestions on this but I think we'll need to at least partially merge the two plugins to ensure that we have functional coverage for all distros.

kyle-walker:
> > The YUM project, which RHEL 7 is based on, is deprecated/obsoleted upstream. DNF is the successor to the project, and what RHEL 8 is based on.
> 
> Sure, but upstream sos is concerned with more than just RHEL7/8 - the information that's needed is currently spread between the yum and dnf plugins, but depending on what you're actually running there is a bit of a mish-mash of data collected:
> 
> * RHEL7 and earlier collects only the yum plugin.
> * RHEL8 collects both yum and dnf
> * Fedora 22 and later only collect dnf
> 
> We need to make sure that the information we collect is as consistent as the versions each distro provides allow, and that we don't rely on anything that may only be available in certain versions.

Actually, that isn't my experience at all. Using a brand-new Fedora 31 system:

```
[root@localhost sosreport-localhost-2020-02-14-illswir]# pwd
/root/sosreport-localhost-2020-02-14-illswir
[root@localhost sosreport-localhost-2020-02-14-illswir]# cat uname 
Linux localhost.localdomain 5.3.7-301.fc31.x86_64 #1 SMP Mon Oct 21 19:18:58 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
[root@localhost sosreport-localhost-2020-02-14-illswir]# ls -l sos_commands/yum/
total 48
-rw-r--r--. 1 root root     0 Feb 14 16:26 package-cleanup_--dupes
-rw-r--r--. 1 root root     0 Feb 14 16:26 package-cleanup_--problems
-rw-r--r--. 1 root root   474 Feb 14 16:26 yum_-C_repolist
-rw-r--r--. 1 root root  3440 Feb 14 16:26 yum_-C_repolist_--verbose
-rw-r--r--. 1 root root   480 Feb 14 16:26 yum_history
-rw-r--r--. 1 root root 36469 Feb 14 16:26 yum_list_installed
```

And checking from the sosreport command directly:
```
[root@localhost sos]# sosreport -l | grep -e "The following plugins" -e yum -e dnf
The following plugins are currently enabled:
 dnf                  dnf package manager
 yum                  yum information
The following plugins are currently disabled:
 dnf.history               off             captures transaction history
 dnf.history-info          off             detailed transaction history
 yum.yumlist               off             list repositories and packages
 yum.yumdebug              off             gather yum debugging data
 yum.yum-history-info      off             gather yum history info
```

I'm not 100% clear on the trigger mechanism for the plugins, but isn't it based at least in part on detection of the packages tuple defined in the plugin? If that is the case, on Fedora 31, you will have the `yum` and `dnf` plugins trigger due to the following packages installed:
```
# rpm -q yum dnf python3-dnf
yum-4.2.9-5.fc31.noarch
dnf-4.2.9-5.fc31.noarch
python3-dnf-4.2.9-5.fc31.noarch
```

Both are included in the `Core` group:
```
# yum repoquery --groupmember yum dnf
Last metadata expiration check: 0:31:06 ago on Fri 14 Feb 2020 04:21:22 PM UTC.
yum-4.2.18-1.fc31.noarch
yum-4.2.9-5.fc31.noarch
  @core
dnf-4.2.18-1.fc31.noarch
dnf-4.2.9-5.fc31.noarch
  @core
  @critical-path-base
  @workstation-product
```

I mean, I'm not against adding this type of repolist change to the dnf plugin, but based on what I see in Fedora I believe that would just cause duplication at the moment.

To the best of my knowledge, there isn't any plan to drop the `yum` compatibility layer in the foreseeable future. In fact, quite the opposite is true. The intention is to carry the `yum` cli entry point so that end users don't have to break years of built up muscle memory so that we can drop a single symlink.

bmr-cymru:
> Actually, that isn't my experience at all. Using a brand-new Fedora 31 system:

Well, it's what you got by default on Fedora 30 (and earlier if memory serves):

```
# cat /etc/fedora-release 
Fedora release 30 (Thirty)
# rpm -q yum dnf
package yum is not installed
dnf-4.2.17-1.fc30.noarch
# sosreport --list-plugins 2>&1 | egrep 'yum|dnf'
 dnf                  dnf package manager
 yum                  inactive       yum information
 dnf.history               off             captures transaction history
 dnf.history-info          off             detailed transaction history
```

So for at least some proportion of users the current situation is something of a mess.

kyle-walker:
Okay, I think I follow now. You are correct, the package name was `dnf-yum` in the `core` group up until fairly recently.

```
$ for comps in comps-f{20..31}.xml.in; do echo $comps;  xmllint --xpath '/comps/group[id="core"]' $comps | grep -e id -e yum; echo ; done
comps-f20.xml.in
    <id>core</id>
      <packagereq type="mandatory">yum</packagereq>

comps-f21.xml.in
    <id>core</id>
      <packagereq type="mandatory">yum</packagereq>

comps-f22.xml.in
    <id>core</id>
      <packagereq type="mandatory">dnf-yum</packagereq>

comps-f23.xml.in
    <id>core</id>
      <packagereq type="mandatory">dnf-yum</packagereq>

comps-f24.xml.in
    <id>core</id>
      <packagereq type="mandatory">dnf-yum</packagereq>

comps-f25.xml.in
    <id>core</id>
      <packagereq type="mandatory">dnf-yum</packagereq>

comps-f26.xml.in
    <id>core</id>
      <packagereq type="mandatory">dnf-yum</packagereq>

comps-f27.xml.in
    <id>core</id>
      <packagereq type="mandatory">dnf-yum</packagereq>

comps-f28.xml.in
    <id>core</id>
      <packagereq type="mandatory">dnf-yum</packagereq>

comps-f29.xml.in
    <id>core</id>
      <packagereq type="mandatory">dnf-yum</packagereq>

comps-f30.xml.in
    <id>core</id>
      <packagereq type="mandatory">dnf-yum</packagereq>

comps-f31.xml.in
    <id>core</id>
      <packagereq type="mandatory">yum</packagereq>
```

This is commit that accounted for the change in package name and it popping back up in Fedora 31:
```
commit 68a2f35b04e35c704ce253a2078c93a3d59a7a93
Author: Adam Williamson <awilliam@redhat.com>
Date:   Mon Oct 14 09:24:34 2019 -0700

    dnf-yum was renamed yum in Fedora 31+
    
    Signed-off-by: Adam Williamson <awilliam@redhat.com>
```

So, how would we like to proceed? Starting with the commit above in Fedora, you would expect that the `yum` and `dnf` plugins will enable. However, I do agree that the F22-F30 range needs to be accounted for as well.

Is there any way of disabling a plugin if another is already detected to be active? The reason I ask is that I would like to:

- Add the "repolist" commands to the `dnf` plugin
- Disable the `yum` plugin if the `dnf` plugin is active

That is, unless the duplication is not of concern at the moment. In which case I can add a commit just covering the former and leave the `yum` and `dnf` conflict for another day...

bmr-cymru:
To further confuse everyone, the `yum` package does in fact exist on f30, and seems to be the full-fat Python2 build in all its glory:

```
$ cat /etc/fedora-release 
Fedora release 30 (Thirty)
$ dnf -y install yum
[...]
$ rpm -qi yum
Name        : yum
Version     : 3.4.3
Release     : 522.fc30
Architecture: noarch
Install Date: Fri 14 Feb 2020 17:14:47 GMT
Group       : Unspecified
Size        : 5870441
License     : GPLv2+
Signature   : RSA/SHA256, Tue 21 May 2019 17:52:43 BST, Key ID ef3c111fcfc659b9
Source RPM  : yum-3.4.3-522.fc30.src.rpm
Build Date  : Tue 21 May 2019 17:44:34 BST
Build Host  : buildhw-05.phx2.fedoraproject.org
Relocations : (not relocatable)
Packager    : Fedora Project
Vendor      : Fedora Project
URL         : http://yum.baseurl.org/
Bug URL     : https://bugz.fedoraproject.org/yum
Summary     : RPM package installer/updater/manager
Description :
Yum is a utility that can check for and automatically download and
install updated RPM packages. Dependencies are obtained and downloaded
automatically, prompting the user for permission as necessary.
```

Which gives us four possible combinations: native yum or dnf, and either dnf and native yum, or dnf and a yum shim. I'm not sure yum and dnf together makes sense to support but the others are potentially valid combinations for distros that we support.

The command adds a couple of seconds to a run for my RHEL8 test systems and a bit more on RHEL7 so I'm happy to merge it for now if it solves a real need, but I think we should fork of an issue to review these two plugins with a view to merging them - at the time the split happened there was a definite effort to replace yum, but in practice that hasn't happened.

bmr-cymru:
> Is there any way of disabling a plugin if another is already detected to be active?

Not today; there have been various discussions about different uses for some form of dependency between plugins but today everything executes independently.

bmr-cymru:
Filed issue #1954 to review the duplication between yum & dnf.

PR #1953, opened 2020-02-14
[policies] fix LinuxPolicy performance regression (test suite)

bmr-cymru:
Commit 285873a introduces a regression in the performance of
LinuxProfile class initialisation:

  commit 285873a4f753822a88d475a1b030ab622bf4c72e
  Author: Bryan Quigley <bryan.quigley@canonical.com>
  Date:   Tue Feb 11 15:03:16 2020 -0800

      [policies] Detect systemd use instead of hardcoding it

All the patch does is to switch from testing self.init to
determine whether to use the SystemdInit() class or the
basic InitSystem() one, to looking for '/run/systemd/system'
in the file system.

This has more broad side effects than it might seem since
the test suite uses LinuxPolicy() as a mock policy object
for a large number of tests. Since SystemdInit() calls out
to systemctl to obtain the state of the init system this
both increases the run time for the tests and causes high
resource use in systemd itself:

PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
  1 root      20   0  177196  13504   4512 R  92.7   0.1   5:24.04 systemd

This causes the run time for the Python2 and 3 test suite
(plus pep8/pycodestyle) to grow from ~5s to over 1m20s on
my test system:

285873a~1 real 0m5.683s
285873a   real 1m20.353s

Allow direct users of LinuxPolicy to override the init
system detection by specifying an init= kwarg initialised
to an InitSystem-like object directly, and use this in the
test suite to avoid the cost of initialising the full
SystemdInit() class.

---

### DIFF
--- a/sos/policies/__init__.py
+++ b/sos/policies/__init__.py
@@ -846,11 +846,13 @@ class LinuxPolicy(Policy):
 
     _preferred_hash_name = None
 
-    def __init__(self, sysroot=None):
+    def __init__(self, sysroot=None, init=None):
         super(LinuxPolicy, self).__init__(sysroot=sysroot)
         self.init_kernel_modules()
 
-        if os.path.isdir("/run/systemd/system/"):
+        if init is not None:
+            self.init_system = init
+        elif os.path.isdir("/run/systemd/system/"):
             self.init_system = SystemdInit()
         else:
             self.init_system = InitSystem()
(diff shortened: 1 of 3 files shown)


bmr-cymru:
As all checks have passed and this is holding up the process of merging the remaining items for 3.9 I'm going to go ahead and merge this without the usual number of acks.

Issue #1687, opened 2019-05-29
[libvirt] Capture process/thread list when vcpu pinning is enabled


links84:
Thread list along with PSR information is very crucial information for NFV/OpenStack troubleshooting to identify if cpus dedicated for KVM or ovs-dpdk is being used by any other process.
Since all openstack deployments with ovd-dpdk are expected to have vms with cpu pinning, we can capture this information in libvirt plugin.

links84:
Can someone assign this to me?

TurboTurtle:
No need to worry about issue assignment - we have almost never used them. If you have a patch for this feel free to open the PR and just reference this issue in the commit message (e.g. 'Fixes: #1687'). 

But also just know that you don't need to open an issue to submit a PR - the commit message with the PR should serve to document the need for the change it's making, so opening a separate issue becomes a bit redundant.

links84:
ack. thanks.

Issue #1322, opened 2018-05-27
Wrap json.load() in with clause


bmr-cymru:
We have a few uses of the JSON module in `sos` (and possibly others) that use a "naked" open, e.g.:

```python
    json.loads(open("foo"))
   ...
```

This is bad since the file object just goes out of scope and is never closed properly.

Fix this by putting it in a `with` clause:

```python
    with open("foo") as f:
        json.loads(f)
        ...
```

PR #1962, opened 2020-02-20
[openvswitch] Collect ovs threads list in ps and in procfs

links84:
Get detailed ps output of all ovs-vswitchd threads including last cpu
known to run on, priority etc from ps.Also copy
/proc/<ovs thread id>/{environ,cgroup,maps,numa_maps,limits,status}
to indicate configured cpu affinity and other details.
These information are critical for troubleshooting NFVi openstack deployments.
    
Closes: #1687

---

### DIFF
--- a/sos/plugins/openvswitch.py
+++ b/sos/plugins/openvswitch.py
@@ -42,6 +42,18 @@ class OpenVSwitch(Plugin):
             "/run/openvswitch/ovs-vswitchd.pid"
         ])
 
+        # Collect ovs-vswitchd's pmd thread details in procfs.
+        proc_files = ["environ", "cgroup", "maps", "numa_maps", "limits",
+                      "status"]
+        try:
+            pid = open("/var/run/openvswitch/ovs-vswitchd.pid").read()\
+                    .splitlines()[0]
+            # List TID & thread command name.
+            self.add_copy_spec(["/proc/%s/task/*/%s"
+                                % (pid, pf) for pf in proc_files])
+        except IOError:
+            self._log_error("Could not read ovs-vswitchd PID file")
+
         self.add_cmd_output([
             # The '-s' option enables dumping of packet counters on the
             # ports.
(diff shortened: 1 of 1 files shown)


sos/plugins/openvswitch.py
```diff
@@ -42,6 +42,18 @@ def setup(self):
             "/run/openvswitch/ovs-vswitchd.pid"
         ])
 
+        # Collect ovs-vswitchd's pmd thread details in procfs.
+        proc_files = ["environ", "cgroup", "maps", "numa_maps", "limits",
+                      "status"]
+        try:
+            pid = open("/var/run/openvswitch/ovs-vswitchd.pid").read()\
+                    .splitlines()[0]
```
TurboTurtle: Like with #1322, this could potentially leave an open file object. Let's wrap this in a `with` clause, E.G.

```python
with open("/var/run/openvswitch/ovs-vswitchd.pid", 'r'), as pidfile:
    pid = pidfile.readline()
    self.add_copy_spec([...])
```

sos/plugins/openvswitch.py
```diff
@@ -95,7 +107,10 @@ def setup(self):
             # Capture DPDK pmd stats
             "ovs-appctl dpif-netdev/pmd-stats-show",
             # Capture DPDK pmd performance counters
-            "ovs-appctl dpif-netdev/pmd-perf-show"
+            "ovs-appctl dpif-netdev/pmd-perf-show",
+            # Collect ovs-vswitchd thread details in ps command.
+            "ps -p %s -Lo pid,tid,user,pcpu,psr,trs,rss,size,sz,drs,vsz,\
+maj_flt,min_flt,class,rtprio,ni,pri,s,stat,wchan,flags,comm" % pid
```
TurboTurtle: Can this be indented to stay with sos-style?

links84: Adding space between ps options will break the ps command. 
Also the ps options are so long that having them in one line will conflict pep8. Would that be okay? Eg:

            "ps -p %s -Lo \
            pid,tid,user,pcpu,psr,trs,rss,size,sz,drs,vsz,maj_flt,min_flt,class,rtprio,ni,pri,s,stat,wchan,flags,comm" \
            % pid

bmr-cymru: You just need to build the final argument string piece-wise: there are various examples of this in sos already, e.g. from the `process` plugin:

```python
        ps_axo = "ps axo"
        # process group and thread options
        ps_group_opts = "pid,ppid,user,group,lwp,nlwp,start_time,comm,cgroup"
        ps_sched_opts = "flags,state,uid,pid,ppid,pgid,sid,cls,pri,addr,sz,"
        ps_sched_opts += "wchan:20,lstart,tty,time,cmd"
[...]
        self.add_cmd_output([
            "ps auxwwwm",
            "ps alxwww",
            "ps -elfL",
            "%s %s" % (ps_axo, ps_group_opts),
            "%s %s" % (ps_axo, ps_sched_opts)
        ])
```

That said, I'm not a huge fan of spreading more `ps` collection around different plugins if it can be avoided - there is a great deal over overlap here between existing data that's collected.

links84: I dont mind adding these options to process plugin's ps_sched_opts, if you find that should be alright.

PR #1964, opened 2020-02-21
[maas-snap] Add snap support to MAAS plugin

sparkiegeek:
From 2.7.0 onwards, MAAS is using snaps for it's releases, and this changes the location of files and logs etc.

Signed-off-by: Adam Collard <adam.collard@canonical.com>

---

### DIFF
--- a/sos/report/plugins/maas.py
+++ b/sos/report/plugins/maas.py
@@ -20,13 +20,16 @@ class Maas(Plugin, UbuntuPlugin):
     packages = ('maas', 'maas-common')
 
     services = (
+        # For the deb:
         'maas-dhcpd',
         'maas-dhcpd6',
         'maas-http',
         'maas-proxy',
         'maas-rackd',
         'maas-regiond',
-        'maas-syslog'
+        'maas-syslog',
+        # For the snap:
+        'snap.maas.supervisor',
     )
 
     option_list = [
@@ -52,19 +55,49 @@ class Maas(Plugin, UbuntuPlugin):
 
         return ret['status'] == 0
 
+    def _is_snap_installed(self):
+        return self.exec_cmd('snap list maas')["status"] == 0
+
+    def check_enabled(self):
+        if super().check_enabled():
+            # deb-based MAAS and existing triggers
+            return True
+        # Do we have the snap installed?
+        return self._is_snap_installed()
+
     def setup(self):
-        self.add_copy_spec([
-            "/etc/squid-deb-proxy",
-            "/etc/maas",
-            "/var/lib/maas/dhcp*",
-            "/var/log/apache2*",
-            "/var/log/maas*",
-            "/var/log/upstart/maas-*",
-        ])
-        self.add_cmd_output([
-            "apt-cache policy maas-*",
-            "apt-cache policy python-django-*",
-        ])
+        self._is_snap = self._is_snap_installed()
+        if self._is_snap:
+            self.add_cmd_output([
+                'snap info maas',
+                'maas status'
+            ])
+            # Don't send secrets
+            self.add_forbidden_path("/var/snap/maas/current/bind/session.key")
+            self.add_copy_spec([
+                "/var/snap/maas/common/log",
+                "/var/snap/maas/common/snap_mode",
+                "/var/snap/maas/current/*.conf",
+                "/var/snap/maas/current/bind",
+                "/var/snap/maas/current/http",
+                "/var/snap/maas/current/supervisord",
+                "/var/snap/maas/current/preseeds",
+                "/var/snap/maas/current/proxy",
+                "/var/snap/maas/current/rsyslog",
+            ])
+        else:
+            self.add_copy_spec([
+                "/etc/squid-deb-proxy",
+                "/etc/maas",
+                "/var/lib/maas/dhcp*",
+                "/var/log/apache2*",
+                "/var/log/maas*",
+                "/var/log/upstart/maas-*",
+            ])
+            self.add_cmd_output([
+                "apt-cache policy maas-*",
+                "apt-cache policy python-django-*",
+            ])
 
         for service in self.services:
             self.add_journal(units=service)
@@ -83,7 +116,11 @@ class Maas(Plugin, UbuntuPlugin):
                     "Cannot login into MAAS remote API with provided creds.")
 
     def postproc(self):
-        self.do_file_sub("/etc/maas/regiond.conf",
+        if self._is_snap:
+            regiond_path = "/var/snap/maas/current/maas/regiond.conf"
+        else:
+            regiond_path = "/etc/maas/regiond.conf"
+        self.do_file_sub(regiond_path,
                          r"(database_pass\s*:\s*)(.*)",
                          r"\1********")
 


sos/plugins/maas.py
```diff
@@ -53,18 +53,42 @@ def _remote_api_login(self):
...
-            "/var/lib/maas/dhcp*",
-            "/var/log/apache2*",
-            "/var/log/maas*",
-            "/var/log/upstart/maas-*",
-        ])
-        self.add_cmd_output([
-            "apt-cache policy maas-*",
-            "apt-cache policy python-django-*",
-        ])
+        snap_list_output = self.exec_cmd('snap list maas')
```
pmoravec: I think it deserves some comment why the plugin behaves so differently based on the outcome of this command.

pmoravec: Further, as this "if" statement changes the plugin behaviour so heavily, does not it make sense to have it e.g. a child/parent class for it? It sounds strange if the same plugin is expected to do quite different stuff during different executions..? (I am OK with the current code, just.. won't a child plugin be a better option?)

sparkiegeek: Good point, have added some commentary.

sos/plugins/maas.py
```diff
@@ -53,18 +53,48 @@ def _remote_api_login(self):
...
+        if snap_list_output["status"] == 0:
```
bmr-cymru: I'd consider calling this something other then `_output`; historically names like `_out` and `_output` were common when we just returned command output, but today it's a dictionary of status and output data - I would suggest just calling it `snap_list` to avoid confusion.

sos/plugins/maas.py
```diff
@@ -53,18 +53,48 @@ def _remote_api_login(self):
...
+                "/var/snap/maas/current/preseeds",
+                "/var/snap/maas/current/proxy",
+                "/var/snap/maas/current/rsyslog",
+            ])
+            # As of 2.7.0 the only systemd service that is in the snap
+            # is supervisor, the processes that it maintains are shown
+            # in the output of `maas status`.
+            self.services = (
+                'snap.maas.supervisor',
+            )
```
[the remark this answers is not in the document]
pmoravec: That was my first thought as well that plugin enablement makes no sense here :) Imho the purpose is collecting the services status and journal logs few lines below : https://github.com/sosreport/sos/blob/master/sos/plugins/maas.py#L69-L72 . But is is quite confusing..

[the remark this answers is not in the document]
bmr-cymru: Sure but the `services` member has its own meaning - since this reference is only used in the scope of `setup()` it's confusing to also set the class member at this point.

sos/plugins/maas.py
```diff
@@ -53,18 +53,48 @@ def _remote_api_login(self):
...
-            "apt-cache policy python-django-*",
-        ])
+        snap_list_output = self.exec_cmd('snap list maas')
+        if snap_list_output["status"] == 0:
+            # We're in the snap, so file paths are different, and all
+            # the related software (i.e. bind, dhcpd) are configured
+            # and maintained by the snap
+            self.add_cmd_output([
+                'snap info maas',
+                'maas status'])
```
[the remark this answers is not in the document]
sparkiegeek: Done, thanks

sos/plugins/maas.py
```diff
@@ -53,18 +53,48 @@ def _remote_api_login(self):
...
+            # Don't send secrets
+            self.add_forbidden_path([
+                "/var/snap/maas/current/bind/session.key"
+            ])
```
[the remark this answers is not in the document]
sparkiegeek: Fair enough, given the first thing this method does is turn a string into a list I figured I'd just create a list to start with :)

sos/plugins/maas.py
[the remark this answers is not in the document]
sparkiegeek: Right, I didn't want the plugin to attempt to get status of services that don't exist in the snap world.

Of course this adds more to the argument that it should be a separate plugin

sos/plugins/maas.py
sparkiegeek: Renamed, thanks.

slashdd:
@BryanQuigley can you help/unblock sparkiegeek on this maas PR ?

sparkiegeek:
Have taken the approach of a child plugin, as suggested by @bmr-cymru - please let me know if this is a better fit?

BryanQuigley:
I think if it's that different, it should just be a fully separate plugin - in my test run I couldn't tell what was done from the original plugin.  What is the advantage of a child plugin (it's new to me...)?

* All these services are run by the one supervisor service?  Is it possible to just stop/start/status a single service?  (I mean this covers ntp/dhcp/postrgres/etc.... that's a lot)
* My preference would be to not activate depending on the service status, but if the maas command exists.  commands = ('maas',)   That is what we have de-facto done for most other snaps so far (or checking for path).

BryanQuigley:
> 1. Define a SnapPlugin class that is similar to SCLPlugin, where the methods used automatically substitute the snap paths.

The snap plugins are completely different from what we see.  but in my view means that we don't need to have Maas<>MaasSnap related at all.

Unfortunately, with a "fork" of postgres/dhcp/etc living within the snap, maybe that's where the SCL bit could help.  

> 2. Define a PackageManager for Snaps, that is then exposed to the distributions that use them.

Yes, we need to figure that out, but I've tried to figure that out several times and failed.  For now checking for if command in path is the way to go.

sparkiegeek:
> I think if it's that different, it should just be a fully separate plugin - in my test run I couldn't tell what was done from the original plugin. What is the advantage of a child plugin (it's new to me...)?

Well that was advocated for in https://github.com/sosreport/sos/pull/1964#discussion_r382616183

Right now it's sharing code in how to grab info out of MAAS using the CLI (`Maas._has_login_options` and `_remote_api_login`).

> 
>     * All these services are run by the one supervisor service?  Is it possible to just stop/start/status a single service?  (I mean this covers ntp/dhcp/postrgres/etc.... that's a lot)

Not right now. 
> 
>     * My preference would be to not activate depending on the service status, but if the maas command exists.  commands = ('maas',)   That is what we have de-facto done for most other snaps so far (or checking for path).

But the `maas` command can exist from both the snap and the deb. It's not a unique identifier for this plugin to be used. 

I'll note that it's not using the service status, as we can see from `check_enabled`, it's running based on the maas snap being installed.

BryanQuigley:
Doh', sorry for that.  I'm also not as familiar with MAAS as would be useful for this review :(.  It might be useful to get an internal meeting setup

>I'll note that it's not using the service status, as we can see from check_enabled

@slashdd correct me if I'm wrong but we never got buy-in from snap team for a specific way that will never change to confirm if a package is installed or not ?
As such we've previously used similar methods to files="/snap/lxd/" I guess that could work for maassnap too.

>Right now it's sharing code in how to grab info out of MAAS using the CLI (Maas._has_login_options and _remote_api_login).

Understood, let's just add a comment to the code to that effect.  It's also not importing that much - I'd be ok having it duplicated for clarity, but am unsure how others feel about that.

BryanQuigley:
We agreed internally that we're keep 1 plugin per application until we determine otherwise.  Will land this and then relook at it as snap packagemanager is implemented.
Relying on snap list <pluginname> status 0 as we believe that's better then checkin if in path or if /snap/maas exists
All the files collect in snap are snap specific and not currently useful to collect via child plugin/SCL like  method IMHO
Get services for everything to see if they are in a transitional state.

Can I get a +1 on this before asking for another round of changes?

BryanQuigley:
Got a +1 on IRC from TurboTurtle, namely:
(10:03:45 AM) turboturtle: but yeah, my view would be to carry this all in the existing maas plugin. It'll likely be a slightly ugly check_enablement() override until we can nail down a snap PackageManager()

sos/report/plugins/maas.py
```diff
@@ -19,16 +19,22 @@ class Maas(Plugin, UbuntuPlugin):
...
         'maas-regiond',
-        'maas-syslog'
+        'maas-syslog',
     )
 
+    snap_services = (
+        'snap.maas.supervisor',
+    )
+
+    services = deb_services + snap_services
```
BryanQuigley, line 36: I think we can just do one services list and collect all of them every time (not if/then on snap/deb).  Information on non-existent services does not get collected (just tried it out).

BryanQuigley:
Just did a range of testing (1 maas snap, 1 maas server) and overall looks good - just my 1 comment on services is all.

sos/report/plugins/maas.py
sparkiegeek: OK - I've squashed them back together and commented to highlight expectations

PR #1967, opened 2020-02-25
[openvswitch] Collect ovs threads list in ps and in procfs

### LINKED ISSUE #1687: [libvirt] Capture process/thread list when vcpu pinning is enabled

links84:
Get detailed ps output of all ovs-vswitchd threads including last cpu
known to run on, priority etc from ps. Also copy 
/proc/<ovs-vswitchd thread id>/{environ,cgroup,maps,numa_maps,limits,
status} to indicate configured cpu affinity and other details.
These information are critical for troubleshooting NFVi openstack
deployments.

Fixes: #1687

---

- [x ] If this commit closes an existing issue, is the line `Closes: #ISSUENUMBER` included in an independent line?
- [x ] If this commit resolves an existing pull request, is the line `Resolves: #PRNUMBER` included in an independent line?

### DIFF
--- a/sos/plugins/openvswitch.py
+++ b/sos/plugins/openvswitch.py
@@ -41,6 +41,18 @@ class OpenVSwitch(Plugin):
             "/run/openvswitch/ovsdb-server.pid",
             "/run/openvswitch/ovs-vswitchd.pid"
         ])
+        proc_files = ["environ", "cgroup", "maps", "numa_maps", "limits",
+                      "status"]
+        with open("/var/run/openvswitch/ovs-vswitchd.pid", 'r') as pidfile:
+            pid = pidfile.readline().splitlines()[0]
+            # List TID & thread command name.
+            self.add_copy_spec(["/proc/%s/task/*/%s"
+                                % (pid, pf) for pf in proc_files])
+
+        ps_Lo = "ps -p %s -Lo" % pid
+        ps_sched_opts = "pid,tid,user,pcpu,psr,trs,rss,size,sz,drs,vsz,"
+        ps_sched_opts += "maj_flt,min_flt,class,rtprio,ni,pri,stat,wchan:20,"
+        ps_sched_opts += "flags,comm"
 
         self.add_cmd_output([
             # The '-s' option enables dumping of packet counters on the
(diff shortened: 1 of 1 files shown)


sos/plugins/openvswitch.py
```diff
@@ -41,6 +41,18 @@ def setup(self):
...
+        with open("/var/run/openvswitch/ovs-vswitchd.pid", 'r') as pidfile:
+            pid = pidfile.readline().splitlines()[0]
+            # List TID & thread command name.
+            self.add_copy_spec(["/proc/%s/task/*/%s"
+                                % (pid, pf) for pf in proc_files])
+
+        ps_Lo = "ps -p %s -Lo" % pid
+        ps_sched_opts = "pid,tid,user,pcpu,psr,trs,rss,size,sz,drs,vsz,"
+        ps_sched_opts += "maj_flt,min_flt,class,rtprio,ni,pri,stat,wchan:20,"
+        ps_sched_opts += "flags,comm"
```
TurboTurtle: I think this is better served going into the `process` plugin.

We do 5 total `ps` calls at the moment, I'd say it's time for a review of those, especially with this new addition, to see where we can merge or drop certain collections. 

That review wouldn't be on you here. In any event however, I don't think adding these calls to individual plugins is the right approach.

sos/plugins/openvswitch.py
```diff
@@ -41,6 +41,18 @@ def setup(self):
             "/run/openvswitch/ovsdb-server.pid",
             "/run/openvswitch/ovs-vswitchd.pid"
         ])
+        proc_files = ["environ", "cgroup", "maps", "numa_maps", "limits",
+                      "status"]
+        with open("/var/run/openvswitch/ovs-vswitchd.pid", 'r') as pidfile:
+            pid = pidfile.readline().splitlines()[0]
+            # List TID & thread command name.
+            self.add_copy_spec(["/proc/%s/task/*/%s"
+                                % (pid, pf) for pf in proc_files])
```
TurboTurtle: Please use sos-style here, E.G.

```python
self.add_copy_spec([
    "/proc/%s/task/*/%s" % (pid, pf) for pf in proc_files
])
```

TurboTurtle (changes requested):

arif-ali:
@links84 it's been a while since this PR was touched, any thoughts on updating and merging this?

TurboTurtle:
Closing as abandoned.

PR #1970, opened 2020-02-28
[Plugin] New Validation Framework plugin

cjeanner:
The Validation Framework provides an overview of deployed services.

For now, this Framework is linked to OpenStack and TripleO[1], but since
it might be used by other products, it's probably better to get a generic
name.

---

### DIFF
--- a/sos/plugins/validation_framework.py
+++ b/sos/plugins/validation_framework.py
@@ -1,0 +1,40 @@
+# Copyright (C) 2020 Red Hat, Inc., Cedric Jeanneret <cjeanner@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.plugins import Plugin, RedHatPlugin
+
+
+# Notes:
+# - The Validation Framework is, for now, linked to openstack and tripleo
+# - Since the intent is to open it to other product (we can validate anything)
+#   this plugin has a generic name.
+# - The Framework is targeted at Red Hat products, at least for now.
+class ValidationFramework(Plugin, RedHatPlugin):
+    """Fetch run logs provided by the Validation Framework
+    """
+
+    plugin_name = 'validation_framework'
+    profiles = ('openstack', 'openstack_controller', 'openstack_compute')
+    packages = ('tripleo-validations',)
+
+    def setup(self):
+        self.add_copy_spec('/var/log/validations/')
+
+    def postproc(self):
+        # Use a generic match in order to clean things up.
+        # It is not expected to get any secrets in here, but we'd better
+        # ensure it's clean.
+        secrets = r'(".*(key|password|pass|secret|database_connection))' \
+                  r'([":\s]+)(.*[^"])([",]+)'
+
+        self.do_path_regex_sub('/var/log/validations/',
+                               secrets, r'\1\3*********\5')
+
+# vim: set et ts=4 sw=4 :


sos/plugins/validations.py
```diff
@@ -0,0 +1,44 @@
...
+# Notes:
+# - The Validation Framework is, for now, linked to openstack and tripleo
+# - Since the intent is to open it to other product (we can validate anything)
+#   this plugin has a generic name.
+# - The Framework is targeted at Red Hat products, at least for now.
+class Validations(Plugin, RedHatPlugin):
+    """Fetch run logs provided by the Validation Framework
+    """
+
+    plugin_name = 'validations'
```
TurboTurtle: I'm not a huge fan of this plugin name. 'Validations', to me at least, makes it sound like sos is doing some, well, validation of the system by itself.

On the other hand, `validations_framework` is bit hefty for a plugin name - but `openstack_validations` isn't any better.

sos/plugins/validations.py
```diff
@@ -0,0 +1,44 @@
...
+    profiles = ('openstack', 'openstack_controller', 'openstack_compute')
+    packages = ('tripleo-validations',)
+
+    def setup(self):
+        self.add_copy_spec([
+            '/var/log/validations/',
+        ])
```
TurboTurtle: Single item calls can be a string. `add_copy_spec` supports both.

sos/plugins/validations.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def postproc(self):
+        # Use a generic match in order to clean things up.
+        # It is not expected to get any secrets in here, but we'd better
+        # ensure it's clean.
+        secrets = r'(".*(key|password|pass|secret|database_connection))' \
+                  r'([":\s]+)(.*[^"])([",]+)'
+        rgxp = re.compile(secrets, re.IGNORECASE)
```
TurboTurtle: `do_path_regex_sub` already does this compilation step, so no need to do it here and we can drop the `re` import for this plugin.

sos/plugins/validations.py
cjeanner: I know :(. Also gave some thoughts when creating this patch, and, well... "let's push it like that and get feedbacks".
Unfortunately, we don't have another name for it, at least not official - I came with "valhalla[1]" as an acronym, but that's probably not better ;).

[1] Validate A Lot Hence Avoid Loads of Liabilities and Accounts

sos/plugins/validations.py
cjeanner: ah, I didn't know about that one, and it wasn't pointed when I created another plugin lately. Lemme correct it.

sos/plugins/validations.py
cjeanner: ah, right - copy-paste is bad ;).

sos/plugins/validations.py
cjeanner: @EmilienM any idea maybe?

EmilienM: A few ideas:
* infra_validations
* cloud_validations

Or just "tripleo_validations"; as for now the validations are part of OpenStack TripleO governance.

cjeanner: Thank you @EmilienM ! infra_validations might fit: generic enough to avoid renaming in the future, and not generic enough to cause issues.
@TurboTurtle would "infra_validations" be OK?

TurboTurtle: I'd say I prefer the longer `validation_framework`, but I'm definitely open to further suggestions.

The plugin looks good to me right now pending a decision on the name. Let's leave this open for a little bit to get some more feedback/suggestions on that end.

cjeanner: hello there! Is there anyone we could ping in order to get feedback about that name "issue"?

pmoravec: We were discussing it internally. TL;DR the consensus is that the `validation_framework` is the best from "not great" names, while we dont have any great one.

cjeanner: hello @pmoravec ! thanks for the update, I'll push that new name shortly then - at least there's some kind of consensus :)

Issue #1971, opened 2020-03-04
Collect the disk usage of "/var/lib/qpidd/.qpidd/"


waldirio:
Basically, the output of the below command
```
# du -sh /var/lib/qpidd/.qpidd/*
```

TurboTurtle:
Closing this issue as the linked PR was closed with another solution.

PR #1973, opened 2020-03-04
[qpid] adding the consumed storage size

waldirio:
Here we are collecting the consumed disk area, very useful when doing Satellite troubleshooting.
```
# du -sh /var/lib/qpidd/.qpidd/*
```
Resolves: #1971


pmoravec:
What is the benefit if we already collect "ls -lanR /var/lib/qpidd" ? So one can get the disk usage via something like:

    grep "^-rw-r" sos_commands/qpid/ls_-lanR_.var.lib.qpidd | awk '{ s+=$5 } END {printf "%d", s}'

TurboTurtle:
Closing this as it was discussed internally that @pmoravec's comment above satisfies the requirements.

PR #1975, opened 2020-03-06
[Plugin|Predicate] Do not collect empty files and add arch sensitivity to Plugin/Predicate

TurboTurtle:
One of our support engineers found that sos was writing empty `lspci` and similar files to the archive when sos was run on ppc/s390 systems.

The command does not exist on those systems, so normally we should have just skipped any output writing, however it was found that if a suggest_filename or root_symlink was specified, we'd write an empty file.

Fixing that led to some architecture fixups in plugins, which in turn then led me to extend `SoSPredicate` to not only allow for architecture checks, but also to include _negative_ predicate checks to allow plugins to specify architectures specific commands can't run on, rather than having to whitelist the ones it can run on.

That then got me thinking of extending the `Plugin` enablement checks to be architecture sensitive as well.

In other words, I went down the rabbit hole on an otherwise straight forward patch and this patchset now:

* Fixes writing empty files to the archive when a command is not found.
* Extends `SoSPredicate` to allow for architecture restraints
* Extends `SoSPredicate` to allow negative checks, I.E. 'collect this only if $thing is **not** present/true'
* Extends `Plugin` to allow an `architectures` tuple to specify what arches to enable on, just like `packages`, `services`, `files`, etc...
* Updates the `powerpc`, `pci`, and `iprconfig` plugins to take advantage of the above.

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -1256,6 +1256,10 @@ class Plugin(object):
                         poller=self.check_timeout
                     )
             self._log_debug("could not run '%s': command not found" % cmd)
+            # Exit here if the command was not found in the chroot check above
+            # as otherwise we will create a blank file in the archive
+            if result['status'] in [126, 127]:
+                return result
 
         self._log_debug("collected output of '%s' in %s (changes=%s)"
                         % (cmd.split()[0], time() - start, changes))
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -125,6 +125,9 @@ class SoSPredicate(object):
     # Command output inclusion pairs {'cmd': 'foo --help', 'output': 'bar'}
     cmd_outputs = []
 
+    #: Allowed architecture(s) of the system
+    arch = []
+
     def __str(self, quote=False, prefix="", suffix=""):
         """Return a string representation of this SoSPredicate with
             optional prefix, suffix and value quoting.
@@ -151,7 +154,11 @@ class SoSPredicate(object):
                                     quotes % cmdoutput['output'])
             for cmdoutput in self.cmd_outputs
         ]
-        pstr += "cmdoutputs=[%s]" % (",".join(cmdoutputs))
+        pstr += "cmdoutputs=[%s], " % (",".join(cmdoutputs))
+
+        arches = self.arch
+        arches = [quotes % a for a in arches] if quote else arches
+        pstr += "arches=[%s]" % (",".join(arches))
 
         return prefix + pstr + suffix
 
@@ -249,6 +256,18 @@ class SoSPredicate(object):
         else:
             return all(_cmds)
 
+    def _eval_arch(self):
+        if not self.arch:
+            return True
+
+        # a test for 'all' against arch does not make sense, so only test to
+        # see if the system's reported architecture is in the last of 'allowed'
+        # arches requested by the predicate
+        if self._owner.policy.get_arch() in self.arch:
+            return True
+        self._failed['architecture'].extend([a for a in self.arch])
+        return False
+
     def report_failed(self):
         """Return a string informing user what caused the predicate to fail
         evaluation
@@ -268,11 +287,12 @@ class SoSPredicate(object):
 
         # Null predicate?
         if not any([self.kmods, self.services, self.packages, self.cmd_outputs,
-                    self.dry_run]):
+                    self.arch, self.dry_run]):
             return True
 
         return ((self._eval_kmods() and self._eval_services() and
-                 self._eval_packages() and self._eval_cmd_outputs())
+                 self._eval_packages() and self._eval_cmd_outputs() and
+                 self._eval_arch())
                 and not self.dry_run)
 
     def __bool__(self):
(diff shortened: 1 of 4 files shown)


TurboTurtle:
Ok, sorry for all the noise. We apparently had a key name mismatch in the `required` and `failed` dicts used in the `SoSPredicate` evals, and I hadn't caught it - but it was throwing off some of the tests.

Interestingly, the tests only failed on Ubuntu and not my local Fedora system though I would have absolutely expected it to fail on Fedora as well.

sos/plugins/__init__.py
```diff
@@ -151,7 +154,11 @@ def __str(self, quote=False, prefix="", suffix=""):
                                     quotes % cmdoutput['output'])
             for cmdoutput in self.cmd_outputs
         ]
-        pstr += "cmdoutputs=[%s]" % (",".join(cmdoutputs))
+        pstr += "cmdoutputs=[%s]," % (",".join(cmdoutputs))
```
pmoravec: Missing space after comma, like

    pstr += "packages=[%s], " % (",".join(pkgs))

has?

sos/plugins/__init__.py
```diff
@@ -1256,6 +1256,10 @@ def _collect_cmd_output(self, cmd, suggest_filename=None,
                         poller=self.check_timeout
                     )
             self._log_debug("could not run '%s': command not found" % cmd)
```
pmoravec: Shouldnt we increase verbosity to `info`, we we (intentionally and correctly) dont create blank files? Like we do when we skip collecting a cmd output due to predicate - default execution of `sosreport` should contain a log about it.

sos/plugins/__init__.py
```diff
@@ -247,14 +262,11 @@ def _eval_cmd_outputs(self):
             res = self._eval_cmd_output(cmd)
             _cmds.append(res)
             if not res:
-                self._failed['cmd_output'].append(
+                self._failed_or_forbidden(
+                    'cmd_outputs',
```
pmoravec: Good unification of `cmd_output` and `cmd_outputs`. +1

sos/plugins/__init__.py
TurboTurtle: I don't think so, personally.

It's not uncommon for plugins to call the same binary multiple times with different options so we could quickly end up spamming users unnecessarily. The predicate warnings have, so far, generally been a low number of reports since a lot of their use is around gating commands that then iterate over some discovered list (like in the docker plugin).

If we changed this to `info`, I think we'd blast to much information at the end user. For the party that ultimately looks at the archive, if there are questions about missing command output they'd have to look at the logs anyways, so I don't think we gain much by increasing the log level.

sos/plugins/__init__.py
TurboTurtle: Yup, will fix.

sos/plugins/pci.py
```diff
@@ -23,6 +24,9 @@ def setup(self):
             "/proc/bus/pci"
         ])
 
+        self.set_cmd_predicate(SoSPredicate(self, arch=['ppc*', 's390*'],
```
bmr-cymru: I'm not certain that this is the right way of handling this - the PPC platform has used PCI as the expansion bus for a long time now, and even Zseries systems can use PCI expansion devices in certain configurations. I believe this is the reason that we build the `pciutils` package for these arches.

Perhaps we should instead gate on the presence of bus directories in `/proc/bus/pci`, or on the output of `lspci` itself? I'm concerned that if we turn all these commands off uniformly for these platforms then we will be losing valuable debugging data for some valid configurations.

sos/plugins/iprconfig.py
```diff
@@ -10,20 +10,18 @@
...
 
 class IprConfig(Plugin, RedHatPlugin, UbuntuPlugin, DebianPlugin):
     """IBM Power RAID storage adapter configuration information
     """
 
     plugin_name = 'iprconfig'
-
-    def check_enabled(self):
-        arch = self.policy.get_arch()
-        return "ppc64" in arch and is_executable("iprconfig")
```
bmr-cymru: I'm not sure we're really doing the right thing here - the restriction to PPC came with the original patch from IBM back in 2014, and the "IPR" branding originally meant "IBM Power RAID". Browsing some hardware articles on IBM's site this still seems to be the case, but I'm confused since it appears all the major distros (us, inc. Fedora/CentOS etc., Debian, SuSE, Ubuntu) all build the package for x86_64 as well.

Interestingly I notice that the packages currently built for Fedora include an `iprsos` command that "is used to generate an sosreport-like report by collecting hardware information from the system, along with other system calls such as lspci, lscsi, and others."

If nothing else I would say that's strong evidence that we need the `lspci` data on Power systems at a minimum...

And sure enough my x86_64 kernel has a module built for "IBM Power RAID" devices...

```$ modinfo ipr
filename:       /lib/modules/5.4.7-100.fc30.x86_64/kernel/drivers/scsi/ipr.ko.xz
version:        2.6.4
license:        GPL
description:    IBM Power RAID SCSI Adapter Driver
author:         Brian King <brking@us.ibm.com>
srcversion:     BAB381AAD15DAC158FA9B8B
...
```
I think ultimately we may need to ask IBM to clarify the support status here, and to help guide us on these decisions. It's really unclear from the publicly available documentation, and it's almost impossible for us to get access to the complete range of possible configurations to find out for ourselves.

TurboTurtle:
I've updated the `lspci` changes to gate on bus directory presence, rather than architecture. I've left the `iprconfig` change as it is because it does not actually change the enablement restrictions we have in place today - it just modifies the checks to use the existing/new mechanisms for it rather than direct overrides. 

We can/should certainly get clarity from IBM as to what the real desire is there, but I don't think that should hold up this patch set when we're really not actually changing the behavior.

bmr-cymru:
Gets my vote.

PR #1977, opened 2020-03-11
[ovirt_engine_backup] Add a new "ovirt_engine_backup" plugin

mmartinv:
---

### DIFF
--- a/sos/plugins/ovirt_engine_backup.py
+++ b/sos/plugins/ovirt_engine_backup.py
@@ -1,0 +1,52 @@
+# Copyright (C) 2019 Red Hat, Inc., Miguel Martin <mmartinv@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+import os
+from sos.plugins import (Plugin, RedHatPlugin)
+from datetime import datetime
+
+
+class oVirtEngineBackup(Plugin, RedHatPlugin):
+    """oVirt Engine database backup"""
+
+    packages = ("ovirt-engine-tools-backup",)
+    plugin_name = "ovirt_engine_backup"
+    option_list = [
+        ("backupdir", "Directory where the backup is generated",
+         "fast", "/var/lib/ovirt-engine-backup"),
+        ("tmpdir", "Directory where the intermediate files are generated",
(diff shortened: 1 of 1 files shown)


sos/plugins/ovirt_engine_backup.py
```diff
@@ -0,0 +1,45 @@
...
+    )
+
+    def setup(self):
+        cmd = "engine-backup --mode=backup --scope=db --file=%s --log=%s" % (
+            self.backup_filename,
+            self.log_filename
+        )
+        os.system(cmd)
+        self.add_copy_spec(self.backup_filename)
+        self.add_copy_spec(self.log_filename)
```
TurboTurtle: It's been a while since I've used `engine-backup`, but is stdout the same as what gets written to the log file?

It'd probably be better to capture that output directly via `Plugin.collect_cmd_output()`. If the `--log` option is absolutely necessary, then we'd still want to use `Plugin.exec_cmd()` here instead of shelling out to `os.system()`.

mmartinv: The stdout is not the same as the log file (which is more detailed) and I would like to capture both of them. The first is to verify the command was executed correctly and the options/flags are correct (if this fails the second file is not even generated). The latter to check what was the exact problem when the backup was taken.

I submitted the changes you asked but I just realized my local repo is too old and I was testing the plugin in a centos 7 server so the changes made are using the old API. I also filed another issue and submitted a patch for the same reason (which I closed/removed). Sorry for the noise.

It would be great to have this plugin for RHEL 7 so I am wondering how should I manage this because if I use Plugin.collect_cmd_output() instead of Plugin.get_cmd_output_now() the plugin won't work for RHEL7. 

Another thing I have noticed: when using Plugin.get_cmd_output() the backup and log files were not collected. That was because it first collected the files and then ran the commands. That's why I used  Plugin.get_cmd_output_now() and I am now wondering if  Plugin.collect_cmd_output() has the same behavior.

TurboTurtle: Yeah, we made some changes to the collection API in the last release of sos. As far as I know, it should be the next sos release in RHEL 7 that contains those changes. @pmoravec can confirm or correct me there.

If you want to save stdout as well, you'd want to use the new `collect_cmd_output()` method. This will save the output, as well as return status, output, and other information. So, what you'd likely want to do is:

```python
res = self.collect_cmd_output(cmd)
if res['status'] == 0:
    self.add_copy_spec([
        backup_filename,
        log_filename
    ])
```

mmartinv: Ok, I have just pulled the changes.

PR #1981, opened 2020-03-12
[console] Add new plugin

k-hagio:
Add a new plugin for collecting information about console and keyboard
including the status of console and Scroll Lock flag.

Signed-off-by: Kazuhito Hagio <k-hagio-ab@nec.com>

---

### DIFF
--- a/sos/plugins/console.py
+++ b/sos/plugins/console.py
@@ -1,0 +1,29 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from glob import glob
+from sos.plugins import Plugin, RedHatPlugin, UbuntuPlugin
+
+
+class Console(Plugin, RedHatPlugin, UbuntuPlugin):
+    """Console and keyboard information
+    """
+
+    plugin_name = 'console'
+    profiles = ('system',)
+    packages = ('kbd',)
+
+    def setup(self):
+        self.add_copy_spec("/proc/consoles")
+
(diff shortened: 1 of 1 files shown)


sos/plugins/console.py
```diff
@@ -0,0 +1,29 @@
...
+    profiles = ('system',)
+    packages = ('kbd',)
+
+    def setup(self):
+        self.add_copy_spec("/proc/consoles")
+
+        self.add_cmd_output("fgconsole")
+
+        for tty in glob("/dev/tty[0-8]"):
+            self.add_cmd_output("kbdinfo -C " + tty + " gkbled")
```
TurboTurtle: We could reduce this to a single `add_copy_spec()` call via:

```python
self.add_copy_spec([
    "kbdinfo -C %s gkbled" % tty for tty in glob.glob("/dev/tty[0-8]")
])

which imo also improves readability.

File: sos/plugins/console.py [hunk 2]
```diff
@@ -0,0 +1,29 @@
...
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from glob import glob
+from sos.plugins import Plugin, RedHatPlugin
+
+
+class Console(Plugin, RedHatPlugin):
```
TurboTurtle: @BryanQuigley @battlemidget any desire to have this in Ubuntu as well?

sos/plugins/console.py
k-hagio: Thanks, will update so.

sos/plugins/console.py
BryanQuigley: Yes, please -- although I'll still +1 if it's not added.

k-hagio: I'll add Ubuntu. Thanks.

PR #2000, opened 2020-04-06
[rabbitmq] Do not run exec for stopped containers

kajinamit:
Currently rabbitmq plugin invokes rabbitmqctl command by docker exec
for all containers whose name start with rabbitmq, but it always tries
to run commands agaist stopped containers, which can never succeed.

This patch makes sure that the plugin only use running containers,
to avoid useless trial with stopped containers.

---

- [ X ] If this commit closes an existing issue, is the line `Closes: #ISSUENUMBER` included in an independent line?
- [ X ] If this commit resolves an existing pull request, is the line `Resolves: #PRNUMBER` included in an independent line?

### DIFF
--- a/sos/plugins/rabbitmq.py
+++ b/sos/plugins/rabbitmq.py
@@ -23,7 +23,7 @@ class RabbitMQ(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
 
     def setup(self):
         container_status = self.exec_cmd(
-            "docker ps -a --format='{{ .Names }}'"
+            "docker ps --format='{{ .Names }}'"
         )
 
         in_container = False


pmoravec:
Good catch. This is already covered in #1993 (see https://github.com/sosreport/sos/pull/1993/files#diff-bd21d686c4c25421b9c266e0e0e7cf42R27 - `get_containers` is called without `get_all=True`).

kajinamit:
@pmoravec 
Thanks for your feedback. Now I confirmed that this topic is covered in the suggested change. 

However I have one concern that the fix seems to be introduced future 4.x, and is not backportable to 3.x. Thought I'm not really familiar with the branch model used in sos, but does it make sense to have this fix somehow in 3.x ?

pmoravec:
Good point, it makes sense to have it in 3.9 branch.

@TurboTurtle , @bmr-cymru : I think we haven't discussed (much) supportability/maintainability of 3.9 additionally to 4.0. As 4.0 might not be applicable for all (at least for some time), does it make sense to maintain 3.9 branch and e.g. release 3.9.[2-3] and even then abandon?

kajinamit:
My point behind the above concern is that the change for 4.x depends on the assumption that python3 is generally used for system python, and I guess it means that 4.x isn't available for el7 but for el8.

However this behaviour also affect OpenStack deployment on el7, it's nice if we can include the fix in 3.x branch, which is still available for el7 .

Please correct me if I misunderstand something.

TurboTurtle:
I've forked a branch for 3.9 (legacy-3.9) for this purpose, to make it easier for downstream to continue to maintain it. I have no problem merging this into the legacy-3.9 branch.

I don't think we'll need to set a definite date on when the stop merging these into legacy-3.9. We can take the stance that only trivial changes, such as this one, will be merged. I.E. no new plugins or features. I anticipate there to only be a trickle of these anyways.

TurboTurtle:
Merged into `legacy-3.9` per e9848da

PR #2006, opened 2020-04-08
[global] Drop use of python six and other 2.X bits

BryanQuigley:
Resolves: #2006

---

### DIFF
--- a/sos/options.py
+++ b/sos/options.py
@@ -69,10 +69,8 @@ class SoSOptions():
             # Convert Python source notation for sequences into plain strings
             vals = [",".join(v) if _is_seq(v) else v for v in vals]
         else:
-            def is_string(val):
-                return isinstance(val, six.string_types)
             # Only quote strings if quote=False
-            vals = ["'%s'" % v if is_string(v) else v for v in vals]
+            vals = ["'%s'" % v if isinstance(val, str) else v for v in vals]
 
         return (args % tuple(vals)).strip(sep) + suffix
 
--- a/sos/utilities.py
+++ b/sos/utilities.py
@@ -18,12 +18,10 @@ import glob
 import tempfile
 import threading
 import time
+import io
 
 from contextlib import closing
 from collections import deque
-
-# PYCOMPAT
-import six
 
 
 def tail(filename, number_of_bytes):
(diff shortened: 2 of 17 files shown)


BryanQuigley:
I'm pretty confident on the main changes, a bit less so on a few of the tests.
I also left six in ovn_central.py because I like to make changes against a test case and didn't have one for that handy.

sos/utilities.py
```diff
@@ -174,7 +169,7 @@ def _child_prep_fn():
             raise e
 
     if p.returncode == 126 or p.returncode == 127:
-        stdout = six.binary_type(b"")
+        stdout = bytes
```
TurboTurtle, line 172: Should this be `stdout = b""`? That way we're still returning a empty string-like rather than the class.

```python
>>> stdout = six.binary_type(b"")
>>> stdout
b''
>>> stdout = bytes
>>> stdout
<class 'bytes'>
>>> stdout = b""
>>> stdout
b''

```

BryanQuigley: Indeed - and that was the test that I wasn't sure about.  Oops!

TurboTurtle:
This LGTM. A quick test shows the `ovn_central` change from `six.iterkeys -> dict.keys` works.

You could either add that to your commit, or I can I add it locally before I merge this - up to you.

sos/options.py
```diff
@@ -70,7 +70,7 @@ def __str(self, quote=False, sep=" ", prefix="", suffix=""):
             vals = [",".join(v) if _is_seq(v) else v for v in vals]
         else:
             def is_string(val):
-                return isinstance(val, six.string_types)
+                return isinstance(val, str)
             # Only quote strings if quote=False
             vals = ["'%s'" % v if is_string(v) else v for v in vals]
```
pmoravec, line 75: Maybe a silly question, but how much sense it has to have `is_string` method at all? Can't be here just:

```
vals = ["'%s'" % v if isinstance(v, str) else v for v in vals]
```

?

BryanQuigley: You are correct, I have no idea why that is like that. Simplified.

BryanQuigley:
Added ovn_central bit and removed unneeded is_string.

Issue #2010, opened 2020-04-09
More python linting...


BryanQuigley:
Was looking at incorporating pylint.. or python coverage.

but there are so many - https://realpython.com/python-code-quality/#linters

Also two meta-tools are either flake8 (shows a bit of unused bits)  or pylama (more output, but not sure how useful).

Anyone have any preferences?

TurboTurtle:
Taking a quick look at flake8, I like it  at first grok- though I haven't dug too deep into it.

pylama seems interesting in that it wraps more projects, but one of the default linters has some rather cryptic errors:

```
sos/report/plugins/__init__.py:885:1: C901 'Plugin._do_copy_path' is too complex (12) [mccabe]
sos/report/plugins/__init__.py:1021:1: C901 'Plugin.add_copy_spec' is too complex (28) [mccabe]
sos/report/plugins/__init__.py:1170:1: C901 'Plugin._add_device_cmd' is too complex (13) [mccabe]
sos/report/plugins/__init__.py:1321:1: C901 'Plugin._collect_cmd_output' is too complex (11) [mccabe]
sos/report/plugins/__init__.py:1596:1: C901 'Plugin.add_journal' is too complex (14) [mccabe]
sos/report/plugins/__init__.py:1738:1: C901 'Plugin.check_enabled' is too complex (11) [mccabe]
```

And looking at [the definition of c901](https://www.flake8rules.com/rules/C901.html), I'm not terribly convinced. Yes, smaller functions == good, but it seems rather arbitrary.

TurboTurtle:
Closing this as we are now using flake8 as part of #2196

PR #2011, opened 2020-04-10
[global] fix some flake8 issues

BryanQuigley:
Mostly unused import or missing variables.
Including 1 mistyped variable by me.

---

### DIFF
--- a/sos/report/plugins/gluster.py
+++ b/sos/report/plugins/gluster.py
@@ -35,9 +35,9 @@ class Gluster(Plugin, RedHatPlugin):
             statedumps_present = statedumps_present+1
             ret = -1
             while ret == -1:
-                last_line = file(
-                    name_dir + '/' + statedump_file, "r").readlines()[-1]
-                ret = string.count(last_line, 'DUMP_END_TIME')
+                with open(name_dir + '/' + statedump_file, 'r') as sfile:
+                    last_line = sfile.readlines()[-1]
+                    ret = string.count(last_line, 'DUMP_END_TIME')
 
     def postproc(self):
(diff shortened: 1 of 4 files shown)


BryanQuigley:
This should be fine to commit (and fixed a mistake I just introduced), but I want to investigate other issues and eventually enable automated testing using this flake command
flake8 sos bin/* --ignore=F841,E121,E123,E126,E226,E24,E704,W503,W504

Remaining issues
sos/options.py:220:49: F821 undefined name 'opts'
sos/policies/__init__.py:917:32: F821 undefined name '_arg_names'

F841 is the one ignore line I effectively added to ignore unused variables for now.

TurboTurtle:
LGTM, will merge shortly.

I am planning on doing a fixup round once I'm done with integrating sos-collector, which I expect will be ready early this week.

Once we get things settled let's talk about the actual linter configuration, and carry on with any discussions surrounding which linter to use (#2010) until then.

sos/report/plugins/gluster.py
```diff
@@ -35,7 +35,7 @@ def wait_for_statedump(self, name_dir):
             statedumps_present = statedumps_present+1
             ret = -1
             while ret == -1:
-                last_line = file(
+                last_line = open(
```
TurboTurtle, line 38: Actually, one small adjustment is needed here before merge.

As written, this leaves us with an out of scope file object that never gets closed. Just need to wrap this in a with clause, e.g.

```python
with open(path, 'r') as sfile:
    last_line = sfile.readlines()[-1]
    ret = string.count(last_line, 'DUMP_END_TIME')
```

PR #2016, opened 2020-04-13
[redis] Collection of redis from scl

jjansky1:
As preparation for Satellite 7 configuration and log
collection adding collection of rh-redis32 from scl.

Also adding fix to __init__.py convert_copyspec_scl
method which prevented of postproc due to
"//" instead of "/".

Resolves: #2016

---

### DIFF
--- a/sos/report/plugins/redis.py
+++ b/sos/report/plugins/redis.py
@@ -9,23 +9,19 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.report.plugins import Plugin, RedHatPlugin
+from sos.report.plugins import Plugin, SCLPlugin
 
 
-class Redis(Plugin, RedHatPlugin):
+class Redis(Plugin, SCLPlugin):
 
     short_desc = 'Redis, in-memory data structure store'
 
     plugin_name = 'redis'
     profiles = ('services',)
 
-    packages = ('redis',)
+    packages = ('redis', 'rh-redis32', 'rh-redis5')
+
     var_puppet_gen = "/var/lib/config-data/puppet-generated/redis"
-    files = (
-        '/etc/redis.conf',
-        '/var/log/redis',
-        var_puppet_gen + '/etc/redis.conf'
-    )
 
     def setup(self):
         self.add_copy_spec([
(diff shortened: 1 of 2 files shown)


sos/report/plugins/redis.py
```diff
@@ -9,23 +9,21 @@
...
-from sos.report.plugins import Plugin, RedHatPlugin
+from sos.report.plugins import Plugin, SCLPlugin
 
 
-class Redis(Plugin, RedHatPlugin):
-
-    short_desc = 'Redis, in-memory data structure store'
+class Redis(Plugin, SCLPlugin):
+    """Redis, in-memory data structure store
+    """
```
TurboTurtle, line 17: Small regression, the `short_desc` attribute is now used instead of docstrings. Let's add that back.

sos/report/plugins/redis.py
```diff
@@ -9,23 +9,21 @@
...
 
     plugin_name = 'redis'
-    profiles = ('services',)
+    profiles = ('services')
```
TurboTurtle, line 20: Needs to be a tuple.

sos/report/plugins/redis.py
```diff
@@ -9,23 +9,21 @@
...
+
+    packages = ('redis',
+                'rh-redis32',
+                'rh-redis5')
```
TurboTurtle, line 24: Can this not all fit on one line?

sos/report/plugins/redis.py
jjansky1: New commit ad3cb2b15dfa6d195c314b6bd44a1825006b9284 proposed, is it ok for you?

sos/report/plugins/redis.py
jjansky1: New commit ad3cb2b15dfa6d195c314b6bd44a1825006b9284 proposed, is it ok for you?

sos/report/plugins/redis.py
jjansky1: New commit ad3cb2b15dfa6d195c314b6bd44a1825006b9284 proposed, is it ok for you?

Issue #1988, opened 2020-03-23
[4.0] Add sos-collector functionality to sos natively


TurboTurtle:
This is to track the addition of `sos-collector` multi-node collection functionality directly into `sos` as a new subcommand for 4.0.

While this should generally be straight forward, there will need to be some consolidation work such as having `collect` leverage `report`'s policy detection and removal of the configuration dict and replacement with a standardized options handling class like what is used for `report`.

This work depends on #1986 being completed.

PR #2017, opened 2020-04-13
[collector] Integrate sos-collector into sos natively

TurboTurtle:
This patchset brings the standalone sos-collector project directly into the sos project, as a new component.

As opposed to the standalone project, the collect component will now make use of `Policy` instead of a separate `SoSHost` definition, meaning any distribution supported by a `Policy` can run sos-collector. Additionally, `PackageManager` has been updated to allow for remote execution in the same manner that `sos collect` executes its own commands, so that we may retrieve package information from the remote node(s).

I have tested this against RHEL and Fedora, on standalone, RHEV, and pacemaker. Additional testing is much appreciated across both distributions and cluster profiles.

Closes: #1988 
---


bin/sos-collector
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/python3
```
pmoravec, line 1: Just a note to the name (how one can comment on a new file _name_? :) ). Doesnt it matter we have `sosreport` without the comma? Should not this be renamed to `soscollector` to have one naming logic?

I am in doubts. Coherent naming is nice, current `sos-collector` is better for eyes.

TurboTurtle:
Latest push makes the following changes:

- Addresses the LGTM double import alert by moving `ClusterOption` into `sos.options`.
- (Re-)adds validation to the cluster option typing
- Updates one of the native runs in Travis to include pexpect. The other run is left without it, reason being the change below
- Importing `sos.collector` is not a hard requirement now, due to the pexpect dependency. If we can't import `sos.collector`, we investigate why and report that back to the user via `--help` and whenever `sos collect` is invoked. In this way, we still allow `sos report` (and other, future components that don't have additional dependencies) to function normally.

BryanQuigley:
Do I need to install this version of sos on all the nodes or can it work if nodes are using an older sosreport version?

What I'm doing:
ssh-add #Needed to make sure ssh key is unlocked?
bin/sos collect -v -b --ssh-user ubuntu --insecure-sudo  --nodes 10.16.169.232,10.16.169.234,10.16.169.73
log: https://pastebin.ubuntu.com/p/5bGwfPFPtk/

Can we rename --insecure-sudo to --nopasswd-sudo?

TurboTurtle:
> Do I need to install this version of sos on all the nodes or can it work if nodes are using an older sosreport version?
> 
> What I'm doing:
> ssh-add #Needed to make sure ssh key is unlocked?
> bin/sos collect -v -b --ssh-user ubuntu --insecure-sudo --nodes 10.16.169.232,10.16.169.234,10.16.169.73
> log: https://pastebin.ubuntu.com/p/5bGwfPFPtk/
> 

You don't need the same version of sos on every node, sos just needs to be installed on the nodes. It should work with any sos version 3.0 or later.

Does 'desktop' have a "normally" installed version of sos on it alongside the git checkout? It's strange to me that the local query failed, but that would explain it if the local installation was removed in favor of the git checkout.

Can you retry with `--cluster-type=none` to avoid the cluster level checks? If that still fails, can you post the full sos.log with `-vvv` enabled?

> Can we rename --insecure-sudo to --nopasswd-sudo?

When adding this functionality to the standalone utility, this was a request from our ProdSec team. Passwordless sudo is _inherently_ insecure, and it should be acknowledged as such.

TurboTurtle:
Wow actually that was, in part, a massive oversight on my end.

Local sudo functionality was just straight broken following the replacement of the former `Configuration` dict. I've pushed a fix commit that should address the ability to run sos reports locally as non-root via sudo (or fail gracefully where appropriate).

Based on some of the cluster checks that were being fouled up by this, I _think_ this should address your errors, but again try using `--cluster-type=none`.

TurboTurtle:
My last comment got a bit messed up, let's try that again.

> Do I need to install this version of sos on all the nodes or can it work if nodes are using an older sosreport version?
> 
> What I'm doing:
> ssh-add #Needed to make sure ssh key is unlocked?
> bin/sos collect -v -b --ssh-user ubuntu --insecure-sudo --nodes 10.16.169.232,10.16.169.234,10.16.169.73
> log: https://pastebin.ubuntu.com/p/5bGwfPFPtk/
> 

No, any sos version 3.0+ should work fine. I believe my latest commit should fix up your issues (see comment above).

If it doesn't, please provide the full sos.log output using `-vvv`.

> Can we rename --insecure-sudo to --nopasswd-sudo?

Our ProdSec team had requested `insecure_sudo` from the standalone tool, as it acknowledges that passwordless sudo is inherently insecure.

I can run it up our side, and I'm sure it would be fine - but I that context is important imo. If you still want to change it, I'm fine with that pending my followup getting an ack.

man/en/sos-collect.1
```diff
@@ -0,0 +1,312 @@
...
+.SH DESCRIPTION
+collect is an sos subcommand to collect sosreports from multiple nodes and package
+them in a single useful tar archive. 
+
+sos collect can be run either on a workstation that has SSH key authentication setup
+for the nodes in a given cluster, or from a "master" node in a cluster that has SSH
+keys configured for the other nodes.
+
+Some sosreport options are supported by sos-collect and are passed directly to 
+the sosreport command run on each node.
```
pmoravec, line 52: Why don't support all `sos report` options? (not necessarily resolved in this PR, for sure)

TurboTurtle: That's a mix of bad wording (should be "applicable sosreport options are supported") and a bit of a hold over from the standalone tool.

At this point, the non-supported options would be `--allow-system-changes`, `--no-env-vars`, `plugin-timeout` (though `-k plugin.timeout` should work), and `--since-`. Easy enough to add those - if you'd prefer I can add them to this patchset.

TurboTurtle:
Fixed a couple edge cases, and added in the missing options @pmoravec highlighted, sos version permitting on a per-node basis.

man/en/sos-collect.1
BryanQuigley: Would it be possible to separate in the help the collector options from the report options?  It would make it a bit cleaner.   *Not a blocker.

BryanQuigley:
It works with updated commit and  --cluster-type=none.  Thanks!

>Our ProdSec team had requested insecure_sudo from the standalone tool, as it acknowledges that passwordless sudo is inherently insecure.
Canonical's security advice is that direct root logins are the bad idea.  Effectively I think what we do is pretty equivalent to root login with keys - our Juju deployments use the known ubuntu user, ssh keys, and nopasswd sudo.

Regardless it's not sos's job to give security advice.

TurboTurtle:
Per the IRC discussion, I've renamed `insecure-sudo` to `nopasswd-sudo` in both the option and the code references. Additionally fixed another edge case that was a perfect storm of non-root, passwords, and sudo usage.

man/en/sos-collect.1
pmoravec: Currently / in this proposal, `sos report` options are maintained at two places: in `man/en/sos-collect.1` and `man/en/sosreport.1`. Isn't it better to defer here the `sos report` options to man pages of `sosreport`? With the notion "this and that option differs/not supported" (like `-t` applies to threads of `sos-collect`)?

pmoravec:
`/var/tmp/sos.*` dirctory is not cleaned up in some scenarios. Running:

    python3 ./bin/sos collect --nodes=$(hostname -f) --batch

left `/var/tmp/sos.85q8ob71` directory. After running multiple `sos collect` commands, I *think* it happens exactly when `Aborting...` logs appear, like:

    2020-04-19 09:38:29,856 ERROR: Node connection for all but localhost failed.
    Aborting...

or in a different use case (on a plain system not knowing any cluster type, git checkout this branch and run `python3 ./bin/sos collect --batch`):

    2020-04-18 16:38:45,564 ERROR: Cluster type could not be determined and no nodes provided
    Aborting...

sos/collector/__init__.py
```diff
@@ -0,0 +1,1091 @@
...
+            self.node_list.remove(self.hostname)
+        for i in self.ip_addrs:
+            if i in self.node_list:
+                self.node_list.remove(i)
+        # remove the master node from the list, since we already have
+        # an open session to it.
+        if self.master:
+            for n in self.node_list:
+                if n == self.master.hostname or n == self.opts.master:
+                    self.node_list.remove(n)
```
pmoravec, line 843: When running:

    python3 ./bin/sos collect --nodes=$(hostname -f),some.existing.system --batch

then this piece of code removes `$(hostname -f)` from the `self.node_list` (confirmed by adding some debugs). Why is it so? From a user experience, it is unexpected.

setup.py
```diff
@@ -70,11 +70,15 @@ def copy_file (self, filename, dirname):
...
+        ('share/man/man1', ['man/en/sosreport.1', 'man/en/sos-report.1',
+                            'man/en/sos.1', 'man/en/sos-collect.1',
+                            'man/en/sos-collector.1']),
         ('share/man/man5', ['man/en/sos.conf.5']),
         ],
-      packages=['sos', 'sos.report.plugins', 'sos.policies', 'sos.report'],
+      packages=['sos', 'sos.policies', 'sos.report', 'sos.report.plugins',
+                'sos.collector', 'sos.collector.clusters'],
       cmdclass={'build': BuildData, 'install_data': InstallData},
+      requires=['pexpect']
```
pmoravec, line 81: This is required just for `sos collect` but not for other subcommand(s). Do we need this requirement for whole `sos`, then?

sos.spec
```diff
@@ -17,6 +17,7 @@ Requires: rpm-python
 Requires: tar
 Requires: bzip2
 Requires: xz
+Requires: python3-pexpect
```
pmoravec, line 20: Ditto above. Maybe I just interpret Requirements wrongly.

On one side, this requirement is not necessary for `sos` in general, "just" for one its feature - so this requirement would be redundant. On the other side, functionality of whole `sos` package will be negatively affected, if this requirement is missing => it should be here.

(on the yet another side, `sos report` does not really require `xz` either way, but we have it among requirements as well - again, to ensure some particular functionality)

So if nobody else sees this as an issue, let ignore both comments.

sos/collector/__init__.py
pmoravec: OK, the reason why `sos report` was skipped from the local host was, that `~/.ssh/known_hosts` contained a valid record for `$(hostname)` but not `localhost`.

Tricky small issue, not blocker for the merge but worth fixing later on, imho.

sos/collector/clusters/satellite.py
```diff
@@ -0,0 +1,39 @@
...
+    cluster_name = 'Red Hat Satellite 6'
+    packages = ('satellite', 'satellite-installer')
+
+    def _psql_cmd(self, query):
+        _cmd = "su postgres -c %s"
+        _dbcmd = "psql foreman -c %s"
+        return _cmd % quote(_dbcmd % quote(query))
+
+    def get_nodes(self):
+        cmd = self._psql_cmd('select name from smart_proxies')
```
pmoravec, line 27: If you have here `copy (select name from smart_proxies) to stdout`, no need to bother by header and footer in the response (one less `idx` shuffling in beow code):

    # su - postgres -c "psql foreman -c 'select name from smart_proxies'"
                         name                      
    -----------------------------------------------
     pmoravec-sat66-on-rhev.gsslab.brq2.redhat.com
     pmoravec-caps66-rhev.gsslab.brq2.redhat.com
    (2 rows)
    
    # su - postgres -c "psql foreman -c 'copy (select name from smart_proxies) to stdout'"
    pmoravec-sat66-on-rhev.gsslab.brq2.redhat.com
    pmoravec-caps66-rhev.gsslab.brq2.redhat.com
    #

But as the current code works well..

sos/collector/clusters/satellite.py
```diff
@@ -0,0 +1,39 @@
...
+class satellite(Cluster):
+    """Red Hat Satellite 6"""
+
+    cluster_name = 'Red Hat Satellite 6'
+    packages = ('satellite', 'satellite-installer')
+
+    def _psql_cmd(self, query):
+        _cmd = "su postgres -c %s"
+        _dbcmd = "psql foreman -c %s"
+        return _cmd % quote(_dbcmd % quote(query))
```
pmoravec, line 24: This does not handle deployments with external `postgres` DB. Cf. https://github.com/sosreport/sos/blob/master/sos/report/plugins/foreman.py#L31 and below.

man/en/sos-collect.1
```diff
@@ -0,0 +1,312 @@
...
+will start collecting from the first X number of nodes and then continue to iterate
+through the remaining nodes as sosreport collection finishes.
+
+Defaults to 4.
+.TP
+\fB\-\-timeout\fR TIMEOUT
+Timeout for sosreport generation on each node, in seconds.
+
+Note that sosreports are collected in parallel, so this can also be considered to be
+approximately the same as a timeout for the entire collection process.
```
pmoravec, line 284: This statement ignores the impact of `--threads`. Defaulting to 4, when I collect `sosreport` from e.g. 8 systems, I can need to wait up to `2*timeout` seconds.

sos/collector/sosnode.py
```diff
@@ -0,0 +1,882 @@
...
+                if ret:
+                    self.file_list.append(filename.split('/')[-1])
+                    self.remove_file(filename)
+                else:
+                    self.log_error("Unable to retrieve file %s" % filename)
+            except Exception as e:
+                msg = 'Error collecting additional data from master: %s' % e
+                self.log_error(msg)
+
+    def make_archive_readable(self, filepath):
```
pmoravec, line 869: `os.chmod(filepath, stat.S_IROTH)` doesnt work?

sos/policies/__init__.py
```diff
@@ -1353,4 +1383,53 @@ def upload_ftp(self, url=None, directory=None, user=None, password=None):
         except IOError:
             raise Exception("could not open archive file")
 
+    def set_sos_prefix(self):
```
pmoravec, line 1386: Is this method used anywhere? I can't find it..

pmoravec:
Log `The following is a list of nodes to collect from:` does not respect `--no-local` option: when running:

    sos collect --nodes=pmoravec-sat66-on-rhev.gsslab.brq2.redhat.com,pmoravec-caps66-rhev.gsslab.brq2.redhat.com -o pulp,qpid,apache,foreman --batch --no-local

it logs:

        pmoravec-rhel8.gsslab.brq2.redhat.com        
        pmoravec-caps66-rhev.gsslab.brq2.redhat.com  
        pmoravec-sat66-on-rhev.gsslab.brq2.redhat.com

and collects reports just from the 2 expected nodes.

pmoravec:
`-o` option is wrongly re-calculated; Using the same command from https://github.com/sosreport/sos/pull/2017#issuecomment-616124402, I see:

    [pmoravec-caps66-rhev.gsslab.brq2.redhat.com:finalize_sos_cmd] Requested plugins ['pulp,qpid,apache,foreman'] were requested to be enabled but do not exist
    [pmoravec-caps66-rhev.gsslab.brq2.redhat.com:sosreport] Final sos command set to sosreport --batch  
    pmoravec-caps66-rhev.gsslab.brq2.redhat.com    : Generating sosreport...
    [pmoravec-caps66-rhev.gsslab.brq2.redhat.com:run_command] Running command /usr/sbin/sosreport --batch  
    [pmoravec-sat66-on-rhev.gsslab.brq2.redhat.com:finalize_sos_cmd] Requested plugins ['pulp,qpid,apache,foreman'] were requested to be enabled but do not exist
    [pmoravec-sat66-on-rhev.gsslab.brq2.redhat.com:sosreport] Final sos command set to sosreport --batch  

while both systems have all four plugins enabled by default.

pmoravec:
In general, ack. I havent checked whole code changes in detail. In particular bits copy&pasted from `sos-collector` were rather quickly scanned and main concern was to the integration with and changes in `sos` directly.

Apart the few comments in code:

- can't be `sos report` options de-duplicated from `sos-collector`? As adding a `report` option needs to add it to `collect` (and its manpages) as well. But I dont see a simple way of doing so, esp. if we apply some logic on the options (i.e. `-o/-e/-n` )
- does it make sense to:
  - support `--threads` option of `report` inside `collector`, and
  - have independent option e.g. `--threads-collect` in `collect`? (here it can have imho higher default value, as each thread performs less intensive work than for a sos plugin)
- does it make sense to add a checksum of the final tarball? (not necessarily now)

setup.py
TurboTurtle: My perspective here is that splitting the components out, or disabling them via not setting certain dependencies, is up to the downstream packagers. For us, I think it makes sense to have it default to required so that the entire functionality of the tool is available when you're running off of the current master.

sos.spec
TurboTurtle: See above, same deal here - I think this should be up to the individual downstream packagers.

sos/collector/clusters/satellite.py
TurboTurtle: Ack, will have to look into that. Do you consider this a merge blocker?

man/en/sos-collect.1
TurboTurtle: Yup, need to reword that.

sos/collector/sosnode.py
TurboTurtle: Unfortunately it does not on remote nodes.

sos/policies/__init__.py
TurboTurtle: I _think_ that's a leftover from the previous design for executing within a container. In the current design, that should be handled by `Policy.format_container_command()`.

I'll double check on some containerized hosts this week.

TurboTurtle:
> Log `The following is a list of nodes to collect from:` does not respect `--no-local` option: when running:
> 

Ack, will fix that up. 
> 
> Moreover, in this use case, the collector redundantly calls `/usr/sbin/sosreport -l` and `/usr/sbin/sosreport --list-presets` on the localhost. That takes some redundant time..

Presets aren't included in `-l` output.

> EDIT: the cause is b'cos `self.opts.only_plugins` is a _string_ with value `'pulp,qpid,apache,foreman'`, not a list of strings.

Oof, that's silly. Easy enough to fix though.


> * can't be `sos report` options de-duplicated from `sos-collector`? As adding a `report` option needs to add it to `collect` (and its manpages) as well. But I dont see a simple way of doing so, esp. if we apply some logic on the options (i.e. `-o/-e/-n` )

Yeah, that's the rub. I don't see an easy way to load the report options without then having to go back through and remove the ones we don't need. So, kind of the reverse operation of what we're doing now.

> * does it make sense to:
>   
>   * support `--threads` option of `report` inside `collector`, and
>   * have independent option e.g. `--threads-collect` in `collect`? (here it can have imho higher default value, as each thread performs less intensive work than for a sos plugin)

That makes a lot of sense, and it shouldn't be too difficult. Let's call it `--jobs` for number of collector threads, and `--threads` can just be passed through to the sos command.

> * does it make sense to add a checksum of the final tarball? (not necessarily now)

Since we gather the checksums for the individual sosreports into the archive, I didn't originally think it was necessary. If it's agreed to be beneficial on it's own then there shouldn't be an issue in adding one.

TurboTurtle:
> `/var/tmp/sos.*` dirctory is not cleaned up in some scenarios.
> 

In those circumstances, my original line of thinking was that we don't want to purge the logs from a failed run if we (or an end user) are trying to dig into that failure. I'm not so sure about that now, especially when one can just crank up the verbosity to spit it all out to console.

TurboTurtle:
Ok, this last PR update should address most of the points @pmoravec brought up, plus an oversight in the use of containerized hosts once we adopted the use a `Policy`.

The two outstanding points from above are:

1. Packaging files setting pexpect as a dependency. To reiterate, I think the splitting out of components by dependencies or sub-packaging is up to the individual downstreams. When running from the source master branch, I believe this should be a default dependency.

2. Satellite support for external databases. I would need some time to get that squared away, but I don't think that is a blocker on this integration PR. @pmoravec please let me know if you feel that this is actually a blocker.

sos/collector/clusters/satellite.py
pmoravec: imho not a blocker. let raise post-merge issue and I can PR it.

pmoravec:
> > `/var/tmp/sos.*` dirctory is not cleaned up in some scenarios.
> 
> In those circumstances, my original line of thinking was that we don't want to purge the logs from a failed run if we (or an end user) are trying to dig into that failure. I'm not so sure about that now, especially when one can just crank up the verbosity to spit it all out to console.

OK, then end user shall get the info the dir persists. The dir is printed at the beginning, at least. But if we can enhance the error handling at the end by printing some info "the outcome of this run is in dir ..", I would see it as beneficial. I.e. whenever `sos` generates some data in `/var/tmp`, it says it to the user at the end.

pmoravec:
One unexpected behaviour (that I understand from devel perspective but it surprised me and also @jjansky1 as users):

Why collecting sos report from localhost requires ssh to it? I.e. either ssh keys to localhost (usually this does not have much sense to have), or provide root password (i.e. user intervention). Could sos report from the localhost be collected directly?

jjansky1:
Maybe jumphost option will be smart move. For example Satellite infrastructure is in some cases as below.

Satellite - network x (accessible by all)
              - network y (accessible by Capsule1 and its clients)
              - network z (accessible by Capsule2 and its clients)
Capsule1(node) - network y
Capsule2(node) - network z

In this scenario which lot of users using, when i run sos-collector --master Satellite from my PC it will correctly recognize master and nodes, but it will collect sosreport only from Satellite as it is trying to connect to Capsules(nodes) from my PC instead of using Satellite(master) as jumphost. As those networks are isolated and accessible only from Satellite as jumphost.

Is possible to create something like option --master-jump-host --nodes-jump-host ?

pmoravec:
(investigation pending on me):

sos collect does not work on Satellite:

    # python3 bin/sos collect --batch
    ..
   	    jjansat66.usersys.redhat.com             
	    could not change directory to "/root/sos"
	    jjansky-capsule.sat66.jjansky.lan        

but no report is collected. That's one problem, 2nd is the `could not change directory to "/root/sos"` node detected. (yes, `$pwd` is that dir).

/me checking both issues now

TurboTurtle:
> OK, then end user shall get the info the dir persists. The dir is printed at the beginning, at least. But if we can enhance the error handling at the end by printing some info "the outcome of this run is in dir ..", I would see it as beneficial. I.e. whenever `sos` generates some data in `/var/tmp`, it says it to the user at the end.

I've updated it to always cleanup, pending below

> if I press Ctrl+C, `/var/tmp/sos.2nm96f6v/` directory is not removed. This works well for `sos report`, reproducible just for `sos collect`.

I just need to hook that up to the proper `exit()` call. Fixup incoming.


> Why collecting sos report from localhost requires ssh to it? I.e. either ssh keys to localhost (usually this does not have much sense to have), or provide root password (i.e. user intervention). Could sos report from the localhost be collected directly?

It doesn't use ssh locally. The 'connection' to localhost is mocked in `SoSNode` initialization, so that we can still run through the same checks and command validations - but no actual ssh session is opened locally.


> Maybe jumphost option will be smart move. For example Satellite infrastructure is in some cases as below.
> Is possible to create something like option --master-jump-host --nodes-jump-host ?


That would be a fairly significant change to the current design. As per the documentation, `collect` (and standalone sos-collector) need network access to the nodes. We can investigate adding this kind of functionality, but I don't think this is in scope for the initial integration.

sos/collector/clusters/satellite.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def get_nodes(self):
+        cmd = self._psql_cmd('copy (select name from smart_proxies) to stdout')
+        res = self.exec_master_cmd(cmd, need_root=True)
```
pmoravec, line 28: Dirty bug: when `root` is in a directory where `postgres` user does not have priviledges to enter, the `su postgres ..` command prints to `stdout` also line `'could not change directory to "/root/sos"'` - and this is taken as a node name, further :-/

sos/collector/sosnode.py
```diff
@@ -0,0 +1,914 @@
...
+            # use the containerized policy's command
+            pkgs = self.run_command(self.host.container_version_command,
+                                    use_container=True, need_root=True)
+            ver = pkgs['stdout'].strip().split('-')[1]
+            if ver:
+                self.sos_info['version'] = ver
+        if 'version' in self.sos_info:
+            self.log_info('sos version is %s' % self.sos_info['version'])
+        else:
+            self.log_error('sos is not installed on this node')
```
pmoravec, line 252: I dont know why, but if a remote system does not have `sos` package installed, then `self.connected = None` and report is not collected, BUT this error is _not_ printed - that is very confusing (and I dont get why)..

sos/collector/clusters/satellite.py
pmoravec: Possible fix:

    def get_nodes(self):
        cmd = self._psql_cmd('copy (select name from smart_proxies) to stdout')
        res = self.exec_master_cmd(cmd, need_root=True)
        nodes = []
        if res['status'] == 0:
            for node in res['stdout'].splitlines():
                if 'could not change directory' not in node:
                    nodes.append(node.strip())
        return nodes

TurboTurtle:
- Added the catch for keyboard interrupts during disclaimer prompt.

- Added a fix to filter out the "could not change directory" error printed from the psql query when the postgres user could not enter the CWD. Thanks @pmoravec
 
- After IRC discussion with Pavel and Jan, clarified the abort message when we only have a single collection possible from localhost. This message now makes it abundantly clear that node enumeration and/or connection failed leaving us with only possibly collecting localhost, which is assumed to not be desired when running `collect`. Note that if we only have one node and it is NOT localhost, we will still collect that node.

- Rebased on current master

I believe that takes care of all raised concerns that have not been pushed to later development (jump host and external db support for satellite).

Calling for final round of reviews/acks for this.

BryanQuigley:
My +1 stands - did another round of testing.  Nice work!

pmoravec:
One (probably the latest) comment:

when `python3-pexpect` is not installed, `sos collect` run with any argument is confusing - it behaves like missing much of functionality:

    # python3 ./bin/sos collect --batch
    usage: sos <component> [options]
    
    Available components:
    	report			Collect files and command output in an archive
    	collect			(unavailable) Collect an sos report from multiple nodes
    sos: error: unrecognized arguments: --batch
    # python3 ./bin/sos collect --help
    usage: sos collect [options]
    
    optional arguments:
      -h, --help            show this help message and exit
    
    Global Options:
      --config-file CONFIG_FILE
                            specify alternate configuration file
      --debug               enable interactive debugging using the python debugger
      -q, --quiet           only print fatal errors
      -s SYSROOT, --sysroot SYSROOT
                            system rootdir path (default='/')
      --tmp-dir TMP_DIR     specify alternate temporary directory
      -t THREADS, --threads THREADS
                            Number of threads to use
      -v, --verbose         increase verbosity
    # 

We should print the `The collect command is unavailable due to a missing dependency on python3-pexpect.` error instead.

TurboTurtle:
Updated locally and verified with @pmoravec via IRC to address this last concern.

Merging this shortly...

Issue #1959, opened 2020-02-18
[RFC] sos-4.0 Planning and Discussion


TurboTurtle:
The last major release (sos-3.0) was made just over six years ago in 2013. This release added the general policy framework allowing sos to be ported to multiple distributions, improved report archive handling and numerous improvements to the plugin and reporting frameworks. In the time since 3.0 was released (when sosreport was available on Fedora and Red Hat Enterprise Linux only), sos has been adopted by numerous additional distributions, products and users.
 
In planning a new major release the team hopes to address several long-standing shortcomings and to deliver a more integrated and streamlined set of tools for gathering, exchanging and reporting crucial support information for today's Linux based systems.
 
Our current goals for sos-4.0 address several long standing limitations in the existing tools, and integrate additional functionality that has historically been covered by third party projects and ad-hoc solutions in specific projects and products. These goals include changes to the core sos reporting tools as used today (for e.g. to improve IO efficiency and memory usage), as well as integrating functionality for collecting sos data from multple host systems and the automated removal of a greater range of potentially sensitive or privacy relevant data included in reports.

Below is the current focus and current goals for a sos-4.0 release.

### Python3 only

For some time now, upstream sos has supported both python 2.7 and python 3. Going forward we will be dropping python 2.7 support. This allows the development team to focus on new features and improvements, rather than backwards compatibility. Further, the beginning of this year marked the end of upstream support for Python 2.7 from the Python Project.

When 4.0 changes start hitting the master branch, the latest 3.9 work will be saved to an alternate branch for ease of maintenance for downstreams.

### Major binary redesign

sosreport has been a staple of diagnostic data collection for years now, and we want to expand on that success. With sos-4.0, we will be introducing a new `sos` command to replace the well-known `sosreport` command as the entry point for the sos project.

What this will allow us to do is add new features from related projects (or those frequently asked for over the years) natively into the sos project. For example, sos-collector will be brought into the project as the `collect` (or similar) sub command, so that users would now run `sos collect [...]` to collect sosreports from multiple hosts.

Similarly, `sosreport` will become `sos report`. We will of course, for transition purposes, continue to carry an `sosreport` which will be marked as deprecated, and point the user to the new `sos report` style. The deprecated `sosreport` will be limited in regards to leveraging the new functionality available via the new `sos` binary.

An overview of some of the subcommands the project could add are as follows - though these are not necessarily slated for 4.0, as the major focus will be focusing on laying the groundwork to easily integrate these in future releases:

[...]

PR #2018, opened 2020-04-14
[policies] Proof of Concept for plugin for any OS

BryanQuigley:
This is an alternative model to allow us to stop specifying what
OSes this should run on in (but we still can choose too).  Not for committing, for discussion.

Relatedly, It may be good to discuss and be explicit about what OSes are supported, and what their support level is.  For instance, do we care about OSX?

---

### DIFF
--- a/sos/policies/redhat.py
+++ b/sos/policies/redhat.py
@@ -12,7 +12,7 @@ import os
 import sys
 import re
 
-from sos.report.plugins import RedHatPlugin
+from sos.report.plugins import RedHatPlugin, LinuxPlugin
 from sos.policies import LinuxPolicy, PackageManager, PresetDefaults
 from sos import _sos as _
 from sos.options import SoSOptions
@@ -56,7 +56,7 @@ class RedHatPolicy(LinuxPolicy):
                                               files_command=self._rpmql_cmd,
                                               chroot=sysroot)
 
-        self.valid_subclasses = [RedHatPlugin]
+        self.valid_subclasses = [RedHatPlugin, LinuxPlugin]
 
         self.pkgs = self.package_manager.all_pkgs()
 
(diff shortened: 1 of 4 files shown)


TurboTurtle:
Would this not prevent us from having the following?

```python

class MyPlugin(Plugin, LinuxPlugin)
    [....]


class DebianMyPlugin(MyPlugin, DebianPlugin)
```

Or is it just acknowledged that as soon as any supported distro needs to do "something" different in a plugin, we can no longer use the `LinuxPlugin` tagging class anywhere in the plugin?

+1 to dropping the OSX policy though. I don't think we've ever had a PR, bug report, or any other kind of GH Issue for it.

bmr-cymru:
If we drop all non-Linux policies then effectively `Plugin == LinuxPlugin`; we might as well just use that and not introduce another layer of complexity if we wanted to allow these kinds of generic plugins.

BryanQuigley:
>Or is it just acknowledged that as soon as any supported distro needs to do "something" different in a plugin, we can no longer use the LinuxPlugin tagging class anywhere in the plugin?

Yes, it doesn't need to be called LinuxPlugin - it could be AlwaysPlugin or AnyDistroPlugin - but it would just mean this should run on any distro.

>If we drop all non-Linux policies then effectively Plugin == LinuxPlugin; we might as well just use that and not introduce another layer of complexity if we wanted to allow these kinds of generic plugins.

I'd be for that, but how would we specify we just want it to run on certain distros? I previously suggested some options https://github.com/sosreport/sos/issues/1959#issuecomment-601760479 but may have missed some.

BryanQuigley:
I had not noticed it before - and it does not appear to work in my quick testing -  IndependentPlugin is basically what I proposed.

BryanQuigley:
I swear I had this working, but now can't get it to do so again :/.  Assuming I can get it to work, would we want to start using IndependentPlugin for any plugins that don't need to depend on a specific OS?

TurboTurtle:
Off the top of my head, I believe we already append `IndependentPlugin` to every policy, so this would just be a matter of cleaning up the actual plugin subclassing rather than any policy-level or plugin enablement changes.

I'm onboard with changing over to using that where possible, as it'll also make adding new distro support an easier process.

PR #2020, opened 2020-04-15
[navicli] replace interactive prompt by plugin option

pmoravec:
List of ClariiOn IP addresses should be provided via
-k navicli.ipaddrs plugopt, as a space separated list.

Resolves: #2020

---

### DIFF
--- a/sos/report/plugins/navicli.py
+++ b/sos/report/plugins/navicli.py
@@ -19,6 +19,8 @@ class Navicli(Plugin, RedHatPlugin):
 
     plugin_name = 'navicli'
     profiles = ('storage', 'hardware')
+    option_list = [("ipaddrs", "list of space separated CLARiiON IP addresses",
+                    '', "")]
 
     def check_enabled(self):
         return is_executable("navicli")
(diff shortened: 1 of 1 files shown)


sos/report/plugins/navicli.py
```diff
@@ -58,29 +60,10 @@ def get_navicli_SP_info(self, SP_address):
...
-            else:
-                if ans != "":
-                    print("The IP address you entered, %s, is not to an "
-                          "active CLARiiON SP." % ans)
-                if ans == "":
-                    CLARiiON_IP_loop = "get_out"
-        # Sort and dedup the list of CLARiiON IP Addresses
-        CLARiiON_IP_address_list.sort()
+        for addr in set(self.get_option("ipaddrs").split()):
+            if self.exec_cmd("navicli -h %s getsptime" % (addr))['status']:
```
TurboTurtle, line 64: Should this be `['status'] == 0`?

pmoravec: Ah sure. Blind copying previous test (that didnt work for years..). Thx for spotting it.

BryanQuigley:
Added closes: #405

PR #2028, opened 2020-04-17
[redhat] fix RH containers without sysroot

robbmanes:
Attempting to run sosreport in a container currently will always
traceback unless ENV_HOST_SYSROOT is set to '/'.

Allow default NoneType sysroot to function as well.

Signed off by: Robb Manes <rmanes@redhat.com>

---

- [x ] If this commit resolves an existing pull request, is the line `Resolves: #PRNUMBER` included in an independent line?

Attempting to run sosreport in a RHEL container will fail unless `HOST_ENV_SYSROOT` is explicitly set to `/`; if one desires to run an sosreport inside of a container to get container-specific values, then it will traceback:
```
$ podman run -it --rm --name sosreport registry.redhat.io/rhel8/support-tools bash
[root@9e2c299cf1c4 /]# sosreport 
  File "/usr/sbin/sosreport", line 19, in <module>
    main(sys.argv[1:])
  File "/usr/lib/python3.6/site-packages/sos/sosreport.py", line 1383, in main
    sos = SoSReport(args)
  File "/usr/lib/python3.6/site-packages/sos/sosreport.py", line 287, in __init__
    self.policy = sos.policies.load(sysroot=self.opts.sysroot)
  File "/usr/lib/python3.6/site-packages/sos/policies/__init__.py", line 44, in load
    cache['policy'] = policy(sysroot=sysroot)
  File "/usr/lib/python3.6/site-packages/sos/policies/redhat.py", line 259, in __init__
    super(RHELPolicy, self).__init__(sysroot=sysroot)
  File "/usr/lib/python3.6/site-packages/sos/policies/redhat.py", line 50, in __init__
    sysroot = self._container_init()
  File "/usr/lib/python3.6/site-packages/sos/policies/redhat.py", line 145, in _container_init
    host_tmp_dir = os.path.abspath(self._host_sysroot + self._tmp_dir)
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```
This is fixed by not just checking for `/` but just allowing `NoneType` default sysroot to function as well.  Let me know if adjustments should be made to this for a scenario I didn't think of, but I can't see why we'd only allow either no sysroot or `/` explicitly, and if we do, we probably shouldn't traceback but fail gracefully.
```
# sosreport 
sosreport (version 3.7)
- - - - 8< - - - -
 Setting up archive ...
 Setting up plugins ...
- - - - 8< - - - -
writing traceback to sos_logs/usb-plugin-errors.txt
caught exception in plugin method "x11.setup()"
writing traceback to sos_logs/x11-plugin-errors.txt
caught exception in plugin method "xfs.setup()"
writing traceback to sos_logs/xfs-plugin-errors.txt
caught exception in plugin method "yum.setup()"
writing traceback to sos_logs/yum-plugin-errors.txt

  Finishing plugins              [Running: systemd]                                       
  Finished running plugins                                                               
caught exception in plugin method "filesys.postproc()"
writing traceback to sos_logs/filesys-plugin-errors.txt
caught exception in plugin method "ldap.postproc()"
writing traceback to sos_logs/ldap-plugin-errors.txt
caught exception in plugin method "openhpi.postproc()"
writing traceback to sos_logs/openhpi-plugin-errors.txt
caught exception in plugin method "openshift.postproc()"
writing traceback to sos_logs/openshift-plugin-errors.txt
caught exception in plugin method "openssl.postproc()"
writing traceback to sos_logs/openssl-plugin-errors.txt
```

### DIFF
--- a/sos/policies/redhat.py
+++ b/sos/policies/redhat.py
@@ -134,7 +134,7 @@ class RedHatPolicy(LinuxPolicy):
                 self._in_container = True
         if ENV_HOST_SYSROOT in os.environ:
             self._host_sysroot = os.environ[ENV_HOST_SYSROOT]
-        use_sysroot = self._in_container and self._host_sysroot != '/'
+        use_sysroot = self._in_container and self._host_sysroot is not None
         if use_sysroot:
             host_tmp_dir = os.path.abspath(self._host_sysroot + self._tmp_dir)
             self._tmp_dir = host_tmp_dir


TurboTurtle:
Generally this is because the thought is "who wants an sosreport of a single container?" and the design is based around the RH support-tools image that is used to run sosreport in a container on Atomic Host/RHCOS but still get information from the host _instead_ of the container.

This happens by reading the `ENV_HOST_SYSROOT` defined environment var (`host` for RH distros) set in the container, which for support-tools is set to `/host`, as the host's filesystem is mounted there. `self._host_sysroot` will be set to the policy defined `sysroot` (obviously defaulting to `/`) unless we detected we're in a container, by reading the environment var specified by `ENV_CONTAINER`. In the case of Red Hat images this is the `container` env var, usually set to `oci`. So, this check is explicitly "we know we're in a container, and we know it's configured to let us capture from the host".

So, first we'd need to validate that this change doesn't break collecting from the host on RHCOS. 

Second, we'd need to be sure that collecting an sosreport of just the container we're executing in is actually useful. When all this work originally went in, it was determined it wasn't useful. If that's changed then we can adjust, or as you say simply not spit out an exception and fail more gracefully.

robbmanes:
> Generally this is because the thought is "who wants an sosreport of a single container?" and the design is based around the RH support-tools image that is used to run sosreport in a container on Atomic Host/RHCOS but still get information from the host instead of the container.

Understood, and this is definitely a conversation probably beyond the scope of just my fix; there are a few scenarios where an sosreport run inside of a containerized environment is useful for things like application configurations, environment settings, etc.  That's what caused me to notice this actually, we might want say `mysql` or `kubernetes` plugins to run in a container to get container-specific information you might not always get from the host (although, arguably, if you capture the `docker` or `podman` plugins, you can infer a lot if not everything).

So that might be a larger discussion beyond the scope of my familiarity of this project and I defer to you.

> So, first we'd need to validate that this change doesn't break collecting from the host on RHCOS.

Agreed, I'll do some additional testing on my own then.

> Second, we'd need to be sure that collecting an sosreport of just the container we're executing in is actually useful.

Wouldn't that be more on the responsibility of the person requesting the sosreport than the tool itself?  What can't be gathered tracebacks and throws an error in `sos_log`; the only way I can think of to make that happen effectively would be to have every plugin have a required list of binaries or something the `sos` tool would have to check for, and that's a pretty hefty architectural change.  I'm of the opinion that it's sane to ask for the sosreport with the understanding that you probably can't get much, but perhaps it's enough or what the individual is looking for; otherwise, we could put in a pretty warning about "sosreport is running within a container: errors are expected as some full installation binaries may not be expected to function normally" or something.

Nonetheless, good points.  I'll do some more testing in-depth, and if you all decide it's best to throw an error rather than allow it to run, I'll defer.

robbmanes:
Just commenting this is related to [Red Hat Bugzilla 1825283](https://bugzilla.redhat.com/show_bug.cgi?id=1825283) for anyone else who wanders across this.

TurboTurtle:
> Understood, and this is definitely a conversation probably beyond the scope of just my fix; there are a few scenarios where an sosreport run inside of a containerized environment is useful for things like application configurations, environment settings, etc. That's what caused me to notice this actually, we might want say `mysql` or `kubernetes` plugins to run in a container to get container-specific information you might not always get from the host (although, arguably, if you capture the `docker` or `podman` plugins, you can infer a lot if not everything).
> 
> So that might be a larger discussion beyond the scope of my familiarity of this project and I defer to you.
> 

Right, there's a lot of point-counterpoint that could revolve around this type of collection. I'm not really all that against it, just weighing the practicality of it. If there's any noticeable amount of real scenarios that would benefit from this then I'm all for removing the impediments.

> > So, first we'd need to validate that this change doesn't break collecting from the host on RHCOS.
> 
> Agreed, I'll do some additional testing on my own then.
> 

Awesome.

> I'm of the opinion that it's sane to ask for the sosreport with the understanding that you probably can't get much, but perhaps it's enough or what the individual is looking for; otherwise, we could put in a pretty warning about "sosreport is running within a container: errors are expected as some full installation binaries may not be expected to function normally" or something.
> 

Definitely a valid point. A notification should at least mitigate some of the end user pains that might crop up.


Like I said, if it's a real scenario where single-container sosreports are beneficial and it doesn't affect our ability to collect from the container's host system, then I have no objections.

robbmanes:
Terribly sorry for the delay, I was very sick for some time.

Reporting that this commit doesn't seem to break `toolbox` on CoreOS in an OpenShift 4.X cluster:

```
[shift-user@coreos-worker-0 ~]$ cat /etc/redhat-release
Red Hat Enterprise Linux CoreOS release 4.3

[shift-user@coreos-worker-0 ~]$ toolbox
Spawning a container 'toolbox-shiftuser' with image 'registry.redhat.io/rhel8/support-tools'
Detected RUN label in the container image. Using that as the default...
command: podman run -it --name toolbox-shift-user --privileged --ipc=host --net=host --pid=host -e HOST=/host -e NAME=toolbox-shift-user -e IMAGE=registry.redhat.io/rhel8/support-tools:latest -v /run:/run -v /var/log:/var/log -v /etc/machine-id:/etc/machine-id -v /etc/localtime:/etc/localtime -v /:/host registry.redhat.io/rhel8/support-tools:latest

[root@coreos-worker-0]# sosreport
...
  /host/var/tmp/sosreport-worker-0-test-2020-04-26-ovzxkgi.tar.xz
```

Only just now was well enough to set up the environment to test.

Anything else I am missing, or anything else should be considered?

pmoravec:
This should go to `legacy-3.9` as well.

TurboTurtle:
> This should go to `legacy-3.9` as well.

Should it? This would qualify to me a 'new functionality' since it's enabling sos to capture inside containers, rather than forcing it to capture from the host. I'm not particularly keen on extending in-container support to the legacy branch, personally.

pmoravec:
Ah right, dropping the request to backport to 3.9

robbmanes:
Did this mean to be closed without being merged; was this denied for a different solution or moved elsewhere?  Thanks.

TurboTurtle:
No, it was merged into master. We don't do merge commits in sos, the the reason GH says closed instead of merged is because the commit IDs change when we rebase locally before merging into master.

robbmanes:
Ah, understood, thanks!

PR #2031, opened 2020-04-24
[openstack_instack] Add new config-download location

rabi:
With the removal of mistral from undercloud,  we would be using $HOME/config-download for
the config downloaded playbooks. Add it to the collected locations on undercloud.


TurboTurtle:
Was going to merge this in this round, however I just noticed the commit is missing a DCO line.

@rabi please update the commit message with a DCO line (`git commit -s --amend`) and force-push the branch to update it.

Issue #1999, opened 2020-04-05
Warnings about invalid cmdline options are printed twice


pmoravec:
```
$ sosreport -k blah.opt=val --batch --build -o nonsense

sosreport (version 3.9)

WARNING: unable to set option for disabled or non-existing plugin (blah)
WARNING: unable to set option for disabled or non-existing plugin (blah)


a non-existing plugin (nonsense) was specified in the command line
a non-existing plugin (nonsense) was specified in the command line
$
```

This was revealed during various testing of #1993 but it applies to current master as well => logging as a separate issue.

PR #2033, opened 2020-04-27
[sos] Don't double log error level messages

TurboTurtle:
Only set the logging handler that prints ERROR level messages to console
if we're running in quiet mode, as otherwise we'll double log from the
normal console handler.

Closes: #1999
Resolves: #2033

---

### DIFF
--- a/sos/component.py
+++ b/sos/component.py
@@ -202,7 +202,9 @@ class SoSComponent():
             else:
                 console.setLevel(logging.WARNING)
             self.soslog.addHandler(console)
-            # log ERROR or higher logs to stderr instead
+        # still log ERROR level message to console, but only setup this handler
+        # when --quiet is used, as otherwise we'll double log
+        else:
             console_err = logging.StreamHandler(sys.stderr)
             console_err.setFormatter(logging.Formatter('%(message)s'))
             console_err.setLevel(logging.ERROR)


TurboTurtle:
Before:
```
$ ./bin/sos report -o nonsense

sosreport (version 3.9)

a non-existing plugin (nonsense) was specified in the command line
a non-existing plugin (nonsense) was specified in the command line
```

After:
```
$ ./bin/sos report -o nonsense

sosreport (version 3.9)

a non-existing plugin (nonsense) was specified in the command line
```

BryanQuigley:
So if quiet is set errors go to stderr
But if it's not - errors (and warnings) go to stdout.

I don't have a strong preference (aside from being consistent), but I'd actually prefer DEBUG/WARNING/ERROR to go to stderr, and the "normal run" text to go to stdout.

TurboTurtle:
> So if quiet is set errors go to stderr
> But if it's not - errors (and warnings) go to stdout.
> 
> I don't have a strong preference (aside from being consistent), but I'd actually prefer DEBUG/WARNING/ERROR to go to stderr, and the "normal run" text to go to stdout.

I kept the errors on non-quiet runs to stdout to keep with what we've done historically, though we could probably split those out with some logging filters on the handlers. The other way to keep consistency between quiet and non-quiet would be to change `console_err` to send to stdout even though it's only ever going to spit out errors. The latter is a much simpler change, but we'd be in the same place we are today - errors on stdout.

I don't agree with debug or warning going to stderr, however. They aren't errors.

Issue #1750, opened 2019-08-11
[plugins] sizelimit not applied when add_copy_spec called to a dir


pmoravec:
Separating this issue from https://github.com/sosreport/sos/pull/1678#issuecomment-519891334:

Calling `add_copy_spec` to a directory, `stat` is applied to the directory "file" itself. That means, sizelimit is compared against the small size of the directory "file" itself, but not against the directory content.

We shall either replace `glob.glob()` at https://github.com/sosreport/sos/blob/master/sos/plugins/__init__.py#L1198 by `os.walk`, or merge parts of `add_copy_spec` with `_do_copy_path` (as `_do_copy_path` does the proper directory/symlink traversal needed to be done before applying the `sizelimit`).

pmoravec:
Will be resolved by #1870.

PR #2035, opened 2020-04-28
[Plugin] Apply size limit to directory specs

TurboTurtle:
Addresses an issue where size limiting was not being applied to whole
directory specifications, as we would only stat the directory directly
when it was returned from a glob. Now, if the given copyspec is just a
directory path, recurse into it with an explicit '*' match.

Updates the forbidden path check to account for the newly extended files
list to match against the directory-only forbidden specifications. This
may result in more log entries for deep directory structures whose
parent directory/directories is/are forbidden.

Closes: #1750
Resolves: #2035

---

### DIFF
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -42,7 +42,8 @@ def _mangle_command(command, name_max):
 
 
 def _path_in_path_list(path, path_list):
-    return any(p == path for p in path_list)
+    return any((p == path or path.startswith(os.path.abspath(p)+os.sep)
+                for p in path_list))
 
 
 def _node_type(st):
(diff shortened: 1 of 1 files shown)


TurboTurtle:
Yep, you're right there. I've updated the PR to check for directory paths within the first glob, and if those are found then we recurse into those directories and add the file contents for subdirs rather than the subdir "files" directly. The exception to that is if the subdir is empty, so we will still capture that directory to show end users the directory is empty, rather than risk implying that the collection failed.

TurboTurtle:
Ok, had a bit of a crazy moment there. Recursive symlinks are still a thing, duh.

I've updated the PR one more time with the proper approach, which is to _replace_ the `/path/to/some/dir` return from `glob` with the actual contents. The easiest way to do that is to check the copyspec first, rather than running it and checking afterwards.

This is a logic improvement rather than a change in approach with this PR, but would appreciate another once-over to make sure I haven't gone completely off the rails...

pmoravec:
Sure, will review in next few days. Thanks for fix!

TurboTurtle:
Ok, after much brain-wracking I finally have what I believe is our ideal state for directory size limiting based on directory-only copyspecs given to `add_copy_spec()`.

Basically, we check to see if the copyspec is a bare directory. If it is *and* it has content, we tack on a `/*` match for a recursive glob, to give us all the content within the directory. If there is no content, we do not add the splat, and `glob` returns just the (empty) directory path, which we will still copy in order to unambiguously show that the collection was successful, with 0 data collected.

Now, assuming the splat was added we then recursively call a new `__expand()` method which iterates over the content found in the directories. For files and links, we accept the path by itself without recursing into symlink dirs thus getting us caught in symlink hell. For directories, we apply the same  not-empty conditional as before - thus giving us arbitrary depth recursion without getting tripped up by recursive symlinks.

A few notes:

* From all my testing, we are not missing any data from archives when compared to current master. Yay.
* We are however getting ADDITIONAL data that was previously skipped over. This data is under `/sys/devices/` and pertains to symlink dirs that point to another location within `/sys` (usually `/sys/class`). Before this symlink was ignored, but we now capture and properly link to the collection point which is usually collected by a different plugin.
* Due to my inability to let this go and being distracted for the last ~24 hours with this and this alone, my wife now knows she too hates recursive symlinks despite not knowing what they actually are.
* Directory-based copyspecs will now return a full list with each individual file location under those directories, this means you can expect more chatty log messages when dealing with matching forbidden paths.

PR #2038, opened 2020-04-30
[Plugin|390] Fix architecture-only enablement checks

TurboTurtle:
First commit fixes the enablement checks for plugins that only specify an `architectures` tuple, and no others like `packages`, `files`, etc...

Second commit updates the s390 plugin to use that style of trigger, like the powerpc plugin.

---

### DIFF
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -1775,6 +1775,11 @@ class Plugin(object):
         return True
 
     def _check_plugin_triggers(self, files, packages, commands, services):
+
+        if not any([files, packages, commands, services]):
+            # no checks beyond architecture restrictions
+            return self.check_is_architecture()
+
         return ((any(os.path.exists(fname) for fname in files) or
                 any(self.is_installed(pkg) for pkg in packages) or
                 any(is_executable(cmd) for cmd in commands) or
--- a/sos/report/plugins/s390.py
+++ b/sos/report/plugins/s390.py
@@ -17,13 +17,7 @@ class S390(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
 
     plugin_name = 's390'
     profiles = ('system', 'hardware')
-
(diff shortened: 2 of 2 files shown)


pmoravec:
Failed Travis test: https://travis-ci.org/github/sosreport/sos/jobs/681621115#L611 : "Disk quota exceeded" isnt relevant to the PR.

sos/report/plugins/__init__.py
```diff
@@ -1775,6 +1775,11 @@ def check_enabled(self):
         return True
 
     def _check_plugin_triggers(self, files, packages, commands, services):
+
+        if not any([files, packages, commands, services]):
```
bmr-cymru, line 1779: This would allow a file, package, command or service trigger to bypass an arch exclusion. Is that what we want, or should we go ahead and check other triggers iff the arch is a match?

I.e. should this be like:

```python
    if not self.check_is_architecture():
        return False

    # Other enablement checks
```

I.e. being on a not-matching architecture excludes us from running *regardless* of any other checks.

TurboTurtle: If we have any files, package, commands, or services, then we skip past this and continue to the existing check which includes the arch check as an AND condition.

Though the simpler arch-only check is easier to understand at a glance...


Yeah, I'll change that over.

TurboTurtle: Wait, actually - no that would still prevent plugins from enabling when they _only_ specify an arch. We'd pass the `check_is_architecture()` conditional, but then we'd fail the following (existing) conditional like we still do today, since all files, commands, packages, and services would evaluate `False`.

bmr-cymru: Nm - `check_is_architecture()` is also (correctly) tested in the main `return` clause so the behaviour is that plugins with just an arch test rely only on that check, whereas those that also define package etc. triggers must match one or more of those AND the arch, which is exactly what we want.

PR #2039, opened 2020-05-01
[tests] Add Timing, summary, and reduce IO for small files

BryanQuigley:
Timing - see how long each test takes
Summary - output the key results at the end
Use /dev/shm to store temporary files - should help ppc tests
Make since command actually meaningful

---


BryanQuigley:
Well, I think it improved it a bit, but it still fails sometimes...

PR #2040, opened 2020-05-02
[containers_common] Add plugin for common containers configs

pmoravec:
Move collection of whole /etc/containers and /usr/share/containers to
one containers_common plugin enabled by the package of the same name.

Since the package is a common dependency for buildah and podman, no regression
in default data collection happens.

Resolves: #2040

---

### DIFF
--- a/sos/report/plugins/containers_common.py
+++ b/sos/report/plugins/containers_common.py
@@ -1,0 +1,27 @@
+# Copyright (C) 2020 Red Hat, Inc., Pavel Moravec <pmoravec@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, RedHatPlugin, UbuntuPlugin
+
+
+class ContainersCommon(Plugin, RedHatPlugin, UbuntuPlugin):
+
+    short_desc = 'Common container configs under {/etc,/usr/share}/containers'
+    plugin_name = 'containers_common'
+    profiles = ('container', )
+    packages = ('containers-common', )
+
+    def setup(self):
+        self.add_copy_spec([
+            '/etc/containers/*',
+            '/usr/share/containers/*',
+        ])
+
+# vim: set et ts=4 sw=4 :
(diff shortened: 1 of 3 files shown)


pmoravec:
The dependency among packages ensures no regression on Red Hat distro. Since `podman` plugin is enabled also for Ubuntu, is this change safe also for this distro? Or should I remove `UbuntuPlugin` from `ContainersCommon` and collect the configs in a separate `PodmanUbuntuPlugin`?

Does e.g. @BryanQuigley  know?

BryanQuigley:
This change appears safe for Ubuntu.  Thanks for checking.  (Although podman/containers-common/buildah are in an external repo.)  This change makes sense to me, but I'm not familiar with podman and friends.

containers-common provides those config files:
# dpkg -S /etc/containers/policy.json 
containers-common: /etc/containers/policy.json

sos/report/plugins/containers_common.py
```diff
@@ -0,0 +1,27 @@
...
+    short_desc = 'Common container configs under {/etc,/usr/share}/containers'
+    plugin_name = 'containers_common'
+    profiles = ('container',)
+    packages = ('containers-common')
+
+    def setup(self):
+        elf.add_copy_spec([
+            '/etc/containers/*',
+            '/usr/share/containers/*',
+            ])
```
TurboTurtle, line 25: Small typo `elf` instead of `self`, and closing indent needs to come back one level.

pmoravec: No elves allowed in `sos`? :D Shame on me for such typos :( And thanks for your good eyes.

TomSweeneyRedHat:
@rhatdan PTAL
@jwhonce @lsm5 FYI

sos/report/plugins/containers_common.py
```diff
@@ -0,0 +1,27 @@
...
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, RedHatPlugin, UbuntuPlugin
+
+
+class ContainersCommon(Plugin, RedHatPlugin, UbuntuPlugin):
+
+    short_desc = 'Common container configs under {/etc,/usr/share}/containers'
```
rhatdan, line 16: Do we care about homedirectories.
/home/*/.config/containers?

control-d: The request to capture non-root/rootless info as a whole is covered in rhbz#1776549.  If/when that's implemented that should take care of ~/.config/containers

rhatdan: Thanks.

rhatdan:
LGTM

PR #2041, opened 2020-05-04
[sos] Write to host filesystem if in container with HOST env var set

TurboTurtle:
Moving tmpdir specifications out of policy gave us the ability to have
logging within `Policy`, however it was unnoticed before that this
stopped the redirection from container filesystem to host filesystem
when sos was running in a container (and we wanted to save the archive
to the host and not the container).

Fix this by standardizing a check for a `HOST` environment variable,
when a `container` env var is set as well. This has been used by the Red
Hat policy for some time to determine chroot locations for when we're
running in a container, and has been working well. The `container`
environment variable should be set at container setup by modern
container runtimes.

If either the `container` or `HOST` environment variable is not set, but
we are running in a container, then we will continue to write to the
container's filesystem (while policy will still determine any chroot
requirements separately).

Resolves: #2041

---

### DIFF
--- a/sos/component.py
+++ b/sos/component.py
@@ -77,7 +77,7 @@ class SoSComponent():
         self.opts = self.load_options()
 
         if self.configure_logging:
-            tmpdir = self.opts.tmp_dir or '/var/tmp'
+            tmpdir = self.get_tmpdir_default()
 
             if not os.path.isdir(tmpdir) \
                     or not os.access(tmpdir, os.W_OK):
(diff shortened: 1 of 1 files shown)


TurboTurtle:
@BryanQuigley since this is just extending the existing checks from the RH-family containerized distros to be a default behavior, I'd like to get a Ubuntu maintainer sign-off for this. Though honestly I'm not sure how common running sos in a container (and wanting host information) is in Ubuntu.

PR #2044, opened 2020-05-06
[cups] Call add_journal() with output set to 'verbose'

ptalbert:
Cups now stores the Job ID as a journal field JID. Unless the journalctl
is captured with the 'verbose' output option the information will not be
available in the report..

---

### DIFF
--- a/sos/report/plugins/cups.py
+++ b/sos/report/plugins/cups.py
@@ -39,6 +39,6 @@ class Cups(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
             "lpstat -d"
         ])
 
-        self.add_journal(units="cups")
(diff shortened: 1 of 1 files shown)


TurboTurtle:
I'm not a fan of just replacing the journal collection with the far less readable verbose output for the one extra field. journalctl supports an `--output-fields` option, however it only works with certain `--output` choices. We'd need to extend `add_journal()` to support this, but I'm curious if there's an invocation using `--output-fields=JOB_ID` that works for you as an _additional_ journal collection rather than replacing the current collection?

TurboTurtle:
Cycling around on this. @ptalbert - would the additional journal call describe above work for you if we extend `add_journal` to support selecting output fields?

TurboTurtle:
Closing as abandoned.

Issue #1990, opened 2020-03-23
[4.0] Create and save metadata about report runs in archive


TurboTurtle:
Following #1959, sos should save metadata information and save it in the final archive. This metadata can be used by both the sos developers and downstream organizations to assist with ongoing maintenance and trend analysis.

TurboTurtle:
I've mocked up an example of what I envision the layout of this to look like:

```
{'version': 4.0
 'cmdline': 'sos report --blah'
 'start_time': 'some_datestamp'
 'end_time': 'some_other_datestamp'
 'components':
	# each subcommand sos supports would have an element here if it ran
	'report':
		'plugins_run': [] # a list of plugin names for those that ran
		'plugins_disabled': [] # list of explicitly disabled plugins
		'plugins_enabled': [] # list of explicitly enabled plugins
		'plugins':
			# one entry per plugin
			'kernel':
				'options': [] # list of enabled plugin options
				'commands': [
					{'command': 'foo', 'parameters': ['bar', '--foobar'], 'exec': 'foo bar --foobar', 'filepath': 'sos_commands/kernel/foo_bar_--foobar', 'exit_status': 0, 'run_time': '12.345s', tags=['tag1','foo_bar']}
				]
				'files': [
		        		{'specification': '/etc/some/*.conf', 'files_copied': ['/etc/some/1.conf', '/etc/some/other.conf'], tags=['some_configs']}
				]
		'hostname': 'foo.bar.example.com',
		'policy: 'RedHat'
		
	'collect':
                'cluster': 'pacemaker',
		'hosts': [
			{'hostname': 'foo.bar.example.com', 'archive': 'relative/path/in/tarball/sosreport.tar.gz'}
		]
}
```

I think this is for the most part very straight forward. One part that might not be is the `tags` parameter  in the `commands` and `files` element for each plugin. This is an idea coming from working with the RH Insights team - by the use of tags such as this, it is far easier for third party analysis tools to locate relevant data. For example, in the `apache` plugin, we can tag the collection of both `/etc/httpd/conf/httpd.conf` and `/etc/httpd/conf.d/*.conf` as, for example `httpd_confs`. Then, these other tools can simply parse the metadata file and look for all things tagged with `httpd_confs` rather than having to grok through file name patterns.

[...]

PR #2050, opened 2020-05-06
[component|collect|report] Add metadata collection

TurboTurtle:
Adds facilities to capture metadata information about a particular sos
run, and save it to any resulting archive as `manifest.json`. This
manifest will include component-specific information that should assist
in analysis of sos runs.

For `report`, this is an extension of existing reporting functionality,
that records information on a per-plugin basis regarding command output
and file collection. However, more information is recorded in this
collection than the JSON or HTML reporting structures that review a
finished archive.

Manifest data is added at the time of collection; for commands this includes
a breakout of command, parameters, full execution syntax, filepath relative
to archive root, and the actual return code of the command, even if non-zero.
For files, the copied files are grouped by the specification set in
`add_copy_spec()` calls, I.E. for any file pattern regex that collects
at least one file, all files collected by that regex are listed.

Additionally, this adds a `tags` parameter to `add_copy_spec()`,
`add_cmd_output()`, and `collect_cmd_output()` that if used will add
tagging information to the manifest entries for those collections. This
is intended to assist in future sos work (E.G. `sos examine`) and third
party utilities in identifying relevant/connected data that is present
in the archive. Note that all collections specified by a single call to
any of these collection methods will inherit the same tags.

For `collect` we now record the cluster type, node list, and a separate
entry for each node in the node list that includes the sos version, node
sos policy, and the archive within the collect tarball that was
collected for that host.

Closes: #1990
Resolves: #2050

---

### DIFF
--- a/sos/archive.py
+++ b/sos/archive.py
@@ -16,6 +16,7 @@ import logging
 import codecs
 import errno
 import stat
+from datetime import datetime
 from threading import Lock
 
 from sos.utilities import sos_get_command_output, is_executable
@@ -130,13 +131,15 @@ class FileCacheArchive(Archive):
     _archive_root = ""
     _archive_name = ""
 
-    def __init__(self, name, tmpdir, policy, threads, enc_opts, sysroot):
+    def __init__(self, name, tmpdir, policy, threads, enc_opts, sysroot,
+                 manifest=None):
         self._name = name
         self._tmp_dir = tmpdir
         self._policy = policy
         self._threads = threads
         self.enc_opts = enc_opts
         self.sysroot = sysroot or '/'
+        self.manifest = manifest
         self._archive_root = os.path.join(tmpdir, name)
         with self._path_lock:
             os.makedirs(self._archive_root, 0o700)
@@ -532,6 +535,19 @@ class FileCacheArchive(Archive):
         if os.path.isdir(self._archive_root):
             shutil.rmtree(self._archive_root)
 
+    def add_final_manifest_data(self, method):
+        """Adds component-agnostic data to the manifest so that individual
+        SoSComponents do not need to redundantly add these manually
+        """
+        end = datetime.now()
+        start = self.manifest.start_time
+        run_time = end - start
+        self.manifest.add_field('end_time', end)
+        self.manifest.add_field('run_time', run_time)
+        self.manifest.add_field('compression', method)
+        self.add_string(self.manifest.get_json(indent=4),
+                        os.path.join('sos_reports', 'manifest.json'))
+
     def finalize(self, method):
         self.log_info("finalizing archive '%s' using method '%s'"
                       % (self._archive_root, method))
@@ -607,9 +623,10 @@ class TarFileArchive(FileCacheArchive):
     method = None
     _with_selinux_context = False
 
-    def __init__(self, name, tmpdir, policy, threads, enc_opts, sysroot):
+    def __init__(self, name, tmpdir, policy, threads, enc_opts, sysroot,
+                 manifest=None):
         super(TarFileArchive, self).__init__(name, tmpdir, policy, threads,
-                                             enc_opts, sysroot)
+                                             enc_opts, sysroot, manifest)
         self._suffix = "tar"
         self._archive_name = os.path.join(tmpdir, self.name())
 
--- a/sos/report/__init__.py
+++ b/sos/report/__init__.py
@@ -129,6 +129,11 @@ class SoSReport(SoSComponent):
 
         self._is_root = self.policy.is_root()
 
+        # add a manifest section for report
+        self.manifest.components.add_section('report')
+        # shorthand reference for ease of maintenance
+        self.report_md = self.manifest.components.report
+
         # user specified command line preset
         if self.opts.preset != self.arg_defaults["preset"]:
             self.preset = self.policy.find_preset(self.opts.preset)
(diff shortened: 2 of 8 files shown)


sos/report/plugins/__init__.py
```diff
@@ -479,6 +495,11 @@ def timeout(self):
             return _timeout
         return self.plugin_timeout
 
+    def set_timeout_hit(self):
+        self._timeout_hit = True
+        self.manifest.add_field('end_time', time())
```
pmoravec, line 500: Doesnt it matter we store in `end_time` a different format (`time.time()`) than for non-timeouted plugins (`datetime.datetime.now()`)?

Is it worth adding also `run_time`?

sos/report/__init__.py
```diff
@@ -903,15 +916,21 @@ def _collect_plugin(self, plugin):
...
                 # Re-type int 0 to NoneType, as otherwise result() will treat
                 # it as a literal 0-second timeout
-                timeout = self.loaded_plugins[plugin[0]-1][1].timeout or None
+                timeout = _plug.timeout or None
+                start = datetime.now()
+                _plug.manifest.add_field('start_time', start)
                 t.result(timeout=timeout)
+                end = datetime.now()
+                _plug.manifest.add_field('end_time', end)
+                _plug.manifest.add_field('run_time', end - start)
```
pmoravec, line 929: A note from an English forever-noob: the `[start|end|run]_time` is well human understandable, but logically incoherent with `setup_[start|end|time]`. But I failed to find out labels that are both logically coherent and such straighforwardly understandable :-/ .

sos/report/plugins/__init__.py
```diff
@@ -1368,10 +1409,22 @@ def _collect_cmd_output(self, cmd, suggest_filename=None,
...
+                        'parameters': cmd.split(' ')[1:],
+                        'exec': cmd,
+                        'filepath': None,
+                        'return_code': result['status'],
+                        'run_time': time() - start,
+                        'tags': tags
+                    })
+                    return result
+
+        end = time() - start
```
pmoravec, line 1424: code de-duplication: isnt it worth having:

    end = time() - start
    manifest_cmd = { 'command': .. .. .. 'tags': tags }
    if <command_not_found>:
        if self.manifest:
            self.manifest.commands.append(manifest_cmd)
        return result
    ..
    if manifest:
        manifest_cmd['filepath'] = outfn
        self.manifest.commands.append(manifest_cmd)

(maybe with better name of `manifest_cmd` variable)

pmoravec: Also, `end` variable has confusing name (cf. `'run_time': end,`) - could it be renamed to e.g. `runtime`?

sos/report/plugins/__init__.py
TurboTurtle: Oh, good catch! Yup, this should all use `datetime`. Adding a `run_time` field sounds good too

sos/report/plugins/__init__.py
TurboTurtle: Ack to both.

sos/report/__init__.py
TurboTurtle: Yeah I don't know of a better way to phrase it, but at the same time I don't think this will be a real issue seeing as the contributors would be the only ones interested in the `setup_` fields, and once you're familiar with the fact that plugins run `setup()` before collection, it tracks easily (at least in my mind).

TurboTurtle:
@pmoravec updated the PR sans the `setup_` field names noted above.

sos/archive.py
```diff
@@ -532,9 +535,24 @@ def cleanup(self):
...
+        self.manifest.add_field('run_time', run_time)
+        self.manifest.add_field('compression', method)
+        self.add_string(self.manifest.get_json(indent=4),
+                        os.path.join('sos_logs', 'manifest.json'))
+
     def finalize(self, method):
         self.log_info("finalizing archive '%s' using method '%s'"
                       % (self._archive_root, method))
+        if self.manifest is not None:
+            self.add_final_manifest_data(method)
```
pmoravec, line 555: No metadata are written when `--build` option is used. Since this `finalize` method is called only when `--build` isnt set (cf. https://github.com/sosreport/sos/blob/master/sos/report/__init__.py#L1099-L1106).

A possible fix is to call at https://github.com/sosreport/sos/blob/master/sos/report/__init__.py#L1122 either `self.archive.add_final_manifest_data(None)` directly, or `self.archive.finalize(None)` (that would require some updates in `finalize` to skip compression when `method=None`).

pmoravec:
Two - rather stylistic - issues that I dont have a strong opinion on: resulting `manifest.json` has two types of duration:

- `"run_time": "0:00:07.225778",` for plugins run time
- `"run_time": 0.07651472091674805,` for commands run time

My comments:
- the high precision of commands run time is redundant (isnt calling such high precision (date)time more expensive, also?)
- we shall have (do we really?) one common format for the run time; but if, so, which one? `H:MM:SS.sssss` has pros as human readable, pure `S.ssss` is better for automatic processing

I dont have strong preferences of either comment and I am fine with the current implementation wrt. this. Just raising as points for a consideration/assessment.

pmoravec:
> @pmoravec updated the PR sans the `setup_` field names noted above.

No other comments than those two (`--build`, and runtime format). Good work!

sos/archive.py
TurboTurtle: Yeah, my hope was that putting it in `archive.finalize()` was to "automagic" it for any components that write one, but I overlooked this.

But, at the same time not a big issue to make the call in each component.

TurboTurtle:
> * we shall have (do we really?) one common format for the run time; but if, so, which one? `H:MM:SS.sssss` has pros as human readable, pure `S.ssss` is better for automatic processing
> 

So, the command execution precision was already there - I just lifted it into the manifest data. It's what we currently log, e.g.:

```
[plugin:kernel] collecting output of 'sysctl -a'
[plugin:kernel] collected output of 'sysctl' in 0.034815073013305664 (changes=False)
[plugin:kernel] unpacked command: binary=False, changes=False, chroot=True, cmd='dkms status', env=None, foreground=False, root_symlink=None, runat=None, sizelimit=25, stderr=True, subdir=None, suggest_filename=None, timeout=300
[plugin:kernel] collecting output of 'dkms status'
[plugin:kernel] collected output of 'dkms' in 0.08468127250671387 (changes=False)
[plugin:kernel] collected plugin 'kernel' in 2.217618465423584
```

I put the plugin's `*_time` in datetime format because that is easier to read for humans. I think for now we should move forward with the current implementation, but we can always pivot if this causes issues or confusion in real world runs.

Unless of course anyone has an objection to doing that, in which case just let us know here before merge.

TurboTurtle:
I've updated the branch to write the manifest during `final_work()` before we decide what to do with `--build`. Also moved the location to `sos_reports/manifest.json` per IRC.

PR #2051, opened 2020-05-07
Ovs debug enhancements

orgcandman:
These represent enhancements to the OvS debug plugin.

---

- [x] If this commit closes an existing issue, is the line `Closes: #ISSUENUMBER` included in an independent line?  -- no issues in the issue tracker being resolved

### DIFF
--- a/sos/report/plugins/openvswitch.py
+++ b/sos/report/plugins/openvswitch.py
@@ -169,6 +169,14 @@ class OpenVSwitch(Plugin):
                 )
                 if port_list_result['status'] == 0:
                     for port in port_list_result['output'].splitlines():
+                        self.add_cmd_output([
+                            "ovs-appctl cfm/show %s" % port,
+                            "ovs-appctl qos/show %s" % port,
+                            # Not all ports are "bond"s, but all "bond"s are
+                            # a single port
+                            "ovs-appctl bond/show %s" % port,
+                        ])
+
                         if check_dpdk:
                             self.add_cmd_output(
                                 "ovs-appctl netdev-dpdk/get-mempool-info %s" %
--- a/sos/report/plugins/openvswitch.py
+++ b/sos/report/plugins/openvswitch.py
@@ -114,14 +114,17 @@ class OpenVSwitch(Plugin):
         if br_list_result['status'] == 0:
             for br in br_list_result['output'].splitlines():
                 self.add_cmd_output([
+                    "ovs-appctl bridge/dump-flows --offload-stats %s" % br,
+                    "ovs-appctl dpif/show-dp-features %s" % br,
                     "ovs-appctl fdb/show %s" % br,
+                    "ovs-appctl fdb/stats-show %s" % br,
+                    "ovs-appctl mdb/show %s" % br,
                     "ovs-ofctl dump-flows %s" % br,
                     "ovs-ofctl dump-ports-desc %s" % br,
                     "ovs-ofctl dump-ports %s" % br,
                     "ovs-ofctl queue-get-config %s" % br,
                     "ovs-ofctl queue-stats %s" % br,
                     "ovs-ofctl show %s" % br,
-                    "ovs-appctl fdb/stats-show %s" % br
                 ])
 
                 # Flow protocols currently supported
--- a/sos/report/plugins/openvswitch.py
+++ b/sos/report/plugins/openvswitch.py
@@ -53,8 +53,11 @@ class OpenVSwitch(Plugin):
             "ovs-vsctl -t 5 show",
             # Gather the database.
             "ovsdb-client -f list dump",
-            # List the contents of runtime directory
+            # List the contents of important runtime directories
             "ls -laZ /run/openvswitch",
+            "ls -laZ /dev/hugepages/",
+            "ls -laZ /dev/vfio",
+            "ls -laZ /var/lib/vhost_sockets",
             # List devices and their drivers
             "dpdk_nic_bind --status",
             "dpdk_devbind.py --status",
--- a/sos/report/plugins/openvswitch.py
+++ b/sos/report/plugins/openvswitch.py
@@ -183,6 +183,16 @@ class OpenVSwitch(Plugin):
                                 port
                             )
 
+                if check_dpdk:
+                    iface_list_result = self.exec_cmd(
+                        "ovs-vsctl -t 5 list-ifaces %s" % br
+                    )
+                    if iface_list_result['status'] == 0:
+                        for iface in iface_list_result['output'].splitlines():
+                            self.add_cmd_output(
+                                "ovs-appctl netdev-dpdk/get-mempool-info %s" %
+                                iface)
+
 
 class RedHatOpenVSwitch(OpenVSwitch, RedHatPlugin):
 
--- a/sos/report/plugins/openvswitch.py
+++ b/sos/report/plugins/openvswitch.py
@@ -129,7 +129,9 @@ class OpenVSwitch(Plugin):
                     "OpenFlow10",
                     "OpenFlow11",
                     "OpenFlow12",
-                    "OpenFlow13"
+                    "OpenFlow13",
+                    "OpenFlow14",
+                    "OpenFlow15"
                 ]
 
                 # List protocols currently in use, if any
--- a/sos/report/plugins/openvswitch.py
+++ b/sos/report/plugins/openvswitch.py
@@ -29,6 +29,11 @@ class OpenVSwitch(Plugin):
             '/usr/local/var/log/openvswitch/',
         ]
 
+        dpdk_enabled = self.collect_cmd_output(
+            "ovs-vsctl -t 5 get Open_vSwitch . other_config:dpdk-init")
+        check_dpdk = (dpdk_enabled["status"] == 0 and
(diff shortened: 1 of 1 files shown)


sos/report/plugins/openvswitch.py
```diff
@@ -29,6 +29,13 @@ def setup(self):
             '/usr/local/var/log/openvswitch/',
         ]
 
+        check_dpdk = False
+        dpdk_enabled = self.collect_cmd_output(
+            "ovs-vsctl -t 5 get Open_vSwitch . other_config:dpdk-init")
+        if dpdk_enabled["status"] == 0 and \
+           dpdk_enabled["output"].startswith("true"):
```
pmoravec, line 36: Isn't it easier to have:

    check_dpdk = (dpdk_enabled["status"] == 0 and \
                            dpdk_enabled["output"].startswith("true"))

? As the line 32 now is superfluous.

sos/report/plugins/openvswitch.py
```diff
@@ -147,18 +166,34 @@ def setup(self):
...
+
+                iface_list_result = self.exec_cmd(
+                    "ovs-vsctl -t 5 list-ifaces %s" % br
+                )
+                if iface_list_result['status'] == 0:
+                    for iface in iface_list_result['output'].splitlines():
+                        if check_dpdk:
+                            self.add_cmd_output(
+                                "ovs-appctl netdev-dpdk/get-mempool-info %s" %
+                                iface)
```
pmoravec, line 196: I would vote to put this whole section under the `if check_dpdk:` test. Since if `check_dpdk` is `False`, whole current section does not store anything but runs redundant commands.

pmoravec:
Two generic comments additionally to the code comments:

1) cant some commits be squashed? The granularity of changes is so low :) (but I am fine with the current approach)

2) I see various commands removed and others added. Can't the removed ones be important for some other use case? (i.e. ideally some other OVS guy shall review the change).

sos/report/plugins/openvswitch.py
bmr-cymru: Backslash escaping a newline is superfluous if you're inside parenthesis ;-)

bmr-cymru:
Commits look fine to me (I far prefer things broken out, especially when it comes to bisecting - it's not like we are charged per commit :-) ).

sos/report/plugins/openvswitch.py
orgcandman: @pmoravec  I can do that

orgcandman:
> Two generic comments additionally to the code comments:
> 
>     1. cant some commits be squashed? The granularity of changes is so low :) (but I am fine with the current approach)

I think the style has been to use commits which are granular.  I tried to keep changes grouped so that the most impactful are separate commits that can be reverted.

>     2. I see various commands removed and others added. Can't the removed ones be important for some other use case? (i.e. ideally some other OVS guy shall review the change).

The stuff that was removed is added using commands that will actually work. For example, the ovs-dpctl commands aren't going to show hardware offload flows, or userspace dp flows.  But ovs-appctl dpctl/show system@ovs-system and ovs-appctl dpctl/show netdev@ovs-netdev (which is discovered in dpctl/dump-dps command) will show all of that.  None of the removed commands are lost - all are recorded, but with more information.  And many of these commands have existed since release 2.6, so it's a many year backward compatibility.  I'll ask another OvS contributor to review, though.

sos/report/plugins/openvswitch.py
```diff
@@ -147,18 +152,16 @@ def setup(self):
...
                     "ovs-vsctl -t 5 list-ports %s" % br
                 )
                 if port_list_result['status'] == 0:
                     for port in port_list_result['output'].splitlines():
-                        self.add_cmd_output(
-                            "ovs-appctl netdev-dpdk/get-mempool-info %s" % port
-                        )
+                        if check_dpdk:
+                            self.add_cmd_output(
+                                "ovs-appctl netdev-dpdk/get-mempool-info %s" %
```
kevintraynor, line 182: The mempool info is missing when I ran with dpdk-init=true and 1x dpdk and 1x dpdkvhostclient ports

sos/report/plugins/openvswitch.py
```diff
@@ -79,17 +79,19 @@ def setup(self):
             # Capture DPDK and other parameters
             "ovs-vsctl -t 5 get Open_vSwitch . other_config",
             # Capture OVS list
-            "ovs-vsctl list Open_vSwitch",
+            "ovs-vsctl -t 5 list Open_vSwitch",
             # Capture OVS interface list
-            "ovs-vsctl list interface",
+            "ovs-vsctl -t 5 list interface",
```
kevintraynor, line 88: "ovs-vsctl list bridge" below and further down also need "-t 5"

sos/report/plugins/openvswitch.py
```diff
@@ -79,17 +79,19 @@ def setup(self):
...
             "ovs-vsctl list bridge",
             # Capture DPDK queue to pmd mapping
             "ovs-appctl dpif-netdev/pmd-rxq-show",
             # Capture DPDK pmd stats
             "ovs-appctl dpif-netdev/pmd-stats-show",
             # Capture DPDK pmd performance counters
-            "ovs-appctl dpif-netdev/pmd-perf-show"
+            "ovs-appctl dpif-netdev/pmd-perf-show",
+            # Capture ofproto tunnel configs
+            "ovs-appctl ofproto/list-tunnels"
```
kevintraynor, line 98: minor: seems like this leaked into the wrong commit

sos/report/plugins/openvswitch.py
```diff
@@ -109,6 +98,18 @@ def setup(self):
         self.add_journal(units="ovs-vswitchd")
         self.add_journal(units="ovsdb-server")
 
+        # Gather the datapath information for each datapath
+        dp_list_result = self.collect_cmd_output('ovs-appctl dpctl/dump-dps')
+        if dp_list_result['status'] == 0:
+            for dp in dp_list_result['output'].splitlines():
+                self.add_cmd_output([
+                    "ovs-appctl dpctl/show -s %s" % dp,
```
kevintraynor, line 112: would be good to add a comment in commit msg on why changing from ovs-dpctl to ovs-appctl

sos/report/plugins/openvswitch.py
kevintraynor: # ls ovsmaster/*mempool*
ovsmaster/ovs-appctl_netdev-dpdk.get-mempool-info_dpdkvhost0  ovsmaster/ovs-appctl_netdev-dpdk.get-mempool-info_myport

# ls ovsaaron/*mempool*
ls: cannot access 'ovsaaron/*mempool*': No such file or directory

# ovs-vsctl -t 5 get Open_vSwitch . other_config:dpdk-init
"true"

# ovs-appctl netdev-dpdk/get-mempool-info dpdkvhost0 | grep mempool
mempool <ovs530eb21900021580262144>@0x23f7707c0

sos/report/plugins/openvswitch.py
orgcandman: will do

sos/report/plugins/openvswitch.py
orgcandman: will do

sos/report/plugins/openvswitch.py
orgcandman: Yep - the test is incorrect.  A correct version needs to account for " being included.

sos/report/plugins/openvswitch.py
orgcandman: Yep, I've moved it to the datapath commit.

orgcandman:
I've rebased and pushed after addressing all comments.

sos/report/plugins/openvswitch.py
```diff
@@ -29,6 +29,11 @@ def setup(self):
             '/usr/local/var/log/openvswitch/',
         ]
 
+        dpdk_enabled = self.collect_cmd_output(
+            "ovs-vsctl -t 5 get Open_vSwitch . other_config:dpdk-init")
+        check_dpdk = (dpdk_enabled["status"] == 0 and
+                      dpdk_enabled["output"].startswith('"true"'))
```
kevintraynor, line 35: I know it's not used much (or at all?), but "try" is a valid value here too, in which case you would want to check the mempools. What is the advantage of the conditional? It may save a few cmds, but not sure if it's more trouble than it's worth.

Tested that with "true" the mempool information is present. Actually, it's called twice in my case due to dpdk ports being in port and interface lists, but I don't see that as an issue.

orgcandman: I agree - I don't think anyone use it in production currently.  Actuallly, until DPDK has a robust uninit and reinit mechanism, "try" will not be recommended or used - I would prefer to wait until the uninit / reinit path is better.  For now "try" is just a debug tool for system integrators.

kevintraynor: ok, as "try" is _intended_ for use as a debug method when DPDK initialization is failing and in that case mempool information will not be available, the current approach seems fine.

kevintraynor:
ACK for this series

PR #2052, opened 2020-05-07
[openshift] Rewrite plugin for OCP 4.x

TurboTurtle:
Rewrites the `openshift` plugin to OCP 4.x, as it currently is written for Openshift 2.x which is an EOL product.

This is done as a separate plugin instead of extending `kubernetes` again due to the significant differences in collection requirements and volume. This plugin was made in direct collaboration with the OCP support team within RH.

The second commit in this set removes the subdir depth limitation so that the new `openshift` plugin can sanely organize the amount of data collected.

---

### DIFF
--- a/sos/report/plugins/openshift.py
+++ b/sos/report/plugins/openshift.py
@@ -49,7 +49,7 @@ class Openshift(Plugin, RedHatPlugin):
     short_desc = 'Openshift Container Platform 4.x'
 
     plugin_name = "openshift"
-    plugin_timeout = 600
+    plugin_timeout = 900
     profiles = ('openshift',)
     packages = ('openshift-hyperkube',)
 
@@ -79,8 +79,8 @@ class Openshift(Plugin, RedHatPlugin):
             return True
 
         # Not logged in currently, attempt to do so
-        if self.get_option('token') or os.getenv('SOSOCPTOKEN', None):
-            token = self.get_option('token') or os.getenv('SOSOCPTOKEN', None)
+        token = self.get_option('token') or os.getenv('SOSOCPTOKEN', None)
+        if token:
             oc_res = self.exec_cmd("oc login %s --token=%s "
                                    "--insecure-skip-tls-verify=True"
                                    % (self.get_option('host'), token))
@@ -316,10 +316,9 @@ class Openshift(Plugin, RedHatPlugin):
                     )
                 # check for podlogs here as a slight optimization to re-running
                 # 'oc get pods' on all namespaces
-                if res == 'pods' and _instances:
-                    if self.get_option('podlogs'):
-                        pod_list = [p.split()[0] for p in _instances]
-                        self.collect_podlogs(namespace, pod_list)
+                if res == 'pods' and _instances and self.get_option('podlogs'):
+                    pod_list = [p.split()[0] for p in _instances]
+                    self.collect_podlogs(namespace, pod_list)
 
     def collect_podlogs(self, namespace, pod_list):
         """For any namespace that has active pods in it, collect the current
@@ -348,21 +347,26 @@ class Openshift(Plugin, RedHatPlugin):
     def postproc(self):
 
         # clear any certificate output
-        self.do_cmd_private_sub('oc')
+        self.do_cmd_private_sub('oc ')
         self.do_file_private_sub('/etc/kubernetes/*')
 
         # clear the certificate data from /etc/kubernetes that does not have
         # the certificate banners that the _private_sub() methods look for
         _fields = [
+            '.*.crt',
             'client-certificate-data',
             'client-key-data',
             'certificate-authority-data',
-            'token'
+            '.*.key',
+            'token',
+            '.*token.*.value'  # don't blind match `.*token.*` and lose names
         ]
 
         regex = r'(\s*(%s):)(.*)' % '|'.join(_fields)
 
         self.do_path_regex_sub('/etc/kubernetes/*', regex, r'\1 *******')
+        # scrub secret content
+        self.do_cmd_output_sub('secrets', regex, r'\1 *******')
 
         # `oc describe` output can include url-encoded file content. For the
         # most part this is not important as the majority of these instances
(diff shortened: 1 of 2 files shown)


TurboTurtle:
Updated the delimiter to be `:` instead of spaces, per discussion with @pmoravec related to #2055

pmoravec:
Not knowing OCP at all, so maybe raising a silly question to:

> Similarly, each default namespace (project) that exists at deployment time will be collected from, ..

Can OCP add projects after deployment (and is it a common practice)? If so, shall not we collect the currently existing projects rather than those existing during deployment?

sos/report/plugins/openshift.py
```diff
@@ -7,151 +7,369 @@
...
+    the token from being recorded in output saved to the archive. Option #1 may
+    be used if this is considered an acceptable risk. It is not recommended to
+    rely on option #3, though it will provide the functionality needed.
+    """
+
+    short_desc = 'Openshift Container Platform 4.x'
 
     plugin_name = "openshift"
-    profiles = ('virt', 'openshift')
+    plugin_timeout = 600
```
pmoravec, line 52: If you stated

> Total runtime of just the openshift plugin from start to finish averages around 10 minutes,

isn't the default timeout to 10 minutes too low?

sos/report/plugins/openshift.py
```diff
@@ -7,151 +7,369 @@
...
+    def _check_oc_logged_in(self):
+        """See if we're logged in to the API service, and if not attempt to do
+        so using provided plugin options
+        """
+        if self._check_oc_function():
+            return True
+
+        # Not logged in currently, attempt to do so
+        if self.get_option('token') or os.getenv('SOSOCPTOKEN', None):
+            token = self.get_option('token') or os.getenv('SOSOCPTOKEN', None)
```
pmoravec, line 83: Isn't bit faster to re-use the statement as:

    token = self.get_option('token') or os.getenv('SOSOCPTOKEN', None)
    if token:
        oc_res = ..
        ..

sos/report/plugins/openshift.py
```diff
@@ -7,151 +7,369 @@
...
+                    _instance_name = _instance.split()[0]
+                    self.add_cmd_output(
+                        "%s %s -o yaml" % (_get_cmd, _instance_name),
+                        subdir=_subdir,
+                        suggest_filename="%s.yaml" % _instance_name
+                    )
+                # check for podlogs here as a slight optimization to re-running
+                # 'oc get pods' on all namespaces
+                if res == 'pods' and _instances:
+                    if self.get_option('podlogs'):
```
pmoravec, line 320: Doesnt it make sense to merge the two `if` commands into one?

sos/report/plugins/openshift.py
```diff
@@ -7,151 +7,369 @@
...
-                         r"(BIND_KEYVALUE\s*=\s*)(.*)",
-                         r"\1********")
-        # LDAP authentication:  AuthLDAPBindPassword "IShouldNotBeHere"
-        ldap_paths = '/var/www/openshift/(broker|console)/httpd/conf.d/.*'
-        self.do_path_regex_sub(ldap_paths,
-                               r"(AuthLDAPBindPassword)\s*(.*)",
-                               r"\1********")
+
+        # clear any certificate output
+        self.do_cmd_private_sub('oc')
```
pmoravec, line 351: This will remove certificates from not only `oc ..` commands but of much more ones (i.e. if a namespace or pod will contain those two letters) - but I expect that redundant work does not matter. If it does, `s/oc/^oc/`?

sos/report/plugins/openshift.py
TurboTurtle: Our `plugin_timeout` doesn't include `setup()` phase time, so that's 10 minutes of `collect()` running. In my tests the `collect()` phase took 5-6 minutes, so we have some headroom there, but I have no problems increasing it further.

sos/report/plugins/openshift.py
TurboTurtle: Ack.

sos/report/plugins/openshift.py
TurboTurtle: Ack.

sos/report/plugins/openshift.py
TurboTurtle: That's a fair point, we'd most likely end up with the same end result, but `^oc` is more explicit so +1. Will update.

TurboTurtle:
> Can OCP add projects after deployment (and is it a common practice)? If so, shall not we collect the currently existing projects rather than those existing during deployment?

Yes, projects can be added and it is quite frequent. However, the OCP support team has noted they rarely need all of an end user's projects whereas they do frequently need the default ones that allow the cluster to run as expected. If we were to capture all end user projects, we'd have an insanely long run time and end up capturing a lot of data that the engineers won't ever look at. That's why the plugin has the `*-namespaces` options to allow for specific user-defined projects to be collected, while defaulting to just those that directly affect how the cluster runs.

pmoravec:
> > Can OCP add projects after deployment (and is it a common practice)? If so, shall not we collect the currently existing projects rather than those existing during deployment?
> 
> Yes, projects can be added and it is quite frequent. However, the OCP support team has noted they rarely need all of an end user's projects whereas they do frequently need the default ones that allow the cluster to run as expected. If we were to capture all end user projects, we'd have an insanely long run time and end up capturing a lot of data that the engineers won't ever look at. That's why the plugin has the `*-namespaces` options to allow for specific user-defined projects to be collected, while defaulting to just those that directly affect how the cluster runs.

Ack then. Didnt have the background info, a good approach chosen.

sos/report/plugins/openshift.py
TurboTurtle: Actually, this would break the obfuscation since the current method appends `.*` in front of whatever cmd regex is provided. We probably want to review the behavior in depth before changing it just for this plugin.

I'm testing another possibility to attempt to avoid redundant scrubbing.

pmoravec: Hm right.. Maybe a smaller improvement in adding a space after the string? i.e.

    self.do_cmd_private_sub('oc ')

Not ideal but better.

TurboTurtle:
I've update the PR to add a space to the certificate scrubbing, as well as improved scrubbing of secret output and other fields in the yaml output of API resources. I've also scoped the final `do_cmd_output_sub()` more finely on `oc describe`, as originally intended.

PR #2054, opened 2020-05-08
[networking] collect iptables when proper kernel modules loaded

pmoravec:
Since kernel-4, iptables / ip6tables is newly provided by nf_tables
kernel module. Therefore, collecting ip[,6]tables commands should
be gated by presence of also this kernel module.

Resolves: #2054

---


pmoravec:
As the kernel modules might be different for other distros, I would appreciate some review from not-RH distros.

PR #2055, opened 2020-05-08
[containers_common] collect rootless containers info

pmoravec:
Add the ability to collect data/info about rootless podman/buildah
containers, in particular:

- containers_common plugopt 'rootlessusers' as a list of users to inspect
- for each user, collect its containers config and [podman|buildah] info
- collect user-status and few user-related config files
- collect UID and GID mapping in buildah and podman

Resolves: #2055

---

### DIFF
--- a/sos/report/plugins/containers_common.py
+++ b/sos/report/plugins/containers_common.py
@@ -9,6 +9,7 @@
 # See the LICENSE file in the source distribution for further information.
 
 from sos.report.plugins import Plugin, RedHatPlugin, UbuntuPlugin
+import os
 
 
 class ContainersCommon(Plugin, RedHatPlugin, UbuntuPlugin):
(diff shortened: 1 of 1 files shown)


sos/report/plugins/containers_common.py
```diff
@@ -17,11 +18,33 @@ class ContainersCommon(Plugin, RedHatPlugin, UbuntuPlugin):
     plugin_name = 'containers_common'
     profiles = ('container', )
     packages = ('containers-common', )
+    option_list = [
+        ('rootlessusers', 'colon-separated list of users\' containers info',
+         '', ''),
+    ]
```
TurboTurtle, line 24: I think we need to decide on a standardized way to handle lists in plugin options like this. Other recent plugins use a space delimiter, is that possible here?

sos/report/plugins/containers_common.py
```diff
@@ -17,11 +18,33 @@ class ContainersCommon(Plugin, RedHatPlugin, UbuntuPlugin):
...
             '/etc/containers/*',
             '/usr/share/containers/*',
+            '/etc/subuid',
+            '/etc/subgid',
         ])
+        self.add_cmd_output(['loginctl user-status'])
+
+        users_opt = self.get_option('rootlessusers')
+        users_list = []
+        if users_opt and users_opt != '':
```
TurboTurtle, line 37: Isn't this a redundant conditional? `''` should return `False` anyways.

sos/report/plugins/podman.py
```diff
@@ -44,7 +44,9 @@ def setup(self):
             'ps -a',
             'stats --no-stream --all',
             'version',
-            'volume ls'
+            'volume ls',
+            'unshare /proc/self/uid_map',
+            'unshare /proc/self/gid_map',
```
TurboTurtle, line 49: These need to be run _by_ the rootless user to get the intended output:

```
[jake@terra sos]$ podman unshare cat /proc/self/uid_map
         0       1000          1
         1     100000      65536
[jake@terra sos]$ sudo podman unshare cat /proc/self/uid_map
Error: please use unshare with rootless

```

sos/report/plugins/buildah.py
```diff
@@ -25,7 +25,9 @@ def setup(self):
             'containers --all',
             'images',
             'images --all',
-            'version'
+            'version',
+            'unshare cat /proc/self/uid_map',
+            'unshare cat /proc/self/gid_map',
```
TurboTurtle, line 30: Ditto. The `unshare` commands only produce useful info as non-root.

sos/report/plugins/containers_common.py
pmoravec: Space as a separator isnt a problem for me. Just the syntax needs extra quotation marks to denote begin and end of the option value, like:

    -k containers_common.rootlessusers='jsmith bjones'

Colon is used in some plugin as well and does not need the quotes. But either works fine for me.

Sadly, the ideal separator - comma - hits an issue in how we parse the arguments (there is some older open issue on this, /me hit a blocker there, afaik).

sos/report/plugins/containers_common.py
pmoravec: Ah right, too blind copy&paste&modify code :)

sos/report/plugins/podman.py
pmoravec: Ah that was misunderstanding on my end. Wll fix along with other comments.

TomSweeneyRedHat:
FYI @rhatdan @giuseppe @baude @mheon

sos/report/plugins/containers_common.py
TurboTurtle: Followup from IRC: colons good, spaces not as good. Let's try to standardize on `:` delimiters in plugin options going forward.

PR #2058, opened 2020-05-12
[insights] collect insights dump output in deterministic filename

pmoravec:
Currently, "insights-client --offline" stdout is stored in filename like
insights-client_--offline_--output-dir_.var.tmp.sos.cbl0ox16. (and so on)

Let make the filename unified and deterministic.

Also add the trailing vim expand tabs comment.

Resolves: #2058

---


pmoravec:
Failed Travis build https://travis-ci.org/github/sosreport/sos/jobs/686005423#L618 :

    sosreport --batch   --debug  --log-size 0  -c always

that was killed (too much memory?). This can't be relevant to this PR.

PR #2060, opened 2020-05-12
[doc] Update path to the sosreport command.

jcastill:
The documentation references ./sosreport , but
the command now lives inside ./bin, so this
commit updates such references.

Resolves: #2060

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,19 +74,19 @@ pull requests.
 
 You can simply run from the git checkout now:
 ```
-$ sudo ./sosreport -a
+$ sudo ./bin/sos report -a
 ```
-
-Or, if you only have python3 installed:
+The command `sosreport` is still available, as a legacy rediretor,
+and can be used like this:
 ```
-$ sudo python3 ./sosreport -a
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -74,19 +74,19 @@ pull requests.
...
 You can simply run from the git checkout now:
 ```
-$ sudo ./sosreport -a
+$ sudo ./bin/sosreport -a
 ```
 
 Or, if you only have python3 installed:
 ```
-$ sudo python3 ./sosreport -a
+$ sudo python3 ./bin/sosreport -a
```
TurboTurtle, line 82: We've moved to python3-only. Let's converge this and the previous section about running from the checkout.

jcastill:
I'm wondering if instead of ./bin/sosreport -a I should write ./bin/sos report -a , since that's the way we are moving on to... what do you think?

TurboTurtle:
I was mixed on it since we **do** have the redirection executable, but yeah - let's move it over to `./bin/sos`.

PR #2065, opened 2020-05-13
[networking] Small change to produce more-useful, numbered ufw rules

desrod:
This allows inspection of the rule order, as well as insert/edit/delete of the rules by that numbered id (ufw delete 23).

This commit also removes the need for using the absolute path to the 'ufw' binary, since this is asserted by the Ubuntu vendor policy. 

More details can be found here on the [Ubuntu Community page on UFW](https://help.ubuntu.com/community/UFW). 

Signed-off-by: David A. Desrosiers <setuid@gmail.com>

### DIFF
--- a/sos/report/plugins/networking.py
+++ b/sos/report/plugins/networking.py
@@ -314,8 +314,8 @@ class UbuntuNetworking(Networking, UbuntuPlugin, DebianPlugin):
             "/run/systemd/network"
         ])
         self.add_cmd_output([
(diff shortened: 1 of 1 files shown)


BryanQuigley:
Can you confirm that all ufw versions support the number option or fallback gracefully?

slashdd:
Look like 14.04LTS and up has it according to manpage:
http://manpages.ubuntu.com/manpages/trusty/man8/ufw.8.html

Issue #2024, opened 2020-04-16
[process] collect iotop


pmoravec:
How useful do you see collecting `iotop` in `process` plugin, e.g. `iotop -n 1 -b -o` ? I think it can help understanding inefficiency of processes, when they wait on I/O.

Does it make sense to collect it?

TurboTurtle:
I like it, however I think we'd want to run with more than just one iteration since otherwise it seems we'd need to get lucky with IO actually happening the moment we run the command.

pmoravec:
Hm right; how many snapshots over what timespan to collect? `iotop -b -o -n 10 -d 0.1`, i.e. 10 snapshots taken within 1 second? (decimal values in `-d` work at least on my Fedora)

BryanQuigley:
Should this plugin run by itself?  I'd somewhat worry about sosreport being the top result - but testing would prove that out.

TurboTurtle:
I was more referring to running for longer than 1 second. We should reach out to groups that are most likely to use this output and get a sense for how long a duration, and how many intervals in that duration, would be best suited for real world use cases.

@BryanQuigley We'd probably see child processes from sos no matter where/when it runs (since we run plugins in parallel, it'd be really hard to isolate this particular collection), but it's definitely something to keep an eye out for in testing.

bu3ny:
I think SOS report usually takes between 1-3 minutes(sometimes longer) based on the collection process. Starting iotop and keep it running for around 1 minute with 5 seconds interval should reveal some info that may help.

pmoravec:
> I think SOS report usually takes between 1-3 minutes(sometimes longer) based on the collection process. Starting iotop and keep it running for around 1 minute with 5 seconds interval should reveal some info that may help.

You should not compare running time of whole `sos report` with one command execution :) Recall that `sos report` (if we ignore some init and finalize phases) runs at most 4 (by default) plugins concurrently "only", and each plugin - independently on others - executes its commands sequentially. So adding say one minute long command to `process` plugin means overall time execution would grow by 30 seconds (on average, if my probabilistic math is correct).

Having commands lasting such longer time can be valuable, sometimes, but I think it is not the case of `iotop` where we should target its running time to seconds, I think.

bu3ny:
[...]

PR #2068, opened 2020-05-15
[process] Collect iotop

jcastill:
Collect the output of the 'iotop' command
by specifying the number of samples in
the command line with the option 'samples',
with 0.5 seconds delay between each one.
By default we capture 20 samples.

Resolves: #2024

---

### DIFF
--- a/sos/report/plugins/process.py
+++ b/sos/report/plugins/process.py
@@ -21,7 +21,9 @@ class Process(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin, CosPlugin):
         ("lsof", "gathers information on all open files", "slow", True),
         ("lsof-threads", "gathers threads' open file info if supported",
          "slow", False),
-        ("smaps", "gathers all /proc/*/smaps files", "", False)
+        ("smaps", "gathers all /proc/*/smaps files", "", False),
+        ("samples", "specify the number of samples that iotop will capture, "
+            "with an interval of 0.5 seconds between samples", "", "20")
     ]
 
     def setup(self):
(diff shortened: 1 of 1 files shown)


sos/report/plugins/process.py
```diff
@@ -21,7 +21,9 @@ class Process(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin, CosPlugin):
         ("lsof", "gathers information on all open files", "slow", True),
         ("lsof-threads", "gathers threads' open file info if supported",
          "slow", False),
-        ("smaps", "gathers all /proc/*/smaps files", "", False)
+        ("smaps", "gathers all /proc/*/smaps files", "", False),
+        ("iotop", "gathers information from iotop, by default 20 "
+            "snapshots within 0.5 seconds", "slow", True)
```
TurboTurtle, line 26: I think rather than toggling the run of `iotop`, we should be allowing users to control the number of samples to take, e.g. `sos report -k process.samples=50` or `sos report -k process.samples=0` the latter of which effectively gracefully disables the command anyways:

```
[jake@terra ~]$ sudo iotop -b -o -n 0 -d 0.5 -t
[jake@terra ~]$ 

```

jcastill: Change done, now it captures by default 20 samples but you can modify the number in the command line with -k process.samples. I've expanded the commit message as well to reflect this, I hope is the right way to do it.

sos/report/plugins/process.py
```diff
@@ -21,7 +21,9 @@ class Process(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin, CosPlugin):
         ("lsof", "gathers information on all open files", "slow", True),
         ("lsof-threads", "gathers threads' open file info if supported",
          "slow", False),
-        ("smaps", "gathers all /proc/*/smaps files", "", False)
+        ("smaps", "gathers all /proc/*/smaps files", "", False),
+        ("samples", "specify the number of samples that iotop will capture, "
+            "with an interval of 0.5 seconds between samples", "", "")
```
pmoravec, line 26: Can't you provide the `20` as the default value directly here? (the latest argument/item in the plugopt list).

jcastill: You are right, I oversaw that possibility. Let me resend with this change

TurboTurtle:
Merged in fc71d8b

PR #2069, opened 2020-05-15
[abrt] Add /var/tmp/abrt into the collection

dougsland:
Files generated under /var/tmp/abrt should be collected
as well.

Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=1644646

---

### DIFF
--- a/sos/report/plugins/abrt.py
+++ b/sos/report/plugins/abrt.py
@@ -36,7 +36,8 @@ class Abrt(Plugin, RedHatPlugin):
         self.add_copy_spec([
             "/etc/abrt/abrt.conf",
             "/etc/abrt/abrt-action-save-package-data.conf",
-            "/etc/abrt/plugins"
+            "/etc/abrt/plugins",
+            "/var/tmp/abrt"
         ])
 
 # vim: set et ts=4 sw=4 :


dougsland:
@TurboTurtle hey Jake, could you please take a look? :)

pmoravec:
Can't be there some customer sensitive data? Isnt coredump being generated there? That would be impossible to scrub secret info from a coredump.. Or from more similar files that `abrt` report contains..

bmr-cymru:
Something a bit off here... The bugzilla is against `ovirt-log-collector`:

```
Running ovirt-log-collector fails on many errors of non existing files in abrt directory.
e.g.

/bin/ls: cannot open directory /var/tmp/abrt/Python-2018-10-30-16:58:35-12807/sos.UensHr/sosreport-lynx10-2018-10-30-mfcvaas/var: No such file or directory
```

And it appears to be a case of something in the log collector racing against an `abrt` initiated run of sosreport - we've had similar problems with `sosreport` on a single node (where `abrt` incorrectly removes the `sos.XXXX` directory before terminating the `sosreport` process).

It's not a request to add collection of `/var/tmp/abrt`, and it's not even clear it's documenting an actual bug in `sos report` or `sos collect`.

pmoravec:
I *assume* there is some legitimate rationale to collect data from `/var/tmp/abrt` dir, but I agree that:

- the underlying bug is not a `sos` issue
- we lack the explanation why `sosreport` should collect that data - having the data, what kind of issues it helps in investigation?

bmr-cymru:
Collecting stuff from `/var/tmp` counts as an extraordinary request - those require extraordinary justification. :-D

There's no explanation of why adding collection of this directory to `sos report` will stop `ovirt-log-collector` failing when it tries to list some subdirectory of this path on a remote system - this looks like a misunderstanding of the bug to me.

dougsland:
Thanks for the comments, we will handle in the ovirt-log-collector-side.

Issue #2061, opened 2020-05-13
[nfsserver] improve check_enabled method relying on runlevels


pmoravec:
The plugin check in https://github.com/sosreport/sos/blob/master/sos/report/plugins/nfsserver.py#L23-L24 the policy methods `default_runlevel` (https://github.com/sosreport/sos/blob/master/sos/policies/__init__.py#L1070) and `runlevel_by_service` (https://github.com/sosreport/sos/blob/master/sos/policies/redhat.py#L153) that rely on initd runlevels. That is not applicable since RHEL7 / systemd.

We should come with some better `check_enabled` method.

PR #2073, opened 2020-05-18
[nfs] merge nfsserver plugin into nfs one

pmoravec:
nfsserver plugin enabledness relies on legacy init scripts. A more
appropriate way to idenfity a NFS server is via nfs-utils package
(though it could be present also on a NFS client).

As that package enables nfs plugin, it is reasonable to merge the
plugins into one.

Resolves: #2061

---


pmoravec:
Cc @pierg75 for review and @BryanQuigley for Nfs plugin extension on Ubuntu/Debian - wont these extra files/dirs and commands harm?

pierg75:
Looks good to me. Thanks!

sos/report/plugins/nfs.py
```diff
@@ -14,7 +14,7 @@ class Nfs(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
     short_desc = 'Network file system information'
     plugin_name = 'nfs'
     profiles = ('storage', 'network', 'nfs')
-    packages = ['nfs-utils']
+    packages = ('nfs-utils')
```
TurboTurtle, line 17: Needs to be a tuple

pmoravec:
Ah, I see, fixed.

PR #2079, opened 2020-05-20
[xdp] Add XDP plugin

tohojo:
This adds a plugin for gathering information about the loaded XDP programs
on a system, using the xdp-loader tool in the xdp-tools package. This is
complementary with the existing eBPF plugin, because this plugin supports
getting status for multiple XDP programs on each interface as supported by
libxdp.

For now, this just prints the output of 'xdp-loader status' which will list
any XDP programs attached. We don't currently support structured
output (such as JSON) from xdp-tools, but should we add that in the future
we can of course update this plugin.

---

### DIFF
--- a/sos/report/plugins/xdp.py
+++ b/sos/report/plugins/xdp.py
@@ -1,0 +1,21 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, RedHatPlugin, DebianPlugin, \
+    UbuntuPlugin
+
+
+class Xdp(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
+
+    short_desc = 'XDP program information'
+    plugin_name = 'xdp'
+    profiles = ('system', 'kernel', 'network')
+    packages = ('xdp-tools',)
+
+    def setup(self):
+        self.add_cmd_output('xdp-loader status')


sos/report/plugins/xdp.py
```diff
@@ -0,0 +1,20 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
```
TurboTurtle, line 9: Looks like Travis didn't run, but this seems to be over 79 characters, so we'll need to wrap it for PEP8.

tohojo:
Jake Hunsaker <notifications@github.com> writes:

> @TurboTurtle commented on this pull request.
>
>
>
>> @@ -0,0 +1,20 @@
> +# This file is part of the sos project: https://github.com/sosreport/sos
> +#
> +# This copyrighted material is made available to anyone wishing to use,
> +# modify, copy, or redistribute it subject to the terms and conditions of
> +# version 2 of the GNU General Public License.
> +#
> +# See the LICENSE file in the source distribution for further information.
> +
> +from sos.report.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
>
> Looks like Travis didn't run, but this seems to be over 79 characters,
> so we'll need to wrap it for PEP8.

Ah, right. Just copied that from the ebpf plugin, but sure, can do :)

sos/report/plugins/xdp.py
bmr-cymru: Perhaps it's time we added an `__all__` member to `sos.report.plugins`? It's not everyone's stylistic cup-of-tea but I use it heavily in other projects that are designed to use `*` imports and shared namespaces.

TurboTurtle:
Merged in af5597b

Issue #2003, opened 2020-04-07
invalid default value for plugin timeout when --list-plugins used


pmoravec:
```
/usr/bin/python3 ./sosreport --list-plugins 
..
The following options are available for ALL plugins:
 timeout                   -1              Timeout in seconds for plugin
 postproc                  True            Enable post-processing collected plugin data
..
```

While the default plugin timeout is 300. `postproc=True` is the correct default, though.

Found during checking #1993 where the same issue appears. Since it is on master as well, logging as a separate issue.

PR #2084, opened 2020-05-22
[Plugin] Fix incorrect default timeout

jcastill:
Timeout should be 300 seconds, but was
set up as '-1' in the option list.

Closes: #2003
Resolves: #2084

---

### DIFF
--- a/man/en/sos-report.1
+++ b/man/en/sos-report.1
@@ -226,7 +226,8 @@ Specify the number of threads sosreport will use for concurrency. Defaults to 4.
 .TP
 .B \--plugin-timeout TIMEOUT
 Specify a timeout in seconds to allow each plugin to run for. A value of 0
-means no timeout will be set.
+means no timeout will be set. A value of -1 is used to indicate the default
+timeout of 300 seconds.
 
 Note that this options sets the timeout for all plugins. If you want to set
 a timeout for a specific plugin, use the 'timeout' plugin option available to
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -415,7 +415,8 @@ class Plugin(object):
     predicate = None
     cmd_predicate = None
     _default_plug_opts = [
-        ('timeout', 'Timeout in seconds for plugin', 'fast', -1),
+        ('timeout', 'Timeout in seconds for plugin. The default value (-1) ' +
+            'defers to the general plugin timeout, 300 seconds', 'fast', -1),
         ('postproc', 'Enable post-processing collected plugin data', 'fast',
          True)
     ]


jcastill:
Ah ok, didn't know about that, sorry! 
What does -1 mean in the context of timeout, no timeout at all?

bmr-cymru:
It's a magic value used to indicate that the timeout is unset - the problem that needs addressing is that it currently leaks out to the UI via `--list-plugins`:

```
The following options are available for ALL plugins:
 timeout                   -1              Timeout in seconds for plugin
 postproc                  True            Enable post-processing collected plugin data
```

We could paper over this with a hack in `SoSReport.list_plugins()` (detect that we are printing the `timeout` option and use some other display value) but it would be nice to come up with a scheme that did not require that.

jcastill:
What about changing -1 to "Unset" inside _default_plug_opts, and then equal "Unset" to "-1" inside **set_option** ? 
This way we present "Unset" to the user, but internally we make the change when we actually set the option. 
Another way I can think to deal with this is to explain the -1 in the option itself, i.e.:

('timeout', 'Timeout in seconds for plugin. The default value of -1 means it's unset, and ...', 'fast', "-1"),

TurboTurtle:
I think `unset` and `-1` are equally unintuitive to any end user who isn't a sos contributor unfortunately, or may even suggest there is *no* timeout for plugins by default.

Perhaps this would be best covered by both an update to the option description ('Use -1 for individual plugin default'?) and then a more thorough explanation in the man page.

jcastill:
> Perhaps this would be best covered by both an update to the option description ('Use -1 for individual plugin default'?) and then a more thorough explanation in the man page.

I think this may be the best solution, so we clarify the situation and address the original issue #, while at the same time can have more time to think and test other ways of not exposing the -1. 
Shall I use this PR for the modification of the option description and the man page, or should I open a new one and close this?

pmoravec:
Reusing this PR number (though rewriting the changes completely from the scratch and forcefully pushing to your branch) is fine.

man/en/sos-report.1
```diff
@@ -226,7 +226,8 @@ Specify the number of threads sosreport will use for concurrency. Defaults to 4.
 .TP
 .B \--plugin-timeout TIMEOUT
 Specify a timeout in seconds to allow each plugin to run for. A value of 0
-means no timeout will be set.
+means no timeout will be set. The default value of -1 is a magic value used
+to indicate that the timeout is unset.
```
TurboTurtle, line 230: Let's make this align with the verbiage in the option description - i.e. "A value of -1 is used to indicate the default timeout of 300 seconds".

sos/report/plugins/__init__.py
```diff
@@ -415,7 +415,8 @@ class Plugin(object):
     predicate = None
     cmd_predicate = None
     _default_plug_opts = [
-        ('timeout', 'Timeout in seconds for plugin', 'fast', -1),
+        ('timeout', 'Timeout in seconds for plugin. The default value (-1) ' +
+            'defers to the general plugin timeout, 300 seconds', 'fast', "-1"),
```
TurboTurtle, line 419: The last tuple element here would convert this from an `int` to a `str`, which I think would break things down the line.

TurboTurtle (commented):
Sorry for the long delay here, I hadn't noticed the updated branch before now.

Comments below.

TurboTurtle (approved):

pmoravec (approved):

jcastill:
Closing this as I removed the branch by mistake. New PR created here:

https://github.com/sosreport/sos/pull/2415

Issue #2080, opened 2020-05-20
[cloud_init] collect information for reporting bugs properly


slashdd:
Maybe we can integrate some of the below in sosreport [cloud_init].
Reporting this bug for me to investigate in the next few days/weeks.

https://cloudinit.readthedocs.io/en/latest/topics/bugs.html

Reporting Bugs
The following documents:

How to collect information for reporting bugs
How to file bugs to the upstream cloud-init project or for distro specific packages
Collect Logs
To aid in debugging, please collect the necessary logs. To do so, run the collect-logs subcommand to produce a tarfile that you can easily upload:

$ cloud-init collect-logs
Wrote /home/ubuntu/cloud-init.tar.gz
If your version of cloud-init does not have the collect-logs subcommand, then please manually collect the base log files by doing the following:

$ dmesg > dmesg.txt
$ sudo journalctl -o short-precise > journal.txt
$ sudo tar -cvf cloud-init.tar dmesg.txt journal.txt /run/cloud-init \
    /var/log/cloud-init.log /var/log/cloud-init-output.log

slashdd:
After investigation `cloud-init collect-logs` create a tarball as follows:

```
cloud-init collect-logs 
Wrote /tmp/cloud-init/cloud-init.tar.gz
```
I tried to dissect what it does so I could adapt to its current plugin but I quickly came to the conclusion that it may reveal sensible informations such as:

`/run/cloud/instance-data-sensitive.json`
```
/var/lib/cloud/instance/user-data.txt

  --include-userdata, -u
                        Optionally include user-data from
                        /var/lib/cloud/instance/user-data.txt which could
                        contain sensitive information.
```
On that note, the actual plugin collect that sensible data information file (implicitly) https://github.com/sosreport/sos/blob/master/sos/report/plugins/cloud_init.py#L24

Unless we have a good obfuscation method in place for that file, I think we should think to remove it.
As you can see `cloud-init collect-logs` default behaviour is to ignore it unless one turn the Boolean to true to include user-data.

And I don't see the value of copy the content of /var/lib/cloud if` cloud-init collect-logs` itself doesn't do it but for 1 optionnal file: https://github.com/canonical/cloud-init/blob/master/cloudinit/cmd/devel/logs.py#L21
and only if the `include_userdata` is turned on

Issue #1681, opened 2019-05-17
sosreport does not collect all cloud-init logs


shoshin71:
sosreport currently does not collect cloud-init logs from /var/lib/cloud, and those have to be collected separately via "cloud-init collect-logs".

On systems where cloud-init is present, I would think that sosreport should collect those logs.

TurboTurtle:
#1630 is adding a cloud-init plugin, which captures from `/var/lib/cloud` as well as a few other locations.

What does `cloud-init collect-logs` generate? Is there anything aside from what's included in the PR we should grab in sos?

shoshin71:
It looks like cloud-init collect-logs collects all of /run/cloud-init, whereas #1630 only collects /run/cloud-init/cloud-init-generator.log. #1630 does collect all of /var/lib/cloud.

TurboTurtle:
I'll admit I'm not super well versed in `cloud-init` - what else ends up under `/run/cloud-init` that we should grab? Is there the potential for sensitive data in there? If not, we can probably just capture the whole thing. If there is, we'd need to obfuscate it in postproc.

shoshin71:
I am still investigating this. It looks like the #1630 may need to be revised, as it collects all of /var/lib/cloud. I have discovered that /var/lib/cloud/instances/.../user-data.txt contains sensitive information that should probably not be shared with an external support provider.

ssh_keys: {dsa_private: '-----BEGIN DSA PRIVATE KEY-----

', ecdsa_private: '-----BEGIN EC PRIVATE KEY-----

', ed25519_private: '-----BEGIN OPENSSH PRIVATE KEY-----

and so forth. I am trying to determine what it might be acceptable/safe to collect.

TurboTurtle:
We can scrub keys and certs and the like, that's no issue. We just need to know the relevant paths. Would it just be `/var/lib/cloud/instances/*/user-data.txt` (or maybe a broader `/var/lib/cloud/*user-data*`?)

BryanQuigley:
I think this has been fixed in #1630, obviously a new Issue should be created if we find we're collecting too much info.

PR #2086, opened 2020-05-25
[cloud_init] Update the plugin

slashdd:
* Snap can found be found as a DEB or SNAP

* Dropping /var/lib/cloud as it may reveal
sensible data

* Integrate commands inspired by
cloud-init collect-logs

Closes: #2080

---

### DIFF
--- a/sos/report/plugins/cloud_init.py
+++ b/sos/report/plugins/cloud_init.py
@@ -18,10 +18,12 @@ class CloudInit(Plugin, RedHatPlugin, UbuntuPlugin, DebianPlugin):
     plugin_name = 'cloud_init'
     packages = ('cloud-init',)
     services = ('cloud-init',)
+    commands = ('cloud-init',)
 
     def setup(self):
+        self.add_cmd_output('cloud-init --version')
+
         self.add_copy_spec([
-            '/var/lib/cloud/',
             '/etc/cloud/',
             '/run/cloud-init/cloud-init-generator.log',
             '/var/log/cloud-init*'


sos/report/plugins/cloud_init.py
```diff
@@ -18,10 +18,15 @@ class CloudInit(Plugin, RedHatPlugin, UbuntuPlugin, DebianPlugin):
     plugin_name = 'cloud_init'
     packages = ('cloud-init',)
     services = ('cloud-init',)
+    commands = ('cloud-init',)
 
     def setup(self):
+        self.add_cmd_output([
+            'cloud-init --version',
+            'journalctl --boot=0 -o short-precise'
```
TurboTurtle, line 26: I'm not sure a full `journal` collection makes sense here - shouldn't this be limited to the `cloud-init` unit?

BryanQuigley:
what is an example of the sensitive data that /var/lib/cloud was capturing and can we just filter specific bad bits?

slashdd:
> what is an example of the sensitive data that /var/lib/cloud was capturing and can we just filter specific bad bits?
```

/var/lib/cloud/instance/user-data.txt

  --include-userdata, -u
                        Optionally include user-data from
                        /var/lib/cloud/instance/user-data.txt which could
                        contain sensitive information.
```

Additionnally, I notice cloud-init collect-logs doesn't even collect `/var/lib/cloud `anymore.

BryanQuigley:
Collecting /var/lib/cloud was requested rather recently - in #1681.    Need to dig deeper into the data  there.

BryanQuigley:
There is also /run/cloud-init/instance-data-sensitive.json,  which on my install is hte same as the non-sensitive version. 
And these partially completed docs..  https://cloudinit.readthedocs.io/en/latest/topics/dir_layout.html

The user-data is the cloud config.. which could have a password in it, but is also really helpful to see IMHO.  Not sure the best way to address.

slashdd:
We could use `do_file_sub()` for the password found in `user-data.txt`

For the entire contents of `/var/lib/cloud` even cloud-init `cloudinit/cmd/devel/logs.py` doesn't collect it which is the upstream way to collect data for debug purposes.

It only focuses on `/var/lib/cloud/instance/user-data.txt` iff `--include-userdata` is used.

PR #2093, opened 2020-05-28
[sos] Add soscleaner functionality natively into sos

TurboTurtle:
This patchset adds functionality similar to `soscleaner` directly into `sos`. This can be run separately as `sos clean|mask`, or by using the `--clean` or `--mask` option when running either `report` or `collect`.

This initial PR should be seen as a WIP, and it is expected to go through a handful of changes based on feedback from maintainers and contributors. I.E. if this design is not suitable, it can be fully changed based on that feedback.

At this time there is support for IPv4 IP addresses, IPv4 and IPv6 MAC addresses, and hostnames based on the hostname of the system that generated the sosreport as well as any domains specified by the user.

This component uses what are known as `maps` and `parsers` to identify and obfuscate sensitive information. In short, `parsers`.... parse files line by line looking for regex matches for whatever it is that parser is intended to obfuscate. Matches are sent to the `maps` for obfuscation (and further validation that the parser hit a correct match). Maps handle the actual obfuscation of the data in question, as well as maintain the store (or 'mapping') or already obfuscated items to allow for consistency.

Currently, this component is explicitly limited to looking at sosreports, and if we are not sure that the given file is either A) an sosreport B) a tarball of sosreports (e.g. `sos collect`) or C) the build directory of an sosreport then we will not process that target.

---


TurboTurtle:
Some things currently missing that need to be added, assuming this general design is acceptable:

- Username obfuscation
- User provided keyword obfuscation (the disclaimer banner currently includes that, but we won't be merging this into master without either adding this obfuscation or removing that bit from the disclaimer)
- IPv6 IP addresses

TurboTurtle:
Hmm, interesting. I don't hit the import error locally that Travis is getting during `nosetests`. I'll try and play around with that to see if I can reproduce it at all.

TurboTurtle:
Ah, ok - so the 3.5 nosetest is failing because on python-3.5 it's `random.choice()` and 3.6+ it's `random.choices()`. Yay.

@BryanQuigley @slashdd are there currently supported versions of Ubuntu out there that ship 3.5? We do a compatibility workaround quite easily, but I remember the discussion of dropping pre-3.6 support came up when we dropped python2 earlier so perhaps this is a good time to look at that again?

sos/cleaner/mappings/ip_map.py
```diff
@@ -0,0 +1,189 @@
...
+    Internally, the ipaddress library is used to manipulate the address objects
+    however, when retrieved by SoSCleaner any values will be strings.
+    """
+
+    ignore_matches = [
+        '127.*',
+        '::1',
+        '0.0.0.0(.*)?',
+        '169.254.*',
+        '255.*'
```
BryanQuigley, line 38: Got some false positives that are 0.*  

Thoughts on adding multicast addresses to the list too?   224.0.0.0/8–239.0.0.0/8 ?  Alternatively we could just stop filtering at 224+ as any above are either multicast or reserved for future use.

sos/cleaner/mappings/mac_map.py
```diff
@@ -0,0 +1,78 @@
...
+
+from sos.cleaner.mappings import SoSMap
+
+
+class SoSMacMap(SoSMap):
+    """Mapping store for MAC addresses
+
+    MAC addresses added to this map will be broken into two halves, vendor and
+    device like how MAC addresses are normally crafted. For the vendor hextets,
+    obfuscation will take the form of 53:4f:53, or 'SOS' in hex. The following
```
BryanQuigley, line 22: >53:4f:53, or 'SOS'
That's a nice touch :)

I was hoping to come up with a proposal to not filter the first 3, but the list (https://gist.github.com/aallan/b4bb86db86079509e6159810ae9bd3e4) is really long and defense orgs have claimed some of them... so oh well.

BryanQuigley:
I swear we weren't collecting as much in /proc and /sys which is slowing my initial pass a bit. (will investigate tomorrow)

--verbose doesn't work in --mask, I'd definitely use it to determine where some of my values have come from (and I think users might appreciate it as well).

sos/cleaner/mappings/mac_map.py
BryanQuigley: I think we may want to not filter 01:00:5e* or 33:33:00*  which from a quick search appear to be multicast macs.

sos/cleaner/__init__.py
```diff
@@ -0,0 +1,552 @@
...
+        """
+        if not filename:
+            # the requested file doesn't exist in the archive
+            return
+        self.log_debug("Obfuscating %s" % filename)
+        subs = 0
+        tfile = tempfile.NamedTemporaryFile(mode='w', dir=self.tmpdir)
+        with open(filename, 'r') as fname:
+            for line in fname:
+                if not line.strip() or line.startswith('#'):
```
pmoravec, line 514: A detail so far, but IMHO comment lines needs to be obfuscated as well. Assume configs like:

    connectTo: hostname.domain
    # connectTo: another.host.domain

sos/cleaner/__init__.py
```diff
@@ -0,0 +1,552 @@
...
+            :param filename str:    Filename the line was read from
+
+        Returns the fully obfuscated line and the number of substitutions made
+        """
+        count = 0
+        for parser in self.parsers:
+            if filename and any([
+                re.match(_s, filename) for _s in parser.skip_files
+            ]):
+                continue
```
pmoravec, line 546: (When) will be `skip_files` used? (so far this is unused) Is it like for skipping binary files, or some other "safe" files that we know they dont collect any sensitive data (but an attempt to obfuscate might harm)?

sos/cleaner/mappings/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+        loopback addresses should not be obfuscated
+        """
+        for skip in self.ignore_matches:
+            if re.match(skip, item):
+                return True
+
+    def item_in_dataset_values(self, item):
+        if item in self.dataset.values():
+            return True
+        return False
```
pmoravec, line 42: Just `return (item in self.dataset.values())` ?

sos/cleaner/mappings/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+
+            :param item:        The plaintext object to obfuscate
+        """
+        return item
+
+    def get(self, item):
+        """Retrieve an item's obfuscated counterpart from the map. If the item
+        does not yet exist in the map, add it by generating one on the fly
+        """
+        if self.ignore_item(item) or self.item_in_dataset_values(item):
```
pmoravec, line 71: Why the `or self.item_in_dataset_values(item)`? Doesn't this return the keys instead of mapped values from `self.dataset`, then?

sos/cleaner/mappings/hostname_map.py
```diff
@@ -0,0 +1,120 @@
...
+    and 'host.foo.example.com' will be separately obfuscated with no relation
+    for example as 'obfuscatedomdain1.com' and 'obfuscatedomain2.com'.
+
+    Top-level domains are left untouched.
+    """
+
+    ignore_matches = [
+        'localhost',
+        '.*localdomain.*',
+        'com..*'
```
pmoravec, line 35: Doesnt this ignore e.g. `www.mobil-com.cz` ?

sos/cleaner/mappings/hostname_map.py
```diff
@@ -0,0 +1,120 @@
...
+                if known_domain in domain:
+                    return True
+        return False
+
+    def get(self, item):
+        if item.startswith(('.', '_')):
+            item = item.lstrip('._')
+        item = item.strip()
+        if not self.domain_name_in_loaded_domains(item):
+            return item
```
pmoravec, line 70: I dont understand here: why do we return `item` (i.e. the not-obfuscated key) when the `item` was evaluated as potential domain we should obfuscated (cf. description of `domain_name_in_loaded_domains`)?

sos/cleaner/mappings/ip_map.py
```diff
@@ -0,0 +1,189 @@
...
+            # otherwise within that obfuscated network grab the next available
+            # address from it
+            for _ip in _obf_network.hosts():
+                if not self.ip_in_dataset(_ip):
+                    # the ipaddress module does not assign the network's
+                    # netmask to hosts in the hosts() generator for some reason
+                    return "%s/%s" % (str(_ip), _obf_network.prefixlen)
+
+        # ip is a single ip address without the netmask
+        return self._new_obfuscated_single_address()
```
pmoravec, line 150: Can't we generate the same obfuscated IP address for different input addresses, here?

Well, the probability is small, as we are picking from a range of 65M (90^4) addresses. But if my math is correct, then e.g. there is 50% chance of hitting already picked address when obfuscating 9538 different IP addresses (or 10% chance for 3719 addresses).

sos/cleaner/obfuscation_archive.py
```diff
@@ -0,0 +1,222 @@
...
+            for filename in files:
+                self.file_list.append(os.path.join(dirname, filename))
+        return self.file_list
+
+    def update_sub_count(self, fname, count):
+        """Called when a file has finished being parsed and used to track
+        total substitutions made and number of files that had changes made
+        """
+        self.file_sub_list.append(fname)
+        self.total_sub_count += count
```
pmoravec, line 195: `total_sub_count` is currently never used. But makes sense to track stats, for sure.

sos/cleaner/parsers/hostname_parser.py
```diff
@@ -0,0 +1,40 @@
...
+from sos.cleaner.mappings.hostname_map import SoSHostnameMap
+
+
+class SoSHostnameParser(SoSCleanerParser):
+
+    name = 'Hostname Parser'
+    map_file_key = 'hostname_map'
+    prep_map_file = 'sos_commands/host/hostname'
+    regex_patterns = [
+        r'(((\b|_)[a-zA-Z0-9-\.]{1,200}\.[a-zA-Z]{1,63}\b))'
```
pmoravec, line 21: I think we can be more precise here (cf. RFC 1123 and 952, or https://stackoverflow.com/questions/106179/regular-expression-to-match-dns-hostname-or-ip-address/106223#106223 or https://stackoverflow.com/questions/11809631/fully-qualified-domain-name-validation/20204811#20204811), but for now it's good.

pmoravec:
In general, I like the design. Some particular minor comments:

- `--domains` requires an extra argument? Do we expect it is often used? (I dont have a user experience with sos-cleaner, just curious if one specific pattern desires an extra cmdline argument)

- Methods shall be re-ordered to declare auxiliary ones first - this would helped me for readability of code.

- Should not we obfuscate also filenames (e.g. even plain `sosreport-<hostname>-<timestamp>..` or others inside the report, like cmdline based filenames (e.g. https://github.com/sosreport/sos/blob/master/sos/report/plugins/pulp.py#L128 generates a filename containing `self.messaging_cert_file` that does contain FQDN of the system; similarly some remote DBs hostnames or IPs can occur in other cmdline based filenames))

- `generate_checksum`: shall not be a common routine also for sosreport, inside `archive` class?

sos/cleaner/parsers/hostname_parser.py
```diff
@@ -0,0 +1,40 @@
...
+        """Force add the domainname found in /sos_commands/host/hostname into
+        the map. We have to do this here since the normal map prep approach
+        from the parser would be ignored since the system's hostname is not
+        guaranteed
+        """
+        domains = hostname_string.split('.')
+        if len(domains) > 3:
+            # make sure we implicitly get example.com if the system's hostname
+            # is something like foo.bar.example.com
+            high_domain = '.'.join(domain[-2:])
```
pmoravec, line 38: `s/domain/domains/` in the `join` (currently it raises `name 'domain' is not defined`)

pmoravec:
We should further update `Makefile` to ensure `No such file or directory: '/etc/sos/cleaner/default_mapping'` does not happen on a system with installed `sos`.

sos/cleaner/__init__.py
TurboTurtle: In my earlier versions, one of the parsers used this however I found it unneeded with another change, however I kept this in here to allow future parsers to easily identify files they need to skip individually.

sos/cleaner/mappings/__init__.py
TurboTurtle: Ack

sos/cleaner/mappings/hostname_map.py
TurboTurtle: Good point, I'll change that to `^com.`. Although, the other consideration is to not skip this (e.g. com.redhat.access), but I couldn't find a good way to balance that with already-obfuscated domains.

sos/cleaner/mappings/ip_map.py
TurboTurtle: Easy enough to add a check here, but yeah my thought initially was that it was an incredibly probability.

sos/cleaner/obfuscation_archive.py
TurboTurtle: Right, prepping this for metadata inclusion.

sos/cleaner/parsers/hostname_parser.py
TurboTurtle: Ack

sos/cleaner/__init__.py
TurboTurtle: > * `--domains` requires an extra argument? Do we expect it is often used? (I dont have a user experience with sos-cleaner, just curious if one specific pattern desires an extra cmdline argument)

--domains allows a user to specify domain names that don't fall under the system's FQDN, which is all we obfuscate by default - there's simply too many false positives generated from any regex for FQDNs to be able to cover all found domains without disrupting a large amount of other data.

> * Methods shall be re-ordered to declare auxiliary ones first - this would helped me for readability of code.

Sure

> * Should not we obfuscate also filenames (e.g. even plain `sosreport-<hostname>-<timestamp>..` or others inside the report, like cmdline based filenames (e.g. https://github.com/sosreport/sos/blob/master/sos/report/plugins/pulp.py#L128 generates a filename containing `self.messaging_cert_file` that does contain FQDN of the system; similarly some remote DBs hostnames or IPs can occur in other cmdline based filenames))

That shouldn't be too difficult, we'd just need to add an `obfuscate_string()` method to handle non-file-content with the parsers. Not sure about the archive name though...

> * `generate_checksum`: shall not be a common routine also for sosreport, inside `archive` class?

 I can rename it, sure

TurboTurtle:
> * `--domains` requires an extra argument? Do we expect it is often used? (I dont have a user experience with sos-cleaner, just curious if one specific pattern desires an extra cmdline argument)

--domains allows a user to specify domain names that don't fall under the system's FQDN, which is all we obfuscate by default - there's simply too many false positives generated from any regex for FQDNs to be able to cover all found domains without disrupting a large amount of other data.

> * Methods shall be re-ordered to declare auxiliary ones first - this would helped me for readability of code.

Sure

> * Should not we obfuscate also filenames (e.g. even plain `sosreport-<hostname>-<timestamp>..` or others inside the report, like cmdline based filenames (e.g. https://github.com/sosreport/sos/blob/master/sos/report/plugins/pulp.py#L128 generates a filename containing `self.messaging_cert_file` that does contain FQDN of the system; similarly some remote DBs hostnames or IPs can occur in other cmdline based filenames))

That shouldn't be too difficult, we'd just need to add an `obfuscate_string()` method to handle non-file-content with the parsers. Not sure about the archive name though...

> * `generate_checksum`: shall not be a common routine also for sosreport, inside `archive` class?

 I can rename it, sure

pmoravec:
Worth extending #2050 or this PR by # of objects obfuscated (e.g. per parser)? Or start+end+running time of cleaner (that would be interesting to have)?

Surely the extension makes sense after either PR is merged.

sos/cleaner/mappings/ip_map.py
TurboTurtle: What's the reasoning behind not filtering multicast? I'm sure users in secure environment would consider those network addresses still sensitive.

TurboTurtle:
> --verbose doesn't work in --mask, I'd definitely use it to determine where some of my values have come from (and I think users might appreciate it as well).

A single `-v` doesn't, no. Use of 2 or more, `-vv`, does. This is because our console log only sets debug level logs when the verbose option count > 1:

```
            if self.opts.verbosity and self.opts.verbosity > 1:
                console.setLevel(logging.DEBUG)
                if flog:
                    flog.setLevel(logging.DEBUG)
```

sos/cleaner/__init__.py
TurboTurtle: Hmm, actually on second thought let's discuss this one a bit further.

Comments in config files has the potential to throw off the used subnet relationships in sosreports.

For example - suppose the system has an actually used subnet of 10.11.12.0/24, with a bunch of addresses in that subnet in use. Now suppose a config file has a line of `# for example 10.11.12.0/26` or a similarly styled smaller subnet than what is actually used on the system.

Now any addresses in the /24 subnet on the system that would mathematically fall into the /26 subnet will now be obfuscated to the network created for /26 network from the comment.

We do already have some form of this problem with some /proc files (`/proc/pid/net/fib_trie`), but this would exacerbate the issue.

pmoravec: I see and agree this is a concern we should handle. But:

- how often could configs contain such subnet comment that cause such problems = compared to comments with a valid hostname / IP we should obfuscate? IMHO the "not cleaning comments" is a bigger issue.

- what if the `10.11.12.0/26` really occurs in a non-commented config? :)

TurboTurtle:
I've pushed an updated branch that:

* fixed an issue with domain name loading when `--clean` was called from `report` rather than via standalone
* added an ability on a per-parser basis to skip parsing certain lines, E.G. the IP parser will not parse lines from dnf logs to avoid package version false positives
* improved the IP regex to not get version substrings from packages that happen to fit the `a.b.c.d` format of IP addresses.

I still need to add handling for filenames, ad-hoc strings, and the user-provided keywords parser.

pmoravec:
I am getting:

    ERROR: Unable to obfuscate report: 'SoSHostnameParser' object has no attribute 'mapping'

Reproducer:

    rm -rf /etc/sos/cleaner/
    mkdir -p /etc/sos/cleaner/
    python3 ./bin/sos report --batch --build --clean
    # check a default_mapping was created
    python3 ./bin/sos report --batch --build --clean

Investigating..

sos/cleaner/parsers/hostname_parser.py
```diff
@@ -0,0 +1,40 @@
...
+    name = 'Hostname Parser'
+    map_file_key = 'hostname_map'
+    prep_map_file = 'sos_commands/host/hostname'
+    regex_patterns = [
+        r'(((\b|_)[a-zA-Z0-9-\.]{1,200}\.[a-zA-Z]{1,63}\b))'
+    ]
+
+    def __init__(self, conf_file=None, opt_domains=None):
+        super(SoSHostnameParser, self).__init__(conf_file)
+        self.mapping = SoSHostnameMap(opt_domains)
```
pmoravec, line 26: These lines need to be swapped (and in other child classes of `SoSCleanerParser`). Since the `self.mapping` is updated in the `SoSCleanerParser` constructor (https://github.com/sosreport/sos/pull/2093/files#diff-48e63d79ded5dde8f6c63d408b0e48ffR52).

This causes the

    ERROR: Unable to obfuscate report: 'SoSHostnameParser' object has no attribute 'mapping'

sos/cleaner/__init__.py
```diff
@@ -0,0 +1,554 @@
...
+        """
+        with open(path, 'w') as mf:
+            mf.write(json.dumps(_map, indent=4))
+        os.chmod(path, 0o600)
+        return path
+
+    def write_map_for_archive(self, _map):
+        try:
+            map_path = os.path.join(self.sys_tmp,
+                                    "private_map_%s" % self.arc_name)
```
pmoravec, line 344: A subjective suggestion: could the `private_map` be appended to the end of the filename, instead of beginning? Similarly like `.md5` checksum? Pros: it will be easier to remove old sosreports (`rm -rf /var/tmp/sos*`) as well as one more easily identifies the files as related to a sosreport. Negs: maybe a customer might make a mistake and upload `sosreport-pmoravec-rhel8-2020-06-02-trfaglz.tar.gz` altogether with `sosreport-pmoravec-rhel8-2020-06-02-trfaglz.tar.gz_private_map`.

TurboTurtle: Ack

sos/cleaner/__init__.py
TurboTurtle: Fair points both, ack to dropping the comment filter

TurboTurtle:
Latest branch update:

* Renames the map to `$archive_private_map`
* Obfuscates the host's short name in the sosreport tarball name, if present
* Similarly, obfuscates the top-level directory name within the tarball
* Extends the file naming obfuscation to both `report --clean` and `collect --clean`
* Adds a few more tests to the suite
* Fixes the parser instantiation bug @pmoravec mentions above
* Corrects and adds a few more ignores to the `ip_map`. Note, I decided to include Google's public DNS servers, `8.8.8.8` and `8.8.4.4` since these are extremely well known public addresses, and to me it seems silly to obfuscate them given that fact

sos/cleaner/mappings/ip_map.py
BryanQuigley: True.. Maybe they should be masked to still be identifiable as multicast addresses?  That's my bigger concern, not being able to tell.

TurboTurtle: That's certainly a reasonable concern. I'm not immediately sure how best to address it though - in no particular order here are some thoughts on it:

* We'd need to set a new range of obfuscated ranges.
    * Current design is addresses with a subnet (or in a known subnet) get assigned to a `100.x.y.z` address, with each new subnet incrementing the first octet. Address for which we aren't sure of the net get randomly generated, with each octet being between 1-99.

* We'd then need to prevent the non-multicast ranges from incrementing into the separate obfuscated range *and* into the actual multicast range. Granted, the latter is less of a concern since that would require 124+ discovered networks.

* Do we set the new obfuscated range above 239 (thus conflicting with other reserved ranges) or below 100 (meaning we'd need to restrict the random generation)?

man/en/sos-clean.1
```diff
@@ -0,0 +1,69 @@
...
+    [\-\-no-update]
+
+.SH DESCRIPTION
+\fBsos clean\fR or \fBsos mask\fR is an sos subcommand used to obfuscate sensitive information from
+previously generated sosreports that is not covered by the standard plugin-based post
+processing executed during report generation, for example IP addresses.
+.LP
+Data obfuscated via this utility is done so consistently, meaning for example an IP address of
+192.168.1.1 in an unprocessed sosreport that gets obfuscated to, for example, 100.0.0.1, will be
+changed to 100.0.0.1 in all occurences found in the report.
```
pmoravec, line 18: Maybe add a note that, by default, the same obfuscation mapping is persisted over subsequent runs of sos clean?

sos/archive.py
```diff
@@ -548,6 +548,16 @@ def add_final_manifest_data(self, method):
         self.add_string(self.manifest.get_json(indent=4),
                         os.path.join('sos_reports', 'manifest.json'))
 
+    def rename_archive_root(self, cleaner):
```
pmoravec, line 551: +1

pmoravec:
A minor comment: we shall catch `Invalid target: must be directory or tar archive` exception likewise `Invalid target: not an sos archive` one. See:

    # echo > empty.tar.gz
    # python3 ./bin/sos clean date.txt.tar.gz --batch
    ..
      File "./bin/sos", line 22, in <module>
        sos.execute()
      File "/root/sos-master/sos/__init__.py", line 174, in execute
        self._component.execute()
      File "/root/sos-master/sos/cleaner/__init__.py", line 244, in execute
        self.inspect_target_archive()
      File "/root/sos-master/sos/cleaner/__init__.py", line 170, in inspect_target_archive
        raise Exception("Invalid target: must be directory or tar archive")
    Exception: Invalid target: must be directory or tar archive

vs.:

    # echo > emtpy.txt
    # tar czf empty.txt.tar.gz empty.txt 
    # python3 ./bin/sos clean empty.txt.tar.gz --batch
    ..
    Invalid target: not an sos archive

pmoravec:
Two issues found during _repetitive_ cleaning a cleaned report (this might be seen as invalid use case of `sos clean`, though I was interested to see the outcome):

    # python3 ./bin/sos report -o host,pam --batch --build --clean
    ..
        /var/tmp/sosreport-host0-2020-06-07-cwdixpl
    # cp -r /var/tmp/sosreport-host0-2020-06-07-cwdixpl /var/tmp/sosreport-host0-2020-06-07-cwdixpl-1
    # python3 ./bin/sos clean --batch /var/tmp/sosreport-host0-2020-06-07-cwdixpl
    ..
    A mapping of obfuscated elements is available at
	/var/tmp/sosreport-host0-2020-06-07-cwdixpl-private_map

    The obfuscated archive is available at
        /var/tmp/sosreport-host0-2020-06-07-cwdixpl
    # diff -r /var/tmp/sosreport-host0-2020-06-07-cwdixpl-1/etc/pam.d/cockpit /var/tmp/sosreport-host0-2020-06-07-cwdixpl/etc/pam.d/cockpit 
    5c5
    < auth       optional     pam_sshadd.so
    ---
    > auth       optional     pamsshadd.so
    15c15
    < session    optional     pam_sshadd.so
    ---
    > session    optional     pamsshadd.so
    # 

Further, the `private_map` contains redundant `"host0.obfuscateddomain1.com": "host0.obfuscateddomain0.com"` (which really can occur only when cleaning a cleaned archive).

If you think the cleaning cleaned archive is not "supported", let disregard this comment,

sos/cleaner/__init__.py
```diff
@@ -0,0 +1,607 @@
...
+    @classmethod
+    def add_parser_options(cls, parser):
+        parser.usage = 'sos clean|mask TARGET [options]'
+        clean_grp = parser.add_argument_group(
+            'Cleaner/Masking Options',
+            'These options control how data obfuscation is performed'
+        )
+        clean_grp.add_argument('--domains', action='extend', default=[],
+                               help='List of domain names to obfuscate')
+        clean_grp.add_argument('target',
```
pmoravec, line 143: Maybe culd we `s/'target'/'TARGET'/g` as we refer to an argument value? See current help:

    # python3 ./bin/sos clean --help
    ..
    Cleaner/Masking Options:
      These options control how data obfuscation is performed
    
      --domains DOMAINS     List of domain names to obfuscate
      target                The directory or archive to obfuscate
      -j JOBS, --jobs JOBS  Number of concurrent archives to clean
      --map MAP_FILE        Provide a previously generated mapping file for
                            obfuscation
      --no-update           Do not update the --map file with new mappings from
                            this run
    #

where `target` is almost invisible.

pmoravec:
Mimicking our QE work (not sure how much of those are a real issues):

    # rm -f foo
    # touch foo
    # python3 ./bin/sos report --clean --batch --build -o host --map foo
    ..
    ERROR: Unable to obfuscate report: Expecting value: line 1 column 1 (char 0)
      File "./bin/sos", line 22, in <module>
        sos.execute()
      File "/root/sos-master/sos/__init__.py", line 174, in execute
        self._component.execute()
      File "/root/sos-master/sos/report/__init__.py", line 1336, in execute
        return self.final_work()
      File "/root/sos-master/sos/report/__init__.py", line 1144, in final_work
        cleaner.obfuscate_file(os.path.join(_dir, 'sos_logs', 'sos.log'),
    UnboundLocalError: local variable 'cleaner' referenced before assignment

(is the raised exception intentional? If so, we shall notify the user "we failed b'cos .." or imho the best output here would be `error in parsing map file: Expecting value: line 1 column 1 (char 0)`)



    # rm -rf /foo; mkdir /foo; python3 ./bin/sos report --clean --batch --build -o host --map /foo
    ..
    Successfully obfuscated 1 report(s)
    
    [cleaner] Could not update mapping config file: [Errno 21] Is a directory: '/foo'
    
    A mapping of obfuscated elements is available at
    ..
    # 

(shouldnt we fail at parsing the map? what map is used here?)

pmoravec:
For your enjoy: When testing `--domains=''` or ' ' or '*' or similar (just in case `cleaner` will survive): the `cleaner` does survive and obfuscates what it is requested to do. The `private_map` file then contains really tons of joy obfuscations like `"md.cpython-36.pyc": "host314.obfuscateddomain17.pyc"` :D

pmoravec:
Apart of the above comments (where half of them could be marked as irrelevant), I am OK with the PR. Great work!

man/en/sos-clean.1
TurboTurtle: Ack.

TurboTurtle:
> If you think the cleaning cleaned archive is not "supported", let disregard this comment,

I think that would/should be the case, perhaps we should put some check in to avoid duplicate processing? E.G. if `hostname` already contains `obfuscateddomain` or similar.

TurboTurtle:
> (is the raised exception intentional? If so, we shall notify the user "we failed b'cos .." or imho the best output here would be `error in parsing map file: Expecting value: line 1 column 1 (char 0)`)
>
> (shouldnt we fail at parsing the map? what map is used here?)

The question this raises for me, is if we can't load the map file - should we just report we couldn't load it and continue on normally (thus generating a completely fresh map), or abort entirely? At that, should we have a different answer based on if the run was `sos clean` vs `sos report --clean`?

arif-ali:
Thought, I would give this a go, may have more from me, over the coming week

**1. first point**

I get the following when cleaning

```
[cleaner] Could not update mapping config file: [Errno 2] No such file or directory: '/etc/sos/cleaner/default_mapping'
```

This is ok, when it is reading, but the man page also suggests that this file would be written to, but by default if the directory doesn't exist, it won't create it

```
By default, sos will write the generated private map file to /etc/sos/cleaner/default_mapping so that consistency is maintained by default.
```
**2. Second point**

I know this would be something that my customer would want (so a wish-list probably, so may require extended discussion, and probably not it in the initial pass). When there are kvm instances running on the node, and we can list them via `virsh list --all` we should get these hosts, and obfuscate any configuration file that either has this hostname and/or rename files that contain the hostname. On my test system, where I have a windows 10 vm (to logon to the customer), I have various files/folders that could potentially have the name. This scenario is showing only 1 instance

directories
```
./sys/fs/cgroup/pids/machine.slice/machine-qemu\x2d1\x2dwin10.scope
./sys/fs/cgroup/memory/machine.slice/machine-qemu\x2d1\x2dwin10.scope
./sys/fs/cgroup/net_cls,net_prio/machine.slice/machine-qemu\x2d1\x2dwin10.scope
./sys/fs/cgroup/cpuset/machine.slice/machine-qemu\x2d1\x2dwin10.scope
./sys/fs/cgroup/devices/machine.slice/machine-qemu\x2d1\x2dwin10.scope
./sys/fs/cgroup/systemd/machine.slice/machine-qemu\x2d1\x2dwin10.scope
./sys/fs/cgroup/unified/machine.slice/machine-qemu\x2d1\x2dwin10.scope
./sys/fs/cgroup/freezer/machine.slice/machine-qemu\x2d1\x2dwin10.scope
./sys/fs/cgroup/blkio/machine.slice/machine-qemu\x2d1\x2dwin10.scope
./sys/fs/cgroup/perf_event/machine.slice/machine-qemu\x2d1\x2dwin10.scope
./sys/fs/cgroup/cpu,cpuacct/machine.slice/machine-qemu\x2d1\x2dwin10.scope
```

files
```
./var/log/libvirt/qemu/win10.log
./var/log/libvirt/qemu/win10.log.0
./etc/libvirt/qemu/win10.xml
./run/libvirt/qemu/win10.xml
./run/libvirt/qemu/win10.pid
```

For a non running instance (instance name zentyal)

```
./var/log/libvirt/qemu/zentyal.log
./etc/libvirt/qemu/zentyal.xml
```

TurboTurtle:
> **1. first point**
> This is ok, when it is reading, but the man page also suggests that this file would be written to, but by default if the directory doesn't exist, it won't create it
> 

Right, that would/will be solved via packaging that will create the `/etc/sos/cleaner` directory upon installation. For RPMs, updating the packaging is next on my list after this (unless anyone else wants to take that on before I do). For Debian/Ubuntu packaging, that would come from their respective distro maintainers.

> **2. Second point**
>  When there are kvm instances running on the node, and we can list them via `virsh list --all` we should get these hosts, and obfuscate any configuration file that either has this hostname and/or rename files that contain the hostname.

This would need to be handled via a mix of the `--domains` option for the hostname parser, and the to-be-written keyword parser that allows users to specify keywords to be obfuscated. The keyword parser will be included in this patchset before it is merged.

BryanQuigley:
I might be overly paranoid - but I am worried about trying to support non-root uses of mask/clean given they will:
 * cause the report permissions to not actually match.
 * possibly expose data to other users with less permissions (especially with it going to /var/tmp..)

I see two possibilities:
 * Require root to mask (I guess this could apply to collector too..); fixes both above
 * Instead output non-root mask/collector to working directory so user is explicitly in control

Namely, the os.getuid() != 0 bits in SoSObfuscationArchive are what got me thinking down this path - if someone kills their sos mask run at the right time, I believe this could result in anyone being able to read the archive left in progress.

On another note, I also think we might want to take a relook at naming before releasing 4.x:
sosreport-host0-2020-06-15-njsbcqd-obfuscated.tar.xz 
@slashdd was recently talking about a similar thing in filename to show it used upload flag.  Putting that all together as is, would be to big IMHO.

These don't have to be fully addressed in this PR, but at the least can we justify (or change?) the chmod 777 more?

TurboTurtle:
> I see two possibilities:
> 
> * Require root to mask (I guess this could apply to collector too..); fixes both above

Enforcing root on collector would break the majority of the use cases for it. It is very rare for sysadmins to have their SSH keys tied to their local root accounts for access to other systems in their environments, it is far more common that their local non-root users have their SSH keys deployed and use _those_ to connect to remote systems as root. 

> * Instead output non-root mask/collector to working directory so user is explicitly in control
> 
I don't see how that would alleviate a potential problem here.

> Namely, the os.getuid() != 0 bits in SoSObfuscationArchive are what got me thinking down this path - if someone kills their sos mask run at the right time, I believe this could result in anyone being able to read the archive left in progress.
> 
They have to have appropriate permissions on the archive to begin with, which means they _already_ have access to the data. The permissions wrangling done via non-root clean runs is to allow us to manipulate the file and have it re-written into the rebuilt archive. We're not exposing any data at this point that is not readily available to the use that is running clean.

> These don't have to be fully addressed in this PR, but at the least can we justify (or change?) the chmod 777 more?

There is no chmod 777 occurring. We're adding _owner_ read/write perms to files and directories (and possibly execute to the dirs missing it) that are missing those from the archive extraction to allow us to properly rewrite the file contents and add them back to the archive.

sos/cleaner/obfuscation_archive.py
```diff
@@ -119,29 +135,40 @@ def compress(self, cmd):
...
+            else:
+                shutil.rmtree(name)
+        if os.getuid() != 0:
+            # this is rather silly, but due to the fact we preserve permissions
+            # in reports, a non-root user would be unable to remove a lot of
+            # paths under proc/ and sys/. Since we're at the point where the
+            # archive has been re-compressed and we're removing data that will
+            # no longer be referenced anywhere, just open the perms to allow
+            # us to delete it all
+            sos_get_command_output("chmod -R 777 %s" % self.extracted_path)
```
BryanQuigley, line 193: This is the chmod 777 I'm most worried about.  chmod can end up taking a while.. and then it's all open unless I'm missing somethng.  I also don't full grep why a full 777 helps with removal.

BryanQuigley:
Our previous security position is that the data is always owned by only root - hence putting in a world readable directory is fine.  With both mask and collector that assumption has changed, but it's still limited to that single user so thinking more about it, it should be fine Nevermind on that bit, still don't like the chmod 777 though.

sos/cleaner/obfuscation_archive.py
TurboTurtle: OH, I thought you were referring to the extraction permission changes.

That chmod was added before I resolved to add the execute bit for directory perms on extraction. It may be possible to drop it at this point, let me run some tests.

TurboTurtle: Ack, we should be able to drop this now.

TurboTurtle:
Well that's embarassing - I some how managed to neglect this PR in the commit messages despite getting the several year-old issue that was opened for it.

This PR has been merged, with the last patch going into master as 2d73f01.

PR #2099, opened 2020-06-03
[man] Update sysroot parameter description

slashdd:
Worth mentionning that 'sysroot' option can also
serve when run from a LiveCD, without having the
LiveCD environment taint the data collection process.

One may want to capture data from the actual
impacted system and not necessarily the LiveCD
environment which may only serves for troubleshooting.

---

### DIFF
--- a/man/en/sosreport.1
+++ b/man/en/sosreport.1
@@ -130,7 +130,7 @@ that will be displayed when using \fB--list-presets\fR.
 .TP
 .B \-s, \--sysroot SYSROOT
 Specify an alternate root file system path. Useful for collecting
-reports from containers and images.
+reports from containers, LiveCD and images.
 .TP
 .B \-c, \--chroot {auto|always|never}
 Set the chroot mode. When \--sysroot is used commands default to


bmr-cymru:
Note that in that case sometimes you would be better off executing `sos report` under `chroot`, for e.g. using the Fedora / RHEL mount point conventions:

```
# chroot /mnt/sysimage sos report -v --batch
```

The differences are subtle and depend on the different binaries that may be present in the host vs. the live (or container) environment as well as the setting of `PATH`. If you use the built-in `--sysroot` then you may find that command output comes from binaries in the LiveCD environment rather than (as you might expect) the host. If you use a chroot then all binaries are guaranteed to be taken from the host (and that they will be linked as they normally would be by the host loader and shared libraries - only kernel-specific data will be "tainted" by the Live environment in this instance). You can also override this on the sos command line by specifying `--sysroot /mnt/sysimage --chroot=never` but that ends up being more to type!

The reason for this is that `--sysroot` and its accompanying chroot framework are designed for use in scenarios like Red Hat Atomic where the actual host environment may be missing critical commands that we rely on for troubleshooting. The current arrangement allows those commands to execute from the container (Live) environment and fill in the gaps.

For conventional support use on a non-containerised OS though I would suggest using chroot as a simpler and potentially less misleading way to execute in the host file system context. This is one reason I never really emphasised this option for other use cases.

bmr-cymru:
All that said it probably would benefit us to improve the documentation around these options - at least to make those caveats clear and to help users understand when it's a good idea.

PR #2109, opened 2020-06-09
[pacemaker] Fix scrubbing when password contains an equal sign

nrwahl2:
If the password contains one or more equal signs ('='), only the substring
after the final equal sign is scrubbed. The rest of the password appears in
plain text.

This patch modifies the scrub regex to scrub all characters after the first
equal sign.

Related to RHBZ#1845386.

---

### DIFF
--- a/sos/report/plugins/pacemaker.py
+++ b/sos/report/plugins/pacemaker.py
@@ -55,14 +55,14 @@ class Pacemaker(Plugin):
     def postproc_crm_shell(self):
         self.do_cmd_output_sub(
             "crm configure show",
-            r"passw(\S*)=\S+",
+            r"passw([^\s=]*)=\S+",
             r"passw\1=********"
         )
 
     def postproc_pcs(self):
         self.do_cmd_output_sub(
             "pcs config",
-            r"passw(\S*)=\S+",
+            r"passw([^\s=]*)=\S+",
             r"passw\1=********"
         )
 


nrwahl2:
Before:
```
[root@fastvm-rhel-8-0-23 ~]# pcs stonith create --disabled testdev fence_vmware_soap ipaddr=1.1.1.1 login=username passwd='asdf=?1234'

[root@fastvm-rhel-8-0-23 tmp]# cat sosreport-fastvm-rhel-8-0-23-2020-06-08-rkllgeq/sos_commands/pacemaker/pcs_config 
...
 Resource: testdev (class=stonith type=fence_vmware_soap)
  Attributes: ipaddr=1.1.1.1 login=username passwd=asdf=********
...
```

After:
```
[root@fastvm-rhel-8-0-23 tmp]# cat sosreport-fastvm-rhel-8-0-23-2020-06-08-tpwyjiz/sos_commands/pacemaker/pcs_config 
...
 Resource: testdev (class=stonith type=fence_vmware_soap)
  Attributes: ipaddr=1.1.1.1 login=username passwd=********
...
```

sos/report/plugins/pacemaker.py
```diff
@@ -55,14 +55,14 @@ def setup_pcs(self):
     def postproc_crm_shell(self):
         self.do_cmd_output_sub(
             "crm configure show",
-            r"passw(\S*)=\S+",
+            r"passw([^\s=]*)=\S+",
```
pmoravec, line 58: What are all the possible options of strings you try to match here? As in some original version, we were scrubbing `incoming_password=` also - that is IMHO why the `\S*` was (as a wide heuristic) there.

Replacing `\S` by it's opposite `\s` simply sounds strange to me..

nrwahl2: `\S` is all non-whitespace characters. `[^\s=]` is all non-whitespace, non-equal-sign characters. Is there a more elegant way to do that?

nrwahl2: See https://github.com/sosreport/sos/pull/2109#issuecomment-641052464 and RHBZ#1845386 for the rationale and the problem I'm trying to solve here.

jcastill: If I understand what Pavel is trying to say, I think he's talking about the distinction between \S and \s, where the former "Matches any character which is not a whitespace character." and the latter "Matches Unicode whitespace characters". So perhaps the regex should be:
[^\S=]
Instead of:

[^\s=]

? As in the regex is perfect, just needs to capitalize the s. Is that what you meant Pavel?

nrwahl2: @jcastill That's the thing -- it shouldn't be `[^\S=]`.
Previously, we used `\S`, which means "any non-whitespace character."
This patch changes that to `[^\s=]`, which means "any character that is not a member of the set {whitespace, equal sign}". `[^\s=]` is equivalent to "\S that is not an equal sign".

jcastill: ah ok, thanks for the clarification!

pmoravec: Ah damn regular expressions - I misinterpreted the `^` /o\. ACK to the patch then, now it makes sense also to me :).

pmoravec:
Shouldn't the same apply also to:

https://github.com/sosreport/sos/blob/master/sos/report/plugins/pacemaker.py#L65

?

sos/report/plugins/pacemaker.py
nrwahl2: No worries; regexes are way too easy to misread :)

nrwahl2:
> Shouldn't the same apply also to:
> 
> https://github.com/sosreport/sos/blob/master/sos/report/plugins/pacemaker.py#L65
> 
> ?

That's already in the commit, isn't it?

pmoravec:
/me feels embarassed for both the over-looks I made.

I was just focused on commenting the first code change, that I overlooked the other one. Sorry for the confusion, the patch is 100% correct.

PR #2111, opened 2020-06-10
[collect] Pass SoSNode object for remote policy loading

TurboTurtle:
Rather than read `/etc/os-release` indiscriminately and pass the entire
contents as a single string to each policy's `check()` method, now just
pass the actual `SoSNode()` object and allow the policy to determine
what to do with it, if provided.

This allows policies to mimic their local checks for remote nodes, be
that substring presence in a file, a file's presence, or anything else.

Resolves: #2111

---

### DIFF
--- a/sos/collector/sosnode.py
+++ b/sos/collector/sosnode.py
@@ -340,7 +340,7 @@ class SosNode():
             return self.commons['policy']
         host = load(cache={}, sysroot=self.opts.sysroot, init=InitSystem(),
                     probe_runtime=False, remote_exec=self.ssh_cmd,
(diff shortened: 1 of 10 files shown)


BryanQuigley:
I was thinking we go in a different direction and require the use of /etc/os-release .. will mock it up

BryanQuigley:
Mocked up in #2113 - I have not tested with Remote options yet.

TurboTurtle:
Closing this in favor of continued development on #2113, which provides a bigger benefit by consolidating our policy checks and the current `collect` design plays into that new approach better than this refactoring would.

PR #2113, opened 2020-06-13
[policies] switch to common method for determining OS

BryanQuigley:
[policies] Switch to common method for determing OS
    
    This uses /etc/os-release exclusively.
    It also requires the module names to match the OS ID exactly
    in this first iteration.
    
    WIP
    This would need to split the policies up to all the invidual
    id options.
    Also, the use of ID_LIKE or partial matches may be considered.

Also drop osx.
---


TurboTurtle:
I like the idea of consolidating these checks, but I'm not sure off the top of my head if this will play nicely with containerized versions of RHEL - e.g. CoreOS. At worst though, that just means we'll still carry overrides for those policies, which is still an improvement since we'd then only have a small handful of policies with overrides, rather than every single policy needing one.

Massive +1 on dropping OSX too

With the current `collect` design, we pass the entire contents of the remote `os-release` file, which would be a better fit with this redesign than my idea of just passing the `SoSNode` object.

BryanQuigley:
Added id_like checking, this may actually capture all modern suse's .  (may need to add a quite empty sles.py)

I did note (https://github.com/nir0s/distro/tree/master/tests/resources/distros) that requiring os-release may not support rhel6?  Is sos 4.x going to make it to rhel6 - I'm not sure of the python3 story there.

I couldn't determine where dist_version is actually used?  We can just pull that from os-release to if it's useful.

I need to better understand the differences between Centos/community and RH official, if it's just the URL, we can easily pull that from os-release too.

TurboTurtle:
> Added id_like checking, this may actually capture all modern suse's . (may need to add a quite empty sles.py)
> 
> I did note (https://github.com/nir0s/distro/tree/master/tests/resources/distros) that requiring os-release may not support rhel6? Is sos 4.x going to make it to rhel6 - I'm not sure of the python3 story there.
> 
No chance of 4.0 landing in RHEL 6. For reference, sos-4.0 also will not land in RHEL 7, it'd be RHEL 8+ only.

> I couldn't determine where dist_version is actually used? We can just pull that from os-release to if it's useful.
> 
We use `dist_version()` in the `python` plugin for RH distros. We use it to know when to capture version output from `platform-python` since RHEL 8+ uses `/usr/libexec/platform-python` as the interpreter rather than `/usr/bin/pythonX`

> I need to better understand the differences between Centos/community and RH official, if it's just the URL, we can easily pull that from os-release too.

For policy loading purposes, it'd only be the URL so we should be good with this for CentOS/RHEL differences. Beyond that, we just keep the `--upload` bits in the RHEL policy so that we don't have CentOS systems trying to upload to our support servers. Should be the same with Fedora once these `check()` changes are made.

There's more to the differences between the two distros in usage, but nothing that affects us here.

BryanQuigley:
If I understand correctly that means we can drop Atomic host. 

Added variant detection because that's apparently how Fedora coreos works, and I'm guessing RHEL coreos is similar. 

Also, anyone against dropping the IBM bits?  The URLs are broken and it should mostly *just work* given it should detect that it's *like* redhat.

bmr-cymru:
> Also, anyone against dropping the IBM bits? The URLs are broken and it should mostly just work given it should detect that it's like redhat.

The right thing to do here is to reach out to IBM either via the sos-devel mailing list or by tagging one of the previous contributors here on GitHub. Then they can have the opportunity to fix things up if they wish - as I understand it all the products that these policies relate to are still extant.

BryanQuigley:
I need to rethink how this should work for the split up RH plugin

kamalesh-babulal:
> Good point. @kamalesh-babulal do you by any chance have the exact os-release ID=? for both IBM defned distros - oh and the URLs are boken now. Alternatively if ID_LIKE="rhel" or similar it should just work without a definition file.

@BryanQuigley Thanks for letting me know, I need to check on the details you have asked for.

PR #2119, opened 2020-06-18
Vdsm include all certs

mmartinv:
---


pmoravec:
Similar comment like to #2118 :).

No other key file or password file under `/etc/pki/vdsm/` that we should avoid to collect? Just the `/etc/pki/vdsm/libvirt-*/*-key.*'`?

PR #2130, opened 2020-06-25
[logs] collect also non-persistent journal logs

pmoravec:
Collect journalctl logs also when journal is configured to store logs
in memory only.

Further, when --all-logs is provided, collect the transient logs in
/var/log/journal dir as well.

Resolves: #2130

---

### DIFF
--- a/sos/report/plugins/logs.py
+++ b/sos/report/plugins/logs.py
@@ -59,13 +59,21 @@ class Logs(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin, CosPlugin):
         self.add_cmd_output("journalctl --disk-usage")
         self.add_cmd_output('ls -alRh /var/log/')
 
-        journal = os.path.exists("/var/log/journal/")
-        if journal and self.is_installed("systemd"):
+        # collect journal logs if:
+        # - there is some data present, either persistent or runtime only
+        # - systemd-journald service exists
+        # otherwise fallback to collecting few well known logfiles directly
+        journal = any([os.path.exists(p + "/log/journal/")
+                      for p in ["/var", "/run"]])
+        if journal and self.is_service("systemd-journald"):
             self.add_journal(since=since)
-            self.add_journal(boot="this", catalog=True)
-            self.add_journal(boot="last", catalog=True)
+            self.add_journal(boot="this", catalog=True, since=since)
+            self.add_journal(boot="last", catalog=True, since=since)
             if self.get_option("all_logs"):
-                self.add_copy_spec("/var/log/journal/*")
+                self.add_copy_spec([
+                    "/var/log/journal/*",
+                    "/run/log/journal/*"
+                ])
         else:  # If not using journal
             if not self.get_option("all_logs"):
                 self.add_copy_spec([


pmoravec:
Cc @BryanQuigley or @slashdd for review from other distros.

We should test for presence and collect also non-persistent journal logs, that's the main goal.

I made the "is `systemd` installed" condition more accurate to _presence_ of `systemd-journald` service - this covers possible scenarios when the service is not running. It could make sense to test _enabledness_ of the service, maybe.

BryanQuigley:
I'd be for just collecting /run/log/journal/ if --all-logs outside of the existing if/else.

If the journal is not persistent, we need to collect the more traditional files /var/log/*.log files which is what the else does - else we can miss log data.

pmoravec:
Well at least in RedHatPolicy, not collecting `journalctl` for transient journals but copying some particular log files is seen as a regression: our support associates started to rely on the `journalctl` outputs.

sos/report/plugins/logs.py
```diff
@@ -59,13 +59,21 @@ def setup(self):
...
+        if journal and self.is_service("systemd-journald"):
             self.add_journal(since=since)
             self.add_journal(boot="this", catalog=True)
             self.add_journal(boot="last", catalog=True)
             if self.get_option("all_logs"):
-                self.add_copy_spec("/var/log/journal/*")
+                self.add_copy_spec([
+                    "/var/log/journal/*",
+                    "/run/log/journal/*"
+                ])
```
bmr-cymru, line 76: I guess I have two questions about this change: if there are files present at either location, what's the harm in just collecting them anyway? Are there valid situations where we *don't* want to do that (for space concerns, for e.g.)?

If would simplify things a bit if we could eliminate that logic.

I'm also a bit curious what the rationale is now for collecting raw journals vs. `journalctl` output; it seems if we have one maybe we don't need the other?

The block that we're discussing here (many commits ago) was originally intended to capture the journal logs on Fedora systems after Fedora moved ahead with disabling/removing syslog by default, so I'm not sure it's the best base for future work since the use of journals has become more pervasive and universal since this was first done.

Anyway; that falls a bit outside the scope of this specific patch review which looks generally fine to me outside the formatting nitpicks.

BryanQuigley:
Ubuntu 16.04 LTS has a runtime journal only, and uses syslog for most real things.   This would break that setup.  Ubuntu in general also has never finished a migration away from rsyslog and still does both by default.

@slashdd - do you plan on trying to backport 4.0 to 16.04?  Dropping 16.04 would likely help streamline other things - it has the oldest python3 version.

pmoravec:
So for transient `journal` logs, Red Hat support requires `journalctl` output, while Ubuntu the `rsyslog` populated logfiles.

We can either split that code for particular policies, or collect *both* `journalctl` and standard logfiles in both current if-then branches, like:

    if self.is_service("systemd-journald"):
            self.add_journal(since=since)
            self.add_journal(boot="this", catalog=True)
            self.add_journal(boot="last", catalog=True)
                self.add_copy_spec([
                    "/var/log/journal/*",
                    "/run/log/journal/*"
                ])

        self.add_copy_spec([
            "/var/log/syslog",
            "/var/log/syslog.1",
            "/var/log/syslog.2*",
            "/var/log/kern.log",
            "/var/log/kern.log.1",
            "/var/log/kern.log.2*",
            "/var/log/auth.log",
            "/var/log/auth.log.1",
            "/var/log/auth.log.2*",
        ])
    else:
        self.add_copy_spec([
            "/var/log/syslog*",
            "/var/log/kern.log*",
            "/var/log/auth.log*",
        ])

I suggest this above approach to ensure no regression for either distro between 3.8 and 3.9. And open a new issue (ideally resolved still in 4.0 timeframe) to decide, what outputs / dirs / logfiles to collect in what particular situation / for what particular distro.

BryanQuigley:
That will unfortunately double the logs collected on Ubuntu again - which I'd be against. This should be a call for @slashdd

pmoravec:
As the distro specific requirements are contradicting, I will re-spin the PR to use distr specific plugins, then.

pmoravec:
Updated.

As far as I understood, then:
- Ubuntu/Debian is not interested in journal logs (currently only transient journals enabled, which the current plugin does not capture, in favour of syslog files directly)
- RedHat is interested in journal logs *only*, as the syslog files are not present / they are Ubuntu/Debian specific only (by default)

pmoravec:
@slashdd and @BryanQuigley : .. or do you require collecting durable journal logs? (i.e. pls. specify conditions for `collect_journals` and `collect_syslogs`).

BryanQuigley:
Ubuntu was interested in Journal logs if they are persistent. (And if so, not to collect the /var/log/syslog/etc

How about always try to collect without any conditions blocking it?
        self.add_journal(since=since)
        self.add_journal(boot="this", catalog=True, since=since)

pmoravec:
> Ubuntu was interested in Journal logs if they are persistent. (And if so, not to collect the /var/log/syslog/etc

Which means requirements to collect journal logs and/or syslogs are mutually exclusive. That can simplify something in a final 

> 
> How about always try to collect without any conditions blocking it?

And what condition to use for collecting syslogs (`syslog*`, `kern*`, `auth*`) directly? Will not unconditional collection of journal logs cause syslogs are also collected, meaning you collect *both* logs, contradicting yours:

> That will unfortunately double the logs collected on Ubuntu again - which I'd be against. This should be a call for @slashdd
?

Further,
> self.add_journal(since=since)
> self.add_journal(boot="this", catalog=True, since=since)

I recall some requirement from Red Hat to collect also previous boot.

Since:
- I see more subtle differences among distros
- I assume the current plugin fits Canonical needs
- Red Hat needs to backport the regression fix to some currently being developed `sos` package

I will suggest a "workaround PR" that:
- behaves the same for Ubuntu/Debian (plus limiting journal logs with `since`)
- collects the journal logs for Red Hat distros

BryanQuigley:
Talked to @slashdd in IRC.  We can drop 4.0 support for Ubuntu 16.04.    Sorry!  I think this means your initial PR would work.

slashdd:
I plan to stop updating Xenial/16.04LTS at 3.9.1 upstream version, and then only cherry-pick fixes here and there as needed.
Hence, it can be safely discarded of the equation for 4.0 and forward.

pmoravec:
OK, original PR with @bmr-cymru formatting suggestions re-spinned.

PR #2135, opened 2020-06-26
[vhostmd] Capture the output of vhostmd

jcastill:
Currently we only use the output of vhostmd temporarily
to check if the machine is a vm or not. With this
patch we also capture its output.

Resolves: #2135 

---


bmr-cymru:
What you need to do is to convert the call to `exec_cmd()` (run command now and return output/status) to a call to `collect_cmd_output()` (run command now, add to report, and return status/output - this is equivalent to the old `get_command_output_now()`). Then you can re-use the dictionary returned from the call for the `vw` tests but still have the output added to the report.

Issue #2125, opened 2020-06-18
[sos4.0] Refactor configuration files/directories


pmoravec:
As we refactored the `sosreport` into `sos [report|collect|clean]` in the tool's behaviour, we should do the same with configuration. Current status:

- `/etc/sos.conf` being used by `sos report`
- `/etc/sos.extras.d/` can be used by `sos_extras` plugin
- `/etc/sos/cleaner/default_mapping` that stores `cleaner`'s obfuscation mapping from previous runs
- I think nothing else..?

That "hierarchy"  is not logical, though it can serve our current purposes well. But as we will release a major version of the tool, it makes sense to refactor this. Some questions that answering can help us to decide:

- Is it worth having an extra directory for each component, to easily *add* (without an interference) config relevant to it? Or is e.g. `/etc/sos.d/` enough (and sections in the configs define the component)? Or is such a dir required at all?
- is an extra (mandatory? optional?) config file (for each component? or just whole `sos`?) required?
- is the current "default config dominance to `report`" useful? Should the default config file refer to `report` or to `sos`, rather?
- should we support reading `sos` Global options from a config file? (if not, then the global config file is less interesting) Or just from the config file of selected component? (I mean: running `sos report` should read `/etc/sos.conf` with global options first, and then `/etc/sos/report.conf` for `report`'s options?)

My proposal:
- `/etc/sos.conf` just being a symlink to `/etc/sos/report.conf` for legacy reasons; `report` should read from the later file
- `/etc/sos/[report|collect|clean].conf` the basic config file for the components; IMHO neither is mandatory (do we even really *need* current `/etc/sos.conf` to be present?)
- `/etc/sos/[report|collect|clean]` as optional directories: files with `*.conf` mask will be traversed in *undefined* ordering after the `/etc/sos/[report|collect|clean].conf` is read, files with different filemask serve specific needs for the component (like `default_mapping`)

But isn't that too complex?

TurboTurtle:
Initially, my reaction is to say we should have a `.conf` file for each component. When that component is loaded, we read exclusively from that file for configuration updates.

For global options though, that can become a pain - but at the same time it seems silly to have _both_ a `sos.conf` and a `$component.conf` where the former only has global definitions.

I'm personally not a fan of the option `.conf` files - one place for each component should be granular enough imho.

I guess my proposal would be:

* Drop `/etc/sos.conf` entirely
[...]

PR #2136, opened 2020-06-26
[global] Update config file location and layout

TurboTurtle:
Moves the default config file we look for to /etc/sos/sos.conf instead
of /etc/sos.conf. Extends the config file to look for a section matching
the name of the component being used. Renames the "general" section to
"global" and the "tunables" section to "plugin_options".

Updates the default sos.conf to this style and adds some comments to the
file. Update the man page for sos.conf.

Note that this commit does NOT update sos.spec to drop the default
configuration file in the new location, as that will be handled by a
later commit to update the specfile wholesale.

Closes: #2125
Resolves: #2136

---

### DIFF
--- a/man/en/sos.conf.5
+++ b/man/en/sos.conf.5
@@ -3,7 +3,76 @@
 sos.conf \- sosreport configuration
 .SH DESCRIPTION
 .sp
-sosreport uses a configuration file at /etc/sos/sos.conf.
+sosreport uses a configuration file at /etc/sos/sos.conf, and there are
+subdirectories under /etc/sos that are used for specific purposes.
+
+Note that non-root users may override options set in /etc/sos/sos.conf by creating
+their own sos.conf under $HOME/.config/sos.
+
+The order in which options are loaded is as follows:
+
+  1. System configuration file at /etc/sos/sos.conf
+  2. User-specific configuration file at $HOME/.config/sos/sos.conf (for sos
+     components that support non-root)
+  3. In the case of running \fBsos report\fR, presets either automatically loaded
+     due to system configuration, or specified via \fB--preset\fR
+  4. Command line values
+
+
+In other words, config files will override defaults, presets override config files,
+and command line values override presets and config files.
+
+.SH SUBDIRECTORIES
+The following subdirectories exist under /etc/sos and are used as noted below
+
+.TP
+\fBextras.d\fP
+This directory is used to store configuration files used by the sos_extras plugin.
+
+The plugin traverses this directory and for each file there it executes commands
+or collects files optionally with sizelimit.
+
+Expected content of an extras file is as follows:
+    - empty lines or those starting with '#' are ignored
+    - add_copy_spec called to lines starting by ':', optionally followed by
+      sizelimit
+    - otherwise, whole line will be executed as a command.
+    Example:
+    command1 --arg1 val1
+    command2
+    :/path/to/file
+    :/path/to/files* sizelimit
+
+    WARNING: be careful what files to collect or what commands to execute:
+    - avoid calling potentially dangerous or system altering commands, like:
+      - using multiple commands on a line (via pipes, semicolon etc.)
+      - executing commands on background
+      - setting env.variables (as those will be ignored)
+      - altering a system (not only by "rm -rf")
+    - be aware, no secret obfuscation is made
+.TP
+\fBgroups.d\fP
+This directory is used to store host group configuration files for \fBsos collect\fP.
+
+These files can specify any/all of the \fBmaster\fP, \fBnodes\fP, and \fBcluster-type\fP
+options.
+
+Users may create their own private host groups in $HOME/.config/sos/groups.d/. If
+a host group of the same name is saved in both the user's homedir and this directory,
+the homedir configuration file will have precedence. When run as non-root, \fBsos collect\fP
+will save host groups to the user's home dir, and create the necessary directory structure
+if required.
+
+Note that non-root users may load host groups defined under /etc/sos/groups.d/, but they
+may not write new groups or update existing groups saved there.
+
+.TP
+\fBpresets.d\fP
+This directory is used to store preset configuration files for \fBsos report\fP.
+
+Presets may be used to save standard sets of options. See \fBman sos-report\fP for
+more information.
+
 .SH PARAMETERS
 .sp
 There are sections for each sos component, as well as global values and
@@ -53,7 +122,7 @@ To disable the 'host' and 'filesys' plugins:
 .LP
 [report]
 .br
-noplugins = host,filesys
+skip-plugins = host,filesys
 .sp
 To disable rpm package verification in the RPM plugin:
 .LP
@@ -64,6 +133,8 @@ rpm.rpmva = off
 .SH FILES
 .sp
 /etc/sos/sos.conf
+.br
+$HOME/.config/sos/sos.conf (optional)
 .SH SEE ALSO
 .sp
 sos-report(1)
--- a/man/en/sos.conf.5
+++ b/man/en/sos.conf.5
@@ -3,11 +3,11 @@
 sos.conf \- sosreport configuration
 .SH DESCRIPTION
 .sp
-sosreport uses a configuration file at /etc/sos.conf.
+sosreport uses a configuration file at /etc/sos/sos.conf.
 .SH PARAMETERS
 .sp
-There are three sections in the sosreport configuration file:
-general, plugins and tunables. Options are set using 'ini'-style
+There are sections for each sos component, as well as global values and
+those for plugin options. Options are set using 'ini'-style
 \fBname = value\fP pairs. Disabling/enabling a boolean option
 is done the same way like on command line (e.g. process.lsof=off).
 
@@ -18,30 +18,30 @@ will result in enabling those options, regardless of value set.
 
 Sections are parsed in the ordering:
 .br
-- \fB[general]\fP
+- \fB[global]\fP
 .br
-- \fB[plugins]\fP (disable)
+- \fB[component]\fP
 .br
-- \fB[plugins]\fP (enable)
-.br
-- \fB[tunables]\fP
+- \fB[plugin_options]\fP
 
 .TP
-\fB[general]\fP
+\fB[global]\fP
 <option>      Sets (long) option value. Short options (i.e. z=auto)
               are not supported.
 .TP
-\fB[plugins]\fP
-disable       Comma separated list of plugins to disable.
-.br
-enable        Comma separated list of plugins to enable.
+\fB[component]\fP
+Each component will have a separate section, and it will support the options
+that particular component provides. These are readily identifiable in the
+\fB--help\fP output for each component, E.G. \fBsos report --help\fP.
 .TP
-\fB[tunables]\fP
-plugin.option Alter available options for defined plugin.
+\fB[plugin_options]\fP
+Alter available options for defined (and loaded) plugins.
+
+Takes the form plugin.option = value, for example \fBrpm.rpmva = true\fP.
 .SH EXAMPLES
 To use quiet and batch mode with 10 threads:
 .LP
-[general]
+[global]
 .br
 batch=yes
 .br
@@ -51,19 +51,21 @@ threads=10
 .sp
 To disable the 'host' and 'filesys' plugins:
 .LP
-[plugins]
+[report]
 .br
-disable = host, filesys
+noplugins = host,filesys
 .sp
 To disable rpm package verification in the RPM plugin:
 .LP
-[tunables]
+[plugin_options]
 .br
 rpm.rpmva = off
 .br
 .SH FILES
 .sp
-/etc/sos.conf
+/etc/sos/sos.conf
 .SH SEE ALSO
 .sp
-sosreport(1)
+sos-report(1)
+sos-collect(1)
+sos-clean(1)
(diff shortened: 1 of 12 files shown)


TurboTurtle:
Note: this commit does *not* implement the per-user overrides that I briefly mentioned in #2125. If this design is acceptable, then I can either extend this PR or we can add that later as we close out 4.0 (or go into 4.1 to ensure stability of everything else going into 4.0).

pmoravec:
I think no option from `[global]` section in `/etc/sos/sos.conf` have an effect, currently. E.g. having there:

    [global]
    batch = true

then the `batch` option is not applied to either `report` or `clean`. The same with `verbose = 3` - default verbosity used instead.

TurboTurtle:
Oops, left off a conditional check there for global. Fixed now.

pmoravec:
./sos/__init__.py:                                dest="config_file", default="/etc/sos.conf",


needs to change `default` to `/etc/sos/sos.conf` :)

pmoravec:
The parser ignores unknown options - we should at least warn about them (cf. `sos report --blah` returns an error, even).

Adding an option twice to the same section, like:

    [report]
    build = true
    build = true

then `sos` fails with:

    sosreport (version 3.9)

    Failed to parse configuration file /etc/sos/sos.conf

and even verbose logging does not suggest why.


Neither issue is a regression, the same behaviour applies to the current code, so I can raise a separate issue for it (until @TurboTurtle decides to fix it within this PR).

sos.conf
```diff
@@ -1,11 +1,36 @@
...
 #verbose = 3
 #verify = yes
 #batch = yes
 #log-size = 15
 
-[plugins]
-#disable = rpm, selinux, dovecot
+[report]
+# Options that will apply to any `sos report` run should be listed here.
+# Note that the option names *must* be the long-form name as seen in --help
```
pmoravec, line 13: There is something wrong.. 2 issues in fact:

- `sos report has cmd option `--skip-plugins` but we support `noplugins` option in config file.
- `sos report has cmd option `--only-plugins` but we support `onlyplugins` option in config file - without the "`-`". Config file option `only-plugins = kernel` is ignored.

TurboTurtle: Nice catch, we'll need to change the destination in the `add_argument()` statement for these, or change the long form syntax.

TurboTurtle:
> The parser ignores unknown options - we should at least warn about them (cf. `sos report --blah` returns an error, even).
> Adding an option twice to the same section, like:
> Neither issue is a regression, the same behaviour applies to the current code, so I can raise a separate issue for it (until @TurboTurtle decides to fix it within this PR).

Let's fix it here, all at once. Notifying of invalid options is no issue, however I'm not sure why duplicate options in the same section would cause that issue, unless it's built into `ConfigParser()` to fail like that.

sos.conf
TurboTurtle: Rather, the above is just for `-n`. We use `skip-plugins` in the long form, but `noplugins` in code. I think `skip-plugins` would be more favorable here, and I'll go with that if no one objects.

TurboTurtle:
Updated the PR with:

* Update the arg default for config file within the report parser as well
* Catch `DuplicateOptionError` separately and report it (and abort, `ConfigParser()` does not offer a way to recover gracefully)
* Catch unknown args and report them, but do not abort
* After config file parsing is done, log the updated effective options
* Standardize the option long-form names and the destination vars they save to, e.g. `noplugins` -> `skip_plugins`.

TurboTurtle:
I'm going to work towards extending this PR to include the per-user config file overrides, it just makes sense to do it all at once here.

Are we good with the proposal in #2125 to move root presets/host groups into `/etc/sos/$thing.d/` and non-root to `~/.config/sos/$thing.d/`, or should we place them elsewhere? Or leave everything alone right now and have it still live in `/var/lib/sos/`?

TurboTurtle:
Added a few more commits to this series, that independently update the `report` and `collect` components to use `/etc/sos` now instead of `/var/lib/sos|sos-collector`.

bmr-cymru:
Since this will affect packagers of `sos` I think we ought to mention it prominently in the release notes - I thought we already had a label for that, but I couldn't find one so I've created the `release note` label and added it to this PR.

TurboTurtle:
Noticed that I had inadvertently swapped the precedence order of cmdline vs config file. That's now resolved, and with it I've moved the user config file handling into `SoSComponent.load_options()` rather than in `_init_component()`.

However, I was not able to do this using the `merge()` function, due to how argparse passes parsed cmdline arguments, which was causing `merge()` to completely override the config file (effectively negating it). So, I've found the need to manually parse the non-default options out from the cmdline option object, and use that to update the `SoSOptions` instantiated and updated from the config files.

pmoravec:
> Implementation looks good. A bit tangential to this, but since we now have config file(s), presets, command line, I wonder if we should clarify in the docs the precedence for options? If I remember rightly it's conf < preset < command line?

+1 to make this documented. IMHO the conf < preset < cmdline is right (or maybe even preset < conf < cmdline, since preset is "product default", while conf is "system default", i.e. more specific settings?).

Current behaviour of the precedence is bit blurred to me:
1) having `/etc/sos/sos.conf`:

    [global]
    verbose = 3
    log-size = 1

    [report]
    log-size = 2

Then the `log-size` option is fully ignored (either I write the option to either section). `verbose = 3` is taken into account from either location (which is good), just the `log-size` from neither.

2) Mimicking I am on Satellite (by installing a dummy "satellite-common" package), then `sosreport --log-size=3` loads and uses `log-size=100` from the presets - i.e. the real precedence is cmdline < presets.

TurboTurtle:
> +1 to make this documented. IMHO the conf < preset < cmdline is right (or maybe even preset < conf < cmdline, since preset is "product default", while conf is "system default", i.e. more specific settings?).
> 

Done. This is now at the top of `man sos.conf`.
 
> Then the `log-size` option is fully ignored (either I write the option to either section). `verbose = 3` is taken into account from either location (which is good), just the `log-size` from neither.

Fixed. Had a typo that was causing the `arg_defaults` to be set in correctly and thus log-size was being ignored.
> 
> 1. Mimicking I am on Satellite (by installing a dummy "satellite-common" package), then `sosreport --log-size=3` loads and uses `log-size=100` from the presets - i.e. the real precedence is cmdline < presets.

Fixed. We now re-apply cmdline overrides after presets are loaded.

sos/__init__.py
```diff
@@ -20,6 +20,7 @@
 import sys
 
 from argparse import ArgumentParser
+from pathlib import Path
```
pmoravec, line 23: I think this is not further required import (while/though I am curious why Travis didnt detect it as superfluous).

sos/component.py
```diff
@@ -169,12 +204,13 @@ def load_options(self):
             if option.default != SUPPRESS:
                 option.default = None
 
-        # load values from cmdline
-        cmdopts = SoSOptions().from_args(self.parser.parse_args(self.cmdline))
-        opts.merge(cmdopts)
+        opts.update_from_conf(self.args.config_file, self.args.component)
+        if os.getuid() != 0:
```
pmoravec, line 208: Why `root` cant use its own config file as well? I am quite OK with this, just dont see much rationale (can't root require different defaults than system ones?).

sos/options.py
```diff
@@ -237,7 +239,7 @@ def merge(self, src, skip_default=True):
             if getattr(src, arg) is not None or not skip_default:
                 self._merge_opt(arg, src, False)
 
-    def dict(self):
+    def dict(self, preset_filter=True):
```
pmoravec, line 242: A nice catch to still apply `--preset` cmdline option.

man/en/sos.conf.5
```diff
@@ -51,19 +119,23 @@ threads=10
 .sp
 To disable the 'host' and 'filesys' plugins:
 .LP
-[plugins]
+[report]
 .br
-disable = host, filesys
+noplugins = host,filesys
```
pmoravec, line 124: This is renamed to `skip-plugins` else-where.

I see this renaming bit controversial, as 1) it is backward incompatible, 2) short option `-n` is not a shortcut from `--skip-plugins` (while we already use `-s` for `--sysroot`). So (not ultimate) +1 for keeping the old "noplugins" or could-be "no-plugins" option.

sos/collector/__init__.py
```diff
@@ -32,7 +33,7 @@
 from sos.component import SoSComponent
 from sos import __version__
 
-COLLECTOR_LIB_DIR = '/var/lib/sos-collector'
+COLLECTOR_LIB_DIR = '/etc/sos/groups.d'
```
pmoravec, line 36: `s/COLLECTOR_LIB_DIR/COLLECTOR_CONFIG_DIR/` or similar?

sos/collector/__init__.py
```diff
@@ -607,12 +614,16 @@ def write_host_group(self):
...
             'name': self.opts.save_group,
             'master': self.opts.master,
-            'cluster_type': self.cluster_type,
+            'cluster_type': self.cluster.cluster_type[0],
             'nodes': [n for n in self.node_list]
         }
-        if not os.path.isdir(COLLECTOR_LIB_DIR):
-            raise OSError("%s no such directory" % COLLECTOR_LIB_DIR)
-        fname = COLLECTOR_LIB_DIR + '/' + cfg['name']
+        if os.getuid() != 0:
```
pmoravec, line 620: Ditto on another place why `root` cant have its own `.config` (but again, I am ok with the current behaviour).

man/en/sos.conf.5
```diff
@@ -3,7 +3,75 @@
...
+Note that non-root users may override options set in /etc/sos/sos.conf by creating
+their own sos.conf under $HOME/.config/sos.
+
+The order in which options are loaded is as follows:
+
+  1. System configuration file at /etc/sos/sos.conf
+  2. User-specific configuration file at $HOME/.config/sos/sos.conf (for sos
+     components that support non-root)
+  3. In the case of running \fBsos report\fR, presets either automatically loaded
+     due to system configuration, or specified via \fB--add-preset\fR
```
pmoravec, line 18: .. specified via `--add-preset` or just `--preset` (also)? The later worked to me as well.

pmoravec:
Sorry for the iterative comments with such delays i between.. And thanks for great PR in general.

Few relics point to old config file location:

- https://github.com/sosreport/sos/blob/master/tests/simple.sh#L85-L91
- https://github.com/sosreport/sos/blob/master/sos/report/plugins/host.py#L33 (worth copying both old and new config file for legacy reasons, or is it obsolete and the new one is enough?)


ACK otherwise!

sos/component.py
TurboTurtle: I don't mind removing the restriction, but my rationale here was simply that it's not a common trend.

man/en/sos.conf.5
TurboTurtle: Whoops, that's a typo.

man/en/sos.conf.5
TurboTurtle: Hmm. I don't think we'll be able to avoid point #1, but as this is a major version update I'm also not too concerned about that. 

The problem I see is that both the short- and long-form name make sense in their own context. `-n` as a "do *n*ot use this" is fairly conventional. `skip-plugins` is also straight forward and intuitive.

However, a long form of `no-plugins` doesn't sound right to an English speaker. I wouldn't say "run sos with no plugins kvm, host, logs", and the first thing that comes to mind with `no-plugins` is "oh, this option means disable all plugins" which obviously isn't the case.

TurboTurtle:
Forgot to mention. This last push:

- Updates the `host` plugin, and adds a forbidden path on `default_mapping`
- Updates `simple.sh`
- Fixes the man page typo, and also cleans up the formatting a bit for the `extras.d/` section
- Changes `COLLECTOR_LIB_DIR` to `COLLECTOR_CONFIG_DIR`
- Updates `noplugins` to `skip-plugins` in the default `sos.conf`. I've left it as `skip-plugins` for the reasons I gave above, but I'm certainly open for counter arguments.

man/en/sos.conf.5
```diff
@@ -51,19 +120,23 @@ threads=10
...
+[plugin_options]
 .br
 rpm.rpmva = off
 .br
 .SH FILES
 .sp
-/etc/sos.conf
+/etc/sos/sos.conf
+.br
+$HOME/sos/sos.conf (optional)
```
pmoravec, line 137: Rather `$HOME/.config/sos/sos.conf` ?

tests/simple.sh
```diff
@@ -83,11 +83,12 @@ run_expecting_sucess () {
 }
 
 # If /etc/sos.conf doesn't exist let's just make it..
```
pmoravec, line 85: The comment lacks "sos" subdir (as well as "Creating .. few lines below").

sos/report/plugins/host.py
```diff
@@ -20,6 +20,8 @@ class Host(Plugin, RedHatPlugin, DebianPlugin):
 
     def setup(self):
 
+        self.add_forbidden_path('/etc/sos/default_mapping')
```
pmoravec, line 23: Great catch!

Just.. isn't the path `/etc/sos/cleaner/default_mapping` instead?

TurboTurtle: Yup, I forgot we moved that into the `cleaner` subdir. Updating these shortly.

pmoravec:
`sos_extras` plugin should replace `sos.extras.d` by `sos/extras.d` (in `short_desc` and namely in `extras_dir`).

pmoravec:
ACK to the PR, appart of the simple / final issues spotted now.

PR #2141, opened 2020-07-07
[leapp] Add preupgrade log to the leapp sos plugin

bocekm:
- `leapp` generates not only leapp-upgrade.log (when running `leapp upgrade`) but also leapp-preupgrade.log (when running `leapp preupgrade`)
- when customers send us a sos report after running `leapp preupgrade`, we don't have the log in the report archive and it makes the investigation of issues harded
- related report that is missing the log: https://bugzilla.redhat.com/show_bug.cgi?id=1846677#c3

---


bocekm:
@pmoravec, is it possible to make this patch appear in 7.9 and 8.3 (z-stream?)?

PR #2143, opened 2020-07-08
[sar] Extend command sadf to get more data out of sar

jcastill:
Currently, the sadf command translates to a XML
file the content of the CPU data captured in sar.
With this patch, all the data captured via sar
is translated into the XML file.
    
Resolves: #2143
    
--


jcastill:
I have two doubts here:
- Is it ok if this is a single patch, solving two problems?
- I've added 'xml' into the list of options, but it is the default, and as far as I know we always capture it. Is there a way to make this more elegant?

PR #2145, opened 2020-07-09
[containers_common] collect user-related commands outputs

pmoravec:
podman and buildah commands should be collected on per-user basis,
as declared by rootlessusers plugin option.

Resolves: #2145

---

### DIFF
--- a/sos/report/plugins/containers_common.py
+++ b/sos/report/plugins/containers_common.py
@@ -30,7 +30,6 @@ class ContainersCommon(Plugin, RedHatPlugin, UbuntuPlugin):
             '/etc/subuid',
             '/etc/subgid',
         ])
-        self.add_cmd_output(['loginctl user-status'])
 
         users_opt = self.get_option('rootlessusers')
         users_list = []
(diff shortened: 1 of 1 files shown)


sos/report/plugins/containers_common.py
```diff
@@ -30,27 +30,43 @@ def setup(self):
...
             self.add_copy_spec(
                 '%s/.config/containers/' % (os.path.expanduser('~%s' % user)))
-            # collect the user's podman/buildah info and uid/guid maps
-            for binary in ['/usr/bin/podman', '/usr/bin/buildah']:
-                for cmd in user_subcmds:
-                    self.add_cmd_output([
-                        'machinectl -q shell %s@ %s %s' % (user, binary, cmd)
-                    ], foreground=True)
+            # collect user-status
+            self.add_cmd_output(['loginctl user-status %s' % user])
```
TurboTurtle, line 65: Doesn't need to be a list here, or are we anticipating expanding this in the future?

sos/report/plugins/containers_common.py
```diff
@@ -30,27 +30,43 @@ def setup(self):
...
+            # collect the user's related commands
+            for cmd in user_subcmds:
+                self.add_cmd_output([
+                    'machinectl -q shell %s@ /usr/bin/%s' % (user, cmd)
+                ], foreground=True)
```
TurboTurtle, line 70: Same here, with this style this doesn't need to be a list. Or it could be refactored to something like

```python
self.add_cmd_output([
    "machinectl -q shell %s@ /usr/bin/%s" % (user, cmd) for cmd in user_subcmds
], foreground=True)
```

pmoravec:
Right, I just blindly copy-pasted the old style to the new code :). Fixed now.

TomSweeneyRedHat:
@rhatdan PTAL

rhatdan:
LGTM
I would also grab the contents of all files in $HOME/.config/containers/*

If they exists.

pmoravec:
@rhatdan : Just to confirm: those directories are not expected to contain some secrets we should obfuscate in sosreport, right?

rhatdan:
No they are just configuration files, default, potentially could have host names within them.

pmoravec:
(we realized that these dirs are already collected: https://github.com/sosreport/sos/blob/master/sos/report/plugins/containers_common.py#L62-L63)

PR #2151, opened 2020-07-13
[hyperv] Add plugin for Hyper-V clients

kamalca:
The hyperv plugin collects config files specific to Hyper-V and runs
commands for debugging the vmbus.

Signed-off-by: Kameron Carr <kameroncarrcollege@gmail.com>

---

### DIFF
--- a/sos/report/plugins/hyperv.py
+++ b/sos/report/plugins/hyperv.py
@@ -1,0 +1,28 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+
+
+class Hyperv(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
+    """Hyper-V client information"""
+
+    plugin_name = "hyperv"
+    files = ('/sys/bus/vmbus/',)
+
+    def setup(self):
+
+        self.add_copy_spec([
+            "/sys/bus/vmbus/drivers/",
+            # copy devices/*/* instead of devices/ to follow link files
+            "/sys/bus/vmbus/devices/*/*"
+        ])
+
+        self.add_cmd_output("lsvmbus -vv")
+
+# vim: set et ts=4 sw=4 :


sos/report/plugins/hyperv.py
```diff
@@ -0,0 +1,30 @@
...
+# See the LICENSE file in the source distribution for further information.
+
+from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+
+
+class Hyperv(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
+    """Hyper-V client information"""
+
+    plugin_name = "hyperv"
+    files = ('/sys/bus/vmbus/', '/boot/config-*')
```
TurboTurtle, line 16: Activating on `/sys/bus/vmbus` is fine (as far as I can tell anyways), but specifying `/boot/config-*` here will enable this on literally every modern Linux installation.

sos/report/plugins/hyperv.py
```diff
@@ -0,0 +1,30 @@
...
+
+        self.add_copy_spec([
+            "/sys/bus/vmbus/drivers/",
+            # copy devices/*/* instead of devices/ to follow link files
+            "/sys/bus/vmbus/devices/*/*",
+            "/boot/config-*"
+        ])
+
+        self.add_cmd_output("lsvmbus -vv")
+        self.add_cmd_output("lspci -vv")
```
TurboTurtle, line 28: Nack to this. It is already collected by the `pci` plugin.

sos/report/plugins/hyperv.py
```diff
@@ -0,0 +1,30 @@
...
+    plugin_name = "hyperv"
+    files = ('/sys/bus/vmbus/', '/boot/config-*')
+
+    def setup(self):
+
+        self.add_copy_spec([
+            "/sys/bus/vmbus/drivers/",
+            # copy devices/*/* instead of devices/ to follow link files
+            "/sys/bus/vmbus/devices/*/*",
+            "/boot/config-*"
```
TurboTurtle, line 24: I'm not sure that the `config-` files should be collected, but in the event they should be, the `boot` plugin would be the place for this collection.

sos/report/plugins/hyperv.py
TurboTurtle, line 16: What about enabling the plugin on presence of the `hyperv-tools` package? That appears to be what provides `lsvmbus` (at least on Fedora, would need to verify package names on the other distros). This would also allow for package verification when sos is run with `--verify`.

TurboTurtle: Following on this - are there any other commands that ship with the `hyperv-tools` package that should be collected, beyond `lsvmbus`?

sos/report/plugins/hyperv.py
bmr-cymru: Those files are the build options for the installed kernels, so if they belong anywhere it's with kernel data rather than boot data (they are installed to `/boot` so that they are available for inspection with the kernel and initramfs images they belong to but from a support stance they are closer to the kernel).

sos/report/plugins/hyperv.py
bmr-cymru: Bizarrely the package seems to include exactly one file, the `lsvmbus` binary:

```
$ rpm -ql hyperv-tools
/usr/sbin/lsvmbus
```

bmr-cymru: Ah, ok - it's a sub-package of the `hyperv-daemons` SRPM.

kamalca: On Ubuntu I believe `lsvmbus` is included in the `linux-cloud-tools-generic` package. I would like the plugin to collect the `/sys/bus/vmbus` files even if `lsvmbus` is unavailable. Also there are not very many cases of a Hyper-V guest not having the `/sys/bus/vmbus` directory, so the case of a machine having `lsvmbus` and not `/sys/bus/vmbus` may not be important to consider.

sos/report/plugins/hyperv.py
kamalca: Thank you for the tip! Is there a reason this isn't currently being collected? It seems to contain a lot of information about the kernel configuration.

sos/report/plugins/hyperv.py
```diff
@@ -0,0 +1,28 @@
...
+# See the LICENSE file in the source distribution for further information.
+
+from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+
+
+class Hyperv(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
+    """Hyper-V client information"""
+
+    plugin_name = "hyperv"
+    files = ('/sys/bus/vmbus/')
```
TurboTurtle, line 16: This needs to be a tuple, even if it's a 1-tuple. Otherwise the enablement check will iterate over this as a string, and you end up with a check to see if `/` exists.

sos/report/plugins/hyperv.py
bmr-cymru: Mostly that they are static files: they are installed by the kernel RPM package (on Red Hat), and they represent the options used to build the kernel. They cannot be changed without rebuilding the kernel so as long as you know the kernel version you can always get the config-* content without needing to collect it from the machine. If the kernel has been rebuilt then we can detect that in other ways and the content of the file is likely not up to date anyway.

andschwa: Hey Bryn, I looked into this a little bit. From the context of kernel developers debugging a VM issue, having the current kernel configurations is very useful. You are right that `/boot/config-*` may be out of date, may have many old configs (from the kernel package getting updated), and could be obtained manually if you know the kernel package installed. However, I would posit that it would be very useful for sosreport to opportunistically collect the _current_ kernel's build configs, which _can be_ available at `/proc/config.gz`. This file being exposed by the kernel is dependent on the kernel having been built with the config module enabled, but may we propose collecting it when it's available in the sosreport kernel plugin?

sos/report/plugins/hyperv.py
andschwa: Jake, to your question about other tools in `hyperv-tools` (and `linux-cloud-tools-generic` etc.), `lsvmbus` is the only one which provides useful information.

sos/report/plugins/hyperv.py
bmr-cymru: Getting the compiled-in configuration is different, because it's strongly bound to the kernel image - but - the last time I looked it was not enabled on many distros. At least historically Red Hat (inc. Fedora/RHEL/CentOS), Debian, and Ubuntu don't enable the CONFIG_IKCONFIG_PROC option that would cause this file to be available.

These distros all provide the `/boot/config-*` files as an alternative.

andschwa:
Hi @TurboTurtle, I was wondering now that it's approved when this will be merged? Thanks!

TurboTurtle:
@andschwa - `sos` operates on a 2-ack process from the maintainers. Now that @pmoravec has also ack'd, I'll get this merged later today.

andschwa:
Fantastic, thank you both so much!

PR #2161, opened 2020-07-22
[pmem] Updated for ipmctl v2.x

sscargal:
ipmctl version 2.x introduces new commands and options.
It also deprecates some older commands.

Signed-off-by: Steve Scargall <steve.scargall@intel.com>

---

- [] If this commit resolves an existing pull request, is the line `Resolves: #PRNUMBER` included in an independent line?

### DIFF
--- a/sos/report/plugins/pmem.py
+++ b/sos/report/plugins/pmem.py
@@ -22,11 +22,12 @@ class pmem(Plugin, RedHatPlugin, UbuntuPlugin, DebianPlugin):
     commands = ('ndctl', 'daxctl', 'ipmctl')
 
     def setup(self):
-        # Copy the contents of the /etc/ndctl directory
-        # and /etc/ipmctl.conf file
+        # Copy the contents of the /etc/ndctl & /var/log/ipmctl
+        # directories and /etc/ipmctl.conf file
         self.add_copy_spec([
             "/etc/ndctl",
-            "/etc/ipmctl.conf"
+            "/etc/ipmctl.conf",
+            "/var/log/ipmctl"
         ])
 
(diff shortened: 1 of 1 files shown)


TurboTurtle:
Does version 1.x still ship on modern distributions?

If not, then it's fine to drop the older commands, but if version 1 is still present on current RHEL/Ubuntu/etc versions, then we'll need to keep the older commands. Perhaps we can do a version check and call the version-specific commands based on that.

sscargal:
@TurboTurtle Version 1 was never shipped in the package repos. It has to be built manually. Version 2 is available in most Linux package repos. If you can point me to an example of supporting multiple binary versions, I'll implement it as I believe it's the cleanest solution.

TurboTurtle:
Generally version checks are done via some package manager checks, e.g. the `etcd` plugin:

```python
ver = self.policy.package_manager.get_pkg_list()['etcd']
ver = ver['version'][0]
```

TurboTurtle:
Cycling around on this as we're aiming to close 4.0 in about 11 days.

Since version 1 was never formally packaged, I'm not opposed to keeping PR as it is. I'll leave it up to you if you want to include the version 1 commands.

TurboTurtle:
Cycling around on this. @sscargal - any desire for updating this for both versions, or do you prefer to keep it as-is?

sscargal:
@TurboTurtle - We can use this version. Version 1.x of ipmctl never shipped and won't ship. It was only ever available if people built it from the source. I'll implement support for multiple versions when v3.x ships.

pmoravec (approved):

Issue #2169, opened 2020-07-28
[drbd] plugin doesn't get triggered in Ubuntu


slashdd:
The drbd plugin doesn't get triggered in ubuntu with the current Packages regex in place:
`packages = ('drbd.*-utils',)`

nniehoff:
I can confirm the plugin does not get triggered

PR #2170, opened 2020-07-28
[drbd] Add drbd-utils to the package triggering list

slashdd:
Closes: #2169

---

### DIFF
--- a/sos/report/plugins/drbd.py
+++ b/sos/report/plugins/drbd.py
@@ -15,7 +15,7 @@ class drbd(Plugin, RedHatPlugin, UbuntuPlugin):
 
     plugin_name = 'drbd'
     profiles = ('storage')
-    packages = ('drbd.*-utils',)
+    packages = ('drbd*-utils',)
 
     def setup(self):
         self.add_cmd_output([


TurboTurtle:
What about correcting the regex to make the version optional in the package name? E.G. `drbd(.*)?-utils`

slashdd:
I have explored the exact same regex `drbd(.*)?-utils` without sucess. I'll try to explore some more regexp pattern.

slashdd:
Look like `packages =` doesn't support regex but `verify_packages = `does.

slashdd:
With d853f4a, the drbd plugin now gets executed:

```
# ls -altr sos_commands/drbd/
total 20
-rw-r--r--  1 root root   68 Jul 30 20:40 drbdadm_dump-xml
-rw-r--r--  1 root root 1508 Jul 30 20:40 drbdsetup_status
drwx------  2 root root 4096 Jul 30 20:40 .
-rw-r--r--  1 root root 1508 Jul 30 20:40 drbdsetup_show
drwxr-xr-x 49 root root 4096 Jul 30 20:40 ..
```

TurboTurtle:
Oh, it looks like we are ultimately calling `PackageManager.pkg_by_name()` which in turn uses `all_pkgs_by_name()` instead of `all_pkgs_by_name_regex()`. The former of which uses `fnmatch.filter()`, which uses Unix shell-style wildcards but not actual regexes.

So, in lieu of changing what `pkg_by_name()` calls (I'll make note to investigate that), we should be fine with simply changing the `.*` to `*`, e.g.:

```python
packages = ('drbd*-utils',)
```

slashdd:
yep it did the trick.

PR #2172, opened 2020-07-28
[docs] Update sphinx docs and standardize docstrings

TurboTurtle:
This set of patches updates the Sphinx documentation setup, provides for building the docs via `setuptools` now instead of `make`, (but makes it optional to not add a `buildrequires` in the spec file), and then standardizes the docstrings for several of the larger parts of the sos API.

---


docs/conf.py
```diff
@@ -59,9 +59,9 @@
 # built documents.
 #
 # The short X.Y version.
-version = '3.2'
+version = '3.9'
 # The full version, including alpha/beta/rc tags.
-release = '3.2'
+release = '3.9.1'
```
BryanQuigley, line 64: did you want this to be 4.0?

TurboTurtle: Not until we close the release, so when we do a bump commit we'd just need to include this version here.

BryanQuigley:
Not familiar with Sphinx - where should the generated docs end up?

As it's conditional, can we output that Sphinx is supposed to run or not?

TurboTurtle:
The docs should eventually end up on readthedocs, however I'll need to track down who has those credentials since I think things are broken on that end as well - this PR at least fixes the build and design issues on our side.

TurboTurtle:
> As it's conditional, can we output that Sphinx is supposed to run or not?

Added a warning when sphinx is not installed locally that docs won't be generated.

PR #2180, opened 2020-08-03
[ceph_podman] Add new plugin for containerized ceph

Meiyan89:
In RHOSP16 with integrated containerized ceph environment, as ceph
command is not installed on host so we need to run ceph command in
ceph-mon or ceph-osd containers.

This new plugin add ability to collect ceph information inside of
podman ceph containers.

Signed-off-by: Meiyan Zheng <mzheng@redhat.com>

---

### DIFF
--- a/sos/report/plugins/ceph_podman.py
+++ b/sos/report/plugins/ceph_podman.py
@@ -1,0 +1,110 @@
+# Copyright (C) 2020 Meiyan Zheng <mzheng@redhat.com>
+#
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, RedHatPlugin
+import re
+
+
+class CephPodman(Plugin, RedHatPlugin):
+    short_desc = 'Containerized Ceph distributed storage'
+    plugin_name = 'ceph_podman'
+    profiles = ('storage', 'container')
+    packages = ('podman',)
+
+    def setup(self):
+        self._container_mon_name = self.get_container_by_name('ceph-mon-*')
+        self._container_osd_name = self.get_container_by_name('ceph-osd-*')
+
+        self.add_copy_spec([
+            "/etc/ceph/",
+            "/var/lib/ceph/",
+            "/run/ceph/"
+        ])
+
+        if self._container_mon_name:
+            cmds = [
+                'ceph mon stat',
+                'ceph mon_status',
+                'ceph quorum_status',
+                'ceph mgr module ls',
+                'ceph mgr metadata',
+                'ceph osd metadata',
+                'ceph osd erasure-code-profile ls',
+                'ceph report',
+                'ceph osd crush show-tunables',
+                'ceph versions',
+                'ceph features',
+                'ceph osd crush dump',
+                'ceph -v',
+                'ceph crash stat',
+                'ceph crash ls',
+                'ceph config log',
+                'ceph config generate-minimal-conf',
+                'ceph config-key dump',
+                'ceph status',
+                'ceph health detail',
+                'ceph osd tree',
+                "ceph osd stat",
+                "ceph osd df tree",
+                "ceph osd dump",
+                "ceph osd df",
+                "ceph osd perf",
+                "ceph osd blocked-by",
+                "ceph osd pool ls detail",
+                "ceph osd numa-status",
+                "ceph device ls",
+                "ceph mon dump",
+                "ceph mgr dump",
+                "ceph mds stat",
+                "ceph df",
+                "ceph df detail",
+                "ceph fs ls",
+                "ceph fs dump",
+                "ceph pg dump",
+                "ceph pg stat",
+            ]
+            container_cmds = self.to_container_cmd(cmds,
+                                                   self._container_mon_name)
+            self.add_cmd_output(container_cmds, foreground=True)
+
+        if self._container_osd_name:
+            cmds = [
+                'ceph -v',
+                'ceph-volume lvm list',
+            ]
+            osd_containers = self.get_all_containers_by_name('ceph-osd-*')
+            for c in osd_containers:
+                container_cmds = self.to_container_cmd(cmds, c)
+                self.add_cmd_output(container_cmds, foreground=True)
+
+        self.add_forbidden_path([
+            "/etc/ceph/*keyring*",
+            "/var/lib/ceph/*keyring*",
+            "/var/lib/ceph/*/*keyring*",
+            "/var/lib/ceph/*/*/*keyring*",
+            "/var/lib/ceph/osd",
+            "/var/lib/ceph/mon",
+            "/var/lib/ceph/tmp/*mnt*",
+            "/etc/ceph/*bindpass*"
+        ])
+
+    def to_container_cmd(self, cmds, container_name):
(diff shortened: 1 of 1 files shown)


sos/report/plugins/ceph_podman.py
```diff
@@ -0,0 +1,110 @@
...
+            "/var/lib/ceph/mon",
+            "/var/lib/ceph/tmp/*mnt*",
+            "/etc/ceph/*bindpass*"
+        ])
+
+    def to_container_cmd(self, cmds, container_name):
+        container_cmds = []
+        for cmd in cmds:
+            container_cmds.append('podman exec %s %s' % (container_name, cmd))
+        return container_cmds
```
TurboTurtle, line 102: There is already a `Policy().format_container_cmd()` method that does this. We could also expose this via a wrapper in `Plugin` if need be.

sos/report/plugins/ceph_podman.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def get_all_containers_by_name(self, name):
+        all_containers = self.get_containers(get_all=True)
+        containers_by_name = []
+        for c in all_containers:
+            if re.match(name, c[1]):
+                containers_by_name.append(c[1])
+        return containers_by_name
```
TurboTurtle, line 110: This can all be condensed down to

```python
return [c for c in self.get_containers() if re.match(name, c[1])]
```

Also, be aware that using `get_all=True` will return terminated containers as well, which you won't be able to run commands inside of.

sos/report/plugins/ceph_podman.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def setup(self):
+        self._container_mon_name = self.get_container_by_name('ceph-mon-*')
+        self._container_osd_name = self.get_container_by_name('ceph-osd-*')
+
+        self.add_copy_spec([
+            "/etc/ceph/",
+            "/var/lib/ceph/",
+            "/run/ceph/"
+        ])
```
TurboTurtle, line 29: These would still be captured from the host, not inside the container. I'm not sure if that's what you want here or not though.

sos/report/plugins/ceph_podman.py
```diff
@@ -0,0 +1,110 @@
...
+        self.add_forbidden_path([
+            "/etc/ceph/*keyring*",
+            "/var/lib/ceph/*keyring*",
+            "/var/lib/ceph/*/*keyring*",
+            "/var/lib/ceph/*/*/*keyring*",
+            "/var/lib/ceph/osd",
+            "/var/lib/ceph/mon",
+            "/var/lib/ceph/tmp/*mnt*",
+            "/etc/ceph/*bindpass*"
+        ])
```
TurboTurtle, line 96: Same as the `add_copy_spec()` above - this will still apply to the host and not inside a particular container.

sos/report/plugins/ceph_podman.py
```diff
@@ -0,0 +1,110 @@
...
+
+from sos.report.plugins import Plugin, RedHatPlugin
+import re
+
+
+class CephPodman(Plugin, RedHatPlugin):
+    short_desc = 'Containerized Ceph distributed storage'
+    plugin_name = 'ceph_podman'
+    profiles = ('storage', 'container')
+    packages = ('podman',)
```
TurboTurtle, line 19: This would cause non-ceph nodes that have `podman` installed to always run this plugin.

sos/report/plugins/ceph_podman.py
Meiyan89: Those files are mounted from host to container, so I guess it's fine to capture those files from host: 

# podman inspect ceph-mon-overcoud-controller-0 | jq .[].HostConfig.Binds
[
  "/var/lib/ceph:/var/lib/ceph:rw,rprivate,rbind",
  "/etc/ceph:/etc/ceph:rw,rprivate,rbind",
  "/var/run/ceph:/var/run/ceph:rw,rprivate,rbind",
  "/etc/localtime:/etc/localtime:ro,rprivate,rbind",
  "/var/log/ceph:/var/log/ceph:rw,rprivate,rbind",
  "/etc/pki/ca-trust/extracted:/etc/pki/ca-trust/extracted:rw,rprivate,rbind"
]

# podman inspect ceph-mgr-overcloud-controller-0 | jq .[].HostConfig.Binds
[
  "/var/log/ceph:/var/log/ceph:rw,rprivate,rbind",
  "/var/lib/ceph:/var/lib/ceph:rw,rprivate,rbind",
  "/etc/ceph:/etc/ceph:rw,rprivate,rbind",
  "/var/run/ceph:/var/run/ceph:rw,rprivate,rbind",
  "/etc/localtime:/etc/localtime:ro,rprivate,rbind"
]

sos/report/plugins/ceph_podman.py
Meiyan89: Same as self.add_copy_spec above.

sos/report/plugins/ceph_podman.py
Meiyan89: I found fmt_container_cmd() will return a command with quote: 

    def fmt_container_cmd(self, container, cmd):
        return "%s %s %s" % (self.run_cmd, container, quote(cmd))

But podman is not working with a quote: 

For example: 
~~~
[root@overcloud-controller-0 tmp]# podman exec  ceph-mon-overcoud-controller-0 'ceph mon stat'
Error: exec failed: container_linux.go:346: starting container process caused "exec: \"ceph mon stat\": executable file not found in $PATH": OCI runtime command not found error
[root@overcloud-controller-0 tmp]# podman exec  ceph-mon-overcoud-controller-0 ceph mon stat
e1: 3 mons at {overcloud-controller-0=[v2:172.24.20.69:3300/0,v1:172.24.20.69:6789/0],overcloud-controller-1=[v2:172.24.20.97:3300/0,v1:172.24.20.97:6789/0],overcloud-controller-2=[v2:172.24.20.49:3300/0,v1:172.24.20.49:6789/0]}, election epoch 34, leader 0 overcloud-controller-2, quorum 0,1,2 overcloud-controller-2,overcloud-controller-0,overcloud-controller-1
~~~

I'm not sure which one is proper, re-write fmt_container_cmd()  or override this function in ceph_podman.

sos/report/plugins/ceph_podman.py
Meiyan89: Thank you the suggestion, and pointing out `get_all=True`, will modify this part.

sos/report/plugins/ceph_podman.py
TurboTurtle: Seems we have a schism as there is both a `format_container_command()` used by `collect`, and then `fmt_container_cmd()` used by plugin. 

I believe the right thing to do here would be to move the former into `ContainerRuntime`, replacing the latter. We can still expose it via `Policy` as well.

TurboTurtle:
Again, I'm not sold on this being a separate plugin. I'd prefer if these changes were merged into the standard `ceph` plugin, to simply detect if the service(s) is/are running in containers and then format the commands appropriately if so - most of the openstack plugins do this today if you need a reference.

pmoravec:
+1 for enhancing `ceph` plugin "only". There is a big overlap in commands taken (once directly, here in a container) - it would be much easier to maintain *one* list of the commands, if there will be some change in it.

I can help merging this proposed plugin into `ceph` one, if welcomed - let me know.

pmoravec:
Bump with an idea: wouldn't it be easier to make `ceph` plugin enabled by default for containerised ceph environment as well? (and if so, how to identify such environment / container?)

TurboTurtle:
Closing this out due to #2646 being merged a while back.

Issue #2181, opened 2020-08-03
[plugin] Missing OpenStack Octavia etc .conf files


cgoncalves:
Some deployments may have multiple .conf files under var_puppet_gen + "/etc/octavia/*.conf".

PR #2182, opened 2020-08-03
[openstack_octavia] Copy all etc .conf files

cgoncalves:
Resolves: #2181

---


cgoncalves:
> Looks good to me. Should we also grab the policy configuration overrides?

Done

PR #2185, opened 2020-08-03
[openstack_octavia] Collect new API resources

cgoncalves:
---

### DIFF
--- a/sos/report/plugins/openstack_octavia.py
+++ b/sos/report/plugins/openstack_octavia.py
@@ -19,6 +19,20 @@ class OpenStackOctavia(Plugin):
     var_config_data = "/var/lib/config-data"
     var_puppet_gen = var_config_data + "/puppet-generated/octavia"
 
+    resources = [
+        'amphora',
+        'availabilityzone',
+        'availabilityzoneprofile',
+        'flavor',
+        'flavorprofile',
+        'healthmonitor',
+        'l7policy',
+        'listener',
+        'pool',
+        'provider',
+        'quota'
+    ]
+
     def setup(self):
         # configs
         self.add_copy_spec([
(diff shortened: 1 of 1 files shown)


sos/report/plugins/openstack_octavia.py
```diff
@@ -60,7 +66,16 @@ def setup(self):
         for loadbalancer in loadbalancers.splitlines():
             loadbalancer = loadbalancer.split()[0]
             self.add_cmd_output(
-                "openstack loadbalancer show %s" % (loadbalancer)
```
johnsom, line 63: This is a big change. The status tree is useful for sure, but by not capturing the load balancer configuration (flavor, az, etc.) I think we are losing a lot of useful information. I'm on this fence on this one.

cgoncalves:
> This is a big change. The status tree is useful for sure, but by not capturing the load balancer configuration (flavor, az, etc.) I think we are losing a lot of useful information. I'm on this fence on this one. 

Good point. My first thought is always that "status show" dumps everything which is not true. Do you think it is best to just do a "loadbalancer show" or both show and status?

sos/report/plugins/openstack_octavia.py
cgoncalves: Good point. My first thought is always that "status show" dumps everything which is not true. Do you think it is best to just do a "loadbalancer show" or both show and status?

cgoncalves: I reverted this change back to "loadbalancer show". This output combined with other resource show commands should provide all needed information.

sos/report/plugins/openstack_octavia.py
```diff
@@ -63,6 +69,15 @@ def setup(self):
...
 
+        # get details from each amphora
+        cmd = "openstack loadbalancer amphora list -f value -c id"
+        amphorae = self.exec_cmd(cmd)['output']
+        for amphora in amphorae.splitlines():
+            amphora = amphora.split()[0]
+            self.add_cmd_output(
+                "openstack loadbalancer amphora show %s" % (amphora)
+            )
+
```
TurboTurtle, line 80: This gets repeated a number of times. Let's consolidate this and make it easier to add more resources in the future. We could do something like the `kubernetes` and `openshift` plugins here, E.G.:


```python
resources = [
    'amphora',
    'flavor',
    'flavorprofile'
   ... etc ...
]

for res in resources:
    cmd = "openstack loadbalancer %s list -f value -c id" % res
    ret = self.exec_cmd(cmd)
    if ret['status'] == 0:
        for ent in ret['output'].splitlines():
            ent = ent.split()[0]
            self.add_cmd_output(
                "openstack loadbalancer %s show %s" % (res, ent)
            )
```

For the `add_cmd_output()`, you could also pass `subdir=res` to get these outputs sorted by resource in `sos_commands/openstack_octavia/$res`

cgoncalves: Agreed. I've addressed this comment. I also fixed one other issue I found with "provider capability list".

PR #2188, opened 2020-08-06
[networking] remove 'ethtool -e' option while running sosreport

suresh2514:
Running NIC EEPROM dump (ethtool -e) can result in hardware pause
and is not recommended for production environment

Signed-off-by: suresh2514 <suresh2514@gmail.com>

---

### DIFF
--- a/sos/report/plugins/networking.py
+++ b/sos/report/plugins/networking.py
@@ -183,7 +183,6 @@ class Networking(Plugin):
                 "ethtool -a " + eth,
                 "ethtool -c " + eth,
                 "ethtool -g " + eth,
-                "ethtool -e " + eth,
                 "ethtool -P " + eth,
                 "ethtool -l " + eth,
                 "ethtool --phy-statistics " + eth,


bmr-cymru:
Do you have a reference to support the claim that this is generally unsafe and should not be used in production? There's no mention of such a risk in the `ethtool` documentation or the kernel `ethtool.h` API docs:

```
/**
 * struct ethtool_eeprom - EEPROM dump
 * @cmd: Command number = %ETHTOOL_GEEPROM, %ETHTOOL_GMODULEEEPROM or
 *      %ETHTOOL_SEEPROM
 * @magic: A 'magic cookie' value to guard against accidental changes.
 *      The value passed in to %ETHTOOL_SEEPROM must match the value
 *      returned by %ETHTOOL_GEEPROM for the same device.  This is
 *      unused when @cmd is %ETHTOOL_GMODULEEEPROM.
 * @offset: Offset within the EEPROM to begin reading/writing, in bytes
 * @len: On entry, number of bytes to read/write.  On successful
 *      return, number of bytes actually read/written.  In case of
 *      error, this may indicate at what point the error occurred.
 * @data: Buffer to read/write from
 *
 * Users may use %ETHTOOL_GDRVINFO or %ETHTOOL_GMODULEINFO to find
 * the length of an on-board or module EEPROM, respectively.  They
 * must allocate the buffer immediately following this structure.
 */
```

If you are seeing problems with one or more buggy drivers then there are more appropriate ways to address that than by removing the command entirely.

suresh2514:
I have  Red Hat cases where customer faces connection issue intermittently or connection totally lost whenever sosreport is executed.  The issue further tracked down to  'ethtool -e' command where strace confirms ioctl returns after 3.4 seconds.  

+++
26621 1595999211.585459 socket(AF_INET, SOCK_DGRAM, IPPROTO_IP) = 3<UDP:[21070872]> <0.000016>
26621 1595999211.585748 ioctl(3<UDP:[21070872]>, SIOCETHTOOL, 0x7fffffffe530) = 0 <0.000026>
26621 1595999211.586002 mmap(NULL, 2101248, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7ffff750a000 <0.000012>
26621 1595999211.587348 ioctl(3<UDP:[21070872]>, SIOCETHTOOL, 0x7fffffffe530 <unfinished ...>
26621 1595999215.032016 <... ioctl resumed> ) = 0 <3.444488>  <<<<<
+++


As of now I can dig out two similar cases. Their NIC details given below:

1.  (RHEL 7)
eth0   0000:04:00.0  link=up 1000Mb/s full (autoneg=Y)  rx ring 453/4078  drv bnx2x v1.713.36-0 storm 7.13.1.0 / fw mbi 7.15.64 bc 7.14.62

2.  (RHEL 8)
em3         0000:01:00.2  link=up 1000Mb/s full (autoneg=Y)   rx ring 128/4078  drv bnx2x v1.713.36-0 storm 7.13.1.0 / fw FFV15.00.14 bc 7.14.12

Not sure how many more are going to be affected as  'ethtool -e' option is recently added to sosreport 

Also refers:
+++
https://www.dell.com/support/manuals/in/en/inbsdt1/red-hat-entps-lx-v7.0/rhel_7.7_rn/reading-eeprom-from-a-broadcom-device-via-ethtool-results-in-soft-lockup?guid=guid-986ca2a9-c9f9-4345-8762-02d286cc0d1f&lang=en-us
+++

bmr-cymru:
Have you reported this problem when running `ethtool -e` to Red Hat?

bmr-cymru:
From the description in the linked article this sounds very strongly like a driver bug:

```
    Ensure that cable is connected to the specific port before reading EEPROM on that port. 
```

If calling an `ioctl` causes a soft lockup that is a driver bug - in this case it appears to be behaviour that is tied to the link state of the port.

The way to deal with driver bugs is to report them so that they can be fixed. Patching other software to avoid the problem is just papering over the cracks and the `ethtool -e` command was added in 2018 at the request of Red Hat support.

We have the ability to control command execution by predicates, which can look at things like software versions and the set of loaded kernel modules in order to decide whether or not to run a particular command. If the problem you describe is so widespread in the field then that is the way to handle it in sos - not by just dropping the command wholesale.

suresh2514:
Hi Bryn,
thanks for the feedback.

I am from RedHat only.
As of now I see three cases reported for Broadcom bnx2x nics.  I have suggested customer to check with hardware vendor and confirm how safe it is to dump EEPROM while in production.

We also have a Bugzilla opened for same. https://bugzilla.redhat.com/show_bug.cgi?id=1846708

regards

bmr-cymru:
Yes: we moved the bug over to the kernel component a couple of months ago because the consensus was that this is a kernel/driver or hardware problem. As I understand things that is still the case.

As I mentioned we can selectively disable collection of this command (and log some message indicating the reason) if certain criteria are met (we use the `SoSPredicate` as the general mechanism for this kind of gating).

If a fix for the problem on the kernel or hardware/firmware side is going to be a long time coming then this is the kind of change we would make in sos to protect affected users while still obtaining the diagnostic information from other unaffected systems.

jbainbri:
We can't discuss that bug publicly here, but it's a different issue to what this PR resolves.

Let's talk about what we're trying to fix on this PR.

We observe that the `bnx2x` driver genuinely blocks for up to 12 seconds while commands sent to the driver complete. See `be_mcc_wait_compl()`. If you ping the system while this is happening, the NIC won't respond.

The ethtool EEPROM collection is one such command which causes this hang. As far as I can see reading the driver, this is the expected operation. You cannot run an ethtool EEPROM collection on `bnx2x` and keep the NIC responsive.

That doesn't seem to match with sos being "production safe".

As far as I know, the `bnx2x` device type is no longer being actively improved by its current intellectual property owner. A significant rewrite of the driver and firmware to solve this (perhaps by turning those commands into an async non-blocking event) probably won't ever happen.

If you wish to make a `SoSPredicate` to **not** run `ethtool -e` when the `bnx2x` driver is loaded, that appears to be a good resolution to the problem we're trying to solve here. Are you okay with that?

bmr-cymru:
> The ethtool EEPROM collection is one such command which causes this hang. As far as I can see reading the driver, this is the expected operation. You cannot run an ethtool EEPROM collection on `bnx2x` and keep the NIC responsive.

Then perhaps this ought to be documented somewhere that users are likely to find it.

> That doesn't seem to match with sos being "production safe".

There's no need for jibes: I suggested a way to change sos almost a week ago that would do exactly that. The project can address problems that are brought to its attention but we can't be held responsible for other people's bad design or implementation decisions.

> If you wish to make a `SoSPredicate` to **not** run `ethtool -e` when the `bnx2x` driver is loaded, that appears to be a good resolution to the problem we're trying to solve here. Are you okay with that?

I don't wish to do anything - I just commented as a former maintainer that the predicate mechanism is normally the correct way to handle a problem of this type. It's up to the current maintainer, @TurboTurtle and the rest of the team to decide what to do with this PR, but since we added the command at the request of Red Hat support it certainly seems to warrant a bit more discussion before we pull it out again due to a single buggy NIC implementation.

mheler:
It is not safe to pull eeprom from Dell network adapters per 

https://www.dell.com/support/manuals/us/en/04/red-hat-entps-lx-v7.0/rhel_7_rn_pub/reading-eeprom-from-a-broadcom-device-via-ethtool-results-in-soft-lockup?guid=guid-986ca2a9-c9f9-4345-8762-02d286cc0d1f&lang=en-us

With that in mind, this flag needs to be removed or have an easier option of being disabled by the end user.

bmr-cymru:
> With that in mind, this flag needs to be removed or have an easier option of being disabled by the end user.

The `sos` tool has a framework for disallowing certain commands based on the properties of the host system it is running on, including what kernel modules are loaded. This is the `SoSPredicate` mechanism.

To address this someone just needs to create a predicate with a list of possibly broken NIC driver kernel modules, and use it to gate the `ethtool -e` execution (and possibly other commands - @jbainbri alluded to there being more potentially unsafe ioctl commands but if that's the case the `sos` team needs to know specifically what those are).

I would look at this myself but I am no longer the maintainer of `sos` and I do not currently have time due to other project commitments. I have already asked the current maintainer @TurboTurtle for his review of this request.

TurboTurtle:
I'm in agreement with @bmr-cymru here - I don't like the idea of pulling out a command due to a single buggy driver.

As has been stated, using a `SoSPredicate` to gate `ethtool -e` collection would be the best workaround for us, while still allowing this collection for non-buggy drivers. This could be as simple as testing for the presence of the `bnx2x` driver, but that would also prevent EEPROM collection from non-problematic NICs, or could be as specific as testing each NIC individually for both the driver loaded and that the NIC has a cable attached - since it appears a bnx2x NIC that does have a cable attached does not present this issue.

 If you'd like to take on updating this PR to use a `SoSPredicate` in that way, you're welcome to do so. If not, I can find some time to work up a solution but I can't make any guarantees as to when that would be complete.

bmr-cymru:
> since it appears a bnx2x NIC that does have a cable attached does not present this issue.

It seems there are multiple issues with EEPROM dump on the `bnx2` devices - as well as possibly other ioctl commands (@jbainbri ?). We really need to get as full a list as possible of both affected drivers/devices and calls so that we can tackle this in one go.

jbainbri:
I had deleted my comment as I wanted to do more research. Sorry, I should have made that clearer.

> I don't like the idea of pulling out a command due to a single buggy driver.

As far as I can tell, there is no driver bug. `bnx2x` may hang for up to 12 seconds when the driver sends some messages to the device. That's the normal expected operation. So it is not production-safe to run `ethtool -e` on a `bnx2x` device.

That's a completely separate thing to the Dell no-cable iDRAC hang described by Matt. That issue does appear to be a Dell firmware problem.

> If you'd like to take on updating this PR to use a `SoSPredicate` in that way, you're welcome to do so.

I agree. Good resolution. Let's do that. Thank you!

> as well as possibly other ioctl commands (@jbainbri ?)

The report we have only shows a disruptive hang when asking the device to provide EEPROM. The other commands which sos currently runs seem fine, they return quickly enough that a disruption is not observed.

jbainbri:
New PR at https://github.com/sosreport/sos/pull/2200

Looks like this is okay to close.

mheler:
This is being seen on tg3 based cards as well.

PR #2189, opened 2020-08-07
Add information on `maybe_stuck()` processes for RMQ

nicolasbock:

---

### DIFF
--- a/sos/report/plugins/rabbitmq.py
+++ b/sos/report/plugins/rabbitmq.py
@@ -37,8 +37,16 @@ class RabbitMQ(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
                     self.fmt_container_cmd(container, 'rabbitmqctl report'),
                     foreground=True
                 )
+                self.add_cmd_output(
+                    self.fmt_container_cmd(
+                        container, "rabbitmqctl eval "
+                        "'rabbit_diagnostics:maybe_stuck().'"),
+                    foreground=True
+                )
         else:
             self.add_cmd_output("rabbitmqctl report")
+            self.add_cmd_output(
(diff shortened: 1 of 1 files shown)


sos/report/plugins/rabbitmq.py
```diff
@@ -37,8 +37,16 @@ def setup(self):
                     self.fmt_container_cmd(container, 'rabbitmqctl report'),
                     foreground=True
                 )
+                self.add_cmd_output(
+                    self.fmt_container_cmd(
+                        container, "rabbitmqctl eval "
+                        "'rabbit_diagnostics:maybe_stuck().'"),
```
TurboTurtle, line 43: Has this been tested? The `fmt_container_cmd()` method already quotes the command being run, so this would result in a decent amount of additional quoting in the command.

There's another report of podman not behaving well with quoted exec commands, so let's dig into this a bit more.

nicolasbock: Thanks @TurboTurtle . I'll run some tests and will update this PR with whether this works or not. I guess in case it doesn't I'll just push a fix :smile:

TurboTurtle: Just wanted to cycle back on this - have you been able to test the quoting here on a system actually running rabbitmq in a container?

nicolasbock: Hi @TurboTurtle . I am still in the process of validating the container part of this change. I think I understand now how to set up a reproducer. Please bear with me for a little longer. Thanks!

nicolasbock: I can confirm that this PR works with #2232 applied as well.

TurboTurtle (approved):

PR #2192, opened 2020-08-10
[Plugin] Add default tags and methods for applying tags via regex

TurboTurtle:
This patchset adds a few mechanisms to `Plugin` for adding tags to `manifest.json`.

1. All commands will now by default add a tag of the binary being run
2. Files collected by a copyspec that matches exactly one file collection will now be tagged with the filename minus any extensions, save for `.conf`, which will be changed to `_conf`. E.G. `httpd.conf` will be tagged with `httpd_conf`.
3. Added `add_cmd_tags()` and `add_file_tags()` that will apply tags to cmds/files based on regex matches of the command being run or the file collected respectively.
4. Commands collected via `add_device_cmd()` will be tagged also with the device name
5. Adds specific tagging to several plugins based on requests from the Red Hat Insights team

---

### DIFF
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -1017,7 +1017,7 @@ class Plugin(object):
         self.copy_paths.update(copy_paths)
 
     def add_copy_spec(self, copyspecs, sizelimit=None, maxage=None,
-                      tailit=True, pred=None, tags=None):
+                      tailit=True, pred=None, tags=[]):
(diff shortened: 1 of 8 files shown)


TurboTurtle:
> Why capture the command name as a tag again?

My thinking is that if we/someone are going to be doing something like "give me XYZ in this sosreport with the tag ABC", then that exact same mechanism can be used for "give me all foobar commands", rather than needing one such mechanism for tags and another one for commands.

Issue #1369, opened 2018-06-28
[rpm] "rpm -Va" to run with bigger timeout


pmoravec:
`rpm -Va` is run only conditionally under `rpm.rpmva` option marked as slow, with 180 seconds timeout. That might not be sufficient for some/many systems.

If the command does really require more time on a usual system (or usual system when sosreport is requested to collect it, i.e. when we suspect some relevant issues), it is worth increasing the timeout.

Some stats from 4 my systems:

- 2m7s for a system with 551 packages
- 0m40s for a system with 643 packages
- 4m28s for a system with 1071 packages
- 11m25s for a system with 831 packages (some older RHEL6, might not be much relevant)

bmr-cymru:
I'm tempted to think that as the user has to ask for it (and should know what to expect), that we run this command without timeout, or with something huge. It would still be helpful in the case of a genuinely stuck `rpm` process, but the problem is that "huge" is a relative term when the time is a function of both system performance and the installed package workload.

jcastill:
The max timeout we have in sos is in the foreman plugin, with two commands having 600s  timeout. It may not be enough still, so perhaps a 12 minute timeout may cover this?

PR #2193, opened 2020-08-10
[rpm] increase timeout of "rpm -Va"

pmoravec:
Since "rpm -Va" takes often more time than the current timeout,
let increase it accordingly.

Resolves: #1369

---

### DIFF
--- a/sos/report/plugins/rpm.py
+++ b/sos/report/plugins/rpm.py
@@ -53,8 +53,9 @@ class Rpm(Plugin, RedHatPlugin):
             add_rpm_cmd(query_fmt, None, None, "package-data")
 
         if self.get_option("rpmva"):
+            self.plugin_timeout = 1000
             self.add_cmd_output("rpm -Va", root_symlink="rpm-Va",
-                                timeout=180)
+                                timeout=900)
 
         if self.get_option("rpmdb"):
             self.add_cmd_output("lsof +D /var/lib/rpm",


sos/report/plugins/rpm.py
```diff
@@ -54,7 +54,7 @@ def add_rpm_cmd(query_fmt, filter_cmd, symlink, suggest):
 
         if self.get_option("rpmva"):
             self.add_cmd_output("rpm -Va", root_symlink="rpm-Va",
-                                timeout=180)
+                                timeout=900)
```
TurboTurtle, line 57: We'll also want to increase the whole plugin timeout here, as otherwise we'll still be limited to the plugin timeout of 300 seconds.

pmoravec: Ah, good point, PR updated.

TurboTurtle: I was thinking more along the lines of conditionally updating `plugin_timeout` if `rpm.va` is used, so that we don't inadvertently wait for 15+ minutes if there are problems with `rpm` besides.

The thread timeouts are set when the plugin's `collect()` method is called, so adjusting it during `setup()` should be perfectly fine.

pmoravec:
Makes sense, here you are ;-)

pmoravec:
Thanks, LGTM, fixed it :)

Issue #2178, opened 2020-07-30
sos now enforces /var/tmp as tmpdir unless --tmp-dir option is used


slashdd:
Debian/Ubuntu systemd's implementation doesn't cleanup `/var/tmp` which makes it risky to archive sosreport at that specific location. We use /tmp for quite some time.

Testing the sosreport master (AKA [4.0]) it seems like `/var/tmp `is enforced -> https://github.com/sosreport/sos/blob/master/sos/component.py#L142

Unless tmp-dir option is used in command-line or in sos.conf
Adding the specification inside the ubuntu/debian policy doesn't seem to be taken into account.

slashdd:
```
$ dpkg -S /usr/lib/tmpfiles.d/tmp.conf
systemd: /usr/lib/tmpfiles.d/tmp.conf
```
```
$ cat /usr/lib/tmpfiles.d/tmp.conf
#  This file is part of systemd.
#
#  systemd is free software; you can redistribute it and/or modify it
#  under the terms of the GNU Lesser General Public License as published by
#  the Free Software Foundation; either version 2.1 of the License, or
#  (at your option) any later version.

# See tmpfiles.d(5) for details

# Clear tmp directories separately, to make them easier to override
D /tmp 1777 root root -
#q /var/tmp 1777 root root 30d

```

slashdd:
A workaround would be to enforce is inside sos.conf at the packaging level

```
[global]
# Set global options here that are not component specific
# If you would like one global default value to be specifically overridden for
# just one component, but not others, you may override that value in the
# component specific section below
#verbose = 3
#verify = yes
#batch = yes
#log-size = 15
tmp-dir = /tmp
```

slashdd:
@BryanQuigley fyi

TurboTurtle:
Yes, this change was made in the initial 4.0 redesign to allow for multiple subcommands, and was highlighted and discussed in #1993.

#1997 is still open and tracking this already.

slashdd:
@TurboTurtle I can add it into the debian/ubuntu package itselves, but would it be worth having a` tmpfiles.d/sosreport.conf `maintained upstream ? If sosreport comes with its tmpfiles.d directive, I'll be more comfortable to do the switch from "/tmp/" to "/var/tmp" once 4.0 version is released.

TurboTurtle:
Yeah, I'm not opposed to carrying a `tmpfiles.d` conf in upstream.

[...]

PR #2195, opened 2020-08-11
[tmpfiles.d] Provide a tmpfiles.d conf example

slashdd:
sos will now enforce /var/tmp to store its content.

This example is for vendors such as Debian/Ubuntu
without a current /var/tmp systemd-tmpfiles
cleaning directive in place.

Debian bug reference: #966621

Closes: #2178

---


slashdd:
I want to note that it will ONLY delete the contents of extracted sosreport archives in /var/tmp, not the archive itself. But it's a start.
I'll see if I can improve it along the way. Anyway it's an example.

PR #2196, opened 2020-08-12
Switch to flake8, enable 20.04 testing and run checks against all python code

BryanQuigley:
Several commits. Happy to squash this more, but figured it's useful separated out.

---

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ jobs:
       os: linux
       dist: focal
       language: shell
-      install: sudo apt-get update; sudo apt install pycodestyle python3-pexpect;
+      install: sudo apt-get update; sudo apt install flake8 python3-pexpect;
       script:
         - "flake8 sos tests bin/*"
         - "sudo ./tests/simple.sh"
@@ -12,7 +12,7 @@ jobs:
       os: linux
       dist: bionic
       language: shell
-      install: sudo apt-get update; sudo apt install pycodestyle python3-pexpect;
+      install: sudo apt-get update; sudo apt install python3-pexpect;
       script: "sudo ./tests/simple.sh"
     - name: "18.04 native run for arm64"
       os: linux
@@ -62,4 +62,4 @@ notifications:
       - "us.freenode.net#sosreport"
     on_success: change
 git:
-  depth: 5
+  depth: 20
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,12 @@
 jobs:
   include:
-    - name: "20.04 pycodestyle and native run (py3.8)"
+    - name: "20.04 flake8 and native run (py3.8)"
       os: linux
       dist: focal
       language: shell
       install: sudo apt-get update; sudo apt install pycodestyle python3-pexpect;
       script:
-        - "pycodestyle sos tests bin/*"
+        - "flake8 sos tests bin/*"
         - "sudo ./tests/simple.sh"
     - name: "18.04 pycodestyle and native run (py3.6)"
       os: linux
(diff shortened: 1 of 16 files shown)


sos/policies/__init__.py
```diff
@@ -717,17 +717,8 @@ def _get_pkg_name_for_binary(self, binary):
...
-            # set compression to -2
-            cmd = "%s -2" % cmd
-            try:
-                xz_package = self._get_pkg_name_for_binary(method)
-                xz_version = self.package_manager\
-                                 .all_pkgs()[xz_package]["version"]
-            except Exception as e:
-                xz_version = [u'0']  # deal like xz version is really old
-            if xz_version >= [u'5', u'2']:
-                cmd = "%s -T%d" % (cmd, threads)
```
TurboTurtle, line 730: Are we 100% sure that `xz` older than 5.2 is no longer available in any of our active downstreams?

.travis.yml
```diff
@@ -1,13 +1,19 @@
...
+    - name: "20.04 flake8 and native run (py3.8)"
       os: linux
-      dist: bionic
+      dist: focal
       language: shell
-      install: sudo apt-get update; sudo apt install pycodestyle python3-pexpect;
+      install: sudo apt-get update; sudo apt install flake8 python3-pexpect;
       script:
-        - "pycodestyle sos"
+        - "flake8 sos tests bin/*"
```
TurboTurtle, line 9: Is `flake8` a complete replacement for `pycodestyle`, as in it checks _all_ the same linter tests and then some? It's been a while since I've looked into it and I can't rightly remember if it's a complete replacement or one that is best used in tandem.

I have no objections to changing to `flake8`, but I just want to be sure we won't be missing something from `pycodestyle`.

sos/report/plugins/__init__.py
```diff
@@ -2021,8 +2021,8 @@ def convert_cmd_scl(self, scl, cmd):
             prefix = open('/etc/scl/prefixes/%s' % scl, 'r').read()\
                      .rstrip('\n')
         except Exception as e:
-            self._log_error("Failed to find prefix for SCL %s, using %s"
-                            % (scl, prefix))
+            self._log_error("Error %s to find prefix for SCL %s, using %s"
+                            % (e, scl, prefix))
```
TurboTurtle, line 2025: This reads a bit weird to me. How about

```python
self._log_error("Failed to find prefix for SCL %s using %s: %s"
                % (scl, prefix, e))
```

TurboTurtle:
Also, I just pushed a trivial fix commit to master for the fibre device failure Travis was hitting, so a rebase should fixup the test failures here.

.travis.yml
BryanQuigley: flake8 includes pycodestyle itself, so it should be equivalent unless ignores are different.  Double checked,

pycodestyle ignores E121,E123,E126,E226,E24,E704,W503,W504
flake8 ignore 'E126', 'E24', 'E226', 'E121', 'E704', 'E123', 'W504', 'W503']

So flake8 seems to base it's ignore list on pycodestyles.

sos/policies/__init__.py
BryanQuigley: I'm pretty confident - 5.2 was released in late 2014/early 2015 - https://github.com/xz-mirror/xz/releases
We're being less aggressive then with Python.

According to this https://pkgs.org/search/?q=xz CentOS 6 has 5.2 it too.  Debian (oldoldstable)/Ubuntu (16.04) are the last ones with old versions _ although in this case it's 5.1alpha.

sos/report/plugins/__init__.py
BryanQuigley: Sure and thanks for the fix

PR #2199, opened 2020-08-12
[ubuntu] Collect /var/log/ubuntu-advantage.log file(s).

arossbell:
This log file is not being collected.
Canonical engineers have needed to ask for the file outside of sosreport.
This addition will collect the log file, as applicable, as part of sosreport.

Resolves: #2199

Signed-off-by: Adam R Bell <adam.bell@canonical.com>

---

### DIFF
--- a/sos/report/plugins/ubuntu.py
+++ b/sos/report/plugins/ubuntu.py
@@ -20,5 +20,15 @@ class Ubuntu(Plugin, UbuntuPlugin):
         self.add_cmd_output([
             "ubuntu-security-status --thirdparty --unavailable",
             "hwe-support-status --verbose",
-            "ubuntu-advantage status"
         ])
+
+        if self.is_installed('ubuntu-advantage-tools'):
+            self.add_cmd_output("ubuntu-advantage status")
+            if not self.get_option("all_logs"):
+                self.add_copy_spec([
+                    "/var/log/ubuntu-advantage.log",
+                    "/var/log/ubuntu-advantage.log.1",
+                    "/var/log/ubuntu-advantage.log.2*",
+                ])
+            else:
+                self.add_copy_spec("/var/log/ubuntu-advantage.log*")


sos/report/plugins/ubuntu.py
```diff
@@ -22,3 +22,12 @@ def setup(self):
             "hwe-support-status --verbose",
             "ubuntu-advantage status"
         ])
+
```
slashdd, line 25: What would you think if we add another check to validate that `ubuntu-advantage-tools `is installed before looking at collecting ubuntu-advantage logs ?

`if self.is_installed('ubuntu-advantage-tools'):`

arossbell: Good idea; I'll get that tested on my side.

slashdd: And move the `ubuntu-advantage status` command under the is_installed() check as well ? 

I'm also starting to think that ubuntu-advantage should probably have its own plugin ? Have you considered that scenario ? so that ubuntu.py will remain community and ua.py will be more ubuntu "entreprise" ish.

arossbell: I can do the first part of that.

For the latter part of that, it isn't clear whether that would be beneficial for sosreport. "ubuntu-advantage-tools" is installed by default on modern community-level/non-enterprise Ubuntu systems, so the additional plugin would be executed along with this plugin. I believe that this is what the output looks like for a non-enterprise, community-level Ubuntu system (UA Tools version 20.3):
```
SERVICE       AVAILABLE  DESCRIPTION
cc-eal        no         Common Criteria EAL2 Provisioning Packages
esm-apps      yes        UA Apps: Extended Security Maintenance
esm-infra     yes        UA Infra: Extended Security Maintenance
fips          no         NIST-certified FIPS modules
fips-updates  no         Uncertified security updates to FIPS modules
livepatch     yes        Canonical Livepatch service

This machine is not attached to a UA subscription.
See https://ubuntu.com/advantage
```
It could certainly be separated into its own plugin (I could work on writing that, too), but it just isn't clear exactly where the benefit would be by doing that (leaving only ...
```
        self.add_cmd_output([
            "ubuntu-security-status --thirdparty --unavailable",
            "hwe-support-status --verbose",
        ])
```
... on this 'ubuntu.py' plugin)

sos/report/plugins/ubuntu.py
```diff
@@ -22,3 +22,12 @@ def setup(self):
...
+        if not self.get_option("all_logs"):
+            self.add_copy_spec([
+                "/var/log/ubuntu-advantage.log",
+                "/var/log/ubuntu-advantage.log.1",
+                "/var/log/ubuntu-advantage.log.2*",
```
TurboTurtle, line 30: Is this trailing glob intentional? It seems odd to me if so, since you'd lose the intermediate rotated logs between 2 and 20.

sos/report/plugins/ubuntu.py
slashdd: If no actual benefit, I'm fine with keeping it as is. I think it was worth the exercise to think about it.
I would add the is_installed() method and also move ua-tool stuff inside the condition statement .

sos/report/plugins/ubuntu.py
arossbell: It's intentional to grab the file whether it's `/var/log/ubuntu-advantage.log.2` or `/var/log/ubuntu-advantage.log.2.gz` in this case.
I actually left it like this to somewhat mimic what it did in [the logs plugin (with syslog)](https://github.com/sosreport/sos/blob/8745eb0ac0baed51f89ff9afaf36220d592a119f/sos/report/plugins/logs.py#L82).
The thought about 2 vs 20 (etc) did cross my mind, though, yes. I can make that a better match to limit to "2 or 2.*" if that would be preferred?

slashdd: That's how `ubuntu-advantage-tools` logrotate logs on my Ubuntu machine:
```
/var/log/ubuntu-advantage.log    
/var/log/ubuntu-advantage.log.1  
=> /var/log/ubuntu-advantage.log.2.gz 
/var/log/ubuntu-advantage.log.3.gz
/var/log/ubuntu-advantage.log.4.gz
```
logrotate conf:
```
/var/log/ubuntu-advantage.log {
    rotate 6
    monthly
    compress
    delaycompress
    missingok
    notifempty
}
```
With current configuration, it will never exceed 6.

TurboTurtle: Ah, ok - makes sense, thanks for the context.

arossbell:
Moved `self.add_cmd_output("ubuntu-advantage status")` and ubuntu-advantage.log collection into `self.is_installed('ubuntu-advantage-tools')` conditional as mentioned by @slashdd.

Issue #2150, opened 2020-07-12
[Plugin] Add designate plugin


bregman-arie:
To collect /var/lib/config-data/puppet-generated/designate/etc/designate/designate.conf and /etc/designate/designate.conf

PR #2201, opened 2020-08-13
[openstack_designate] Add list command output

johnsom:
This patch adds the various "list" command output for OpenStack
designate to the SOS report.
    
Closes: #2150
    
Signed-off-by: Michael Johnson <johnsomor@gmail.com>

---

### DIFF
--- a/sos/report/plugins/openstack_designate.py
+++ b/sos/report/plugins/openstack_designate.py
@@ -6,7 +6,7 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.report.plugins import Plugin, RedHatPlugin
+from sos.report.plugins import Plugin, RedHatPlugin, UbuntuPlugin
 
 
 class OpenStackDesignate(Plugin):
@@ -82,5 +82,10 @@ class OpenStackDesignate(Plugin):
 class RedHatdesignate(OpenStackDesignate, RedHatPlugin):
 
     packages = ('openstack-selinux',)
+
+
+class Ubuntudesignate(OpenStackDesignate, UbuntuPlugin):
+
+    packages = ('designate-common',)
 
 # vim: set et ts=4 sw=4 :
--- a/sos/report/plugins/openstack_designate.py
+++ b/sos/report/plugins/openstack_designate.py
@@ -18,6 +18,18 @@ class OpenStackDesignate(Plugin):
 
     var_puppet_gen = "/var/lib/config-data/puppet-generated/designate"
 
+    subcmds = ['dns service list',
+               'dns quota list',
+               'ptr record list',
+               'tld list',
+               'tsigkey list --column name --column algorithm --column scope',
+               'zone blacklist list',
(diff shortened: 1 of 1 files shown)


sos/report/plugins/openstack_designate.py
```diff
@@ -0,0 +1,95 @@
...
+                            subdir='blacklist')
+        self.add_cmd_output('openstack zone transfer request list '
+                            '--all-projects', subdir='transfer-request')
+        self.add_cmd_output('openstack zone transfer accept list '
+                            '--all-projects', subdir='transfer-accept')
+        self.add_cmd_output('openstack tld list --all-projects',
+                            subdir='tld')
+        self.add_cmd_output('openstack tsigkey list --all-projects '
+                            '--column name --column algorithm --column scope',
+                            subdir='tsig')
```
TurboTurtle, line 64: The `subdir` parameter is typically used when there will be numerous, iterative, outputs for a particular item. It looks here like we'll just have a single output in each of these subdirs, so we really don't gain anything here by using them.

If there are individual command outputs to be captured for each item being listed here, then those should be added and contained in the same subdir - otherwise I think it's better if these outputs are in the top-level plugin directory together.

sos/report/plugins/openstack_designate.py
```diff
@@ -0,0 +1,95 @@
...
+        regexp = r"((?m)^\s*(%s)\s*=\s*)(.*)" % "|".join(protect_keys)
+
+        self.do_path_regex_sub("/etc/designate/*", regexp, r"\1*********")
+        self.do_path_regex_sub(
+            self.var_puppet_gen + "/etc/designate/*",
+            regexp, r"\1*********"
+        )
+
+
+class RedHatdesignate(OpenStackDesignate, RedHatPlugin):
```
TurboTurtle, line 91: From the other designate PR, there is a request to add Ubuntu support. Let's add that here as well, since you're already carrying the initial commit from the other PR.

sos/report/plugins/openstack_designate.py
```diff
@@ -0,0 +1,91 @@
...
+        self.add_cmd_output('openstack zone list --all-projects')
+        self.add_cmd_output('openstack ptr record list --all-projects')
+        self.add_cmd_output('openstack zone export list --all-projects')
+        self.add_cmd_output('openstack zone import list --all-projects')
+        self.add_cmd_output('openstack zone blacklist list --all-projects')
+        self.add_cmd_output('openstack zone transfer request list '
+                            '--all-projects')
+        self.add_cmd_output('openstack zone transfer accept list '
+                            '--all-projects')
+        self.add_cmd_output('openstack tld list --all-projects')
```
TurboTurtle, line 53: `add_cmd_output()` also accepts a list, and the preferred style in sos is to use a list when there are no additional parameters being passed, so let's condense these into a single call using a list of these commands.

After that, I believe this looks good now.

PR #2210, opened 2020-08-20
[collect] Allow collect to upload via policy

TurboTurtle:
Extends the upload functionality to `sos collect`, to work the same way
that it does for `report`. Uploads will default to policy definitions,
then user provided overrides.

Resolves: #2210

---

### DIFF
--- a/sos/policies/__init__.py
+++ b/sos/policies/__init__.py
@@ -1387,7 +1387,7 @@ class LinuxPolicy(Policy):
                                            "that you are generating this "
                                            "report for [%s]: ") % caseid)
                 # Policies will need to handle the prompts for user information
-                if cmdline_opts.upload or self.upload_url:
+                if cmdline_opts.upload and self.get_upload_url():
                     self.prompt_for_upload_user()
                     self.prompt_for_upload_password()
                 self._print()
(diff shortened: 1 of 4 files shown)


sos/policies/__init__.py
```diff
@@ -1402,7 +1402,7 @@ def pre_work(self):
                                            "that you are generating this "
                                            "report for [%s]: ") % caseid)
                 # Policies will need to handle the prompts for user information
-                if cmdline_opts.upload or self.upload_url:
+                if cmdline_opts.upload and self.get_upload_url():
```
pmoravec, line 1405: Why this change (esp. compared to "upload or upload_url" at https://github.com/sosreport/sos/pull/2210/files#diff-e07b4b5df8f32d4dab336b60233ae2aeR1130), please?

Otherwise nice extension.

TurboTurtle: Hmm, the prompts should set the necessary bits to allow the latter addition to execute appropriately - but from a consistency standpoint you're absolutely right. I'll change the latter to also check `get_upload_url()`.

pmoravec (approved):

PR #2213, opened 2020-08-25
[powerpc] Collecting logs for POWER systems Hardware components

mamatha4:
This patchset enables sosreport to capture
logs for power specific Hardware components

Signed-off-by: Mamatha Inamdar <mamatha4@linux.vnet.ibm.com>

---

- [ x] If this commit resolves an existing pull request, is the line `Resolves: #PRNUMBER` included in an independent line?

### DIFF
--- a/sos/report/plugins/arcconf.py
+++ b/sos/report/plugins/arcconf.py
@@ -1,0 +1,28 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+
+# This sosreport plugin is meant for sas adapters.
+# This plugin logs inforamtion on each adapter it finds.
+
+from sos.report.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+
+
+class arcconf(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
+
+    short_desc = 'arcconf Integrated RAID adapter information'
+
+    plugin_name = "arcconf"
+    commands = ("arcconf",)
+
+    def setup(self):
+
+        # get list of adapters
+        self.add_cmd_output("arcconf getconfig 1")
+
+# vim: et ts=4 sw=4
--- a/sos/report/plugins/iprconfig.py
+++ b/sos/report/plugins/iprconfig.py
@@ -32,6 +32,7 @@ class IprConfig(Plugin, RedHatPlugin, UbuntuPlugin, DebianPlugin):
             "iprconfig -c show-af-disks",
             "iprconfig -c show-all-af-disks",
             "iprconfig -c show-slots",
+            "iprconfig -c dump"
         ])
 
         show_ioas = self.collect_cmd_output("iprconfig -c show-ioas")
--- a/sos/report/plugins/mvcli.py
+++ b/sos/report/plugins/mvcli.py
@@ -1,0 +1,35 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+
+# This sosreport plugin is meant for sas adapters.
+# This plugin logs inforamtion on each adapter it finds.
+
+from sos.report.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+
+
+class mvCLI(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
+
+    short_desc = 'mvCLI Integrated RAID adapter information'
+
+    plugin_name = "mvcli"
+    commands = ("/opt/marvell/bin/mvcli",)
+
+    def setup(self):
+
+        # get list of adapters
+        subcmds = [
+            'info -o vd',
+            'info -o pd',
+            'info -o hba',
+            'smart -p 0',
+        ]
+
+        self.add_cmd_output(["/opt/marvell/bin/mvcli %s" % s for s in subcmds])
+
+# vim: et ts=4 sw=4
--- a/sos/report/plugins/kernel.py
+++ b/sos/report/plugins/kernel.py
@@ -107,6 +107,7 @@ class Kernel(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin, CosPlugin):
             "/proc/misc",
             "/var/log/dmesg",
             "/sys/fs/pstore",
+            "/var/log/hcnmgr",
             clocksource_path + "available_clocksource",
             clocksource_path + "current_clocksource"
         ])
--- a/sos/report/plugins/powerpc.py
+++ b/sos/report/plugins/powerpc.py
@@ -50,7 +50,10 @@ class PowerPC(Plugin, RedHatPlugin, UbuntuPlugin, DebianPlugin):
                 "ppc64_cpu --run-mode",
                 "ppc64_cpu --frequency",
                 "ppc64_cpu --dscr",
-                "diag_encl -v"
+                "diag_encl -v",
+                "lsvpd -D",
+                "lsmcode -A",
+                "lscfg -v"
             ])
 
         if ispSeries:


sos/report/plugins/system.py
```diff
@@ -23,6 +23,14 @@ def setup(self):
             "/etc/sysconfig",
             "/etc/default",
             "/etc/environment",
+            "info -o vd",
+            "info -o pd",
+            "info -o hba",
```
TurboTurtle, line 28: The `info` command is traditionally a man-like command, and in particular the `-o` option pushes output to a file, which definitely would not make sense in an sos plugin.

Does `info` map to a different command on powerpc? If so, I'd recommend these go in the `powerpc` plugin, not `system.

sos/report/plugins/system.py
```diff
@@ -23,6 +23,14 @@ def setup(self):
...
+            "arcconf getconfig 1",
+            "smart -p 0",
+            "sas3ircu list",
```
TurboTurtle, line 31: This is already captured by the `sas3ircu` plugin

sos/report/plugins/system.py
```diff
@@ -23,6 +23,14 @@ def setup(self):
...
+            "storcli show all",
+            "storcli /cx[/eall]/sall show",
```
TurboTurtle, line 33: If I'm not mistaken, `storcli` is the replacement for `megacli`, in which case these should either go into the `megacli` plugin, or a new `storcli` plugin should be created to house them. In either case, they don't belong in `system`.

sos/report/plugins/system.py
```diff
@@ -23,6 +23,14 @@ def setup(self):
             "/etc/sysconfig",
             "/etc/default",
             "/etc/environment",
+            "info -o vd",
+            "info -o pd",
+            "info -o hba",
+            "arcconf getconfig 1",
```
TurboTurtle, line 29: I'm not familiar with `arcconf`, but a quick search shows it is for adaptec RAID controllers. It does not belong in `system`.

sos/report/plugins/system.py
bmr-cymru: These look like Marvell mvCLI command verbs (`-o vd`, `-o pd` and `-o hba` are all valid arguments to the `mvcli info` command, and nothing else that I know).

These adapters are common on IBM systems but if that's the case then they won't work without (at least) the `mvcli` command as prefix - it's also possibly common that the binary is installed in a non-standard bin directory, meaning that we would either need to provide an absolute path or modify the pre-set PATH used by `sos`.

More information on the commands can be found [here](https://www.ibm.com/support/knowledgecenter/9040-MR9/p9ej8/p9ej8_recreate_drive.htm) and [here](https://www.ibm.com/support/knowledgecenter/TI0003K/p8eip/p8eip_drive_commands_mvcli.htm).

bmr-cymru: So yes: definitely not approariate for the system plugin. These need to find a new home, either in a plugin of their own or one of the existing plugins that's more specific to the systems where these devices are found.

bmr-cymru (approved):
Ack so long as it's just the `iprconfig -c dump` command (and not the `iprsos` wrapper script that appears to duplicate a bunch of information from sos).

sos/report/plugins/system.py
TurboTurtle: Oh, also I didn't notice this before, but these are also in the wrong call here as well. Command output needs to be collected via `add_cmd_output()`, not `add_copy_spec()`

mamatha4:
Thanks for the review, 
will make suggested changes and submit new version of patch set

mamatha4:
Hi,
I have updated all review comments and pushed new version of patch set, can you please check

mamatha4:
Hi, I have pushed 6 commits but I am able to see only one commit  in my latest version of branch, am I missing something here? do I need to push all commits once again?

bmr-cymru:
If you are rebasing again for any reason please drop the `:` from the commit subject lines - the convention we use in the `sos` project is `[what] how it changed`. There's no need for anything but whitespace between the closing `]` and the change description.

sos/report/plugins/storcli.py
```diff
@@ -0,0 +1,29 @@
...
+
+    plugin_name = "storcli"
+    commands = ("storcli",)
+
+    def setup(self):
+
+        # Display list of adapters
+        self.collect_cmd_output("storcli show all", timeout=5)
+        self.collect_cmd_output("storcli /cx[/eall]/sall show", timeout=5)
+
```
bmr-cymru, line 28: You can use a single `collect_cmd_output()` call here since all the other arguments are the same - just pass both commands as a list.

Do you have example output for the 2nd command (or the whole plugin)?

bmr-cymru (changes requested):

sos/report/plugins/arcconf.py
```diff
@@ -0,0 +1,28 @@
...
+
+    short_desc = 'arcconf Integrated RAID adapter information'
+
+    plugin_name = "arcconf"
+    commands = ("arcconf",)
+
+    def setup(self):
+
+        # get list of adapters
+        self.collect_cmd_output("arcconf getconfig 1", timeout=5)
```
bmr-cymru, line 26: Same

sos/report/plugins/mvcli.py
```diff
@@ -0,0 +1,31 @@
...
+    plugin_name = "mvcli"
+    commands = ("/opt/marvell/bin/mvcli",)
+
+    def setup(self):
+
+        # get list of adapters
+        self.collect_cmd_output("/opt/marvell/bin/mvcli info -o vd", timeout=5)
+        self.collect_cmd_output("/opt/marvell/bin/mvcli info -o pd", timeout=5)
+        self.collect_cmd_output("/opt/marvell/bin/mvcli info -o hba", timeout=5)
+        self.collect_cmd_output("/opt/marvell/bin/mvcli smart -p 0", timeout=5)
```
bmr-cymru, line 29: Same here: use a single call and don't use `collect_cmd_output()` unless you are going to do something with the result.

bmr-cymru (changes requested):

sos/report/plugins/storcli.py
```diff
@@ -0,0 +1,29 @@
...
+# See the LICENSE file in the source distribution for further information.
+
+
+# This sosreport plugin is meant for sas adapters.
+# This plugin logs inforamtion on each adapter it finds.
+
+from sos.report.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+
+
+class SAS3ircu(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
```
TurboTurtle, line 16: I assume this is a copy/paste error? Please make sure this matches the plugin name.

mamatha4: oh sorry for that...yes will update the patch

sos/report/plugins/mvcli.py
mamatha4: ok  sure..will make changes

sos/report/plugins/arcconf.py
mamatha4: ok

sos/report/plugins/storcli.py
mamatha4: Thanks for the review..sure will make changes and send new  version of patchset

mamatha4: Yes as you mentioned we have to rework on this patch
please find output of the second commands

storcli /c0/eall/sall show
CLI Version = 007.1316.0000.0000 Mar 12, 2020
Operating system = Linux 4.14.0-115.el7a.ppc64le
Controller = 0
Status = Success
Description = Show Drive Information Succeeded.
Drive Information :
=================
-----------------------------------------------------------------------------------------------------
EID:Slt DID State DG       Size Intf Med SED PI SeSz Model                                   Sp Type
-----------------------------------------------------------------------------------------------------
252:0     8 JBOD  -  223.570 GB SATA SSD N   N  512B MTFDDAK240TCB        01EJ260 01EJ260IBM U  -
252:2    10 JBOD  -  223.570 GB SATA SSD N   N  512B MTFDDAK240TCB        01EJ260 01EJ260IBM U  -
252:3    11 JBOD  -  223.570 GB SATA SSD N   N  512B MTFDDAK240TCB        01EJ260 01EJ260IBM U  -
252:4    15 UBad  F    2.182 TB SAS  HDD N   Y  4 KB ST2400MM0129                            U  -
252:5    14 Onln  0    2.182 TB SAS  HDD N   Y  4 KB ST2400MM0129                            U  -
252:6    13 Onln  1    2.182 TB SAS  HDD N   N  4 KB ST2400MM0129                            U  -
252:7    20 JBOD  -    2.182 TB SAS  HDD N   Y  4 KB ST2400MM0129                            U  -
-----------------------------------------------------------------------------------------------------

mamatha4: I will rework and send this patch later. right now I will update all other comments and send new version of patchset except storcli

mamatha4:
Hi,
I have worked on review comments and pushed new version of patch set, can you please check

sos/report/plugins/iprconfig.py
```diff
@@ -32,6 +32,7 @@ def setup(self):
             "iprconfig -c show-af-disks",
             "iprconfig -c show-all-af-disks",
             "iprconfig -c show-slots",
+            "iprconfig -c dump"
```
TurboTurtle, line 35: Can we get a confirmation on @bmr-cymru 's earlier concern about this:

> Ack so long as it's just the iprconfig -c dump command (and not the iprsos wrapper script that appears to duplicate a bunch of information from sos).

If this is just a separate command then ack, but if it's a sos-like script we'd want to drop this and hopefully add the actual collections it is doing to sos natively.

TurboTurtle (commented):
Code LGTM, just one highlight that I'd like confirmation on.

sos/report/plugins/iprconfig.py
mamatha4: Do we need to collect output of following commands instead of adding iprconfig -c dump?
iprconfig -c show-alt-config
iprconfig -c show-config
iprconfig -c show-ioas
iprconfig -c show-arrays
iprconfig -c show-hot-spares
iprconfig -c show-jbod-disks
iprconfig -c show-details
iprconfig -c show-battery-info
iprconfig -c query-path-status
iprconfig -c query-path-details
iprconfig -c show-perf

following commands which are already added to sos/report/plugins/iprconfig.py
 "iprconfig -c show-af-disks",
"iprconfig -c show-all-af-disks",
"iprconfig -c show-slots",

"iprconfig -c dump"  will collect  output of all the above mentioned commands

bmr-cymru: Using `dump` if it's just a shorthand for those commands is fine - the only concern was whether it might also run the `iprsos` script in the background (which is more of a problem since it duplicates `sos` functionality).

mamatha4: Here is definition of "dump" argument of iprconfig command
dump
Display detailed hardware and system information on standard output. In case a report file is needed, the iprsos command will create one at /var/log/iprsos.log.
Example:
iprconfig -c dump

bmr-cymru: That was the comment in the docs that confused me - it wasn't clear whether that implied it would be created automatically or if the user is supposed to run `iprsos` only if one is needed...

mamatha4: ok, got the system today and verified once again iprconfig -c dump will not run the iprsos script in the background and it will not create /vat/log/iprsos.log file, it will just dump output of all the commands which are mentioned above

TurboTurtle (approved):

mamatha4:
Thanks for the review

Issue #2214, opened 2020-08-25
[sos-collector] discussion to use sos report or sosreport


arif-ali:
Hi all,

Started testing `sos collecter` from 4.0 branch as per @slashdd work on getting sos into Ubuntu

Using the `--verbose`, I noticed that it was using `sosreport` as a command instead of `sos report` as shown below

```
<snip>
[immense-owl:sosreport] Final sos command set to sosreport --batch -c auto 
[arif-laptop:sosreport] Final sos command set to sosreport --batch -c auto
<snip>
```

The first thing that came into my mind is, why are we using `sosreport` when the new sos 4.0 is `sos report`? maybe it's already been discussed

I guess a things to consider why we might to keep `sosreport` or move to `sos report`

* We may want to use `sos collecter` to collect sosreports from machines that have sos 3.9.1 and older (i.e. current stable operating systems)
* We use sos 4.0 version of `sos collect` just for nodes that have sos 4.0, and hence use `sos report ` command
* I guess the decision may come when the `sosreport` command gets deprecated and removed from sos 4.0?

TurboTurtle:
Yeah, the 4.0 detection is on the list - there are a few "gotcha" points where simply changing the initial sos command to `sos report` won't work (these points are in the per-node command formatting and execution bits), but it should all function normally today since `sosreport` is still shipped with the packaging.

The framework is already there to detect sos 4.x, it's just a matter of testing to work out the "gotcha"s. However, we _absolutely_ want to preserve 3.x compatibility.

PR #2221, opened 2020-08-31
[collect] Detect sos-4.0 and use new binary

TurboTurtle:
Updates the `collect` module to detect sos-4.0 installation, and if
present update the binary called to `sos report`.

As part of this change, move the update of the binary path to the
policy's full filesystem path into `finalize_sos_cmd()`

Closes: #2214
Resolves: #2221

---

### DIFF
--- a/sos/collector/sosnode.py
+++ b/sos/collector/sosnode.py
@@ -62,6 +62,7 @@ class SosNode():
             'presets': [],
             'sos_cmd': commons['sos_cmd']
         }
+        self.sos_bin = 'sosreport'
         filt = ['localhost', '127.0.0.1']
         self.soslog = logging.getLogger('sos')
         self.ui_log = logging.getLogger('sos_ui')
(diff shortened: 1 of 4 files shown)


sos/collector/sosnode.py
```diff
@@ -649,6 +647,15 @@ def finalize_sos_cmd(self):
             if self.opts.since:
                 sos_opts.append('--since=%s' % quote(self.opts.since))
 
+        # sos-4.0 changes the binary
+        if self.check_sos_version('4.0'):
+            self.sos_bin = 'sos report'
+
+        sos_cmd = sos_cmd.replace(
+            'sosreport',
```
pmoravec, line 655: You relax from `if cmd.startswith('sosreport'):` test which is good ATM since `self.sos_cmd = 'sosreport --batch ..'`

sos/policies/redhat.py
```diff
@@ -39,7 +39,7 @@ class RedHatPolicy(LinuxPolicy):
     upload_directory = '/incoming'
     default_container_runtime = 'podman'
     sos_pkg_name = 'sos'
-    sos_bin_path = '/usr/sbin/sosreport'
+    sos_bin_path = '/usr/sbin'
```
pmoravec, line 42: Shouldn't we appy the same also to https://github.com/sosreport/sos/blob/master/sos/policies/debian.py#L20 ?

TurboTurtle: Nice catch! I actually just need to remove that line from Debian entirely, and leave it as the policy default.

pmoravec (approved):

PR #2223, opened 2020-08-31
[openstack_ironic] Missing ironic-inspector configs

valleedelisle:
We're missing the ironic-inspector configurations, probably because they were in the `RedHatPlugin` class, at the bottom of the file and they were probably missed when updating this plugin. Moving them at the top with the other `add_copy_spec` will help tracking them in case something change again.
    
Revamping also the way we grab logs to check if we're in a container first.

---

### DIFF
--- a/sos/report/plugins/openstack_ironic.py
+++ b/sos/report/plugins/openstack_ironic.py
@@ -20,41 +20,95 @@ class OpenStackIronic(Plugin):
     profiles = ('openstack', 'openstack_undercloud')
 
     var_puppet_gen = "/var/lib/config-data/puppet-generated/ironic"
+    ins_puppet_gen = var_puppet_gen + "_inspector"
 
     def setup(self):
-        self.conf_list = [
-            "/etc/ironic/*",
-            self.var_puppet_gen + "/etc/ironic/*",
-            self.var_puppet_gen + "_api/etc/ironic/*"
-        ]
-        self.add_copy_spec([
-            "/etc/ironic/",
-            self.var_puppet_gen + "/etc/xinetd.conf",
-            self.var_puppet_gen + "/etc/xinetd.d/",
-            self.var_puppet_gen + "/etc/ironic/",
-            self.var_puppet_gen + "/etc/httpd/conf/",
-            self.var_puppet_gen + "/etc/httpd/conf.d/",
-            self.var_puppet_gen + "/etc/httpd/conf.modules.d/*.conf",
-            self.var_puppet_gen + "/etc/my.cnf.d/tripleo.cnf",
-            self.var_puppet_gen + "_api/etc/ironic/",
-            self.var_puppet_gen + "_api/etc/httpd/conf/",
-            self.var_puppet_gen + "_api/etc/httpd/conf.d/",
-            self.var_puppet_gen + "_api/etc/httpd/conf.modules.d/*.conf",
-            self.var_puppet_gen + "_api/etc/my.cnf.d/tripleo.cnf"
-        ])
 
-        if self.get_option("all_logs"):
+        in_container = self.container_exists('.*ironic_api')
+
+        if in_container:
+            self.conf_list = [
+                self.var_puppet_gen + "/etc/ironic/*",
+                self.var_puppet_gen + "/etc/ironic-inspector/*",
+                self.var_puppet_gen + "_api/etc/ironic/*",
+                self.ins_puppet_gen + "/etc/ironic-inspector/*",
+                self.ins_puppet_gen + "/var/lib/httpboot/inspector.ipxe"
+            ]
             self.add_copy_spec([
-                "/var/log/ironic/",
+                "/var/lib/ironic-inspector/",
+                "/var/log/containers/ironic-inspector/ramdisk/",
+                self.var_puppet_gen + "/etc/xinetd.conf",
+                self.var_puppet_gen + "/etc/xinetd.d/",
+                self.var_puppet_gen + "/etc/ironic/",
+                self.var_puppet_gen + "/etc/ironic-inspector/",
+                self.var_puppet_gen + "/etc/httpd/conf/",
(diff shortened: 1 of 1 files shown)


valleedelisle:
> Please rebase to current master, the tree layout has changed and this commit is unmergable until the rebase.

My apologies, it's rebased now.

sos/report/plugins/openstack_ironic.py
```diff
@@ -74,10 +122,28 @@ def setup(self):
...
 
+    def running_in_container(self):
+        for runtime in ["docker", "podman"]:
+            container_status = self.exec_cmd(runtime + " ps")
+            if container_status['status'] == 0:
+                for line in container_status['output'].splitlines():
+                    if line.endswith("ironic_api"):
+                        self.container_runtime = runtime
+                        return True
+        return False
```
TurboTurtle, line 145: There are methods in `Plugin` for detecting containerized workloads. This can all be done away with and replaced with

```python
in_container = self.container_exists('ironic_api')
```

sos/report/plugins/openstack_ironic.py
```diff
@@ -74,10 +122,28 @@ def setup(self):
             self.add_cmd_output("openstack baremetal port list --long")
             self.add_cmd_output("openstack baremetal port group list --long")
 
+        if in_container:
+            for container_name in ['ironic_inspector_dnsmasq',
+                                   'ironic_inspector', 'ironic_pxe_http',
+                                   'ironic_pxe_tftp', 'ironic_neutron_agent',
+                                   'ironic_conductor', 'ironic_api']:
+                self.add_cmd_output('%s exec %s rpm -qa'
+                                    % (self.container_runtime, container_name))
```
TurboTurtle, line 131: Similarly, there is no need to manually construct these.

```python
self.add_cmd_output(self.fmt_container_cmd(container_name, 'rpm -qa'))
```

sos/report/plugins/openstack_ironic.py
```diff
@@ -20,41 +20,89 @@ class OpenStackIronic(Plugin):
...
+                "/etc/ironic-inspector/*",
+            ]
             self.add_copy_spec([
-                "/var/log/ironic/*.log",
+                "/etc/ironic/",
+                "/etc/ironic-inspector/",
+                "/var/lib/ironic-inspector/",
+                "/var/log/ironic-inspector/ramdisk/",
+                "/etc/xinetd.conf",
+                "/etc/xinetd.d/",
```
TurboTurtle, line 85: Capturing these from the `/var/log` location if running containerized is fine, but these are already captured by the `xinetd` plugin and should not be duplicated here.

sos/report/plugins/openstack_ironic.py
```diff
@@ -20,41 +20,89 @@ class OpenStackIronic(Plugin):
...
+                "/etc/httpd/conf/",
+                "/etc/httpd/conf.d/",
+                "/etc/httpd/conf.modules.d/*.conf",
```
TurboTurtle, line 88: Same here for the `apache` plugin.

sos/report/plugins/openstack_ironic.py
valleedelisle: That's even better, thanks!

valleedelisle: Also, regarding the `RedHatPlugin` comment, what I meant is that, since they were at the bottom of the plugin, they were probably missed by the person that added the containerized config at the top. So moving them at the top will help the next time we need to add/change the paths. I'll clarify the commit message.

pmoravec (approved):
Assuming there is some rationale behind the changes (as not knowing OSP ironic at all, I cant judge here) and assuming the change works well on both containerised and non-containerised OSP deployment, ACKing the PR - the resulting code sounds sane.

TurboTurtle:
Cycling around - @valleedelisle has this been tested on containerized and non-containerized OSP?

valleedelisle:
I wanted to test this on OSP13 where the undercloud isn't containerized and it doesn't work because we don't have `python3` on RHEL7, at least with the supported packages. The latest release of `sos` doesn't work without `python3` as you already know it.

The only other OSP release where the undercloud is containerized is OSP15 and it's EOL in 10 days.

If this commit isn't going to make it to RHEL7, ideally, we should remove all the non-containerized code from the openstack plugins because from now, everything is going to be in containers.

Let me know if that works for you guys.

```
(.venv) [stack@undercloud-0 sos]$ ./bin/sosreport --batch -o openstack_ironic
-bash: ./bin/sosreport: /usr/bin/python3: bad interpreter: No such file or directory
(.venv) [stack@undercloud-0 sos]$ python2 ./bin/sosreport --batch -o openstack_ironic
  File "./bin/sosreport", line 20, in <module>
    from sos import SoS
  File "/home/stack/sos/sos/__init__.py", line 23, in <module>
    from sos.options import SosListOption
  File "/home/stack/sos/sos/options.py", line 118
    opts = SoSOptions(**vars(args), arg_defaults=arg_defaults)
                                  ^
SyntaxError: invalid syntax
```

pmoravec:
FYI for pre-(onlyPY3) stuff, we have [legacy-3.9](https://github.com/sosreport/sos/tree/legacy-3.9) branch that we keep for these purposes. By default we backport there selective commits important to appear on a distro without python3, roughly saying.

TurboTurtle:
Yeah, if any of this is needed for EL7 we can look at `legacy-3.9` backports.

Going forward however, I'm not keen on ripping out the non-containerized bits just yet. However, I'd still like confirmation that the changes to capture from containers is working. There's been a few conflicting reports on functionality surrounding the automatic quoting we do for containerized commands and such.

valleedelisle:
@TurboTurtle yup it works, I'm sorry this slipped between too-many-emails. I tested it in an OSP16.x lab with containerized undercloud services.

PR #2224, opened 2020-09-01
[policies] Make IndependentPlugin work and switch 17 to use it

BryanQuigley:
Switch valid_subclasses to be something we can just add to and default
it to include IndependentPlugin.

17 plugins switched are those that generally are going to activate via
files =, commands or the like.  Not packages. This will allow them to run
without sos being able to determine the OS release.

Closes: #2018

---

### DIFF
--- a/sos/policies/suse.py
+++ b/sos/policies/suse.py
@@ -29,7 +29,7 @@ class SuSEPolicy(LinuxPolicy):
         self.package_manager = PackageManager(
             'rpm -qa --queryformat "%{NAME}|%{VERSION}\\n"',
             remote_exec=remote_exec)
-        self.valid_subclasses = [SuSEPlugin, RedHatPlugin]
+        self.valid_subclasses += [SuSEPlugin, RedHatPlugin]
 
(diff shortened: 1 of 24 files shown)


sos/policies/suse.py
```diff
@@ -29,7 +29,7 @@ def __init__(self, sysroot=None, init=None, probe_runtime=True,
         self.package_manager = PackageManager(
             'rpm -qa --queryformat "%{NAME}|%{VERSION}\\n"',
             remote_exec=remote_exec)
-        self.valid_subclasses = [SuSEPlugin, RedHatPlugin]
+        self.valid_subclasses += [SuSEPlugin]
```
TurboTurtle, line 32: AFAIK, the SuSE policy was loading the `RedHatPlugin` tagging class as well on purpose, as there is a large overlap of compatibility there, and there are only 5 SuSE-tagged plugins to date.

Changing this would result in a loss of about 80 plugins on SuSE. I'd say for now let's keep the RH tag in the policy.

PR #2230, opened 2020-09-10
Add snapped dockerd service

nicolasbock:
When the docker service is installed via snap then the systemd service
is called `snap.docker.dockerd.service` and the test whether the
service is running fails. This change adds the snap service name so
that the `DockerContainerRuntime` can detect it.

---

### DIFF
--- a/sos/policies/__init__.py
+++ b/sos/policies/__init__.py
@@ -217,7 +217,8 @@ class DockerContainerRuntime(ContainerRuntime):
     def check_is_active(self):
         # the daemon must be running
         if (is_executable('docker') and
-                self.policy.init_system.is_running('docker')):
+                (self.policy.init_system.is_running('docker') or
+                 self.policy.init_system.is_running('snap.docker.dockerd'))):
             self.active = True
             return True
         return False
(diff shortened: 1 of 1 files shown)


TurboTurtle:
Not sure why Travis isn't part of the automatic checks, but the jobs are still running. There's a PEP8 issue:

```
sos/policies/__init__.py:221:80: E501 line too long (80 > 79 characters)
The command "flake8 sos tests bin/*" exited with 1.
```

Dropping the indentation level on line 220 will do the trick:

```python
    def check_is_active(self):
        # the daemon must be running
        if (is_executable('docker') and
            (self.policy.init_system.is_running('docker') or
                self.policy.init_system.is_running('snap.docker.dockerd'))):
            self.active = True
            return True
        return False
```

nicolasbock:
Oh, thanks for the info @TurboTurtle . Will fix in the next iteration.

nicolasbock:
Would there be interest in using `tox` in this project? That would make testing this kind of stuff easier locally.

TurboTurtle (approved):

pmoravec (approved):

Issue #2231, opened 2020-09-10
Quoting for container commands seems to be broken for `docker` runtime


nicolasbock:
The call to [`fmt_container_cmd`](https://github.com/sosreport/sos/blob/master/sos/report/plugins/rabbitmq.py#L37) quotes the command including the whitespace between `rabbitmqctl` and `report` which breaks the `docker exec` command. The quoted command is:
```
docker exec  rabbitmq-container 'rabbitmqctl report'
```

Running this command from the commandline results in the following:
```
$ docker exec  rabbitmq-container 'rabbitmqctl report'
OCI runtime exec failed: exec failed: container_linux.go:349: starting container process caused "exec: \"rabbitmqctl report\": executable file not found in $PATH": unknown
```

According to upstream docker docs quotes are not supported for the `exec` command [[1]](https://docs.docker.com/engine/reference/commandline/exec/#extended-description).

TurboTurtle:
Looking into this on the podman side, it looks to be similar. However, on older versions I do see quoting working (oddly, intermittently).

I think the best way to address this would be to make the default to not quote but have an option to enable quoting - if for no other reason than being ready in the event of new container runtimes that use or even want quoting.

nicolasbock:
Thanks. I'll see if I can change it.

nicolasbock:
Please have a look at #2232 @TurboTurtle . I hope I caught all of the places where the extra `quote` parameter are needed.

Thanks for the hint!

PR #2232, opened 2020-09-10
Make quoting of container cmd optional

nicolasbock:
In case of the `docker` runtime (and potentially also in case of
`podman`) quoting a command that consists of multiple words separated
by whitespace leads to an incorrectly quoted command that fails to
execute. Whitespaces are explicitly not supported by upstream `docker`
for the `docker exec` command [1].

[1] https://docs.docker.com/engine/reference/commandline/exec/#extended-description

Fixes: #2231

---

### DIFF
--- a/sos/policies/__init__.py
+++ b/sos/policies/__init__.py
@@ -181,7 +181,7 @@ class ContainerRuntime(object):
                     vols.append(ent[-1])
         return vols
 
-    def fmt_container_cmd(self, container, cmd):
+    def fmt_container_cmd(self, container, cmd, quotecmd):
         """Format a command to run inside a container using the runtime
 
         :param container: The name or ID of the container in which to run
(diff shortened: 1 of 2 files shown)


sos/policies/__init__.py
```diff
@@ -190,10 +190,17 @@ def fmt_container_cmd(self, container, cmd):
...
+
         :returns: Formatted string to run `cmd` inside `container`
         :rtype: ``str``
         """
-        return "%s %s %s" % (self.run_cmd, container, quote(cmd))
+        if quote:
+            quoted_cmd = quote(cmd)
+        else:
+            quoted_cmd = cmd
+        return "%s %s %s" % (self.run_cmd, container, quoted_cmd)
```
TurboTurtle, line 203: There's a potential of a scoping issue (and slightly readability to) by using `quote` as both a variable and a method. It's fine if the parameter to `Plugin.fmt_container_runtime()` is `quote`, but for `ContainerRuntime.fmt_container_runtime()` let's name it something else to avoid that issue. E.G.

```python
    def fmt_container_cmd(self, container, cmd, quotecmd=False):
        """Format a command to run inside a container using the runtime

        :param container: The name or ID of the container in which to run
        :type container: ``str``

        :param cmd: The command to run inside `container`
        :type cmd: ``str``

        :param quotecmd: Should the command be quoted for execution
        :type quotecmd: ``bool``

        :returns: Formatted string to run `cmd` inside `container`
        :rtype: ``str``
        """
        return "%s %s %s" % (self.run_cmd,
                             container,
                             quote(cmd) if quotecmd else cmd)
```

nicolasbock:
For consistency I changed all parameters to `quotecmd`.

TurboTurtle (approved):

pmoravec (approved):
I would believe that the behaviour described for `docker` applies the same way also for `podman`.

Issue #1992, opened 2020-03-23
[4.0] Add Snap and Flatpak support


TurboTurtle:
Adding flatpak and snaps packaging as requested in #1959. As noted, this would lower the barrier to `sos` shipping in additional distributions.

slashdd:
Would be really nice to get that support feature in next release. Plus having sosreport support more than 1 package manager for the same host. (e.g. In Ubuntu DEB and snap co-exist within the same system)

BryanQuigley:
FYI, I gave up on adding Flatpak support because I couldn't find a current sos plugin where it would actually be useful to be used.

slashdd:
This issue is not fixed. The intent of this bug was to make sure sos support SNAP FlatPak in PackageManager() just like it does for RPM and DEB.

The PR suggested by @arif-ali is the configuration to build sos as a SNAP, which is a different thing.

slashdd:
Main discussion: https://github.com/sosreport/sos/issues/2237

BryanQuigley:
The big issue that stopped me from pursuing flatpak was the apps are not designed to be called without a GUI of some kind. So once you get the list, what would you use it for it?

If there are multiple apps with the same name it does seem wise to flag that in some way - there can also be user installed flatpaks - but I'm not sure if there is a good way to pick up on them.

arif-ali:
Now that we have both snap and flatpack package managers, I presume we can close this one now?

PR #2233, opened 2020-09-10
Add snap configuration file

nicolasbock:
Closes: #1992

---

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ jobs:
       dist: bionic
       language: python
       python: "3.6"
-      install: pip install -r requirements.txt; python3 setup.py install;
+      install: pip install -r test-requirements.txt; python3 setup.py install;
       script:
         - "nosetests -v --with-cover --cover-package=sos --cover-html"
         - "sudo ./tests/simple.sh ~/virtualenv/python$TRAVIS_PYTHON_VERSION/bin/python"
@@ -40,7 +40,7 @@ jobs:
       dist: bionic
       language: python
       python: "3.7"
-      install: pip install -r requirements.txt; python3 setup.py install;
+      install: pip install -r test-requirements.txt; python3 setup.py install;
       script:
         - "nosetests -v --with-cover --cover-package=sos --cover-html"
         - "sudo ./tests/simple.sh ~/virtualenv/python$TRAVIS_PYTHON_VERSION/bin/python"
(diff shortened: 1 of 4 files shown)


snapcraft.yaml
```diff
@@ -0,0 +1,24 @@
+name: sosreport
```
TurboTurtle, line 1: For new packaging, I think using the project name of `sos` is a better approach, as the project is now more than the report functionality.

.travis.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       dist: bionic
       language: python
       python: "3.6"
-      install: pip install -r requirements.txt; python3 setup.py install;
+      install: pip install -r test-requirements.txt; python3 setup.py install;
```
TurboTurtle, line 34: I don't see why we'd need to split the requirements into 2 files. Not least of which, this stops pulling in pexpect which once we have testing for `collect` figured out, we'd need to add it to the new file as well. I don't see a benefit in splitting out the requirements for a single different module to install.

TurboTurtle (changes requested):
I realize this is still a draft, but a few comments below.

Additionally, please include an informative commit message beyond an issue reference, for historical tracking when we look back at this in the future.

snapcraft.yaml
nicolasbock: :+1:

.travis.yml
nicolasbock: I am still experimenting here. The main reason I split the requirements though is that packages such as `pycodestyle` are not necessary for `sos` and are only needed for testing (unless I am missing something).

nicolasbock:
I am getting an error when building the snap:
```
/root/parts/sosreport/install/usr/bin/python3 setup.py --no-user-cfg install --single-version-externally-managed --user --record install.txt
usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
   or: setup.py --help [cmd1 cmd2 ...]
   or: setup.py --help-commands
   or: setup.py cmd --help

error: option --single-version-externally-managed not recognized
```

It's not clear to me what this error message means at this point...

nicolasbock:
@TurboTurtle Eventually, once this here is working properly, the `sos` project needs to register the `sos` snap in the snap store [1].Please note also that the snap requires `classic` confinement which means less sandboxing basically so snapped `sos` can access the various system files it needs to access. However, `classic` confinement is discouraged because of the potential security implications but can be requested for apps that actually need it. I have done so [2].

[1] https://snapcraft.io/register-snap
[2] https://forum.snapcraft.io/t/request-for-classic-confinement-for-sos/19912

TurboTurtle:
Wanted to cycle back around on this, is this still desired for inclusion in 4.1? What's outstanding for this change?

nicolasbock:
Closing because of #2676

PR #2238, opened 2020-09-17
[cleaner] Fix two conditions where files can be incorrectly skipped

TurboTurtle:
Fixes two separate issues leading to incorrect file skipping.

First, fixes a condition where a file would be skipped if the name of the archive being cleaned appears more than once in the filename.

Second, fixes a regex issue that was too broad for matching against tarballs.

---

### DIFF
--- a/sos/cleaner/obfuscation_archive.py
+++ b/sos/cleaner/obfuscation_archive.py
@@ -73,7 +73,9 @@ class SoSObfuscationArchive():
             'sys/kernel/debug',
             'sys/module',
             'var/log/.*dnf.*',
-            '.*.tar.*',  # TODO: support archive unpacking
+            '.*\.tar$',  # TODO: support archive unpacking
+            # Be explicit with these tar matches to avoid matching commands
+            '.*\.tar\.xz',
             '.*.gz'
         ]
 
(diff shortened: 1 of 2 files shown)


slashdd (approved):
lgtm

sos/cleaner/obfuscation_archive.py
```diff
@@ -73,7 +73,9 @@ def _load_skip_list(self):
             'sys/kernel/debug',
             'sys/module',
             'var/log/.*dnf.*',
-            '.*.tar.*',  # TODO: support archive unpacking
+            '.*\.tar$',  # TODO: support archive unpacking
+            # Be explicit with these tar matches to avoid matching commands
+            '.*\.tar\.xz',
```
slashdd, line 78: What about `.*\.tar\.[g,x]z` ?

slashdd: Ignore this, I just noticed line 79.

TurboTurtle: Yeah, I thought about doing that too but I'm err'ing on the side of caution of not wanting another surprise like this down the road.

pmoravec (approved):

PR #2239, opened 2020-09-18
[rabbitmq] Add 10 second timeout to call to `maybe_stuck()`.

nicolasbock:
The `maybe_stuck()` function might run for a long time if it is
tracking down a lot of potentially stuck processes. In order to
prevent long run times of `sos report` the `cmd` timeout should be set
explicitly. As a compromise between acceptable execution times and
information gathered from the plugin, this change introduces a timeout
of 10 seconds for the `maybe_stuck()` call.

---


pmoravec (approved):
IMHO it is fine to have even higher timeouts for a command (though timeout per container can bluw up in sum), but as having no experience with the command usual duration, I rely on the commiter judgement.

ACK from me.

TurboTurtle (approved):

slashdd (approved):
lgtm

PR #2242, opened 2020-09-21
[openvswitch] Adding 6wind commands

valleedelisle:
This adds the 6wind's commands to openvswitch plugin.

---

### DIFF
--- a/sos/report/plugins/openvswitch.py
+++ b/sos/report/plugins/openvswitch.py
@@ -33,6 +33,31 @@ class OpenVSwitch(Plugin):
             "ovs-vsctl -t 5 get Open_vSwitch . other_config:dpdk-init")
         check_dpdk = (dpdk_enabled["status"] == 0 and
                       dpdk_enabled["output"].startswith('"true"'))
+        check_6wind = any([self.is_installed(p) for p in
+                           ['6windgate-fp', 'nuage-openvswitch']])
+        actl = "ovs-appctl"
+
+        files_6wind = [
+            "/etc/systemd/system/multi-user.target.wants/openvswitch.service",
+            "/etc/sysctl.d/60-6wind-system-auto-reboot.conf",
+            "/etc/openvswitch/system-id.conf",
+            "/etc/openvswitch/*.db",
+            "/etc/ld.so.conf.d/linux-fp-sync-fptun.conf",
+            "/etc/NetworkManager/conf.d/fpn0.conf",
+            "/etc/default/openvswitch",
+            "/etc/logrotate.d/openvswitch",
+            "/etc/linux-fp-sync.env",
+            "/etc/fp-daemons.env",
+            "/etc/fp-vdev.ini",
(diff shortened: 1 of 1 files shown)


TurboTurtle (changes requested):
This contains basically the entirety of the existing `openvswitch` plugin, sans a handful of omissions.

Rather than duplicate all that, and then have 2 places to maintain openvswitch code, we should take 1 of 2 paths:

1. Simply add a package check within the existing `openvswitch` plugin for the new `6windgate-fp` package, and gate the `fp *` commands et al behind that.

2. Trim out all the duplicated `openvswitch` collections here, leaving just the actual 6wind bits.

As I understand it 6wind is an extension of openvswitch (ie everything in ovs is in 6wind, but not the other way around), so I have a slight preference for option 1, but I don't have any particular opposition to option 2. I'll leave that decision up to you but as for carrying a second copy of the "main" openvswitch collections, that's a hard nack.

valleedelisle:
@TurboTurtle that makes total sense. I wasn't sure if I should modify the current ovs plugin or create a new one. I'll add the 6wind bits to the current ovs package and push a new commit.

sos/report/plugins/openvswitch.py
```diff
@@ -33,6 +33,31 @@ def setup(self):
             "ovs-vsctl -t 5 get Open_vSwitch . other_config:dpdk-init")
         check_dpdk = (dpdk_enabled["status"] == 0 and
                       dpdk_enabled["output"].startswith('"true"'))
+        check_6wind = any([self.is_installed(p) for p in
+                           ['6windgate-fp', 'nuage-openvswitch']])
+        actl = "ovs-appctl"
```
pmoravec, line 38: Why introducing this variable, that is in fact a constant that makes the `add_cmd_output` commands less readable? Some preparation for conditional assignment of the variable in future?

I am not against it per se, just.. the code sounds less obvious with the constant variable.

ACK otherwise.

valleedelisle: That's because of the 80 column limit. It was easier to do this than try to split the strings IMHO.

pmoravec (approved):
That makes sense, then.

TurboTurtle:
Code looks fine to me, just wondering if we can't make the nested 6wind `ovs-appctl` commands a bit easier to read.

I'll play with this a bit on my end to see if I can come up with anything, but if I don't get something by EOD then this looks good to me for merge.

valleedelisle:
Thanks guys. I'll submit a PR for legacy-3.9 as well if you don't mind. Is there anything special in the process? That woud be great to land this in RHEL7 if possible at somepoint.

pmoravec:
Nothing special with PRs to `legacy-3.9`. Just we generally recommend filing it after upstream PR is merged, just in case some change in commit happens that should be reflected in the backport to `legacy-3.9`.

TurboTurtle (approved):

PR #2248, opened 2020-09-25
[openvswitch] Adding 6wind commands

valleedelisle:
This adds the 6wind's commands to openvswitch plugin.

---

### DIFF
--- a/sos/plugins/openvswitch.py
+++ b/sos/plugins/openvswitch.py
@@ -13,10 +13,13 @@ from sos.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
 from os.path import join as path_join
 from os import environ
 
+import re
+
 
 class OpenVSwitch(Plugin):
     """ OpenVSwitch networking
     """
+
     plugin_name = "openvswitch"
     profiles = ('network', 'virt')
 
@@ -27,6 +30,36 @@ class OpenVSwitch(Plugin):
         log_dirs = [
             '/var/log/openvswitch/',
             '/usr/local/var/log/openvswitch/',
+        ]
+
+        dpdk_enabled = self.collect_cmd_output(
+            "ovs-vsctl -t 5 get Open_vSwitch . other_config:dpdk-init")
+        check_dpdk = (dpdk_enabled["status"] == 0 and
+                      dpdk_enabled["output"].startswith('"true"'))
+        check_6wind = any([self.is_installed(p) for p in
+                           ['6windgate-fp', 'nuage-openvswitch']])
+        actl = "ovs-appctl"
+
+        files_6wind = [
+            "/etc/systemd/system/multi-user.target.wants/openvswitch.service",
+            "/etc/sysctl.d/60-6wind-system-auto-reboot.conf",
+            "/etc/openvswitch/system-id.conf",
+            "/etc/openvswitch/*.db",
+            "/etc/ld.so.conf.d/linux-fp-sync-fptun.conf",
+            "/etc/NetworkManager/conf.d/fpn0.conf",
+            "/etc/default/openvswitch",
+            "/etc/logrotate.d/openvswitch",
+            "/etc/linux-fp-sync.env",
+            "/etc/fp-daemons.env",
+            "/etc/fp-vdev.ini",
+            "/etc/fpm.env",
+            "/etc/6WINDGate/fp.config",
+            "/etc/6WINDGate/fpnsdk.config",
+            "/etc/dms.d/fp-dms.conf",
+            "/etc/dms.d/fpmd-dms.conf",
+            "/etc/dms.d/fpsd-dms.conf",
+            "/etc/fast-path.env",
+            "/etc/fps-fp.env",
         ]
 
         if environ.get('OVS_LOGDIR'):
@@ -68,6 +101,7 @@ class OpenVSwitch(Plugin):
             "ovs-appctl bond/show",
             # Capture LACP details
             "ovs-appctl lacp/show",
+            "ovs-appctl lacp/show-stats",
             # Capture coverage stats"
             "ovs-appctl coverage/show",
             # Capture cached routes
@@ -90,6 +124,13 @@ class OpenVSwitch(Plugin):
             "ovs-appctl dpctl/show -s",
             # Capture DPDK datapath flows
             "ovs-appctl dpctl/dump-flows",
+            # The '-s' option enables dumping of packet counters on the
+            # ports.
+            "ovs-dpctl -s show",
+            # Capture the in-kernel flow information if it exists
+            "ovs-dpctl dump-flows -m",
+            # Capture the flow information also for offloaded rules
+            "ovs-dpctl dump-flows type=offloaded -m",
             # Capture DPDK queue to pmd mapping
             "ovs-appctl dpif-netdev/pmd-rxq-show",
             # Capture DPDK pmd stats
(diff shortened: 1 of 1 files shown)


sos/plugins/openvswitch.py
```diff
@@ -13,10 +13,12 @@
 from os.path import join as path_join
 from os import environ
 
+import re
+
```
pmoravec, line 17: Just curious: why is this import required here, but it is missing in (otherwise in-par) the plugin in `master` branch?

TurboTurtle: I pushed a fixup commit to `master` after this one that added the import. The plugin does a regex match later one:

https://github.com/sosreport/sos/pull/2248/files#diff-0a9522d996653ba1be2e296998f68ea1R171

sos/plugins/openvswitch.py
```diff
@@ -13,10 +13,12 @@
...
 
 class OpenVSwitch(Plugin):
-    """ OpenVSwitch networking
-    """
+
+    short_desc = 'OpenVSwitch networking'
```
TurboTurtle, line 21: The change to `short_desc` is a sos-4.0 thing, the legacy branch still uses the class docstrings.

sos/plugins/openvswitch.py
```diff
@@ -29,6 +31,36 @@ def setup(self):
             '/usr/local/var/log/openvswitch/',
         ]
 
+        dpdk_enabled = self.call_ext_prog(
```
TurboTurtle, line 34: `legacy-3.9` does not include `call_ext_prog()` in the API.

If you need to run something during `setup()`, get the output returned but _not_ write that output to the archive, use `Plugin.exec_cmd()`. If you do need to save the output as well, use `Plugin.collect_cmd_output()`.

sos/plugins/openvswitch.py
```diff
@@ -43,20 +75,16 @@ def setup(self):
         ])
 
         self.add_cmd_output([
-            # The '-s' option enables dumping of packet counters on the
-            # ports.
-            "ovs-dpctl -s show",
-            # Capture the in-kernel flow information if it exists
-            "ovs-dpctl dump-flows -m",
-            # Capture the flow information also for offloaded rules
-            "ovs-dpctl dump-flows type=offloaded -m",
```
TurboTurtle, line 52: It's not clear to me why these are getting dropped with this commit.

sos/plugins/openvswitch.py
```diff
@@ -104,32 +131,93 @@ def setup(self):
...
+                "fp-cli iface",
+                "ovs-appctl memory/show",
+            ])
+            self.add_journal(units="virtual-accelerator")
+            for table in ['filter', 'mangle', 'raw', 'nat']:
+                self.add_cmd_output(["fpcmd nf4-rules %s" % table])
+
+            # 6wind doesn't care on which bridge the ports are, there's only
+            # one bridge and it's alubr0
+            port_list = self.call_ext_prog("fp-cli fp-vswitch-ports")
```
TurboTurtle, line 168: Same here, use `exec_cmd()` or `collect_cmd_output()`

sos/plugins/openvswitch.py
```diff
@@ -104,32 +131,93 @@ def setup(self):
...
+                    "%s dpctl/show -s %s" % (actl, dp),
+                    "%s dpctl/dump-flows -m %s" % (actl, dp),
+                    "%s dpctl/dump-conntrack -m %s" % (actl, dp),
+                    "%s dpctl/ct-stats-show -m %s" % (actl, dp),
+                    "%s dpctl/ipf-get-status %s" % (actl, dp),
+                ])
+
         # Gather additional output for each OVS bridge on the host.
-        br_list_result = self.collect_cmd_output("ovs-vsctl list-br")
+        br_list_result = self.call_ext_prog("ovs-vsctl -t 5 list-br")
```
TurboTurtle, line 191: Same here - this should be reverted back to the original `collect_cmd_output()`

sos/plugins/openvswitch.py
```diff
@@ -104,32 +131,93 @@ def setup(self):
...
 
                 # Flow protocols currently supported
                 flow_versions = [
                     "OpenFlow10",
                     "OpenFlow11",
                     "OpenFlow12",
-                    "OpenFlow13"
+                    "OpenFlow13",
+                    "OpenFlow14",
+                    "OpenFlow15"
```
TurboTurtle, line 215: Is this intentional, or a result of copying the `master` plugin over the legacy one?

sos/plugins/openvswitch.py
```diff
@@ -104,32 +131,93 @@ def setup(self):
...
                 ]
 
                 # List protocols currently in use, if any
-                ovs_list_bridge_cmd = "ovs-vsctl list bridge %s" % br
-                br_info = self.collect_cmd_output(ovs_list_bridge_cmd)
+                ovs_list_bridge_cmd = "ovs-vsctl -t 5 list bridge %s" % br
+                br_info = self.call_ext_prog(ovs_list_bridge_cmd)
```
TurboTurtle, line 220: Same, use `collect_cmd_output()`

sos/plugins/openvswitch.py
```diff
@@ -147,24 +235,81 @@ def setup(self):
                             "ovs-ofctl -O %s dump-ports-desc %s" % (flow, br)
                         ])
 
-        # Gather info on the DPDK mempools associated with each DPDK port
-        br_list_result = self.collect_cmd_output("ovs-vsctl -t 5 list-br")
-        if br_list_result['status'] == 0:
-            for br in br_list_result['output'].splitlines():
-                port_list_result = self.exec_cmd(
+                port_list_result = self.call_ext_prog(
```
TurboTurtle, line 238: Same, in this case `exec_cmd()` is the correct approach.

sos/plugins/openvswitch.py
```diff
@@ -147,24 +235,81 @@ def setup(self):
...
+                        ])
+
+                        if check_dpdk:
+                            self.add_cmd_output(
+                                "ovs-appctl netdev-dpdk/get-mempool-info %s" %
+                                port
+                            )
+
+                if check_dpdk:
+                    iface_list_result = self.call_ext_prog(
```
TurboTurtle, line 258: Same here, do not try to use `call_ext_prog()` as it does not exist in the 3.9 API.

pmoravec:
Ah I see..

Mocking the commands, I found one gotcha in upstream version of `openvswitch` plugin that might deserve backporting as well : #2249. Worth adding the commit to this branch or squashing it into one?

TurboTurtle (changes requested):
This commit does a lot more than just add the 6wind commands. It seems to try to backport the entire `master` copy of the plugin, but it gets a bit mangled in the process. There's bits of the old API bleeding through, and there are changes being made that are unrelated to 6wind.

For the purpose of this commit, this should be exclusively focused on adding the 6wind commands and nothing else. If there are other issues with the `legacy-3.9` version of this plugin then separate commits should correct and document those.

sos/plugins/openvswitch.py
valleedelisle: This was semi-intentional, I don't see any downside in adding this.

sos/plugins/openvswitch.py
valleedelisle: These commands were were removed from the master with this commit eb145d9484cae850f6fc4984e2e229486241c833

I believe that if we put them back here, we should put them back on master, but I'll let you judge.

sos/plugins/openvswitch.py
valleedelisle: I tested this with sos-3.6 which didn't have collect_cmd_output(), I'll change this.

sos/plugins/openvswitch.py
valleedelisle: > This commit does a lot more than just add the 6wind commands. It seems to try to backport the entire `master` copy of the plugin, but it gets a bit mangled in the process. There's bits of the old API bleeding through, and there are changes being made that are unrelated to 6wind.
> 
> For the purpose of this commit, this should be exclusively focused on adding the 6wind commands and nothing else. If there are other issues with the `legacy-3.9` version of this plugin then separate commits should correct and document those.

Just wondering if it's an issue to sync up master's version with legacy? I thought this was a good idea.

valleedelisle:
> Ah I see..
> 
> Mocking the commands, I found one gotcha in upstream version of `openvswitch` plugin that might deserve backporting as well : #2249. Worth adding the commit to this branch or squashing it into one?
> 
> ACK otherwise.

Added #2249 to this commit.

sos/plugins/openvswitch.py
TurboTurtle: Well, the main thing is sync'ing the plugin with `master` is entirely separate from adding the 6wind commands, which is what this commit documents it does. If a sync is desired, then it should be handled in a commit that explicitly highlights that it is syncing the legacy version to the current master, and why that is needed.

Then there's the matter of determining if the dropped commands are indeed not useful for any system that might install a packaged version of sos from the `legacy-3.9` branch.

valleedelisle: That makes sense. I re-added them. We should consider to put them back on the master branch actually, this is lacking.

pmoravec (approved):

TurboTurtle (changes requested):
My main issue with this PR remains - it does more than the commit details it does.

If the plugin needs to be sync'd with master for legacy, then that needs to be done in either 1) a separate commit or 2) in a replacement commit that brings the current master plugin, with the 6wind changes, into legacy.

In either case this means at minimum

1. The commit message needs to clearly state the change being made
2. If a full sync is desired, we need to be sure that this won't cause issues on systems installing a `sos` package built from legacy. For RHEL, this means RHEL 7.9 systems. Are we sure that sync'ing the master branch version of this plugin for 7.9 systems is appropriate?

valleedelisle:
> My main issue with this PR remains - it does more than the commit details it does.
> 
> If the plugin needs to be sync'd with master for legacy, then that needs to be done in either 1) a separate commit or 2) in a replacement commit that brings the current master plugin, with the 6wind changes, into legacy.
> 
> In either case this means at minimum
> 
> 1. The commit message needs to clearly state the change being made
> 2. If a full sync is desired, we need to be sure that this won't cause issues on systems installing a `sos` package built from legacy. For RHEL, this means RHEL 7.9 systems. Are we sure that sync'ing the master branch version of this plugin for 7.9 systems is appropriate?

I reverted most of the master sync change, the only thing added are the OpenFlow. Should I remove them and resync in another commit?

TurboTurtle:
There's still a number of other changes being brought forward:

* The OpenFlow updates from https://github.com/sosreport/sos/commit/e1b474af6bccee152fcdaa247e743d924348a71c
* There's the updates from https://github.com/sosreport/sos/commit/bc0c02453e1746083173301857edfc7fdccb8a9c
* The `ls` updates from https://github.com/sosreport/sos/commit/27ef2569734f3d99eb7bfcdee721b364beaa8114
* The datapath changes from https://github.com/sosreport/sos/commit/eb145d9484cae850f6fc4984e2e229486241c833

jjansky1:
How are we with this PR ? What can i do for successful merge? It will not be in current z-stream batch, but will be glad if we can move forward in next batch.

sos/plugins/openvswitch.py
```diff
@@ -68,6 +101,7 @@ def setup(self):
             "ovs-appctl bond/show",
             # Capture LACP details
             "ovs-appctl lacp/show",
+            "ovs-appctl lacp/show-stats",
```
pmoravec, line 104: This was added by a different upstream PR, but let's ok..

sos/plugins/openvswitch.py
```diff
@@ -90,6 +124,13 @@ def setup(self):
             "ovs-appctl dpctl/show -s",
             # Capture DPDK datapath flows
             "ovs-appctl dpctl/dump-flows",
+            # The '-s' option enables dumping of packet counters on the
+            # ports.
+            "ovs-dpctl -s show",
+            # Capture the in-kernel flow information if it exists
+            "ovs-dpctl dump-flows -m",
+            # Capture the flow information also for offloaded rules
+            "ovs-dpctl dump-flows type=offloaded -m",
```
pmoravec, line 133: I cant find them in upstream anywhere - why do you need them in legacy branch but not in upstream?

sos/plugins/openvswitch.py
```diff
@@ -147,24 +236,81 @@ def setup(self):
                             "ovs-ofctl -O %s dump-ports-desc %s" % (flow, br)
                         ])
 
-        # Gather info on the DPDK mempools associated with each DPDK port
-        br_list_result = self.collect_cmd_output("ovs-vsctl -t 5 list-br")
-        if br_list_result['status'] == 0:
-            for br in br_list_result['output'].splitlines():
-                port_list_result = self.exec_cmd(
+                port_list_result = self.collect_cmd_output(
```
pmoravec, line 239: You need extra collecting the `"ovs-vsctl -t 5 list-ports %s"` command, that you replace `exec_cmd` (as in upstream) by `collect_cmd_output`?

pmoravec:
> Update: just discussed @valleedelisle and we agreed to change this PR to backport just #2242 . If others will be needed to backport new BZs should be created or at least 1 major BZ and 4 other PRs to backport 1 by 1.

I do see the PR contains (all?) #2242 but also some changes from other PRs and also at least two changes not present in upstream - see my code comments.

TurboTurtle:
Closing in favor of #2484 which has been merged into `legacy-3.9`

Issue #2049, opened 2020-05-06
[conntrackd] Support Debian/Ubuntu and capture conntrack info


hillpd:
The plug-in checks for "conntrack-tools" but doesn't capture any debug information from the `conntrack` utility. On Debian/Ubuntu the tools package is simply named "conntrack" which should be included.

Recommend adding the following command output:
`conntrack -L -o extended`
`conntrack -S`

ddstreet:
@slashdd @BryanQuigley

jay-vosburgh:
"conntrack" is a separate package (on Ubuntu), so perhaps the cleanest way forward is to add a plugin for conntrack itself, and capture -L, -S, etc, via the separate plugin.

BryanQuigley:
@TurboTurtle I'm thinking just add to current plugin makes sense.  It could also be separated as two plugins `conntract`  and `contractd` (On ubuntu conntract is supported, while conntractd is not).

TurboTurtle:
Wow, this excerpt from [the manual](http://conntrack-tools.netfilter.org/manual.html) really hits the nail on the head

```
Although the name of both tools is very similar - and you can blame me for that, I'm not a marketing guy - they are used for very different tasks.
```

At any rate, since both utilities are at least logically coupled together I'd be in favor of adding this to the current plugin, rather than splitting out a new one.

slashdd:
sosreport is fault tolerant (if cmd doesn't exist, it will continue) so I personally would simply add the `conntrack` command in the current plugin with the 'conntrack' package name as one of the possible triggers. And maybe rename the plugin from `conntrackd.py `to `conntrack.py` to make it more connection tracking specifics ?

pmoravec:
+1 to have just one plugin, and rename it to `contrack`.

PR #2251, opened 2020-09-27
[conntrackd] add conntrack info

hemanthnakkina:
The plugin conntrackd is renamed to conntrack. Added the following
conntrack commands to the plugin.
conntrack -L -o extended
conntrack -S

Closes: #2049

Signed-off-by: Hemanth Nakkina hemanth.nakkina@canonical.com

---

### DIFF
--- a/sos/report/plugins/conntrack.py
+++ b/sos/report/plugins/conntrack.py
@@ -11,16 +11,17 @@ from sos.report.plugins import (Plugin, RedHatPlugin, DebianPlugin,
                                 UbuntuPlugin, SuSEPlugin)
 
 
-class Conntrackd(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin, SuSEPlugin):
+class Conntrack(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin, SuSEPlugin):
 
-    short_desc = 'conntrackd - netfilter connection tracking user-space daemon'
+    short_desc = 'conntrack - netfilter connection tracking'
 
-    plugin_name = 'conntrackd'
+    plugin_name = 'conntrack'
     profiles = ('network', 'cluster')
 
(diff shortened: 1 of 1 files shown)


TurboTurtle (commented):
The code change looks fine, however is there actually a reason the plugin needs to be renamed?

hemanthnakkina:
> The code change looks fine, however is there actually a reason the plugin needs to be renamed? EDIT: and if there is, it should be included in the commit message.

The plugin is renamed as per the comments in the bug.

TurboTurtle:
The issue suggests a rename, sure, but it doesn't provide an actual reason for it. Usually plugin renames follow upstream conventions (see openswan, snap). It seems both `conntrack` and `conntrackd` are still actively used.

If we're going to rename the plugin, I'd like it to be for a tangible reason.

slashdd:
It all came down because Ubuntu offers them into two separate binary package as follows:
```
conntrack - Program to modify the conntrack tables
conntrackd - Connection tracking daemon
```

- Scenario 1: Keep it as is and add both `conntrackd` and `conntrack` within the same `conntrackd` plugin (With appropriate package triggering)
- Scenario 2: Add a new plugin `conntrack`
- Scenario 3: Rename it to `conntrack` because it is more general and more "Connection tracking" oriented than specific to a connection tracking daemon, like the actual name suggest.

I'm +1 with all scenarios, but have a preference for Scenario 3, so that conntrack.py will gather all "Connection tracking" collection (`conntrack` , `conntrackd`) under the same plugin. And changing the name will make the plugin less specific to a deamon and more general and open to add stuff outside the daemon context.

TurboTurtle:
Works for me

TurboTurtle (approved):

pmoravec (approved):

PR #2253, opened 2020-09-30
[cleaner] Add username parser

TurboTurtle:
Adds a username parser for native obfuscation of usernames that appear
in `lastlog` output as collected by the `login` plugin.

Users may also supply additional usernames not appearing in `lastlog`
output via the `--usernames` option for either report, clean, or
collect.

Resolves: #2253

---

### DIFF
--- a/sos/cleaner/parsers/username_parser.py
+++ b/sos/cleaner/parsers/username_parser.py
@@ -1,0 +1,58 @@
+# Copyright 2020 Red Hat, Inc. Jake Hunsaker <jhunsake@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+
+from sos.cleaner.parsers import SoSCleanerParser
+from sos.cleaner.mappings.username_map import SoSUsernameMap
+
+
+class SoSUsernameParser(SoSCleanerParser):
+    """Parser for obfuscating usernames within an sosreport archive.
+
+    Note that this parser does not rely on regex matching directly, like most
+    other parsers do. Instead, usernames are discovered via scraping the
+    collected output of lastlog. As such, we do not discover new usernames
+    later on, and only usernames present in lastlog output will be obfuscated,
+    and those passed via the --usernames option if one is provided.
+    """
+
+    name = 'Username Parser'
+    map_file_key = 'username_map'
+    prep_map_file = 'sos_commands/login/lastlog_-u_1000-60000'
+    regex_patterns = []
+    skip_list = [
+        'nobody',
+        'nfsnobody',
+        'root'
+    ]
+
+    def __init__(self, conf_file=None, opt_names=None):
+        self.mapping = SoSUsernameMap()
+        super(SoSUsernameParser, self).__init__(conf_file)
+        self.mapping.load_names_from_options(opt_names)
+
+    def load_usernames_into_map(self, fname):
+        """Since we don't get the list of usernames from a straight regex for
+        this parser, we need to override the initial parser prepping here.
+        """
+        with open(fname, 'r') as lastfile:
+            for line in lastfile.read().splitlines()[1:]:
+                user = line.split()[0]
+                if user in self.skip_list:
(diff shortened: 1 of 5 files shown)


TurboTurtle:
Two things I'm unsure of in the current commit:

1. Is `sos_commands/login/lastlog_-u_1000-60000` actually the right file to reference for usernames? Should we instead iterate through the `lastlog` outputs for everything above UID 999? Meaning extend this to UIDs 1000-4294967295?

2. I have added `root` to the list of usernames to skip. In the current implementation, this should never actually appear in the list of usernames so it is more of a fail safe than anything else. My primary thought here is "I always want to know what root is doing", but perhaps that's wrong and root should be obfuscated like any other user? Of course, then we run into the argument that we should also be obfuscating all users with UIDs lower than 1000 as well which I don't think is beneficial.

sos/cleaner/parsers/username_parser.py
```diff
@@ -0,0 +1,58 @@
...
+    Note that this parser does not rely on regex matching directly, like most
+    other parsers do. Instead, usernames are discovered via scraping the
+    collected output of lastlog. As such, we do not discover new usernames
+    later on, and only usernames present in lastlog output will be obfuscated,
+    and those passed via the --usernames option if one is provided.
+    """
+
+    name = 'Username Parser'
+    map_file_key = 'username_map'
+    prep_map_file = 'sos_commands/login/lastlog_-u_1000-60000'
```
slashdd, line 28: Excellent choice to go with what I had implement in soscleaner to only obfuscate "internal" users created by useradd and such.

```
$ grep UID /etc/login.defs 
UID_MIN			 1000
UID_MAX			60000
```

Obfuscating public known users such as www-data, ubuntu, ... or else is unnecessary and based on my experience with soscleaner it will vastly contribute in having less false positive situations.

I'm definitely +1 on this approach.

slashdd:
> Two things I'm unsure of in the current commit:
> 
> 1. Is `sos_commands/login/lastlog_-u_1000-60000` actually the right file to reference for usernames? Should we instead iterate through the `lastlog` outputs for everything above UID 999? Meaning extend this to UIDs 1000-4294967295?

I think the focus should be on internal users created by the admin, and avoid anything publicly known (at least for now) my experience with soscleaner was a false positive nightmare without that.

```
/etc/login.defs
#
# Min/max values for automatic uid selection in useradd
#
UID_MIN                  1000
UID_MAX                 60000
```


> 2. I have added `root` to the list of usernames to skip. In the current implementation, this should never actually appear in the list of usernames so it is more of a fail safe than anything else. My primary thought here is "I always want to know what root is doing", but perhaps that's wrong and root should be obfuscated like any other user? Of course, then we run into the argument that we should also be obfuscating all users with UIDs lower than 1000 as well which I don't think is beneficial.

arif-ali:
What about users that are from sssd/ldap/Samba PDC, they will not be necessarily local users created by admin. So would be good to cater for those, if we can

slashdd:
> What about users that are from sssd/ldap/Samba PDC, they will not be necessarily local users created by admin. So would be good to cater for those, if we can

If the goal is to play safe and start small and humble w/ user obfuscation, I would definitely set barrier 1000-60000, because I predict false positive otherwise even if most false positive will be majority coming from UID lt 1000. I think it's fine to say we are obfuscation user added by useradd/useradd and such tools.

IMHO, I could see 1000-4294967295 as a logic second phase, once we have a good user experience sample

That's my 2 cents on this.

slashdd:
Ubuntu cloud packages uses a very high set of UID. If we don't set barrier with MAX_UID, we will face situation where nova user will be used to be obfuscated.

```
nova-common.postinst:NOVA_UID=64060

    if ! getent passwd nova > /dev/null 2>&1; then
        adduser --quiet --system \
            --home /var/lib/nova \
            --no-create-home \
            --uid $NOVA_UID \
            --gid $NOVA_GID \
            --shell /bin/false nova 2>/dev/null
    fi
```

pmoravec (approved):

TurboTurtle:
> Unless we maintain a very specific skip_list which IMHO will be a pain.

Yeah, this is where my hesitation comes in. Handling AD/IPA/IDM/etc... users would be a huge benefit, but the maintenance overhead is likely to be non-trivial, and at that we'd probably also have at least a handful of distro-dependent items to skip which means extending `Policy` to provide the right level of separation.

PR #2256, opened 2020-10-06
[plugins] Use `IndependentPlugin` tag in more plugins

TurboTurtle:
This commit updates 94 plugins to use `IndependentPlugin` as the tagging
class, instead of the distro-specific tags.

This tag is applied to plugins where _all_ of `RedHatPlugin`,
`DebianPlugin`, and `UbuntuPlugin` are imported and there is only a
single class definition for the plugin which uses all of those tagging
classes. `SuSEPlugin` is also accepted where it is explicitly imported,
as the SuSE policy already uses the `RedHatPlugin` tagging class as
well.

Resolves: #2256

---

### DIFF
--- a/sos/report/plugins/anacron.py
+++ b/sos/report/plugins/anacron.py
@@ -6,10 +6,10 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.report.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+from sos.report.plugins import Plugin, IndependentPlugin
 
 
-class Anacron(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
+class Anacron(Plugin, IndependentPlugin):
 
     short_desc = 'Anacron job scheduling service'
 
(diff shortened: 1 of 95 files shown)


pmoravec:
Why not also these plugins (`grep "RedHatPlugin, UbuntuPlugin, DebianPlugin" -c` returns 2 :) )?

- ata
- btrfs
- cloud_init
- iprconfig
- java
- libvirt
- nvmetcli
- os_net_config
- perl
- pmem
- powerpc
- scsi
- snap
- sos_extras
- sunrpc
- virsh
- zfs

TurboTurtle:
Probably because I'm a space cadet and looked for the pattern of RH, Deb, Ubuntu - but not RH, Ubuntu, Deb. ;-)

pmoravec:
Further i found also sos/report/plugins/kimchi.py, no idea why I overlooked on previous review /o\ .

I expect explicitly mentioned `CosPlugin` also falls into `IndependentPlugin`, per the logic of plugin's enablement.

Issue #2241, opened 2020-09-21
[snappy] Collect state.json file


slashdd:
I'm reporting this issue on behalf of the Canonical/Ubuntu snap team:

Please include` /var/lib/snapd/state.json` if possible. This will allow easier debugging of snap related issues.

Please note that the content of `/var/lib/snapd/state.json` may contain sensitive information (macaroons) so this data needs to be handled with care, or we need to provide a way to scrub this data.

Origin request/ Launchpad bug:
[0] - https://bugs.launchpad.net/ubuntu/+source/sosreport/+bug/1891296

slashdd:
I'll look at it when time permits.

TurboTurtle:
One thing to keep in mind is do we need to skip this file in `cleaner` like we do with `sos_commands/dpkg` or all the `dnf` logs?

slashdd:
Yep good point, will keep that in mind as well.

slashdd:
@arossbell could you please look at it when time permit ?

arossbell:
Sure thing when time permits it!

PR #2257, opened 2020-10-07
[snap] Collect state.json file, omitting sensitive data.

arossbell:
Closes: #2241
Resolves: #2257

Signed-off-by: Adam R Bell <adam.bell@canonical.com>

---

### DIFF
--- a/sos/report/plugins/snap.py
+++ b/sos/report/plugins/snap.py
@@ -22,6 +22,7 @@ class Snap(Plugin, IndependentPlugin):
     services = ('snapd',)
 
     def setup(self):
+        self.add_copy_spec("/var/lib/snapd/state.json")
         self.add_cmd_output("snap list --all", root_symlink="installed-snaps")
         self.add_cmd_output([
             "snap --version",
@@ -75,5 +76,12 @@ class Snap(Plugin, IndependentPlugin):
                 change_id, change_status = change[0], change[1]
                 if change_status == "Doing" or change_status == "Error":
                     self.add_cmd_output(f"snap tasks {change_id} --abs-time")
+
+    def postproc(self):
+        self.do_file_sub(
+            "/var/lib/snapd/state.json",
+            (r"\"(macaroon|store-macaroon|key-id|session-macaroon|macaroon-key"
+                r"|store-discharges)\":\"?\[?([\"?A-Za-z0-9_=\-,]*)\"?\]?"),
+            r'"\1":"***"')
 
 # vim: set et ts=4 sw=4 :


arossbell:
I'm currently checking on those regex patters to be sure that there's no additional sensitive fields that I missed.

arossbell:
(Corrected Travis CI-error'd code)

arossbell:
Per feedback by @slashdd, I'll change "OMITTED" to "***" shortly

arossbell:
Changed "OMITTED"-like things to "***"-like things.

arossbell:
@mvo5, you originally opened [the LaunchPad ticket](https://bugs.launchpad.net/ubuntu/+source/sosreport/+bug/1891296) about this. Could you check to be sure that we're removing/omitting all fields that we need to remove from the state.json file, please? Thank you!

arossbell:
> `[]` aren't needed here.

Cleaned that up

sos/report/plugins/snap.py
```diff
@@ -28,4 +29,15 @@ def setup(self):
         self.add_service_status("snapd")
         self.add_journal(units="snapd")
 
+    def postproc(self):
+        self.do_file_sub(
+            "/var/lib/snapd/state.json",
+            r"\"(macaroon|store-macaroon|key-id|session-macaroon"
+            + "|macaroon-key)\":\"([A-Za-z0-9_=-]*)\"",
```
TurboTurtle, line 36: I believe this also needs to be explicitly a raw string, given the escapes in it. You can also use normal string concatenation here.

```python
            (r"\"(macaroon|store-macaroon|key-id|session-macaroon"
             r"|macaroon-key)\":\"([A-Za-z0-9_=-]*)\""),
```

arossbell: Changes done; thanks for catching that and the suggestion!

mvo5:
This cleanup looks sensible - you *may* consider to remove all of "config" again except maybe "config.core". It's not entirely clear though, config can hold snap specific config that may contain secrets to login to services etc. OTOH the misfunction of a system may exactly happen because of some configuration. With cofig.core we know the most important configuration (i.e. what experimental features are enabled etc).

sos/report/plugins/snap.py
```diff
@@ -28,4 +29,15 @@ def setup(self):
         self.add_service_status("snapd")
         self.add_journal(units="snapd")
 
+    def postproc(self):
+        self.do_file_sub(
+            "/var/lib/snapd/state.json",
+            (r"\"(macaroon|store-macaroon|key-id|session-macaroon"
```
mvo5, line 35: Should this feature have a test? Also I would feel slightly better if this would use a json parser and not just a regexp?

TurboTurtle: All plugin file postprocessing is done via regex - we don't use file specific parsers. For one, what if the underlying issue that the sos report is being generated for is malformed json? The parser would fail during post processing and we'd likely end up with sensitive data being shipped.

sos/report/plugins/snap.py
```diff
@@ -28,4 +29,15 @@ def setup(self):
...
+                r"|macaroon-key)\":\"([A-Za-z0-9_=-]*)\""),
+            r'"\1":"***"')
+        self.do_file_sub(
+            "/var/lib/snapd/state.json",
+            r"\"(store-discharges)\":\[([\"A-Za-z0-9_\-=,]*)\]",
+            r'"\1":[***]')
```
TurboTurtle, line 41: Why not just pull `store-discharges` into the same `do_file_sub()` above? If we're going to be obfuscating the value anyways, do the brackets really matter?

TurboTurtle:
@arossbell checking in on this one. We close sos-4.3 in ~3 weeks. If you want this to be in 4.3, there appear to be only minor changes needed here.

arif-ali:
@arossbell, do you think you can do the last bits on this, or you happy for someone else to take this on?

sos/report/plugins/snap.py
TurboTurtle: Yeah, since the PR allows edits from maintainers. Basically the flow is to clone their fork, update this branch locally, and then force push it back to *their* fork, which will update this PR.

If you're unsure about it, I can tackle updating this this weekend.

pmoravec (approved):

TurboTurtle (approved):

PR #2245, opened 2020-09-24
[policy] Fix several failure conditions with upload

TurboTurtle:
This patchset fixes several error conditions with `--upload`. First, a too-strict logic check for `RHELPolicy` specifically, then followed by several global fixes for FTP uploads.

The FTP upload fixes include honoring the user-provided upload directory, and then correcting existing error handling while also adding more handlers for specific error conditions.

---
Please place an 'X' inside each '[]' to confirm you adhere to our [Contributor Guidelines](https://github.com/sosreport/sos/wiki/Contribution-Guidelines)

- [x] Is the commit message split over multiple lines and hard-wrapped at 72 characters?
- [x] Is the subject and message clear and concise?
- [x] Does the subject start with **[plugin_name]** if submitting a plugin patch or a **[section_name]** if part of the core sosreport code?
- [x] Does the commit contain a **Signed-off-by: First Lastname <email@example.com>**?
- [ ] If this commit closes an existing issue, is the line `Closes: #ISSUENUMBER` included in an independent line?
- [x] If this commit resolves an existing pull request, is the line `Resolves: #PRNUMBER` included in an independent line?

PR #2265, opened 2020-10-08
[policy] Fix several failure conditions with upload

jjansky1:
The logic for determining if an archive should be uploaded to the
Customer Portal was too strict, ease it to now properly only block on a
missing case number since username and passwords may now be provided via
env vars.

Fixes an issue whereby we ignore a user-provided FTP directory.

Adds a timeout and a timeout handler for FTP connections, rather than
letting the connection attempt continue indefinitely.

Second, adds exception handling for an edge case where the connection to
the FTP server fails, but does not generate an exception from the ftplib
module.

Additionally, correct the type-ing of the error numbers being checked so
that we actually match them.

Caling "sos report --upload --case-id=123 --batch" should fallback
to uploading to FTP server as the upload user is unknown and can't
be prompted in batch mode.

Related: #2276 
Related: #2245
Resolves: #2265

---

### DIFF
--- a/sos/policies/__init__.py
+++ b/sos/policies/__init__.py
@@ -972,7 +972,8 @@ class LinuxPolicy(Policy):
         """Should be overridden by policies to determine if a user needs to
         be provided or not
         """
-        if not self.upload_user and not self._upload_user:
+        if not self.get_upload_password() and (self.get_upload_user() !=
+                                               self._upload_user):
             msg = "Please provide upload user for %s: " % self.get_upload_url()
             self.upload_user = input(_(msg))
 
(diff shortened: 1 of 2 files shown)


pmoravec (approved):
ACK to backport #2245 itself.

Maybe there is an upstream issue in relevant area affecting legacy-3.9 as well: leaving `upload-[user|password]` both empty, `RedHatPolicy` tries to upload to the Red Hat Customer Portal (while it should be clever enough to use anonymous FTP server).

TurboTurtle:
This looks good to me now with the additional backport included to solve the fallback issues.

pmoravec (approved):

TurboTurtle:
Merged into `legacy-3.9` via 8288b5b

PR #2267, opened 2020-10-09
[azure] Include AEM (Azure Enhanced Monitoring for Linux) data (if available)

ofalk:
If available, use AEM to fetch PerfCounters.

Signed-off-by: Oliver Falk <oliver@linux-kernel.at>

---

- [ x Does the subject start with **[plugin_name]** if submitting a plugin patch or a **[section_name]** if part of the core sosreport code?

### DIFF
--- a/sos/report/plugins/azure.py
+++ b/sos/report/plugins/azure.py
@@ -27,7 +27,8 @@ class Azure(Plugin, UbuntuPlugin):
             "/etc/default/kv-kvp-daemon-init",
             "/etc/waagent.conf",
             "/sys/module/hv_netvsc/parameters/ring_size",
-            "/sys/module/hv_storvsc/parameters/storvsc_ringbuffer_size"
+            "/sys/module/hv_storvsc/parameters/storvsc_ringbuffer_size",
+            "/var/lib/AzureEnhancedMonitor"
         ])
 
         # Adds all files under /var/log/azure to the sosreport


ofalk:
Given AEM is *required* for SAP workloads and we see more SAP HANAs in Azure, it makes sense to fetch the PerfCounters if available.
Please let me know what you think!
Of course, I've tested this on some Azure VM!

sos/report/plugins/azure.py
```diff
@@ -66,4 +67,21 @@ def setup(self):
                 'openssl x509 -noout -text -in ' + crt_path
             ])
 
+
+class AzureEnhancedMonitoring(Azure, RedHatPlugin):
```
bmr-cymru, line 71: This isn't a common way to add new functionality to sos - normally only one subclass of a plugin (with the appropriate tagging class) is executed per `sos report` run. Normally this kind of feature would either be merged into the existing (`Azure`) plugin, or if it's relatively orthogonal to the parent product then it may make sense for it to be a completely separate plugin in its own Python source file. I'm not 100% sure without testing what the plugin loader would do with this class - it's not necessarily wrong in Python terms, but it's not a model we promote or test for using the `sos` plugin class hierarchy.

There doesn't seem to be anything shared between the plugin and the parent in terms of state (just the trigger conditions in the class preamble) - if that was the case then that's a definite sign that the feature belongs in a parent class but my gut feeling is that this probably does belong as an addition to either `AzurePlugin` or if it's somehow specific to Red Hat on Azure then it ought to be in the `RedHatAzure` class.

sos/report/plugins/azure.py
```diff
@@ -66,4 +67,21 @@ def setup(self):
...
+    def setup(self):
+        super(AzureEnhancedMonitoring, self).setup()
+        aem_settings = '/var/lib/waagent/' \
+            'Microsoft.OSTCExtensions.AzureEnhancedMonitorForLinux-*/' \
+            'config/0.settings'
+        # Check if AEM (Azure Enhanced Monitoring for Linux) is active
+        aem_glob = glob.glob(aem_settings)
+        if aem_glob:
+            if os.path.isfile(aem_glob[0]):
+                # Add AEM PerfCounters
```
bmr-cymru, line 82: What are we guarding against here? Presumably there are non-file type path entries (directories? sockets?) that cannot be discriminated by pattern and that mean that you don't want to use the automatic glob expansion in `add_copy_spec()` here - that kind of deviation from the normal flow should have a comment explaining why it's being done (big subdirectories, socket files or whatever we're trying to avoid collecting).

In other words what is the condition that could cause `if os.path.isfile(aem_glob[0])` to be `False`?

sos/report/plugins/azure.py
```diff
@@ -66,4 +67,21 @@ def setup(self):
...
+                self.add_copy_spec([
+                    '/var/lib/AzureEnhancedMonitor/PerfCounters'
+                ])
```
bmr-cymru, line 85: No need for a list here - `add_copy_spec()` accepts either a list or a single string value as its first argument.

bmr-cymru (changes requested):

sos/report/plugins/azure.py
ofalk: It's specific to Azure and it's AEM feature. I agree it might make sense to have a separate azure_aem plugin, 
I tested it and it worked :-)

What do you vote for? Separate plugin, given I can imagine to add more around AEM in future?

sos/report/plugins/azure.py
ofalk: AEM may not be available and then this file doesn't exist AFAIK.
So, are you suggesting to just do a `add_copy_spec('/var/lib/AzureEnhancedMonitor/')`?

sos/report/plugins/azure.py
ofalk: OK. I wasn't aware of that and will adapt, once we resolve the discussion around the overall glob-ing.

sos/report/plugins/azure.py
bmr-cymru: Sure - but lots of plugins have optional parts that may or may not be present. This is already gated by the glob into `/var/lib` - if those paths aren't present then none of the extra collection runs and Azure users without AEM will not be troubled by it.

It's not so much what I vote for - the question is why does this need to be a separate plugin at all? Imho it should go into the parent `Azure` plugin class unless there is some reason not to do that (I can't see any from the information in the PR).

sos/report/plugins/azure.py
bmr-cymru: All collection in `sos` is speculative - if paths don't exist then that's fine (and explicitly not an error). This is how we support components that have optional features, or feature variation across versions.

It makes the code more complicated to review and maintain to put these extra checks in the plugin so I would suggest not doing that unless there is a good reason to (like for e.g. huge files or subdirectories that cannot be excluded by the glob pattern).

ofalk: I get your point.
So I'm better off by adding `var/lib/AzureEnhancedMonitor/` to `add_copy_spec` in the Azure class, right?

bmr-cymru: So long as that is safe yes - the only reason to elevate glob handling out of `add_copy_spec()` and into the plugin itself is if there is special logic needed to skip certain paths due to size or other side-effects. If it's just performance counters and they are reasonably sized (and don't scale to 1000s of instances or whatever) then it's fine to just try to scoop them all up in one go.

Btw isn't the path `/var/lib/waagent/Microsoft.OSTCExtensions.AzureEnhancedMonitorForLinux-*`?

ofalk: Got it. Then it will boil down to a single line commit :-P
There are two paths:

    /var/lib/waagent/Microsoft.OSTCExtensions.AzureEnhancedMonitorForLinux-*/config/0.settings

for determining if AEM is enabled on that host

and 

    /var/lib/AzureEnhancedMonitor/

containing the data.

bmr-cymru: Ah, gotcha - thanks - I didn't look closely at the path in the `add_copy_spec()`.

ofalk: Only one question left... The paths in `add_copy_spec` are currently using double quotes, although not necessary. If I add my path now, shall I add it with double quotes, with single quote and if the later, shall I change the other paths also?

bmr-cymru: We've not been all that consistent on this down the years so the general rule of thumb (unless the contributor guide has been updated & I missed it) is to go with the style of the code you are changing - so double quotes in this case.

For new plugins I would tend to prefer double quotes - simply because there are more of them in the tree right now, so when it does eventually get cleaned up there should be fewer single quote cases to fix:

```
$ git grep "add_copy_spec('" | wc -l
35
$ git grep 'add_copy_spec("' | wc -l
152
```

ofalk: Ack. Then I'll change the code accordingly - add the single line with double quotes!
Thanks for your help @bmr-cymru !

ofalk:
Changed as requested, rebased, changed title of PR accordingly.

TurboTurtle (approved):

PR #2268, opened 2020-10-09
Update readme

pipopopo:
---

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,12 +74,17 @@ pull requests.
 
 You can simply run from the git checkout now:
 ```
-$ sudo ./bin/sos report -a
+$ sudo ./bin/sos report 
 ```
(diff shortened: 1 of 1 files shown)


bmr-cymru:
You seem to have two almost identical commits in your branch, one with a Signed-off-by (DCO), and one without. There is also a merge commit which is not allowed in master - please use `git rebase -i` to tidy up the history of your branch so that there is just one commit (with a DCO) that applies cleanly to the current master branch. You'll also want to change the subject of the commit to something like `[docs] update README` to comply with the patch formatting rules.

bmr-cymru (changes requested):
See comment in PR.

bmr-cymru (approved):

pmoravec (approved):

Issue #2203, opened 2020-08-17
[global] have an option to exclude specific files


arif-ali:
There may be situations where a specific file is too big, and want to exclude this file. If it is not excluded the filesystem would be full and the `sos report` will fail to collect. Excluding the plugin is not feasible, as we want to collect the information that the plugin is collecting

maybe have an option called `--exclude-files` as a command line option?

pmoravec:
`sos report` applies a size limit (by default 25MB) for a file / group of files. So until it is e.g. https://github.com/sosreport/sos/blob/master/sos/report/plugins/foreman.py#L79 case where we surpress the limit by `sizelimit=0`, just a limited amount of data should be collected. And that is configurable via `--log-size` option.

Or is the "too big file" one of those with unlimited sizelimit? Then two enhancements are possible:

- option `--exclude-files` as suggested
- enhance `sos_extras` plugin (https://github.com/sosreport/sos/blob/master/sos/report/plugins/sos_extras.py) that currently can collect *new* files and cmdoutputs - it should be easy to enhance it to blocklist some file patterns. BUT doing so would break plugin independence (no plugin shall affect any other one, we rely on this premise) - such an implementation shall work well, just the paradigm will be restricted.

al3jandrosg:
An example "too big" file we found to be collected is `lastlog`, we've seen it causing free disk space issues. I guess that `--exclude-files` would be a better approach in such case as it is a plugin independent file

bmr-cymru:
The `lastlog` file often *appears* huge since it is sparsely allocated on-disk and the maximum record (which determines the file size) depends on the highest numbered UID and GID in use on the system - **BUT** we have never collected the file in `sosreport`, so I am not sure what you are seeing that is causing it to be picked up?

Could you add the version of `sos` that you are running, and the set of plugins (including any 3rd party plugins not sourced from the upstream repository)?

There was a bug in an external component (the `ipsec barf` command provided by the OpenSwan package) that caused long runtimes when running a `grep` command on this file, however we fixed `sos` to not run this by default a number of years ago in sos-3.1:

```
commit 4858ce293a77cc9b973ab7552f3e0b041360b20e
Author: Bryn M. Reeves <bmr@redhat.com>
Date:   Tue Aug 13 15:31:37 2013 +0100

    Make 'ipsec barf' collection non-default in openswan plug-in
    
    The 'ipsec barf' debugging command has a nasty bug where it runs
[...]

PR #2271, opened 2020-10-12
[report] Allow users to specify commands and files to skip

TurboTurtle:
Adds two new options, `--skip-commands` and `--skip-files`, that allow
users to selectively skip specific command or file collection instead of
having to disable whole plugins to skip those collections.

These options are also exposed via `sos collect`, being gated by a
version of 4.1 since that is the next scheduled release where we can
guarantee this functionality will be present.

Closes: #2203
Resolves: #2271

---

### DIFF
--- a/man/en/sos-collect.1
+++ b/man/en/sos-collect.1
@@ -30,6 +30,8 @@ sos collect \- Collect sosreports from multiple (cluster) nodes
     [\-\-password]
     [\-\-password\-per\-node]
     [\-\-preset PRESET]
+    [\-\-skip-commands COMMANDS]
+    [\-\-skip-files FILES]
     [\-s|\-\-sysroot SYSROOT]
     [\-\-ssh\-user SSH_USER]
     [\-\-sos-cmd SOS_CMD]
@@ -260,6 +262,18 @@ defined, or has a version of sos prior to 3.6, this option is ignored for that n
 .TP
 \fB\-p\fR SSH_PORT, \fB\-\-ssh\-port\fR SSH_PORT
 Specify SSH port for all nodes. Use this if SSH runs on any port other than 22.
+.TP
+\fB\-\-skip-commands\fR COMMANDS
+A comma delimited list of commands to skip execution of, but still allowing the
+rest of the plugin that calls the command to run. This will generally need to
+be some form of UNIX shell-style wildcard matching. For example, using a value
+of \fBhostname\fR will skip only that single command, while using \fBhostname*\fR
+will skip all commands with names that begin with the string "hostname".
+.TP
+\fB\-\-skip-files\fR FILES
+A comma delimited list of files or filepath wildcard matches to skip collection
+of. Values may either be exact filepaths or paths using UNIX shell-style wildcards,
+for example \fB/etc/sos/*\fR.
 .TP
 \fB\-\-ssh\-user\fR SSH_USER
 Specify an SSH user for sos collect to connect to nodes with. Default is root.
--- a/man/en/sos-report.1
+++ b/man/en/sos-report.1
@@ -26,6 +26,8 @@ sosreport \- Collect and package diagnostic and support data
           [--log-size]\fR
           [--all-logs]\fR
           [--since YYYYMMDD[HHMMSS]]\fR
+          [--skip-commands commands]\fR
+          [--skip-files files]\fR
           [--allow-system-changes]\fR
           [-z|--compression-type method]\fR
           [--encrypt-key KEY]\fR
@@ -179,6 +181,18 @@ archive is any file not found in /etc, that has either a numeric or a
 compression-type file extension for example ".zip". ".1", ".gz" etc.).
 This also affects \--all-logs. The date string will be padded with zeros
 if HHMMSS is not specified.
+.TP
+.B \--skip-commands COMMANDS
+A comma delimited list of commands to skip execution of, but still allowing the
+rest of the plugin that calls the command to run. This will generally need to
+be some form of UNIX shell-style wildcard matching. For example, using a value
+of \fBhostname\fR will skip only that single command, while using \fBhostname*\fR
+will skip all commands with names that begin with the string "hostname".
+.TP
+.B \--skip-files FILES
+A comma delimited list of files or filepath wildcard matches to skip collection
+of. Values may either be exact filepaths or paths using UNIX shell-style wildcards,
+for example \fB/etc/sos/*\fR.
 .TP
 .B \--allow-system-changes
 Run commands even if they can change the system (e.g. load kernel modules).
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -495,6 +495,8 @@ class Plugin(object):
         self.policy = commons['policy']
         self.devices = commons['devices']
         self.manifest = None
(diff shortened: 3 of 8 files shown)


man/en/sos-report.1
```diff
@@ -180,6 +182,17 @@ compression-type file extension for example ".zip". ".1", ".gz" etc.).
...
+A comma delimited list of commands to skip execution of, but still allowing the
+rest of the plugin that calls the command to run. This will generally need to
+be a regular expression, and unix-style shell regex syntax is supported. For
+example, to skip all 'foobar' commands, use '--skip-commands foobar*'
+.TP
+.B \--skip-files FILES
+A comma delimited list of files or filepath regexes to skip collection of, but
+still allowing the rest of the plugin that tries to collect the files to run.
+These values may be explicit filepaths, or unix-style regexes.
+.TP
```
bmr-cymru, line 195: Looking at the decision code later in the patch we're using `fnmatch.fnmatch` for the pattern matching - this does UNIX style file name pattern matching using glob syntax so I think it's best to avoid calling these regular expressions at all. The doc text for the module says this in the introduction: "This module provides support for Unix shell-style wildcards, which are not the same as regular expressions (which are documented in the re module)."

bmr-cymru (changes requested):
Looks good & potentially provides a mechanism for plugins to define global forbidden files/commands. Only suggestion is to tighten up on the regex/glob terminology in the docs & comments.

man/en/sos-report.1
TurboTurtle: That's a very fair point.

I am not 100% sold on using `fnmatch` here; however it makes perfect sense for `--skip-files` and we don't want to introduce 2 different styles for very similar options (i.e. using `re` for `--skip-commands`). But, if we do go down that route, it should definitely be as clear as possible in the docs.

bmr-cymru: Yeah agree - I think the way you've done it is likely to be the most intuitive for admins; I think it would be confusing to have e.g. `--skip-files /etc/systemd*` but then insist on `--skip-commands system.*` - just a case of having the docs clear. I'm not a huge fan of the term "glob" in docs (even though it's part of POSIX) because it's a bit obscure - I think "shell style pattern matching" is one of the phrases I've used in the past?

TurboTurtle: Updated the phrasing to "UNIX shell-style wildcards" which I think is sufficient.

man/en/sos-collect.1
```diff
@@ -258,8 +260,17 @@ are not used.
 If \fB\-\-preset\fR is specified and a given node either does not have that preset
 defined, or has a version of sos prior to 3.6, this option is ignored for that node.
 .TP
-\fB\-p\fR SSH_PORT, \fB\-\-ssh\-port\fR SSH_PORT
-Specify SSH port for all nodes. Use this if SSH runs on any port other than 22.
```
pmoravec, line 262: Removed lines due to an omission?

man/en/sos-collect.1
```diff
@@ -258,8 +260,17 @@ are not used.
...
+\fB\-\-skip-commands\fR COMMANDS
+A comma delimited list of commands to skip execution of, but still allowing the
+rest of the plugin that calls the command to run. This will generally need to
+be some form of UNIX shell-style wildcard matching. For example, using a value
+of \fBhostname\fR will skip only that single command, while using \fBhostname*\fR
+will skip any any all hostname commands.
```
pmoravec, line 268: "any any" sounds a typo for me, non-native English speaker..?

sos/report/plugins/__init__.py
```diff
@@ -1577,6 +1596,14 @@ def _add_cmd_output(self, **kwargs):
         pred = kwargs.pop('pred') if 'pred' in kwargs else None
         soscmd = SoSCommand(**kwargs)
         self._log_debug("packed command: " + soscmd.__str__())
+        for _skip_cmd in self.skip_cmds:
+            # This probably seems weird to be doing filename matching on the
+            # commands, however we want to remain consistent with our regex
+            # matching with file paths, which sysadmins are almost guaranteed
+            # to assume will use shell-style unix matching
+            if fnmatch.fnmatch(soscmd.cmd, _skip_cmd):
+                self._log_debug("skipping exlcuded command '%s'" % soscmd.cmd)
```
pmoravec, line 1605: s/exlcuded/excluded/    :)

sos/report/plugins/__init__.py
```diff
@@ -1577,6 +1596,14 @@ def _add_cmd_output(self, **kwargs):
         pred = kwargs.pop('pred') if 'pred' in kwargs else None
         soscmd = SoSCommand(**kwargs)
         self._log_debug("packed command: " + soscmd.__str__())
+        for _skip_cmd in self.skip_cmds:
+            # This probably seems weird to be doing filename matching on the
+            # commands, however we want to remain consistent with our regex
+            # matching with file paths, which sysadmins are almost guaranteed
+            # to assume will use shell-style unix matching
+            if fnmatch.fnmatch(soscmd.cmd, _skip_cmd):
```
pmoravec, line 1604: Does it make sense to match against suggest_filename as well (if set)? I am not convinced on either yes or no here..

sos/report/plugins/__init__.py
```diff
@@ -1425,6 +1441,9 @@ def time_filter(path):
                 if self._is_forbidden_path(_file):
                     self._log_debug("skipping forbidden path '%s'" % _file)
                     continue
+                if self._is_skipped_path(_file):
+                    self._log_debug("skipping excluded path '%s'" % _file)
```
pmoravec, line 1445: Worth adding this info to `manifest.json` (or optionally to `sos.[html|txt|json` reports as well)?  (and ditto for skipped cmds)

pmoravec:
Generally ACK to the good idea, just some nit-picks and few questions.

man/en/sos-collect.1
TurboTurtle: Yup, that should be "any and".

sos/report/plugins/__init__.py
TurboTurtle: I don't think so, no. The manifest captures collected files, and executed commands (even if they fail). In this case we aren't executing any commands, and we aren't attempting a file collection.

sos/report/plugins/__init__.py
TurboTurtle: I'd say no, to avoid potential clashes/confusion. But we can certainly keep an eye on this and add it later if need be.

man/en/sos-collect.1
bmr-cymru: How about something like:

```
while using \fBhostname*\fR will skip all commands with names that begin with the string "hostname".
```

To more accurately describe the pattern match for that string? (I know we all get it, but to a user who's less familiar with typical UNIX/Linux conventions the phrase "any and all hostname commands" might imply that it would exclude `oshostname` or `whatsthehostname` when the example pattern will not).

pmoravec: +1 for the "names that begin".

TurboTurtle: Ack.

pmoravec (approved):

Issue #2266, opened 2020-10-08
[pulp] Get dynaconf list into pulp plugin debug


sjha4:
We need to add the below command to pulp debug output.
**sudo -u pulp PULP_SETTINGS='/etc/pulp/settings.py' DJANGO_SETTINGS_MODULE='pulpcore.app.settings' dynaconf list**

ekohl:
Note that this is really Pulp 3. I don't know if there will be a distinction between Pulp 2 and Pulp 3 in sosreport.

For understanding: the name pulpcore was chosen so Pulp 4 can be released on the same codebase. Pulp 3 plugins are named pulp-$myplugin (pulp-rpm, pulp-deb, etc).

Pulp 2 and Pulp 3 can live in parallel on the same system.

The settings can contain a `SECRET_KEY` and also a database password.

pmoravec:
I think it is safe to collect it on pulp-2 (where I think `dynaconf` isnt even installed, hence the output will be empty and wont appear in sosreport, even).

Currently, `pulp` plugin in `sos` is enabled by presence of either `pulp-server` or `pulp-katello` package (https://github.com/sosreport/sos/blob/master/sos/report/plugins/pulp.py#L21). I guess pulp 3 does contain either such package, thus `pulp` plugin will be enabled on a `pulp 3` server by default?

Re obfuscation of sensitive data: could you please provide example output including a `SECRET_KEY` and also a database password, so we can build a r.e. to scrub it? Just the two pieces of data to obfuscate?

ekohl:
> I guess pulp 3 does contain either such package, thus `pulp` plugin will be enabled on a `pulp 3` server by default?

I think `python3-pulpcore` is currently the closest. I know their Ansible based installer can also install using pip rather than RPMs but I don't touch that. From a Foreman perspective we always deal with packages.

pmoravec:
OK, that will need to be added to `packages` list.

Reviewing code of `pulp` plugin, a lot of stuff is `pulp-2`-only related, I think - I will need to review this. But still it makes sense to keep one plugin for `pulp-2` and `pulp-3`.

Pls. provide the examples of data to obfuscate, such that I can prepare a PR for this.

ekohl:
@sjha4 could you take a look at the data and what to obfuscate?

sjha4:
@ekohl , @pmoravec 
So the secret key and Database password live in the /etc/pulp/settings.py file which looks like this:
[...]

PR #2277, opened 2020-10-19
[pulp] collect dynaconf list in pulp-3

pmoravec:
- enable the plugin for pulp-3
- collect dynaconf list output and scrub passwords there

Resoves: #2277
Closes: #2266

---

### DIFF
--- a/sos/report/plugins/pulp.py
+++ b/sos/report/plugins/pulp.py
@@ -18,7 +18,8 @@ class Pulp(Plugin, RedHatPlugin):
     short_desc = 'Pulp platform'
 
     plugin_name = "pulp"
-    packages = ("pulp-server", "pulp-katello")
+    packages = ("pulp-server", "pulp-katello", "python3-pulpcore")
+    files = ("/etc/pulp/settings.py")
     option_list = [
         ('tasks', 'number of tasks to collect from DB queries', 'fast', 200)
     ]
@@ -63,6 +64,7 @@ class Pulp(Plugin, RedHatPlugin):
 
         self.add_copy_spec([
             "/etc/pulp/*.conf",
+            "/etc/pulp/settings.py",
             "/etc/pulp/server/plugins.conf.d/",
             "/etc/default/pulp*",
             "/var/log/httpd/pulp-http.log*",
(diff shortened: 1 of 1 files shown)


sos/report/plugins/pulp.py
```diff
@@ -18,7 +18,7 @@ class Pulp(Plugin, RedHatPlugin):
     short_desc = 'Pulp platform'
 
     plugin_name = "pulp"
-    packages = ("pulp-server", "pulp-katello")
+    packages = ("pulp-server", "pulp-katello", "python3-pulpcore")
```
ekohl, line 21: I don't know sos that well. Does this mean that if any of these packages is installed, it runs?

ekohl (commented):
Pulp 2 and Pulp 3 are really different and are nothing alike. They only share a name and a purpose so it feels very odd to me to combine them. For Pulp 3 you would need to check PostgreSQL for example so in a pure Pulp 3 deployment, all the mongo config and output will fail. That can be very confusing for anyone analyzing the logs. We will also have installs where they are both present.

sos/report/plugins/pulp.py
pmoravec: Yes. Presence of either package will trigger this plugin automatically / by default.

sjha4:
@pmoravec: To add context, the pulp3 details we collect in foreman-debug currently are:
```
add_files /etc/pulp/settings.py
add_cmd "systemctl list-units -t service --full  | grep pulpcore"
add_cmd "systemctl cat pulpcore*"
add_cmd "sudo -u pulp PULP_SETTINGS='/etc/pulp/settings.py' DJANGO_SETTINGS_MODULE='pulpcore.app.settings' dynaconf list"
```

sos/report/plugins/pulp.py
```diff
@@ -128,6 +130,11 @@ def setup(self):
             "qpid-stat -%s --ssl-certificate=%s -b amqps://localhost:5671" %
             (opt, self.messaging_cert_file) for opt in "quc"
         ])
+        self.add_cmd_output(
+            "sudo -u pulp PULP_SETTINGS='/etc/pulp/settings.py'"
+            "DJANGO_SETTINGS_MODULE='pulpcore.app.settings' dynaconf list",
```
TurboTurtle, line 135: Does this not need a space between the env vars? This will concatenate as `PULP_SETTINGS='/etc/pulp/settings.py'DJANGO_SETTINGS_MODULE='pulpcore.app.settings'` currently.

pmoravec: Good eyes! Thanks, fixed now.

TurboTurtle (approved):

TurboTurtle:
Merged into `master` via cc8cc68.

PR #2283, opened 2020-10-27
[kdump] Collect the list of files compressed by squashfs

s-morishima:
Initramfs for kdump has "squash/root.img" compressed by squashfs.
It includes files needed by kdump kernel to boot. The list of them
is collected.

Signed-off-by: MORISHIMA shigeki <s.morishima@fujitsu.com>

---

### DIFF
--- a/sos/report/plugins/kdump.py
+++ b/sos/report/plugins/kdump.py
@@ -8,6 +8,8 @@
 
 import os
 import platform
+import tempfile
+import subprocess
 from sos.report.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
 
 
@@ -67,6 +69,11 @@ class RedHatKDump(KDump, RedHatPlugin):
                         + "kdump.img"
         if os.path.exists(initramfs_img):
             self.add_cmd_output("lsinitrd %s" % initramfs_img)
+            with tempfile.NamedTemporaryFile(delete=False) as tempfp:
+                p = subprocess.Popen(['lsinitrd', '-f', 'squash/root.img', initramfs_img], stdout=tempfp)
+                out, err = p.communicate()
+                if not err:
+                    self.add_cmd_output("unsquashfs -lls %s" % tempfp.name, suggest_filename="unsquashfs_-lls_squash_root.img")
 
         self.add_copy_spec([
             "/etc/kdump.conf",


s-morishima:
Could you please provide me with your suggestions to improve
this patch?

Unfortunately, this patch has one problem. A temporary file
remains existing after collecting this diagnostic information
due to "delete=False" specified in NamedTemporaryFile.
At the moment, collecting this diagnostic information needs
"delete=False" because the temporary file is deleted
before actually collecting this diagnostic information
if "delete=False" isn't specified. I don't think that
this problem can be solved only on this plugin.

sos/report/plugins/kdump.py
```diff
@@ -8,6 +8,8 @@
 
 import os
 import platform
+import tempfile
+import subprocess
```
bmr-cymru, line 12: Nack: do not import subprocess directly in plugin files. Use the APIs provided for this purpose by the `Plugin` class (that set the correct `Popen()` arguments and that handle all the necessary path conversions and logging).

sos/report/plugins/kdump.py
```diff
@@ -67,6 +69,11 @@ def setup(self):
                         + "kdump.img"
         if os.path.exists(initramfs_img):
             self.add_cmd_output("lsinitrd %s" % initramfs_img)
+            with tempfile.NamedTemporaryFile(delete=False) as tempfp:
+                p = subprocess.Popen(['lsinitrd', '-f', 'squash/root.img', initramfs_img], stdout=tempfp)
```
bmr-cymru, line 73: You can't leave temporary files lying around after an sos run. This either has to be done using a pipe between the `lsinitrd -f` and `unsquashfs -ll` commands, or by using temporary paths that are contained in the `sos` working directory (and so are cleaned up during exit handling).

bmr-cymru (changes requested):
I understand what you're trying to do here but the implementation needs a re-think to fit in with `sos` plugin conventions.

bmr-cymru:
> Unfortunately, this patch has one problem. A temporary file
> remains existing after collecting this diagnostic information
> due to "delete=False" specified in NamedTemporaryFile.
> At the moment, collecting this diagnostic information needs
> "delete=False" because the temporary file is deleted
> before actually collecting this diagnostic information
> if "delete=False" isn't specified. I don't think that
> this problem can be solved only on this plugin.

Unfortunately `unsquashfs` does not seem to support reading the compressed file system data from standard input: this would be the ideal way for this to work (it might be possible with `squashfs-tools-ng`, but that is not available in many distributions).

Without this the best option is to place the tmpfile.NamedTemporaryFile into the sos working directory by passing the `dir=` keyword argument. This can be done using existing sos and plugin interfaces but it's fine to suggest improvements to the APIs if they make this case easier to handle.

s-morishima:
Thank you for your suggestion.

> This can be done using existing sos and plugin interfaces
> but it's fine to suggest improvements to the APIs if they make this case easier to handle.

I'll consider both of them. First, I'll consider to add new APIs to somewhere (e.g. Plugin class)
to handle this problem more properly.

> Nack: do not import subprocess directly in plugin files. Use the APIs provided for this purpose 
> by the Plugin class (that set the correct Popen() arguments and that handle all the necessary 
> path conversions and logging).

I see. I'll address this comment as well by using the APIs of the Plugin class instead.

s-morishima:
I apologize for my late response.

Recently, I found that lsinitrd command became able to list the squash content
on Fedora33. lsinitrd has already been fixed on dracut upstream, and 
my purpose also has already been achieved. 
Therefore, I will close this pull request.

Thank you for your cooperation.

PR #2284, opened 2020-10-28
[fibrechannel] Add remote_ports for logging

darina-becker:
Add collecting information about remote ports

Resolves: #2284

Signed-off-by: Daria Bukharina d.bukharina@yadro.com

---

### DIFF
--- a/sos/report/plugins/fibrechannel.py
+++ b/sos/report/plugins/fibrechannel.py
@@ -17,7 +17,7 @@ class Fibrechannel(Plugin, RedHatPlugin):
 
     plugin_name = 'fibrechannel'
     profiles = ('hardware', 'storage', 'system')
-    files = ('/sys/class/fc_host')
+    files = ('/sys/class/fc_host', '/sys/class/fc_remote_ports')
 
     def setup(self):
         self.add_blockdev_cmd("udevadm info -a %(dev)s", devices='fibre')


sos/report/plugins/fibrechannel.py
```diff
@@ -17,7 +17,7 @@ class Fibrechannel(Plugin, RedHatPlugin):
 
     plugin_name = 'fibrechannel'
     profiles = ('hardware', 'storage', 'system')
-    files = ('/sys/class/fc_host')
+    files = ('/sys/class/fc_host', '/sys/class/fc_remote_ports')
```
TurboTurtle, line 20: The `files` tuple only affects plugin enablement, not what gets captured.

This plugin currently only captures `udevadm info` output for each fibre device identified during sos initialization. In particular that is handled here https://github.com/sosreport/sos/blob/59c5c461e36103d453cd9ef9b6cb4624a8171fbe/sos/report/__init__.py#L323

We do currently recurse through `/sys/class/fc_remote_ports` as part of our fibre device discovery.

Is there a case where a system using fibrechannel could potentially have the `fc_remote_ports` directory but _not_ the `fc_host` directory? Or is there information under this directory that we should be capturing separately from `udevadm`?

TurboTurtle: On second look, this does also fix a bug in the current plugin - `files` isn't a tuple currently, which is problematic.

darina-becker: @TurboTurtle Oh, sorry, you're right. I just ran it on the target machine before changes where remote ports are not avaliable and on the initiator after my additions that's why it looked working version! That was my fault sorry for taking your time.

TurboTurtle: Not at all :)

Like I said, this actually indirectly fixes a bug in the plugin right now. It certainly doesn't hurt anything by adding that path as an enablement check. I'm just not personally very familiar with fibrechannel so I wasn't/am not sure if it is possible to have a system with the `fc_remote_ports` path but not the `fc_host` path.

darina-becker: I asked my colleagues - The situation we have system with the `fc_remote_ports` path but not the `fc_host` path is unlikely except there is any critical bag in a kernel or drivers.

Issue #2289, opened 2020-11-01
presets ignore verbosity level


pmoravec:
sos report --add-preset sostestpreset -vvv -o host --batch --build

should add a preset that also enables debug logs. But:

    sos report --preset sostestpreset

prints just default logs - both to the terminal and to logfile.

Curiously, generated logfile shows:

    2020-11-01 13:59:47,913 INFO: [sos.report:setup] effective options now: --batch --build --log-size 100 --only-plugins host --plugopts apache.log=on --preset sostestpreset -vvv

that is the same like log from

    sos report --preset sostestpreset -vvv

which *does* print the debug logs, both to the logfile and to the terminal output..

pmoravec:
The reason is here: https://github.com/sosreport/sos/blob/master/sos/component.py#L214 just returns cmdline options but does not apply presets like:

https://github.com/sosreport/sos/blob/master/sos/report/__init__.py#L139-L151

Therefore, https://github.com/sosreport/sos/blob/master/sos/component.py#L277 has default verbosity, not updated from the preset.

Worth to put `__init__.py#L139-L151` code into a component method that is executed from there and also newly from `component.py#L214` ?

rhdojun:
Hi Pavel,

I had a look at this today.

To me it looks like verbosity is set correctly and the reason must be something else.

[Report is a Component](https://github.com/sosreport/sos/blob/master/sos/report/__init__.py#L69). So [first it initializes the Components `__init__()`](https://github.com/sosreport/sos/blob/master/sos/report/__init__.py#L123) then continues with its own. The reason it doesn't load the `preset` in the Component's `__init__()` is because [`preset` only exists in the context of a Report](https://github.com/sosreport/sos/blob/master/sos/report/__init__.py#L106). It [does not exist in the context of a vanilla Component](https://github.com/sosreport/sos/blob/master/sos/component.py#L54-L66).

I verified that verbosity is set to three correctly via a [pdb](https://docs.python.org/3/library/pdb.html) interactive debugging session on today's master:

```
> /path/to/sosrepo/sos/report/__init__.py(1344)execute()
-> self.prework()
(Pdb) self.opts.verbosity
3
(Pdb) next
 Setting up archive ...
> /path/to/sosrepo/sos/report/__init__.py(1345)execute()
-> self.setup()
(Pdb) self.opts.verbosity
3
(Pdb) next
 Setting up plugins ...
> /path/to/sosrepo/sos/report/__init__.py(1346)execute()
[...]

PR #2294, opened 2020-11-05
Increasing Usage+Readability of ThreadPoolExecutor

rhdojun:
While checking out another issue (#2289) I stumbled over the ThreadPoolExecutor that is used in the Report to execute plugins asynchronously. I knew about Futures from other languages and I know that there was an asyncio trend a few years ago, but I didn't know that Python decided to implement something like this into the core. Loving the idea, so I [checked out the 3.6 docs](https://docs.python.org/3.6/library/concurrent.futures.html).

There I found some differences to how the framework is used in the sos codebase.

 * I added the ContextManager wrapping through a `with` statement. The advantage of that version is that the cleanup code will be executed even if an exception occurs during execution.
 * I removed it as attribute of the Report object. I only saw it being used in these few lines, so it's not necessary to have this variable to exist outside the context where it was used.
 * I give it the pluglist as expanded list.

The last point shouldn't be underestimated and might result in code breaking for users. In the old version the whole `pluglist` was treated as one item that was given to one thread of `_collect_plugin`. In the new version *each item* of `pluglist` is handed separately to its own thread of `_collect_plugin`. I think this was the intention of using ThreadPoolExecutor in the first place. But if the items are not split up by expanding the list, actually only one thread would run and not "up to max_workers" in parallel.

Example:

```
>>> def asdf(a, *b):
...   print(str(a))
...   print(str(b))
... 
>>> asdf(1,[2,3,4,5])
1
([2, 3, 4, 5],)    # <----- the whole list is given as one item in the tuple
>>> asdf(1, *[2,3,4,5])
1
(2, 3, 4, 5)      # <-----  each item is treated separately in the tuple so map() can apply "1" to each item individually
```

I really expect that there might be some non-obvious bugs coming out of this change since asynchronous processing is super hard to get right. But the test suite runs through successfully:

``` 
(venv) 11:13:54:sosreport🍔  nosetests
.................................................................................................................
----------------------------------------------------------------------
Ran 113 tests in 4.676s

OK
```

What do you think?

---

### DIFF
--- a/sos/report/__init__.py
+++ b/sos/report/__init__.py
@@ -910,11 +910,13 @@ class SoSReport(SoSComponent):
             plugruncount += 1
             self.pluglist.append((plugruncount, i[0]))
         try:
-            self.plugpool = ThreadPoolExecutor(self.opts.threads)
-            # Pass the plugpool its own private copy of self.pluglist
-            results = self.plugpool.map(self._collect_plugin,
-                                        list(self.pluglist))
-            self.plugpool.shutdown(wait=True)
+            results = []
+            with ThreadPoolExecutor(
+                max_workers=self.opts.threads) as executor:
+                results = executor.map(
+                    self._collect_plugin,
+                    *list(self.pluglist)
+                )
             for res in results:
                 if not res:
                     self.soslog.debug("Unexpected plugin task result: %s" %


pmoravec:
I would defer the last review to @TurboTurtle as from my limted knowledge of this topic, I dont see particular situations when the patch improves some particular behaviour (this can be surely my limited knowledge, hence deferring the review to Jake). My comments:

- patch makes the report generation bit slower (few percents of time)
- https://github.com/sosreport/sos/pull/2294/files#diff-2034dd4a58a26ac2cf5bb7873e3ff743986cef40100ae8135bcde35660168522R918 breaks execution:

    python3 bin/sos report -o host --batch

raises exception https://travis-ci.org/github/sosreport/sos/jobs/741625148#L548-L567 .

Removing the asterisk, the patch is sane.

TurboTurtle:
@rhdojun - just cycling around here, is there an intent on updating this PR? As discussed on IRC and as @pmoravec mentions, dropping the `*` makes the patch sane, and I agree the `with` mapping makes it more readable.

If you'd like to continue with this PR, let's make that change and rebase on current master to pick up the CI changes that we've recently done (which should also make looking at the speed aspect Pavel brings up easier as well).

PR #2295, opened 2020-11-05
[component] execute() as Abstract class

rhdojun:
Just a little clean up point I stumbled upon while investigating what the `execute()` method does.

---


TurboTurtle (commented):
Ack, though I'd suggest a slight wording change, e.g. `sos components must define their own execute() method`. Just a tiny bit more specific.

Good catch on this one!

sos/component.py
```diff
@@ -123,6 +123,9 @@ def __init__(self, parser, parsed_args, cmdline_args):
             self.manifest.add_field('policy', self.policy.distro)
             self.manifest.add_section('components')
 
+    def execute(self):
+        raise NotImplementedError("subclasses should define their own execute()")
```
TurboTurtle, line 127: Ah, the Travis test is failing due to this being 2 characters too long. In addition to my re-wording suggestion, we'll also want to wrap this under 80 characters.

bmr-cymru: Do we even want a custom string for this exception? I'm of the opinion that `NotImplementedError` is in the does-what-it-says-on-the-tin class. We do a tonne of this in the Boom project, e.g. `boom.osprofile` defines an ABC `BoomProfile` that is the parent class of `OsProfile` and `HostProfile` - many of the methods defined by the `BoomProfile` interface are stubbed out with `NotImplementedError`, e.g.:

```python
class BoomProfile(object):
    """Class ``BoomProfile`` is the abstract base class for Boom template
        profiles. The ``BoomProfile`` class cannot be instantiated by
        itself but serves as the base class for both ``OsProfile`` and
        ``HostProfile`` instances.
    """
    # ...

    def __str__(self):
        """Format this profile as a human readable string.

            This method must be implemented by concrete profile classes.

            :returns: A human readable string representation of this
                      ``BoomProfile``.

            :rtype: string
        """
        raise NotImplementedError

    def __repr__(self):
        """Format this ``BoomProfile`` as a machine readable string.

            This method must be implemented by concrete profile classes.

            :returns: a string representation of this ``BoomProfile``.
            :rtype: ``string``
        """
        raise NotImplementedError

    def _from_data(self, profile_data, dirty=True):
        """
        This method must be implemented by concrete profile classes.
        """
        raise NotImplementedError

    def __len__(self):
        """Return the length (key count) of this profile.

            :returns: the profile length as an integer.
            :rtype: ``int``
        """
        return len(self._profile_data)
```

See [`boom.osprofile.BoomProfile`](https://github.com/snapshotmanager/boom/blob/main/boom/osprofile.py#L471) for more examples.

When thrown they look like this:

```python
>>> from boom.osprofile import *
>>> bp = BoomProfile(profile_keys=["BOOM_OS_NAME"],required_keys=["BOOM_OS_NAME"], identity_key="BOOM_OS_NAME")
>>> str(bp)
  File "<stdin>", line 1, in <module>
  File "/home/breeves/src/git/boom/boom/osprofile.py", line 501, in __str__
    raise NotImplementedError
NotImplementedError
```

(Side note: I never found a good way to prevent the abstract class from being instantiated as shown in the example. Playing around with `__new__` and `abc.ABCMeta` seems more trouble than it's worth but I'm open to ideas on more Pythonic approaches to this area. Side side note: I should change "cannot" to "should not" in the docstring).

rhdojun: Alright, thanks for the feedback, followed the Boom example and left out the comment at all.

pmoravec (approved):

TurboTurtle (approved):

PR #2298, opened 2020-11-09
[openstack_ceilometer] Add backend_url to protected keys

jcastill:
This patch adds the password and url in backend_url to the
list of protected keys, so it gets properly masked.

Related: RHBZ#1892692
Resolves: #2298 

---

### DIFF
--- a/sos/report/plugins/openstack_ceilometer.py
+++ b/sos/report/plugins/openstack_ceilometer.py
@@ -51,10 +51,9 @@ class OpenStackCeilometer(Plugin):
             "admin_password", "connection_password", "host_password",
             "memcache_secret_key", "os_password", "password", "qpid_password",
             "rabbit_password", "readonly_user_password", "secret_key",
-            "ssl_key_password", "telemetry_secret", "metering_secret",
-            "transport_url"
+            "ssl_key_password", "telemetry_secret", "metering_secret"
         ]
-        connection_keys = ["connection"]
+        connection_keys = ["connection", "backend_url", "transport_url"]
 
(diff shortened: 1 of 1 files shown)


pmoravec (commented):
For both `transpert_url` and `backend_url`, I think there is no need to obfuscate hostname and URI. If am I right, isn't it worth moving them from `protect_keys` to `connection_keys`?

jcastill:
I was in doubt between these two, but what made me decide for protected_keys was that transport_url was added there. That said, the hostname/IP can be an internal customer's IP, so I thought having it masked as well was a good idea. But I can change backend_url (and transport_url as well?) to connection_keys if you prefer

TurboTurtle:
I think we should rely on `--clean` to obfuscate the hostnames here if it is desired by the user.

jcastill:
ok, I'll move backend_url to connection_keys. What about transport_url?

pmoravec (approved):
IMHO it would make sense to move `transport_url` to `connection_keys` as well, but a review from OSP is needed (if the format of the obfuscated data match).

ACK to this change itself.

TurboTurtle:
@jcastill this looks ready to merge once `transport_url` is added to `connection_keys`.

pmoravec (approved):

pmoravec:
The packit failure seems an infra problem not on our side, we can ignore it.

TurboTurtle (approved):
Yeah, I'm disabling the packit service again until we get a better way to auth builds for PRs from contributors without write access to the repo.

This LGTM as well.

PR #2300, opened 2020-11-09
[openstack_tripleo] Add new path for log collection

cloudnull:
This change essentially adds /var/lib/tripleo to the collection path.

All of the collection targets have been moved into a class object
which is easily shared across methods. This will allow us to easily
collect more in the future and ensure all of our collected objects
are sanitized.

Signed-off-by: Kevin Carter <kecarter@redhat.com>

---

### DIFF
--- a/sos/report/plugins/openstack_tripleo.py
+++ b/sos/report/plugins/openstack_tripleo.py
@@ -23,12 +23,14 @@ class OpenStackTripleO(Plugin, IndependentPlugin):
     def setup(self):
         # Notes: recursion is max 2 for container-puppet and tripleo-config
         # Those directories are present on all OpenStack nodes
-        self.add_copy_spec([
+        self.tripleo_log_paths = [
             '/var/log/paunch.log',
             '/var/lib/container-puppet/',
             '/var/lib/tripleo-config/',
+            '/var/lib/tripleo/',
             '/etc/puppet/hieradata/'
-        ])
+        ]
+        self.add_copy_spec(self.tripleo_log_paths)
 
     def postproc(self):
         # Ensures we do not leak passwords from the tripleo-config and
@@ -38,10 +40,7 @@ class OpenStackTripleO(Plugin, IndependentPlugin):
                   r'([":\s]+)(.*[^"])([",]+)'
         rgxp = re.compile(secrets, re.IGNORECASE)
 
-        self.do_path_regex_sub('/var/lib/tripleo-config/',
-                               rgxp, r'\1\3*********\5')
-        self.do_path_regex_sub('/etc/puppet/hieradata/',
-                               rgxp, r'\1\3*********\5')
-
+        for path in self.tripleo_log_paths:
+            self.do_path_regex_sub(path, rgxp, r'\1\3*********\5')
 
 # vim: set et ts=4 sw=4 :


weshayutin:
+1

sos/report/plugins/openstack_tripleo.py
```diff
@@ -19,16 +19,18 @@ class OpenStackTripleO(Plugin, IndependentPlugin):
     plugin_name = 'openstack_tripleo'
     profiles = ('openstack', 'openstack_controller', 'openstack_compute')
     packages = ('openstack-selinux',)
+    tripleo_log_paths = [
+        '/var/log/paunch.log',
+        '/var/lib/container-puppet/',
+        '/var/lib/tripleo-config/',
+        '/var/lib/tripleo/',
+        '/etc/puppet/hieradata/'
+    ]
```
TurboTurtle, line 28: Pulling the paths into a list like this is fine, however stylistically I would prefer `self.tripleo_log_paths` be set inside `setup()` - a handful of plugins do this today and we should remain consistent.

sos/report/plugins/openstack_tripleo.py
```diff
@@ -38,10 +40,7 @@ def postproc(self):
                   r'([":\s]+)(.*[^"])([",]+)'
         rgxp = re.compile(secrets, re.IGNORECASE)
 
-        self.do_path_regex_sub('/var/lib/tripleo-config/',
-                               rgxp, r'\1\3*********\5')
-        self.do_path_regex_sub('/etc/puppet/hieradata/',
-                               rgxp, r'\1\3*********\5')
-
+        for path in self.tripleo_log_paths:
+            self.do_path_regex_sub(path, rgxp, r'\1\3*********\5')
```
TurboTurtle, line 44: This would significantly expand the reach of the obfuscation. Have you verified that this produces the desired effect on an actual OSP node?

Code-wise this looks fine.

cloudnull: I have verified this is working on OSP16.1. That said if we think the increase in scope is too large I can scale this back.

sos/report/plugins/openstack_tripleo.py
cloudnull: done.

sos/report/plugins/openstack_tripleo.py
TurboTurtle: Scope is fine - just always like to verify we're not accidentally scrubbing unintentional data when we expand postprocessing.

Thanks!

TurboTurtle (commented):
Code looks good. The commit message just needs a couple tweaks to adhere to our [contribution guidelines](https://github.com/sosreport/sos/wiki/Contribution-Guidelines):

1. Please preface the actual commit title with `[openstack_tripleo]`.
2. Please include a line of `Resolves: #2300` after your commit message, above the DCO line.

cloudnull:
> 1. [openstack_tripleo]

done

TurboTurtle (approved):

pmoravec (approved):

cloudnull:
@TurboTurtle was this intended to be closed?

TurboTurtle:
@cloudnull Yes, it was merged in the commit referenced - 7c88d13

We don't do merge commits in sos, we when we go to add our sign-off for merging we end up rebasing which is what changes the commit IDs, so GH can't link them. It's one of the reasons we have the `Resolves: #XYZ` line in commit messages to make sure the PRs get closed properly.

cloudnull:
Ah this all makes sense; thanks for the clarification @TurboTurtle. This was my first commit to SOS, so I appreciate the insight.

PR #2304, opened 2020-11-10
[component] Add log verbosity from presets

rhdojun:
A simple fix suggestion for #2289 . See commit message for more details.

---

### DIFF
--- a/sos/report/__init__.py
+++ b/sos/report/__init__.py
@@ -153,6 +153,12 @@ class SoSReport(SoSComponent):
         self.opts.merge(self.preset.opts)
         # re-apply any cmdline overrides to the preset
         self.opts = self.apply_options_from_cmdline(self.opts)
+        if hasattr(self.preset.opts, 'verbosity') and \
+            self.preset.opts.verbosity > 0:
+            print('\nWARNING: It is not recommended to set verbosity via the '
+                  'preset as it might have\nunforseen consequences for your '
+                  'report logs.\n')
+            self._setup_logging()
 
         self._set_directories()
 


sos/report/__init__.py
```diff
@@ -153,6 +153,7 @@ def __init__(self, parser, args, cmdline):
         self.opts.merge(self.preset.opts)
         # re-apply any cmdline overrides to the preset
         self.opts = self.apply_options_from_cmdline(self.opts)
+        self._setup_logging()
```
TurboTurtle, line 156: I haven't run this yet myself, but my initial thought is that this would create an additional temp file within our temp directory.

Just thinking out loud here - but perhaps what we should instead do here if check the verbosity level after the merging and `apply_options_from_cmdline()` is done, and simply set the level of the logging handlers appropriately (like we do initially in the base component class).

rhdojun: Searching the code base (`grep -Inr Handler |grep -v venv`) shows me there is only a [stdout StreamHandler](https://github.com/sosreport/sos/blob/master/sos/component.py#L278), a [log file Stream Handler for sos](https://github.com/sosreport/sos/blob/master/sos/component.py#L271), as well as a [log file StreamHandler for sos_ui](https://github.com/sosreport/sos/blob/master/sos/component.py#L304).

And all the variable files give me the impression that their file names are cached in the Component as variables and when recreating the Handlers one would just need to reuse these file names and ensure that that all previous writes have been flushed to the files and the files are closed. I'll think about how I can put this in code.

pmoravec: From my testing, no extra file is created and the patch works well except for:

    020-11-11 14:35:10,109 INFO: [sos.report:setup] effective options now: --batch --build --log-size 100 --plugopts apache.log=on --preset sostestpreset -v


(there should be `-vvv`). But that is a minor issue..

Not sure if the patch can have some side-effect I dont see now, though.

pmoravec (approved):
We should wrap the warning text to 80chars like other texts (e.g. https://github.com/sosreport/sos/blob/master/sos/policies/__init__.py#L799-L801), otherwise ACK.

TurboTurtle:
I guess I'm still wondering why we need to call the logging setup again entirely? Does changing the verbosity level (`self.soslog.setLevel()`) based on the preset value not work?

rhdojun:
`self.soslog.setLevel()` is documented with `Sets the threshold for this logger to level.` Also it doesn't work in my tests.

You first have a tree of loggers, starting with the root logger. When you are at the leaf logger it has handlers which for instance write to file or stdout. It would make sense if each logger in the tree would filter out messages that are beyond its set level. The same goes for the Handler which finally writes messages to stdout or files. So you need at least the Handler level, and in a perfect world probably all the levels reset for the new value. And since I can't see how you can get to the child items in the tree via an official interface I think the only clean values is to create the tree from scratch with new values.

While typing of course I also have other ideas. E.g. one could use a very simple single-logger-stdout-handler setup until all config values are collected and then only create it once. A little like a boot loader for the logging system. Or maybe one could store the whole tree nodes manually in another structure and then access the desired handler from there.

So there's not a good solution and the choice is basically to find the least bad solution. I chose the one with the least new code, which at least keeps the technical debt at a minimum for the maintainers and allows to still implement another solution without much overhead if bug reports later show that another option would have been better.

TurboTurtle:
Ah, ok - that tracks then.

Then my only comment would be to echo @pmoravec - please wrap the text inline with the indentation level. e.g.

```python
            print('WARNING: It is not recommended to set verbosity via the '
                  'preset as it might have unforseen consequences for your '
                  'report logs.')
```

rhdojun:
Well, check out [what Pavel said before](https://github.com/sosreport/sos/pull/2304#pullrequestreview-530568052). I could do either, but we must decide for one way.

TurboTurtle:
For warning messages, I prefer wrapping with the indention. The style used in the disclaimer text is fine for blocks of text.

rhdojun:
@pmoravec What do you say?

pmoravec:
Direct `print` like @TurboTurtle suggests makes more sense to me - I posted not-100%-ideal example. This is not policy related..

rhdojun:
Alright. Interesting that it works without the `\` at the line endings. Thanks for that. Also added some in-string line breaks so the printed message also stays within 80 chars.

TurboTurtle (approved):

TurboTurtle:
Merged into `master` via 169898b

PR #2308, opened 2020-11-12
Improvements to `ip netns` code

mfoliveira:
This PR:
 - Fixes one issue where the networking plugin runs commands on `Error` namespaces;
 - Gathers per-namespace data in the conntrack plugin too.
 - Deduplicates code to get/filter the list of namespaces in the conntrack/ebpf/networking plugins.

Each commit message contains testing details, and a final `# ./tests/simple.sh` ends in `Everything worked!`.

Comparing the contents of `sos_commands` before/after patches, the only changes are the new per-namespace commands in the `conntrack` plugin:

Before and After:
```
$ sudo ./bin/sos report -o conntrack,ebpf,networking --batch
```
Compare:
```
$ find sosreport-<OLD>/sos_commands/ | cut -d/ -f2- | sort \
    > find-sosreport.old

$ find sosreport-<NEW>/sos_commands/ | cut -d/ -f2- | sort \
    > find-sosreport.new

$ diff find-sosreport.{old,new}
4a5,10
> sos_commands/conntrack/ip_netns_exec_ns1_conntrack_-L_-o_extended
> sos_commands/conntrack/ip_netns_exec_ns1_conntrack_-S
> sos_commands/conntrack/ip_netns_exec_ns2_conntrack_-L_-o_extended
> sos_commands/conntrack/ip_netns_exec_ns2_conntrack_-S
> sos_commands/conntrack/ip_netns_exec_ns3_conntrack_-L_-o_extended
> sos_commands/conntrack/ip_netns_exec_ns3_conntrack_-S
```
And the contents of the `ip_netns_exec` files is identical before/after:

```
$ md5sum sosreport-<OLD>/sos_commands/*/* | grep ip_netns_exec | sed 's, [^/]\+/, ,' > md5sum-sosreport.old
$ md5sum sosreport-<NEW>/sos_commands/*/* | grep ip_netns_exec | sed 's, [^/]\+/, ,' > md5sum-sosreport.new

$ diff md5sum-sosreport.{old,new}
0a1,6
> 70e61b0a6cfc43f1fbaa7884f5296495 sos_commands/conntrack/ip_netns_exec_ns1_conntrack_-L_-o_extended
> 84e5f5c039b6d1f35f836f0a2a7c4cdc sos_commands/conntrack/ip_netns_exec_ns1_conntrack_-S
> 70e61b0a6cfc43f1fbaa7884f5296495 sos_commands/conntrack/ip_netns_exec_ns2_conntrack_-L_-o_extended
> 84e5f5c039b6d1f35f836f0a2a7c4cdc sos_commands/conntrack/ip_netns_exec_ns2_conntrack_-S
> 70e61b0a6cfc43f1fbaa7884f5296495 sos_commands/conntrack/ip_netns_exec_ns3_conntrack_-L_-o_extended
> 84e5f5c039b6d1f35f836f0a2a7c4cdc sos_commands/conntrack/ip_netns_exec_ns3_conntrack_-S
```

---

### DIFF
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -2662,8 +2662,40 @@ class Plugin(object):
                 continue
         return pids
 
-    def get_network_namespaces(self):
-        return self.commons['namespaces']['network']
+    def get_network_namespaces(self, ns_pattern=None, ns_max=0):
+        return self.filter_namespaces(self.commons['namespaces']['network'],
+                                      ns_pattern, ns_max)
+
+    def filter_namespaces(self, ns_list, ns_pattern=None, ns_max=0):
+        """Filter a list of namespaces by regex pattern or max number of
+        namespaces (options originally present in the networking plugin.)
+        """
+        out_ns = []
+
+        # Regex initialization outside of for loop
+        if ns_pattern:
+            pattern = (
+                '(?:%s$)' % '$|'.join(ns_pattern.split()).replace('*', '.*')
+                )
+        for ns in ns_list:
+            # if ns_pattern defined, append only namespaces
+            # matching with pattern
+            if ns_pattern:
+                if bool(re.match(pattern, ns)):
+                    out_ns.append(ns)
+
+            # if ns_max is defined and ns_pattern is not defined
+            # remove from out_ns namespaces with higher index than defined
+            elif ns_max != 0:
+                out_ns.append(ns)
+                if len(out_ns) == ns_max:
+                    self._log_warn("Limiting namespace iteration "
+                                   "to first %s namespaces found"
+                                   % ns_max)
+                    break
+            else:
+                out_ns.append(ns)
+        return out_ns
 
 
 class RedHatPlugin(object):
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -2662,6 +2662,9 @@ class Plugin(object):
                 continue
         return pids
 
+    def get_network_namespaces(self):
+        return self.commons['namespaces']['network']
+
 
 class RedHatPlugin(object):
     """Tagging class for Red Hat's Linux distributions"""
--- a/sos/report/plugins/networking.py
+++ b/sos/report/plugins/networking.py
@@ -231,53 +231,52 @@ class Networking(Plugin):
         # per-namespace.
         self.add_cmd_output("ip netns")
         cmd_prefix = "ip netns exec "
-        if self.get_network_namespaces():
-            out_ns = self.get_network_namespaces(
-                        self.get_option("namespace_pattern"),
-                        self.get_option("namespaces"))
+        for namespace in self.get_network_namespaces(
+                            self.get_option("namespace_pattern"),
+                            self.get_option("namespaces")):
+            ns_cmd_prefix = cmd_prefix + namespace + " "
+            self.add_cmd_output([
+                ns_cmd_prefix + "ip address show",
+                ns_cmd_prefix + "ip route show table all",
+                ns_cmd_prefix + "ip -s -s neigh show",
+                ns_cmd_prefix + "ip rule list",
+                ns_cmd_prefix + "iptables-save",
+                ns_cmd_prefix + "ip6tables-save",
+                ns_cmd_prefix + "netstat %s -neopa" % self.ns_wide,
+                ns_cmd_prefix + "netstat -s",
+                ns_cmd_prefix + "netstat %s -agn" % self.ns_wide,
+            ])
 
-            for namespace in out_ns:
+            ss_cmd = ns_cmd_prefix + "ss -peaonmi"
+            # --allow-system-changes is handled directly in predicate
+            # evaluation, so plugin code does not need to separately
+            # check for it
+            self.add_cmd_output(ss_cmd, pred=ss_pred)
+
+        # Collect ethtool commands only when ethtool_namespaces
+        # is set to true.
+        if self.get_option("ethtool_namespaces"):
+            # Devices that exist in a namespace use less ethtool
+            # parameters. Run this per namespace.
+            for namespace in self.get_network_namespaces(
+                                self.get_option("namespace_pattern"),
+                                self.get_option("namespaces")):
                 ns_cmd_prefix = cmd_prefix + namespace + " "
-                self.add_cmd_output([
-                    ns_cmd_prefix + "ip address show",
-                    ns_cmd_prefix + "ip route show table all",
-                    ns_cmd_prefix + "ip -s -s neigh show",
-                    ns_cmd_prefix + "ip rule list",
-                    ns_cmd_prefix + "iptables-save",
-                    ns_cmd_prefix + "ip6tables-save",
-                    ns_cmd_prefix + "netstat %s -neopa" % self.ns_wide,
-                    ns_cmd_prefix + "netstat -s",
-                    ns_cmd_prefix + "netstat %s -agn" % self.ns_wide,
-                ])
-
-                ss_cmd = ns_cmd_prefix + "ss -peaonmi"
-                # --allow-system-changes is handled directly in predicate
-                # evaluation, so plugin code does not need to separately
-                # check for it
-                self.add_cmd_output(ss_cmd, pred=ss_pred)
-
-            # Collect ethtool commands only when ethtool_namespaces
-            # is set to true.
-            if self.get_option("ethtool_namespaces"):
-                # Devices that exist in a namespace use less ethtool
-                # parameters. Run this per namespace.
-                for namespace in out_ns:
-                    ns_cmd_prefix = cmd_prefix + namespace + " "
-                    netns_netdev_list = self.exec_cmd(
-                        ns_cmd_prefix + "ls -1 /sys/class/net/"
-                    )
-                    for eth in netns_netdev_list['output'].splitlines():
-                        # skip 'bonding_masters' file created when loading the
-                        # bonding module but the file does not correspond to
-                        # a device
-                        if eth == "bonding_masters":
-                            continue
-                        self.add_cmd_output([
-                            ns_cmd_prefix + "ethtool " + eth,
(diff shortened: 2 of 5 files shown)


sos/utilities.py
```diff
@@ -210,6 +210,45 @@ def get_human_readable(size, precision=2):
         size = size/1024.0
     return "%.*f%s" % (precision, size, suffixes[suffixindex])
 
+def get_network_namespaces(namespace_pattern = None, namespaces_max = 0,
+                           log = None):
+    """Used to get a list of network namespaces, optionally filtered
+    by a regex pattern or max number of namespaces (options from the
+    networking plugin.)
+    If you specify max number of namespaces (namespaces_max) specify
+    a log function too, to log if the limit was reached.
```
TurboTurtle, line 219: Nack. Use the logging already setup by sos. This will be available inside the `Plugin` class already.

TurboTurtle (commented):
Some initial comments follow.

I like the aim of `get_network_namespaces()` however I think this should either:
1. Go into `Plugin()` rather than into `utilities.py`
2. Be called during initialization and passed to plugins via `commons` like we do for disks and fibre devices, then have a wrapper in `Plugin` to filter the cached results as needed.

I prefer option 2 personally, but I wouldn't object to option 1.

sos/utilities.py
```diff
@@ -210,6 +210,45 @@ def get_human_readable(size, precision=2):
...
+    """
+    out_ns = []
+
+    ip_netns = sos_get_command_output("ip netns")
+    if ip_netns['status'] == 0:
+        # Regex initialization outside of for loop
+        if namespace_pattern:
+            pattern = '(?:%s$)' % '$|'.join(
+                    namespace_pattern.split()
+                    ).replace('*', '.*')
```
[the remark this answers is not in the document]
mfoliveira: 1. Indeed; did I get it right, here? Sending it this way for now.
```
+            pattern = (
+                '(?:%s$)' % '$|'.join(ns_pattern.split()).replace('*', '.*')
+                )
```

2. Ack; indeed, I just copied it over, trying to be conservative. :)

sos/utilities.py
```diff
@@ -210,6 +210,45 @@ def get_human_readable(size, precision=2):
...
+                if bool(re.match(pattern, line)):
+                    out_ns.append(line.partition(' ')[0])
+
+            # if namespaces is defined and namespace_pattern is not defined
+            # remove from out_ns namespaces with higher index than defined
+            elif namespaces_max != 0:
+                out_ns.append(line.partition(' ')[0])
+                if len(out_ns) == namespaces_max:
+                    log("Limiting namespace iteration " +
+                        "to first %s namespaces found"
```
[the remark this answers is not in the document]
mfoliveira: Ack.

sos/utilities.py
mfoliveira: Face palm. I missed that came from `Plugin`. Fixed it.

mfoliveira:
@TurboTurtle 

Thanks for reviewing!

> Some initial comments follow.
> 
> I like the aim of `get_network_namespaces()` however I think this should either:
> 
>     1. Go into `Plugin()` rather than into `utilities.py`
> 
>     2. Be called during initialization and passed to plugins via `commons` like we do for disks and fibre devices, then have a wrapper in `Plugin` to filter the cached results as needed.
> 
> 
> I prefer option 2 personally, but I wouldn't object to option 1.

Option 2 indeed looks better and more consistent with what's already available.
I implemented that, along with your suggestions above. Pushing the changes.

mfoliveira:
> This pull request **introduces 1 alert** when merging [99de64b](https://github.com/sosreport/sos/commit/99de64b29b117bfe1d92fdd672d7148a729db5e3) into [7314865](https://github.com/sosreport/sos/commit/7314865bab8569e2d6f64e0e30f4118c18bdc2f6) - [view on LGTM.com](https://lgtm.com/projects/g/sosreport/sos/rev/pr-81601b49d2c1ffdddfe380bd06135cc277bf432a)
> 
> **new alerts:**
> 
>     * 1 for Unused import

Removed `from re import match` from `networking.py`, not used anymore.

sos/report/plugins/__init__.py
```diff
@@ -2662,6 +2662,37 @@ def get_process_pids(self, process):
...
+        if ns_pattern:
+            pattern = (
+                '(?:%s$)' % '$|'.join(ns_pattern.split()).replace('*', '.*')
+                )
+        for ns in namespaces:
+            # if ns_pattern defined, append only namespaces
+            # matching with pattern
+            if ns_pattern:
+                if bool(re.match(pattern, ns)):
+                    out_ns.append(ns)
```
pmoravec, line 2681: This means, if *both* `ns_pattern` and `ns_max` is set, the later option is ignored? I am not sure how broad scenario it is to call networking plugin with both limits applied, but still I see it as a legitimate one.

Could this scenario be supported as well?

mfoliveira: Hi Pavel,

Thanks for reviewing.

Good point. I wondered about that as well, as to why the options are mutually exclusive, but their combination does seem  a legitimate use case, as you mentioned. (e.g., all namespaces matching a regex, but at most N, in case the system has way too many.)

I guess it can be supported, but I'd have to defer to more experienced folks/author/maintainers, as I'm fresh new to sosreport code, as you may have seen from previous suggestions. :)

For now/this PR, I focused more on keeping the current behavior (sure enough) and just changing places for current things; but would be happy to take that route / add another commit with your suggestion, if that is the consensus.

cheers,
Mauricio

sos/report/plugins/__init__.py
```diff
@@ -2662,6 +2662,37 @@ def get_process_pids(self, process):
                 continue
         return pids
 
+    def filter_namespaces(self, namespaces, ns_pattern = None, ns_max = 0):
```
TurboTurtle, line 2665: Take a look at the device related commands in plugin for a reference - this should be automatically using the namespaces discovered earlier rather than having the plugins pass that same value in every time.

E.G.

```python
def get_namespaces(self, ns_pattern=None, ns_max=0):
[...]
    for ns in self.commons[namespaces][network]
```

Then in the plugins you'd use `for namespace in self.get_namespaces()` for the whole list, passing only a regex if you want to filter those down.

sos/report/plugins/networking.py
```diff
@@ -230,38 +229,13 @@ def setup(self):
...
-                        break
-                else:
-                    out_ns.append(line.partition(' ')[0])
+        if self.commons['namespaces']['network']:
+            out_ns = self.filter_namespaces(
+                        self.commons['namespaces']['network'],
+                        self.get_option("namespace_pattern"),
+                        self.get_option("namespaces"))
 
             for namespace in out_ns:
```
TurboTurtle, line 240: This would all be condensed down to the following for example with the changes mentioned above

```python
for namespace in self.get_namespaces(self.get_option("namespace_pattern"),
                                     self.get_option("namespaces")):
    [... collections ...]
```

TurboTurtle (changes requested):

sos/report/plugins/__init__.py
pmoravec: Ah right, now I recall there was some discussion in #1916 without a consensus. Let dont change the behaviour and just copy the current one, here.

sos/report/plugins/__init__.py
mfoliveira: I see; I tried to make the function generic in case other non-network, namespace types are implemented/used in the future.

But this approach looks much simpler for users, and it can use `get_network_namespaces()` to particularize for the network namespace type.

I'll resubmit with those changes; thanks!

sos/report/plugins/networking.py
mfoliveira: I was lazy and kept `if` and `out_ns` to avoid the large indentation changes, and due to the other use of `out_ns` for `ethtool_namespaces` option.
I'll update both more properly this time. :)

mfoliveira:
@TurboTurtle 

Just pushed the changes with your suggestions, and updated testing in PR description. Thanks!

TurboTurtle:
Code-wise this looks good to me, and a couple test runs look right.

Just 2 things:

1. One more rebase to pick up the CirrusCI changes, so this PR will then kick off a set of the automated testing
2. There are some PEP8 cleanups needed:

```
sos/report/__init__.py:20:80: E501 line too long (81 > 79 characters)
sos/report/plugins/__init__.py:2665:48: E251 unexpected spaces around keyword / parameter equals
sos/report/plugins/__init__.py:2665:50: E251 unexpected spaces around keyword / parameter equals
sos/report/plugins/__init__.py:2665:63: E251 unexpected spaces around keyword / parameter equals
sos/report/plugins/__init__.py:2665:65: E251 unexpected spaces around keyword / parameter equals
sos/report/plugins/__init__.py:2669:52: E251 unexpected spaces around keyword / parameter equals
sos/report/plugins/__init__.py:2669:54: E251 unexpected spaces around keyword / parameter equals
sos/report/plugins/__init__.py:2669:67: E251 unexpected spaces around keyword / parameter equals
sos/report/plugins/__init__.py:2669:69: E251 unexpected spaces around keyword / parameter equals
```

If Travis was working this would have been caught earlier, but this will now also be caught by Cirrus once point 1 is done.

mfoliveira:
> Code-wise this looks good to me, and a couple test runs look right.

Cool; thanks for promptly reviewing!

> Just 2 things:
> 
>     1. One more rebase to pick up the CirrusCI changes, so this PR will then kick off a set of the automated testing
> 
>     2. There are some PEP8 cleanups needed:

Uh, I had the impression I did rebase; just re-did, and CirrusCI is running now.  And fixed the PEP8 warnings.

TurboTurtle (approved):

pmoravec (approved):

mfoliveira:
@TurboTurtle @pmoravec 

Thanks for speedy reviews, corrections, and suggestions!

pmoravec:
.. and thanks for the fast implementation of our feedback, though it might be seen as punctilious sometimes.

PR #2309, opened 2020-11-15
[postgresql] reorganize postgres from SCL

pmoravec:
- enable `RedHatPostgreSQL` also by `rh-postgresql12` SCL
- collect configs and logs regardless of running service
  - this is required for newer `foreman` / Satellite6 deployments, where PostgreSQL backend is installed from `rh-postgresql12` SCL but run as `postgresql.service`; technically enhancing collection of logs+configs to *all* SCLs does not hurt
  - `pg_dump` will **not** be collected for that scenario, but that does not matter (there are other supported means to get postgres backup, if required)
- collect "du -sh" for SCL directories as well

Resolves: #2309

---

### DIFF
--- a/sos/report/plugins/postgresql.py
+++ b/sos/report/plugins/postgresql.py
@@ -83,47 +83,41 @@ class RedHatPostgreSQL(PostgreSQL, SCLPlugin):
     packages = (
         'postgresql',
         'rh-postgresql95-postgresql-server',
-        'rh-postgresql10-postgresql-server'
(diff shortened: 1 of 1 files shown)


pmoravec:
Cc @jjansky1 and @sandrobonazzola and @TurboTurtle for review / feedback.

TurboTurtle (approved):

sandrobonazzola (approved):

PR #2315, opened 2020-11-18
Align sosreport plugins with Mirantis Cloud Platform installation:

wkonitzer:
* Update Docker plugin;
* Update Salt/SaltMaster plugins;
* Update OpenStack services plugins;
* Add OpenStack Barbican plugin;
* Update Keepalived as IndependentPlugin class

---

### DIFF
--- a/sos/report/plugins/keepalived.py
+++ b/sos/report/plugins/keepalived.py
@@ -8,10 +8,10 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.report.plugins import Plugin, RedHatPlugin
+from sos.report.plugins import Plugin, IndependentPlugin
 
 
-class Keepalived(Plugin, RedHatPlugin):
+class Keepalived(Plugin, IndependentPlugin):
 
     short_desc = 'Keepalived routing server'
 
--- a/sos/report/plugins/openstack_aodh.py
+++ b/sos/report/plugins/openstack_aodh.py
@@ -10,16 +10,15 @@
 # See the LICENSE file in the source distribution for further information.
 
 import os
-from sos.report.plugins import Plugin, RedHatPlugin
+from sos.report.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
 
 
-class OpenStackAodh(Plugin, RedHatPlugin):
+class OpenStackAodh(Plugin):
 
     short_desc = 'OpenStack Alarm service'
     plugin_name = "openstack_aodh"
     profiles = ('openstack', 'openstack_controller')
 
-    packages = ('openstack-selinux',)
     var_puppet_gen = "/var/lib/config-data/puppet-generated/aodh"
 
     def setup(self):
@@ -87,5 +86,20 @@ class OpenStackAodh(Plugin, RedHatPlugin):
             r"\1********",
         )
 
+
+class DebianOpenStackAodh(OpenStackAodh, DebianPlugin, UbuntuPlugin):
+
+    packages = (
+        'aodh-api',
+        'aodh-evaluator',
+        'aodh-notifier',
+        'aodh-listener',
+        'python-aodhclient'
+    )
+
+
+class RedHatOpenStackAodh(OpenStackAodh, RedHatPlugin):
+
+    packages = ('openstack-selinux',)
 
 # vim: set et ts=4 sw=4 :
--- a/sos/report/plugins/openstack_barbican.py
+++ b/sos/report/plugins/openstack_barbican.py
@@ -1,0 +1,54 @@
+# Copyright (C) 2019 Mirantis, Inc., Denis Egorenko <degorenko@mirantis.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, DebianPlugin, UbuntuPlugin
+
+
+class OpenStackBarbican(Plugin, DebianPlugin, UbuntuPlugin):
+    """OpenStack Barbican Secure storage service"""
+    plugin_name = "openstack_barbican"
+    profiles = ('openstack', 'openstack_controller')
+
+    packages = (
+        'barbican-common',
+        'barbican-keystone-listener',
+        'barbican-worker'
+    )
+
+    requires_root = False
+
+    def setup(self):
+        self.add_copy_spec("/etc/barbican/")
+
+        if self.get_option("all_logs"):
+            self.add_copy_spec("/var/log/barbican/*")
+        else:
+            self.add_copy_spec("/var/log/barbican/*.log")
+
+        self.add_forbidden_path("/etc/barbican/*.pem")
+        self.add_forbidden_path("/etc/barbican/alias/*")
+
+    def postproc(self):
+        self.do_file_sub(
+            "/etc/barbican/barbican.conf",
+            r"(password|rabbit_password[\t\ ]*=[\t\ ]*)(.+)",
+            r"\1********"
+        )
+
+        connection_keys = ["transport_url", "sql_connection"]
+
+        self.do_path_regex_sub(
+            "/etc/barbican/barbican.conf",
+            r"((?m)^\s*(%s)\s*=\s*(.*)://(\w*):)(.*)(@(.*))" %
+            "|".join(connection_keys),
+            r"\1*********\6")
+
+
+# vim: set et ts=4 sw=4 :
--- a/sos/report/plugins/salt.py
+++ b/sos/report/plugins/salt.py
@@ -6,17 +6,17 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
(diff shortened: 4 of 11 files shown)


sos/report/plugins/keepalived.py
```diff
@@ -8,10 +8,10 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.report.plugins import Plugin, RedHatPlugin
+from sos.report.plugins import Plugin, IndependentPlugin
 
 
-class Keepalived(Plugin, RedHatPlugin):
+class Keepalived(Plugin, IndependentPlugin):
```
TurboTurtle, line 14: Using `IndependentPlugin` enables this plugin for all distributions. Is the package name the same on all distributions? Along with the file paths being collected?

sos/report/plugins/openstack_aodh.py
```diff
@@ -10,16 +10,15 @@
 # See the LICENSE file in the source distribution for further information.
 
 import os
-from sos.report.plugins import Plugin, RedHatPlugin
+from sos.report.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
 
 
-class OpenStackAodh(Plugin, RedHatPlugin):
+class OpenStackAodh(Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin):
```
TurboTurtle, line 16: If you're going to be adding distro specific subclasses, these tagging classes go in the subclasses, not the base class.

sos/report/plugins/openstack_aodh.py
```diff
@@ -88,4 +87,19 @@ def postproc(self):
         )
 
 
+class DebianAodh(OpenStackAodh, DebianPlugin, UbuntuPlugin):
```
TurboTurtle, line 90: The preferred style is to use the full previous class name, e.g. this would be `DebianOpenStackAodh`

sos/report/plugins/openstack_barbican.py
```diff
@@ -0,0 +1,62 @@
...
+    profiles = ('openstack', 'openstack_controller')
+
+    packages = ('barbican-common', 'barbican-keystone-listener', 'barbican-worker',)
+
+    requires_root = False
+
+    def setup(self):
+        self.add_copy_spec([
+            "/etc/barbican/",
+        ])
```
TurboTurtle, line 26: No need to use a list for single-items. `add_copy_spec()` takes a str as well.

sos/report/plugins/openstack_barbican.py
```diff
@@ -0,0 +1,62 @@
...
+
+        if self.get_option("all_logs"):
+            self.add_copy_spec([
+                "/var/log/barbican/*",
+            ])
+        else:
+            self.add_copy_spec([
+                "/var/log/barbican/*.log",
+            ])
```
TurboTurtle, line 35: Ditto.

sos/report/plugins/salt.py
```diff
@@ -6,17 +6,17 @@
...
-class Salt(Plugin, RedHatPlugin, DebianPlugin):
+class Salt(Plugin, IndependentPlugin):
 
     short_desc = 'Salt'
 
     plugin_name = 'salt'
     profiles = ('sysmgmt',)
 
-    packages = ('salt',)
+    packages = ('salt-minion', 'salt-common',)
```
TurboTurtle, line 19: There is a package named `salt` on RHEL distributions, in addition to `salt-minion` (but there is no `salt-common`). I'd prefer we leave the original in place as well given this.

sos/report/plugins/saltmaster.py
```diff
@@ -6,20 +6,33 @@
...
 
     plugin_name = 'saltmaster'
     profiles = ('sysmgmt',)
 
-    packages = ('salt-master',)
+    packages = ('salt-master', 'salt-api',)
 
     def setup(self):
-        self.add_copy_spec("/var/log/salt/master")
+        all_logs = self.get_option("all_logs")
```
TurboTurtle, line 22: Unnecessary - use `get_option()` in line with your conditionals.

sos/report/plugins/docker.py
```diff
@@ -82,7 +82,7 @@ def setup(self):
         for container in containers:
             self.add_cmd_output("docker inspect %s" % container,
                                 subdir='containers')
-            if self.get_option('logs'):
+            if self.get_option('logs') or self.get_option("all_logs"):
```
TurboTurtle, line 85: I'm unsure about this one. I see the rationale for it, but `all-logs` has traditionally been more related to size considerations than gating command output.

sos/report/plugins/openstack_barbican.py
```diff
@@ -0,0 +1,62 @@
...
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.plugins import Plugin, DebianPlugin, UbuntuPlugin
```
TurboTurtle, line 11: This needs to be `sos.report.plugins`

TurboTurtle (changes requested):
In addition to the comments below, please see the [Contributor Guidelines](https://github.com/sosreport/sos/wiki/Contribution-Guidelines).

We don't allow merge commits in sos, so you'll need to rebase against master and remove those.

Similarly, the addition of `openstack_barbican` should have it's own commit. Please squash any fixup commits as well.

The commit('s) titles should be in the form `[Component] <short description>`, E.G. `[openstack_barbican] Add new plugin`

In at least one of the commits, please include `Resolves: #2315`

sos/report/plugins/keepalived.py
wkonitzer: Yes, it looks to be the same everywhere I've checked.

sos/report/plugins/openstack_aodh.py
wkonitzer: Done.

sos/report/plugins/openstack_aodh.py
wkonitzer: Done.

sos/report/plugins/openstack_barbican.py
wkonitzer: Done.

sos/report/plugins/salt.py
wkonitzer: Done.

sos/report/plugins/docker.py
wkonitzer: I've left this the same, but if it needs changing let me know.

sos/report/plugins/openstack_barbican.py
wkonitzer: Done.

sos/report/plugins/openstack_barbican.py
wkonitzer: Done

sos/report/plugins/saltmaster.py
wkonitzer: Reworked.

sos/report/plugins/docker.py
TurboTurtle, line 85: After some more thought and discussion on this with another maintainer, I don't think this is desirable for most users of sos. The plugin option serves as sufficient enough gating for collecting container logs, which are historically not needed to diagnose host issues.

If a user/consumer of sosreports is looking into containerized applications, then it's reasonable to assume they know to use `-k docker.logs`. Conversely, `--all-logs` is going to be used more often and in more use cases, in which getting (non-size-limited) container logs is likely going to be surprising behavior to users.

TurboTurtle (commented):
Rest of the code looks fine, once the `--all-logs` addition is dropped.

Beyond that, please squash the fixup commit into your first commit. The commit title could also use a tag like `[plugins]` or somesuch.

sos/report/plugins/docker.py
wkonitzer: ok, I've reverted the changes.

TurboTurtle (approved):

sos/report/plugins/openstack_barbican.py
```diff
@@ -0,0 +1,60 @@
...
+    def postproc(self):
+        self.do_file_sub(
+            "/etc/barbican/barbican.conf",
+            r"(password[\t\ ]*=[\t\ ]*)(.+)",
+            r"\1********"
+        )
+
+        self.do_file_sub(
+            "/etc/barbican/barbican.conf",
+            r"(rabbit_password[\t\ ]*=[\t\ ]*)(.+)",
```
pmoravec, line 47: Isn't this regexp already included in the previous one?

If not included, does not it make sense to merge the two REs into one, e.g.:

    `r"([password|rabbit_password][\t\ ]*=[\t\ ]*)(.+)"`

?

pmoravec:
Upon the one comment to scrub passwords, ACK.

sos/report/plugins/openstack_barbican.py
wkonitzer: Hi @pmoravec, so I've merged those into one, verified the regex works and resubmitted.

pmoravec (approved):
Thanks!

PR #2324, opened 2020-11-30
[fibrechannel] Add optional arguments for plugin

darina-becker:
Add a possibility to enable collecting debug logs from vendor specific
paths. And add a posibility to extend collecting debug logs with
optional argument.
Second argument has to get additional paths. The input format of
argument is <path_1>:...:<path_n>. The default value is 'None'.
It is useful when you need to collect additional debug information.

Resolves: #2324

Signed-off-by: Daria Bukharina d.bukharina@yadro.com

---

### DIFF
--- a/sos/report/plugins/fibrechannel.py
+++ b/sos/report/plugins/fibrechannel.py
@@ -18,8 +18,18 @@ class Fibrechannel(Plugin, RedHatPlugin):
     plugin_name = 'fibrechannel'
     profiles = ('hardware', 'storage', 'system')
     files = ('/sys/class/fc_host', '/sys/class/fc_remote_ports')
+    option_list = [
+        ("debug", "enable debug logs", "fast", True)
+    ]
+
+    # vendor specific debug paths
+    debug_paths = [
+        '/sys/kernel/debug/qla2*/'
+    ]
 
     def setup(self):
         self.add_blockdev_cmd("udevadm info -a %(dev)s", devices='fibre')
+        if self.get_option('debug'):
+            self.add_copy_spec(self.debug_paths)
 
 # vim: set et ts=4 sw=4 :


darina-becker:
We want to use it like this:
`sos report -kfibrechannel.add_path=/sys/kernel/debug/qla2xxx`.
This folder contains debug information about QLogic fibre channel adapter (include fc sessions).

```
Target-22-pci2742-ip64:/sys/kernel/debug/qla2xxx # ls -R
.:
qla2xxx_0  qla2xxx_1

./qla2xxx_0:
fce  fw_resource_count	naqp  tgt_counters  tgt_port_database  tgt_sess

./qla2xxx_1:
fce  fw_resource_count	naqp  tgt_counters  tgt_port_database  tgt_sess
```

sos/report/plugins/fibrechannel.py
```diff
@@ -18,8 +18,15 @@ class Fibrechannel(Plugin, RedHatPlugin):
     plugin_name = 'fibrechannel'
     profiles = ('hardware', 'storage', 'system')
     files = ('/sys/class/fc_host', '/sys/class/fc_remote_ports')
+    option_list = [("add_path", "additional path to collect", "fast", None)]
 
     def setup(self):
         self.add_blockdev_cmd("udevadm info -a %(dev)s", devices='fibre')
+        additional_path = self.get_option("add_path")
+        if additional_path is not None:
+            path_list = additional_path.split(',')
```
TurboTurtle, line 27: Delimiters within plugin options have recently taken on the style of using `:` as the delimiting character. This avoids conflicts with `argparse` splitting list arguments earlier on.

sos/report/plugins/fibrechannel.py
```diff
@@ -18,8 +18,15 @@ class Fibrechannel(Plugin, RedHatPlugin):
...
+            if path_list:
+                for path in path_list:
```
TurboTurtle, line 29: We can tighten this up a bit:

```python
for path in self.get_option('add_path').split(':'):
    self.add_copy_spec(path)
```

`add_copy_spec` has the logic to skip non existing paths and the like, so no need to put that into plugins.

TurboTurtle (commented):
I'm leaning towards saying just make this a default action of the plugin, rather than rely on users passing in paths via plugin options, especially if common use cases for this all have well-known paths.

If a path doesn't exist on a system, `add_copy_spec()` handles that already, so we could do something like this:

```python
# vendor specific debug paths
self.debug_paths = [
    '/sys/kernel/debug/qla2*/,
    '/some/other/vendor/path'
]
self.add_copy_spec(self.debug_paths)
```

sos/report/plugins/fibrechannel.py
darina-becker: Thank you a lot! I've fixed as you wrote.

darina-becker: Sorry, I haven't noticed the first comment. I'll fix it a bit later.

sos/report/plugins/fibrechannel.py
```diff
@@ -18,8 +18,20 @@ class Fibrechannel(Plugin, RedHatPlugin):
     plugin_name = 'fibrechannel'
     profiles = ('hardware', 'storage', 'system')
     files = ('/sys/class/fc_host', '/sys/class/fc_remote_ports')
+    option_list = [("debug", "enable debug logs", "fast", True),
+                   ("add_dbg_path", "add debug path to collect", "fast", None)
```
TurboTurtle, line 22: So we still end up with the same end issue here - user defined paths. All it would take is a user typo here to land us in a lot of hot water (imagine a user does `-k fibrechannel.add_dbg_path=/` for example).

What I am saying for this is to not allow user defined paths, and simply maintain a list of well known vendor paths for this kind of information.

darina-becker: Well, I agree with you. I wanted to have an easy way to add paths for collecting specific logs in future without pull requests. You're right it may lead to negative effects if someone uses this mindlessly. So I've removed it.

TurboTurtle (approved):
Ack from a code perspective.

One last sanity check - are there any known "gotchas" under `/sys/kernel/debug/qla2*`? E.G. pipes like `/sys/kernel/debug/tracing/trace_pipe` that we need to skip to avoid hanging?

pmoravec (approved):
ACK, +1 for the Jake's question; can't concatenating a file-like object in the dirs be end-less?

darina-becker:
@TurboTurtle @pmoravec No, there are only regular files in `/sys/kernel/debug/qla2*`.

TurboTurtle:
Great, thanks for the confirmation. We'll get this merged in the next few days.

PR #2326, opened 2020-12-02
[tests] Add test for newly loaded kernel modules

TurboTurtle:
Adds a test to `simple.sh` to check if we loaded any new kernel modules
during a normal exeuction. This test will be somewhat limited in the
fact that this currently only tests the plugins that are loaded by
default on each tested distribution.

Also included is a minor typo fix for reporting where IP addresses were
not obfuscated, if any. This typo did not affect the actual test
however.

Resolves: #2326

---

### DIFF
--- a/tests/simple.sh
+++ b/tests/simple.sh
@@ -1,3 +1,4 @@
+#!/bin/bash
 # This file is part of the sos project: https://github.com/sosreport/sos
 #
 # This copyrighted material is made available to anyone wishing to use,
@@ -5,7 +6,6 @@
 # version 2 of the GNU General Public License.
 #
 # See the LICENSE file in the source distribution for further information.
-#/bin/bash
 # A quick port of the travis tests to bash, requires root
 # TODO
 # * look into using a framework..
@@ -101,6 +101,8 @@ add_failure () {
 # Test a no frills run with verbosity and make sure the expected items exist
 test_normal_report () {
     cmd="-vvv"
+    # get a list of initial kmods loaded
+    kmods=( $(lsmod | cut -f1 -d ' ' | sort) )
     run_expecting_success "$cmd" extract
     if [ $? -eq 0 ]; then
         if [ ! -f /var/tmp/sosreport_test/sos_reports/sos.html ]; then
@@ -114,6 +116,12 @@ test_normal_report () {
         fi
         if [ ! "$(grep "DEBUG" /var/tmp/sosreport_test/sos_logs/sos.log)" ]; then
             add_failure "did not find debug logging when using -vvv"
+        fi
+        # new list, see if we added any
+        new_kmods=( $(lsmod | cut -f1 -d ' ' | sort) )
+        if [ "$(printf '%s\n' "${kmods[@]}" "${new_kmods[@]}" | sort | uniq -u)" ]; then
+            add_failure "new kernel modules loaded during execution"
+            echo "$(printf '%s\n' "${kmods[@]}" "${new_kmods[@]}" | sort | uniq -u)"
         fi
         update_failures
     update_summary "$cmd"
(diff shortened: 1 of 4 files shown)


TurboTurtle:
Hmmm....

So that's interesting, if not all that surprising.

On Fedora we're apparently loading `cfg80211` which is a wireless kmod. This unsurprisingly comes from the `wireless` plugin and the `iw` commands.

On Ubuntu we're loading a number of networking related modules as well. At first glance, could these be getting loaded by the `ufw` commands? @BryanQuigley @slashdd  

```
bpfilter
ebtable_filter
ebtables
ip6table_filter
ip6table_mangle
ip6table_nat
ip6table_raw
ip6_tables
iptable_filter
iptable_mangle
iptable_nat
iptable_raw
nf_nat
nf_tables
```

tests/simple.sh
```diff
@@ -101,6 +101,8 @@ add_failure () {
 # Test a no frills run with verbosity and make sure the expected items exist
 test_normal_report () {
     cmd="-vvv"
+    # get a list of initial kmods loaded
+    kmods=( $(lsmod | cut -f1 -d ' ') )
```
pmoravec, line 105: We do rely on the same ordering of modules in the `lsmod` output; isnt it safer to `sort` them (here and in `new_komds` assignment)?

BryanQuigley:
Just noticed (and minor).. can we move the bash shebang to the top of the simple.sh?

I'm running debian and got the following modules loaded additionally:
> dm_mod                159744  0
> binfmt_misc            24576  1

tests/simple.sh
TurboTurtle: afaik, the `sort` in the array comparison should be fine, but it would only help things to sort them on collection too, adding.

TurboTurtle:
> Just noticed (and minor).. can we move the bash shebang to the top of the simple.sh?
> 

Oh wow, that was weird. Fixed.

> I'm running debian and got the following modules loaded additionally:
> 
> > dm_mod                159744  0

I'm assuming this one is coming from `devicemapper`. Potentially also `dmraid`? Are either of these enabled by default on Debian?

> > binfmt_misc            24576  1

It's actually really surprising that's not loaded by default. I'm not entirely sure if that could even be narrowed down to a plugin or two.

pmoravec (approved):

BryanQuigley:
On a fresh boot both dm_raid and binfmt_misc they are both already there.. which is as to be expected.

I have no idea what that was about...  diving into to an Ubuntu VM in a bit.

BryanQuigley:
On Ubuntu:
A simple lxc image list (or maybe any lxd command)?  That seems quite bad as lxd is installed by default, but not setup on the majority of installs...

binfmt is brought in by the kernel plugin, need to dive more into that.  I wonder if it only starts when we access that dir?

BryanQuigley:
Really didn't expect this.. afaict it's running sysctl -a (for both Ubuntu 20.04 and Debian unstable) that loads the binfmt_misc module.

BryanQuigley:
binfmt_misc is automounted on debian/ubuntu now based on this upstream code (although it seems to exist upstream for a while).. https://github.com/systemd/systemd/blob/c5b6b4b6d08cf4c16a871401358faeb5a186c02a/units/proc-sys-fs-binfmt_misc.automount

I guess we could special case binfmt_misc in the tests?  Not seeing another way around it.

TurboTurtle:
Was it just running `sysctl -a` from your tests that loaded it? In keeping with our no-change statement I'd think we should lean towards adding a predicate to that collection.

BryanQuigley:
I ended up debugging binfmt_misc first:
it's also brought in from system.py, needs exclusion: "/proc/sys/fs/binfmt_misc",  We've been previously discussing excluding that too.. but I can make a new issue on it if we want that separate. 

sysctl --system doesn't trigget it, but captures much less.. will try to determine if we already capture most of it.

All the lxd/lxc commands cause modules to load AFAICT (including lxd --help..).

TurboTurtle:
\o/ There we go.

So `lxd` was loading the full list above (as were the `lxc` commands), whereas the `ufw` commands loaded only 2 of them. Predicates in place and the test is now showing no new modules being loaded across the board.

BryanQuigley:
I swear my tests didn't show ufw doing anything, but eh.

LGTM!

BryanQuigley (approved):

PR #2330, opened 2020-12-08
[plugins] Dont stop collecting by empty specfile when sizelimit=0

pmoravec:
When sizelimit=0, collecting an empty file would set limit_reached
wrongly and stop collecting further files. Let fix this corner case.

Resolves: #2330

---


pmoravec:
Trivial reproducer (though bit artificial):

    touch /var/log/sa/sar32
    sos report -o sar -k sar.all_sar=on --batch --build

That should collect all `/var/log/sa/*` files, but it collects just the empty `sar32` one.

TurboTurtle (approved):
Ack.

I don't want to imagine the scenario that lead to this discovery ;-)

Issue #2334, opened 2020-12-09
--log-size=0 is silently ignored


pmoravec:
Reproducer: have a huge `/var/log/messages` and call:

    sos report -o logs --batch --build --log-size=0 -vvv

Then the zero value is lost during:

    [sos.report:setup] executing 'sos report -o host --batch --build -vvv --log-size=0'
    [sos.report:setup] using 'rhel' preset defaults ()
    [sos.report:setup] effective options now: --batch --build --only-plugins host -vvv

and the default logsize limit is applied.

TurboTurtle:
Oof, that's a rough one. Good find though.

This came about because in `Component.apply_options_from_cmdline()` was checking for `if val`. Well in almost all cases that's right, but for `log-size=0` that (now obviously) won't fly. So we can change that to `if val is not None`, which is what _should_ be the null value for anything coming from argparse.

I've opened #2335 for this, let's get some testing on this to make sure we're copacetic with the change.

PR #2335, opened 2020-12-09
[options] Fix --log-size=0 being ignored and unreported otherwise

TurboTurtle:
The `--log-size` option was being silently ignored, due to a too-loose
conditional in `Component.apply_options_from_cmdline()` which was
inadvertently filtering out the option when a user set it to 0. Note
that this did not affect `sos.conf` settings for this same value.

Similarly, reporting the effective options after preset and cmdline
merging was skipping log-size when it was set to 0, since we normally
want to filter out null-value options (which imply they were not
invoked). Adding an explicit check for `log-size` here is the easiest
route forward to allow the reporting we expect.

Closes: #2334
Resolves: #2335

---

### DIFF
--- a/sos/options.py
+++ b/sos/options.py
@@ -282,6 +282,9 @@ class SoSOptions():
             """
             if name in ("add_preset", "del_preset", "desc", "note"):
                 return False
+            # Exception list for options that still need to be reported when 0
+            if name in ['log_size', 'plugin_timeout'] and value == 0:
+                return True
             return has_value(name, value)
 
         def argify(name, value):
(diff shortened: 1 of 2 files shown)


sos/options.py
```diff
@@ -282,6 +282,9 @@ def filter_opt(name, value):
             """
             if name in ("add_preset", "del_preset", "desc", "note"):
                 return False
+            # Exception list for options that still need to be reported when 0
+            if name in ['log_size'] and value == 0:
```
pmoravec, line 286: Could we add here also `plugin_timeout`? It suffers by the same issue.

Otherwise great catch in both places. All my testing passed.

TurboTurtle: Ack

pmoravec (approved):

Issue #2333, opened 2020-12-09
sos report does not collect broken symlinks


pmoravec:
Another rather question than an issue (so far): should we collect broken symlinks?

Reproducer:

    ln -s /var/log/brokenlinkdest /var/log/messagesbrokenlink
    sos report -o logs --batch --build

Generated report does not contain the broken symlink from `/var/log/messagesbrokenlink`. The cause is that https://github.com/sosreport/sos/blob/master/sos/report/plugins/__init__.py#L1454 fails to stat the file to determine its size. Easy to fix by prep-ing the "`continue`" there by `if not (os.path.islink(_file)):` test but the question is: should we collect broken symlinks? I would say so, as the information about the broken symlink (and where it points) is useful.

PR #2338, opened 2020-12-10
[report] collect broken symlinks

pmoravec:
Information about broken symlink destination is useful information
that sos report should collect. Currently it stops doing so as
stat-ing the symlink to determine filesize fails.

Resolves: #2333

---

### DIFF
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -1449,11 +1449,16 @@ class Plugin(object):
                     continue
 
                 try:
-                    filestat = os.stat(_file)
+                    file_size = os.stat(_file)[stat.ST_SIZE]
                 except OSError:
-                    self._log_info("failed to stat '%s'" % _file)
-                    continue
-                current_size += filestat[stat.ST_SIZE]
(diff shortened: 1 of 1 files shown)


sos/report/plugins/__init__.py
```diff
@@ -1449,11 +1449,16 @@ def time_filter(path):
...
-                    filestat = os.stat(_file)
+                    file_size = os.stat(_file)[stat.ST_SIZE]
                 except OSError:
                     self._log_info("failed to stat '%s'" % _file)
-                    continue
-                current_size += filestat[stat.ST_SIZE]
+                    # if _file is a broken symlink, we should collect it,
+                    # otherwise continue
+                    if os.path.islink(_file):
+                        file_size = 0
```
TurboTurtle, line 1458: I think we should only log a failure if it is not a broken symlink in this case, no?

The logged message would make me think "ok, we didn't collect that", but then seeing that (broken) collection in the archive would make me stop for a moment.

pmoravec: Ah, good point, will fix it.

PR #2339, opened 2020-12-10
[containers_common] Gather container digests

badone:
It can be useful to have the digest for each container image for
verifying exact binary compatibility, etc.

Signed-off-by: Brad Hubbard <bhubbard@redhat.com>

---

### DIFF
--- a/sos/report/plugins/containers_common.py
+++ b/sos/report/plugins/containers_common.py
@@ -41,6 +41,7 @@ class ContainersCommon(Plugin, RedHatPlugin, UbuntuPlugin):
             'podman unshare cat /proc/self/uid_map',
             'podman unshare cat /proc/self/gid_map',
             'podman images',
+            'podman images --digests',
             'podman pod ps',
             'podman port --all',
             'podman ps',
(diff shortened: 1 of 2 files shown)


badone:
I added this as an additional command as it can create pretty wide output that has the normal wrapping issues and I reasoned it may be a good idea to leave the existing command as is where wrapping is less of an issue.

TurboTurtle (commented):
So this will capture the command for _only_ non-root users given to the containers_common plugin via `-k containers_common.rootlessusers=foo`.

I think this should also then be added to the `podman` plugin, where we get the info as root as well.

badone:
> So this will capture the command for _only_ non-root users given to the containers_common plugin via `-k containers_common.rootlessusers=foo`.
> 
> I think this should also then be added to the `podman` plugin, where we get the info as root as well.

I see. Let me look into this.

badone:
@TurboTurtle How's that?

TurboTurtle (approved):

pmoravec (approved):

badone:
Thanks.

PR #2346, opened 2020-12-15
[component] Use sysroot from Policy when opts doesn't specify it

pmoravec:
Until --sysroot option is specified, Archive (sub)classes should
be called with sysroot determined from Policy.

Resolves: #2346

---

### DIFF
--- a/sos/component.py
+++ b/sos/component.py
@@ -246,13 +246,13 @@ class SoSComponent():
             auto_archive = self.policy.get_preferred_archive()
             self.archive = auto_archive(archive_name, self.tmpdir,
                                         self.policy, self.opts.threads,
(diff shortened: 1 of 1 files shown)


pmoravec:
Compare the changed code with original one in https://github.com/sosreport/sos/blob/legacy-3.9/sos/sosreport.py#L417 / https://github.com/sosreport/sos/blob/legacy-3.9/sos/sosreport.py#L422 .

Particular reproducer: I would say very generic one: run `sos report` inside a container, where it is supposed to collect *any* file accessible in *directory* that exists under the sysroot dir only.

TurboTurtle (approved):

PR #2348, opened 2020-12-16
[networking/ufw/firewall_tables] Split out firewall from networking

BryanQuigley:
Drop fallback to nat as that can bring in more kernel modules
Left ip/6-table save in networking for now as it can be filtered by
networking options and I don't want to break that.

---

### DIFF
--- a/sos/report/plugins/firewall_tables.py
+++ b/sos/report/plugins/firewall_tables.py
@@ -1,0 +1,87 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import (Plugin, IndependentPlugin, SoSPredicate)
+
+
+class firewall_tables(Plugin, IndependentPlugin):
+
+    short_desc = 'firewall tables'
+
+    plugin_name = "firewall_tables"
+    profiles = ('network', 'system')
+
+    def collect_iptable(self, tablename):
+        """ Collecting iptables rules for a table loads either kernel module
+        of the table name (for kernel <= 3), or nf_tables (for kernel >= 4).
+        If neither module is present, the rules must be empty."""
+
+        modname = "iptable_" + tablename
+        cmd = "iptables -t " + tablename + " -nvL"
+        self.add_cmd_output(
+            cmd,
+            pred=SoSPredicate(self, kmods=[modname, 'nf_tables']))
+
+    def collect_ip6table(self, tablename):
+        """ Same as function above, but for ipv6 """
+
+        modname = "ip6table_" + tablename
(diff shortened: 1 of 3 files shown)


sos/report/plugins/firewall_tables.py
```diff
@@ -0,0 +1,79 @@
+from sos.report.plugins import (Plugin, IndependentPlugin, SoSPredicate)
+
+
+class firewall_tables(Plugin, IndependentPlugin):
```
TurboTurtle, line 4: Since we already have a `firewalld` plugin, and this PR is adding a `ufw` plugin, I think we should should name this new plugin simply `iptables`, to keep consistent with the component <-> plugin naming.

sos/report/plugins/ufw.py
```diff
@@ -0,0 +1,26 @@
+from sos.report.plugins import (Plugin, IndependentPlugin, SoSPredicate)
```
TurboTurtle, line 1: Needs the license header. Optionally a copyright as well if you like.

sos/report/plugins/firewall_tables.py
```diff
@@ -0,0 +1,79 @@
+from sos.report.plugins import (Plugin, IndependentPlugin, SoSPredicate)
```
TurboTurtle, line 1: Needs the license header. Optionally a copyright line as well if you like.

TurboTurtle (commented):
Logical change to make - anything that simplifies the `networking` plugin always has my support ;-)

1 minor change noted, and just need the license header in the new plugins (note to self: see if we're missing these anywhere else currently). Otherwise this LGTM.

sos/report/plugins/firewall_tables.py
BryanQuigley: Including nftables in iptables seemed a bit odd to me, but if you think it won't be confusing, happy to do that.

TurboTurtle: That's a fair point. Part of my hesitation with the use of `firewall_tables` is we (will) have 2 other firewall plugins. Rock and a hard place with this naming it seems. I'm open for other suggestions as well of course.

`nftables` is supposed to outright replace `iptables` though, if I'm not mistaken. That, coupled with the fact that pretty much any support engineer or sysadmin is going to know the name `iptables` and its context (but the same can't be necessarily said for `nftables`, is what makes me lean towards naming it `iptables`.

@pmoravec @bmr-cymru thoughts?

pmoravec: I lean towards `firewall_tables`, as that is technically accurate while `[ip|nf]tables` is / will be confusing. But I deal the name purely from the technical perspective, hard to say what is the first word a user comes up when thinking about firewall (now or in next few years)..

pmoravec:
ACK up to the license headers.

sos/report/plugins/firewall_tables.py
TurboTurtle: Works for me then. Ack to the name.

TurboTurtle (approved):

PR #2349, opened 2020-12-17
[Policy] Break up monolithic policy into more organized pieces

TurboTurtle:
2-commit patchset that simply moves `InitSystem`, `ContainerRuntime`, and their related subclasses out of `sos/policies/__init__.py` and into their own files.

Just a bit of housekeeping to better organize.

---


BryanQuigley:
+1 makes it nicer to understand how things are organized.

I don't think it will have any negative impacts today, but this means runtime and init_system are iterated through potentially as part of the check what policy to use (line 50 policies/__init__.py).

TurboTurtle:
That's a good point actually that I hadn't considered, and we should avoid that.

Curious if we should place these in something like `sos/policies/runtimes/...` or somewhere else in the tree entirely?

BryanQuigley:
In one of my redo os detection attempts.  I had considered moving the os policies to their own os directory.

TurboTurtle:
So something like this then?

```
sos/
    - policies/
        - __init__.py
        - os/
            - redhat.py
            - ubuntu.py
        - runtimes/
            - podman.py
        - initsystems/
            - systemd.py
```

At first grok that looks good to me, and it'd make it easier to organize policy utilities (e.g. `load()`) without circling back to where we are today with a bloated `sos/policies/__init__.py`.

BryanQuigley:
That structure LGTM

pmoravec:
+1 for the proposal re-organized to the new directory structure.

TurboTurtle:
Well that certainly spread out more than I thought it would.

I've updated this branch to:

* Split out `ContainerRuntime` under `sos/policies/runtimes` with distinct `docker` and `podman` subclasses in that dir
* Split out `InitSystem` under `sos/policies/init_systems` with a distinct `systemd` subclass
* Split out `PackageManager` under `sos/policies/package_managers` and *added* subclasses for `dpkg` and `rpm`, and then updated the policies as needed
* Moved the actual policy classes under `sos/policies/distros`. Using `sos/policies/os` ended up with a name conflict with the regular `os` module unfortunately.
* Split out `PresetDefaults` under `sos/presets`. Then moved the RH-family presets defined in the policy under `sos/presets/redhat`, and imported those back into the policy. I think this will allow us to have better organized presets in the future, with an easier time of sharing and/or separating presets among distributions

BryanQuigley (approved):

pmoravec:
Browsing source code for leftovers:

We need to update https://github.com/sosreport/sos/blob/master/sos/report/plugins/python.py#L45 (and maybe the `import` line there?).

Not sure if some change should be done to https://github.com/sosreport/sos/blob/master/docs/policies.rst and/or https://github.com/sosreport/sos/blob/master/setup.py#L100

pmoravec:
Commit https://github.com/sosreport/sos/pull/2349/commits/746ba6261045d0525f149eabd26c785bbeb07a4d) merges changes on different places (rpm and SuSe, here), but I would let it be..

Otherwise ACK.

PR #2350, opened 2020-12-17
[redhat] Automatically failover to dropbox when RHCP upload fails

TurboTurtle:
When an upload to the RHCP fails for some reason, now automatically
failover to attempt an upload to the FTP dropbox.

This failover has been intended for some time, as we already failover
to the FTP location if we detect we're missing, for example, login
credentials for the Customer Portal. This change just extends that
intention to when we run into errors with the actual upload process.

Resolves: #2350

---

### DIFF
--- a/sos/policies/redhat.py
+++ b/sos/policies/redhat.py
@@ -364,6 +364,21 @@ support representative.
             return os.getenv('SOSUPLOADUSER', None) or self.upload_user
         return self._upload_user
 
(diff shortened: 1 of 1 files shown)


pmoravec (approved):

TurboTurtle:
Small update: only do a failover if our upload target was actually the RHCP.

pmoravec (approved):

Issue #1204, opened 2018-02-05
[RFE] provide SHA256 checksum 


knumskull:
The default checksum for sosreport is `md5sum`. On some support-portals, (e.g. Red Hat), the default checksum after uploading the file is `sha256sum`.
This would require an extra step for verification the successful upload. 

An additional `sha256sum` beside the already existing `md5sum` would be great. If generating to much overhead, defining the default checksum-method in `sos.conf` would be sufficient instead.

pmoravec:
I understand MD5 checksum is so legacy feature we shall not replace until very solid rationale behind. *additional* checksum is an option, but:
- that might confuse users what checksum(s) file are relevant or why there are two
- usability of that file on a customer portal relies on customers to really upload the extra file (how many times a customer uploads the md5 file?)

Note that when using default "xz" compressing algorithm, it is calculating crc64 checksum of the archive by default - see man pages:

       -C check, --check=check
              Specify  the  type of the integrity check.  The check is calculated from the uncompressed data and stored in the .xz file.  This option has an effect only when compressing
              into the .xz format; the .lzma format doesn't support integrity checks.  The integrity check (if any) is verified when the .xz file is decompressed.

              Supported check types:

              none   Don't calculate an integrity check at all.  This is usually a bad idea.  This can be useful when integrity of the data is verified by other means anyway.

              crc32  Calculate CRC32 using the polynomial from IEEE-802.3 (Ethernet).

              crc64  Calculate CRC64 using the polynomial from ECMA-182.  This is the default, since it is slightly better than CRC32 at detecting damaged files and the speed difference
                     is negligible.

              sha256 Calculate SHA-256.  This is somewhat slower than CRC32 and CRC64.

              Integrity of the .xz headers is always verified with CRC32.  It is not possible to change or disable it.

One can check how it works as follows:

    $ date > date.txt
    $ xz date.txt
    $ file date.txt.xz       # check for file format / xz archive
    date.txt.xz: XZ compressed data
    $ xz -l date.txt.xz      # check if it contains a checksum
    Strms  Blocks   Compressed Uncompressed  Ratio  Check   Filename
        1       1         88 B         30 B  2.933  CRC64   date.txt.xz
    
    $ xz -t date.txt.xz      # test integrity checksum
    $ echo $?                # return value 0 means OK
    0
    $ cat date.txt.xz | sed "s/f/ /g" > date-corrupted.txt.xz  # corrupt the archive by replacing some character (this wont work the same way every time)
[...]

PR #2351, opened 2020-12-17
[policy] switch to sha256 sum

BryanQuigley:
WIP - need cleaner/sosnode changes

Resolves: #2351
Closes: #1204

---

### DIFF
--- a/sos/collector/sosnode.py
+++ b/sos/collector/sosnode.py
@@ -773,6 +773,7 @@ class SosNode():
         self.ui_msg('Generating sosreport...')
         try:
             path = False
+            checksum = False
             res = self.run_command(self.sos_cmd,
                                    timeout=self.opts.timeout,
                                    get_pty=True, need_root=True,
@@ -781,6 +782,19 @@ class SosNode():
                 for line in res['stdout'].splitlines():
                     if fnmatch.fnmatch(line, '*sosreport-*tar*'):
                         path = line.strip()
+                    if line.startswith((" sha256\t", " md5\t")):
+                        checksum = line.split("\t")[1]
+                    elif line.startswith("The checksum is: "):
+                        checksum = line.split()[3]
+
+                if checksum is not None:
+                    self.manifest.add_field('checksum', checksum)
+                    if len(checksum) == 32:
+                        self.manifest.add_field('checksum_type', 'md5')
+                    elif len(checksum) == 64:
+                        self.manifest.add_field('checksum_type', 'sha256')
+                    else:
+                        self.manifest.add_field('checksum_type', 'unknown')
             else:
                 err = self.determine_sos_error(res['status'], res['stdout'])
                 self.log_debug("Error running sosreport. rc = %s msg = %s"
@@ -857,10 +871,6 @@ class SosNode():
                 except Exception:
                     self.log_error('Failed to make archive readable')
                     return False
-                try:
-                    self.make_archive_readable(self.sos_path + '.md5')
-                except Exception:
-                    self.log_debug('Failed to make md5 readable')
             self.soslog.info('Retrieving sosreport from %s' % self.address)
             self.ui_msg('Retrieving sosreport...')
             ret = self.retrieve_file(self.sos_path)
@@ -870,9 +880,6 @@ class SosNode():
             else:
                 self.log_error('Failed to retrieve sosreport')
                 raise SystemExit
-            self.hash_retrieved = self.retrieve_file(self.sos_path + '.md5')
-            if self.hash_retrieved:
-                self.file_list.append(self.sos_path.split('/')[-1] + '.md5')
             return True
         else:
             # sos sometimes fails but still returns a 0 exit code
@@ -901,7 +908,9 @@ class SosNode():
     def cleanup(self):
         """Remove the sos archive from the node once we have it locally"""
         self.remove_sos_archive()
-        if self.hash_retrieved:
+        if os.path.isfile(self.sos_path + '.sha256'):
+            self.remove_file(self.sos_path + '.sha256')
+        elif os.path.isfile(self.sos_path + '.md5'):
             self.remove_file(self.sos_path + '.md5')
         cleanup = self.host.set_cleanup_cmd()
         if cleanup:
--- a/sos/cleaner/__init__.py
+++ b/sos/cleaner/__init__.py
@@ -215,7 +215,8 @@ third party.
         nested_archives = []
         for _file in archive.getmembers():
             if (re.match('sosreport-.*.tar', _file.name.split('/')[-1]) and not
-                    _file.name.endswith('.md5')):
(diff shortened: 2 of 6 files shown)


TurboTurtle:
Thus far I haven't found anything internal on the RH side that would be adversely affected by this change. I'll update again if that changes, but so far this is looking like it is/will be an easy low-impact change.

BryanQuigley:
Haven't tested the collector changes yet

TurboTurtle:
I ran a quick test on the `collect` bits, couple notes:

- There is an attempt to grab the `.sha256` file from the master node, regardless of whether it exists or not. This is likely an existing bug just getting highlighted by the fact that the master in my case was not running this branch. 
- The `.md5` sums that are generated by current releases of sos are not gathered. For this change we'd need `collect` to gather both `.md5` and `.sha256` files if they exist, to account for mix-version node lists.

BryanQuigley:
I found it easier to just switch to the hash outputed via stdout.  In this iteration the checksum just gets added to the json manifest.    Thoughts?

TurboTurtle:
Hmm, I hadn't considered that, but it is a nice alternative to collecting the actual checksum files.

I don't _think_ there'd be any issues with dropping the file collection, and it'd arguably make things easier on any automated inspection of sos-collect tarballs.

I'll find some time this week to run this updated branch and in the meantime see if I can find any "gotcha"s. If not, then I think writing it to the manifest instead is a nice improvement.

TurboTurtle:
Hmmm. Ran this and I'm not seeing a checksum added to the `collect` manifest.

BryanQuigley:
It's certainly possible I missed some bits.. my test case was to a VM running an older version and the localhost:
bin/sos collect --nodes 192.168.122.183 --ssh-user bryan --password

That results in the local collection using from git so sha256sum, and the remote one using md5

TurboTurtle:
Yeah, I'd expect there to be different checksum types for older versions based on the commit, but what I mean is neither are getting written to the `manifest.json` created for the `collect` run.

```
$ sudo ./bin/sos collect --master=redacted.redhat.com --password

[...]
The following archive has been created. Please provide it to your support team.
	/var/tmp/sos-collector-123456-2021-01-29-ppbbu.tar.xz

[.. extract it ..]

[jake@terra sostmp]$ grep md5 sos-collector-123456-2021-01-29-ppbbu/sos_reports/manifest.json 
[jake@terra sostmp]$ grep sha sos-collector-123456-2021-01-29-ppbbu/sos_reports/manifest.json 
[jake@terra sostmp]$ 
```

BryanQuigley:
Hmm, I switched mine to do master node with type none and still find the md5 present. What cluster type are you using?

TurboTurtle:
It's a RHV (oVirt) cluster, however that shouldn't be making a difference since the manifest is handled outside of any cluster profiles.

Just tried again using `--cluster-type none` and I get the same result. No md5/sha sum added to the manifest generated for `collect`.

TurboTurtle:
Just to make sure it wasn't something with my specific workstation, I loaded up a fresh RHEL 8 VM and tried it there as well - same results.

BryanQuigley:
I just did a bunch of attempts including:
One was from my debian desktop to two centos 8 VMs. (bin/sos collect --no-local --nodes 192.168.122.233 --nodes 192.168.122.25 --cluster-type none --password)
One was from a CentOS 8 VM to include another centos VM (bin/sos collect --nodes 192.168.122.233 --cluster-type none --password)

All have md5s added to the manifest.json. Just to confirm your new VM test was just a single node with no cluster setup on it?

TurboTurtle:
Correct, new VM was just a single node.

Out of curiosity can you post a snippet of the manifest with the checksum?

BryanQuigley:
{
    "version": "4.0",
    "cmdline": "collect --no-local --nodes 192.168.122.233 --nodes 192.168.122.25 --cluster-type none --password",
    "start_time": "2021-02-05 10:13:24.422109",
    "end_time": "2021-02-05 10:14:21.263212",
    "run_time": "0:00:56.841103",
    "compression": "auto",
    "policy": "Debian",
    "components": {
        "collect": {
            "cluster_type": null,
            "node_list": [
                "192.168.122.233",
                "192.168.122.25"
            ],
            "initial_sos_cmd": "sosreport --batch -c auto",
            "master": "localhost",
            "nodes": {
                "localhost": {
                    "hostname": "desktop",
                    "policy": "Debian",
                    "sos_version": null,
                    "final_sos_command": ""
                },
                "192.168.122.25": {
                    "hostname": "192.168.122.25",
                    "policy": "CentOS",
                    "sos_version": "3.9.1",
                    "final_sos_command": "/usr/sbin/sosreport --batch -c auto ",
                    "label": "",
                    "md5": "262cdc11fae401a47e982072b1fe660f",
                    "collected_archive": "sosreport-localhost-2021-02-05-hzlnftl.tar.xz"
                },
                "192.168.122.233": {
                    "hostname": "192.168.122.233",
                    "policy": "CentOS",
                    "sos_version": "3.9.1",
                    "final_sos_command": "/usr/sbin/sosreport --batch -c auto ",
                    "label": "",
                    "md5": "8cf67fe3ab0f28b03c6bb7c54b1409a8",
                    "collected_archive": "sosreport-localhost-2021-02-05-mbatihh.tar.xz"
                }
            }
        }
    }
}

TurboTurtle:
Ok, so it's not an ordering issue or anything - on my RHEL family boxes we're just straight up missing the checksum keys, yet the `collected_archive` which is added afterwards is present as expected.

About the only other thing I can think of right now is python version - what version is on the system you're running collect from?

sos/collector/sosnode.py
```diff
@@ -781,6 +781,15 @@ def execute_sos_command(self):
...
                         path = line.strip()
+                    if line.startswith(" sha256\t"):
+                        self.manifest.add_field('checksum', line.split("\t")[1])
+                        self.manifest.add_field('checksum_type', 'sha256')
+                    if line.startswith(" md5\t"):
+                        self.manifest.add_field('checksum', line.split("\t")[1])
+                        self.manifest.add_field('checksum_type', 'md5')
+                    if line.startswith("The checksum is: "):
+                        self.manifest.add_field('checksum', line.split()[3])
+                        self.manifest.add_field('checksum_type', 'unspecified_guess_by_size')
```
TurboTurtle, line 792: This should be automatically set, as the hash sizes are well-known. E.G. something like:

```python
if line.startswith((" sha256\t", " md5\t")):
    checksum = line.split("\t")[1]
if line.startswith("The checksum is: "):
    checksum = line.split()[3]
self.manifest.add_field('checksum', checksum)
self.manifest.add_field('checksum_type', 'md5' if len(checksum) == 32 else 'sha256')
```

TurboTurtle (approved):
Ack, tested against ovirt and a random list of nodes.

PR #2355, opened 2021-01-04
[ipa] dogtag debug logfile name format change

tscherf:
Starting with Dogtag 10.6 a new logging framework is used. As a result,
debug logfiles now have the format 'debug-<date>' rather than just 'debug'.

Signed-off-by: Thorsten Scherf <tscherf@redhat.com>

---

### DIFF
--- a/sos/report/plugins/ipa.py
+++ b/sos/report/plugins/ipa.py
@@ -50,13 +50,13 @@ class Ipa(Plugin, RedHatPlugin):
     def retrieve_pki_logs(self, ipa_version):
         if ipa_version == "v4":
             self.add_copy_spec([
-               "/var/log/pki/pki-tomcat/ca/debug",
+               "/var/log/pki/pki-tomcat/ca/debug*",
                "/var/log/pki/pki-tomcat/ca/system",
                "/var/log/pki/pki-tomcat/ca/transactions",
                "/var/log/pki/pki-tomcat/ca/selftests.log",
                "/var/log/pki/pki-tomcat/catalina.*",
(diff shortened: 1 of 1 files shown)


TurboTurtle:
Is there any log rotating happening with these files? The new glob will match those as well. Not saying that's an issue, but we just want to make sure we aren't going to get unexpected collections here.

Come to think of it, has the plain jane `/var/log/pki/pki-tomcat/ca/debug` ever been an actual file by itself? I.E. Have we ever been capturing what we need, or are the (old) files actually named something like `/var/log/pki/pki-tomcat/ca/debug123456`?

tscherf:
> Is there any log rotating happening with these files? The new glob will match those as well. Not saying that's an issue, but we just want to make sure we aren't going to get unexpected collections here.

Yes, there is log rotation. But I don't see this as a problem because usually when PKI logs are not part of sosreport (for whatever reason) what happens is that people ask for whole /var/log/pki/pki-tomcat/ folder anyway (attached are archive to a case). So I think we can save this one extra ping-pong with the proposed change.
 
> Come to think of it, has the plain jane `/var/log/pki/pki-tomcat/ca/debug` ever been an actual file by itself? I.E. Have we ever been capturing what we need, or are the (old) files actually named something like `/var/log/pki/pki-tomcat/ca/debug123456`?

In the past there was a file called 'debug' which is the current logfile used by the process and rotated logs called 'debug.<date>'. With the change of the logging framework the former format is not available anymore and only the latter format is used.

TurboTurtle (approved):
Ack.

pmoravec (approved):

PR #2359, opened 2021-01-05
[component] honour plugopts from config file

pmoravec:
Currently, config file plugopts are ignored as we overwrite it
in apply_options_from_cmdline by empty list default value from
cmdline.

Resolves: #2359

---

### DIFF
--- a/sos/component.py
+++ b/sos/component.py
@@ -192,7 +192,7 @@ class SoSComponent():
         for opt, val in codict.items():
             if opt not in cmdopts.arg_defaults.keys():
                 continue
-            if val is not None and val != opts.arg_defaults[opt]:
+            if val not in [None, [], ''] and val != opts.arg_defaults[opt]:
                 setattr(opts, opt, val)
 
         return opts


sos/component.py
```diff
@@ -192,7 +192,7 @@ def apply_options_from_cmdline(self, opts):
         for opt, val in codict.items():
             if opt not in cmdopts.arg_defaults.keys():
                 continue
-            if val is not None and val != opts.arg_defaults[opt]:
+            if val and val != opts.arg_defaults[opt]:
```
TurboTurtle, line 195: This was recently changed in c2ddb50 to fix the `log-size=0` issue, #2334. Changing this back will cause a regression.

TurboTurtle:
After some testing, I believe this is actually what we need:

```python
         for opt, val in codict.items():
             if opt not in cmdopts.arg_defaults.keys():
                 continue
-            if val is not None and val != opts.arg_defaults[opt]:
+            if val not in [None, [], ''] and val != opts.arg_defaults[opt]:
                 setattr(opts, opt, val)
 
         return opts

```

This way we handle both empty values, like the "overriding" plugopts default value of `[]` and the 0-and-None mashing that we have to deal with due to argparse as well as the theoretical empty string if anyone is setting string values in config files (label maybe?). Does it look stupid? Yes. Does it work? Apparently. ;-)

pmoravec:
Good point. I felt my PR is just fixing one use case but could remember the other ones. PR updated accordingly.

I hope nobody will want to set `verbose=0` in config file, that would be an ultimate break :D

TurboTurtle (approved):

PR #2361, opened 2021-01-08
[sapnw] call sapconf only when a limits.d file exists

pmoravec:
sapconf command creates /etc/security/limits.d/99-sap-limits.conf
on its own when missing, so calling the command must be gated
by a predicate testing the file presence.

Resolves: #2361

---

### DIFF
--- a/sos/report/plugins/sapnw.py
+++ b/sos/report/plugins/sapnw.py
@@ -129,6 +129,12 @@ class sapnw(Plugin, RedHatPlugin):
         self.collect_list_dbs()
 
         # run sapconf in check mode
-        self.add_cmd_output("sapconf -n", suggest_filename="sapconf_checkmode")
+        #
+        # since the command creates a limits.d file on its own,
+        # we must predicate it by presence of the file
+        if os.path.exists('/etc/security/limits.d/99-sap-limits.conf') \
+                or self.get_option('allow_system_changes'):
+            self.add_cmd_output("sapconf -n",
+                                suggest_filename="sapconf_checkmode")
 
 # vim: et ts=4 sw=4


TurboTurtle (commented):
I think maybe we should instead add file presence to `SoSPredicate`. E.G. `SoSPredicate(self, files=['foo', 'bar'])`

It would track with our other parts of plugin checks for both enabled and items like `is_executable`.

pmoravec:
I was thinking about that idea as well, but wasnt sure if it is worth of the effort. But no problem to add it that way, for sure.

TurboTurtle:
Yeah...the alternative is to just use the existing `os.path.exists()`. The only reason I see to add it to `SoSPredicate` now that I think about it a bit more is to have it combined with other checks, like "this package is installed AND this file exists", which I'm not sure we'd really use in practice.

pmoravec:
Good point with the simple `os.path.exists` test, which is the most easy one. Until (if ever) we will need to combine predicates like "if file exists and .., then ..", I think using this simple test is easier than to expand `SoSPredicate` redundantly.

bmr-cymru:
`sapconf` describes itself as a tool for preparing SLES systems to run SAP. Perhaps we shouldn't be running it at all? (or only on SLES?).

bmr-cymru:
Also `sapconf -n` is supposed to be "check" mode - no changes to system config. So that's looking like a `sapconf` bug also.

pmoravec:
Afaik there is a bug for the tool as well. The tool is shipped also in a RHEL repository.

TurboTurtle (approved):

PR #2362, opened 2021-01-11
[postfix] forbid collecting SSL keys files declared in main.cf

pmoravec:
Collecting whole /etc/postfix, we might collect some SSL keys placed
to this directory. Traverse main.cf to identify all such potential
files we must add to forbidden paths.

Resolves: #2362

---

### DIFF
--- a/sos/report/plugins/postfix.py
+++ b/sos/report/plugins/postfix.py
@@ -17,6 +17,41 @@ class Postfix(Plugin):
 
     packages = ('postfix',)
 
+    def forbidden_ssl_keys_files(self):
+        # list of attributes defining a location of a SSL key file
+        # we must forbid from collection
+        forbid_attributes = [
+            "lmtp_tls_dkey_file",
+            "lmtp_tls_eckey_file",
+            "lmtp_tls_key_file",
+            "smtp_tls_dkey_file",
+            "smtp_tls_eckey_file",
+            "smtp_tls_key_file",
+            "smtpd_tls_dkey_file",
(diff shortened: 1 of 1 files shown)


pmoravec:
The PR assumes `main.cf` has option only in form

    option = value

i.e. with spaces around `=`. If other format is possible, please let me know and I will amend the PR accordingly.

sos/report/plugins/postfix.py
```diff
@@ -17,6 +17,43 @@ class Postfix(Plugin):
...
+            cffile = open('/etc/postfix/main.cf', 'r')
+            lines = cffile.readlines()
+            cffile.close()
+            for line in lines:
+                # ignore comments and empty lines
+                # treat only lines like
+                # lmtp_tls_dkey_file = /path/to/file
+                words = line.split('#')[0].split()
+                if (len(words) == 3) and (words[0] in forbid_attributes):
+                    fp.append(words[2])
```
TurboTurtle, line 53: We could split on the `=` and side-step potential formatting problems in terms of whitespace and length:

```python3
with open('/etc/postfix/main.cf', 'r') as cffile:
    for line in cffile.readlines():
        if line.startswith('#'):
            continue
        words = line.split('=')
        if words[0] in forbid_attributes:
            fp.append(words[1])
```

pmoravec:
Makes sense. Some minor striping added, to remove trailing `\n` and split words after `=` (like `smtp_tls_key_file = /path/to/file # some comment`).

TurboTurtle (approved):

PR #2367, opened 2021-01-13
[block] Add lsblk all column output command

akshay196:
"lsblk -O -P" provides all available column output in key pair
format.

Signed-off-by: Akshay Gaikwad <akgaikwad001@gmail.com>

--


TurboTurtle (approved):

pmoravec (approved):
The -O option is not available on older versions of the `lsblk` command (i.e. afaik on any RHEL7 version), but thats not a big issue, I think.

Issue #2352, opened 2020-12-18
[process] Capture lsof output without -b


TurboTurtle:
A request has come in from the RHV support team to capture `lsof` without the use of the `-b` option. The use of this option causes the loss of some helpful data, as reported in RHBZ1890616, E.G.:

```
# lsof -b +M -p 3544479 -n -l -c '' | grep qemu | grep 'dm\-'
............
qemu-kvm   955375      107   29   unknown                           /dev/dm-20 lstat: Resource temporarily unavailable) (stat: Resource temporarily unavailable)
qemu-kvm   955375      107   90   unknown                           /dev/dm-21 lstat: Resource temporarily unavailable) (stat: Resource temporarily unavailable)
qemu-kvm  1017578      107   16   unknown                           /dev/dm-22 lstat: Resource temporarily unavailable) (stat: Resource temporarily unavailable)
qemu-kvm  1017578      107   89   unknown                           /dev/dm-23 lstat: Resource temporarily unavailable) (stat: Resource temporarily unavailable)
qemu-kvm  1697954      107   40   unknown                           /dev/dm-26 lstat: Resource temporarily unavailable) (stat: Resource temporarily unavailable)
qemu-kvm  3544479      107   39   unknown                           /dev/dm-25 lstat: Resource temporarily unavailable) (stat: Resource temporarily unavailable)
qemu-kvm  3544479      107   40   unknown                           /dev/dm-27 lstat: Resource temporarily unavailable) (stat: Resource temporarily unavailable)


# lsof -n -l -c '' | grep qemu | grep 'dm\-'
qemu-kvm   955375      107   29u      BLK             253,20         0t0  251083474 /dev/dm-20
qemu-kvm   955375      107   90u      BLK             253,21         0t0  251888993 /dev/dm-21
qemu-kvm  1017578      107   16u      BLK             253,22         0t0  252815128 /dev/dm-22
qemu-kvm  1017578      107   89u      BLK             253,23         0t0  252845504 /dev/dm-23
qemu-kvm  1697954      107   40u      BLK             253,26         0t0  271818320 /dev/dm-26
qemu-kvm  3544479      107   39r      BLK             253,25         0t0  448559389 /dev/dm-25
qemu-kvm  3544479      107   40u      BLK             253,27         0t0  448569269 /dev/dm-27
```

[...]

Issue #1979, opened 2020-03-12
Plugin.get_cmd_output_now does not honor "suggest_filename" param


mmartinv:
Passing "suggest_filename" to Plugin.get_cmd_output has no effect on the filename created.

mmartinv:
Sorry my copy was too old

PR #2370, opened 2021-01-13
[process] Drop -b and consolidate duplicate calls

TurboTurtle:
2-commit patchset that:

1. Drops `-b` per #2352 and adds a timeout to those `lsof` commands as a safeguard
2. Consolidates 2 duplicate `ps` calls to get some headway on #1979 

---


TurboTurtle:
@bmr-cymru any concerns over dropping `-b`? I am assuming it was initially used before we had timeout controls in place, and the RHV support team has reported it strips useful information for them (so I imagine it would also affect RHOSP).

pmoravec (approved):

PR #2376, opened 2021-01-18
[networking] Collect 'ethtool -e <device>' conditionally only

pmoravec:
As the command can hang and cause OOM issues on ovs-dpdk / tg3
NICs, we should gate its execution for that NIC type.

Resolves: #2376

---

### DIFF
--- a/sos/report/plugins/networking.py
+++ b/sos/report/plugins/networking.py
@@ -27,7 +27,8 @@ class Networking(Plugin):
         ("namespaces", "Number of namespaces to collect, 0 for unlimited. " +
          "Incompatible with the namespace_pattern plugin option", "slow", 0),
         ("ethtool_namespaces", "Define if ethtool commands should be " +
-         "collected for namespaces", "slow", True)
+         "collected for namespaces", "slow", True),
+        ("eepromdump", "collect 'ethtool -e' for all devices", "slow", False)
     ]
 
     # switch to enable netstat "wide" (non-truncated) output mode
@@ -141,16 +142,15 @@ class Networking(Plugin):
                 "ethtool --show-eee " + eth
             ], tags=eth)
 
-            # skip EEPROM collection for 'bnx2x' NICs as this command
-            # can pause the NIC and is not production safe.
-            bnx_output = {
-                "cmd": "ethtool -i %s" % eth,
-                "output": "bnx2x"
-            }
-            bnx_pred = SoSPredicate(self,
-                                    cmd_outputs=bnx_output,
-                                    required={'cmd_outputs': 'none'})
-            self.add_cmd_output("ethtool -e %s" % eth, pred=bnx_pred)
+            # skip EEPROM collection by default, as it might hang or
+            # negatively impact the system on some device types
+            if self.get_option("eepromdump"):
+                cmd = "ethtool -e %s" % eth
+                self._log_warn("WARNING (about to collect '%s'): collecting "
+                               "an eeprom dump is known to cause certain NIC "
+                               "drivers (e.g. bnx2x/tg3) to interrupt device "
+                               "operation" % cmd)
+                self.add_cmd_output(cmd)
 
         # Collect information about bridges (some data already collected via
         # "ip .." commands)


pmoravec:
I don't like the predicate runs `ethtool -i` twice for a NIC, but I don't see a way how to prevent that (apart of not using `SoSPredicate` that fits here pretty well otherwise).

mheler:
I worry that there may be other network card vendors that have this problem. Is this the best approach? Thinking out loud, would a boolean to toggle eeprom dumps (ethtool -e) make more sense?

pmoravec:
Thinking again and again on the topic, I lean towards the same idea; disable `ethtool -e` by default and enable it via a new plugin option (like `-k networking.ethtoole=on`). As this is the 2nd such problem in quite short time.

Though the original issue is in `ethtool` itself that isnt sufficiently robust, we should not allow `sos report` to trigger those issues.

@bmr-cymru or @TurboTurtle , your suggestions?

bmr-cymru:
`eepromdump` might be a more descriptive name for the option, or even just `eeprom`.

I think it's also worth filing a bug against the kernel component with Red Hat - if this ioctl isn't expected to not have side effects then it really should be mentioned in the documentation (and if it isn't expected to have side effects then there are some broken drivers/NICs out there).

It would also be useful to get more of an idea if/when this is actually useful - we were asked to add it relatively recently by Patrick Talbert (works on/supports OpenVSwitch iirc):

```
commit 8b989aebc9c152430fc57f918a8e90210a792a9f
Author: Patrick Talbert <ptalbert@redhat.com>
Date:   Thu Dec 6 13:14:38 2018 +0100

    [networking] Extend ethtool command set
    
    Update the list of ethtool commands to include:
    
    ethtool -e (EEPROM dump)
    ethtool -P (permanent MAC address)
    ethtool -l (channel/queue settings)
    ethtool --phy-statistics
    ethtool --show-priv-flags
    ethtool --show-eee
    
    All of the above are helpful in understanding the state of modern NICs.
    And -P is nice to have as otherwise there is no reliable way to see the
    permanent MAC of team ports.
    
    Resolves: #1503
    
```

bmr-cymru:
Commit message should be `skip 'ethtool -e'`?

pmoravec:
Collection of the command conditioned by the `eepromdump` plugopt, which is disabled by default.

sos/report/plugins/networking.py
```diff
@@ -27,7 +27,8 @@ class Networking(Plugin):
         ("namespaces", "Number of namespaces to collect, 0 for unlimited. " +
          "Incompatible with the namespace_pattern plugin option", "slow", 0),
         ("ethtool_namespaces", "Define if ethtool commands should be " +
-         "collected for namespaces", "slow", True)
+         "collected for namespaces", "slow", True),
+        ("eepromdump", "collect 'ethtool -e' for all devices", "fast", False)
```
bmr-cymru, line 31: We don't really do anything with the "fast"/"slow" metadata but since the plugin is already using the "slow" string for other options (and since this collection is known to cause "hiccups" of various kinds), perhaps we should use it for this new option?

sos/report/plugins/networking.py
```diff
@@ -141,16 +142,10 @@ def setup(self):
...
-                "output": "bnx2x"
-            }
-            bnx_pred = SoSPredicate(self,
-                                    cmd_outputs=bnx_output,
-                                    required={'cmd_outputs': 'none'})
-            self.add_cmd_output("ethtool -e %s" % eth, pred=bnx_pred)
+            # skip EEPROM collection by default, as it might hang or
+            # negatively impact the system on some device types
+            if self.get_option("eepromdump"):
+                self.add_cmd_output("ethtool -e %s" % eth)
```
bmr-cymru, line 148: Should we be enabling this for devices that are already "known bad", even when the option is used? I.e. bnx2, possibly tg3 and others?

(genuine question - I could argue it either way...)

mheler: I would personally leave that up to the operator to decide that. If the default is to skip the eeprom dumps, then that seems pretty safe and sane to me.

TurboTurtle: We could add a warning message here perhaps? That way sysadmins still have a chance to abort the collection.

Something like this maybe?

`WARNING: collecting an eeprom dump is known to cause certain NIC drivers (e.g. bnx2x/tg3) to interrupt device operation`

bmr-cymru: > I would personally leave that up to the operator to decide that. If the default is to skip the eeprom dumps, then that seems pretty safe and sane to me.

The problem is that nothing currently describes these actions as potentially dangerous (which we know from experience they are). Neither the kernel ioctl documentation, nor the ethtool docs mention any possible side effects from the eeprom dump collection. Given that we are now aware of these problems I think we should communicate that to the user - a warning message is an option, but I think the option description should also flag that using this plugin option has the potential for very serious side effects.

bmr-cymru: iow "leave it to the operator" assumes that the operator is properly informed and understands what they are consenting to. I don't think that's the case currently.

pmoravec:
Thanks for productive feedback. "slow" option makes sense, as well as the warning.

An extra predicate about the NIC type : I see both pros (to provide really conservative behaviour) and cons (we properly warned an operator that just explicitly enabled the option), bit leaning towards "just" the warning message.

pmoravec:
OK, additionally I added to the warning message the explicit command in question, to help the operator to check if the cmd might be hanging (and also to let the operator to understand, why the warning message is printed multiple times - we *can* change it to one print, but not in a trivial way).

TurboTurtle (approved):
This LGTM at this point, though I'm still open to hearing more opinions about warning message(s) or further restricting from ever running on known "bad" drivers.

pmoravec:
Just a FYI update: it seems yet another device type (i40e) can trigger similar unwanted misbehaviour, see https://bugzilla.redhat.com/show_bug.cgi?id=1917196 .

PR #2379, opened 2021-01-19
[powerpc] Collect logs for power specific components (HNV and SCSI)

mamatha4:
This patch set enables sosreport to capture
logs for power specific components (HNV and SCSI)

signed-off-by: Mamatha Inamdar mamatha4@linux.vnet.ibm.com

### DIFF
--- a/sos/report/plugins/powerpc.py
+++ b/sos/report/plugins/powerpc.py
@@ -63,7 +63,10 @@ class PowerPC(Plugin, IndependentPlugin):
                 "/proc/ppc64/systemcfg",
                 "/var/log/platform",
                 "/var/log/drmgr",
-                "/var/log/drmgr.0"
+                "/var/log/drmgr.0",
+                "/var/log/hcnmgr",
+                "/var/ct/IBM.DRM.stderr",
+                "/var/ct/IW/log/mc/IBM.DRM/trace*"
             ])
             ctsnap_path = self.get_cmd_output_path(name="ctsnap", make=True)
             self.add_cmd_output([
@@ -74,8 +77,10 @@ class PowerPC(Plugin, IndependentPlugin):
                 "serv_config -l",
                 "bootlist -m both -r",
                 "lparstat -i",
-                "ctsnap -xrunrpttr -d %s" % (ctsnap_path)
+                "ctsnap -xrunrpttr -d %s" % (ctsnap_path),
+                "lsdevinfo"
             ])
+            self.add_service_status("hcn-init")
 
         if isPowerNV:
             self.add_copy_spec([
--- a/sos/report/plugins/scsi.py
+++ b/sos/report/plugins/scsi.py
@@ -29,10 +29,18 @@ class Scsi(Plugin, IndependentPlugin):
         ])
 
         self.add_cmd_output("lsscsi -i", suggest_filename="lsscsi")
-        self.add_cmd_output("sg_map -x")
-        self.add_cmd_output("lspath")
-        self.add_cmd_output("lsmap -all")
-        self.add_cmd_output("lsnports")
+
+        self.add_cmd_output([
+            "sg_map -x",
+            "lspath",
+            "lsmap -all",
+            "lsnports",
+            "lsscsi -H",
+            "lsscsi -g",
+            "lsscsi -d",
+            "lsscsi -s",
+            "lsscsi -L"
+        ])
 
         scsi_hosts = glob("/sys/class/scsi_host/*")
         self.add_blockdev_cmd("udevadm info -a %(dev)s", devices=scsi_hosts,
--- a/sos/report/plugins/kernel.py
+++ b/sos/report/plugins/kernel.py
@@ -106,7 +106,6 @@ class Kernel(Plugin, IndependentPlugin):
             "/proc/misc",
             "/var/log/dmesg",
             "/sys/fs/pstore",
-            "/var/log/hcnmgr",
             clocksource_path + "available_clocksource",
             clocksource_path + "current_clocksource"
         ])


sos/report/plugins/powerpc.py
```diff
@@ -63,7 +63,10 @@ def setup(self):
                 "/proc/ppc64/systemcfg",
                 "/var/log/platform",
                 "/var/log/drmgr",
-                "/var/log/drmgr.0"
+                "/var/log/drmgr.0",
+                "/var/log/hcnmgr",
+                "/var/ct/IBM.DRM.stderr",
+                "/var/ct/IW/log/mc/IBM.DRM/trace*"
```
TurboTurtle, line 69: Are these regular log files that happen to be named `trace*`, or actual tracing like what is seen in `/sys/kernel/debug/tracing/trace_pipe`?

sos/report/plugins/powerpc.py
```diff
@@ -74,7 +77,10 @@ def setup(self):
                 "serv_config -l",
                 "bootlist -m both -r",
                 "lparstat -i",
-                "ctsnap -xrunrpttr -d %s" % (ctsnap_path)
+                "ctsnap -xrunrpttr -d %s" % (ctsnap_path),
+                "lsdevinfo",
+                "journalctl",
+                "systemctl is-enabled hcn-init.service"
```
TurboTurtle, line 83: There is `Plugin.add_service_status()`, which should provide this information while also using the existing policy hooks.

sos/report/plugins/scsi.py
```diff
@@ -33,6 +33,11 @@ def setup(self):
         self.add_cmd_output("lspath")
         self.add_cmd_output("lsmap -all")
         self.add_cmd_output("lsnports")
+        self.add_cmd_output("lsscsi -H")
+        self.add_cmd_output("lsscsi -g")
+        self.add_cmd_output("lsscsi -d")
+        self.add_cmd_output("lsscsi -s")
+        self.add_cmd_output("lsscsi -L")
```
TurboTurtle, line 40: Let's do this as a single call to `add_cmd_output()` with a list instead.

sos/report/plugins/powerpc.py
```diff
@@ -74,7 +77,10 @@ def setup(self):
                 "serv_config -l",
                 "bootlist -m both -r",
                 "lparstat -i",
-                "ctsnap -xrunrpttr -d %s" % (ctsnap_path)
+                "ctsnap -xrunrpttr -d %s" % (ctsnap_path),
+                "lsdevinfo",
+                "journalctl",
```
TurboTurtle, line 82: Didn't see this earlier, but nack to this blind `journalctl` call. We get the system journal in the `logs` plugin already, and if you want/need a service specific journal you can use `Plugin.add_journal()`.

sos/report/plugins/powerpc.py
mamatha4: these are trace.sp files as below... 
trace.10.sp  trace.13.sp  trace.16.sp  trace.19.sp  trace.21.sp  trace.24.sp  trace.27.sp  trace.2.sp   trace.32.sp  trace.35.sp  trace.38.sp  trace.40.sp  trace.6.sp  trace.9.sp
trace.11.sp  trace.14.sp  trace.17.sp  trace.1.sp   trace.22.sp  trace.25.sp  trace.28.sp  trace.30.sp  trace.33.sp  trace.36.sp  trace.39.sp  trace.4.sp   trace.7.sp
trace.12.sp  trace.15.sp  trace.18.sp  trace.20.sp  trace.23.sp  trace.26.sp  trace.29.sp  trace.31.sp  trace.34.sp  trace.37.sp  trace.3.sp   trace.5.sp   trace.8.sp

sos/report/plugins/powerpc.py
mamatha4: sure...will update this with add_service_status() function

sos/report/plugins/scsi.py
mamatha4: ok

sos/report/plugins/powerpc.py
TurboTurtle: `journalctl --no-pager` is the full content of the journal, it just pushes the entire content to stdout without paging - it doesn't truncate the output in any way.

TurboTurtle: Any truncating you're seeing is from the size limiting in place, not the command being run. Journals are limited by default to the latest 100MB. If you truly need the entire journal regardless of size then use `--log-size=0` in your sos command.

In any event, the `logs` plugin already does sufficient journal collection for non-unit-specific collections, and the `powerpc`  plugin would not be the place for such a collection anyways.

mamatha4: ok ..Thanks for the info
I think we have all data collected in following path
$ls sos_commands/logs
journalctl_--disk-usage  journalctl_--no-pager  journalctl_--no-pager_--catalog_--boot  journalctl_--no-pager_--catalog_--boot_-1  ls_-alRh_.var.log

sos/report/plugins/scsi.py
TurboTurtle: It looks like these got dropped in the most recent branch update, was that intentional?

TurboTurtle:
Updated branch looks good, just need to know if dropping the added `lsscsi` collections was intentional? Those were fine, just wanted to have them passed as a list, e.g.

```python
self.add_cmd_output([
    'lsscsi -H',
    'lsscsi -g'
[...]
])
```

mamatha4:
Hi,
Thanks for the update..I was trying to push branch with two commits with following command, but it's always updating with only one commit, that is why lsscsi commit is missing here.. I will try once again to push the branch with two commits

git push -f origin HNV-SCSI-log

mamatha4:
Hi,
Can you please check now and confirm..

TurboTurtle (approved):

mamatha4:
Any updates on integration of patches to upstream?

pmoravec (approved):

pmoravec:
> Any updates on integration of patches to upstream?

We use 2-ACK requirement before merging a PR into upstream. I added the 2nd approve now, so I expect we will merge the PR within a week during a regular bulk-merge activity.

mamatha4:
Sure.. Thanks for the update

Issue #2378, opened 2021-01-19
Add a plugin to collect bcache perf stats


pponnuvel:
Collecting perf stats on bcache [0] [1] is useful to identify performance problems if they're deployed as a caching layer.
Useful stats are:

- `/sys/block/bcache*/bcache/{cache_mode,dirty_data,io_errors,sequential_cutoff,state,writeback_percent,stats_hour/{bypassed,cache_{hit_ratio,hits,misses}},cache/internal/copy_gc_enabled}`

- `/sys/fs/bcache/*/{block_size,bucket_size,cache_available_percent,congested_*_threshold_us}`

- `/sys/fs/bcache/*/{*,stats_*/*,internal/*,bdev*/*,bdev*/stat_*/*}`

I plan to work on adding a new plugin to collect these.

[0] https://wiki.archlinux.org/index.php/bcache
[1] https://bcache.evilpiepirate.org/

hillpd:
This data would definitely be useful. 

Be mindful of lp#1840043 [0]. It is an older issue where stat collection can severely impact system performance. The plug-in may need to check and prevent stat collection for exposed kernels.

[0] https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1840043

bmr-cymru:
Executing the stat collection can be gated by a predicate - we just need the right criteria for when to exclude the collection. The `SoSPredicate` supports various checks that are used to determine whether or not a command should run.

pponnuvel:
@hillpd Thanks, Dan, for the note. `priority_stats` isn't in the list of the stats I planned to collect on bcache (which degrades performance when queried). So we're OK.

pponnuvel:
@bmr-cymru It'd be useful if bcache are deployed as a caching device. I am thinking of `IndependentPlugin` with `files` set one of the sysfs entry something like as: 

https://github.com/sosreport/sos/commit/b6a7a8158111ee55ad6f471d5cd577a39b256e45

(I haven't tested it yet - will make a PR after test).

PR #2384, opened 2021-01-25
[bcache] Add a new plugin for bcache

pponnuvel:
bcache is used as a caching device (typically an SSD) for
HDDs; bcache stats are useful to identify performance problems.

Closes: #2378
Resolves: #2384

Signed-off-by: Ponnuvel Palaniyappan <ponnuvel.palaniyappan@canonical.com>

---

### DIFF
--- a/sos/report/plugins/bcache.py
+++ b/sos/report/plugins/bcache.py
@@ -1,0 +1,56 @@
+# Copyright (C) 2021, Canonical ltd
+# Ponnuvel Palaniyappan <ponnuvel.palaniyappan@canonical.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, IndependentPlugin, SoSPredicate
+
+
+class Bcache(Plugin, IndependentPlugin):
+
+    short_desc = 'Bcache statistics'
+
+    plugin_name = 'bcache'
+    profiles = ('storage', 'hardware')
+    files = ('/sys/fs/bcache',)
+
+    def setup(self):
+
+        # Caution: reading /sys/fs/bcache/*/cache0/priority_stats is known
+        # to degrade performance on old kernels. Needs care if that's ever
+        # considered for inclusion here.
+        # see: https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1840043
+        self.add_forbidden_path([
+            '/sys/fs/bcache/*/*/priority_stats',
+        ])
+
+        self.add_copy_spec([
+            '/sys/block/bcache*/bcache/cache/internal/copy_gc_enabled',
+            '/sys/block/bcache*/bcache/cache_mode',
+            '/sys/block/bcache*/bcache/dirty_data',
+            '/sys/block/bcache*/bcache/io_errors',
+            '/sys/block/bcache*/bcache/sequential_cutoff',
+            '/sys/block/bcache*/bcache/stats_hour/bypassed',
+            '/sys/block/bcache*/bcache/stats_hour/cache_hit_ratio',
+            '/sys/block/bcache*/bcache/stats_hour/cache_hits',
+            '/sys/block/bcache*/bcache/stats_hour/cache_misses',
+            '/sys/block/bcache*/bcache/writeback_percent',
+            '/sys/fs/bcache/*/average_key_size',
+            '/sys/fs/bcache/*/bdev*/*',
+            '/sys/fs/bcache/*/bdev*/stat_*/*',
+            '/sys/fs/bcache/*/block_size',
+            '/sys/fs/bcache/*/bucket_size',
+            '/sys/fs/bcache/*/cache_available_percent',
+            '/sys/fs/bcache/*/congested_*_threshold_us',
+            '/sys/fs/bcache/*/internal/*',
+            '/sys/fs/bcache/*/stats_*/*',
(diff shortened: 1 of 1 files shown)


pponnuvel:
I have tested the changes using a setup that uses 2 x 1:1 mapping (1 bcache device per backing device).

The sos report file (tar.xz) size increase with this setup is ~36KB. The files gathered for bcache plugin are text files, so even if the system has many number of bcache devices, it's not going to massively increase the sos report.

Attaching a sample sos report here:
[sosreport-bcache-test.zip](https://github.com/sosreport/sos/files/5868847/sosreport-bcache-test.zip)

(had to zip a tar.xz file because github doesn't allow tar .xz file ;)

sos/report/plugins/bcache.py
```diff
@@ -0,0 +1,49 @@
+# Copyright (C) 2021 Ponnuvel Palaniyappan <ponnuvel.palaniyappan@canonical.com>
```
TurboTurtle, line 1: One character too long. Could either split over multi-line or use initials, whichever you prefer.

sos/report/plugins/bcache.py
```diff
@@ -0,0 +1,49 @@
...
+
+class Bcache(Plugin, IndependentPlugin):
+
+    short_desc = 'Bcache statistics'
+
+    plugin_name = 'bcache'
+    profiles = ('storage', 'hardware')
+    files = '/sys/fs/bcache'
+
+    # Caution: reading /sys/fs/bcache/*/cache0/priority_stats is known
```
TurboTurtle, line 22: Perhaps we should add this to a forbidden list, as a safeguard beyond the note here? I'm thinking what if this location for one reason or another changes and a `priority_stats` is added in an otherwise captured spot? The forbidden path glob could be fairly wide open, e.g. `/sys/fs/bcache/*/priority_stats` or similar

TurboTurtle (commented):
General ack overall. Just 2 minor notes.

sos/report/plugins/bcache.py
pponnuvel: Yes, forbidden list makes sense. Does glob expansion considered as access/read?

Did quick test with:
```
import glob

for name in glob.glob('/sys/fs/bcache/*/cache0/priority_stats'):
    print name
```
and it does `lstat` and `fstat` on it:

```
...
lstat("/sys/fs/bcache/register_quiet/cache0", 0x7ffdd00ee980) = -1 ENOTDIR (Not a directory)
lstat("/sys/fs/bcache/register/cache0", 0x7ffdd00ee980) = -1 ENOTDIR (Not a directory)
lstat("/sys/fs/bcache/583c9130-fa01-455f-9e96-e0b5c84b2902/cache0", {st_mode=S_IFLNK|0777, st_size=0, ...}) = 0
lstat("/sys/fs/bcache/583c9130-fa01-455f-9e96-e0b5c84b2902/cache0/priority_stats", {st_mode=S_IFREG|0444, st_size=4096, ...}) = 0
fstat(1, {st_mode=S_IFCHR|0620, st_rdev=makedev(136, 4), ...}) = 0
write(1, "/sys/fs/bcache/583c9130-fa01-455"..., 74) = 74
...
```
Just want to make sure the forbidden list doesn't end up doing what we want to avoid here.

TurboTurtle: I wouldn't think the glob would trigger the referenced issue, stat'ing just returns the contents of the file's i-node and not the content of the file directly.

pponnuvel: Thanks. Addressed both comments now.

slashdd:
Would it be best to execute `add_copy_spec()` iff the bcache module is loaded via `SosPredicate' kmods` ?

pponnuvel:
Thanks, @slashdd. Gated via a predicate (I believe having `files` also ensures the same).

sos/report/plugins/bcache.py
```diff
@@ -0,0 +1,54 @@
...
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, IndependentPlugin
```
TurboTurtle, line 12: You'll need to explicitly import `SoSPredicate` here. That's the reason the `flake8` test is failing.

After that, this LGTM.

pponnuvel: Done, Thanks!

pmoravec (approved):
ACK, relying on the reporter's knowledge that the collecting files other than `priority_stats` can't harm a similar way like `priority_stats` would.

sos/report/plugins/bcache.py
```diff
@@ -0,0 +1,56 @@
...
+    short_desc = 'Bcache statistics'
+
+    plugin_name = 'bcache'
+    profiles = ('storage', 'hardware')
+    files = ('/sys/fs/bcache',)
+
+    def setup(self):
+
+        # Caution: reading /sys/fs/bcache/*/cache0/priority_stats is known
+        # to performance degradation on some kernels. Needs care if that's
```
hillpd, line 26: A small nit, recommend changing to:
`degrade performance on older kernels.`

I'm suggesting this change just to be clear that the issue has been fixed in recent kernels.

pponnuvel: Hi Dan,
My understanding is that performance is much improved with that fix, but it still has some impact. I am referring this message in that thread: https://lkml.org/lkml/2019/8/14/636

hillpd: Ah, you are right. There is still a minor performance hit due to locking.

hillpd (approved):

TurboTurtle (approved):

PR #2388, opened 2021-01-28
[rpm] Capture the output of 'rpm --showrc'

jcastill:
This request adds the output of --showrc  to sos
so we can verify the default values of options set
via rpmrc and macros config files.

Resolves: RHBZ#1921496
Resolves: #2388 

---

### DIFF
--- a/sos/report/plugins/rpm.py
+++ b/sos/report/plugins/rpm.py
@@ -65,4 +65,6 @@ class Rpm(Plugin, RedHatPlugin):
                                 suggest_filename='lsof_D_var_lib_rpm')
             self.add_copy_spec("/var/lib/rpm")
 
+        self.add_cmd_output("rpm --showrc")
+
 # vim: set et ts=4 sw=4 :


pmoravec (approved):

sos/report/plugins/rpm.py
```diff
@@ -19,7 +19,9 @@ class Rpm(Plugin, RedHatPlugin):
     option_list = [("rpmq", "queries for package information via rpm -q",
                     "fast", True),
                    ("rpmva", "runs a verify on all packages", "slow", False),
-                   ("rpmdb", "collect /var/lib/rpm", "slow", False)]
+                   ("rpmdb", "collect /var/lib/rpm", "slow", False),
+                   ("rpmshowrc", "shows the options in rpmrc and macros files",
+                    "fast", True)]
```
TurboTurtle, line 24: Does this really need to be gated behind a plugin option? We usually reserve those options for long running items, potentially sensitive collections, or historically ones that could make changes to the system (before the addition of `--allow-system-changes` and the like).

From quick testing this looks like it should be safe for a default collection.

pmoravec: I do agree that the extra option (though enabled by default) is bit redundant - I cant figure out a use case when one would need to prevent collecting the cmdoutput.

jcastill: I chose this approach to make sure that it was consistent with the other options, but you are right, so let me re-send this PR.

pmoravec (approved):

TurboTurtle (approved):

PR #2391, opened 2021-01-29
[md] Add "mdadm -E" command output

akshay196:
The "mdadm -E" print contents of the metadata stored in devices. -E
applies to devices which are components of an array, where -D applies
to whole array.

Signed-off-by: Akshay Gaikwad <akgaikwad001@gmail.com>

---

### DIFF
--- a/sos/report/plugins/md.py
+++ b/sos/report/plugins/md.py
@@ -18,6 +18,8 @@ class Md(Plugin, IndependentPlugin):
 
     def setup(self):
         self.add_cmd_output("mdadm -D /dev/md*")
+        self.add_blockdev_cmd("mdadm -E %(dev)s",
+                              blacklist=['ram.*', 'zram.*'])
         self.add_copy_spec([
             "/proc/mdstat",
             "/etc/mdadm.conf",


akshay196:
What cause "Report functionality test" to fail? 

```
Summary
 
 failures false 	 time 2 	 Size 0.00MiB 	 List plugins 
 failures false 	 time 1 	 Size 0.00MiB 	 List presets 
 failures false 	 time 1 	 Size 0.00MiB 	 List profiles 
 failures false 	 time 4 	 Size 0.00MiB 	 --build -t1 -o host,kernel,filesys,hardware,date,logs 
 failures true 	 time 17 	 Size 2.50MiB 	 -vvv 
 failures false 	 time 17 	 Size 2.72MiB 	 -e opencl -v -k kernel.with-timer,libraries.ldconfigv --no-postproc 
 failures false 	 time 2 	 Size 10.00KiB 	 --no-report --label TEST -o hardware 
 failures false 	 time 14 	 Size 3.07MiB 	 --log-size 0 --no-env-vars -z gzip 
 failures false 	 time 30 	 Size 2.48MiB 	 --mask

Tests Failed: 1
Failures within each test:

 	 test_normal_report: 		 new kernel modules loaded during execution
```

Can someone help me understanding the failure?

bmr-cymru:
This is the text to focus on:

  "new kernel modules loaded during execution"

Possibly the system where the test ran had no MD devices, and the `mdadm` commands run by the plugin caused them to be loaded.

That said, I don't think that running any `mdadm` command on *every device in /dev* is a very good idea. This runs the command on character devices, block devices, and every udev alias present in the directory. This has the potential to generate a lot of IO and possibly other side effects.

To make this effective you would at least need to filter this to just block devices that could possibly be a part of an MD array.

If you look at the `block` plugin and the `cmds` list that is fed to the `Plugin.add_blockdev_cmd()` method that is the normal way of collecting command data for every block device in the system, although I am still not convinced that this is worthwhile for the `mdadm -E` information.

akshay196:
@bmr-cymru  Thanks for the information. I have updated PR to use `add_blockdev_cmd()`.

I know the best approach to collect "mdadm -E" is parse `/proc/mdstat` file for device components. Although the parsing file for component names is too complicated,  See https://github.com/RedHatInsights/insights-core/blob/master/insights/parsers/mdstat.py#L104-L154 

However if any block device is not a valid for "mdadm -E" then it will show below line in stdout:
```
mdadm: No md superblock detected on /dev/xx.
```
I hope this is just OK. What do you think?

TurboTurtle:
Ack to the use of `add_blockdev_cmd()`, and it appears that this has addressed the kernel module loading check - though I'm not entirely comfortable with the thought that mdadm simply targeting non-block devices was the root cause there. I'll need to dig into that a bit more.

TurboTurtle (approved):
I've not been able to reproduce the kernel module loading at all. I'm still not 100% comfortable with running this against every block device, but I also don't have a better alternative.

I don't _think_ it will cause an issue, though, so I'm ok with adding it at this point unless someone can point out a real world consideration as to why this might be a bad idea.

akshay196:
Can we merge this now?

pmoravec (approved):
ACK.

We use 2 approval mode to get PRs merged. Once that happens, it is a matter of few days or a week to merge all such PRs in a bulk.

Thanks for patience.

TurboTurtle:
Yup, now that we have 2 acks I'll get this merged later today or tomorrow. I try to do at least one batch of merges a week.

akshay196:
@TurboTurtle @pmoravec Thanks for reviewing and merging. :)

PR #2392, opened 2021-02-01
[report] add a script to print plugins' overview

pmoravec:
Let's have a script that generates an overview of each plugin's
activity. I.e. for each plugin, print list of items the plugin collects
in add_copy_spec or add_cmd_output and other similar methods.

Resolves: #2392

---

### DIFF
--- a/plugins_overview.py
+++ b/plugins_overview.py
@@ -1,0 +1,73 @@
+# this script generates for each plugin:
+# - list of paths it collects (add_copy_spec)
+# - list of paths it forbits to collect (add_forbidden_path)
+# - list of commands it calls (add_cmd_output)
+#
+# Output of the script: a JSON object with plugins in keys
+#
+# TODO:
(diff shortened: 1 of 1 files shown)


pmoravec:
Few my notes:

- The script performs a simple static analysis, no rocket science. Anyway the outcome is quite usable already.
- I am not sure of **location** of the script inside `sos` source code tree (should it be there, even?). Surely it makes a little sense to provide it in a package.
- Also I know the code quality is not such high as we require else-where. I think it is ok here as this script wont be used "in production" and is "just" an auxiliary static analysis one.

TurboTurtle:
+1 useful for auditing purposes.

As for where, I don't think there would be an issue leaving it in the repo root, since it'd be unpackaged. If we see ourselves adding more scripts like this we could add a 'tools' or similar directory (still in repo root).

TurboTurtle (approved):

Issue #2393, opened 2021-02-02
[cleaner] Automatically create directory for default_mapping


pmoravec:
Having a system without `/etc/sos/cleaner` directory, an attempt to run `sos clean` fails in storing the `default_mapping` file:

    [cleaner] Could not update mapping config file: [Errno 2] No such file or directory: '/etc/sos/cleaner/default_mapping'

The reason is really in missing the directory.

In case this specific error is raised, `cleaner` should create the directory and re-try storing the mapping file.

We can argue if creating directories isnt a packaging work we should handle e.g. in `spec` file, but:
- this directory is redundant for non-cleaner `sos` usage, so creating it ad hoc is less meaningfull
- `sos` does create `/etc/sos` in case it is missing - we should provide consistent behaviour here

PR #2397, opened 2021-02-02
[cleaner] Automatically create directory for sos-cleaner default_mapping

jcastill:
When /etc/sos/cleaner directory is missing, cleaner fails to store
new default_mapping file due to:

[cleaner] Could not update mapping config file:
[Errno 2] No such file or directory: '/etc/sos/cleaner/default_mapping'

This patch adds the code to check the existence of the directory
and create it if necessary.

Resolves: RHBZ#1923937
Closes: #2393
Resolves: #2397 

---

### DIFF
--- a/sos/cleaner/__init__.py
+++ b/sos/cleaner/__init__.py
@@ -405,7 +405,12 @@ third party.
         able to provide the same consistent mapping
         """
         if self.opts.map_file and not self.opts.no_update:
+            cleaner_dir = os.path.dirname(self.opts.map_file)
+            """ Attempt to create the directory /etc/sos/cleaner
+            just in case it didn't exist previously
+            """
             try:
+                os.makedirs(cleaner_dir, exist_ok=True)
                 self.write_map_to_file(_map, self.opts.map_file)
                 self.log_debug("Wrote mapping to %s" % self.opts.map_file)
             except Exception as err:


jcastill:
Some doubts I have about this approach:

Should I add a try/except around os.makedirs as well, or will it be caught in the other one?
Should I add a 'mode', or is the default enough?
Is write_map_for_config() the right place to declare cleaner_dir?

pmoravec:
Maybe worth setting `cleaner_dir = os.path.dirname(self.opts.map_file)` in case one overwrites the default one?

Answering the questions in comment:

- if directory create fails, the `except` block catches it and that is imho legitimate way of handling it
- I think the code change is on the right place. 
- default mode `o700` seems OK to me - don't expose potentially sensitive data to any potential other user on the system

TurboTurtle:
> Maybe worth setting `cleaner_dir = os.path.dirname(self.opts.map_file)` in case one overwrites the default one?
> 
Good idea, I was thinking that if a provided `map_file` doesn't exist, we'd fail before this with reading it so why write it? But perhaps thats the intent? Use a blank mapping and then save it fresh - +1.

> 
> * if directory create fails, the `except` block catches it and that is imho legitimate way of handling it

Right, the try block is fine. I'm saying the `if not os.path.exists(...)` is redundant.

jcastill:
Changes made, and tested it - seems to work well when the directory exist and when it doesn't, without 'os.path.exists(...), specially I guess because of the "exist_ok=True" inside os.makedirs()

TurboTurtle (approved):

Issue #2288, opened 2020-11-01
"effective options now" to log valid variants of cmdline options


pmoravec:
effective options now: --batch --build --log-size 100 --only-plugins host --plugopts apache.log=on -vvv

But:

    sosreport --batch --build --log-size 100 --only-plugins host --plugopts apache.log=on -vvv
    ..
    sosreport: error: unrecognized arguments: --plugopts apache.log=on

Not sure if some other option is affected the same.

bmr-cymru:
Awkward. __Most__ of our long option names map directly to the `dest` value used in `add_argument()` calls, but this is one that doesn't. The bug has existed for quite a long time. We can either bring the `dest` values into sync with the long options or introduce another mapping layer in `to_args()` to fix up the mismatches.

TurboTurtle:
Any opposition to instead add `--plugopts` as a valid long form name, perhaps eventually deprecating the existing `--plugin-option` for it?

pmoravec:
ACK to either/both.

bmr-cymru:
No preference for which way it's resolved, but the intent of logging the message in `--command-line --notation` was that it should be possible to copy & paste from the logs to a new command line - so long as they are consistent and that works I am fine with whichever option names are chosen.

pmoravec:
Same issue is `--no-report` that is printed as `effective options now: --noreport`. Here, `s/noreport/no-report/g` in `./sos/report/__init__.py` does the fix.

Since we print cmdline options based on ther `dest` where we replace `_` by `-` (https://github.com/sosreport/sos/blob/master/sos/options.py#L298), I spot few other options affected the same - listing all of them with my suggestions:

- `--plugin-option` / `plugopts` (no preference)
- `--no-report` / `noreport` (`dest=no_report`)
- `--profile` / `profiles` (`--profiles`, as it can be a list)
- `--skip-commands` / `skip_cmds` (`dest=skip_commands`)
- `--map` / `map_file` (`--map-file` to be more specific in option name)

PR #2398, opened 2021-02-02
[sos] Align plugin options with destination names

TurboTurtle:
This commit alters several option long-form names or destination names
to align those values in a sensible way. This serves to not only remove
some abiguity in option naming in code, but also to make it so that the
"effective options" line logged in every sos execution can be direction
copy-pasted as a working command.

Closes: #2288
Resolves: #2398

---

### DIFF
--- a/man/en/sos-clean.1
+++ b/man/en/sos-clean.1
@@ -4,7 +4,7 @@ sos clean - Obfuscate sensitive data from one or more sosreports
 .SH SYNOPSIS
 .B sos clean TARGET [options]
     [\-\-domains]
-    [\-\-map]
+    [\-\-map-file]
     [\-\-jobs]
     [\-\-no-update]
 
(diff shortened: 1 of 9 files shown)


pmoravec:
I guess the failed test

    AttributeError: 'MockOptions' object has no attribute 'skip_commands'

is due to missing `s/skip_cmds/skip_commands/` on two places:

https://github.com/sosreport/sos/blob/master/tests/plugin_tests.py#L118
https://github.com/sosreport/sos/blob/master/tests/option_tests.py#L20

?

TurboTurtle:
Yup, that seems like it, updated and pushed.

pmoravec (approved):
LGTM.

We might mention the alternate cmdline option(s) at:

https://github.com/sosreport/sos/blob/master/man/en/sos-report.1#L11
https://github.com/sosreport/sos/blob/master/man/en/sos-report.1#L23
https://github.com/sosreport/sos/blob/master/man/en/sos-report.1#L70

but hard to say if it that would improve readability of the manpage.

PR #2403, opened 2021-02-03
[ds] For Directory Server plugin, add the output from the DS's healthcheck tool

mreynolds389:
---

- [X ] If this commit closes an existing issue, is the line `Closes: #ISSUENUMBER` included in an independent line?
- [ X] If this commit resolves an existing pull request, is the line `Resolves: #PRNUMBER` included in an independent line?

### DIFF
--- a/sos/report/plugins/ds.py
+++ b/sos/report/plugins/ds.py
@@ -1,6 +1,7 @@
 # Copyright (C) 2007 Red Hat, Inc., Kent Lamb <klamb@redhat.com>
 # Copyright (C) 2014 Red Hat, Inc., Bryn M. Reeves <bmr@redhat.com>
-
+# Copyright (C) 2021 Red Hat, Inc., Mark Reynolds <mreynolds@redhat.com>
+#
 # This file is part of the sos project: https://github.com/sosreport/sos
 #
 # This copyrighted material is made available to anyone wishing to use,
@@ -48,6 +49,7 @@ class DirectoryServer(Plugin, RedHatPlugin):
                 if d[0:5] == 'slapd':
                     certpath = os.path.join("/etc/dirsrv", d)
                     self.add_cmd_output("certutil -L -d %s" % certpath)
+                    self.add_cmd_output("dsctl %s healthcheck" % d)
         except OSError:
             self._log_warn("could not list /etc/dirsrv")
 


TurboTurtle (changes requested):
Code change looks ok, however please take a look at the [Contribution Guidelines](https://github.com/sosreport/sos/wiki/Contribution-Guidelines) for how the commit message should be formatted.

E.G. `[ds] Add dsctl healthcheck output` as the commit title. You can then reference the RHBZ in the commit message, and please include a line of `Resolves: #2403` so that GH properly closes the PR automatically when we merge (maintainers add their own DCO for merges, which changes the commit id which in turn breaks the merge tagging).

sos/report/plugins/ds.py
```diff
@@ -48,6 +48,10 @@ def setup(self):
                 if d[0:5] == 'slapd':
                     certpath = os.path.join("/etc/dirsrv", d)
                     self.add_cmd_output("certutil -L -d %s" % certpath)
+                    if "ds8" in self.check_version():
```
vashirov, line 51: `ds8` is returned by `check_version()` if package `redhat-ds-base` is installed. It's RHDS 8 (RHEL 5), where the healthcheck is not available. I think we should extend `check_version()` to support recent versions of RHDS to correctly use the healthcheck tool.

mreynolds389:
Changes made.  Instead of checking for a DS version we check for the "healthcheck tool" instead.  If it is present, we use it, otherwise we just skip over it.  Also cleaned up commit message.

sos/report/plugins/ds.py
```diff
@@ -43,11 +45,20 @@ def setup(self):
...
+            dsctl_output = subprocess.check_output(['dsctl', '--help'])
+            if 'healthcheck' in str(dsctl_output):
+                HAVE_HEALTHCHECK = True
+
         try:
             for d in os.listdir("/etc/dirsrv"):
                 if d[0:5] == 'slapd':
                     certpath = os.path.join("/etc/dirsrv", d)
                     self.add_cmd_output("certutil -L -d %s" % certpath)
+                    if HAVE_HEALTHCHECK:
```
pmoravec, line 60: You can use `SoSPredicate` rather than calling `subprocess` methods. See e.g.:

https://github.com/sosreport/sos/blob/master/sos/report/plugins/lvm2.py#L54-L55

(you need to import `SoSPredicate` from `sosreport.plugins`)

mreynolds389: OK, changes made, tests pass, please review...

sos/report/plugins/ds.py
```diff
@@ -43,11 +44,20 @@ def setup(self):
             "/etc/dirsrv/admin-serv/password.conf"
         ])
 
+        # check for dsctl, and if it has the healthcheck option
+        HAVE_HEALTHCHECK = False
+        if os.path.exists('/usr/sbin/dsctl'):
```
TurboTurtle, line 49: This is unnecessary - sos handles command failures automatically, especially for when commands do not exist.

mreynolds389: Ok, round 3, removed these checks, please review...

pmoravec (approved):
If it is the intention to collect `dsctl %s healthcheck` without any condition (other than the `slapd` one), let it be!

Thanks for the patience!

mreynolds389:
> If it is the intention to collect `dsctl %s healthcheck` without any condition (other than the `slapd` one), let it be!

Correct, the complete command to be run would be:    dsctl slapd-INSTANCE healthcheck

> Thanks for the patience!

Not a problem

TurboTurtle (approved):

PR #2406, opened 2021-02-04
[clean] Fix hostnames not being obfuscated when host plugin wasn't run

TurboTurtle:
It was discovered that if a report was generated without the host plugin
being run, then the hostname parser would not obfuscate the localhost
hostname or domain at all. This was due to the fact that while the
default_mapping values were being read into the parser's map's dataset
correctly, they weren't being loaded into the 'intermediary' dicts that
the parser uses to separate short names from domain names.

Fix this by reading the contents of the map's dataset dict (which is
populated by the map file) into those intermediary dicts that the parser
uses.

Resolves: #2406

---

### DIFF
--- a/sos/cleaner/mappings/hostname_map.py
+++ b/sos/cleaner/mappings/hostname_map.py
@@ -40,13 +40,46 @@ class SoSHostnameMap(SoSMap):
     _domains = {}
     hosts = {}
 
-    def __init__(self, opt_domains):
-        super(SoSHostnameMap, self).__init__()
-        self.load_domains_from_options(opt_domains)
+    def load_domains_from_map(self):
+        """Because we use 'intermediary' dicts for host names and domain names
+        in this parser, we need to re-inject entries from the map_file into
+        these dicts and not just the underlying 'dataset' dict
+        """
+        for domain in self.dataset:
+            if len(domain.split('.')) == 1:
+                self.hosts[domain.split('.')[0]] = self.dataset[domain]
+            else:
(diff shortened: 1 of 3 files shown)


TurboTurtle:
Additional testing revealed that loading this way has the potential to overload host obfuscation numbers (e.g. `hostX`). Adding a safeguard against that, which will also apply to domains for consistency. This has the added effect of removing the presence of `obfuscateddomain1.net` alongside `obfuscateddomain1.com` which was previously possible.

sos/cleaner/mappings/hostname_map.py
```diff
@@ -88,6 +119,7 @@ def sanitize_item(self, item):
             # obfuscate the short name
             ob_hostname = self.sanitize_short_name(hostname)
             ob_domain = self.sanitize_domain(domain)
+            self.dataset[item] = ob_domain
```
pmoravec, line 122: `ob_domain` only, or also `ob_hostname` (like the return value)? I need to review it first, I might be wrong..

TurboTurtle: This specifically was for getting subdomains added to the dataset in addition to the original domain sourced during initialization. Short names are/should be handled elsewhere.

In my tests now, I'm not seeing the hostname persist in the archives anywhere, save for local rpm builds since we're currently skipping all rpm files for all parsers (which is a separate issue to this one).

pmoravec (approved):
Ah I understand, that makes sense.

ACK

Issue #2409, opened 2021-02-08
Extend OVN Plugins


dosaboy:
The existing ovn plugins are a bit limited e.g. they don't collect any logs. I'd like to extend them a little to include logs and a bit more debug info i.e. these plugins:

sos/report/plugins/ovn_central.py
sos/report/plugins/ovn_host.py

PR #2410, opened 2021-02-08
[ovn-{central,host}] Include logs

dosaboy:
Closes: #2409

---

- [ X] If this commit closes an existing issue, is the line `Closes: #ISSUENUMBER` included in an independent line?
- [ X] If this commit resolves an existing pull request, is the line `Resolves: #PRNUMBER` included in an independent line?

### DIFF
--- a/sos/report/plugins/ovn_central.py
+++ b/sos/report/plugins/ovn_central.py
@@ -76,6 +76,11 @@ class OVNCentral(Plugin):
             if ovs_rundir:
                 self.add_copy_spec(os.path.join(ovs_rundir, pidfile))
 
+        if self.get_option("all_logs"):
+            self.add_copy_spec("/var/log/ovn/")
+        else:
+            self.add_copy_spec("/var/log/ovn/*.log")
+
         # Some user-friendly versions of DB output
         cmds = [
             'ovn-nbctl show',
@@ -86,6 +91,9 @@ class OVNCentral(Plugin):
             'ovn-sbctl get-ssl',
             'ovn-sbctl get-connection',
             'ovn-nbctl list loadbalancer',
+            'ovn-nbctl list Load_Balancer',
+            'ovn-nbctl list ACL',
+            'ovn-nbctl list Logical_Switch_Port',
         ]
 
         schema_dir = '/usr/share/openvswitch'
--- a/sos/report/plugins/ovn_host.py
+++ b/sos/report/plugins/ovn_host.py
@@ -29,6 +29,12 @@ class OVNHost(Plugin):
     def setup(self):
         if os.environ.get('OVS_RUNDIR'):
             pid_paths.append(os.environ.get('OVS_RUNDIR'))
+
+        if self.get_option("all_logs"):
+            self.add_copy_spec("/var/log/ovn/")
+        else:
+            self.add_copy_spec("/var/log/ovn/*.log")
+
         self.add_copy_spec([os.path.join(pp, pidfile) for pp in pid_paths])
 
         self.add_copy_spec('/etc/sysconfig/ovn-controller')


slashdd (commented):
+1, but if RHEL share the same logs location. I would be in favour to simply move this piece of code in the 'OVN*' specific class, so that it isn't only available Debian/Ubuntu distribution.

slashdd:
Based on the quick test I did, they do:

```
[root@mycentos ovn]# pwd
/var/log/ovn

[root@mycentos ovn]# ls -altr
total 12
drwxr-xr-x 5 root        root        4096 Feb  8 13:06 ..
drwxr-x--- 2 openvswitch openvswitch 4096 Feb  8 13:06 .
-rw-r----- 1 openvswitch openvswitch  427 Feb  8 13:06 ovn-controller.log

[root@mycentos ovn]# more /etc/redhat-release 
CentOS Linux release 8.3.2011
```

bmr-cymru:
>     * 2 for First argument to super() is not enclosing class

The first argument to `super()` is the class you want to retrieve the parent class of - so usually it's the class whose method you are implementing (because you want to call an identically named method in the parent).

For these two it should be:

```python
        super(DebianOVNCentral, self).setup()
```

and:

```python
        super(DebianOVNHost, self).setup()
```

With all that said, @slashdd is right that generally functionality like this belongs in the superclass unless there's some split between the distros that cannot be handled in a single class.

I believe RHEL deploys these logs to different places according to configuration: `/var/log/openvswitch` is one, but I think in containerised scenarios that becomes `/var/log/containers/openvswitch` - I would have to defer to the RH OVN experts for a definitive answer but this is suggesting that the way to handle these logs is to put the collection in the main `OVN*.setup()` method, and have the `RedHatOVN*` and `DebianOVN*` classes set member variables that indicate the location (or set those methods in the subclass `setup()` method if it can't be determined in a class initialiser - that might be the case for Red Hat due to the need to check the type of deployment).

dosaboy:
Thanks both, i wasn't sure if the RHEL paths were the same as Ubuntu. Ill get that fixed up and thanks for correcting my super().

dosaboy:
I've fixed up those bits and added some extra table queries that I think will be handy for debug. I notice that the existing nbctl commands include table "loadbalancer" which I don't think exists (it doesn't on Ubuntu anyway). I've left it in just in case but could perhaps be removed.

sos/report/plugins/ovn_central.py
```diff
@@ -76,6 +76,15 @@ def setup(self):
...
                 self.add_copy_spec(os.path.join(ovs_rundir, pidfile))
 
+        if self.get_option("all_logs"):
+            self.add_copy_spec([
+                "/var/log/ovn/",
+            ])
+        else:
+            self.add_copy_spec([
+                "/var/log/ovn/*.log",
+            ])
```
TurboTurtle, line 86: For singleton collections, `add_copy_spec()` can take a string instead of a list.

sos/report/plugins/ovn_host.py
```diff
@@ -29,6 +29,16 @@ class OVNHost(Plugin):
...
             pid_paths.append(os.environ.get('OVS_RUNDIR'))
+
+        if self.get_option("all_logs"):
+            self.add_copy_spec([
+                "/var/log/ovn/",
+            ])
+        else:
+            self.add_copy_spec([
+                "/var/log/ovn/*.log",
+            ])
```
TurboTurtle, line 40: Same here - single item collection can be passed as a string

sos/report/plugins/ovn_central.py
```diff
@@ -86,6 +95,9 @@ def setup(self):
             'ovn-sbctl get-ssl',
             'ovn-sbctl get-connection',
             'ovn-nbctl list loadbalancer',
+            'ovn-nbctl list Load_Balancer',
```
TurboTurtle, line 98: Do both `loadbalancer` and `Load_Balancer` produce different output? Or is this a versioning change and the former `loadbalancer` is/will be deprecated?

dosaboy: Yeah table name is defined at https://github.com/ovn-org/ovn/blob/44ea2ec88136f83e7eab9790473025b6c95bdcc0/ovn-nb.xml#L1516 as Load_Balancer. If I try "loadbalancer" on Ubuntu I get the following so i suspect that other distros maybe be aliasing the name for ease of use - I see that bhaley added that feature [1] pretty recently and I'm sure he did it for a good reason. Maybe I should move this to the Ubuntu/Debian class.

# ovn-nbctl list loadbalancer
ovn-nbctl: unknown table "loadbalancer"

[1] https://github.com/sosreport/sos/commit/bcc897bc4

I will fixup those add_copy_spec() parts

TurboTurtle (approved):

slashdd (approved):

PR #2411, opened 2021-02-08
[gluster] Add glusterd public keys and status files

jcastill:
This patch helps capture some missing files in the
gluster plugin, i.e.:

Files inside /var/lib/glusterd/glusterfind, like
*.status files, that store the required timestamps,
*.pem.pub files, that store ssh public keys.
We also need to omit the glusterfind_*_secret.pem,
which contains the openssh private key.

Files inside /var/lib/glusterd/.keys, that contains
*.pem.pub,  the ssh public key.

Closes: RHBZ#1925035, RHBZ#1925419
Resolves: #2411 

---

### DIFF
--- a/sos/report/plugins/gluster.py
+++ b/sos/report/plugins/gluster.py
@@ -55,6 +55,9 @@ class Gluster(Plugin, RedHatPlugin):
 
     def setup(self):
         self.add_forbidden_path("/var/lib/glusterd/geo-replication/secret.pem")
+        self.add_forbidden_path(
+            "/var/lib/glusterd/glusterfind/glusterfind_*_secret.pem"
+        )
 
         self.add_cmd_output([
             "gluster peer status",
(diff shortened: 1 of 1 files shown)


sos/report/plugins/gluster.py
```diff
@@ -55,6 +55,8 @@ def postproc(self):
 
     def setup(self):
         self.add_forbidden_path("/var/lib/glusterd/geo-replication/secret.pem")
+        self.add_forbidden_path("/var/lib/glusterd/glusterfind/"
+                                "glusterfind_*_secret.pem")
```
jcastill, line 59: Not really sure about this way of splitting the line, to be honest. Any way to improve it?

TurboTurtle:
```python
        self.add_forbidden_path(
            "/var/lib/glusterd/glusterfind/glusterfind_*_secret.pem"
        )
```

sos/report/plugins/gluster.py
```diff
@@ -72,7 +74,10 @@ def setup(self):
             "/etc/glusterfs",
             "/var/lib/glusterd/",
             # collect nfs-ganesha related configuration
-            "/run/gluster/shared_storage/nfs-ganesha/"
+            "/run/gluster/shared_storage/nfs-ganesha/",
+            # collect status files and public ssh keys
+            "/var/lib/glusterd/.keys/",
```
TurboTurtle, line 79: Are we absolutely, 100% sure, that _only_ public keys are stored here and that no private key _ever_ is stored here by the service?

TurboTurtle (commented):
An internal RH resource has confirmed no private keys exist in the directory in question. So, this looks good to merge with the small stylistic change noted above.

TurboTurtle (approved):

pmoravec (approved):

Issue #2407, opened 2021-02-05
SCLPlugin doesn't work on EL8 and makes plugins using it not auto-load


evgeni:
Ohai,

we (@theforeman) recently noticed that the `foreman` plugin is not automatically enabled on EL8 (CentOS8). A bit of digging revealed that this is due to the fact that the plugin does subclass `SCLPlugin`
https://github.com/sosreport/sos/blob/00a25deaba41cd34a2143b5324d22a7c35098c1c/sos/plugins/foreman.py#L309-L316
And that seems to confuse `sos` on EL8 as there are no SCL there.

Dropping the `SCLPlugin` (and the `add_cmd_output_scl` call) makes `sos` (`sos-3.9.1-6.el8.noarch`) auto-enable the plugin as desired when one of the listed `packages` is installed.

We see the same behavior with the `postgresql` plugin, which also uses `SCLPlugin`.

What's the correct solution here? Fixing `SCLPlugin` not to fail on EL8? Making two `foreman` plugins (well, three, we already have `RedHatForeman` and `DebianForeman`)?

TurboTurtle:
I wonder if this is due to the fact that we have:

```python
class RedHatForeman(Foreman, SCLPlugin, RedHatPlugin):
```

and

```python
class SCLPlugin(RedHatPlugin)
```

This might be causing an MRO exception during instantiation. If you instead remove `RedHatPlugin` and leave `SCLPlugin`, does the plugin load properly on CentOS for you?

evgeni:
Nope, that's not it. The PostgreSQL plugin doesn't do this double-inheritance from `RedHatPlugin`:
https://github.com/sosreport/sos/blob/00a25deaba41cd34a2143b5324d22a7c35098c1c/sos/plugins/postgresql.py#L80

FWIW, this is rather easily reproducible with a CentOS 8 container:

```console
% podman run -ti --rm centos:8 bash
# yum install -y sos postgresql
# sosreport --profile services --list-plugins | grep PostgreSQL
WARNING: unable to set option for disabled or non-existing plugin (boot)

 postgresql           inactive       PostgreSQL RDBMS
```

(the `--profile services` is required, as it's not loaded by default in a container that has no `init`)

Making the plugin definition be

```python
class RedHatPostgreSQL(PostgreSQL, RedHatPlugin)
```

makes it work (after importing `RedHatPlugin`)

TurboTurtle:
[...]

PR #2412, opened 2021-02-09
[SCLPlugin] Fix SCLPlugin enablement triggers

TurboTurtle:
It was found that if a system has `scl` available to it, and has a
non-scl package installed for a plugin that subclasses `SCLPlugin`, but
the scl packages are not installed, then the plugin will not be enabled
even for the non-scl collections.

This was due to a too-restrictive check if the plugin subclasses
`SCLPlugins` that prevented the normal plugin triggers from being
checked.

The future of Software Collections is uncertain to the sos project at
this time, but our current use of it is only 3 plugins; foreman,
postgresql, and redis.

As such, rather than splitting out `SCLPlugin` further into a distinct
separate subclass of a plugin and then having to deal with how to
determine which plugin class should be instantiated if both the scl and
non-scl plugin would otherwise be enabled, make the `SCLPlugin`
enablement conditions not prevent normal plugin enablement, and add
gates to the `_scl` methods provided by `SCLPlugin` to abort any
collection attempt if the specified scl is not detected.

If, at a later date, the project sees further (new) use of the
`SCLPlugin` then we can/will review this approach at that time.

Closes: #2407
Resolves: #2412

---

### DIFF
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -2568,7 +2568,8 @@ class Plugin(object):
                                                    commands,
                                                    services):
                         type(self)._scls_matched.append(scl)
-                return len(type(self)._scls_matched) > 0
+                    if type(self)._scls_matched:
+                        return True
 
             return self._check_plugin_triggers(self.files,
                                                self.packages,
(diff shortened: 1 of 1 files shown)


bmr-cymru (approved):
Seems reasonable as an interim fix until we can make a decision on the future of SCL support.

Issue #2254, opened 2020-10-04
Cleaner does not obfuscate kerberos realms


pmoravec:
Having `/etc/krb5.conf`:

    [libdefaults]
     default_realm = BAR.COM

    [realms]
     BAR.COM = {
      kdc = kerberos.foo.bar.com:88
      admin_server = kerberos.foo.bar.com:749
      default_domain = foo.bar.com
     }
    
    [domain_realm]
     .bar.com = FOO.BAR.COM
     bar.com = FOO.BAR.COM

Then `sosreport -o host,krb5 --batch --build --clean` does not obfuscate realms in capitals.

The cause is case sensitive matching of hostnames primarily fetched from `hostname`. Not sure if it is safe to regexp match case insensitively..?

Further, having `hostname` set to `pmoravec-rhel8.foo.bar.com`, the above command did not obfuscate *anything* (that sounds strange even to me and I suppose some issue with my mocking I did)?

TurboTurtle:
I think matching case insensitively is fine, however I'm not sure I'm comfortable with the replacement being completely lowercase when the source is completely upper.

This is theoretically an easy fix on that front via `isupper()` or `islower()` or something. Remains to be seen what "gotchas" we might face.

pmoravec:
Two other possible approaches:

- have dedicated parser/mapper for krb5 that will match&replace either lowercase or uppercase hostnames
- let an user to obfuscate it via `--keywords` or `--keyword-file` (not ideal user experience)

PR #2416, opened 2021-02-18
[cleaner|hostname] Fix duplication issue, allow for matching all uppercase domains

TurboTurtle:
This 2-commit patchset addresses two issues with the hostname parser:

1. Refines the fix for #2406 to avoid duplicating TLDs over subsequent no-host-plugin executions
2. Handles #2254 by detecting an all-uppercase convention for kerberos realms and obfuscating appropriately.

---


pmoravec (approved):
Nice!

PR #2420, opened 2021-02-23
[rear] Update the rear plugin to collect needed files.

ryan-blakley:
The plugin didn't collect anything under /var/lib/rear,
it also didn't collect anything under /etc/rear/mappings.

Signed-off-by: Ryan Blakley <rblakley@redhat.com>

---

### DIFF
--- a/sos/report/plugins/rear.py
+++ b/sos/report/plugins/rear.py
@@ -20,12 +20,14 @@ class Rear(Plugin, RedHatPlugin):
     def setup(self):
         # don't collect recovery ISOs or tar archives
         self.add_forbidden_path([
-            '/var/log/rear/*.iso',
-            '/var/log/rear/*.tar.gz'
+            '/var/lib/rear/output/*'
         ])
 
         self.add_copy_spec([
             '/etc/rear/*conf',
+            '/etc/rear/mappings/*',
+            '/var/lib/rear/layout/*',
+            '/var/lib/rear/recovery/*',
             '/var/log/rear/*log*'
         ])
 


sos/report/plugins/rear.py
```diff
@@ -20,12 +20,14 @@ class Rear(Plugin, RedHatPlugin):
     def setup(self):
         # don't collect recovery ISOs or tar archives
```
pmoravec, line 21: Good idea to forbid whole output dir (as far as I read man pages and `OUTPUT_URL=file://`). Is the "recovery ISOs or tar archives" still valid description of content of the dir?

sos/report/plugins/rear.py
```diff
@@ -20,12 +20,14 @@ class Rear(Plugin, RedHatPlugin):
...
         self.add_forbidden_path([
-            '/var/log/rear/*.iso',
-            '/var/log/rear/*.tar.gz'
+            '/var/lib/rear/output/*'
         ])
 
         self.add_copy_spec([
             '/etc/rear/*conf',
+            '/etc/rear/mappings/*',
```
pmoravec, line 28: I guess no sensitive data can be present here (just to doublecheck if we shouldnt scrub something but imho not).

sos/report/plugins/rear.py
ryan-blakley: So the iso will pretty much always be in the output directory, but the backup tar file is rarely stored locally since it could fill the filesystem up. So I figured just excluding everything in that directory just in case.

sos/report/plugins/rear.py
ryan-blakley: So the mappings directory is a spot to set a custom ip and routes, so yes there will be sensitive data, but it's needed in support if the files are there.

TurboTurtle: IP addresses and such are fine - they'd be obfuscated by `--clean`. But are there any passwords or such that might be set in `/etc/rear/*conf`?

ryan-blakley: It looks like there is options for specifying passwords in the /etc/rear/local.conf file, I've personally never seen them used. But they're a few options with password in the name, in the example conf file /usr/share/rear/conf/default.conf.

TurboTurtle: Oh, my fault there - read the git diff wrong and thought you were adding the `conf` files as well.

Seems we should be good then.

TurboTurtle (approved):

pmoravec (approved):

Issue #2365, opened 2021-01-12
[RFC] Testing approach and design overhaul


TurboTurtle:
Recently, Red Hat QE has been highlighting several pain points with ongoing testing of `sos` releases, and that eventually led to the idea of overhauling the upstream testing suite and pushing several internal test cases upstream to better catch major issues before releases are sent to downstreams.

It's no secret that current `sos` testing is not in a great place. It definitely has value in its current form, but we can do better and provide better assurances to downstreams that `sos` is ready for production releases.

This RFC will summarize the results of a couple meetings with Red Hat QE and provide them for review by other upstream contributors, as well as provide a point to make other suggestions for improvements.

## Where we are right now

- Integration tests are contained in a monolithic `simple.sh` bash script. This makes things easy in terms of running tests locally, but makes maintenance of extension of the test cases a bit onerous.
- Integration tests are only covering `sos` runs with default package sets and configurations
- Minimal testing coverage of `sos clean`
- No testing coverage of `sos collect` 
- Unit tests are not strictly enforced or even monitored. New features are not required by project policy to add new tests, though it is encouraged. Right now we sit at about ~45% unittest coverage

## Proposal for where we want to go

First and foremost I want to highlight a _major_ proposal coming from these talks. Since sos-4.1 is closing in a month, it is proposed that we make the *primary* focus of sos-4.2 this testing overhaul. Yes, that is an ambitious suggestion given we're a month out from starting that work, however the capacity requirements for QE to test each new release is becoming untenable. By primary focus, I mean that all other new functionality is put on the back burner until the testing situation is improved and merged into master. Bug fixes will of course still be a high priority item for merges. This would kick back `sos help|info` and `sos examine` again.

Now, on to the specific proposal:

We migrate from bash scripts to an actual (python) testing framework. Framework is TBD, and we are open to suggestions. This migration would be done in several stages.

**Stage 1**
Somewhere between a POC and actually accepted testing posture, stage 1 would consist of basic-functionality integration tests. This would practically be a migration of the current tests in `simple.sh` to the chosen framework, in addition to a number of basic checks used by Red Hat QE (for example, does `sos --help` and `sos report --help` work and provide expected output?).

This would define the basis of our testing approach:

- Build the basic function integration tests to have multiple tests attached to each run, but also accept that many tests will need their own individual runs of sos. We'll need to define our setup and teardown procedures here.
- The preference will be for multiple smaller tests rather than larger tests, except where those larger tests are just common sense to be run together.
[...]

PR #2431, opened 2021-03-02
[tests] Start using avocado for test suite

TurboTurtle:
This commit represents the start of an overhaul of the test suite used
by sos. Note that several more commits to follow will be required in
order for the test suite to be considered stable.

The new test suite will use the avocado-framework to build out new
tests.

This first part adopts a new 'stageX' naming scheme for our tests as
follows:

	stage0 -> Unittests
	stage1 -> Basic function tests, no mocking allowed
	stage2 -> Mocked tests for specific scenarios/regressions
	stage3 -> Complex setups for layed products/environments

At the moment, these unittests are not updated for avocado, though most
should still work with `nosetest` directly.

A new set of base classes is defined in tests/sos_tests.py which provide
the foundation for actual tests cases. This approach entails new test
cases subclassing a base class, such as the new `StageOneReportTest`,
and setting the `sos_cmd` class attr to the _options_ for an sos report
run. By default `sos report --batch` will be run, and targeted to the
test job's directory as a tmpdir.

Each sos command will be executed once, and all test_* methods within a
test case that subclasses `StageOneReportTest` will be checked against
the output of that execution. Note that this diverges from avocado's
typical approach where each test_* method is performed against a brand
new instance of the class (thus meaning any setup including our sos
report run would normally be run fresh). However, after speaking with
the avocado devel team, this is still seen as a valid pattern for the
framework.

The current organizational approach is to separate the tests by
component rather than stage. For example. `tests/report_tests/` should
hold any report-centric tests, and the `plugin_tests` directory therein
should be used for plugin-specific tests. As of this commit, there are
basic functionality tests under `tests/report_tests/` and a single
plugin test under `tests/report_tests/plugin_tests/` to act as a POC.

Further, there is a `tests/vendor_tests/` directory for organizing
vendor-specific bug/feature tests that are not covered by the generic
project-wide tests. A POC test from RHBZ1928628 is available with this
commit.

Note that in order for these tests to be run properly _without_
installing the current branch to the local system, you will need to run
the following command:

`PYTHONPATH=tests/ avocado run -t stageone tests/`

---


TurboTurtle:
This is a first draft POC for #2365. Expect numerous changes over time to this PR. It is *not* in a mergable state and is being presented here as an easy reference for initial feedback on the design approach.

Please see the commit message for more details, as well as an explanation of how to run these tests.

TurboTurtle:
Known issues:

1. Running the unittests with avocado breaks them, but them (mostly) pass with nosetests still. The tests that fail with nose are because I have yet to update the filepaths for the test data files used in those unit tests.

2. The idea with `vendor_tests/` is that they would subclass other tests, e.g. `report_tests/plugin_tests/networking.py:NetworkingPluginTest`, and extend bug-specific tests there. However, subclassing that test case doesn't seem to give the expected effect as avocado does not recognize it as an instrumented test (a subclass of avocado.Test). Further, if I additionally subclass `StageOneReportTest` as well, only the tests defined in `StageOneReportTest` get run in addition to those being defined in the new class - in other words, nothing seems to be inherited by subclassing `NetworkingPluginTest`.

pmoravec:
A nit-pick from applying `.patch` : ` tests/ziptest_link → tests/unittests/ziptest_link` now refers to invalid symlink destination (that moved as well).

tests/sos_tests.py
```diff
@@ -0,0 +1,464 @@
...
+import json
+import os
+import pickle
+import re
+
+SOS_TEST_DIR = os.path.dirname(os.path.realpath(__file__))
+SOS_BIN = os.path.realpath(os.path.join(SOS_TEST_DIR, '../bin/sos'))
+
+
+def skipIf(cond, message=None):
```
clebergnu, line 25: Any reason for not using [`avocado.skipIf()`](https://avocado-framework.readthedocs.io/en/86.0/api/test/avocado.html#avocado.skipIf) ?

tests/sos_tests.py
```diff
@@ -0,0 +1,464 @@
...
+        
+
+class BaseSoSReportTest(BaseSoSTest):
+    """This is the class to use for building sos report tests with.
+
+    An instance of this test is expected to set at minimum a ``sos_cmd`` class
+    attribute that represets the options handed to a specific execution of an
+    sos command. This should be anything following ``sos report --batch``.
+
+    :avocado: disable
```
clebergnu, line 143: I don't think this is needed, because there are no `test_` methods defined in this class.

tests/sos_tests.py
```diff
@@ -0,0 +1,464 @@
...
+    """
+
+    sos_cmd = ''
+
+    def test_archive_created(self):
+        """Ensure that the archive tarball was created and has the right owner
+
+        :avocado: tags=stageone
```
clebergnu, line 422: Is it expected that all classes that inherit from this will be `stageone`? If so, it should be possible to move the tags definition to the class' docstring (and avoid repetition).

tests/sos_tests.py
TurboTurtle: This test is defined in the `StageOneReportTest` class that our other tests subclass:

```python
    @skipIf(lambda x: '--no-report' in x.sos_cmd, '--no-report used in command')
    def test_html_reports_created(self):
```

The problem I was running into was that `avocado.skipIf()` was not properly inspecting the child class' `sos_cmd` attribute,  so it was not properly skipping if the `--no-report` option was being used.

If there is another way around this using the avocado provided decorator, I'm all ears.

tests/sos_tests.py
TurboTurtle: Ack.

tests/sos_tests.py
TurboTurtle: Ack. This is likely a hold over from when I was first playing with tags and trying to get these tests to run in anything that subclasses `StageOneReportTest`, but not run `StageOneReportTest` directly. I.E. before I found `:avocado: disable:` for the base class.

tests/sos_tests.py
clebergnu: So no, there was no way to do that with the Avocado decorators.  But, I think it's a valid feature, so I proposed this:

https://github.com/avocado-framework/avocado/pull/4467

If you could take a quick look (or even a formal review), it'd be very much appreciated.

TurboTurtle:
Ok, so after all that noise we have a bit better of a situation to build more tests off of now:

- Tests are running for Ubuntu now, thanks @ddstreet for the `apt update` tip for the prep script on those nodes
- nosetests are now running properly. Still via nose directly and not avocado, but I'm not sure if that matters?
- By using `:avocado: enable` on the vendor tests, those now get enabled. However, they only run the tests they themselves define, and we still don't get any inherited tests. I'm still looking into this and talking with @clebergnu about this

- One strange bit is the obfuscation test for renaming a file with a obfuscated word is failing only on Ubuntu 18.04. Note immediately sure why this in particular would fail, since the other renames (e.g. the archive) is working as expected there.

clebergnu:
> Ok, so after all that noise we have a bit better of a situation to build more tests off of now:
> 
> * Tests are running for Ubuntu now, thanks @ddstreet for the `apt update` tip for the prep script on those nodes
> * nosetests are now running properly. Still via nose directly and not avocado, but I'm not sure if that matters?
> * By using `:avocado: enable` on the vendor tests, those now get enabled. However, they only run the tests they themselves define, and we still don't get any inherited tests. I'm still looking into this and talking with @clebergnu about this

Could you do a quick check and replace `:avocado: enable` with `:avocado: recursive`?  It's still on my queue to reproduce/describe/fix the failures, but this different workaround may help. 

> * One strange bit is the obfuscation test for renaming a file with a obfuscated word is failing only on Ubuntu 18.04. Note immediately sure why this in particular would fail, since the other renames (e.g. the archive) is working as expected there.

TurboTurtle:
> 
> Could you do a quick check and replace `:avocado: enable` with `:avocado: recursive`? It's still on my queue to reproduce/describe/fix the failures, but this different workaround may help.
> 

Gave that a shot, no change in behavior.

clebergnu:
> > Could you do a quick check and replace `:avocado: enable` with `:avocado: recursive`? It's still on my queue to reproduce/describe/fix the failures, but this different workaround may help.
> 
> Gave that a shot, no change in behavior.

OK, thanks.  I'll proceed with a more complete investigation.

pmoravec:
Some comments from a chat with an avocado engineer (something can be applicable, something might be irrelevant):

- maintainability of possible changes in Avocado API & using Avocado LTS: before an upgrade between Avocado versions, https://avocado-framework.readthedocs.io/en/86.0/releases/lts/82_0.html should list the diff. Also we can look in advance to "what next version will change" list like https://avocado-framework.readthedocs.io/en/86.0/releases/lts/next.html .
- if tests will generate lots of data (I expect so), we shall set `TMPDIR` variable to force Avocado to write to the disk and not to the ramdisk
- if various tests (within the same job) are in demand to share some common info, we can use https://trello.com/c/P1JOtVY7/863-per-job-tmpdir-for-direct-use-from-tests - but I think we left this idea already

TurboTurtle:
This commit actually already uses the `tmpdir` approach to write to disk and have tests within the same job use the same sos report.

The changelog notes are however a very welcome reference, as I have been assuming we'd go from LTS -> LTS, but if we need to break from that cadence then having the strong documentation behind the frequent updates alleviates some of the concern.

BryanQuigley:
My quick test had etc/fstab -> etc/obfuscatedword1  correctly on 18.04.  Will dig in more later.

BryanQuigley:
Haven't reproduced the fstab issue, but with 18.04 (hostname localhost, have found) - this is Linode's default. 
 (38/69) tests/report_tests/report_with_mask.py:ReportWithMask.test_localhost_was_obfuscated: FAIL: Content 'host0' does not appear in /var/tmp/avocado__6a6zy0ksos_tests.py.ReportWithMask/sosreport-ReportWithMask/sosreport-localhost-2021-03-24-rmcwfdy/etc/hostname\nlocalhost\n (0.03 s)
 (39/69) tests/report_tests/report_with_mask.py:ReportWithMask.test_ip_address_was_obfuscated: FAIL: Content '127.0.0.1' appears in file /var/tmp/avocado__6a6zy0ksos_tests.py.ReportWithMask/sosreport-ReportWithMask/sosreport-localhost-2021-03-24-rmcwfdy/ip_addr (0.03 s)
 (06/69) tests/report_tests/basic_report_tests.py:NormalSoSReport.test_no_new_kmods_loaded: FAIL: Element counts were not equal:\nFirst has 0, Second has 1:  'devlink' (0.05 s)

And locally rhbz1928628Enabled failed because I don't have ethtool installed.

I really really like the error reporting of this system.

TurboTurtle:
> (38/69) tests/report_tests/report_with_mask.py:ReportWithMask.test_localhost_was_obfuscated: FAIL: Content 'host0' does not appear in /var/tmp/avocado__6a6zy0ksos_tests.py.ReportWithMask/sosreport-ReportWithMask/sosreport-localhost-2021-03-24-rmcwfdy/etc/hostname\nlocalhost\n (0.03 s)

Ack, we're skipping `localhost` as an obfuscation, so that test needs to be a bit more precise.

> (39/69) tests/report_tests/report_with_mask.py:ReportWithMask.test_ip_address_was_obfuscated: FAIL: Content '127.0.0.1' appears in file /var/tmp/avocado__6a6zy0ksos_tests.py.ReportWithMask/sosreport-ReportWithMask/sosreport-localhost-2021-03-24-rmcwfdy/ip_addr (0.03 s)

Same - we skip obfuscating `127.0.0.1`. However, this implies that there is no network connection on the VM by default outside of the loopback address.

> (06/69) tests/report_tests/basic_report_tests.py:NormalSoSReport.test_no_new_kmods_loaded: FAIL: Element counts were not equal:\nFirst has 0, Second has 1: 'devlink' (0.05 s)
> 

I'm going to assume that this is related to the above, as devlink should be loaded when a NIC is active afaik.

> And locally rhbz1928628Enabled failed because I don't have ethtool installed.
> 

Is this standard for Ubuntu distros generally? Or is this a Linode-specific Ubuntu image that trims it out?

I'm finishing up the framework for stage 2 (mocking allowed, e.g. placing files and installing packages for a particular test), so we could potentially move this test to stage 2 if it's the former.

> I really really like the error reporting of this system.

TurboTurtle:
I pushed the framework for stage 2 tests, i.e. `tests/sos_tests.py:StageTwoReportTest`. I don't have many use cases in front of me at the moment for which to begin writing tests, so I'm going to reach out to RH QE for some of the existing regression tests and try to move those over.

In the meantime, I'll work on fleshing out some more stage 1 tests. Review on the stage 2 design would be very welcome.

TurboTurtle:
> My quick test had etc/fstab -> etc/obfuscatedword1 correctly on 18.04. Will dig in more later.

I believe I've found a solution, I just can't explain why this is only affecting 18.04. My solution is this:

```
diff --git a/sos/cleaner/parsers/keyword_parser.py b/sos/cleaner/parsers/keyword_parser.py
index 4315f2a4..21583299 100644
--- a/sos/cleaner/parsers/keyword_parser.py
+++ b/sos/cleaner/parsers/keyword_parser.py
@@ -29,7 +29,10 @@ class SoSKeywordParser(SoSCleanerParser):
         for _keyword in self.mapping.dataset.keys():
             self.user_keywords.append(_keyword)
         if keywords:
-            self.user_keywords.extend(keywords)
+            for keyword in keywords:
+                if keyword not in self.user_keywords:
+                    self.mapping.get(keyword)
+                    self.user_keywords.append(keyword)
         if keyword_file and os.path.exists(keyword_file):
             with open(keyword_file, 'r') as kwf:
                 self.user_keywords.extend(kwf.read().splitlines())
```

E.G. pre-generate a keyword map for each keyword given. This is because `obfuscate_string()` which is what we use to get an obfuscated filename ultimately looks directly in the mapping's `dataset` dict key list. 

**Why is this failing for /etc/fstab?**

Because when we go to rename the file, the string `fstab` has not yet been encountered and is thus not in the dataset. This is why pre-generating the obfuscation map resolves this.

**Why is this only failing on 18.04?**

No idea. The content of fstab between 18.04 and 20.04 as far as the GCE stock/public images is concerned is identical (does not contain the string `fstab`).

This does not fail on Fedora because the file there has the string `# /etc/fstab` directly in the content, which kicks off the initial mapping.


In any event, I think pre-generating user-added keywords is the right move for precisely this reason for non-standard files. A user could reasonably expect to use the keyword functionality to scrub filenames rather than actual content.

BryanQuigley:
Ethtool can be assumed in the basic tests -it's in ubuntu-server metapackage - so should be in every cloud image.  (It's not in desktop which is what i was testing there)

TurboTurtle:
I've pushed a small change to `assertFileCollected()`, that will first make sure the file actually exists on the test system, as this mimics the behavior of `add_copy_spec()`. This way we have a bit more flexibility in writing tests across distributions.

clebergnu:
@TurboTurtle [this PR](https://github.com/avocado-framework/avocado/pull/4490) may fix the issue you're experiencing.  I'll try to setup an `sos` development environment to confirm it.  About that, I noticed that you have pre-defined VM images on GCE where the tests are run.  Is there any special sauce on those?

TurboTurtle:
Yes and no. They're built from the stock Fedora server net install ISOs, but have the `google-compute-engine-tools` and `google-compute-engine-oslogin` packages added. Additionally these services need to be enabled: `google-accounts-daemon google-clock-skew-daemon google-instance-setup google-network-daemon google-shutdown-scripts google-startup-scripts` otherwise the VMs hang indefinitely on boot.

I seem to remember reading somewhere (literally a week after I worked out what was needed for the images to boot properly on GCE) that the Fedora project was pushing official images but I don't actually see any listed on GCE now...

TurboTurtle:
> I wonder how much we could inline into the report plugins themselves? Much of the info might be helpful as a base to sos examine as well. That's the only part I'm worried about scaling.
>

The problem I ran into for that is the fact that we have distro-dependent versions on plugins, and while we have the ability to install distro-specific packages within the test suite, I was not able to define a way to make distro-specific tests.
 
> Could a test be auto-generated for some plugins like gdm? In many cases the StageOne test will just be a mirror image of the plugin.
> 
> * We know two packages we could try installing to get it.

Installing packages would make it a StageTwo test.

> * We know some expected files. (We could start splitting up file collections to always expected, and optional - this would be something we could build on for sos examine.)
>

I like the idea on the surface, but I think we'd run into similar cross-distro problems. This is partly why attempted collections of non-existent files is not considered an error already - and if we say a missing "expected" file is still not an error, then I'm not sure what we gain. I'm definitely keen on hearing how this might be fleshed out further though.
 

> Could the test functions be added to the report plugins directly? Again this might allow for easier reuse of other functions for sos examine - if we need a function to parse something, might as well use it for both testing and examine.
> 

Only if we added the test classes (as in `StageOneReportTest`) to the plugins directly. Which I think would muddy the waters a bit. We'd also then need to re-design several aspects of test and plugin enablement (e.g. package presence/installation) to not interfere with each other.

> The rest of the structure makes sense to me so far.

BryanQuigley:
>I was not able to define a way to make distro-specific tests.
Is the plan for the RH specific tests to just be defined in vendor_tests/ and then not included upstream?

>and if we say a missing "expected" file is still not an error, then I'm not sure what we gain. I'm definitely keen on hearing how this might be fleshed out further though.

I think it should definitely be a user facing warning and could be highlighted in the report./.  The first version of a sos compare tool would very likely just compare if files exist - for many plugins that might be enough.

I see the current path eventually looking like:
tests/report_tests/plugin_tests/*allplugins*
sos/report/plugins/*allplugins*
sos/examine/plugins/*allplugins*
tests/examine/plugin_tests/*allplugins*

And for some plugins like logs that does make sense (but logs are really really special!).  But for many plugins they are more like config definitions then python code.  I'd expect many of their tests or "examine" functions to be generic as well - like a generic conf file checker, etc.

TurboTurtle:
> Is the plan for the RH specific tests to just be defined in vendor_tests/ and then not included upstream?

Not exactly. The plan is that under vendor_tests/ we could have tests for specific Bugzilla or LaunchPad bug reports and that they would be included in the upstream tests. The motivation behind this is that now upstream enjoys a greater contribution from our downstreams in the form of better regression testing.

However, this would _also_ be a defacto spot for downstreams to place their own test cases for their own internal testing that may not function in upstream. For example, right now upstream is limited to x86_64 since that's the only arch that GCE provides, however downstreams may have internal infrastructure to test on other arches, and arch-specific tests could be placed here.


> And for some plugins like logs that does make sense (but logs are really really special!). But for many plugins they are more like config definitions then python code. I'd expect many of their tests or "examine" functions to be generic as well - like a generic conf file checker, etc.

It's definitely going to be a balancing act for sure between code readability/maintainability and what is effectively code reuse. While I know QE teams would love to see a 1:1 mapping between `plugins/` and `plugin_tests/`, I don't think we really need to go that far. If we're talking bare "did add_copy_spec() really collect XYZ file", then there's not much use to repeating that test a hundred times for simple plugins that just define static files to collect.

What I see us doing more of is "did this logic that extrapolates multiple commands across multiple configs/devices/etc work correctly?", or "did the postproc() method for this plugin really sanitize the password here?", etc...

As for `sos examine`, I think we really need to flesh out what that is going to look like before we get into the weeds about how and where those checks are defined.

I certainly see plugins being expanded to include any `examine` methods as a possibility, perhaps even something very much desired, since those would both be first-class sos components/functionality. But for the test cases I don't think we gain much by extending plugins to additionally subclass the test classes in terms of maintainability. I also don't think we'd gain anything by shipping the test suite along with the tool by default (whereas it can also be added in in downstream builds if desired or separated out into a sub-package).

BryanQuigley:
>If we're talking bare "did add_copy_spec() really collect XYZ file", then there's not much use to repeating that test a hundred times for simple plugins that just define static files to collect.

That's what I thought was going to happen - that we would aspire to write a test for every plugin - so not doing that removes my concerns.  (I actually do see some usefulness to ensure that all files copied previously on RHEL8 are still copied but I get that's harder to do)

Agreed, it's better to let tests not need to be included in downstream work.

TurboTurtle:
Ubuntu tests are failing due to the `psacct` plugin. The distro-specific plugin is overriding the plugin name:

```python

class DebianPsacct(Psacct, DebianPlugin, UbuntuPlugin):

    plugin_name = "acct"
    packages = ["acct"]
```

Which I'm not sure is something we ever intended to support? Theoretically this _should_ be fine at first glance, but our plugin discovery is highlighting `acct` as in invalid plugin name (despite being properly displayed as the override in `--list-plugins` output).

@slashdd @BryanQuigley any objections to just removing the `plugin_name` override, or would you prefer we fix whatever is not working for that kind of override with plugin enablement? Taking a quick look I don't see any other plugins doing this.

BryanQuigley:
I'm +1 on not supporting name changes like that.

TurboTurtle:
\o/ another kmod corner case discovered as part of this framework. Added a predicate to `libreswan`. I have a couple of remaining test cases I want to write and then I'll transition this PR out of draft.

TurboTurtle:
Final noise for the day: I've added helper decorators to make it easy to specify if specific `test_*` methods within a test class should run only on a certain family of distributions.

TurboTurtle:
I've moved this out from draft, as I believe this is now in a merge-able state.

The current test suite covers all aspects previously covered by `simple.sh`, and expands those tests in many areas. Additionally, the set of tests that get run against every test case (e.g. the no new kmods test) gives us extra confidence beyond the single test iteration we had before.

Mocking functionality is present in `StageTwoReportTest`, we can define desitro-specific tests if need be, and the structure to allow downstream vendors to onboard specific bug tests I think allows for a great contribution vector back into upstream from those distributions.

I think from here out, we can build new plugin tests as needed (especially as new plugins get added). `sos clean` functionality can be almost entirely handled through the existing report-centric test classes - the only thing we wouldn't be able to test directly at the moment is generating an archive and then feeding that directly to `sos clean`. That's something that could be built out in the future if we determine that is specifically needed. As for `sos collect`, we're blocked until/if Cirrus allows for multiple instance creation for a single task. I reached out to Cirrus support in the past and the answer was that that kind of functionality was not on their road map, but I can touch base again and see if anything has changed.

BryanQuigley:
I had to install python3-systemd to get the log tests to work - which Ubuntu cloud images come with - still worth documenting somewhere as the error was not obvious.

I also noted that "sudo PYTHONPATH=tests/ avocado run  tests/" runs all the tests  - although many of the nosetests don't work that way right now - but I understand that's not part of this PR - but a cool future possibility.

TurboTurtle:
> I had to install python3-systemd to get the log tests to work - which Ubuntu cloud images come with - still worth documenting somewhere as the error was not obvious.
> 

Good catch. I assumed the module would be available by default but it could definitely be a change.

> I also noted that "sudo PYTHONPATH=tests/ avocado run tests/" runs all the tests - although many of the nosetests don't work that way right now - but I understand that's not part of this PR - but a cool future possibility.

Yeah, that's another part of the reasoning behind the tags. I'm not sure why the nosetests perform differently but for now I think staying with nose for the unittests is sufficient.

TurboTurtle:
Updated to include `python3-systemd` as a required package for the stage 2 logs plugin test.

BryanQuigley (approved):
I didn't mean to make it a review.. but this all LGTM.  I have some ideas for more tests, but figure it makes sense to just use this as is with the next group of PRs.

.cirrus.yml
```diff
@@ -94,8 +94,11 @@ report_function_task:
...
     # Ubuntu has sos installed by default, so remove that to avoid any conflicts
-    setup_script: |
+    remove_sos_script: |
         if [ $(command -v apt) ] && [ $(command -v sosreport) ]; then
             apt -y purge sosreport
+            apt update
+            apt -y install python3-pip
         fi
-    main_script: ./tests/simple.sh
+    setup_script: pip3 install avocado-framework==84.0
```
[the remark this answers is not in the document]
TurboTurtle: 84 is the current avocado LTS version.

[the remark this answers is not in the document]
BryanQuigley: It seems relatively stable to me - I guess I'd prefer to do their "regular" versions to get features/fixes faster.  We can always stop on the next LTS.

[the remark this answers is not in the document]
clebergnu: > It seems relatively stable to me - I guess I'd prefer to do their "regular" versions to get features/fixes faster. We can always stop on the next LTS.

Avocado tries very hard to follow a deprecate, document, and then exclude/change/replace features, even for non LTS releases.  Before every release, we test how the latest Avocado would behave if enabled in some of the projects that depend on it.  For example this is checking how the latest cut before Avocado 87.0 would behave on QEMU:

https://gitlab.com/cleber.gnu/qemu/-/pipelines/283543999

Also please notice that this specific release has at least a new feature and a bugfix inspired by sos:

* https://github.com/avocado-framework/avocado/pull/4467
* https://github.com/avocado-framework/avocado/pull/4490

So I tend to agree that you would benefit from using the non-LTS version to pickup those (and future enhancements).  It may be that after a few development cycles for the tests, you stabilize and then end up settling on a future LTS version.

Anyway, just my 2 cents.

TurboTurtle:
After hemming and hawing about this, I've updated the patchset to not version lock avocado on the test nodes - i.e. pip should install the latest version available. While I'd like to keep as stable as possible of a testing framework, the fact is that even the sprint releases of avocado are well tested and dependable.

I think if we set a soft goal that we stabilize on the _next_ LTS, that should be sufficient in allowing us to have a stable platform, the ability to build out more tests and use cases, and provide  feedback to the avocado project if there are further requests from us that could make it into the sprint releases before the next LTS cut.

PR #2434, opened 2021-03-05
[openstack_designate] Add current pool configuration

michaeltchapman:
Use designate-manage to grab the current pools from designate, in
case they differ from the pools stored in pools.yaml

Signed-off-by: Michael Chapman <woppin@gmail.com>

---

### DIFF
--- a/sos/report/plugins/openstack_designate.py
+++ b/sos/report/plugins/openstack_designate.py
@@ -19,6 +19,16 @@ class OpenStackDesignate(Plugin):
     var_puppet_gen = "/var/lib/config-data/puppet-generated/designate"
 
     def setup(self):
+        # collect current pool config
+        pools_cmd = self.fmt_container_cmd(
+            self.get_container_by_name(".*designate_central"),
+            "designate-manage pool generate_file --file /dev/stdout")
+
+        self.add_cmd_output(
+            pools_cmd,
+            suggest_filename="openstack_designate_current_pools.yaml"
+        )
+
         # configs
         self.add_copy_spec([
             "/etc/designate/*",


sos/report/plugins/openstack_designate.py
```diff
@@ -19,6 +19,25 @@ class OpenStackDesignate(Plugin):
...
+                                                      create_pools_cmd)
+            collect_pools_cmd = self.fmt_container_cmd("designate_central",
+                                                       collect_pools_cmd)
+
+        self.exec_cmd(create_pools_cmd + self.pools_path)
+
+        self.add_cmd_output(
+            collect_pools_cmd + self.pools_path,
+            suggest_filename="openstack_designate_current_pools.yaml"
+        )
```
pmoravec, line 39: If I understood this right, you generate a file here that is cat-ed (and this is collected). Can't that be done directly like https://github.com/sosreport/sos/blob/master/sos/report/plugins/subscription_manager.py#L40-L42 ? I.e. something like:

    ..
    self.add_cmd_output(create_pools_cmd + os.path.join(self.get_cmd_output_path(), openstack_designate_current_pools.yaml),
                        suggest_filename="openstack_designate_current_pools.yaml"
    )

?

michaeltchapman: Yep you understood correctly, I had a quick look for a simpler way but didn't find anything. I'll give your suggestion a go, thanks.

michaeltchapman: I think that method will only work in the non-containerized case since the command within the container won't usually have access to the output directory.

TurboTurtle: This might be a good time for us to explore adding a way to leverage a `ContainerRuntime` to copy files out of a container.

michaeltchapman: @TurboTurtle maybe an extra param for copy spec with the container name? I'm not sure I'm in a position to implement but I can try.

michaeltchapman: I filed https://github.com/sosreport/sos/issues/2439 to track a new feature for copying files from a container, though it might not be worth the effort. It's only needed when the desired file isn't within a mounted directory, which I guess isn't that common since it's not come up before, and the workaround of catting the file is probably good enough for non binary cases.

sos/report/plugins/openstack_designate.py
```diff
@@ -19,6 +19,17 @@ class OpenStackDesignate(Plugin):
     var_puppet_gen = "/var/lib/config-data/puppet-generated/designate"
 
     def setup(self):
+        # collect current pool config
+        pools_cmd = "designate-manage pool generate_file --file /dev/stdout"
+
+        if self.container_exists(".*designate_central"):
+            pools_cmd = self.fmt_container_cmd("designate_central", pools_cmd)
```
TurboTurtle, line 26: Has this been tested on a real OSP environment? Is the container name really, always, just `designate_central`? If so we can drop the `.*` in the `container_exists()` call - otherwise we'll want to do something like:

```python
pools_cmd = self.fmt_container_cmd(self.get_container_by_name('.*designate_central'), pools_cmd)
```

Also, `fmt_container_cmd()` already does a `container_exists()` check internally, so we can drop line 25. If the container doesn't exist, it just returns the same `cmd` that you hand to it.

TurboTurtle:
Pending the above, this looks good to me.

sos/report/plugins/openstack_designate.py
michaeltchapman: Fixed, thanks.

TurboTurtle: Sorry, let me clarify a bit:

`self.container_exists()` can take a regex and simply checks if a given string matches _any_ container's name.
`self.fmt_container_cmd()` does _not_ take a regex, but must take an explicit container name. It does a check to make sure a container with that exact name actually exists before doing any formatting of the command string.
`self.get_container_by_name()` takes a regex and returns the first matching container.

The key here is that if the container name is not truly static, e.g. `designate_central`, but does have some predictable element, e.g. `qwerty-designate_central`, then you'd use `get_container_by_name()` to retrieve the actual name of the container to use with `fmt_container_cmd()`.

michaeltchapman: My bad, read your previous reply too quickly. Should be fixed now.

TurboTurtle (approved):

PR #2435, opened 2021-03-08
[profiles] Add nvme and devices plugins to storage profile

jcastill:
This patch adds nvme and devices to the list of plugins
in the storage profile.

Resolves: #2435 

---

### DIFF
--- a/sos/report/plugins/devices.py
+++ b/sos/report/plugins/devices.py
@@ -14,7 +14,7 @@ class Devices(Plugin, IndependentPlugin):
     short_desc = 'devices specific commands'
 
     plugin_name = 'devices'
-    profiles = ('system', 'hardware', 'boot')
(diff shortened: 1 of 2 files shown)


jcastill:
I'm not so convinced about 'devices', because it not only captures the storage devices in udevadm but a lot of other things as well. But thinking about udev, we are not capturing the actual rules, i.e. what we have in:

            "/etc/udev/udev.conf",
            "/lib/udev/rules.d",
            "/etc/udev/rules.d/*"

And this may be very useful as well. Thoughts on this?

TurboTurtle (approved):

TurboTurtle:
I think including `devices` in the `storage` profile is reasonable - logically it tracks even if the actual collections within `devices` somewhat goes beyond `storage`.

As far as capturing rules, I don't have any objections immediately come to mind.

BryanQuigley (approved):

PR #2443, opened 2021-03-12
[openctack_octavia] Gather rsyslog settings for amphora log offloading

kajinamit:
Recently TripleO added support for Octavia amphora log offloading.
When this feature is enabled an additional rsyslog container is started
to receive logs from amphora instances.
This change ensures that settings for the rsyslog container is also
captured in sosreport.

---

- [ X ] If this commit closes an existing issue, is the line `Closes: #ISSUENUMBER` included in an independent line?
- [ X ] If this commit resolves an existing pull request, is the line `Resolves: #PRNUMBER` included in an independent line?

### DIFF
--- a/sos/report/plugins/openstack_octavia.py
+++ b/sos/report/plugins/openstack_octavia.py
@@ -42,6 +42,7 @@ class OpenStackOctavia(Plugin):
             "/var/lib/octavia",
             self.var_config_data + "/octavia/etc/octavia",
             self.var_puppet_gen + "/etc/octavia",
+            self.var_puppet_gen + "/etc/rsyslog.d",
             self.var_puppet_gen + "/etc/my.cnf.d/tripleo.cnf",
         ])
 


sos/report/plugins/openstack_octavia.py
```diff
@@ -42,6 +42,7 @@ def setup(self):
             "/var/lib/octavia",
             self.var_config_data + "/octavia/etc/octavia",
             self.var_puppet_gen + "/etc/octavia",
+            self.var_puppet_gen + "/etc/octavia/rsyslog.d",
```
pmoravec, line 45: Isn't this directory already collected by the previous line? sos report does collect whole path until sizelimit - or was the limit hit, in your case?

kajinamit: Sorry the path was wrong. I updated the patch to point the correct path.

pmoravec (approved):
Ah this makes sense to me, then :)

TurboTurtle (approved):

PR #2445, opened 2021-03-12
[sssd] Add individual SSSD log files

thalman:
The issue is that SSSD creates individual log files for its
components. To be able to track the issue we need all of them.

With one wildcard copy set we usually get just one truncated
log file and this is not very useful for solving issues. We
need to track the request accross logs to understand the
problem. Also log file names are specific for paricular
configuration.

With this patch we list log files under /var/log/sssd and
we add them one by one.

Signed-off-by: Tomas Halman <thalman@redhat.com>

---

### DIFF
--- a/sos/report/plugins/sssd.py
+++ b/sos/report/plugins/sssd.py
@@ -10,6 +10,7 @@
 
 from sos.report.plugins import (Plugin, RedHatPlugin, DebianPlugin,
                                 UbuntuPlugin, SoSPredicate)
+from glob import glob
 
 
 class Sssd(Plugin):
@@ -23,11 +24,13 @@ class Sssd(Plugin):
     def setup(self):
         self.add_copy_spec([
             "/etc/sssd/sssd.conf",
-            "/var/log/sssd/*",
             "/var/lib/sss/pubconf/krb5.include.d/*",
             # SSSD 1.14
             "/etc/sssd/conf.d/*.conf"
         ])
+
+        # add individual log files
+        self.add_copy_spec(glob("/var/log/sssd/*log*"))
 
         # call sssctl commands only when sssd service is running,
         # otherwise the command timeouts


sos/report/plugins/sssd.py
```diff
@@ -23,12 +24,16 @@ class Sssd(Plugin):
...
-            "/var/log/sssd/*",
             "/var/lib/sss/pubconf/krb5.include.d/*",
             # SSSD 1.14
             "/etc/sssd/conf.d/*.conf"
         ])
 
+        # add individual log files
+        logdir = "/var/log/sssd/"
+        for logfile in listdir(logdir):
+            self.add_copy_spec([ logdir + logfile ])
```
alexey-tikhonov, line 35: Probably skip rotated "*-<date>" logs?

thalman: current solution pack them (if they fit in the limit or --all-logs is used). I thought of it but I choose to pack them and not change the behavior.

pmoravec:
I dont like plugins have to do the work that plugin's API should do :) so I am thinking about other possibilities to prevent `listdir`:

One trivial: `self.add_copy_spec("/var/log/sssd/*", sizelimit=0)` - this will collect all the files regardless of size. Which might not be desired, if a logfile is really huge. Otherwise, elegant trivial resolution.

Another: do we know the list of SSSD components (and is it quite stable)? Then we can collect `/var/log/sssd/foo, /var/log/sssd/bar` etc.

Or maybe yet having there `glob.glob("/var/log/sssd/*log*")`?

sos/report/plugins/sssd.py
```diff
@@ -23,12 +24,16 @@ class Sssd(Plugin):
...
-            "/var/log/sssd/*",
             "/var/lib/sss/pubconf/krb5.include.d/*",
             # SSSD 1.14
             "/etc/sssd/conf.d/*.conf"
         ])
 
+        # add individual log files
+        logdir = "/var/log/sssd/"
+        for logfile in listdir(logdir):
+            self.add_copy_spec([logdir + logfile])
```
pmoravec, line 35: While the concatenation of strings with path+filename should work, it is safer to have there `os.path.join(logdir,logfile)`. Anyway, we might end up in a different approach that does not need this at all.

thalman:
> I dont like plugins have to do the work that plugin's API should do :) so I am thinking about other possibilities to prevent `listdir`:
> 
> One trivial: `self.add_copy_spec("/var/log/sssd/*", sizelimit=0)` - this will collect all the files regardless of size. Which might not be desired, if a logfile is really huge. Otherwise, elegant trivial resolution.
> 
> Another: do we know the list of SSSD components (and is it quite stable)? Then we can collect `/var/log/sssd/foo, /var/log/sssd/bar` etc.
Unfortunately not. Some of them are known ahead like "sssd.log" but others can contain the "domain" name and this depends on the deployment and sssd configuration. Parsing the config to get exact log names would be too complicated for sosreport purposes.

> 
> Or maybe yet having there `glob.glob("/var/log/sssd/*log*")`?
glob looks good to me. I was not aware of glob function. That would be good solution to our problem. I will update the PR.

Thanks Tom

tscherf:
I'm not sure we really want to collect all the logs, regardless of size, in *all* cases when a sosreport is created. This can increase the size of the sosreport enormously and it might not be required to have all the logs all the time. E.g. in cases where debug logs were provided already and customers are asked to provide new sosreports at a later point in time but independent of the sssd issue that was already analysed (and resolved) earlier. 

I see the benefit of having all the logs available right from the beginning, but I also see some risk to collect all logs all the time unconditionally. We already have sos '--all-logs' option and it's also common that engineers ask customers to create an archive from /var/log/sssd/ which will then also provide all the logs.

alexey-tikhonov:
What I would care about: to get all SSSD logs with default log level in initial sos-report. This should be low size.

thalman:
> I'm not sure we really want to collect all the logs, regardless of size, in _all_ cases when a sosreport is created. This can increase the size of the sosreport enormously and it might not be required to have all the logs all the time. E.g. in cases where debug logs were provided already and customers are asked to provide new sosreports at a later point in time but independent of the sssd issue that was already analysed (and resolved) earlier.
> 
> I see the benefit of having all the logs available right from the beginning, but I also see some risk to collect all logs all the time unconditionally. We already have sos '--all-logs' option and it's also common that engineers ask customers to create an archive from /var/log/sssd/ which will then also provide all the logs.

I agree with you @tscherf, This is why I did not implement the "sizelimit = 0" suggestion. Provided patch still uses the limit. But this way limit is applied to log-files one-by-one. So in case of huge logs we will get tailed logs, but at least we will have tail from each. So we can hope that if customer creates sosreport shortly after an incident, we may have all necessary information.

I also think that this approach will not increase the sosreport size significantly. We can eventually exclude logrotated files if the size is an issue here (simply using *log instead of *log*.

HTH
Tom

tscherf:
> > I'm not sure we really want to collect all the logs, regardless of size, in _all_ cases when a sosreport is created. This can increase the size of the sosreport enormously and it might not be required to have all the logs all the time. E.g. in cases where debug logs were provided already and customers are asked to provide new sosreports at a later point in time but independent of the sssd issue that was already analysed (and resolved) earlier.
> > I see the benefit of having all the logs available right from the beginning, but I also see some risk to collect all logs all the time unconditionally. We already have sos '--all-logs' option and it's also common that engineers ask customers to create an archive from /var/log/sssd/ which will then also provide all the logs.
> 
> I agree with you @tscherf, This is why I did not implement the "sizelimit = 0" suggestion. Provided patch still uses the limit. But this way limit is applied to log-files one-by-one. So in case of huge logs we will get tailed logs, but at least we will have tail from each. So we can hope that if customer creates sosreport shortly after an incident, we may have all necessary information.

Ok, when the size limit is still taken into account than I'm ok with this change.
 
> I also think that this approach will not increase the sosreport size significantly. We can eventually exclude logrotated files if the size is an issue here (simply using *log instead of _log_.

That probably makes sense because usually customers are asked to reproduce an issue and then create the sosreport right after. For corner cases we can still explicitly request a full archive from /var/log/sssd/.

pmoravec:
With this approach (`add_copy_spec(glob(/var/log/sssd/*log*))`), `sos` will collect up to 25MB of each individual file(*) there. If a file has >25MB, then 25MB tail of it will be collected. I dont know typical content of the logdir, but based on the discussion above, it sounds me like a good compromise/approach.

(*) If there will be a *directory* `/var/log/sssd/foologbar`, then 25MB of newest files from it *altogether* will be collected as well.

pmoravec (approved):

PR #2452, opened 2021-03-17
[cups] Add cups-browsed from journal logs

jcastill:
Gather logs from the service cups-browsed sent
to the journal.

Resolves: RHBZ#1939963
Closes: #2452 

---

### DIFF
--- a/sos/report/plugins/cups.py
+++ b/sos/report/plugins/cups.py
@@ -40,5 +40,6 @@ class Cups(Plugin, IndependentPlugin):
         ])
 
         self.add_journal(units="cups")
+        self.add_journal(units="cups-browsed")
 
 # vim: set et ts=4 sw=4 :


sos/report/plugins/cups.py
```diff
@@ -36,7 +36,8 @@ def setup(self):
         self.add_cmd_output([
             "lpstat -t",
             "lpstat -s",
-            "lpstat -d"
+            "lpstat -d",
+            "journalctl -u cups-browsed"
```
pmoravec, line 40: Why not `self.add_journal(units="cups-browsed")`?

(or optionally enhance the `self.add_journal(units="cups")` to `self.add_journal(units=["cups", "cups-browsed"])`, but that will generate *one* file with `journalctl -u cups -u cups-browsed` output altogether. I dont know if it is better to split the logs or merge..).

jcastill: Actually I missed that, so I'm sending the patch again. Tested as well, and it works.

TurboTurtle (approved):

pmoravec (approved):

PR #2453, opened 2021-03-18
[WIP] [cups] Add gathering job logs into separate files

zdohnal:
All CUPS logs are sent to journal in Fedora/RHEL, so job related logs
are hidden within other logs. The job specific logs can be obtained by
`journalctl -u cups JID=N`, where N is a job id.
This feature introduces gathering 10 latest job logs (to prevent big
sosreports on print servers) one by one via `journalctl` command. I
enhanced `add_journal()` arguments with `otherargs`, which is a list of
commands which can be added to `journalctl` command. It contains a list
of JID strings for CUPS ['JID=1', 'JID=2', 'JID=3'].

Closes: RHBZ#1940310
Signed-off-by: Zdenek Dohnal <zdohnal@redhat.com>

---

### DIFF
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -2332,7 +2332,7 @@ class Plugin(object):
     def add_journal(self, units=None, boot=None, since=None, until=None,
                     lines=None, allfields=False, output=None,
                     timeout=cmd_timeout, identifier=None, catalog=None,
-                    sizelimit=None, pred=None, tags=[]):
+                    sizelimit=None, otherargs=None, pred=None, tags=[]):
         """Collect journald logs from one of more units.
 
         :param units:   Which journald units to collect
@@ -2420,6 +2420,10 @@ class Plugin(object):
 
         if output:
             journal_cmd += output_opt % output
+
+        if otherargs:
+            for arg in otherargs:
+                journal_cmd += ' {}'.format(arg)
 
         self._log_debug("collecting journal: %s" % journal_cmd)
         self._add_cmd_output(cmd=journal_cmd, timeout=timeout,
--- a/sos/report/plugins/cups.py
+++ b/sos/report/plugins/cups.py
@@ -6,6 +6,8 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
+import os
+from re import compile
 from sos.report.plugins import Plugin, IndependentPlugin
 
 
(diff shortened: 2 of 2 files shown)


pmoravec:
"job related logs are hidden within other logs." - so `add_journal(units="cups") does already collect the job related logs and you want to copy/separate them to extra file(s)? Why the `cups` journal output isnt sufficient (a dummy question)?

About the code itself: I do feel it can be simplified (i.e. return / deal with `[]` to prevent *both* plain `return` statements - esp. the 2nd one is dangerous for future adding code below the return - as that code will be ignored on an unrelated situation), but I dont see a quick fix atm.

sos/report/plugins/__init__.py
```diff
@@ -2332,7 +2332,7 @@ def add_service_status(self, services, **kwargs):
     def add_journal(self, units=None, boot=None, since=None, until=None,
                     lines=None, allfields=False, output=None,
                     timeout=cmd_timeout, identifier=None, catalog=None,
-                    sizelimit=None, pred=None, tags=[]):
+                    sizelimit=None, otherargs=None, pred=None, tags=[]):
```
TurboTurtle, line 2335: Nack to grouping everything under an umbrella argument. If `add_journal()` needs to be extended to support `journalctl` options, the arguments should match the options being supported.

Additionally, we need to properly handle typing (can this take a list? a string? both?) in `add_journal()`. The option handling must also do the string formatting here, not relying on doing this in various plugin's `setup()` method which would be more error prone.

sos/report/plugins/__init__.py
```diff
@@ -2421,6 +2421,10 @@ def add_journal(self, units=None, boot=None, since=None, until=None,
         if output:
             journal_cmd += output_opt % output
 
+        if otherargs:
+            for arg in otherargs:
+                journal_cmd += ' {}'.format(arg)
+
```
TurboTurtle, line 2427: As mentioned above, this is fairly error prone based on what gets handed to `add_journal()`. The option string formatting should be performed here, not in individual plugins.

sos/report/plugins/cups.py
```diff
@@ -18,6 +20,37 @@ class Cups(Plugin, IndependentPlugin):
...
     packages = ('cups',)
 
+    def get_jids(self, dir="/var/spool/cups", n_jids=10,
+                 exists_f=os.path.exists, listdir_f=os.listdir):
+        jids = []
+
+        if not exists_f(dir):
+            return
+
+        spool_files = listdir_f(dir)
```
TurboTurtle, line 30: I'm not sure why we're passing `os.path.exists` or `os.listdir` as arguments just to use them like this? Just call the `os` methods directly.

sos/report/plugins/cups.py
```diff
@@ -41,4 +74,12 @@ def setup(self):
 
         self.add_journal(units="cups")
 
+        jid_arguments = self.get_jid_strings()
+
+        if not jid_arguments:
+            return
```
TurboTurtle, line 80: Nack. Do not perform `return` in `setup()` manually. The loop immediately following this will simply not call execute anything if `jid_arguments` is empty.

TurboTurtle (changes requested):
The addition to `add_journal` needs to be more specific in its implementation, and would also need to be in a commit by itself (including it in the same PR is fine, but separate it out into another commit for historical tracking).

zdohnal:
Hi @pmoravec and @TurboTurtle !

Thank you for the review! I had a talk with Bryan, who works on CUPS cases too and it looks like there is an use case when all cups logs are in one file so an engineer can see what happens in job and in scheduler at the same time.

So the only thing we need to update for this use case is cups right now (to make it show job ids in normal logs). And because job ids will be in logs after that, it looks simpler to run ```grep``` instead of getting a separate files by ```sos```.

So please ignore this PR till ```[WIP]``` disappears from the title or I'll close the PR.

sos/report/plugins/cups.py
zdohnal: It is for possible mocking of functions during testing - AFAIK the convention is the testing should not touch files on the system where the tests are run, so this way you can pass a fake function, which you will implement in the test - it will accept the same args as os methods and will give the same output.

PR #2459, opened 2021-03-23
[ostree] New plugin

BryanQuigley:
This variant only works if the /ostree directory exists which
should cover Gnome OS, Endless OS, and Fedora SilverBlue.

Resolves: #2459

---

### DIFF
--- a/sos/report/plugins/ostree.py
+++ b/sos/report/plugins/ostree.py
@@ -1,0 +1,30 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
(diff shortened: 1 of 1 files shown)


sos/report/plugins/ostree.py
```diff
@@ -0,0 +1,30 @@
...
+class OSTree(Plugin, IndependentPlugin):
+
+    short_desc = 'OSTree'
+
+    plugin_name = 'ostree'
+    profiles = ('system', 'sysmgmt', 'packagemanager')
+    files = ('/ostree',)
+
+    def setup(self):
+        self.add_copy_spec("/ostree/repo/config")
```
pmoravec, line 21: Just an ostree-newbie question: Cant be there some credentials to obfuscate, in a repo config?

BryanQuigley: Looks like a no from the man page.. but I'm very new to it too.

PR #2444, opened 2021-03-12
[sssd] Enable collecting SSSD memory cache

elkoniu:
SSSD plugin by default collects only logs and configuration.
This patch enables collecting memory cache maintained
by SSSD daemon. Cache does not contain any client sensible
data so can be safely included in the sos-report.

Signed-off-by: Paweł Poławski <ppolawsk@redhat.com>

---
Please place an 'X' inside each '[]' to confirm you adhere to our [Contributor Guidelines](https://github.com/sosreport/sos/wiki/Contribution-Guidelines)

- [X] Is the commit message split over multiple lines and hard-wrapped at 72 characters?
- [X] Is the subject and message clear and concise?
- [X] Does the subject start with **[plugin_name]** if submitting a plugin patch or a **[section_name]** if part of the core sosreport code?
- [X] Does the commit contain a **Signed-off-by: First Lastname <email@example.com>**?
- [X] If this commit closes an existing issue, is the line `Closes: #ISSUENUMBER` included in an independent line?
- [X] If this commit resolves an existing pull request, is the line `Resolves: #PRNUMBER` included in an independent line?

PR #2462, opened 2021-03-24
[sssd] Collect memory cache individual files

elkoniu:
By default SSSD collects all memory cache files:
* /var/lib/sss/mc/passwd
* /var/lib/sss/mc/group
* /var/lib/sss/mc/initgroups

They all are included in 25MB size limit for sosreport.
This commits add memory cache files one - by - one,
this way 25MB size limit will be aplied per file.

---

### DIFF
--- a/sos/report/plugins/sssd.py
+++ b/sos/report/plugins/sssd.py
@@ -23,18 +23,21 @@ class Sssd(Plugin):
 
     def setup(self):
         self.add_copy_spec([
-            # Main config file
+            # main config file
             "/etc/sssd/sssd.conf",
             # SSSD 1.14
             "/etc/sssd/conf.d/*.conf",
-            # Memory cache
-            "/var/lib/sss/mc/*",
-            # Dynamic Kerberos configuration
+            # dynamic Kerberos configuration
             "/var/lib/sss/pubconf/krb5.include.d/*"
         ])
 
         # add individual log files
         self.add_copy_spec(glob("/var/log/sssd/*log*"))
+
+        # add memory cache
+        self.add_copy_spec(["/var/lib/sss/mc/passwd",
+                            "/var/lib/sss/mc/group",
+                            "/var/lib/sss/mc/initgroups"])
 
         # call sssctl commands only when sssd service is running,
         # otherwise the command timeouts


elkoniu:
This PR is combination of both:
https://github.com/sosreport/sos/pull/2444
https://github.com/sosreport/sos/pull/2445

It extends logic from https://github.com/sosreport/sos/pull/2445 to affect memory cache files we use in SSSD daemon.

sos/report/plugins/sssd.py
```diff
@@ -23,19 +23,22 @@ class Sssd(Plugin):
...
             "/var/lib/sss/pubconf/krb5.include.d/*"
         ])
 
         # add individual log files
         self.add_copy_spec(glob("/var/log/sssd/*log*"))
 
+        # add memory cache
+        self.add_copy_spec(glob("/var/lib/sss/mc/passwd"))
+        self.add_copy_spec(glob("/var/lib/sss/mc/group"))
+        self.add_copy_spec(glob("/var/lib/sss/mc/initgroups"))
```
TurboTurtle, line 40: Are these files or directories? If they're straight files there's no reason to glob here, and you can just pass the individual file names in a list in a single `add_copy_spec()` call. Each item in a list has the `--log-size` limitation applied to it individually.

elkoniu: Thank you for this suggestion, I just updated the code. Please take a look and share you thoughts.

pmoravec (approved):
Assuming each binary file is under 25MB sizelimit every time, ACK.

pmoravec:
The failed tests are irelevant to the PR; `sssd` plugin was inactivated in both of them.

thalman:
@pmoravec, @elkoniu normally this should be OK, it is much better like this.

But I wonder if we can handle the situation when files are bigger due to some user's customization. I know there is this sizelimit=0 - that would pack everything. The price will be bigger sosreport. Can we somehow "not include" files if they do not fit the limit?

I'm afraid  that people can conclude that memory cache is broken on the system just because it did not fit into sosreport. And I think that in this case it is better to pack nothing rather than tail.

Tom

pmoravec:
> @pmoravec, @elkoniu normally this should be OK, it is much better like this.
> 
> But I wonder if we can handle the situation when files are bigger due to some user's customization. I know there is this sizelimit=0 - that would pack everything. The price will be bigger sosreport. Can we somehow "not include" files if they do not fit the limit?
> 
> I'm afraid that people can conclude that memory cache is broken on the system just because it did not fit into sosreport. And I think that in this case it is better to pack nothing rather than tail.
> 
> Tom

I understand the concern. But `sos report` does log something like:

    2021-02-23 10:42:49,734 INFO: [plugin:rpm] collecting tail of '/var/lib/rpm/Packages' due to size limit
    2021-02-23 10:42:50,117 INFO: [plugin:rpm] skipping '/var/lib/rpm/Conflictname' over size limit

into `sos_logs/sos.log`. That informs a user sufficiently - *if* they check the logs.

I would change the default sizelimit primarily to prevent potential lack of data collected. How often either file can be >25MB ? How often it could be over, say, 100MB (or anyhow huge that it would become impractical to collect it)? If we speak about very rare circumstances, maybe we can start with the default 25MB and optionally amend per future experience.

alexey-tikhonov:
> maybe we can start with the default 25MB and optionally amend per future experience.

+1

Default sizes are within (6mb..10mb) range.
The option to configure this size was introduced very recently and I expect this to be customized in extreme rare cases.

(Still "all or nothing" option would be nice too, because cache tail really doesn't make sense)

thalman:
I'm fine with this PR as it is too.

Thanks for all the work.
Tom

elkoniu:
@pmoravec I think that on SSSD side we agree to continue with this PR in current shape. We can continue with review process.

TurboTurtle (approved):
Ack. Minor stylistic nitpick: sos-style for the list would have it like so:

```python
self.add_copy_spec([
    'foo',
    'bar'
])
```

But I'll take care of that one during a batch merge later today, no worries.

elkoniu:
> Ack. Minor stylistic nitpick: sos-style for the list would have it like so:
> 
> ```python
> self.add_copy_spec([
>     'foo',
>     'bar'
> ])
> ```
> 
> But I'll take care of that one during a batch merge later today, no worries.

Thank you for the fast merge and taking care of this sos-style mismatch :)

PR #2466, opened 2021-03-29
[report] add --cmd-timeout option

pmoravec:
Add --cmd-timeout option to configure command timeout. Plugin-specific
option of the same name (i.e. -k logs.cmd-timeout=60) can control the
timeout per plugin.

Option defaults and global/plugin-specific option preference follows the
--plugin-timeout rules.

Resolves: #2466

---

### DIFF
--- a/man/en/sos-report.1
+++ b/man/en/sos-report.1
@@ -17,6 +17,7 @@ sosreport \- Collect and package diagnostic and support data
           [--label label] [--case-id id]\fR
           [--threads threads]\fR
           [--plugin-timeout TIMEOUT]\fR
+          [--cmd-timeout TIMEOUT]\fR
           [-s|--sysroot SYSROOT]\fR
           [-c|--chroot {auto|always|never}\fR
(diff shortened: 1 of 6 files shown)


TurboTurtle (approved):
This LGTM. The only caveat I see is not being able to override existing timeouts set on commands, since we only reference the `cmdtimeout` property if the command's `timeout` is unset.

That being said, I don't necessarily think that's a real world concern. I think we can implement this as is today, and if we get reports that users are wanting to override the default per-command timeout values then address it at that time. Don't want to over-engineer this from the get-go, basically.

pmoravec:
> This LGTM. The only caveat I see is not being able to override existing timeouts set on commands, since we only reference the `cmdtimeout` property if the command's `timeout` is unset.

Yeah, the same applies to `--log-size` not being applied to `a_c_s` calls with `sizelimit` set. If we will change the current behaviour (that is hard to judge if is OK or not), we should change it in both places the same way / consistently.

Issue #2468, opened 2021-03-31
new kernel module (devlink) loaded after sosreport execution on ubuntu Bionic/18.04LTS


slashdd:
It was reported via our autopkgtest env:
https://objectstorage.prodstack4-5.canonical.com/v1/AUTH_77e2ada1e7a84929a74ba3b87153c0ac/autopkgtest-bionic/bionic/arm64/s/sosreport/20210331_013736_df618@/log.gz

but was also able to reproduce manually as follows:
The networking plugin somehow load devlink module by happenstance, as I don't think it is intentional.

```

root@srv:~# lsmod | grep -i devlink
root@srv:~# sos report -o networking

sosreport (version 4.1)

This command will collect system configuration and diagnostic
information from this Ubuntu system.

For more information on Canonical visit:

        Community Website  : https://www.ubuntu.com/
        Commercial Support : https://www.canonical.com

The generated archive may contain data considered sensitive and its
content should be reviewed by the originating organization before being
passed to any third party.

No changes will be made to system configuration.


Press ENTER to continue, or CTRL-C to quit.

Please enter the case id that you are generating this report for []: 

 Setting up archive ...
 Setting up plugins ...
[plugin:networking] skipped command 'ip -s macsec show': required kmods missing: macsec.  Use '--allow-system-changes' to enable collection.
[plugin:networking] skipped command 'ss -peaonmi': required kmods missing: udp_diag, unix_diag, af_packet_diag, netlink_diag, inet_diag, tcp_diag.  Use '--allow-system-changes' to enable collection.
 Running plugins. Please wait ...

  Starting 1/1   networking      [Running: networking]                                    
  Finished running plugins                                                               
Creating compressed archive...

Your sosreport has been generated and saved in:
	/tmp/sosreport-srv-2021-03-31-tvjgeyv.tar.xz

 Size	29.75KiB
 Owner	root
 sha256	72d49c26463f8fe42f11696dbc69aaa6eaf4e621551d88822b72a6f27ed7a136

Please send this file to your support representative.

root@srv:~# lsmod | grep -i devlink
devlink                45056  0
root@srv:~# lsb_release -cs
bionic

```

slashdd:
```
root@srv:~# devlink dev
root@srv:~# lsmod | grep -i devlink
devlink                45056  0
```

slashdd:
Behaviour not observe other than in Bionic.

slashdd:
I think it is related to how the kernel was configured.

[...]

PR #2469, opened 2021-03-31
[tests] ignore devlink for certains kernel config

slashdd:
Ignore devlink module for kernel configuration where
devlink is set as loadable module (=m). The module
may automatically load after devlink cmd is performed
by the networking plugin.

sosreport reproducer:
 $ lsmod | grep -i devlink
 $ sos report -o networking

sosreport (version 4.1)

.....

 /tmp/sosreport-srv-2021-03-31-tvjgeyv.tar.xz

 Size 29.75KiB
 Owner root
 sha256 72d49c26463f8fe42f11696dbc69aaa6eaf4e621551d88822b72a6f27ed7a136

 $ lsmod | grep -i devlink
devlink 45056 0

More precisely caused after this specific code line:
https://github.com/sosreport/sos/blob/master/sos/report/plugins/networking.py#L109

Manual reproducer:
 $ lsmod | grep -i devlink
 $ devlink dev
 $ lsmod | grep -i devlink
 devlink 45056 0

Closes: #2468

---

### DIFF
--- a/tests/simple.sh
+++ b/tests/simple.sh
@@ -102,7 +102,7 @@ add_failure () {
 test_normal_report () {
     cmd="-vvv"
     # get a list of initial kmods loaded
-    kmods=( $(lsmod | cut -f1 -d ' ' | sort) )
+    kmods=( $(lsmod | grep -v devlink | cut -f1 -d ' ' | sort) )
     run_expecting_success "$cmd" extract
     if [ $? -eq 0 ]; then
         if [ ! -f /var/tmp/sosreport_test/sos_reports/sos.html ]; then
@@ -118,7 +118,7 @@ test_normal_report () {
             add_failure "did not find debug logging when using -vvv"
         fi
         # new list, see if we added any
-        new_kmods=( $(lsmod | cut -f1 -d ' ' | sort) )
+        new_kmods=( $(lsmod | grep -v devlink | cut -f1 -d ' ' | sort) )
         if [ "$(printf '%s\n' "${kmods[@]}" "${new_kmods[@]}" | sort | uniq -u)" ]; then
             add_failure "new kernel modules loaded during execution"
             echo "$(printf '%s\n' "${kmods[@]}" "${new_kmods[@]}" | sort | uniq -u)"


pmoravec:
Won't we skip testing for the module loaded also when it matters / when sosreport shouldnt load it? Shouldnt be the `grep -v` conditional based on some kernel settings? (or will that be too complex check, error prone?)

TurboTurtle:
Actually, wait... that would likely break collection on systems where `devlink` is built-in. This requires a bit more investigation, I'm not a fan of adding more exceptions to our kmod test.

bmr-cymru:
It's unfortunate that there's nothing reflected in either `lsmod` or `/proc/modules` for modules that support either modular or linked-in builds. There are a few other traces left on the system that could be of use in making this determination (in rough order of unsavouriness):

 * Look for `/sys/class/devlink`. I haven't tested but I'm fairly certain from code inspection that this *is* always present with either a loadable module or built-in, and is never present otherwise.
 * Look for devlink symbols in `/proc/kallsyms`
   - As long as we pick a good symbol to look for these are present regardless of whether the object is built-in or a loadable module, but only if the module is already loaded. E.g. `devlink_init`
 * Look for devlink debug artefacts in `/sys`. These aren't so specific, in that I think it's technically possible to build a kernel _with_ built-in devlink, but _without_ some of these extra paths. I'm not aware of any distribution that actually does that however.
   - Things like `/sys/kernel/tracing/events/devlink`

TurboTurtle:
This also brings up the question of if we should include built-ins in the kmod list for both plugin enablement and predicate usage? We could parse `/lib/modules/$(uname -r)/modules.builtin` and append those to a policies kmod list from `init_kernel_modules()`.

I don't know how reliable this would be though, as `devlink` for example is not in `modules.builtin` either (on F33 at least).

bmr-cymru:
Good point - also not sure how dependable those kernel package files are on other distro families, e.g. Debian/Ubuntu or SuSE.

slashdd:
Let's move the discussion back to the bug. I'll close the PR. I agree that there is better viable solution.

Issue #2470, opened 2021-03-31
ibmkvm policies effectiveness?


BryanQuigley:
Aside from showing an IBM specific support link (but it's been broken for a while), I'm not sure what purpose the ibmkvm distro policies actually do today - namely after the changes to IndependentPlugin.  If it's properly marked as based on RHEL it should do the *right thing* based on the existing distro policies.

BryanQuigley:
@kamalesh-babulal  thoughts?

TurboTurtle:
Yeah I don't know if it's been useful for a while now that you bring it up. No objections to drop it from me (potentially ZKVM as well?), but I'll try and reach out to some IBM folks to get more feedback on it.

hegdevasant:
We created `ibmkvm` policy when we created initial PowerKVM project (few year ago). Now that all support is integrated into  distro , it may be OK to drop this support.

I will poke around internally and get back to you,

zKVM : I don't know the story around zKVM. I will try to check.

-Vasant

kamalesh-babulal:
@BryanQuigley Sorry for the delayed response, I was out of office for a week. I will let @hegdevasant to get back with response. Vasant Thank you.

hegdevasant:
@BryanQuigley I had contacted various folks in Power and Z side. Based on our internal discussion we think its safe to remove ibmkvm policies.

-Vasant

TurboTurtle:
@hegdevasant one more consideration - I'm working on porting `dbginfo.sh` for #2228. Most of what it collects is already in sos, with the exception of the z/VM commands, which are easily added in a new plugin. If we port dbginfo.sh, does that change anything?

If no, ack to removing.

hegdevasant:
@TurboTurtle 

> @hegdevasant one more consideration - I'm working on porting `dbginfo.sh` for #2228. Most of what it collects is already in sos, with the exception of the z/VM commands, which are easily added in a new plugin. If we port dbginfo.sh, does that change anything?

Sorry. I don't know about z/VM side. I have forwarded this request to system z folks . Will let you know once I get response from them.

-Vasant

mihajlov:
The KVM for IBM z product was discontinued, so there's no impact to current or future sos activities.

PR #2474, opened 2021-04-03
[IBM] Remove unneeded policies

BryanQuigley:
These policies haven't done anything since
IndependentPlugin was made a more common default.

Closes: #2470
Resolves: #2474

---


BryanQuigley:
Wait for feedback in Issue before merging.

TurboTurtle:
We'll also want to removing the tagging classes from `sos/report/plugins/__init__.py`.

bmr-cymru:
Can we be a bit more precise about the language in the commit message? The commits are not removing any of the IBM *policies* (which would concern me, if we did not have agreement from the groups that submitted them), they are just deleting the IBM-specific tagging classes that they define.

I think the commit subject should be something like "[plugins,policy] remove IBM policy tagging classes".

Also, if you are going to delete those classes then you also need to remove the references to them from the corresponding policy class to avoid breaking stuff:

```python
class PowerKVMPolicy(RedHatPolicy):
    distro = "PowerKVM"
    vendor = "IBM"
    vendor_url = "http://www-03.ibm.com/systems/power/software/linux/powerkvm"

    def __init__(self, sysroot=None):
        super(PowerKVMPolicy, self).__init__(sysroot=sysroot)
        self.valid_subclasses = [PowerKVMPlugin, RedHatPlugin]
```

```python
class ZKVMPolicy(RedHatPolicy):
    distro = "IBM Hypervisor"
    vendor = "IBM Hypervisor"
    vendor_url = "http://www.ibm.com/systems/z/linux/IBMHypervisor/support/"

    def __init__(self, sysroot=None):
        super(ZKVMPolicy, self).__init__(sysroot=sysroot)
        self.valid_subclasses = [ZKVMPlugin, RedHatPlugin]
```

One concern I have with making changes to this area is that I don't think any of us have access to an environment where we can test these changes on the systems they affect. I would suggest trying to reach out to someone at IBM to find out if the policies are still relevant/used in the corresponding products, and if so to get these changes tested before the next release. If they aren't being used any more then we might as well cull the whole thing.

BryanQuigley:
Sorry, I could have been more clear - my first comment was we should wait some time to see if we hear back from IBM per #2470.

This commit does fully remove the IBM policy classes you mention. The links are one of the few changes they make and they have been broken for some time.

bmr-cymru:
IBM always contributed to sos more via the mailing list than via GitHub. I think it would be courteous to reach out to them and get positive confirmation that this is no longer used before deleting it. The archives give some email addresses to try, and failing that me or @TurboTurtle should be able to find a contact.

Afaik the products still exist and although they have not been keeping the upstream maintenance up on these I think removing support is one place where it's better to ask for permission than forgiveness.

If there's no response from anyone at IBM after sending that notice then it's appropriate to just drop it.

TurboTurtle:
I've been attempting to find an up to date contact internally, no responses yet. I'll try again and send a devel list email as well later today/tomorrow to see if we can drum something up.

TurboTurtle (approved):

PR #2476, opened 2021-04-06
[sssd] legacy-3.9 merge of PR #2445 #2444 #2462

jjansky1:
The issue is that SSSD creates individual log files for its
components. To be able to track the issue we need all of them.

With one wildcard copy set we usually get just one truncated
log file and this is not very useful for solving issues. We
need to track the request accross logs to understand the
problem. Also log file names are specific for paricular
configuration.

With this patch we list log files under /var/log/sssd and
we add them one by one.

SSSD plugin by default collects only logs and configuration.
This patch enables collecting memory cache maintained
by SSSD daemon. Cache does not contain any client sensible
data so can be safely included in the sos-report.

By default SSSD collects all memory cache files:

/var/lib/sss/mc/passwd
/var/lib/sss/mc/group
/var/lib/sss/mc/initgroups
They all are included in 25MB size limit for sosreport.
This commits add memory cache files one - by - one,
this way 25MB size limit will be aplied per file

Related: #2445
Related: #2444
Related: #2462
Resolves: #2476

Signed-off-by: Jan Jansky jjansky@redhat.com

---

### DIFF
--- a/sos/plugins/sssd.py
+++ b/sos/plugins/sssd.py
@@ -23,20 +23,21 @@ class Sssd(Plugin):
 
     def setup(self):
         self.add_copy_spec([
-            # Main config file
+            # main config file
             "/etc/sssd/sssd.conf",
             # SSSD 1.14
             "/etc/sssd/conf.d/*.conf",
-            # Main logs directory
-            "/var/log/sssd/*",
-            # Memory cache
-            "/var/lib/sss/mc/*",
-            # Dynamic Kerberos configuration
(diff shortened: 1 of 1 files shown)


pmoravec (approved):
Also verified the PR is a merge of the three small commits on main branch.

TurboTurtle (approved):

TurboTurtle:
Merged into `legacy-3.9` via 7d6cb4e

PR #2480, opened 2021-04-09
[cleaner] Do not break iteration of parse_string_for_keys on first match

TurboTurtle:
Previously, `parse_string_for_keys()`, called by `obfuscate_string()`
for non-regex based obfuscations, would return on the first match in the
string found for each parser.

Instead, continue iterating over all items in each parser's dataset
before returning the (now fully) obfuscated string.

Resolves: #2480

---


pmoravec (approved):
I also checked no other error like this.

PR #2487, opened 2021-04-13
[networking] check for the presence of devlink

slashdd:
On certain kernel configuration, devlink cmds may trigger
the module to load.

Closes: #2468

---

### DIFF
--- a/sos/report/plugins/networking.py
+++ b/sos/report/plugins/networking.py
@@ -9,6 +9,7 @@
 from sos.report.plugins import (Plugin, RedHatPlugin, UbuntuPlugin,
                                 DebianPlugin, SoSPredicate)
 from os import listdir
+from os import path
 
 
 class Networking(Plugin):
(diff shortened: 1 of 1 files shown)


TurboTurtle:
This would prevent collection on systems where `devlink` is built into the running kernel. @bmr-cymru mentioned that the presence of `/sys/class/devlink/` _might_ be a somewhat reliable marker in place of the kernel module.

Longer term we need to figure out a way to reliably populate our kmods list with built-ins as well.

slashdd:
> This would prevent collection on systems where `devlink` is built into the running kernel. @bmr-cymru mentioned that the presence of `/sys/class/devlink/` _might_ be a somewhat reliable marker in place of the kernel module. (EDIT: so we may want to use that here as a temporary measure until we can flesh out a better approach using a predicate)
> 
> Longer term we need to figure out a way to reliably populate our kmods list with built-ins as well.

Okay, I changed the PR accordingly to your note.

TurboTurtle (approved):

pmoravec (approved):

Issue #2495, opened 2021-04-14
version.txt contains lots of unhelpful noise


slashdd:
See [Ubuntu LP:# 1922925](https://bugs.launchpad.net/ubuntu/+source/sosreport/+bug/1922925) for context.

slashdd:
'unversioned' is the default behaviour, but I can't see any plugin having version number, and see the value of such information.

```
    def version(self):
        """Fetch version information from all plugins and store in the report
        version file"""

        versions = []
        versions.append("sosreport: %s" % __version__)

        for plugname, plug in self.loaded_plugins:
            versions.append("%s: %s" % (plugname, plug.version))

        self.archive.add_string(content="\n".join(versions),
                                dest='version.txt')
```

```
    :cvar version:      The version of the plugin, defaults to 'unversioned'
    :vartype version:   ``str``

```

slashdd:
Only jars, seems to used this feature:

```
class Jars(Plugin, RedHatPlugin):

    short_desc = 'Collect information about available Java archives'

    plugin_name = "jars"
    version = "1.0.0"
```

slashdd:
One thing we can do, is to put the 'unversioned' plugin into silence to stop the unecessary noise.

```
        for plugname, plug in self.loaded_plugins:
+         if plug.version != 'unversioned':
              versions.append("%s: %s" % (plugname, plug.version))
```
Or drop this version (which seems to be like legacy) option that very few plugins uses anyway.

TurboTurtle:
Yeah, plugin versions never really took off. I think it would be more relevant if there was wider (any?) adoption of users writing their own local plugins, or plugins being shipped with individual components/packages, but in the current packaging design I'd be in favor of dropping it.

We did experiment for a little while in products shipping their own plugins, but in reality they went unmaintained and even those that did ship did not keep updated versioning.

+1 to rip it out for plugins. This does raise the question in my mind though, do we want to separately version sos components like `report`, `collect`, or `clean`, or just keep them all under the "umbrella" version of the sos packaging? I'm in favor of the latter but I'm open to hearing reasons for making that kind of change.

slashdd:
I'm in favor of the latter as well.

PR #2496, opened 2021-04-15
[global] Drop plugin version

slashdd:
Plugin version never really took off. A very few plugins
had it configured (e.g. jars, pacemaker)

Closes: #2495

---


TurboTurtle (approved):
Ack.

@bmr-cymru mentioned an interesting idea on IRC about this. What if we recorded a hash for each plugin pyfile? That would give us a very strong indicator of if the plugin was really a sos-4.X plugin or if it had been modified.

I rather like that as a replacement once this is removed.

PR #2499, opened 2021-04-15
[cirrus] Add testing on CentOS Stream 8

TurboTurtle:
As CentOS Stream now runs ahead of RHEL, instead of behind, using it as
a testing distribution on GCE allows us to have more confidence on those
distributions than testing on Fedora alone.

As CentOS Stream does not have version locks, there is only one version
to test on, rather than two as is the case with Fedora and Ubuntu. The
image we build from will need to be updated more regularly however as
new versions are released following RHEL minor version releases.

Resolves: #2499

---

### DIFF
--- a/tests/report_tests/plugin_tests/logs.py
+++ b/tests/report_tests/plugin_tests/logs.py
@@ -12,7 +12,6 @@ import os
 
 from sos_tests import StageOneReportTest, StageTwoReportTest
 from string import ascii_uppercase, digits
-from systemd import journal
 
 
 class LogsPluginTest(StageOneReportTest):
@@ -43,15 +42,15 @@ class LogsSizeLimitTest(StageTwoReportTest):
 
     sos_cmd = '-o logs'
     packages = {
-        'rhel': 'python3-systemd',
-        'ubuntu': 'python3-systemd'
+        'rhel': ['python3-systemd'],
+        'ubuntu': ['python3-systemd']
     }
 
-    # override the stage2 mocking here to inject a string into the journal
-    def setup_mocking(self):
+    def pre_sos_setup(self):
         # write 20MB at a time to side-step rate/size limiting on some distros
         # write over 100MB to ensure we will actually size limit inside sos,
         # allowing for any compression or de-dupe systemd does
+        from systemd import journal
         rsize = 20 * 1048576
         for i in range(3):
             # generate 20MB, write it, then write it in reverse.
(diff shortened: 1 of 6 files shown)


TurboTurtle:
`dm_mod` as an unloaded kmod on centos is honestly shocking to me, need to think on how to approach that one.

For the failing journal test, we just need to extend the "only fail if we don't collect this if it really does exist on the host" bit from `assertFileCollected` to `assertFileGlobInArchive` as well.

TurboTurtle:
So apparently `grub2-mkconfig` loads `dm_mod`. And `dm_mod` has the same issue as `devlink` - if it's builtin it doesn't get reported in `/usr/lib/modules/$(uname -r)/modules.builtin`, `/proc/modules`, `lsmod`, etc...

I don't want to end up peppering `if os.path.isdir()` checks all over plugins but at the moment I don't see a way to resolve this module loading otherwise for systems that don't have it loaded that doesn't also skip those collections for systems that have it as a builtin.

TurboTurtle:
Strange. I've manually run the test suite on a freshly-provisioned CentOS Stream VM on GCE and it passes each time. I've added the size of the resulting journal as part of the fail message to try and dig into this a bit more.

TurboTurtle:
Based on the reported size of the failed test I can only surmise that the `pre_sos_setup()` action of writing garbage to the journal is not actually occurring. Why this is only happening on centos stream, and why this only happens with the automated testing (as again, manually running this branch on a CS VM runs as expected) is beyond me at the moment.

TurboTurtle:
Switching to writing to the journal via `journal.stream()` instead of `journal.send()` seems to have resolved the issue of not writing to the system journal for large sets of data.

Ready for review. Note this also contains the change for #2500, included in this patchset since it is related to the `dm_mod` loading issue discovered as part of this testing.

TurboTurtle:
Well that was a frustrating few days, I apologize to anyone who was still on the notification list for this even when it was back in draft.

At any rate, the tests are now more reliable and centos stream testing should, finally, be ready for merging into master.

tests/report_tests/plugin_tests/logs.py
```diff
@@ -51,22 +51,26 @@ def pre_sos_setup(self):
...
+        rsize = 10 * 1048576
+        for i in range(6):
+            # generate 10MB, write it, then write it in reverse.
             # Spend less time generating new strings
             rand = ''.join(random.choice(ascii_uppercase + digits) for _ in range(rsize))
-            journal.send(rand)
-            journal.send(rand[::-1])
+            sosfd.write(rand + '\n')
+            # sleep to avoid burst rate-limiting
+            sleep(15)
```
BryanQuigley, line 62: Was this sleep actually required?

TurboTurtle: Not strictly, but that's more due to how long it takes to generate the test data strings. Previously, we were hitting the burst limit rates when the 20MB strings got split up into X number of messages, the change to stream alleviated that somewhat but not completely. We'd hit the limits somewhere in the second write (when we reverse the generated string). If we had prepared strings that were just rapid-fired off, we'd hit those burst limits again.

So at the moment, we could remove it, but if some change happens that increases the performance of generating that random test data, we could hit burst limits again.

We could also reduce the sleep timer here though if you'd prefer. I chose 15 seconds here based on the journald default of 30 seconds for the burst window and the previous behavior seen.

BryanQuigley: No strong preferences

BryanQuigley (approved):
LGTM (worried about long term kernel module handling as it seems we are running into more issues with it, but not going to fix that here).

PR #2503, opened 2021-04-21
[subscription_manager] call curl to the RHSM server

pmoravec:
Collect a verbose curl command output, run against RHSM server like
subscription-manager does. This helps troubleshooting with RHSM proxy
issues.

Resolves: #2503

---

### DIFF
--- a/sos/report/plugins/subscription_manager.py
+++ b/sos/report/plugins/subscription_manager.py
@@ -8,6 +8,7 @@
 
 from sos.report.plugins import Plugin, RedHatPlugin
 import glob
+from configparser import NoOptionError, NoSectionError
 
 
 class SubscriptionManager(Plugin, RedHatPlugin):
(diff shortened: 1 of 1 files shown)


pmoravec:
Not sure about few things:
- should we call it only based on a plugin option which is disabled by default? (sufficient for me)
- if opening RHSM conf fails, shouldnt we fail over to default `curl -vv https://subscription.rhsm.redhat.com:443/subscription --cacert /etc/rhsm/ca/redhat-uep.pem` ? (I think it makes sense?)
- as we are usually interested in connection details and not the response itself, is it worth removing the prefix / URI (`/subscription`) to not receive RHSM server's capabilities? (the capabilities can sometimes help but often ridiculous info received)

TurboTurtle:
> Not sure about few things:
> 
> * should we call it only based on a plugin option which is disabled by default? (sufficient for me)

If this isn't expected to take a while to run, I think having it as a default capture is fine - especially if this is a common data point to get for RHSM issues

> * if opening RHSM conf fails, shouldnt we fail over to default `curl -vv https://subscription.rhsm.redhat.com:443/subscription --cacert /etc/rhsm/ca/redhat-uep.pem` ? (I think it makes sense?)

Agreed. I'm assuming the logic is here for the cases where a system is registered to a Satellite rather than RHSM directly, right? In that case I'd say a default location would be helpful.

> * as we are usually interested in connection details and not the response itself, is it worth removing the prefix / URI (`/subscription`) to not receive RHSM server's capabilities? (the capabilities can sometimes help but often ridiculous info received)

No opinion here, as I'm not sure what the data looks like normally. Would it be at all beneficial to have 2 collections - one with and one without the `/subscription` end point?


Overall this looks ok, I just think we might be able to make it a bit more readable for at-a-glance reviews in the future. Perhaps splitting out the proxy and server string determinations into different methods? E.G. something like:

```python
try:
    from rhsmlib.services import config
    import rhsm.config
    config = config.Config(rhsm.config.initConfig())
    proxy = self.get_proxy_string(config)
    server_url = self.get_server_url(config)
    curlcmd = "curl -vv %s --cacert %s %s" % (server_url, cert, proxy)
except (ModuleNotFoundError, ImportError):
    pass 
```

pmoravec:
Sadly there is no method like `get_proxy_string` or `get_server_url` so the constitution of the `curl` command seems to be required here. The best I found in RHSM API is getting connection object but that cant be used for a verbose check via `curl`.

Default fallback added, timeout preventing plugin delays added, imho the /subscriptions is sufficient to have.

TurboTurtle:
> Sadly there is no method like `get_proxy_string` or `get_server_url` so the constitution of the `curl` command seems to be required here. The best I found in RHSM API is getting connection object but that cant be used for a verbose check via `curl`.
> 

Right, I just meant moving the string construction out into a new `get_server_url()` or similarly named method within the plugin - we'd still need to write the construction logic ourselves.. Just so that the flow is easier to read.

pmoravec:
Ah I got it, I misread the previous comment on the string construction. PR updated.

TurboTurtle (approved):

PR #2504, opened 2021-04-22
[kernel] Collect kernel log for page allocation failure

mamatha4:
This patch is to update kernel plugin to collect
page allocation failure info

Signed-off-by: Mamatha Inamdar mamatha4@linux.vnet.ibm.com

- [ x] If this commit resolves an existing pull request, is the line `Resolves: #PRNUMBER` included in an independent line?

### DIFF
--- a/sos/report/plugins/kernel.py
+++ b/sos/report/plugins/kernel.py
@@ -107,6 +107,8 @@ class Kernel(Plugin, IndependentPlugin):
             "/var/log/dmesg",
             "/sys/fs/pstore",
             "/sys/kernel/debug/dynamic_debug/control",
+            "/sys/kernel/debug/extfrag/unusable_index",
+            "/sys/kernel/debug/extfrag/extfrag_index",
             clocksource_path + "available_clocksource",
             clocksource_path + "current_clocksource"
         ])
--- a/sos/report/plugins/snapper.py
+++ b/sos/report/plugins/snapper.py
@@ -1,0 +1,27 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, IndependentPlugin
+
+
+class Snapper(Plugin, IndependentPlugin):
+
+    short_desc = 'System snapper'
+
+    plugin_name = 'snapper'
+    commands = ("snapper",)
+
+    def setup(self):
+
+        self.add_cmd_output([
+            "ls -la /usr/lib/snapper/",
+            "snapper --version",
+            "snapper list"
+        ])
+
+# vim: set et ts=4 sw=4 :
--- a/sos/report/plugins/block.py
+++ b/sos/report/plugins/block.py
@@ -41,6 +41,7 @@ class Block(Plugin, IndependentPlugin):
             "/sys/block/*/queue/",
             "/sys/block/sd*/device/timeout",
             "/sys/block/hd*/device/timeout",
+            "/sys/block/sd*/device/state",
         ])
 
         cmds = [


sos/report/plugins/kernel.py
```diff
@@ -107,6 +107,10 @@ def setup(self):
             "/var/log/dmesg",
             "/sys/fs/pstore",
             "/sys/kernel/debug/dynamic_debug/control",
+            "/proc/pagetypeinfo",
```
pmoravec, line 110: We collect this in `memory` plugin; we can not collect it in two plugins as that cause some issues during collection. If you think the file is worth to collect here, please provide a rationale and remove the file from `memory` plugin.

sos/report/plugins/kernel.py
```diff
@@ -107,6 +107,10 @@ def setup(self):
...
+            "/proc/buddyinfo",
```
pmoravec, line 111: We already collect it 20 lines above :)

sos/report/plugins/logs.py
```diff
@@ -56,6 +56,7 @@ def setup(self):
 
         self.add_cmd_output("journalctl --disk-usage")
         self.add_cmd_output('ls -alRh /var/log/')
+        self.add_cmd_output('ls -la /usr/lib/snapper/')
```
pmoravec, line 59: Why excatly `logs` plugin shall list directory for a user libraries of a system recovery and snapshot management tool?

sos/report/plugins/kernel.py
mamatha4: ah yes..sorry forgot to check this..will update and send new version of patch

sos/report/plugins/kernel.py
mamatha4: yes.. we already have this info.. will remove this info

mamatha4: will update and send new version of patch..Thanks

sos/report/plugins/logs.py
mamatha4: Please let us know in which plugin we can collect this data..

TurboTurtle: It looks like a new `snapper` plugin would be appropriate to capture this as well as other `snapper` relevant config files and/or command output.

TurboTurtle:
@mamatha4 looks like your commits got split somehow, as this branch only has the one commit attached to it.

mamatha4:
ok..let me push the branch once again

mamatha4:
Can you please check now

sos/report/plugins/snapper.py
```diff
@@ -0,0 +1,24 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from glob import glob
```
TurboTurtle, line 9: Unused import

sos/report/plugins/snapper.py
```diff
@@ -0,0 +1,24 @@
...
+from sos.report.plugins import Plugin, IndependentPlugin
+
+
+class Snapper(Plugin, IndependentPlugin):
+
+    short_desc = 'System snapper'
+
+    plugin_name = 'snapper'
+    profiles = ('system', 'hardware')
+
```
TurboTurtle, line 19: There needs to be some enablement triggers here. `packages`, or `files`, or `commands`, etc... Without one, this will always be enabled.

sos/report/plugins/snapper.py
```diff
@@ -0,0 +1,24 @@
...
+    def setup(self):
+
+        self.add_cmd_output('ls -la /usr/lib/snapper/')
```
TurboTurtle, line 22: Surely there is more relevant data to collect for `snapper` than just a directory listing?

Taking a quick look there's:

`snapper list-configs`
`snapper get-config` (is this iterable for each config listed from the above command?)
`snapper list` (is there any value is iterating over this list with some other `snapper` command for more info?)

sos/report/plugins/snapper.py
mamatha4: ok.. will update this
    short_desc = 'System snapper'

    plugin_name = 'snapper'
    commands = ("snapper",)

sos/report/plugins/snapper.py
mamatha4: ok..right now I will be adding following commands, 
  ls -al /usr/lib/snapper/
 snapper --version
 snapper list
Later will  work on other commands and add the same

TurboTurtle (approved):

mamatha4:
Thanks

Issue #704, opened 2015-12-16
[apache] factor out common code in RedHatApache


bmr-cymru:
Adding support for JBoss web server introduced code duplication in the `RedHatApache` class, e.g.:

``` python
        self.add_copy_spec([
            "/etc/httpd/conf/httpd.conf",
            "/etc/httpd/conf.d/*.conf",
            "/etc/httpd/conf.modules.d/*.conf",
            # JBoss Enterprise Web Server 2.x
            "/etc/httpd22/conf/httpd.conf",
            "/etc/httpd22/conf.d/*.conf",
            # Red Hat JBoss Web Server 3.x
            "/etc/httpd24/conf/httpd.conf",
            "/etc/httpd24/conf.d/*.conf",
            "/etc/httpd24/conf.modules.d/*.conf",
        ])
```

Factor this out to remove the repetition and instead loop over the three possible top-level directories in `/etc`.

bmr-cymru:
btw: attempting to collect e.g. `conf.modules.d/*.conf` for a release (2.2) that does not use it is explicitly not an error in sos - this helps to keep the code simple since there are no harmful side-effects to trying to collect something that is not there.

Amitgb14:
@bmr-cymru, I write some lines to reduce duplication, Please let me know it's correct.

```
paths     = ['/etc/httpd{}/conf/httpd.conf', '/etc/httpd{}/conf.d/*.conf', 
                  '/etc/httpd{}/conf.modules.d/*.conf']
versions  = ['', '22', '24']
conf_path = [path.format(ver) for ver in versions for path in paths]
self.add_copy_spec(conf_path)

example;

for path in conf_path:
    print path

output:
/etc/httpd/conf/httpd.conf
/etc/httpd/conf.d/*.conf
/etc/httpd/conf.modules.d/*.conf
/etc/httpd22/conf/httpd.conf
/etc/httpd22/conf.d/*.conf
/etc/httpd22/conf.modules.d/*.conf
/etc/httpd24/conf/httpd.conf
/etc/httpd24/conf.d/*.conf
/etc/httpd24/conf.modules.d/*.conf
```

bmr-cymru:
When you're making a change like this the important thing is to step back and ask "Is this simpler?", "Is this more readable?", "is there less string duplication?"

Although there is duplication in the original I think the answer to the questions for the above snippet is "no".

[...]

PR #2505, opened 2021-04-22
[apache] Refactor duplicate code for Jboss file collections

TurboTurtle:
Refactor the collections for the different httpd versions collected by
the RedHatApache instance of this plugin, to remove the repetition of
similar collection points.

Moving forward, collection of newer httpd versions will only need to
specify the version number, rather than explicitly repeat the same file
collection paths.

Closes: #704
Resolves: #2505

---

### DIFF
--- a/sos/report/plugins/apache.py
+++ b/sos/report/plugins/apache.py
@@ -48,48 +48,44 @@ class RedHatApache(Apache, RedHatPlugin):
     def setup(self):
         super(RedHatApache, self).setup()
 
-        self.add_copy_spec([
-            "/etc/httpd/conf/httpd.conf",
-            "/etc/httpd/conf.d/*.conf",
-            "/etc/httpd/conf.modules.d/*.conf",
-            # JBoss Enterprise Web Server 2.x
-            "/etc/httpd22/conf/httpd.conf",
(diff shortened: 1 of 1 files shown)


sos/report/plugins/apache.py
```diff
@@ -48,48 +48,42 @@ class RedHatApache(Apache, RedHatPlugin):
...
+        # relevant log files within each
+        logdirs = ["/var/log/httpd%s" % ver for ver in vers]
+        logs = [
+            "access_log",
+            "error_log",
+            "ssl_access_log",
+            "ssl_error_log"
+        ]
 
         self.add_forbidden_path("/etc/httpd/conf/password.conf")
```
pmoravec, line 73: Not relevant to this patch itself, but does not the plugin collect `/etc/httpd2[2,4]/conf/password.conf` that we should forbid (using the `vers` variable)?

pmoravec (approved):

sos/report/plugins/apache.py
TurboTurtle: Yep - we should totally expand that.

pmoravec (approved):

Issue #1332, opened 2018-06-06
[sosreport] log command size output limiting


bmr-cymru:
Make sure that the user is aware that command output has been size limited by either:

* Logging a message indicating limits have been exceeded (once only)

Or:

* Placing the limited output in a file named like `command_--options.tailed` with a symbolic link from `command_--options`

knoha-rh:
From support side, we would like to know whether the output are truncated/size-limited by sos.
To make it clear, I'd like to vote to the filename which describes that the file is truncated/size-limited.

k-hagio:
I also think that the filename approach is good because of ease to notice it and consistency with `add_copy_spec()`, and it is better to have both the ways to describe it more clearly if possible.

jcastill:
I think logging a message so customer sees that a size limit has been exceeded (perhaps with a suggestion to attach the file separately, or how to call sos with an increase log size?), plus adding a filename as knoha-rh suggests, like messages.truncated, will be very useful for customers and support.

PR #2506, opened 2021-04-22
[Plugin] Make command output size limiting obvious

TurboTurtle:
When command output is size limited it should be made obvious that it is
not the whole output of the command, beyond the collected output (most
likely) starting in the middle of a string.

To do this, we add a 'truncated' key to the results dict returned by
`sos_get_command_output()`, which is set based on if the `AsyncReader`
was filled during command execution.

From that, first log if a command was truncated. Second, instead of
dropping the collected output in the normal plugin directory, instead
drop it into `sos_strings` with a `.tailed` extention to mimic how we
handled tailed file collections. Finally, add a symlink from the plugin
dir with the original filename that points to the `sos_strings` path.

Closes: #1332
Resolves: #2506

---

### DIFF
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -1935,6 +1935,7 @@ class Plugin(object):
             'parameters': cmd.split(' ')[1:],
             'exec': cmd,
             'filepath': None,
+            'truncated': result['truncated'],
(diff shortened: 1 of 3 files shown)


pmoravec (approved):
Works great, even for cmdoutputs in `subdir`!

PR #2507, opened 2021-04-22
[archive|logging] Fix archive debug logging, change meaning of different logging levels

TurboTurtle:
This is a 2-patch set that accomplishes 2 changes:

* Fixes archive debug logging to not be tied to `--debug`, but some level of `--verbose`
* Gives more (and documented) meaning to the different verbosity levels

One of the more pertinent changes is that a single `-v` will now write debug logs to `sos.log`, but will *no longer* print info level messages to console. Using `-vv` is now required for printing logs to console. Using `-vvv` is what enables the archive debug logs being recorded, as these messages generate a whole lot more noise than any other part of logging currently.

---

### DIFF
--- a/sos/component.py
+++ b/sos/component.py
@@ -256,7 +256,7 @@ class SoSComponent():
                                           enc_opts, self.sysroot,
                                           self.manifest)
 
-        self.archive.set_debug(True if self.opts.debug else False)
+        self.archive.set_debug(True if self.opts.verbosity > 2 else False)
 
     def _setup_logging(self):
         """Creates the log handler that shall be used by all components and any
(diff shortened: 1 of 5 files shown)


sos/component.py
```diff
@@ -256,7 +256,7 @@ def setup_archive(self, name=''):
                                           enc_opts, self.sysroot,
                                           self.manifest)
 
-        self.archive.set_debug(True if self.opts.debug else False)
+        self.archive.set_debug(True if self.opts.verbosity > 2 else False)
```
pmoravec, line 259: A monkey questions: why not just `self.archive.set_debug(self.opts.verbosity > 2)` ?

pmoravec (approved):
ACK otherwise.

Issue #2508, opened 2021-04-23
[insights] insights plugin performs undesirable system changes


superjamie:
As discussed in https://github.com/sosreport/sos/issues/1435 having sos perform system changes such as loading kernel modules is undesirable. The `--allow-system-changes` flag was implemented for this purpose.

Since https://github.com/sosreport/sos/commit/0766cc739d424932e18c250c88f0227da06ecaed the insights plugin runs `insights-client --offline` which then runs a large amount of commands (many of which are redundantly already included in other sos plugins).

These commands include `ss_-tupna` which loads the networking diag modules, and `brctl show` which loads the bridge module. I have not audited all commands to determine if others perform system changes which sos promises it does not perform.

The insights plugin requires at least a `SoSPredicate()` around `insights-client --offline` similar to https://github.com/pmoravec/sos/commit/f3b1977d2e4ca4522c4777ba97655cc9a3ddb400 but also adding the `bridge` module. However, that just solves the immediate problem.

The wider issue is that the command `insights-client --offline` is not part of sos, so could theoretically do anything, and sos is not protected from future system changes which such a large-scope command may run.

It's currently running the equivalent of a second "mini-sosreport" inside an existing sosreport. It seems counter to a sos plugin's purpose to run a command which then re-does much of the same things which other sos plugins are already doing in the first place.

My impression of the sos plugin infrastructure is that each plugin should collect data specifically useful to that plugin domain, and not large system-wide commands or commands which are the scope of other plugin domains, as is currently being done here.

I'm not sure the best way to address this. Argubaly `insights-client --offline` should be removed from this plugin altogether, and replaced with specific commands which are useful to the Insights software only?

superjamie:
Discussed with Insights developer Paul Wayper, resolved by https://github.com/sosreport/sos/pull/2509

pmoravec:
Agreed. If some specific command/config for debugging the insights software is required (i.e. test connection to insights server/proxy), let us know and we can add it to the plugin.

PR #2509, opened 2021-04-23
**[insights]** Removing insights-client command as it's superfluous

PaulWay:
The `insights-client` command should not be run within sosreport,
as it collects data sosreport already collects and is not cautious
about how it does so.  Running it is therefore superfluous.

Closes: #2508
Resolves: #2508 

Signed-off-by: Paul Wayper <paulway@redhat.com>

---

### DIFF
--- a/sos/report/plugins/insights.py
+++ b/sos/report/plugins/insights.py
@@ -32,11 +32,6 @@ class RedHatInsights(Plugin, RedHatPlugin):
         else:
             self.add_copy_spec("/var/log/insights-client/insights-client.log")
 
-        # Collect insights-client report data into given dump dir
-        path = self.get_cmd_output_path(name="insights-client-dump")
-        self.add_cmd_output("insights-client --offline --output-dir %s" % path,
-                            suggest_filename="insights-client-dump.log")
-
     def postproc(self):
         for conf in self.config:
             self.do_file_sub(


pmoravec (approved):

TurboTurtle:
We really need to hash out what the future looks like. This was added about a year ago at the request of the Insights team. Since then there's been some discussion about a better way to collect this or similar information for Insights (a few talks with miclark and csams), some kicked off from the very issue this is being removed for; nothing prevents running commands that load kmods or otherwise makes system changes.

I'm fine with removing this collection, but I'd like us to explore the different approaches to still being able to provide Insights with the data that it needs that is missing from any given release of sos.

Please note though, the commit message itself is what needs to be formatted per the contrib guidelines not the PR message (i.e. the closes/resolves tags and such, not the checklist). If you have a moment please copy those bits over to an amended commit message, otherwise I can take care of it when I go to merge this.

TurboTurtle (approved):

PaulWay:
Hi there,
Last time I looked, the Insights client still had the ability to read a sosreport.  It may not have every command in exactly the same format, but it'll give the user 90% of what they want.  We can check with Chris Sams to see what it needs, but that should not be the concern of sosreport.

If the Insights client needs to read something specific from a sosreport I would advise them to submit their own PRs to add those commands, and have each request judged on its own merit.

Hope this helps,

Paul

PR #2510, opened 2021-04-27
[mysql] Update collection paths for Debian/Ubuntu

lathiat:
- All configuration files from /etc/mysql/
- Default error log path: /var/log/mysql/error.log
- Alternative data directory location /var/lib/percona-xtradb-cluster
- gvwstate.dat for galera debugging
- Update the various server package names (though all seem to depend on
  mysql-common anyway)

Signed-off-by: Trent Lloyd <trent.lloyd@canonical.com>

---

### DIFF
--- a/sos/report/plugins/mysql.py
+++ b/sos/report/plugins/mysql.py
@@ -35,7 +35,8 @@ class Mysql(Plugin):
             "/var/log/mysqld.log",
             "/var/log/mysql/mysqld.log",
             "/var/log/mariadb/mariadb.log",
-            "/var/lib/mysql/grastate.dat"
+            "/var/lib/mysql/grastate.dat",
+            "/var/lib/mysql/gvwstate.dat"
         ])
 
         if self.get_option("all_logs"):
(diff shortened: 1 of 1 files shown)


sos/report/plugins/mysql.py
```diff
@@ -36,6 +36,7 @@ def setup(self):
             "/var/log/mysql/mysqld.log",
             "/var/log/mariadb/mariadb.log",
             "/var/lib/mysql/grastate.dat"
+            "/var/lib/mysql/gvwstate.dat"
```
TurboTurtle, line 39: Missing a comma at the end of line 38, hence the implicit concatenation error from LGTM.

sos/report/plugins/mysql.py
```diff
@@ -98,14 +99,25 @@ def setup(self):
 class DebianMysql(Mysql, DebianPlugin, UbuntuPlugin):
 
     packages = (
-        'mysql-server',
+        'mysql-server.*',
```
TurboTurtle, line 102: What are the other package names expected here for mysql-server?

sos/report/plugins/mysql.py
```diff
@@ -98,14 +99,25 @@ def setup(self):
...
+        self.add_copy_spec([
+            "/etc/mysql/",
+            "/var/log/mysql/error.log",
+            "/var/lib/mysql/*.err",
+            "/var/lib/percona-xtradb-cluster/*.err",
+            "/var/lib/percona-xtradb-cluster/grastate.dat",
+            "/var/lib/percona-xtradb-cluster/gvwstate.dat",
+            "/var/lib/percona-xtradb-cluster/innobackup.*.log",
+        ])
+        self.add_cmd_output("du -s /var/lib/percona-xtradb-cluster/*")
```
TurboTurtle, line 120: This looks like it would be better suited for a new `percona` (or `percona_xtradb` if you prefer) plugin.

TurboTurtle (commented):
The 18.04 tests are failing due to a timeout of the `sos report` command within those tests. From the notes above I'm not immediately sure what would be causing the delay there however.

Let's address the other bits first though and see what we can shake out before digging more into the timeouts directly.

sos/report/plugins/mysql.py
lathiat: 'mysql-server' is a virtual package which depends on the current/latest server package version, with 2 package names that could be installed independently, the virtual package may not always be present:
mysql-server-8.0
mysql-server-core-8.0
(there are also versions for -5.6, -5.7, etc)

sos/report/plugins/mysql.py
lathiat: Ultimately any variant stores pretty much the same data and files, just in two possible data directory locations. So a separate plugin would duplicate basically everything except for the data directory path, so that seemed extreme.

I'd prefer to keep a single plugin but I guess I could in theory try to auto detect the data directory path, however, even some versions of percona exist in /var/lib/mysql so this would mostly end up being a stat() on the directory paths and i'm not sure there's ultimately much of a performance win versus just asking for all the various files. do you have any thoughts/precedent for that?

lathiat:
Pushed new commit with the string concatentation/missing comma issue fixed. But after you review my replies to the other questions let me know if you want further updates made.

sos/report/plugins/mysql.py
TurboTurtle: Ack, here is fine.

TurboTurtle (approved):

Issue #2278, opened 2020-10-19
[pulp] Update the plugin for pulp-3 properly


pmoravec:
There were multiple bigger changes in `pulp-3` with respect what `sos report` should collect in the plugin. E.g.:

- mongo->postgres DB change
- pulp celery workers replaced by redis (what to collect as `qpid-stat` replacement?)
- not sure what else changed

This is a tracker issue for collecting such changes.

@goosemania or @bmbouter, could you please review what commands or files are specific to pulp-2 only, and what commands or files should newly be collected (optionally with scrubbing some secrets), due to pulp-2 -> pulp-3 shift? Cf. #2266 / #2277 as some starting point.

The plugin will collect both pulp-2 and pulp-3 specific stuff for legacy reasons, I just need to know what is pulp-2 and/or pulp-3 specific, to be able to prepare a PR.

pmoravec:
Also, what shall trigger *automatic* enablement of the plugin? Just presence of `python3-pulpcore` package isnt sufficient, when `pulpcore` is installed via `pip`.

Is there either a specific command existing exactly when pulp-3 installed?

Or specific file (e.g. `/etc/pulp/settings.py`?) present?

Those checks could be done in sos report well..

PR #2512, opened 2021-04-27
[pulpcore] add plugin for pulp-3

pmoravec:
Pulp-3 / pulpcore as a revolution from pulp-2 needs a separate
plugin, since both plugins have nothing in common and there might
be deployments where is active both pulp-2 and pulp-3.

Resolves: #2278

---

### DIFF
--- a/sos/report/plugins/pulpcore.py
+++ b/sos/report/plugins/pulpcore.py
@@ -1,0 +1,120 @@
+# Copyright (C) 2021 Red Hat, Inc., Pavel Moravec <pmoravec@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, IndependentPlugin
+from pipes import quote
+from re import match
+
+
+class PulpCore(Plugin, IndependentPlugin):
+
+    short_desc = 'Pulp-3 aka pulpcore'
+
+    plugin_name = "pulpcore"
+    commands = ("pulpcore-manager",)
+    files = ("/etc/pulp/settings.py",)
+    option_list = [
+        ('task-days', 'days of tasks history', 'fast', 7)
+    ]
+
+    def parse_settings_config(self):
+        databases_scope = False
+        self.dbhost = "localhost"
+        self.dbport = 5432
+        self.dbpasswd = ""
+        # TODO: read also redis config (we dont expect much customisations)
+        # TODO: read also db user (pulp) and database name (pulpcore)
+        self.staticroot = "/var/lib/pulp/assets"
+        self.uploaddir = "/var/lib/pulp/media/upload"
+
+        def separate_value(line, sep=':'):
+            # an auxiliary method to parse values from lines like:
+            #       'HOST': 'localhost',
+            val = line.split(sep)[1].lstrip().rstrip(',')
+            if (val.startswith('"') and val.endswith('"')) or \
+               (val.startswith('\'') and val.endswith('\'')):
+                val = val[1:-1]
+            return val
+
+        try:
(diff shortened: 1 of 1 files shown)


pmoravec:
This is so far a WIP version of the plugin, that needs some review from pulp devels and some more testing.

@goosemania , could you please (ask for a) review of the plugin?

For testing on a sos-3.9 version, it is sufficient to:
- copy the file to `sos/plugins/pulpcore.py`
- `sed 's/sos.report.plugins/sos.plugins/1'`
- `sed 's/IndependentPlugin/RedHatPlugin, DebianPlugin, UbuntuPlugin/g'

sos/report/plugins/pulpcore.py
```diff
@@ -0,0 +1,121 @@
+# Copyright (C) 2018 Red Hat, Inc., Jake Hunsaker <jhunsake@redhat.com>
```
TurboTurtle, line 1: Hmm. ;-)

sos/report/plugins/pulpcore.py
```diff
@@ -0,0 +1,121 @@
...
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, IndependentPlugin
+from pipes import quote
+from re import match
+
+
+class PulpCore(Plugin, IndependentPlugin):
+    """Pulp-3 aka pulpcore"""
+
```
TurboTurtle, line 18: Needs a `short_desc` class attr

TurboTurtle (commented):
Initial comments below, very minor points. Only other question I have is why `pulpcore` instead of `pulp3`? I don't necessarily mind using the former, I'm just curious as to the reasoning.

sos/report/plugins/pulpcore.py
```diff
@@ -0,0 +1,121 @@
...
+        self.add_copy_spec("/etc/pulp/settings.py")
+
+        _hname = self.exec_cmd('hostname')['output']
+        _hname = _hname.strip()
+        self.add_cmd_output("rq info -u redis://localhost:6379/8",
+                            env={"LC_ALL": "en_US.UTF-8"},
+                            suggest_filename="rq_info")
+        self.add_cmd_output("curl -s https://%s/pulp/api/v3/status/" % _hname,
+                            suggest_filename="pulp_status")
```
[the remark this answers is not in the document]
pmoravec: Yes and no: `curl`-ing localhost fails on SSL certificate/hostname check. But `--insecure` option skips that check, so I will use it.

Technically, the very proper solution is to use `CONTENT_HOST = "my.fqdn"` from `/etc/pulp/settings.py` but 1) we would have to have a failover value and 2) the hostname will very rarely differ from the local host. So till `--insecure` works, I am in favour of `localhost`, for the sake of simplicity.

pmoravec:
> Initial comments below, very minor points. Only other question I have is why `pulpcore` instead of `pulp3`? I don't necessarily mind using the former, I'm just curious as to the reasoning.

No strong argument or opinion here. `pulp-3` is the "branding" name, `pulpcore` is the classes name. Maybe `pulp3` is more informative to users of sosreport, to stress it is the "replacement" of pulp plugin / pulp-2?

IMHO short description should have both terms, plugin name: no much preference (now I tend to `pulp3` bit). If somebody agrees with `pulp3`, let comment it and I will rename the plugin.

evgeni:
We've been using "pulpcore" in various places (Puppet module, Repository name, etc), as we didn't want to hardcode the *3* to differentiate from pulp(2) which is usually referred to as "Pulp".

TurboTurtle (approved):

TurboTurtle:
Merged via 808d9f3

ekohl:
I missed this PR. Is there a good way to stay on top of these (Foreman, Pulp and the subsystems it depends on)?

pmoravec:
> I missed this PR. Is there a good way to stay on top of these (Foreman, Pulp and the subsystems it depends on)?

Yeah I thought we get some feedback before merging the PR, but never mind, we can iterate / fine-tune the plugin further, for sure. Let open an issue if you spot something (and we can make the filename a variable as well, as a side-fix).

ekohl:
Next time, feel free to ping me in a PR.

Issue #1723, opened 2019-07-08
[sosreport] option to ignore unknown plugins


pmoravec:
Automation tools running sosreport nowadays need to distinguish among sos minor versions (or do other tricks) to identify what plugins the sosreport knows. Otherwise, "-o new_plugin" will cause

    a non-existing plugin (new_plugin) was specified in the command line

halt of older sosreport execution.

We should have an option to suppress this behaviour and to simply ignore plugins the (older) sosreport is not aware of. (well that option wont help to older sos versions, but can help to future).

BryanQuigley:
Why not just grep the list of plugins first, this seems to gets a list of plugins (with a couple bits to ignore)
sudo python3 sosreport -l | cut -d' ' -f2

If needed, a more machine-readable version of this list would be my preference (could be reused for travis QA checks as well).

Issue #537, opened 2015-03-27
Can we have an option for --disable-plugin to not kill the command when non-existent or already disabled?


portante:
Lately have been working on RPMs which use sosreport across multiple versions of RHEL and Fedora, with multiple versions of sosreport.

It would really help to start now relaxing the behavior of killing the sosreport command when --disable-plugin specifies a non-existent plugin, or one that is already disabled.

bmr-cymru:
Let me think about this some; I know some users have used error returns as a "plugin detection" mechanism so my main concern would be that we aren't breaking any downstream use cases.

Otherwise it seems a reasonable request.

sandrobonazzola:
I think that relaxing the check will make a bit harder to see that sos changed from a caller program point of view. You'll never see that your super-useful-plugin changed name in extremely-useful-plugin unless you actually check the report content or inspect carefully the execution output. I would prefer to keep existing behavior.

bmr-cymru:
@sandrobonazzola that's exactly my concern - that downstream users could be relying on the "fail if unknown" behaviour.

Personally I would much rather see this tackled by providing an API for programmatic execution of sos than by changing very long standing command line behaviour.

That said if we cannot find any examples of that reliance I would consider making this change to ease transitions for users who do need to interrogate the plugin set.

bmr-cymru:
A possible compromise would be to add a new command line option, say `--mask-plugins`, that accepts a list of plugins (just like `-n` does now), to be "masked off" - but with the difference that masking a non-existent plugin is explicitly not treated as error.

This may also then tie into another feature I've been thinking of for a while - a mechanism to allow administrators to persistently mask certain plugins (Solaris' explorer script has had this for some time).

bmr-cymru:
Obviously this didn't make 3.5, but I'm going to re-flag it for 3.5 as I think it's worth considering.

bmr-cymru:
Still no progress here & getting a bit close to 3.5... I think the final proposal for a `-m/--mask-plugins` is worth considering (a list of plugin names to mask out whether or not they actually exist).

bmr-cymru:
Unfortunately out of time for 3.5 but this will be reconsidered for 3.6 (next April).

PR #2517, opened 2021-04-29
[report] Do not fail on skipping non-existing plugins

TurboTurtle:
Specifying a non-existing plugin to skip is no longer considered a fatal
error. This was previously in place, in part, for third party utilities
calling sos to ensure adherence to known plugins. However, since
`collect` has been integrated into sos natively, this is far less of a
concern.

Specfying non-existing plugins will now only generate a warning message
for plugins wanting to be skipped (`-n`). If non-existing plugins are
specified for enablement via `-e` or `-o`, that will still generate a
fatal error.

Closes: #537
Closes: #1723
Resolves: #2517

---

### DIFF
--- a/sos/report/__init__.py
+++ b/sos/report/__init__.py
@@ -643,7 +643,7 @@ class SoSReport(SoSComponent):
                     del opts[plugname]
             for plugname in opts.keys():
                 self.soslog.error('WARNING: unable to set option for disabled '
-                                  'or non-existing plugin (%s)' % (plugname))
+                                  'or non-existing plugin (%s).' % (plugname))
             # in case we printed warnings above, visually intend them from
             # subsequent header text
             if opts.keys():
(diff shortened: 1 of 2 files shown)


sos/report/__init__.py
```diff
@@ -652,13 +652,18 @@ def _set_tunables(self):
...
             if plugin_name not in self.plugin_names:
                 self.soslog.fatal('a non-existing plugin (%s) was specified '
                                   'in the command line' % (plugin_name))
                 self._exit(1)
+        for plugin in self.opts.skip_plugins:
+            if plugin not in self.plugin_names:
+                self.soslog.warning(
+                    "Requested to skip non-existing plugin '%s'" % plugin
+                )
+        self.ui_log.info('')
```
pmoravec, line 666: Minor comment 2: this will add the empty line unconditionally, to:

    sosreport (version 4.1)


    This command will collect diagnostic and configuration information from
    this Red Hat Enterprise Linux system and installed applications.

    An archive containing the collected information will be generated in
    ..

I would prefer adding the line only when a warning has been printed.

sos/report/__init__.py
```diff
@@ -652,13 +652,18 @@ def _set_tunables(self):
...
                                       self.opts.enable_plugins):
             plugin_name = plugin.split(".")[0]
             if plugin_name not in self.plugin_names:
                 self.soslog.fatal('a non-existing plugin (%s) was specified '
                                   'in the command line' % (plugin_name))
                 self._exit(1)
+        for plugin in self.opts.skip_plugins:
+            if plugin not in self.plugin_names:
+                self.soslog.warning(
+                    "Requested to skip non-existing plugin '%s'" % plugin
```
pmoravec, line 664: Minor comment 1: add a dot to finish the sentence?

sos/report/__init__.py
TurboTurtle: Yeah, I was trying to avoid tracking reports just to print a blank line, but the way the policy banner msg is built and printed makes this difficult to do neatly. I'm leaning towards just dropping the empty line and letting the banner just immediately follow any printed warning (as adding a newline instead makes multiple 'bad' plugins super spaced out)

sos/report/__init__.py
TurboTurtle: Our other plugin enablement warnings don't have a period, so I was matching that. I'll add a period to the other messages as well then for consistency sake ;-)

pmoravec (approved):

Issue #1256, opened 2018-04-01
[RFE][openstack] Collect output of clustercheck in openstack controller


shatadru:
Collect output of clustercheck command in an openstack controller.

clustercheck lets one know status of galera cluster (if  galera cluster is in sync or not)

This is very useful information understanding status of openstack controller. 
This should be installed in OSP controller. 
~~~
rpm -qf /bin/clustercheck
mariadb-galera-server-5.5.42-5.el7ost.x86_64
~~~

To be noted this runs mysql command and has a timeout of 10 seconds by default in case it is hung. 

Reference : https://github.com/olafz/percona-clustercheck

PR #2519, opened 2021-04-29
[openstack_database] Add new plugin

TurboTurtle:
Adds a new plugin for openstack deployments that collects information
about/from the underlying database for the deployment. This plugin
crosses multiple openstack components (for which there is typically one
component for each), since that is preferable to adding a host of
database-related options to each individual plugin.

By default, this plugin will only collect the output of `clustercheck`.
When using either the `dump` or `dumpall` plugin options, this plugin
will also collect a mysql dump of the database; either for specific
tables or all tables present respectively.

Closes: #1256
Closes: #1984
Resolves: #2519

---

### DIFF
--- a/sos/report/plugins/openstack_database.py
+++ b/sos/report/plugins/openstack_database.py
@@ -1,0 +1,91 @@
+# Copyright (C) 2021 Red Hat, Inc., Jake Hunsaker <jhunsake@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+import re
+
+
+from sos.report.plugins import Plugin, RedHatPlugin
+
+
+class OpenStackDatabase(Plugin):
+
+    short_desc = 'Openstack Database Information'
+    plugin_name = 'openstack_database'
+    profiles = ('openstack', 'openstack_controller')
+
+    option_list = [
+        ('dump', 'Dump select databases to a SQL file', 'slow', False),
+        ('dumpall', 'Dump ALL databases to a SQL file', 'slow', False)
+    ]
+
+    databases = [
+        'cinder',
+        'glance',
+        'heat',
+        'ironic',
+        'keystone',
+        'mistral',
+        '(.*)?neutron',
+        'nova.*'
+    ]
+
+    def setup(self):
+
+        in_container = False
+        # determine if we're running databases on the host or in a container
+        _db_containers = [
+            'galera-bundle-.*',  # overcloud
+            'mysql'  # undercloud
+        ]
+
+        for container in _db_containers:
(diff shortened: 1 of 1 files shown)


TurboTurtle:
@camorris-redhat @valleedelisle PTAL. I am basing the container name of  `galera-bundle.*` off my test env which is OSP 16.1, but I am unsure if that name pattern is consistent/stable/etc across versions that this would ship on.

@sosreport/team-ubuntu I'm not sure what the plugin enablement trigger should be. If this is desired for Ubuntu-based OSP deployments as well, please let me know what to look for.

sos/report/plugins/openstack_database.py
```diff
@@ -0,0 +1,83 @@
...
+        'mistral',
+        '(.*)?neutron',
+        'nova.*'
+    ]
+
+    def setup(self):
+
+        # determine if we're running databases on the host or in a container
+        in_container = self.container_exists('galera-bundle-.*')
+        cname = None
```
pmoravec, line 43: Isn't it easier to call directly here:

    cname = self.get_container_by_name('galera-bundle-.*')
    in_container = (cname != None)

One iteration over containers is prevented.

sos/report/plugins/openstack_database.py
```diff
@@ -0,0 +1,83 @@
...
+
+    databases = [
+        'cinder',
+        'glance',
+        'heat',
+        'ironic',
+        'keystone',
+        'mistral',
+        '(.*)?neutron',
+        'nova.*'
```
valleedelisle, line 36: This can be nova_api, nova_cellX, etc.

sos/report/plugins/openstack_database.py
```diff
@@ -0,0 +1,83 @@
...
+    ]
+
+    def setup(self):
+
+        # determine if we're running databases on the host or in a container
+        in_container = self.container_exists('galera-bundle-.*')
+        cname = None
+
+        if in_container:
+            cname = self.get_container_by_name('galera-bundle-.*')
```
valleedelisle, line 46: On the undercloud, the database container is called mysql. The schema inside this database is pretty much the same as the overcloud's

```
[stack@undercloud-0 ~]$ sudo podman ps --filter name=mysql
CONTAINER ID  IMAGE                                                                                      COMMAND      CREATED      STATUS          PORTS  NAMES
4722b2cbe23d  undercloud-0.ctlplane.redhat.local:8787/rh-osbs/rhosp16-openstack-mariadb:16.1_20210322.1  kolla_start  2 weeks ago  Up 2 weeks ago         mysql
```

sos/report/plugins/openstack_database.py
TurboTurtle: Ack, valid point.

sos/report/plugins/openstack_database.py
TurboTurtle: Ack. Since this has to be enabled by a plugin option I have less hesitation than a default collection, given that the container name just being `mysql` is a bit concerning. Will fix this up in a short bit.

pmoravec (approved):

PR #2521, opened 2021-05-04
[foreman] fix unmatched group error in scrubbing installer logs

pmoravec:
Having just one whitespace before "Found key:" failed to match
the first group.

Resolves: #2521

---


pmoravec:
The original regexp worked fine for logs like:

    [DEBUG 2021-04-22T08:23:06 main]           Found key: "foreman_proxy::puppetca_token_ttl" value: 360

but fails on a different (newer) log format:

    2021-04-26 16:32:07 [DEBUG ] [configure] Found key: "foreman_proxy::puppetca_token_ttl" value: 360

TurboTurtle (approved):

Issue #2478, opened 2021-04-08
[cleaner] do we want to obfuscate binary files?


arif-ali:
As discussed on IRC

While obfuscating a sosreport, and running tests for ubuntu packaging, we noticed that when searching for files with the hostname, the journal files would respond back with having the hostname. This is shown in the pastebin [1] below.

The journal files may be a simple one to obfuscate, but there may be other binary files that may need some attention too

[1] https://paste.ubuntu.com/p/h5QbP6DDKg/

TurboTurtle:
Generally speaking, obfuscation of binary files is _extremely_ hard. More specifically, it's almost impossible if we don't know the format.

There's at least the potential for overcoming this with journals, if we can work it out with the systemd team, but that leaves us with a consistency question. Do we (potentially) solve this for journals, and still find ourselves with the same problem with other binary files? I think it'd be an unreasonable position to take that we'd work with every project that generates binary output we collect specifically to undo their binary output. Or do we accept the position of "Nope, can't do any obfuscation of binary files" and go from there?

My gut is saying, we should go with the latter. Even **if** we're able to solve this for journals, there's the maintenance overhead just for those to consider, and again we would still have to wrangle with non-journal binary collections. With that in mind, how do we proceed?

* Do we remove all binary files by default when `--clean` is used and have an option to keep them in?
* Do we have an option to remove them, but leave them in by default?
* Do we keep the same posture we have today (leave them in and have no option) and add disclaimer text for binary files?
* Specific to journals, does it matter to us that these binary files are only collected when `--all-logs` is used? Meaning do we expand the understanding of that option to include "using `--all-logs` may leak sensitive information otherwise obfuscated by `sos clean`"? (Ew.)


Thinking this over some, my preference would be to go with option 1 - remove any binary files encountered during a `clean` run by default. However, if anyone can offer a realistic and reliable way for us to approach binary obfuscation, I'm all ears.

BryanQuigley:
4- All logs and Clean should not be recommended together IMHO.  

--all-logs from a customer perspective means I just want to give the support group everything they could want. 

clean/mask means I am required to be selective and may be required to look through the report manually.

___
Also for the journal files they are specifically designed to be sealed to allow for detecting modification.  One time I could see a customer who usually masks their sosreport use --all-logs is if they are requesting help from a security incident. In that case, we absolutely need them not to use mast.

TurboTurtle:
> 4- All logs and Clean should not be recommended together IMHO.
> 
[...]

PR #2524, opened 2021-05-05
[cleaner] Remove binary files by default

TurboTurtle:
Binary files generally speaking cannot be obfuscated, and as such we
should remove them from archives being obfuscated by default so that
sensitive data is not mistakenly included in an obfuscated archive.

This commits adds a new `--keep-binary-files` option that if used will
keep any encountered binary files in the final archive. The default
option of `false` will ensure that encountered binary files are removed.

The number of removed binary files per archive is reported when
obfuscation is completed for that archive.

Closes: #2478
Resolves: #2524

---


BryanQuigley:
Quick thoughts...
I would just do - anything with .tar anywhere in it.  (We don't care what kind..)

Also, the .journal / .journal~ files. 

Binary data can still be written to files identified as text.

TurboTurtle:
Updated the regex list for auto-removal. Refined the exception handling to just `UnicodeDecodeError` for failure to read binary data from a text file.

TurboTurtle:
Noticed the new option wasn't exposed via `collect` - now it is

pmoravec (approved):
ACK with two minor comments:

- `file_is_binary` is worth having a short explanation that read() method on text file (`open(..'tr')`) fails for binary content - it took me a while to get this.
- worth mentioning this change in default behaviour in release notes of 4.2?

TurboTurtle:
> ACK with two minor comments:
> 
> * `file_is_binary` is worth having a short explanation that read() method on text file (`open(..'tr')`) fails for binary content - it took me a while to get this.

Ack, I'll add a comment before merge.

> * worth mentioning this change in default behaviour in release notes of 4.2?

Oh, absolutely.

PR #2528, opened 2021-05-06
[ssh] configurations may be fragmented

kevin-c-myers:
SSH and SSHD configurations may be fragmented

The SSH configurations may have an include directive.  This
adjustment to the plugin will copy in the main configs and
then read those for any include keywords and attempt to
copy those as well.

Signed-off-by: Kevin C Myers <kevin dot myers at redhat dt com>

---


kevin-c-myers:
Adjusted to only pull *.conf from these directories, we don't need anything else that might be in there.

kevin-c-myers:
WIP on a few minor adjustments to follow based on include keywords in the ssh(d) configs.  Will also squash these into a single commit.

sos/report/plugins/ssh.py
```diff
@@ -19,9 +19,28 @@ class Ssh(Plugin, IndependentPlugin):
...
+            ]
+
+        # Include main config files
+        self.add_copy_spec(sshcfgs)
+
+        # Read configs for any includes and copy those
+        try:
+            for sshcfg in sshcfgs:
+                with open(sshcfg, 'r') as cfgfile:
+                    for line in cfgfile.read().splitlines():
```
TurboTurtle, line 34: This reads the entire content of the file into memory at once. In 99.9% of cases this will be fine for the expected files, but if this instead uses the form `for line in cfgfile`, we'll iterate line-by-line (what `.read().splitlines()` ends up doing anyways) in a safer and more efficient manner.

TurboTurtle (changes requested):
One change noted. Below.

Also, the commit message title needs to have the plugin tag. What you had before, `[ssh] configurations may be fragmented`, is fine.

Please also add a line of `Resolves: #2528` to the commit message.

sos/report/plugins/ssh.py
kevin-c-myers: Thank you for this information.  

I adjusted this to be:
```
   try:
            for sshcfg in sshcfgs:
                for line in open(sshcfg, 'r').read().splitlines():
```
Reran and plugin tested out OK grabbing the extra include files.  Please let me know if I misunderstood or if there are any other ideas on this.  Appreciate the assistance and review!

TurboTurtle: The `.read().splitlines()` is what reads the entire file into memory at once, and is what I was suggesting to change. The `with` context manager is definitely beneficial as it ensures that the open file handle is always closed. This is what I was suggesting:

```python
        try:
            for sshcfg in sshcfgs:
                with open(sshcfg, 'r') as cfgfile:
                    for line in cfgfile:
```

If you use an open file in a `for` loop like this, python uses it as an iterator so we'll only read a single line at a time rather than reading everything into memory at once (`.read()`) and then splitting that into an iterable list (`.splitlines()`).

kevin-c-myers: Thanks for taking time to explain that further.  I didn't catch that it iterates this way but I see what you mean.  This makes sense and is more efficient.  
Made the adjustments and tested it out.  Thanks!

pmoravec:
Sounds good to me. No idea why `test_journal_size_limit` test fails (or why it could be caused by this PR/commit).

TurboTurtle (approved):
LGTM.

CentOS logs test failure seems unrelated, it's been unreliable so I'll need to take another look at that if we see continued failures there.

PR #2530, opened 2021-05-07
[candlepin] move tomcat logs from foreman plugin

pmoravec:
Candlepin as a tomcat application needs to collect the logs, not
foreman.

Ideally, the tomcat plugin should be responsible for colecting the
logs, but doing so would pollute non-candlepin usage (cf. apache
logs in foreman plugin as well).

Resolves: #2530

---


pmoravec:
Kudos to @ewoud for spotting it.

To have the PR 100% correct, shall not we also forbid these logs collection from `tomcat` plugin (cf. with https://github.com/sosreport/sos/blob/master/sos/report/plugins/apache.py#L45-L47)?

ekohl (approved):

TurboTurtle (approved):
I don't think we should forbid these in the `tomcat` plugin, as the names are very generic.

LGTM

PR #2531, opened 2021-05-07
[tests] Add CI testing for foreman 

TurboTurtle:
Adding an integration test for the Foreman project, which feeds into RH Satellite.

These tests will be gated to only run when a PR is posted that affects the plugins related to Foreman/Satellite, or the cirrus config itself.

---

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -20,6 +20,8 @@ env:
     # These are generated images pushed to GCP from Red Hat
     FEDORA_IMAGE_NAME: "f${FEDORA_VER}-server-sos-testing"
     FEDORA_PRIOR_IMAGE_NAME: "f${FEDORA_PRIOR_VER}-server-sos-testing"
+    FOREMAN_CENTOS_IMAGE_NAME: "foreman-24-centos-8-sos-testing"
+    FOREMAN_DEBIAN_IMAGE_NAME: "foreman-24-debian-sos-testing"
 
     # Images exist on GCP already
     CENTOS_IMAGE_NAME: "centos-stream-8-v20210316"
@@ -97,16 +99,16 @@ report_stageone_task:
             BUILD_NAME: ${UBUNTU_PRIOR_NAME}
             VM_IMAGE_NAME: ${UBUNTU_PRIOR_IMAGE_NAME}
     remove_sos_script: &remove_sos |
-        if [ $(command -v apt) ] && [ $(command -v sosreport) ]; then
+        if [ $(command -v apt) ]; then
             apt -y purge sosreport
             apt update
             apt -y install python3-pip
         fi
-        if [ $(command -v dnf) ] && [ $(command -v sos) ]; then
+        if [ $(command -v dnf) ]; then
             dnf -y remove sos
         fi
     setup_script: &setup 'pip3 install avocado-framework'
-    main_script: PYTHONPATH=tests/ avocado run -t stageone tests/
+    main_script: PYTHONPATH=tests/ avocado run -t stageone tests/report_tests tests/vendor_tests/
 
 # IFF the stage one tests all pass, then run stage two for latest distros
 report_stagetwo_task:
@@ -129,4 +131,25 @@ report_stagetwo_task:
             VM_IMAGE_NAME: ${UBUNTU_IMAGE_NAME}
     remove_sos_script: *remove_sos
     setup_script: *setup
-    main_script: PYTHONPATH=tests/ avocado run -t stagetwo tests/
+    main_script: PYTHONPATH=tests/ avocado run -t stagetwo tests/report_tests tests/vendor_tests/
+
+report_foreman_task:
+    skip: "!changesInclude('.cirrus.yml', '**/{__init__,apache,foreman,foreman_tests,candlepin,pulp,pulpcore}.py')"
+    alias: "foreman_integration"
+    name: "Integration Test - Foreman ${FOREMAN_VER}"
+    depends_on: stageone_report
+    gce_instance: &bigvm
+        <<: *standardvm
+        memory: "8Gb"
+    matrix:
+        - env:
+            PROJECT: ${SOS_PROJECT}
+            VM_IMAGE_NAME: ${FOREMAN_CENTOS_IMAGE_NAME}
+            FOREMAN_VER: "2.4 - CentOS Stream 8"
+        - env:
+            PROJECT: ${SOS_PROJECT}
+            VM_IMAGE_NAME: ${FOREMAN_DEBIAN_IMAGE_NAME}
+            FOREMAN_VER: "2.4 - Debian 10"
+    remove_sos_script: *remove_sos
+    setup_script: *setup
+    main_script: PYTHONPATH=tests/ avocado run -t foreman tests/product_tests/foreman/
--- a/tests/product_tests/foreman/foreman_tests.py
+++ b/tests/product_tests/foreman/foreman_tests.py
@@ -1,0 +1,106 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+
+from sos_tests import StageOneReportTest, redhat_only, ubuntu_only
+
+# known values in our CI test images
+FOREMAN_DB_PASSWORD = r'S0Sdb=p@ssw0rd!'
+FOREMAN_ADMIN_PASSWORD = r'S0S@dmin\\p@ssw0rd!'
+CANDLEPIN_DB_PASSWORD = r'S0SKatello%sp@ssw0rd!'
+
+FOREMAN_PASSWORDS = [FOREMAN_DB_PASSWORD, FOREMAN_ADMIN_PASSWORD, CANDLEPIN_DB_PASSWORD]
+
+
+class ForemanBasicTest(StageOneReportTest):
+    """Ensure that a basic execution runs as expected with all TFM related
+    plugins. For the Red Hat tests, it assumes Foreman has been deployed on a
+    Katello system. On Debian systems, these tests are skipped.
+
+    :avocado: tags=foreman
+    """
+
+    sos_cmd = '-v'
+
+    def test_tfm_plugins_ran(self):
+        self.assertPluginIncluded([
+            'apache',
+            'foreman',
+            'postgresql',
+            'puppet',
+            'ruby'
+        ])
+
+    @redhat_only
+    def test_candlepin_plugin_ran(self):
+        self.assertPluginIncluded('candlepin')
+
+    def test_foreman_keys_skipped(self):
+        self.assertFileGlobNotInArchive("/etc/foreman*/*key.pem")
+
+    def test_foreman_database_sizes_collected(self):
+        self.assertFileCollected('sos_commands/foreman/foreman_db_tables_sizes')
+
+    def test_foreman_installer_dirs_collected(self):
+        self.assertFileGlobInArchive("/etc/foreman-installer/*")
+        self.assertFileGlobInArchive("/var/log/foreman-installer/*")
+
+    def test_foreman_production_log_collected(self):
+        self.assertFileCollected('/var/log/foreman/production.log')
+
+    def test_foreman_database_dump_collected(self):
+        self.assertFileCollected('sos_commands/foreman/foreman_settings_table')
+
+    def test_foreman_tasks_csv_collected(self):
+        self.assertFileCollected('sos_commands/foreman/foreman_tasks_tasks')
+
+    def test_proxyfeatures_not_collected(self):
+        self.assertFileGlobNotInArchive("sos_commands/foreman/smart_proxies_features/*")
+
+    def test_foreman_config_postproc_worked(self):
+        self.assertFileNotHasContent('/etc/foreman/database.yml', FOREMAN_DB_PASSWORD)
+
+    def test_foreman_password_postproc_worked(self):
+        for _check in ['/var/log/foreman-installer/foreman.log', '/etc/foreman-installer/scenarios.d/foreman-answers.yaml']:
+            for passwd in FOREMAN_PASSWORDS:
+                self.assertFileNotHasContent(_check, passwd)
+
+    @redhat_only
+    def test_candlepin_table_sizes_collected(self):
+        self.assertFileCollected('sos_commands/candlepin/candlepin_db_tables_sizes')
+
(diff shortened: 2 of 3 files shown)


pmoravec:
Cc @ekohl.

.cirrus.yml
```diff
@@ -119,14 +120,29 @@ report_stagetwo_task:
...
+      #  - env:
+      #      PROJECT: ${UBUNTU_PROJECT}
+      #      BUILD_NAME: ${UBUNTU_NAME}
+      #      VM_IMAGE_NAME: ${UBUNTU_IMAGE_NAME}
     remove_sos_script: *remove_sos
     setup_script: *setup
     main_script: PYTHONPATH=tests/ avocado run -t stagetwo tests/
+
+report_foreman_task:
+    skip: "!changesInclude('.cirrus.yml', '**/__init__.py', '**/{foreman,foreman_tests,candlepin,pulp,pulpcore}.py')"
```
pmoravec, line 136: I would add also `apache.py`, since `foreman` and `pulp*` relies on `httpd` config a lot.

tests/vendor_tests/redhat/foreman_tests.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def test_candlepin_table_sizes_collected(self):
+        self.assertFileCollected('sos_commands/candlepin/candlepin_db_tables_sizes')
+
+    def test_foreman_password_postproc_worked(self):
+        # known value for our sealed test image
+        self.assertFileNotHasContent(
+            '/var/log/foreman-installer/katello.log',
+            'rs74B8sAiaEJtDzdVgu5SAP2TZHQSDS4'
+        )
```
pmoravec, line 53: This is now failing (I *think* as my Satellite instance has `satellite.log` but not `katello.log`) due to still not obfuscated:

    2021-04-26 16:37:11 [INFO  ] [configure] +  password: "QaUerNfEdZg4q78Sqsw6VskRpLYxwp9B"

tests/vendor_tests/redhat/foreman_tests.py
```diff
@@ -0,0 +1,61 @@
...
+        self.assertFileCollected('sos_commands/foreman/foreman_db_tables_sizes')
+
+    def test_foreman_database_dumps_collected(self):
+        self.assertFileCollected('sos_commands/foreman/foreman_settings_table')
+
+    def test_foreman_csv_collected(self):
+        self.assertFileCollected('sos_commands/foreman/foreman_tasks_tasks')
+
+    def test_proxyfeatures_not_collected(self):
+        self.assertFileGlobNotInArchive("*/smart_proxies_features/*")
```
pmoravec, line 43: Nice (to test something that is disabled by default).  Maybe comment it here like "smart_proxies_features" are collected only when `proxyfeatures` is enabled") or similar? It took me a look to the code to get this..

tests/vendor_tests/redhat/foreman_tests.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def test_candlepin_table_sizes_collected(self):
+        self.assertFileCollected('sos_commands/candlepin/candlepin_db_tables_sizes')
```
ekohl, line 46: Foreman doesn't run Candlepin by default. In the Katello scenario it does, but this feels like something that belongs in the test description.

tests/vendor_tests/redhat/foreman_tests.py
```diff
@@ -0,0 +1,61 @@
...
+#
+# See the LICENSE file in the source distribution for further information.
+
+
+from sos_tests import StageOneReportTest, redhat_only
+
+
+class ForemanBasicTest(StageOneReportTest):
+    """Ensure that a basic execution runs as expected with all TFM related
+    plugins
```
ekohl, line 15:
```suggestion
    plugins. It assumes it runs on a Katello installation.
```

tests/vendor_tests/redhat/foreman_tests.py
```diff
@@ -0,0 +1,61 @@
...
+
+    :avocado: tags=foreman
+    """
+
+    sos_cmd = '-v'
+
+    def test_tfm_plugins_ran(self):
+        self.assertPluginIncluded([
```
ekohl, line 23: Should it also assert the Apache and PostgreSQL plugins are loaded?

pmoravec: +1 to both apache and postgresql plugins.

TurboTurtle: Ack

tests/vendor_tests/redhat/foreman_tests.py
TurboTurtle: I've pulled in the updates from that PR, however I had another thought.

I pulled this password directly from the logs after the test setup. Is there a standard way to control this (and other passwords?) in the foreman installation?

I was thinking this might be a pain when it comes to adding foreman 2.5 and later to the test matrix. If we can statically set an array of passwords to well-known values, then it makes maintenance easier and allows us to validate a lot more passwords.

TurboTurtle:
@ekohl what does a Debian-based Foreman deployment look like? Is it against stable? Testing? Is python3 present?

I know one of the pain points you mentioned was sos on Debian is too old. Stables ships 3.6, but sos-4.0 does appear in debian-testing. Doesn't appear to be a 4.1 release yet though even for unstable.

We can definitely add a debian test, but I want to make sure it actually makes sense to do so with the current design of sos and how Foreman gets deployed on that distro.

ekohl:
> @ekohl what does a Debian-based Foreman deployment look like? Is it against stable? Testing? Is python3 present?

My Debian 10 boxes no longer have Python 2 installed, only Python 3.

As for the installation, https://theforeman.org/manuals/2.4/index.html#2.1Installation has the steps. Select Debian 10 in the dropdown and follow them.

tests/vendor_tests/redhat/foreman_tests.py
pmoravec: `foreman-installer` has option `--foreman-db-password`. Optionally, we can fetch it directly from the config file:

    /etc/foreman/database.yml:  password: "QaUerNfEdZg4q78Sqsw6VskRpLYxwp9B"

and then use that value in the `assertFileNotHasContent` test.

tests/product_tests/foreman/foreman_tests.py
```diff
@@ -0,0 +1,87 @@
...
+        self.assertFileGlobNotInArchive("*/smart_proxies_features/*")
+
+    def test_candlepin_table_sizes_collected(self):
+        self.assertFileCollected('sos_commands/candlepin/candlepin_db_tables_sizes')
+
+    def test_foreman_password_postproc_worked(self):
+        # known value for our sealed test image
+        self.assertFileNotHasContent(
+            '/var/log/foreman-installer/katello.log',
+            'rs74B8sAiaEJtDzdVgu5SAP2TZHQSDS4'
```
pmoravec, line 61: +1 for having a variable for the password (can't see the Ewoud's comment on this).

tests/product_tests/foreman/foreman_tests.py
```diff
@@ -0,0 +1,100 @@
...
+#
+# See the LICENSE file in the source distribution for further information.
+
+
+from sos_tests import StageOneReportTest, redhat_only
+
+# known values in our CI test images
+FOREMAN_DB_PASSWORD = r'S0Sdb=p@ssw0rd!'
+FOREMAN_ADMIN_PASSWORD = r'S0S@dmin\\p@ssw0rd!'
+KATELLO_DB_PASSWORD = r'S0SKatello%sp@ssw0rd!'
```
ekohl, line 15: Where is this used? Katello doesn't have a DB. In a Katello install there are 3 DBs: Foreman, Candlepin and Pulp (3). Each has their own password. Katello is a plugin for Foreman so it uses that DB.

tests/product_tests/foreman/foreman_tests.py
```diff
@@ -0,0 +1,100 @@
...
+
+    @redhat_only
+    def test_katello_password_postproc_worked(self):
+        for _check in ['/var/log/foreman-installer/katello.log', '/etc/foreman-installer/scenarios.d/katello-answers.yaml']:
+            for passwd in FOREMAN_PASSWORDS:
+                self.assertFileNotHasContent(_check, passwd)
+
+    @redhat_only
+    def test_foreman_httpd_collected(self):
+        self.assertFileGlobInArchive("/var/log/httpd*/foreman-ssl_*_ssl*log*")
```
ekohl, line 84: Shuold a Debian-only test be added that checks `/var/log/apache2`?

tests/product_tests/foreman/foreman_tests.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def test_proxyfeatures_not_collected(self):
+        self.assertFileGlobNotInArchive("*/smart_proxies_features/*")
+
+    def test_foreman_config_postproc_worked(self):
+        self.assertFileNotHasContent('/etc/foreman/database.yml', FOREMAN_DB_PASSWORD)
+
+    def test_foreman_password_postproc_worked(self):
+        for passwd in FOREMAN_PASSWORDS:
+            self.assertFileNotHasContent('/var/log/foreman-installer/foreman.log', passwd)
```
ekohl, line 70: Should this also check `foreman-answers.yaml` like you did for `katello-answers.yaml` below?

tests/product_tests/foreman/foreman_tests.py
```diff
@@ -0,0 +1,100 @@
...
+    """Enable Foreman/Katello/Candlepin specific options and ensure they are
+    working
+
+    :avocado: tags=foreman
+    """
+
+    sos_cmd = '-v -k foreman.proxyfeatures=on'
+
+    def test_proxyfeatures_collected(self):
+        self.assertFileGlobInArchive("*/smart_proxies_features/*")
```
ekohl, line 100: Looking at the code, I'd expect this to fail on non-Katello installs:
https://github.com/sosreport/sos/blob/61ff5ce165e654a02fe80b9de5ec8e49ed808ec9/sos/report/plugins/foreman.py#L256-L257

Those paths do not exist in non-katello. Technically you need to check it in the database but in most installations you can find `ssl_certificate`, `ssl_ca_file` and `ssl_priv_key` in `/etc/foreman/settings.yaml`.

If this test passes, then perhaps it should also verify that it at least got data for one proxy.

tests/product_tests/foreman/foreman_tests.py
TurboTurtle: This was passed to `--katello-candlepin-db-password`, so I could rename it to `CANDLEPIN_DB_PASSWORD` if that tracks easier.

tests/product_tests/foreman/foreman_tests.py
TurboTurtle: Ack.

tests/product_tests/foreman/foreman_tests.py
TurboTurtle: Can do.

tests/product_tests/foreman/foreman_tests.py
TurboTurtle: That's my fault, that test actually needs to have the plugin dir in the glob check, but also yeah that needs to be tagged `redhat_only`

tests/product_tests/foreman/foreman_tests.py
ekohl: It would. There's also (at least with pulp 3) `--foreman-proxy-content-pulp-db-password`. With Pulp 2 it's more complicated because in the katello/satellite scenario it's `--katello-something` and in the foreman-proxy-content/capsule it's `--foreman-proxy-content-something`.

At least in upstream Pulp 2 is only used in 3.18 and once Katello 4.1 goes GA that's EOL. Right now 4.1 rc1 is in the process of being released. That means that if you care about upstream and the future, the focus should be on Pulp 3.

tests/product_tests/foreman/foreman_tests.py
ekohl: If you follow my suggestion, then you use the configured certificates and it would also work on Debian. Reading something that's configurable from the config files feels more correct to me.

TurboTurtle: Oh, I follow now.

For the moment I'll leave that since @pmoravec is planning on splitting out proxy stuff into a separate plugin. We'll update that check there and then can extend the relevant test(s) back out to debian.

ekohl: That's fair, it's really a bugfix separate from CI testing.

tests/product_tests/foreman/foreman_tests.py
```diff
@@ -0,0 +1,106 @@
...
+        self.assertFileCollected('sos_commands/foreman/foreman_db_tables_sizes')
+
+    def test_foreman_installer_dirs_collected(self):
+        self.assertFileGlobInArchive("/etc/foreman-installer/*")
+        self.assertFileGlobInArchive("/var/log/foreman-installer/*")
+
+    def test_foreman_production_log_collected(self):
+        self.assertFileCollected('/var/log/foreman/production.log')
+
+    def test_foreman_database_dumps_collected(self):
```
pmoravec, line 56: Just bit confusing method name - foreman_settings_table is a dump of one table "only"; maybe `test_foreman_settings_dump_collected`?

tests/product_tests/foreman/foreman_tests.py
```diff
@@ -0,0 +1,106 @@
...
+        self.assertFileCollected('sos_commands/foreman/foreman_settings_table')
+
+    def test_foreman_csv_collected(self):
```
pmoravec, line 59: `s/test_foreman_csv_collected/test_foreman_tasks_csv_collected/` ?

tests/sos_tests.py
```diff
@@ -313,7 +313,9 @@ def assertFileCollected(self, fname):
         :param fname:  The name of the file within the archive
         :type fname:  ``str``
         """
-        if os.path.exists(fname):
+        if fname.startswith(('sos_', '/sos_')):
+            self.assertFileExists(self.get_name_in_archive(fname))
+        elif os.path.exists(fname):
             self.assertFileExists(self.get_name_in_archive(fname))
```
pmoravec, line 319: Can't we simplify this to:

    if fname.startswith(('sos_', '/sos_')) or os.path.exists(fname):
        self.assertFileExists(self.get_name_in_archive(fname))
    else:
        assert True

?

pmoravec (approved):
Commit message has a typo `projcet` and very minor comments that neither block merging, imho.

Good work!

TurboTurtle:
Good eyes, changes made and pushed.

PR #2536, opened 2021-05-11
[Plugin|tags] Add tagging to numerous plugins for Insights

TurboTurtle:
This patchset adds tags to several plugins for aiding the Insights project when it analyzes provided sos archives.

Resolves: #2536
---

### DIFF
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -1515,6 +1515,8 @@ class Plugin(object):
                     # in the corner case we just reached the sizelimit, we
                     # should collect the whole file and stop
                     limit_reached = (sizelimit and current_size == sizelimit)
+
+            _spec_tags = list(set(_spec_tags))
             if self.manifest:
                 self.manifest.files.append({
                     'specification': copyspec,
@@ -1650,7 +1652,7 @@ class Plugin(object):
                        chroot=True, runat=None, env=None, binary=False,
                        sizelimit=None, pred=None, subdir=None,
                        changes=False, foreground=False, tags=[],
-                       priority=10):
+                       priority=10, cmd_as_tag=False):
         """Run a program or a list of programs and collect the output
 
         Output will be limited to `sizelimit`, collecting the last X amount
@@ -1711,6 +1713,10 @@ class Plugin(object):
         :param priority:  The priority with which this command should be run,
                           lower values will run before higher values
         :type priority: ``int``
+
+        :param cmd_as_tag: Should the command string be automatically formatted
+                           to a tag?
+        :type cmd_as_tag: ``bool``
         """
         if isinstance(cmds, str):
             cmds = [cmds]
@@ -1727,7 +1733,7 @@ class Plugin(object):
                                  env=env, binary=binary, sizelimit=sizelimit,
                                  pred=pred, subdir=subdir, tags=tags,
                                  changes=changes, foreground=foreground,
-                                 priority=priority)
+                                 priority=priority, cmd_as_tag=cmd_as_tag)
 
     def add_cmd_tags(self, tagdict):
         """Retroactively add tags to any commands that have been run by this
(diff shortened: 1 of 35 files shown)


sos/report/plugins/__init__.py
```diff
@@ -1516,6 +1516,8 @@ def time_filter(path):
                     # in the corner case we just reached the sizelimit, we
                     # should collect the whole file and stop
                     limit_reached = (sizelimit and current_size == sizelimit)
+
+            _spec_tags = list(set(_spec_tags))
```
pmoravec, line 1520: Why this list unification? (well, it doesnt make sense to have a tag multiple times - but then we can use a set type directly..?)

TurboTurtle: Yeah, it's just to scrub duplicates. The various tag-adding bits allow for strings or lists of strings, so doing the reduction here I feel is easier than transitioning the methods to using sets directly.

sos/report/plugins/__init__.py
```diff
@@ -1912,6 +1922,16 @@ def _collect_cmd_output(self, cmd, suggest_filename=None,
         _tags.append(cmd.split(' ')[0])
         _tags.extend(self.get_tags_for_cmd(cmd))
 
+        if cmd_as_tag:
+            _cmdtag = cmd
+            for s in (' ', '/', '-'):
+                _cmdtag = _cmdtag.replace(s, '_')
+            # reduce duplicates to a single underscore
+            _cmdtag = _cmdtag.replace('__', '_')
```
pmoravec, line 1930: Can't we use `_mangle_command` from https://github.com/sosreport/sos/blob/master/sos/report/plugins/__init__.py#L36, as being used for `outfn` in https://github.com/sosreport/sos/blob/master/sos/report/plugins/__init__.py#L1803 ? Then tag would match the `outfn` filename.

sos/report/plugins/foreman.py
```diff
@@ -326,6 +334,12 @@ class RedHatForeman(Foreman, SCLPlugin, RedHatPlugin):
     apachepkg = 'httpd'
 
     def setup(self):
+
+        self.add_file_tags({
+            '/var/log/foreman-installer/satellite.log.*': 'satellite_log',
```
pmoravec, line 339: Maybe tagname `insataller_satellite_log` or `satellite_installer_log`? As under the term "the" Satellite log, I would imagine rather `/var/log-foreman-production.log`.

sos/report/plugins/processor.py
```diff
@@ -19,6 +19,18 @@ class Processor(Plugin, IndependentPlugin):
     packages = ('cpufreq-utils', 'cpuid')
 
     def setup(self):
+
+        cpupath = '/sys/devices/system/cpu'
+
+        self.add_file_tags({
+            "%s/smt/control" % cpupath: 'cpu_smt_control',
+            "%s/smt/active": 'cpu_smt_active',
```
pmoravec, line 27: ` % cpupath` is missing here.

sos/report/plugins/processor.py
```diff
@@ -19,6 +19,18 @@ class Processor(Plugin, IndependentPlugin):
...
+            "%s/vulnerabilities/.*" % cpupath: 'cpu_vulns',
+            "%s/vulnerabilities/spectre_v2" % cpupath: 'cpu_vulns_spectre_v2',
+            "%s/vulnerabilities/meltdown" % cpupath: 'cpu_vulns_meltdown',
+            "%s/cpu.*/online": 'cpu_cores'
```
pmoravec, line 31: .. and also missing it here :)

sos/report/plugins/__init__.py
TurboTurtle: Hadn't thought of that, but yeah that makes sense.

sos/report/plugins/foreman.py
TurboTurtle: This was to match the name Insights uses for it already. We could do `insights_satellite_log` and `satellite_installer_log` though just as easily.

pmoravec (approved):

PR #2545, opened 2021-05-14
[networking] Add nstat command support

sikarash:
As netstat command is being deprecated,
we need nstat as an alternative to "netstat -s".

Signed-off-by: Seiichi Ikarashi <s.ikarashi@fujitsu.com>

---

### DIFF
--- a/sos/report/plugins/networking.py
+++ b/sos/report/plugins/networking.py
@@ -88,6 +88,7 @@ class Networking(Plugin):
                             root_symlink="netstat")
 
         self.add_cmd_output([
+            "nstat -zas",
             "netstat -s",
             "netstat %s -agn" % self.ns_wide,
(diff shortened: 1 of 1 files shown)


TurboTurtle (approved):
Didn't know `netstat` was getting deprecated, but LGTM.

BryanQuigley (approved):
Thanks for your contribution!

slashdd:
While I have no strong opinion on adding `nstat` to sos, please note that `netstat` is indeed depracate, but its true replacement is `ss`.

sikarash:
I agree ss is known to be a replacement for netstat, however, ss does not display protocol statistics that nstat does.
Since ss and nstat are distributed within the same package iproute, I expect ss will not implement the function in the future and we need to use both of them.

Issue #2423, opened 2021-02-25
[foreman] separate foreman-proxy to a new plugin


pmoravec:
Split foreman plugin into `foreman` and `foreman-proxy`.

While every `foreman` instance is deployed with its integrated `foreman-proxy`, there are external proxies (smart proxy / Capsule) where executing most of `foreman` plugin content is redundant (i.e. most of postgres stuff). Possible negative outcome apart of the redundant commands to be executed on an external proxy: see #2421. Therefore we should split the plugin into two.

I plan to prepare the PR in next weeks.

pmoravec:
I hit an ambiguity problem:

- files should be collected by exactly one plugin
- SmartProxy has various `/var/log/foreman-installer/` or `/var/log/httpd/*` or `/etc/foreman-installer/*` files, i.e. `foreman-proxy` plugin must collect these
- hence `foreman` plugin itself shouldn't collect them
- BUT `foreman` itself does log to the same files and use the same installer configs that we can very hardly separate between the plugins

We can move all of these logs and configs to `foreman-proxy` plugin only, but we risk a user running `sos report -o foreman` woudl lack those important stuff.

Is that an acceptable cons of the split?

pmoravec:
OK, a draft PR proposed, with the notice above applicable.

Some review from @ekohl or @jjansky1 is welcomed.

ekohl:
Does it make sense to introduce a foreman-installer plugin as well to deal with that ambiguity? Within the platform team we've been talking about a standalone scenario for the external PostgreSQL server. Then there is neither Foreman nor Foreman Proxy but understanding the installation when it fails can be useful.

PR #2546, opened 2021-05-16
[foreman-proxy] split the plugin from foreman

pmoravec:
Most of foreman plugin is not applicable to a smart proxy, so
split that functionality to a separate plugin.

Resolves: #2546
Closes: #2423

---

### DIFF
--- a/sos/report/plugins/foreman.py
+++ b/sos/report/plugins/foreman.py
@@ -10,19 +10,19 @@
 # See the LICENSE file in the source distribution for further information.
 
 from sos.report.plugins import (Plugin, RedHatPlugin, DebianPlugin,
-                                UbuntuPlugin, SCLPlugin)
+                                UbuntuPlugin)
 from pipes import quote
 from re import match
 
 
 class Foreman(Plugin):
 
-    short_desc = 'Foreman/Satellite 6 systems management'
+    short_desc = 'Foreman/Satellite systems management'
 
     plugin_name = 'foreman'
     plugin_timeout = 1800
     profiles = ('sysmgmt',)
-    packages = ('foreman', 'foreman-proxy')
+    packages = ('foreman',)
     option_list = [
         ('months', 'number of months for dynflow output', 'fast', 1),
         ('proxyfeatures', 'collect features of smart proxies', 'slow', False),
@@ -65,15 +65,13 @@ class Foreman(Plugin):
 
         self.add_file_tags({
             '/var/log/foreman/production.log.*': 'foreman_production_log',
-            '/var/log/foreman-proxy/proxy.log.*': 'foreman_proxy_log',
-            '/etc/foreman-proxy/settings.yml': 'foreman_proxy_conf',
             '/etc/sysconfig/foreman-tasks': 'foreman_tasks_config',
             '/etc/sysconfig/dynflowd': 'foreman_tasks_config'
         })
 
         self.add_forbidden_path([
-            "/etc/foreman*/*key.pem",
-            "/etc/foreman*/encryption_key.rb"
+            "/etc/foreman/*key.pem",
+            "/etc/foreman/encryption_key.rb"
         ])
 
         _hostname = self.exec_cmd('hostname')['output']
@@ -90,12 +88,9 @@ class Foreman(Plugin):
         # Allow limiting these
         self.add_copy_spec([
             "/etc/foreman/",
-            "/etc/foreman-proxy/",
             "/etc/sysconfig/foreman",
             "/etc/sysconfig/dynflowd",
-            "/etc/smart_proxy_dynflow_core/settings.yml",
             "/etc/default/foreman",
-            "/etc/foreman-installer/",
             "/var/log/foreman/dynflow_executor*log*",
             "/var/log/foreman/dynflow_executor*.output*",
             "/var/log/foreman/apipie_cache*.log*",
@@ -103,20 +98,9 @@ class Foreman(Plugin):
             "/var/log/foreman/db_migrate*log*",
             "/var/log/foreman/db_seed*log*",
             "/var/log/foreman/production.log[.-]*",
-            "/var/log/foreman-proxy/cron*log*",
-            "/var/log/foreman-proxy/migrate_settings*log*",
-            "/var/log/foreman-proxy/proxy*log*",
-            "/var/log/foreman-proxy/smart_proxy_dynflow_core*log*",
             "/var/log/foreman-selinux-install.log",
             "/var/log/foreman-proxy-certs-generate*",
-            "/var/log/foreman-installer/*",
-            "/var/log/foreman-maintain/*",
-            "/var/log/syslog*",
             "/usr/share/foreman/Gemfile*",
-            "/var/lib/puppet/ssl/certs/ca.pem",
-            "/etc/puppetlabs/puppet/ssl/certs/ca.pem",
-            "/etc/puppetlabs/puppet/ssl/certs/{}.pem".format(_hostname),
-            "/var/lib/puppet/ssl/certs/{}.pem".format(_hostname),
             "/var/log/{}*/foreman*".format(self.apachepkg),
             "/var/log/{}*/katello-reverse-proxy_access_ssl.log*".format(
                 self.apachepkg),
@@ -128,10 +112,8 @@ class Foreman(Plugin):
         ])
 
         self.add_cmd_output([
-            'bundle --local --gemfile=/usr/share/foreman/Gemfile*',
             'hammer ping',
             'foreman-selinux-relabel -nv',
-            'foreman-maintain service status',
             'passenger-status --show pool',
             'passenger-status --show requests',
             'passenger-status --show backtraces',
@@ -286,42 +268,17 @@ class Foreman(Plugin):
         return _dbcmd % (self.dbhost, quote(query))
 
     def postproc(self):
-        install_logs = "/var/log/foreman-installer/"
-        satreg = r"((foreman.*)?(\"::(foreman(.*?)|katello).*)?((::(.*)::.*" \
-                 r"(passw|cred|token|secret|key).*(\")?:)|(storepass )" \
-                 r"|(password =)))(.*)"
-        self.do_path_regex_sub(install_logs, satreg, r"\1 ********")
-        # need to do two passes here, debug output has different formatting
-        sat_debug_reg = (r"(\s)+(Found key: (\"(foreman(.*?)|katello)"
-                         r"::(.*(token|secret|key|passw).*)\") value:) "
-                         r"(.*)")
-        self.do_path_regex_sub(install_logs, sat_debug_reg, r"\1 \2 ********")
-        # also hide passwords in yet different formats
-        self.do_path_regex_sub(
-            install_logs,
-            r"(\.|_|-)password(=\'|=|\", \")(\w*)",
-            r"\1password\2********")
-        self.do_path_regex_sub(
-            "/var/log/foreman-installer/foreman-proxy*",
-            r"(\s*proxy_password\s=) (.*)",
-            r"\1 ********")
-        self.do_path_regex_sub(
-            "/var/log/foreman-maintain/foreman-maintain.log*",
-            r"(((passw|cred|token|secret)=)|(password ))(.*)",
-            r"\1********")
         self.do_path_regex_sub(
             "/var/log/%s*/foreman-ssl_access_ssl.log*" % self.apachepkg,
             r"(.*\?(passw|cred|token|secret|key).*=)(.*) (HTTP.*(.*))",
             r"\1******** \4")
-        # all scrubbing applied to configs must be applied to installer logs
-        # as well, since logs contain diff of configs
         self.do_path_regex_sub(
-            r"(/etc/foreman(.*)((conf)(.*)?))|(%s)" % install_logs,
+            r"/etc/foreman/(.*)((conf)(.*)?)",
             r"((\:|\s*)(passw|cred|token|secret|key).*(\:\s|=))(.*)",
             r"\1********")
         # yaml values should be alphanumeric
         self.do_path_regex_sub(
-            r"(/etc/foreman(.*)((yaml|yml)(.*)?))|(%s)" % install_logs,
+            r"/etc/foreman/(.*)((yaml|yml)(.*)?)",
             r"((\:|\s*)(passw|cred|token|secret|key).*(\:\s|=))(.*)",
             r'\1"********"')
 
@@ -329,21 +286,17 @@ class Foreman(Plugin):
 # attr so we can keep all log definitions centralized in the main class
 
 
-class RedHatForeman(Foreman, SCLPlugin, RedHatPlugin):
+class RedHatForeman(Foreman, RedHatPlugin):
 
     apachepkg = 'httpd'
 
     def setup(self):
 
         self.add_file_tags({
-            '/var/log/foreman-installer/satellite.log.*':
-                ['insights_satellite_log' 'satellite_installer_log'],
             '/usr/share/foreman/.ssh/ssh_config': 'ssh_foreman_config',
         })
 
         super(RedHatForeman, self).setup()
-        self.add_cmd_output_scl('tfm', 'gem list',
-                                suggest_filename='scl enable tfm gem list')
 
 
 class DebianForeman(Foreman, DebianPlugin, UbuntuPlugin):
--- a/sos/report/plugins/foreman_installer.py
+++ b/sos/report/plugins/foreman_installer.py
@@ -1,0 +1,98 @@
+# Copyright (C) 2021 Red Hat, Inc., Pavel Moravec <pmoravec@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import (Plugin, RedHatPlugin, DebianPlugin,
+                                UbuntuPlugin)
+
+
+class ForemanInstaller(Plugin, DebianPlugin, UbuntuPlugin):
+
+    short_desc = 'Foreman installer and maintainer'
+
+    plugin_name = 'foreman_installer'
+    profiles = ('sysmgmt',)
+    packages = ('foreman-installer', 'rubygem-foreman_maintain')
+
+    def setup(self):
+        self.add_copy_spec([
+            "/etc/foreman-installer/*",
+            "/var/log/foreman-installer/*",
+            "/var/log/foreman-maintain/*",
+            # specifically collect .applied files
+            # that would be skipped otherwise as hidden files
+            "/etc/foreman-installer/scenarios.d/*/.applied",
+        ])
+
+        # skip collecting individual migration scripts;
+        # .applied file in each dir is still
+        self.add_forbidden_path(
+                "/etc/foreman-installer/scenarios.d/*.migrations/*.rb"
+        )
+
+        self.add_cmd_output([
+            'foreman-maintain service status',
+        ])
+
+    def postproc(self):
+        install_logs = "/var/log/foreman-installer/"
+        logsreg = r"((foreman.*)?(\"::(foreman(.*?)|katello).*)?((::(.*)::.*" \
+                  r"(passw|cred|token|secret|key).*(\")?:)|(storepass )" \
+                  r"|(password =)))(.*)"
+        self.do_path_regex_sub(install_logs, logsreg, r"\1 ********")
+        # need to do two passes here, debug output has different formatting
+        logs_debug_reg = (r"(\s)+(Found key: (\"(foreman(.*?)|katello)"
+                          r"::(.*(token|secret|key|passw).*)\") value:) "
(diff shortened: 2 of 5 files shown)


sos/report/plugins/foreman_proxy.py
```diff
@@ -0,0 +1,123 @@
...
+        self.add_copy_spec([
+            "/etc/foreman-proxy/",
+            "/etc/smart_proxy_dynflow_core/settings.yml",
+            "/etc/foreman-installer/",
+            "/var/log/foreman-proxy/cron*log*",
+            "/var/log/foreman-proxy/migrate_settings*log*",
+            "/var/log/foreman-proxy/proxy*log*",
+            "/var/log/foreman-proxy/smart_proxy_dynflow_core*log*",
+            "/var/log/foreman-selinux-install.log",
+            "/var/log/foreman-installer/",
```
TurboTurtle, line 43: From a non-foreman-user perspective, I would think that installation logs would come from the main foreman plugin rather than one for the proxy.

sos/report/plugins/foreman_proxy.py
```diff
@@ -0,0 +1,123 @@
+# Copyright (C) 2015 Red Hat, Inc., Pavel Moravec <pmoravec@redhat.com>
```
TurboTurtle, line 1: 2021? ;-)

sos/report/plugins/foreman_proxy.py
```diff
@@ -0,0 +1,123 @@
...
+
+    short_desc = 'Foreman SmartProxy systems management'
+
+    plugin_name = 'foreman_proxy'
+    profiles = ('sysmgmt',)
+    packages = ('foreman-proxy', )
+
+    option_list = [
+        ('proxyfeatures', 'list features of the smart proxy', 'slow', False),
+    ]
```
TurboTurtle, line 25: Is it intentional that this option is also left in the foreman plugin? Or should it be removed there and left only in this new proxy plugin? If the latter we'll need to update the test in tests/product_tests/foreman/foreman_tests.py.

TurboTurtle (commented):
Minor feedback.

Should we modify the existing Satellite preset at all here? Like default enabling the proxyfeatures option?

sos/report/plugins/foreman_proxy.py
pmoravec: Yeah that belongs to the concern I raised. Capsule / smart-proxy installation logs appear in this dir, particular filenames overlap with main foreman system, so in order to collect the logfiles also on a Caps/proxy, we need it in `foreman-proxy` plugin. BUT `foreman` plugin not collecting foreman-installer logs sounds.. strange, non-intuitive,.. The same for configs.

sos/report/plugins/foreman_proxy.py
pmoravec: OK, now I updated the name from copy&paste, there is *always* something I can forget on.. :)

pmoravec:
> Minor feedback.
> 
> Should we modify the existing Satellite preset at all here? Like default enabling the proxyfeatures option?

I think it makes sense to Satellite preset to enable foreman-proxy.proxyfeatures, as that queries for the local SmartProxy quite quickly (and can rpovide some valuable info). The foreman.proxyfeatures option would query e.g. 10 external SmartProxies (depending on the deployment), and that can take a while - hence imho worth disabled by default.

I will update Satellite preset that way.

Anyway still I am quite concerned if the (necessary) move of some `foreman` data collection to `foreman-proxy` plugin, if that is a good step or not..

sos/report/plugins/foreman_proxy.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+class RedHatForemanProxy(ForemanProxy, RedHatPlugin):
+
+    apachepkg = 'httpd'
+
+
+class DebianForemanProxy(ForemanProxy, DebianPlugin, UbuntuPlugin):
+
+    apachepkg = 'apache'
```
ekohl, line 121: Apache is not used in a non-Katello installation. Since Katello is not supported on Debian, this feels odd.

sos/report/plugins/foreman.py
```diff
@@ -82,53 +85,29 @@ def setup(self):
...
             "/var/log/foreman/cron*.log*",
             "/var/log/foreman/db_migrate*log*",
             "/var/log/foreman/db_seed*log*",
             "/var/log/foreman/production.log[.-]*",
-            "/var/log/foreman-proxy/cron*log*",
-            "/var/log/foreman-proxy/migrate_settings*log*",
-            "/var/log/foreman-proxy/proxy*log*",
-            "/var/log/foreman-proxy/smart_proxy_dynflow_core*log*",
-            "/var/log/foreman-selinux-install.log",
             "/var/log/foreman-proxy-certs-generate*",
```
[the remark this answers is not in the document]
pmoravec: Yeah, we have somehow merged katello plugin into foreman one. Since katello plugin would be just few logfiles, I *think*.

Anyway, we can also separate katello plugin, if we figure out what precisely falls into what plugins scope.

sos/report/plugins/foreman_proxy.py
```diff
@@ -0,0 +1,123 @@
...
+            "/var/log/foreman-proxy/migrate_settings*log*",
+            "/var/log/foreman-proxy/proxy*log*",
+            "/var/log/foreman-proxy/smart_proxy_dynflow_core*log*",
+            "/var/log/foreman-selinux-install.log",
+            "/var/log/foreman-installer/",
+            "/var/log/foreman-maintain/",
+            "/var/lib/puppet/ssl/certs/ca.pem",
+            "/var/lib/puppet/ssl/certs/{}.pem".format(_hostname),
+            "/etc/puppetlabs/puppet/ssl/certs/ca.pem",
+            "/etc/puppetlabs/puppet/ssl/certs/{}.pem".format(_hostname),
```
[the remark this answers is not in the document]
pmoravec: So there can be a foreman instance without an "internal" foreman-proxy? That would make the split almost impossible (until we duplicate many logs and configs collection to both plugins).

As I think about it, maybe we can have just one plugin and collect there some data (e.g. all DB queries) only when `foreman` package is present?

sos/report/plugins/foreman_proxy.py
```diff
@@ -0,0 +1,123 @@
...
+        self.add_cmd_output([
+            'foreman-maintain service status',
+            'ls -lanR /root/ssl-build',
+        ])
+
+        if self.get_option('proxyfeatures'):
+            _host_f = self.exec_cmd('hostname -f')['output']
+            _host_f = _host_f.strip()
+            _cmd = 'curl --cacert /etc/foreman-proxy/foreman_ssl_ca.pem ' \
+                   'https://%s:9090/features' % _host_f
```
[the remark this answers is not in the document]
pmoravec: Where to get the port and CA filename from generic configs? We can make the command parametrised, likewise e.g. DB queries against localhost/external host.

sos/report/plugins/foreman_proxy.py
pmoravec: What webserver does foreman on Debian without katello use, then? How to identify what webserver is used by any foreman deployment? We should update it accordingly...

ekohl: Foreman uses Apache, but Foreman Proxy is its own webserver. On a Katello deployment it just happens to also deploy a reverse proxy in Apache that is there for Pulp and a reverse proxy to Foreman itself.

sos/report/plugins/foreman_proxy.py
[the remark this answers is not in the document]
ekohl: In https://community.theforeman.org/t/introducing-fp-curl/13958 I linked to some code I wrote for it. https://gist.github.com/ekohl/e22b230798a2553c6e698b817fbb5c9a is the direct link but you may find the discussion around it interesting.

sos/report/plugins/foreman_proxy.py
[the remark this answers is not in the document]
ekohl: > So there can be a foreman instance without an "internal" foreman-proxy? That would make the split almost impossible (until we duplicate many logs and configs collection to both plugins).

Yes, the whole "internal" foreman-proxy is something that is a default but there's no technical reason they need to live on the same machine. With Pulp 2 there was some functionality (exports IIRC) that relies on the same filesystem but with Pulp 3 that's gone.

What I would expect is that the foreman plugin collects all information for the `foreman` package and its plugins. The `foreman_proxy` plugin should collect all information for `foreman-proxy` and its plugins.

There is no overlap in files that they log: `foreman` logs to `/var/log/foreman` and `foreman-proxy` to `/var/log/foreman-proxy`. Both services also have their own posix user + group.

The only overlap that they can have is certificates in use. However, it wouldn't be that bad to collect those twice IMHO.

sos/report/plugins/foreman.py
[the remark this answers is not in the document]
ekohl: I think that Foreman plugins belong in the `foreman` plugin. Foreman Proxy plugins belong in the `foreman_proxy`. However, we also have `foreman-installer` (technically the `foreman-installer-katello` subpackage) and do ship the `foreman-proxy-certs-generate` command with that. AFAIK today that package is installed on both what we call downstream Satellite and Capsule deployments even if should only be used on Satellite installations (today).

pmoravec:
I feel we should stick to one plugin only, as it is very hard to specify the border line what data to collect for `foreman` itself and what for `foreman-proxy` only. So we should rather amend the current plugin by conditions like "collect this or that only when `foreman` is present".

ekohl:
To me it's very clear what Foreman and Foreman Proxy are. I really see them as strictly separate projects. They are both networked services that talk to each other over sockets.

The only shared parts can be that they use the same certificate files and there's the installer.

As for the installer, I still want a scenario that only deploys a PostgreSQL instance (remote DB) without Foreman, Foreman Proxy or any other system. To me, foreman-installer is actually a third project.

What may be a bit ambiguous is the "capsule" part. It can deploy a reverse proxy setup in Apache to reach back to Foreman. I don't consider that part of Foreman Proxy. I really see that as just an Apache vhost and the Apache plugin feels like a more natural place.

pmoravec:
If you can split the current plugin into `foreman`, `foreman-proxy` and also `foreman-installer`, that would properly collect both upstream and downstream files, that would be great. As I hit few common areas that should be collected by both foreman and proxy plugins (maybe they are all from potential installer plugin?).

sos/report/plugins/foreman_proxy.py
```diff
@@ -0,0 +1,123 @@
...
+        self.add_forbidden_path([
+            "/etc/foreman-proxy/*key.pem",
+        ])
+
+        _hostname = self.exec_cmd('hostname')['output']
+        _hostname = _hostname.strip()
+        self.add_copy_spec([
+            "/etc/foreman-proxy/",
+            "/etc/smart_proxy_dynflow_core/settings.yml",
+            "/etc/foreman-installer/",
```
ekohl, line 37: This would live in an installer plugin

sos/report/plugins/foreman_proxy.py
```diff
@@ -0,0 +1,123 @@
...
+            "/var/log/foreman-proxy/cron*log*",
+            "/var/log/foreman-proxy/migrate_settings*log*",
+            "/var/log/foreman-proxy/proxy*log*",
+            "/var/log/foreman-proxy/smart_proxy_dynflow_core*log*",
+            "/var/log/foreman-selinux-install.log",
```
ekohl, line 42: foreman-selinux is **only** for Foreman. There is no Foreman Proxy SELinux policy (it runs unconfined)

sos/report/plugins/foreman_proxy.py
```diff
@@ -0,0 +1,123 @@
...
+        _hostname = self.exec_cmd('hostname')['output']
+        _hostname = _hostname.strip()
+        self.add_copy_spec([
+            "/etc/foreman-proxy/",
+            "/etc/smart_proxy_dynflow_core/settings.yml",
+            "/etc/foreman-installer/",
+            "/var/log/foreman-proxy/cron*log*",
+            "/var/log/foreman-proxy/migrate_settings*log*",
+            "/var/log/foreman-proxy/proxy*log*",
+            "/var/log/foreman-proxy/smart_proxy_dynflow_core*log*",
```
ekohl, line 41: Perhaps it should just capture 
```suggestion
            "/var/log/foreman-proxy/*log*",
```

sos/report/plugins/foreman_proxy.py
```diff
@@ -0,0 +1,123 @@
...
+            "/var/log/foreman-selinux-install.log",
+            "/var/log/foreman-installer/",
+            "/var/log/foreman-maintain/",
```
ekohl, line 44: foreman-maintain is its own tool. Perhaps you can mash it into the installer plugin.

sos/report/plugins/foreman_proxy.py
```diff
@@ -0,0 +1,123 @@
...
+                self.apachepkg),
+            "/var/log/{}*/katello-reverse-proxy_error_ssl.log*".format(
+                self.apachepkg),
+            "/var/log/{}*/error_log*".format(self.apachepkg),
+            "/etc/{}*/conf/".format(self.apachepkg),
+            "/etc/{}*/conf.d/".format(self.apachepkg)
+        ])
+
+        self.add_cmd_output([
+            'foreman-maintain service status',
```
ekohl, line 59: Also foreman-maintain.

sos/report/plugins/foreman_proxy.py
```diff
@@ -0,0 +1,123 @@
...
+            'ls -lanR /root/ssl-build',
```
ekohl, line 60: This is IMHO more related to the installer. Today only the installer uses this.

sos/report/plugins/foreman_proxy.py
```diff
@@ -0,0 +1,123 @@
...
+            _cmd = 'curl --cacert /etc/foreman-proxy/foreman_ssl_ca.pem ' \
+                   'https://%s:9090/features' % _host_f
+            self.add_cmd_output(_cmd, suggest_filename='smart_proxy_features',
+                                timeout=10)
+
+        # collect http[|s]_proxy env.variables
+        self.add_env_var(["http_proxy", "https_proxy"])
+
+    def postproc(self):
+        install_logs = "/var/log/foreman-installer/"
```
ekohl, line 75: Also for an installer plugin IMHO.

ekohl (commented):
You may be able to combine foreman-maintain and foreman-installer in a single plugin.

As someone who strongly dislikes foreman-maintain and always avoids it I really can't tell you what it does and doesn't do. It's also not supported on Debian so I always consider it an optional element that can be ignored while I know this isn't true for downstream.

pmoravec:
A new split to `foreman[|-proxy|-installer]` that is more WIP now (need to resolve the merge conflict and a review if the split is sane).

I dropped the "local smart proxy feature list" command as it would be more complicated than we might gain from it.

pmoravec:
Few changes I should fix in the next iteration:
- https://github.com/sosreport/sos/pull/2546/files#r643061665 to merge `foreman-proxy` logs items in `a_c_s`.
- "Add Red Hat Insights tags for RedHatPlugin only" to move closer to the `RedHatForemanInstaller` plugin
- rename `sat_debug_reg` in a `postproc`

sos/report/plugins/foreman_proxy.py
pmoravec: We might risk that a huge file content (there were such use cases) will cause skipping some other logfile type due to 25MB or 100MB default sizelimit. On the other side, such a risk is small and we migh skip collecting some other logfiles with different pattern that might appear there.  So ACK.

pmoravec:
OK, I should have a final version of the PR.

Could you @ekohl and @jjansky1 pls review?

TurboTurtle (approved):
Plugin changes LGTM.

I would just suggest updating the foreman tests to reflect the new plugins in the expected list, and perhaps a couple collection checks for the collections now handled by different plugins.

pmoravec:
Good point, foreman tests updated.

sos/report/plugins/foreman_installer.py
```diff
@@ -0,0 +1,87 @@
...
+
+    short_desc = 'Foreman installer and maintainer'
+
+    plugin_name = 'foreman_installer'
+    profiles = ('sysmgmt',)
+    packages = ('foreman-installer', 'rubygem-foreman_maintain')
+
+    def setup(self):
+        self.add_copy_spec([
+            "/etc/foreman-installer/",
```
ekohl, line 25: Why does this line not have `/*`? Is that not needed? Something I recently noticed is that backups also include `/etc/foreman-installer/scenarios.d/*.migrations/*.rb` but those are shipped by the RPM. I think it's highly unlikely that it's ever needed. Something you may want to consider. Note that `/etc/foreman-installer/scenarios.d/*.migrations` are actually symlinks to `/usr/share/foreman-installer`.

sos/report/plugins/foreman_installer.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+# Add Red Hat Insights tags for RedHatPlugin only
+
+class RedHatForemanInstaller(ForemanInstaller, RedHatPlugin):
+
+    def setup(self):
+
+        self.add_file_tags({
+            '/var/log/foreman-installer/satellite.log.*':
```
ekohl, line 80: Is this a glob or regex? If it's a regex, then it's ok and will match `/var/log/foreman-installer/satellite.log` itself.

sos/report/plugins/foreman.py
```diff
@@ -90,33 +88,19 @@ def setup(self):
...
-            "/etc/smart_proxy_dynflow_core/settings.yml",
             "/etc/default/foreman",
-            "/etc/foreman-installer/",
             "/var/log/foreman/dynflow_executor*log*",
             "/var/log/foreman/dynflow_executor*.output*",
             "/var/log/foreman/apipie_cache*.log*",
             "/var/log/foreman/cron*.log*",
             "/var/log/foreman/db_migrate*log*",
             "/var/log/foreman/db_seed*log*",
             "/var/log/foreman/production.log[.-]*",
```
ekohl, line 100: Should this be simplified to `/var/log/foreman/*log*`?

sos/report/plugins/foreman.py
```diff
@@ -90,33 +88,19 @@ def setup(self):
...
-            "/var/log/foreman-installer/*",
-            "/var/log/foreman-maintain/*",
-            "/var/log/syslog*",
             "/usr/share/foreman/Gemfile*",
-            "/var/lib/puppet/ssl/certs/ca.pem",
-            "/etc/puppetlabs/puppet/ssl/certs/ca.pem",
-            "/etc/puppetlabs/puppet/ssl/certs/{}.pem".format(_hostname),
-            "/var/lib/puppet/ssl/certs/{}.pem".format(_hostname),
             "/var/log/{}*/foreman*".format(self.apachepkg),
             "/var/log/{}*/katello-reverse-proxy_access_ssl.log*".format(
```
ekohl, line 105: I think these only exist on what we call Capsules in downstream. So the `foreman` package won't be installed on them and probably won't be collected.

ekohl (commented):
I also noticed the `short_desc` mentions Satellite 6. "Soon" we'll have Satellite 7 based on the same stack. Not sure if we should we just make it `Foreman/Satellite` or `Foreman/Satellite 6/7`.

sos/report/plugins/foreman_installer.py
pmoravec: I recall there were some issues where `add_copy_spec` really needs '/*' for dirs, but I dont recall why. Just to be sure, I will add the asterisk there.

To the `*rb` files: indeed, these files shall be redundant. The only harm from them is they occupy some small space, but.. why not forbid collecting them.

The `.applied` symlinks collection: indeed, that will be a replacement of the `*rb` files (with particular info what script was applied). We didnt collect it until now as we dont collect hidden files until explicitly mentioned. Adding it now..

To the "Satellite 6" - good point, removing the six.

sos/report/plugins/foreman_installer.py
pmoravec: Yes, I do see that tag being created.

sos/report/plugins/foreman.py
pmoravec: That would create a conflict with explicitly collecting whole `production.log` without a sizelimit. 

Neither that is ideal (if the logfile has 10GBs (have seen similars..), we do copy it whole. And sometimes even the current logfile isnt enough.. Hard to find a reasonable balance between minimal time of production.log covered and maximal size of the data collected..

sos/report/plugins/foreman.py
pmoravec: Right, so I will move them to the proxy.

sos/report/plugins/foreman_installer.py
ekohl: I do wonder if we should also do the same for `capsule.log`.

TurboTurtle (approved):
LGTM

sos/report/plugins/foreman_installer.py
pmoravec: Afaik Insights does not demand/use it, but we can enhance the tags for future (and also to refer to it as `capsule_installer_log`).

TurboTurtle:
@pmoravec  @ekohl 

Any further concerns for this PR, or anything additional edge cases to consider?

PR #2547, opened 2021-05-17
[report,collect,Plugin] Add global namespaces options

TurboTurtle:
Adds a new `--namespaces` option that controls the number of namespaces to iterate over for anything that calls `get_network_namespaces()` (and in the future, any other `get_*_namespaces()` methods that might be added).

This global option may be overridden by plugin specific options, and plugin options are added for the `ebpf` and `conntrack` plugins. 

Finally, fixup the display of options with magic/non-standard defaults in `--list` output.

---

### DIFF
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -11,8 +11,10 @@
 """ This exports methods available for use by plugins for sos """
 
 from sos.utilities import (sos_get_command_output, import_module, grep,
-                           fileobj, tail, is_executable, path_exists,
-                           path_isdir, path_isfile, path_islink, listdir)
+                           fileobj, tail, is_executable, TIMEOUT_DEFAULT,
+                           path_exists, path_isdir, path_isfile, path_islink,
+                           listdir)
+
 import os
 import glob
 import re
@@ -458,8 +460,8 @@ class Plugin(object):
     archive = None
     profiles = ()
     sysroot = '/'
-    plugin_timeout = 300
-    cmd_timeout = 300
+    plugin_timeout = TIMEOUT_DEFAULT
+    cmd_timeout = TIMEOUT_DEFAULT
     _timeout_hit = False
     cmdtags = {}
     filetags = {}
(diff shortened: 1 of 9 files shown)


sos/report/plugins/__init__.py
```diff
@@ -2843,11 +2848,13 @@ def get_process_pids(self, process):
                 continue
         return pids
 
-    def get_network_namespaces(self, ns_pattern=None, ns_max=0):
+    def get_network_namespaces(self, ns_pattern=None, ns_max=None):
+        if ns_max is None and self.commons['cmdlineopts'].namespaces:
+            ns_max = self.commons['cmdlineopts'].namespaces
```
pmoravec, line 2852: Why the "`and self.commons[..]` option? If this evals to False, then the commons namespace is `None` or `0`. Either value can be assigned to `ns_max`, then, as "this allows the use of both '0' and `None` to mean unlimited".

Or do I skip some corner case..?

pmoravec (approved):
ACK up to the one minor comment (where the proposed code can be worth to be such robust).

sos/report/plugins/__init__.py
TurboTurtle: Remember ns_max is the internally set limit. This allows users to override plugin settings of unlimited collections.

ns_max is `None` and commons is `None` -> ns_max is unlimited
ns_max is `None` and commons is `0` -> ns_max is unlimited, as the user intended by setting it to `0`
ns_max is `None` and commons is > `0` -> ns_max is limited, as the user intended
ns_max is not `None` and commons is `None` -> ns_max is limited by the call, the user never set an option value.

So it looks like yes there is a use case that is getting missed actually. Where ns_max is not None, and commons is set to something else. I'll get this fixed up.

TurboTurtle: Oh, wait - duh. Sorry this escaped me at first since it's been a bit.

This is because of the per-plugin option to set a namespace limit. This allows users set individual plugin namespace limits separate from any global limit set. Using the plugin option is the only way `ns_max` has a non `None` value.

PR #2548, opened 2021-05-17
[rpm] Use query tags and drop use of piping to awk

TurboTurtle:
Removes the use of awk in the `rpm` command that generates the
`installed-rpms` output, instead relying on rpm's builtin query tags.

Further removes the `add_rpm_cmd` method and directly calls
`add_cmd_output` with the proper parameters now.

Resolves: #2548

---

### DIFF
--- a/sos/report/plugins/rpm.py
+++ b/sos/report/plugins/rpm.py
@@ -26,34 +26,25 @@ class Rpm(Plugin, RedHatPlugin):
     def setup(self):
         self.add_copy_spec("/var/log/rpmpkgs")
 
-        def add_rpm_cmd(query_fmt, filter_cmd, symlink, suggest, tags=[]):
-            rpmq_cmd = 'rpm --nodigest -qa --qf=%s' % query_fmt
-            shell_cmd = rpmq_cmd
-            if filter_cmd:
-                shell_cmd = "sh -c '%s'" % (rpmq_cmd + "|" + filter_cmd)
-            self.add_cmd_output(shell_cmd, root_symlink=symlink,
-                                suggest_filename=suggest,
-                                tags=tags)
-
         if self.get_option("rpmq"):
+            rpmq = "rpm --nodigest -qa --qf=%s"
             # basic installed-rpms
-            query_fmt = '"%{NAME}-%{VERSION}-%{RELEASE}.%{ARCH}~~'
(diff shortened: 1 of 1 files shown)


TurboTurtle:
I was originally hoping to also drop the use of the nested shell, but `rpm` doesn't directly provide a way to sort the output. I'm pretty sure I'd have more than a few people beating down my door if I got rid of the sorting that's been in place for years.

sos/report/plugins/rpm.py
```diff
@@ -26,34 +26,25 @@ class Rpm(Plugin, RedHatPlugin):
...
-
         if self.get_option("rpmq"):
+            rpmq = "rpm --nodigest -qa --qf=%s"
             # basic installed-rpms
-            query_fmt = '"%{NAME}-%{VERSION}-%{RELEASE}.%{ARCH}~~'
-            query_fmt = query_fmt + '%{INSTALLTIME:date}\n"'
-
-            filter_cmd = 'awk -F "~~" ' \
-                r'"{printf \"%-59s %s\n\",\$1,\$2}"|sort -V'
+            nevra = '"%-59{NEVRA} %{INSTALLTIME:date}\n"'
```
pmoravec, line 32: The NEVRA format means change like:

    < wireshark-cli-2.6.2-12.el8.x86_64                           Tue 29 Sep 2020 09:48:04 AM CEST
    ---
    > wireshark-cli-1:2.6.2-12.el8.x86_64                         Tue 29 Sep 2020 09:48:04 AM CEST

See the dot -> colon /o\. Checking other format..

TurboTurtle: NEVRA uses the epoch if one is set, otherwise it's skipped and the output for that package is the same as NVRA. If epoch is not useful then sure I can drop it back to NVRA.

pmoravec: My probably main concern here is that `installed_packages` might be used by various tooling I am not aware of, and a change in formatting (even adding some information to the middle of each line) can break the tooling. Until we have pros of the Epoch presence there..

TurboTurtle: Ack, good point. I'll make the change later today.

pmoravec (approved):

PR #2549, opened 2021-05-18
[plugins,tests] Fixup plugin conformance issues and add test

TurboTurtle:
This patchset adds a series of fixups for plugin design conformance, such as fixing malformed tuples, across several plugins and adds a new unit test to ensure future conformance of new plugins.

---

### DIFF
--- a/tests/unittests/conformance_tests.py
+++ b/tests/unittests/conformance_tests.py
@@ -1,0 +1,57 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+import unittest
+
+
+from sos.report.plugins import import_plugin
+from sos.utilities import ImporterHelper
+
+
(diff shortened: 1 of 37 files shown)


tests/unittests/conformance_tests.py
```diff
@@ -0,0 +1,57 @@
...
+                    _attr, tuple,
+                    "%s.%s is type %s" % (plug.__name__, tup, type(_attr))
+                )
+
+    def test_plugin_description_is_str(self):
+        for plug in self.plug_classes:
+            self.assertIsInstance(plug.short_desc, str, "%s name not string" % plug.__name__)
+            # make sure the description is not empty
+            self.assertNotEqual(plug.short_desc, '',
+                                "%s description unset" % plug.__name__)
```
pmoravec, line 41: Will this work (will the test fail) also when `short_desc` will be *missing* in some plugin? (I think so, but unsure)

pmoravec (approved):

pmoravec:
Nice test and fixups.

tests/unittests/conformance_tests.py
TurboTurtle: Yep, in fact that's how I found one of the Openstack plugins didn't get updated to move the docstring to `short_desc` in the original patch for that.

PR #2551, opened 2021-05-19
[utilities,plugins] Provider wrappers for common `os.path` methods

### LINKED ISSUE #494: [plugins] review os.path and other os API uses
```
$ grep -r exists sos-3.2/sos/plugins/
sos-3.2/sos/plugins/cs.py:from os.path import exists
sos-3.2/sos/plugins/cs.py:        if self.is_installed("redhat-cs") or exists("/opt/redhat-cs"):
sos-3.2/sos/plugins/cs.py:        elif self.is_installed("pki-common") or exists("/usr/share/java/pki"):
sos-3.2/sos/plugins/named.py:from os.path import exists, join, normpath
sos-3.2/sos/plugins/named.py:            if exists(cfg):
sos-3.2/sos/plugins/gluster.py:        if not os.path.exists(self.statedump_dir):
sos-3.2/sos/plugins/openshift.py:        return os.path.exists("/etc/openshift/broker.conf")
sos-3.2/sos/plugins/openshift.py:        return os.path.exists("/etc/openshift/node.conf")
sos-3.2/sos/plugins/__init__.py:        if os.path.exists(outfn):
sos-3.2/sos/plugins/__init__.py:                if not os.path.exists(newfn):
sos-3.2/sos/plugins/__init__.py:            return (any(os.path.exists(fname) for fname in self.files) or
sos-3.2/sos/plugins/libvirt.py:        if os.path.exists(self.join_sysroot(libvirt_keytab)):
sos-3.2/sos/plugins/kernelrt.py:    # this file exists only when the realtime kernel is booted
sos-3.2/sos/plugins/openstack_neutron.py:        if not os.path.exists("/etc/neutron/") or self.get_option("quantum"):
sos-3.2/sos/plugins/logs.py:                    or os.path.exists("/etc/rsyslog.conf"):
sos-3.2/sos/plugins/logs.py:        if not os.path.exists(messages) and self.is_installed("systemd"):
sos-3.2/sos/plugins/ds.py:                os.path.exists("/etc/dirsrv"):
sos-3.2/sos/plugins/ds.py:                os.path.exists("/opt/redhat-ds"):
```

We have high-level methods for e.g. package checks. We might consider doing the same for path checks.

TurboTurtle:
Use of `os.path.exists` or similar methods is fairly prevalent in plugins today, however the issue is that these do not account for the use of `--sysroot` (e.g. when sos is run in a container), and expecting plugins to manually check for a sysroot for everything path needing validation in untenable.

These commits add wrapper methods to `sos.utilities`, and then provider helper methods in `Plugin` to make this as easy as possible on plugin authors.

The new plugin method names follow their `os.path` names in a hopefully logical manner, for example `os.path.exists` becomes `self.path_exists`.

For this iteration, plugins using the `exists`, `isfile`, `isdir`, `islink` methods of `os.path`, or `os.listdir`, have been updated to use the new wrapper methods. Many internals of `Plugin` have also been updated. I have not however extended this to either `archive` or the base `SoSReport` or `SoSCollector` methods involving policy/plugin/cluster module discovery as those are not use cases to check for containerization.

 Closes #494

---

### DIFF
--- a/sos/report/plugins/pcp.py
+++ b/sos/report/plugins/pcp.py
@@ -10,7 +10,6 @@
 
 from sos.report.plugins import Plugin, RedHatPlugin, DebianPlugin
 import os
-import os.path
 from socket import gethostname
 
 
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -11,7 +11,8 @@
 """ This exports methods available for use by plugins for sos """
 
 from sos.utilities import (sos_get_command_output, import_module, grep,
-                           fileobj, tail, is_executable)
+                           fileobj, tail, is_executable, path_exists,
+                           path_isdir, path_isfile, path_islink, listdir)
 import os
 import glob
 import re
@@ -1034,13 +1035,13 @@ class Plugin(object):
             reldest = linkdest
 
         self._log_debug("copying link '%s' pointing to '%s' with isdir=%s"
-                        % (srcpath, linkdest, os.path.isdir(absdest)))
+                        % (srcpath, linkdest, self.path_isdir(absdest)))
 
         dstpath = self.strip_sysroot(srcpath)
         # use the relative target path in the tarball
         self.archive.add_link(reldest, dstpath)
 
-        if os.path.isdir(absdest):
+        if self.path_isdir(absdest):
             self._log_debug("link '%s' is a directory, skipping..." % linkdest)
             return
 
@@ -1073,7 +1074,7 @@ class Plugin(object):
 
     def _copy_dir(self, srcpath):
         try:
-            for name in os.listdir(srcpath):
+            for name in self.listdir(srcpath):
                 self._log_debug("recursively adding '%s' from '%s'"
                                 % (name, srcpath))
                 path = os.path.join(srcpath, name)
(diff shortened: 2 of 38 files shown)


pmoravec:
We might also replace `os.path.isfile(path)` in https://github.com/sosreport/sos/blob/master/sos/report/plugins/jars.py#L90 but that wont gain much, since `os.path` is used in the plugin in either way as well.

sos/report/plugins/pcp.py
```diff
@@ -10,7 +10,6 @@
 
 from sos.report.plugins import Plugin, RedHatPlugin, DebianPlugin
 import os
-import os.path
```
pmoravec, line 13: Hmm... why some CI job does not detect this automatically? There is a warning like "unused import specified"..?

TurboTurtle: > We might also replace `os.path.isfile(path)` in https://github.com/sosreport/sos/blob/master/sos/report/plugins/jars.py#L90 but that wont gain much, since `os.path` is used in the plugin in either way as well.

Can't use the helpers there since `is_jar()` is a staticmethod.

> Hmm... why some CI job does not detect this automatically? There is a warning like "unused import specified"..?

I was a bit surprised this wasn't caught either by `flake8`, but I'm not sure why it isn't off hand.

pmoravec (approved):

bmr-cymru:
Missed this before merge but nice improvement/cleanup 👍

PR #2552, opened 2021-05-20
[Policy,redhat] Add support for SFTP uploads, replace legacy RH FTP fallback

TurboTurtle:
This patchset adds support for SFTP uploads, using `pexpect` to handle the connection, auth, and upload monitoring.

Further, the `RedHat` policy is updated to drop the legacy FTP server as a fallback and instead use the new SFTP server if there are any issues using the RH Customer Portal, or if credentials are not supplied.

A new `--upload-protocol` option is added which can be used to skip protocol determination and use a specific protocol. In the case of RHEL systems, this means  using `--upload-protocol sftp` will entirely skip any attempts to use the customer portal even if valid credentials are supplied.

---


pmoravec (approved):
Sounds good! All my testing and playing with the functionality worked well.

PR #2553, opened 2021-05-20
[Plugin,plugins] Allow plugins to influence command execution order

TurboTurtle:
Adds a `priority` parameter to `add_cmd_output()` and `add_journal()`,
that can be used to influence the order in which commands are executed
within a plugin. This allows for plugins to specify that long-running
commands should be run last, regardless of where in the plugin those
commands are defined.

As part of accounting for this, minor fixups in the evaluation of run
times for individual commands and how those are reflected in the
report's manifest are included.

Assigns new `priority` values to known long running commands, to ensure
those commands run last in their respective plugins' execution.

Finally, adds a new test to ensure all of this is and remains working.

---

### DIFF
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -1525,7 +1525,7 @@ class Plugin(object):
     def add_blockdev_cmd(self, cmds, devices='block', timeout=None,
                          sizelimit=None, chroot=True, runat=None, env=None,
                          binary=False, prepend_path=None, whitelist=[],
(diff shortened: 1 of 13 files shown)


TurboTurtle:
Of course, I am happy to update this before merge with any additional requests for specifying a new `priority` value for any current plugins. @sosreport/downstreams

BryanQuigley:
This LGTM and makes sense, but I didn't run through live tests with it (yet).

pmoravec (approved):

PR #2554, opened 2021-05-24
[foreman] Follow sizelimit to foreman maintain and installer logs

pmoravec:
Specifying add_copy_spec to a directory means all files from the
directory are collected. That is ridiculous for foremain-installer and
namely foreman-maintain logs, where we should stick to the size limit by
default. Hence collect files from the dir "specifically".

Resolves: #2554

---


pmoravec:
I added to the same PR yet another "opposite" commit on a similar topic (foreman plugin to collect proper logs), that we discovered just after sending the original PR. Let me know if splitting the PR to 2 is preferred.

TurboTurtle (approved):

PR #2556, opened 2021-05-24
[tests] Improve failed command reporting in test output

TurboTurtle:
Improves error reporting for failed sos commands by logging stderr (or
stdout if stderr is not populated) to the console, which was previously
being truncated by the builtin error handling of avocado. Printed output
is limited to the last 8k to avoid dumping several MBs at a time for
scenarios such as timeouts where command failure may generate
significant logging prior to failing.

Included with this are 2 minor changes to existing tests. First, remove
verbose output from the expected plugins test to reduce otherwise
irrelevant output for command failures. Second limit the number of
plugins run for the LogLevelTest, both to reduce overall run time for a
test where we aren't testing specific plugins and to improve readability
of failures for such a test.

Resolves: #2556

---

### DIFF
--- a/tests/report_tests/basic_report_tests.py
+++ b/tests/report_tests/basic_report_tests.py
@@ -39,7 +39,7 @@ class LogLevelTest(StageOneReportTest):
     :avocado: tags=stageone
     """
 
-    sos_cmd = '-vvv'
+    sos_cmd = '-vvv -o kernel,host,boot,filesys'
 
     def test_archive_logging_enabled(self):
         self.assertSosLogContains('DEBUG: \[archive:.*\]')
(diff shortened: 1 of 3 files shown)


TurboTurtle:
This should also let us eventually track down why we're getting intermittent timeout errors on 18.04 via Cirrus (that I've never been able to reproduce manually).

TurboTurtle:
Well, I wasn't expecting to catch that 18.04 issue quite so fast....nor for it to be almost completely devoid of output:

```
 ERROR:
WARNING: Unable to read configuration file /etc/sos/sos.conf : No such file or directory
sosreport (version 4.1)

ERROR: Timeout exceeded, see output above (300.20 s)
```

So we're just spinning here, until we hit 300 seconds, without having even printed the banner message.

Issue #1399, opened 2018-08-20
[filesys] collecting /proc/*/mountinfo raises AVC denials


pmoravec:
When sosreport attempts to collect `/proc/[0-9]*/mountinfo` via `filesys` plugin (enabled by default) on Enforcing SELinux system, below AVCs are raised:

    type=AVC msg=audit(1534763891.040:3455): avc:  denied  { relabelto } for  pid=12759 comm="sosreport" name="mountinfo" dev="dm-0" ino=15039276 scontext=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 tcontext=system_u:system_r:init_t:s0 tclass=file permissive=0

It is due to the fact sosreport collects any file also _with_ its SELinux context, and the relabel fails.

The files are collected by sosreport, just with wrong label (that shouldnt matter here).

I understand sos should not be allowed to label files to context like e.g. `system_u:system_r:init_t:s0` while the `mountinfo` files can be interesting (they were added in #1212).

There are 3 options:

- leave the harmless AVCs (as we are not interested in the context that relabel fails) - but sosreport shouldnt raise AVCs
- dont collect those files
- collect them without preserving permissions incl. SElinux context (needs new option of `add_copy_spec`)

I personally vote for the 3rd option as the feature makes sense - if we will be interested in files content but not their permissions.

@bmr-cymru , @TurboTurtle : any thoughts?

TurboTurtle:
I agree with option 3. Particularly because these files were requested by our filesystem support team and were mentioned to be "extremely valuable", so I don't think option 2 should be considered even if we don't go with 3.

bmr-cymru:
Simple blacklist (global or per-policy: some discussion going on on IRC now): if it's in `/proc` or `/sys` we need to avoid using `shutil.copystat()` (which attempts to copy xattrs, resulting in AVCs when the containing file system is forbidden by policy from containing that label).

We'll need to grow our own code to copy the remaining file properties (mode, times etc.), but this shouldn't be a big deal (we've been ignoring the exceptions that these errors generate in `sos.archive` for years).

PR #2560, opened 2021-05-26
[archive] skip copying SELinux context for /proc and /sys everytime

pmoravec:
A supplement of #1399 fix, now also for adding strings or special
device files.

Resolves: #2560

---

### DIFF
--- a/sos/archive.py
+++ b/sos/archive.py
@@ -326,6 +326,20 @@ class FileCacheArchive(Archive):
             return None
         return dest
 
+    def _copy_attributes(self, src, dest):
+        # copy file attributes, skip SELinux xattrs for /sys and /proc
+        try:
+            stat = os.stat(src)
+            if src.startswith("/sys/") or src.startswith("/proc/"):
+                shutil.copymode(src, dest)
+                os.utime(dest, ns=(stat.st_atime_ns, stat.st_mtime_ns))
+            else:
+                shutil.copystat(src, dest)
+            os.chown(dest, stat.st_uid, stat.st_gid)
+        except Exception as e:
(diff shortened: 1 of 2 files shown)


TurboTurtle:
LGTM, but is there a (redhat_only for selinux labels) test that we could add that checks to ensure the file ownership/mode/labeling is as we expect it to be?

pmoravec:
Good point we shall have a test case for it. As far as I played with the reproducer, `copystat` first sets file ownership and mode and even *then* SELinux label - which fails (under properly strict policy). So checking ownership/mode/labeling of the collected files will pass every time.

We can check SELinux denials, like:

    type=AVC msg=audit(1622112718.062:16254): avc:  denied  { associate } for  pid=573921 comm="sosreport" name="compact_memory" dev="tmpfs" ino=9780 scontext=system_u:object_r:sysctl_vm_t:s0 tcontext=system_u:object_r:tmpfs_t:s0 tclass=filesystem permissive=0
    type=AVC msg=audit(1622112718.066:16255): avc:  denied  { associate } for  pid=573921 comm="sosreport" name="drop_caches" dev="tmpfs" ino=9790 scontext=system_u:object_r:sysctl_vm_t:s0 tcontext=system_u:object_r:tmpfs_t:s0 tclass=filesystem permissive=0


that are present also in Fedora (at least since `selinux-policy-targeted-3.14.5-46.fc32.noarch` but probably from older). Reproducer:

    TmpDir=$(mktemp -d)
    sosreport -o system --batch -vvv --tmp-dir=$TmpDir

Checking if/how to prepare a test case for this..

TurboTurtle (approved):

PR #2561, opened 2021-05-26
[Plugin] Add container enablement trigger

TurboTurtle:
Adds an enablement trigger tuple, `containers`, that can be used to enable plugins based on the presence of a running container matching any of the names or regexes in the tuple.

Also updates several plugins to make use of this new enablement tuple.

---


TurboTurtle:
Fixed a potential iteration issue for SCL plugins against container names.

PR #2573, opened 2021-06-04
[tests] Add tests for checking console output

TurboTurtle:
This patchset adds a base test case class for checking the output of a given command. As part of this, several items are moved from `BaseSoSReportTest` to the lower level `BaseSoSTest` to make future additions of new test classes easier, and less directly tied to `report`.

Some re-organization of the `tests/` directory structure is done as well to better define what tests are for what component.

---


TurboTurtle:
Oh, duh. The CI images don't all have python3-pexpect installed on them.

Normally I'd say this is a good opportunity to have the StageOne test check that collect is unavailable, and a StageTwo test would ensure pexpect is available and the `--list-options` output is as expected.... but that would break against contributor systems running the stageone test suite when they already have pexpect installed.

Need to think on how best to approach this...

TurboTurtle:
Decided to go with a heavy use of `skipIf` to control test executions based on `python3-pexpect` already being installed. The stagetwo tests will now also install the package as part of their setup, so from a CI perspective we can be sure both the actual `collect` component and the placeholder component outputs are being tested.

PR #2575, opened 2021-06-07
[veritas] Fix "veritas" plugin check_enabled() with correct path

rajanikantchirmade:
Veritas plugin use wrong "/opt/VRTSspt/VRTSexplorer" path.
Instead actual path is "/opt/VRTSspt/VRTSexplorer/VRTSexplorer".

Signed-off-by: Rajanikant Chirmade <rajanikantchirmade@gmail.com>

---

### DIFF
--- a/sos/report/plugins/veritas.py
+++ b/sos/report/plugins/veritas.py
@@ -20,14 +20,19 @@ class Veritas(Plugin, RedHatPlugin):
     # Information about VRTSexplorer obtained from
     # http://seer.entsupport.symantec.com/docs/243150.htm
     option_list = [("script", "Define VRTSexplorer script path", "",
-                    "/opt/VRTSspt/VRTSexplorer")]
+                    "/opt/VRTSspt/VRTSexplorer/VRTSexplorer")]
 
     def check_enabled(self):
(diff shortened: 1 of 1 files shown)


TurboTurtle:
Does this location change at all with Veritas version, or has the plugin just always been wrong? If it's a version thing we'll want to not break collection from older installations.

rajanikantchirmade:
Good point.
To work it with both version old & new,  check_enabled() check for both files. If any one present it is true.

TurboTurtle:
That would solve the enablement, but the plugin option would still point only to the new location, unless it was overridden.

Looking over the docs, it's unclear to me if the existing location was ever correct. I'd be fine with the original form of this PR, where we're just updating the default value of the plugin option, but I just want to be sure that we're not inadvertently stopping collection on older hosts.

TurboTurtle:
Cycling around on this. There's a `flake8` issue to resolve and also a rebase needed for the proposed changes.

arif-ali:
@rajanikantchirmade it's been a while on this PR, would you be updating this at all?

TurboTurtle:
Closing as abandoned.

PR #2585, opened 2021-06-16
[collect,clean,ocp] Enhance OCP cluster collection and archive cleaning

TurboTurtle:
This patchset touches several aspects of `collect` and `clean` that do not solely related to OCP or the (new) cluster profile, however the enhancements added we discovered as needed (or needed to be improved/fixed) as part of further enhancements to the OCP profile. Normally, these would be broken out into separate PRs, however given how linked these are to the ultimate changes being made for OCP collections and cleaning, it would be drastically higher overhead to split these out.

Of particular note, is the fix for not clobbering sos command options between nodes, which was inadvertently using a shared variable scope rather than private copies for each node.

Further, this includes allowing cluster profiles to individually set master/worker options, command env vars, and vastly improved handling of cleaning symlinks (of which OCP collections have a heavy presence of). Additionally, we will now explicitly go through directory name obfuscation rather than relying on it being performed as part of file renaming.

Finally, there is a vast improvement to the detection and enablement of `oc` command output among cluster nodes returned by the OCP cluster profile.

---

### DIFF
--- a/sos/cleaner/__init__.py
+++ b/sos/cleaner/__init__.py
@@ -612,28 +612,55 @@ third party.
         if not filename:
             # the requested file doesn't exist in the archive
             return
-        self.log_debug("Obfuscating %s" % short_name or filename,
-                       caller=arc_name)
         subs = 0
-        tfile = tempfile.NamedTemporaryFile(mode='w', dir=self.tmpdir)
-        with open(filename, 'r') as fname:
-            for line in fname:
-                try:
-                    line, count = self.obfuscate_line(line)
-                    subs += count
-                    tfile.write(line)
-                except Exception as err:
-                    self.log_debug("Unable to obfuscate %s: %s"
-                                   % (short_name, err), caller=arc_name)
-        tfile.seek(0)
-        if subs:
-            shutil.copy(tfile.name, filename)
-        tfile.close()
-        _ob_filename = self.obfuscate_string(short_name)
-        if _ob_filename != short_name:
+        if not os.path.islink(filename):
+            # don't run the obfuscation on the link, but on the actual file
+            # at some other point.
+            self.log_debug("Obfuscating %s" % short_name or filename,
+                           caller=arc_name)
+            tfile = tempfile.NamedTemporaryFile(mode='w', dir=self.tmpdir)
+            with open(filename, 'r') as fname:
+                for line in fname:
+                    try:
+                        line, count = self.obfuscate_line(line)
+                        subs += count
+                        tfile.write(line)
+                    except Exception as err:
+                        self.log_debug("Unable to obfuscate %s: %s"
+                                       % (short_name, err), caller=arc_name)
+            tfile.seek(0)
+            if subs:
+                shutil.copy(tfile.name, filename)
+            tfile.close()
+
+        _ob_short_name = self.obfuscate_string(short_name.split('/')[-1])
+        _ob_filename = short_name.replace(short_name.split('/')[-1],
+                                          _ob_short_name)
+        _sym_changed = False
+        if os.path.islink(filename):
+            _link = os.readlink(filename)
+            _ob_link = self.obfuscate_string(_link)
+            if _ob_link != _link:
+                _sym_changed = True
+
+        if (_ob_filename != short_name) or _sym_changed:
             arc_path = filename.split(short_name)[0]
             _ob_path = os.path.join(arc_path, _ob_filename)
-            os.rename(filename, _ob_path)
+            # ensure that any plugin subdirs that contain obfuscated strings
+            # get created with obfuscated counterparts
+            if not os.path.islink(filename):
+                os.rename(filename, _ob_path)
+            else:
+                # generate the obfuscated name of the link target
+                _target_ob = self.obfuscate_string(os.readlink(filename))
+                # remove the unobfuscated original symlink first, in case the
+                # symlink name hasn't changed but the target has
+                os.remove(filename)
+                # create the newly obfuscated symlink, pointing to the
+                # obfuscated target name, which may not exist just yet, but
+                # when the actual file is obfuscated, will be created
+                os.symlink(_target_ob, _ob_path)
+
         return subs
 
     def obfuscate_string(self, string_data):
--- a/sos/cleaner/__init__.py
+++ b/sos/cleaner/__init__.py
@@ -226,8 +226,7 @@ third party.
         nested_archives = []
         for _file in archive.getmembers():
             if (re.match('sosreport-.*.tar', _file.name.split('/')[-1]) and not
-               (_file.name.endswith('.md5') or
-               _file.name.endswith('.sha256'))):
+                    (_file.name.endswith(('.md5', '.sha256')))):
                 nested_archives.append(_file.name.split('/')[-1])
 
         if nested_archives:
@@ -235,6 +234,8 @@ third party.
             nested_path = self.extract_archive(archive)
             for arc_file in os.listdir(nested_path):
                 if re.match('sosreport.*.tar.*', arc_file):
+                    if arc_file.endswith(('.md5', '.sha256')):
+                        continue
                     self.report_paths.append(os.path.join(nested_path,
                                                           arc_file))
             # add the toplevel extracted archive
--- a/sos/cleaner/__init__.py
+++ b/sos/cleaner/__init__.py
@@ -562,6 +562,11 @@ third party.
                 except Exception as err:
                     self.log_debug("Unable to parse file %s: %s"
                                    % (short_name, err))
+            try:
+                self.obfuscate_directory_names(archive)
+            except Exception as err:
+                self.log_info("Failed to obfuscate directories: %s" % err,
+                              caller=archive.archive_name)
 
             # if the archive was already a tarball, repack it
             method = archive.get_compression()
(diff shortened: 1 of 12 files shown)


sos/cleaner/__init__.py
```diff
@@ -612,28 +612,48 @@ def obfuscate_file(self, filename, short_name=None, arc_name=None):
...
-                    self.log_debug("Unable to obfuscate %s: %s"
-                                   % (short_name, err), caller=arc_name)
-        tfile.seek(0)
-        if subs:
-            shutil.copy(tfile.name, filename)
-        tfile.close()
-        _ob_filename = self.obfuscate_string(short_name)
+        if not os.path.islink(filename):
+            # don't run the obfuscation on the link, but on the actual file
+            # at some other point.
```
pmoravec, line 620: Maybe my bad understanding of the English, but here we do obfuscate the real file content, not the symlink. While I read the comment "run obfuscation on the actual file at some other point, not now"..?

sos/cleaner/parsers/__init__.py
```diff
@@ -102,7 +103,9 @@ def parse_string_for_keys(self, string_data):
         :returns: The obfuscated line
         :rtype: ``str``
         """
-        for key, val in self.mapping.dataset.items():
+        for pair in sorted(self.mapping.dataset.items(), reverse=True,
+                           key=lambda x: len(x[0])):
+            key, val = pair
```
pmoravec, line 108: This commit does not work in the scenario where `sos_commands/host/hostnamectl_status` has:

        Virtualization: kvm

When I run:

    python3 bin/sos report -o host --build --batch --clean --keywords Virt,Virtualization

I get:

        obfuscatedword2ualization: kvm

instead of:

        obfuscatedword3: kvm

sos/cleaner/__init__.py
```diff
@@ -656,6 +658,27 @@ def obfuscate_file(self, filename, short_name=None, arc_name=None):
 
         return subs
 
+    def obfuscate_directory_names(self, archive):
```
pmoravec, line 661: We need to correct symlinks to the just-obfuscated directory names. Reproducer:

    mkdir /var/foo
    date > /var/foo/date.txt
    ln -s /var/foo /var/barlink

Add to e.g. `qpid` plugin collection of `/var/barlink`, and then:

    python3 bin/sos report -o qpid --build --batch --clean --keywords foo,bar

it leaves:

1) `var/barlink` with not obfuscated symlink
2) the symlink is broken:

    # file /var/tmp/sosreport-host0-2021-06-16-daowtyx/var/barlink
    /var/tmp/sosreport-host0-2021-06-16-daowtyx/var/barlink: broken symbolic link to foo
    #

Sadly I am not aware of a simple solution to fix both issues when having a chain of symlinks, though.. (or maybe just dummy "traverse all symlinks in the archive and apply obfuscation to source and destination once" does work well in all cases..?).

sos/collector/clusters/ocp.py
```diff
@@ -106,4 +123,37 @@ def check_node_is_master(self, sosnode):
         return 'master' in self.node_dict[sosnode.address]['roles']
 
     def set_master_options(self, node):
-        node.opts.enable_plugins.append('openshift')
+        node.enable_plugins.append('openshift')
```
pmoravec, line 126: Good point (from the other commit) :)

pmoravec:
My comments inline. Just the 2 factual (obfuscation ordering counterexample, symlinks obfuscation missing/broken) and one to the code comment.

sos/cleaner/__init__.py
TurboTurtle: From `Archive.get_file_list()` we'll end up iterating over _both_ the link and the real file the link points to. Since there are potentially 2 filenames to obfuscate (both real and the link), skip the real file content obfuscation until our iteration hits the real file in the file list.

sos/cleaner/parsers/__init__.py
TurboTurtle: Hmm, that change is supposed to address that exact scenario. This will need some further digging.

sos/cleaner/__init__.py
TurboTurtle: The symlink name obfuscation should be accounting for the to-be-obfuscated directories:

```python
                # generate the obfuscated name of the link target
                _target_ob = self.obfuscate_string(os.readlink(filename))
                # create the newly obfuscated symlink, pointing to the
                # obfuscated target name, which may not exist just yet, but
                # when the actual file is obfuscated, will be created
                os.symlink(_target_ob, _ob_path)
```

The testing for this was a highly populated `/var/log/containers` with links pointing to `/var/log/pods/` directories that include the hostname as part of the directory name. The symlinks are actually created as broken until `obfuscate_directory_names()` comes through and renames those directories under `/var/log/pods/`, at which point the links become valid.

Do you see this happening for any real collections?

sos/cleaner/parsers/__init__.py
TurboTurtle: Oh! I see why. The keyword parser overrides `parse_line()` without eventually `super()`ing back into the base method.

That should be a quick fix, and is most likely the reason for the other problem below.

sos/cleaner/__init__.py
TurboTurtle: Ah, got it. The problem is that dir-level symlinks will not pass `isfile()` like other symlinks, so we were skipping over these types of symlinks entirely.

I have this fixed locally and will push an update shortly.

sos/cleaner/parsers/__init__.py
TurboTurtle: This should be resolved in the most recent branch update.

sos/cleaner/__init__.py
TurboTurtle: This should also be fixed in the most recent branch update.

pmoravec (approved):
Sounds good to me!

(just very unrelated observation from my tests: cleaner does obfuscate also "Linux" :) - even current master branch, not relevant to this patch-set; I just discovered it when playing with / testing the PR)

TurboTurtle:
> Sounds good to me!
> 

\o/ yay! There's so many edge cases surrounding these commits that I'm sure there will be some edge-of-an-edge case that we'll need to revisit at some point, but glad to have the extra confidence from more eyes.

> (just very unrelated observation from my tests: cleaner does obfuscate also "Linux" :) 

Meaning if there is a "linux" user present? Or is something else triggering the obfuscation of that string? If an encountered FQDN had a hostname of 'linux' that would also do it. We should probably make a note to make a "global excludes" list that prevents any obfuscation of items like that at some point.

TurboTurtle:
Found one minor nitpick, addressing locally before merge.

pmoravec:
> > Sounds good to me!
> 
> \o/ yay! There's so many edge cases surrounding these commits that I'm sure there will be some edge-of-an-edge case that we'll need to revisit at some point, but glad to have the extra confidence from more eyes.

I would expect some edge-cases to happen as well, since the PR has many corner cases that migh exhibit some issue we dont see neither tested. But as much as I reviewed it, I havent found anything (other than reported or denied by myself as a false alarm).

> 
> > (just very unrelated observation from my tests: cleaner does obfuscate also "Linux" :)
> 
> Meaning if there is a "linux" user present? Or is something else triggering the obfuscation of that string? If an encountered FQDN had a hostname of 'linux' that would also do it. We should probably make a note to make a "global excludes" list that prevents any obfuscation of items like that at some point.

I take this back. I had to manually add that keyword to obfuscate, in past, for some testing. Since I see it in `/etc/sos/cleaner/default_mapping` and when I delete this file, I cant reproduce this further more.

TurboTurtle:
Ok so after many, many tests I think I've nailed the big-ticket "gotcha" items that can completely blow up our obfuscation attempts (like a mangled command name that gets extra characters brought in via a word-boundary regex that in turn reduces down to a hostname of nothing (literally `''`)) in a way that is both more robust and more stable, if at the cost of a bit of efficiency, but overall performance seems the same.

However, on the note of performance I've found something else. The `TarFile` module absolutely chokes when doing multiple very-large `extractall()` calls in parallel. So if we have a few sos archives that uncompresses to over 1GB and has say 70K+ entities inside of it (so, most OCP archives that collect API data), what we end up with is the extraction processes will continually become slower and slower, until we're barely extracting 1MB of data per minute for each archive. The workaround to this is to just extract one archive at a time (`--jobs `). In the mean time I'll be doing some more performance tests and bringing this to the python interpreter team and/or the upstream python bug tracker.

pmoravec:
I understood you discovered this performance issue when testing this PR. But I dont much understand how (or if) is the performance problem related to the PR? Or does it occur on sos upstream as well?

TurboTurtle:
With large enough archives, this affects already released versions and current master. It seems that we just haven't been doing testing against collections large enough to surface the issue yet.

pmoravec:
patchset sounds good to me, nice catch-ups in the latest PR.

One scenario remains unfixed, imho not required to be covered in this already-huge patchset: obfuscating broken symlink names does not work "at all" - neither symlink source or (broken) destination is obfuscated. Reproducer:

    rm -rf /etc/qpid/qpidd.conf /nonsense
    ln /nonsense /etc/qpid/qpidd.conf
    python3 bin/sos report -o qpid --build --batch --clean --keywords qpidd

This will generate:

    /var/tmp/sosreport-host0-2021-06-21-ufajopb/etc/qpid/qpidd.conf:                                 broken symbolic link to ../../nonsense

But as I wrote, I would resolve this in a new PR rather.

PR #2593, opened 2021-06-23
Sos pmoravec rhui fixups

pmoravec:
Two commits patchset that fixes some findings from using `pulpcore` and `rhui` plugins on a real environment.

Individual commit messages provide more descriptive explanation.

Resolves: #2593

---

### DIFF
--- a/sos/report/plugins/rhui.py
+++ b/sos/report/plugins/rhui.py
@@ -27,6 +27,7 @@ class Rhui(Plugin, RedHatPlugin):
             "/var/log/rhui-subscription-sync.log",
             "/var/cache/rhui/*",
             "/root/.rhui/*",
+            "/etc/ansible/facts.d/rhui_*.fact",
         ])
         # skip collecting certificate keys
         self.add_forbidden_path("/etc/pki/rhui/**/*.key", recursive=True)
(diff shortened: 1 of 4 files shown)


sos/report/plugins/rhui.py
```diff
@@ -27,15 +27,23 @@ def setup(self):
...
             "/var/cache/rhui/*",
             "/root/.rhui/*",
+            "/etc/ansible/facts.d/rhui_*.fact",
         ])
         # skip collecting certificate keys
-        self.add_forbidden_path("/etc/pki/rhui/*.key")
+        self.add_forbidden_path([
+            "/etc/pki/rhui/*.key",
+            "/etc/pki/rhui/*/*.key",
+            "/etc/pki/rhui/*/*/*.key",
```
TurboTurtle, line 36: Should we instead be modifying `add_forbidden_path` to recurse into directories with `glob`?

If we pass `recursive=True` to `glob` [here](https://github.com/sosreport/sos/blob/master/sos/report/plugins/__init__.py#L1205), then this could be:

```python
self.add_forbidden_path("/etc/pki/rhui/**/*.key")
```

Which would match all of this in addition to any potential new subdirs that may be added by RHUI later.

pmoravec: Good point, I will update the PR accordingly.

TurboTurtle (approved):

PR #2595, opened 2021-06-23
[tests] Add full execution clean test

TurboTurtle:
Adds a full `sos report --clean` execution test, that is one that does not limit what the report is gathering like the current tests do. Then, inspect the contents of the archive more deeply, ensuring the local hostname and primary IP address do not appear anywhere within the archive. 

Also includes two fixes found during test case development. First, catch a json decoding exception like we do a non-existing file exception. Second, allow test cases to define their own timeout if needed, still defaulting back to 300 seconds.

---

### DIFF
--- a/sos/cleaner/__init__.py
+++ b/sos/cleaner/__init__.py
@@ -129,6 +129,16 @@ class SoSCleaner(SoSComponent):
                 self.log_error(
                     "ERROR: map file %s does not exist, will not load any "
                     "obfuscation matches" % self.opts.map_file)
+        else:
+            with open(self.opts.map_file, 'r') as mf:
+                try:
+                    json.load(mf)
+                except json.JSONDecodeError:
+                    self.log_error("ERROR: Unable to parse map file, json is "
+                                   "malformed. Will not load any mappings.")
+                except Exception as err:
+                    self.log_error("ERROR: Could not load '%s': %s"
+                                   % (self.opts.map_file, err))
 
     def print_disclaimer(self):
         """When we are directly running `sos clean`, rather than hooking into
--- a/tests/cleaner_tests/full_report_run.py
+++ b/tests/cleaner_tests/full_report_run.py
@@ -1,0 +1,86 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+import json
+import re
+
+from avocado.utils import process
+from sos_tests import StageTwoReportTest
+
+
+class FullCleanTest(StageTwoReportTest):
+    """Run an unrestricted report execution through sos clean, ensuring that
+    our obfuscation is reliable across arbitrary plugin sets and not just the
+    'main' plugins that tend to collect data needing obfuscation
+
+    :avocado: tags=stagetwo
+    """
+
+    sos_cmd = '-v --clean'
+    sos_timeout = 600
+    # replace with an empty placeholder, make sure that this test case is not
+    # influenced by previous clean runs
+    files = ['/etc/sos/cleaner/default_mapping']
+
+    def _grep_for_content(self, search):
+        """Call out to grep for finding a specific string 'search' in any place
+        in the archive
+        """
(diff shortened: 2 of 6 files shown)


TurboTurtle:
I'm not able to reproduce the `RuntimeError` manually on GCE instances, so I'm pushing a temporary commit to this branch to enable avocado debugging, which will be removed prior to merge.

TurboTurtle:
Ok, so after a lot more digging, I've found the culprit for these timeouts - it's the random data inserted into the log from the `LogSizeLimitTest` in stagetwo. It trips up the compression algorithms - though **significantly** moreso `xz`.

In my testing, `gzip` will finish compressing the archive after ~5 or so minutes. That's just compression time, so that plus the time the generate and clean the report will still send it over the timeout. `xz` on the other hand does not do well _at all_ and takes upwards of 15 minutes on the same data.

All of this boils down to, how do we handle the situation here?

- An immediate answer is to change the order in which the tests get run by avocado. Specify the `report_tests` last, so that the logs test runs after the cleaner tests. It's effective to a point, but as we build out more stagetwo tests, we'll likely hit this again at some point.
- Drop the `LogSizeLimitTest` test case, thus avoiding the problematic situation altogether. However, we would no longer be testing our journal size limiting which I think is important.
- Increase the timeout of this test more and force the use of `gzip` - however this could be a recurring theme as we add more stagetwo tests.

...and I don't have any other immediate solutions. Ideas? @sosreport/downstreams

BryanQuigley:
This is related to the random failures with 18.04 or does this effect all distros?

I don't think I've ever seen this happen on a real system - so making it better match a real system might help. I tried to play around with making the journal input less random - but I haven't really reduced the time much.

TurboTurtle:
> This is related to the random failures with 18.04 or does this effect all distros?
> 

This is all distros on the stage 2 tests. The 18.04 timeouts are different.

> I don't think I've ever seen this happen on a real system - so making it better match a real system might help. I tried to play around with making the journal input less random - but I haven't really reduced the time much.


Yeah, that's the problem. I can't think of a way to have the ephemeral test VMs mimic real systems in this capacity for large journals.

TurboTurtle:
I pushed a branch update that re-orders the component test execution, so that `clean` tests run before `report` tests. This is a workaround to the problem above, but not really a solution. However, I've not been able to think of any other approach that provides a better end result.

For now, I suggest we carry this change and re-visit if our test suite builds out to the point that this issue gets hit again.

sos/cleaner/__init__.py
```diff
@@ -129,6 +129,16 @@ def validate_map_file(self):
                 self.log_error(
                     "ERROR: map file %s does not exist, will not load any "
                     "obfuscation matches" % self.opts.map_file)
+        else:
+            with open(self.opts.map_file, 'r') as mf:
+                try:
+                    json.load(mf)
```
pmoravec, line 135: It is pitty we load+parse the `json` file repeatedly (here just for a check and to throw away its content, and in https://github.com/sosreport/sos/blob/master/sos/cleaner/parsers/__init__.py#L60 to really store the `json` for each and every parser individually). If the file becomes huge, this costs us a bit (still irelevant low time compared to the execution time of cleaner).

tests/cleaner_tests/full_report_run.py
```diff
@@ -0,0 +1,86 @@
...
+class FullCleanTest(StageTwoReportTest):
+    """Run an unrestricted report execution through sos clean, ensuring that
+    our obfuscation is reliable across arbitrary plugin sets and not just the
+    'main' plugins that tend to collect data needing obfuscation
+
+    :avocado: tags=stagetwo
+    """
+
+    sos_cmd = '-v --clean'
+    sos_timeout = 600
```
pmoravec, line 25: Will 10 minutes suffice? As I see some CI jobs taking 10m now (well they run various tests and not just one, but afaik much limited)?

Either way, we can increase it any time once we realize it is too low.

pmoravec (approved):

tests/cleaner_tests/full_report_run.py
TurboTurtle: Yeah, the 10m is the whole job execution time. With re-ordering the tests, the `clean` tests are executing much faster due to not consuming the garbage from the stage 2 logs test. But absolutely we can adjust this as we need going forward.

PR #2598, opened 2021-06-29
[cleaner] Add error message when using empty mappings file

jcastill:
This patch solves the situation where we attempt to
use an empty json file for cleaner matches. Before this
it issues the following error:

Could not initialize 'clean': Expecting value: line 1
column 1 (char 0)

With the patch, it prints an error message similar to
the one we use when there's no mapping file.

Resolved: RHBZ#1929172
Resolved: #2598 

---

### DIFF
--- a/sos/cleaner/__init__.py
+++ b/sos/cleaner/__init__.py
@@ -129,6 +129,11 @@ class SoSCleaner(SoSComponent):
                 self.log_error(
                     "ERROR: map file %s does not exist, will not load any "
                     "obfuscation matches" % self.opts.map_file)
+        if os.stat(self.opts.map_file).st_size == 0:
+            self.log_error(
+                "ERROR: map file %s exists, but it is empty, will not "
+                "load any obfuscation matches" % self.opts.map_file)
+            self.opts.map_file = None
 
     def print_disclaimer(self):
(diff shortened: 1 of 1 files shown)


jcastill:
I'm not really sure about reassigning None to self.opts.map_file, but without this, the code kept using it anyway. Is there any other way to do this, more elegant?

TurboTurtle:
Another approach to this is in #2595 - https://github.com/sosreport/sos/pull/2595/commits/2f2dc77cfc9b3b3a9084e4ff533af944687e7ee8

In that case we're doing a pre-emptive attempt to read the (expected) json, which will also handle empty files. We do end up re-iterating over the attempt to read the file for the parsers, which is where the exception handling in `parsers/__init__.py` comes in, however this is necessary because we can't set the `map_file` option to `None`, as we will later attempt to write the discovered mappings to that file name. This is done so that we can transparently write a fresh `default_mapping` on the first use of `sos clean`.

I believe the tests are failing here because `clean` aborts when trying to effectively `open(None, 'w')` when it goes to write the contents of the discovered mappings out to disk.

TurboTurtle:
Closing this out per #2595 which added a similar message. If there is further concern or warnings that need to be displayed, please feel free to open a new PR or re-open this one.

PR #2604, opened 2021-07-01
kata_containers: Update to the 2.x release

fidencio:
Kata Containers 1.x, which is the base for this plugin, has reached its
EOL on May 12th 2021*.

Based on that, let's update the sos plugin to match the current Fedora
RPM (kata-containers, instead of kata-runtime), and set the identifiers
we'd be interested in, which are the crio and containerd ones.

Currently, kata-containers can only be used in the kubernetes world and
it'll always be spawn either by CRI-O or containerd, and then it'll log
to the journal using their identifier.

*: https://medium.com/kata-containers/about-kata-containers-releases-8b3d91d1166e

Signed-off-by: Fabiano Fidêncio <fidencio@redhat.com>

---

### DIFF
--- a/sos/report/plugins/kata_containers.py
+++ b/sos/report/plugins/kata_containers.py
@@ -15,7 +15,7 @@ class KataContainers(Plugin, IndependentPlugin):
 
     plugin_name = 'kata_containers'
     profiles = ('system', 'virt', 'container')
-    packages = ('kata-runtime',)
+    packages = ('kata-containers',)
 
     def setup(self):
         self.limit = self.get_option('log_size')
@@ -42,10 +42,5 @@ class KataContainers(Plugin, IndependentPlugin):
                     config_files.add(config)
 
             self.add_copy_spec(config_files)
-
-        self.add_journal(identifier='kata-proxy')
-        self.add_journal(identifier='kata-shim')
-        self.add_journal(identifier='kata-runtime')
-        self.add_journal(units='kata-ksm-throttler')
 
 # vim: set et ts=4 sw=4 :


sos/report/plugins/kata_containers.py
```diff
@@ -43,9 +43,7 @@ def setup(self):
 
             self.add_copy_spec(config_files)
 
-        self.add_journal(identifier='kata-proxy')
-        self.add_journal(identifier='kata-shim')
-        self.add_journal(identifier='kata-runtime')
-        self.add_journal(units='kata-ksm-throttler')
+        self.add_journal(identifier='crio')
```
TurboTurtle, line 46: The `crio` plugin already captures the `crio` journal.

TurboTurtle (commented):
By "always be spawned either by cri-o or containerd" - does this mean that the existing config file collection is no longer valid? Or does kata still write to the host filesystem despite being spawned by a container runtime?

sos/report/plugins/kata_containers.py
fidencio: So, I'd also remove the `containerd` journal and leave that for a possible `containerd` plugin.  This simplifies even more things on the kata-containers side. :-)

fidencio:
> By "always be spawned either by cri-o or containerd" - does this mean that the existing config file collection is no longer valid? 

The existing config file collection is still valid.

> Or does kata still write to the host filesystem despite being spawned by a container runtime?

What I meant there is that as kata-containers will be spawned by the CRI runtime, its logs will go together with the CRI runtime logs to journal, using the very same identifier.  So, if using CRI-IO, kata-containers logs will have the `crio` identifier, rather than `kata-runtime`.

Is this more clear?

TurboTurtle:
Ah! Ok, yes that makes more sense to me now.

pmoravec (approved):

PR #2607, opened 2021-07-12
[container_log] Capture Erlang crash dump from RabbitMQ

kajinamit:
When RabbitMQ process crashes, it sometimes generates erl_crash.dump
in its log directory. This file is is useful to understand the cause.

This change ensures that file is captured in sosreport even without
--all-logs.

---

- [ X ] Are any related Issues or existing PRs [properly referenced](https://docs.github.com/en/issues/tracking-your-work-with-issues/creating-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) via a Closes (Issue) or Resolved (PR) line?


TurboTurtle:
We can certainly add collection of these logs, but `container_log` isn't the right place. This collection should go in the `rabbitmq` plugin.

kajinamit:
I modified PR title several times but ended up with reverting the change to make the title consistent with git commit title... sorry for annoying notifications.

TurboTurtle (approved):

pmoravec (approved):
ACK in general. Keep in mind that `sos` captures up to 25MB of a file by default. So if the dump would be bigger, `sos` would collect just 25MB unuseable tail of it.

(I dont know the typical/maximum RabbitMQ dump size nor its format, just raising a potential warning)

kajinamit:
@pmoravec That's a very good point. I was looking at the actual dump file I found during debugging an issue and actually the dump file size was more than 100MiB. In this specific case RabbitMQ crashed because of huge number of items(atoms) it was handling but it is likely the dump file can be big enough.

I'll give another thought on this but a good option would be 
 - get output of ls so that we can notice that dump file is generated
 - get first 8 lines from dump file to capture its header, which usually explains why rabbitmq crashed

TurboTurtle:
We can also just set the `sizelimit` parameter of this collection higher - the sizelimit default is there as a sanity catch against huge recursive collections - for individual files it's fine to increase it if the data being collected is actually useful in its full size.

TurboTurtle:
@kajinamit - cycling around on this. How large can we expect the dump file(s) to be for a "standard" collection? If they aren't super massive, then I'd still be fine with increasing the `sizelimit` for this particular collection.

PR #2608, opened 2021-07-12
[cleaner] Read map file only once for parsers

TurboTurtle:
Instead of re-reading the mapping file for each parser, read it once and
store the contents, then hand those contents over to each parser.

This allows us to side-step handling the same exception for malformed
config files over and over for each parser loaded.

---

### DIFF
--- a/sos/cleaner/__init__.py
+++ b/sos/cleaner/__init__.py
@@ -76,7 +76,7 @@ class SoSCleaner(SoSComponent):
             # when obfuscating a SoSCollector run during archive extraction
             os.makedirs(os.path.join(self.tmpdir, 'cleaner'), exist_ok=True)
 
-        self.validate_map_file()
+        self.cleaner_mapping = self.load_map_file()
         os.umask(0o77)
         self.in_place = in_place
         self.hash_name = self.policy.get_preferred_hash_name()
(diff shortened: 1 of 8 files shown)


sos/cleaner/__init__.py
```diff
@@ -132,13 +133,14 @@ def validate_map_file(self):
...
                 try:
-                    json.load(mf)
+                    _conf = json.load(mf)
                 except json.JSONDecodeError:
                     self.log_error("ERROR: Unable to parse map file, json is "
                                    "malformed. Will not load any mappings.")
                 except Exception as err:
                     self.log_error("ERROR: Could not load '%s': %s"
                                    % (self.opts.map_file, err))
+        return _conf
```
pmoravec, line 143: As we newly load the map file here, I suggest renaming the `validate_map_file` to e.g. `load_map_file`.

sos/cleaner/parsers/keyword_parser.py
```diff
@@ -22,10 +22,10 @@ class SoSKeywordParser(SoSCleanerParser):
     map_file_key = 'keyword_map'
     prep_map_file = ''
 
-    def __init__(self, conf_file=None, keywords=None, keyword_file=None):
+    def __init__(self, config, keywords=None, keyword_file=None):
         self.mapping = SoSKeywordMap()
         self.user_keywords = []
-        super(SoSKeywordParser, self).__init__(conf_file)
+        super(SoSKeywordParser, self).__init__(config=config)
```
pmoravec, line 28: Very nitpick: on other places, you call parent class init method with `(config)` - is there some reason to specifically assign `config` here?

TurboTurtle: Good point, not sure.

sos/cleaner/__init__.py
TurboTurtle: Ack, good idea.

pmoravec (approved):

PR #2610, opened 2021-07-13
[qaucli] Add QLogic plugin support

vikas-goel:

---

### DIFF
--- a/sos/report/plugins/qaucli.py
+++ b/sos/report/plugins/qaucli.py
@@ -1,0 +1,41 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, IndependentPlugin
+
+
+class Qaucli(Plugin, IndependentPlugin):
+
+    short_desc = 'QLogic information'
+
+    plugin_name = 'qaucli'
+    profiles = ('system', 'storage', 'hardware',)
+    packages = ('QConvergeConsoleCLI',)
+
+    def setup(self):
+        cmd = "/opt/QLogic_Corporation/QConvergeConsoleCLI/qaucli"
+        subcmds = ['-c', '-g', '-pr fc -z', '-t']
+
+        for subcmd in subcmds:
+            self.add_cmd_output(
+                "%s %s" % (cmd, subcmd),
+                suggest_filename="qaucli_%s" % subcmd)
+
+        result = self.collect_cmd_output(
+                     "%s -i" % cmd,
+                     suggest_filename="qaucli_-i")
+
+        if result['status'] == 0:
+            for line in result['output'].splitlines():
+                if "HBA Instance" in line:
+                    hba = line.split(':')[1].strip()
+                    self.add_cmd_output(
+                        "%s -l %s" % (cmd, hba),
+                        suggest_filename="qaucli_-l_%s" % hba)
+
+# vim: set et ts=4 sw=4 :


sos/report/plugins/qaucli.py
```diff
@@ -0,0 +1,37 @@
...
+
+    short_desc = 'QLogic information'
+
+    plugin_name = 'qaucli'
+    profiles = ('system', 'storage', 'hardware',)
+    packages = ('QConvergeConsoleCLI',)
+
+    def setup(self):
+        os.environ['PATH'] += os.pathsep + \
+            "/opt/QLogic_Corporation/QConvergeConsoleCLI"
```
TurboTurtle, line 23: Nack. Do not manually modify the `PATH` for this, the sos API already accounts for standardized `PATH`s, and additionally doing things this way can interfere with non-standard `--sysroot`s.

sos/report/plugins/qaucli.py
```diff
@@ -0,0 +1,37 @@
...
+
+        subcmds = ['-c', '-g', '-pr fc -z', '-t']
+        self.add_cmd_output([
+            "qaucli %s" % subcmd for subcmd in subcmds
+        ])
```
TurboTurtle, line 28: Ack.

sos/report/plugins/qaucli.py
```diff
@@ -0,0 +1,37 @@
...
+
+        result = self.collect_cmd_output('qaucli -i')
+        if result['status'] == 0:
+            for line in result['output'].splitlines():
+                if "HBA Instance" in line:
+                    hba = line.split(':')[1].strip()
+                    self.add_cmd_output("qaucli -l %s" % hba)
```
TurboTurtle, line 35: Ack.

TurboTurtle (changes requested):
Overall this looks good, just need to remove the `PATH` modifications.

The plugin name I think should be changed to just `qlogic`, but I can certainly be persuaded otherwise if there's a solid reason it shouldn't be `qlogic`.

sos/report/plugins/qaucli.py
vikas-goel: The CLI is `/opt/QLogic_Corporation/QConvergeConsoleCLI/qaucli`. What would be appropriate way to set the path?

vikas-goel:
> Overall this looks good, just need to remove the `PATH` modifications.
> 
> The plugin name I think should be changed to just `qlogic`, but I can certainly be persuaded otherwise if there's a solid reason it shouldn't be `qlogic`.

There are other packages too to manage QLogic. For example, SANsurfer CLI (`scli`). Someone may want to develop a plugin for that. I can change the plugin name to `qlogic`. I think the filename should still be `qaucli.py`. What do you think?

sos/report/plugins/qaucli.py
TurboTurtle: You can just use that full path as the command.

If you wanted to make the resulting filenames more human friendly, you'd need to change your command loop a bit to something like:

```python
cmd = "/opt/QLogic_Corporation/QConvergeConsoleCLI/qaucli"
for subcmd in subcmds:
    self.add_cmd_output("%s %s" % (cmd, subcmd),
                        suggest_filename="quacli_%s" % subcmd)
```

vikas-goel: That won't work for options (`subcmd`) with space. Is there a possibility to change the sosreport framework to
1. pick command basename when creating output file,
2. create command alias as a function parameter, or
3. update PATH env variable?

vikas-goel: I just tried your suggestion. It worked for the space too. Thank you @TurboTurtle .

vikas-goel:
> Overall this looks good, just need to remove the `PATH` modifications.
> 
> The plugin name I think should be changed to just `qlogic`, but I can certainly be persuaded otherwise if there's a solid reason it shouldn't be `qlogic`.

Done

vikas-goel:
Hi @TurboTurtle ,
All suggestions incorporated here.

TurboTurtle (approved):
LGTM, just need a second ack from one of the other maintainers.

vikas-goel:
Hi @TurboTurtle ,
How can we get another reviewer ack this and other similar PRs?

TurboTurtle:
The other maintainers will look over and review as they are able - but I know at least 2 of them were on vacation this week. I imagine a second ack should be coming in the next week or so.

In any event, our next release is slated for mid-August, so these should have no issue being merged by that time.

vikas-goel:
> In any event, our next release is slated for mid-August, so these should have no issue being merged by that time.

Any idea which Red Hat release will pick the mid-August sos release?

vikas-goel:
> The other maintainers will look over and review as they are able - but I know at least 2 of them were on vacation this week. I imagine a second ack should be coming in the next week or so.
> 
> In any event, our next release is slated for mid-August, so these should have no issue being merged by that time.

Is there any update on when this can be merged?

sos/report/plugins/qaucli.py
```diff
@@ -0,0 +1,41 @@
...
+from sos.report.plugins import Plugin, IndependentPlugin
+
+
+class Qaucli(Plugin, IndependentPlugin):
+
+    short_desc = 'QLogic information'
+
+    plugin_name = 'qaucli'
+    profiles = ('system', 'storage', 'hardware',)
+    packages = ('QConvergeConsoleCLI',)
```
pmoravec, line 18: Will the package name be matched on Debian/Ubuntu? Since we do case sensitive package name check and I see `qconvergeconsolecli` DEB package ("contrary" to `QConvergeConsoleCLI` RPM) in some manuals.

pmoravec:
Up to the package name case sensitivity, ack.

sos/report/plugins/qaucli.py
vikas-goel: I don’t know about that. Worst case if the package name doesn’t match, it won’t execute. I don’t have access to that package nor do I know how on the disk the package extraction looks like. So, I don’t want to add the lowercase package name.

What’s your opinion?

pmoravec: > Worst case if the package name doesn’t match, it won’t execute.

Indeed. It isnt a bug, "just" possible lack of new functionality. Until @slashdd or @adam-stokes object (if they know `qconvergeconsolecli` DEB is really in lowercase), I am OK with merging it as is.

pmoravec (approved):

PR #2611, opened 2021-07-13
[storcli] Add LSI MegaRAID device support

vikas-goel:

---

### DIFF
--- a/sos/report/plugins/storcli.py
+++ b/sos/report/plugins/storcli.py
@@ -1,0 +1,44 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, IndependentPlugin
+
+
+class StorCLI(Plugin, IndependentPlugin):
+
+    short_desc = 'LSI MegaRAID devices'
+
+    plugin_name = 'storcli'
+    profiles = ('system', 'storage', 'hardware',)
+    packages = ('storcli',)
+
+    def setup(self):
+        cmd = '/opt/MegaRAID/storcli/storcli64'
+        subcmds = [
+            'show ctrlcount J',
+            '/call show AliLog',
+            '/call show all',
+            '/call show events',
+            '/call show termlog',
+            '/call/bbu show all',
+            '/call/cv show all',
+            '/call/dall show J',
+            '/call/eall show all',
+            '/call/eall/sall show all',
+            '/call/sall show all',
+            '/call/vall show all',
+        ]
+
+        logpath = self.get_cmd_output_path()
+
+        for subcmd in subcmds:
+            self.add_cmd_output(
+                "%s %s" % (cmd, subcmd),
+                suggest_filename="storcli64_%s" % subcmd, runat=logpath)
+
+# vim: set et ts=4 sw=4 :


sos/report/plugins/storcli64.py
```diff
@@ -0,0 +1,46 @@
...
+class StorCLI64(Plugin, IndependentPlugin):
+
+    short_desc = 'LSI MegaRAID devices'
+
+    plugin_name = 'storcli64'
+    profiles = ('system', 'storage', 'hardware',)
+    packages = ('storcli',)
+
+    def setup(self):
+        os.environ['PATH'] += os.pathsep + "/opt/MegaRAID/storcli"
```
TurboTurtle, line 22: Nack to `PATH` modifications - same as in #2610

sos/report/plugins/storcli64.py
```diff
@@ -0,0 +1,46 @@
...
+            '/call/eall/sall show all',
+            '/call/sall show all',
+            '/call/vall show all',
+        ]
+
+        logpath = self.get_cmd_output_path()
+
+        self.add_cmd_output([
+            "%s %s" % (cmd, subcmd) for subcmd in subcmds
+        ], runat=logpath)
```
TurboTurtle, line 44: What's the rationale behind using `runat` here? It's certainly acceptable to do so if needed, but it's very rare to see so I just would like to understand the requirement here.

sos/report/plugins/storcli64.py
```diff
@@ -0,0 +1,46 @@
...
+
+from sos.report.plugins import Plugin, IndependentPlugin
+import os
+
+
+class StorCLI64(Plugin, IndependentPlugin):
+
+    short_desc = 'LSI MegaRAID devices'
+
+    plugin_name = 'storcli64'
```
TurboTurtle, line 17: So a while back we updated the `megacli` plugin, and ran into feedback that `storcli` was abandoned even though it was originally intended to _replace_ `megacli`.

Has that changed? Is there actual use/development going on for `storcli`?

If so, I'd suggest a slight modification to the `short_desc`. The `megacli` plugin currently has the same description string, which might confuse users to see two different plugins with the same description.

TurboTurtle:
Also, the Ubuntu 18.04 failure seems unrelated to this PR - it's a not uncommon thing for us to see transient timeouts on 18.04.

I've restarted the job.

sos/report/plugins/storcli64.py
vikas-goel: Looking forward for your recommendation to fix the path.

sos/report/plugins/storcli64.py
vikas-goel: The command generates a log file `storcli.log` in current directory. If `runat` is not passed, then the log file is generated in the plugin directory or some other.

sos/report/plugins/storcli64.py
vikas-goel: Working with the vendor (LSI) recommendation is to use storcli64.
`LSI has created a tool to help manage RAID devices called StorCLI (previously MegaCLI).`

What change in the description would you recommend?

vikas-goel:
@TurboTurtle , thank you for review comments.

vikas-goel:
@TurboTurtle ,
Should I change the plugin name to 'storcli' and still keep the filename `storcli64.py`? Or, would you prefer the plugin name be `storcli64`?

sos/report/plugins/storcli64.py
vikas-goel: This problem is there with the megacli plugin too. The plugin generates a couple of log files in the current working directory.

TurboTurtle:
I think `storcli` is better here, for both the plugin and filename.

vikas-goel:
Hi @TurboTurtle ,
All suggestions incorporated here.

TurboTurtle (approved):

vikas-goel:
@pmoravec , please review

pmoravec (approved):

PR #2616, opened 2021-07-16
[cleaner] Use a nested ProcessPoolExecutor for extraction

TurboTurtle:
This commit inserts a nested ProcessPoolExecutor into the extraction
workflow for archives that are being obfuscated by `sos clean`.

Previously, the extraction was handled inside the same thread as the
rest of the obfuscation routines for each archive. However, it has been
found that when very large archives are manipulated concurrently,
performance can take a massive hit during the extraction process. This
is due to GIL limitations.

In this aspect 'very large archives' implies many tens of thousands of
files - e.g. 50K+. Because TarFile uses a 10K internal buffer, we end up
spinning a lot of time processing each file via the interpreter.

By shunting each extraction off into a new process space, we can avoid
the GIL issues altogether.

---

### DIFF
--- a/sos/cleaner/obfuscation_archive.py
+++ b/sos/cleaner/obfuscation_archive.py
@@ -15,6 +15,19 @@ import stat
 import tarfile
 import re
 
+from concurrent.futures import ProcessPoolExecutor
+
+
+# python older than 3.8 will hit a pickling error when we go to spawn a new
+# process for extraction if this method is a part of the SoSObfuscationArchive
+# class. So, the simplest solution is to remove it from the class.
+def extract_archive(archive_path, tmpdir):
+    archive = tarfile.open(archive_path)
+    path = os.path.join(tmpdir, 'cleaner')
+    archive.extractall(path)
+    archive.close()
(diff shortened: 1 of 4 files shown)


TurboTurtle:
One thing I would like to note, is I did attempt to replace the existing TPE with a PPE (within `sos/clean/__init__.py::obfuscate_report_paths()`), however this lead to many pickle-related issues, including things outside our direct purview like argparse.

I wouldn't be opposed to making that kind of switch later down the line if we can determine the correct way to do it, while still ensuring consistency across multiple archives.

TurboTurtle:
Following a traceback encountered by @pmoravec, I've update the PR to account for python runtimes older than 3.8 (see code comment).

I've also included a new test in this push that tests `sos clean` against an already existing archive, rather than just having the tests against an inline `sos report --clean`. My only hesitation is pushing the 11M file to the git repo, but I tend to fall on the belief that including a known-configuration sos report for the test suite is more advantageous than creating a test that creates an ad-hoc report on the test systems for each run of this particular test. Thoughts welcome on this point.

pmoravec (approved):

PR #2621, opened 2021-07-21
[policies] add uniontech os policy support

viikei:
Signed-off-by: liugang <liuganga@uniontech.com>

---

### DIFF
--- a/sos/policies/distros/uniontechserver.py
+++ b/sos/policies/distros/uniontechserver.py
@@ -1,0 +1,40 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.policies.distros.redhat import RedHatPolicy, OS_RELEASE
+import os
+
+
+class UnionTechPolicy(RedHatPolicy):
+    distro = "UnionTech OS Server"
+    vendor = "The UnionTech Project"
+    vendor_urls = [('Distribution Website', 'https://www.chinauos.com/')]
+
+    def __init__(self, sysroot=None, init=None, probe_runtime=True,
+                 remote_exec=None):
+        super(UnionTechPolicy, self).__init__(sysroot=sysroot, init=init,
+                                              probe_runtime=probe_runtime,
+                                              remote_exec=remote_exec)
+
+    @classmethod
+    def check(cls, remote=''):
+
+        if remote:
+            return cls.distro in remote
+
+        if not os.path.exists(OS_RELEASE):
+            return False
+
+        with open(OS_RELEASE, 'r') as f:
+            for line in f:
+                if line.startswith('NAME'):
+                    if 'UnionTech OS Server' in line:
+                        return True
+        return False
+
+# vim: set et ts=4 sw=4 :


viikei:
@TurboTurtle PLS, Thanks

sos/policies/distros/uniontech.py
```diff
@@ -0,0 +1,42 @@
+# Copyright (C) Red Hat, Inc. 2019
```
bmr-cymru, line 1: This copyright claim seems incorrect - please update it with a correct copyright holder for this new file.

sos/policies/distros/uniontech.py
```diff
@@ -0,0 +1,42 @@
...
+        if not os.path.exists(OS_RELEASE):
+            return False
+
+        with open(OS_RELEASE, 'r') as f:
+            for line in f:
+                if line.startswith('NAME'):
+                    if 'UnionTech' in line:
+                        return True
+        return False
+
```
bmr-cymru, line 41: It looks as thought this policy was based on the Amazon policy (`sos/policies/distros/amazon.py`). In that case this works because Amazon's Linux distribution is based on and shares a lot of similarity with the Red Hat family of distributions. They have a common package manager (RPM), and similarities in file system layout that mean sharing policy makes sense.

If I understand correctly UOS/Union Tech Linux is based on Deepin, which itself is based on Debian Stable. If that is still the case then it probably makes more sense to base the Union Tech policy on one of the Debian family of distributions, since there will be greater commonality (of fs layout, packaging etc.) than with a Red Hat based distro.

TurboTurtle:
> @TurboTurtle PLS, Thanks

Sorry, I've been a bit swamped internally the last couple days. I'll try and take a look at this in the next few days, but for the moment I'd just echo @bmr-cymru's comment.

sos/policies/distros/uniontech.py
viikei: We have two versions. Debian is our desktop version system. I also have a server version. This time, the server version is supported. The server version is a secondary development version based on the openEuler system.

viikei:
@bmr-cymru

sos/policies/distros/uniontech.py
viikei: @bmr-cymru 
I have revised this question. Thank you very much.

sos/policies/distros/uniontech-server.py
```diff
@@ -0,0 +1,40 @@
...
+                 remote_exec=None):
+        super(UnionTechPolicy, self).__init__(sysroot=sysroot, init=init,
+                                              probe_runtime=probe_runtime,
+                                              remote_exec=remote_exec)
+
+    @classmethod
+    def check(cls, remote=''):
+
+        if remote:
+            return cls.distro in remote
```
TurboTurtle, line 28: Does `/etc/os-release` actually contain the string `Union Tech Linux`, or should this be looking for `UnionTech` like the check below?

viikei: image

bmr-cymru: Hyphens are problematic (although not outright forbidden) in Python filenames. This is because they are ambiguous in source code with respect to the subtraction operator (python does not strictly require whitespace around this operator, so "import foo-bar" could be interpreted as "import the module named foo-bar" or "import the module name resulting from subtracting bar from foo". For this reason they are highly discouraged - consider using "uniontechserver.py" instead.

TurboTurtle: Right, so for the `remote` check you'd want to check for `UnionTech`, or perhaps `UnionTech OS`, rather than `Union Tech Linux` . The `distro` value can stay the same so banner messages are correct, but this particular check needs to actually match the file content in order for `collect` to be able to work for remote nodes running this distro.

viikei: @bmr-cymru @TurboTurtle You are right,I accept your suggestion。Thank you very much again.^_^

TurboTurtle:
Code looks fine to me now. Only thing left is you'll need to squash the fixup commits into the original, as the project doesn't allow fixup or merge commits into master.

If you're unsure how to do this, you can do the following:

1. On your local branch run `git rebase -i HEAD~4`
2. In the editor that opens, for the 3 commits following the original one, change `pick` to `fixup`
3. Save and exit the editor
4. `git push origin $branch --force`

This will squash those commits into the first.

viikei:
> Code looks fine to me now. Only thing left is you'll need to squash the fixup commits into the original, as the project doesn't allow fixup or merge commits into master.
> 
> If you're unsure how to do this, you can do the following:
> 
> 1. On your local branch run `git rebase -i HEAD~4`
> 2. In the editor that opens, for the 3 commits following the original one, change `pick` to `fixup`
> 3. Save and exit the editor
> 4. `git push origin $branch --force`
> 
> This will squash those commits into the first.


@TurboTurtle @bmr-cymru Thank you very much.You two are really really sweet.

TurboTurtle (approved):

TurboTurtle:
Received second ack on IRC, merging shortly.

PR #2628, opened 2021-07-26
[pulp] adding obfuscation of proxy_password in /etc/pulp/server/plugins.conf.d/

### LINKED ISSUE #2564: [pulp] no obfuscation of proxy_password in /etc/pulp/server/plugins.conf.d/*.json files
Reproducer: `/etc/pulp/server/plugins.conf.d/iso_importer.json` with:

        "proxy_password": 'my_SECRET_password',

Then `sos report` does not obfuscate this password.

bvassova:
Closes #2564

Signed-off-by: Barbora Vassova bvassova@redhat.com

---

### DIFF
--- a/sos/report/plugins/pulp.py
+++ b/sos/report/plugins/pulp.py
@@ -160,8 +160,8 @@ class Pulp(Plugin, RedHatPlugin):
         # Now handle JSON-formatted data in the same /etc/pulp directory
         # structure. We use a different substitution string here to preserve
         # the file's JSON syntax.
-        jreg = r"(\s*\".*(passw|cred|token|secret).*\"\s*:\s*\")(.*)(\")"
-        repl = r"\1********\4"
+        jreg = r"(\s*\".*(passw|cred|token|secret).*\"\s*:(\s))(.*)(\w+)"
+        repl = r"\1********"
         self.do_path_regex_sub("/etc/pulp(.*)(.json$)", jreg, repl)
 
         # obfuscate SECRET_KEY = .. and 'PASSWORD': .. in dynaconf list output


sos/report/plugins/pulp.py
```diff
@@ -160,8 +160,8 @@ def postproc(self):
         # Now handle JSON-formatted data in the same /etc/pulp directory
         # structure. We use a different substitution string here to preserve
         # the file's JSON syntax.
-        jreg = r"(\s*\".*(passw|cred|token|secret).*\"\s*:\s*\")(.*)(\")"
-        repl = r"\1********\4"
+        jreg = r"(\"proxy_password\"(\s)*:(\s))(.*)(\w+)"
+        repl = r"\1********"
```
TurboTurtle, line 164: This removes obfuscation of anything aside from `proxy_password`, which is not an acceptable change.

What needs to be investigated is why `proxy_password` is not being matched by the current regex (`.*passw.*`, which should be matching it).

pmoravec: Hm.. maybe just different quotation mark? The current `jreg` uses `"`, my example in #2564 has `'`.

Maybe (writing from head):

            jreg = r"(\s*\".*(passw|cred|token|secret).*\"\s*:\s*(\w+)"

will cover anything after `:`, hence just enhancing the regexp to cover also my example?

pmoravec (approved):
We end up with obfuscation like:

    "proxy_password": ********',

but when I tried to improve the RE, I started to see some passwords again :)

So ACK, all passwords obfuscated, nothing unexpected hidden.

TurboTurtle (approved):

PR #2638, opened 2021-08-05
[tests] Allow entire test classes to be distro specific

TurboTurtle:
Previously new test cases had to be written in such a way that they
could be run on any supported distribution, even if their individual
tests were all marked for specific distributions. Use of the distro-only
decorators would fail when applied to test cases as a whole due to how
test instantiation errors are handled.

To address this, provide a distro check within `BaseSoSTest` so that
tests can be made distro-specific and we can signal to avocado to skip
all tests within that test case without failing the test suite entirely.

---

### DIFF
--- a/tests/sos_tests.py
+++ b/tests/sos_tests.py
@@ -66,6 +66,8 @@ class BaseSoSTest(Test):
     _exception_expected = False
     sos_cmd = ''
     sos_timeout = 300
+    redhat_only = False
+    ubuntu_only = False
 
     @property
     def klass_name(self):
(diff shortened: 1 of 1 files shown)


TurboTurtle:
I had originally planned to add this alongside a test for SCL plugins, however since we don't do testing directly on RHEL (yet) that doesn't really shake out properly. Still, this change by itself should lend itself to more/better test creation in the future.

pmoravec (approved):
(interesting that "UBUNTU_DIST = ['Ubuntu', 'debian']" (with the capital "U" as the only capital letter among all distro names) is really correct, per https://github.com/avocado-framework/avocado/blob/master/avocado/utils/distro.py#L379)

PR #2640, opened 2021-08-10
[sos] Read config section for clean when `--clean` is used

TurboTurtle:
When a `report` or `collect` run would use `--clean` or `--mask` to do
in-line obfuscation of collected reports, sos would not read the config
section for clean - it would only be read if `sos clean` was called
directly. As such, users would need to manually specify config file
values for each run.

Alleviate this gap by reading the config section for `clean` if either
of the cleaner options are used. Do this before we apply cmdline options
so that we maintain our order of precedence.

Related: RHBZ#1950350

---

### DIFF
--- a/sos/component.py
+++ b/sos/component.py
@@ -210,6 +210,14 @@ class SoSComponent():
                 option.default = None
 
         opts.update_from_conf(self.args.config_file, self.args.component)
+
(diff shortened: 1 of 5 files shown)


TurboTurtle:
Added a test for this, which required adding a few other test files.

pmoravec (approved):

Issue #2641, opened 2021-08-10
[collector] unable to pass --plugin-option to cluster nodes


pmoravec:
Reproducer:

    python3 bin/sos collect --batch -vvv --node sos-node1 -o services -k services.servicestatus=on | grep "Final sos command set to"
    [pmoravec-rhel8:finalize_sos_cmd] Final sos command set to /usr/sbin/sos report --batch -c auto --only-plugins=services

And no notion (or configuring) `-k ..` option.

The same for `--plugin-option` instead of `-k`.

So far I was able to check that https://github.com/sosreport/sos/blob/master/sos/collector/__init__.py#L484 already gets `self.opts` with `only_plugins=services` (correct) but `plugin_options=None` (already lost).

TurboTurtle:
This is due to the use of `--only-plugins` - https://github.com/sosreport/sos/blob/8793ff269d6acc0bf2f12ea98a72dc60a8b5ab77/sos/collector/sosnode.py#L769

We should probably move the plugin option block on line 795 up before the `--only-plugins` check.

pmoravec:
Hm.. that is the *second* place that needs fixing, indeed.

The problem is that `self.plugin_options` is empty there (per my debugs), so the line 795 will be no-op. Some passing options has to drop plugopts (while keeping all/most other opts).

PR #2642, opened 2021-08-11
[sosnode] Fix passing of plugin options when using `--only-plugins`

TurboTurtle:
Fixes the handling of plugin options passed by `sos collect` to each
node by first aligning the SoSOption name to those of `report`
(`plugopts`), and second re-arranges the handling of plugin options and
preset options passed by the user when also using `--only-plugins` so
that the former are preserved and passed only with the `--only-plugins`
option value.

Resolves: #2641

---


pmoravec (approved):
Nice!

PR #2643, opened 2021-08-12
[sapnw] Fix IndexError exception

jcastill:
This patch solves the situation where the
command:

/usr/sap/hostctrl/exe/saphostctrl
		-function ListInstances

returns:

No instances found

And a non-zero return code.

Fixes: RHBZ#1992938
Closes: #2643 

---


pmoravec:
I guess it returned zero return code, otherwise we would catch it at https://github.com/sosreport/sos/blob/master/sos/report/plugins/sapnw.py#L25-L26 :)

TurboTurtle (approved):

PR #2646, opened 2021-08-13
[Ceph] Add support for containerized Ceph setup

AsM0DeUz:
RH Ceph 4 can either be installed as RPM or as containers. The changes
permits to collect the ceph config on both kind of setup.

Signed-off-by: Teoman ONAY <tonay@redhat.com>

This is an alternative to PR https://github.com/sosreport/sos/pull/2180 which didn't progress for a year. Instead of having a new plugin for containerized Ceph, I followed the request from the maintainers asking to put everything in one plugin.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1882544

---

### DIFF
--- a/sos/report/plugins/ceph.py
+++ b/sos/report/plugins/ceph.py
@@ -8,6 +8,7 @@
 
 from sos.report.plugins import Plugin, RedHatPlugin, UbuntuPlugin
 from socket import gethostname
+import re
 
 
 class Ceph(Plugin, RedHatPlugin, UbuntuPlugin):
@@ -15,7 +16,8 @@ class Ceph(Plugin, RedHatPlugin, UbuntuPlugin):
     short_desc = 'CEPH distributed storage'
 
     plugin_name = 'ceph'
-    profiles = ('storage', 'virt')
+    profiles = ('storage', 'virt', 'container')
+    containers = ('ceph-(mon|rgw|osd)*',)
     ceph_hostname = gethostname()
 
     packages = (
@@ -35,6 +37,12 @@ class Ceph(Plugin, RedHatPlugin, UbuntuPlugin):
         'ceph-mgr@%s' % ceph_hostname,
         'ceph-radosgw@*',
         'ceph-osd@*'
+    )
+
+    # This check will enable the plugin regardless of being
+    # containerized or not
+    files = (
+        '/etc/ceph/ceph.conf',
     )
 
     def setup(self):
@@ -117,9 +125,9 @@ class Ceph(Plugin, RedHatPlugin, UbuntuPlugin):
             "time-sync-status",
         ]
 
-        self.add_cmd_output([
-            "ceph %s" % s for s in ceph_cmds
-        ])
+        ceph_osd_cmds = [
+            "ceph-volume lvm list",
+        ]
 
         self.add_cmd_output([
             "ceph %s --format json-pretty" % s for s in ceph_cmds
@@ -137,5 +145,30 @@ class Ceph(Plugin, RedHatPlugin, UbuntuPlugin):
             "/var/lib/ceph/tmp/*mnt*",
             "/etc/ceph/*bindpass*"
         ])
+
+        # If containerized, run commands in containers
+        containers_list = self.get_all_containers_by_regex("ceph-*")
+        if containers_list:
+            # Avoid retrieving multiple times the same data
+            got_ceph_cmds = False
+            for container in containers_list:
+                if re.match("ceph-(mon|rgw|osd)", container[1]) and \
+                        not got_ceph_cmds:
+                    self.add_cmd_output([
+                        self.fmt_container_cmd(container[1], "ceph %s" % s)
+                        for s in ceph_cmds
+                    ])
+                    got_ceph_cmds = True
+                if re.match("ceph-osd", container[1]):
+                    self.add_cmd_output([
+                        self.fmt_container_cmd(container[1], "%s" % s)
+                        for s in ceph_osd_cmds
+                    ])
+                    break
+        # Not containerized
+        else:
+            self.add_cmd_output([
+                "ceph %s" % s for s in ceph_cmds
+            ])
 
 # vim: set et ts=4 sw=4 :


sos/report/plugins/ceph.py
```diff
@@ -37,6 +38,12 @@ class Ceph(Plugin, RedHatPlugin, UbuntuPlugin):
         'ceph-osd@*'
     )
 
+    # This check will enable the plugin regardless of being
+    # containerized or not
+    files = (
+        '/etc/ceph/ceph.conf'
+    )
```
TurboTurtle, line 45: Needs to be a tuple, at the moment it's a str.

sos/report/plugins/ceph.py
```diff
@@ -117,14 +124,17 @@ def setup(self):
...
+            "ceph-volume lvm list",
+        ]
 
         self.add_cmd_output([
             "ceph %s --format json-pretty" % s for s in ceph_cmds
         ], subdir="json_output", tags="insights_ceph_health_detail")
 
+        for service in self.services:
+            self.add_journal(units=service)
+
```
TurboTurtle, line 137: This is automatically handled already.

sos/report/plugins/ceph.py
```diff
@@ -138,4 +148,32 @@ def setup(self):
             "/etc/ceph/*bindpass*"
         ])
 
+        # If containerized, run commands in containers
+        containers_list = self.list_ceph_containers("ceph-*")
```
TurboTurtle, line 152: Why not just call `self.get_all_containers_by_regex("ceph-*")` directly here?

sos/report/plugins/ceph.py
```diff
@@ -138,4 +148,32 @@ def setup(self):
...
+                        self.fmt_container_cmd(container[1], "ceph %s" % s)
+                        for s in ceph_cmds
+                    ])
+                    got_ceph_cmds = True
+                if re.match("ceph-osd", container[1]):
+                    self.add_cmd_output([
+                        self.fmt_container_cmd(container[1], "%s" % s)
+                        for s in ceph_osd_cmds
+                    ])
+                    break
```
TurboTurtle, line 169: Why break here? If we iterate over `ceph-osd` before `ceph-mon` then we'd skip collection of the latter.

sos/report/plugins/ceph.py
```diff
@@ -8,14 +8,15 @@
...
+import re
 
 
 class Ceph(Plugin, RedHatPlugin, UbuntuPlugin):
 
     short_desc = 'CEPH distributed storage'
 
     plugin_name = 'ceph'
-    profiles = ('storage', 'virt')
+    profiles = ('storage', 'virt', 'container')
```
TurboTurtle, line 19: As an enablement trigger, you can use the `containers` class attr here. Also a tuple, elements should be patterns for container names.

sos/report/plugins/ceph.py
AsM0DeUz: Fixed

sos/report/plugins/ceph.py
AsM0DeUz: This is why it was removed in my git diff ... I was wondering why. Fixed

sos/report/plugins/ceph.py
AsM0DeUz: Because at first I thought that I might have to use it multiple times and at the end ... I didn't. Thx for the highlight.

sos/report/plugins/ceph.py
AsM0DeUz: Because if the first container in the list is an OSD then both list of commands will be retrieved and the iteration will stop (which is the expected result). If it is another kind of container (mon or mgr) it will retrieve "ceph_cmds" and continue iterate until it finds (or not) an OSD container and gets commands specific to OSDs (ceph_osd_cmds) then stop. I did put the break there it is because there can be a lot of OSD containers running (minimum 1 / disk on the server) and running "ceph_osd_cmds" on all of them will be useless as the output will always be the same.

sos/report/plugins/ceph.py
AsM0DeUz: You would leave just 'container' as a profile?

AsM0DeUz:
@TurboTurtle 

Regarding the class:

class Ceph(Plugin, RedHatPlugin, UbuntuPlugin):

I wonder if I shouldn't create a separate Class for UbuntuPlugin (with the old content) as I don't know if what I wrote will work on an Ubuntu.

sos/report/plugins/ceph.py
TurboTurtle: No, I'm saying you can do this:

```python
profiles = ('storage', 'virt', 'container')
containers = ('ceph-(mon|rgw|osd)*',)
```

So it will be enabled with the container profile, but _also_ if there are any running containers with names matching that pattern.

sos/report/plugins/ceph.py
TurboTurtle: Ack.

sos/report/plugins/ceph.py
AsM0DeUz: OK clear now. Thx

TurboTurtle (approved):

TurboTurtle:
> @TurboTurtle
> 
> Regarding the class:
> 
> class Ceph(Plugin, RedHatPlugin, UbuntuPlugin):
> 
> I wonder if I shouldn't create a separate Class for UbuntuPlugin (with the old content) as I don't know if what I wrote will work on an Ubuntu.

The changes would work _if_ the container names are the same on Ubuntu based hosts running Ceph. Otherwise, it will be a no-op. From that perspective, no need to split out another class.

However, if the container names are different then we would probably want to split it out, with each distro defining container names.

@slashdd @arif-ali any insight on the above?

AsM0DeUz:
If they use one of the same installation method we (RH) support (ceph-ansible), the containers names will be the same. If they do the installation manually it can be anything.

pmoravec (approved):

PR #2648, opened 2021-08-16
[sapnw] Fix IndexError exception

bvassova:
This patch solves the situation where the
command:

/usr/sap/hostctrl/exe/saphostctrl
		-function ListInstances

returns:

No instances found

And a non-zero return code.

Resolves: #2648
Related: #2643
Relevant to: rhbz#1992938

---

### DIFF
--- a/sos/plugins/sapnw.py
+++ b/sos/plugins/sapnw.py
@@ -39,7 +39,8 @@ class sapnw(Plugin, RedHatPlugin):
         # Cycle through all the instances, get 'sid', 'instance_number'
         # and 'vhost' to determine the proper profile
         for inst_line in inst_out['output'].splitlines():
-            if "DAA" not in inst_line:
+            if ("DAA" not in inst_line and not
+                    inst_line.startswith("No instances found")):
                 fields = inst_line.strip().split()
                 sid = fields[3]
                 inst = fields[5]


TurboTurtle (approved):

sos/plugins/sapnw.py
```diff
@@ -39,7 +39,8 @@ def collect_list_instances(self):
         # Cycle through all the instances, get 'sid', 'instance_number'
         # and 'vhost' to determine the proper profile
         for inst_line in inst_out['output'].splitlines():
-            if "DAA" not in inst_line:
+            if "DAA" not in inst_line and not
+                    inst_line.startswith("No instances found")):
```
TurboTurtle, line 43: The trailing `)` is causing the LGTM failure. A leading `(` at the beginning of the conditional is probably the best soluition.

TurboTurtle (changes requested):

sos/plugins/sapnw.py
bvassova: Hi, thank you, I made the change.

pmoravec (approved):
Just "And a non-zero return code." should be "And a zero return code." (copy&paste error from previous PR). Since when the command returns non-zero, we don't reach this code due to https://github.com/sosreport/sos/pull/2648/files#diff-b780ae8196dfb05ae81755b73569ca17e73f29d7ccd3b3198a8c1e123891ef19L35-L36 ;-)

TurboTurtle (approved):

PR #2650, opened 2021-08-18
[policies] add openeuler policy support

marsCatXdu:
Signed-off-by: lijingwei <lijingwei@uniontech.com>

---

### DIFF
--- a/sos/policies/distros/openeuler.py
+++ b/sos/policies/distros/openeuler.py
@@ -1,0 +1,40 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.policies.distros.redhat import RedHatPolicy, OS_RELEASE
+import os
+
+
+class OpenEulerPolicy(RedHatPolicy):
+    distro = "openEuler"
+    vendor = "The openEuler Project"
+    vendor_urls = [('Distribution Website', 'https://openeuler.org/')]
+
+    def __init__(self, sysroot=None, init=None, probe_runtime=True,
+                 remote_exec=None):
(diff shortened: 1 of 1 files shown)


TurboTurtle (commented):
The policy by itself looks fine. However, are the existing plugins sufficient for the needs of openEuler?

With the current setup, openEuler systems will run `RedHatPlugin` tagged plugins - which is fine, especially since the policy inherits from `RedHatPolicy` - but this assumes that any configuration files/layouts, package names, etc... are all the same.

If there are instances where package names, directory layouts, config file paths, log paths, etc... differ, then we'd likely want to add a new `OpenEulerPlugin` tagging class as well to be able to define the differences in how plugins should run on these systems.

pmoravec (approved):
Likewise what Jake said. Policy itself fine, worth to identify if/when it differs from `RedHatPolicy`.

marsCatXdu:
> The policy by itself looks fine. However, are the existing plugins sufficient for the needs of openEuler?
> 
> With the current setup, openEuler systems will run `RedHatPlugin` tagged plugins - which is fine, especially since the policy inherits from `RedHatPolicy` - but this assumes that any configuration files/layouts, package names, etc... are all the same.
> 
> If there are instances where package names, directory layouts, config file paths, log paths, etc... differ, then we'd likely want to add a new `OpenEulerPlugin` tagging class as well to be able to define the differences in how plugins should run on these systems.

Well... I wanted to make my first contribution simple. I would make a further look around writing the plugin these days and try to make one.

So, should I close this pull request?

pmoravec (approved):
Sounds good then (wrt explanation of no real diff from `RedHatPolicy` for now).

TurboTurtle:
> 
> Well... I wanted to make my first contribution simple. I would make a further look around writing the plugin these days and try to make one.
> 
It's fine to add the tagging class later as we need to with any new plugins or changes to existing ones. I was just highlighting it here in case there were any known plugins that you or other openEuler users may have planned to submit contributions for later on.

> So, should I close this pull request?

No, we can take this as is.

TurboTurtle (approved):

PR #2651, opened 2021-08-18
[block] Capture udisks configuration and output

jcastill:
This patch enables the capture of the udisks config
in:

/etc/udisks/

And status and object dump via udisksctl command.

Resolved: #2651 

---

### DIFF
--- a/sos/report/plugins/block.py
+++ b/sos/report/plugins/block.py
@@ -66,4 +66,14 @@ class Block(Plugin, IndependentPlugin):
                     dev = line.split()[0]
                     self.add_cmd_output('cryptsetup luksDump /dev/%s' % dev)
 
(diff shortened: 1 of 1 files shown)


TurboTurtle:
Code looks fine by itself, but I'm wavering on if this should actually be in `block`, and not a new `udisks` plugin? Is there more data that can be collected from udisksctl or other commands provided by the package?

jcastill:
At the moment there isn't a lot of other data that we can get from udisks, but I can see it expanding in the future. A new plugin for udisks may be a better idea. Should I close this PR and send a new one, or reuse this one?

TurboTurtle:
I'd say let's go with a new plugin then, a new PR is fine.

PR #2655, opened 2021-08-19
[collect] Remove double prompt for case ID

TurboTurtle:
Remove the case ID prompt within `collect`, since the one from `Policy`
will always apply. This avoids a double prompt for a case ID if the user
does not originally provide one.

---


pmoravec (approved):
ACK. The reproducer is (it took me a while to find it) to not provide `--batch` either `--case-id` on command line, as well as **leave** the first `Optionally, please enter the case id ..` prompt empty.

Issue #2656, opened 2021-08-19
[tests] Foreman testing suite aborting due to Debian 11 release


TurboTurtle:
The current Foreman tests will now abort before the test suite actually gets run. This is due to Debian 11 releasing, thus causing this in the prep scripts:

```
Reading package lists...
E: Repository 'http://security.debian.org/debian-security buster/updates InRelease' changed its 'Suite' value from 'stable' to 'oldstable'
E: Repository 'http://deb.debian.org/debian buster InRelease' changed its 'Suite' value from 'stable' to 'oldstable'
E: Repository 'http://deb.debian.org/debian buster-updates InRelease' changed its 'Suite' value from 'stable-updates' to 'oldstable-updates'
```
Which causes `apt-get update` to exit with a status code of 100, thus triggering cirrus to fail the task.

We have a few options here:

1. Add `--allow-releaseinfo-change` to the `apt-get update` step
2. Rebuild the images on Debian 11 (does Foreman 2.4 support Debian 11?)

pmoravec:
>     2\. Rebuild the images on Debian 11 (does Foreman 2.4 support Debian 11?)

Does not seem so: https://theforeman.org/manuals/2.4/index.html#2.Quickstart says Debian 10 is supported. (and foreman 2.5 has the same).

ekohl:
The problem is that in your `/etc/apt/sources.list` you have listed `stable`. If you make that `buster` it shouldn't try to update to the new stable. How do you get the image? I think that the cloud image should have this by default.

As for Debian 11, we don't support it yet, but it's being discussed. The groundwork for that has yet to start so Foreman 3.1 earliest (Foreman 3.0 is at rc1 now).

PR #2657, opened 2021-08-20
[tests] Update Cirrus testing for Foreman 2.5

TurboTurtle:
Updates the Foreman tests to use new images created for Foreman 2.5.

Additionally, adds an option to the prep script to prevent new Debian
release suite changes from causing the prep scripts to abort the tests.
For out purposes, suite changes are not important during test runs as
these images are built on known versions.

Resolves: #2656

---

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -20,8 +20,8 @@ env:
     # These are generated images pushed to GCP from Red Hat
     FEDORA_IMAGE_NAME: "f${FEDORA_VER}-server-sos-testing"
     FEDORA_PRIOR_IMAGE_NAME: "f${FEDORA_PRIOR_VER}-server-sos-testing"
-    FOREMAN_CENTOS_IMAGE_NAME: "foreman-24-centos-8-sos-testing"
-    FOREMAN_DEBIAN_IMAGE_NAME: "foreman-24-debian-sos-testing"
+    FOREMAN_CENTOS_IMAGE_NAME: "foreman-25-centos-8-sos-testing"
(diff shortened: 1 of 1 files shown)


pmoravec (approved):

.cirrus.yml
```diff
@@ -118,7 +116,7 @@ report_stageone_task:
     remove_sos_script: &remove_sos |
         if [ $(command -v apt) ]; then
             apt -y purge sosreport
-            apt update
+            apt update --allow-releaseinfo-change
```
ekohl, line 119: I don't think this is correct. You will pull in Debian 11 packages into Debian 10. You'll end up with a weird hybrid 10/11. I recommend you choose a base image that has `buster` in `/etc/apt/sources.list` rather than `stable`.

TurboTurtle: Doesn't `update` just refresh package availability info? We're only installing `python3-pip` to in turn pull in `avocado`.

I can edit the images in the GCP project, but I was under the impression `update` would not actually install new versions.

ekohl: Correct, but you then have metadata info from Debian 11. Any `apt-get install` command after that will install the Debian 11 version of it. That's why you'll end up with a weird hybrid. It's kind of like using Fedora 34 but pointing `/etc/yum.repos.d` to Fedora 35 (without actually running a distro upgrade).

TurboTurtle: Ack. I'll update the images in the background, and add a note for future image/foreman updates.

PR #2662, opened 2021-08-29
[omnipath_client] Ensure opacapture runs only with allow-system-changes

jcastill:
While omnipath_client plugin is collecting "opacapture",
`depmod -a` command is executed to regenerates some files
under /usr/lib/modules/$kernel.

modules.dep
modules.dep.bin
modules.devname
modules.softdep
modules.symbols
modules.symbols.bin

This patch ensures that the command is only run when
the option --allow-system-changes is used.

Fixes: RHBZ#1998433
Closes: #2662 

---


jcastill:
This is based on the approach I saw in sos/report/plugins/sapnw.py . The other way I imagine is to use predicates, right? But I cannot find a kernel module or a service that we can use as predicate.

pmoravec (approved):

TurboTurtle (approved):

PR #2664, opened 2021-08-30
[unpackaged] deal with recursive loop of symlinks properly

pmoravec:
When the plugin processes a recursive loop of symlinks, it currently
hangs in an infinite loop trying to follow the symlinks. Use
pathlib.Path.resolve() method to return the target directly.

Resolves: #2664

---

### DIFF
--- a/sos/report/plugins/unpackaged.py
+++ b/sos/report/plugins/unpackaged.py
@@ -10,6 +10,7 @@ from sos.report.plugins import Plugin, RedHatPlugin
 
 import os
 import stat
+from pathlib import Path
 
 
 class Unpackaged(Plugin, RedHatPlugin):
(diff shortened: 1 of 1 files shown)


pmoravec:
One potential concern: the `pathlib` library is present since Python 3.4. Do all distros support/contain at least 3.4? (I expect so but havent verified distro-by-distro).

TurboTurtle (approved):
Code change LGTM.

For all RH-based downstreams, 3.4 is available where sos-4.0+ ships. Since this plugin is not tagged for any Debian or Ubuntu distros, that's good enough in my book.

But, for reference, if this plugin should be extended to those distros at anytime, I believe Debian has shipped python 3.4 since Debian 8?

Issue #2663, opened 2021-08-30
plugopts from config file not mentioned in "effective options" debug log


pmoravec:
Currently, sos report prints a debug:

    [sos.report:setup] effective options now: --batch --build --config-file /tmp/sos.conf --only-plugins qpid -vvv

before the config file is evaluated. We dont know what particular options were really used at the end.

Shall not we (maybe rename the "effective"? and ..) add a debug similar to above that will print options after all options processing is done, incl. config file?

Reproducer:

    echo "[report]" > /tmp/sos.conf
    echo "only-plugins = qpid" >> /tmp/sos.conf
    python3 bin/sos report --conf /tmp/sos.conf -v -v -v --batch --build | grep ffect
    [sos.report:setup] effective options now: --batch --build --config-file /tmp/sos.conf --only-plugins qpid -vvv

(or have the default `/etc/sos/sos.conf` of the same content)

pmoravec:
I apologize for confusion - the bug is in not printing plugin options only, the above example shows a correct behaviour. Proper reproducer:

    echo "[plugin_options]" > /tmp/sos.conf; echo "filesys.lsof = on" >> /tmp/sos.conf
    python3 bin/sos report --conf /tmp/sos.conf -v -v -v --batch --build | grep effect
    [sos.report:setup] effective options now: --batch --build --config-file /tmp/sos.conf -vvv

(same happens when updating the default `/etc/sos/sos.conf` file - other sections are shown in the `effective options", but `plugin_options` not)

TurboTurtle:
Two findings:

1. This is caused by https://github.com/sosreport/sos/blob/main/sos/options.py#L103 - we need to instead hand a private copy of this value over here, otherwise we update the `arg_default` value when we later set the plugin options at https://github.com/sosreport/sos/blob/main/sos/options.py#L236

This can be solved by using `self.arg_defaults.get(arg)` on line 103. Easy enough.

2. However, if we have a plugin option set in the config file _and_ we pass `-k plugin.opt=foo` on the cmdline, the config file value is overwritten. This is inline with how we handle options generally right now, but I wonder if this is actually desired for plugin options specifically? Should we continue to overwrite, or should the cmdline options be appended (since we allow multiple `-k` invocations on the cmdline)?

pmoravec:
1. Whatever I tried on line 103 there, I got no improvement :( .

[...]

PR #2669, opened 2021-08-31
[report,options] Fix reporting of plugopts with effective options message

TurboTurtle:
This patchset fixes the issue raised in #2663 by adding an explicit handler for plugopts in the `to_args()` logic used to generate the command line logged for the ending effective options. We need to do this because the way `SoSOptions()` builds from the `arg_defaults` compared to reading config file sections does not work for how we parse the `plugin_options` section, as it ends up updating the `arg_defaults` inadvertently. During testing however, fixing this more generically lead to either command line values for `plugopts` not being honored or the section handling methods breaking. So, I see this as the least bad option.

Additionally, a new test class is added to ensure logging and option handling from config files behaves as we expect it to. To do this, a companion commit that allows files from `tests/test_data` to be dropped to differently named locations on the test system is included.

---

### DIFF
--- a/sos/options.py
+++ b/sos/options.py
@@ -281,6 +281,8 @@ class SoSOptions():
             null_values = ("False", "None", "[]", '""', "''", "0")
             if not value or value in null_values:
                 return False
+            if name == 'plugopts' and value:
+                return True
             if name in self.arg_defaults:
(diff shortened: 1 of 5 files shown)


pmoravec (approved):
Nice!

PR #2671, opened 2021-09-01
[cleaner] Allow obfuscation of non-sos archives

TurboTurtle:
This patchset enables `sos clean` to perform obfuscation on archives and directories that were not generated by `sos`. This comes after some discussion with maintainers following interest from other groups wanting to use `sos clean` similar to how they leveraged the standalone soscleaner in the past.

As part of this initial PR, supported archives are `sos report|collect` archives or build directories, Insights archives, and arbitrary directories and tarballs - the latter 2 of which will not provide the same level of obfuscation as well-known archive types, instead relying on existing mappings from previous `sos clean` executions.

Also included in this is a catch for truncated shortnames within the hostname parser that became evident during the introduction of archive abstractions.

---


sos/cleaner/__init__.py
```diff
@@ -92,6 +95,17 @@ def __init__(self, parser=None, args=None, cmdline=None, in_place=False,
             SoSUsernameParser(self.cleaner_mapping, self.opts.usernames)
         ]
 
+        self.archive_types = [
+            SoSDirectoryArchive,
```
pmoravec, line 100: Just a nitpick: following the naming logic for `SoSCollector[Directory|Archive]`, does not have sense to name this `SoSReportDirectory`?

sos/cleaner/__init__.py
```diff
@@ -218,59 +237,26 @@ def inspect_target_archive(self):
...
-            archive = tarfile.open(archive)
-        path = os.path.join(self.tmpdir, 'cleaner')
-        archive.extractall(path)
-        return os.path.join(path, archive.name.split('/')[-1].split('.tar')[0])
+        force = False
+        if self.opts.archive_type != 'auto':
+            check_type = self.opts.archive_type.replace('-', '_')
+            self.archive_types = [
+                a for a in self.archive_types if a.type_name == check_type
+            ]
```
pmoravec, line 247: (it took me a while to understand this, hence the below request)

Could you add a comment explaining why it is a *list*? (b'cos `check_type == report` can mean `type_name = SoSReportArchive` **or** `SoSReportDirectory` - curiously it works well even for the current `SoSDirectoryArchive` value :-o)

sos/cleaner/archives/sos.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+class SoSReportArchive(SoSObfuscationArchive):
+    """This is the class representing an sos report, or in other words the
+    type the archive the SoS project natively generates
+    """
+
+    type_name = 'report'
+    description = 'sos report archive'
+    prep_files = {
```
pmoravec, line 25: A good idea to put all prep files to one place.

sos/cleaner/__init__.py
TurboTurtle: I'm trying to see if I can merge this with `SoSCollectDirectory`, but if I can't then ack to the rename

sos/cleaner/__init__.py
TurboTurtle: Ack.

TurboTurtle: Instead of a comment, I refactored the logic here to be a bit more robust and readable.

pmoravec (approved):
Sounds great!

PR #2673, opened 2021-09-05
[report] Implement --estimate-mode

pmoravec:
Add report option `--estimate-mode` to estimate disk space requirements
when running a sos report.

Reasoning for changing some cmdline options when `--estimate-mode` is used:

- `--threads=1` is required to limit disk usage during the estimation and to simplify the calculation
- `--build` enabled as the outcome is rather in UI logs than in packing the (almost empty) tarball; anyway we can agree on any solution here (to optionally pack the almost empty directory, or to delete the tarball/directory (but it contains `sos.log` that can be valuable for understanding of the disk space consumption)
- `--no-postproc` enabled as we dont care about any password obfuscation (in files we will delete either way)
- `--clean` disabled due to the same reason
- `--upload` disabled as a redundant feature (or does it make sense to upload the almost-empty dir/tarball?)

Resolves: #2673

---

### DIFF
--- a/man/en/sos-report.1
+++ b/man/en/sos-report.1
@@ -14,7 +14,7 @@ sos report \- Collect and package diagnostic and support data
           [--preset preset] [--add-preset add_preset]\fR
           [--del-preset del_preset] [--desc description]\fR
           [--batch] [--build] [--debug] [--dry-run]\fR
-          [--label label] [--case-id id]\fR
+          [--estimate-only] [--label label] [--case-id id]\fR
           [--threads threads]\fR
           [--plugin-timeout TIMEOUT]\fR
           [--cmd-timeout TIMEOUT]\fR
@@ -316,6 +316,17 @@ Execute plugins as normal, but do not collect any file content, command
 output, or string data from the system. The resulting logs may be used
 to understand the actions that sos would have taken without the dry run
 option.
+.TP
+.B \--estimate-only
+Estimate disk space requirements when running sos report. This can be valuable
+to prevent sosreport working dir to consume all free disk space. No plugin data
+is available at the end.
+
+Plugins will be collected sequentially, size of collected files and commands outputs
+will be calculated and the plugin files will be immediatelly deleted prior execution
+of the next plugin. This still can consume whole free disk space, though. Please note,
+size estimations may not be accurate for highly utilized systems due to changes between
+an estimate and a real execution.
 .TP
 .B \--upload
 If specified, attempt to upload the resulting archive to a vendor defined location.
--- a/sos/report/__init__.py
+++ b/sos/report/__init__.py
@@ -86,6 +86,7 @@ class SoSReport(SoSComponent):
         'desc': '',
         'domains': [],
         'dry_run': False,
+        'estimate_only': False,
         'experimental': False,
         'enable_plugins': [],
         'keywords': [],
@@ -137,6 +138,7 @@ class SoSReport(SoSComponent):
         self._args = args
         self.sysroot = "/"
         self.preset = None
+        self.estimated_plugsizes = {}
 
         self.print_header()
         self._set_debug()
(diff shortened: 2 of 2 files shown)


pmoravec:
Please propose improvements to variable names, manpage and UI texts - I am not much happy with the wording either where.

man/en/sos-report.1
```diff
@@ -14,7 +14,7 @@ sos report \- Collect and package diagnostic and support data
           [--preset preset] [--add-preset add_preset]\fR
           [--del-preset del_preset] [--desc description]\fR
           [--batch] [--build] [--debug] [--dry-run]\fR
-          [--label label] [--case-id id]\fR
+          [--estimate-mode] [--label label] [--case-id id]\fR
```
TurboTurtle, line 17: Since we are not generating any usable collection, perhaps `--estimate-only` here? Just to be incredibly clear users should not expect an archive.

man/en/sos-report.1
```diff
@@ -317,6 +317,16 @@ output, or string data from the system. The resulting logs may be used
...
 .TP
+.B \--estimate-mode
+Estimate disk space requirements when running sos report. This can be valuable
+to prevent sosreport working dir to consume all free disk space. No plugin data
+is available at the end.
+
+Plugins will be collected sequentially, size of collected files and commands outputs
+will be calculated and the plugin files will be immediatelly deleted prior execution
+of the next plugin. This still can consume whole free disk space, though. Also the
+estimation might not be accurate in several ways.
```
TurboTurtle, line 328: Last bit could use a little more detail. Perhaps "Please note, size estimations may not be accurate for highly utilized systems due to changes between an estimate and real execution".

sos/report/__init__.py
```diff
@@ -1402,7 +1472,7 @@ def execute(self):
                 return False
 
             self.add_manifest_data()
-            self.batch()
+            self.batch(self._check_estimate_mode())
```
TurboTurtle, line 1475: Rather than passing this as a parameter, I think it'd be better to call `_check_estimate_mode()` inside `batch()`, e.g.:

```python
def batch(self):
    msg = self.policy.get_msg()
    if self.opts.estimate_mode:
        msg += self._check_estimate_mode()
```

sos/report/__init__.py
```diff
@@ -1333,6 +1385,24 @@ def final_work(self):
             self.policy.display_results(archive, directory, checksum,
                                         map_file=map_file)
 
+        if self.opts.estimate_mode:
+            from sos.utilities import get_human_readable
+            _sum = get_human_readable(sum(self.estimated_plugsizes.values()))
+            self.ui_log.info("Estimated disk space requirement for whole sos "
+                             "report: %s" % _sum)
```
TurboTurtle, line 1392: Let's specify that this is for an _uncompressed_ archive. If we're reporting e.g. 100MB but a user then ends up with a significantly smaller tarball, I can foresee a number of reports of inaccurate estimates and us routinely explaining this difference.

"Estimated (uncompressed) disk requirements: XXMB" maybe?

TurboTurtle (commented):
Overall this looks good, just a few recommendations.

man/en/sos-report.1
pmoravec: Yes, that was the term I was looking for.

sos/report/__init__.py
pmoravec: or rather "(uncompressed) sos report directory", to be specific what is uncompressed. Ack.

TurboTurtle (approved):

bmr-cymru:
I'm concerned that this doesn't adequately address a common case with sos and ENOSPC type errors - we often see cases where just a single plugin "goes off into the weeds" (often because something has been moved around in the file system layout, so that a path that was formerly capped to some acceptable size now includes e.g. dozens of multi-GB VM images). I don't think this approach will help substantially with that kind of case, because we tend to overrun  by a very large margin so the `--estimate-mode` run would take a very long time and potentially fill up the fs.

It's quite a lot of work to separate out all the collected data that sos uses internally (e.g. enumeration of devices or network interfaces) from stuff that's "just" collected data but I can't currently see a robust way forward on this that doesn't somehow tackle that.

pmoravec:
I agree it is far from an ideal solution for those typical usages of the feature that the option shall implement. Anyway it was the best (the only reliable?) idea how to estimate the disk space that we can (easily) implement.

If even in the worst case the `sos report --estimate-only` fills up the disk, we at least know what particular plugin is responsible for it.

PR #2674, opened 2021-09-06
[foreman] Correct Apache package on Debian

ekohl:
On Debian the Apache package is named apache2 and this is also used in directory paths such as /etc/apache2 and /var/log/apache2.

---

### DIFF
--- a/sos/report/plugins/foreman.py
+++ b/sos/report/plugins/foreman.py
@@ -303,6 +303,6 @@ class RedHatForeman(Foreman, RedHatPlugin):
 
 class DebianForeman(Foreman, DebianPlugin, UbuntuPlugin):
 
-    apachepkg = 'apache'
+    apachepkg = 'apache2'
 
 # vim: set et ts=4 sw=4 :
--- a/sos/report/plugins/foreman_proxy.py
+++ b/sos/report/plugins/foreman_proxy.py
@@ -64,7 +64,7 @@ class RedHatForemanProxy(ForemanProxy, RedHatPlugin):
 
 class DebianForemanProxy(ForemanProxy, DebianPlugin, UbuntuPlugin):
 
-    apachepkg = 'apache'
+    apachepkg = 'apache2'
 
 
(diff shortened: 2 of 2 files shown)


pmoravec:
ACK to the change itself. But should not we update also:

https://github.com/sosreport/sos/blob/main/sos/report/plugins/apache.py#L112
https://github.com/sosreport/sos/blob/main/sos/report/plugins/gnocchi.py#L105

? And do we really have an invalid trigger for `apache` plugin itself on Debian? Cc @slashdd or @BryanQuigley .

ekohl:
(I took the liberty of changing your URLs to absolute URLs - you can press `y` on a page and it'll expand the reference to a commit ID. Then line numbers are guaranteed not to change and Github will inline the actual code)

> https://github.com/sosreport/sos/blob/7d025933c2cdc7bf07a7b6b9c0d8b2637ade372b/sos/report/plugins/apache.py#L112

I didn't spot this. This is what the console tells me:
```console
$ apt show apache
Package: apache
State: not a real package (virtual)
N: Can't select candidate version from package apache as it has no candidate
N: Can't select versions from package 'apache' as it is purely virtual
N: No packages found
```

So I think you're right that it may be broken altogether.

Looking deeper, the `packages` variable isn't even overridden from here:
https://github.com/sosreport/sos/blob/7d025933c2cdc7bf07a7b6b9c0d8b2637ade372b/sos/report/plugins/apache.py#L18

It's interesting that the `apachepkg` variable is used only here:
https://github.com/sosreport/sos/blob/7d025933c2cdc7bf07a7b6b9c0d8b2637ade372b/sos/report/plugins/apache.py#L47

However, it could be used in more places.

> https://github.com/sosreport/sos/blob/7d025933c2cdc7bf07a7b6b9c0d8b2637ade372b/sos/report/plugins/gnocchi.py#L105

I saw this as well and that looks wrong yes. I tried to keep this scoped to only Foreman.

TurboTurtle (approved):
Ack to this change, we'll need to make another for other places.

PR #2676, opened 2021-09-07
Add snap configuration file

arif-ali:
After discussion with the snapcraft team, this fits the bill
for a classic snap

Closes: #1992
Supersedes: #2233 

Reference: https://forum.snapcraft.io/t/25673
Reference: https://forum.snapcraft.io/t/26447

Signed-off-by: Nicolas Bock nicolas.bock@canoical.com

---

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,37 @@
+name: sosreport
+summary: Sos is an extensible, portable, support data collection tool
+description: |
+  Sos is an extensible, portable, support data collection tool
+  primarily aimed at Linux distributions and other UNIX-like operating
+  systems.
+grade: stable
+base: core20
+confinement: classic
+adopt-info: sos
+
+parts:
+  sos:
+    plugin: python
+    source: .
+    override-pull: |
+      snapcraftctl pull
+      snapcraftctl set-version $(git describe --tags --always)
+    build-packages:
+      - git
+      - python3
+      - snapcraft
+      - gettext
+
+apps:
+  sos:
+    environment:
+      PYTHONPATH: ${SNAP}/lib/python3.8
+    command: bin/sos
+  sosreport:
+    environment:
+      PYTHONPATH: ${SNAP}/lib/python3.8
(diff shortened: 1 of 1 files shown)


arif-ali:
/cc @slashdd @nicolasbock 

For testing, we can install the snap using the following command

```
snap install sosreport --channel latest/edge --classic
```

This is while we can agree to push it to stable

snap/snapcraft.yaml
```diff
@@ -0,0 +1,37 @@
+name: sosreport
```
TurboTurtle, line 1: Since this is shipping v4.0+, should this be called `sos` instead? Since we're providing the `sos` binary, of which `report` is a component, the same as `clean` and `collect`

arif-ali: I hear you, in-fact, I already requested `sos`, and that was denied on the snap store as it was too short of a name, see below. Hence why I went for sosreport.

We will still be able to use the sos command, once the aliasing will be automatically done as per the second reference in the commit.

> Your "sos" registration has been revoked.
>
> The reviewer provided the following feedback:
>
> Hello,
>
> Three-letter and shorter snap names are too generic and nondescript and don't provide a good idea of what the snap actually is or does. For that reason, we prefer and suggest you pick a more descriptive name for your snap (sos-collector for example; sosreport would be ideal but I understand the tool can do more than just generate reports). Keep in mind that even if the snap has one of those names, a command called "sos" can still be provided by the snap - this needs creating an "alias" but it's entirely feasible to do so.
> 
> Let me know if some of the suggested names are OK with you or if you want to pick a new one; in any case, you can simply register the new name and it should be automatically approved - only 3-letter or shorter names (well and a few other exceptions) require manual name review/approval.

TurboTurtle: Well that seems rather.....arbitrary.

Regardless it's beyond our control, so ack.

TurboTurtle (approved):

slashdd (approved):
lgtm.

PR #2680, opened 2021-09-08
[Ceph] Add support for containerized Ceph setup

bvassova:
RH Ceph 4 can either be installed as RPM or as containers.
The changes permits to collect the ceph config
on both kind of setup.

Resolves: #2680
Related: #2646

---

### DIFF
--- a/sos/plugins/ceph.py
+++ b/sos/plugins/ceph.py
@@ -8,6 +8,7 @@
 
 from sos.plugins import Plugin, RedHatPlugin, UbuntuPlugin
 from socket import gethostname
+import re
 
 
 class Ceph(Plugin, RedHatPlugin, UbuntuPlugin):
@@ -15,7 +16,8 @@ class Ceph(Plugin, RedHatPlugin, UbuntuPlugin):
     """
 
     plugin_name = 'ceph'
-    profiles = ('storage', 'virt')
+    profiles = ('storage', 'virt', 'container')
+    containers = ('ceph-(mon|rgw|osd)*',)
     ceph_hostname = gethostname()
 
     packages = (
@@ -35,6 +37,12 @@ class Ceph(Plugin, RedHatPlugin, UbuntuPlugin):
         'ceph-mgr@%s' % ceph_hostname,
         'ceph-radosgw@*',
         'ceph-osd@*'
+    )
+
+    # This check will enable the plugin regardless of being
+    # containerized or not
+    files = (
+        '/etc/ceph/ceph.conf',
     )
 
     def setup(self):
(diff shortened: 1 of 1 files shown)


TurboTurtle (approved):

sos/plugins/ceph.py
```diff
@@ -8,14 +8,16 @@
...
 
 
 class Ceph(Plugin, RedHatPlugin, UbuntuPlugin):
     """CEPH distributed storage
     """
 
     plugin_name = 'ceph'
-    profiles = ('storage', 'virt')
+    profiles = ('storage', 'virt', 'container')
+    containers = ('ceph-(mon|rgw|osd)*',)
```
pmoravec, line 20: Sadly, legacy-3.9 does not have the ability to enable a plugin based on a container presence. Cf. https://github.com/sosreport/sos/blob/legacy-3.9/sos/plugins/__init__.py#L1603 .

https://github.com/sosreport/sos/commit/c10c1a39a338cb52ca398a5568b1dc31001612fa needs to be backported from main to legacy-3.9 to have this working.

TurboTurtle: Oh, good catch!

pmoravec (approved):
ACK, assuming #2687 will be merged first (or its commit added to this branch).

TurboTurtle (changes requested):
Upon further review, 3.9 does not contain the container runtime abstraction at all. I had originally thought that went in at the end of 3.9 but I was wrong there.

Without that abstraction, none of the container-minded changes here will actually function.

pmoravec:
An idea / code snippet to overcome this:

    ceph_containers_re = 'ceph-*' # I guess it is a *safe* enhancement of   containers = ('ceph-(mon|rgw|osd)*',)  that we need for 'self.get_all_containers_by_regex("ceph-*")'

    def _find_ceph_containers(self):
        self.ceph_conatiners = []
        _runtime = if executable('docker') then 'docker' elif executable('podman') then 'podman' else None    #TODO rewrite to python language :)
        if _runtime is not None:
            out = sos_get_command_output('%s ps' % _runtime)     # TODO: "ps -a"? or just running? check what we really use
            if out['status'] == 0:
                for ent in out['output'].splitlines()[1:]:
                    # cotainer name is the latest item
                    ent = ent.split()[-1]
                    if re.match(ceph_containers_re, c[1]):
                        self.ceph_conatiners.append(ent)

    def check_enabled(self):
        self._find_ceph_containers()
        if self.ceph_conatiners:
          return True
        return False  # TODO: call parent class method to check enabledness per services or packages

arif-ali:
@bvassova Any plans on making the improvements as @pmoravec suggested.

With this back-port to legacy-3.9, not sure how many systems out there are running 3.9 that can benefit from this now

arif-ali:
Looking at the following URL (if it's correct), there's only 3 distros using 3.9.1, which I doubt are supported anymore.

https://repology.org/project/sos-system-data-collection/badges

bvassova:
@arif-ali @jcastill we discussed this with @pmoravec and since rhel 7 is end of maintenance now, this issue might probably be closed

arif-ali:
@bvassova thanks for the update, I'll close this PR then; if however, there is a new update, then don't hesitate in re-opening, and we can continue work on this.

PR #2682, opened 2021-09-09
[cephadm] add plugin to collect cephadm config, logs and diagnostic o…

haklein:
…utput

Signed-off-by: Harald Klein <hari@vt100.at>

---

### DIFF
--- a/sos/report/plugins/cephadm.py
+++ b/sos/report/plugins/cephadm.py
@@ -1,0 +1,55 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, RedHatPlugin, UbuntuPlugin
+
+
+class Cephadm(Plugin, RedHatPlugin, UbuntuPlugin):
+
+    short_desc = 'CEPH distributed storage managed via cephadm'
+
+    plugin_name = 'cephadm'
+    profiles = ('storage')
+
+    packages = (
+        'cephadm'
+    )
+
+    def setup(self):
+        all_logs = self.get_option("all_logs")
+
+        if not all_logs:
+            self.add_copy_spec([
+                "/var/log/ceph/cephadm.log"
+            ])
+        else:
+            self.add_copy_spec([
+                "/var/log/ceph/cephadm*"
+            ])
+
+        self.add_copy_spec([
+            "/etc/ceph/"
+        ])
+
+        cephadm_cmds = [
+            "version",
+            "gather-facts",
+            "ls",
+            "ceph-volume lvm list",
+            "check-host"
+        ]
+
+        for s in cephadm_cmds:
+            self.add_cmd_output(cmds="cephadm --timeout 5 %s" % s, suggest_filename="%s" % s)
+
+        self.add_forbidden_path([
+            "/etc/ceph/*keyring*",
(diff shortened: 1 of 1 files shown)


sos/report/plugins/cephadm.py
```diff
@@ -0,0 +1,55 @@
...
+class Cephadm(Plugin, RedHatPlugin, UbuntuPlugin):
+
+    short_desc = 'CEPH distributed storage managed via cephadm'
+
+    plugin_name = 'cephadm'
+    profiles = ('storage')
+
+    packages = (
+        'cephadm'
+    )
```
TurboTurtle, line 21: These need to be tuples, so you'll need a trailing `,` even for 1-tuples.  E.g.

```python
profiles = ('storage',)
packages = ('cephadm',)
```

sos/report/plugins/cephadm.py
```diff
@@ -0,0 +1,55 @@
...
+        all_logs = self.get_option("all_logs")
+
+        if not all_logs:
+            self.add_copy_spec([
+                "/var/log/ceph/cephadm.log"
+            ])
+        else:
+            self.add_copy_spec([
+                "/var/log/ceph/cephadm*"
+            ])
```
TurboTurtle, line 33: This is already captured by the `ceph` plugin.

sos/report/plugins/cephadm.py
```diff
@@ -0,0 +1,55 @@
...
+
+        self.add_copy_spec([
+            "/etc/ceph/"
+        ])
```
TurboTurtle, line 37: Likewise, this is captured by the `ceph` plugin today.

sos/report/plugins/cephadm.py
```diff
@@ -0,0 +1,55 @@
...
+        cephadm_cmds = [
+            "version",
+            "gather-facts",
+            "ls",
+            "ceph-volume lvm list",
+            "check-host"
+        ]
+
+        for s in cephadm_cmds:
+            self.add_cmd_output(cmds="cephadm --timeout 5 %s" % s, suggest_filename="%s" % s)
```
TurboTurtle, line 48: `add_cmd_output()` supports a `timeout` parameter, so you don't need to use the `cephadm` option here or do any filename mangling. `self.add_cmd_output(["cephadm %s" % cmd for cmd in cephadm_cmds], timeout=5)`

TurboTurtle (commented):
There's a lot of overlap with the existing `ceph` plugin - would it not make sense to add the `cephadm` commands there, or should we instead be looking at potentially breaking up the `ceph` plugin and moving some of the items there into this (and/or other plugins)?

sos/report/plugins/cephadm.py
haklein: `cephadm` is only available in very recent Ceph versions. Hence I was considering a separate plugin. Thanks for the comment on the tuples.

sos/report/plugins/cephadm.py
haklein: This would only be captured via the `ceph` plugin when the `ceph-common` package is installed (to trigger the ceph plugin), which might not be the case in all situations (there is no package dependency between `cephadm` and `ceph-common`. `ceph-common` would usually be installed via playbook from `cephadm-ansible`).

sos/report/plugins/cephadm.py
haklein: Isn't the "native" timeout of a command preferred? cephadm would give some hints on where the timeout did occur (e.g. if it couldn't acquire a lock).

arif-ali:
@haklein were you planning to work on this. I think having `cephadm` data in the ceph plugins could be useful

TurboTurtle:
Closing as abandoned.

Issue #2668, opened 2021-08-31
[request] Add juju ssh support in sos-collect


slashdd:
In a Canonical juju context, nodes part of a Juju deployment can be accessible via 'juju ssh' command.

Might be nice for Canonical support to have sos-collect support 'juju ssh' to collect sosreport on multiples nodes accross a juju model.

https://juju.is/docs/olm/accessing-individual-machines-with-ssh

```
Example:
juju ssh mysql/0
```

TurboTurtle:
Would this be a change "only" in the execution of the actual `ssh` commands used to connect to the nodes, or would this be (or **also** be) a new cluster profile? Would remote command execution also use the `juju` command?

If it's the former, we'll need to devise a way to toggle `SoSNode` between building traditional `ssh` commands and `juju` commands for:

* Creating the initial connection - on that point does `juju` support something akin to Control Persist where a "background" connection is kept open and subsequent commands are routed through that connection rather than creating a whole new one for every command?
    * Do we need to have special considerations for authentication?
* Capturing and handling command results/ouput (currently we rely on pexpect for traditional `ssh`)
* Remote file copying


I'm all for extending this functionality where it would be useful, but I want to make sure we scope this correctly from the get-go.

slashdd:
When using Juju’s ssh (or scp) command, Juju’s internal user rights system imposes an extra degree of security by permitting access solely from a Juju user. This user must also have ‘admin’ model access. If 'user' is specified then the connection is made to that user account; otherwise, the default 'ubuntu' account, created by Juju, is used.

>> Would remote command execution also use the juju command?

No, `juju ssh mysql/0 sos report ...` will suffice.

>> Creating the initial connection - on that point does juju support something akin to Control Persist where a "background" connection is kept open and subsequent commands are routed through that connection rather than creating a whole new one for every command?

```
juju ssh mysql/0
juju ssh mysql/1
and so on would need to be executed from the juju "jumphost"
```
Let's say we add the `juju ssh` support and make sure the node's name can handle service/unit_number (mysql/0), then it should collect like any sos-collect standard SSH.

TurboTurtle:
> No, `juju ssh mysql/0 sos report ...` will suffice.
> 

Does this echo command output and reflect return codes like traditional ssh? If so, we can probably retain our current pexpect handling which would be awesome.


> ```
> juju ssh mysql/0
> juju ssh mysql/1
> and so on would need to be executed from the juju "jumphost"
> ```
> 

[...]

PR #2684, opened 2021-09-10
[collect] Abstract transport protocol from SoSNode

TurboTurtle:
Since its addition to sos, collect has assumed the use of a system
installation of SSH in order to connect to the nodes identified for
collection. However, there may be use cases and desires to use other
transport protocols.

As such, provide an abstraction for these protocols in the form of the
new `RemoteTransport` class that `SoSNode` will now leverage. So far an
abstraction for the currently used SSH ControlPersist function is
provided, along with a psuedo abstraction for local execution so that
SoSNode does not directly need to make more "if local then foo" checks
than are absolutely necessary.

Related: #2668

---

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -101,8 +101,8 @@ setup(
         'sos.policies.distros', 'sos.policies.runtimes',
         'sos.policies.package_managers', 'sos.policies.init_systems',
         'sos.report', 'sos.report.plugins', 'sos.collector',
-        'sos.collector.clusters', 'sos.cleaner', 'sos.cleaner.mappings',
-        'sos.cleaner.parsers', 'sos.cleaner.archives'
+        'sos.collector.clusters', 'sos.collector.transports', 'sos.cleaner',
+        'sos.cleaner.mappings', 'sos.cleaner.parsers', 'sos.cleaner.archives'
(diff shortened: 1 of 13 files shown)


TurboTurtle:
Found a small bug in `reconnect()` that would have not incremented the attempt counter if the transport returned anything other than True/False. Fixed, re-pushed.

TurboTurtle:
Packit failure for F35 x86 appears unrelated - the build environment for that one ran out of space. The others all passed.

pmoravec (approved):
Sounds good to me but I would prefer another pair of eyes to look at it.

(I also run a few test cases and all went fine)

TurboTurtle:
Agreed, this is a rather significant change so I'd like a bit more review, especially from the perspective of being able to build off of it, from the other parties.

@slashdd @BryanQuigley

TurboTurtle:
Added some better handling of non-root execution for detecting and sending sudo or su passwords.

BryanQuigley:
This makes sense to me on a high level - didn't dive deep though.

TurboTurtle:
@slashdd does this look workable to you for adding a `juju` transport?

PR #2687, opened 2021-09-14
[Plugin] Allow plugin activation by container presence

bvassova:
Adds an enablement trigger tuple, `containers`, that allows for
enabling a plugin based on the presence of a _running_ container
matching a given name or regex. Note that this is only functional
for the default runtime discovered by a policy during initialization.

Resolves: #2687
Related: #2680

---

### DIFF
--- a/sos/plugins/__init__.py
+++ b/sos/plugins/__init__.py
@@ -424,6 +424,7 @@ class Plugin(object):
     commands = ()
     kernel_mods = ()
     services = ()
+    containers = ()
     architectures = None
     archive = None
     profiles = ()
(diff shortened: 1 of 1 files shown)


pmoravec (approved):
ACK. Though as this is required to backport purely due to #2680, it is generally better to update that PR by adding this commit there.

TurboTurtle (changes requested):
The container runtime bits that enable all this to actually work are not present in sos-3.9. The abstraction was added in sos-4.0, via a 7-patch set - https://github.com/sosreport/sos/pull/1873 - which has since been updated several times making any needed backport fairly large.

Our backporting policy is for bugs, not backporting new features. This for me is a strong nack.

PR #2690, opened 2021-09-17
[ceph] split the ceph plugin into multiple plugins

nkshirsagar:
Split the ceph plugin into individual plugins
for each service.

Closes: #1945

---


sos/report/plugins/ceph_mds.py
```diff
@@ -0,0 +1,107 @@
...
+            inner_files = os.listdir(mds_folder)
+            for item in inner_files:
+                full_path = mds_folder + "/" + item
+                if(os.path.isdir(full_path)):
+                    # verify the folder has files
+                    if not os.listdir(full_path):
+                        continue
+                    else:
+                        return True
+        return False
```
TurboTurtle, line 48: I believe this can all be reduced to the following, for each of the new plugins:

`return True if glob.glob('/var/lib/ceph/mds/*/*') else False`

as `glob.glob` will return `[]` if there is nothing matched at the final location, regardless of if parent directories exist or not.

sos/report/plugins/ceph_mds.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def setup(self):
+        # all_logs = self.get_option("all_logs")
```
TurboTurtle, line 51: if this is unused, let's just remove it entirely.

sos/report/plugins/ceph_mds.py
```diff
@@ -0,0 +1,107 @@
...
+    plugin_name = 'ceph_mds'
+    profiles = ('storage', 'virt', 'container')
+
+    containers = ('ceph-fs.*',)
+
+    ceph_hostname = gethostname()
+
+    services = (
+        'ceph-mds@%s' % ceph_hostname,
+    )
```
TurboTurtle, line 27: Stylistically this can all be on one line (ditto for the other new plugins).

sos/report/plugins/ceph_rgw.py
```diff
@@ -0,0 +1,100 @@
...
+    def setup(self):
+        # all_logs = self.get_option("all_logs")
+
+        self.add_file_tags({
+            '/var/log/ceph/ceph-client.rgw*.log': 'ceph_rgw_log',
+        })
+
+        self.add_copy_spec([
+                "/var/log/ceph/ceph-client.rgw*.log",
+        ])
```
TurboTurtle, line 55: Since we have a single spec here, this can be condensed:

```python
self.add_copy_spec('/var/log/ceph/ceph-cleint.rgw*.log', tags='ceph_rgq_log')
```

sos/report/plugins/ceph_rgw.py
```diff
@@ -0,0 +1,100 @@
...
+        #        self.add_cmd_output([
+        #            self.fmt_container_cmd(container[1], "ceph %s" % s)
+        #            for s in ceph_cmds
+        #        ])
+        #        break
+        # Not containerized
+        # else:
+        #   self.add_cmd_output([
+        #       "ceph %s" % s for s in ceph_cmds
+        #    ])
```
TurboTurtle, line 98: If this isn't relevant to the plugin, let's remove this entirely rather than comment it out. Same for lines 62-69 above.

TurboTurtle (changes requested):
Overall this is going in the right direction, and splitting up this plugin will likely be more helpful as more ceph components or collections are added.

A few stylistic changes noted, as well as a much reduced `check_enabled()` override - though we should probably look at adding glob support for `files` again. I don't recall if there are any historical reasons we don't support them today.

Beyond that, the only thing that needs to happen is squashing the fixup commits, as we don't allow those in the project. If you aren't sure how to do this, then in your branch you can run `git rebase -i HEAD~4`  (or however many commits there are from HEAD) and in the editor for each fixup commit change `pick` to `fixup`, then save and git will fold the fixup commits into your first commit.

sos/report/plugins/ceph_mds.py
nkshirsagar: ack, I will test this approach

sos/report/plugins/ceph_mds.py
nkshirsagar: ack thank you I will make these changes

sos/report/plugins/ceph_rgw.py
nkshirsagar: I will keep this and in fact uncomment it. Instead of "ceph xxx" commands, I plan to collect "rados xxx" commands in the rgw plugi, like for eg "rados lspools", etc. If they might need the "fmt_container_cmd()" invokation instead of simply invoking through a "self.add_cmd_output()" call, this loop will be changed accordingly. I'm trying to understand why "some" ceph commands are added directly using "add_cmd_output() (without checking if its a container) while some need the fmt_container_cmd if its running in a container. Once I resolve that, I'll update this code accordingly.

sos/report/plugins/ceph_mds.py
nkshirsagar: ack. I will remove this in the individual plugins. Its also left in ceph_common just for the calamari logs (entire /var/log/calamari folder vs the log files only).

Since the files to collect are now spread over many plugins, common doesn't collect any other full folders. The individual plugins will collect the /var/lib/ceph/XXX and /var/lib/ceph/bootstrap-XXX/ folders, (for eg. osd will collect the folder /var/lib/ceph/osd) but those are not large sized folders.

sos/report/plugins/ceph_mds.py
nkshirsagar: ack thank you

sos/report/plugins/ceph_rgw.py
nkshirsagar: ack thank you

sos/report/plugins/ceph_rgw.py
nkshirsagar: ack.

TurboTurtle (approved):
LGTM. Packit failures are unrelated.

sos/report/plugins/ceph_osd.py
```diff
@@ -0,0 +1,100 @@
...
+            "osd dump",
+            "osd df",
+            "osd perf",
+            "osd blocked-by",
+            "osd pool ls detail",
+            "osd pool autoscale-status",
+            "osd numa-status"
+        ]
+
+        # is this not to be collected when no containers?
```
pponnuvel, line 60: I suspect this was an oversight. I hink you can add this to `ceph_cmds` instead.

sos/report/plugins/ceph_osd.py
```diff
@@ -0,0 +1,100 @@
...
+
+        self.add_cmd_output([
+            "ceph osd metadata",
+            "ceph osd erasure-code-profile ls",
+            "ceph osd crush show-tunables",
+            "ceph osd crush dump"
+        ])
+
+        ceph_cmds = [
+            "osd tree",
```
pponnuvel, line 48: I think most of these commands, if not all,  are actually meant to be run on the mon. You can test by running these commands directly on a osd node.

(With admin keys, you can anything anywhere - typically you'd expect to have it only on mon nodes).

sos/report/plugins/ceph_rgw.py
```diff
@@ -0,0 +1,44 @@
...
+    containers = ('ceph-rgw.*',)
+
+    def check_enabled(self):
+        return True if glob.glob('/var/lib/ceph/radosgw/*/*') else False
+
+    def setup(self):
+        self.add_copy_spec('/var/log/ceph/ceph-client.rgw*.log',
+                           tags='ceph_rgw_log')
+
+        # stub for now
```
pponnuvel, line 28: I think this stub can be removed. Commands can be included when they are needed.

sos/report/plugins/ceph_common.py
```diff
@@ -41,57 +42,38 @@ class Ceph(Plugin, RedHatPlugin, UbuntuPlugin):
...
             "/etc/calamari/",
-            "/var/lib/ceph/",
-            "/run/ceph/"
+            "/var/lib/ceph/tmp/",
         ])
 
         self.add_cmd_output([
-            "ceph mon stat",
-            "ceph mon_status",
             "ceph quorum_status",
```
pponnuvel, line 69: Does the ceph_common meant to be run on all different nodes (OSD, mon, mgr, etc)?

If so, most of these commands probably need to go elsewhere such as mon & mgr.

pponnuvel, line 69: move to mon?

sos/report/plugins/ceph_common.py
```diff
@@ -41,57 +42,38 @@ class Ceph(Plugin, RedHatPlugin, UbuntuPlugin):
...
-            "ceph mgr module ls",
-            "ceph mgr metadata",
             "ceph balancer status",
```
pponnuvel, line 70: Move to mgr?

pponnuvel (commented):
Looks OK. 

But I think a number of the commands aren't in the right plugin and need to be moved around to the right ones. 

Check the documentation to see what node  (OSD, mon, mgr, etc) a given command belongs to. Deploying a simple ceph cluster and running each command directly would help verify.

sos/report/plugins/ceph_osd.py
nkshirsagar: ack.  agreed, these would only run on a mon, i will move them all there. So then we have a few commands that run on OSDs like  ceph-disk list and ceph-volume inventory , ceph -v that runs in common.. they do not need a keyring passed to them. Shall I look for more examples, so we have at least one or two commands in each plugin for now? Or for the mgr/rgw/mds would you suggest keeping them totally empty stubs for now, with just some specific log files collected?

sos/report/plugins/ceph_rgw.py
nkshirsagar: ack

sos/report/plugins/ceph_common.py
nkshirsagar: yes common will run on all nodes. Yes, ack.

nkshirsagar: ack

sos/report/plugins/ceph_common.py
nkshirsagar: ack

sos/report/plugins/ceph_osd.py
nkshirsagar: Understood. But then except mon, everything else mostly becomes an empty stub for now and the ceph_common one just would collect just ceph -v for now, and logs like  

            "/var/log/ceph/ceph.audit.log*",
            "/var/log/calamari/*.log",
            "/etc/ceph/",
            "/etc/calamari/",
            "/var/lib/ceph/tmp/",

Does that sound ok?

I also think the OSD plugin could collect command output like ceph-disk list and ceph-volume inventory? Those would run without needing the admin keys. If you don't want to add any new ones though, osd would become a stub as well since there's nothing that wouldn't run on mon (but would on OSD) in the current list.

pponnuvel: > [...]
> Does that sound ok?

Yes, that's OK. OSD list wouldn't be empty - at least, `ceph-volume lvm list` and `ceph-disk list` commands can be run OSD.

sos/report/plugins/ceph_common.py
```diff
@@ -41,57 +42,38 @@ class Ceph(Plugin, RedHatPlugin, UbuntuPlugin):
...
+    files = ('/etc/ceph/ceph.conf',)
 
     def setup(self):
         all_logs = self.get_option("all_logs")
 
         self.add_file_tags({
             '.*/ceph.conf': 'ceph_conf',
-            '/var/log/ceph/ceph.log.*': 'ceph_log',
-            '/var/log/ceph/ceph-osd.*.log': 'ceph_osd_log'
+            '/var/log/ceph/ceph.log*': 'ceph_log',
```
dosaboy, line 52: by changing .* to * you are changing this from a regex to a glob. Is that intentional? The method at [1] suggests this should be a regex pattern.

[1] https://github.com/sosreport/sos/blob/aaeb8cb57ed55598ab744b96d4f127aedebcb292/sos/report/plugins/__init__.py#L1350

dosaboy (changes requested):

sos/report/plugins/ceph_common.py
nkshirsagar: thanks for spotting it, its a mistake. Shall correct it!

sos/report/plugins/ceph_osd.py
nkshirsagar: ack

sos/report/plugins/ceph_osd.py
nkshirsagar: ack

sos/report/plugins/ceph_common.py
```diff
@@ -0,0 +1,84 @@
...
+    # This check will enable the plugin regardless of being
+    # containerized or not
+    files = ('/etc/ceph/ceph.conf',)
+
+    def setup(self):
+        all_logs = self.get_option("all_logs")
+
+        self.add_file_tags({
+            '.*/ceph.conf': 'ceph_conf',
+            '/var/log/ceph/ceph.log.*': 'ceph_log',
```
TurboTurtle, line 51: Just noticed - the tagging for this is in `ceph_common`, but not the collection for `/var/log/ceph/ceph.log`. However, the collection is in `ceph_mon`, `ceph_mgr`, and `ceph_mds`.

Should the collection for `/var/log/ceph/ceph.log` not be solely in `ceph_common`?

nkshirsagar: The reason I did not put the collection of /var/log/ceph/ceph.log in ceph_common is because rgw and osd nodes do not have the /var/log/ceph/ceph.log file. But common does run on each node and the file would exist if it were to be running on mon/mgr/mds nodes, so I could put the collection of that /var/log/ceph/ceph.log in common if you think that makes more sense.

slashdd (approved):
LGTM.
I ran the plugins in a cloud environment running ceph (osd, mgr, mds, mon, ...) and it work well /wo no errors found in sos_logs/

sos/report/plugins/ceph_rgw.py
```diff
@@ -0,0 +1,103 @@
...
+from socket import gethostname
+import re,os
+
+
+class CephRGW(Plugin, RedHatPlugin, UbuntuPlugin):
+
+    short_desc = 'CEPH rgw'
+
+    plugin_name = 'ceph_rgw'
+    profiles = ('storage', 'virt', 'container')
```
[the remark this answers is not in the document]
nkshirsagar: ack. this is done.

PR #2695, opened 2021-09-22
[origin] Removing unused configuration

vwalek:
Removed the diagnostics part as it is no longer maintained and doesn't
work on Openshift.

Adding additional projects to collect.

Removed getting all namespaces as it is not needed for troubleshooting
and project names are sensitive for some customers.

Adding condition to collect the logs from systemd openshift services if
not running as static pods.

Signed-off-by: Vladislav Walek <22072258+vwalek@users.noreply.github.com>

---

### DIFF
--- a/sos/report/plugins/origin.py
+++ b/sos/report/plugins/origin.py
@@ -8,7 +8,7 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.report.plugins import Plugin, RedHatPlugin, PluginOpt
+from sos.report.plugins import Plugin, RedHatPlugin
 import os.path
 
 # This plugin collects static configuration and runtime information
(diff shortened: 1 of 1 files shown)


sos/report/plugins/origin.py
```diff
@@ -40,14 +40,7 @@ class OpenShiftOrigin(Plugin):
...
 
-    option_list = [
-        PluginOpt('diag', default=True,
-                  desc='Collect oc adm diagnostics output'),
-        PluginOpt('diag-prevent', default=True,
-                  desc='Use --prevent-modification with oc adm diagnostics'),
-        PluginOpt('all-namespaces', default=False,
-                  desc='collect dc output for all namespaces')
-    ]
+    option_list = []
```
pmoravec, line 43: No need of this line, as `Plugin` class already sets that (https://github.com/sosreport/sos/blob/main/sos/report/plugins/__init__.py#L551).

sos/report/plugins/origin.py
```diff
@@ -145,32 +141,26 @@ def setup(self):
             ]
 
             self.add_cmd_output([
-                '%s get -o json %s' % (oc_cmd_admin, jcmd) for jcmd in jcmds
+                '%s get -o yaml %s' % (oc_cmd_admin, jcmd) for jcmd in jcmds
```
TurboTurtle, line 144: Is there anything currently using the json outputs? Anything in Insights or other tooling?


Also, the `postproc()` method will need to be updated to account for not only the file name change but also any corrections to ensure it keeps working.

TurboTurtle (changes requested):

sos/report/plugins/origin.py
vwalek: Ok, I will remove that line. Thx

sos/report/plugins/origin.py
vwalek: Hey, I will put it back to `json`. I put it as `yaml`, but it should be `json`.

pmoravec (approved):

TurboTurtle (approved):

Issue #2696, opened 2021-09-23
Fix OVN DBs schema path


pmannidi:
currently, the OVN DBs schema path is set /usr/share/openvswitch. It should have been /usr/share/ovn

pmoravec:
Does it apply to all versions of OpenVSwitch? Or just for some newer ones?

Cc @amorenoz for another view.

amorenoz:
> Does it apply to all versions of OpenVSwitch? Or just for some newer ones?

The default path is, indeed, version-dependent (2.11 still uses `/usr/share/openvswitch`). Furthermore, I'd say it is also distribution-dependent.
So both paths should be checked.

Also note that we're duplicating most of this information anyway since we're collecting the entire db file (which is safer IMHO).

PR #2698, opened 2021-09-23
[ovn_central] Fix OVN DBs Schema Path

pmannidi:
Currently sosreport tries to search for OVN DB schemas in
/usr/share/openvswitch instead of /usr/share/ovn. Since the schemas are
not available in /usr/share/openvswitch, they are not collected.

This patch fixes the schema path.

Resolves: #2696

Signed-off-by: Purandhar Sairam Mannidi <pmannidi@redhat.com>

---

- [X Are any related Issues or existing PRs [properly referenced](https://docs.github.com/en/issues/tracking-your-work-with-issues/creating-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) via a Closes (Issue) or Resolved (PR) line?

### DIFF
--- a/sos/report/plugins/ovn_central.py
+++ b/sos/report/plugins/ovn_central.py
@@ -102,7 +102,7 @@ class OVNCentral(Plugin):
             'ovn-sbctl get-connection',
         ]
 
-        schema_dir = '/usr/share/openvswitch'
+        schema_dir = '/usr/share/ovn'
 
         nb_tables = self.get_tables_from_schema(os.path.join(
             schema_dir, 'ovn-nb.ovsschema'))


TurboTurtle (commented):
The change itself is fine, but I want to verify if this was __never__ the correct path, or if the path is version dependent?

pmoravec:
Cc @amorenoz and @putnopvut (per git blame) - could you please comment about the history of the OVN DBs schema path? (we want to avoid situation that `sos` with updated path will be run on some older OVN with the old/current path, which would prevent collecting some data).

amorenoz:
@pmoravec see https://github.com/sosreport/sos/issues/2696#issuecomment-926370584

amorenoz:
In summary, both paths should be checked as old versions still use `/usr/share/openvswitch`

pmoravec:
Thanks for the comment.

> The default path is, indeed, version-dependent (2.11 still uses `/usr/share/openvswitch`). Furthermore, I'd say it is also distribution-dependent.
> So both paths should be checked.

Just these two paths can exist? So is it worth to have a code like:

    for path in [`/usr/share/openvswitch`, '/usr/share/ovn']:
        if os.path.isdir(path):
            self.get_tables_from_schema(..)  #  for both calls of the methos

? (the `os.path.isdir` will prevent redundant errors "Cannot parse JSON file" or "Could not open DB schema file")

> 
> Also note that we're duplicating most of this information anyway since we're collecting the entire db file (which is safer IMHO).

Most or all information duplicated? If most, it makes sense to collect also this (to still get *something* new). If all, does it make sense to collect the schema?

amorenoz:
> Thanks for the comment.
> 
> > The default path is, indeed, version-dependent (2.11 still uses `/usr/share/openvswitch`). Furthermore, I'd say it is also distribution-dependent.
> > So both paths should be checked.
> 
> Just these two paths can exist? So is it worth to have a code like:
> 

Those are the two default ones (for old and new versions)

Having glanced over some distro packages, I don't see any other path being used:
https://rpmfind.net/linux/RPM/opensuse/15.3/aarch64/openvswitch-ovn-central-2.11.5-3.15.3.aarch64.html
https://packages.ubuntu.com/focal/amd64/ovn-central/filelist
https://cbs.centos.org/koji/rpminfo?rpmID=270672

So I'd say yes.


> ```
> for path in [`/usr/share/openvswitch`, '/usr/share/ovn']:
>     if os.path.isdir(path):
>         self.get_tables_from_schema(..)  #  for both calls of the methos
> ```
> 
> ? (the `os.path.isdir` will prevent redundant errors "Cannot parse JSON file" or "Could not open DB schema file")
> 

Yes, this looks good to me

> > Also note that we're duplicating most of this information anyway since we're collecting the entire db file (which is safer IMHO).
> 
> Most or all information duplicated? If most, it makes sense to collect also this (to still get _something_ new). If all, does it make sense to collect the schema?

All of the info from table list is redundant compared with collecting the db file. Maybe @pmannidi can comment but, the way I see it is:
 
Adding redundant commands that has relatively small output and provides a good, quick overview of the OVN config makes sense (e.g: "ovn-nbcl show"). For anything more deep, I think starting an ovsdb-server on the copied `ovnnb_db.db` file would be the best way to inspect OVN data. That would allow you to perform filtering (and of course dump all tables just like the script does).

That approach comes with two extra benefits: smaller/faster sosreport, and less potentially intrusive interaction with the ovsdb-server who is single-threaded.

TurboTurtle:
I'm ok with keeping both collections if the ability to analyze the actual db file has some setup behind it, whereas the other collections could be used for quick analysis.

As for the schema_dir, yes we'll want to keep both dirs for version compatibility. The loop Pavel suggests looks fine - as a note to myself we need to extend plugin's `os.path` wrappers to include `os.path.join()` to account for non-default sysroots.

pmannidi:
After looking at the big picture, capturing the schema files is not going to help us figure out much anyway and moreover we are capturing OVN DBs as is. In my opinion, its best to remove the logic of collecting schema as it's of no use and can be made redundant.

arif-ali:
@pmannidi is this PR now redundant, or does it need any further work?

arif-ali:
@amorenoz thanks for double checking, in that case I'll close this one off then, if that's ok, if needs doing then it can be re-opened

Issue #2237, opened 2020-09-16
PackageManager support, how we want it to evolve


BryanQuigley:
I'd like to see sos plugins being able to package trigger from rpm/dpkg/flatpak/snaps/etc.  From my attempt at consolidating the OSes I think we really need to stop having packages related to OSes. I can install rpms, snaps, flatpaks and debs on one machine - sos should be able to handle that.

I see two possible ways to do this:
1. define a new series of policies, pkg_deb.py, pkg_rpm.py, etc that trigger based on either file existing or path. Similar to what we have for OSes today (which might need to be renamed to os_debian etc).
2. Add the package query code for each package system to the core sos code.

1 is more extensible but in practice the set of package managers stays pretty constant. My slight preference is for #2 as it let's us put all that logic in one place.

TurboTurtle:
We do have a third option, allow policies to define multiple package managers and iterate over them for queries (and/or build out ways to specify a package manager for each query like we do with `ContainerRuntime`s). Then have the specific policies dictate which package managers they want to use/support. This is kind of like option 2, but provides more flexibility for us and downstreams.

For the RH family there are a couple different views. For Fedora, snaps and flatpaks are available and used to varying degrees, but it is considered bad practice to install debs. For RHEL, anything beyond RPMs is strongly discouraged and unsupported. I imagine the same is true for Suse and Amazon Linux as well.

The way sos behaves on a particular distribution should match that distribution's best practices, which is why we have policies at all - so we don't have a one-size-fits-all approach.

BryanQuigley:
I don't see the use case where one would not want to collect the deb information from a RHEL system.  That would be very useful to help support realize there is something very wrong - and suggest what that person should do instead.

I do like the concept of OS policies being able to specify what package managers *should* be the only ones present and highlight when that is not the case.

TurboTurtle:
> I don't see the use case where one would not want to collect the deb information from a RHEL system. That would be very useful to help support realize there is something very wrong - and suggest what that person should do instead.
> 

Assume we have plugin `foo` that collects a bunch of command output. If I get a support case dealing with `foo`, let's say getting "bad" output from some command in production, I'm going to first look for/in `sos_commands/foo` for the output I expect. If that is present, I can assume that `foo` is at least installed properly, since sos activated the plugin.

[...]

PR #2706, opened 2021-09-29
[gcp] Adding Google Cloud plugin

### LINKED ISSUE #2699: [gcp] New plugin for Google Cloud Platform
Hi!

I'm planning to write a new plugin to better support Virtual Machines running on Google Cloud Platform. Similar to what is already implemented for [Azure](https://github.com/sosreport/sos/blob/main/sos/report/plugins/azure.py).

I plan to capture following info:
 * All information that can be extracted from [Metadata server](https://cloud.google.com/compute/docs/metadata/overview).
 * Get output of `gcloud auth list` - if it's available
 * Grab journal logs for any Google-related services 

When would it run?
Only on VMs running on GCP - meaning a `curl` to `http://metadata.google.internal/computeMetadata/v1/` is successful.

My questions:
 * Is `gcp` a good name for such plugin, or would `google-cloud` be better?
 * Should all this data be collected inside one plugin, or is it better to divide those into for example: `gcp-metadata` and `gcp-services`?

m-strzelczyk:
Adding a plugin for Google Cloud Compute Engine VMs.
The plugin will collect data about Google services
running on the system (journalctl -u google*), data from
the Metadata Server that's available for every instance
and output of `gcloud auth list` - if available.

Closes #2699

Signed-off-by: Maciej Strzelczyk <strzelczyk@google.com>
---


m-strzelczyk:
Example report from the GCP plugin run on GCE instance:
[sosreport-instance-2-2-2021-09-29-jeonzle.zip](https://github.com/sosreport/sos/files/7252975/sosreport-instance-2-2-2021-09-29-jeonzle.zip).

sos/report/plugins/gcp.py
```diff
@@ -0,0 +1,107 @@
...
+                                tags=['gcp'])
+
+        # Add journal entries
+        self.add_journal(units="google*", tags=['gcp'])
+
+    def get_metadata(self) -> dict:
+        try:
+            response = self._query_address(self.METADATA_ROOT)
+            if response.code != 200:
+                raise RuntimeError(
```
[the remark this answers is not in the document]
m-strzelczyk: You're right, it'll be better not to crash the whole plugin because of an error in metadata collection. Fix pushed.

sos/report/plugins/gcp.py
```diff
@@ -0,0 +1,111 @@
...
+    def _query_address(url: str) -> HTTPResponse:
+        """
+        Query the given url address with headers required by Google Metadata
+        Server.
+        """
+        req = request.Request(url, headers={'Metadata-Flavor': 'Google'})
+        return request.urlopen(req)
+
+    @classmethod
+    def _get_metadata(cls, url: str = '') -> Union[dict, str]:
```
pmoravec, line 90: Putting my paranoic hat on: we call some URI in a directory-recursive way. What if - due to a proxy or network or DNS issue - each query takes many seconds, causing the whole `self._get_metadata(self.METADATA_ROOT)` takes minutes?

(but I have no idea how to easily protect that behaviour to happen, so I think we can live with this tiny risk)

pmoravec (approved):
ACK otherwise (not sure if we can easily resolve the "what if the traversal is slow").

sos/report/plugins/gcp.py
m-strzelczyk: The thing about querying Metadata Server on GCP is that the requests **never** leave the physical machine. The metadata server is just a simple process running next to the VM on the physical host. It'd be an extremely weird event for those local requests to be slow.

DNS is also not an issue, since `metadata.google.internal` name gets resolved locally. 

Because the metadata server is so close to the VM, it can be used to acquire authentication token for GCP Service Accounts (that's why I redact them in line 91). This method will end fast, either with an error or success.

TurboTurtle:
Looking at the provided example report (thanks!) I see there are SSH keys present in the metadata. We should probably be removing those, right?  If not in the collection method directly, then via `postproc()`.

Within `metadata.json` are there any other sensitive fields we should be obfuscating? The hostnames and ip addresses can be handled via `sos clean`, but should we be removing items like service accounts or the related email addresses?

m-strzelczyk:
@TurboTurtle thanks for reviewing :)

The SSH keys in the report are all public keys only. I see no harm in keeping them, but if that's the best practice, I guess we can cut them out.

Names of service accounts (those are always in the form of email address) are rather important for Google Cloud Support, they should IMO stay. 

The only data that is really sensitive can be found in the `instance/service-accounts/*/token` fields and those are instantly redacted.

I will review GCP policies and GCP Support policies and check what is considered PII and I'll get back to this. Perhaps you are right and there are more things to be redacted.

TurboTurtle:
> @TurboTurtle thanks for reviewing :)
> 
> The SSH keys in the report are all public keys only. I see no harm in keeping them, but if that's the best practice, I guess we can cut them out.
> 

Yeah, I figured (well, hoped anyways :wink:) they were public keys, but wanted to double check that. With that being the case, I don't have an objection to keeping them.

> Names of service accounts (those are always in the form of email address) are rather important for Google Cloud Support, they should IMO stay.
> 

Ack.

> The only data that is really sensitive can be found in the `instance/service-accounts/*/token` fields and those are instantly redacted.
> 

Ack.

> I will review GCP policies and GCP Support policies and check what is considered PII and I'll get back to this. Perhaps you are right and there are more things to be redacted.

Sounds good. This is the only concern I had/have with the PR, everything else looks good (and yay for the first commit to sos that includes type hints :smiley: ) to me.

Incidentally, does Google Support make regular use of sos? We've had COS support in the project for some time, but it'd be good to get some feedback as to if we can improve the adoption of sos there. One thing that comes to mind is we recently abstracted the transport for `sos collect`, and I'm curious if there'd be a benefit to adding a `gcloud compute ssh` wrapper to allow multi-node report collection that way?

pmoravec:
A side note: I really appreciate the code quality and clarity and very prompt reaction to our feedback (during my review, I wrote one comment and before writing the other in 15mins(?), I had to refresh the PR that just had been modified based on my feedback - what a turnaround :) ). Obviously such prompt feedback is not required from us, just I am impressed by (not only) it.

m-strzelczyk:
Google support uses SOS with increasing frequency I'd say. For a long time, it was only part of debugging GKE, since the COS system support was available. Some time ago, this was added to gcloud: [gcloud alpha compute diagnose sosreport](https://cloud.google.com/sdk/gcloud/reference/alpha/compute/diagnose/sosreport). Seeing this was one of the reasons I decided to get engaged with the SOS project.

Thank you for the kind words about my speed :)

I have read a bit more about PII in GCP and it seems things like project number, project name etc. are treated as PII. I will proceed to add proper support to `sos clean` so those values can be removed if customer wishes.

m-strzelczyk:
I have prepared new version. The main changes are:

1. The Metadata collection now is just a one call to the Metadata Server. Should be faster and less error prone.
2. Plugin will by default remove project name and project number from the `metadata.json` output file. This can be stopped by using the `keep-pii` option. I added this is an option, because these info might be important for Google Cloud Support for faster issue resolution - they know project name and number when handling a support case anyway.

I'm attaching two output examples, one with and one without the PII flag:
[sosreport-instance-2-keep-pii-2021-10-05-uloogsa.zip](https://github.com/sosreport/sos/files/7286789/sosreport-instance-2-keep-pii-2021-10-05-uloogsa.zip) and 
[sosreport-instance-2-no-pii-2021-10-05-rhswnbn.zip](https://github.com/sosreport/sos/files/7286790/sosreport-instance-2-no-pii-2021-10-05-rhswnbn.zip)

m-strzelczyk:
As mentioned in #2237 I have added a little piece of code that makes sure my `gcloud auth list` command gets executed on new Ubuntus that have `gcloud` installed via `snap` as well.

sos/report/plugins/gcp.py
```diff
@@ -0,0 +1,155 @@
...
+         * Any google services output from journal
+        """
+
+        # Capture gcloud auth list
+        if self._has_gcloud_snap():
+            self.add_cmd_output("gcloud auth list", tags=['gcp'])
+        else:
+            predicate = SoSPredicate(self, packages=['google-cloud-sdk'])
+            self.add_cmd_output("gcloud auth list", pred=predicate,
+                                tags=['gcp'])
```
TurboTurtle, line 70: `add_cmd_output()` automatically gracefully handles when a called command does not exist. So this can be a "plain" `self.add_cmd_output("gcloud auth list", tags=['gcp'])` and it will function properly regardless of where (or if) `gcloud` is installed from.

m-strzelczyk: OK, I'll just switch to the "plain" version you described. Makes life easier, thanks :)

PR #2711, opened 2021-10-04
[report] Overwrite pred=None before refering predicate attributes

pmoravec:
During a dry run, add_journal method sets pred=None whilst log_skipped_cmd
refers to predicate attributes. In that case, replace None predicate
by a default / empty predicate.

Resolves: #2711

---


pmoravec:
Reproducer:

    pyhon3 bin/sos report -o foreman --batch -build --dry-run

generates traceback:

      File "/root/sos-main/sos/report/__init__.py", line 974, in setup
        plug.setup()
      File "/root/sos-main/sos/report/plugins/foreman.py", line 301, in setup
        super(RedHatForeman, self).setup()
      File "/root/sos-main/sos/report/plugins/foreman.py", line 135, in setup
        self.add_journal(units="dynflow-sidekiq@*")
      File "/root/sos-main/sos/report/plugins/__init__.py", line 2581, in add_journal
        priority=priority)
      File "/root/sos-main/sos/report/plugins/__init__.py", line 1714, in _add_cmd_output
        self.log_skipped_cmd(pred, soscmd.cmd, kmods=bool(pred.kmods),
    AttributeError: 'NoneType' object has no attribute 'kmods'

and the explicit `pred=None` comes from https://github.com/sosreport/sos/blob/main/sos/report/plugins/__init__.py#L2580 .


There are multiple ways how to fix this and I am uncertain which is the best one:

- the proposed one
- set the same just after `pred = kwargs.pop(.....)`
- replace `kmods=bool(pred.kmods)` by `kmods=bool(pred.kmods) if pred is not None else False` and similar for `services`

Either solution has its pros and cons wrt. code readiness or efficiency. Let me know if you prefer another fix.

TurboTurtle:
We should already have an "empty" predicate due to https://github.com/sosreport/sos/blob/34d5d978c5d500129ba1a95dc3eab7727331c1f3/sos/report/plugins/__init__.py#L1695

But, that is checking for the `pred` kwarg, not that the kwarg actually has value. I think we should instead refine the check in line 1695, that way we're only imposing an empty predicate once.

pmoravec:
Updated.

A side note: when reviewing validity of my patch, I come up to a `py2` relic we might remove? https://github.com/sosreport/sos/blob/main/sos/report/plugins/__init__.py#L352-L355

PR #2712, opened 2021-10-04
[foreman] Collect puma status and stats

pmoravec:
Collect foreman-puma-status and 'pumactl [gc-|]stats', optionally using
SCL (if detected).

Resolves: #2712

---

### DIFF
--- a/sos/report/plugins/foreman.py
+++ b/sos/report/plugins/foreman.py
@@ -13,6 +13,7 @@ from sos.report.plugins import (Plugin, RedHatPlugin, DebianPlugin,
                                 UbuntuPlugin, PluginOpt)
 from pipes import quote
 from re import match
+from sos.utilities import is_executable
 
 
 class Foreman(Plugin):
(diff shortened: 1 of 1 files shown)


sos/report/plugins/foreman.py
```diff
@@ -297,6 +312,10 @@ def setup(self):
         self.add_file_tags({
             '/usr/share/foreman/.ssh/ssh_config': 'ssh_foreman_config',
         })
+        # if we are on RHEL7 with scl, wrap some Puma commands by
+        # scl enable tfm 'command'
+        if self.policy.dist_version() == 7 and is_executable('scl'):
+            self.pumactl = "scl enable tfm '%s'" % self.pumactl
```
pmoravec, line 318: We can utilise `SCLPlugin.convert_cmd_scl` method but:
- we dont need to modify `PATH` here
- it might be a bit overkill to use the class at all, if majority of (future) execution will not use SCL at all

TurboTurtle (approved):

sos/report/plugins/foreman.py
ekohl: I see you already use `foreman-puma-status` (good!). If it's easier, we can probably introduce wrappers for other Puma things. I suspect other users may find them useful too.

PR #2713, opened 2021-10-04
[transports] Add a transport for `oc`

TurboTurtle:
This patchset adds the ability to use the `oc` binary as a transport for `sos collect` so that OCP nodes can be collected from without directly SSH'ing to the nodes.

Patch 1 adds the `--transport` option, and basic logic for auto-setting the transport in use (will default to SSH ControlPersist) if not set by the user or the cluster. Cluster profiles may now directly set the transport type.

Patch 2 adds the `oc` transport which will deploy a pod on each node using `oc create`, and will then perform our host checks and sos execution via `oc exec` wrappers.

Patch 3 moves the automatic nesting of a remote command in a bash shell from `SoSNode` to `RemoteTransport`.

Patch 4 fixes a minor typo and logic break discovered during testing on an OCP cluster.

---


TurboTurtle:
I have tested this in a small OCP cluster, but admittedly the platform it is running on is very under powered and there may be some bits in this initial PR that are artifacts of having to deal with such an environment - i.e. a "real" environment will likely not be affected in the same way.

I am hoping that others may be able to provide more accurate testing and provide feedback for this.

sos/collector/transports/__init__.py
```diff
@@ -189,7 +196,8 @@ def _format_cmd_for_exec(self, cmd):
                         transport
         :rtype:         ``str``
         """
-        cmd = "%s %s" % (self.remote_exec, quote(cmd))
+        cmd = "%s %s" % (self.remote_exec,
+                         quote(cmd) if self.quote_cmds else cmd)
```
pmoravec, line 200: `self.quote_cmds` is a dead variable, always set to `True`?

man/en/sos-collect.1
```diff
@@ -350,6 +351,20 @@ Note that sosreports are collected in parallel, so you can approximate the total
...
+\fBsos collect\fR uses locally installed binaries to connect to and interact with remote
+nodes, instead of directly establishing those connections. By default, OpenSSH's ControlPersist
+feature is preferred, however certain cluster types may have preferences of their own for how
+remote sessions should be established.
+
+The types of transports supported are currently as follows:
+
+    \fBauto\fR                  Allow the cluster type to determine the transport used
+    \fBcontrol_persist\fR       Use OpenSSH's ControlPersist feature. This is the default behavior
+
```
pmoravec, line 367: The "Add 'oc' as a transport option" patch should add 'oc' option here.

man/en/sos-collect.1
```diff
@@ -350,6 +351,20 @@ Note that sosreports are collected in parallel, so you can approximate the total
...
+\fBsos collect\fR uses locally installed binaries to connect to and interact with remote
+nodes, instead of directly establishing those connections. By default, OpenSSH's ControlPersist
+feature is preferred, however certain cluster types may have preferences of their own for how
+remote sessions should be established.
+
+The types of transports supported are currently as follows:
+
+    \fBauto\fR                  Allow the cluster type to determine the transport used
+    \fBcontrol_persist\fR       Use OpenSSH's ControlPersist feature. This is the default behavior
```
pmoravec, line 366: Subjective opinion: From the type name `control_persist`, I would not deduce we use `ssh` for the transport. Could "ssh" be part of the name? (But `ssh_control_persist` is too long, `ssh` might lack the ControlPersist feature info (is the info in type name required?)]

sos/collector/transports/oc.py
```diff
@@ -0,0 +1,206 @@
...
+                        "hostPath": {
+                            "path": "/etc/machine-id",
+                            "type": "File"
+                        }
+                    }
+                ],
+                "containers": [
+                    {
+                        "name": "sos-collector-tmp",
+                        "image": "registry.redhat.io/rhel8/support-tools",
```
pmoravec, line 89: Just a question from an OCP noob: is it OK to put distro-specific image name to distro-generic code / response?

pmoravec:
Do I get it right that the 'oc' transport means a new pod / container is downloaded from `registry.redhat.io` end `sos` is executed in it? If so, how do we handle situation when OCP cluster can't download the pod due to whatever reason? (via the `"pod/%s created" % self.pod_name not in out['output'])` test in https://github.com/sosreport/sos/pull/2713/commits/b9ee96cc516721735d3d6c631f38192ec0665b0a#diff-fa63f5d69985ee0b15ccb811b20781fa9541b6b88556858d2c08762b37d5beb7R150-R151 ?)

sos/collector/transports/oc.py
```diff
@@ -0,0 +1,206 @@
...
+        return super(OCTransport, self)._format_cmd_for_exec(cmd)
+
+    def run_command(self, cmd, timeout=180, need_root=False, env=None,
+                    get_pty=False):
+        # debug pod setup is slow, extend all timeouts to account for this
+        if timeout:
+            timeout += 10
+
+        # since we always execute within a bash shell, force disable get_pty
+        # to avoid double-quoting
```
pmoravec, line 190: Ah, now I understand the `self.quote_cmd` relic :)

sos/collector/transports/oc.py
```diff
@@ -0,0 +1,206 @@
...
+        get_pty = False
+        return super(OCTransport, self).run_command(cmd, timeout, need_root,
+                                                    env, get_pty)
```
pmoravec, line 193: I understand we need `get_pty` as method argument since the method can be called somewhere with the argument that we do ignore here. Isn't it easier to ignore `get_pty` variable, then, and cal `run_command(cmd, timeout, need_root, env, False)` directly?

sos/collector/transports/oc.py
```diff
@@ -0,0 +1,206 @@
...
+
+    def _disconnect(self):
+        os.unlink(self.pod_tmp_conf)
+        return "deleted" in shell_out("oc delete pod %s" % self.pod_name)
```
pmoravec, line 197: A question from an OCP / pod noob: is it the right ordering (or safe) that we delete the pod after we remove its config?

sos/collector/transports/__init__.py
TurboTurtle: This is a bit of a left-over from an earlier draft - initially I was using `oc debug` which required some "interesting" quoting work. At the moment yes it's a dead variable but I wonder if it would be useful for future transports (of course, in that case we can always add it back at that time).

I'll pull it out in the next branch update.

man/en/sos-collect.1
TurboTurtle: Ack.

man/en/sos-collect.1
TurboTurtle: There's the possibility that in the future we could add a transport that uses, for example, the libssh python bindings. In which case we need an obvious way to differentiate between the two, but I couldn't think of anything besides this or as you mentioned `ssh_control_persist` which is a bit heavy.

sos/collector/transports/oc.py
TurboTurtle: This transport is for OCP clusters only - which only ships on RHEL. This wouldn't be enabled for any other kubernetes-based platform, so I think we're good here.

sos/collector/transports/oc.py
TurboTurtle: Yes - the pod is already created and exists entirely within the OPC API. This is just a local file reference, so we can delete it without concern or impact on the API.

TurboTurtle:
> Do I get it right that the 'oc' transport means a new pod / container is downloaded from `registry.redhat.io` end `sos` is executed in it?

Correct.

> If so, how do we handle situation when OCP cluster can't download the pod due to whatever reason? (via the `"pod/%s created" % self.pod_name not in out['output'])` test in [b9ee96c#diff-fa63f5d69985ee0b15ccb811b20781fa9541b6b88556858d2c08762b37d5beb7R150-R151](https://github.com/sosreport/sos/commit/b9ee96cc516721735d3d6c631f38192ec0665b0a#diff-fa63f5d69985ee0b15ccb811b20781fa9541b6b88556858d2c08762b37d5beb7R150-R151) ?)

If we fail to deploy the pod, we report the error and drop the node from the collection list - basically the same as today if we can't open an SSH session to a particular node.

If the `oc` binary is not available, configured properly, or otherwise has issues, the `ocp` cluster profile will attempt to fallback to `control_persist` as the transport. Users can also manually override the `--transport` if needed.

man/en/sos-collect.1
pmoravec: I understand. `ssh` is too generic, `ssh_control_persist` too long. So let use the currently proposed term.

sos/collector/transports/__init__.py
```diff
@@ -212,6 +212,9 @@ def _run_command_with_pexpect(self, cmd, timeout, need_root, env):
         :type env:      ``dict``
         """
         cmd = self._format_cmd_for_exec(cmd)
+        print('!!')
+        print(cmd)
+        print('!!!')
```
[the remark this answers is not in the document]
TurboTurtle: Total debug leftover, my fault. ;-)

PR #2724, opened 2021-10-14
[firewall_tables] call iptables -t <table> based on nft list

pmoravec:
---


pmoravec:
Aside to the failed tests (that I am unable to identify/reproduce, so far), there might be a problem on systems without `nft` (could they exist?) where we should not guard `iptables` commands by non-existing `nft` command output - but we do.

I must extend the PR by "if the nft command returned error, ignore the checks from it".

christianhorn:
Hm.. I assumed the copr repo should be populated with the build which succeeded now, but on rhel8 and Fedora34:
[root@guybrush ~]# dnf copr enable packit/sosreport-sos-2724
[..]
Do you really want to enable copr.fedorainfracloud.org/packit/sosreport-sos-2724? [y/N]: y
Error: This repository does not have any builds yet so you cannot enable it now.

christianhorn:
packit-as-service above commented "rhel8 and Fedora", so I tried rhel8 and fedora34. https://copr.fedorainfracloud.org/coprs/packit/sosreport-sos-2724/ seems to say that only Fed35 was build, and indeed on that I can access the build, will look at that build.

christianhorn:
For rhel8:

- build of sos-4.2-1.20211015203833789910.pr2724.55.gb3b8e4aa.fc34.src.rpm, these 2 tables still get created: "ip filter", "ip6 filter"
- build of sos-4.1 leaves 3 nft tables: "ip6 mangle", "ip6 filter", "ip filter"
- sos-4.0-11 (native from repos) leaves 4 nft tables: ip6 nat/mangle/filter, ip filter

> Could you give me access to such system to let me try?
Let's first try without.. I mostly play with quickly provisioned KVM guests on my laptop at home.

> Or let check from sos_logs/sos.log what commands (from plugin:networking, I guess?)
> would create those tables, when you run the cmds manually?

I bisected, plugin:firewall_tables is executing
  iptables -vnxL
  ip6tables -vnxL

Running a build of sos-4.2-1.20211015203833789910.pr2724.55.gb3b8e4aa,
"sos report -o networking" is not creating any nft tables.

> [pmoravec]
> You mean sos-4.1 already behaves properly? And sos-4.2 as well?
> (hence this PR is redundant)?

Ah sorry, I was solely judging from Fed34, Fed35 and EL9 with sos 4.1 or 4.2 not having the issue.  I see now that a build
of sos-4.1 on rhel8 also has the issue. So we need patches to solve the issue on rhel8, even if we rebase to newer sos.

PR #2726, opened 2021-10-18
[ceph] Add support for containerized Ceph setup

pmoravec:
RH Ceph 4 can either be installed as RPM or as containers.
The changes permits to collect the ceph config
on both kind of setup.

Resolves: #2680
Related: #2646

---

### DIFF
--- a/sos/plugins/ceph.py
+++ b/sos/plugins/ceph.py
@@ -8,6 +8,8 @@
 
 from sos.plugins import Plugin, RedHatPlugin, UbuntuPlugin
 from socket import gethostname
+from sos.utilities import (sos_get_command_output, is_executable)
+import re
 
 
 class Ceph(Plugin, RedHatPlugin, UbuntuPlugin):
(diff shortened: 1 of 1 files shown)


sos/plugins/ceph.py
```diff
@@ -132,4 +165,29 @@ def setup(self):
...
+        if self.ceph_containers:
+            # Avoid retrieving multiple times the same data
+            got_ceph_cmds = False
+            for container in self.ceph_containers:
+                if re.match("ceph-(mon|rgw|osd)", container[1]) and \
+                        not got_ceph_cmds:
+                    self.add_cmd_output([
+                        self.fmt_container_cmd(container[1], "ceph %s" % s)
+                        for s in ceph_cmds
+                    ])
```
TurboTurtle, line 178: `Plugin.fmt_container_cmd()` does not exist in `legacy-3.9`. It was added in as part of the 4.0 update.

pmoravec: Ah, good catch, the original PR is a trap aside a trap :)

Issue #1506, opened 2018-12-10
Large command outputs should not be completely stored in memory


pmoravec:
Despite memory utilization improvements done in past year, there is still one(?) principal bottleneck causing whole outputs of commands are kept in memory in one chunk:

https://github.com/sosreport/sos/blob/master/sos/utilities.py#L246

This implies that e.g. on OpenStack, due to RHOSP preset and --all-logs option enabled by default, journalctl verbosed output is collected (usually the biggest output I have seen in a sosreport). This requires few-to-many GB of memory, usually. And it can easily cause OOM killer stopping sosreport execution.

Sosreport should handle (big or all?) command outputs either:

- in memory by smaller batches (this can be problematic for regexp substitution when the regexp to be matched is spread over few batches processed independently)
- or write them to the disk (and apply regexp sub on the file)

Several options are here:
- keep in memory just small batches OR write everything to the disk
- apply this change just to outputs bigger than some threshold OR to all outputs (as there is a performance penalty behind this change)

What also plays a role: we need to keep the output in memory just for the sake of possible regexp substitution. In other cases, we can simply write the output to the disk without storing it completely in memory.

(my deduction/opinion): This fact suggests we shall write everything directly to a file (so AsyncReader should do so, all chunks of self.dequeue in a row) - ideally to the final destination within the sosreport build tree - and then regexp sub to be applied on the file.

bmr-cymru:
Yeah. That kinda defeats the entire object. I think the answer here is to refactor that as a method as a generator, *but* I don't think that really solves the problem: it's the straw that breaks the camel's back. I didn't check in detail but my guess what's happening here is `sizelimit` is disabled, the entire command output is pulled into memory, and then we do the join and format another giant string - it's this 2nd allocation that takes us down the OOM route.

If that is the case then we need to find a way to stream output to the file from the pipeline. I'll try to look properly later on once I've got outstanding PRs done.

bmr-cymru:
> in memory by smaller batches

Long term I think that's the only answer. Do we have many multi-line regexes (or regexes that address single lines that may be excessively large)? I'm not sure that's the case.

pmoravec:
> Do we have many multi-line regexes (or regexes that address single lines that may be excessively large)? I'm not sure that's the case.

IMHO "just" scrubbing of SSL certs belong to this.

pmoravec:
Just as a cross-reference: this issue is planned to be resolved in sos 4.0, see #1959 and "IO improvements".

PR #2729, opened 2021-10-19
[Plugin, utilities] Allow writing command output directly to disk

TurboTurtle:
This patchset aims to address a long standing ask for sos - that command output is not _required_ to be stored entirely in memory while we are collecting it.

With these commits, when `--all-logs` is used, or if `sizelimit` is zero for specific commands, `sos_get_command_output()` will write the output directly to a temp file rather than storing the output in memory and returning that.

This is so far only exposed via `add_cmd_output()` by a new `to_file` parameter, but this is automatically handled in the cases above for both `add_cmd_output()` and `add_journal()`. This functionality is not exposed via `exec_cmd()` or `collect_cmd_output()`.

After the command output temp file has been copied to the archive, the temp file is removed via the second commit extending `TempFileUtil`.

Resolves: #1506

---

### DIFF
--- a/sos/utilities.py
+++ b/sos/utilities.py
@@ -110,7 +110,8 @@ def is_executable(command, sysroot=None):
 
 def sos_get_command_output(command, timeout=TIMEOUT_DEFAULT, stderr=False,
                            chroot=None, chdir=None, env=None, foreground=False,
-                           binary=False, sizelimit=None, poller=None):
+                           binary=False, sizelimit=None, poller=None,
+                           to_file=False):
     """Execute a command and return a dictionary of status and output,
     optionally changing root or current working directory before
     executing command.
@@ -123,6 +124,12 @@ def sos_get_command_output(command, timeout=TIMEOUT_DEFAULT, stderr=False,
             os.chroot(chroot)
         if (chdir):
             os.chdir(chdir)
+
+    def _check_poller(proc):
+        if poller():
+            proc.terminate()
+            raise SoSTimeoutError
+        time.sleep(0.01)
 
     cmd_env = os.environ.copy()
     # ensure consistent locale for collected command output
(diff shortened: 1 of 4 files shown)


TurboTurtle:
@pmoravec cycling back around on this.

Do we want to pursue a call to `Archive._check_path()` outside of `add_file()`, or are we ok with the copy-then-delete approach? I'm personally still torn on this.

pmoravec:
First, I did some performance measurements and I really appreciate the results: 1GB big command output collected in 10 seconds instead of 30, and with basically no extra memory needed (compared to extra 2GB as now).

I agree size limiting is a blocker to this approach, so I take back it.

About copy&delete vs. populate-in-final-destination: isn't it "just" a matter of careful reordering (and putting under few "`if to_file`" tests) `_collect_cmd_output` ? Now, we generate `outfn` after the command is executed and its output is kept in tempfile or in memory. But the `outfn` relies on command execution outcome only if:

- command not found or runnable or chroot-ed: if we create the leading path in advance, we might end up with empty subdirs - imho no big deal
- result is truncated: this can't happen (as applied to sizelimit=0 only)

Or did I miss something?

TurboTurtle:
One concern I had was that we could end up creating empty _files_ for the collections, which may confuse end users into thinking the command was successfully run but produced no output. However, we can always just remove that file if we have no output to write.

I'll work on something for this this week.

TurboTurtle:
Sorry for the long silence on this, I wasn't able to dig back into this until now with the other work going on in the project.

Anyways, I've changed this up to write directly to the archive for when `sizelimit=0` or when `--all-logs` is used. From my tests, these scenarios are greatly improved in terms of performance and resource consumption - and "normal" runs are left unaffected.

I dropped the changes to `TempFileUtil` as they were no longer necessary. I think I may also see a way forward for dropping our use of `timeout` for all commands, but that requires more investigation and is out of scope for this particular PR.

sos/utilities.py
```diff
@@ -124,6 +125,12 @@ def _child_prep_fn():
         if (chdir):
             os.chdir(chdir)
 
+    def _check_poller(proc):
+        if poller():
+            proc.terminate()
+            raise SoSTimeoutError
+        time.sleep(0.1)
```
pmoravec, line 132: We had `time.sleep(0.01)`, some reason to increase this? (but I think even the increase does not matter).

TurboTurtle: I ran a couple tests with this while re-writing this section. The TL;DR of it is using `0.1` gives us lower CPU consumption, however at the cost of runtime. This makes sense as we're doing 10x less calls to `poller()`, however we stack up the extra fractions of a second across all the hundreds/thousands of commands we run.

Initially, my opinion is/was that this is preferable - however, I'm now thinking of this in the context of nodes with extremely high numbers of namespaces - which would definitely cause the runtime to increase dramatically more than "normal" systems.

Reverting this back to 0.01 for now, and I'll make a note to do some more investigation on this.

pmoravec (approved):
Passed all my smoke tests. Performance gain isnt such such great as previously (but I tested it on smaller setup), but still evident improvement.

Great!

PR #2731, opened 2021-10-21
[foreman] Collect 'scl enable tfm gem list' again

pmoravec:
---

### DIFF
--- a/sos/report/plugins/foreman.py
+++ b/sos/report/plugins/foreman.py
@@ -9,8 +9,8 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.report.plugins import (Plugin, RedHatPlugin, DebianPlugin,
-                                UbuntuPlugin, PluginOpt)
+from sos.report.plugins import (Plugin, RedHatPlugin, SCLPlugin,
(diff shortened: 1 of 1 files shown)


sos/report/plugins/__init__.py
```diff
@@ -3004,24 +3004,9 @@ def _get_scls(self):
...
-            self._log_error("Failed to find prefix for SCL %s using %s: %s"
-                            % (scl, prefix, e))
-
-        # expand PATH by equivalent prefixes under the SCL tree
-        path = os.environ["PATH"]
-        for p in path.split(':'):
-            path = '%s/%s%s:%s' % (prefix, scl, p, path)
-
-        scl_cmd = "scl enable %s \"PATH=%s %s\"" % (scl, path, cmd)
+        scl_cmd = "scl enable %s \"%s\"" % (scl, cmd)
```
TurboTurtle, line 3009: Is `scl` doing its own PATH handling something it has always done, or is this a somewhat new development? We've had SCL collections in the project for years (even though the number of those collections is fairly low), without a report of an issue so this makes me wonder about other `scl` collections outside of the one that triggered this specific change.

TurboTurtle:
Ack, works for me.

PR #2733, opened 2021-10-23
[block] Fix error while gathering data on loop devices

jcastill:
When we run parted command on detached loop devices,
some error messages are printed in the kernel ring buffer,
like:

[ +10.838245] blk_update_request: I/O error, dev loop0,
	sector 0 op 0x1: (WRITE) flags 0x800
	phys_seg 0 prio class 0
Warning: Error fsyncing/closing /dev/loop0: Input/output error

These error messages are printed as well when parted gets run
while collecting sosreport.

This patch solves the problem by checking the list
of attached loop devices and then running commands over them,
after blacklisting loop devices from a previous, more
general call to collect data.

Fixes: RHBZ#2010735
Fixes: #2733 

---

### DIFF
--- a/sos/report/plugins/block.py
+++ b/sos/report/plugins/block.py
@@ -56,7 +56,14 @@ class Block(Plugin, IndependentPlugin):
             "udevadm info %(dev)s",
             "udevadm info -a %(dev)s"
         ]
-        self.add_blockdev_cmd(cmds, blacklist='ram.*')
+        self.add_blockdev_cmd(cmds, blacklist=['ram.*', 'loop.*'])
+        loop_devices = self.exec_cmd('losetup --all --noheadings')
+        if loop_devices['status'] == 0:
+            loop_cmds = ["fdisk -l", "udevadm info", "udevadm info -a"]
+            for loop_dev in loop_devices['output'].splitlines():
+                loop_device = loop_dev.split()[0].replace(':', '')
+                self.add_cmd_output(["%s %s" % (cmd, loop_device)
+                                     for cmd in loop_cmds])
 
         lsblk = self.collect_cmd_output("lsblk -f -a -l")
         # for LUKS devices, collect cryptsetup luksDump


TurboTurtle:
@jcastill If this needs to land by sos-4.3, please update the PR as 4.3 is closing Feb 11th.

jcastill:
I'll finish this soon.

jcastill:
I'm fixing my branch to sync with main and will be pushing the changes soon

sos/report/plugins/block.py
```diff
@@ -56,7 +56,14 @@ def setup(self):
             "udevadm info %(dev)s",
             "udevadm info -a %(dev)s"
         ]
-        self.add_blockdev_cmd(cmds, blacklist='ram.*')
+        self.add_blockdev_cmd(cmds, blacklist=['ram.*', 'loop.*'])
+        loop_devices = self.exec_cmd('losetup --all --noheadings')
+        if loop_devices['status'] == 0:
+            loop_cmds = ["fdisk -l", "udevadm info", "udevadm info -a"]
```
pmoravec, line 62: Why we cant use `cmds` list and set here `loop_cmds = cmds[1:]` (and add a warning comment to `cmds` declaration that `parted` command must be the first)?

This will prevent adding a command to two lists in future.

pmoravec:
When running against original reproducer, I still see *one* `blk_update_request` error log:

Originally, `dmesg` contain:

	[Feb12 09:42] loop: module loaded
	[  +5.020437] blk_update_request: I/O error, dev loop0, sector 0 op 0x1:(WRITE) flags 0x800 phys_seg 0 prio class 0
	[  +0.529661] blk_update_request: I/O error, dev loop0, sector 0 op 0x1:(WRITE) flags 0x800 phys_seg 0 prio class 0

With the patch, it contains:

	[Feb12 09:43] loop: module loaded
	[  +5.469202] blk_update_request: I/O error, dev loop0, sector 0 op 0x1:(WRITE) flags 0x800 phys_seg 0 prio class 0

I am checking what *else* command from `block` plugin is responsible for that.

jcastill:
These are the commands that "touch" loop devices:

blockdev_--report
fdisk_-l_.dev.loop1
losetup_-a
lsblk
lsblk_-D
lsblk_-f_-a_-l
lsblk_-O_-P
lsblk_-t
ls_-lanR_.dev
ls_-lanR_.sys.block
udevadm_info_-a_.dev.loop1
udevadm_info_.dev.loop1


Running any of these by hand in my test server show nothing in dmesg. And the same running the PR code, nothing in dmesg. 
Can you try these commands by hand and see if you get anything?

pmoravec:
I apologize, the dmesg error was caused by the explicit call `parted -s /dev/loop0 unit s printM` from the reproducer :)

`sos` behaves well alone, with this PR.

pmoravec (approved):
ACK also for the current PR version.

sos/report/plugins/block.py
TurboTurtle: Why are we re-detecting `loopX` devices here? Is there something wrong with the list determined and used by `add_blockdev_cmd()` normally?

I.E. why can't we simply do this?

```python
self.add_blockdev_cmd(cmds, blacklist=['ram.*', 'loop.*'])
self.add_blockdev_cmd(cmds[1:], whitelist='loop.*')
```

pmoravec: +1 for this simplistic elegant proposal.

jcastill: It didn't occur to me to do it this way, and is super elegant. I'll make this change and test it, then commit again.

jcastill: OK I remember why I decided to gather the devices again. The problem is that when you detach an image from a loop device, the actual loop device continues existing in /dev. An example:

```
[root@voyager ~]# ll /dev/loop*
brw-rw----. 1 root disk   7,   0 Feb 14 14:42 /dev/loop0
brw-rw----. 1 root disk 259,   0 Feb 14 14:42 /dev/loop0p1
brw-rw----. 1 root disk   7,   1 Feb 12 10:00 /dev/loop1
brw-rw----. 1 root disk   7,   2 Feb 14 14:42 /dev/loop2
crw-rw----. 1 root disk  10, 237 Oct 22 15:49 /dev/loop-control

[root@voyager ~]# losetup -d /dev/loop1

[root@voyager ~]# losetup --all --noheadings
/dev/loop2: [64513]:2152 (/root/loopbackfile2.img)
/dev/loop0: [64513]:84 (/root/loopbackfile.img)

[root@voyager ~]# ll /dev/loop*
brw-rw----. 1 root disk   7,   0 Feb 14 14:42 /dev/loop0
brw-rw----. 1 root disk 259,   0 Feb 14 14:42 /dev/loop0p1
brw-rw----. 1 root disk   7,   1 Feb 14 14:43 /dev/loop1
brw-rw----. 1 root disk   7,   2 Feb 14 14:42 /dev/loop2
crw-rw----. 1 root disk  10, 237 Oct 22 15:49 /dev/loop-control
```

Result when running with the proposed change:

```
[root@voyager block]# cat fdisk_-l_.dev.loop1
fdisk: cannot open /dev/loop1: Inappropriate ioctl for device
```

As you can see, using 'losetup --all --noheadings' is the only way to have a proper list of loop devices in use.

pmoravec: Hm.. should not we improve detection of (loop only?) devices in https://github.com/sosreport/sos/blob/ade857c82926bb7de2c45232f00a3f346db4e59a/sos/report/__init__.py#L490 , then? Can't this issue affect some other calls of `add_blockdev_cmd` against loop devices?

TurboTurtle: Agreed @pmoravec

@jcastill - let's move this detection change into `sos/report/__init__.py.get_block_devs()`, then we can use the simplified calls above in the plugin while having better `loopX` device detection across all plugins.

jcastill: @TurboTurtle @pmoravec ack, I'll start work on this right away, then I'll resubmit this PR when the changes on the other one are added.

sos/report/plugins/block.py
```diff
@@ -56,7 +56,14 @@ def setup(self):
             "udevadm info %(dev)s",
             "udevadm info -a %(dev)s"
         ]
-        self.add_blockdev_cmd(cmds, blacklist='ram.*')
+        self.add_blockdev_cmd(cmds, blacklist=['ram.*', 'loop.*'])
+        loop_devices = self.exec_cmd('losetup --all --noheadings')
```
vikas-goel, line 60: Is `--noheadings` a valid option in RHEL?

sos/report/plugins/block.py
vikas-goel: My setup has loop device count up to 10K. The plugin times out while collecting data. What will be impact of this change? Is there an option to skip loop devices?

jcastill: @vikas-goel I'm rewriting this PR at the moment to deal with loop devices outside of the block plugin, like we do with block devices etc. While the PR is almost ready, I'm doing some tests before sending it, so I'll test with 10K loop devices and see if my new approach times out as well.

sos/report/plugins/block.py
jcastill: @vikas-goel no, but this command is irrelevant now because I'm using a different approach to gather loop devices.

pmoravec:
Hi @TurboTurtle , could you please review this PR?

TurboTurtle:
> Hi @TurboTurtle , could you please review this PR?

I thought we were waiting on a rewrite from @jcastill

TurboTurtle:
I don't believe there's an urgency here. If we can get this merged by sos-4.4 (in August), then I think we're fine.

pmoravec:
Ping @jcastill , could you fix it within 1-2 weeks / until 4.4 release?

TurboTurtle:
I think this may now be superfluous due to #2980 and #2972

pmoravec:
Ah right, so worth closing this.

Issue #2735, opened 2021-10-27
Collect OSD specific information in the ceph_osd plugin


nkshirsagar:
https://github.com/sosreport/sos/issues/1945 split the ceph plugin into individual component plugins. One of the reasons this was done was so we could continue to add component specific data collection for each individual ceph service.

This issue has been opened to now enhance the ceph_osd plugin to collect OSD related detailed data. Some of the data  this plugin will collect is listed below.

ceph daemon osd.<id> ...

1) dump_blocked_ops
2) config diff
3) bluestore bluefs available
4) dump_ops_in_flight
5) dump_historic_ops_by_duration (sorted by duration) 
6) dump_historic_slow_ops
7) dump_op_pq_state
8) dump_mempools
9) dump_osd_network
10) perf dump
11) dump_reservations
12) dump_watchers
13) status
14) config show
15) config diff
16) bluestore bluefs stats
17) log dump
18) ops
19) dump_blacklist
20) objecter_requests

Suggestions for additions/removals are welcome!

PR #2736, opened 2021-10-28
    [ceph] Add more data collection to ceph_osd

nkshirsagar:
Enhance the ceph_osd plugin to collect more data
    specific to OSD nodes.

    Closes: #2735

---

### DIFF
--- a/sos/report/plugins/ceph_osd.py
+++ b/sos/report/plugins/ceph_osd.py
@@ -39,8 +39,58 @@ class CephOSD(Plugin, RedHatPlugin, UbuntuPlugin):
 
         self.add_cmd_output([
             "ceph-disk list",
-            "ceph-volume lvm list",
+            "ceph-volume lvm list"
         ])
+
+        ceph_cmds = [
+           "bluestore bluefs available",
+           "config diff",
+           "config show",
+           "dump_blacklist",
+           "dump_blocked_ops",
+           "dump_historic_ops_by_duration",
+           "dump_historic_slow_ops",
+           "dump_mempools",
+           "dump_ops_in_flight",
+           "dump_op_pq_state",
+           "dump_osd_network",
+           "dump_reservations",
+           "dump_watchers",
+           "log dump",
+           "perf dump",
+           "perf histogram dump",
+           "objecter_requests",
+           "ops",
+           "status",
+           "version",
+        ]
+
+        osd_ids = []
+        # Get the ceph user processes
+        out = self.exec_cmd('ps -u ceph -o args')
+
+        if out['status'] == 0:
+            # Extract the OSD ids from valid output lines
+            for procs in out['output'].splitlines():
+                proc = procs.split()
+                if len(proc) < 6:
+                    continue
+                if proc[4] == '--id' and proc[5].isdigit():
+                    osd_ids.append("osd.%s" % proc[5])
+
+        containers_list = self.get_all_containers_by_regex("ceph-osd*")
+
+        if containers_list:
+            self.add_cmd_output([
+                self.fmt_container_cmd(
+                    containers_list[0][1], "ceph daemon %s %s"
+                    % (osid, cmd)) for osid in osd_ids for cmd in ceph_cmds
+            ])
+        else:
+            self.add_cmd_output([
+                "ceph daemon %s %s" % (
+                    osid, cmd) for osid in osd_ids for cmd in ceph_cmds
+            ])
 
         self.add_forbidden_path([
             "/etc/ceph/*keyring*",


sos/report/plugins/ceph_osd.py
```diff
@@ -39,9 +40,47 @@ def setup(self):
 
         self.add_cmd_output([
             "ceph-disk list",
-            "ceph-volume lvm list",
         ])
 
+        ceph_cmds = [
```
hillpd, line 45: This list looks good. Please add the following two commands:
"version",
"perf histogram dump",

For details, see: https://docs.ceph.com/en/latest/dev/perf_histograms/

sos/report/plugins/ceph_osd.py
```diff
@@ -39,9 +40,47 @@ def setup(self):
...
+           "dump_watchers",
+           "log dump",
+           "perf dump",
+           "objecter_requests",
+           "ops",
+           "status"
+        ]
+
+        osd_ids = []
+        output = self.collect_cmd_output('ceph-volume lvm list --format json')
```
hillpd, line 67: There have been a number of performance issues with ceph-volume in the past. I do not recommend using this command to build a list of osd ids. 

It is a bit uglier, but you can build a list of ids by parsing `mount` output. As an example, this is the command systemd uses to generate a list of the ceph-osd services that need to be started:
`sudo mount|grep /var/lib/ceph/osd/ceph-|awk "{print \$3}"|cut -d- -f2|sort -n|xargs`

hillpd (changes requested):

sos/report/plugins/ceph_osd.py
pponnuvel: An equally ugly but slightly more efficient version of it can be
```
sudo mount|awk -v p=/var/lib/ceph/osd/ceph- '$0~p{split($3, x, "-"); printf x[2]" "}'
```
(sorting isn't necessary here).

nkshirsagar: Hi @pponnuvel  , @hillpd 

thank you for the review and suggestions!

The collect_cmd_output() method provided by sosreport does not allow pipes. So we need to get the output of a command, then implement the output extraction in the sos plugin.

Another approach might be to extract it from output of ps -u ceph -o args, and extract the id numbers that way (The ceph-osd process belong to the user ceph). 

output of "ps -u ceph -o args" looks like - 

COMMAND
/usr/bin/ceph-osd -f --cluster ceph --id 0 --setuser ceph --setgroup ceph
/usr/bin/ceph-osd -f --cluster ceph --id 1 --setuser ceph --setgroup ceph
/usr/bin/ceph-osd -f --cluster ceph --id 2 --setuser ceph --setgroup ceph

If we use this approach, we would not collect data for an OSD that is stopped by admin, or even down/crashed, possibly saving the plugin some time. Please let me know what you think.

TurboTurtle: Chiming in: pipes and bash foo in general aren't allowed in plugins.

You could do something like the following, if `ceph-volume lvm list` is not desired:

```python
out = self.collect_cmd_output('mount')
mounts = [
    m.split()[2] for m in out['output'].splitlines() if 'ceph/osd/ceph-' in m
]
```

However, the actual collection of `ceph-volume lvm list` seems important as it's been in sos for a while. What kind of performance concerns are there? Is it just related to the `ceph-volume` command, or does this affect the entire host? If we're going to keep collecting the output of `ceph-volume lvm list`, I have no issues with using it to build the list.

We have several safe guards in place to prevent any under-performing command collection from impacting the rest of the plugin, or the report as a whole, too much. At worst, we'd hit a timeout and report that timeout to the user, then continue on as normal.

sos/report/plugins/ceph_osd.py
nkshirsagar: ack done

sos/report/plugins/ceph_osd.py
```diff
@@ -39,9 +40,47 @@ def setup(self):
...
+        containers_list = self.get_all_containers_by_regex("ceph-osd*")
+
+        if containers_list:
+            self.add_cmd_output([self.fmt_container_cmd(
+                containers_list[0][1], "ceph daemon %s %s"
+                % (osid, cmd)) for osid in osd_ids for cmd in ceph_cmds])
+
+        else:
+            self.add_cmd_output(["ceph daemon %s %s" % (
+                osid, cmd) for osid in osd_ids for cmd in ceph_cmds])
```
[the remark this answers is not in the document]
nkshirsagar: ack, thanks.

sos/report/plugins/ceph_osd.py
hillpd: @nkshirsagar 
I think parsing either `ps` or `mount` output is a reasonable approach.

@TurboTurtle
I'm not suggesting that we stop collecting `ceph-volume lvm list`. We should just avoid using this command to build the `osd_ids` list. A time-out would prevent all of the `ceph_cmds` from being collected on lines 78 and 82.

> What kind of performance concerns are there? Is it just related to the `ceph-volume` command, or does this affect the entire host?

My concern is that the `ceph-volume` command may be prone to time-out on hosts running older Ceph code levels that sosreport can still target. There are several bugs I'm concerned about: [lp#1850754](https://bugs.launchpad.net/ubuntu/+source/ceph/+bug/1850754) and [lp#1908375](https://bugs.launchpad.net/ubuntu/+source/ceph/+bug/1908375).

slashdd (approved):
lgtm

sos/report/plugins/ceph_osd.py
```diff
@@ -39,9 +39,62 @@ def setup(self):
...
+           "ops",
+           "status",
+           "version",
+        ]
+
+        osd_ids = []
+        out = self.exec_cmd('ps -u ceph -o args')
+
+        if out['status'] != 0:
+            return
```
pmoravec, line 72: This will mean **whole remaining code in `setup()`** will be ignored and no command output from below commands will be collected - is that intentional?

sos/report/plugins/ceph_osd.py
```diff
@@ -39,9 +39,62 @@ def setup(self):
...
+
+        # Get the running OSD process IDs from ps
+        osd_nums = [
+                m.split()[5] for m in out['output'].splitlines()
+                if 'ceph --id' in m
+        ]
+
+        for element in osd_nums:
+            osd_ids.append('osd.' + element)
```
TurboTurtle, line 81: Shorten this up into a single list:

```python
if out['status'] == 0:
    osd_ids = [
        'osd.%s' % m.split()[5] for m in out['output'].splitlines()
        if 'ceph --id' in m
    ]
else:
    osd_ids = []
```

sos/report/plugins/ceph_osd.py
```diff
@@ -39,9 +39,62 @@ def setup(self):
...
+
+        osd_ids = []
+        out = self.exec_cmd('ps -u ceph -o args')
+
+        if out['status'] != 0:
+            return
+
+        # Get the running OSD process IDs from ps
+        osd_nums = [
+                m.split()[5] for m in out['output'].splitlines()
```
pmoravec, line 76: Are we sure each line will have at least 6 words, and no `IndexError` exception can be raised?

sos/report/plugins/ceph_osd.py
```diff
@@ -39,9 +39,62 @@ def setup(self):
...
+
+        for element in osd_nums:
+            osd_ids.append('osd.' + element)
+
+        containers_list = self.get_all_containers_by_regex("ceph-osd*")
+
+        if containers_list:
+            self.add_cmd_output([
+                self.fmt_container_cmd(
+                    containers_list[1][1], "ceph daemon %s %s"
```
pmoravec, line 88: Can't `containers_list` contain exactly one element / one container? (if so, also `IndexError` could happen)

sos/report/plugins/ceph_osd.py
TurboTurtle: Agreed, we don't need to return here. If the command status is non-zero, then the following logic to populate osd_ids can be skipped, which will in turn make the `add_cmd_output()` calls a no-op. But, by not issuing a return here, we still hit the `add_forbidden_path()` call, without needing to move it around as well within the flow of `setup()`.

sos/report/plugins/ceph_osd.py
nkshirsagar: agreed. this crept in through the changes made for https://github.com/sosreport/sos/pull/2736#discussion_r738476601 .. I guess 

self.fmt_container_cmd(
                    containers_list[0][1], "ceph daemon %s %s"

would work, since we just need any one container so the first one is as good as any.

sos/report/plugins/ceph_osd.py
nkshirsagar: ack thank you

sos/report/plugins/ceph_osd.py
nkshirsagar: Yes I think we can expect this output to be consistent like the format I see today, but I will check again with @pponnuvel or @hillpd if this might not be the case. I'd expect the id number always to be element number 6.

root@bronzor:~/sos/sos# ps -u ceph -o args
COMMAND
/usr/bin/ceph-osd -f --cluster ceph --id 0 --setuser ceph --setgroup ceph
/usr/bin/ceph-osd -f --cluster ceph --id 1 --setuser ceph --setgroup ceph
/usr/bin/ceph-osd -f --cluster ceph --id 2 --setuser ceph --setgroup ceph

sos/report/plugins/ceph_osd.py
nkshirsagar: done, thanks.

sos/report/plugins/ceph_osd.py
pponnuvel: It may be possible that there are other ceph processes running on the same host, not just OSDs. So it's possible to hit `IndexError`.

This could be handled by checking:
- the line contains at least 6 fields
- 5th field is `--id`
- 6th field is a number (not a string e.g. mon which could be running on the same machine in a hyperconverged env)

Which would rule out others except OSDs.

nkshirsagar: ack @pponnuvel , done

sos/report/plugins/ceph_osd.py
```diff
@@ -39,9 +39,60 @@ def setup(self):
...
+        osd_ids = []
+        out = self.exec_cmd('ps -u ceph -o args')
+
+        if out['status'] == 0:
+            # Get the running OSD process IDs from ps
+            osd_ids = [
+                'osd.%s' % m.split()[5] for m in out['output'].splitlines()
+                if 'ceph --id' in m and len(m.split()) > 5
+                and m.split()[5].isdigit() and m.split()[4] == "--id"
+            ]
```
pmoravec, line 77: We are redundantly calling `m.split()` many times in one iteration of `for m in ..`. But calling it once would mean less obvious/elegant code and I hope/expect the data are kept in some python's/kernel cache, and the "`ps`" output isnt too big.

TL;DR: I am not happy with this, but I see it as best option.

sos/report/plugins/ceph_osd.py
```diff
@@ -39,9 +39,60 @@ def setup(self):
...
+        else:
+            osd_ids = []
```
pmoravec, line 79: Dead code due to "`osd_ids = []`" in line 68 :)

pmoravec (approved):
ACK even for the current code (removing the "dead"(*) code is preferred, though).

(*) it is rather no-op than dead, program execution *can* get there but will apply nothing.

sos/report/plugins/ceph_osd.py
TurboTurtle: If we're going to expand the safeguard checks this much, we don't _need_ to keep everything in a single list constructor. The request to do this originally was because there was only a single check occurring and there were 2 lists being used to create the single list of osd ids.

```python
osd_ids = []
for procs in out['output'].splitlines():
    proc = procs.split()
    if len(proc) < 6:
        continue
    if proc[4] == '--id' and proc[5].isdigit():
        osd_ids.append("osd.%s" % proc[5])
```

Single `split()` call, and is very easily grok'd.

pmoravec: +1 to this.

sos/report/plugins/ceph_osd.py
nkshirsagar: removed, thanks.

sos/report/plugins/ceph_osd.py
nkshirsagar: leaving as is, thank you

pmoravec (approved):
> If we're going to expand the safeguard checks this much, we don't _need_ to keep everything in a single list constructor. The request to do this originally was because there was only a single check occurring and there were 2 lists being used to create the single list of osd ids.
> 
> ```python
> osd_ids = []
> for procs in out['output'].splitlines():
>     proc = procs.split()
>     if len(proc) < 6:
>         continue
>     if proc[4] == '--id' and proc[5].isdigit():
>         osd_ids.append("osd.%s" % proc[5])
> ```
> 
> Single `split()` call, and is very easily grok'd.

I would prefer to have this applied, but ACK even the current code.

nkshirsagar:
> > If we're going to expand the safeguard checks this much, we don't _need_ to keep everything in a single list constructor. The request to do this originally was because there was only a single check occurring and there were 2 lists being used to create the single list of osd ids.
> > ```python
> > osd_ids = []
> > for procs in out['output'].splitlines():
> >     proc = procs.split()
> >     if len(proc) < 6:
> >         continue
> >     if proc[4] == '--id' and proc[5].isdigit():
> >         osd_ids.append("osd.%s" % proc[5])
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Single `split()` call, and is very easily grok'd.
> 
> I would prefer to have this applied, but ACK even the current code.

Thank you for the review. I have done the changes as you suggested.

pmoravec (approved):

TurboTurtle (approved):

hillpd (approved):

PR #2739, opened 2021-11-01
[discovery] Add new discovery plugin

jcastill:
This commit creates as new plugin for the
discovery tool. This tool is an inspection
and reporting tool, designed to find, identify,
and report environment data.

Fixes: RHBZ#2018549

---

### DIFF
--- a/sos/report/plugins/discovery.py
+++ b/sos/report/plugins/discovery.py
@@ -1,0 +1,31 @@
+# Copyright (C) 2021 Red Hat, Inc., Jose Castillo <jcastillo@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, RedHatPlugin
+
+
+class Discovery(Plugin, RedHatPlugin):
+
+    short_desc = 'Discovery inspection and reporting tool'
+    plugin_name = 'discovery'
+    packages = ('discovery', 'discovery-tools',)
+
+    def setup(self):
+        self.add_copy_spec([
+            "/root/discovery/db/volume/data/userdata/pg_log/",
+            "/root/discovery/server/volumes/log/app.log",
+            "/root/discovery/server/volumes/log/discovery-server.log"
+        ])
+
+        self.add_container_logs([
+            'discovery',
+            'dsc-db'
+        ])
+# vim: set et ts=4 sw=4 :


sos/report/plugins/discovery.py
```diff
@@ -0,0 +1,32 @@
...
+        self.add_copy_spec([
+            "/root/discovery/db/volume/data/userdata/pg_log/",
+            "/root/discovery/server/volumes/log/app.log",
+            "/root/discovery/server/volumes/log/discovery-server.log"
+        ])
+
+        self.add_cmd_output([
+            "podman logs discovery",
+            "podman logs dsc-db"
+        ])
```
TurboTurtle, line 30: Hmm. I'm a little torn on this. One the one hand, I can see the relevance and I assume that discovery is run primarily in a container, so this collection would likely be considered significant. On the other, container logs are collected by the `podman` plugin, which has optional filtering of which containers to collect from.

I need to think on this for a bit, but at any rate I'd say we want to link this to a policy's loaded `ContainerRuntime` at a minimum, rather than hard coding it to be `podman`.

jcastill: I think having this as separated may be helpful, specially if we extend the discovery tool in the future. But I can see the value of having it inside the podman plugin. 
A question about this - is there a possibility to make an alias like this:

sos report -o discovery

That doesn't call a discovery plugin, but rather does 'sos report -o podman --log=discovery' ? Like an alias? Or is this making something simple too  complex?

TurboTurtle: It'd be pretty complex to do something like that, and is something users definitely would not expect out of the box.

Since the logs option for `podman` is default off, I think what would make sense here is if we added something like a `add_container_logs()` method to `Plugin` that uses the loaded container runtime to collect logs for a given container name or list of names.

Off the top of my head, perhaps something like this (untested):

```python
def add_container_logs(self, containers, **kwargs):
    _runtime = self._get_container_runtime()
    if _runtime is None:
        return None
    if isinstance(containers, str):
        containers = [containers]
    for container in containers:
        # allow regex matching container names
        _cons = self.get_all_containers_by_regex(container)
        for _con in _cons:
            self.add_cmd_output("%s logs %s" % (_runtime, _con[1]), **kwargs)
```

pmoravec: The idea of `add_container_logs` sounds good to me.

pmoravec: .. and to the "plugins' competencies to collect same data" - a possible approach would be having the `discovery` plugin collecting data as proposed, and enhance `podman`/`docker` plugins in prohibiting collection of `discovery` and `dsc-db` containers logs (similarly like https://github.com/sosreport/sos/blob/main/sos/report/plugins/apache.py#L33-L48)?

TurboTurtle: ...oh, heh. We actually already have the infrastructure in place for this - it's just badly named so it's easy to overlook. `Plugin.get_container_logs()` is a basic implementation of this already. However, I do see some shortcomings in it. I'm prepping an update to make it more robust, after which we can start using it here.

TurboTurtle: #2740 will allow this to be:

```python
self.add_container_logs([
    'discovery',
    'dsc-db'
])
```

TurboTurtle: #2740 has been merged. This can be updated with the `add_container_logs()` call mentioned above. Once that is done, this LGTM.

pmoravec: Hi Jose @jcastill , could you please rebase your PR and modify to utilise the #2740 ?

TurboTurtle:
@jcastill reminder ping. If this needs to land by sos-4.3, please update the PR as 4.3 is closing Feb 11th.

jcastill:
I'll work on this asap.

pmoravec (approved):

TurboTurtle (approved):

Issue #2744, opened 2021-11-03
-k plugin.timeout is applied to all plugins


pmoravec:
reproducer:

    python3 bin/sos report -o kernel,selinux -k selinux.timeout=1 --batch --build

causes *both* kernel and selinux plugins to timeout (within one second). Moreover, this leads to segfault like https://github.com/sosreport/sos/issues/2722 due to an attempt to write to closed file (will be resolved in that issue).

TurboTurtle:
Looks like this is because the default plugin options are built before the individual plugins are instantiated. I'm hammering out a fix for this on my end, and trying to make a test case for it going forward. I'll post a PR later today for it, as this is quite significant.

PR #2745, opened 2021-11-03
[Plugin] Ensure specific plugin timeouts are only set for that plugin

TurboTurtle:
It was discovered that setting a specific plugin timeout via the `-k
$plugin.timeout` option could influence the timeout setting for other
plugins that are not also having their timeout explicitly set. Fix this
by moving the default plugin opts into `Plugin.__init__()` so that each
plugin is ensured a private copy of these default plugin options.

Additionally, add more timeout data to plugin manifest entries to allow
for better tracking of this setting.

Adds a test case for this scenario.

Closes: #2744

---

### DIFF
--- a/sos/report/__init__.py
+++ b/sos/report/__init__.py
@@ -775,7 +775,7 @@ class SoSReport(SoSComponent):
         if self.all_options:
             self.ui_log.info(_("The following options are available for ALL "
                                "plugins:"))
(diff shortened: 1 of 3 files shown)


pmoravec (approved):

TurboTurtle:
Had an incorrect tag on the new test. Fixed, and I've also moved it into `vendor_tests/`. Everything else remains the same, will merge once the CI tests are complete.

Issue #2722, opened 2021-10-08
unhandled exception when plugin timeout happens during logging a collected file


pmoravec:
Reproducer:

    sos report -o selinux --batch --build -k selinux.timeout=1 -vv
    ..
    [plugin:selinux] copying path '/var/lib/selinux/targeted/active/modules/100/webalizer/hll' to archive:'/var/lib/selinux/targeted/active/modules/100/webalizer/hll'
    
     Plugin selinux timed out
    
    [plugin:selinux] collecting path '/var/lib/selinux/targeted/active/modules/100/postgresql/hll'
    
    [plugin:selinux] collecting path '/var/lib/selinux/targeted/active/modules/100/uuidd/hll'
    ..
    [plugin:selinux] collecting path '/var/lib/selinux/targeted/active/modules/100/sudo/hll'
    Your sosreport build tree has been generated in:
    	/var/tmp/sosreport-pmoravec-rhel8-2021-10-08-uyaoush
    
    
    [plugin:selinux] collecting path '/var/lib/selinux/targeted/active/modules/100/snort/cil'
    --- Logging error ---
    Traceback (most recent call last):
      File "/usr/lib64/python3.6/logging/__init__.py", line 996, in emit
        stream.write(msg)
    ValueError: I/O operation on closed file.
    Call stack:
      File "/usr/lib64/python3.6/threading.py", line 884, in _bootstrap
        self._bootstrap_inner()
      File "/usr/lib64/python3.6/threading.py", line 916, in _bootstrap_inner
        self.run()
      File "/usr/lib64/python3.6/threading.py", line 864, in run
        self._target(*self._args, **self._kwargs)
      File "/usr/lib64/python3.6/concurrent/futures/thread.py", line 69, in _worker
        work_item.run()
      File "/usr/lib64/python3.6/concurrent/futures/thread.py", line 56, in run
        result = self.fn(*self.args, **self.kwargs)
      File "/root/sos-main/sos/report/__init__.py", line 1083, in collect_plugin
        plug.collect()
      File "/root/sos-main/sos/report/plugins/__init__.py", line 2667, in collect
        self._collect_copy_specs()
      File "/root/sos-main/sos/report/plugins/__init__.py", line 2631, in _collect_copy_specs
        self._log_info("collecting path '%s'" % path)
      File "/root/sos-main/sos/report/plugins/__init__.py", line 706, in _log_info
        self.soslog.info(self._format_msg(msg))
[...]

PR #2746, opened 2021-11-03
[report] shutdown threads for timeouted plugins

pmoravec:
Wait for shutting down threads of timeouted plugins, to prevent
them in writing to moved auxiliary files like sos_logs/sos.log

Resolves: #2722
Closes: #2746

---


TurboTurtle:
At first this struck me as a possible upstream python bug, since documentation states that you don't need to call `shutdown()` with using `with` manager - https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.shutdown

However, I think this may be an edge case for us specifically since we're handling the thread termination still within the `with` statement. I wonder if using `shutdown()` makes the subsequent `pool._threads.clear()` unnecessary?

TurboTurtle (approved):

Issue #2738, opened 2021-10-30
TMPDIR


ciq-mcarifio:
sos report doesn't seem to use the value of TMPDIR to write reports. I understand `--tmp-dir` is a switch and `/etc/sos/sos.conf` will allow for a setting. Still seems like `TMPDIR` should be used. Ty.

TurboTurtle:
Where are you seeing `--tmpdir` not being used? Running against current `main`, it works as expected for me:

```
$ sudo ./bin/sos report --batch -q -o kernel 

sosreport (version 4.2)

Your sosreport has been generated and saved in:
	/var/tmp/sosreport-terra-2021-10-30-shyobvm.tar.xz

$ sudo ./bin/sos report --batch -q -o kernel --tmp-dir=/tmp

sosreport (version 4.2)

Your sosreport has been generated and saved in:
	/tmp/sosreport-terra-2021-10-30-tyuepgz.tar.xz

$ sudo ./bin/sos report --batch -q -o kernel --tmp-dir=/home/jake

sosreport (version 4.2)

Your sosreport has been generated and saved in:
	/home/jake/sosreport-terra-2021-10-30-ebtxocg.tar.xz

```

ciq-mcarifio:
Yup. That works. I was thinking: `TMPDIR=/tmp sudo -E ./bin/sos report --batch -q -o kernel`. Often the environment variable is picked up. But it's not guaranteed.

TurboTurtle:
Hmm. I can see the desire for that, but at least for me I think a cmdline option and a config file setting is sufficient. Though I'm not solidly against it, either. 

@pmoravec @BryanQuigley @bmr-cymru - any thoughts either way?

pmoravec:
`sos` would be nice if it supports `TMPDIR`, as the variable is commonly used as defined in [Single UNIX Specification](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html#tag_08_03).

Though I think it is lower prio but it makes sense to support the env.var. If neither cmdline option or config option is specified, we should source `TMPDIR` env.var (or maybe we can use the env.var default, to be overwritten by either cmdline or config option?).

PR #2747, opened 2021-11-04
[sos] Reference TMPDIR environment variable if set

TurboTurtle:
If the `TMPDIR` env var is set, we should reference it if the user has
not provided `--tmp-dir` by the cmdline or sos.conf.

The order of precedence is now:

	1. cmdline use of `--tmp-dir`
	2. setting `tmp-dir` in `/etc/sos/sos.conf`
	3. the `TMPDIR` environment variable
	4. `/var/tmp` as a default

Closes: #2738

---

### DIFF
--- a/sos/component.py
+++ b/sos/component.py
@@ -14,6 +14,7 @@ import logging
 import os
 import tempfile
 import sys
+import time
 
 from argparse import SUPPRESS
 from datetime import datetime
@@ -22,7 +23,7 @@ from pathlib import Path
 from sos import __version__
 from sos.archive import TarFileArchive
 from sos.options import SoSOptions
-from sos.utilities import TempFileUtil
+from sos.utilities import TempFileUtil, shell_out
 
 
 class SoSComponent():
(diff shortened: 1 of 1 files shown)


pmoravec (approved):

bmr-cymru:
The reason we switched from `/tmp` to `/var/tmp` a while back was twofold:

  * Preserves content across reboots (so a partial tree can be recovered in the event of a fatal system error during report gathering).
  * Avoids system memory pressure from the tmp-on-tmpfs change that was made back in Fedora 18

I think maybe along with this change it's worth revisting this and adding a check and warning if `TMPDIR` or `--tmp-dir` evaluate to a `tmpfs` backed file system. The effects can be pretty bad on low memory systems: if they have swap they can grind to a halt and without they can wind up OOM killing.

TurboTurtle:
I like the idea of a warning here. I'll update this commit to include that check.

TurboTurtle:
Updated to check fs type and print a warning if that is `tmpfs`.

pmoravec (approved):

PR #2748, opened 2021-11-04
[report] fix filter_namespace per pattern

pmoravec:
Curently, -k networking.namespace_pattern=.. is broken as the R.E. test
forgets to add the namespace in case of positive match.

Resolves: #2748

---

### DIFF
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -2953,21 +2953,20 @@ class Plugin():
                 )
         for ns in ns_list:
             # if ns_pattern defined, skip namespaces not matching the pattern
-            if ns_pattern:
-                if not bool(re.match(pattern, ns)):
-                    continue
+            if ns_pattern and not bool(re.match(pattern, ns)):
+                continue
+            out_ns.append(ns)
 
-            # if ns_max is defined at all, limit returned list to that number
+            # if ns_max is defined at all, break the loop when the limit is
+            # reached
             # this allows the use of both '0' and `None` to mean unlimited
-            elif ns_max:
-                out_ns.append(ns)
+            if ns_max:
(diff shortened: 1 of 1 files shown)


pmoravec:
Trivial reproducer:

    ip netns add foobar
    sos report --batch --build -o networking -k networking.namespace_pattern='foo*'

does not collect "`ip netns exec foobar ..`" command(s).

sos/report/plugins/__init__.py
```diff
@@ -2954,8 +2954,8 @@ def filter_namespaces(self, ns_list, ns_pattern=None, ns_max=None):
         for ns in ns_list:
             # if ns_pattern defined, skip namespaces not matching the pattern
             if ns_pattern:
-                if not bool(re.match(pattern, ns)):
-                    continue
+                if bool(re.match(pattern, ns)):
+                    out_ns.append(ns)
```
TurboTurtle, line 2958: I don't think this is right. We're issuing a `continue` here because we aren't matching the pattern. We don't directly add the namespace to `out_ns` here because there are still other criteria.

I think instead we need to change separate this conditional, rather than a continuation of the pattern check
 https://github.com/sosreport/sos/blob/dd8defdeec62a6f086bc2afebfb87b4af74178c0/sos/report/plugins/__init__.py#L2962 - i.e. make `elif` a new `if` statement.

pmoravec: Right; my patch works with the (wrong) assumption that the pattern and max can not be set concurrently (at least the original code suggested me that assumption). So when the options are used in conjunction, `ns_max` is ignored.

Updating the PR accordingly.

TurboTurtle (approved):
LGTM

Issue #2205, opened 2020-08-19
Scope Discussion: new info|help component


TurboTurtle:
Now that 4.0 has closed, I'd like to prioritize one of the components that slipped from the release.

First is there a preference for either `sos info` or `sos help`? Or should we alias one to the other like we do with `clean|mask`?

Second, let's nail down the desired output. What I'm imagining for the use case here is an end use wants details on a specific plugin - such as expected collections, details on plugin options, etc... A rough template could be like:

```
Kernel Plugin Info - <short_desc>

<some longer description here - perhaps use docstrings?>

The following plugin options are available:
    with-timer            <short desc from option tuple>        <default value>
            <longer detailed description>
    [...other plugin options...]

This plugin is expected to collect the following output:

    commands:
        <list of add_cmd_output() calls>

    files:
       <list of add_copy_spec() specs>

```

This raises a few questions to me:

1. Should we use docstrings for the long description, or another class attr?
2. How should we handle longer plugin option descriptions? Extend the tuple, or perhaps make plugin options a distinct class for easier/better manipulation?
3. Do we just grok the plugin for a_c_o and a_c_s calls, or do we run the plugin in dry run mode to account for iterative collections over interfaces? On that note, do we actually want to list all individual files being collected by a_c_s, or just the high level specs that may include globs or dirs? Do we even _want_ this output?
4. Would it be beneficial to add a method ala `help_output()` to `Plugin` that allows plugin authors to manually control output formatting and content?

For reference, my initial thoughts are:

1. Docstrings
2. Make it a class
3. grok both, but have an option to dry run commands to get all iterations across interfaces
4. I think such a method should _extend_ the help output, not override, if we add one. I say this, but then I think of how bloated the `openshift` plugin will be using defaults like the above, in which case an override would be better.

Third, what kind of command invocation should be used here. Do we interpret bare strings, or make some kind of template? E.G. `sos info|help kernel` vs. `sos info|help plugins.kernel` vs `sos info|help --plugin kernel`. 

With the first, we may run into conflicts, e.g. `sos info|help ubuntu` - does this refer to the plugin or the policy? Do we even want to provide help output for policies? Does `sos info|help ovirt` refer to the plugin or the cluster profile for `collect`?

I'm just trying to brainstorm here before we dive wholesale into development of this component, so please provide any and all comments or concerns.

pmoravec:
> For reference, my initial thoughts are:
> 
>     1. Docstrings
> 
>     2. Make it a class
> 
[...]

PR #2749, opened 2021-11-05
[sos] Add a new `help` component

TurboTurtle:
This patchset adds a new `help` component, enabling the execution of `sos help $topic`.

This subcommand is intended to provide more detailed information on the various parts of the SoS project that cannot reasonably be maintained in a manpage or `--help` output.

For this initial implementation, I have added help for `report`, `collect`, `clean|mask`, as well as various constructs within those commands. I have not yet however added help sections for parsers within `clean`, as there is very little user control or interaction there. As we build that out, we can add help sections at that time.

Generally speaking, this makes it so items like plugins have help pages built automatically from the plugins themselves - i.e. enablement criteria, available options, etc... I would eventually like to see if we can include a basic representation of collections the plugin will do. For automatically constructed help sections (plugins, cluster profiles, transports, policies, etc...) the class docstrings are used, so this implies we should put a greater emphasis on including those in future development.

I have updated a few common plugins to serve as examples of how to build out this documentation. Doing this for all 300+ plugins currently in the project is out of scope for this PR, and we can extend those over time.

Closes: #2205 

---

### DIFF
--- a/sos/collector/clusters/__init__.py
+++ b/sos/collector/clusters/__init__.py
@@ -11,6 +11,7 @@
 import logging
 
 from sos.options import ClusterOption
+from sos.utilities import bold
 from threading import Lock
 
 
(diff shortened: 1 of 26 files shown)


sos/policies/distros/redhat.py
```diff
@@ -206,6 +206,32 @@ def get_tmp_dir(self, opt_tmp_dir):
...
+    policies.
+
+    Plugins activated by installed packages will only be activated if those
+    packages are installed via RPM (dnf/yum inclusive). Packages installed by
+    other means are not considered by this policy.
+
+    By default, --upload will be directed to using the SFTP location provided
+    by Red Hat for technical support cases. Users who provide login credentials
+    for their Red Hat Customer Portal account will have their archives uploaded
+    to a user-specific directory.
```
pmoravec, line 223: By default, `--upload` will rely on default `--upload-protocol=auto`, which will attempt an upload to the Red Hat Customer Portal (if credentials provided). Only when this fails or when credentials are missing, we failover to uploading to the SFTP server.

pmoravec: I mean, starting the comment with "default is SFTP" sounds confusing to me, despite it is properly explained in the next paragraph.

pmoravec:
(maybe irrelevant comment)

Assuming syntax will be `sos help report.plugins.$plugin`, it might be difficult for users to "guess" the hierarchy - what prefix to use for `ovirt` cluster or `RHELPolicy` (which is a distro and also policy)? Can we have a subcommand to enumerate all "help pages" / show the "help tree subcommands/topics"?

sos/policies/distros/redhat.py
TurboTurtle: Yeah, this could go either way. The reason I went with saying SFTP is the 'default', is the is because we require 3 non-default options in order to attempt the customer portal.

TurboTurtle:
> (maybe irrelevant comment)
> 
> Assuming syntax will be `sos help report.plugins.$plugin`, it might be difficult for users to "guess" the hierarchy - what prefix to use for `ovirt` cluster or `RHELPolicy` (which is a distro and also policy)? Can we have a subcommand to enumerate all "help pages" / show the "help tree subcommands/topics"?

~~~
$ ./bin/sos help

[...]
How to search using sos help
To get more information on a given topic, use the form '$component.$topic.$subtopic'.

For example 'sos help report.plugins.kernel' will provide more information on the kernel plugin for the
report function.

        report                      Detailed help on the report command
        report.plugins              Information on the plugin design of sos
        report.plugins.$plugin      Information on a specific $plugin
[...]
~~~

Should this be built out more?

sos/collector/clusters/__init__.py
```diff
@@ -81,6 +82,106 @@ def name(cls):
...
+    @classmethod
+    def display_help(cls, section):
+        if cls is Cluster:
+            cls.display_self_help(section)
+            return
+        section.set_title("%s Cluster Profile Detailed Help"
+                          % cls.cluster_name)
+        if cls.__doc__ and cls.__doc__ is not Cluster.__doc__:
+            section.add_text(cls.__doc__)
+        elif cls.__mro__[1].__doc__ and cls.__mro__[1] is not Cluster:
```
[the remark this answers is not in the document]
TurboTurtle: Ack

sos/help/__init__.py
```diff
@@ -0,0 +1,303 @@
...
+                    cls = _help[_sec]()
+                    break
+        return cls
+
+    def _get_collect_transport(self):
+        from sos.collector.sosnode import TRANSPORTS
+        _transport = self.opts.topic.split('.')[-1]
+        if _transport in TRANSPORTS:
+            return TRANSPORTS[_transport]
+        raise Exception("No such transport '%s'" % _transport)
```
pmoravec, line 129: Worth raising an equivalent exception also in `_get_collect_cluster`, `_get_plugin_variant` and `_get_policy_by_name`?

pmoravec:
"Global Options:" are redundant when running `sos help --help` :) (but it could be tricky to not print them, I would leave it as is).

I like the `help` subcommand! Nice work.

sos/help/__init__.py
TurboTurtle: Actually, I think removing this exception is better as that gives us our stock "No help section found" handling.

TurboTurtle:
Rebased and updated to address the 2 points above.

sos/collector/clusters/__init__.py
```diff
@@ -81,6 +82,107 @@ def name(cls):
...
+        section.add_text(
+            'Cluster profiles are enabled similarly to SoS report plugins; '
+            'usually by package, command, or configuration file presence. '
+            'Clusters may also define default transports for SoS collect.'
+        )
+
+        from sos.collector import SoSCollector
+        import sos.collector.clusters
+        clusters = SoSCollector._load_modules(sos.collector.clusters,
+                                              'clusters')
```
pmoravec, line 170: LGTM complains here on self-import. Not sure how legitimate it is, but an alternative here is:

        import inspect
        clusters = SoSCollector._load_modules(inspect.getmodule(cls),
                                              'clusters')

pmoravec:
Sounds good otherwise.

sos/collector/clusters/__init__.py
TurboTurtle: :+1:

pmoravec (approved):
Nice work!

PR #2760, opened 2021-11-16
[networking] collect nft list ruleset per each namespace

pmoravec:
.. but only when nf_tables kmod is loaded

Resolves: #2760

---

### DIFF
--- a/sos/report/plugins/firewall_tables.py
+++ b/sos/report/plugins/firewall_tables.py
@@ -16,7 +16,7 @@ class firewall_tables(Plugin, IndependentPlugin):
     plugin_name = "firewall_tables"
     profiles = ('network', 'system')
 
-    def collect_iptable(self, tablename):
+    def collect_iptable(self, tablename, kmod):
         """ Collecting iptables rules for a table loads either kernel module
         of the table name (for kernel <= 3), or nf_tables (for kernel >= 4).
         If neither module is present, the rules must be empty."""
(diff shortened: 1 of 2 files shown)


pmoravec:
Adding one independent commit in similar area (that alters same code). See RHBZ 2001096 for explanation.

TurboTurtle:
```
(049/264) tests/report_tests/basic_report_tests.py:NormalSoSReport.test_no_new_kmods_loaded:  FAIL: Element counts were not equal:\nFirst has 0, Second has 1:  'nfnetlink' (0.03 s)
```

Interesting that we only loaded this on F34. I've restarted the job just to double check this is valid. If so, we'll need to investigate what specifically is loading it.

pmoravec:
Hm there is still the

     (049/264) tests/report_tests/basic_report_tests.py:NormalSoSReport.test_no_new_kmods_loaded:  FAIL: Element counts were not equal:\nFirst has 0, Second has 1:  'nfnetlink' (0.03 s)

so I will investigate it.

pmoravec:
OK, the typo was found.

But..

If `nfnetlink` kmod is not loaded, then `networking` plugin will load it. Due to the test `iptables -V | grep nf_tables`. Cf. https://github.com/sosreport/sos/pull/2760/files#diff-ac22bccaa91329cc6c801081c9399a16d2777fca9d6be962846cf315dfaece50R68-R72 and:

    # lsmod | grep -e nfnetlink -e nf_tables
    # iptables -V
    iptables v1.8.4 (nf_tables)
    # lsmod | grep -e nfnetlink -e nf_tables
    nfnetlink              16384  0
    # 

This is a minefield. Have to we guard an execution of a guard of a command, even?

pmoravec:
.. and the failed test is exactly due to the `nfnetlink` kmod.

arif-ali:
@pmoravec do we still want this to land? now we have conflict

TurboTurtle:
Closing as abandoned.

PR #2761, opened 2021-11-17
[clean] Improve hostname obfuscation sourcing and reliability

TurboTurtle:
This patchset addresses two issues:

1. Adds sourcing of `/etc/hosts`, which we aren't doing currently, provides valuable information for prepping the hostname parser, in particular with short names where we would otherwise have a race condition for obfuscation.
2. Fixes a condition where our shortname handling was unintentionally case sensitive.

---

### DIFF
--- a/sos/cleaner/mappings/hostname_map.py
+++ b/sos/cleaner/mappings/hostname_map.py
@@ -169,13 +169,13 @@ class SoSHostnameMap(SoSMap):
 
     def sanitize_item(self, item):
         host = item.split('.')
-        if all([h.isupper() for h in host]):
+        if len(host) > 1 and all([h.isupper() for h in host]):
             # by convention we have just a domain
             _host = [h.lower() for h in host]
             return self.sanitize_domain(_host).upper()
         if len(host) == 1:
             # we have a shortname for a host
-            return self.sanitize_short_name(host[0])
+            return self.sanitize_short_name(host[0].lower())
         if len(host) == 2:
             # we have just a domain name, e.g. example.com
             return self.sanitize_domain(host)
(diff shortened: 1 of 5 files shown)


sos/cleaner/mappings/hostname_map.py
```diff
@@ -169,13 +169,13 @@ def get(self, item):
 
     def sanitize_item(self, item):
         host = item.split('.')
-        if all([h.isupper() for h in host]):
+        if len(host) > 1 and all([h.isupper() for h in host]):
             # by convention we have just a domain
             _host = [h.lower() for h in host]
             return self.sanitize_domain(_host).upper()
```
pmoravec, line 175: We upper-case domains due to LDAP/Kerberos/AD realms being in uppercase, while we lower-case shortnames to have a canonical representation otherwise?

I can't find a counterexample to the approach we use, but I am not sure why we call "upper()" for domains and "lower()" for shortnames.

TurboTurtle: The convention is that a string like `FOO.EXAMPLE.COM` is a domain, but I've recently encountered some instances where just the shortname was all uppercase. The addition of `len(host) > 1` is to make sure we aren't passing short names to `sanitize_domain()`.

We `upper()` the returned domain to match the specific context of these types of regex matches, however we `lower()` everything for obfuscation so that we aren't creating duplicate mappings, e.g. we don't want `myhost` to map to `host0`, `MYHOST` to `host1` and something like `MyHoSt` to `host2`, we want all of it to map to `host0`.

pmoravec (approved):
I see, that makes sense. Thx for explanation.

PR #2762, opened 2021-11-18
[global] Refactor sysroot determination and usage across different components

TurboTurtle:
This patchset addresses a small number of issues surrounding sysroot handling inside containers:

1. Use the correct sysroot (the policy determined sysroot, not just blindly passing the default) for `PackageManager`. I am unsure where/when this broke, but this resolves sos using the container-local package list instead of the host package list.
2. Extend kmod detection to account for non-`/` sysroots.
3. Enable `ContainerRuntime`s to activate when sos is run in a container, and the runtime is accessible through the policy-determined sysroot.

---

### DIFF
--- a/sos/policies/distros/__init__.py
+++ b/sos/policies/distros/__init__.py
@@ -71,19 +71,18 @@ class LinuxPolicy(Policy):
     def __init__(self, sysroot=None, init=None, probe_runtime=True):
         super(LinuxPolicy, self).__init__(sysroot=sysroot,
                                           probe_runtime=probe_runtime)
-        self.init_kernel_modules()
 
-        # need to set _host_sysroot before PackageManager()
         if sysroot:
-            self._container_init()
-            self._host_sysroot = sysroot
+            self.sysroot = sysroot
         else:
-            sysroot = self._container_init()
+            self.sysroot = self._container_init()
+
+        self.init_kernel_modules()
 
         if init is not None:
             self.init_system = init
         elif os.path.isdir("/run/systemd/system/"):
-            self.init_system = SystemdInit(chroot=sysroot)
+            self.init_system = SystemdInit(chroot=self.sysroot)
         else:
             self.init_system = InitSystem()
 
@@ -149,27 +148,30 @@ class LinuxPolicy(Policy):
             if os.environ[ENV_CONTAINER] in ['docker', 'oci', 'podman']:
                 self._in_container = True
         if ENV_HOST_SYSROOT in os.environ:
-            self._host_sysroot = os.environ[ENV_HOST_SYSROOT]
-        use_sysroot = self._in_container and self._host_sysroot is not None
+            _host_sysroot = os.environ[ENV_HOST_SYSROOT]
+        use_sysroot = self._in_container and _host_sysroot is not None
         if use_sysroot:
-            host_tmp_dir = os.path.abspath(self._host_sysroot + self._tmp_dir)
+            host_tmp_dir = os.path.abspath(_host_sysroot + self._tmp_dir)
             self._tmp_dir = host_tmp_dir
-        return self._host_sysroot if use_sysroot else None
+        return _host_sysroot if use_sysroot else None
 
     def init_kernel_modules(self):
         """Obtain a list of loaded kernel modules to reference later for plugin
         enablement and SoSPredicate checks
         """
         self.kernel_mods = []
+        release = os.uname().release
 
         # first load modules from lsmod
-        lines = shell_out("lsmod", timeout=0).splitlines()
+        lines = shell_out("lsmod", timeout=0, chroot=self.sysroot).splitlines()
         self.kernel_mods.extend([
             line.split()[0].strip() for line in lines[1:]
         ])
 
         # next, include kernel builtins
-        builtins = "/usr/lib/modules/%s/modules.builtin" % os.uname().release
+        builtins = self.join_sysroot(
+            "/usr/lib/modules/%s/modules.builtin" % release
+        )
         try:
             with open(builtins, "r") as mfile:
                 for line in mfile:
@@ -186,7 +188,7 @@ class LinuxPolicy(Policy):
             'dm_mod': 'CONFIG_BLK_DEV_DM'
         }
 
-        booted_config = "/boot/config-%s" % os.uname().release
+        booted_config = self.join_sysroot("/boot/config-%s" % release)
         kconfigs = []
         try:
             with open(booted_config, "r") as kfile:
@@ -199,6 +201,11 @@ class LinuxPolicy(Policy):
         for builtin in config_strings:
             if config_strings[builtin] in kconfigs:
                 self.kernel_mods.append(builtin)
+
+    def join_sysroot(self, path):
+        if self.sysroot and self.sysroot != '/':
+            path = os.path.join(self.sysroot, path.lstrip('/'))
+        return path
 
     def pre_work(self):
         # this method will be called before the gathering begins
--- a/sos/policies/runtimes/__init__.py
+++ b/sos/policies/runtimes/__init__.py
@@ -64,7 +64,7 @@ class ContainerRuntime():
         :returns: ``True`` if the runtime is active, else ``False``
         :rtype: ``bool``
         """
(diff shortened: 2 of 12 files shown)


TurboTurtle:
Digging into this a bit more, I found that we've splintered a bit over time in sysroot determination and usage. I've elected to squash the previous commits into one with the additional changes needed to consolidate sysroot determination and usage based on `Policy`.

sos/policies/runtimes/__init__.py
```diff
@@ -55,7 +55,7 @@ def load_container_info(self):
         self.images = self.get_images()
         self.volumes = self.get_volumes()
 
-    def check_is_active(self):
+    def check_is_active(self, sysroot=None):
```
npinaeva, line 58: From my tests it looks like only passing sysroot here isn't enough, since get_containers, get_images, ... functions are not using this sysroot and therefore can't get right output

sos/policies/distros/__init__.py
```diff
@@ -71,19 +71,19 @@ class LinuxPolicy(Policy):
     def __init__(self, sysroot=None, init=None, probe_runtime=True):
         super(LinuxPolicy, self).__init__(sysroot=sysroot,
                                           probe_runtime=probe_runtime)
-        self.init_kernel_modules()
 
         # need to set _host_sysroot before PackageManager()
```
npinaeva, line 75: This comment seems to get here by mistake, since PackageManager is initialized in child classes

npinaeva (commented):
PackageManager works, container runtimes, not sure (left some suggestions)

If you're interested, PackageManager was broken here https://github.com/sosreport/sos/pull/2702/files#diff-9e337c4eedfbd3c072d5d88a69b34a2d859ed1c5f5cae9180a62a9a25f468ef8L66

sos/policies/distros/__init__.py
TurboTurtle: Yup, left over from moving the other bits around.

sos/policies/runtimes/__init__.py
TurboTurtle: Ack, I think I see where this breaks.

sos/policies/distros/__init__.py
pmoravec: Still (well, a nitpick) `s/set _host_sysroot/set sysroot/1` (or `set self.sysroot`)?

TurboTurtle: I was just going to remove the comment entirely in my next update

sos/policies/distros/__init__.py
```diff
@@ -200,6 +203,11 @@ def init_kernel_modules(self):
             if config_strings[builtin] in kconfigs:
                 self.kernel_mods.append(builtin)
 
+    def join_sysroot(self, path):
```
pmoravec, line 206: Why we can't use `archive`'s `join_sysroot` directly?

pmoravec (approved):
ACK otherwise.

sos/policies/distros/__init__.py
TurboTurtle: 1. `Policy` does not have access to an `Archive()`  directly (but `Archive()` _does_ have access to `Policy()`)
2. This is being used for kmod detection using the local files - `Archive()` isn't in the equation for accessing these. 
3. This commit has `Policy` be the ultimate source of `sysroot` truth, so it'd be a bit weird to wire up `Archive` here when `Archive` is dependent on `Policy` to know what `sysroot` even is.

pmoravec: Mea culpa, I should think before I write - esp. at Friday's evening :)

TurboTurtle:
I'm currently coming up short as to why we're still seeing failures in the `foreman_installer` plugin with this change. The failures are around symlinks for `/etc/foreman-installer/scenarios.d/foreman*.migrations`, which point to locations from `/usr/share/foreman-installer/config/foreman*.migrations`.

Reviewing the CI failure tarballs, the symlink is failing to be created in `/etc/...` because there is **already** a symlink there by that name, but it is broken because for some reason it is pointing to the wrong location (specifically, it is missing a `../` level). Further, the actual content under `/usr/share/foreman-installer/config/foreman*.migrations/` is missing, even though the directory is created and present in the archive.

TurboTurtle:
```
2021-11-19 21:44:43,557 DEBUG: [plugin:foreman_installer] made link target '/usr/share/foreman-installer/config/foreman-proxy-content.migrations' relative as '../../usr/share/foreman-installer/config/foreman-proxy-content.migrations'
2021-11-19 21:44:43,557 DEBUG: [plugin:foreman_installer] copying link '/etc/foreman-installer/scenarios.d/foreman-proxy-content.migrations' pointing to '/usr/share/foreman-installer/config/foreman-proxy-content.migrations' with isdir=True
2021-11-19 21:44:43,558 DEBUG: [plugin:foreman_installer] link '/usr/share/foreman-installer/config/foreman-proxy-content.migrations' is a directory, skipping...
2021-11-19 21:44:43,558 INFO: [plugin:foreman_installer] collecting path '/etc/foreman-installer/scenarios.d/foreman-proxy-content.migrations/.applied'
2021-11-19 21:44:43,559 DEBUG: [plugin:foreman_installer] made link target '/etc/foreman-installer/scenarios.d/foreman-proxy-content-migrations-applied' relative as '../../../../etc/foreman-installer/scenarios.d/foreman-proxy-content-migrations-applied'
2021-11-19 21:44:43,559 DEBUG: [plugin:foreman_installer] copying link '/etc/foreman-installer/scenarios.d/foreman-proxy-content.migrations/.applied' pointing to '/etc/foreman-installer/scenarios.d/foreman-proxy-content-migrations-applied' with isdir=False
2021-11-19 21:44:43,562 ERROR: caught exception in plugin method "foreman_installer.collect()"
2021-11-19 21:44:43,562 ERROR: writing traceback to sos_logs/foreman_installer-plugin-errors.txt

[...]

  File "/tmp/cirrus-ci-build/sos/report/__init__.py", line 1079, in collect_plugin
    plug.collect()
  File "/tmp/cirrus-ci-build/sos/report/plugins/__init__.py", line 2663, in collect
    self._collect_copy_specs()
  File "/tmp/cirrus-ci-build/sos/report/plugins/__init__.py", line 2628, in _collect_copy_specs
    self._do_copy_path(path)
  File "/tmp/cirrus-ci-build/sos/report/plugins/__init__.py", line 1228, in _do_copy_path
    self._copy_symlink(srcpath)
  File "/tmp/cirrus-ci-build/sos/report/plugins/__init__.py", line 1115, in _copy_symlink
    self.archive.add_link(reldest, dstpath)
  File "/tmp/cirrus-ci-build/sos/archive.py", line 412, in add_link
    dest = self._check_path(link_name, P_LINK)
  File "/tmp/cirrus-ci-build/sos/archive.py", line 299, in _check_path
    self._make_leading_paths(src_dir)
  File "/tmp/cirrus-ci-build/sos/archive.py", line 246, in _make_leading_paths
    os.symlink(target, abs_path)
FileExistsError: [Errno 17] File exists: '../../../usr/share/foreman-installer/config/foreman-proxy-content.migrations' -> '/var/tmp/avocado_9697ahlzsos_tests.py.ForemanWithOptionsTest/sos.ruvwv61t/sosreport-cirrus-task-5435606386868224-2021-11-19-wxvjvfh/etc/foreman-installer/scenarios.d/foreman-proxy-content.migrations'

```

So, we try to copy a link, /etc/foreman-installer/scenarios.d/foreman-proxy-content.migrations`, targeting `/usr/share/foreman-installer/config/foreman-proxy-content.migrations/`, which is a real directory. We specify that it `isdir` just before we call `strip_sysroot()` on it and then `Archive.add_link()`.

This is _apparently_ successful, as the `add_link()` call does not raise an exception and we continue on to https://github.com/sosreport/sos/blob/cd9e9f0301f94fc593de4d0756f8919a1fb687f7/sos/report/plugins/__init__.py#L1118 which ends this copy.

However, the very next copy is a file within the directory we just processed and it fails with...trying to create the parent directory symlink?

npinaeva (approved):
Works!

Issue #2759, opened 2021-11-15
Collect MGR specific information in the ceph_mgr plugin


nkshirsagar:
#1945 split the ceph plugin into individual component plugins. One of the reasons this was done was so we could add component specific data collection for each individual ceph service.

This issue has been opened to enhance the ceph_mgr plugin to collect the ceph manager related detailed data. Data the ceph_mgr plugin will collect includes

ceph daemon mgr.id < ... >

config diff
config show
dump_cache
dump_mempools
dump_osd_network
mds_requests
mds_sessions
objecter_requests
get_command_descriptions
mds_requests
mds_sessions
perf dump
perf histogram dump
perf histogram schema
perf schema
status
version

If the mgr is colocated with the mon (as is the usual case) and the keyring is present on the machine, additionally we might also want to collect,

ceph mgr dump
ceph mgr metadata
ceph mgr module ls
ceph mgr services
ceph mgr versions

TurboTurtle:
> This issue has been opened to enhance the ceph_mgr plugin to collect the ceph manager related detailed data. Data the ceph_mgr plugin will collect includes
> 
> ceph daemon mgr.id < ... >
> 

Is this literally `ceph daemon mgr.id` or is `id` variable, i.e. we'd actually be executing `ceph daemon mgr.1234` or somesuch?

> config diff config show dump_cache dump_mempools dump_osd_network mds_requests mds_sessions objecter_requests get_command_descriptions mds_requests mds_sessions perf dump perf histogram dump perf histogram schema perf schema status version
> 

Do any of these potentially contain sensitive info? Like passwords or private keys? Particularly interested in `config show`.

Are the `perf` commands particularly expensive to run? As in they have a long runtime or could cause a spike in memory usage? If they do, we'll want to gate them behind a plugin option likely.

> If the mgr is colocated with the mon (as is the usual case) and the keyring is present on the machine, additionally we might also want to collect,
> 

How do we detect this?

> ceph mgr dump ceph mgr metadata ceph mgr module ls ceph mgr services ceph mgr versions

nkshirsagar:
Hi Jake, 

> > This issue has been opened to enhance the ceph_mgr plugin to collect the ceph manager related detailed data. Data the ceph_mgr plugin will collect includes
> > ceph daemon mgr.id < ... >
> 
> Is this literally `ceph daemon mgr.id` or is `id` variable, i.e. we'd actually be executing `ceph daemon mgr.1234` or somesuch?

[...]

PR #2767, opened 2021-11-22
[ceph_mgr] Add more data collection to ceph_mgr

nkshirsagar:
Enhance the ceph_mgr plugin to collect more data specific to
mgr nodes. Also collect some additional manager data in the
ceph_mon plugin.

Related: #1945
Resolves: #2759

---

### DIFF
--- a/sos/report/plugins/ceph_mgr.py
+++ b/sos/report/plugins/ceph_mgr.py
@@ -37,18 +37,26 @@ class CephMGR(Plugin, RedHatPlugin, UbuntuPlugin):
         # more commands to be added later
         self.add_cmd_output([
             "ceph balancer status",
-            "ceph mgr metadata",
         ])
 
-        # more commands to be added later
         ceph_cmds = [
-            "mgr module ls",
-            "mgr dump",
+            "config diff",
+            "config show",
+            "dump_cache",
+            "dump_mempools",
+            "dump_osd_network",
+            "mds_requests",
+            "mds_sessions",
+            "objecter_requests",
+            "mds_requests",
+            "mds_sessions",
+            "perf dump",
+            "perf histogram dump",
+            "perf histogram schema",
+            "perf schema",
+            "status",
+            "version"
         ]
-
-        self.add_cmd_output([
-            "ceph %s --format json-pretty" % s for s in ceph_cmds
-        ], subdir="json_output", tags="insights_ceph_health_detail")
 
         self.add_forbidden_path([
             "/etc/ceph/*keyring*",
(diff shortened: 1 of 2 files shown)


sos/report/plugins/ceph_mgr.py
```diff
@@ -40,16 +40,26 @@ def setup(self):
             "ceph mgr metadata",
         ])
 
-        # more commands to be added later
-        ceph_cmds = [
-            "mgr module ls",
-            "mgr dump",
+        ceph_d_cmds = [
```
pponnuvel, line 42: `ceph_d_cmds`? Did you mean to say `ceph_cmds` or `ceph_mgr_cmds` or something like that?

sos/report/plugins/ceph_mgr.py
```diff
@@ -40,16 +40,26 @@ def setup(self):
...
+            "config diff",
+            "config show",
+            "dump_cache",
+            "dump_mempools",
+            "dump_osd_network",
+            "mds_requests",
+            "mds_sessions",
+            "objecter_requests",
+            "get_command_descriptions",
```
pponnuvel: Remove `get_command_descriptions`?

sos/report/plugins/ceph_mon.py
```diff
@@ -51,6 +51,11 @@ def setup(self):
             "ceph osd erasure-code-profile ls",
             "ceph osd crush show-tunables",
             "ceph osd crush dump"
+            "ceph mgr dump",
```
pponnuvel, line 54: Maybe add a note why these are here instead of ceph mgr.

sos/report/plugins/ceph_mgr.py
nkshirsagar: using ceph_d_cmds for ceph_daemon commands list.

sos/report/plugins/ceph_mgr.py
nkshirsagar: ack, done.

sos/report/plugins/ceph_mon.py
nkshirsagar: ack

sos/report/plugins/ceph_mgr.py
nkshirsagar: changed it the way you suggested on IRC - i.e,

                    containers_list[0][1], "ceph daemon %s %s" % (mgrid, cmd))
                    for mgrid in mgr_ids for cmd in ceph_daemon_cmds

sos/report/plugins/ceph_mon.py
```diff
@@ -50,7 +53,12 @@ def setup(self):
             "ceph osd metadata",
             "ceph osd erasure-code-profile ls",
             "ceph osd crush show-tunables",
-            "ceph osd crush dump"
+            "ceph osd crush dump",
+            "ceph mgr dump",
+            "ceph mgr metadata",
+            "ceph mgr module ls",
+            "ceph mgr services",
+            "ceph mgr versions"
```
TurboTurtle, line 61: This seems very strange to me to move `mgr` commands _out_ of the `ceph_mgr` plugin, even with the note provided. It is something I think users would certainly not expect at the least.

pponnuvel: @TurboTurtle It's really being taken "out" of ceph-mgr plugun - previously it was all part of the single ceph plugin where each cmd was every unit and and we had all the info w/ sosreports from all units. Now that the plugin has been split, only the relevant commands are run on given nodes. This is all fine.

But the `ceph mgr ...` commands require admin keyring which may not be present on mgr units. If so, we would lose out `ceph mgr ...` data.

I agree that this sounds odd to say `ceph mgr ...` commands may not work on a ceph-mgr node. But this is probably a historic as ceph-mgr was initially part of the ceph-mon daemon and why these commands don't work with just ceph-mgr keyring.

The bottom line is if we kept the `ceph mgr ...` commands in ceph-mgr plugin, these commands may not be a part of *any* sosreport unless ceph-mgr happens to have the admin keyring. So we might miss out those data completely.

TurboTurtle: Ack. In the end, the folks who use this data will, more likely than not, understand the apparent schism here.

TurboTurtle (approved):

pmoravec (approved):

PR #2770, opened 2021-11-23
[Plugin] Add tagging class for openEuler distro

marsCatXdu:
I'm planning to do more works on add support for openEuler distribution, including making some plugins and improve existing openEuler policy. Adding this tagging class for the first step.

openEuler has made some projects that not published on other distros yet, like iSulad, A-Tune, stratovirt , etc. So I think make a tagging might be necessary

Signed-off-by: lijingwei <lijingwei@uniontech.com>

---

### DIFF
--- a/sos/policies/distros/openeuler.py
+++ b/sos/policies/distros/openeuler.py
@@ -6,6 +6,7 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
+from sos.report.plugins import OpenEulerPlugin
 from sos.policies.distros.redhat import RedHatPolicy, OS_RELEASE
 import os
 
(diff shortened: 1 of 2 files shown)


marsCatXdu:
The check "packit/rpm-build-fedora-rawhide-s390x" have ran for more than 5 hours, is that right?

TurboTurtle:
You'll also want to add the new tagging class to `valid_subclasses` within the openEuler policy, e.g. like we do with Red Hat here:

https://github.com/sosreport/sos/blob/main/sos/policies/distros/redhat.py#L63

marsCatXdu:
> You'll also want to add the new tagging class to `valid_subclasses` within the openEuler policy, e.g. like we do with Red Hat here:
> 
> https://github.com/sosreport/sos/blob/main/sos/policies/distros/redhat.py#L63

Done.

TurboTurtle (approved):

PR #2774, opened 2021-12-01
[cleaner] Fix username casing and sourcing, simplify hostname casing

TurboTurtle:
First, simplify the case matching of hostnames to all upper case conventions for domains.

Second, make username matching similarly case insensitive like we did with hostnames. Additionally, fix and extend the sourcing of usernames to not skip the first entry in `last` by erroneously accounting for a header that is only in `lastlog`, and include `/etc/cron.{allow|deny}` as a source for usernames for an edge case where (ldap/ad) users are allowed or denied to modify crontab but have not actually logged in.

---

### DIFF
--- a/sos/cleaner/mappings/hostname_map.py
+++ b/sos/cleaner/mappings/hostname_map.py
@@ -169,16 +169,15 @@ class SoSHostnameMap(SoSMap):
 
     def sanitize_item(self, item):
         host = item.split('.')
-        if len(host) > 1 and all([h.isupper() for h in host]):
-            # by convention we have just a domain
-            _host = [h.lower() for h in host]
-            return self.sanitize_domain(_host).upper()
         if len(host) == 1:
             # we have a shortname for a host
             return self.sanitize_short_name(host[0].lower())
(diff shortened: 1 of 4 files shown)


sos/cleaner/mappings/hostname_map.py
```diff
@@ -194,7 +190,10 @@ def sanitize_item(self, item):
                 ob_hostname = 'unknown'
             ob_domain = self.sanitize_domain(domain)
             self.dataset[item] = ob_domain
-            return '.'.join([ob_hostname, ob_domain])
+            _fqdn = '.'.join([ob_hostname, ob_domain])
+            if all([h.isupper() for h in host]):
+                _fqdn = _fqdn.upper()
```
pmoravec, line 195: Shall not we apply this also for `len(host) == 2`? The current code does apply so, the proposed one doesn't.

TurboTurtle: Ack, updated and pushed.

pmoravec (approved):

PR #2779, opened 2021-12-03
[runtimes:crio][plugins:openshift_ovn] Add crio runtime and openshift_ovn plugin

npinaeva:
`openshift_ovn` plugin collects logs from crio containers
Fix `get_container_by_name` function returning container_id and not name

Tested with new `--container-runtime` option, works well.

---

### DIFF
--- a/sos/policies/runtimes/__init__.py
+++ b/sos/policies/runtimes/__init__.py
@@ -100,7 +100,7 @@ class ContainerRuntime():
             return None
         for c in self.containers:
             if re.match(name, c[1]):
-                return c[1]
+                return c[0]
         return None
 
     def get_images(self):
--- a/sos/policies/runtimes/crio.py
+++ b/sos/policies/runtimes/crio.py
@@ -1,0 +1,79 @@
+# Copyright (C) 2021 Red Hat, Inc., Nadia Pinaeva <npinaeva@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.policies.runtimes import ContainerRuntime
+from sos.utilities import sos_get_command_output
+from pipes import quote
+
+
+class CrioContainerRuntime(ContainerRuntime):
+    """Runtime class to use for systems running crio"""
+
+    name = 'crio'
+    binary = 'crictl'
+
+    def get_containers(self, get_all=False):
+        """Get a list of containers present on the system.
+
+        :param get_all: If set, include stopped containers as well
+        :type get_all: ``bool``
+        """
+        containers = []
+        _cmd = "%s ps %s" % (self.binary, '-a' if get_all else '')
+        if self.active:
+            out = sos_get_command_output(_cmd, chroot=self.policy.sysroot)
+            if out['status'] == 0:
+                for ent in out['output'].splitlines()[1:]:
+                    ent = ent.split()
+                    # takes the form (container_id, container_name)
+                    containers.append((ent[0], ent[-3]))
+        return containers
+
+    def get_images(self):
+        """Get a list of images present on the system
+
+        :returns: A list of 2-tuples containing (image_name, image_id)
+        :rtype: ``list``
+        """
+        images = []
+        if self.active:
+            out = sos_get_command_output("%s images" % self.binary,
+                                         chroot=self.policy.sysroot)
+            if out['status'] == 0:
+                for ent in out['output'].splitlines():
+                    ent = ent.split()
+                    # takes the form (image_name, image_id)
+                    images.append((ent[0] + ':' + ent[1], ent[2]))
+        return images
+
+    def fmt_container_cmd(self, container, cmd, quotecmd):
+        """Format a command to run inside a container using the runtime
+
+        :param container: The name or ID of the container in which to run
+        :type container: ``str``
+
+        :param cmd: The command to run inside `container`
+        :type cmd: ``str``
+
+        :param quotecmd: Whether the cmd should be quoted.
+        :type quotecmd: ``bool``
+
+        :returns: Formatted string to run `cmd` inside `container`
+        :rtype: ``str``
+        """
+        if quotecmd:
+            quoted_cmd = quote(cmd)
+        else:
+            quoted_cmd = cmd
+        container_id = self.get_container_by_name(container)
+        return "%s %s %s" % (self.run_cmd, container_id,
+                             quoted_cmd) if container_id is not None else ''
+
+# vim: set et ts=4 sw=4 :
--- a/sos/report/plugins/openshift_ovn.py
+++ b/sos/report/plugins/openshift_ovn.py
@@ -1,0 +1,41 @@
+# Copyright (C) 2021 Nadia Pinaeva <npinaeva@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, RedHatPlugin
+
+
+class OpenshiftOVN(Plugin, RedHatPlugin):
+    """This plugin is used to collect OCP 4.x OVN logs.
+    """
+    short_desc = 'Openshift OVN'
+    plugin_name = "openshift_ovn"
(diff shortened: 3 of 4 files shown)


sos/policies/runtimes/crio.py
```diff
@@ -0,0 +1,80 @@
...
+        """
+        containers = []
+        _cmd = "%s ps %s" % (self.binary, '-a' if get_all else '')
+        if self.active:
+            out = sos_get_command_output(_cmd, chroot=self.policy.sysroot)
+            if out['status'] == 0:
+                for ent in out['output'].splitlines()[1:]:
+                    ent = ent.split()
+                    # takes the form (container_id, container_name)
+                    containers.append((ent[0], ent[-3]))
```
pmoravec, line 36: As the `-3` is the only difference from the parent original method: does not it make sense to have the index as a class attribute like `binary`, defaulting to `-1` and overwritten in `CrioContainerRuntime` to `-3`?

sos/policies/runtimes/crio.py
```diff
@@ -0,0 +1,80 @@
...
+        """
+        images = []
+        if self.active:
+            out = sos_get_command_output("%s images" % self.binary,
+                                         chroot=self.policy.sysroot)
+            if out['status'] == 0:
+                for ent in out['output'].splitlines():
+                    ent = ent.split()
+                    # takes the form (image_name, image_id)
+                    images.append((ent[0] + ':' + ent[1], ent[2]))
```
pmoravec, line 53: Here we rely on `len(ent) > 2`. Is it a safe assumption in either use case? (I don't know `crio` enough, hence asking)

sos/report/plugins/openshift_ovn.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    short_desc = 'Openshift OVN'
+    plugin_name = "openshift_ovn"
+    containers = ('ovnkube-master', 'ovn-ipsec')
+    profiles = ('openshift',)
+
+    def setup(self):
+        self.add_copy_spec([
+            "/var/lib/ovn/etc/ovnnb_db.db",
+            "/var/lib/ovn/etc/ovnsb_db.db",
```
pmoravec, line 25: These is already collected by `ovn_central` plugin (https://github.com/sosreport/sos/blob/main/sos/report/plugins/ovn_central.py#L135-L141). We should collect any file by no more than 1 plugin.

sos/policies/runtimes/__init__.py
```diff
@@ -100,7 +100,7 @@ def get_container_by_name(self, name):
             return None
         for c in self.containers:
             if re.match(name, c[1]):
-                return c[1]
+                return c[0]
```
pmoravec, line 103: Good catch!

sos/policies/runtimes/crio.py
```diff
@@ -0,0 +1,80 @@
...
+        :rtype: ``str``
+        """
+        if quotecmd:
+            quoted_cmd = quote(cmd)
+        else:
+            quoted_cmd = cmd
+        container_id = self.get_container_by_name(container)
+        return "%s %s %s" % (
+            self.run_cmd, container_id
+            if container_id is not None else container, quoted_cmd)
```
TurboTurtle, line 78: Why are we getting the ID here? Does crio not support the use of container names for exec?

sos/policies/runtimes/crio.py
TurboTurtle: Agreed. I think having these indexes as a part of the class would simplify things.

sos/report/plugins/openshift_ovn.py
```diff
@@ -0,0 +1,53 @@
...
+                    'ovnkube-master',
+                    'ovn-appctl -t /var/run/ovn/ovnnb_db.ctl ' +
+                    'cluster/status OVN_Northbound'),
+                self.fmt_container_cmd(
+                    'ovnkube-master',
+                    'ovn-appctl -t /var/run/ovn/ovnsb_db.ctl ' +
+                    'cluster/status OVN_Southbound')
+            ])
+        if self.container_exists('ovn-ipsec'):
+            self.add_cmd_output([
```
TurboTurtle, line 43: Ack, however let me see about adding a `container` parameter to `add_cmd_output()` so this reads a little easier. We could then simply do things like:

```python
self.add_cmd_output([
   'ipsec status',
   'ovs-appctl [...]'
], container='ovn'ipsec')
```

TurboTurtle (commented):
Looks mostly good to me, just need a few clarifications.

sos/policies/runtimes/crio.py
npinaeva: `crio images` has the following columns in the output
`IMAGE                                                     TAG                 IMAGE ID            SIZE`
So it's the same as parsing output of any other command

sos/policies/runtimes/crio.py
npinaeva: I'm afraid it doesn't
```
USAGE:
   crictl exec [command options] CONTAINER-ID COMMAND [ARG...]
```
I'd like to return some kind of error here if container id is not found (command won't work anyway), but I found that general function https://github.com/sosreport/sos/blob/main/sos/report/plugins/__init__.py#L2423 just returns command if container is not found and followed the same rule here
Maybe you can help me figuring out what is the best option for error here?

sos/report/plugins/openshift_ovn.py
npinaeva: Yeah, so the difference is that this plugin is used to collect ovn logs from container, and these pathes are actually just where container mounts its directories.
I'd prefer to have `add_copy_spec_container` to copy files directly from containers, so that changing host mount directory doesn't break everything, but with `crictl` it's not possible to copy files from containers https://github.com/kubernetes-sigs/cri-tools/issues/677, and using `kubectl cp` also have limitations.

About collecting files only once, we have this problem about using the same piece of software but in different "configurations" for different systems. Files I'm trying to collect are openshift-specific (and plugin is named accordingly) meaning that code that configures ovn and defines where required files will be located is a part of OpenShift code.
It also means that these plugins for ovn are mutually exclusive, so these same files will never be collected twice (if system configuration makes any sense ofc).

There is also an example of collecting the same files in different plugins: `ovn_host` and `ovn_central` copying the same logs: https://github.com/sosreport/sos/blob/main/sos/report/plugins/ovn_host.py#L33-L36 and https://github.com/sosreport/sos/blob/main/sos/report/plugins/ovn_central.py#L82-L85, but I think these plugins are also never supposed to be active on the same host

sos/report/plugins/openshift_ovn.py
npinaeva: Do you want to have this change before this PR is merged, so that I can rebase?

sos/policies/runtimes/crio.py
npinaeva: I'm not sure about how many other runtimes we'll have so that this option is used more than once and how general this suggestion that all runtimes `ps` output is only different in columns order (because that's the only when this option will work), but if you think it's reasonable, let's do that.

sos/report/plugins/openshift_ovn.py
TurboTurtle: Ack. Just merged the PR for this.

sos/policies/runtimes/crio.py
TurboTurtle: For now, return an empty string. The changes (just merged) to `add_cmd_output()` will log an error.

Alternatively, if you want we could enclose the call to `fmt_container_cmd()` in `add_cmd_output()` in a `try` block, and you could raise an exception here in the crio runtime, to get catch by that block.

I'm fine either way.

sos/policies/runtimes/crio.py
TurboTurtle: Fair point. At the moment while it is duplication, it is small and confined and I don't see future expansion without further adding new runtimes, at which point we can re-evaluate this override or the suggestion above.

sos/policies/runtimes/crio.py
TurboTurtle: Is there a condition where `crio images` reports an error message, but still gives a `0` exit code? I don't believe we've had this with other runtime executables, but we have seen this in other collections in general.

npinaeva: I definitely can't guarantee that, but I also have no other means to verify its output, even if I check for `ent` len, in case of error it'll probably have more than 3 words, so I'll just return the wrong `images` list

TurboTurtle (approved):

sos/policies/runtimes/crio.py
pmoravec: ACK

sos/report/plugins/openshift_ovn.py
pmoravec: OK

pmoravec (approved):

PR #2783, opened 2021-12-04
[dellrac] Dell Remote Access Controller Administration

vikas-goel:

---


sos/report/plugins/dellrac.py
```diff
@@ -0,0 +1,63 @@
...
+                '%s %s' % (self.racadm, subcmd),
+                suggest_filename='racadm_%s' % subcmd)
+
+        if self.get_option("debug"):
+            self.do_debug()
+
+    def do_debug(self):
+        self.add_cmd_output(
+            '%s supportassist accepteula',
+            suggest_filename='racadm_supportassist_accepteula')
```
TurboTurtle, line 40: Does this actually accept the eula or does it only report on if it's been accepted?

If it's the former - absolutely no way.

sos/report/plugins/dellrac.py
```diff
@@ -0,0 +1,63 @@
...
+        if status['status'] == 0:
+            jid = re.search(r'JID_[0-9]*', status['output']).group()
+
+        done_collect = False
+        while not done_collect and jid:
+            time.sleep(10)
+            status = self.collect_cmd_output(
+                '%s jobqueue view -i %s' % (self.racadm, jid),
+                suggest_filename='racadm_jobqueue_view')
+            done_collect = re.search(r'Status=Completed', status['output'])
```
TurboTurtle, line 57: What are we waiting for to be completed here? Does the previous command automatically background itself? Is there someway to not return until that command is collected? I'd prefer that method as that allows sos to leverage its timeout controls.

sos/report/plugins/dellrac.py
```diff
@@ -0,0 +1,63 @@
...
+
+        self.add_cmd_output(
+            '%s supportassist collect -f %s/sar.zip' % (self.racadm, logpath),
```
TurboTurtle, line 60: What exactly is the difference between this and the `supportassist collect` command from above? At first glance it seems the first one is writing output to stdout, which we'll collect anyways, and the second is writing straight to disk?

TurboTurtle (changes requested):
Need some clarification on this plugin. Additionally, the commit message needs to be filled out so that reading the git history gives the proper context.

Usually for new plugins, the format is:

```
[plugin_name] Add new plugin

<...descriptive text of the plugin and a short explanation
    of what the use case is...>
```

sos/report/plugins/dellrac.py
vikas-goel: It does accept EULA if not already done. The `supportassist` sub command does not work if not already accepted.

sos/report/plugins/dellrac.py
vikas-goel: Yes, the previous command goes in background. There is no option to run that in foreground. The only way is to keep checking job id status.

sos/report/plugins/dellrac.py
vikas-goel: The first command starts collection of required data. This command with `-f` option creates zip of the data collected earlier. It's all writing to disk.

TurboTurtle: So there seems to be some sort of caching occurring in order for subsequent executions to be able to compress earlier generated data. Does the `-f` form not generate the data if it's not already present?

sos/report/plugins/dellrac.py
TurboTurtle: 100% not going to happen in sos. We go through great lengths to not change system state, but we _absolutely_ cannot accept legal agreements on the behalf of users, not even if this were gated behind a plugin option.

sos/report/plugins/dellrac.py
vikas-goel: No, it doesn’t as far as I know.

sos/report/plugins/dellrac.py
vikas-goel: Sure. There is a eula status command. We can use that prior to collecting report. Okay?

bmr-cymru: 100% agree with @TurboTurtle - if users have to accept license terms in order to use functionality then that is something that we cannot automate. It's fine to check the status and log a warning message if that's possible, but we cannot perform an action that would agree to license terms on a user's behalf.

sos/report/plugins/dellrac.py
```diff
@@ -0,0 +1,63 @@
...
+        PluginOpt('debug', default=False, desc='capture support assist data')
+    ]
+
+    racadm = '/opt/dell/srvadmin/bin/idracadm7'
+
+    def setup(self):
+        for subcmd in ['getniccfg', 'getsysinfo']:
+            self.add_cmd_output(
+                '%s %s' % (self.racadm, subcmd),
+                suggest_filename='racadm_%s' % subcmd)
```
TurboTurtle, line 32: Since the binary is ultimately `idracadm7`, we should use that in the filename. Otherwise users could be confused and think they are in fact running `racadm` here.

vikas-goel: The Dell documents refer to `racadm` command. There are other versions idracadm8 and 9 for other platforms where the RPM creates a symlink racadm. If and when the plugin supports those platforms or newer RPMs, then we want to make sure we be consistent with the filename. So, I would like to keep the output filename agnostic of the version.

sos/report/plugins/dellrac.py
vikas-goel: Yup

sos/report/plugins/dellrac.py
TurboTurtle: If there are symlinks dropped, then use call the symlink. If not, then we shouldn't be conflating the command that is run. If I see `racadm_getniccfg` then I assume the command that generated the output was exactly `racadm getniccfg`, not `idracadm7 getniccfg` or `idracadm8 getniccfg`. If I need to run one of the latter commands, then looking at `sos_commands/dellrac/racadm_*` does not track.

vikas-goel: The Dell guide refers to `racadm` everywhere. There are different versions of the RPM with the same usage and purpose. The (manual/automated) consumer of the data should not bother about whether the command was symlink, racadm, or versioned. There needs to be consistency in the output filename so that regardless of how the data was generated, the consumers are not impacted now and in future. If it helps, the plugin can be renamed to `racadm`. But I think having vendor name in the plugin name is apt.

TurboTurtle: If the Dell guide refers to `racadm` - then this should be calling `racadm` if you want `racadm_$cmd` in the plugin directory. If there are different versions of the command out there then that is valuable information to pass on to whomever is reviewing the collected data. 

`suggest_filename` is not meant to obfuscate what is being run. It's used to make unruly or human-unfriendly command strings more readable.

Either actually call `racadm` here, or don't override the binary string in the resulting filename.

vikas-goel: The command will change to racadm with v8 and v9 support. The usage and output will remain same. Can we live with this until then?

TurboTurtle: No.

As I've been saying we do not use `suggest_filename` to hide what we're executing from users. If we're running `idracadm7`, then the file here should be called `idracadm7_$cmd`. If we're running `racadm`, then at that point the file should be called `racadm_$cmd`.

What would be acceptable is to have a static _tag_ for this collection. You can add tags via the `tags` parameter (accepts `str` or `list` values), and those tags will appear in the manifest for this/these collections.

vikas-goel: Updated

pmoravec (approved):

TurboTurtle (approved):

PR #2794, opened 2021-12-08
Foreman new vhosts

taftsanders:
---

### DIFF
--- a/sos/report/plugins/foreman_proxy.py
+++ b/sos/report/plugins/foreman_proxy.py
@@ -38,6 +38,10 @@ class ForemanProxy(Plugin):
                 self.apachepkg),
             "/var/log/{}*/katello-reverse-proxy_error_ssl.log*".format(
                 self.apachepkg),
+            "/var/log/{}*/rhsm-pulpcore-https-*_access_ssl.log*".format(
+                self.apachepkg),
+            "/var/log/{}*/rhsm-pulpcore-https-*_error_ssl.log*".format(
+                self.apachepkg),
         ])
 
         # collect http[|s]_proxy env.variables


wbclark (approved):
This looks good to me

See https://github.com/theforeman/puppet-foreman_proxy_content/commit/76e2a6852d1d2ca33935ccf8a6ab69992c32ec1d for more context

sos/report/plugins/foreman.py
```diff
@@ -108,10 +108,6 @@ def setup(self):
             "/var/log/foreman-proxy-certs-generate*",
             "/usr/share/foreman/Gemfile*",
             "/var/log/{}*/foreman*".format(self.apachepkg),
-            "/var/log/{}*/katello-reverse-proxy_access_ssl.log*".format(
-                self.apachepkg),
-            "/var/log/{}*/katello-reverse-proxy_error_ssl.log*".format(
-                self.apachepkg),
```
pmoravec, line 114: I would leave collecting the logs for a while due to legacy reasons. A new `sos` can run also on older foreman/Satellite/Capsule that still logs there.

E.g. checking (in downstream) current Capsule 6.10, it does log there. But it runs on RHEL7 where we wont probably backport this change. But.. we would have to be aware of "no backport" requirement, and for `sos` it is much easier to let collect potentially non-existing files (which is handled well) than risking we forget to collect something.

sos/report/plugins/foreman_proxy.py
```diff
@@ -38,6 +38,14 @@ def setup(self):
...
                 self.apachepkg),
+            "/var/log/{}*/rhsm-pulpcore-https-443_access_ssl.log*".format(
+                self.apachepkg),
+            "/var/log/{}*/rhsm-pulpcore-https-443_error_ssl.log*".format(
+                self.apachepkg),
+            "/var/log/{}*/rhsm-pulpcore-https-8443_access_ssl.log*".format(
+                self.apachepkg),
+            "/var/log/{}*/rhsm-pulpcore-https-8443_error_ssl.log*".format(
+                self.apachepkg),
         ])
```
pmoravec, line 49: ACK to adding this. Maybe we can shrink it to `rhsm-pulpcore-https-*_access_ssl.log*` and `rhsm-pulpcore-https-*_error_ssl.log*` ? Just in case of port change?

ehelms: I wonder why be this specific? Foreman always controls httpd when present so it should be safe to grab everything under `/var/log/{}*/*` and reduce the churn you need whenever we make changes. I'd also want to see what @ekohl thinks

ekohl: Honestly, I'd expect such a thing in the apache plugin. That this is in the foreman_proxy pains me a lot since it has nothing to do with a Foreman Proxy. It just happens that on some systems there is also an Apache deployed which has some config that relates to it.

TurboTurtle: We've explicitly moved these kinds of collections out of `apache` because of historical issues with multiple different components logging there, and the `apache` plugin either becoming unruly in its maintenance or insufficient in its handling of component-specific logs.

I do agree however, let's replace `8443` with a `*` glob so we don't miss these on non-standard configurations.

pmoravec (approved):

TurboTurtle (approved):

PR #2799, opened 2021-12-14
[utilities] WIP do we really need glob expansion for commands?

BryanQuigley:
I think being more precise might be important.

Checking next... (not complete list)
sos/report/plugins/apport.py:        self.add_cmd_output("bash -c 'grep -B 50 -m 1 ProcMaps /var/crash/*'")
sos/report/plugins/candlepin.py:        self.add_cmd_output("du -sh /var/lib/candlepin/*/*")
sos/report/plugins/ds.py:        self.add_cmd_output("ls -l /var/lib/dirsrv/slapd-*/db/*")
sos/report/plugins/foreman.py:        self.add_cmd_output('systemctl list-units dynflow*',
sos/report/plugins/krb5.py:        self.add_cmd_output("klist -ket %s/.k5*" % self.kdcdir)
sos/report/plugins/ldap.py:        self.add_cmd_output("ldapsearch -x -b '' -s base 'objectclass=*'")
sos/report/plugins/md.py:        self.add_cmd_output("mdadm -D /dev/md*")
sos/report/plugins/mysql.py:        self.add_cmd_output("du -s /var/lib/mysql/*")
sos/report/plugins/mysql.py:        self.add_cmd_output("du -s /var/lib/percona-xtradb-cluster/*")
sos/report/plugins/ipa.py:            "ls -la /etc/dirsrv/slapd-*/schema/",
sos/report/plugins/juju.py:            "ls -alRh /var/log/juju*",
sos/report/plugins/juju.py:            "ls -alRh /var/lib/juju*"
sos/report/plugins/smartcard.py:            "ls -nl /usr/lib*/pam_pkcs11/",

---

### DIFF
--- a/sos/report/plugins/pam.py
+++ b/sos/report/plugins/pam.py
@@ -37,7 +37,7 @@ class Pam(Plugin):
 
 
 class RedHatPam(Pam, RedHatPlugin):
-    security_libs = "/lib*/security"
+    security_libs = "/lib64/security"
 
     def setup(self):
         super(RedHatPam, self).setup()
--- a/sos/utilities.py
+++ b/sos/utilities.py
@@ -143,16 +143,8 @@ def sos_get_command_output(command, timeout=TIMEOUT_DEFAULT, stderr=False,
         )
 
     args = shlex.split(command)
-    # Expand arguments that are wildcard paths.
-    expanded_args = []
-    for arg in args:
-        expanded_arg = glob.glob(arg)
-        if expanded_arg:
-            expanded_args.extend(expanded_arg)
-        else:
-            expanded_args.append(arg)
     try:
-        p = Popen(expanded_args, shell=False, stdout=PIPE,
+        p = Popen(args, shell=False, stdout=PIPE,
                   stderr=STDOUT if stderr else PIPE,
                   bufsize=-1, env=cmd_env, close_fds=True,
                   preexec_fn=_child_prep_fn)


TurboTurtle:
Is there a problem being caused by this? I think plugin authors would be more surprised that globs aren't supported here.

BryanQuigley:
I noticed it because I was trying to figure out why in an strace sos was looking for a bunch of files called 30s. (I was also looking into if the timeout command could be replaced by some python..)

At least a few of the places it's used - it isn't needed, or also uses glob to do the same thing (ipa.py) - which specifying a specific directory for glob seems better anyway, something like:
for dev in glob(self.path_join('/dev/watchdog*')):
    self.add_cmd_output("wdctl %s" % dev)

I also know that shell expansion has been shown to create security issues in the past. I'm more worried about commands where they are not expecting the option to be a valid directory/filename/etc.

Thoughts?

TurboTurtle:
> I noticed it because I was trying to figure out why in an strace sos was looking for a bunch of files called 30s. (I was also looking into if the timeout command could be replaced by some python..)
> 
Would be great if we could replace wrapping via `timeout` with something python native, for sure.

> At least a few of the places it's used - it isn't needed, or also uses glob to do the same thing (ipa.py) - which specifying a specific directory for glob seems better anyway, something like: for dev in glob(self.path_join('/dev/watchdog*')): self.add_cmd_output("wdctl %s" % dev)
> 

Definitely a valid point there with regards to a better approach in plugin design.

However, this glob expansion (from https://github.com/sosreport/sos/commit/04d412816de2e4053f43956adf536272a2ddc19c), and was (re-)added to sos as a direct result of removing full shell-style glob support for commands which broke certain collections like `ls /some/foo*/path`.

> I also know that shell expansion has been shown to create security issues in the past. I'm more worried about commands where they are not expecting the option to be a valid directory/filename/etc.
> 

Ack, following the commit above we can see that security concerns were the crux of why we originally removed full shell expansion support in the first place.

At the moment, I'd be in favor of refining existing collections and updating our plugin guidelines to suggest not putting globs in the actual commands whenever possible, but I'm not sure if outright removing the expansion is the right play.

BryanQuigley:
How about limiting in some way like I tried in #2803 ?

Issue #2800, opened 2021-12-15
Collect mds specific information in the ceph_mds plugin


nkshirsagar:
#1945 split the ceph plugin into individual component plugins. One of the reasons this was done was so we could add component specific data collection for each individual ceph service.

This issue has been opened to enhance the ceph_mds plugin to collect the ceph mds related detailed data. Data the ceph_mds plugin will collect includes,

ceph daemon mds.id < ... > 

            "cache status",
            "client ls",
            "config diff",
            "config show",
            "damage ls",
            "dirfrag ls",
            "dump cache",
            "dump inode",
            "dump loads",
            "dump snaps",
            "dump tree",
            "dump_blocked_ops",
            "dump_historic_ops",
            "dump_historic_ops_by_duration",
            "dump_mempools",
            "dump_ops_in_flight",
            "get subtrees",
            "log dump",
            "objecter_requests",
            "openfiles ls",
            "ops",
            "perf histogram dump",
            "perf histogram schema",
            "perf schema",
            "perf dump",
            "status",
            "version",
            "session ls"

the ID will be parsed out from the ps -u ceph -o args command, (like already done in the ceph_mgr plugin, for the mgr id), as shown in the below example where z-rotomvm17 is the mds ID.

~~~
$ps -u ceph -o args
COMMAND
/usr/bin/ceph-mds -f --cluster ceph --id z-rotomvm17 --setuser ceph --setgroup ceph
~~~

PR #2801, opened 2021-12-15
[ceph_mds] Collect mds specific data in ceph_mds

nkshirsagar:
Enhance the ceph_mds plugin to collect data specific
to mds nodes.

Related: #1945
Resolves: #2800

---

### DIFF
--- a/sos/report/plugins/ceph_mds.py
+++ b/sos/report/plugins/ceph_mds.py
@@ -25,7 +25,6 @@ class CephMDS(Plugin, RedHatPlugin, UbuntuPlugin):
         })
 
         self.add_copy_spec([
-            "/var/log/ceph.log",
             "/var/log/ceph/ceph-mds*.log",
             "/var/lib/ceph/bootstrap-mds/",
             "/var/lib/ceph/mds/",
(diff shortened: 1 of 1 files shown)


sos/report/plugins/ceph_mds.py
```diff
@@ -45,4 +45,60 @@ def setup(self):
...
+        # If containerized, run commands in containers
+        try:
+            cname = self.get_all_containers_by_regex("ceph-mds*")[0][1]
+        except Exception:
+            cname = None
+
+        self.add_cmd_output([
+            "ceph daemon %s %s"
+            % (mdsid, cmd) for mdsid in mds_ids for cmd in ceph_cmds
+        ], container=cname)
```
TurboTurtle, line 101: This looks fine by itself, however how many mds ids would a "normal" environment potentially have here?

If there's a decent number of them, we may want to put these collections in an id-specific subdir within the plugin directory, for better organization.

TurboTurtle (commented):
Overall this looks fine, just a couple minor comments.

sos/report/plugins/ceph_mds.py
```diff
@@ -25,7 +25,7 @@ def setup(self):
         })
 
         self.add_copy_spec([
-            "/var/log/ceph.log",
+            # "/var/log/ceph.log",
```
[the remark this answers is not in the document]
nkshirsagar: ack

sos/report/plugins/ceph_mds.py
nkshirsagar: I would suppose there would not be more than one, in most cases. I will check with @hillpd to confirm.

taodd:
@nkshirsagar Nikhil, 

There are some cephfs admin socket commands that can not be collected 
E.g dump_cache
Many fs production cluster has millions or 10's millions of inode cached in the memory
The impact would be:
1. The dump will take a very long time, file output file will be very large
2. could slow down the mds significantly 

There are also some command require an extra parameter from user
like dump_inode and dirfrag ls ..

Most options looks good to me
My suggestion is to remove below:
dump_cache
dump_inode
dirfrag ls
dump snaps
log dump
openfiles ls [this doesn't work as expected any more, would be misleading, though I added :)]

nkshirsagar:
> 

I've made the changes you suggested. Thanks!

taodd (approved):

sos/report/plugins/ceph_mds.py
nkshirsagar: @hillpd would there ever be multiple mds ids on a single node?

hillpd: Normal environments will not have more than one MDS on a single node.

TurboTurtle (approved):

pmoravec (approved):

Issue #2777, opened 2021-12-02
Sos collect errors for OCP cluster


npinaeva:
I tried to run the latest sos collect for OCP cluster from my local machine using
`bin/sos collect --no-local --batch -vvv --cluster-type ocp` command. I found a couple of errors and even fixed on my way:

1. Cleanup problem: when sos collect fails e.g. with `No nodes connected. Aborting...`, cluster cleanup isn't called and the next run you get this error:
```
[ux331un:run_command] Running command oc new-project sos-collect-tmp
[ocp] Failed to create project: Error from server (AlreadyExists): project.project.openshift.io "sos-collect-tmp" already exists

Traceback (most recent call last):
  File "bin/sos", line 22, in <module>
    sos.execute()
  File "/home/nadia/ocp/sos/sos/__init__.py", line 186, in execute
    self._component.execute()
  File "/home/nadia/ocp/sos/sos/collector/__init__.py", line 1100, in execute
    self.prep()
  File "/home/nadia/ocp/sos/sos/collector/__init__.py", line 812, in prep
    self.cluster.setup()
  File "/home/nadia/ocp/sos/sos/collector/clusters/ocp.py", line 83, in setup
    raise Exception("Failed to create temporary project for collection. "
Exception: Failed to create temporary project for collection. 
Aborting...
```
I solved this by adding `self.cluster.cleanup()` here https://github.com/sosreport/sos/blob/main/sos/collector/__init__.py#L444 , but in general, sos collect doesn't seem to have a reliable way to cleanup, at least because there are multiple different exit points in the same function calling different shutdown functions, e.g. https://github.com/sosreport/sos/blob/main/sos/collector/__init__.py#L1140 and https://github.com/sosreport/sos/blob/main/sos/collector/__init__.py#L1164-L1166.

2. Nodes are listed wrong for ocp
Tried to fix that: https://github.com/sosreport/sos/pull/2776

3. `The command was not found or was not executable: oc`
`oc` is executable in the beginning but then sometimes it's getting lost. 
- during connecting to nodes `[ip-10-0-213-88.eu-central-1.compute.internal:read_file] Exception while reading /etc/os-release: The command was not found or was not executable: oc`
- during cleanup: 
```
No nodes connected. Aborting...
[ux331un:run_command] Running command oc delete project sos-collect-tmp
EXC: The command was not found or was not executable: oc.
```

[...]

PR #2804, opened 2021-12-17
[transport] Detect retrieval failures and automatically retry

TurboTurtle:
If a paritcular attempt to retrieve a remote file fails, we should
automatically retry that collection up to a certain point. This provides
`sos collect` more resiliency for the collection of sos report archives.

This change necessitates a change in how we handle the SoSNode flow for
failed sos report retrievals, and as such contains minor fixes to
transports to ensure that we do not incorrectly hit exceptions in error
handling that were not previously possible with how we exited the
SoSNode retrieval flow.

Closes: #2777

---

### DIFF
--- a/sos/collector/transports/__init__.py
+++ b/sos/collector/transports/__init__.py
@@ -303,7 +303,20 @@ class RemoteTransport():
         :returns:   True if file was successfully copied from remote, or False
         :rtype:     ``bool``
         """
-        return self._retrieve_file(fname, dest)
+        attempts = 0
+        try:
+            while attempts < 5:
+                attempts += 1
+                ret = self._retrieve_file(fname, dest)
+                if ret:
+                    return True
+                self.log_info("File retrieval attempt %s failed" % attempts)
+            self.log_info("File retrieval failed after 5 attempts")
+            return False
+        except Exception as err:
+            self.log_error("Exception encountered during retrieval attempt %s "
+                           "for %s: %s" % (attempts, fname, err))
+            raise err
 
     def _retrieve_file(self, fname, dest):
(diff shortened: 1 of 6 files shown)


sos/collector/transports/__init__.py
```diff
@@ -303,7 +303,20 @@ def retrieve_file(self, fname, dest):
...
-        return self._retrieve_file(fname, dest)
+        attempts = 0
+        try:
+            while attempts < 5:
+                attempts += 1
+                ret = self._retrieve_file(fname, dest)
+                if ret:
+                    return True
+                self.log_info("File retrieval attempt %s failed" % attempts)
+            self.log_info("File retrieval failed after 5 attempts")
```
npinaeva, line 314: maybe set it to `log_error` so that it's visible that file wasn't downloaded correctly?

TurboTurtle: Will be visible via https://github.com/sosreport/sos/pull/2804/files#diff-c64b04e837e3b06a12632c6b923d50b18f88884c482c71973d43a6738b49f3dcR807

npinaeva: Ah, right, I thought at first that `ui_msg` has lower visibility than `log_err`
But it doesn't really say ERROR, so log message looks exactly like "Retrieving sos report..." for example, maybe add something like "ERROR:" to the message itself?

npinaeva:
1-2 retries is usually enough to download archives correctly, only 1 of ~15 runs hit 5 retries and failed. Conclusion - it works.
Thank you for this feature!

pmoravec (approved):
ACK.

Maybe worth having the number of retrires as a constant, but rather it would be an overkill.

PR #2809, opened 2021-12-22
[systemd] Collect systemctl list-units --all

kolyshkin:
Systemd has a hard limit of units that it can keep, and the limit is 128K.

Once that limit is hit (which is usually happening due to various bugs),
systemd returns E2BIG ("Argument list too long") when trying to start a
new unit. Until recently, it was not possible to figure out what is
going on without looking at the "systemctl list-units --all" output.

That's why it's nice to have it in sos report.

Found while working on https://bugzilla.redhat.com/show_bug.cgi?id=2028153

Signed-off-by: Kir Kolyshkin <kolyshkin@gmail.com>

---

### DIFF
--- a/sos/report/plugins/systemd.py
+++ b/sos/report/plugins/systemd.py
@@ -39,6 +39,7 @@ class Systemd(Plugin, IndependentPlugin):
             # status --all mostly seems to cover the others.
             "systemctl list-units",
             "systemctl list-units --failed",
+            "systemctl list-units --all",
             "systemctl list-unit-files",
             "systemctl list-jobs",
             "systemctl list-dependencies",


sos/report/plugins/systemd.py
```diff
@@ -39,6 +39,7 @@ def setup(self):
             # status --all mostly seems to cover the others.
             "systemctl list-units",
             "systemctl list-units --failed",
+            "systemctl list-units --all",
             "systemctl list-unit-files",
```
pmoravec, line 43: A question from `systemctl` noob: what *additional* information the `list-units --all` provide, that is not covered in the three other command outputs? (we try to limit redundant duplicate data collection, if possible)

TurboTurtle (approved):

sos/report/plugins/systemd.py
```diff
@@ -39,6 +39,7 @@ def setup(self):
             # status --all mostly seems to cover the others.
             "systemctl list-units",
             "systemctl list-units --failed",
+            "systemctl list-units --all",
```
pmoravec, line 42: I am OK to add the `--all`. Just thinking about possible duplication data collection: is there some reason to collect similar info three times (`list-units`, `--failed` and `--all`)? Are there unique pieces of information in each of the command outputs?

pmoravec: Disregard, somehow answered already.

pmoravec (approved):

PR #2810, opened 2021-12-22
[crio] Collect goroutine information

haircommander:
According to instructions in https://github.com/cri-o/cri-o/blob/main/tutorials/debugging.md

Signed-off-by: Peter Hunt <pehunt@redhat.com>

---

### DIFF
--- a/sos/report/plugins/crio.py
+++ b/sos/report/plugins/crio.py
@@ -78,6 +78,8 @@ class CRIO(Plugin, RedHatPlugin, UbuntuPlugin):
         images = self._get_crio_list(img_cmd)
         pods = self._get_crio_list(pod_cmd)
 
+        self._get_crio_goroutine_stacks()
+
         for container in containers:
             self.add_cmd_output("crictl inspect %s" % container,
                                 subdir="containers")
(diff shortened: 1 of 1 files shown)


sos/report/plugins/crio.py
```diff
@@ -50,6 +50,12 @@ def setup(self):
             "crio config"
         ])
 
+        result = self.exec_cmd("kill -USR1 $(pidof crio)")
+        if result['status'] == 0:
```
pmoravec, line 54: Are you sure the `$(..)` will be properly evaluated? As when I test it on a mocked `crio` command, I always get `result` as:

    {'status': 1, 'output': 'kill: cannot find process "$(pidof"\nkill: cannot find process "crio)"\n', 'truncated': False}

TurboTurtle: Nack to bash-foo in plugins. Get the pid first, and if one exists, make another call to `kill`.

sos/report/plugins/crio.py
```diff
@@ -50,6 +50,12 @@ def setup(self):
...
+            self.add_copy_spec([
+                "/tmp/crio-goroutine-stacks*.log"
+            ])
```
TurboTurtle, line 57: `add_copy_spec()` supports single strings, so no need for a list here. Glob expansion is done later on.

sos/report/plugins/crio.py
haircommander: makes sense, updated

haircommander:
updated for comments. The test failures don't look related, are they?

TurboTurtle (approved):
LGTM.

Test failure is unrelated and a known issue - restarted the job.

pmoravec (approved):

PR #2811, opened 2021-12-28
[cleaner] detect longer known domain names properly

pmoravec:
For domains consisting of >2 names (foo.bar.com or longer), detection
of them in input data does not currently work. Since we grep for a
string inside a list (split by '.') instead of the domainname (separated
from its top-level domain).

Resolves: #2811

---

### DIFF
--- a/sos/cleaner/mappings/hostname_map.py
+++ b/sos/cleaner/mappings/hostname_map.py
@@ -106,7 +106,6 @@ class SoSHostnameMap(SoSMap):
             # don't block on host's shortname
             return host[0] in self.hosts.keys()
         else:
-            domain = host[0:-1]
             for known_domain in self._domains:
                 if known_domain in domain:
                     return True


pmoravec:
Reproducer:

    mv /etc/sos/cleaner/default_mapping{,.prev}
    mkdir /var/log/anaconda
    echo "$(date) $(hostname -d) some_test_log" > /var/log/anaconda/some_test_log.log
    echo "$(date) some.example.domain.com some_test_log" >> /var/log/anaconda/some_test_log.log
    python3 bin/sos report --build --batch --clean --domains some.example.domain.com,$(hostname -d) -o anaconda

And check if collected `/var/log/anaconda/some_test_log.log` file has the domains obfuscated.

pmoravec:
To the repeatedly-failing `tests/cleaner_tests/full_report_run.py:FullCleanTest.test_private_map_was_generated` test case on F34: I can not reproduce it on F34, so far.

I will give it few more attempts but if I won't succeed, I would blame the "known timeout false positive issue with an unknown reason".

pmoravec:
Rebased the local branch to current `main` branch with the "Do not remove temp dirs" commit inside - let see where/why the CI build on F34 constantly fails.

pmoravec:
Discarding in favour of #2819 .

PR #2812, opened 2021-12-30
[system] Collect glibc tuning decisions

pmoravec:
Resolves: #2812

---

### DIFF
--- a/sos/report/plugins/system.py
+++ b/sos/report/plugins/system.py
@@ -33,5 +33,12 @@ class System(Plugin, IndependentPlugin):
             "/proc/sys/net/ipv6/neigh/*/base_reachable_time"
         ])
 
+        # collect glibc tuning decisions
+        self.add_cmd_output([
+            "ld.so --help",
+            "ld.so --list-diagnostics",
+            "ld.so --list-tunables"
+        ])
+
 
 # vim: set et ts=4 sw=4 :


pmoravec:
I am not sure if we should stick to the hardcoded paths or not. `ld.so` is shipped exactly under `/usr/bin` on Red Hat distros, where this path is quite primary choice to look at among `PATH`s in sos (https://github.com/sosreport/sos/blob/main/sos/policies/distros/redhat.py#L75-L77). Original request in https://bugzilla.redhat.com/show_bug.cgi?id=2032913#c2 isn't much conclusive if the path does matter or not - but I feel it is improbable one to have a system with `ld.so` binary different than the `/usr/bin/ld.so`.

I havent found a different distro shipping this fairly new binary, yet. Maybe @BryanQuigley could comment wrt Debian/Ubuntu?

sos/report/plugins/system.py
```diff
@@ -33,5 +33,11 @@ def setup(self):
             "/proc/sys/net/ipv6/neigh/*/base_reachable_time"
         ])
 
+        # collect glibc tuning decisions
+        self.add_cmd_output([
+            "/usr/bin/ld.so --help",
```
TurboTurtle, line 38: Collecting `--help` output seems strange. Is there a particular reason?

sos/report/plugins/system.py
```diff
@@ -33,5 +33,11 @@ def setup(self):
...
+            "/usr/bin/ld.so --list-diagnostics"
```
TurboTurtle, line 39: Should we also be collecting `ld.so --list-tunables`?

TurboTurtle (commented):
Since this is in a canonical location, I think dropping the absolute path is the right move here. Let `Policy` handle `PATH` like we do for typical collections.

sos/report/plugins/system.py
pmoravec: I had the same question to the original reporter, quoting his answer:

> --help output contains some diagnostic output, too, perhaps in a more digestable form. It shows the search paths being used.

sos/report/plugins/system.py
pmoravec: Nothing like that has been asked for.

TurboTurtle: We should probably clarify - the commit and comment both reference tuning, but doesn't collect the only thing in the help output that references tuning:

```
  --list-tunables       list all tunables with minimum and maximum values
```

fweimer-rh: I wouldn't mind adding that, too, but the minimums/maximums currently aren't dynamic, so they won't confer additional system-specific information.

Cc @siddhesh

TurboTurtle: mins/maxes might not, but surely the actual value would?

```
# ld.so --list-tunables
glibc.rtld.nns: 0x4 (min: 0x1, max: 0x10)
glibc.elision.skip_lock_after_retries: 3 (min: 0, max: 2147483647)
glibc.malloc.trim_threshold: 0x0 (min: 0x0, max: 0xffffffffffffffff)
```

siddhesh: >> I wouldn't mind adding that, too, but the minimums/maximums currently aren't dynamic, so they won't confer additional system-specific information.

> mins/maxes might not, but surely the actual value would?

The x86-specific tunables (`x86_data_cache_size` for example) will depend on the target CPU since it gets the default value from `_SC_LEVEL1_DCACHE_SIZE`, so it could return different values on different machines.  `glibc.malloc.trim_threshold` though will just return 0 if not set since the default is dynamic, i.e. it starts at 128K and the malloc implementation changes it at runtime.  A single value is misleading since it implies that it is the single default value.  Many tunables however should have a constant default value.

So in summary, "it depends".

fweimer-rh: Okay, then let's include it, too.

pmoravec:
Thanks for the feedback (and hi @siddhesh ;-) ), `ld.so --list-tunables` was added.

TurboTurtle (approved):

PR #2813, opened 2022-01-04
[collector] Add cluster cleanup for all exit() calls

npinaeva:
OCP project wasn't deleted e.g. if node filter returned empty list

---


TurboTurtle (commented):
I think this would be better put within `exit()`, that way we additionally catch any other "gotcha"s.

npinaeva:
Yeah, I thought that too, but this `exit()` function is called 17 times, so I wasn't sure if it's safe to call `cluster.cleanup()` in all those places. But I think since you say so, it should be :)

pmoravec (approved):

TurboTurtle (approved):

PR #2816, opened 2022-01-05
[cirrus,tests] Update testing matrix to Fedora 35

TurboTurtle:
F35 has been out for a bit, we should update our testing matrix
accordingly.

New images have been pushed to GCP, so this change updates Cirrus to use
those new images.

---

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -3,8 +3,8 @@
 # Main environment vars to set for all tasks
 env:
 
-    FEDORA_VER: "34"
-    FEDORA_PRIOR_VER: "33"
+    FEDORA_VER: "35"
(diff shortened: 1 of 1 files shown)


pmoravec (approved):

pmoravec:
I can reproduce the `xsk_diag` kmod being loaded on F35 when running default `sos report`. It has nothing common to the PR itself, that just discovered the issue.

Trying to narrow down the problem to particular cmd.

TurboTurtle:
Interesting. Apparently F35 doesn't have `ethtool` by default anymore. Adding it to the install list, since the eeprom test is one we should definitely keep running.

pmoravec (approved):

PR #2819, opened 2022-01-06
[cleaner] Fix long domain handling/detection, parser handling, and known skips

TurboTurtle:
This patchset includes an alternative fix to #2811 which also carries some fixes for other issues found while investigating the CI failures from that change.

We should now properly account for long `--domains` values, and better handle parser file skips. Additionally, some well-known files/strings are added to skip lists that were discovered in tandem with this work.

---

### DIFF
--- a/sos/cleaner/__init__.py
+++ b/sos/cleaner/__init__.py
@@ -12,6 +12,7 @@ import hashlib
 import json
 import logging
 import os
+import re
 import shutil
 import tempfile
 
@@ -640,10 +641,16 @@ third party.
             self.log_debug("Obfuscating %s" % short_name or filename,
                            caller=arc_name)
             tfile = tempfile.NamedTemporaryFile(mode='w', dir=self.tmpdir)
+            _parsers = [
+                _p for _p in self.parsers if not
+                any([
+                    re.match(p, short_name) for p in _p.skip_files
+                ])
+            ]
             with open(filename, 'r') as fname:
                 for line in fname:
                     try:
-                        line, count = self.obfuscate_line(line)
+                        line, count = self.obfuscate_line(line, _parsers)
                         subs += count
                         tfile.write(line)
                     except Exception as err:
@@ -713,7 +720,7 @@ third party.
                 pass
         return string_data
 
-    def obfuscate_line(self, line):
+    def obfuscate_line(self, line, parsers=None):
         """Run a line through each of the obfuscation parsers, keeping a
         cumulative total of substitutions done on that particular line.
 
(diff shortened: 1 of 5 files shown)


sos/cleaner/__init__.py
```diff
@@ -643,7 +644,13 @@ def obfuscate_file(self, filename, short_name=None, arc_name=None):
             with open(filename, 'r') as fname:
                 for line in fname:
                     try:
-                        line, count = self.obfuscate_line(line)
+                        _parsers = [
+                            _p for _p in self.parsers if not
+                            any([
+                                re.match(p, short_name) for p in _p.skip_files
+                            ])
+                        ]
```
pmoravec, line 652: Shouldnt we move these lines before `with open` cycle, to let feed static `_parsers` just once (per filename)? Or maybe yet "higher" as that move would still mean regenerating `_parsers` once per each file?

sos/cleaner/mappings/hostname_map.py
```diff
@@ -102,9 +106,12 @@ def domain_name_in_loaded_domains(self, domain):
         and should be obfuscated
         """
         host = domain.split('.')
+        no_tld = '.'.join(domain.split('.')[0:-1])
         if len(host) == 1:
             # don't block on host's shortname
             return host[0] in self.hosts.keys()
+        elif any([no_tld.endswith(_d) for _d in self._domains]):
```
pmoravec, line 113: This was my real aim :)

sos/cleaner/mappings/hostname_map.py
```diff
@@ -50,10 +50,14 @@ def load_domains_from_map(self):
         in this parser, we need to re-inject entries from the map_file into
         these dicts and not just the underlying 'dataset' dict
         """
-        for domain in self.dataset:
+        for domain, ob_pair in self.dataset.items():
             if len(domain.split('.')) == 1:
                 self.hosts[domain.split('.')[0]] = self.dataset[domain]
             else:
+                if ob_pair.startswith('obfuscateddomain'):
```
pmoravec, line 57: Not required now, but in future, it would make sense to use a constant for `obfuscateddomain`, to declare it just once and use any times in code (this is 2nd usage).

pmoravec:
Just the one "move `_parsers` to be more static" is really relevant.

sos/cleaner/__init__.py
TurboTurtle: Oh, yep - this should totally be outside of the `for line` loop.

TurboTurtle: Regenerating once per file is the aim here - so that we have per-file selection of parsers.

pmoravec:
Ah right, the `_parsers` do rely on filename, my second advice was wrong.

pmoravec (approved):

Issue #2439, opened 2021-03-09
[plugin] Support copying files from containers


michaeltchapman:
Currently there is no method to provide a container from which to copy a file. This leads to adding cat commands within the container or other workarounds that may not be as clean.

I think this would involve extending `add_copy_spec`  to allow a container name similar to the command methods, extending the file copy manifest to include optional container names, and handling in the archive classes for performing the copy from the container into the archive.

TurboTurtle:
I'm playing catchup from being on PTO last week and need to get some more progress on the test suite updates, but this is on my radar and I'm hoping to get to it early next week.

michaeltchapman:
Thanks for looking into this. It does seem like recreating the experience for copying files is more trouble than it's worth. I would expect the interesting artifacts will usually be mounted either in config directories or data directories so I think just having a quick wrapper around cp is fine to cover the edge cases. I did think maybe adding an option to zip a directory and copy the zip across would be a reasonable alternative to directory copies, but it's hard to guarantee the requisite commands are available within the container. Even the shell version could be anything under the sun.

TurboTurtle:
I'm cycling back around on this following some other container-centric work and internal discussions.

I think I've got a good approach in mind, however there's one big decision to make here - _where_ do we put the copied files in the archive?

- In `sos_commands/$plugin/` since we'd be wrapping `podman cp` (or similar)? This is super easy and requires no extra work on our part, but I worry about accessibility in the sense of we go from "where ever the file exists on the host, is where it exists in the archive" (very straight forward and long standing) to that plus "* except when it comes from a container, in which case you need to search through the plugin directory with the container name and parse a command for a file name". Which, while not only not succinct will likely be flow-disrupting for analysis.
- In the "normal" path, but append some string to highlight it's a container copy? E.G. `/etc/foo/bar` becomes `/etc/foo/bar_container_$name`? I think this is _slightly_ better than the above, but there'd be a decent amount of shuffling in `Plugin()` to get this working like that, and I don't know how usable that naming scheme is for any kind of automation or script work that may be in use.
- Something else entirely? I'm open to ideas here. @pmoravec @jjansky1 @sferich888 @michaeltchapman

pmoravec:
"where ever the file exists on the **host**, is where it exists in the archive" - so if the file is "hidden" inside a container, how can I get to it in the host?

[...]

PR #2828, opened 2022-01-20
[Plugin] Add support for copying files from containers

TurboTurtle:
2-patch set to add support for copying files from containers via `add_copy_spec()` with a new `container` parameter.

The second patch in this set provides symlinking of container-based command collections to the new `sos_containers/` location introduced by the first commit from the normal location inside `sos_commands/$plugin/`.

Closes: #2439

---

### DIFF
--- a/sos/policies/runtimes/__init__.py
+++ b/sos/policies/runtimes/__init__.py
@@ -69,6 +69,12 @@ class ContainerRuntime():
             return True
         return False
 
+    def check_can_copy(self):
+        """Check if the runtime supports copying files out of containers and
+        onto the host filesystem
+        """
+        return True
+
     def get_containers(self, get_all=False):
         """Get a list of containers present on the system.
 
@@ -199,5 +205,31 @@ class ContainerRuntime():
         """
         return "%s logs -t %s" % (self.binary, container)
 
+    def get_copy_command(self, container, path, dest, sizelimit=None):
+        """Generate the command string used to copy a file out of a container
+        by way of the runtime.
+
+        :param container:   The name or ID of the container
+        :type container:    ``str``
+
+        :param path:        The path to copy from the container. Note that at
+                            this time, no supported runtime supports globbing
+        :type path:         ``str``
+
+        :param dest:        The destination on the *host* filesystem to write
+                            the file to
+        :type dest:         ``str``
+
+        :param sizelimit:   Limit the collection to the last X bytes of the
+                            file at PATH
+        :type sizelimit:    ``int``
+
+        :returns:   Formatted runtime command to copy a file from a container
+        :rtype:     ``str``
+        """
+        if sizelimit:
+            return "%s %s tail -c %s %s" % (self.run_cmd, container, sizelimit,
+                                            path)
+        return "%s cp %s:%s %s" % (self.binary, container, path, dest)
 
 # vim: set et ts=4 sw=4 :
--- a/sos/policies/runtimes/docker.py
+++ b/sos/policies/runtimes/docker.py
@@ -27,4 +27,7 @@ class DockerContainerRuntime(ContainerRuntime):
             return True
         return False
 
+    def check_can_copy(self):
+        return self.check_is_active(sysroot=self.policy.sysroot)
+
 # vim: set et ts=4 sw=4 :
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -561,6 +561,7 @@ class Plugin():
         self.commons = commons
         self.forbidden_paths = []
         self.copy_paths = set()
+        self.container_copy_paths = []
         self.copy_strings = []
         self.collect_cmds = []
         self.options = {}
@@ -621,6 +622,7 @@ class Plugin():
         self.manifest.add_list('commands', [])
         self.manifest.add_list('files', [])
         self.manifest.add_field('strings', {})
+        self.manifest.add_field('containers', {})
 
     def timeout_from_options(self, optname, plugoptname, default_timeout):
         """Returns either the default [plugin|cmd] timeout value, the value as
@@ -1558,7 +1560,7 @@ class Plugin():
                 self.manifest.files.append(manifest_data)
 
     def add_copy_spec(self, copyspecs, sizelimit=None, maxage=None,
-                      tailit=True, pred=None, tags=[]):
+                      tailit=True, pred=None, tags=[], container=None):
         """Add a file, directory, or regex matching filepaths to the archive
 
         :param copyspecs: A file, directory, or regex matching filepaths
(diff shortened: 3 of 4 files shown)


TurboTurtle:
So we use the `cp` function of the runtime here to copy files that do not exceed a provided sizelimit.

I've used `tail` via a runtime exec instead of doing something like `podman cp mycontainer:/my/file -` which would send the output to stdout - in the case of size limiting is because when the file is sent to stdout, currently supported runtimes forcibly write it to stdout as a tar archive. Which makes everything binary data which we can't size limit - we'd just end up with the tail end of a tar archive that we can't uncompress because the header would be missing.

This does however assume that `tail` is present in any given container. I'm open to hearing any objections to this or any other avenues we might have, but I don't have a better approach in mind.

@npinaeva the `crio` runtime doesn't have a `cp` function, so on OCP nodes we effectively would still be unable to copy from a container. I'm assuming we want to ask the crio team to add support for copying from those containers as well, but wanted to check in with you first on that front if it's actually desired.

npinaeva:
Yeah I complained about not being able to copy files from containers, that adds a dependency on where container volume is (and if) mounted. 
I have only found this Issue for crictl https://github.com/kubernetes-sigs/cri-tools/issues/677 that doesn't look like going to be resolved. And using `kubectl cp` instead requires `tar` to be installed in container https://github.com/kubernetes/kubernetes/issues/58512 and some comments say that it should be implemented in crio instead.
So I don't see any reliable way of copying files now (maybe there is some workaround), if we can ask `cp` to be implemented for `crio` that would be really nice

sos/policies/runtimes/__init__.py
```diff
@@ -199,5 +205,31 @@ def get_logs_command(self, container):
...
+                            this time, no supported runtime supports globbing
+        :type path:         ``str``
+
+        :param dest:        The destination on the *host* filesystem to write
+                            the file to
+        :type dest:         ``str``
+
+        :param stdout:      Print the contents of the file to stdout, instead
+                            of writing the file to ``dest``
+        :type stdout:       ``bool``
```
pmoravec, line 225: Not an argument of the method. `sizelimit` is there instead :)

sos/policies/runtimes/__init__.py
```diff
@@ -199,5 +205,31 @@ def get_logs_command(self, container):
...
+
+        :returns:   Formatted runtime command to copy a file from a container
+        :rtype:     ``str``
+        """
+        if sizelimit:
+            return "%s %s tail -c %s %s" % (self.run_cmd, container, sizelimit,
```
pmoravec, line 231: Is `sizelimit` in bytes here (`tail -c`), or in MB (like `--size-limit` or `a_c_s` argument)?

sos/policies/runtimes/docker.py
```diff
@@ -27,4 +27,7 @@ def check_is_active(self, sysroot=None):
             return True
         return False
 
+    def check_can_copy(self):
+        return self.check_is_active(sysroot=self.policy.sysroot)
```
pmoravec, line 31: Smart!

sos/policies/runtimes/__init__.py
TurboTurtle: Bytes. In `a_c_s` we convert the MB value to bytes, and we hand that off to the container collections as part of the tuple. But we only use the sizelimit value if we've determined we actually need to use tail.

sos/policies/runtimes/__init__.py
TurboTurtle: Ah, yep - `stdout` was from an earlier revision. I'll update the docstrings.

sos/report/plugins/__init__.py
```diff
@@ -1453,28 +1462,79 @@ def add_copy_spec(self, copyspecs, sizelimit=None, maxage=None,
...
 
+        def get_filename_tag(fname):
+            """Generate a tag to add for a single file copyspec
+
+            This tag will be set to the filename, minus any extensions
+            except '.conf' which will be converted to '_conf'
+            """
+            fname = fname.replace('-', '_')
+            if fname.endswith('.conf'):
+                return fname.replace('.', '_')
```
pmoravec, line 1473: This will for `fname = "foo.bar.conf"` return `foo_bar_conf` - are *both* / all replacements intentional, or just the latest one?

sos/report/plugins/__init__.py
```diff
@@ -1453,28 +1462,79 @@ def add_copy_spec(self, copyspecs, sizelimit=None, maxage=None,
...
-                if fname.endswith('.conf'):
-                    return fname.replace('.', '_')
-                return fname.split('.')[0]
+            if container:
+                if isinstance(container, str):
+                    container = [container]
+                for con in container:
+                    if not self.container_exists(con):
+                        continue
+                    _tail = False
```
pmoravec, line 1502: What happens if we aim to copy *more* than sizelimit is? Then we set `_tail` to True and tail zero size from any other "container over limit"? Should not we set `_tail = False` before the `for con` cycle and have here `is _tail: break`?

sos/report/plugins/__init__.py
```diff
@@ -1453,28 +1462,79 @@ def add_copy_spec(self, copyspecs, sizelimit=None, maxage=None,
...
+                                continue
+                        else:
+                            self._log_debug(
+                                "stat of '%s' in container '%s' failed, "
+                                "skipping collection: %s"
+                                % (copyspec, con, ret['output'])
+                            )
+                            continue
+                    self.container_copy_paths.append(
+                        (con, copyspec, sizelimit, _tail, _spec_tags)
```
pmoravec, line 1528: How does this will work for the copyspec being tailed? Will we really tail just `cosize-sizelimit` bytes or collect it whole?

sos/report/plugins/__init__.py
TurboTurtle: Since copying from containers currently only supports explicit files - no globs or similar - each `copyspec` is treated as a separate collection up to `sizelimit`, and is distinct from other containers. So if the stat of the file comes back over `sizelimit`, we set `_tail = True` for that collection within the container where the file is larger, and collect the last X bytes of the file.

sos/policies/runtimes/__init__.py
TurboTurtle: Updated in latest push.

sos/report/plugins/__init__.py
pmoravec: Ah I understand. I was already one step ahead, trying to collect `/foo/bar*` in one glob that one has to explicitly enumerate to `/foo/bar1`, `/foo/bar2` and `/foo/bar3` and `a_s_c` will collect up to `sizelimit` of each file - that is subtle different behaviour due to the "manual glob expansion". But it makes sense..

pmoravec (approved):

PR #2830, opened 2022-01-21
[foreman] Use psql-msgpack-decode wrapper for dynflow >= 1.6

pmoravec:
In dynflow >=1.6.3, dynflow* tables in postgres are encoded by
msgpack which makes plain CSV dumps unreadable. In such a case,
psql-msgpack-decode wrapper tool from dynflow-utils (of any
version) must be used instead of the plain psql command.

Resolves: #2830

---

### DIFF
--- a/sos/report/plugins/foreman.py
+++ b/sos/report/plugins/foreman.py
@@ -244,8 +244,16 @@ class Foreman(Plugin):
             self.add_cmd_output(_cmd, suggest_filename=table, timeout=600,
                                 sizelimit=100, env=self.env)
 
+        # dynflow* tables on dynflow >=1.6.3 are encoded and hence in that
+        # case, psql-msgpack-decode wrapper tool from dynflow-utils (any
+        # version) must be used instead of plain psql command
+        dynutils = self.is_installed('dynflow-utils')
         for dyn in foremancsv:
-            _cmd = self.build_query_cmd(foremancsv[dyn], csv=True)
+            binary = "psql"
+            if dyn != 'foreman_tasks_tasks' and dynutils:
+                binary = "/usr/libexec/psql-msgpack-decode"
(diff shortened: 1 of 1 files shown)


sos/report/plugins/foreman.py
```diff
@@ -244,8 +244,17 @@ def setup(self):
             self.add_cmd_output(_cmd, suggest_filename=table, timeout=600,
                                 sizelimit=100, env=self.env)
 
+        # dynflow* tables on dynflow >=1.6.3 are encoded and hence in that
+        # case, psql-msgpack-decode wrapper tool from dynflow-utils (any
+        # version) must be used instead of plain psql command
+        dynutils = 'dynflow-utils' in \
+            self.policy.package_manager.get_pkg_list().keys()
```
TurboTurtle, line 251: Is `self.is_installed('dynflow-utils')` not sufficient here?

pmoravec: Ah right. It was a relic of previous testing "dynutils of version .." which is not required.

TurboTurtle (approved):

adamruzicka:
dynflow-utils is already available in nightly foreman repos http://yum.theforeman.org/nightly/el7/x86_64/dynflow-utils-1.6.3-1.el7.x86_64.rpm . For testing it can be installed anywhere, it has no dependencies and contains only a static binary and a shell script

pmoravec:
> dynflow-utils is already available in nightly foreman repos http://yum.theforeman.org/nightly/el7/x86_64/dynflow-utils-1.6.3-1.el7.x86_64.rpm . For testing it can be installed anywhere, it has no dependencies and contains only a static binary and a shell script

Thanks. The package and binary works great on my reproducer of this PR.

Issue #2829, opened 2022-01-21
[cleaner] The regex fails to match if mac/IPv6 is at start position 


NikhilKakade-1:
The current regex fails to obscure the mac or IPv6 at the start of the line.
https://github.com/sosreport/sos/blob/105075539e0cf7525657aa2004ca6503edf6e6b2/sos/cleaner/parsers/mac_parser.py#L26
The regex should be:

`(([^:|-])?([0-9a-fA-F]{2}([:-])){5}([0-9a-fA-F]){2}(.)?(\s|$|\W))`

**Before:**

**NOT Matching**

<img width="816" alt="Screenshot 2022-01-21 at 2 01 34 PM" src="https://user-images.githubusercontent.com/66179696/150525189-3707d304-8aac-4897-97ea-e5c620a33e62.png">

**Matches only If we do have character at start:**

<img width="816" alt="Screenshot 2022-01-21 at 2 01 42 PM" src="https://user-images.githubusercontent.com/66179696/150525348-966598df-d567-4aef-8aae-788280868cce.png">

> ### FIX

<img width="1651" alt="Screenshot 2022-01-21 at 5 47 33 PM" src="https://user-images.githubusercontent.com/66179696/150525860-e7d8a0b9-5e0e-4ce7-b93e-fe28fe997954.png">

@TurboTurtle what's your suggestion? am I missing something here?

TurboTurtle:
This looks like a valid and correct fix to me, yes.

If you open a PR with this change our test suite will verify that we're not causing a regression here with our current expectation of mac obfuscations.

UcanD0:
@TurboTurtle 
I have raised PR could you please, look into it?

TurboTurtle:
Closing per #2833

PR #2833, opened 2022-01-24
[cleaner] Update regex for finding IPv4/v6 addresses

NikhilKakade-1:
The regex trys to match a single character not present
in the given list [^:|-]. The string which starts with
addresses are left out as there is
no character to match.

Adding ? to matches the previous token between zero and one time.

Resolves Issue: #2829

Signed-off-by: Nikhil Kakade <nikhilkaka5@outlook.com>

---

### DIFF
--- a/sos/cleaner/parsers/mac_parser.py
+++ b/sos/cleaner/parsers/mac_parser.py
@@ -20,10 +20,10 @@ class SoSMacParser(SoSCleanerParser):
     name = 'MAC Parser'
     regex_patterns = [
         # IPv6
-        r'(([^:|-])([0-9a-fA-F]{2}(:|-)){7}[0-9a-fA-F]{2}(\s|$))',
+        r'(([^:|-])?([0-9a-fA-F]{2}(:|-)){7}[0-9a-fA-F]{2}(\s|$))',
         r'(([^:|-])([0-9a-fA-F]{4}(:|-)){3}[0-9a-fA-F]{4}(\s|$))',
         # IPv4, avoiding matching a substring within IPv6 addresses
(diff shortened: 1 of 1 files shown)


TurboTurtle:
This change does pick up additional mac addrs, however it appears to have the unintended side effect of matching IPv6 mac addrs against the tail end of IPv6 addresses as well.

For example, on a test box:

```
proc/1402905/net/nf_conntrack:ipv6     10 udp      17 19 src=fe80:0000:0000:0000:dee7:0579:5c20:d01d dst=ff02:0000:0000:0000:0000:0000:0000:00fb  [....]
```

Becomes with this change:

```
proc/1412340/net/nf_conntrack:ipv6     10 udp      17 19 src=fe80:0000:0000:0000:534f:53ff:fe31:926e dst=ff02:0000:0000:0000:534f:53ff:fee7:2fa4  [....]
```

TurboTurtle:
Ok, so the change is valid for the ipv4 syntax mac addresses, and the 8 2-char hextet ipv6 pattern. The 4 4-char hextet syntax is what needs refinement and we'll need to use a negative lookbehind here I think.

I'm working that out on my end now. For now, drop the changes from line 24, and I think this should be good for merge (though of course I'll do some more testing on that).

UcanD0:
Sure @TurboTurtle will keep looking for the 4 4-char hextet refinement.
By the way, I have dropped the changes from line 24 as you suggested.

TurboTurtle (approved):

pmoravec (approved):

UcanD0:
@pmoravec @TurboTurtle 
Could you please help to merge this PR?

pmoravec:
Usually we (well, almost solely Jake :) ) do merge PRs in some batches, say once or twice per week.

PR #2835, opened 2022-01-25
[mac_parser] Avoid substring matches with look-behinds

TurboTurtle:
This commit adds a negative look-behind to the regex patterns for mac
addresses so we avoid matching substrings from both non-mac-address
strings, certain ipv6 address strings, and longer strings of hextets
that we could potentially match a substring of.

---


pmoravec (approved):
I hope I got the cryptic regexp right :) as it newly aims to prevent false matches of strings like `aa::aa::aa::aa::aa::aa::aa::aa::aa::aa` that should not be interpreted as IPv6 addresses.

PR #2842, opened 2022-02-04
[cleaner] Improve concurrent archive cleaning performance

TurboTurtle:
Prior to this change, concurrent archives were handled with a thread
pool. While this worked, the performance gains were not significant as
we were still locked into a single processor for all archives.

With this commit, we now use a process pool to obfuscate archives. `sos
clean` will now use up to `--jobs` processes (default 4) for concurrent
archive obfuscation. This results in a significant performance increase
for cleaning, for example, `sos collect` tarballs.

This requires moving several aspect of `SoSCleaner()` into different/new
objects so that the process pool can correctly pickle everything needed
in the new process space.

This in turn requires a shared memory object for the parser maps to
store their internal datasets. Multiprocessing dicts are inherently much
slower for access than standard dict objects, so in the event of single
archive obfuscations, parser maps retain the standard dict objects, and
we will only replace those with shared dicts for 2 or more concurrent
archives.

---

### DIFF
--- a/sos/cleaner/__init__.py
+++ b/sos/cleaner/__init__.py
@@ -16,7 +16,7 @@ import re
 import shutil
 import tempfile
 
-from concurrent.futures import ThreadPoolExecutor
+from concurrent.futures import ProcessPoolExecutor
 from datetime import datetime
 from pwd import getpwuid
 from sos import __version__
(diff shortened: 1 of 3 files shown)


TurboTurtle:
Interesting... in python != 3.6, `threading.Lock()`s are pickled, but in 3.6 they aren't.

Come to think of it, with this change we don't *need* that type of Lock at all, since we only use it for single obfuscations, and in multi-archive obfuscations we're using a `Manager.Lock()` instead to share locking across the processes.

I'll look into the simplest way to do a conditional lock here in `SoSMap.add()` and update the PR shortly.

TurboTurtle:
I originally thought this was due to the `Lock()` object being used in `SoSMap().__init__()`, but that's now removed and we're still hitting the same issue on python-3.6.

What's more, is that this seems to be within the `multiprocessing` lib directly - we don't have any threading locks ourselves now, especially not in `obfuscate_report()` which is what is getting pickled - yet we're still getting this exception.

TurboTurtle:
Confirmed, it's in python-3.6. Using the RH appstream python38, this commit behaves as expected.

This presents an issue - we *have* to use multiprocessing (either directly or indirectly via the `concurreny.futures` library) in some way in order to do what this commit intends. However, the version of python in at least 2 of our downstreams (RHEL 8 and Ubuntu 18.04 (plus derivatives of both, actually)) has this issue.

I can reach out to the RH python team to see about how feasible a backport fix would be, but before I do so I think we should see if that is at all possible for 18.04. @slashdd @BryanQuigley ?

TurboTurtle:
It turns out that we inherit an `RLock()` via the `logging` module. I do not know, nor does the team reviewing the initial report I made, why the code here works (and to be clear, works **as expected**) on python 3.7+ but not 3.6, but the general consensus is that it should _not_ work anywhere.

The obvious initial response is to remove the logger from the `SoSReportArchive` objects - but that's not really an option in my opinion as we need something to highlight "we obfuscated $file in $time" and/or "we had an issue obfuscating $file".

So, my other thought was rather than shunting the entire archive off into a into process and running there, we could keep the the archive handling in the main process and use a `ProcessPoolExecutor` for the file list each archive provides. That is....well, honestly it's awful. Single report obfuscation time sky rockets, which makes this a non-starter to begin with before even looking at multi archive times. I _believe_ this is due to my use of shared objects via a `multiprocessing.Manager()` which uses proxy objects to control access, but in turn has a massive performance penalty. Theoretically, this _might_ be overcome by the use of shared ctypes in some capacity, but that's well beyond my wheelhouse, personally.


Unfortunately I think this means that as long as we support a downstream that is based on python-3.6, we can't move forward using the `multiprocessing` module.

If anyone has any input or thoughts on other implementations, I'd love to hear it as I would really like to get multi-archive obfuscation times down to something reasonable.

TurboTurtle:
Closing this out for now, given the hurdles above. I think this is a good candidate for a re-visit in the not-to-distant future, but I have some ideas percolating around refactoring some of our approach in `clean`, that should be settled before we tackle this again.

PR #2848, opened 2022-02-10
[tests] Minor fixes for avocado and unittests

TurboTurtle:
`SoftwareManager()` from the avocado test suite is now under a different
module, so we need to update the help output test accordingly.

Additionally, the legacy runner is being dismantled in avocado in favor
of the nrunner test runner. Upstream indicates it will be fully removed
sooner rather than later, so we should get ahead of it and switch to
nrunner. To do so successfully, since our tests are designed to have a
single `sos` execution per test case (Test() superclass), we need to
disable parallel execution of tests for this runner.

As part of local testing, it was found that the previous sysroot fix had
an unintentional side-effect of causing our unittests to fail if they
were run by a non-root user. This is because of a call to `os.chroot()`
in `_child_prep_fn()`. This has the edge case possibility of also
throwing an exception in certain invocations of `sos collect`, so add a
second conditional that we aren't trying to chroot to `/`. As a knock-on
effect, this will mean we are not constantly calling `os.chroot()` for
every command executed during report collections.

---

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -118,7 +118,7 @@ report_stageone_task:
     # run the unittests separately as they require a different PYTHONPATH in
     # order for the imports to work properly under avocado
     unittest_script: PYTHONPATH=. avocado run tests/unittests/
-    main_script: PYTHONPATH=tests/ avocado run --test-runner=runner -t stageone tests/{cleaner,collect,report,vendor}_tests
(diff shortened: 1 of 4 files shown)


pmoravec:
"disable parallel execution of tests for this runner." - that would mean slower execution of the tests, right? Sigh..

TurboTurtle:
Closing this at this time - once avocado hits their next LTS I'll revisit the update work on our end.

PR #2853, opened 2022-02-14
[utilities] Don't try to chroot to /

TurboTurtle:
With the recent fix for sysroot being `None` to always being (correctly)
`/`, we should guard against situations where `sos_get_command_output()`
would now try to chroot to `/` before running any command. Incidentally,
this would also cause our unittests to fail if they were run by a
non-root user.

---


pmoravec (approved):
ACK, just dont forget to remove this change from https://github.com/sosreport/sos/pull/2848 ;-)

PR #2857, opened 2022-02-15
[cleaner] Use compiled regex lists for parsers by default

TurboTurtle:
This commit follows the initial change made to the username and keyword
parsers in #2823 and applies it to all parsers by default.

When a new match is found, a new `regex.Pattern()` object will now be
compiled and saved in the parser's map, and this object will be used for
ongoing obfuscations from that point forward, rather than rebuilding
regexes for every line we iterate over with that parser.

This is now built into the base `SoSMap` and leveraged by parsers,
rather than being handled directly by parsers. Further, this will be
enabled by default for all existing and new parsers. This shows decent
improvements to hostname parser performance in local testing. Note that
this functionality is explicitly disabled for the mac and ip parsers.

---

### DIFF
--- a/sos/cleaner/mappings/__init__.py
+++ b/sos/cleaner/mappings/__init__.py
@@ -24,21 +24,23 @@ class SoSMap():
     ignore_matches = []
     # used for filename obfuscations in parser.parse_string_for_keys()
     skip_keys = []
+    compile_regexes = True
 
     def __init__(self):
         self.dataset = {}
+        self._regexes_made = set()
+        self.compiled_regexes = []
         self.lock = Lock()
 
     def ignore_item(self, item):
         """Some items need to be completely ignored, for example link-local or
         loopback addresses should not be obfuscated
         """
+        if not item or item in self.skip_keys or item in self.dataset.values():
+            return True
         for skip in self.ignore_matches:
             if re.match(skip, item):
                 return True
-
-    def item_in_dataset_values(self, item):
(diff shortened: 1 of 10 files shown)


pmoravec:
> Note that this functionality is explicitly disabled for the mac and ip parsers.

Why so, please?

TurboTurtle:
> > Note that this functionality is explicitly disabled for the mac and ip parsers.
> 
> Why so, please? (I am just starting to review the changes, maybe this will be self-answered)

Because those parsers don't make use of secondary checks or similar - the "main" regexes used are the only form that strings needing substitutions can take (and if we did do this we'd have to add a negative check for an obfuscated match that would still hit the "main" patterns, which would counteract the intended performance gain the compiled lists provide).

sos/cleaner/mappings/__init__.py
```diff
@@ -48,12 +50,52 @@ def add(self, item):
...
+        :param item:    The unobfuscated item to generate a regex for
+        :type item:     ``str``
+        """
+        if self.ignore_item(item):
+            return
+        if item not in self._regexes_made:
+            # save the item in a set to avoid clobbering existing regexes,
+            # as searching this set is significantly faster than searching
+            # through the actual compiled_regexes list, especially for very
+            # large collections of entries
```
pmoravec, line 74: Nice.

pmoravec (approved):

pmoravec:
Ah that makes sense.

PR #2859, opened 2022-02-15
[runtimes] Allow container IDs to be used with `container_exists()`

TurboTurtle:
As container runtimes can interchange container names and container IDs,
sos should also allow the use of container IDs when checking for the
presence of a given container.

In particular, this change unblocks the use of `Plugin.exec_cmd()` when
used in conjunction with `Plugin.get_container_by_name()` to pick a
container based on a provided regex that the container name may match.

Related: #2856

---

### DIFF
--- a/sos/policies/runtimes/__init__.py
+++ b/sos/policies/runtimes/__init__.py
@@ -147,6 +147,23 @@ class ContainerRuntime():
                     vols.append(ent[-1])
         return vols
 
+    def container_exists(self, container):
+        """Check if a given container ID or name exists on the system from the
+        perspective of the container runtime.
+
+        Note that this will only check _running_ containers
+
+        :param container:       The name or ID of the container
+        :type container:        ``str``
+
+        :returns:               True if the container exists, else False
+        :rtype:                 ``bool``
+        """
+        for _contup in self.containers:
+            if container in _contup:
+                return True
+        return False
+
     def fmt_container_cmd(self, container, cmd, quotecmd):
         """Format a command to run inside a container using the runtime
 
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -2593,7 +2593,7 @@ class Plugin():
         """If a container runtime is present, check to see if a container with
         a given name is currently running
 
-        :param name:    The name of the container to check presence of
+        :param name:    The name or ID of the container to check presence of
         :type name: ``str``
 
         :returns: ``True`` if `name` exists, else ``False``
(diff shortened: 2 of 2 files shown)


sos/policies/runtimes/__init__.py
```diff
@@ -147,6 +147,23 @@ def get_volumes(self):
...
+        Note that this will only check _running_ containers
+
+        :param container:       The name or ID of the container
+        :type container:        ``str``
+
+        :returns:               True if the container exists, else False
+        :rtype:                 ``bool``
+        """
+        for _contup in self.containers:
+            if container in _contup:
```
pmoravec, line 163: When testing this with `podman`, the `_contup` is a pair of container ID and image name, not the container name, like:

    # podman ps
CONTAINER ID  IMAGE                                       COMMAND    CREATED        STATUS            PORTS  NAMES
    902ab039c904  registry.access.redhat.com/ubi8/ubi:latest  /bin/bash  4 minutes ago  Up 4 minutes ago         nice_einstein
    e214612764d3  registry.access.redhat.com/ubi8/ubi:latest  /bin/bash  4 minutes ago  Up 4 minutes ago         xenodochial_bohr
    # 

While printing the pair gets me:

    902ab039c904 registry.access.redhat.com/ubi8/ubi:latest
    e214612764d3 registry.access.redhat.com/ubi8/ubi:latest

Did I something wrong or ar we really storing image name instead of container name? See https://github.com/sosreport/sos/blob/main/sos/policies/runtimes/__init__.py#L89-L92 how we feed the `self.containers` list.

fernandoroyosanchez: I also tested and you are right about _contup is a tuple (ID, container name). 

[self.containers ](https://github.com/sosreport/sos/blob/main/sos/policies/runtimes/__init__.py#L89-L92 ) are ok as ([0],**[-1]**) so we are storing ID and container name.

TurboTurtle: The name should always be the last field for podman:

```
$ podman ps
CONTAINER ID  IMAGE       COMMAND     CREATED     STATUS      PORTS       NAMES
```

I've run a few more sanity tests locally, and from that I'm always seeing the name in the last field.

What version of podman are you testing with?

pmoravec: Ah I apologize. Typo in my debug (`1` instead of `-1`, hence image name instead of container name). This works well.

pmoravec (approved):

Issue #2863, opened 2022-02-17
[report] Not able to set timeout using "plugin_timeout" attribute. 


NikhilKakade-1:
@TurboTurtle @pmoravec 
Not able to set plugin timeout.
As per [document](https://sos.readthedocs.io/en/main/plugins.html), the timeout of an individual plugin can be set using `plugin_timeout`.

Explicitly added `sleep 10` and `plugin_timeout=1`

<img width="632" alt="Screenshot 2022-02-17 at 9 55 53 AM" src="https://user-images.githubusercontent.com/66179696/154406320-7a7e35b4-537f-4373-a03f-d9976baf6f22.png">

### RESULT
<img width="632" alt="Screenshot 2022-02-17 at 9 55 29 AM" src="https://user-images.githubusercontent.com/66179696/154406165-befdcd2f-5638-4f32-af55-4678582360f9.png">

pmoravec:
Indeed, when I collect e.g. `foreman` plugin, I see in generated `sos_reports/manifest.json` :

            "plugins": {
                "foreman": {
                    ..
                    "timeout": 300,

PR #2864, opened 2022-02-17
[report] Honor plugins' hardcoded plugin_timeout

pmoravec:
Currently, plugin's plugin_timeout hardcoded default is superseded by
whatever --plugin-timeout value, even when this option is not used and
we eval it to TIMEOUT_DEFAULT.

In this case of not setting --plugin-timeout, honour plugin's
plugin_timeout instead.

Resolves: #2863
Closes: #2864

---

### DIFF
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -636,7 +636,10 @@ class Plugin():
             if opt_timeout is None:
                 _timeout = own_timeout
             elif opt_timeout is not None and own_timeout == -1:
-                _timeout = int(opt_timeout)
+                if opt_timeout == TIMEOUT_DEFAULT:
+                    _timeout = default_timeout
+                else:
+                    _timeout = int(opt_timeout)
             elif opt_timeout is not None and own_timeout > -1:
                 _timeout = own_timeout
             else:


sos/report/plugins/__init__.py
```diff
@@ -633,7 +633,7 @@ def timeout_from_options(self, optname, plugoptname, default_timeout):
         try:
             opt_timeout = self.get_option(optname)
             own_timeout = int(self.get_option(plugoptname))
-            if opt_timeout is None:
+            if opt_timeout is None or opt_timeout == TIMEOUT_DEFAULT:
                 _timeout = own_timeout
```
TurboTurtle, line 637: I believe this needs to be `_timeout = default_timeout`, because here we are passing `self.plugin_timeout` as `default_timeout`, whereas `own_timeout` would be something passed to `-k plugin.timeout`.

TurboTurtle: Either that, or this:

```python
if opt_timeout is None:
    _timeout = own_timeout
elif opt_timeout == TIMEOUT_DEFAULT:
    _timeout = default_timeout
    [...]
```

TurboTurtle: Actually, we may want to nest the check against `TIMEOUT_DEFAULT` in the second conditional, so that we can properly override with `-k plugin.timeout` and `--plugin-timeout`:

```
diff --git a/sos/report/plugins/__init__.py b/sos/report/plugins/__init__.py
index cc5cb65b..336b4d22 100644
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -636,7 +636,10 @@ class Plugin():
             if opt_timeout is None:
                 _timeout = own_timeout
             elif opt_timeout is not None and own_timeout == -1:
-                _timeout = int(opt_timeout)
+                if opt_timeout == TIMEOUT_DEFAULT:
+                    _timeout = default_timeout
+                else:
+                    _timeout = int(opt_timeout)
             elif opt_timeout is not None and own_timeout > -1:
                 _timeout = own_timeout
             else:
```

pmoravec: Riight, I knew I havent thought all combinations.

pmoravec: :+1:

TurboTurtle (approved):

PR #2866, opened 2022-02-17
[tests] Add a test for non-default plugin_timeout compliance

TurboTurtle:
There was a gap in our testing that allowed #2863 to escape our notice -
that a `Plugin()`'s `plugin_timeout` attribute would be ignored if it
wasn't set to `TIMEOUT_DEFAULT`.

As that was resolved by #2864, add a test to ensure it remains working
as expected. The expected resolution order for a plugin's whole timeout
is as follows:

 1. The value set by `-k plugin.timeout`
 2. The value set by `--plugin-timeout`
 3. The value hardcoded in the plugin via the `plugin_timeout` attr
 4. `TIMEOUT_DEFAULT`

Related: #2863
Related: #2864

---

### DIFF
--- a/tests/report_tests/timeout_tests.py
+++ b/tests/report_tests/timeout_tests.py
@@ -18,6 +18,10 @@ class PluginTimeoutTest(StageTwoReportTest):
     install_plugins = ['timeout_test']
     sos_cmd = '-o timeout_test -vvv --plugin-timeout=10'
 
+    def test_correct_plugin_timeout(self):
+        man = self.get_plugin_manifest('timeout_test')
+        self.assertEquals(man['timeout'], 10)
+
     def test_plugin_timed_out(self):
         self.assertSosLogNotContains('collected plugin \'timeout_test\' in')
         self.assertSosUILogContains('Plugin timeout_test timed out')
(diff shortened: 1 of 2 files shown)


tests/report_tests/timeout_tests.py
```diff
@@ -35,10 +39,27 @@ class NativeCmdTimeoutTest(StageTwoReportTest):
     install_plugins = ['timeout_test']
     sos_cmd = '-o timeout_test -vvv'
 
+    def test_correct_plugin_timeout(self):
```
pmoravec, line 42: Worth to add to some other `fake_test` this?

    from sos.utilities import TIMEOUT_DEFAULT

        def test_correct_plugin_timeout(self):
            man = self.get_plugin_manifest('binary_test')
            self.assertEquals(man['timeout'], TIMEOUT_DEFAULT)

? Well, that would introduce some dependency among fake plugins that isnt ideal, so maybe have another `defaulttimeout_test` plugin to have also similar tests like here for a plugin without hardcoded timeout?

TurboTurtle: I folded these types of tests into the existing ones by enabling the `host` plugin and checking its timeout setting where appropriate. I think that should suffice?

pmoravec (approved):

PR #2867, opened 2022-02-21
[sosnode] Handle downstream versioning for runtime option check

TurboTurtle:
Adds a fallback version check for 4.2-13 for RHEL downstreams that
backport the `container-runtime` option into sos-4.2.

Carry this in upstream to account for use cases where a workstation used
to run `collect` from may be from a different stream than those used by
cluster nodes.

---

### DIFF
--- a/sos/collector/sosnode.py
+++ b/sos/collector/sosnode.py
@@ -275,21 +275,34 @@ class SosNode():
     def _load_sos_info(self):
         """Queries the node for information about the installed version of sos
         """
+        ver = None
+        rel = None
         if self.host.container_version_command is None:
             pkg = self.host.package_manager.pkg_version(self.host.sos_pkg_name)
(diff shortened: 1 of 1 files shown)


pmoravec (approved):

sos/collector/sosnode.py
```diff
@@ -587,7 +607,8 @@ def finalize_sos_cmd(self):
                 sos_opts.append('--cmd-timeout=%s'
                                 % quote(str(self.opts.cmd_timeout)))
 
-        if self.check_sos_version('4.3'):
+        # handle downstream versions that backported this option
+        if self.check_sos_version('4.3') or self.check_sos_version('4.2-13'):
```
npinaeva, line 611: It works if you change version to "4.2.13"

pmoravec (approved):

PR #2868, opened 2022-02-28
[Plugins] Adjust paths for containerized setup

mrunge:
---

### DIFF
--- a/sos/report/plugins/collectd.py
+++ b/sos/report/plugins/collectd.py
@@ -21,14 +21,17 @@ class Collectd(Plugin, IndependentPlugin):
     # or being inside Super Proviledged Container that does not have
     # the package but logs to the host's logfile
     packages = ('collectd',)
-    files = ('/var/log/containers/collectd/collectd.log',)
+    files = ('/var/log/containers/collectd/collectd.log',
+             '/var/log/collectd/collectd.log')
 
     def setup(self):
         self.add_copy_spec([
             '/etc/collectd.conf',
             '/etc/collectd.d/*.conf',
             '/var/log/containers/collectd/collectd.log',
-            '/var/lib/config-data/collectd',
+            '/var/lib/config-data/puppet-generated/collectd/etc/collectd.conf',
+            '/var/lib/config-data/puppet-generated/collectd/etc/collectd.d/'
+            + '*.conf',
         ])
 
         p = re.compile('^LoadPlugin.*')
--- a/sos/report/plugins/openstack_aodh.py
+++ b/sos/report/plugins/openstack_aodh.py
@@ -102,4 +102,17 @@ class RedHatOpenStackAodh(OpenStackAodh, RedHatPlugin):
 
     packages = ('openstack-selinux',)
 
+    def setup(self):
+        super(RedHatOpenStackAodh, self).setup()
+        if self.get_option("all_logs"):
+            self.add_copy_spec([
+                "/var/log/containers/httpd/aodh-api/*",
+                "/var/log/containers/aodh/*"
+            ])
+        else:
+            self.add_copy_spec([
+                "/var/log/containers/httpd/aodh-api/*.log",
+                "/var/log/containers/aodh/*.log"
+            ])
+
 # vim: set et ts=4 sw=4 :
--- a/sos/report/plugins/openstack_ceilometer.py
+++ b/sos/report/plugins/openstack_ceilometer.py
@@ -86,11 +86,11 @@ class RedHatCeilometer(OpenStackCeilometer, RedHatPlugin):
         super(RedHatCeilometer, self).setup()
         if self.get_option("all_logs"):
             self.add_copy_spec([
-                "/var/log/httpd/ceilometer*",
+                "/var/log/containers/ceilometer/*",
             ])
         else:
             self.add_copy_spec([
-                "/var/log/httpd/ceilometer*.log",
+                "/var/log/containers/ceilometer/*.log",
             ])
 
 # vim: set et ts=4 sw=4 :


csibbitt (approved):
LGTM

sos/report/plugins/collectd.py
```diff
@@ -28,7 +28,7 @@ def setup(self):
             '/etc/collectd.conf',
             '/etc/collectd.d/*.conf',
             '/var/log/containers/collectd/collectd.log',
-            '/var/lib/config-data/collectd',
```
pmoravec, line 31: Can't `sos` run on some older version of `collectd` where the data are yet stored in this location?

sos/report/plugins/openstack_aodh.py
```diff
@@ -33,13 +33,13 @@ def setup(self):
 
         if self.get_option("all_logs"):
             self.add_copy_spec([
-                "/var/log/aodh/*",
-                "/var/log/httpd/aodh*",
```
pmoravec, line 37: Same comment like to `collectd` plugin.

sos/report/plugins/openstack_aodh.py
```diff
@@ -33,13 +33,13 @@ def setup(self):
...
+                "/var/log/containers/aodh/*",
+                "/var/log/containers/httpd/aodh-api/*",
             ])
         else:
             self.add_copy_spec([
-                "/var/log/aodh/*.log",
-                "/var/log/httpd/aodh*.log",
```
pmoravec, line 42: Ditto.

sos/report/plugins/openstack_ceilometer.py
```diff
@@ -26,11 +26,11 @@ def setup(self):
         # Ceilometer
         if self.get_option("all_logs"):
             self.add_copy_spec([
-                "/var/log/ceilometer/*",
+                "/var/log/containers/ceilometer/*",
             ])
         else:
             self.add_copy_spec([
-                "/var/log/ceilometer/*.log",
```
pmoravec, line 33: .. and also here. When `OpenStack Ceilometer` switched to *only* containerised deployment? What is the latest OSP version that logs into `/var/log/ceilometer/`?

mrunge: The last OSP version logging to /var/log/ceilometer was OSP10, which is out of support since December 15, 2021, see https://access.redhat.com/support/policy/updates/openstack/platform/non-current

sos/report/plugins/collectd.py
mrunge: /var/lib/config-data/collectd should be a typo, it should have been /var/lib/config-data/puppet-generated/collectd in the first place. 
container based setups are the only setups used in OSP since OSP 10 was discontinued in December 2021

sos/report/plugins/openstack_aodh.py
mrunge: same here: the only version using /var/log/... was OSP 10, which is out of support since Dec 2021, see https://access.redhat.com/support/policy/updates/openstack/platform/non-current

sos/report/plugins/openstack_ceilometer.py
TurboTurtle: As these changes are changing the base plugin, we need to consider the non-RH deployments that may consume this change.

Do we have an indication on upstream's stock configuration, or if there is (sizeable) usage of the older version(s)? @ddstreet @slashdd any ideas on Ubuntu based deployments for this?

mrunge: I should probably add the changed path below to the Red Hat section. The path mentioned there does not exist anymore.

sos/report/plugins/openstack_aodh.py
TurboTurtle: Since we're doing the override to `/var/log/containers` in the `RedHatOpenStackAodh` class, these should be left as their originals, yes?

mrunge:
Thank you for your reviews. I have addressed the raised concerns here.

pmoravec (approved):
Assuming all supported versions of OSP do log under `/var/log/containers` only, ACK (just fyi https://github.com/openstack/aodh/search?q=%2Fvar%2Flog%2Faodh still *refers* to the old log path).

TurboTurtle (commented):
Code change looks ok, but please squash the 2 fixup commits into the first one.

mrunge:
Thank you for the review, squashed and force-pushed.

That highly depends on the projects style. Some don't want to loose the review context, some prefer single commits. I can live with both; in collectd, I am using the "squash and merge" strategy hidden under the "merge" button.

sos/report/plugins/openstack_aodh.py
TurboTurtle: Sorry for the noise, reading through this again.

We're changing it to the `/var/log/containers` path here for the base plugin, which means for *all* distributions. We're then _again_ specifying it in the RH-specific plugin. So we have the same collection defined in two spots now.

We either would have the `/var/log/containers` collection in the base plugin and *not* have the override in `RedHatOpenStackAodh.setup()`, _or_ we would have the base plugin stay on the "old" `/var/log` paths and keep the override in `setup()`.

Without acks from other distros, I'd say we should do the latter. Have the `setup()` override like you have now, but keep the old paths in the base plugin so we don't inadvertently break collections on non-RH environments that aren't using a containerized setup.

sos/report/plugins/openstack_ceilometer.py
dosaboy: Yeah this won't work with Ubuntu Openstack deployments that are not using /var/log/containers. @mrunge would it be possible to move all these changes under the RedHat* implementation class so that it does not impact other distros?

mrunge:
I'll submit a new commit asap, I don't want to break anyone here.

dosaboy (changes requested):
@mrunge the changes in OpenStackAodh will not work with Ubuntu Openstack deployments since they are not using the /var/log/containers path. I think either we need to keep both paths or  those changes need to move to a distro-specific implementation.

dosaboy:
@mrunge dunno if its useful but here's an example of a different plugin that has support for both paths - https://github.com/sosreport/sos/blob/1682e78d2a63aa771a8209b46382c109bd78e84d/sos/report/plugins/openstack_placement.py#L44

mrunge:
Okay, I've added a Red Hat setup to openstack_aodh and (compared to my previous commits) reverted to the status we had before *this* PR, the change also adds /var/log/containers/ceilometer for ceilometer (in the Red Hat section). Also I've added /var/log/containers/collectd/ for the collectd plugin. This should not harm any other setups.

mrunge:
Anything else I can do to help this getting forward?

pmoravec (approved):

mrunge:
Thank you!

PR #2872, opened 2022-03-02
smartcards: Add more related files and packages

### LINKED ISSUE #2871: smartcard plugin is using tools removed/not available from rhel8 and newer
The pam_pkcs11 was last available in RHEL7 and is no longer available in RHEL8 and newer so there is no point for checking it and its configuration files in:

https://github.com/sosreport/sos/blob/b96926f151da984491de3082af7d06ecb07fc517/sos/report/plugins/smartcard.py#L21-L22

The `pcsc-tools` package above is not available in RHEL (only from EPEL). Instead, this list should contain `pcsc-lite` and `pcsc-lite-ccid` packages.

Also the following commands/path do not have any significance so I propose to remove them (the `pklogin_finder` is part of pam_pkcs11`, which is not available in RHEL8 and newer). The respective replacement would be some command from sssd, but most of the sssd logs are already collected so I do not think we need a separate command:

https://github.com/sosreport/sos/blob/b96926f151da984491de3082af7d06ecb07fc517/sos/report/plugins/smartcard.py#L32-L33
https://github.com/sosreport/sos/blob/b96926f151da984491de3082af7d06ecb07fc517/sos/report/plugins/smartcard.py#L40

The `pcsc_scan` is a tool that is shipped in `pcsc-tools` package, which is available in EPEL only so I suggest to remove it:

https://github.com/sosreport/sos/blob/b96926f151da984491de3082af7d06ecb07fc517/sos/report/plugins/smartcard.py#L34

Let me know if this sounds ok to you or I should open a PR with the proposed changes.

Note, that these are still needed for RHEL7, but I think for RHEL7 you have a different branch.

Jakuje:
Fixes #2871

### DIFF
--- a/sos/report/plugins/smartcard.py
+++ b/sos/report/plugins/smartcard.py
@@ -19,7 +19,9 @@ class Smartcard(Plugin, RedHatPlugin):
     profiles = ('security', 'identity', 'hardware')
 
     files = ('/etc/pam_pkcs11/pam_pkcs11.conf',)
-    packages = ('pam_pkcs11', 'pcsc-tools', 'opensc')
+    # The pam_pkcs11 is available only in RHEL7
+    packages = ('pam_pkcs11', 'pcsc-tools', 'opensc', 'pcsc-lite',
+                'pcsc-lite-ccid')
 
     def setup(self):
         self.add_copy_spec([
--- a/sos/report/plugins/smartcard.py
+++ b/sos/report/plugins/smartcard.py
@@ -28,7 +28,9 @@ class Smartcard(Plugin, RedHatPlugin):
             "/etc/reader.conf",
             "/etc/reader.conf.d/",
             "/etc/pam_pkcs11/",
-            "/etc/opensc-*.conf"
+            "/etc/opensc-*.conf",
+            "/etc/pkcs11/modules/*.module",
+            "/usr/share/p11-kit/modules/*.module"
         ])
         self.add_cmd_output([
             "pklogin_finder debug",


pmoravec (approved):

sos/report/plugins/smartcard.py
```diff
@@ -19,14 +19,17 @@ class Smartcard(Plugin, RedHatPlugin):
     profiles = ('security', 'identity', 'hardware')
 
     files = ('/etc/pam_pkcs11/pam_pkcs11.conf',)
-    packages = ('pam_pkcs11', 'pcsc-tools', 'opensc')
+    # The pam_pkcs11 is available only in RHEL7
+    packages = ('pam_pkcs11', 'pcsc-tools', 'opensc', 'pcsc-lite', 'pcsc-lite-ccid')
```
pmoravec, line 23: There is a PEP error / too long line:

    sos/report/plugins/smartcard.py:23:80: E501 line too long (84 > 79 characters)

Please split the line like e.g. https://github.com/sosreport/sos/blob/main/sos/report/plugins/openvswitch.py#L365-L367 and verify via e.g. `pycodestyle sos/report/plugins/smartcard.py`.

Jakuje: I was trying to set up the git commit hooks according to the guide, but it is not updated for 5 years and no longer reasonably working -- can you update the scripts to use the pycodestyle instead of the pep8 at least?

https://github.com/sosreport/sos/wiki/Using-git-hooks-to-automate-tests

Jakuje: Fixed with current version. The tests did not work for me locally on Fedora so I gave up and hope they will run here somehow.

pmoravec (approved):

TurboTurtle (approved):

PR #2877, opened 2022-03-10
[ceph_mon] Replace an obsolete command

pponnuvel:
mon_status command is no longer available since Octopus
release. Replaced it with its ceph tell equivalent that
works in both older as well as newer Ceph releases.

Signed-off-by: Ponnuvel Palaniyappan <pponnuvel@gmail.com>

---

### DIFF
--- a/sos/report/plugins/ceph_mon.py
+++ b/sos/report/plugins/ceph_mon.py
@@ -49,7 +49,6 @@ class CephMON(Plugin, RedHatPlugin, UbuntuPlugin):
             "ceph config log",
             "ceph config generate-minimal-conf",
             "ceph config-key dump",
-            "ceph mon_status",
             "ceph osd metadata",
             "ceph osd erasure-code-profile ls",
             "ceph osd crush show-tunables",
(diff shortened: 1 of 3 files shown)


sos/report/plugins/ceph_mon.py
```diff
@@ -49,7 +49,7 @@ def setup(self):
             "ceph config log",
             "ceph config generate-minimal-conf",
             "ceph config-key dump",
-            "ceph mon_status",
+            "ceph tell mon.* mon_status",
```
TurboTurtle, line 52: What does `mon.*` typically expand out to in this context? Or rather, what is the expected output of this command (and is that any different from the syntax being replaced?)

TurboTurtle:
Restarted the Ubuntu stage one test - timeout is a known issue and not related to this change.

sos/report/plugins/ceph_mon.py
pponnuvel: That (`mon.*`) would send the `mon_status` to all the mon daemons. While it'd work, I realzied it could more than the *current* mon's status (aka, where the command was issued). So I modified to work out the specific mon daemons' IDs.

Similarly, I modified the other 2 daemons too as `--id` doesn't necessarily have to be in any specific order (kind of of future proofing if args are re-ordered and/or more args are inserted at the beginning of the cmdline for example). It now looks at index of `--id` and then picks the next arg.

sos/report/plugins/ceph_mon.py
```diff
@@ -90,6 +89,31 @@ def setup(self):
             "ceph %s --format json-pretty" % s for s in ceph_cmds
         ], subdir="json_output", tags="insights_ceph_health_detail")
 
+        ceph_tell_cmds = [
+            "mon_status"
+        ]
```
pmoravec, line 94: Until you plan to enhance the list by some other subcommands in some future, this is bit overkill. But even that is ACK from me.

pmoravec (approved):

pmoravec (approved):

TurboTurtle (approved):

PR #2880, opened 2022-03-14
[distros] Pass --upload-directory to SFTP uploads

pmoravec:
Apart of that, print a heads up message during SFTP upload when --upload-directory
was set.

Resolves: #2880

---

### DIFF
--- a/sos/policies/distros/__init__.py
+++ b/sos/policies/distros/__init__.py
@@ -577,7 +577,10 @@ class LinuxPolicy(Policy):
         :returns:       Filename as it will exist on the SFTP server
         :rtype:         ``str``
         """
-        return self.upload_archive_name.split('/')[-1]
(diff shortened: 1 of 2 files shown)


TurboTurtle:
Since we leverage an open sftp session for these uploads, I think we should instead try to use any passed `upload_directory`.

We could either do this by passing a `cd` command to the sftp session, or perhaps prepend the upload directory via `Policy._get_sftp_upload_name()`.

Or did we avoid supporting this originally for some specific reason that I'm forgetting?

pmoravec:
I am not aware of any particular reason why not use it. So let enhance the functionality.

I chosen pre-pending remote filename by the path, less error-prone than an `pexpect` interactive call.

TurboTurtle (approved):

PR #2881, opened 2022-03-15
[md] Restrict data capture to raid members

jcastill:
When generating sos report, md-raid plugin won't capture the
metadata information from the actual RAID device. Also, it
tries to fetch metadata information for all block devices
available in the system.
Sometimes RAID is created using the disk partition,
but sos won't collect metadata from those devices.

This patch takes the list of raid members from lsblk,
and captures the output only for these devices.

Fixes: RHBZ#2062283
Fixes: #2881 

---

### DIFF
--- a/sos/report/plugins/md.py
+++ b/sos/report/plugins/md.py
@@ -18,8 +18,13 @@ class Md(Plugin, IndependentPlugin):
 
     def setup(self):
         self.add_cmd_output("mdadm -D /dev/md*")
-        self.add_blockdev_cmd("mdadm -E %(dev)s",
-                              blacklist=['ram.*', 'zram.*'])
+        mdadm_members = self.exec_cmd("lsblk -o NAME,FSTYPE -r")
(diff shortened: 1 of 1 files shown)


jcastill:
This one has been a bit tricky, because add_blockdev_cmd() returns all devices, but using whitelists/blacklists couldn't provide the granularity needed (some times we will use partitions, and/or sd* devices, hd*, nvme*, etc). So I had two possibilities:
Get the output of 'blkid', where the raid members are listed, or
Get the output of 'lsblk', where the same happens. 
I opted for 'lsblk' because the output, with the options I used, made it easier to manage the list, whereas using 'blkid' needed to do some post processing to remove separators. 

I've performed some basic tests but the original reporter in bugzilla is going to do some more to verify this PR doesn't break anything.

TurboTurtle (approved):

pmoravec (approved):

PR #2887, opened 2022-03-16
Sos pmoravec dont obfuscate tmpdir path

pmoravec:
Two commits preventing to apply cleaner obfuscation to the working directory path itself. Creating files or moving directories must happen within the (non-obfuscated) working dir.

The `archive` commit changes purpose of `self._name` variable from `/var/tmp/sos.oq9ae6gj/sosreport-pmoravec-rhel8-2022-03-16-qzazpxi` to just `sosreport-pmoravec-rhel8-2022-03-16-qzazpxi`. Since the full path is already captured in `_archive_root` variable.

---

### DIFF
--- a/sos/archive.py
+++ b/sos/archive.py
@@ -135,6 +135,9 @@ class FileCacheArchive(Archive):
     def __init__(self, name, tmpdir, policy, threads, enc_opts, sysroot,
                  manifest=None):
         self._name = name
+        # truncate the name just relative to the tmpdir in case of full path
+        if os.path.commonprefix([self._name, tmpdir]) == tmpdir:
+            self._name = os.path.relpath(name, tmpdir)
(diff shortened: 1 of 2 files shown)


pmoravec:
Hm, the failed unittests **do** fail also on current main, independently on this PR.

sos/archive.py
```diff
@@ -135,6 +135,9 @@ class FileCacheArchive(Archive):
     def __init__(self, name, tmpdir, policy, threads, enc_opts, sysroot,
                  manifest=None):
         self._name = name
+        # truncate the name just relative to the tmpdir in case of full path
+        if os.path.commonprefix([self._name, tmpdir]) == tmpdir:
```
pmoravec, line 139: This condition is required due to `TarFileArchive.add(`testfile`) where we add files without their path.

PR #2888, opened 2022-03-17
Adding ability to use the localhost.kubeconfig 

vwalek:
Opening the PR to implement the usage of localhost.kubeconfig present only on masters to use it when collecting data from the API.

Localhost kubeconfig is file containing the cert/key for system:admin user.

Signed-off-by: Vladislav Walek <22072258+vwalek@users.noreply.github.com>

---

### DIFF
--- a/sos/report/plugins/openshift.py
+++ b/sos/report/plugins/openshift.py
@@ -53,12 +53,19 @@ class Openshift(Plugin, RedHatPlugin):
     profiles = ('openshift',)
     packages = ('openshift-hyperkube',)
 
+    master_localhost_kubeconfig = (
+        '/etc/kubernetes/static-pod-resources/'
+        'kube-apiserver-certs/secrets/node-kubeconfigs/localhost.kubeconfig'
+        )
+
     option_list = [
         PluginOpt('token', default=None, val_type=str,
                   desc='admin token to allow API queries'),
+        PluginOpt('kubeconfig', default=None, val_type=str,
+                  desc='Path to a locally available kubeconfig file'),
         PluginOpt('host', default='https://localhost:6443',
                   desc='host address to use for oc login, including port'),
-        PluginOpt('no-oc', default=False, desc='do not collect `oc` output'),
+        PluginOpt('no-oc', default=True, desc='do not collect `oc` output'),
         PluginOpt('podlogs', default=True, desc='collect logs from each pod'),
         PluginOpt('podlogs-filter', default='', val_type=str,
                   desc='only collect logs from pods matching this pattern'),
@@ -73,6 +80,10 @@ class Openshift(Plugin, RedHatPlugin):
         """Check to see if we can run `oc` commands"""
         return self.exec_cmd('oc whoami')['status'] == 0
 
+    def _check_localhost_kubeconfig(self):
+        """Check if the localhost.kubeconfig exists with system:admin user"""
+        return self.path_exists(self.get_option('kubeconfig'))
+
     def _check_oc_logged_in(self):
         """See if we're logged in to the API service, and if not attempt to do
         so using provided plugin options
@@ -80,8 +91,38 @@ class Openshift(Plugin, RedHatPlugin):
         if self._check_oc_function():
             return True
 
-        # Not logged in currently, attempt to do so
+        if self.get_option('kubeconfig') is None:
+            # If admin doesn't add the kubeconfig
+            # use default localhost.kubeconfig
+            self.set_option(
+                'kubeconfig',
+                self.master_localhost_kubeconfig
+            )
+
+        # Check first if we can use the localhost.kubeconfig before
+        # using token. We don't want to use 'host' option due we use
+        # cluster url from kubeconfig. Default is localhost.
+        if self._check_localhost_kubeconfig():
+            self.set_default_cmd_environment({
+                'KUBECONFIG': self.get_option('kubeconfig')
+            })
+
+            oc_res = self.exec_cmd(
+                "oc login -u system:admin "
+                "--insecure-skip-tls-verify=True"
+            )
+            if oc_res['status'] == 0 and self._check_oc_function():
+                return True
+
+            self._log_warn(
+                "The login command failed with status: %s and error: %s"
+                % (oc_res['status'], oc_res['output'])
+            )
+            return False
+
+        # If kubeconfig is not defined, check if token is provided.
         token = self.get_option('token') or os.getenv('SOSOCPTOKEN', None)
+
         if token:
             oc_res = self.exec_cmd("oc login %s --token=%s "
                                    "--insecure-skip-tls-verify=True"


sos/report/plugins/openshift.py
```diff
@@ -82,6 +89,18 @@ def _check_oc_logged_in(self):
...
         token = self.get_option('token') or os.getenv('SOSOCPTOKEN', None)
+
+        # Check first if we can use the localhost.kubeconfig before
+        # using token. We don't need to use 'host' option due we
+        # use localhost from kubeconfig
+        if self._check_localhost_kubeconfig():
+            self.exec_cmd(env={'KUBECONFIG': self.master_localhost_kubeconfig})
+
+            oc_res = self.exec_cmd("oc login -u system:admin \
+                --insecure-skip-tls-verify=True")
```
TurboTurtle, line 100: We talked a little about this over GChat so I'm going to rehash a bit of that now so we can get some wider input. However, one note is that I should have been more specific here - I meant adding the `env` parameter to the `self.exec_cmd()` call for `oc login`, e.g.:

```python
self.exec_cmd(
    'oc login -u system:admin --insecure-skip-tls-verify=True',
    env={'KUBECONFIG': self.master_localhost_kubeconfig}
)
```

That still brings us around to the issue of needing to have the `KUBECONFIG` env var set for any subsequent `oc` commands.

Generally speaking there are only 2 solutions at the top level here:

1. Set the env var on the system
2. Pass the env var with each `add_cmd_output()` command.

Option 1 is a non-starter for me, as we have to maintain the guarantee of "no system changes". I understand that in this specific context and use case that is not as large of a concern as it is normally, but I don't want to set a precedent (or a precedent for an exception), especially since we have historically always rejected directly setting env vars in sos in the past.

Option 2 gives us 3 routes. Either we add `--kubeconfig=$path` to each command and then have to do a lot of tinkering with `suggest_filename`, which is also a non-starter in my opinion. Or we pass `env={'KUBECONFIG': self.master_localhost_kubeconfig}` to every `add_cmd_output()` call, which is _very_ ugly and error prone, especially for future maintenance.

I then had a 3rd idea of allowing `Plugin` to set a "default environment" configuration, which would get automatically passed via `env` to every `add_cmd_output()` call. Sort of similar to how we have `Plugin.set_predicate()` today.

I think this is a workable solution and overall a Good Thing(tm) for plugins, but I'd like to hear if there are any counter-arguments or better ideas first.

To be clear, my thought is something like this:


```python

class MyPlugin(Plugin):

    def setup(self):
        self.set_default_environment({'KUBECONFIG': self.master_localhost_kubeconfig})
        self.add_cmd_output('oc something') 
```
And in `Plugin.add_cmd_output()` we add logic to pass that dict to `sos_get_command_output()` via the `env` parameter.

vwalek: Hey @TurboTurtle, thx for summarizing our gchat discussion. I totally agree with your 3rd idea. Having a env variable for the sos report would be the best way to do that. I agree, options 1 and 2 are not feasible.

TurboTurtle: I've posted #2894 which should fit this usecase. If you want to try that out with a working cluster ontop of this branch, you'd do something like this:

```python
if self._check_localhost_kubeconfig():
            self.set_default_cmd_environment({'KUBECONFIG': self.master_localhost_kubeconfig})

            oc_res = self.exec_cmd("oc login -u system:admin \
                --insecure-skip-tls-verify=True")
```

TurboTurtle:
I realize we're still refining this, but should we expose this via a plugopt for the kubeconfig location? And if it is not set, but `no-oc` is False, _then_ use this location as a default?

Just a thought that I wanted to write out before I forgot about it.

E.G.

```python
PluginOpt('kubeconfig', default=None, val_type=str,
          desc='Path to a locally available kubeconfig file')

if self.get_option('kubeconfig') is None:
    # using `self.path_join()` here will properly wrap this to `/host/` when run in a container
    self.set_option('kubeconfig', self.path_join(self.master_localhost_kubeconfig))

[...]
# later, after the check against `no-oc`...
self.set_default_cmd_environment({'KUBECONFIG': self.get_option('kubeconfig')})
```

sos/report/plugins/openshift.py
vwalek: Hey @TurboTurtle , yeah, I think it fits perfectly to our use case, having the default variable defined. I will check it on my lab.

vwalek:
> I realize we're still refining this, but should we expose this via a plugopt for the kubeconfig location? And if it is not set, but `no-oc` is False, _then_ use this location as a default?
> 
> Just a thought that I wanted to write out before I forgot about it.
> 
> E.G.
> 
> ```python
> PluginOpt('kubeconfig', default=None, val_type=str,
>           desc='Path to a locally available kubeconfig file')
> 
> if self.get_option('kubeconfig') is None:
>     # using `self.path_join()` here will properly wrap this to `/host/` when run in a container
>     self.set_option('kubeconfig', self.path_join(self.master_localhost_kubeconfig))
> 
> [...]
> # later, after the check against `no-oc`...
> self.set_default_cmd_environment({'KUBECONFIG': self.get_option('kubeconfig')})
> ```

Yeah, maybe it would be better to make it as plugin opt if the location changes in any way.
If the sos runs that it defines some options as default, then we can skip the definition of the file in the code directly and use the plugin opt and default value from the config. (would be better if this changes in future, to only change the config and not the code).

vwalek:
Hey @TurboTurtle - I've tried to run it, unfortunately it doesn't work for some reason. It won't pick the kubeconfig and it will completely skip the oc with kubeconfig. I need to dig in more, but for some reason it won't work.

TurboTurtle:
> Hey @TurboTurtle - I've tried to run it, unfortunately it doesn't work for some reason. It won't pick the kubeconfig and it will completely skip the oc with kubeconfig. I need to dig in more, but for some reason it won't work.

The issue is in the string break here - https://github.com/sosreport/sos/blob/c57187bed8fae03b553125b9b02a7b08cbe010d9/sos/report/plugins/openshift.py#L56-#L57

This ultimately gets passed to `path_exists()` as `/host/etc/kubernetes/static-pod-resources/         kube-apiserver-certs/secrets/node-kubeconfigs/localhost.kubeconfig` (there are many spaces after `static-pod-resources/`, which GH comment formatting is stripping). Since this all eventually ends up in an `os.path.exists()` call:

```python
>>> os.path.exists('/host/etc/kubernetes/static-pod-resources/        kube-apiserver-certs/secrets/node-kubeconfigs/localhost.kubeconfig')
False
>>> os.path.exists('/host/etc/kubernetes/static-pod-resources/kube-apiserver-certs/secrets/node-kubeconfigs/localhost.kubeconfig')
True
>>> 
```

So should just need to reformat the string breaking, e.g.:


```python
self.master_localhost_kubeconfig = (
    '/etc/kubernetes/static-pod-resources/kube-apiserver-certs/'
    'secrets/node-kubeconfigs/localhost.kubeconfig'
)
```

If I make that change, then the default command environment variable seems to be functioning for passing the kubeconfig file to each command.

TurboTurtle:
Oh, so this is a fun quirk. There is no need to call `self.path_join()` on the kubeconfig location when setting the `KUBECONFIG` env var for the plugin, because the plugin will automatically chroot to the sysroot location. `self.path_exists()` does already account for the sysroot, so no changes there.

Basically, do this for setting the env var:

```python
self.set_default_cmd_environment(
    {'KUBECONFIG': self.master_localhost_kubeconfig}
)
```

Otherwise we'll end up trying to set the config file to `/host` _from_ `/host`, which won't work.

vwalek:
> Oh, so this is a fun quirk. There is no need to call `self.path_join()` on the kubeconfig location when setting the `KUBECONFIG` env var for the plugin, because the plugin will automatically chroot to the sysroot location. `self.path_exists()` does already account for the sysroot, so no changes there.
> 
> Basically, do this for setting the env var:
> 
> ```python
> self.set_default_cmd_environment(
>     {'KUBECONFIG': self.master_localhost_kubeconfig}
> )
> ```
> 
> Otherwise we'll end up trying to set the config file to `/host` _from_ `/host`, which won't work.

Yeah, it does not work. The KUBECONFIG variable is set, but the `oc` won't pick it up for some reason. 
https://github.com/vwalek/sos/blob/localhost-kubeconfig-testing/sos/report/plugins/openshift.py#L121-L125

The output is:

```
[plugin:openshift] does the kubeconfig exists? True
[plugin:openshift] Something wrong happened when running the oc with KUBECONFIG. {'KUBECONFIG': '/etc/kubernetes/static-pod-resources/kube-apiserver-certs/secrets/node-kubeconfigs/localhost.kubeconfig'}
[plugin:openshift] output of oc command {'status': 1, 'output': 'error: dial tcp [::1]:6443: connect: connection refused - verify you have provided the correct host and port and that the server is currently running.\n', 'truncated': False}
```

The `oc` still can't use the kubeconfig.

TurboTurtle:
How are you running `sos` here? Are you directly running this on an OCP master? Are you running in a container? Is that container using the `registry.redhat.io/rhel8/support-tools` image and launched via the `RUN` label?

Fixing the spacing issue in `self.master_localhost_kubeconfig` and calling `self.set_default_cmd_environment()` with that location results in a successful collection for me on my test cluster when running in a support-tools based container (which is the supported way to run sos on RHCOS). Also, for reference the collection took approx. 15 minutes from start to finish.

TurboTurtle:
Ah, didn't see the update. Good to hear this is working for you now with that change.

vwalek:
@TurboTurtle it is just me being stupid in the way how the toolbox container is started. I will commit my changes to this branch.

vwalek:
@TurboTurtle PTAL

sos/report/plugins/openshift.py
```diff
@@ -73,6 +80,10 @@ def _check_oc_function(self):
         """Check to see if we can run `oc` commands"""
         return self.exec_cmd('oc whoami')['status'] == 0
 
+    def _check_localhost_kubeconfig(self):
+        """Check if the localhost.kubeconfig exists with system:admin user"""
+        return self.path_exists(self.master_localhost_kubeconfig)
```
TurboTurtle, line 85: This should instead be checking the `kubeconfig` option value. Otherwise, we're still gating on a static file even if a user uses a different config file.

TurboTurtle (commented):
A little re-arranging I think is in order, but beyond that this looks correct to me.

You'll want to rebase ontop of current `main` to pull in #2894 when you repush.

Also, please reformat the commit message to align with the [Contribution Guidelines](https://github.com/sosreport/sos/wiki/Contribution-Guidelines)

sos/report/plugins/openshift.py
```diff
@@ -82,6 +93,37 @@ def _check_oc_logged_in(self):
...
+        # use localhost from kubeconfig
+        if self._check_localhost_kubeconfig():
+
+            if self.get_option('kubeconfig') is None:
+                # using `self.path_join()` here will properly wrap
+                # this to `/host/` when run in a container
+                self.set_option(
+                    'kubeconfig',
+                    self.master_localhost_kubeconfig
+                )
```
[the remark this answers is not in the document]
TurboTurtle: Oh, and let's drop the comment about `path_join()` since we're not actually using it. Don't want to confuse anyone reading this in the future.

vwalek:
@TurboTurtle the failed test doesn't run the openshift plugin, is it desired?

pmoravec:
> @TurboTurtle the failed test doesn't run the openshift plugin, is it desired?

I do see just skipped tests but no failed one.

pmoravec (approved):

TurboTurtle (approved):

vwalek:
> > @TurboTurtle the failed test doesn't run the openshift plugin, is it desired?
> 
> I do see just skipped tests but no failed one.

Yeah, when I wrote that comment one of the tests failed, but it is ok now. But it was not related because it didn't run the `openshift` plugin.
So ok now.

PR #2891, opened 2022-03-21
[pacemaker] Update collect cluster profile for pacemaker

TurboTurtle:
This 2-patch set overhauls how the `pacemaker` profile enumerates nodes.

Rather than rely on `pcs status` output which is liable to change (and apparently has changed recently), instead use `crm_mon` which is what the former called under the hood anyways. If that fails for any reason, directly reference `/etc/corosync/corosync.conf`. Note however that this fallback method does not let us separate online nodes from offline ones.

Further, due to pacemaker's ability to have node names reported that are not the node's actual resolvable hostnames (they can be named arbitrarily by sysadmins) there is a potential for duplicate collections of the primary node. As such, add a toggle to cluster profiles via the `strict_node_list` class attr that can be used to tell collect to not add the local host to the node list (as we need to do with several other profiles).

---

### DIFF
--- a/sos/collector/clusters/pacemaker.py
+++ b/sos/collector/clusters/pacemaker.py
@@ -8,7 +8,11 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
+import re
+
 from sos.collector.clusters import Cluster
+from setuptools._vendor.packaging import version
+from xml.etree import ElementTree
 
 
 class pacemaker(Cluster):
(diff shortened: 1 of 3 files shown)


sos/collector/clusters/pacemaker.py
```diff
@@ -18,42 +22,80 @@ class pacemaker(Cluster):
...
+            )
+        return self.nodes
+
+    def get_nodes_from_crm(self):
+        """
+        Try to parse crm_mon output for node list and status.
+        """
+        xmlopt = '--output-as=xml'
+        # older pacemaker had a different option for xml output
+        _ver = self.exec_primary_cmd('crm_mon --version')
```
nrwahl2, line 62: You might want to use XML for this too, just in case.
```
# crm_mon --version --output-as=xml
<pacemaker-result api-version="2.12" request="crm_mon --version --output-as=xml">
  <version program="Pacemaker" version="2.1.0-8.el8" author="Andrew Beekhof" build="7c3f660707" features="agent-manpages cibsecrets compat-2.0 corosync-ge-2 default-concurrent-fencing default-sbd-sync generated-manpages monotonic nagios ncurses remote systemd"/>
  <status code="0" message="OK"/>
</pacemaker-result>
```

nrwahl2: Never mind, I just realize that would be circular.

So, maybe a `try`/`except` block instead of a version check.

sos/collector/clusters/pacemaker.py
```diff
@@ -18,42 +22,80 @@ class pacemaker(Cluster):
...
+                self.nodes.append(_node['name'])
+            elif self.get_option('offline') and _node['online'] == 'false':
+                self.nodes.append(_node['name'])
+
+    def get_nodes_from_corosync(self):
+        """
+        As a fallback measure, read corosync.conf to get the node list. Note
+        that this prevents us from separating online nodes from offline nodes.
+        """
+        self.log_warn("WARNING: using corosync.conf as a fallback measure to "
```
nrwahl2, line 94: Maybe don't say fallback if the `only-corosync` option is enabled.

TurboTurtle: Fair point. I'll change this to still print a warning that using corosync means no online/offline options, but we can say we're falling back to it specifically when parsing `crm_mon` fails.

sos/collector/clusters/pacemaker.py
```diff
@@ -18,42 +22,80 @@ class pacemaker(Cluster):
...
+                      "enumerate nodes. This method is unable to separate "
+                      "online nodes from offline nodes.")
+        cc = self.primary.read_file('/etc/corosync/corosync.conf')
+        nodes = re.findall(r'((\sring0_addr:)(.*))', cc)
```
nrwahl2, line 98: These should be resolvable (from **within** the cluster), but may not be the node names from pacemaker's perspective. They may not be names at all -- they could be IP addresses. If that doesn't create any issues for the collection or resulting report bundle (i.e., if we're using these addrs only to connect to each target node), then that's fine. I haven't had a chance to test.

TurboTurtle: IP addresses are fine. We use these to make the connections to the nodes and in our initial logging for those connections only.

sos/collector/clusters/pacemaker.py
nrwahl2: Tl;dr, probably fine.

I was concerned about futureproofing, not pythonicity. The text output format of `crm_mon --version` is subject to change (though it doesn't seem particularly likely to change). If it changes, then we may fail to parse out the version that we need later to determine `--output-as=xml` versus `--as-xml`.

A solution would be to get XML output from `crm_mon --version`, but that becomes circular... you'd need the version in order to know which "as xml" option to use in `crm_mon --version`.

That's why I suggested skipping the whole version check and just using try/except.

In practice, it's unlikely to matter. If the `crm_mon --version` format has changed, then that means we're on a newer version and so we won't hit the `version < 2.3` condition to choose `--as-xml`. The only way it would matter is if `crm_mon --version` format changes and then the `--output-as=xml` syntax also changes.

PR #2894, opened 2022-03-22
[Plugin] Allow plugins to define default command environment vars

TurboTurtle:
Adds the ability for plugins to define a default set of environment vars
to pass to all commands executed by the plugin. This may be done either
via the new `set_default_cmd_environment()` or
`add_default_cmd_environment()` methods. The former will override any
previously set values, whereas the latter will add/update/modify any
existing values.

---

### DIFF
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -571,6 +571,7 @@ class Plugin():
         self.manifest = None
         self.skip_files = commons['cmdlineopts'].skip_files
         self.skip_commands = commons['cmdlineopts'].skip_commands
+        self.default_environment = {}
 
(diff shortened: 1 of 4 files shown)


TurboTurtle:
Note: in the current design the environment variables are compiled at the time of execution, not at the time of the command being added to the list of commands to collect.

Given we only have a single use case for this at this point (passing a kubeconfig file for the `openshift` plugin), I think this is alright. That said, if there is a desire to change this to be done in the order commands are added let's hash that out now before this gets merged.

pmoravec (approved):

TurboTurtle:
Added some tests for this to the suite.

pmoravec (approved):
Nice test (incl. the message in it :) ).

PR #2896, opened 2022-03-28
[presets] Adjust OCP preset options

TurboTurtle:
Adjust the options used by the 'ocp' preset to better reflect the
current collection needs and approach.

This includes disabling the `cgroups` plugin due to the large amount of
mostly irrelevant data captured due to the high number of containers
present on OCP nodes, ensuring the `--container-runtime` option is set
to `crio` to align container-based collections, disabling HTML report
generation and increasing the base log size rather than blindly enabling
all-logs.

---

### DIFF
--- a/sos/presets/redhat/__init__.py
+++ b/sos/presets/redhat/__init__.py
@@ -36,10 +36,15 @@ RHOSP_OPTS = SoSOptions(plugopts=[
 
 RHOCP = "ocp"
 RHOCP_DESC = "OpenShift Container Platform by Red Hat"
-RHOCP_OPTS = SoSOptions(all_logs=True, verify=True, plugopts=[
-                             'networking.timeout=600',
-                             'networking.ethtool_namespaces=False',
-                             'networking.namespaces=200'])
+RHOCP_OPTS = SoSOptions(
+    verify=True, skip_plugins=['cgroups'], container_runtime='crio',
+    no_report=True, log_size=100,
+    plugopts=[
+        'crio.timeout=600',
+        'networking.timeout=600',
+        'networking.ethtool_namespaces=False',
+        'networking.namespaces=200'
+    ])
 
 RH_CFME = "cfme"
 RH_CFME_DESC = "Red Hat CloudForms"


pmoravec (approved):

npinaeva:
Hey Jake, can you please explain what log-size option applies to and what happens when it's exceeded?

pmoravec:
> Hey Jake, can you please explain what log-size option applies to and what happens when it's exceeded?

By default, `sos` collects 25MB of each command output or `copyspec` (i.e. having `self.add_copy_spec(['/foo/dir/*', '/bar/*log'])`, then 25MB applies **independently** on (mtime-sorted) `/foo/dir/*` and independently on `/bar/*log` files). You can overwrite the default via plugin option `-k plugin.log_size=123`. Further, some collections have increased sizelimit explicitly - then this sizelimit is used every time.

If size of a `copyspec` or a `cmdoutput` exceeds `logsize`, we collect tail of the cmdoutput or tail of the "current file from copyspec that is on the sizelimit" (recall: files are collected from the newest mtime to older ones). The collected data are stored under sos_strings directory and the usual location is a symlink there.

npinaeva:
My main concern here is size limiting for directories, because if there is a huge file in the folder with the latest mtime, I will miss all the other files from the same folder. So maybe limiting size per file will help here? Or any other suggestions to solve this?

pmoravec:
> My main concern here is size limiting for directories, because if there is a huge file in the folder with the latest mtime, I will miss all the other files from the same folder. So maybe limiting size per file will help here? Or any other suggestions to solve this?

Do you know *file* patterns you would like to collect? I.e. dont call `add_copy_spec('dir')` but `add_copy_spec(['dir/foo*log*', 'dir/bar*log*',...])`?

npinaeva:
I am actually talking about other plugins we rely on, e.g. https://github.com/sosreport/sos/blob/main/sos/report/plugins/openvswitch.py#L71, https://github.com/sosreport/sos/blob/main/sos/report/plugins/networking.py#L56 and others that may have directories as copy_spec (so changing that, I guess, doesn't make a lot of sense).
Also, for this file patterns instead of directory, will size limits apply per file?

pmoravec:
> Also, for this file patterns instead of directory, will size limits apply per file?

Per filepattern. See https://github.com/sosreport/sos/blob/main/sos/report/plugins/__init__.py#L1696 / https://github.com/sosreport/sos/blob/main/sos/report/plugins/__init__.py#L1707 / https://github.com/sosreport/sos/blob/main/sos/report/plugins/__init__.py#L1784 / https://github.com/sosreport/sos/blob/main/sos/report/plugins/__init__.py#L1790 / https://github.com/sosreport/sos/blob/main/sos/report/plugins/__init__.py#L1818-L1820 .

TurboTurtle:
I agree with @pmoravec that this particular concern is better addressed by more specific/controlled specs for `add_copy_spec()`.

One nuance of this is that `add_copy_spec('/var/foo')` would apply the `log-size` to the whole (recursive) collection, whereas `add_copy_spec('/var/foo/*')` would individually apply `log-size` to everything within the directory - so we still have flexibility without having to get exceedingly specific for things like `/proc/net` or what have you.

npinaeva:
I think this behaviour is a little confusing (than `size_limit` applies not per-file but per `add_copy_spec`), but that's probably the only way to strictly limit archive size, so we will adjust `size_limit` or `copy_spec` if we have any problems in the future.
Thanks for all the explanations, and let's try these new presets :)

npinaeva:
I tried to test presets, but I guess they only work if the node I am running sos collect from is using RHCOS? Anyway presets didn't work for my local machine :(

pmoravec:
> I tried to test presets, but I guess they only work if the node I am running sos collect from is using RHCOS? Anyway presets didn't work for my local machine :(

See head of `sos_logs/sos.log`:

    2022-04-01 12:37:49,913 INFO: [sos.report:setup] using 'satellite' preset defaults (--log-size 100 --plugopts apache.log=on)

for chosen preset. You can also overwrite it via `sos report --preset ocp` but I assume you want to have that loaded by default..?

`ocp` preset is chosen only here: https://github.com/sosreport/sos/blob/main/sos/policies/distros/redhat.py#L484 , when we detected we are on an Atomic host - which is detected by this method: https://github.com/sosreport/sos/blob/main/sos/policies/distros/redhat.py#L466 .

Is either of the detection suboptimal?

npinaeva:
Ah I guess I misunderstood the option - I thought it is for sos collect, but it's for sos report, I will test with report and get back, thanks for clarifying

pmoravec:
`sos collect` offers `--preset` option as well, to pass it to `sos report` instances run within the cluster. If not provided, `sos report` run on individual cluster nodes will chose the preset based on `probe_preset` method like https://github.com/sosreport/sos/blob/main/sos/policies/distros/redhat.py#L415 or the https://github.com/sosreport/sos/blob/main/sos/policies/distros/redhat.py#L484 one.

npinaeva:
I was able to verify presets with this fix https://github.com/sosreport/sos/pull/2899

npinaeva:
I think presets we use here (especially `container_runtime`) are more node-related, and if somebody somehow configured a node to use different options, maybe the best way is to use them and not overwrite with `sos collect` presets.

So I think only use presets for `sos report` should be enough here and makes the most sense

pmoravec:
Agreed. Presets are relevant to the system where `sos report` is executed only (until we introduce a preset with collect or clean option..?). So passing a preset option via `collect1 is meaningful only when it differs from default (by either config file or cmdline option).

TurboTurtle:
Minor update, changing `log-size` to 100 instead of 50.

Issue #2851, opened 2022-02-12
[low prio] a_s_c should skip collecting (all) binary files instead of tailing them due to sizelimit


pmoravec:
Reproducer on Red Hat distros:

    sos report -o rpm -k rpm.rpmdb=true

will attempt to collect `/var/lib/rpm/Packages` which is a binary file often >25MB big. So we get an unusable tail.

Currently, we skip collecting tail of *some* binary files, those matching https://github.com/sosreport/sos/blob/main/sos/report/plugins/__init__.py#L69. It is worth expanding the check to more evident binary files, BUT we should be careful here:

- there is no simple and/or reliable heuristic to detect a file is binary (cf. the discussion about `file_is_binary` in cleaner, though there the heuristic can be more strict)
- the heuristic "is_binary" in `report` should rather return false negative than *any* false positive result. We should skip tailing *just* files we are sure they are binary.

I dont see much reliable way how to improve current situation in `report` in this aspect. Until somebody gets an idea, I am also fine closing this as a known limitation.

TurboTurtle:
It's crude, but would it be viable for us to call `file --mime-type $filename` in this case, and filter out anything that has an `application/*` mime-type?

Issue #2839, opened 2022-02-01
[cleaner] Some evidently binary files are attempted to be obfuscated


pmoravec:
Two issues in fact:

1) `/etc/hostid` is being collected by two plugins `host` and `networking` (https://github.com/sosreport/sos/blob/main/sos/report/plugins/networking.py#L62), the later is redundant.

2) Cleaner attempts to obfuscate the file, despite it has evidently binary content:

```
# hexdump -C /etc/hostid 
00000000  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
*
07b00000
#
```

and it takes significant amount of time:

    2022-02-01 14:17:40,764 DEBUG: [cleaner:sosreport-pmoravec-rhel8-2022-02-01-sjjppwd] Obfuscating sos_strings/host/etc.hostid.tailed
    2022-02-01 14:18:26,599 DEBUG: [cleaner:sosreport-pmoravec-rhel8-2022-02-01-sjjppwd] Obfuscating sos_strings/jars/jars.json
    2022-02-01 14:18:26,616 DEBUG: [cleaner:sosreport-pmoravec-rhel8-2022-02-01-sjjppwd] Obfuscating sos_strings/networking/etc.hostid.tailed
    2022-02-01 14:19:12,231 DEBUG: [cleaner:sosreport-pmoravec-rhel8-2022-02-01-sjjppwd] Obfuscating sos_strings/logs/var.log.messages-20220130.tailed

pmoravec:
Fix for 1) is trivial, skip collecting the file in `networking` plugin.

For 2) , my concern is that `file_is_binary` method/heuristic simply does not work properly to files with empty content (where very first bytes are '\0`, to be precise).

TurboTurtle:
For point 2, is this a matter of simply reading more than 1 byte in `file_is_binary()`, or do we need to get a bit more complex here?

pmoravec:
Reading any number of `\0` bytes that way, no exception is raised (at least I replaced there 1 by random values up to 10k and got the same results).

Searching for other approaches, I found these two promising:

- https://stackoverflow.com/questions/898669/how-can-i-detect-if-a-file-is-binary-non-text-in-python#comment52258726_7392391
- https://stackoverflow.com/questions/1446549/how-to-identify-binary-and-text-files-using-python/1446870#1446870 (referenced from https://gist.github.com/magnetikonline/7a21ec5f5bcdbf7adb92f9d617e6198f)

[...]

PR #2897, opened 2022-03-30
[utilities, Plugin] Improve binary file detection, don't tail binary files

TurboTurtle:
First, implement better binary file detection for `sos clean` via the `python-magic` module. Which is the python bindings for `libmagic` - which among other things is the backend behind the `file` command.

Then add a check to `add_copy_spec()` so we don't tail binary files.

Closes: #2839
Closes: #2851

---

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -60,7 +60,8 @@ py_break_task:
             PY_VERSION: "latest"
         - env:
             PY_VERSION: "3.9"
-    setup_script: pip install -r requirements.txt
+    # This image has 2 py environments. Install to the one sos uses.
+    setup_script: pip3 install -t /usr/lib/python3/dist-packages -r requirements.txt
     main_script: ./bin/sos report --batch
 
 # Make sure a user can manually build an rpm from the checkout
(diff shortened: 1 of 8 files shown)


TurboTurtle:
Hmm. I don't know why we're getting an import error in the breakage tests when we're clearly installing the module:

```
Successfully installed [...] pyparsing-3.0.7 python-magic-0.4.25 pytz-2022.1 requests-2.27.1 [...]
```

TurboTurtle:
AH - ok. The community python image the breakage test runs in has 2 python environments. Using `pip` to install here uses the `/usr/local/bin/python` environment, whereas `sos` is expecting `/usr/bin/python`. Interestingly, there is not a `/usr/bin/pip` in the image to leverage that python environment which does legitimately exist.

I think this is most easily resolved by re-targeting the installation directory to /usr/lib/python3/dist-packages for this containerized test.

pmoravec:
Sadly we are enlarging dependencies of `sos`. But as the "best" alternative is the native `mimetypes` with so limited usage.. /o\

TurboTurtle:
Yeah, I'm not sure why the F35 test is failing now - especially with the change to not writing binary data. I'm spinning up a testing VM in GCE and will be looking at this today.

TurboTurtle:
It appears that on F35, the string of '-' characters is being detected as binary content. I've tested using a randomly generated string and that seems to work as expected (detected as ascii text, not binary) however it does introduce a slightly longer runtime. It's ugly, but I don't see another way forward given that the python module is leveraging the underlying `libmagic` to do this detection.

pmoravec (approved):

PR #2902, opened 2022-04-05
[ocp, openshift] Re-align API collection options and rename option

TurboTurtle:
Previously, in #2888, the `openshift` plugin was extended to allow API
collections by using a default-available kubeconfig file rather than
relying on user-provided tokens. This also included flipping the default
value of the `no-oc` plugin option to `True` (meaning do not collect API
output by default).

This worked for the plugin, but it introduced a gap in `sos collect`
whereby the cluster profile could no longer reliably enable API
collections when trying to leverage the new functionality of not
requiring a user token.

Fix this by updating the cluster profile to align with the new
default-off approach of API collections.

Along with this, add a toggle to the cluster profile directly to allow
users to toggle API collections on or off (default off) directly. This
is done via a new `with-api` cluster option (e.g. `-c ocp.with-api`).
Further, rename the `openshift` plugin option from `no-oc` to
`with-api`. This change not only makes the option use case far more
obvious, it will also align the use of the option to both `collect` and
`report` so that users need only be aware of a single option for either
method.

The cluster profile also has logic to detect which plugin option,
`no-oc` or `with-api` to use based on the (RHEL) sos version installed
on the nodes being inspected by the `ocp` cluster profile.

---

### DIFF
--- a/sos/collector/clusters/ocp.py
+++ b/sos/collector/clusters/ocp.py
@@ -30,7 +30,11 @@ class ocp(Cluster):
     clusterAdmin privileges.
 
     If this requires the use of a secondary configuration file, specify that
-    path with the 'kubeconfig' cluster option.
+    path with the 'kubeconfig' cluster option. This config file will also be
+    used on a single master node to perform API collections if the `with-api`
+    option is enabled (default disabled). If no `kubeconfig` option is given,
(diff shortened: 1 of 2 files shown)


vwalek:
Seem fine @TurboTurtle , LGTM

pmoravec (approved):

PR #2904, opened 2022-04-06
[collector] Use --force-pull-image option for pods created with oc

npinaeva:
Set --force-pull-image=True by default, can be turned off with
--force-pull-image=False

---


TurboTurtle (commented):
Ack to the code change.

We should update `man/en/sos-collect.1` as well to reflect that it is now enabled by default, and users can use this like `--pull=off` or similar to disable.

pmoravec (approved):

PR #2909, opened 2022-04-14
[sg_ses] Added sg_ses plugin to collect info of scsi devices

BhushanSKale:
The plugin collects ses page 2 info using sg_ses command

Signed-off-by: Bhushan Kale bhushanskale@hotmail.com

---

### DIFF
--- a/sos/report/plugins/scsi.py
+++ b/sos/report/plugins/scsi.py
@@ -39,6 +39,14 @@ class Scsi(Plugin, IndependentPlugin):
             "/sys/class/scsi_generic"
         ])
 
+        scsi_types = ["enclosu"]
+        result = self.collect_cmd_output('lsscsi -g')
+        if result['status'] == 0:
+            for line in result['output'].splitlines():
+                if (line.split()[1] in scsi_types):
+                    devsg = line.split()[-1]
+                    self.add_cmd_output("sg_ses -p2 -b1 %s" % devsg)
+
         self.add_cmd_output("lsscsi -i", suggest_filename="lsscsi")
 
         self.add_cmd_output([
@@ -47,7 +55,6 @@ class Scsi(Plugin, IndependentPlugin):
             "lsmap -all",
             "lsnports",
             "lsscsi -H",
-            "lsscsi -g",
             "lsscsi -d",
             "lsscsi -s",
             "lsscsi -L"


TurboTurtle (commented):
This addition looks ok by itself, but we also have #2907 open for, apparently, the same component - Seagate SES.

I would say this collection should either be added to that plugin (after it gets merged), or this should instead be added to the `scsi` plugin which already has collections from commands from `sg3_utils` and calls `lsscsi -g`.

sos/report/plugins/scsi.py
```diff
@@ -27,6 +27,7 @@ class Scsi(Plugin, IndependentPlugin):
 
     plugin_name = 'scsi'
     profiles = ('storage', 'hardware')
+    packages = ('sg3_utils')
```
TurboTurtle, line 30: This would disable the plugin for all systems without this package, which would not be desirable.

sos/report/plugins/scsi.py
```diff
@@ -39,6 +40,14 @@ def setup(self):
             "/sys/class/scsi_generic"
         ])
 
+        scsi_types = ["enclosu"]
+        result = self.collect_cmd_output('lsscsi -g')
+        if result['status'] == 0:
+            for line in result['output'].splitlines():
+                if (line.split()[1] in scsi_types):
+                    devsg = line.split()[-1]
+                    self.add_cmd_output("sg_ses -p2 -b1 %s" % devsg)
```
TurboTurtle, line 49: We can leverage a predicate here to control the command execution based on the presence of the `sg3_utils` package. E.G.

```python
sg3_pred = SoSPredicate(packages='sg3_utils')
[...]
        scsi_types = ["enclosu"]
        result = self.collect_cmd_output('lsscsi -g')
        if result['status'] == 0:
            for line in result['output'].splitlines():
                if (line.split()[1] in scsi_types):
                    devsg = line.split()[-1]
                    self.add_cmd_output("sg_ses -p2 -b1 %s" % devsg,
                                        pred=sg3_pred)
```

sos/report/plugins/scsi.py
```diff
@@ -39,6 +40,14 @@ def setup(self):
             "/sys/class/scsi_generic"
         ])
 
+        scsi_types = ["enclosu"]
+        result = self.collect_cmd_output('lsscsi -g')
```
TurboTurtle, line 44: Ack, but remove the other call to `lsscsi -g` on line 59 so we don't duplicate collections.

TurboTurtle:
The CI failures are due to the `packages` class attr being a string, not a tuple (missing `,`), but that should be resolved with removing that addition and instead using the `SoSPredicate` as mentioned above.

Aside from that, you'll need to squash your fixup commits into a single commit for this change please. If you aren't sure how to do that:

1. From your branch run `git rebase -i HEAD~3`
2. In the editor, drop the commits that create/modify the previous `sg_ses` plugin (change `pick` to `drop`).
3. If the commits have changed such that there are multiple commits changing `scsi.py`, change the commits after the first one making those changes to `fixup`. I.E. there should be only one commit that makes changes to `scsi.pi`
4. Save and exit the editor, git should then squash/remove the commits as instructed.
5. `git push origin $your_branch --force` to override this PR while avoiding merge commits/conflicts.

sos/report/plugins/scsi.py
vikas-goel: The other `sg_` commands used in the plugin belong to the same sg3_utils RPM. I think we can live without the predicate in that case.

TurboTurtle: I was actually originally going to suggest adding the predicate to those commands, but then it appears we don't support `pred` as an arg for `add_blockdev_cmd()` currently.

So, I have an AI on me to add that support, and then I'll make a follow up commit that reuses the predicate there as well too.

sos/report/plugins/scsi.py
BhushanSKale: Removed the duplicate call to lsscsi -g

BhushanSKale:
> The CI failures are due to the `packages` class attr being a string, not a tuple (missing `,`), but that should be resolved with removing that addition and instead using the `SoSPredicate` as mentioned above.
> 
> Aside from that, you'll need to squash your fixup commits into a single commit for this change please. If you aren't sure how to do that:
> 
> 1. From your branch run `git rebase -i HEAD~3`
> 2. In the editor, drop the commits that create/modify the previous `sg_ses` plugin (change `pick` to `drop`).
> 3. If the commits have changed such that there are multiple commits changing `scsi.py`, change the commits after the first one making those changes to `fixup`. I.E. there should be only one commit that makes changes to `scsi.pi`
> 4. Save and exit the editor, git should then squash/remove the commits as instructed.
> 5. `git push origin $your_branch --force` to override this PR while avoiding merge commits/conflicts.

vikas-goel:
LGTM

vikas-goel (approved):

sos/report/plugins/scsi.py
BhushanSKale: Hi @TurboTurtle , could you please review again?

pmoravec (approved):

PR #2914, opened 2022-04-20
[policies] Simplify flow in _container_init()

TurboTurtle:
Simplifies the logic and flow of `LinuxPolicy._container_init()` so that
it is both easier to follow and less prone to edge cases that could
break in-container collections that are not setup to capture from the
host system.

---

### DIFF
--- a/sos/policies/distros/__init__.py
+++ b/sos/policies/distros/__init__.py
@@ -218,13 +218,15 @@ class LinuxPolicy(Policy):
         if ENV_CONTAINER in os.environ:
             if os.environ[ENV_CONTAINER] in ['docker', 'oci', 'podman']:
                 self._in_container = True
-        if ENV_HOST_SYSROOT in os.environ:
-            _host_sysroot = os.environ[ENV_HOST_SYSROOT]
-        use_sysroot = self._in_container and _host_sysroot is not None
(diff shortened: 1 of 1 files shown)


sos/policies/distros/__init__.py
```diff
@@ -218,13 +218,12 @@ def _container_init(self):
...
-        if ENV_HOST_SYSROOT in os.environ:
-            _host_sysroot = os.environ[ENV_HOST_SYSROOT]
-        use_sysroot = self._in_container and _host_sysroot is not None
-        if use_sysroot:
-            host_tmp_dir = os.path.abspath(_host_sysroot + self._tmp_dir)
-            self._tmp_dir = host_tmp_dir
-        return _host_sysroot if use_sysroot else None
+                if ENV_HOST_SYSROOT in os.environ:
+                    self._tmp_dir = os.path.abspath(
+                        os.environ[ENV_HOST_SYSROOT] + self._tmp_dir
```
pmoravec, line 223: Can't be in some strange use case `os.environ[ENV_HOST_SYSROOT] = None`? The current code detects this, the new one doesnt.

Otherwise ACK and +1 to the simplification, this will be more readable.

pmoravec (approved):

PR #2916, opened 2022-04-22
[collect] Add cluster profile for RHOSP

TurboTurtle:
Adds a cluster profile for Red Hat OpenStack Platform to identify
controller and (optionally) compute nodes to collect sos reports from.

---


TurboTurtle:
@BryanQuigley and/or @slashdd - I made this with RHOSP in mind, but mainly because I'm not familiar with other flavors of  OpenStack, personally. If this same profile can be used for Ubuntu/Deb deployments (meaning, the use of the ansible inventory yaml is not RH-specific) then I can drop the RH-specific verbiage.

sos/collector/clusters/openstack.py
```diff
@@ -0,0 +1,68 @@
...
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+import yaml
```
pmoravec, line 11: Some tests are failing with `Could not initialize 'collect': No module named 'yaml'`. The module is provided by `python3-pyyaml` which is a requirement of `python3-magic` we (newly) need for file type detection. So how the module could be missed?

pmoravec: ACK to the PR otherwise, no idea why the SCI tests could fail this way.

TurboTurtle: I've been looking into this a bit more.

There is a possibility of using something like `openstack overcloud profiles list` and then a follow up `openstack` command that maps the node names to addresses - however the architecture of OSP in regards to where this command is functional is not immediately clear to me.

I've also clarified with the support team that this ansible inventory file is updated as the environment changes, and is the defacto source of truth.

I'm a bit torn here. One the one hand I am _really_ not fond of adding _another_ dependency, but this approach to enumeration is a lot more straight forward and reliable than relying on a chain of commands that output a table and having to grok that.

pmoravec:
Maybe a stupid idea: `python3-pyyaml` does not need to be present on all systems, but maybe it is present on all OSP nodes?

Could we assume that / is that correct? If so, we can conditionally collect the data only when "try import yaml" succeeded.

TurboTurtle:
No, because `sos collect` can be run from a non-OSP node - pyyaml needs to be available locally. In the case that `collect` is being run from a workstation, we still handle the yaml content locally be reading it from the remote node and parsing it as a string instead of an open file.

pmoravec:
I feel the option 1 (new weak dependency) has less negs than the 2, indeed.

pmoravec (approved):

PR #2918, opened 2022-04-25
[report,collect] Add env var and prompt options for encryption

TurboTurtle:
Adds a new `--encrypt` option that may be used in place of the existing
encrypt-related options. If used alongside `--batch`, this new option
will cause sos to set encryption options based on the `SOSENCRYPTKEY` or
`SOSENCRYPTPASS` environment variables. This allows users to leverage
this functionality without potentially leaking the values into `ps`-like
output inside the archive.

If `--batch` is not used, then this option prompts the user to select
the type of method to use and then provide the value in-line, or to use
the env var option.

---


pmoravec (approved):
Nice!

PR #2919, opened 2022-04-25
ostree: Collect logs from ostree-finalize-staged in particular

cgwalters:
See https://github.com/ostreedev/ostree/pull/2589 for rationale;
this is a key systemd unit and we always want its full logs.

---

### DIFF
--- a/sos/report/plugins/ostree.py
+++ b/sos/report/plugins/ostree.py
@@ -27,4 +27,11 @@ class OSTree(Plugin, IndependentPlugin):
         if self.get_option("verify"):
             self.add_cmd_output("ostree fsck")
 
+        units = [
+            'ostree-finalize-staged',
+            'ostree-boot-complete',
+        ]
+        for unit in units:
+            self.add_journal(unit)
+
 # vim: set et ts=4 sw=4 :


TurboTurtle (approved):

pmoravec (approved):

TurboTurtle:
@cgwalters we were just talking about this - it may be better to stick those units in the `services` tuple for the plugin enablement trigger.

This would mean not only would we get the journals, but we'd also get the service status and have a way to enable the plugin if something has gone completely off the rails and `/ostree` is missing or in some way not detected.

TurboTurtle:
Cycling around on this - ping @cgwalters - see comment above.

cgwalters:
> @cgwalters we were just talking about this - it may be better to stick those units in the services tuple for the plugin enablement trigger.

Since you know this code base and I don't, do you mind force-pushing any changes on top of my commit?

TurboTurtle:
> > @cgwalters we were just talking about this - it may be better to stick those units in the services tuple for the plugin enablement trigger.
> 
> Since you know this code base and I don't, do you mind force-pushing any changes on top of my commit?

Ack, sure thing. I'll get to that in a bit if I can.

For reference, the change would just be adding `services = ($serv1, $serv2)` to the class constructor (similar to the `files` tuple) instead of adding the changes to `setup()`.

TurboTurtle:
I haven't forgotten about this, but I've been having problems force-pushing over this branch. I'll push a separate PR shortly rather than continue fighting with GH over this.

Issue #2922, opened 2022-04-27
PlugOpts val_type=str disallows numeric values (as strings)


pmoravec:
Example:

    # python3 bin/sos report -o skydive -k skydive.username=foo -k skydive.password='456' --batch --build
    
    sosreport (version 4.3)
    
    Plugin option 'skydive.password' takes string (no spaces), not int
    # python3 bin/sos report -o skydive -k skydive.username='123' -k skydive.password=foo --batch --build
    
    sosreport (version 4.3)
    
    Plugin option 'skydive.username' takes string (no spaces), not int
    #

It is arguable if numeric-only password is safe, or how meaningful is to have numeric-only username, but we should allow it. '123' is also a string.

TurboTurtle:
Is the right answer to expand the `val_type` list for these options, or to change the type-handling flow when setting these options?

On a one-off basis, adding `int` as a `val_type` here makes no _functional_ difference, because the use of the value will be as a string anyways. I'm leaning towards changing the logic to just accept anything that could possibly be a string when the `val_type` is `str`, but I'm hesitant as I'm not sure if there are other edge cases we could catch then.

pmoravec:
The change of logic "`val_type=str` => accept anything" does make sense to me.

jcastill:
Do you mean modifying the logic inside sos/report/plugins/__init__.py in the class PluginOpt directly?

pmoravec:
Yes.

PR #2923, opened 2022-04-27
[skydive] Fix PlugOpts val_type=str to accept numeric values

jcastill:
This PR fixes the cases where either the username or the
password are numeric values, and avoid receiving error
messages like:

Plugin option 'skydive.password' takes string (no spaces),
not int

Resolves: #2922 

---


jcastill:
I based this off the postgresql.py plugin, that uses this approach for dbport. Testing for numeric username and password worked with this change.

pmoravec:
I would rather accept a generic approach where any PlugOpt with `val_type=str` does accept numbers as a string value. The `skydive` plugin is one example, there are few more with password set as `val_type=str`.

Issue #2921, opened 2022-04-26
sos report -l to honour -a/-k


pmoravec:
Since sos-4.2, there is a difference in `sos report -l` behaviour when `-a` or `-k` option is used (in cmdline, config file or preset). Previously, `-l`/`--list-options` was printing the currently set plugin options while nowadays it is printing their defaults. That is inconsistent with `-o`/`-e` options that affect the `-l` output.

I suggest reverting back this change to be consistent, such that either `-o`/`-e`/`-a`/`-k` option affects `-l` output. Then `-l` output can be used as a check of currently set option values also.

If one would like to see bare defaults, let use plain `-l` option (with empty config file, no cmdline option or any preset).

PR #2924, opened 2022-04-27
[report] --list-plugins should report used, not default, option values

TurboTurtle:
When using `--list-plugins`, sos should report the values that will be
used in a given command, or with a given config file, not what the
default values are.

By reporting the set value, users can be sure their configuration or
commandline settings are being honored correctly before executing a
report collection.

Closes: #2921

---

### DIFF
--- a/sos/report/__init__.py
+++ b/sos/report/__init__.py
@@ -868,24 +868,32 @@ class SoSReport(SoSComponent):
             _defaults = self.loaded_plugins[0][1].get_default_plugin_opts()
             for _opt in _defaults:
                 opt = _defaults[_opt]
-                val = opt.default
-                if opt.default == -1:
-                    val = TIMEOUT_DEFAULT
+                val = opt.value
+                if opt.value == -1:
+                    if _opt == 'timeout':
+                        val = self.opts.plugin_timeout or TIMEOUT_DEFAULT
+                    elif _opt == 'cmd-timeout':
+                        val = self.opts.cmd_timeout or TIMEOUT_DEFAULT
+                    else:
+                        val = TIMEOUT_DEFAULT
+                if opt.name == 'postproc':
+                    val = not self.opts.no_postproc
                 self.ui_log.info(" %-25s %-15s %s" % (opt.name, val, opt.desc))
             self.ui_log.info("")
 
             self.ui_log.info(_("The following plugin options are available:"))
             for opt in self.all_options:
                 if opt.name in ('timeout', 'postproc', 'cmd-timeout'):
-                    continue
+                    if opt.value == opt.default:
+                        continue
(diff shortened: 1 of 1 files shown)


sos/report/__init__.py
```diff
@@ -879,13 +879,13 @@ def list_plugins(self):
                 if opt.name in ('timeout', 'postproc', 'cmd-timeout'):
                     continue
```
jjansky1, line 880: Can we also add 1 more condition to show if for example plugin.timeout is changed?

in meaning sos report -l -o dnf -k dnf.timeout=400 do not show that dnf.timeout is different.

My proposal is 

```
                if opt.name in ('timeout', 'postproc', 'cmd-timeout'):
                    if opt.value == opt.default:
                        continue
```

jjansky1 (changes requested):
Can we look at it? I worked on my own PR when you posted yours.

sos/report/__init__.py
TurboTurtle: Ack, good idea.

sos/report/__init__.py
```diff
@@ -868,8 +868,8 @@ def list_plugins(self):
             _defaults = self.loaded_plugins[0][1].get_default_plugin_opts()
             for _opt in _defaults:
                 opt = _defaults[_opt]
-                val = opt.default
-                if opt.default == -1:
+                val = opt.value
+                if opt.value == -1:
                     val = TIMEOUT_DEFAULT
                 self.ui_log.info(" %-25s %-15s %s" % (opt.name, val, opt.desc))
```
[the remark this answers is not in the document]
TurboTurtle: Ack.

TurboTurtle:
Updated to handle both suggestions, @jjansky1.

pmoravec (approved):

pmoravec:
Nice collaboration!

sos/report/__init__.py
```diff
@@ -868,24 +868,30 @@ def list_plugins(self):
...
                 opt = _defaults[_opt]
-                val = opt.default
-                if opt.default == -1:
-                    val = TIMEOUT_DEFAULT
+                val = opt.value
+                if opt.value == -1:
+                    if _opt == 'timeout':
+                        val = self.opts.plugin_timeout or TIMEOUT_DEFAULT
+                    elif _opt == 'cmd-timeout':
+                        val = self.opts.cmd_timeout or TIMEOUT_DEFAULT
```
jjansky1, line 876: Looks better as timeout and cmd-timeout is not respected, however should not be there also postproc? I am not sure why anyone will do it but --no-postproc is still not respected.

jjansky1 (changes requested):

PR #2936, opened 2022-05-08
[postgresql] Simplify val_type for dbport

jcastill:
After the change introduced via PR #2935
(Allow 'str' PlugOpt type to accept any value)
the postgresql plugin was the only one specifying
both str and int as possible values. This patch
simplifies the val_type field to be in line
with the rest of plugins where str covers int as well.

Depends on: #2935 
Closes: #2936

---


pmoravec:
Why not reducing to `int` rather?

jcastill:
Actually this was my original idea but then I started testing your path with both int (first) then str, and then committed with the last test. I can change it back if makes things clearer, but at the moment all int values will be treated as str anyway, right? Other than the port being a number, is there an advantage on using 'int' as val_type?

pmoravec (approved):

TurboTurtle (approved):

PR #2939, opened 2022-05-10
[ceph_mon/ceph_osd] Collect metata type used in the backend

pponnuvel:
Ceph uses leveldb (old versions) or rocksdb (recent to current ones)
for metadata. The older filestore backend is deprecated since
Quincy (17.2.0) which uses leveldb. So it's useful to know
if, for example, ceph-mon is still using leveldb for metadata.
This can happen even after upgrading to newer Ceph versions
if the leveldb isn't migrated to leveldb.

Signed-off-by: Ponnuvel Palaniyappan <pponnuvel@gmail.com>

---

### DIFF
--- a/sos/report/plugins/ceph_mon.py
+++ b/sos/report/plugins/ceph_mon.py
@@ -28,9 +28,9 @@ class CephMON(Plugin, RedHatPlugin, UbuntuPlugin):
         })
 
         self.add_copy_spec([
-            "/var/log/ceph/ceph-mon*.log",
-            "/var/lib/ceph/mon/",
-            "/run/ceph/ceph-mon*"
+            "/run/ceph/ceph-mon*",
+            "/var/lib/ceph/mon/*/kv_backend",
+            "/var/log/ceph/ceph-mon*.log"
         ])
 
         self.add_cmd_output([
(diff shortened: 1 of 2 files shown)


sos/report/plugins/ceph_mon.py
```diff
@@ -28,9 +28,9 @@ def setup(self):
         })
 
         self.add_copy_spec([
-            "/var/log/ceph/ceph-mon*.log",
-            "/var/lib/ceph/mon/",
```
pponnuvel, line 32: This is nullified due to the fact `/var/lib/ceph/mon/` is in the forbidden list as well. But there's not much else interesting there anyway besides `kv_backend`.

sos/report/plugins/ceph_mon.py
```diff
@@ -110,14 +110,12 @@ def setup(self):
...
         ], subdir="json_output", tags="insights_ceph_health_detail")
 
-        # these can be cleaned up too but leaving them for safety for now
         self.add_forbidden_path([
             "/etc/ceph/*keyring*",
             "/var/lib/ceph/*keyring*",
             "/var/lib/ceph/*/*keyring*",
             "/var/lib/ceph/*/*/*keyring*",
             "/var/lib/ceph/osd",
-            "/var/lib/ceph/mon",
```
pponnuvel, line 120: Only the `keyring` file(s) contain sensitive info. Since that's explicitly excluded, no need to forbid the whole directory. This would otherwise prevent collecting `/var/lib/ceph/mon/*/kv_backend`.

sos/report/plugins/ceph_osd.py
```diff
@@ -28,13 +28,10 @@ def setup(self):
 
         # Only collect OSD specific files
         self.add_copy_spec([
+            "/run/ceph/ceph-osd*",
+            "/var/lib/ceph/osd/*/kv_backend",
             "/var/log/ceph/ceph-osd*.log",
             "/var/log/ceph/ceph-volume*.log",
-
-            "/var/lib/ceph/osd/",
-            "/var/lib/ceph/bootstrap-osd/",
```
pmoravec, line 35: Is this really desired to stop collecting?

pponnuvel: This doesn't contain anything useful as it's used during the osd bootstrapping. Typically contains just the keyring (which is excluded already deliberately).

pmoravec: OK then :+1: . I wasn't sure if it is deliberate or a mistake.

pmoravec (approved):

pmoravec:
Cc @nkshirsagar for a potential review or awareness.

nkshirsagar:
> Cc @nkshirsagar for a potential review or awareness.
Thanks! This looks good to me.

Issue #2806, opened 2021-12-20
[report] Unhandled exception when concurrently removing temp dir


pmoravec:
When running sos report with `--tmp-dir` and the directory is "accidentally" removed, `sos report` raises an uncaught exception. It is just matter of removing the temp dir when `sos` generates the final archive.

To reliably & easily reproduce, let slow down execution on the right place of the code:

    sed -i '/Creating compressed archive/a\                from time import sleep; sleep(20)' sos/report/__init__.py

and when execution:

    mkdir /tmp/sostmp
    sos report --batch -o host --tmp-dir=/tmp/sostmp

prints:

     Creating compressed archive...

then `rm -rf /tmp/sostmp` in another terminal. `sos report` will finish with:

    [archive:TarFileArchive] An error occurred compressing the archive: [Errno 2] No such file or directory: '/tmp/sostesttmp/sos.z5xi26ru/sosreport-pmoravec-rhel8-2021-12-20-okpzkaj.tar.xz'
    Traceback (most recent call last):
      File "bin/sos", line 22, in <module>
        sos.execute()
      File "/root/sos-main/sos/__init__.py", line 186, in execute
        self._component.execute()
      File "/root/sos-main/sos/report/__init__.py", line 1540, in execute
        return self.final_work()
      File "/root/sos-main/sos/report/__init__.py", line 1390, in final_work
        checksum = self._create_checksum(archive, hash_name)
      File "/root/sos-main/sos/report/__init__.py", line 1260, in _create_checksum
        return digest.hexdigest()
    UnboundLocalError: local variable 'digest' referenced before assignment

(optionally, to yet simplify the reproduce more, dont inject `sleep` but `import shutil;shutil.rmtree(\"/tmp/sostmp\")` directly)

PR #2941, opened 2022-05-12
[sos] Fix unhandled exception when concurrently removing temp dir

jcastill:
This patch tries to capture the exception when running
the sosreport at the same time the temp directory
specified as an option is deleted.

Resolves: #2806

---

### DIFF
--- a/sos/report/__init__.py
+++ b/sos/report/__init__.py
@@ -1453,9 +1453,14 @@ class SoSReport(SoSComponent):
             if not archive:
                 print("Creating archive tarball failed.")
             else:
-                # compute and store the archive checksum
-                hash_name = self.policy.get_preferred_hash_name()
-                checksum = self._create_checksum(archive, hash_name)
+                try:
+                    # compute and store the archive checksum
+                    hash_name = self.policy.get_preferred_hash_name()
+                    checksum = self._create_checksum(archive, hash_name)
+                except Exception:
+                    print(_("Error generating archive checksum after "
+                            "archive creation.\n"))
+                    return False
                 try:
                     self._write_checksum(archive, hash_name, checksum)
                 except (OSError, IOError):


jcastill:
I'm not really sure about the actual error message and formatting, but at least I think I've captured the exception in the right place. The result of trying to reproduce with this patch is:

` Setting up archive ...
 Setting up plugins ...

  Starting 1/1   host            [Running: host]

  Finished running plugins

[archive:TarFileArchive] An error occurred compressing the archive: [Errno 2] No such file or directory: '/tmp/sostmp/sos.zg8hovoe/sosreport-voyager-2022-05-12-lsciawt.tar.xz'

Error while accessing temp dir specified. Please check that the directory exists.`

sos/report/__init__.py
```diff
@@ -1453,9 +1453,16 @@ def final_work(self):
...
-                checksum = self._create_checksum(archive, hash_name)
+                try:
+                    # compute and store the archive checksum
+                    hash_name = self.policy.get_preferred_hash_name()
+                    checksum = self._create_checksum(archive, hash_name)
+                except Exception:
+                    print("")
+                    print(_("Error while accessing temp dir specified. "
+                            "Please check that the directory exists."))
+                    print("")
```
TurboTurtle, line 1464: You can use `\n` in the single print statement with the message to generate the newlines, e.g.

`print(_("\nError while [...] exists.\n"))`

We could probably make this a bit more specific. `Error generating archive checksum after archive creation.` I think gives us a bit more insight here.

pmoravec: ACK to both `\n` and message content.

Apart of that, +1 for the PR. Thanks!

TurboTurtle (approved):

pmoravec (approved):

PR #2947, opened 2022-05-26
[ovirt] answer files: Filter out all password keys

didib:
Instead of hard-coding specific keys and having to maintain them over
time, replace the values of all keys that have 'password' in their name.
I think this covers all our current and hopefully future keys. It might
add "false positives" - keys that are not passwords but have 'password'
in their name - and I think that's a risk worth taking.

A partial list of keys added since the replaced code was written:
- grafana-related stuff
- keycloak-related stuff
- otopi-style answer files

---

### DIFF
--- a/sos/report/plugins/ovirt.py
+++ b/sos/report/plugins/ovirt.py
@@ -241,19 +241,22 @@ class Ovirt(Plugin, RedHatPlugin):
                 r'{key}=********'.format(key=key)
             )
 
-        # Answer files contain passwords
-        for key in (
-            'OVESETUP_CONFIG/adminPassword',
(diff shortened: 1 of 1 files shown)


didib:
@sandrobonazzola can you please review? Thanks.

sandrobonazzola (approved):
LGTM

didib:
Sorry, this won't work - we have passwords which do not have 'password' in their key. Will fix later.

sandrobonazzola (approved):
LGTM

TurboTurtle (approved):

pmoravec (approved):

PR #2949, opened 2022-06-02
[runtimes:crio] switch from parsing output in table format to json

npinaeva:
crictl added a field to the `crictl ps` output and that broke parsing - now we use json output format that (hopefully) won't change in the future

---

### DIFF
--- a/sos/policies/runtimes/crio.py
+++ b/sos/policies/runtimes/crio.py
@@ -7,6 +7,7 @@
 # version 2 of the GNU General Public License.
 #
 # See the LICENSE file in the source distribution for further information.
+import json
 
 from sos.policies.runtimes import ContainerRuntime
 from sos.utilities import sos_get_command_output
@@ -29,14 +30,15 @@ class CrioContainerRuntime(ContainerRuntime):
         :type get_all: ``bool``
         """
         containers = []
-        _cmd = "%s ps %s" % (self.binary, '-a' if get_all else '')
+        _cmd = "%s ps %s -o json" % (self.binary, '-a' if get_all else '')
         if self.active:
             out = sos_get_command_output(_cmd, chroot=self.policy.sysroot)
-            if out['status'] == 0:
-                for ent in out['output'].splitlines()[1:]:
-                    ent = ent.split()
+            if out["status"] == 0:
+                out_json = json.loads(out["output"])
+                for container in out_json["containers"]:
                     # takes the form (container_id, container_name)
-                    containers.append((ent[0], ent[-3]))
+                    containers.append(
+                        (container["id"], container["metadata"]["name"]))
         return containers
 
     def get_images(self):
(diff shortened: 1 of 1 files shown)


sos/policies/runtimes/crio.py
```diff
@@ -47,13 +49,21 @@ def get_images(self):
...
+                    # takes the form (repository:tag, image_id)
+                    if len(image["repoTags"]) > 0:
+                        for repo_tag in image["repoTags"]:
+                            images.append((repo_tag, image["id"]))
+                    else:
+                        if len(image["repoDigests"]) == 0:
+                            image_name = "<none>"
+                        else:
+                            image_name = image["repoDigests"][0].split("@")[0]
+                        images.append((image_name + ":<none>", image["id"]))
```
TurboTurtle, line 66: I don't think adding `:<none>` here as part of the image name is the right move. That would require plugins to either use that specifically (self.add_cmd_output('foobar', container=`mycontainer:<none>`), or make their relevant container calls all use globs (`self.add_cmd_output('foobar', container=`mycontainer*`) which will probably add to problems down the line.

sos/policies/runtimes/crio.py
```diff
@@ -29,14 +30,15 @@ def get_containers(self, get_all=False):
...
-            if out['status'] == 0:
-                for ent in out['output'].splitlines()[1:]:
-                    ent = ent.split()
+            if out["status"] == 0:
+                out_json = json.loads(out["output"])
+                for container in out_json["containers"]:
                     # takes the form (container_id, container_name)
-                    containers.append((ent[0], ent[-3]))
+                    containers.append(
+                        (container["id"], container["metadata"]["name"]))
```
TurboTurtle, line 41: Ack.

sos/policies/runtimes/crio.py
npinaeva: so here I was trying to reproduce the output of table format for `crictl images` (also to be compatible with docker output), so adding none is what `crictl` does https://github.com/kubernetes-sigs/cri-tools/blob/ccb94be8f9c365f243d0c5c65fa7c26993d6094c/cmd/crictl/image.go#L522, image name can also be none https://github.com/kubernetes-sigs/cri-tools/blob/ccb94be8f9c365f243d0c5c65fa7c26993d6094c/cmd/crictl/image.go#L542
docker for command https://github.com/sosreport/sos/blob/main/sos/policies/runtimes/__init__.py#L122-L123 will return e.g.

```
# docker images --format '{{lower .Repository}}:{{lower .Tag}} {{lower .ID}}'
<none>:<none> dbfda3a8826c
quay.io/npinaeva/ovn-k:<none> a5851f835871

```
and for `self.add_cmd_output` we need container name and not image name - actually I don't think images are used in sos right now, just added that for completeness

TurboTurtle: Yep, I went cross eyed there and was mixing images with IDs.

PR #2952, opened 2022-06-06
[vdsm] Set LVM option use_devicesfile=0

jorti:
Since RHV 4.4 SP1, vdsm configures LVM to use devicesfile, causing that
the LVM filter configuration used by sos is ignored.

This change disables the use of the devicesfile, so that the information
of the devices used for RHV storage domains can be collected.

Fixes: RHBZ#2093993

Signed-off-by: Juan Orti <jortialc@redhat.com>

---

### DIFF
--- a/sos/report/plugins/vdsm.py
+++ b/sos/report/plugins/vdsm.py
@@ -29,7 +29,8 @@ import re
 # use_lvmetad is set to 0 in order not to show cached, old lvm metadata.
 # use_lvmetad=0
 #
-# preferred_names and filter config values are set to capture Vdsm devices.
+# preferred_names, use_devicesfile and filter config values are set to
+# capture Vdsm devices.
 # preferred_names=[ '^/dev/mapper/' ]
 # filter=[ 'a|^/dev/mapper/.*|', 'r|.*|' ]
 LVM_CONFIG = """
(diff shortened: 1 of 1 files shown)


pmoravec:
Does this "config option change" works well also on older RHV / `vdsm` versions (i.e. what if the new sos is run on older RHV?)?

As you write the change is driven by RHV: can't be `vdsm` used outside RHV where that change would break anything?

jorti:
> Does this "config option change" works well also on older RHV / `vdsm` versions (i.e. what if the new sos is run on older RHV?)?

Previous RHV versions based on RHEL 8 already used `use_devicesfile=0`, which was the default.

I've tested it on a RHEL 7.6 hypervisor, where the lvmdevices concept didn't exist yet, and LVM information is obtained correctly, although this warning is displayed:

    Configuration setting "devices/use_devicesfile" unknown.

> As you write the change is driven by RHV: can't be `vdsm` used outside RHV where that change would break anything?

Using vdsm outside of a RHV/oVirt environment doesn't make much sense, but even in that situation, I don't see any problem obtaining the LVM information ignoring the lvmdevices file.

TurboTurtle (approved):

Issue #2913, opened 2022-04-20
option --only-plugins does not work


anxstj:
I like to execute only one plugin and it seems that `-o | --only-plugins` should be used for that purpose. But

```
# sos report --only-plugins yum --dry-run

sosreport (version 4.3)

WARNING: unable to set option for disabled or non-existing plugin (boot).

This command will collect diagnostic and configuration information from
this CentOS system and installed applications.

An archive containing the collected information will be generated in
/var/tmp/sos.3uzrisxk and may be provided to a CentOS support
representative.

Any information provided to CentOS will be treated in accordance with
the published support policies at:

        Community Website : https://www.centos.org/

The generated archive may contain data considered sensitive and its
content should be reviewed by the originating organization before being
passed to any third party.

No changes will be made to system configuration.

Press ENTER to continue, or CTRL-C to quit.

Optionally, please enter the case id that you are generating this report for []: 

 Setting up archive ...
 Setting up plugins ...
[plugin:anaconda] skipped command 'rpm -V ': (dry run)
[plugin:block] skipped command 'lsblk': (dry run)
[plugin:block] skipped command 'lsblk -t': (dry run)
[plugin:block] skipped command 'lsblk -D': (dry run)
[plugin:block] skipped command 'blkid -c /dev/null': (dry run)
[plugin:block] skipped command 'blockdev --report': (dry run)
[plugin:block] skipped command 'ls -lanR /dev': (dry run)
[plugin:block] skipped command 'ls -lanR /sys/block': (dry run)
[plugin:block] skipped command 'lsblk -O -P': (dry run)
[plugin:block] skipped command 'losetup -a': (dry run)
[plugin:block] skipped command 'parted -s /dev/dm-1 unit s print': (dry run)
[plugin:block] skipped command 'parted -s /dev/sr0 unit s print': (dry run)
[plugin:block] skipped command 'parted -s /dev/dm-0 unit s print': (dry run)
[plugin:block] skipped command 'parted -s /dev/sda unit s print': (dry run)
[plugin:block] skipped command 'fdisk -l /dev/dm-1': (dry run)
[plugin:block] skipped command 'fdisk -l /dev/sr0': (dry run)
[plugin:block] skipped command 'fdisk -l /dev/dm-0': (dry run)
[plugin:block] skipped command 'fdisk -l /dev/sda': (dry run)
[plugin:block] skipped command 'udevadm info /dev/dm-1': (dry run)
[...]

PR #2954, opened 2022-06-10
[InitSystem] Allow toggle of default `is_running` result

TurboTurtle:
The base `InitSystem` used when systemd is not present (or at least
functional), defaults to returning `True` for all probes, in an effort
to not inadvertantly block service queries.

This works well for most scenarios, however it backfires when checking
for runlevel-analogous systemd targets for items such as the `cantboot`
preset - in that the base `InitSystem` will always enable that preset.

Fix that by allowing calls to `InitSystem.is_running()` to specify a
default value to return, instead of blindly always returning `True`.
Similarly, update the preset check to default to `False` to avoid
improperly enabling that preset when systemd is unavailable.

Closes: #2913

---

### DIFF
--- a/sos/policies/init_systems/systemd.py
+++ b/sos/policies/init_systems/systemd.py
@@ -43,8 +43,11 @@ class SystemdInit(InitSystem):
             except IndexError:
                 pass
 
-    def is_running(self, name):
-        svc = self.get_service_status(name)
-        return svc['status'] == 'active'
+    def is_running(self, name, default=False):
+        try:
+            svc = self.get_service_status(name)
+            return svc['status'] == 'active'
+        except Exception:
+            return default
 
 # vim: set et ts=4 sw=4 :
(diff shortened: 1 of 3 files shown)


sos/policies/init_systems/systemd.py
```diff
@@ -43,8 +43,11 @@ def load_all_services(self):
...
 
-    def is_running(self, name):
-        svc = self.get_service_status(name)
-        return svc['status'] == 'active'
+    def is_running(self, name, default=False):
+        try:
+            svc = self.get_service_status(name)
+            return svc['status'] == 'active'
+        except Exception:
+            return default
```
pmoravec, line 51: Is there some known call flow that could lead to an exception we are catching here, or is it there "just in case" / to fallback to `default`?

TurboTurtle: More of a just in case - haven't seen anything yet, but there is the potential situation in which we are able to load the systemd `InitSystem`, but actual calls to querying service statuses fails.

pmoravec (approved):

Issue #1817, opened 2019-10-04
[policies] review pkg_by_name() wildcard behaviour


bmr-cymru:
It was noticed while reviewing #1816 that the `PackageManager.pkg_by_name()` method applies shell-style wildcards to the package names it returns by applying `fnmatch.filter()` to the results of `all_pkgs_by_name()`. This can lead to confusing results when several packages exist which match a supplied pattern, since the function arbitrarily returns the last entry of the list of matching packages:

```python
>>> p.package_manager.pkg_by_name("glibc")
u'glibc'
>>> p.package_manager.pkg_by_name("glibc*")
u'glibc-langpack-en'
>>> p.package_manager.all_pkgs_by_name("glibc*")
[u'glibc-devel', u'glibc-all-langpacks', u'glibc', u'glibc-common', u'glibc-headers', u'glibc-langpack-en']
```

Review this behaviour to see if anything relies on it currently and consider making the `pkg_by_name()` method an exact match.

TurboTurtle:
Reviewing this, there are currently 10 plugins that use wildcards in their enablement triggers for packages - drbd, java, mysql, nss, openssl, openvswitch, ovn_central, ovn_host, pam, and perl.

Some of these are old, and some are relatively newer. So, we have a couple options here:

1. Do nothing, and keep allowing `pkg_by_name()` to accept wildcards
2. Switch `is_installed()` to directly leverage `all_pkgs_by_name()` and have `pkg_by_name()` only do exact matches. There are uses of `pkg_by_name()` outside plugin enablement, such as cluster profiles and presets loading/enablement that today do not have any wildcards in use
3. Switch `pkg_by_name()` to be an exact match, and update these plugins to not use wildcards in their package triggers.

I'm personally leaning towards option 2. While not a very common use case, allowing wildcards in packages tuples for plugins is handy when the package has version information baked into the name (for some forsaken reason), e.g. - openvswitch with `packages = ('openvswitch', 'openvswitch2.*', [...]`

PR #2958, opened 2022-06-21
[PackageManager] Make pkg_by_name() more predictable

TurboTurtle:
As highlighted in #1817, `pkg_by_name()` could provide unpredictable
results, when using wildcards. As such, limited this method to only
returning package info for exact package name matches. In turn, change
`Plugin.is_installed()` to leverage `PackageManager.all_pkgs_by_name()`
which does explicitly support wildcards and returns information on _all_
matching packages, not just the last one found.

In so doing, clean up the `PackageManager` design to use a new
`packages` property for these lookups, and update the former usage of
`all_pkgs()` accordingly. Similarly, signal `get_pkg_list()` should be
    private (in any sense that a python method can be) by renaming to
    `_get_pkg_list()` and update the single Plugin (`etcd`) referencing this
    method.

Closes: #1817

---

### DIFF
--- a/sos/policies/package_managers/__init__.py
+++ b/sos/policies/package_managers/__init__.py
@@ -22,7 +22,7 @@ class PackageManager():
 
         package name|package.version
 
-    You may also subclass this class and provide a get_pkg_list method to
+    You may also subclass this class and provide a _generate_pkg_list method to
     build the list of packages and versions.
 
     :cvar query_command: The command to use for querying packages
@@ -56,7 +56,7 @@ class PackageManager():
     def __init__(self, chroot=None, query_command=None,
                  verify_command=None, verify_filter=None,
                  files_command=None, remote_exec=None):
-        self.packages = {}
+        self._packages = {}
         self.files = []
 
         self.query_command = query_command or self.query_command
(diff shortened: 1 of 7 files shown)


sos/policies/package_managers/__init__.py
```diff
@@ -112,13 +118,12 @@ def pkg_by_name(self, name):
...
-        if (len(pkgmatches) != 0):
-            return self.all_pkgs_by_name(name)[-1]
-        else:
+        try:
+            return self.packages[name]
+        except Exception:
             return None
 
-    def get_pkg_list(self):
+    def _get_pkg_list(self):
```
pmoravec, line 126: Technically the method does not *get* the pkg list, but it "just" generates it. Worth changing its name? (and definitely change its description - that (both together) is confusing).

pmoravec (commented):
Otherwise ACK :+1: .

sos/policies/package_managers/__init__.py
TurboTurtle: Ack, fair point.

pmoravec (approved):

PR #2969, opened 2022-07-06
[clean] Don't re-create regex for parsers on every skip check

TurboTurtle:
When determining if a parser should be skipped for a given file, we were
previously doing so ineffeciently by regenerating the regex object for
every file from the parser's `skip_files` class attr.

Instead, during parser initialization, use those string definitions to
create a list of regex objects to be used directly by the skip file
checks. This way, we only generate the regex objects once for each
parser for the entire run.

---

### DIFF
--- a/sos/cleaner/__init__.py
+++ b/sos/cleaner/__init__.py
@@ -12,7 +12,6 @@ import hashlib
 import json
 import logging
 import os
-import re
 import shutil
 import tempfile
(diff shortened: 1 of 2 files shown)


pmoravec (approved):
Nice finding - on some testing data, I see improvement from 19mins to 18mins in cleaner time.

PR #2972, opened 2022-07-08
[report] Fix loop devices data gathering

jcastill:
When we run parted command on detached loop devices,
some error messages are printed in the kernel ring buffer,
like:

[ +10.838245] blk_update_request: I/O error, dev loop0,
	sector 0 op 0x1: (WRITE) flags 0x800
	phys_seg 0 prio class 0
Warning: Error fsyncing/closing /dev/loop0: Input/output error

These error messages are printed as well when parted gets run
while collecting sosreport.

This patch solves the problem by removing detached
devices from the list returned from get_block_devs().

Fixes: RHBZ#2010735

---


jcastill:
I tried to reduce the amount of lines as much as possible without affecting code legibility, but as always, any suggestion you have let me know.

sos/report/__init__.py
```diff
@@ -499,7 +499,20 @@ def get_block_devs(self):
         These devices are used by add_blockdev_cmd() in the Plugin class.
         """
         try:
-            return os.listdir('/sys/block/')
+            device_list = os.listdir('/sys/block')
+            loop_devices = sos_get_command_output('losetup --all --noheadings')
+            if loop_devices['status'] == 0:
+                real_loop_devices = []
```
TurboTurtle, line 505: This line needs to be brought outside the conditional. If `losetup` fails for whatever reason, we still need to initialize the empty list for the presence conditional lower on line 512.

sos/report/__init__.py
```diff
@@ -499,7 +499,20 @@ def get_block_devs(self):
...
+                for loop_dev in loop_devices['output'].splitlines():
+                    loop_device = loop_dev.split()[0].replace(':', '')
+                    loop_device = loop_device.replace('/dev/', '')
+                    real_loop_devices.append(loop_device)
+            ghost_loop_devs = [dev for dev in device_list
+                               if dev.startswith("loop")
+                               if dev not in real_loop_devices]
+            dev_list = []
```
TurboTurtle, line 513: This is superfluous given the next line.

TurboTurtle (changes requested):
Ack to the approach, just a few minor changes suggested below.

TurboTurtle (approved):

pmoravec (approved):
ACK. We might add a comment describing what/why ghost loop devices are being removed, but git blame could be used to point to the PR at least :)

PR #2973, opened 2022-07-11
[Plugin] Refactor add_blockdev_cmd into add_device_cmd

TurboTurtle:
Initially, the design behind `add_blockdev_cmd()` made the assumption
that there would (eventually) be separate methods for different kinds of
devices. On review, this is less practical and less maintainable than a
single method, `add_device_cmd()` paired with improved functionality for
handling device lists.

As such, rework `add_blockdev_cmd()` into `add_device_cmd()` and with
it, change the logic on how device lists are sourced. The `devices` dict
handed to plugins now has a top-level separation between device types,
in this revision putting `block` and `fibre` devices under `storage` in
the `devices` dict.

In using `add_device_cmd()`, plugins may now specify either `block`,
`fibre`, or `storage` for both. In addition, any devices passed via the
`devices` parameter that aren't keys in the devices dict are returned,
meaning a plugin may specify a list of devices and add `storage` to
include all storage devices for command iteration.

This change should allow easier addition of network device iteration.

---

### DIFF
--- a/sos/report/__init__.py
+++ b/sos/report/__init__.py
@@ -434,8 +434,10 @@ class SoSReport(SoSComponent):
 
     def _get_hardware_devices(self):
         self.devices = {
-            'block': self.get_block_devs(),
-            'fibre': self.get_fibre_devs()
+            'storage': {
+                'block': self.get_block_devs(),
+                'fibre': self.get_fibre_devs()
+            }
         }
         # TODO: enumerate network devices, preferably with devtype info
 
@@ -474,7 +476,7 @@ class SoSReport(SoSComponent):
         """Enumerate a list of fibrechannel devices on this system so that
         plugins can iterate over them
 
-        These devices are used by add_fibredev_cmd() in the Plugin class.
+        These devices are used by add_device_cmd() in the Plugin class.
         """
         try:
             devs = []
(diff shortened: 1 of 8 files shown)


TurboTurtle:
Note, once #2972 is merged I'll need to update this one to include loop devices. The only question in my mind is do we put loop devices into the block device list always, or keep them separated in a (new) `devices['storage']['loop']` element?

jcastill:
> Note, once #2972 is merged I'll need to update this one to include loop devices. The only question in my mind is do we put loop devices into the block device list always, or keep them separated in a (new) `devices['storage']['loop']` element?

I don't think we should separate them, since loop devices are block devices, And I cannot think what command (other than losetup) would need to be run that other block devs cannot.

pmoravec (approved):
A minor nitpick: we shall update https://github.com/sosreport/sos/blob/main/plugins_overview.py#L15 :)

TurboTurtle:
Rebased to include the loop device fix, waiting for remaining CI jobs to finish.

sos/report/__init__.py
```diff
@@ -496,16 +498,15 @@ def get_block_devs(self):
         """Enumerate a list of block devices on this system so that plugins
         can iterate over them
 
-        These devices are used by add_blockdev_cmd() in the Plugin class.
+        These devices are used by add_device_cmd() in the Plugin class.
         """
         try:
-            device_list = os.listdir('/sys/block')
+            device_list = ["/dev/%s" % d for d in os.listdir('/sys/block')]
```
pmoravec, line 504: Nice improvement :)

sos/report/plugins/scsi.py
```diff
@@ -61,14 +61,13 @@ def setup(self):
         ])
 
         scsi_hosts = glob("/sys/class/scsi_host/*")
-        self.add_blockdev_cmd("udevadm info -a %(dev)s", devices=scsi_hosts,
-                              prepend_path='/sys/class/scsi_host')
+        self.add_device_cmd("udevadm info -a %(dev)s", devices=scsi_hosts)
```
pmoravec, line 64: I dont have means to test this, but I believe the `prepend_path` option is really not further needed.

pmoravec (approved):

sos/report/plugins/scsi.py
TurboTurtle: At the moment, correct. I'm a bit hesitant to remove it because I think there may be a use for it in the future, potentially (e.g. a container-focused sos report that may need some host-level collections as well), but I don't know yet how practical that actually is or if it's entirely irrelevant.

I'd rather leave it for now and remove it after awhile of no usage than pulling it out now and having to replace it later.

PR #2974, opened 2022-07-13
[ssacli] Add ssacli plugin to gather status and config detail for HPE Smart Storage Array controllers.

scality-trevorbenson:
---

- [X] Are any related Issues or existing PRs [properly referenced](https://docs.github.com/en/issues/tracking-your-work-with-issues/creating-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) via a Closes (Issue) or Resolved (PR) line? (No existing Issues or PRs)

### DIFF
--- a/sos/report/plugins/hpssm.py
+++ b/sos/report/plugins/hpssm.py
@@ -7,10 +7,17 @@
 # See the LICENSE file in the source distribution for further information.
 
 from sos.report.plugins import Plugin, IndependentPlugin, PluginOpt
+import re
 
 
 class Hpssm(Plugin, IndependentPlugin):
-
+    """
+    This plugin will capture details for each controller from Smart Storage
+    Array Administrator, an Array diagnostic report from Smart Storage
+    Administrator Diagnostics Utility and, when the plugins debug option is
+    enabled will gather the Active Health System log via the RESTful Interface
+    Tool (iLOREST).
+    """
     short_desc = 'HP Smart Storage Management'
 
     plugin_name = 'hpssm'
@@ -22,25 +29,53 @@ class Hpssm(Plugin, IndependentPlugin):
     ]
 
     def setup(self):
-        self.add_cmd_output([
-            'ssacli ctrl slot=0 array all show detail',
-            'ssacli ctrl slot=0 ld all show detail',
-            'ssacli ctrl slot=0 pd all show detail',
-            'ssacli ctrl slot=0 show detail',
-        ])
+        cmd = 'ssacli'
+        subcmds = [
+            'ctrl all show status'
+        ]
+        slot_subcmds = [
+            'ld all show',
+            'ld all show detail',
+            'pd all show',
+            'pd all show detail'
+        ]
+        self.add_cmd_output(
+            ["%s %s" % (cmd, subcmd) for subcmd in subcmds]
+        )
+
+        pattern = re.compile("^HP.*Smart Array (.*) in Slot ([0123456789])")
+        config_detail_cmd = cmd + ' ctrl all show config detail'
+        config_detail = self.collect_cmd_output(config_detail_cmd)
+        ctrl_slots = []
+        if config_detail['status'] == 0:
+            ctrl_slots = [m.group(2)
+                          for line in config_detail['output'].splitlines()
+                          for m in [pattern.search(line)] if m]
+        ssacli_ctrl_slot_cmd = cmd + ' ctrl slot='
+        self.add_cmd_output(
+            ["%s%s %s" % (
+                ssacli_ctrl_slot_cmd,
+                slot,
+                slot_subcmd
+            )
+             for slot in ctrl_slots
+             for slot_subcmd in slot_subcmds]
+        )
 
         logpath = self.get_cmd_output_path()
 
         self.add_cmd_output(
             'ssaducli -v -adu -f %s/adu-log.zip' % logpath,
-            suggest_filename='ssaducli_-v_-adu.log')
+            suggest_filename='ssaducli_-v_-adu.log'
+        )
 
         if self.get_option("debug"):
             self.do_debug(logpath)
 
     def do_debug(self, logpath):
         self.add_cmd_output(
-            'ilorest serverlogs --selectlog=AHS',
-            runat=logpath, suggest_filename='ilorest.log')
+            'ilorest serverlogs --selectlog=AHS --directorypath=%s' % logpath,
+            runat=logpath, suggest_filename='ilorest.log'
+        )
 
 # vim: set et ts=4 sw=4 :


sos/report/plugins/ssacli.py
```diff
@@ -0,0 +1,50 @@
+# ssacli.py
+
```
TurboTurtle, line 2: Plugins need to have the standard banner message in them at the top, optionally with a copyright line above it. Any other plugin will serve as a fine reference.

sos/report/plugins/ssacli.py
```diff
@@ -0,0 +1,50 @@
...
+            'ld all show detail',
+            'pd all show',
+            'pd all show detail'
+        ]
+        self.add_cmd_output([
+            "%s %s" % (cmd, subcmd) for subcmd in subcmds
+        ])
+
+        pattern = re.compile("^HP.*Smart Array (.*) in Slot ([0123456789])")
+        config_detail_cmd = cmd + 'ctrl all show config detail'
```
TurboTurtle, line 34: Missing a space here before `ctrl`.

sos/report/plugins/ssacli.py
```diff
@@ -0,0 +1,50 @@
...
+    controller(s) configuration on the system.
+    """
+    short_desc = 'HPE Smart Array devices'
+    plugin_name = 'ssacli'
+    profiles = ('system', 'storage', 'hardware')
+    files = ('/opt/smartstorageadmin/ssacli/bin/ssacli',)
+    packages = ('ssacli',)
+
+    def setup(self):
+        cmd = '/opt/smartstorageadmin/ssacli/bin/ssacli'
```
TurboTurtle, line 19: Is this really not typically installed to a user's PATH? Having a bunch of `opt_smartstorage_admin_ssacli_bin_ssacli_*` prefixed outputs isn't exactly easy on the eyes. Perhaps leveraging `add_cmd_output()`'s `runat` parameter could help here? E.G. have `cmd` be the stock `ssacli` and set `runat` to the `/opt/....` path?

TurboTurtle (commented):
Couple thoughts below, however the big one here is that this seems to be largely the same as the `hpssm` plugin. The main differences being that `hpssm` just calls `ssacli` directly, rather than using the `/opt/...` path and the commands are all for `slot=0`, rather than iterating over slots.

What's the difference between this and `hpssm`? My first thought is the current `hpssm` plugin should simply be extended to handle multiple slots, but I'm also not super familiar with hpssm vs. ssacli from a product/vendor perspective.

sos/report/plugins/ssacli.py
scality-trevorbenson: > Couple thoughts below, however the big one here is that this seems to be largely the same as the `hpssm` plugin. The main differences being that `hpssm` just calls `ssacli` directly, rather than using the `/opt/...` path and the commands are all for `slot=0`, rather than iterating over slots.

- My use of `/opt/...` comes from forgetting that HPE actually added a `/usr/bin/ssacli` to the RPM package at some point. So using the full `/opt/` path should no longer be required, it's just a habit I've picked up from when `ssacli` wasn't part of the path.
- `hpssm.py` lists 3 packages `packages = ('ilorest', 'ssacli', 'ssaducli',)` 
  - Only the ssacli plugin contains the binary. 
  - The **ilorest** and **ssaducli** RPM packages do not contain the `ssacli` binary and do not have dependencies requiring `ssacli` RPM be installed with them. 

I believe this would run `ssacli` even when it is not present, generating errors in `sos_log` for the plugin unnecessarily.

> 
> What's the difference between this and `hpssm`? My first thought is the current `hpssm` plugin should simply be extended to handle multiple slots, but I'm also not super familiar with hpssm vs. ssacli from a product/vendor perspective.

Agreed on using existing plugins vs. adding duplicated functionality/intent. I had forgotten (and overlooked) the **hpssm** plugin when reviewing for HPE controllers the other week. I'd be happy to rework this PR into a modification of the existing plugin if nobody minds the change. I also suspect it will be helpful to many, as I have observed systems where slot 0 has no card at all, while slot 1 does and various combinations when more than 1 card is present.


**TL;DR**
 
I am mostly familiar with the **HP SSM** abbreviation being **System Software Manager** which was replaced in 2000 with **Image Assistant**. The plugin says its **Smart Storage Management**, but I cannot find that term used by HPE when referring to Smart Array controllers. The abbreviation I would have expected when searching would have been **HPE SSA** (Smart Storage Administrator). 

Maybe this is a combination of two different abbreviations, or an acronym HPE got rid of enough years ago that even google does not provide a historical reference.

sos/report/plugins/ssacli.py
scality-trevorbenson:
```suggestion
        config_detail_cmd = cmd + ' ctrl all show config detail'
```

sos/report/plugins/ssacli.py
scality-trevorbenson: It is in current versions of the RPM. My pedantic memory recalls when it was not, so I tend to use the full path excessively. Given it is now part of `/usr/sbin/` for a few years I'd be happy to drop this entirely and rely on the defined PATH.

sos/report/plugins/ssacli.py
TurboTurtle: Sorry for the delay.

> I believe this would run ssacli even when it is not present, generating errors in sos_log for the plugin unnecessarily.

Generally not a concern for this project to attempt to run non-present commands - it is handled gracefully within `Plugin()`.

> Maybe this is a combination of two different abbreviations, or an acronym HPE got rid of enough years ago that even google does not provide a historical reference.

Interesting. Well, I also cannot find a reference to SSM in the sense of storage management so perhaps we should take the opportunity to realign what this plugin does.

Are `ilorest` and/or `ssaducli` in anyway _related_ to `ssacli`? If not, perhaps we should split `hpssm` into multiple plugins instead. If it helps, we could keep the `hp` prefixing for these plugins (e.g. `hpssa`, `hpilo`, etc...). 

> I'd be happy to rework this PR into a modification of the existing plugin if nobody minds the change.

Let's do that for the `ssacli` commands at least. Depending upon the answer to the above question, feel free to either split out the others into new plugins as well or just stick to the multiple-slots update (and rename as you see fit).

scality-trevorbenson: > Interesting. Well, I also cannot find a reference to SSM in the sense of storage management so perhaps we should take the opportunity to realign what this plugin does.
> 
> Are ilorest and/or ssaducli in anyway related to ssacli? If not, perhaps we should split hpssm into multiple plugins instead. If it helps, we could keep the hp prefixing for these plugins (e.g. hpssa, hpilo, etc...).

Slightly.

- `ssaducli` is a debug tool for Smart Storage Arrays which comes in a separate RPM. I've rarely installed it and even more rarely used it. 
- `ilorest` is a tool to allow CLI or scripted iLo actions. iLo can reconfigure storage controllers as one area of management.

Neither of these tools are required to investigate physical or logical devices with `ssacli` commands used in the current plugin.

> Let's do that for the ssacli commands at least. Depending upon the answer to the above question, feel free to either split out the others into new plugins as well or just stick to the multiple-slots update (and rename as you see fit).

Sounds good. I'll get something put together for submission this week.

scality-trevorbenson:
@TurboTurtle I just realized another reason why I overlooked the `hpssm` plugin. The `hpssm` plugin apparently doesn't get packaged into the sos RPM for RHEL 7,  8 or 9 distributions. The only "HP" plugin which exists on all three appears to be the `hpasm` plugin. This explains how I overlooked `hpssm` as I was checking only the files the distro's sos RPM installed, but not the github repo's plugins directory.

- Initially I was worried about modifying the `short_description` or plugin file to use the accurate naming.
- It's never been included in RHEL distro's, and only recent Fedora.

Changing the `short_desc` or even the entire filename has less potential negative impact given its never existed in enterprise environments yet. I'll still submit changes to the branch as we discussed, probably including name corrections. I prefer when the community can benefit from code to have it in the upstream repo, unless the code is unwelcome or maintaining a separate fork has a solid reason.

Since the repo is community based (not distro) the following parts of this comment may be off topic for this PR and repository, please let me know if so and I will look for a more appropriate place to hold this discussion.

OFF TOPIC CONTENT
-------------------------------------------------------------------------------------------------------------------------------------------------------

- With a creation date of Oct 26, 2021 I won't expect it to ever reach RHEL7 as its beyond **Full Support Phase**.
- With RHEL8 not yet transitioned to **Maintenance Support Phase** (still within 5 year **Full Support Phase** including RHEA) is it possible this plugin will get included in an upcoming minor point release, like 8.7?
- With RHEL9 barely entered its **Full Support Phase** , but based on Fedora 34, is it likely to make it into an RHEA?
- With Fedora 36 distro the `hpssm` plugin is included. I believe that only suggests that RHEL 10 or newer is likely to have it on GA releases.

Is there a specific process to request or vote on getting items added to RHEA's? I'd hate to see this excluded until a RHEL 10 release in a few more years given the amount of time this hardware has been in production already without any sos plugin gathering data.

TurboTurtle:
> I won't expect it to ever reach RHEL7 as its beyond Full Support Phase.

Correct.

> it possible this plugin will get included in an upcoming minor point release, like 8.7?

If the changes make it into this upstream repo by sos-4.4, it would be included in any enterprise release of that version. Tentatively, we'd be looking at RHEL 8.8 for sos-4.4. If you need these changes for 8.7 or earlier, then you (or someone else) would need to open a bugzilla requesting that backport along with justification as to why it can't wait for a later release.

> With RHEL9 barely entered its Full Support Phase , but based on Fedora 34, is it likely to make it into an RHEA?

Similarly to the above, assuming the changes are merged by the sos-4.4 cut, then it'd be included in the RHEL 9 update that includes the rebase to 4.4. Tentatively, this would align with 9.2.

> With Fedora 36 distro the hpssm plugin is included. I believe that only suggests that RHEL 10 or newer is likely to have it on GA releases.

`sos` is a bit special in RHEL in that we generally do a rebase for sos with each minor release of RHEL (e.g. RHEL 8.4 shipped sos-4.0 and when RHEL 8.5 went GA sos was updated to sos-4.1 while upstream sos-4.2 had just been cut). This is because we have a bit more freedom than other packages/projects in terms of rebases and API changes since 1) no one is leveraging sos as a library and 2) the only real use case for sos on RHEL is interactions with support.

To be clear here, RHEL 8 and RHEL 9 will at some point ship sos-4.4, even if it's not the tentative release above.

Outside of these minor release rebases, anyone is free to raise a backport request on bugzilla.redhat.com to bring specific changes (or new plugins) into the current minor release. However, the main determining factor there is testing capacity.

scality-trevorbenson:
@TurboTurtle OK I came up with 2 separate plugins.

1. **hpssa** plugin
   - Contains both ssacli and ssaducli as both are Smart Storage Administrator utilities.
   - Keeps the current HP naming convention instead of  HPE
   - Changes plugin name. 
     - Felt odd to leave a toolset acronym which HPE does not use and neither does the community at large.
3. **hpilorest** plugin
   - Contains the ilorest functionality to grab server logs
   - Uses existing debug logic. 
   - No default output for the ilorest plugin without debug option True (Default set to False)

TurboTurtle (changes requested):
Overall looks good, just some minor adjustments noted.

One thing though is please squash this changes into either a single commit or two (1 for rewriting hpssm -> hpssa and 1 for adding the new plugin for example)

sos/report/plugins/hpilorest.py
```diff
@@ -9,32 +9,21 @@
...
 
-    plugin_name = 'hpssm'
+    plugin_name = 'hpilorest'
     profiles = ('system', 'storage', 'hardware',)
-    packages = ('ilorest', 'ssacli', 'ssaducli',)
+    packages = ('ilorest',)
 
     option_list = [
         PluginOpt('debug', default=False, desc='capture debug data')
     ]
```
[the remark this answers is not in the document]
scality-trevorbenson: Agreed, this is why I mentioned it above. The other thought I had is if this was only collected by setting the debug option true, because the original plugin was enabled by default for `ssacli` collection, would be to remove the debug altogether, but also set the `ilorest` plugin default to disabled. This would continue to mimic some of the behavior of not collecting `ilorest` unless requested.

I'm open to any of the methods here honestly.

- Leaving it it in `hpssa` with debug
- Remove debug from `ilorest` and
  - Plugin defaults to enabled and collects command output
  - Plugin defaults to disabled, requiring activation to collect output.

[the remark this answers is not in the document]
TurboTurtle: Let's leave it in `hpssa` with the debug option. After thinking on it a bit more, I think it might be seen as a bit heavy handed for a default collection.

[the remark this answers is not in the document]
scality-trevorbenson: I thought through the adjustments, and moving ilorest back into the "main" plugin and leaving debug enabled, it seemed to make more sense to refactor and reduce the changes down.

- Left the original hpssm plugin/file name.
- Updated the section for controller enumeration and gathering details from each with ssacli.
- Added an accurate (although somewhat verbose) docstring
- Updated the style of all `add_cmd_output()` as requested
- Updated the `ilorest` output gathering to use the `--directorypath` flag. I suspect `runat` will handle it but this ensures the output is where it is expected if the working directory is no longer the default output path.

TurboTurtle (approved):
LGTM.

flake8 failures are unrelated, and are already fixed in `main`.

scality-trevorbenson:
> LGTM.
> 
> flake8 failures are unrelated, and are already fixed in `main`.

Great, I'll rebase shortly.

Do you prefer a comment post rebase, since review is complete, or do you prefer a review request "refresh"?

TurboTurtle:
No need to rebase here. Your changes don't touch the already-fixed-in-main files.

PR #2977, opened 2022-07-18
[openvswitch] Extend package version pattern

david-marchand:
The next OVS release will be numbered 3.0.
In a near future, the Red Hat Fast Datapath channel will provide
packages versionned in a same manner.

Signed-off-by: David Marchand <david.marchand@redhat.com>

---

### DIFF
--- a/sos/report/plugins/openvswitch.py
+++ b/sos/report/plugins/openvswitch.py
@@ -365,7 +365,7 @@ class OpenVSwitch(Plugin):
 
 class RedHatOpenVSwitch(OpenVSwitch, RedHatPlugin):
 
-    packages = ('openvswitch', 'openvswitch2.*',
+    packages = ('openvswitch', 'openvswitch[2-9].*',
                 'openvswitch-dpdk', 'nuage-openvswitch'
                 '6windgate-fp')
 


TurboTurtle (commented):
The change works, but is there a reason we shouldn't instead change this to `openswitch(\d\.*)?`  or similar? I.E. are there other openvswitch packages out there that we _shouldn't_ enable this plugin on?

david-marchand:
- I know there is a openvswitch-selinux-extra-policy package, which does not have a dependency on openvswitch packages.
So its presence does not imply that ovs is available.
I guess this is why we don't have a simple openvswitch* pattern.


- Now, we can improve with a larger pattern for the version part.
But I am not sure it works as suggested.

Afaics packages pattern ends up being passed to fnmatch (check_enabled -> _check_plugin_triggers -> is_installed -> policy.package_manager.all_pkgs_by_name -> fnmatch.filter).
I would expect python fnmatch matches behavior of the C fnmatch and so this proposal that looks like a regexp would not work.

So maybe something like, replacing 'openvswitch2.*' with 'openvswitch[2-9].*' is better?

david-marchand:
Yes, I'll update accordingly.

pmoravec (approved):

TurboTurtle (approved):

Issue #2790, opened 2021-12-06
Uncaught exception when collecting file with non-UTF-8 name


pmoravec:
Reproducer:

    date > '/etc/yum.repos.d/team'$'\303''_epel.repo' 
    python3 bin/sos report -o yum --batch
    ..
    --- Logging error ---
    Traceback (most recent call last):
      File "/usr/lib64/python3.6/logging/__init__.py", line 996, in emit
        stream.write(msg)
    UnicodeEncodeError: 'utf-8' codec can't encode character '\udcc3' in position 81: surrogates not allowed
    Call stack:
      File "/usr/lib64/python3.6/threading.py", line 884, in _bootstrap
        self._bootstrap_inner()
      File "/usr/lib64/python3.6/threading.py", line 916, in _bootstrap_inner
        self.run()
      File "/usr/lib64/python3.6/threading.py", line 864, in run
        self._target(*self._args, **self._kwargs)
      File "/usr/lib64/python3.6/concurrent/futures/thread.py", line 69, in _worker
        work_item.run()
      File "/usr/lib64/python3.6/concurrent/futures/thread.py", line 56, in run
        result = self.fn(*self.args, **self.kwargs)
      File "/root/sos-main/sos/report/__init__.py", line 1120, in collect_plugin
        plug.collect()
      File "/root/sos-main/sos/report/plugins/__init__.py", line 2676, in collect
        self._collect_copy_specs()
      File "/root/sos-main/sos/report/plugins/__init__.py", line 2640, in _collect_copy_specs
        self._log_info("collecting path '%s'" % path)
      File "/root/sos-main/sos/report/plugins/__init__.py", line 712, in _log_info
        self.soslog.info(self._format_msg(msg))
    Message: "[plugin:yum] collecting path '/etc/yum.repos.d/team\udcc3_epel.repo'"
    Arguments: ()

The problem is we use Python `logging` library that does not support nonUTF-8 characters.

One not-so-nice fix is in sos/report/plugins/__init__.py : 

        def _format_msg(self, msg):
-            return "[plugin:%s] %s" % (self.name(), msg)
+           return "[plugin:%s] %s" % (self.name(), msg.encode('utf-8', 'ignore'))

while the same change needs to be applied to `generate_reports` method, to prevent:

    Traceback (most recent call last):
      File "bin/sos", line 22, in <module>
        sos.execute()
      File "/root/sos-main/sos/__init__.py", line 186, in execute
        self._component.execute()
[...]

PR #2979, opened 2022-07-20
[logging] Safeguard against attempting to log non-UTF characters

TurboTurtle:
The `logging` module does not support non UTF-8 characters, and as such
there was an exception being thrown when trying to log for example a
copied file that has such characters in the file name.

Safeguard against this by converting all messages logged via
`Plugin._format_msg()` to UTF-8, and replace any problematic characters
with `?`s to attempt to preserve context within the logs.

Closes: #2790

---

### DIFF
--- a/sos/report/__init__.py
+++ b/sos/report/__init__.py
@@ -1296,6 +1296,8 @@ class SoSReport(SoSComponent):
             try:
                 fd = self.get_temp_file()
                 output = class_(report).unicode()
+                # safeguard against non-UTF characters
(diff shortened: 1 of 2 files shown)


TurboTurtle:
I tried to come up with a test case for this, however `shutil.copy()` breaks on something like `'/etc/yum.repos.d/team'$'\303''_epel.repo' ` from the original issue report, and any work around to that introduces stuff that would invalidate any test for it. :\

pmoravec (approved):
Nice!

PR #2980, opened 2022-07-21
[report,Plugin] Enumerate network devices for plugins

TurboTurtle:
This patchset first adds enumeration of network devices to the `devices` dict handed to plugins for device iteration. Second, it updates `add_device_cmd()` to allow handling of the `subdir` parameter, and finally it updates various network-related plugins to use `add_device_cmd()` instead of enumerating devices themselves.

---

### DIFF
--- a/sos/report/__init__.py
+++ b/sos/report/__init__.py
@@ -18,7 +18,8 @@ from datetime import datetime
 import glob
 import sos.report.plugins
 from sos.utilities import (ImporterHelper, SoSTimeoutError, bold,
-                           sos_get_command_output, TIMEOUT_DEFAULT)
+                           sos_get_command_output, TIMEOUT_DEFAULT, listdir,
+                           is_executable)
 from shutil import rmtree
 import hashlib
 from concurrent.futures import ThreadPoolExecutor, TimeoutError
@@ -190,8 +191,8 @@ class SoSReport(SoSComponent):
             self._exit(1)
 
         self._check_container_runtime()
-        self._get_hardware_devices()
         self._get_namespaces()
+        self._get_hardware_devices()
 
     @classmethod
     def add_parser_options(cls, parser):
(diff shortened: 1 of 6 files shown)


TurboTurtle:
Hmmm...it's strange that centos is getting a new kmod loaded now (team) only on one test...especially when there are no `team` devices on the test VMs...

sos/report/__init__.py
```diff
@@ -522,6 +524,131 @@ def _get_namespaces(self):
...
+                dev for dev in listdir('/sys/class/net', self.opts.sysroot)
+                if dev not in ['bonding_masters']
+            ]
+        else:
+            try:
+                _nscmd = "ip netns exec %s ls /sys/class/net" % namespace
+                _nsout = sos_get_command_output(_nscmd)
+                if _nsout['status'] == 0:
+                    for _nseth in _nsout['output'].split():
+                        _eth_devs.append(_nseth)
```
pmoravec, line 617: Devices in a namespace can't be `bonding_masters` (cf. line 609)?

sos/report/__init__.py
```diff
@@ -522,6 +524,131 @@ def _get_namespaces(self):
...
+
+        :returns:   All valid ethernet devices found, potentially within a
+                    given namespace
+        :rtype:     ``list``
+        """
+        _eth_devs = []
+        if not namespace:
+            _eth_devs = [
+                dev for dev in listdir('/sys/class/net', self.opts.sysroot)
+                if dev not in ['bonding_masters']
```
pmoravec, line 609: Isn't `if dev != 'bonding_master'` easier comparison?

sos/report/plugins/networking.py
```diff
@@ -141,33 +141,28 @@ def setup(self):
...
+        ], devices='ethernet')
+
+        self.add_device_cmd([
+            "ethtool %(dev)s",
+            "ethtool --phy-statistics %(dev)s",
+            "ethtool --show-priv-flags %(dev)s",
+            "ethtool --show-eee %(dev)s",
+            "tc -s filter show dev %(dev)s",
+            "tc -s filter show dev %(dev)s ingress",
+        ], devices="ethernet")
```
pmoravec, line 156: .. and the (here) missing `tags=eth` is added automatically via `_add_device_cmd default method option `tags=_dev_tags`, so no "regression" :)

pmoravec:
Just the two very minor comments, otherwise great!

sos/report/__init__.py
TurboTurtle: Initial thinking was that we'd have a handful of skip patterns, or at least one that might build up over time.

Issue #2882, opened 2022-03-15
Improve sos collect cleanup


npinaeva:
Problem:
interrupting sos collect on the `prep()` stage causes no cleanup even though cluster prep has already been executed and needs cleanup.

With a lot of different exit points and complex exceptions handling, it is not assured now that all the code pathes have cleanup calls. 
E.g. `collector.exit()` calls `cluster.cleanup()` first then some more actions then `self.cleanup()` https://github.com/sosreport/sos/blob/main/sos/collector/__init__.py#L492-L500.
and cleanup for `collector.execute()` looks different https://github.com/sosreport/sos/blob/main/sos/collector/__init__.py#L1157-L1158
and cleanup in `collector.collect()` is also different https://github.com/sosreport/sos/blob/main/sos/collector/__init__.py#L1215-L1216

Suggestion:
It would be nice to have some common `cleanup()` function that can be called before every exit point. Also adding try/finally block at least here https://github.com/sosreport/sos/blob/main/sos/collector/__init__.py#L1140-L1156 with cleanup being called in `finally`.
Some more changes may be required to be sure there is no code path without a cleanup call.

Unfortunately I am not aware about all resources sos collect uses, so I can't implement it myself

TurboTurtle:
> Problem: interrupting sos collect on the `prep()` stage causes no cleanup even though cluster prep has already been executed and needs cleanup.
> [...]
> Suggestion: It would be nice to have some common `cleanup()` function that can be called before every exit point. Also adding try/finally block at least here https://github.com/sosreport/sos/blob/main/sos/collector/__init__.py#L1140-L1156 with cleanup being called in `finally`. Some more changes may be required to be sure there is no code path without a cleanup call.
> 

Ack, we should handle failures in `prep()` more distinctly. 


> With a lot of different exit points and complex exceptions handling, it is not assured now that all the code pathes have cleanup calls. E.g. `collector.exit()` calls `cluster.cleanup()` first then some more actions then `self.cleanup()` https://github.com/sosreport/sos/blob/main/sos/collector/__init__.py#L492-L500. and cleanup for `collector.execute()` looks different https://github.com/sosreport/sos/blob/main/sos/collector/__init__.py#L1157-L1158 and cleanup in `collector.collect()` is also different https://github.com/sosreport/sos/blob/main/sos/collector/__init__.py#L1215-L1216
> 

Let me break this out a bit.

[...]

PR #2985, opened 2022-07-28
[collect] Standardize use of `exit()`

TurboTurtle:
There were multiple exit paths and behaviors within sos collect, that
were not the easiest to logically follow. Fix this by standardizing on
leveraging `self.exit()` to exit in all situations. If a message is
provided to this call, it is regarded as an error message and logged as
such. This method will still set the exit code according to the `error`
parameter, however that has been changed to default to 0 to signify a
"clean" end of execution.

Finally, there is a new `force` parameter which is meant to be used when
exiting from within a child thread when the entire process needs to exit
(and thus needs to leverage the lower-level `os._exit()` instead).

Closes: #2882

---

### DIFF
--- a/sos/collector/__init__.py
+++ b/sos/collector/__init__.py
@@ -493,17 +493,35 @@ class SoSCollector(SoSComponent):
                 newline=False
             )
 
-    def exit(self, msg, error=1):
-        """Used to safely terminate if sos-collector encounters an error"""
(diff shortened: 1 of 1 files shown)


pmoravec (approved):

TurboTurtle:
Minor change to keep the tempdir in cases of keyboard interrupts, like we currently do and have done for report for many years.

PR #2986, opened 2022-07-29
[auditd] Capture log files when configured to a non default location

jcastill:
The location of the audit.log files can be changed
in the configuration file /etc/audit/audit.conf.
This change ensures that we capture the log files
when the user specifies a different location via
log_file.

---

### DIFF
--- a/sos/report/plugins/auditd.py
+++ b/sos/report/plugins/auditd.py
@@ -33,9 +33,23 @@ class Auditd(Plugin, IndependentPlugin):
             "auditctl -l"
         ])
 
+        config_file = "/etc/audit/auditd.conf"
+        log_file = "/var/log/audit/audit.log"
+        try:
+            with open(config_file, 'r') as cf:
+                for line in cf.read().splitlines():
+                    if not line:
+                        continue
+                    words = line.split('=')
+                    if words[0].strip() == 'log_file':
+                        log_file = words[1].strip()
+        except IOError as error:
+            self._log_error('Could not open conf file %s: %s' %
+                            (config_file, error))
+
         if not self.get_option("all_logs"):
-            self.add_copy_spec("/var/log/audit/audit.log")
+            self.add_copy_spec(log_file)
         else:
-            self.add_copy_spec("/var/log/audit")
+            self.add_copy_spec(log_file+'*')
 
 # vim: set et ts=4 sw=4 :


sos/report/plugins/auditd.py
```diff
@@ -33,9 +33,20 @@ def setup(self):
             "auditctl -l"
         ])
 
+        config_file = "/etc/audit/auditd.conf"
+        log_file = "/var/log/audit/audit.log"
+        try:
+            for line in open(config_file).read().splitlines():
```
TurboTurtle, line 39: This can leave an open file handle since there is no context manager or an explicit close. Using a context manager like below will ensure it gets closed once we're out of this loop.

```python
with open(config_file, 'r') as cf:
    for line in cf.read().splitlines()
        [...]
```

sos/report/plugins/auditd.py
```diff
@@ -33,9 +33,20 @@ def setup(self):
...
+                words = line.split('=')
+                if words[0].strip() == 'log_file':
```
TurboTurtle, line 41: I think this may have issues on empty lines. We'll want to either wrap this in a try block, or put some other safeguard against empty lines in this loop.

sos/report/plugins/auditd.py
jcastill: And the 'with open...' still needs to happen inside the  try section, to check if the file exist, right?

TurboTurtle: Correct

sos/report/plugins/auditd.py
jcastill: I've done some tests for the empty lines and so far I haven't got any backtrace. It looks like read() manages well these lines. But just to be on the safe side, I can add a safeguard. What about a check that word is not '' before splitting? Would that work?

TurboTurtle: Either `if not line: continue` or a try/except block would be fine.

jcastill:
I think I've added all suggestions. Tested them as well and works as expected. But let me know if I missed something.

TurboTurtle (approved):
LGTM.

The Fedora test failed due to a webhook issue it looks like, so I've restarted it.

pmoravec (approved):

PR #2993, opened 2022-08-04
[teamd,tests] Add a test for teamd plugin and device enumeration

TurboTurtle:
Adds a new test case for the `teamd` plugin that also inherently acts to
test team device enumeration.

Included with this test case is an addition to the base test classes
that allows for tests to define a `post_test_tear_down()` method that
will be run at the end of each test execution to allow for manual
cleanup - in this case deleting a 'fake' team device created for the
test.

---

### DIFF
--- a/tests/report_tests/plugin_tests/teamd.py
+++ b/tests/report_tests/plugin_tests/teamd.py
@@ -1,0 +1,46 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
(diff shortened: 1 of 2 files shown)


TurboTurtle:
It seems there's some ambiguity revolving around if/when `nmcli` will report the fake team device in `nmcli dev` output. I'm trying to dig into this now. It appears that Ubuntu doesn't ship with NetworkManager, so for the time being I've set this test to redhat only.

TurboTurtle:
Turns out there's a `NetworkManager-team` package which, if installed, will cause `nmcli` to always report team connections as devices. What is happening under the hood for this is beyond me, but that package is the key for it.

I've left the test disable for Ubuntu/Debian for the reason above. Should be good now.

pmoravec (approved):

PR #3005, opened 2022-08-17
[libvirt] Fix missing libvirt.all_logs option in libvirt plugin

osherdp:
Seems like there's use of 'all_logs' option in the code of the libvirt
plugin, but there's no definition of it as an available option in the
class. This change just adds it so that an external user can use it.

I tested it with:
```
sudo `which sos` report -o libvirt  -k libvirt.all_logs
```

and by checking:

```
sudo `which sos` report -l | grep "libvirt.all_logs
```

---

### DIFF
--- a/sos/report/plugins/libvirt.py
+++ b/sos/report/plugins/libvirt.py
@@ -6,7 +6,7 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.report.plugins import Plugin, IndependentPlugin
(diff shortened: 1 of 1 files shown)


TurboTurtle (commented):
Nack. The `all_logs` option is a global option, specified via `--all-logs` on the command line. It is not plugin specific.

osherdp:
> Nack. The `all_logs` option is a global option, specified via `--all-logs` on the command line. It is not plugin specific.

ACK, I guess I have missed this option
I'll give it a try. Thank you!

PR #3006, opened 2022-08-17
[tests] Allow testing from local installation, use PR-built packages in Cirrus CI

TurboTurtle:
We've built RPMs of each PR in Cirrus for some time, and recently added snaps as well.

We should use these built artifacts for more complete testing than what can be gained from running
from a git checkout of the PR.

This patchset allows the test suite to use a locally installed version of sos instead of forcing the use of the git checkout. This can be leveraged by adding `-p TESTLOCAL=true` to a given `avocado run` command. This is then brought forward in to the tasks run by Cirrus CI, using the rpm and snap built for each PR.

---

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -33,6 +33,13 @@ env:
     UBUNTU_IMAGE_NAME: "ubuntu-2204-jammy-v20220622"
     UBUNTU_PRIOR_IMAGE_NAME: "ubuntu-2004-focal-v20220701"
 
+    # Curl-command prefix for downloading task artifacts, simply add the
+    # the url-encoded task name, artifact name, and path as a suffix.
+    # This approach originally seen in the podman project.
+    ARTCURL: >-
+        curl --fail --location -O
+        --url https://api.cirrus-ci.com/v1/artifact/build/${CIRRUS_BUILD_ID}
+
 # Default task timeout
 timeout_in: 30m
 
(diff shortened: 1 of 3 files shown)


TurboTurtle:
Hmm, the rpm build is now failing moving it to a VM instead of a container. I would prefer if we could leverage the PackIt built copr repo in the various stages, but I don't know how/if Cirrus can be made to depend on external tasks. I'm going to look into that so that we can maybe avoid duplicating rpm builds.

TurboTurtle:
Hmm I had tried to consolidate the avocado commands to make expansion of tags easier, but it appears there's something else going on when putting the command in an env var that prevents test resolution.

I'll revert that to stop spamming the updates, I think we're at a place here where we can move forward with the use of built RPMs.

Looking at leveraging the packit builds, there's an issue in that GCP availability of Fedora images doesn't match the builds produced by packit. Notably, F36 is still not available on GCP, so we continue to test on F34 which packit doesn't build for any more. Our options would be 1) Take on the maintenance overhead of rebuilding Fedora images as packit moves forward, or 2) drop packit and keep doing the builds "manually" with Cirrus.

Option 2 here I personally think is preferable as we can easily control the testing matrix in a single place and ensure that the rpm we build are in fact usable by the versions we test them on.

pmoravec (approved):
+1 for the option 2.

PR #3009, opened 2022-08-19
[soscollect:ocp] Fix ocp get nodes and sos container priviledges

npinaeva:
---

### DIFF
--- a/sos/collector/clusters/ocp.py
+++ b/sos/collector/clusters/ocp.py
@@ -9,6 +9,8 @@
 # See the LICENSE file in the source distribution for further information.
 
 import os
+import tempfile
+import json
 
 from pipes import quote
 from sos.collector.clusters import Cluster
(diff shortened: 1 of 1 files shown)


sos/collector/clusters/ocp.py
```diff
@@ -122,6 +124,41 @@ def check_enabled(self):
         _who = self.fmt_oc_cmd('whoami')
         return self.exec_primary_cmd(_who)['status'] == 0
 
+    def create_clusterrolebinding(self, namespace):
```
TurboTurtle, line 127: What's the need for this? Is there some deployment/config that would cause the existing approach to fail?

sos/collector/clusters/ocp.py
```diff
@@ -152,6 +188,16 @@ def cleanup(self):
...
+            # cleanup cluster role binding
+            if os.path.exists(self.rbac_tmp):
+                os.unlink(self.rbac_tmp)
+            ret = self.exec_primary_cmd(
+                self.fmt_oc_cmd("delete clusterrolebinding %s"
+                                % self.rbac_name))
+            if ret['status'] != 0 or "deleted" not in ret['output']:
+                self.log_debug("Calling delete on clusterrolebinding '%s' "
+                               "failed: %s" % (self.rbac_name, ret['output']))
+                return False
```
TurboTurtle, line 200: If we exit here, we won't delete the project, which could cause issues for subsequent runs. Shouldn't we always (attempt to) delete the project at the end of a collection?

Also, while the existing debug message is fine so that we can dig into a particular failure, I think a user-level error notification would be helpful. It doesn't have to specify the reason, a simple "failed to delete sos clusterrolebinding" or similar would suffice.

TurboTurtle (commented):
Couple comments follow below.

If the reason this is needed could be captured in the commit message, that'd be helpful - especially when looking back through the git history at some point in the future.

Issue #1350, opened 2018-06-18
[process] record package mapping for running process binaries


bmr-cymru:
Collect data for the package corresponding to each running process's binary. This needs to be reasonably efficient: i.e. not invoking a separate RPM process for every PID. One way to do this would be to pack each binary path into a single RPM call, and have the package manager sort out the details:

```
$ ps --no-headers aex | grep -v '\[.*\]' | awk '{print $5}' | xargs rpm -qf
systemd-233-7.fc26.x86_64
systemd-233-7.fc26.x86_64
systemd-udev-233-7.fc26.x86_64
audit-2.8.2-1.fc26.x86_64
audit-2.8.2-1.fc26.x86_64
setroubleshoot-server-3.3.14-1.fc26.x86_64
dbus-1.11.18-1.fc26.x86_64
mcelog-153-1.fc26.x86_64
rng-tools-5-9.fc26.x86_64
udisks2-2.6.5-4.fc26.x86_64
rtkit-0.11-15.fc26.x86_64
accountsservice-0.6.42-2.fc26.x86_64
ModemManager-1.6.10-1.fc26.x86_64
python3-3.6.5-1.fc26.x86_64
systemd-233-7.fc26.x86_64
alsa-utils-1.1.4-1.fc26.x86_64
bluez-5.46-6.fc26.x86_64
switcheroo-control-1.1-2.fc26.x86_64
gssproxy-0.7.0-9.fc26.x86_64
chrony-3.2-1.fc26.x86_64
abrt-2.10.3-4.fc26.x86_64
error: file /home/breeves/avahi-daemon:: No such file or directory
polkit-0.113-8.fc26.x86_64
abrt-addon-xorg-2.10.3-4.fc26.x86_64
abrt-addon-ccpp-2.10.3-4.fc26.x86_64
abrt-addon-kerneloops-2.10.3-4.fc26.x86_64
abrt-dbus-2.10.3-4.fc26.x86_64
NetworkManager-1.8.2-4.fc26.x86_64
libvirt-daemon-3.2.1-7.fc26.x86_64
openssh-server-7.5p1-4.fc26.x86_64
at-3.1.20-3.fc26.x86_64
cronie-1.5.1-5.fc26.x86_64
gdm-3.24.3-1.fc26.x86_64
systemd-233-7.fc26.x86_64
systemd-233-7.fc26.x86_64
error: file /home/breeves/(sd-pam): No such file or directory
gdm-3.24.3-1.fc26.x86_64
[...]

PR #3011, opened 2022-08-19
[Plugin] Allow plugins to perform manual collections

TurboTurtle:
This patchset implements changes that allow plugins to perform manual data collections/compilations that will get saved in the archive alongside normal command output collections.

First, adjust the behavior of collecting `tail`s of file specs that exceed the given log size so that it is no longer performed during `setup()`, but instead the actual collection takes place during the collection phase.

Second, renamed the `collect()` method to `collect_plugin()`, and implement `collect()` as a new entry point for plugin authors to define their manual collections.

The patches following the first two adjust existing plugins to use this new functionality. Additionally, the `process` plugin has been updated to collect a pid->package mapping, as has been long requested in #1350. This includes an update to `PackageManager` to facility these kinds of look ups. 

---

### DIFF
--- a/sos/policies/package_managers/__init__.py
+++ b/sos/policies/package_managers/__init__.py
@@ -50,12 +50,13 @@ class PackageManager():
     verify_command = None
     verify_filter = None
     files_command = None
+    query_path_command = None
     chroot = None
     files = None
 
-    def __init__(self, chroot=None, query_command=None,
-                 verify_command=None, verify_filter=None,
-                 files_command=None, remote_exec=None):
+    def __init__(self, chroot=None, query_command=None, verify_command=None,
+                 verify_filter=None, files_command=None,
+                 query_path_command=None, remote_exec=None):
         self._packages = {}
         self.files = []
 
@@ -63,6 +64,7 @@ class PackageManager():
         self.verify_command = verify_command or self.verify_command
         self.verify_filter = verify_filter or self.verify_filter
         self.files_command = files_command or self.files_command
+        self.query_path_command = query_path_command or self.query_path_command
 
         # if needed, append the remote command to these so that this returns
         # the remote package details, not local
@@ -191,6 +193,24 @@ class PackageManager():
             files = shell_out(cmd, timeout=0, chroot=self.chroot)
             self.files = files.splitlines()
         return self.files
+
+    def pkg_by_path(self, path):
+        """Given a path, return the package that owns that path.
+
+        :param path:    The filepath to check for package ownership
+        :type path:     ``str``
+
+        :returns:       The package name or 'unknown'
+        :rtype:         ``str``
+        """
+        if not self.query_path_command:
+            return 'unknown'
+        try:
+            cmd = f"{self.query_path_command} {path}"
+            pkg = shell_out(cmd, timeout=5, chroot=self.chroot)
+            return pkg.splitlines() or 'unknown'
+        except Exception:
+            return 'unknown'
 
     def build_verify_command(self, packages):
         """build_verify_command(self, packages) -> str
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -562,6 +562,7 @@ class Plugin():
         self.skip_files = commons['cmdlineopts'].skip_files
         self.skip_commands = commons['cmdlineopts'].skip_commands
         self.default_environment = {}
+        self._tail_files_list = []
 
         self.soslog = self.commons['soslog'] if 'soslog' in self.commons \
             else logging.getLogger('sos')
@@ -1836,24 +1837,21 @@ class Plugin():
                 if sizelimit and current_size > sizelimit:
                     limit_reached = True
 
-                    if tailit and not file_is_binary(_file):
-                        self._log_info("collecting tail of '%s' due to size "
-                                       "limit" % _file)
-                        file_name = _file
-                        if file_name[0] == os.sep:
-                            file_name = file_name.lstrip(os.sep)
-                        strfile = (
-                            file_name.replace(os.path.sep, ".") + ".tailed"
+                    if tailit:
+                        if file_is_binary(_file):
+                            self._log_info(
+                                "File '%s' is over size limit and is binary. "
+                                "Skipping collection." % _file
+                            )
+                            continue
+
+                        self._log_info(
+                            "File '%s' is over size limit, will instead tail "
+                            "the file during collection phase." % _file
                         )
                         add_size = sizelimit + file_size - current_size
-                        self.add_string_as_file(tail(_file, add_size), strfile)
-                        rel_path = os.path.relpath('/', os.path.dirname(_file))
-                        link_path = os.path.join(rel_path, 'sos_strings',
-                                                 self.name(), strfile)
-                        self.archive.add_link(link_path, _file)
+                        self._tail_files_list.append((_file, add_size))
                         _manifest_files.append(_file.lstrip('/'))
-                    else:
-                        self._log_info("skipping '%s' over size limit" % _file)
                 else:
                     # size limit not exceeded, copy the file
                     _manifest_files.append(_file.lstrip('/'))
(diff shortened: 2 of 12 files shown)


sos/report/plugins/__init__.py
```diff
@@ -3105,14 +3107,79 @@ def _collect_strings(self):
...
+            _file = open(_name, 'w')
+            self._log_debug(f"manual collection file opened: {_name}")
+            yield _file
+            _file.close()
+            end = time()
+            run = end - start
+            self._log_info(f"manual collection '{fname}' finished in {run}")
+            if isinstance(tags, str):
+                tags = [tags]
+            self.manifest.collections[fname.split('.')[0]] = {
```
pmoravec, line 3167: Why truncating the `fname` at the first `.`? What if a plugin wants to collect `foo.bar1.txt` and `foo.bar2.txt`?

tests/unittests/plugin_tests.py
```diff
@@ -388,10 +388,9 @@ def test_glob_file_over_limit(self):
         create_file(2, dir=tmpdir)
         create_file(2, dir=tmpdir)
         self.mp.add_copy_spec(tmpdir + "/*", 1)
-        self.assertEquals(len(self.mp.copy_strings), 1)
-        content, fname, _tags = self.mp.copy_strings[0]
-        self.assertTrue("tailed" in fname)
```
pmoravec, line 393: We stop testing such an assert here - worth comparing filenames like 

    fn = create_file(2, dir=tmpdir)
    ..
    self.assertTrue(fname == fn)

(likewise changed in `test_single_file_over_limit`)?

sos/report/plugins/process.py
```diff
@@ -88,4 +89,28 @@ def setup(self):
...
+                pfile.write("Unable to get process list: %s" % _ps['output'])
+                return
+            for proc in _ps['output'].splitlines():
+                proc = proc.strip().split()
+                pid = proc[0]
+                path = proc[4]
+                if not self.path_exists(path):
+                    continue
+                if path not in paths:
+                    paths[path] = self.policy.package_manager.pkg_by_path(path)
```
pmoravec, line 111: Rename `paths` variable to `pkgs`? As it tells us (for a given path) a package.

sos/policies/package_managers/__init__.py
```diff
@@ -192,6 +194,24 @@ def all_files(self):
...
+
+        :returns:       The package name or 'unknown'
+        :rtype:         ``str``
+        """
+        if not self.query_path_command:
+            return 'unknown'
+        try:
+            cmd = f"{self.query_path_command} {path}"
+            pkg = shell_out(cmd, timeout=5, chroot=self.chroot)
+            return pkg.replace('\n', ',') or 'unknown'
```
pmoravec, line 211: Why the replace? For a case when a file path is owned by multiple packages? As this generates `pids_to_packages.json` with unexpected commas:

        "1": {
            "path": "/usr/lib/systemd/systemd",
            "package": "systemd-239-51.el8.x86_64,"
        },

(or add there `.rstrip(',')` ?)

pmoravec:
Ah this felt down from my radar, reviewed it now.

Generally big +1, just minor comments. And one doubt: should we collect the manual collection files under `sos_commands`? As it is not a `sos`s command (output).

The manual collections of a plugin *should* sit in the directory with the plugin commands, that is the most natural or implicit place. Renaming `sos_commands` to, say, `sos_plugins` would be more appropriate, BUT we would break very stable user experience (and tools relying on `sos_commands/<plugin>` path).

sos/report/plugins/__init__.py
TurboTurtle: Hmm. I had only been considering a plugin writing something like `foo.txt` or `foo.tar.gz` or `foo.json`, as that's what we have today for these manual collections. But it's a valid concern that a future plugin may want to format file names like that.

tests/unittests/plugin_tests.py
TurboTurtle: Not for this test anymore - we're testing size collection size here, and we no longer write the file to disk (so no name to compare to) within `add_copy_spec()` for files over the limit any more. We'd have to have this test go all the way thru `collect_plugin()` now to test the file naming, which is still handled elsewhere.

sos/report/plugins/process.py
TurboTurtle: `paths` is named that way, as we're holding the path string from the process listing there, which is what we're using to check the package. In other words "this (paths) dict has keys that are paths, and those keys' values are package(s) that own that path".

sos/policies/package_managers/__init__.py
TurboTurtle: Ack to stripping a trailing comma.

TurboTurtle:
> Ah this felt down from my radar, reviewed it now.
> 
> Generally big +1, just minor comments. And one doubt: should we collect the manual collection files under `sos_commands`? As it is not a `sos`s command (output).
> 
> The manual collections of a plugin _should_ sit in the directory with the plugin commands, that is the most natural or implicit place. Renaming `sos_commands` to, say, `sos_plugins` would be more appropriate, BUT we would break very stable user experience (and tools relying on `sos_commands/<plugin>` path).

Yeah, changing the dir name of `sos_commands` I think is too heavy a change in this PR given the historical precedence. Perhaps we could look into this as a separate change - but we'd need to be very careful and have, in my opinion, at least several releases where we carry both (one being a symlink to the other).

sos/report/plugins/__init__.py
TurboTurtle: Dropped the stripping of `fname` - collections entries in the manifest are now the regular file name.

sos/report/plugins/process.py
```diff
@@ -88,4 +89,28 @@ def setup(self):
...
+                pfile.write(f"Unable to get process list: {_ps['output']}")
+                return
+            for proc in _ps['output'].splitlines():
+                proc = proc.strip().split()
+                pid = proc[0]
+                path = proc[4]
+                if not self.path_exists(path):
+                    continue
+                if path not in paths:
+                    paths[path] = self.policy.package_manager.pkg_by_path(path)
```
pmoravec, line 111: Calling external packaging command (like, say, `rpm -qf <binary>`) for each and every binary *separately* takes some time. I suggest (once reviewed and improved) idea "call rpm -qf on all binaries once, parse the output", as that is quite faster.

I did some comparison on a small RHEL8 system and also on a bit busy one (320 processes), comparing `sos report -o process --batch --build` execution times for:
1) `process` plugin empty (`return` at the beginning of `setup()` and `collect()` methods)
2) the plugin calling just `collect()` from this PR
3) the plugin calling just `collect()` which calls the package query once

idle RHEL8 system, average values from 30 tests:
- no `collect`: `3.15687s`
- `collect` from this PR: `3.8576s`
- `collect` from my suggestion: `3.27247s`

busier RHEL system, average values from 30 tests:
- no `collect`: `3.65807s`
- `collect` from this PR: `4.60553s`
- `collect` from my suggestion: `3.8297s`


My `collect()` method (which relies three lists have always the same length, maybe too fragile construction?). The implementation does contain identical paths multiple times within the one query (if multiple processes run the same binary), yet still it is faster than original approach:

    def collect(self):
        with self.collection_file('pids_to_packages.json') as pfile:
            if not self.policy.package_manager.query_path_command:
                pfile.write('Package manager not configured for path queries')
                return
            _ps = self.exec_cmd('ps --no-headers aex')
            pidpkg = {}
            paths = {}
            if not _ps['status'] == 0:
                pfile.write(f"Unable to get process list: {_ps['output']}")
                return
            paths = []
            pids = []
            for proc in _ps['output'].splitlines():
                proc = proc.strip().split()
                pid = proc[0]
                path = proc[4]
                if not self.path_exists(path):
                    continue
                pids.append(pid)
                paths.append(path)

            pkgs = self.policy.package_manager.pkg_by_path(' '.join(paths))
            for i in range(len(pids)):
                pidpkg[pids[i]] = {'path': paths[i], 'package': pkgs[i]}
            pfile.write(json.dumps(pidpkg, indent=4))

TurboTurtle: Testing this out, multiple packages for the same path seem to throw this off due to the requirement of equal-length lists:

```
    "1208": {
        "path": "/usr/bin/python3",
        "package": "python3-3.10.6-1.fc35.x86_64"
    },
... later ...
    "3657": {
        "path": "/usr/bin/python3",
        "package": "evolution-data-server-3.42.5-2.fc35.x86_64"
    },


and from CLI:

$ rpm -qf /usr/bin/python3
python3-3.10.6-1.fc35.x86_64
python3-3.10.6-1.fc35.i686
```

pmoravec: I see - the argument "multiple packages output" from elsewhere breaks my idea, then. Sad to see the improvement wont work (and ACK to the whole PR itself, then).

pmoravec (approved):

PR #3012, opened 2022-08-22
[policies] Add Anolis OS policy

UOS-liang:
Anolis OS is an enterprise Linux distribution based on RHEL. This adds
a Anolis OS policy, which inherits the RedHatPolicy base and uses the
RedHatPlugin tagging class.

### DIFF
--- a/sos/policies/distros/anolis.py
+++ b/sos/policies/distros/anolis.py
@@ -1,0 +1,46 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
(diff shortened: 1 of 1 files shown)


TurboTurtle (commented):
The base policy looks fine, however please see the contribution guidelines on the wiki. Commit messages should be prefix with a tag like `[anolis]`, e.g. `[anolis] Add new policy` as this makes it easier to search through the git history later on.

While not strictly required, you may want to add an informative docstring to the class definition, as this would be used for `sos help` output. See `sos help policies.redhat.rhel` for a good example.

TurboTurtle:
What needs to change here is the actual _commit_ message, not the GH PR title.

TurboTurtle:
I'm not sure if got dropped in the edit, or if I just overlooked it earlier, but per our [contribution guidelines](https://github.com/sosreport/sos/wiki/Contribution-Guidelines), all commits must have a DCO line on them. Once that's in place, this looks fine to me.

TurboTurtle (approved):

PR #3014, opened 2022-08-24
[policy] add support for logs, kdump, release in CosPlugin

vteratipally:
Add support for logs, kdump, release in CosPlugin

Signed-off-by: Varsha Teratipally <teratipally@google.com>

---


sos/report/plugins/kdump.py
```diff
@@ -102,4 +103,25 @@ def setup(self):
...
         ])
 
+
+class CosKDump(KDump, CosPlugin):
+
+    option_list = [
+        ("all_dumps", "enable capture for all kernel dumps", "", False),
+        ("latest_dump", "enable capture for latest kernel crash dump",
+         "", False),
+    ]
```
TurboTurtle, line 113: Plugin options changed in 4.3. They are now distinct `PluginOpt` objects. See [the boot plugin](https://github.com/sosreport/sos/blob/main/sos/report/plugins/boot.py#L22) for a good example.

sos/report/plugins/kdump.py
```diff
@@ -102,4 +103,25 @@ def setup(self):
...
+        super(CosKDump, self).setup()
+        self.add_cmd_output('ls -alRh /var/kdump*')
+        if self.get_option("all_dumps"):
+            self.add_copy_spec([
+                "/var/kdump-*"
+            ])
+        if self.get_option("latest_dump"):
+            self.add_copy_spec([
+                "/var/kdump"
+            ])
```
TurboTurtle, line 125: Collections like this should leverage the `--all-logs` global option instead of using plugin-specific options. E.G.

```python
    self.add_copy_spec('/var/kdump-*')
else:
    self.add_copy_spec('/var/kdump')
```

The above would result in always _trying_ to capture a kdump file, but may be prevented if that kdump file is larger than the sizelimit (25MB by default). The behavior of `add_copy_spec()` will ensure that the newest file is always taken first. If you want, you can also pass `sizelimit=XYZ` to the `add_copy_spec()` to increase the threshold for capturing a kdump file or not. Since this is limited to the CoS policy variant of the plugin, feel free to set this as your org needs.

sos/report/plugins/logs.py
```diff
@@ -105,4 +105,20 @@ def postproc(self):
             r"pwd=[******]"
         )
 
+class CosLogs(Logs, CosPlugin):
+    option_list = [
+        ("log_days", "the number of days logs to collect", "", 3)
+    ]
```
TurboTurtle, line 111: Same as above. Needs to be a `PluginOpt`.

sos/report/plugins/logs.py
```diff
@@ -105,4 +105,20 @@ def postproc(self):
...
+
+    def setup(self):
+        super(CosLogs, self).setup()
+        if self.get_option("all_logs"):
+            self.add_cmd_output("journalctl -o export")
+        else:
+            try:
+                days = int(self.get_option("log_days"))
+            except ValueError:
```
TurboTurtle, line 120: When using `PluginOpt`, this type checking should not be needed.

TurboTurtle (changes requested):
Comments below. The test suite is failing due to new form of defining plugin options.

sos/report/plugins/logs.py
vteratipally: updated the file

sos/report/plugins/kdump.py
vteratipally: Thank you for the suggestion but we would want to use the all_dumps to collect the kdump file rather than all_logs.

sos/report/plugins/kdump.py
```diff
@@ -102,4 +103,26 @@ def setup(self):
...
         ])
 
+
+class CosKDump(KDump, CosPlugin):
+
+    option_list = [
+        PluginOpt("all_dumps", "enable capture for all kernel dumps",
+                  "", False),
+        PluginOpt("latest_dump", "enable capture for latest kernel crash dump",
+                  "", False),
```
TurboTurtle, line 113: `PluginOpt` does not take the same order of parameters as the option tuples. Please re-look at the reference options in other plugins. `PluginOpt("opt-name", default=<default value>, desc="short description")` is what's needed at minimum.

sos/report/plugins/kdump.py
TurboTurtle: Plugin options are not meant to duplicate global options.

If you want this collection to be option-gated, then you can use a single option to do the gating and rely on the existing behavior of `--all-logs` and `add_copy_spec()`:

```python
PluginOpt('collect-kdumps', default=False, desc='Collect existing kdump files')
[...]
if self.get_option('collect-kdumps'):
    self.add_copy_spec('/var/kdump-*')
```

The above would capture capture files matching `/var/kdump-*` starting with the most recent kdump file, up to `sizelimit`. When `--all-logs` is used, this means _all_ kdump files will be collected.

Having two separate plugin options to dictate a single copy specification is not a user experience we want to encourage in plugins.

sos/report/plugins/logs.py
```diff
@@ -105,4 +105,18 @@ def postproc(self):
             r"pwd=[******]"
         )
 
+
+class CosLogs(Logs, CosPlugin):
+    option_list = [
+        PluginOpt("log_days", "the number of days logs to collect", "", 3)
+    ]
```
TurboTurtle, line 112: Same as above, the parameters for `PluginOpt` do not match the previous tuple format.

TurboTurtle (changes requested):

sos/report/plugins/logs.py
vteratipally: makes sense.

TurboTurtle (approved):
LGTM.

The centos test timed out, which appears unrelated, so I've restarted it.

pmoravec (approved):

Issue #2850, opened 2022-02-11
Add crictl plugin without crio data


samkunz:
Example Use Case: User wants to collect `crictl` data, but does not want to use [crio plugin](https://github.com/sosreport/sos/blob/main/sos/report/plugins/crio.py) to do so, because does not want to clutter results with "empty" crio related information (e.g. `add_copy_spec(["/etc/crio/crio.conf"])`, e.g. `add_cmd_output(["crio config"])`) and/or does not want to introduce risk of crio plugin failing if crio is not installed on system. 

Google COS has already implemented a workaround by patching in [a separate crictl plugin](https://cos.googlesource.com/cos/overlays/board-overlays/+/master/project-lakitu/app-admin/sosreport/files/0011-add-crictl-plugin.patch). 

I don't mind opening a pull request to get this functionality into upstream sosreport, but I wanted to check in with the maintainers about the preferred way to do this, since I'm not very familiar with the sosreport design philosophies.

**One way to accomplish this** is to transition the existing crio plugin into crio, crictl, containerd plugins. E.g. `add_cmd_output(["crio config"])` goes into crio plugin, `add_cmd_output(["crictl info"])` goes into crictl plugin, `self.add_journal(units='containerd')` goes into containerd plugin.
-> Pro of this approach: minimize duplicate information collected by multiple plugins.
-> Con of this approach: existing users of crio plugin would also need to make a change to use crio and crictl plugins to collect the same info.
-> Note that this approach doesn't require creation of a containerd plugin -- but to me it seemed like one way to reduce duplication of data collection (e.g. crio plugin and crictl plugin users both might want data on containerd and may not want to commit to crictl and/or crio plugins to get it.)

**Another way to accomplish this** is to simply add a crictl plugin, similar to the patch linked above, without modifying the crio plugin. 
-> Pro of this approach: introduces crictl specific plugin to satisfy use case while not changing crio plugin behavior. 
-> Con of this approach: duplicate data collected between crio plugin and crictl plugin.

Please advise on the preferred approach, thank you!

TurboTurtle:
First, let me preface all of this with saying we'd welcome and appreciate any contributions from Google/COS coming back into upstream. And looking at the other patches on that branch there are at least a couple that are very good candidates for upstream contribution (such as the COS package manager).


I think there may be some misunderstanding of how sos plugins work, or I may be reading this issue incorrectly:

> because does not want to clutter results with "empty" crio related information

[...]

PR #3016, opened 2022-08-26
[crio] add crio and containerd plugin support for Cos

vteratipally:
Add crio support for COS

Signed-off-by: Varsha Teratipally <teratipally@google.com>

---

### DIFF
--- a/sos/report/plugins/containerd.py
+++ b/sos/report/plugins/containerd.py
@@ -1,0 +1,29 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import (Plugin, RedHatPlugin, UbuntuPlugin, CosPlugin)
+
+
+class Containerd(Plugin, RedHatPlugin, UbuntuPlugin, CosPlugin):
+
+    short_desc = 'Containerd containers'
+    plugin_name = 'containerd'
+    profiles = ('container',)
+    packages = ('containerd',)
+
+    def setup(self):
+        self.add_copy_spec([
+            "/etc/containerd/",
+        ])
+
+        self.add_cmd_output('containerd config dump')
+
+        # collect the containerd logs.
+        self.add_journal(units='containerd')
+
+# vim: set et ts=4 sw=4 :
--- a/sos/report/plugins/crio.py
+++ b/sos/report/plugins/crio.py
@@ -9,10 +9,10 @@
 # See the LICENSE file in the source distribution for further information.
 
 from sos.report.plugins import (Plugin, RedHatPlugin, UbuntuPlugin,
-                                SoSPredicate, PluginOpt)
+                                SoSPredicate, PluginOpt, CosPlugin)
 
 
-class CRIO(Plugin, RedHatPlugin, UbuntuPlugin):
+class CRIO(Plugin, RedHatPlugin, UbuntuPlugin, CosPlugin):
 
     short_desc = 'CRI-O containers'
     plugin_name = 'crio'
--- a/sos/report/plugins/docker.py
+++ b/sos/report/plugins/docker.py
@@ -135,7 +135,6 @@ class UbuntuDocker(Docker, UbuntuPlugin):
     def setup(self):
         super(UbuntuDocker, self).setup()
         self.add_copy_spec([
-            "/etc/containerd/",
             "/etc/default/docker",
             "/run/docker/libcontainerd/containerd/events.log"
         ])


sos/report/plugins/crio.py
```diff
@@ -104,6 +78,44 @@ def _get_crio_list(self, cmd):
...
+        ])
+
+        self.add_cmd_output([
+            "ls -alhR /etc/cni",
+            "crio config"
+        ])
+
+        # base cri-o installation does not require cri-tools, which is what
+        # supplies the crictl utility
+        self.set_cmd_predicate(SoSPredicate(self, packages=['cri-tools']))
```
TurboTurtle, line 113: This would need to be called before the `super().setup()`, otherwise the `crictl` commands this predicate is intended to gate would have already been added to the collection phase.

sos/report/plugins/crio.py
```diff
@@ -113,4 +125,13 @@ def _get_crio_goroutine_stacks(self):
         if result['status'] == 0:
             self.add_copy_spec("/tmp/crio-goroutine-stacks*.log")
 
+
+class CosCRIO(Plugin, CosPlugin):
+
+    def setup(self):
+        super(CosCRIO, self).setup()
+
+        self.add_journal(units='containerd')
```
TurboTurtle, line 132: Are there other `containerd` related collections that'd be useful for COS? The `docker` plugin is what captures `/etc/containerd`, so I'm thinking it may be best to instead create a new `containerd` plugin at this point and put both of these collections there.

TurboTurtle (commented):
Couple notes below.

If we split out a separate `containerd` plugin, then we can still add COS support to the `crio` plugin, just by adding the tagging class. At that point we wouldn't need to split the plugin out into separate distro-specific classes.

If the files in the `add_copy_spec()` call don't exist on COS, then sos will handle that gracefully - same with the env vars. If they don't exist, sos doesn't raise an error or anything and the end result is there's simply nothing in the final archive for it.

sos/report/plugins/crio.py
vteratipally: That would be great but as of now, we would just need the `/etc/containerd`, container pods as well as the containerd service logs.

TurboTurtle (commented):
The latest force push doesn't seem to have included changes, at least none that stick out to me.

I am still of the opinion that if the only thing COS needs to add to this plugin is the containerd journal, then the better approach is to add the `CosPlugin` tag to the existing crio plugin, and instead create a new `containerd` plugin for the journal collection, as well as pulling the existing containerd collections out of the `docker` plugin and  into this new one.

As I mentioned, if the provided paths or env vars in their relevant calls don't exist or aren't set on COS, then there's no issue in keeping them in the "stock" plugin - they will simply be skipped on COS systems.

If the cmd predicate is not relevant to COS, then we can address that with a policy check. Alternatively, and I'm thinking preferably, the COS policy should instead be extended here to include some variation of the downstream changes COS is carrying, as highlighted in #2850, so that the predicate is applicable/valid for COS as well.

vteratipally:
> The latest force push doesn't seem to have included changes, at least none that stick out to me.
> 
> I am still of the opinion that if the only thing COS needs to add to this plugin is the containerd journal, then the better approach is to add the `CosPlugin` tag to the existing crio plugin, and instead create a new `containerd` plugin for the journal collection, as well as pulling the existing containerd collections out of the `docker` plugin and into this new one.
> 
> As I mentioned, if the provided paths or env vars in their relevant calls don't exist or aren't set on COS, then there's no issue in keeping them in the "stock" plugin - they will simply be skipped on COS systems.
> 
> If the cmd predicate is not relevant to COS, then we can address that with a policy check. Alternatively, and I'm thinking preferably, the COS policy should instead be extended here to include some variation of the downstream changes COS is carrying, as highlighted in #2850, so that the predicate is applicable/valid for COS as well.

Thank you for the feedback. It made sense, I created a new plugin containerd and added support for COS plugin in crio.

sos/report/plugins/containerd.py
```diff
@@ -0,0 +1,33 @@
...
+                  desc='collect stdout/stderr logs for containerd')
+    ]
+
+    def setup(self):
+        self.add_copy_spec([
+            "/etc/containerd/",
+        ])
+
+        if self.get_option('logs'):
+            self.add_journal(units='containerd')
```
TurboTurtle, line 30: Let's just make this a default collection, no need for a plugin option here.

sos/report/plugins/containerd.py
```diff
@@ -0,0 +1,33 @@
...
+from sos.report.plugins import (Plugin, PluginOpt, RedHatPlugin,
+                                UbuntuPlugin, CosPlugin)
+
+
+class Containerd(Plugin, RedHatPlugin, UbuntuPlugin, CosPlugin):
+
+    short_desc = 'Containerd containers'
+    plugin_name = 'containerd'
+    profiles = ('container',)
+
```
TurboTurtle, line 18: There should be some enablement trigger here. Either a `packages`, `files`, or `commands` tuple will suffice. If the package name differs between distros, that's still fine to have a single plugin - just put all the possible package names in the tuple and sos will figure it out.

sos/report/plugins/containerd.py
```diff
@@ -0,0 +1,33 @@
...
+    option_list = [
+        PluginOpt('logs', default=False,
+                  desc='collect stdout/stderr logs for containerd')
+    ]
+
+    def setup(self):
```
TurboTurtle, line 24: Would collecting `containerd config dump` output be helpful here alongside the directory?

sos/report/plugins/crio.py
```diff
@@ -9,10 +9,10 @@
 # See the LICENSE file in the source distribution for further information.
 
 from sos.report.plugins import (Plugin, RedHatPlugin, UbuntuPlugin,
-                                SoSPredicate, PluginOpt)
+                                SoSPredicate, PluginOpt, CosPlugin)
 
 
-class CRIO(Plugin, RedHatPlugin, UbuntuPlugin):
+class CRIO(Plugin, RedHatPlugin, UbuntuPlugin, CosPlugin):
```
TurboTurtle, line 15: Ack.

pmoravec (approved):

TurboTurtle (approved):

PR #3019, opened 2022-08-31
[cleaner] Check --domains values for validity

TurboTurtle:
The domain parser is built to understand actual names, e.g.
'example.com', and not operate off of just 'example' despite the fact
that the parser handles TLDs separately.

To safeguard against potential errors when trying to parse the latter
example above as an actual domain, validate any values passed to
`--domains` during cleaner initialization.

---


pmoravec (approved):
Good idea to push users to use `--keywords` instead mis-using `--domains`.

PR #3022, opened 2022-09-07
[cleaner] Prevent stripping '_' on no match in hostname parser

pmoravec:
Hostname parser works with stripped leading/trailing '.' and '_' characters for a match. When no pattern match is found, we must return the original unstripped item.

Resolves: #3022

---

### DIFF
--- a/sos/cleaner/mappings/hostname_map.py
+++ b/sos/cleaner/mappings/hostname_map.py
@@ -147,7 +147,9 @@ class SoSHostnameMap(SoSMap):
         if item in self.dataset:
             return self.dataset[item]
         if not self.domain_name_in_loaded_domains(item.lower()):
-            return item
+            # no match => return the original string with optional
+            # leading/trailing '.' or '_' characters
+            return ''.join([prefix, item, suffix])
         if item.endswith(self.strip_exts):
             ext = '.' + item.split('.')[-1]
             item = item.replace(ext, '')
--- a/tests/unittests/cleaner_tests.py
+++ b/tests/unittests/cleaner_tests.py
@@ -102,7 +102,8 @@ class CleanerParserTests(unittest.TestCase):
     def setUp(self):
         self.ip_parser = SoSIPParser(config={})
         self.mac_parser = SoSMacParser(config={})
-        self.host_parser = SoSHostnameParser(config={}, opt_domains='foobar.com')
+        self.host_parser = SoSHostnameParser(config={},
+                                             opt_domains=['foobar.com'])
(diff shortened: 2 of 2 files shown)


pmoravec:
Reproducer:

    # sos report -o pam --batch --build --clean --disable-parsers=ip,keyword,mac,username
    ..
    # diff /etc/pam.d/su /var/tmp/sosreport-pmoravec-rhel8-2022-09-07-fzfervg/etc/pam.d/su
    2,3c2,3
    < auth		required	pam_env.so
    < auth		sufficient	pam_rootok.so
    ---
    > auth		required	pamenv.so
    > auth		sufficient	pamrootok.so
    ..

pmoravec:
Hmm, does it deserve a dedicated avocado test?

TurboTurtle:
> Hmm, does it deserve a dedicated avocado test?

I'd say yes. Since we have a simple reproducer via the `pam` plugin, we can leverage that for it.

pmoravec:
I wanted to add to `tests/unittests/cleaner_tests.py` either:

    class CleanerMapTests(unittest.TestCase):
        ..
        def test_hostname_no_obfuscate_underscore(self):
            line = 'pam_env.so _why.not_'
            _test = self.host_map.get(line)[0]
            self.assertEqual(line, _test)

or:

    class CleanerParserTests(unittest.TestCase):

        def setUp(self):
            self.ip_parser = SoSIPParser(config={})
            self.mac_parser = SoSMacParser(config={})
            self.host_parser = SoSHostnameParser(config={}, opt_domains=['foobar.com'])    ### a bug fixed: added []

        ..

        def test_hostname_no_obfuscate_underscore(self):
            line = 'pam_env.so _why.not_'
            _test = self.host_parser.parse_line(line)[0]
            self.assertEqual(line, _test)

but in either case, `assert` fails as `_test` has removed the problematic underscores.

When I run the test as a python script:

    from sos.cleaner.parsers.hostname_parser import SoSHostnameParser
    host_parser = SoSHostnameParser(config={}, opt_domains=['foobar.com'])
    line = 'pam_env.so _why.not_'
    print(host_parser.parse_line(line))

it prints correct output:

    # python3 test_underscore.py
    ('pam_env.so _why.not_', 2)
    #

What did I wrong..?

TurboTurtle:
I'm not sure why it's failing for you - if I pull your branch and add the second test you provide, it passes for me locally.

pmoravec:
Really, all tasks passed. It had to be some local problem on my system, then.

TurboTurtle (approved):

PR #3023, opened 2022-09-08
[snap] Update base to core22

arif-ali:
core20 has python3-magic 0.4.15, and hence the binary detection will not work by default. Moving to core22 goes to 0.4.24-2 which would mean this would work nevertheless

---

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -117,13 +117,13 @@ snap_build_task:
     name: "snap Build From Checkout"
     gce_instance:
         image_project: "${UBUNTU_PROJECT}"
-        image_name: "${UBUNTU_PRIOR_IMAGE_NAME}"
+        image_name: "${UBUNTU_IMAGE_NAME}"
         type: e2-medium
     setup_script: |
         apt update
         apt -y install snapd
         systemctl start snapd
-        sed -i 's/adopt-info.*/version: test/g' snap/snapcraft.yaml
+        sed -i -e 's/adopt-info.*/version: test/g' -e '/set version/d' snap/snapcraft.yaml
         snap install snapcraft --classic
     main_script: |
         snapcraft --destructive-mode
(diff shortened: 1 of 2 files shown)


arif-ali:
hmm, that's interesting, worked in my lab ...

I'll have a another go, and see what I can find on the issue

TurboTurtle:
For the Foreman test, do we need to switch Debian 10 to not use snaps? We can probably safely use the git checkout on that test if core22 is incompatible with Debian 10.

arif-ali:
> For the Foreman test, do we need to switch Debian 10 to not use snaps? We can probably safely use the git checkout on that test if core22 is incompatible with Debian 10.

Yeah, I think Debian will not have the latest snapd available to test with `core22`. If it is easier to shift to git checkout for Debian, that might be the easier option.

Ideally, it would be good to do everything in `core22`, based on my findings below

base | size
----|---
core20 (magic==0.4.15) | 9meg
core20+pip install magic 0.4.20 | 26meg
core22 | 6.9meg

arif-ali:
Reading https://www.debian.org/releases/, version 10 is oldstable, and 11 is stable. Would it make sense to now test on 11 instead of 10?

TurboTurtle:
The debian version for that test is primarily determined due to Foreman support.

@pmoravec @ekohl - is Foreman still running on Debian 10, or can the image be updated to Debian 11?

ekohl:
> is Foreman still running on Debian 10, or can the image be updated to Debian 11?

Foreman 3.2 introduced support for Debian 11 and Foreman 3.4 (just released) has dropped support for Debian 10. So :+1: to upgrading.

TurboTurtle:
@arif-ali - I just merged #3032 which should resolve the Foreman test issues with Debian 10. Try rebasing to current `main`.

arif-ali:
@TurboTurtle, thanks, should be rebased and looks like it has all gone through now

TurboTurtle (approved):

pmoravec (approved):

Issue #3021, opened 2022-09-06
Use older filetype detection for linux distributions which do not have python3-magic version 0.4.20


nkshirsagar:
On sos 4.3, filetype detection does not use magic.detect_from_filename() 
~~~
    def file_is_binary(self, fname):
        """Determine if the file is a binary file or not.


        :param fname:          Filename relative to the extracted archive root
        :type fname:           ``str``

        :returns:   ``True`` if file is binary, else ``False``
        :rtype:     ``bool``
        """
        with open(self.get_file_path(fname), 'tr') as tfile:
            try:
                # when opened as above (tr), reading binary content will raise
                # an exception
                tfile.read(1)
                return False
            except UnicodeDecodeError:
                return True
~~~
but it does on 4.4, 
~~~
def file_is_binary(fname):
    """Helper to determine if a given file contains binary content or not.

    This is especially helpful for `sos clean`, which cannot obfuscate binary
    data and instead, by default, will remove binary files.

    :param fname:   The full path of the file to check binaryness of
    :type fname:    ``str``

    :returns:   True if binary, else False
    :rtype:     ``bool``
    """
    try:
        _ftup = magic.detect_from_filename(fname)
~~~
( due to https://github.com/sosreport/sos/commit/0689d8b3e10e969b7fed0af9ac906222858e4fb9 ) 

Older linux distributions may not be having python3-magic version 0.4.20 which provides the detect_from_filename() method (introduced in 0.4.19 , thanks @arif-ali !)

So sos should detect if an older python3-magic version than 0.4.20 is installed and if so, use the older way to detect if a file is binary.

TurboTurtle:
Copying from IRC -

python3-magic has the current dependency on 0.4.20 due to the use of a properly functioning `detect_from_filename()` method. 0.4.19 introduced this method and at the same time replacing the older methods. Meaning, moving forward our only option via this library is this function - the previous approaches are not even no longer maintained but are completely removed.

I'd first suggest seeing if 0.4.20 (or newer) can be pushed as an upgrade in the older distro releases. If not, then we could certainly accept a PR that implements fallback logic for the older versions of the magic module.

nkshirsagar:
We have decided to patch downstream old distributions for Ubuntu like so, (basically reverting to older approach for binary file detection)

`--- sosreport-4.4.orig/requirements.txt
+++ sosreport-4.4/requirements.txt
@@ -2,5 +2,4 @@ pycodestyle>=2.4.0
[...]

PR #3025, opened 2022-09-12
[utilities] Relax from hard dependency of python3-magic

pmoravec:
For compatibility reasons on some distros, sos should not have a hard dependency on 'magic' python library. It should attempt to use it for detection of binary file content, but should fall back to previous "read the very first byte" method otherwise.

Resolves: #3025
Relates: #3021

---

### DIFF
--- a/sos/utilities.py
+++ b/sos/utilities.py
@@ -19,10 +19,25 @@ import tempfile
 import threading
 import time
 import io
-import magic
-
 from contextlib import closing
 from collections import deque
+
+# try loading magic>=0.4.20 which implements detect_from_filename method
+magic_mod = False
+try:
+    import magic
+    magic.detect_from_filename(__file__)
+    magic_mod = True
+except (ImportError, AttributeError):
+    log = logging.getLogger('sos')
+    from textwrap import fill
+    msg = ("""\
+WARNING: Failed to load 'magic' module version >= 0.4.20 which sos aims to \
+use for detecting binary files. A less effective method will be used. It is \
+recommended to install proper python3-magic package with the module.
+""")
+    log.warn('\n' + fill(msg, 72, replace_whitespace=False) + '\n')
+
 
 TIMEOUT_DEFAULT = 300
 
@@ -75,17 +90,26 @@ def file_is_binary(fname):
     :returns:   True if binary, else False
     :rtype:     ``bool``
     """
-    try:
-        _ftup = magic.detect_from_filename(fname)
-        _mimes = ['text/', 'inode/']
-        return (
-            _ftup.encoding == 'binary' and not
-            any(_ftup.mime_type.startswith(_mt) for _mt in _mimes)
-        )
-    except Exception:
-        # if for some reason this check fails, don't blindly remove all files
-        # but instead rely on other checks done by the component
-        return False
+    if magic_mod:
+        try:
+            _ftup = magic.detect_from_filename(fname)
+            _mimes = ['text/', 'inode/']
+            return (
+                _ftup.encoding == 'binary' and not
+                any(_ftup.mime_type.startswith(_mt) for _mt in _mimes)
+            )
+        except Exception:
+            pass
+    # if for some reason the above check fails or magic>=0.4.20 is not present,
+    # fail over to checking the very first byte of the file content
+    with open(fname, 'tr') as tfile:
+        try:
+            # when opened as above (tr), reading binary content will raise
+            # an exception
+            tfile.read(1)
+            return False
+        except UnicodeDecodeError:
+            return True
 
 
 def find(file_pattern, top_dir, max_depth=None, path_pattern=None):
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,5 +2,4 @@ pycodestyle>=2.4.0
 coverage>=4.0.3
 Sphinx>=1.3.5
 pexpect>=4.0.0
-python_magic>=0.4.20
 pyyaml
--- a/setup.py
+++ b/setup.py
@@ -107,7 +107,7 @@ setup(
     ],
     cmdclass=cmdclass,
     command_options=command_options,
-    requires=['pexpect', 'python_magic', 'pyyaml']
+    requires=['pexpect', 'pyyaml']
     )
 
 
--- a/sos.spec
+++ b/sos.spec
@@ -16,7 +16,7 @@ Requires: python3-rpm
 Requires: tar
 Requires: xz
 Requires: python3-pexpect
-Requires: python3-magic
+Recommends: python3-magic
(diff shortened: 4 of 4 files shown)


pmoravec:
Cc @nkshirsagar - please let me know if this PR would resolve the #3021 (in upstream).

pmoravec:
> Having the failover method makes sense, but is there a way we can make `python3-magic` a similarly weak dep for `setup.py`?

`distutils.core` can define just hard dependencies (`'requires', 'provides', 'obsoletes', 'depends'`) but not a weak one :(

> Since the current approach depends on a method that only exists in 0.4.19, should we leave the dependency in there, but simply fallback to the "read a byte" approach when an older version is installed?

You mean "try importing magic library, if succeeds and version is >=0.4.20, always use the method from magic; otherwise always fallback to the 'read a byte' approach"? That makes sense, and keep that in a variable to test inside `file_is_binary` method.

TurboTurtle:
> You mean "try importing magic library, if succeeds and version is >=0.4.20, always use the method from magic; otherwise always fallback to the 'read a byte' approach"? That makes sense, and keep that in a variable to test inside `file_is_binary` method.

Yes, but also something that provides for "Hey, $user, the sos team _really_ suggests having python3-magic installed here, and it should be installed alongside sos if at all possible, but just in case that can't be done, you're still covered." Effectively, try to ensure that python3-magic gets installed (so, I guess, keep it as a dep), but provide failover in case that's not possible but we still want to support generating an archive.

nkshirsagar:
> Cc @nkshirsagar - please let me know if this PR would resolve the #3021 (in upstream).
yes it would. Thanks!

We have patched downstream with a very similar patch for bionic/focal Ubuntu versions, pasted below,

~~~
Description: Revert to old way of binary file detection
 Reverting to older (4.3) way of detecting binary files 
 because sos 4.4 uses python3-magic version 0.4.20 which is 
 not present in bionic.
 .
 sosreport (4.4-1ubuntu0.18.04.1) bionic; urgency=medium
 .
   * New 4.4 upstream. (LP: #1986611)
 .
   * For more details, full release note is available here:
     - https://github.com/sosreport/sos/releases/tag/4.4
 .
   * Former patches, now fixed:
     - d/p/0002-fix-setup-py.patch
     - d/p/0003-mention-sos-help-in-sos-manpage.patch
 .
   * Remaining patches:
     - d/p/0001-debian-change-tmp-dir-location.patch
 .
   * New patches:
     - d/p/0002-revert-to-old-style-binary-file-detection.patch
Author: Nikhil Kshirsagar <nikhil.kshirsagar@canonical.com>
Bug-Ubuntu: https://bugs.launchpad.net/bugs/1986611

--- sosreport-4.4.orig/requirements.txt
+++ sosreport-4.4/requirements.txt
@@ -2,5 +2,4 @@ pycodestyle>=2.4.0
 coverage>=4.0.3
 Sphinx>=1.3.5
 pexpect>=4.0.0
-python_magic>=0.4.20
 pyyaml
--- sosreport-4.4.orig/setup.py
+++ sosreport-4.4/setup.py
@@ -107,7 +107,7 @@ setup(
     ],
     cmdclass=cmdclass,
     command_options=command_options,
-    requires=['pexpect', 'python_magic', 'pyyaml']
+    requires=['pexpect', 'pyyaml']
     )
 
 
--- sosreport-4.4.orig/sos/utilities.py
+++ sosreport-4.4/sos/utilities.py
@@ -19,7 +19,6 @@ import tempfile
 import threading
 import time
 import io
-import magic
 
 from contextlib import closing
 from collections import deque
@@ -75,17 +74,14 @@ def file_is_binary(fname):
     :returns:   True if binary, else False
     :rtype:     ``bool``
     """
-    try:
-        _ftup = magic.detect_from_filename(fname)
-        _mimes = ['text/', 'inode/']
-        return (
-            _ftup.encoding == 'binary' and not
-            any(_ftup.mime_type.startswith(_mt) for _mt in _mimes)
-        )
-    except Exception:
-        # if for some reason this check fails, don't blindly remove all files
-        # but instead rely on other checks done by the component
-        return False
+    with open(fname, 'tr') as tfile:
+        try:
+            # when opened as above (tr), reading binary content will raise
+            # an exception
+            tfile.read(1)
+            return False
+        except UnicodeDecodeError:
+            return True
 
 
 def find(file_pattern, top_dir, max_depth=None, path_pattern=None):
~~~

So we are going ahead with our 4.4 releases with that for now, and for 4.5 we will revert the patch so as to use the upstream source.

pmoravec:
Good point with the warning, I added it.

Yes the downstream Ubuntu patch was an inspiration :) - my question was rather to ensure the upstream solution will properly work on Ubuntu without a need of a downstream patch.

nkshirsagar:
> Good point with the warning, I added it.
> 
> Yes the downstream Ubuntu patch was an inspiration :) - my question was rather to ensure the upstream solution will properly work on Ubuntu without a need of a downstream patch.

The warning "It is recommended to install proper python3-magic package with the module." is moot on Ubuntu bionic/focal versions because we know they will not have the required python3-magic package. The upstream patch caters to both situations but the debian patch simply assumes no such python3-magic version will be available, so no warning is needed, and we simply revert to older approach for those distros. Thinking ahead, for 4.5 we could potentially revert the debian patch, but we'd end up spewing a warning.

pmoravec:
Feel free to propose any better warning text (in either better meaning or better English), I know I am not the best writer for either :).

nkshirsagar:
> Feel free to propose any better warning text (in either better meaning or better English), I know I am not the best writer for either :).

The warning is perfect for distributions where the upgrade of that package is actually an option :-)

sos/utilities.py
```diff
@@ -19,11 +19,26 @@
...
 from contextlib import closing
 from collections import deque
 
+# try loading magic>=0.4.20 which implements detect_from_filename method
+magic_mod = False
+try:
+    import magic
+    magic.detect_from_filename(__file__)
+    magic_mod = True
+except ImportError:
```
TurboTurtle, line 31: We'd also need to catch `AttributeError` for cases where older `magic` is installed and `detect_from_filename()` is not present.

sos/utilities.py
```diff
@@ -19,11 +19,26 @@
...
 from contextlib import closing
 from collections import deque
 
+# try loading magic>=0.4.20 which implements detect_from_filename method
+magic_mod = False
+try:
+    import magic
+    magic.detect_from_filename(__file__)
+    magic_mod = True
+except ImportError, AttributeError:
```
TurboTurtle, line 31: Syntax error is due to these being non-parenthesized.

TurboTurtle (approved):

PR #3026, opened 2022-09-13
fix: change setup.py requires to install_requires

assafgi:
---

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -107,7 +107,7 @@ setup(
     ],
     cmdclass=cmdclass,
     command_options=command_options,
-    requires=['pexpect', 'python_magic', 'pyyaml']
+    install_requires=['pexpect', 'python_magic', 'pyyaml']
     )
 
 


setup.py
```diff
@@ -107,7 +107,7 @@ def copy_file (self, filename, dirname):
     ],
     cmdclass=cmdclass,
     command_options=command_options,
-    requires=['pexpect', 'python_magic', 'pyyaml']
+    install_requires=['pexpect', 'python_magic', 'pyyaml']
```
TurboTurtle, line 110: AFAIK, this would be correct for `setuptools`, but sos to date still uses `distutils` where `requires` is the keyword to use. I believe `distutils` will be removed in `python-3.12`, but at the moment we still use it.

In order for this change to be complete, we'd need to update the entire `setup.py` to use `setuptools`. Which is a PR we'd be happy to accept, of course.

TurboTurtle (commented):
Thanks for the PR - please see my note below.

Additionally, please review our [Contribution Guidelines](https://github.com/sosreport/sos/wiki/Contribution-Guidelines). We require a DCO for all commits, and that commit message must be formatted in a clear `[tag] <short title>` format that includes a short description of the changes being made.

TurboTurtle:
Closing this as another commit has converted us over to setuptools entirely.

PR #3030, opened 2022-09-20
[clean,username_parser] Handle usernames with domain prefixes

TurboTurtle:
In some configurations, we can expect usernames to be written as `DOMAIN\user`, as well as the plain `user` form. Previously, the domain-prefixed format would throw an exception in our obfuscation if the letter after the `\` would cause the interpreter to regard that as a special character.

Fix this by escaping the values passed to `get_regex_result()` so that we can successfully compile a regex pattern object for domain-prefixed usernames as well.

Further, when we encounter one of these usernames, the username parser should automatically add an entry for the non-prefixed username as well. While this will result in a secondary obfuscation match (due to the current design of `sos clean`), this is preferable to potentially leaving non-prefixed usernames in plaintext, if they happen to appear in logs but not in any of our sourced files.

Related: RHBZ#2127977

---

### DIFF
--- a/sos/cleaner/mappings/__init__.py
+++ b/sos/cleaner/mappings/__init__.py
@@ -94,7 +94,7 @@ class SoSMap():
         :returns:       A compiled regex pattern for the item
         :rtype:         ``re.Pattern``
         """
-        return re.compile(item, re.I)
+        return re.compile(re.escape(item), re.I)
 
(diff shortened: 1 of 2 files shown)


pmoravec (commented):
Nice!

pmoravec (approved):

PR #3031, opened 2022-09-20
[dnf] Collect legacy yum config and dnf global vars

pmoravec:
- collect legacy yum->dnf symlinks
- collect dnf global vars
- obfuscate all passwords in all vars

---

### DIFF
--- a/sos/report/plugins/dnf.py
+++ b/sos/report/plugins/dnf.py
@@ -55,7 +55,12 @@ class DNFPlugin(Plugin, RedHatPlugin):
             '/etc/dnf/modules.d/.*.module': 'dnf_modules'
         })
 
-        self.add_copy_spec("/etc/dnf/")
+        self.add_copy_spec([
+            "/etc/dnf/",
+            "/etc/yum.conf",
+            "/etc/yum/pluginconf.d/",
+            "/etc/yum/vars/",
+        ])
         self.add_copy_spec("/etc/yum.repos.d/",
                            tags=['yum_repos_d', 'dnf_repos_d', 'dnf_repo'])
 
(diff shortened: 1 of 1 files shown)


sos/report/plugins/dnf.py
```diff
@@ -55,7 +55,13 @@ def setup(self):
             '/etc/dnf/modules.d/.*.module': 'dnf_modules'
         })
 
-        self.add_copy_spec("/etc/dnf/")
+        self.add_copy_spec([
+            "/etc/dnf/",
+            "/etc/dnf/vars/",
```
TurboTurtle, line 60: This should already be collected by the previous copyspec - unless we're hitting size limiting somewhere?

sos/report/plugins/dnf.py
```diff
@@ -120,8 +126,9 @@ def setup(self):
         self.get_modules_info(modules['output'])
 
     def postproc(self):
-        regexp = r"(proxy_password(\s)*=(\s)*)(\S+)\n"
+        regexp = r"(password(\s)*=(\s)*)(\S+)\n"
```
TurboTurtle, line 129: Does this need a leading `(.*)?` optional match in order to still hit `proxy_password` along with all others?

pmoravec: Until we have there `^` or something similar, it should work well - as I checked:

    # cat /etc/dnf/vars/test.conf 
    proxy_password = test
    password = different
    # python3 bin/sos report -o dnf --batch --build
    ..
    # cat /var/tmp/sosreport-pmoravec-rhel8-2022-09-21-vfswezo/etc/dnf/vars/test.conf 
    proxy_password = ********
    password = ********
    #

sos/report/plugins/dnf.py
pmoravec: Ah, I was blind.. thanks for pointing to the dup.

TurboTurtle (approved):

PR #3034, opened 2022-09-28
[testing] Fine tune the testing for ubuntu

arif-ali:
There was no testing for ubuntu 18.04, and previous manual testing found issues, and this is still supported. Add envs similar to fedora and centos testing. This change also ensures that ubuntu 22.04 is the one that is used for snap builds. We don't need python3-magic on the OS, as this should now come in-built within the snap, which will hav ehe right version for binary checks.

---


.cirrus.yml
```diff
@@ -8,8 +8,9 @@ env:
 
     DEBIAN_NAME: "debian-11"
 
-    UBUNTU_NAME: "ubuntu-22.04"
-    UBUNTU_PRIOR_NAME: "ubuntu-20.04"
+    UBUNTU_1804_NAME: "ubuntu-18.04"
+    UBUNTU_2004_NAME: "ubuntu-20.04"
+    UBUNTU_2204_NAME: "ubuntu-22.04"
```
TurboTurtle, line 13: Drop versioning from the field names. We use `DISTRO_NAME` and `DISTRO_PRIOR_NAME` as it makes ongoing maintenance easier when we need to update cirrus.yml. The reason we have 8/9 in the Centos fields is because it's literally baked into the Stream names, but if we need to update images at any point we don't need to rename the fields in multiple places. We just need to update the string in `*_NAME` and the relevant image in `*_IMAGE_NAME`.

For this, we could use something like `UBUNTU_*`, `UBUNTU_PRIOR_*`, and `UBUNTU_LTS_*`.

.cirrus.yml
```diff
@@ -143,21 +147,22 @@ report_stageone_task:
...
+            VM_IMAGE_NAME: ${UBUNTU_2004_IMAGE_NAME}
+        - env: *ubuntu2204
     setup_script: &setup |
         if [ $(command -v apt) ]; then
             echo "$ARTCURL/snap%20Build%20From%20Checkout/packages/sosreport_test_amd64.snap"
             $ARTCURL/snap%20Build%20From%20Checkout/packages/sosreport_test_amd64.snap
             apt -y purge sosreport
             apt update --allow-releaseinfo-change
-            apt -y install python3-pip snapd python3-magic
+            apt -y install python3-pip snapd
```
TurboTurtle, line 165: Ack

TurboTurtle (commented):
Comments below.

Generally speaking we keep testing matrices to the current and n-1 release, but the overhead of a single extra release here for stageone testing is very minor so it's fine.

My only concern is module version availability in 18.04. My initial thought is that 18.04 should be generally regarded the same as RHEL 7 and early RHEL 8 releases - any packaging or module differences need to be handled downstream instead of here. But, 18.04 looks to be at python 3.6, which should be common enough ground with our EL 8 testing, so I'm good to add it back on for stage one.

.cirrus.yml
arif-ali: ack, will think about how I can refactor, although all of these versions are LTS releases, with 5 years of support from initial release

.cirrus.yml
```diff
@@ -28,6 +29,8 @@ env:
     FEDORA_PRIOR_IMAGE_NAME: "fedora-cloud-base-gcp-35-1-2-x86-64"
     UBUNTU_IMAGE_NAME: "ubuntu-2204-jammy-v20220622"
     UBUNTU_PRIOR_IMAGE_NAME: "ubuntu-2004-focal-v20220701"
+    UBUNTU_PRIOR2_IMAGE_NAME: "ubuntu-1804-bionic-v20220901"
+    UBUNTU_SNAP_IMAGE_NAME: "ubuntu-2204-jammy-v20220622"
```
TurboTurtle, line 33: Is there a reason to not just use the 'regular' `UBUNTU_IMAGE_*` fields for the snap build? If that image can/should be updated for either/both the snap build and the actual tests, we can do that just as easily. For example, I've started reviewing the images for Fedora/Centos every 3 months (speaking of which...it's about that time for me) where we'll pick up the newer images for those releases.

arif-ali: I could, but the SNAP is based on `core22`, i.e. 22.04, and will stay that way until there are dependency issues. I could change to UBUNTU_IMAGE_NAME, but will change in 2024, when the next LTS release comes, and hence the SNAP build would change to `UBUNTU_PRIOR_IMAGE_NAME`. So, was looking to future proof that, so that it will stay consistent.

Now, when the we decide to update the `snapcraft.yaml` to use `core24`, which is unlikely (unless there is a major change for some package to be required, such as magic this time round), that variable won't change for a while.

I hope that makes sense

TurboTurtle: Ah, ok - yep, that tracks.

pmoravec (approved):

arif-ali:
Any preferences in using `UBUNTU_PRIOR2_NAME` or `UBUNTU_OLD_NAME`? based on conversation on IRC :)

PR #3039, opened 2022-10-06
[mac_parser, plugins] Refine parser skipping, add kmods to plugins

TurboTurtle:
This patchset first refines the mac parser's file skipping to skip all `modinfo` commands, instead of just the one from the kernel plugin.

Second, it updates several plugins to define modules in their respective `kernel_mods` enablement trigger, which will also cause separate `modinfo` collections within those plugins per #3037 

---

### DIFF
--- a/sos/cleaner/parsers/mac_parser.py
+++ b/sos/cleaner/parsers/mac_parser.py
@@ -44,7 +44,7 @@ class SoSMacParser(SoSCleanerParser):
         '534f:53'
     )
     skip_files = [
-        'sos_commands/kernel/modinfo.*'
+        'sos_commands/.*/modinfo.*'
     ]
     map_file_key = 'mac_map'
     compile_regexes = False
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -3305,7 +3305,8 @@ class Plugin():
                 self.add_service_status(service)
                 self.add_journal(service)
         for kmod in self.kernel_mods:
-            self.add_cmd_output(f"modinfo {kmod}")
+            if self.is_module_loaded(kmod):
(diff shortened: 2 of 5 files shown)


TurboTurtle:
Small update to refine the `modinfo` collection from earlier, to only do so if the module is loaded.

sos/cleaner/parsers/mac_parser.py
```diff
@@ -44,7 +44,7 @@ class SoSMacParser(SoSCleanerParser):
         '534f:53'
     )
     skip_files = [
-        'sos_commands/kernel/modinfo.*'
+        'sos_commands/.*modinfo.*'
```
pmoravec, line 47: Being bit paranoid, if a plugin would collect say `blah-modinfo.txt` file, we would skip cleaning it. Could be there `'sos_commands/.*/modinfo.*'` instead?

sos/report/plugins/__init__.py
```diff
@@ -3305,7 +3305,8 @@ def add_default_collections(self):
                 self.add_service_status(service)
                 self.add_journal(service)
         for kmod in self.kernel_mods:
-            self.add_cmd_output(f"modinfo {kmod}")
+            if self.is_module_loaded(kmod):
+                self.add_cmd_output(f"modinfo {kmod}")
```
pmoravec, line 3309: +1

pmoravec (approved):
ACK up to the tiny improvement.

sos/cleaner/parsers/mac_parser.py
TurboTurtle: Fair point, I'll push an update.

pmoravec (approved):

PR #3040, opened 2022-10-06
[logs] grab journalctl logs if not --all-logs

arif-ali:
If we are using --all-logs, we are collecting /var/log/journal, so we don't need to collect the journalctl output of the logs, when these can easily be grabbed from the files in /var/log/journal

---

### DIFF
--- a/sos/report/plugins/logs.py
+++ b/sos/report/plugins/logs.py
@@ -63,12 +63,14 @@ class Logs(Plugin, IndependentPlugin):
         journal = any([self.path_exists(self.path_join(p, "log/journal/"))
                        for p in ["/var", "/run"]])
         if journal and self.is_service("systemd-journald"):
-            self.add_journal(since=since, tags='journal_full', priority=100)
-            self.add_journal(boot="this", since=since,
-                             tags='journal_since_boot')
-            self.add_journal(boot="last", since=since,
-                             tags='journal_last_boot')
-            if self.get_option("all_logs"):
+            if not self.get_option("all_logs"):
+                self.add_journal(since=since, tags='journal_full',
+                                 priority=100)
+                self.add_journal(boot="this", since=since,
+                                 tags='journal_since_boot')
+                self.add_journal(boot="last", since=since,
+                                 tags='journal_last_boot')
+            else:
                 self.add_copy_spec([
                     "/var/log/journal/*",
                     "/run/log/journal/*"


TurboTurtle:
I think most end users would still have an expectation of the typical journal outputs in `sos_commands/logs/` regardless of size limiting.

@pmoravec @jjansky1 - thoughts?

jcastill:
> I think most end users would still have an expectation of the typical journal outputs in `sos_commands/logs/` regardless of size limiting.


Not only that, but also we need to bear in mind if journalctl in the machine opening the sosreport will be able to read the files inside /var/log/journal. Is guaranteed that it will work across versions? I.e. I tried to open a fedora 36 journal file from that dir in a RHEL 8 server, and I got:

`# journalctl --file user-1002@e56992149d674f4080d30ed83d521667-00000000007c0633-0005e1b54cce5adc.journal
Failed to open files: Protocol not supported`

Now, I'm aware this may be a extreme case - we'll normally be going the other way around perhaps, opening the files with newer versions of journalctl. But I think we should avoid the situation we have with sar, where you need a specific version to decode sa files, and have journalctl decode the logs in the server where sos is run.

pmoravec:
I am not *strongly* against the idea / PR. I understand we can save a lot of time by implementing it. I do agree with "legacy reasons" that `sos_commands/logs/journalctl*` files are expected to be present every time, but that is just a matter of generating them locally from `/[var|run]/logs/journal/`. Until some tooling system (Insights?) does rely on presence of the files (I am checking this internally).

`journalctl` seems to be backward compatible in reading old journal logs:, at least I can run on F36 `journalctl --file ..` for a journal file from F28,. *Usually* (but not always!) we investigate sosreports from "older systems" on a "newer system". If that does not happen, we have a problem like RHEL8 can't decode journalctl logs from RHEL9.

Also we can't be guaranteed with the backward compatibility. And from my personal experience with before mentioned `sar`, the burden "I need to copy the binary file to an appropriate system to decode it" is something I would like to avoid - esp. for something like system logs.

lathiat:
> I am not _strongly_ against the idea / PR. I understand we can save a lot of time by implementing it. I do agree with "legacy reasons" that `sos_commands/logs/journalctl*` files are expected to be present every time, but that is just a matter of generating them locally from `/[var|run]/logs/journal/`. Until some tooling system (Insights?) does rely on presence of the files (I am checking this internally).

I also agree with the premise. We're basically doubling up on the logs stored (and in the case of Ubuntu it's even worse as the journal still gets written into separate log files in /var/log as well, so now we are tripling the usage). Plus the processing time.

However I also agree that randomly having this file there or not there depending on --all-logs is going to be painful. If we make the change I'd like to see us always having a consistent dataset regardless of log filtering or --all-logs options. 

To that end I had a look to see if there were any guarantees of backwards/forwards compatability and found that journals support "Journal Export Format" into a binary-mostly-text or JSON format. Perhaps we can consider standardising on such an export and then excluding /var/log/journal from --all-logs.

https://systemd.io/JOURNAL_EXPORT_FORMATS/

```
           export
               serializes the journal into a binary (but mostly text-based) stream suitable for backups and network
               transfer (see Journal Export Format[1] for more information). To import the binary stream back into
               native journald format use systemd-journal-remote(8).

           json
               formats entries as JSON objects, separated by newline characters (see Journal JSON Format[2] for more
               information). Field values are generally encoded as JSON strings, with three exceptions:

                1. Fields larger than 4096 bytes are encoded as null values. (This may be turned off by passing
                   --all, but be aware that this may allocate overly long JSON objects.)

                2. Journal entries permit non-unique fields within the same log entry. JSON does not allow non-unique
                   fields within objects. Due to this, if a non-unique field is encountered a JSON array is used as
                   field value, listing all field values as elements.

                3. Fields containing non-printable or non-UTF8 bytes are encoded as arrays containing the raw bytes
                   individually formatted as unsigned numbers.

               Note that this encoding is reversible (with the exception of the size limit).
```

I'd like to do a little more research here to understand the forward and backwards compatability of the export/json/etc options.. and which of those options can be used in combination with the time-based exports.

But if it works out, this would seem to work towards a possibly better ultimate solution as then you can parse/filter the received logs based on date/unit/etc with journalctl in all cases - with or without all-logs/time limits.

> `journalctl` seems to be backward compatible in reading old journal logs:, at least I can run on F36 `journalctl --file ..` for a journal file from F28,. _Usually_ (but not always!) we investigate sosreports from "older systems" on a "newer system". If that does not happen, we have a problem like RHEL8 can't decode journalctl logs from RHEL9.

This has also been my experience however it's not uncommon for our team at least on the Ubuntu side to potentially be a major release behind the investigated machine and perhaps is not entirely unexpected for you either. 

> Also we can't be guaranteed with the backward compatibility. And from my personal experience with before mentioned `sar`, the burden "I need to copy the binary file to an appropriate system to decode it" is something I would like to avoid - esp. for something like system logs.

For what it's worth, sar includes a mode to convert those files since v9.1.6 - just run "sadf -c old_datafile > new_datafile". Discovered that a little later than I wished :)

A bunch of tooling like sar-jitsu had re-implemented parsers or other things tried to use a relevant sar version but I eventually figured out you can just convert them. Though this conversion is a little slow it's super CPU heavy for probably a silly reason - it's been on my todo list to try and fix that - I think it's because it does a lot of tiny small writes.

pmoravec:
> I think most end users would still have an expectation of the typical journal outputs in `sos_commands/logs/` regardless of size limiting.
> 
> @pmoravec @jjansky1 - thoughts?

I did some more research within Red Hat (support and Insights team) and got a negative feedback on this change / interest in keeping the `sos_commands/logs/` files in either case - at least on `RedHat` distros.

arif-ali:
abandoning the change based on legacy reasons, and for purpose of parsing data quickly

PR #3046, opened 2022-10-12
[ovn_central] Add ovsdb directory for Ubuntu

lathiat:
Debian & Ubuntu OVN packages store the ovsdb files in /var/lib/ovn

Signed-off-by: Trent Lloyd <trent.lloyd@canonical.com>

---

### DIFF
--- a/sos/report/plugins/ovn_central.py
+++ b/sos/report/plugins/ovn_central.py
@@ -150,7 +150,8 @@ class OVNCentral(Plugin):
                 self.path_join('/usr/local/etc/openvswitch', dbfile),
                 self.path_join('/etc/openvswitch', dbfile),
                 self.path_join('/var/lib/openvswitch', dbfile),
-                self.path_join('/var/lib/ovn/etc', dbfile)
+                self.path_join('/var/lib/ovn/etc', dbfile),
+                self.path_join('/var/lib/ovn', dbfile)
             ])
             if ovs_dbdir:
                 self.add_copy_spec(self.path_join(ovs_dbdir, dbfile))


lathiat:
Note: I am happy to accept this PR as is, however, it seems ripe for breaking out the path into the distro-specific classes, however, there are so many paths I am not really confident that I could cover all bases for versions and distros but if someone more familiar with the RedHat packaging over time could give me some advice on that I'd be happy to restructure it.

It seems at least in part this long list is because it originates from the openvswitch plugin which was duplicated to the oVN plugin and had a more historical set of paths on top of OVN - though seems some early versions of OVN (or maybe packages) had the ovn dbs in the openvswitch paths where as at some point they moved into the OVN specific paths so we maybe still need those.

```
        ovs_dbdir = os.environ.get('OVS_DBDIR')
        for dbfile in ['ovnnb_db.db', 'ovnsb_db.db']:
            self.add_copy_spec([
                self.path_join('/var/lib/openvswitch/ovn', dbfile),
                self.path_join('/usr/local/etc/openvswitch', dbfile),
                self.path_join('/etc/openvswitch', dbfile),
                self.path_join('/var/lib/openvswitch', dbfile),
                self.path_join('/var/lib/ovn/etc', dbfile)
                self.path_join('/var/lib/ovn', dbfile)
            ])
            if ovs_dbdir:
                self.add_copy_spec(self.path_join(ovs_dbdir, dbfile))
                
        self.add_journal(units="ovn-northd")
```

pmoravec (approved):
Not sure if Red Hat distros do store a db file there so I can serve as "the" reviewer requested in https://github.com/sosreport/sos/pull/3046#issuecomment-1275535658). But collecting a filepath that does not exist is safe. So purely from sos perspective, ACK.

TurboTurtle (approved):

PR #3049, opened 2022-10-25
[ovn_central] Add --no-leader-only to ovn-nbctl/ovn-sbctl commands

fernandoroyosanchez:
The option --no-leader-only will allow to run ovn-nbctl or 
ovn-sbctl in all nodes, independently if we are in a clustered 
or non-clustered enviroment.

The current check of executing commands only on nodes identified 
as Leader, causes that in non-cluster environments (active/backup 
mode) ovn-sbctl commands are not executed, not reporting the dump 
of the tables.

By adding this parameter to the execution, which does not harm 
non-clustered environments (active/backup), the command is executed 
over the local database.

---


pmoravec (approved):
ACK from my side, worth reviewing from Ubuntu/Debian side (cc @lathiat and @dosaboy) if the changes are compatible with those distros.

PR #3052, opened 2022-10-26
[ovn_central] Query ovn_controller SBDB connection with ovn-appctl

fernandoroyosanchez:
This additional command will tell us the ovn_controller connection 
status to OVN SBDB.

This is done thanks to ovn-appctl command using -t option to specify
the target over the ovn_controller socket, a function has been created
in order to located the ovn_controller socket path that is name is
different from each controller node.

---

### DIFF
--- a/sos/report/plugins/ovn_central.py
+++ b/sos/report/plugins/ovn_central.py
@@ -16,6 +16,7 @@ from sos.report.plugins import (
 )
 import json
 import os
+import re
 
 
 class OVNCentral(Plugin):
@@ -24,6 +25,26 @@ class OVNCentral(Plugin):
     plugin_name = "ovn_central"
     profiles = ('network', 'virt')
     containers = ('ovn-dbs-bundle.*', 'ovn_cluster_north_db_server')
+
+    def _find_ovn_controller_sock(self):
+        _sfile = os.path.join(self.ovn_controller_sock_path,
+                              self.ovn_controller_sock_regex)
+        if self._container_name:
+            res = self.exec_cmd("ls %s" % self.ovn_controller_sock_path,
+                                container=self._container_name)
+            if res['status'] != 0 or '\n' not in res['output']:
+                self._log_error(
+                    "Could not retrieve ovn_controller socket path "
+                    "from container %s" % self._container_name
+                )
+            else:
+                pattern = re.compile(self.ovn_controller_sock_regex)
+                for filename in res['output'].split('\n'):
+                    if pattern.match(filename):
+                        return os.path.join(self.ovn_controller_sock_path,
+                                            filename)
+        # File not found, return the regex full path
+        return _sfile
 
     def get_tables_from_schema(self, filename, skip=[]):
         if self._container_name:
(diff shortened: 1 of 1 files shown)


sos/report/plugins/ovn_central.py
```diff
@@ -25,6 +25,20 @@ class OVNCentral(Plugin):
     profiles = ('network', 'virt')
     containers = ('ovn-dbs-bundle.*', 'ovn_cluster_north_db_server')
 
+    def _find_ovn_controller_sock(self):
+        if self._container_name:
+            cmd = "find %s -name %s" % (self.ovn_controller_sock_path,
+                                        self.ovn_controller_sock_regex)
+            res = self.exec_cmd(cmd, timeout=None, foreground=True,
+                                container=self._container_name)
```
TurboTurtle, line 33: Use `stat` here - we aren't guaranteed to have `find` available in a given container, whereas `stat` is from `coreutils` which is much more common.

```python
_sfile = os.path.join(self.ovn_controller_sock_path,
                      self.ovn_controller_sock_regex)
res = self.exec_cmd("stat -t %s" % _sfile",
                    container=self._container_name)
[...]
```

sos/report/plugins/ovn_central.py
```diff
@@ -25,6 +25,20 @@ class OVNCentral(Plugin):
...
+            if res['status'] != 0:
+                self._log_error(
+                    "Could not retrieve ovn_controller socket path "
+                    "from container %s" % self._container_name
+                )
+                return
```
TurboTurtle, line 39: Should this be `return ''` instead? Otherwise the `format()` below will convert the string to `ovn-appctl -t None connection-status`.

Alternatively, break out the connection-status command and gate it with a conditional.

sos/report/plugins/ovn_central.py
fernandoroyosanchez: Good point, I make some test, but stat not accept the * included in the self.ovn_controller_sock_regex, when I test on shell the command expand the '*' but running in code the output looks as:

`stat: cannot statx '/var/run/openvswitch/ovn-controller.*.ctl': No such file or directory`

So taking your tip about use a command on coreutils I move the solution to ls + pattern matching over the regex.

sos/report/plugins/ovn_central.py
fernandoroyosanchez: It is not really necessary, None will be appended to the output file name, but to make life easier for future parsing of the output file I will return the proposed _sfile in case we are not able to locate the real sock path.

TurboTurtle (commented):
Ack to the code change.

Please add a DCO line to the commit message, as it is required via our [Contribution Guidelines](https://github.com/sosreport/sos/wiki/Contribution-Guidelines)

fernandoroyosanchez:
> Ack to the code change.
> 
> Please add a DCO line to the commit message, as it is required via our [Contribution Guidelines](https://github.com/sosreport/sos/wiki/Contribution-Guidelines)

Sorry but after read the Contribution-Guidelines I'm a little confused, By DCO do you mean the Signed-off-by?

pmoravec (approved):

fernandoroyosanchez:
> Correct. The `-s` flag to `git commit` (or `git commit --amend`) will add that for you.

done! thx!

TurboTurtle (approved):

PR #3053, opened 2022-10-27
cmd fwdownloader may never end sometimes, add timeout for it

gaojun0827:
[seagate_ses] Add timeout for fwdownloader in seagate_ses plugin.

Recently, I found the **fwdownloader -ses**  command would never finish sometime when run sos report .
It will cause sos report hang.  
I have manual test in my environment , The parameter plugin_timeout can't restrict plugin execution time if the inside command execution time over plugin_timeout.

Here is invoke stack when hang problem occurred.
image

### DIFF
--- a/sos/utilities.py
+++ b/sos/utilities.py
@@ -176,7 +176,7 @@ def sos_get_command_output(command, timeout=TIMEOUT_DEFAULT, stderr=False,
             os.chdir(chdir)
 
     def _check_poller(proc):
-        if poller():
+        if poller() or proc.poll() == 124:
             proc.terminate()
             raise SoSTimeoutError
         time.sleep(0.01)
@@ -240,6 +240,7 @@ def sos_get_command_output(command, timeout=TIMEOUT_DEFAULT, stderr=False,
                     _output.close()
                 # until we separate timeouts from the `timeout` command
                 # handle per-cmd timeouts via Plugin status checks
+                reader.running = False
                 return {'status': 124, 'output': reader.get_contents(),
                         'truncated': reader.is_full}
         if to_file:
--- a/bin/sos
+++ b/bin/sos
@@ -20,5 +20,6 @@ except KeyboardInterrupt:
 if __name__ == '__main__':
     sos = SoS(sys.argv[1:])
     sos.execute()
+    os._exit(0)
 
 # vim:ts=4 et sw=4


TurboTurtle:
> The parameter plugin_timeout can't restrict plugin execution time if the inside command execution time over plugin_timeout.

This doesn't track. `plugin_timeout` is handled by the top-level thread pool and is entirely separate from command timeout, which is controlled via the `timeout` wrapper. Setting a higher timeout for the command wrapper would not resolve this as the thread timeout would trigger before the command wrapper.

If the plugin is hanging despite _both_ timeouts in place, that points to something else at play here with the command in question. I don't have `fwdownloader` available to me, so we'll need to investigate this together.

- If you run `fwownloader -ses` manually yourself, what is the behavior? Does it hang? Is it waiting for some input?
- We've had some reports in the past of some commands not behaving nicely with timeouts if they are backgrounded. Instead of increasing `timeout`, what happens if you set `foreground=True` in the `collect_cmd_output()` call?

gaojun0827:
Hello,

The answers about your question:

1. If you run fwownloader -ses manually yourself, what is the behavior? Does it hang? Is it waiting for some input?
This command is used to collect Seagate device list , here is example output.
image
It does not hang, it doesn't need any input.
This is an intermittent problem in sosreport.
In most of the time, seagate_ses plugin can be executed successfully, On rare occasion(maybe under 10%), it will hang in seagate_ses plugin forever.

2. We've had some reports in the past of some commands not behaving nicely with timeouts if they are backgrounded. Instead of increasing timeout, what happens if you set foreground=True in the collect_cmd_output() call?
I have not test this, I will test this in my environment.

About `plugin_timeout`, `cmd_timeout` and `timeout` parameter in `collect_cmd_output()`

you said " thread timeout would trigger before the command wrapper." but I have different test result. 
here is code
```
from sos.report.plugins import Plugin, IndependentPlugin
class JuniusPlugin(Plugin, IndependentPlugin):
    plugin_name = "junius_plugin"
    short_desc = 'Junius test'
    plugin_timeout = 10
    #cmd_timeout = 10
    def setup(self):
        res = self.collect_cmd_output('date')
        print(res)
        res = self.collect_cmd_output('sleep 60')
        print(res)
        res = self.collect_cmd_output('date')
        print(res)

```
Run code: `time sos report --only-plugin junius_plugin `

This plugin didn't finish in 60 seconds.
I think both `plugin_timeout`, `cmd_timeout` is necessary  for a plugin.
The cmd_timeout have same effect with `timeout` parameter in `collect_cmd_output()`

TurboTurtle:
Ah - I see the discrepancy.

`plugin_timeout` is applied to actually _running_ the plugin, i.e. `Plugin.collect()` which is when the commands defined by the plugin are executed. The `setup()` method is used to _define_ what files and commands are collected via the `add_copy_spec()` and `add_cmd_output()` methods respectively.

The `collect_cmd_output()` method is intended to be used to run a command in `setup()` that needs to run right then and there in order to use that output to define some additional command collections. E.G. running a command that lists partitions on a system to then iterate over that output to define another series of commands that needs to use the list of discovered partitions as part of those commands.

So in your example, the `collect_cmd_output()` methods are being used incorrectly since their output is not being used in the `setup()` method to define further collections. There are (currently) no time restrictions on `setup()` - meaning your example would be expected to run for the whole 60 seconds of the `sleep` command.


> This is an intermittent problem in sosreport.

sos is just shelling out to execute the specified command. If that hangs at any point, it's because the command being executed is hanging. If this is intermittent, that means the command has intermittent problems.

> The cmd_timeout have same effect with timeout parameter in collect_cmd_output()

`cmd_timeout` sets the default timeout for any command executed by a plugin. `collect_cmd_output()` as stated earlier executes during `setup()` - the `timeout` parameter, be it directly set by the call or inheriting the default `cmd_timeout` is what would apply there.

> I think both plugin_timeout, cmd_timeout is necessary for a plugin.

Both of these are always set for every plugin. The default for both is 300 seconds. Defining them in a plugin class simply overrides this default.

gaojun0827:
Hi, 

Thanks for those really detailed explanation!!! 

if command ` fwdownloader -ses`  really hang sometimes, to avoid sos stuck in this plugin , we could try this `collect_cmd_output('fwdownloader -ses',foreground=True)` to see if this problem re-occurred.

Do you have other advice to avoid or investigate this hang problem?  logs, setting ... any thing that I can do 

Thanks

TurboTurtle:
That's where I'd start - try and catch when the command hangs and work backwards from there.

gaojun0827:
Hello, 

After nearly a month， I finally catch the error on the spot.
The "fwdownload -ses" command would hang when fibre channel device is busy .

Invoke stack
1. fwdownloader ses
System call : clone(child_stack=NULL, flags=CLONE_PARENT_SETTID|SIGCHLD, parent_tid=[183985]) = 183985
image
2.sh -c sg_map -x | grep -iw 13 | awk -F " " '{if ($6=="13") print $1}' > /tmp/jbifscan183984
image
System call: openat(AT_FDCWD, "/dev/nst4", O_RDONLY|O_NONBLOCK
3. /dev/nst4
this is device that is connected by fiber channel, it can't be access at that time.

I have try the `collect_cmd_output('fwdownloader -ses',foreground=True)` during the time that error occurred and It doesn't work.

Test steps:
cp /usr/lib/python3.6/site-packages/sos/report/plugins/seagate_ses.py /usr/lib/python3.6/site-packages/sos/report/plugins/seagate_ses_foreground.py
vim /usr/lib/python3.6/site-packages/sos/report/plugins/seagate_ses_foreground.py
sos report --batch --only-plugin seagate_ses_foreground


We could separate  code to follow layer:
layer1: seagate_ses.py
layer2: fwdownload -ses
layer3: sh 
layer4: sg_map -x 

There is a default cmd_timeout=600 in seagate_ses(layer1),  timeout would not work if hang occur in the deeper layers( layer4: sg_map -x).

Do we have any method to avoid this situation?

gaojun0827:
I think I finally find the root cause of this problem.
If a command(e.g. fwdownload -ses) can't be executed correctly with /usr/bin/timeout command, plugin will hang and sleep forever.

Some commands are work will with timeout command, e.g.
image
Some commands will hang with timeout command, it can't finish with required time  e.g.
image

**The sos/utilities.py sos_get_command_output function have two mechanisms to prevent timeouts.**
1. Use linux command /usr/bin/timeout as implementation 
image

This implementation have a big drawback, this command may execute forever if SIGTERM was ignore by subprocess.
timeout --foreground 300s fwdownloader -ses
If no signal is given, timeout sends the SIGTERM signal to the managed command when the time limit is reached. You can specify which signal to send using the -s (--signal) option.
SIGTERM, the default signal sent when the time limit is exceeded, can be caught or ignored by some processes. In those situations, the process continues to run after the termination signal is sent.
reference: https://linuxize.com/post/timeout-command-in-linux/


2. Use _check_poller function to check if process reach timeout.
It also have a drawback, When run `sos report` This method point to `sos/report/plugins/__init__.py class plugin check_timeout()` . It return the value of _timeout_hit which is always false.

image
image


I think maybe we can enhance the first timeout mechanism to avoid this problem.
image
The SIGKILL will force sub-process exit if command reach timeout.


What do you think ? is this a good idea?

TurboTurtle:
Great investigative work!

I don't have any immediate concerns about using sigkill, however one part of the analysis jumps out at me:

>  It return the value of _timeout_hit which is always false.

This should not be the case. `_timeout_hit` should be toggled by the nested thread executor that runs the plugin - however, this is apparently not functional when an executing command does not respond to sigterm.

I wonder if another solution to this would be to add logic to the `_timeout_hit` check to consider actual running time within the plugin, rather than solely at the thread executor level. I.E., we'd check for both `_timeout_hit` _and_ a new "has this plugin been running for more than the configured timeout value" directly in `_check_poller()` which would map to here - https://github.com/sosreport/sos/blob/main/sos/report/plugins/__init__.py#L753

We'd need to capture a start time within the plugin (we already do this for the manifest in https://github.com/sosreport/sos/blob/main/sos/report/__init__.py#L1283 so we would just need to move a few bits around) and then compare that with the current time when `check_timeout()` is called.


I'm honestly not sure if that would solve the issue in this specific case or not. If you can test that out, I think that'd be a preferred approach if it works. If it doesn't, then ack to a PR to using sigkill via the `timeout` command.

pmoravec (approved):
The change will affect `sos` on a few use cases, but I cant see any negative impact - ACK.

Worth updating the commit message per https://github.com/sosreport/sos/wiki/Contribution-Guidelines .

gaojun0827:
### Problem 2: The sos/utilities.py class AsyncReader deadlock when sub-process(parent) have deadlock child process

The problem also occurred when invoke  sos/utilities.py sos_get_command_output function.

Here is how to reproduce this problem.
**1. Create a child process Ignore SIGTERM. uninterrupted_sleep120**
Code segment see problem 1
**2. Create a parent process invoke child process**
nbapp625:/home/maintenance # cat father_sleep120.sh
/home/maintenance/uninterrupted_sleep120
**3. Run father process(sub-process) and use AsyncReader  to check process running state.**
Enter  python3.6
```
import subprocess
import collections
import threading
class AsyncReader(threading.Thread):
    def __init__(self, channel, sizelimit, binary):
        super(AsyncReader, self).__init__()
        self.chan = channel
        self.binary = binary
        self.chunksize = 2048
        self.slots = None
        if sizelimit:
            sizelimit = sizelimit * 1048576  # convert to bytes
            self.slots = int(sizelimit / self.chunksize)
        self.deque = collections.deque(maxlen=self.slots)
        self.running = True
        self.start()
    def run(self):
        try:
            while True:
                print("Reader Thread in while loop")
                line = self.chan.read(self.chunksize)
                print(line)
                print("Reader Thread, finish read cmd output")
                if not line:
                    # Pipe can remain open after output has completed
                    break
                self.deque.append(line)
        except (ValueError, IOError):
            print("Reader Thread in Except")
            # pipe has closed, meaning command output is done
            pass
        print("Reader Thread set self.running = False")
        self.running = False
```
Test code

```
p = subprocess.Popen( ['timeout', '10s', 'bash', '/home/maintenance/father_sleep120.sh'],stdout=subprocess.PIPE,stdin= subprocess.PIPE)
reader = AsyncReader(p.stdout, None, None)
p.poll()
reader.running
```
image

reader.running would be set to False after 120 seconds.

**Root Cause analysis:**
The p.stdout.read() block state can't represent process is running， it  means the process is running or its child-process is running.
https://github.com/sosreport/sos/blob/main/sos/utilities.py#L462
line = self.chan.read(self.chunksize) <== it will read child process output even father process is finish.

**How the problem influence seagate_ses plugin** 
This command 'fwdownloader -ses' may have some deadlock child-process sometimes, so this plugin will hang forever.
https://github.com/sosreport/sos/blob/main/sos/report/plugins/seagate_ses.py#L27

**What I need ?**
I don't know how to fix this problem now. The mechanism of AsyncReader seems to not perfect.
I would like your guys can give me some advices to fix this problem .
I need fix this in our product.

Thanks

TurboTurtle:
First, with regards to the sigkill change. Ack to the change, but please see the [contribution guidelines](https://github.com/sosreport/sos/wiki/Contribution-Guidelines) for formatting of the commit message. The commit message should be titled like `[utilities] Change timeout signal to SIGKILL` and the message body describe that change (like what you currently have). We also require a DCO (signed-off-by) line on all commits, you can add this with the `-s` for `git commit`.


Second, I'm not entirely following the remaining issue. Are you saying that even using SIGKILL, the `fwdownloader` command may deadlock and not respond to that signal from `timeout`? If so, then that is not something we can address in sos - that's a problem that the `fwdownloader` maintainers need to correct. It's one thing for us to take a firmer stance on terminating processes, but if the commands being run don't adhere to well-defined and expected NIX signals then we can't be expected to pick up that slack.

gaojun0827:
Frist, I update the my pull request and format my commit message.

Second, Use SIGKILL can terminate process  `fwdownloader` , but can't terminate its child-process `sg_map` which is deadlock.
In this situation, the sos/utilities.py --> sos_get_command_output() function will execute forever.
```
p = Popen(expanded_args, shell=False, stdout=_output,
                  stderr=STDOUT if stderr else PIPE,
                  bufsize=-1, env=cmd_env, close_fds=True,
                  preexec_fn=_child_prep_fn)

        if not to_file:
            reader = AsyncReader(p.stdout, sizelimit, binary)  <=== this class mechanism have problem
        else:
            reader = FakeReader(p, binary)

        if poller:
            while reader.running:    <== reader.running is always True, even process(fwdownloader) is terminated
                print('p.poll():', p.poll())   <== I add this to debug.   p.poll() return None if process is running , p.poll() return integer if process end. 
                _check_poller(p)  
```
The `reader.running` should be True when process is running, and should be False when process is terminated. 
This work for normal situation that process don't have deadlock child-process 
When we run process `fwdownloader`, the process is terminated, but the `reader.running` is always True .
Because `fwdownloader` have a child-process `sg_map`  and it is in deadlock state. the program just stuck in while loop.
image

The problem cause by this statement.
https://github.com/sosreport/sos/blob/main/sos/utilities.py#L462
line = self.chan.read(self.chunksize) <== it will read child process output even father process is terminated.

gaojun0827:
Thanks, I think we made a big step to the final solution.

This change would let code pass the  seagate_ses plugin . but it will hang in the end of the sos report.

> The `124` return there indicates a timeout. I think it'd be safe and sensible to add a check for this in `_check_poller()`:
> 
> ```python
>     def _check_poller(proc):
>         if poller() or proc.poll() == 124:  # extend the conditional here with the check for 124
>             proc.terminate()
>             raise SoSTimeoutError
>         time.sleep(0.01)
> ```
image

The AsyncReader is a thread, it deadlock when `fwdownloader`  child-process deadlock , python program must wait for all thread finish.

I am looking at some theading.Thread API to kill the thread, but I didn't find any.
https://docs.python.org/3/library/threading.html

Do you have any idea to kill this thread?

TurboTurtle:
So, first thing's first - we'll need to modify the try block on line 233 of `sos/utilities.py` to include the entire conditional. E.G.:

```python
        try:
            if poller:
                while reader.running:
                    _check_poller(p)
            else:
                # override timeout=0 to timeout=None, as Popen will treat the
                # former as a literal 0-second timeout
                p.wait(timeout if timeout else None)
        except Exception:
            p.terminate()
            if to_file:
                _output.close()
            # until we separate timeouts from the `timeout` command
            # handle per-cmd timeouts via Plugin status checks
            return {'status': 124, 'output': reader.get_contents(),
                    'truncated': reader.is_full}
```


As for the AsyncReader...first, I just want to make sure here - you're seeing that we never come out of the `self.chan.read()` right? It's not that we're simply looping on that, but that it never returns, yes? If it is actually looping, then we just need to change the `while True` to `while self.running` and add a mechanism to set that to `False` independently of reviewing the return of the `read()`. But I'm assuming here that's not what is happening, unfortunately.

Generally speaking, it's not advisable to kill a thread directly. So our options boil down to refactoring `AsyncReader`, or forcibly terminating ourselves at the end of every sos run with something like `os._exit()`.

Ironically, `AsyncReader` is itself a thread to avoid deadlocking sos when compiling the size-limited output of a command being run. As far as I'm aware, there is no option available to us to keep these reads in the main thread while also being able to size limit the output in real-time and not deadlocking ourselves. For reference, before we had this approach, we stored _all_ output in memory before reducing that down to match the size limit, however that meant a single wayward command or output and we could very easily hit OOM.


I'm not a huge fan of the second option because we could conceivably hit a situation where we build up a bunch of deadlocked threads during an execution. I get that this situation is very unlikely, and those threads would be terminated via the `os._exit()`, but it's still not a great design choice.

gaojun0827:
It seems there is no perfect  solution

I change follow code and it work for me.

```
        try:
            if poller:
                while 
                    print("enter _check_poller()")
                    _check_poller(p)
            else:
                # override timeout=0 to timeout=None, as Popen will treat the
                # former as a literal 0-second timeout
                p.wait(timeout if timeout else None)
        except Exception:
            p.terminate()
            if to_file:
                _output.close()
            # until we separate timeouts from the `timeout` command
            # handle per-cmd timeouts via Plugin status checks
            return {'status': 124, 'output': reader.get_contents(),
                    'truncated': reader.is_full}
```
In AsyncReader, I delete sleep statement.

```
    def get_contents(self):
        """Returns the contents of the deque as a string"""
        # block until command completes or timesout (separate from the plugin
        # hitting a timeout)

        if not self.binary:
            return ''.join(ln.decode('utf-8', 'ignore') for ln in self.deque)
        else:
            return b''.join(ln for ln in self.deque)
```
In bin\sos
```
if __name__ == '__main__':
    sos = SoS(sys.argv[1:])
    sos.execute()
    os._exit(0)
```

Do you guys have any schedule to refactoring `AsyncReader`? 
My manager is pushing me to fix this problem in our product.   
We have hundreds of machines that deployed Red hat ISO 2 times in one week. I can't manually change all those machines.

I'd like to know an approximate completion time.
Or If you don't have a this refactoring `AsyncReader` plan, I'll come up with something else

again I really appreciate your help

TurboTurtle:
Personally, I don't have to space right now to look at refactoring as I have a number of other things in flight for sos both here and internally.

For the moment, I think I'd be ok with taking the changes to `sos_get_command_output()` (we'll also need to remove the initial `p.terminate()` in `_check_poller()`), and calling `os._exit(0)` as a temporary workaround. I think it'd be best if it goes under `sos/__init__.py` in `execute()`.

Feel free to put that change in this PR along with the timeout signal.

If you're able to find some time to look at refactoring AsyncReader, that'd be fantastic. At the very least we can collaborate on that in a new PR like we've done here.

gaojun0827:
I update my pull request.

Let me explain some details in this PR.

** proc.poll() equal -9 if we use `command = "timeout --signal SIGKILL %s %ds %s"` .**

Last time test, I forgot to add ` --signal SIGKILL` , so when the process is terminated , the proc.poll() equal 124.
This time, I change it to right way.

```
def _check_poller(proc):
        if poller():
            proc.terminate()
        # timeout command kill child-process with SIGKILL signal
        if poller() or proc.poll() == -9:
            raise SoSTimeoutError
        time.sleep(0.01)
```
image

According to the **man timeout** , timeout command will exit with 137 if sub-process is killed by SIGKILL.
 https://manpages.courier-mta.org/htmlman1/timeout.1.html
137 if COMMAND (or timeout itself) is sent the KILL (9) signal (128+9)
image


In python3.6, the timeout sub-process poll() function will return -9 if  sub-process is killed by SIGKILL
image
This also make sense , because 128 + 9 = 137. I guess python3.6 convert this internally.


Now , I also have a tight schedule in our project.  we could use this PR as a workaround. 
If I have free time in future , I also love to make some contributions in opensource community, Then we could refactor the AsyncReader.

I also open a ticket for Red hat for this question. Maybe Red hat will solve this refactoring problem before we do.

sos/utilities.py
```diff
@@ -471,8 +471,6 @@ def get_contents(self):
         """Returns the contents of the deque as a string"""
         # block until command completes or timesout (separate from the plugin
         # hitting a timeout)
-        while self.running:
-            time.sleep(0.01)
```
TurboTurtle, line 475: This is what blocks until the command completes or is killed by timeout. This should stay.

sos/utilities.py
```diff
@@ -226,22 +226,22 @@ def _check_poller(proc):
...
-                # handle per-cmd timeouts via Plugin status checks
-                return {'status': 124, 'output': reader.get_contents(),
-                        'truncated': reader.is_full}
+        except Exception:
+            p.terminate()
+            if to_file:
+                _output.close()
+            # until we separate timeouts from the `timeout` command
+            # handle per-cmd timeouts via Plugin status checks
+            return {'status': p.poll(), 'output': reader.get_contents(),
```
TurboTurtle, line 243: For this we need to return 124 for any timeouts, as many other parts of sos rely on that to identify the timeout as that is the standard exit code.

If `Popen()` is changing this internally, we need to change it back to what's expected.

sos/utilities.py
```diff
@@ -176,8 +176,8 @@ def _child_prep_fn():
             os.chdir(chdir)
 
     def _check_poller(proc):
-        if poller():
-            proc.terminate()
+        # timeout command kill child-process with SIGKILL signal
+        if poller() or proc.poll() == -9:
```
TurboTurtle, line 180: > Last time test, I forgot to add --signal SIGKILL , so when the process is terminated , the proc.poll() equal 124.

So, the SIGKILL is unnecessary here? If that's the case, I'd prefer we keep with the default SIGTERM that timeout uses, that way the exit code handling is "normal".

TurboTurtle (commented):
It seems there are several approaches here that work, and we've inadvertently come across a couple of them whereas only one might be required. Comments below.


> Maybe Red hat will solve this refactoring problem before we do.

That's Pavel and myself. As mentioned there is a decent amount of other efforts in flight so it may be a bit before we get to this in more depth.

Also, the underlying culprit in this is the deadlocking of the `fwdownloader` command's child processes - so the ultimate fix here lies in those utilities by way of preventing the deadlocks.

sos/utilities.py
gaojun0827: If the 'fwdownloader' command's child-process is deadlock,  self.running is always True, this would cause deadlock in this get_contents function.

In the exception handling,  this get_contents function is invoked. 
return {'status': p.poll(), 'output': reader.get_contents(),'truncated': reader.is_full}

We have two option to prevent deadlock
1. Delete running validation in get_contents function
2. Delete reader.get_contents() in Exception handling 
OR
We could create a new Exception (let's say ChildDeadlockException )
If exception = ChildDeadlockException , output='', else output=reader.get_contents()

sos/utilities.py
gaojun0827: Sure, I agree with you, I will change it in update.

sos/utilities.py
gaojun0827: For , 'fwdownloader'  command , yes , it is unnecessary  , because 'fwdownloader' process will be killed by SIGTERM or SIGKILL, the deadlock process is its child process 'sg_map'.
I add ' --signal SIGKILL' for potential problem in the future , if a process can't be killed by SIGTERM, This change would work.

If you prefer the default SIGTERM, I will change it back in update.

gaojun0827:
Pull request update, please  help me to review this.

gaojun0827:
I discovered another disappointing fact, The sos\report\plugins\scsi.py also use the `sg_map`, this is the deadlock process !!
https://github.com/sosreport/sos/blob/main/sos/report/plugins/scsi.py#L53
Now, after my code change, Our sos rereport can pass seagate_ses plugin , but it will hang in scsi plugin.
image

sos/utilities.py
```diff
@@ -226,22 +228,27 @@ def _check_poller(proc):
...
-                    _output.close()
-                # until we separate timeouts from the `timeout` command
-                # handle per-cmd timeouts via Plugin status checks
-                return {'status': 124, 'output': reader.get_contents(),
-                        'truncated': reader.is_full}
+        except ChildDeadlockError:
+            if to_file:
+                _output.close()
+            return {'status': 124, 'output': 'Error,Child process deadlock',
+                    'truncated': False}
```
TurboTurtle, line 243: It is entirely possible for us to receive usable output up until the point a specific command deadlocks. This would eliminate that output, which would not be helpful to the person(s) reviewing an sos report.

In the case of a deadlocked child process, the parent process (the one we actually directly care about) still terminates via `timeout`, so it should all still be handled by the existing exception handler.

sos/report/plugins/scsi.py
```diff
@@ -50,7 +50,6 @@ def setup(self):
         self.add_cmd_output("lsscsi -i", suggest_filename="lsscsi")
 
         self.add_cmd_output([
-            "sg_map -x",
```
TurboTurtle, line 53: Nack. Keep this, as it has been expressly requested by support teams. The fact we wrap this in a `timeout` command means we should still catch the timeout, and the deadlocked process would be killed when sos exits with the newly added `os._exit(0)`.

sos/utilities.py
gaojun0827: In existing exception handler, The reader.get_contents() is invoked.  The 'sos report' will stuck in the `while self.running:`
https://github.com/sosreport/sos/blob/main/sos/utilities.py#L462

This is why I create the ChildDeadlockError, 
If command is timeout, reader.get_contents() can work.
if command have deadlock child process. reader.get_contents() will execute forever.

TurboTurtle: `get_contents()` blocks on the internal `running` status, but that is entirely controlled by us. In the exception handling we can just directly set `reader.running = False`.

gaojun0827: I just modify my pull request as your ask, please check.

TurboTurtle (approved):
Ack, LGTM.

TurboTurtle:
I've restarted the Stage2 Cirrus tests - I'm not sure why they failed, but the packit builds also failed - those with git errors.

I'll keep an eye on the re-running tests and we'll go from there.

pmoravec (approved):

TurboTurtle:
Hmm..the stagetwo test failures are interesting:

```
2023-01-05 15:18:05,646 DEBUG: [plugin:timeout_test] unpacked command: binary=False, changes=False, chroot=True, cmd='sleep 15', cmd_as_tag=False, container_cmd=None, env=None, foreground=False, priority=10, root_symlink=None, runat=None, sizelimit=25, stderr=True, subdir=None, suggest_filename=None, tags=[], timeout=None, to_file=False
2023-01-05 15:18:05,646 INFO: [plugin:timeout_test] collecting output of 'sleep 15'
2023-01-05 15:18:15,650 INFO: Plugin timeout_test timed out
2023-01-05 15:18:15,653 WARNING: [plugin:timeout_test] command 'sleep 15' timed out after 300s
2023-01-05 15:18:15,653 DEBUG: [plugin:timeout_test] collected output of 'sleep' in 10.006226539611816 (changes=False)
2023-01-05 15:18:15,653 DEBUG: [archive:TarFileArchive] Making leading paths for sos_commands/timeout_test
2023-01-05 15:18:15,653 DEBUG: [archive:TarFileArchive] Making path /var/tmp/avocado_mpln1vcssos_tests.py.PluginTimeoutTest/sos.7n5nw17o/sosreport-cirrus-task-6466148359733248-2023-01-05-rqkuldd/sos_commands/timeout_test
2023-01-05 15:18:15,654 DEBUG: [archive:TarFileArchive] Making directory /var/tmp/avocado_mpln1vcssos_tests.py.PluginTimeoutTest/sos.7n5nw17o/sosreport-cirrus-task-6466148359733248-2023-01-05-rqkuldd/sos_commands/timeout_test
2023-01-05 15:18:15,654 DEBUG: [archive:TarFileArchive] added string at 'sos_commands/timeout_test/sleep_15' to FileCacheArchive '/var/tmp/avocado_mpln1vcssos_tests.py.PluginTimeoutTest/sos.7n5nw17o/sosreport-cirrus-task-6466148359733248-2023-01-05-rqkuldd'
2023-01-05 15:18:15,654 DEBUG: [plugin:timeout_test] unpacked command: binary=False, changes=False, chroot=True, cmd='echo I slept great', cmd_as_tag=False, container_cmd=None, env=None, foreground=False, priority=10, root_symlink=None, runat=None, sizelimit=25, stderr=True, subdir=None, suggest_filename='echo_good', tags=[], timeout=None, to_file=False
2023-01-05 15:18:15,654 INFO: [plugin:timeout_test] collecting output of 'echo I slept great'
2023-01-05 15:18:15,655 DEBUG: [plugin:timeout_test] unpacked command: binary=False, changes=False, chroot=True, cmd='sleep 30', cmd_as_tag=False, container_cmd=None, env=None, foreground=False, priority=10, root_symlink=None, runat=None, sizelimit=25, stderr=True, subdir=None, suggest_filename=None, tags=[], timeout=10, to_file=False
2023-01-05 15:18:15,655 INFO: [plugin:timeout_test] collecting output of 'sleep 30'
```

So we hit the timeout, we trigger the the time out handling (hence the `Plugin timeout_test timed out` message)...but then we still continue.

Ok, small adjustment needed. I was wrong about capturing the entire conditional in the try block. If we have a `poller` available, that still needs to raise the exception back to `Plugin`, regardless of if the it's the poller or `proc.poll()` that causes us to raise it.

This means that the total changes needed in `sos_get_command_output()` are:

```
@@ -176,7 +176,7 @@ def sos_get_command_output(command, timeout=TIMEOUT_DEFAULT, stderr=False,
             os.chdir(chdir)
 
     def _check_poller(proc):
-        if poller():
+        if poller() or proc.poll() == 124:
             proc.terminate()
             raise SoSTimeoutError
         time.sleep(0.01)
@@ -240,6 +240,7 @@ def sos_get_command_output(command, timeout=TIMEOUT_DEFAULT, stderr=False,
                     _output.close()
                 # until we separate timeouts from the `timeout` command
                 # handle per-cmd timeouts via Plugin status checks
+                reader.running = False
                 return {'status': 124, 'output': reader.get_contents(),
                         'truncated': reader.is_full}
         if to_file:

```

Meaning:

- Keep the terminate in `_check_poller()`
- Keep your changes to check for the exit status, **and** setting `reader.running = False` (which will unblock the `get_contents()` call in case of a deadlock)
- Keep the `os._exit(0)` addition, so that we don't hang the entire `sos` command waiting on a deadlocked thread.


Sorry for all the back and forth on this. Timeout handling gets a bit convoluted when you need to contend with multiple failure states.

sos/__init__.py
```diff
@@ -191,5 +191,6 @@ def _init_component(self):
 
     def execute(self):
         self._component.execute()
+        os._exit(0)
```
TurboTurtle, line 194: Ah, this would apparently actually need to go in `bin/sos` after `sos.execute()` is called there. We use this same hook when calling cleaner in-line from report, so we're prematurely exiting in that case.

TurboTurtle:
Ok, so the most recent errors are due to avocado not capturing the last block of `print()`s from `cleaner.execute()` that reports the final archive location and related stats. By that I mean, if you run with `avocado --show=test` to show "all" output, you'll see all of sos' output, including debugging, but not the final block of print statements.

..but I'm not sure _why_ that has stopped happening with this PR. Even sticking an empty `log_debug()` at the end of `execute()` resolves this, which only further confuses me.

gaojun0827:
I am not familiar with those unittest, just give me some time to understand it.
So， how could we solve this problem?

TurboTurtle:
This should be resolved now by #3106 being merged. Please rebase this PR on current `main` - that should be all that's needed.

gaojun0827:
Hi, Sorry for late reply, busy on other programming.
I just rebase my code, and update PR, let us wait unit-test result .

gaojun0827:
All test passed, how could I merge this Pull request.

gaojun0827:
@TurboTurtle Thank you for guiding me. I really appreciate that~

PR #3055, opened 2022-10-28
[ovn_central] Add commands to get OVN DBs file information

fernandoroyosanchez:
In order to automate the parsing of the plugin output files, by systems could only read file contentit, seems useful to add an output cmd file where data, such as the size or permissions, of the OVN DBs is directly available.

This patch modifies the current behaviour to copy the db files of each OVN DB to create a file containing the ls -lan output of the files. The list of possible locations of these *.db files is maintained for backwards compatibility but it has been necessary to iterate over them to only obtain the information file of those that really exist.

---

### DIFF
--- a/sos/report/plugins/ovn_central.py
+++ b/sos/report/plugins/ovn_central.py
@@ -145,15 +145,20 @@ class OVNCentral(Plugin):
         self.add_copy_spec("/etc/sysconfig/ovn-northd")
 
         ovs_dbdir = os.environ.get('OVS_DBDIR')
-        for dbfile in ['ovnnb_db.db', 'ovnsb_db.db']:
-            self.add_copy_spec([
-                self.path_join('/var/lib/openvswitch/ovn', dbfile),
-                self.path_join('/usr/local/etc/openvswitch', dbfile),
-                self.path_join('/etc/openvswitch', dbfile),
-                self.path_join('/var/lib/openvswitch', dbfile),
-                self.path_join('/var/lib/ovn/etc', dbfile),
-                self.path_join('/var/lib/ovn', dbfile)
-            ])
+        for dbfile in ["ovnnb_db.db", "ovnsb_db.db"]:
+            for path in [
+                "/var/lib/openvswitch/ovn",
+                "/usr/local/etc/openvswitch",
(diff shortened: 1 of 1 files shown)


sos/report/plugins/ovn_central.py
```diff
@@ -145,15 +145,20 @@ def setup(self):
...
+                "/etc/openvswitch",
+                "/var/lib/openvswitch",
+                "/var/lib/ovn/etc",
+                "/var/lib/ovn",
+            ]:
+                dbfilepath = self.path_join(path, dbfile)
+                if os.path.exists(dbfilepath):
+                    self.add_copy_spec(dbfilepath)
+                    self.add_cmd_output(
+                        "ls -lan %s" % dbfilepath, foreground=True)
```
TurboTurtle, line 161: The `os.path.exists()` check is superfluous. `add_copy_spec()` gracefully handles when a provided file path does not exist, and we won't capture output from any failed `ls` commands for the same reason.

TurboTurtle (commented):
Minor note below, otherwise looks fine.

Also, please add a DCO line to the actual commit message.

sos/report/plugins/ovn_central.py
fernandoroyosanchez: Yeah, but without that os.path.exists(), we can trust on add_copy_spect() to get the file copy, but how we can jump the add_cmd_output("ls -lan...") for files that dont exist without that check? because if we run add_cmd_output("ls -lan...") over an non-existing file I already checked that a output file was written to sosreport with a "file not found". Or maybe I missunderstand something?

TurboTurtle: Ah, I thought `ls` gave a different error code than it does in that scenario, my mistake.

TurboTurtle: Ack, this is fine.

fernandoroyosanchez:
> Minor note below, otherwise looks fine.
> 
> Also, please add a DCO line to the actual commit message.

done! thx!

pmoravec (approved):

TurboTurtle (approved):

PR #3057, opened 2022-11-01
[build] Build snap and push to latest/edge

arif-ali:
When a any MRs or commits are added to the main branch then automatically create a snap, and push to latest/edge channel

---

### DIFF
--- a/.github/workflows/snap.yaml
+++ b/.github/workflows/snap.yaml
@@ -1,0 +1,26 @@
+name: snap
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: snapcore/action-build@v1
+      id: build-snap
+    # Make sure the snap is installable
+    - run: |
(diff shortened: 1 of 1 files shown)


arif-ali:
This is based on the following 2 GH Actions

1. https://github.com/snapcore/action-build
1. https://github.com/snapcore/action-publish

I can provide the `STORE_LOGIN` details privately, if we want to go ahead with this.

This will will to get the latest sos as a snap, and we don't need to publish this manually

stable releases will still be done manually

.github/workflows/snap.yaml
```diff
@@ -0,0 +1,31 @@
+name: snap
+on:
+  push:
+    tags:
+      - '*'
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
```
TurboTurtle, line 10: Would this kick off a new build on any new PR looking to merge into main? If the intent is for only merged commits to create a build, shouldn't the `push` section above be sufficient?

TurboTurtle (commented):
Ack to the end goal of this, just one question about the action configuration.

.github/workflows/snap.yaml
arif-ali: ack, makes total sense :) totally missed that point

TurboTurtle (approved):

Issue #3008, opened 2022-08-18
Multus network-attach-def ipv6 is not obfuscated


ansvu:
Does sos clean 4.4 support ipv6 yet? 
I ran the test using sos clean 4.4, but it seems not cleaned those ipv6.
 Add: mvnr-du:du-ddc7f8cb6-qqw8p:3a88000e-bfe3-436e-b810-b2bd290fb66a:mvnr-du/du-f1c-conf(sriov):f1c {"cniVersion":"0.4.0","interfaces":[{"name":"f1c","sandbox":"/var/run/netns/4239d766-68e8-4c74-a983-6e4edf2583d3"}],"ips":[{"version":"6","interface":0,"address":"2607:c540:8c00:3318::34/64","gateway":"2607:c540:8c00:3318::1"}],"routes":[{"dst":"::/0"}],"dns":{}}

Thanks.

TurboTurtle:
No, `clean` does not yet have support for ipv6 network address. There is support for ipv6 mac addrs, but that's it.

ansvu:
Thank you, TT. Just wonder if you have this ipv6 feature support in the future?

TurboTurtle:
It's definitely on the list. The actual matching of addresses is not really an issue - the stance we have on consistent network topology after obfuscation is however.

By this I mean, with ipv4 we maintain network topology and relationships with addresses. If we find an address from a new subnet, we create a new obfuscation subnet and assign addresses from there - it's not complete random assignment or even scrubbing an address to simply "--obfuscated_addr_0001--". This was a conscious decision made at the request of RH support teams (as a quick example, having topology consistent addressing can highlight a routing issue whereas scrubbing all addresses to "addr_XYZ" would hinder if not outright prevent that).

So, for ipv6 we would want to maintain that consistency. Given how much more complex ipv6 addressing is, and how the address format is malleable, this becomes fairly difficult to do within sos.

If there is another python module that has the ability to spit out addresses (e.g. "create an ipv6subnet object, and from it be able to generate addresses within that subnet on the fly") then that would make this implementation pretty straightforward actually.

TurboTurtle:
@ansvu - I've just opened https://github.com/sosreport/sos/pull/3060 that I hope resolves this. If you are able, please try pulling either that branch or one of the generated RPMs and test it out and let us know if it works for you.

PR #3060, opened 2022-11-04
[cleaner,ipv6] Add support for IPv6 obfuscation

TurboTurtle:
This commit adds a new parser and accompanying map for obfuscating IPv6 addresses.

This new parser will attempt to capture valid IPv6 networks and addresses, and produce a mostly-randomized obfuscated pair. Due to the multiple formats an IPv6 address can take, some identifiers are necessary to preserve relevant information while still obfuscating actual addresses and networks.

For example, global unicast addresses that have more than one defined hextet (greater than /16 prefix) will always generate an obfuscated address starting with `534f` (or 'so', continuing the style of our mac address handling that uses 'sos' as an identifier). Addresses with a /16 prefix or less, will start with simply '53'. Private addresses, which start with `fd` will generate an obfuscated address starting with `fd53`, so that the contextual understanding that it is a private network/address can remain. Link-local addresses which start with `fe80::` will remain that way, only having the device hextets obfuscated, again keeping the contextual information that it is a link-local interface intact, as otherwise these obfuscations may confuse end users reviewing an sos report for problems.

Note that the address `::1` and `::/0` are explicitly skipped and never obfuscated, for the same reasons given above.

Additionally, this parser/map will write data to the default map (and any per-run private maps) differently than previous parsers. Rather than simply dumping the obfuscation pairs into the map, it is broken up via network, with hosts belonging to that network nested inside those network entries (still being json-formatted). Users will also note that the ipv6 entries in the map also have a `version` key, which is intended to be used for handling future updates to the parser/map when upgrading from an older sos version to a newer one. This may or may not be carried over to future updates to other parsers.

Closes: #3008
Related: RHBZ#2134906

---

### DIFF
--- a/sos/cleaner/mappings/ipv6_map.py
+++ b/sos/cleaner/mappings/ipv6_map.py
@@ -1,0 +1,282 @@
+# Copyright 2022 Red Hat, Inc. Jake Hunsaker <jhunsake@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+import ipaddress
+
+from random import getrandbits
+from sos.cleaner.mappings import SoSMap
+
+
+def generate_hextets(hextets):
+    """Generate a random set of hextets, based on the length of the source
+    hextet. If any hextets are compressed, keep that compression.
+
+    E.G. '::1234:bcd' will generate a leading empty '' hextet, followed by two
+    4-character hextets.
+
+    :param hextets:     The extracted hextets from a source address
+    :type hextets:      ``list``
+
+    :returns:           A set of randomized hextets for use in an obfuscated
+                        address
+    :rtype:             ``list``
+    """
+    return [random_hex(4) if h else '' for h in hextets]
+
+
+def random_hex(length):
+    """Generate a string of size length of random hex characters.
+
+    :param length:  The number of characters to generate
+    :type length:   ``int``
+
+    :returns:       A string of ``length`` hex characters
+    :rtype:         ``str``
+    """
+    return f"{getrandbits(4*length):0{length}x}"
+
+
+class SoSIPv6Map(SoSMap):
+    """Mapping for IPv6 addresses and networks.
+
+    Much like the IP map handles IPv4 addresses, this map is designed to take
+    IPv6 strings and obfuscate them consistently to maintain network topology.
+    To do this, addresses will be manipulated by the ipaddress library.
+
+    If an IPv6 address is encountered without a netmask, it is assumed to be a
+    /64 address.
+    """
+
+    networks = {}
+
+    ignore_matches = [
+        r'^::1/.*',
+        r'::/0',
+        r'fd53:.*',
+        r'^53..:'
+    ]
+
+    first_hexes = ['534f']
+
+    compile_regexes = False
+    version = 1
+
+    def conf_update(self, config):
+        """Override the base conf_update() so that we can load the existing
+        networks into ObfuscatedIPv6Network() objects for the current run.
+        """
+        if 'networks' not in config:
+            return
+        for network in config['networks']:
+            _orig = ipaddress.ip_network(network)
+            _obfuscated = config['networks'][network]['obfuscated']
+            _net = self._get_network(_orig, _obfuscated)
+            self.dataset[_net.original_address] = _net.obfuscated_address
+            for host in config['networks'][network]['hosts']:
+                _ob_host = config['networks'][network]['hosts'][host]
+                _net.add_obfuscated_host_address(host, _ob_host)
+                self.dataset[host] = _ob_host
+
+    def sanitize_item(self, ipaddr):
+        _prefix = ipaddr.split('/')[-1] if '/' in ipaddr else ''
+        _ipaddr = ipaddr
+        if not _prefix:
+            # assume a /64 default per protocol
+            _ipaddr += "/64"
+        try:
+            _addr = ipaddress.ip_network(_ipaddr)
+            # ipaddr was an actual network per protocol
+            _net = self._get_network(_addr)
+            _ipaddr = _net.obfuscated_address
+        except ValueError:
+            # A ValueError is raised from the ipaddress module when passing
+            # an address such as 2620:52:0:2d80::4fe/64, which has host bits
+            # '::4fe' set - the /64 is generally interpreted only for network
+            # addresses. We use this behavior to properly obfuscate the network
+            # before obfuscating a host address within that network
+            _addr = ipaddress.ip_network(_ipaddr, strict=False)
+            _net = self._get_network(_addr)
+            if _net.network_addr not in self.dataset:
+                self.dataset[_net.original_address] = _net.obfuscated_address
+            # then, get the address within the network
+            _hostaddr = ipaddress.ip_address(_ipaddr.split('/')[0])
+            _ipaddr = _net.obfuscate_host_address(_hostaddr)
+
+        if _prefix and '/' not in _ipaddr:
+            return f"{_ipaddr}/{_prefix}"
+        return _ipaddr
+
+    def _get_network(self, address, obfuscated=''):
+        """Attempt to find an existing ObfuscatedIPv6Network object from which
+        to either find an existing obfuscated match, or create a new one. If
+        no such object already exists, create it.
+        """
+        _addr = address.compressed
+        if _addr not in self.networks:
+            self.networks[_addr] = ObfuscatedIPv6Network(address, obfuscated,
+                                                         self.first_hexes)
+        return self.networks[_addr]
+
+
+class ObfuscatedIPv6Network():
+    """An abstraction class that represents a network that is (to be) handled
+    by sos.
+
+    Each distinct IPv6 network that we encounter will have a representative
+    instance of this class, from which new obfuscated subnets and host
+    addresses will be generated.
+
+    This class should be built from an ``ipaddress.IPv6Network`` object. If
+    an obfuscation string is not passed, one will be created during init.
+    """
+
+    def __init__(self, addr, obfuscation='', used_hexes=None):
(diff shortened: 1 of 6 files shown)


TurboTurtle:
Added a couple more unittests. I'm not sure if we'd be able to rely on any kind of ipv6 in GCP for a potential integration test, so I'll look into that a bit more. For the time being though, I _think_ the unittests are sufficient.

pmoravec:
(WIP comment, thinking loudly) Is it ok that same parts of an address are obfuscated differently for different whole address, like:

            "ff02::/64": {
                "obfuscated": "537b::/64",
                "hosts": {
                    "ff02::1:ff22:3aee": "537b::806f:7d58:269a",
                    "ff02::1:ff22:392b": "537b::a5c7:dd9a:70f3",
                }

I.e. shouldn't be the second line:

                    "ff02::1:ff22:392b": "537b::806f:7d58:70f3",

? (imho this is ok)

sos/cleaner/mappings/ipv6_map.py
```diff
@@ -0,0 +1,264 @@
...
+        if not _prefix:
+            # assume a /64 default per protocol
+            _ipaddr += "/64"
+        try:
+            _addr = ipaddress.ip_network(_ipaddr)
+            # ipaddr was an actual network per protocol
+            _net = self._get_network(_addr)
+            _ipaddr = _net.obfuscated_address
+        except ValueError:
+            # host bits are set, first get the network
```
pmoravec, line 97: Worth expanding the explanation when the exception can be raised? `ipaddress.ip_network(_ipaddr)` raises

    ValueError: 2620:52:0:2d80::4fe/64 has host bits set

but it took me some debugging to understand it..

sos/cleaner/mappings/ipv6_map.py
```diff
@@ -0,0 +1,264 @@
...
+        _hextets = self.network_addr.split(':')[1:]
+        _ob_hex = ['534f']
+        if all(not c for c in _hextets):
+            # we have only a single defined hextet, e.g. ff00::/64, so we need
+            # to not use the standard first-hex identifier or we'll overlap
+            # every similar address obfuscation
+            _ob_hex = f"53{random_hex(2)}"
+            while _ob_hex in self._first_hexes:
+                # prevent duplicates
+                _ob_hex = f"53{random_hex(2)}"
```
pmoravec, line 201: We are limited here by 256 options only, and I can imagine some systems referring to more uniq global addresses - in that case, cleaner would loop here forever. Not sure how realistic the scenario / problem is, though.

sos/cleaner/mappings/ipv6_map.py
```diff
@@ -0,0 +1,264 @@
...
+            return ''.join([
+                self._obfuscated_network,
+                ':'.join(generate_hextets(_host.split(':')))
+            ])
+
+        if addr.compressed not in self.hosts:
+            try:
+                _, _host = addr.compressed.split(self.network_addr.rstrip(':'))
+            except ValueError:
+                # network addr is simply '::'
```
pmoravec, line 251: A good catch :)

TurboTurtle:
> First, I dont know IPv6 in much detail, e.g. if the differentiation to link/global/local addresses is right for `sos`. Apart of that and a few code comments, here are some generic ones from user experience:
> 
> 0. +1 for the `networks` subsections, that improves clarity
> 1. Is it ok that same parts of an address are obfuscated differently for different whole address, like:
>    ```
>         "ff02::/64": {
>             "obfuscated": "537b::/64",
>             "hosts": {
>                 "ff02::1:ff22:3aee": "537b::806f:7d58:269a",
>                 "ff02::1:ff22:392b": "537b::a5c7:dd9a:70f3",
>             }
>    ```
> 

By obfuscated differently, do you mean that we obfuscate `:1:` to `:a5c7:` here, i.e. different lengths? I went with this approach for consistency with our obfuscations - we could potentially limit the obfuscated hextets to the same length as the source address, but I think we buy ourselves a decent amount of breathing room in terms of surety that our obfuscations are both sane and not easily picked apart by forcing every obfuscated hextet to the same length.


> 
> 2. Why just the first record is obfuscated?
> 
> ```
>     fdd9:90e3:5690:ed79::1
>     fdd9:90e3:5690:ed79:ffff:ffff:ffff:ffff
> ```
> 
> 3. Why these records are not obfuscated either?
> 
> ```
>     fdda:5cc1:23:4::1f
>     2001:0db8:85a3:0000:0000:8a2e:0370:7334
>     2001:db8:a::123    # while 2001:db8::2:1 is obfuscated well..
> 
>     # neither of a randomly generated address is obfuscated (are these really valid?)
>     cacc:9913:bf2f:f67b:9cb2:f0a8:9d7b:806a
>     cd0b:730b:bc32:5025:1a58:5296:fe28:73c8
>     d789:c9e9:d3f2:8409:2fab:d988:88f7:e5f2
>     0d17:449a:4651:61f9:db82:e794:86f5:86fc
>     e8d3:f121:908a:bf98:923f:6ed9:7867:7883
> ```
> 

Looks for both this, and point 2 above, that a parenthesis imbalance when I added the negative lookbehind came up that I didn't catch. I'll fix this early tomorrow.

> 4. `foreman` can log:
> 
> ` Automatically imported trusted_ca::ca from trusted_ca/ca into production` that is wrongly detected as address `ca::ca`.
> 

This is what the negative look behind is for, but it might be that I have it in the wrong position in the regex (or, really, another symptom of incorrect parens placement). I _think_ this should get cleared up with the same change as above.

Thanks for highlighting this in a practical log file - it helps refine the exclusionary pieces of the regex.

> 5. Obfuscating just a network, private map contains empty `hosts` map, which might be confusing but logical. Like when the only occurence of `2002:a00` is in `sos_commands/firewall_tables/nft_list_ruleset`:
> 
> ` ip6 daddr { ::/96, ::ffff:0.0.0.0/96, 2002::/24, 2002:a00::/24, ..` then private map contains:
> 
> ```
>             "2002:a00::/24": {
>                 "obfuscated": "534f:a09f::/24",
>                 "hosts": {}
>             },
> ```

Ack, it took me a bit to realize this as well when I first started testing this. There are several places in a "standard" sos report that can have network addresses without any hosts being encountered within them. A lot of these will come from examples in config file comments, but I think it'd be far riskier (more fragile) to put in more exceptions for this, rather than just excepting the network-only obfuscations.

sos/cleaner/mappings/ipv6_map.py
TurboTurtle: Good idea, we wouldn't be obfuscating anything else with that prefix as the first bits in the address in any condition, so makes sense to increment from there.

sos/cleaner/mappings/ipv6_map.py
TurboTurtle: Ack, can do.

pmoravec:
> > First, I dont know IPv6 in much detail, e.g. if the differentiation to link/global/local addresses is right for `sos`. Apart of that and a few code comments, here are some generic ones from user experience:
> > 
> > 0. +1 for the `networks` subsections, that improves clarity
> > 1. Is it ok that same parts of an address are obfuscated differently for different whole address, like:
> >    ```
> >         "ff02::/64": {
> >             "obfuscated": "537b::/64",
> >             "hosts": {
> >                 "ff02::1:ff22:3aee": "537b::806f:7d58:269a",
> >                 "ff02::1:ff22:392b": "537b::a5c7:dd9a:70f3",
> >             }
> >    ```
> 
> By obfuscated differently, do you mean that we obfuscate `:1:` to `:a5c7:` here, i.e. different lengths?

No, I meant that `:1:` is obfuscated once to `:806f:` and once to `:a5c7:` (likewise `:ff22:` is obfuscated once to `:7d58:` and once to `:dd9a:`). But since we are obfuscating address bits within their /64 address block, we does not need to have such refined identical replacement.

TurboTurtle:
Oh, I see.

The `::1:` and `:ff22` in this case are entirely device/host bits when considering this is a /64 (expanded this would be `ff02:0000:0000:0000:0000:1:ff22:3aee` so there is nothing to actually maintain in terms of address space (host bits are not related to each other, generally).

What _may_ be the case here is that `ff02::1:ff22` is supposed to all be network bits (a /96 network) but it was not logged that way, hence why we assumed a /64.

sos/cleaner/mappings/ipv6_map.py
TurboTurtle: Added incrementing from 53 on up for every 256 addresses generated like this.

sos/cleaner/mappings/ipv6_map.py
TurboTurtle: Added

pmoravec:
Should we clean also IPv4/IPv6 translation addresses (like, e.g. `64:ff9b:1234::123.45.67.89` (Global) or `64:ff9b:1:234::123.45.67.89` (Private)? Here neither ip parser or ipv6 one applies. (imho not required now, this is a spare use case).

Should we obfuscate Unique Local Addresses like `fdda:5cc1:23:4::1f` ? We don't do now, but I havent dug into its purpose if we should or not.

Why is `2001:db8::2:1` obfuscated (to e.g. `fd53:9369::bcbe`) but `2001:db8:a::123` isn't? (is it allowed address? dunno now..)

Everything else passed! \o/

TurboTurtle:
> Should we clean also IPv4/IPv6 translation addresses (like, e.g. `64:ff9b:1234::123.45.67.89` (Global) or `64:ff9b:1:234::123.45.67.89` (Private)? Here neither ip parser or ipv6 one applies. (imho not required now, this is a spare use case).
> 

I think for now this will remain a known limitation.

> Should we obfuscate Unique Local Addresses like `fdda:5cc1:23:4::1f` ? We don't do now, but I havent dug into its purpose if we should or not.
> 

Can you provide context for that - like an excerpt of where it was encountered? An address like that should be getting obfuscated.

> Why is `2001:db8::2:1` obfuscated (to e.g. `fd53:9369::bcbe`) but `2001:db8:a::123` isn't? (is it allowed address? dunno now..)
> 

Same here. Having the textual context of how these are encountered would help a lot. Both addresses are valid.

> Everything else passed! \o/

\o/ Getting there!

pmoravec (approved):
Works great for me!

TurboTurtle:
Minor docs update. Will merge once the test suite passes again.

PR #3065, opened 2022-11-13
[report] Prevent obfuscating tmpDir path before tarbal move

pmoravec:
When moving sos tarball from a private directory to /var/tmp, apply filename obfuscation just to the file and not the tmpDir path itself.

Resolves: #3065

---

### DIFF
--- a/sos/report/__init__.py
+++ b/sos/report/__init__.py
@@ -1619,12 +1619,12 @@ class SoSReport(SoSComponent):
 
                 # output filename is in the private tmpdir - move it to the
                 # containing directory.
-                final_name = os.path.join(self.sys_tmp,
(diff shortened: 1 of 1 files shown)


pmoravec:
Reproducer:

```
python3 bin/sos report --clean --keywords tmp -o system,host --batch --disable-parsers=ip,mac,username
..
Error moving archive file: /var/tmp/sos.rmyfph6e/sosreport-host0-2022-11-13-lkvrnxp.tar.xz
```

The reason is `final_name = cleaner.obfuscate_string(...` is applied to full name including the `/var/tmp` path.

Here it is bit artificial reproducer, but when ones uses custom --tmp-dir to a sensitive location, it becomes a real issue.

PR #3066, opened 2022-11-14
[testing] Switch to CodeQL from LGTM

TurboTurtle:
LGTM is shutting down shortly, as it has been folded into GH CodeQL. As such, we need to switch over to using CodeQL to maintain automatic code analysis.

Do this by using the provided CodeQL workflow configuration, and add a config file to limit the analysis to the appropriate `sos/` directory.

---

### DIFF
--- a/sos/report/plugins/openstack_instack.py
+++ b/sos/report/plugins/openstack_instack.py
@@ -81,16 +81,17 @@ class OpenStackInstack(Plugin):
             # capture all the possible stack uuids
             get_stacks = "openstack stack list"
             stacks = self.collect_cmd_output(get_stacks)['output']
-            stack_ids = re.findall(r'(\s(\w+-\w+)+\s)', stacks)
+            stack_ids = re.findall(r'(\|\s(((\w+-){4})\w+)\s\|)', stacks)
             # get status of overcloud stack and resources
-            for sid in stack_ids:
+            for _sid in stack_ids:
+                sid = _sid[1]
                 self.add_cmd_output([
-                    "openstack stack show %s" % sid[0],
-                    "openstack stack resource list -n 10 %s" % sid[0]
+                    "openstack stack show %s" % sid,
+                    "openstack stack resource list -n 10 %s" % sid
                 ])
 
                 # get details on failed deployments
-                cmd = "openstack stack resource list -f value -n 5 %s" % sid[0]
+                cmd = "openstack stack resource list -f value -n 5 %s" % sid
                 deployments = self.exec_cmd(cmd)
                 for deployment in deployments['output'].splitlines():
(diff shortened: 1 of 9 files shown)


TurboTurtle:
The config file seems to be correct now to restrict scanning to `sos/` and not include our test suite. However, there's still a lot of new warnings that LGTM was not presenting to us.

I think with this PR, I'll make another update to tackle the 2 alerts marked as 'high' - from there we can whittle down the new alerts over time.

sos/report/plugins/openstack_instack.py
```diff
@@ -81,16 +81,17 @@ def setup(self):
...
                 self.add_cmd_output([
-                    "openstack stack show %s" % sid[0],
-                    "openstack stack resource list -n 10 %s" % sid[0]
+                    "openstack stack show %s" % sid,
+                    "openstack stack resource list -n 10 %s" % sid
                 ])
 
                 # get details on failed deployments
-                cmd = "openstack stack resource list -f value -n 5 %s" % sid[0]
+                cmd = "openstack stack resource list -f value -n 5 %s" % sid
```
pmoravec, line 94: Why the change from `sid[0]` to `sid[1]` value in commands arguments?

TurboTurtle: The new regex groups the matches differently - I grabbed some output from some docs and built against, but it would be great if we could have some "real" validation on it.

pmoravec: Ah I see - but why `re.search` (that searches for the first instance only) when we need to spot all stack IDs?

Moreover, the code doesnt work for me on an an example output like:

```
+--------------------------------------+------------+-----------------+---------------------+--------------+
| ID                                   | Stack Name | Stack Status    | Creation Time       | Updated Time |
+--------------------------------------+------------+-----------------+---------------------+--------------+
| 70b9feca-8f99-418e-b2f1-cc38d61b3ffb | MYSTACK    | CREATE_COMPLETE | 2016-06-08T09:54:15 | None         |
| 82f75a62-9e03-4031-afb1-a63022f8dd68 | SOMESTACK  | ALSO_COMPLETE   | 2022-11-21T08:46:01 | None         |
+--------------------------------------+------------+-----------------+---------------------+--------------+
```

where the `for sid in stack_ids:` fails with `TypeError: '_sre.SRE_Match' object is not iterable`.

TurboTurtle: Good catch. I was overlooking this was scanning the entire output, not line by line like our other matching methods, so this should go back to `findall()`.

pmoravec (approved):
ACK to the current changes.

Many CodeQL warnings are (still) raised, we can fix them later on (or within this PR).

Issue #3072, opened 2022-11-16
[dnf] password/proxy_password directives in dnf.conf are not hidden by sosreport


miztake:
The dnf plugin tries to hide the password/proxy_password directive with ********.

https://github.com/sosreport/sos/blob/2556e64b36a1f312e331589dc77146e35c6941e8/sos/report/plugins/dnf.py#L127-L131

However, the /etc/dnf/dnf.conf where password/proxy_password would be written is out of scope.
We need to add /etc/dnf/dnf.conf as a file that hides password/proxy_password.

brozs:
Working on a patch.

PR #3074, opened 2022-11-16
[dnf] Scrub passwords from dnf.conf

brozs:
Currently the dnf plugin scrubs passwords from the repository files and DNF variables, however "password" and "proxy_password" can be defined in "/etc/dnf/dnf.conf".

This patch ensures that passwords are scrubbed from dnf.conf too.

Example of scrubbing:

Before:
  proxy_password = hackme
After:
  proxy_password = ********

Resolves: #3072

Signed-off-by: Stepan Broz <sbroz@redhat.com>

---

### DIFF
--- a/sos/report/plugins/dnf.py
+++ b/sos/report/plugins/dnf.py
@@ -125,9 +125,28 @@ class DNFPlugin(Plugin, RedHatPlugin):
         self.get_modules_info(modules['output'])
 
     def postproc(self):
+        # Scrub passwords in repositories and yum/dnf variables
+        # Example of scrubbing:
+        #
+        #   password=hackme
+        # To:
+        #   password=********
+        #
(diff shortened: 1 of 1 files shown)


sos/report/plugins/dnf.py
```diff
@@ -125,9 +125,28 @@ def setup(self):
...
 
+        # Scrub password and proxy_password from /etc/dnf/dnf.conf.
+        # This uses the same regex patterns as above.
+        #
+        # Example of scrubbing:
+        #
+        #   proxy_password = hackme
+        # To:
+        #   proxy_password = ********
+        #
```
pmoravec, line 149: I think we can refer to the example above and be more concise here, but OK.

pmoravec (approved):
Good catch.

TurboTurtle (approved):

PR #3076, opened 2022-11-21
[openstack_mistral] Collect mistral content itself

cjeanner:
Logs as well as deploy information are located in /var/lib/mistral as well. Until now, this location was overlooked, and important information about the deploy state, configuration and potential errors were missing from the initial SOS-Report, leading to time loss for both Customer and Support.

This patch intends to correct this situation, by ensuring the content is taken from the Undercloud.

Notes:
- this is especially important for OSP<17.0
- the location may content multiple subdirectories
- the location will contain the history of the different actions done by the operator

---

### DIFF
--- a/sos/report/plugins/openstack_mistral.py
+++ b/sos/report/plugins/openstack_mistral.py
@@ -1,0 +1,48 @@
+# Copyright (C) 2022 Red Hat, Inc.
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, RedHatPlugin
+
+
+MISTRAL_DIRECTORIES = [
+        '/var/log/mistral/',
+        '/var/lib/mistral/',
+        ]
+MISTRAL_LOGS = [
+        '/var/log/mistral/*.log',
+        '/var/lib/mistral/*/*.log',
+        ]
+
+
+class OpenStackMistral(Plugin, RedHatPlugin):
+    '''Gather Mistral directories content, both data from /var/lib/mistral
+    and its log from /var/log/mistral if it exists (older OSP).
+    The data also embed logs for the ansible runs launched via the service,
+    meaning we'll be able to properly debug failures therein. The rest of the
+    data are the generated environment files, also really useful in order
+    to debug an issue at deploy or day-2 operations.
+    We filter out on the presence of any "mistral" related container on the
+    host - usually the Undercloud presents mistral_engine, mistral_executor
+    and mistral_api.
+    '''
+
+    short_desc = 'OpenStack Mistral'
+
+    plugin_name = "openstack_mistral"
+    profiles = ('openstack', 'openstack_undercloud')
+    containers = ('.*mistral_engine',)
+
+    def setup(self):
+        if self.get_option('all_log'):
+            self.add_copy_spec(MISTRAL_DIRECTORIES)
+        else:
+            self.add_copy_spec(MISTRAL_LOGS)
+
+# vim: set et ts=4 sw=4 :
--- a/sos/report/plugins/openstack_instack.py
+++ b/sos/report/plugins/openstack_instack.py
@@ -44,12 +44,10 @@ class OpenStackInstack(Plugin):
 
         if self.get_option("all_logs"):
             self.add_copy_spec([
-                "/var/log/mistral/",
                 "/var/log/zaqar/",
             ])
         else:
             self.add_copy_spec([
-                "/var/log/mistral/*.log",
                 "/var/log/zaqar/*.log",
             ])
 


sos/report/plugins/openstack_mistral.py
```diff
@@ -0,0 +1,36 @@
...
+
+    short_desc = 'OpenStack Mistral'
+
+    plugin_name = "openstack_mistral"
+    profiles = ('openstack', 'openstack_undercloud')
+    logdir = '/var/lib/mistral/'
+    files = (logdir, )
+
+    def setup(self):
+        if self.ge_option('all_log'):
```
TurboTurtle, line 25: Typo, should be `get_option()`.

sos/report/plugins/openstack_mistral.py
```diff
@@ -0,0 +1,36 @@
...
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin
+import os
+
+
+class OpenStackMistral(Plugin):
```
TurboTurtle, line 15: This needs to also have at least either `IndependentPlugin` to be enabled by default across distributions, or a distro tagging class to enable on specific distributions, e.g. `RedHatPlugin`

TurboTurtle, line 15: A class docstring would be helpful to detail the plugin a bit more. For example, what is Mistral in the context of both an OSP component and the data collected in an sos report? Take a look at the [kernel plugin](https://github.com/sosreport/sos/blob/main/sos/report/plugins/kernel.py) for a good example of what we look for in new plugin docstrings - these will be used for `sos help` output (e.g. `sos help report.plugins.openstack_mistral`)

cjeanner:
Hello @TurboTurtle 

I did consider the _instack for this addition, but it's an old, deprecated name (I'll have to double-check with my team to be 100% sure, but iirc, it's down to osp-10, which is EOL for some time now).

So I'd rather keep mistral in its own thing and, on the contrary, move the `/var/log/mistral` and related from _instack to _mistral if you're OK with that.

As for container: yes, newer OSP versions run in container, and we already scrap the content of /var/log/containers (in the `container_log` plugin).
Thing is, at some point mistral, an openstack service, was used to schedule and run workflows used to deploy openstack from the Director node - and thus, we have some needed data in /var/lib/mistral. Without them, Support is usually to ask Customer to provide them manually.

I'll correct the comments you pushed (thanks for the review!) and move mistral log to this new plugin. And probably do some follow-up regarding the openstack_instack plugin itself. This name is dead and might confuse ppl.

Cheers,

C.

sos/report/plugins/openstack_mistral.py
cjeanner: Right - added one.

sos/report/plugins/openstack_mistral.py
cjeanner: woops....

TurboTurtle (approved):

cjeanner:
Hello there,

Any news on that one? I see it's "approved", so I guess the code is suitable for the project - do I need to do anything more to get it in?

Thank you!

Cheers,

C.

pmoravec:
The `openstack_mistral` plugin would be enabled by default on any system. While it technically would be a no-op on most of systems (until there is a `/var/[log|lib]/mistral` directory), still I feel this OpenStack-specific plugin shall not be enabled by default on each and every system.

Could you please add the same "trick" like:

https://github.com/sosreport/sos/blob/main/sos/report/plugins/openstack_instack.py#L150-L155

to condition the plugin by presence of that package (which would be still generic and RH-distro related only :) but if we dont have a better diagnostics for "Am I running on an OSP [Mistral|Instack|Sahara|..]", we should use what we have).

cjeanner:
@pmoravec updated accordingly, also "locking" it to RedHatPlugin instead of IndependantPlugin. I pushed a comment about "why openstack-selinux", that should make the choice clearer hopefully.

sos/report/plugins/openstack_mistral.py
```diff
@@ -0,0 +1,58 @@
...
+    profiles = ('openstack', 'openstack_undercloud')
+
+    def setup(self):
+        if self.get_option('all_log'):
+            self.add_copy_spec(MISTRAL_DIRECTORIES)
+        else:
+            self.add_copy_spec(MISTRAL_LOGS)
+
+
+class RedHatRDOManager(OpenStackMistral, RedHatPlugin):
```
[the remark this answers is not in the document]
cjeanner: Ah, let's move that package thing there then. Would be better/more readable. Note: OSP *is* redhat specific - it's a product. The way we use mistral there is also specific to OSP. Hence the need to get files.

Lemme check some of the other plugins related to openstack, and I'll push something shortly.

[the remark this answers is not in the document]
cjeanner: Could even do it in  a better way, by filtering using the "containers" keyword. Lemme know if it suits better.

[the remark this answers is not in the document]
TurboTurtle: If this component *only* runs in a container, then yes that is a great way to control plugin enablement.

If it can also run outside of a container, then having the `packages = ('openstack-selinux',)` control _as well_ as the `containers` control would be best.

TurboTurtle (approved):

pmoravec (approved):

Issue #3073, opened 2022-11-16
[postfix] /etc/postfix/sasl_passwd is collected by sosreport


miztake:
The postfix plugin collects /etc/postfix/.
However, the following paths are not collected.

https://github.com/sosreport/sos/blob/2556e64b36a1f312e331589dc77146e35c6941e8/sos/report/plugins/postfix.py#L64-L68

On the other hand, smtp_sasl_password_maps in /etc/postfix/main.cf is typically set to /etc/postfix/sasl_passwd.
(refer to https://www.postfix.org/SASL_README.html)
/etc/postfix/sasl_passwd is not specified as a forbidden path, so it is collected by sosreport.

/etc/postfix/sasl_passwd contains the password,
so it should be specified in the forbidden path so that sosreport does not collect it.

brozs:
There are multiple configuration options which may point to a map containing passwords. The following come to mind:

`lmtp_sasl_password_maps`
`smtp_sasl_password_maps`
`postscreen_dnsbl_reply_map`
`smtp_sasl_auth_cache_name`

A custom path is allowed, there are no defaults. It would require parsing of the configuration file.
Also, it would be better to scrub the individual map files rather than adding those to forbidden paths.

TurboTurtle:
I believe this is resolved by #3077, which is now merged, so I'm closing this.

if there is still some part of this unaddressed, please feel free to re-open.

PR #3077, opened 2022-11-22
[postfix] Exclude password files

Iwao-Miyake:
Changed to sosreport exclude files set by following options.
- `lmtp_sasl_password_maps`
- `smtp_sasl_password_maps`
- `postscreen_dnsbl_reply_map`
- `smtp_sasl_auth_cache_name`

Resolvs : #3073

Signed-off-by: Iwao Miyake <miyake.iwao@fujitsu.com>

---

### DIFF
--- a/sos/report/plugins/postfix.py
+++ b/sos/report/plugins/postfix.py
@@ -8,6 +8,8 @@
 
 from sos.report.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
 
+import re
+
 
 class Postfix(Plugin):
 
@@ -52,6 +54,42 @@ class Postfix(Plugin):
         finally:
             return fp
 
+    def forbidden_password_files(self):
+        forbid_attributes = (
+            "lmtp_sasl_password_maps",
+            "smtp_sasl_password_maps",
+            "postscreen_dnsbl_reply_map",
+            "smtp_sasl_auth_cache_name",
+        )
+        fp = []
+        prefix = 'hash:'
+        option_format = re.compile(r"^(.*)=(.*)")
+        try:
+            with open(self.path_join('/etc/postfix/main.cf'), 'r') as cffile:
+                for line in cffile.readlines():
+                    # ignore comment and check option format
+                    line = re.sub('#.*', '', line)
+                    option = option_format.match(line)
+                    if option is None:
+                        continue
+
+                    # sieving
+                    attribute = option.group(1).strip()
+                    if attribute in forbid_attributes:
+                        filepath = option.group(2).strip()
+                        # ignore no filepath
+                        if len(filepath) == 0:
+                            continue
+                        # remove prefix
+                        if filepath.startswith(prefix):
+                            filepath = filepath[len(prefix):]
+                        fp.append(filepath)
+        except Exception as e:
+            # error log
+            msg = f"Error parsing main.cf: {e.args[0]}"
+            self._log_error(msg)
+        return fp
+
     def setup(self):
         self.add_copy_spec([
             "/etc/postfix/",
@@ -67,6 +105,7 @@ class Postfix(Plugin):
             "/etc/postfix/ssl/",
         ])
         self.add_forbidden_path(self.forbidden_ssl_keys_files())
+        self.add_forbidden_path(self.forbidden_password_files())
 
 
 class RedHatPostfix(Postfix, RedHatPlugin):


sos/report/plugins/postfix.py
```diff
@@ -52,6 +52,31 @@ def forbidden_ssl_keys_files(self):
...
+        prefix = 'hash:'
+        try:
+            with open(self.path_join('/etc/postfix/main.cf'), 'r') as cffile:
+                for line in cffile.readlines():
+                    # ignore comments and take the first word after '='
+                    if line.startswith('#'):
+                        continue
+                    words = line.split('=')
+                    if words[0].strip() in forbid_attributes:
+                        filepath = words[1].split()[0]
```
pmoravec, line 72: Assume a corner case:

    lmtp_sasl_password_maps = 

which would raise `IndexError: list index out of range` error ;-)

What about:

    words = line.replace(' ', '').split('=')
    if words[0] in forbid_attributes and len(words) > 1:
        filepath = words[1]

(this won't work if the file path contains space - is that syntax allowed?)

pmoravec:
It seems the commits stepon each others toes. Let do this on your local branch:

1) Copy the desired final `postfix.py` file somewhere (it will be updated during the process).
2) rebase your local branch:

    git rebase --interactive HEAD~4

3) Now, either:
- edit the file such that you `drop` all four commits, then copy the backed `postfix.py` to the proper location, write `git add sos/report/plugins/postfix.py` and `git commit -s` and type the commit message again
- or edit the rebase file by `drop`-ing 3 of 4 commits and `edit`-ing just one commit with the right text inside, then  copy the backed `postfix.py` to the proper location, write `git add sos/report/plugins/postfix.py` and `git commit --amend` (to amend the left commit)

4) `git rebase --continue` to replay on top of your branch head
5) `git push --force` your branch

Iwao-Miyake:
Thank you to guide me.
I did not understand about git.
So your guidance makes me helped.

I tried modifying the process to regex to solve the input no filepath.

sos/report/plugins/postfix.py
```diff
@@ -52,6 +54,38 @@ def forbidden_ssl_keys_files(self):
...
+            "smtp_sasl_auth_cache_name",
+        )
+        fp = []
+        prefix = 'hash:'
+        option_format = re.compile(r"^(.*)=(.*)")
+        try:
+            with open(self.path_join('/etc/postfix/main.cf'), 'r') as cffile:
+                for line in cffile.readlines():
+                    # ignore comment and check option format
+                    line = re.sub('#.*', '', line)
```
pmoravec, line 71: It makes sense to scrub passwords also in comments, no? Two use cases:

1) Sometimes, previously used passwords are kept in comments as a kind of history. Even the historic passwords are worth to obfuscate, since they might be reused or still in use else-where (yes, neither that is a good password policy, but I have seen that in practise)

2) Sometimes, for debugging purposes, one can temporarily comment out existing password to test a password-less authentication.

Iwao-Miyake: I think it may be a different issue from excluding files.
In my humble opinion, If necessary, we should create a new issue.

pmoravec: Ah I apologize for my mental shortcut (we usually scrub passwords using regexp, while here we stop collecting files using regexp). I meant "stop collecting forbidden files mentioned also in comments". But I agree there would be less valid use cases for it, so I am OK with the current PR.

pmoravec (approved):

sos/report/plugins/postfix.py
```diff
@@ -52,6 +54,38 @@ def forbidden_ssl_keys_files(self):
...
+                        filepath = option.group(2).strip()
+                        # ignore no filepath
+                        if len(filepath) == 0:
+                            continue
+                        # remove prefix
+                        if filepath.startswith(prefix):
+                            filepath = filepath[len(prefix):]
+                        fp.append(filepath)
+        finally:
+            return fp
```
TurboTurtle, line 87: A return in a `finally` block will swallow raised exceptions. I think we should instead have an exception handler (log it for traceability) and have the return outside the try/except block.

TurboTurtle (commented):
Overall looks good, just one comment below regarding the try block.

Iwao-Miyake:
I compiled the commit added logs, possibly is not this needed?

In forbidden_password_files(), I have added to check non-exists path.
And the following logging process have been added for traceability.
- warning non-exists path
- error when the process catch exception

and changed the 'return' position.

sos/report/plugins/postfix.py
```diff
@@ -52,6 +55,51 @@ def forbidden_ssl_keys_files(self):
...
+                                    (attribute, filepath)
+                            msg = "[plugin:%s] %s" % (self.name(), msg)
+                            self.soslog.warning(msg)
+                            continue
+                        fp.append(filepath)
+        except Exception as e:
+            # error log
+            msg = "catch '%s' in forbidden_password_files()" % e.args[0]
+            msg = "[plugin:%s] %s" % (self.name(), msg)
+            self.soslog.error(msg)
```
TurboTurtle, line 100: `_log_*()` methods handle proper formatting here.

E.G.

```python
self._log_info(f"Filepath {filepath} specified by {attribute} does not exist.")
```

or/and

```python
self._log_error(f"Error parsing main.cf: {e}")
```

sos/report/plugins/postfix.py
```diff
@@ -52,6 +55,51 @@ def forbidden_ssl_keys_files(self):
...
+                        if filepath.startswith(prefix):
+                            filepath = filepath[len(prefix):]
+                        # ignore non-existent files
+                        if not os.path.exists(filepath):
+                            # warning log for traceability
+                            msg = "'%s' has non-existent path '%s'" % \
+                                    (attribute, filepath)
+                            msg = "[plugin:%s] %s" % (self.name(), msg)
+                            self.soslog.warning(msg)
+                            continue
```
pmoravec, line 94: This is not needed, imho. The worst what would happen is you feed `add_forbidden_path` with a non-existing file, and any file to be collected would be verified about bit bigger list (https://github.com/sosreport/sos/blob/91cce33a1c9b3cff0795c7e573813d3394cc4107/sos/report/plugins/__init__.py#L1413-L1416).

TurboTurtle (approved):

pmoravec (approved):

PR #3082, opened 2022-12-01
[collect] Remove --sos-cmd option

TurboTurtle:
Removes the `--sos-cmd` option for `sos collect`. Allowing passthru options in this manner is inherently flawed, and any attempts at sanitizing potentially malicious/dangerous values will always be a losing battle. Instead, `sos collect` should leverage available `report` options that are vetted and handled via the existing per-node capabilities checks that is well-defined for explicit passthru options.

---


pmoravec (approved):
ACK. This - as a deprecated option - should go to Release Notes of 4.5, imho.

PR #3083, opened 2022-12-02
[testing] Change location of mocked files for tests

TurboTurtle:
Previously, mocked files were kept under the `tests/test_data/` directory and generally mimic'd the file location they would be temporarily copied to during the execution of their relevant tests.

This has a few maintainability drawbacks, and the handling of the `files` attribute for test cases as either strings or tuples is at best confusing.

Improve on this by instead making the `files` references relative to where the test case file is written. This enables easier maintenance by keeping all test requirements closer together, rather than spread across the repo. As such, the `files` attribute now requires a list of tuples, taking the form `(relative_src, absolute_dest)`. Additionally, fake plugins for tests that need them to artificially test a specific criteria should also be included in the test's subdir now.

Along with this change, move several StageTwo tests to their own subdirs that now contain both the test cases and the needed files for mocking. This should be the new design pattern going forward - if a test needs to mock files of any kind, put it in a new subdirectory (and if it doesn't need to mock files, continue to keep it in the relevant directory within the test suite).

---


pmoravec (approved):
For maintainability purposes, does it make sense to have a symlink

`tests/vendor_tests/redhat/rhbz2018033/timeout_test.py -> tests/report_tests/timeout/timeout_test.py`

rather than a hard copy of the same file? (but I doubt we will change the file (much), and also we must be careful not to delete the target, in case we decide so)

TurboTurtle:
It does. Rebased and made that change in the latest push.

mhradile (approved):
Looks good to me.

PR #3084, opened 2022-12-06
[plugins] Count in estimate mode real disk usage

pmoravec:
Currently, --estimate-only mode summarizes sizes of all files as returned by stat. That corresponds to "du --apparent-size" calculation.

Meanwhile, the user is rather concerned about the real disk usage reported by "du" output.

Let update the estimate mode from "du --apparent-size" style to "du".

Resolves: #3084

---


pmoravec:
Some background arguments: https://www.redhat.com/sysadmin/du-vs-df

jcastill (approved):

sos/report/__init__.py
```diff
@@ -1297,16 +1297,21 @@ def _collect_plugin(self, plugin):
...
         if self.opts.estimate_only:
-            from pathlib import Path
-            tmpdir_path = Path(self.archive.get_tmp_dir())
-            self.estimated_plugsizes[plugin[1]] = sum(
-                    [f.lstat().st_size for f in tmpdir_path.glob('**/*')])
+            # call "du -s -B1" for the tmp dir to get the disk usage of the
+            # data collected by the plugin - if the command fails, count with 0
+            tmpdir = self.archive.get_tmp_dir()
+            try:
+                du = _plug.exec_cmd('du -sB1 %s' % tmpdir)
```
TurboTurtle, line 1304: Rather than running this in a plugin, I think we should just directly use `sos_get_command_output()` here.

TurboTurtle:
The foreman tests are failing due to `Warning: failed loading '/etc/yum.repos.d/puppet6-stable.repo', skipping.` which means we can't install puppet to setup the envs.

It's unrelated to this change, or even to sos, but we'll want to look into that. At first glance this makes me think that a foreman update broke the repo file in some way.

TurboTurtle (approved):

PR #3087, opened 2022-12-12
[sos collector] - Add SaltStack Transport 

### LINKED ISSUE #3085: [sos-collect] Adding a new transport
I'm attempting to add a transport for saltstack master. The transport connects to nodes but an issue occurs during the sosnodes.py `collect()` function. Comparing to **control_persist** transport with `-vvv` I can see both have identical behavior parsing and setting the hostname. Both transports succeed to `read_file(/etc/os-release)` with identical content and during `collect()` both have 4 jobs in (when I set 3 hosts w/ `--nodes`) . 

However at that point control_persist `collect()` have 4 jobs in `self.opts.jobs`. However `self.client_list` for transport saltstack always shows 0 or 1 depending on the `--no-local` flag, where transport control_persist always has the correct number in `self.client_list`.

All nodes connect and gather a hostname and read the `/etc/os-release` file. The `self.client_list` ends up with only the localhost or nobody in this transport as I can see the `connected()` property is True either via log_debug or even simply testing the class manually in python shell.

The `node` passed to `SosNode` is a tuple, and contains the node name and **None** for password for both transports. I've tracked it to function `determine_host_policy()` in file `sosnodef.py`. I'm at a loss to understand why creating the host object is failing

```
+       self.log_debug("Policy not local, loading host object")
        host = load(cache={}, sysroot=self.opts.sysroot, init=InitSystem(),
                    probe_runtime=True,
                    remote_exec=self._transport.remote_exec,
                    remote_check=self.read_file('/etc/os-release'))
+       self.log_debug("The host object successfully loaded")
```

1. The `log_debug("Policy is not local, loading host object")` appears
2. A few other messages but then then a successful `read_file('/etc/os-release')` message
3. The final `log_debug("The host object successfully loaded")`  never appears on the console. It goes directly back to `collect()` and reports back that the client_list is only the localhost.

```
[sos_collector:_connect_to_node] Attempting to connect to node: store-0
[store-0:_connect] Transport is locally supported and service running. 
[store-0:run_command] Running command hostname
[EL8_store-0:_get_hostname] Hostname set to EL8_store-0
[store-0:determine_host_policy] Attempting to determine host policy   <-- start of function
[store-0:determine_host_policy] Policy not local, loading host object     
[store-0:read_file] Reading file /etc/os-release
[EL8_store-0:read_file] Reading file /etc/os-release
[EL8_store-0:run_command] Running command cat /etc/os-release
[EL8_store-0:_read_file] success reading - fname: /etc/os-release            <-- when the status == 0
[sos_collector:collect] Number of nodes in client_list: 1                              <-- drops out of determine
[sos_collector:collect] Number of nodes to collect from: 1
Collection would only gather from localhost due to failure to either enumerate or connect to cluster nodes. Assuming single collection from localhost is not desired.
Aborting...
```

I then decided to tweak `load()` from file `policies/__init__.py` slightly to try and examine what happens when remote_exec is defined. 

```
def load(cache={}, sysroot=None, init=None, probe_runtime=True,
         remote_exec=None, remote_check=''):
    if 'policy' in cache:
        return cache.get('policy')
    print("remote_exec: %s" % remote_exec)
    import sos.policies.distros
    helper = ImporterHelper(sos.policies.distros)
    for module in helper.get_modules():
        for policy in import_policy(module):
            if policy.check(remote=remote_check):
                cache['policy'] = policy(sysroot=sysroot, init=init,
                                         probe_runtime=probe_runtime,
                                         remote_exec=remote_exec)
            else:
                if remote_exec is not None:
                    print("Policy %s does not match this remote system" %
                          policy)
                    print("Policy check returned: %s" % policy.check())
```

What is odd is non of the listed policies match the RockyLinux policy that the localhost immediately determines when included:

```
remote_exec: salt --out json --static store-0 cmd.shell 
Policy <class 'sos.policies.distros.amazon.AmazonPolicy'> does not match this remote system
Policy check returned: False
Policy <class 'sos.policies.distros.cos.CosPolicy'> does not match this remote system
Policy check returned: False
Policy <class 'sos.policies.distros.debian.DebianPolicy'> does not match this remote system
Policy check returned: False
Policy <class 'sos.policies.distros.openeuler.OpenEulerPolicy'> does not match this remote system
Policy check returned: False
Policy <class 'sos.policies.distros.redhat.CentOsAtomicPolicy'> does not match this remote system
Policy check returned: False
Policy <class 'sos.policies.distros.redhat.CentOsPolicy'> does not match this remote system
Policy check returned: False
Policy <class 'sos.policies.distros.redhat.FedoraPolicy'> does not match this remote system
Policy check returned: False
Policy <class 'sos.policies.distros.redhat.RHELPolicy'> does not match this remote system
Policy check returned: False
Policy <class 'sos.policies.distros.redhat.RedHatAtomicPolicy'> does not match this remote system
Policy check returned: False
Policy <class 'sos.policies.distros.redhat.RedHatCoreOSPolicy'> does not match this remote system
Policy check returned: False
Policy <class 'sos.policies.distros.redhat.RedHatPolicy'> does not match this remote system
Policy check returned: False
```

Clearly when `control_persist` is used however it does find a RockyLinux policy:

```
[store-0:_connect] Opening SSH session to create control socket
[store-0:_connect] Successfully created control socket at /var/tmp/sos.b1rp19fm/.sos-collector-store-0
[store-0:run_command] Running command hostname
[EL8_store-0:_get_hostname] Hostname set to EL8_store-0
[store-0:determine_host_policy] Attempting to determine host policy
[store-0:determine_host_policy] Policy not local, loading host object
[store-0:read_file] Reading file /etc/os-release
[EL8_store-0:read_file] Reading file /etc/os-release
[EL8_store-0:run_command] Running command cat /etc/os-release
[EL8_store-0:_read_file] read success for fname: /etc/os-release
[store-0:determine_host_policy] The host object contains: <sos.policies.distros.rocky.RockyPolicy object at 0x7f69690a9780>
[store-0:determine_host_policy] loaded policy Rocky Linux for host
[EL8_store-0:_load_sos_info] sos version is 4.2.0-20
```

Thanks in advance for any suggestions or pointers which might help.

TrevorBenson:
Closes #3085 

Signed-off-by: Trevor Benson <trevor.benson@gmail.com>

---

### DIFF
--- a/sos/collector/transports/saltstack.py
+++ b/sos/collector/transports/saltstack.py
@@ -1,0 +1,136 @@
+# Copyright Red Hat 2022, Trevor Benson <trevor.benson@gmail.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+import contextlib
+import json
+import os
+import shutil
+from sos.collector.transports import RemoteTransport
+from sos.collector.exceptions import (ConnectionException,
+                                      SaltStackMasterUnsupportedException)
+from sos.utilities import (is_executable,
+                           sos_get_command_output)
+
+
+class SaltStackMaster(RemoteTransport):
+    """
+    A transport for collect that leverages SaltStack's Master Pub/Sub
+    functionality to send commands to minions.
+
+    This transport will by default assume the use cmd.shell module to
+    execute commands on the minions.
+    """
+
+    name = 'saltstack'
+
+    def _convert_output_json(self, json_output):
+        return list(json.loads(json_output).values())[0]
+
+    def run_command(
+            self, cmd, timeout=180, need_root=False, env=None, get_pty=False):
+        """
+        Run a command on the remote host using SaltStack Master.
+        If the output is json, convert it to a string.
+        """
+        ret = super(SaltStackMaster, self).run_command(
+            cmd, timeout, need_root, env, get_pty)
+        with contextlib.suppress(Exception):
+            ret['output'] = self._convert_output_json(ret['output'])
+        return ret
+
+    def _salt_retrieve_file(self, node, fname, dest):
+        """
+        Execute cp.push on the remote host using SaltStack Master
+        """
+        cmd = f"salt {node} cp.push {fname}"
+        res = sos_get_command_output(cmd)
+        if res['status'] == 0:
+            cachedir = f"/var/cache/salt/master/minions/{self.address}/files"
+            cachedir_file = os.path.join(cachedir, fname.lstrip('/'))
+            shutil.move(cachedir_file, dest)
+            return True
+        return False
+
+    @property
+    def connected(self):
+        """Check if the remote host is responding using SaltStack Master."""
+        up = self.run_command("echo Connected", timeout=10)
+        return up['status'] == 0
+
+    def _check_for_saltstack(self, password=None):
+        """Checks to see if the local system supported SaltStack Master.
+
+        This check relies on feedback from the salt binary. The command being
+        run should always generate stderr output, but depending on what that
+        output reads we can determine if SaltStack Master is supported or not.
+
+        For our purposes, a host that does not support SaltStack Master is not
+        able to run sos-collector.
+
+        Returns
+            True if SaltStack Master is supported, else raise Exception
+        """
+
+        cmd = 'salt-run manage.status'
+        res = sos_get_command_output(cmd)
+        if res['status'] == 0:
+            return res['status'] == 0
+        else:
+            raise SaltStackMasterUnsupportedException
+
+    def _connect(self, password=None):
+        """Connect to the remote host using SaltStack Master.
+
+        This method will attempt to connect to the remote host using SaltStack
+        Master. If the connection fails, an exception will be raised.
+
+        If the connection is successful, the connection will be stored in the
+        self._connection attribute.
+        """
+        if not is_executable('salt'):
+            self.log_error("salt command is not executable. ")
+            return False
+
+        try:
+            self._check_for_saltstack()
+        except ConnectionException:
+            self.log_error("Transport is not locally supported. ")
+            raise
+        self.log_info("Transport is locally supported and service running. ")
+        cmd = "echo Connected"
+        result = self.run_command(cmd, timeout=180)
+        return result['status'] == 0
+
+    def _disconnect(self):
+        return True
+
+    @property
+    def remote_exec(self):
+        """The remote execution command to use for this transport."""
+        salt_args = "--out json --static --no-color"
+        return f"salt {salt_args} {self.address} cmd.shell "
+
+    def _retrieve_file(self, fname, dest):
+        """Retrieve a file from the remote host using saltstack
+
+        Parameters
+            fname       The path to the file on the remote host
+            dest        The path to the destination directory on the master
+
+        Returns
+            True if the file was retrieved, else False
+        """
+        return (
+            self._salt_retrieve_file(self.address, fname, dest)
+            if self.connected
+            else False
+        )
+
+# vim: set et ts=4 sw=4 :
--- a/sos/collector/exceptions.py
+++ b/sos/collector/exceptions.py
@@ -104,6 +104,14 @@ class InvalidTransportException(Exception):
         super(InvalidTransportException, self).__init__(message)
 
 
+class SaltStackMasterUnsupportedException(Exception):
+    """Raised when SaltStack Master is unsupported locally"""
+
+    def __init__(self):
+        message = 'Master unsupported by local SaltStack installation'
+        super(SaltStackMasterUnsupportedException, self).__init__(message)
+
+
 __all__ = [
     'AuthPermissionDeniedException',
     'CommandTimeoutException',
@@ -113,6 +121,7 @@ __all__ = [
     'ControlSocketMissingException',
     'InvalidPasswordException',
     'PasswordRequestException',
+    'SaltStackMasterUnsupportedException',
     'TimeoutPasswordAuthException',
     'UnsupportedHostException',
     'InvalidTransportException'
--- a/sos/collector/sosnode.py
+++ b/sos/collector/sosnode.py
@@ -21,6 +21,7 @@ from sos.policies.init_systems import InitSystem
 from sos.collector.transports.control_persist import SSHControlPersist
 from sos.collector.transports.local import LocalTransport
 from sos.collector.transports.oc import OCTransport
+from sos.collector.transports.saltstack import SaltStackMaster
 from sos.collector.exceptions import (CommandTimeoutException,
                                       ConnectionException,
                                       UnsupportedHostException,
@@ -29,7 +30,8 @@ from sos.collector.exceptions import (CommandTimeoutException,
 TRANSPORTS = {
     'local': LocalTransport,
     'control_persist': SSHControlPersist,
-    'oc': OCTransport
+    'oc': OCTransport,
+    'saltstack': SaltStackMaster
 }
 
 


sos/collector/transports/saltstack.py
```diff
@@ -0,0 +1,175 @@
...
+
+    name = 'saltstack'
+
+    def run_salt(self, cmd, **kwargs):
+        """Format and run a command with `salt` on the remote host.
+        """
+        return sos_get_command_output_json(
+            "salt --out json --static %s %s" % (self.address, cmd),
+            **kwargs
+        )
```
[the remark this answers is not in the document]
TrevorBenson: Agreed. Intention was to place this into the `salstack.py` transport. Added below original code only for initial tests. I left it there ATM so I could get the WIP PR opened without delay, instead of moving it to the transport and another round of local tests.

> That does in turn raise the question of why this transport uses --out json, just to convert that json output back to a string before anything else is done to it?

The saltstack transports (salt master) output does not match the **control_persist** transport (ssh).

Without `--out json`  (ie.`salt store-0 cmd.shell 'hostname'`) returns:
```
store-0:
    EL8_store-0
```

- The first line is identical to the `self.address + ':'`.
- The second line contains the hostname. 

This felt like a poor match comparing to **control_persist** transport. Including `--out json` produces the json string. After conversion it provides identical output to **control_persist**. 

With `--out json` (ie. `salt --out json --static store-0 cmd.shell 'hostname'`) returns:

```
{
    "store-0": "EL8_store-0"
}
```

And of course `salt store-0 cmd.shell 'cat /etc/os-release'` returns:

```
store-0:
    NAME="Rocky Linux"
    VERSION="8.5 (Green Obsidian)"
    ID="rocky"
    ID_LIKE="rhel centos fedora"
    VERSION_ID="8.5"
    PLATFORM_ID="platform:el8"
    PRETTY_NAME="Rocky Linux 8.5 (Green Obsidian)"
    ANSI_COLOR="0;32"
    CPE_NAME="cpe:/o:rocky:rocky:8:GA"
    HOME_URL="https://rockylinux.org/"
    BUG_REPORT_URL="https://bugs.rockylinux.org/"
    ROCKY_SUPPORT_PRODUCT="Rocky Linux"
    ROCKY_SUPPORT_PRODUCT_VERSION="8"
```

To eliminate differences in data returned as the root cause for transport not working I created the `sos_get_command_output_json()` function. I am open to doing this another way as this was the first time I've looked at sos collect code so I may have gone a complex or unnecessary route.

sos/collector/transports/saltstack.py
```diff
@@ -0,0 +1,175 @@
...
+
+        if not self._hostname:
+            self._hostname = self.address
+        self.log_info("Hostname set to %s" % self._hostname)
+        return self._hostname
+
+    def _read_file(self, fname):
+        res = self.run_command("cat %s" % fname, timeout=10)
+        output = list(json.loads(res['output']).values())[0]
+        output += output + '\n'
```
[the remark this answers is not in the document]
TrevorBenson: Good catch, not root cause. This was added this at the very last minute last night before calling it quits. For some reason **control_persist** transport always shows a blank line in `/etc/os-release` even when the file has no blank line. The **saltstack** transport always shows the file exactly, no extra blank line. This was a hack to just try to confirm the blank line at the end didn't fix finding the RockyLinux policy from the data returned in `os-release`.

I'll remove this shortly since an extra blank line did not change the behavior.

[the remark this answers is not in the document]
TrevorBenson: By "shows" I mean if I just added `print(output)` to any `_read_file()` to examine the data obtained by the transport.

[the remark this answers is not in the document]
TrevorBenson: > If we can drop --out json then it looks like we should also be able to use the base _read_file() as well.

I didn't look deep enough into collect & transport code to confirm all possible regex patterns and if salt modifying the data would cause failures. I will gladly use the default functions if you think this is an unlikely issue.

sos/collector/transports/saltstack.py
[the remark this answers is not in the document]
TurboTurtle: On this note, I'd recommend doing something like this in the saltstack transport, and dropping `sos_get_command_output_json()` from utilities:

```python

def _convert_output(self, json_output):
    return list(json.loads(json_output).values())[0]


def run_command(self, cmd, timeout=180, need_root=False, env=None,
                    get_pty=False):
    ret = super(SaltStackMasterTransport, self).run_command(cmd, timeout, need_root, env, get_pty)
    try:
        ret['output'] = self._convert_output(ret['output'])
    except Exception:
        # not json somehow?
        pass
    return ret
```

You can then drop `run_salt()` and leverage `run_command()` for everything (aside from the check against different salt binaries for support of the transport). This gives the benefit of always ensuring that the timeouts are handled along with sudo/root needs, pty obtaining, etc... it also makes debugging easier to trace through a single flow (ultimately thru `pexpect.spawn()` instead of `subprocess.Popen()`.)

Having both `run_salt()` and `run_command()` is not in and of itself a problem - `OCTransport` does something similar. However, that is so that there is some controlled difference between commands being executed on remote nodes (in pods) vs locally executed commands for various transport functions.

Once `run_command()` has this in place, we should be able to drop the `_read_file()` definition and use the base method that `control_persist` also uses. If we still have problems at that point, we at least know the issue is in the presentation of the output, not obtaining it.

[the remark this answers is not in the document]
TrevorBenson: Implemented this change. Removed all other changes except the required import statement and adding the transport in the **TRANSPORTS** dictionary of `sosnode.py` file.

TrevorBenson:
To be sure the test box didn't have anything remaining I did `yum reinstall sos` and then restored the current `saltstack.py` and `sosnodes.py`, and then ran another `-vvv` for good measure:

```
[sos_collector:configure_sos_cmd] Initial sos cmd set to sosreport --batch -c auto
[sos_collector:prep] password not specified, assuming SSH keys
sos-collector ASSUMES that SSH keys are installed on all nodes unless the
--password option is provided.

[localhost:run_command] Running command hostname
[EL8_supervisor:_get_hostname] Hostname set to EL8_supervisor
[localhost:determine_host_policy] Attempting to determine host policy
[localhost:determine_host_policy] Using local policy Rocky Linux
[localhost:determine_host_policy] using local policy Rocky Linux
[EL8_supervisor:_load_sos_info] sos version is 4.3.0-5
[EL8_supervisor:run_command] Running command sos report -l
[EL8_supervisor:run_command] Running command sos report --list-presets
[EL8_supervisor:run_command] Running command oc whoami
[EL8_supervisor:_run_command_with_pexpect] The command was not found or was not executable: oc.
Cluster type could not be determined, but --nodes is provided. Attempting to continue using JBON cluster type and the node list
[sos_collector:get_nodes_from_cluster] Node list: []
[sos_collector:get_nodes] Force adding store-0 to node list
[sos_collector:reduce_node_list] Node list reduced to ['store-0']

	EL8_supervisor
	store-0                  

[archive:TarFileArchive] initialised empty FileCacheArchive at '/var/tmp/sos.c015c2vo/sos-collector-2022-12-12-sbqiv'
[archive:TarFileArchive] created directory at 'sos_logs' in FileCacheArchive '/var/tmp/sos.c015c2vo/sos-collector-2022-12-12-sbqiv'

Connecting to nodes...
[store-0:_connect] Transport is locally supported and service running. 
[store-0:run_command] Running command test.ping
[store-0:run_command] Running command hostname
[EL8_store-0:_get_hostname] Hostname set to EL8_store-0
[store-0:determine_host_policy] Attempting to determine host policy
[store-0:determine_host_policy] Policy not local, loading host object
[store-0:read_file] Reading file /etc/os-release
[EL8_store-0:read_file] Reading file /etc/os-release
[EL8_store-0:run_command] Running command cat /etc/os-release
Collection would only gather from localhost due to failure to either enumerate or connect to cluster nodes. Assuming single collection from localhost is not desired.
Aborting...
[sos_collector:close_all_connections] Closing connection to localhost
[EL8_supervisor:disconnect] Successfully disconnected from node
```

TurboTurtle:
Ok, so we're definitely suppressing an exception somewhere.

From the log messages, we enter `_read_file()` and call `run_command()` just fine, but from there we die and we're not hitting any of the catches around that flow. So we have to be in `_run_command_with_pexpect()` somewhere.

Let's start probing there - does the command get formatted correctly via `_format_cmd_for_exec()` which relies on the transport's `remote_exec` property? Does `pexpect.spawn()` get called successfully? Do we get a return from `result.expect()` (line 289)? Are we getting a response that is not accounted for?

TurboTurtle:
I do apologize for not being able to dig deeper on this directly, however I don't have a salt setup available to me that I could test this all myself with.

TrevorBenson:
> Ok, so we're definitely suppressing an exception somewhere.
> 
> From the log messages, we enter `_read_file()` and call `run_command()` just fine, but from there we die and we're not hitting any of the catches around that flow. So we have to be in `_run_command_with_pexpect()` somewhere.
> 
> Let's start probing there - does the command get formatted correctly via `_format_cmd_for_exec()` which relies on the transport's `remote_exec` property? Does `pexpect.spawn()` get called successfully? Do we get a return from `result.expect()` (line 289)? Are we getting a response that is not accounted for?

I'll start adding some log_debug for testing and try to track down the answers.

TrevorBenson:
> I do apologize for not being able to dig deeper on this directly, however I don't have a salt setup available to me that I could test this all myself with.

No apologies necessary, I have an EC2 lab solely just to test this out. The fast response and improvement feedback has been exceptional!

TrevorBenson:
@TurboTurtle I'm back at the `load()` function in `policies/__init__.py` again. 

```
+               print("loading policy into cache['policy']")
                cache['policy'] = policy(sysroot=sysroot, init=init,
                                         probe_runtime=probe_runtime,
                                         remote_exec=remote_exec)
+               print("loaded policy into cache['policy']")
```

The above code sample when using `control_persist` prints both **loading policy** and **loaded policy** messages. 

```
policy: <class 'sos.policies.distros.rocky.RockyPolicy'>
policy.check(remote=remote_check) returned True
loading policy into cache['policy']
loaded policy into cache['policy']
... output omitted ....
Checking sys.platform
- sys.platform == linux
[store-0:determine_host_policy] The host object returned from load()
[store-0:determine_host_policy] The host object contains: <class 'sos.policies.distros.rocky.RockyPolicy'>
[store-0:determine_host_policy] The host object distro: Rocky Linux
```

Using saltstack transport the **loaded policy** message never prints.

```
policy: <class 'sos.policies.distros.rocky.RockyPolicy'>
policy.check(remote=remote_check) returned True
loading policy into cache['policy']
No nodes connected. Aborting...
```

 The only obvious difference above this is the string vs. json string returned from `_run_command_with_pexpect`. This can be reviewed in the below gists if helpful.

[Gist of saltstack transport debug](https://gist.github.com/TrevorBenson/3ead359401d4628a5f1d85896ad3faee)
[Gist of control_persist transport debug](https://gist.github.com/TrevorBenson/ff8e891d16ab2df1267407681d071dd3)

Thanks

TurboTurtle:
Ok, so something is amiss in our error handling with `load()` as it is called within the initializer for `SosNode()`. It is likely getting suppressed by this - https://github.com/sosreport/sos/blob/main/sos/collector/__init__.py#L1117-#L1119

So, evidently, _not_ all errors are logged within SoSNode. For the immediate issue, replace the `pass` there with actually reporting the exception for your tests. This was likely written this way to avoid double logging, but `SosNode()` has changed a lot since the original implementation so we may be able to just keep that debugging in place.

TrevorBenson:
I changed that block of code to:

```
        except Exception as e:
            self.log_debug("Error connecting to %s: %s" % (node[0], e))
            trace = traceback.format_exc()
            self.log_debug(trace)
```

which resulted in:

```
loading policy into cache['policy']
[sos_collector:_connect_to_node] Error connecting to store-0: local variable 'name' referenced before assignment
[sos_collector:_connect_to_node] Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/sos/collector/__init__.py", line 1109, in _connect_to_node
    client = SosNode(node[0], self.commons, password=node[1])
  File "/usr/lib/python3.6/site-packages/sos/collector/sosnode.py", line 86, in __init__
    self.host = self.determine_host_policy()
  File "/usr/lib/python3.6/site-packages/sos/collector/sosnode.py", line 391, in determine_host_policy
    remote_check=self.read_file('/etc/os-release'))
  File "/usr/lib/python3.6/site-packages/sos/policies/__init__.py", line 52, in load
    remote_exec=remote_exec)
  File "/usr/lib/python3.6/site-packages/sos/policies/distros/rocky.py", line 27, in __init__
    remote_exec=remote_exec)
  File "/usr/lib/python3.6/site-packages/sos/policies/distros/redhat.py", line 65, in __init__
    self.pkgs = self.package_manager.all_pkgs()
  File "/usr/lib/python3.6/site-packages/sos/policies/package_managers/__init__.py", line 174, in all_pkgs
    self.packages = self.get_pkg_list()
  File "/usr/lib/python3.6/site-packages/sos/policies/package_managers/__init__.py", line 144, in get_pkg_list
    'name': name,
UnboundLocalError: local variable 'name' referenced before assignment

No nodes connected. Aborting...
```

TurboTurtle:
>   File "/usr/lib/python3.6/site-packages/sos/policies/distros/redhat.py", line 65, in __init__
>     self.pkgs = self.package_manager.all_pkgs()
>   File "/usr/lib/python3.6/site-packages/sos/policies/package_managers/__init__.py", line 174, in all_pkgs
>     self.packages = self.get_pkg_list()
>   File "/usr/lib/python3.6/site-packages/sos/policies/package_managers/__init__.py", line 144, in get_pkg_list
>     'name': name,
> UnboundLocalError: local variable 'name' referenced before assignment
> 

This flow includes methods that are no longer part of the current `main` branch of sos. Off the top of my head, I think this goes back to a time before `package_manager` abstractions were (fully) integrated with remote execution.  I should have realized this earlier, but I assume you're making these changes on a packaged downstream version? Try cloning the repo and running straight from there with your changes.

TrevorBenson:
You are correct I was overlapping versions between upstream and the distro. I presumed when I saw that sos-4.3-5.el8 contained the `--transport` option I could drop the new transport in for testing.

I now 

1. Set aside the 4.3-5.el8 site-packages sos folder.
2. Added the sos directory from the upstream repository feature/transport-saltstack-master branch.
3. Added traceback module to `sos/collector/__init__.py` and just the traceback on the same exception.

The line numbers for each layer of the traceback are different, but the traceback is otherwise identical between upstream and 4.3-5.el8 when it comes to which file, function and bit of code the traceback refers to.

**New traceback**
```
[sos_collector:_connect_to_node] Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/sos/collector/__init__.py", line 1109, in _connect_to_node
    client = SosNode(node[0], self.commons, password=node[1])
  File "/usr/lib/python3.6/site-packages/sos/collector/sosnode.py", line 86, in __init__
    self.host = self.determine_host_policy()
  File "/usr/lib/python3.6/site-packages/sos/collector/sosnode.py", line 388, in determine_host_policy
    remote_check=self.read_file('/etc/os-release'))
  File "/usr/lib/python3.6/site-packages/sos/policies/__init__.py", line 42, in load
    remote_exec=remote_exec)
  File "/usr/lib/python3.6/site-packages/sos/policies/distros/rocky.py", line 27, in __init__
    remote_exec=remote_exec)
  File "/usr/lib/python3.6/site-packages/sos/policies/distros/redhat.py", line 65, in __init__
    self.pkgs = self.package_manager.packages
  File "/usr/lib/python3.6/site-packages/sos/policies/package_managers/__init__.py", line 83, in packages
    self._generate_pkg_list()
  File "/usr/lib/python3.6/site-packages/sos/policies/package_managers/__init__.py", line 151, in _generate_pkg_list
    'name': name,
UnboundLocalError: local variable 'name' referenced before assignment

```

When time permits today I'll take a look at package_managers and distros and see if I can isolate what is happening with **name**. :crossed_fingers: I can submit a patch along with a new transport type.

TurboTurtle:
It's this - https://github.com/sosreport/sos/blob/main/sos/policies/package_managers/__init__.py#L138

When using the transport, we'd get json formatted output back directly here, which would throw off the slicing we do on the returned output. If the output was returned similarly to how you described `salt` functions without `--out json`, I'm reasonably sure at a glance that this would work as expected.

TrevorBenson:
> It's this - https://github.com/sosreport/sos/blob/main/sos/policies/package_managers/__init__.py#L138
> 
> When using the transport, we'd get json formatted output back directly here, which would throw off the slicing we do on the returned output. If the output was returned similarly to how you described `salt` functions without `--out json`, I'm reasonably sure at a glance that this would work as expected.

I can test this out, but will pose 2 questions which is partly why I started trying to use json and convert it to try and create an identical output for both transports:

1. What happens when hostname call expects a single line of text but the return is two lines, the final being the hostname?

   ```
   minion-id222:
     server01
   ```

2. What happens if the minion name overlaps with the policy distros? 

   ```
   CentOS:
     NAME="Rocky Linux"
     VERSION="8.5 (Green Obsidian)"
     ID="rocky"
     ID_LIKE="rhel centos fedora"
     VERSION_ID="8.5"
     PLATFORM_ID="platform:el8"
     PRETTY_NAME="Rocky Linux 8.5 (Green Obsidian)"
     ANSI_COLOR="0;32"
     CPE_NAME="cpe:/o:rocky:rocky:8:GA"
     HOME_URL="https://rockylinux.org/"
     BUG_REPORT_URL="https://bugs.rockylinux.org/"
     ROCKY_SUPPORT_PRODUCT="Rocky Linux"
     ROCKY_SUPPORT_PRODUCT_VERSION="8"
   ```
OR:

- Linux
- Unknown
- CentOS
- Debian
- Fedora
- SuSe
- Ubuntu

While it is uncommon to see a minion_id in saltstack which contains whitespace, it is not prevented. So in theory even the remaining distro identities could be identified from a minion_id at the top of `/etc/os-release` files:

- Amazon Linux
- Anolis OS
- Red Hat
- Red Hat Atomic Host
- Red Hat CoreOS
- Rocky Linux
- UnionTech OS SErver

etc. etc.

TrevorBenson:
If the problem above is not only theoretical this code chunk might prove useful to normalize the the default `highstate` output (vs. `json`).

I'll try out dropping `json` syntax first and keep this in my back pocket for later.


```
"\n".join([x.replace(' ','', 4) for x in output.split("\n",1)[1].split("\n")])
```

TrevorBenson:
Some issues as I suspected from above. I'm working out some changes which I will submit into the branch for discussion on if it is the proper way to handle it.

Thanks again for all the support along the way.

TurboTurtle:
About the only way I can think of that would be somewhat reliable and extensible for the next transport that has some similar additions/formatting changes to output (thinking `juju` maybe? @slashdd) would be to alter `remote_exec` from a string that we use to craft the command run locally, to a method reference that we execute via, so that the transport takes care of formatting and all that as well.

What it wouldn't be, necessarily, is altogether clean. But perhaps that's the tradeoff we need to make here.

TrevorBenson:
Looks like it is definitely related to the returned data and `remote_exec` string.

I modified the transports `run_command()` to have:

```
    def run_command(
            self, cmd, timeout=180, need_root=False, env=None, get_pty=False):
        """
        Run a command on the remote host using SaltStack Master.
        If the output is json, convert it to a string.
        """
        ret = super(SaltStackMasterTransport, self).run_command(
            cmd, timeout, need_root, env, get_pty)
        with contextlib.suppress(Exception):
            ret['output'] = self._convert_output_highstate(ret['output'])
        return ret

    def _convert_output_highstate(self, highstate_output):
        return "\n".join([x.replace(' ','', 4) for x in highstate_output.split("\n",1)[1].split("\r\n")])
```

This fixed the hostname being set to the entire multiline string from saltstack output, and the `/etc/os-release` file parsing to find the Rocky Linux policy/distro. Anything reviewed at the end of `run_command` appears identical to output returned from **control_persist** transport.

This appears to be the reason why the `pkg` from the distro is not found as a key, because those 4 extra spaces from the saltstack **highstate** output style are being embedded into the key names:

```
[HOSTNAME:pkg_version] Package list: {'    alsa-lib': {'name': '    alsa-lib', 'version': ['1', '2', '6', '1'], 'release': '3.el8'}, '    python3-pip-wheel': {'name': '    python3-pip-wheel', 'version': ['9', '0', '3'], 'release': '20.el8.rocky.0'}, '    libipt': {'name': '    libipt', 'version': ['1', '6', '1'], 'release': '8.el8'}, '    rocky-repos': {'name': '    rocky-repos', 'version': ['8', '5'], 'release': '3.el8'}, '    guile': {'name': '    guile', 'version': ['2', '0', '14'], 'release': '7.el8'},
... output omitted ...
```

With the key being `'    sos': ` and not `'sos': ` causes `pkg_version()` to return None, causing the following check to fail:

[https://github.com/sosreport/sos/blob/main/sos/collector/sosnode.py#L282](https://github.com/sosreport/sos/blob/main/sos/collector/sosnode.py#L282)

TurboTurtle:
Ok, I'm going to start looking at utilizing the transport's actual `run_command()` for policy/package_manager remote executions a bit closer. That will make sure that we have fully consistent ways to execute the remote inspections. If I can put something together this week I'll post it, otherwise if I can't then it'd be the first week of January.

TrevorBenson:
> Ok, I'm going to start looking at utilizing the transport's actual `run_command()` for policy/package_manager remote executions a bit closer. That will make sure that we have fully consistent ways to execute the remote inspections. If I can put something together this week I'll post it, otherwise if I can't then it'd be the first week of January.

If you need anything from me, including a salstack master+minion test bed, please do not hesitate to ask.

TurboTurtle:
I haven't forgotten about this, just making my way through backlog stuff from the holidays. As I've run through this a bit more in my head, I'm less nervous about it so hopefully that translates through to the actual implementation.

TrevorBenson:
Apologies for the delayed response A I took a quick look at your commits on Monday but didnt get back to it until today. 

I applied main to the lab with the Saltstack transport (in whatever state I had left it for testing) and it appears to work perfectly. 

- `sos collect -vvv` shows it continuing through all the echo tests and gathering the tarballs, cleaning them up from the nodes, etc. 
- The manifest file looks correct.
  - The node names align with the minions names saltstack uses.
  - The hostnames are correct host shortnames for the server when minion identities differ.

I'll spend some time reviewing the branch and try to clean it prior to moving the draft to full PR. I'll do a once over on the contribution guidelines again and try to minimize the number of "oopsies" I submit for review :smile: 

Thanks @TurboTurtle !

TrevorBenson:
@TurboTurtle Moved from Draft to ready to review. 

I have not reviewed the `sos/collector/clusters/` for **ocp** and **jbon** clusters yet to wrap my head around what to implement for Saltstack, but will understand if a cluster implementation is preferred to be part of this PR instead of a later one. Otherwise I can add that to my todo list for a later PR, probably after uploads to an S3 bucket.

Thanks

TrevorBenson:
I just pushed a new commit to fix my sign off which had my username, not my First Last.

I verified locally from the prior commit flake8, nosetests, avacado. I noticed some failures which I am unsure if was related to the code in this branch or not as the checks from Git Actions were successful.

```
RESULTS    : PASS 244 | ERROR 0 | FAIL 4 | SKIP 11 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB TIME   : 1742.24 s
```

sos/collector/transports/saltstack.py
```diff
@@ -0,0 +1,173 @@
...
+                self.log_info("File %s copied into dest: %s" % (file, dest))
+                return True
+            else:
+                self.log_debug(
+                    "File copy was not successful for: %s" % cachedir_file
+                )
+                return False
+        else:
+            self.log_debug("Error copying file: %s" % fname)
+            return False
```
TurboTurtle, line 171: Logging is handled in the base `RemoteTransport().retrieve_file()` which calls `_retrieve_file()`, as well as retries. A new transport just needs `_retrieve_file()` to execute the command and return True/False depending on if the retrieval was successful.

TurboTurtle (commented):
Main parts look good overall, some small changes noted as part of file retrieval.

sos/collector/transports/saltstack.py
```diff
@@ -0,0 +1,173 @@
...
+        """
+        Execute cp.push on the remote host using SaltStack Master
+        """
+        command = "salt %s cp.push %s" % (node, file)
+        command = command.split()
+        cmd = Popen(
+            command, stdout=PIPE, stderr=PIPE)
+        out, err = cmd.communicate()
+        err = err.decode('utf-8')
+        return {'status': 0} if out and not err else {'status': 1}
```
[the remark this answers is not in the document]
TrevorBenson: I removed this mistakenly thinking it was part of testing when trying to gather more data around the initial failures. I'll restore that part shortly.

Thanks

sos/collector/transports/saltstack.py
```diff
@@ -0,0 +1,173 @@
...
+                command, stdout=PIPE, stderr=PIPE)
+            out, err = cmd.communicate()
+            err = err.decode('utf-8')
+            if 'command not found' in err or 'Usage:' in err:
+                raise ConnectionException(failure_message)
+
+        salt_cmd = ['salt', '--version']
+        salt_run_cmd = ['salt-run', 'manage.status']
+        _check_for_salt_master(
+            salt_cmd, "The salt binary is not available on this system")
```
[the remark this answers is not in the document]
TrevorBenson: Mostly redundant, at least for `salt`. I'll remove this and rely on **_connect()**'s **is_executable('salt')**

The additional check for `salt-run` confirms the **salt-master** service is responding local to the sos collector. Otherwise the master may not be in a functional state to send commands to minions. Unless you have a preferred placement for this, I'll leave the **_check_for_salt_master()** verifying the service responds.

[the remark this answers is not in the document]
TrevorBenson: Going to leave this checking `salt-run manage.status` for the time being as its possible the service active in terms of systemd, but potentially not functional in which case I think the current check results in a more thorough test.

TrevorBenson:
@TurboTurtle I believe I've addressed each comment.

flake8 has zero output.
nosetests return:
```

----------------------------------------------------------------------
Ran 131 tests in 622.344s

OK
```
Running avocado now, here is an example of one of the failures I observe locally:
```
 (024/259) tests/cleaner_tests/basic_function_tests/report_with_mask.py:ReportWithMask.test_localhost_was_obfuscated: FAIL: /var/tmp/avocado_x1r_56vasos_tests.py.ReportWithMask/sosreport-ReportWithMask/sosreport-host1124-2023-01-18-qhenvll/hostname does not exist (0.03 s)
```

re-requesting review now, thanks.

TrevorBenson:
Results from Avacado:
```
RESULTS    : PASS 235 | ERROR 10 | FAIL 3 | SKIP 11 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB TIME   : 1713.42 s
```

sos/collector/transports/saltstack.py
```diff
@@ -0,0 +1,140 @@
...
+
+import contextlib
+import json
+import os
+from sos.collector.transports import RemoteTransport
+from sos.collector.exceptions import ConnectionException
+from sos.utilities import (is_executable,
+                           sos_get_command_output)
+
+TIMEOUT_DEFAULT = 300
```
TurboTurtle, line 19: This can be dropped now.

sos/collector/transports/saltstack.py
```diff
@@ -0,0 +1,140 @@
...
+        Execute cp.push on the remote host using SaltStack Master
+        """
+        cmd = f"salt {node} cp.push {fname}"
+        res1 = sos_get_command_output(cmd)
+        if res1['status'] == 0:
+            cachedir = f"/var/cache/salt/master/minions/{self.address}/files"
+            cachedir_file = os.path.join(cachedir, fname.lstrip('/'))
+            cmd = f"mv {cachedir_file} {dest}"
+            res2 = sos_get_command_output(cmd)
+        return res2['status'] == 0
```
TurboTurtle, line 61: This needs to be indented into the `if res1['status'] == 0` conditional. If `res1['status']` is not `0`, then we'll hit a `NameError` as `res2` is yet to be defined.

If `res1['status']` is not `0`, then we should explicitly return False. I think `shutil` here is better than directly calling `mv`, also.

Something like this (untested):

```python
cmd = f"salt {node} cp.push {fname}"
res = sos_get_command_output(cmd)
if res['status'] == 0:
    cachedir = f"/var/cache/salt/master/minions/{self.address}/files"
    cachedir_file = os.path.join(cachedir, fname.lstrip('/'))
    shutil.move(cachedir_file, dest)
    return True
return False
```

TurboTurtle (commented):
Looking good, just a couple changes.

When ready, please squash the fixup commits into the first commit as we don't allow fixup or merge commits into `main`.


> Additional errors seem to be python3-magic, timeouts and file not found. Not sure if this is expected outside of CI.

All those are in turn caused by the timeout in the logs test, which is due to it being a badly designed test ;). We generate random data to test that size limiting is working, but random data crushes compression performance which then causes a timeout.

sos/collector/transports/saltstack.py
TrevorBenson: I'll nest it. Does this comment also mean to suggest usage of **shutil.move()** instead of **sos_get_command_output()** which replaced shutil?

TrevorBenson: Doh, reading the comment in vscode, missed the explicit statement about shutil, fixing.

TrevorBenson:
Single commit and changes applied. However I'm trying to track down in git configs and vscode where my sign-off by Illuminatus (GitHub Name) is set. 

I'll do a force push with another single commit once the second sign off is removed.

sos/collector/transports/saltstack.py
```diff
@@ -0,0 +1,139 @@
...
+        """
+
+        def _check_for_salt_master(command):
+            """Check for various salt master binaries on the local system."""
+            res = sos_get_command_output(command)
+            return res['status'] == 0
+
+        cmd = 'salt-run manage.status'
+        res = _check_for_salt_master(cmd)
+        return res
```
pmoravec, line 89: Until there are plans for possible re-usage of the `_check_for_salt_master` routine, why not calling directly:

```
res = sos_get_command_output('salt-run manage.status')
return res['status'] == 0
```

?

sos/collector/transports/saltstack.py
```diff
@@ -0,0 +1,139 @@
...
+
+        This check relies on feedback from the salt binary. The command being
+        run should always generate stderr output, but depending on what that
+        output reads we can determine if SaltStack Master is supported or not.
+
+        For our purposes, a host that does not support SaltStack Master is not
+        able to run sos-collector.
+
+        Returns
+            True if SaltStack Master is supported, else raise Exception
```
pmoravec, line 79: Is the "else raise Exception" still correct?

(I came late to the review party so I just guess there were changes with return value and with re-usage of the `_check_for_salt_master`?)

pmoravec (approved):
Just very minor comments, ACK to the code as is now.

sos/collector/transports/saltstack.py
TrevorBenson: Now that salt binary is handled by **is_executable()** instead, no reason. I'll move this to **sos_get_command_output()** and leave any alternative logic for a cluster implementation.

sos/collector/transports/saltstack.py
```diff
@@ -0,0 +1,139 @@
...
+        """
+        ret = super(SaltStackMasterTransport, self).run_command(
+            cmd, timeout, need_root, env, get_pty)
+        with contextlib.suppress(Exception):
+            ret['output'] = self._convert_output_json(ret['output'])
+        return ret
+
+    def _salt_retrieve_file(
+            self, node, fname, dest,
+            timeout=180, need_root=False, env=None, get_pty=False):
```
TurboTurtle, line 49: Sorry, missed this earlier. With current use of `sos_get_command_output()`, the parameters after `timeout` are unnecessary and can be removed. `timeout` can be as well, unless we want to use that for the `cp.push` command instead of the default 300 seconds.

After this, everything looks good to me for merge.

sos/collector/transports/saltstack.py
TrevorBenson: I think I was mimicking **oc**, or **ControlPersist**  or ocp cluster logic initially. While cleaning up debug statements and refactoring I overlooked the docstring defined Return. 

I'll review again the other transports this morning and determine what makes the most sense, to raiseerror or return bool False.

TrevorBenson: Confirmed that I was mimicking the logic from the **_check_for_control_persist()** function. 

I will restore the raise Exception, keeping similar behavior when transport is not available.

TrevorBenson: Added a new SaltStackMasterUnsupportedException to exceptions and restored original return logic.

sos/collector/transports/saltstack.py
TrevorBenson: Relying on default of 300 seconds for rare cases where network is slow and log collection has balooned sos report size.

sos/collector/transports/saltstack.py
```diff
@@ -0,0 +1,136 @@
...
+        For our purposes, a host that does not support SaltStack Master is not
+        able to run sos-collector.
+
+        Returns
+            True if SaltStack Master is supported, else raise Exception
+        """
+
+        cmd = 'salt-run manage.status'
+        res = sos_get_command_output(cmd)
+        if res:
```
TurboTurtle, line 83: This should be `if res['status'] == 0`, as the method will always return a dict of `{'status': <int>, 'output': <str>}`

TurboTurtle:
Also, ack for the new exception pending the PEP8 error.

TrevorBenson:
@TurboTurtle I believe your and @pmoravec final comments have been addressed.

TrevorBenson:
> Also, ack for the new exception pending the PEP8 error.

Yep fixed. 

TL;DR
My vscode gives me the "flake8 has crashed more than X times in..." message when starting and I have not trackewd down what I did before xmas break to cause it. I'll track it down or add a manual `flake8 sos` to a local debug step to remind myself to stop pushing since I know automatic linting isn't working ATM.

sos/collector/transports/saltstack.py
TrevorBenson: Yep, I overlooked the swap away from the saltstack check no longer returned bool but dict, fixup and squashing into single commit now.

TurboTurtle (approved):

sos/collector/transports/saltstack.py
```diff
@@ -0,0 +1,136 @@
...
+        able to run sos-collector.
+
+        Returns
+            True if SaltStack Master is supported, else raise Exception
+        """
+
+        cmd = 'salt-run manage.status'
+        res = sos_get_command_output(cmd)
+        if res['status'] == 0:
+            return res['status'] == 0
```
pmoravec, line 84: This can be simplified to "return True", but let have it as is :).

pmoravec (approved):

PR #3089, opened 2022-12-13
[ceph] Add new collect cluster profile for RHCS5

TurboTurtle:
Adds a new cluster profile for RHCS 5 for `sos collect`. This profile depends upon the use of `cephadm` which is used to both deploy and manage the cluster. Users may optionally restrict the list of nodes to collect from by using the `-c ceph.labels` option to specify a set of label(s) to filter node results with.

---

### DIFF
--- a/sos/collector/__init__.py
+++ b/sos/collector/__init__.py
@@ -934,30 +934,34 @@ class SoSCollector(SoSComponent):
 
     def configure_sos_cmd(self):
         """Configures the sosreport command that is run on the nodes"""
-        self.sos_cmd = 'sosreport --batch '
+        sos_cmd = 'sosreport --batch '
 
-        sos_opts = []
+        sos_options = {}
 
         if self.opts.case_id:
-            sos_opts.append('--case-id=%s' % (quote(self.opts.case_id)))
+            sos_options['case-id'] = quote(self.opts.case_id)
         if self.opts.alloptions:
-            sos_opts.append('--alloptions')
+            sos_options['alloptions'] = ''
         if self.opts.all_logs:
-            sos_opts.append('--all-logs')
+            sos_options['all-logs'] = ''
         if self.opts.verify:
-            sos_opts.append('--verify')
+            sos_options['verify'] = ''
         if self.opts.log_size:
-            sos_opts.append(('--log-size=%s' % quote(str(self.opts.log_size))))
+            sos_options['log-size'] = quote(str(self.opts.log_size))
         if self.opts.sysroot:
-            sos_opts.append('-s %s' % quote(self.opts.sysroot))
+            sos_options['sysroot'] = quote(self.opts.sysroot)
         if self.opts.chroot:
-            sos_opts.append('-c %s' % quote(self.opts.chroot))
+            sos_options['chroot'] = quote(self.opts.chroot)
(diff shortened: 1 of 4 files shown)


sos/collector/clusters/ceph.py
```diff
@@ -0,0 +1,66 @@
...
+    such an environment and this profile does not work for you, please file a
+    bug report detailing what is failing.
+
+    By default, all nodes in the cluster will be returned for collection. This
+    may not be desireable, so users are encouraged to use the `labels` option
+    to specify a colon-delimited set of ceph node labels to restrict the list
+    of nodes to.
+
+    For example, using `-c ceph.labels=osd:mgr` will return only nodes labeled
+    with *either* `osd` or `mgr`.
```
pmoravec, line 31: Why colon and not comma (that we use in other multi-value arguments, like `sos report -o qpid,logs,pulp`)?

Or is this syntax more convenient for ceph users or for cluster..?

I am ok with whatever works well / the best for the users (where I lack their preferences).

pmoravec (approved):
Sounds good otherwise.

sos/collector/clusters/ceph.py
TurboTurtle: This is a hold over from the structure of plugin options that was also applied to cluster options. Because plugin options are passed as  _values_ of the `-k`/`--plugin-option` option, we can't have commas in the values for those plugin options. Argparse instead uses the comma delimiter for additional values of `-k`. E.G. `sos report -k foo.bar=123,abc.opt=456` equates to `sos report -k foo.bar=123 -k abc.opt=456`. It's only internally (external to argparse) that we sort out that the string `foo.bar=123` is itself an option that needs further parsing.

We do the same colon delimitation for the OCP cluster for this same reason.

pmoravec: I see, that makes sense. And worth to have the consistency within - among clusters.

TurboTurtle:
Updated to have the profile set a bigger log-size, as RH Ceph support engineers implied the defaults could sometimes be low for them. This in turn required adding that mechanism to the base `Cluster()` class.

sos/collector/__init__.py
```diff
@@ -934,30 +934,33 @@ def display_nodes(self):
...
         if self.opts.compression_type != 'auto':
-            sos_opts.append('-z %s' % (quote(self.opts.compression_type)))
-        self.sos_cmd = self.sos_cmd + ' '.join(sos_opts)
-        self.log_debug("Initial sos cmd set to %s" % self.sos_cmd)
-        self.commons['sos_cmd'] = self.sos_cmd
-        self.collect_md.add_field('initial_sos_cmd', self.sos_cmd)
+            sos_options['compression-type'] = quote(self.opts.compression_type)
+
+        for k, v in sos_options.items():
+            sos_cmd += f"--{k} {v} "
```
pmoravec, line 959: Just a nitpick that can be ignored: this would leave space at the end of the command - removing the trailing space in `        sos_cmd = 'sosreport --batch '` and moving the trailing space here to the leading one would fix it. No issue at the end of the day, can be ignored.

pmoravec (approved):

pmoravec (approved):

PR #3092, opened 2022-12-16
[ceph] Update Ceph plugins for newer versions of Ceph

TurboTurtle:
This patchset first corrects an overall issue with the enablement of the ceph plugins for newer versions of Ceph, and then delves into updating the individual plugins to ensure collections remain consistent with expectations across those ceph versions.

---

### DIFF
--- a/sos/report/plugins/ceph_common.py
+++ b/sos/report/plugins/ceph_common.py
@@ -17,7 +17,7 @@ class Ceph_Common(Plugin, RedHatPlugin, UbuntuPlugin):
     plugin_name = 'ceph_common'
     profiles = ('storage', 'virt', 'container')
 
-    containers = ('ceph-(mon|rgw|osd).*',)
+    containers = ('ceph-(.*-)?(mon|rgw|osd).*',)
     ceph_hostname = gethostname()
 
     packages = (
--- a/sos/report/plugins/ceph_mds.py
+++ b/sos/report/plugins/ceph_mds.py
@@ -7,17 +7,14 @@
 # See the LICENSE file in the source distribution for further information.
 
 from sos.report.plugins import Plugin, RedHatPlugin, UbuntuPlugin
-import glob
 
 
 class CephMDS(Plugin, RedHatPlugin, UbuntuPlugin):
     short_desc = 'CEPH mds'
     plugin_name = 'ceph_mds'
     profiles = ('storage', 'virt', 'container')
-    containers = ('ceph-fs.*',)
-
-    def check_enabled(self):
-        return True if glob.glob('/var/lib/ceph/mds/*/*') else False
+    containers = ('ceph-(.*-)?fs.*',)
+    files = ('/var/lib/ceph/mds/',)
 
     def setup(self):
         self.add_file_tags({
--- a/sos/report/plugins/ceph_mgr.py
+++ b/sos/report/plugins/ceph_mgr.py
@@ -6,28 +6,67 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
+import os
+
 from sos.report.plugins import Plugin, RedHatPlugin, UbuntuPlugin
 
 
 class CephMGR(Plugin, RedHatPlugin, UbuntuPlugin):
+    """
+    This plugin is for capturing information from Ceph mgr nodes. While the
+    majority of this plugin should be version-agnostic, several collections are
+    dependent upon the version of Ceph installed. Versions that correlate to
+    RHCS 4 or RHCS 5 are explicitly handled for differences such as those
+    pertaining to log locations on the host filesystem.
+
+    Note that while this plugin will activate based on the presence of Ceph
+    containers, commands are run directly on the host as those containers are
+    often not configured to successfully run the `ceph` commands collected by
+    this plugin. These commands are majorily `ceph daemon` commands that will
+    reference discovered admin sockets under /var/run/ceph.
+    """
 
     short_desc = 'CEPH mgr'
 
     plugin_name = 'ceph_mgr'
     profiles = ('storage', 'virt', 'container')
-    files = ('/var/lib/ceph/mgr/',)
-    containers = ('ceph-(.*)?mgr.*',)
+    files = ('/var/lib/ceph/mgr/', '/var/lib/ceph/*/mgr*')
+    containers = ('ceph-(.*-)?mgr.*',)
 
     def setup(self):
+
+        self.ceph_version = self.get_ceph_version()
+
+        logdir = '/var/log/ceph'
+        libdir = '/var/lib/ceph'
+        rundir = '/run/ceph'
+
+        if self.ceph_version >= 16:
+            logdir += '/*'
+            libdir += '/*'
+            rundir += '/*'
+
         self.add_file_tags({
-            '/var/log/ceph/ceph-mgr.*.log': 'ceph_mgr_log',
+            f'{logdir}/ceph-mgr.*.log': 'ceph_mgr_log',
         })
 
+        self.add_forbidden_path([
+            "/etc/ceph/*keyring*",
+            f"{libdir}/*keyring*",
+            f"{libdir}/**/*keyring*",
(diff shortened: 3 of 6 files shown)


sos/report/plugins/ceph_mgr.py
```diff
@@ -6,28 +6,68 @@
...
+        ])
+
         self.add_copy_spec([
-            "/var/log/ceph/ceph-mgr*.log",
-            "/var/lib/ceph/mgr/",
-            "/var/lib/ceph/bootstrap-mgr/",
-            "/run/ceph/ceph-mgr*",
+            f"{logdir}/ceph-mgr*.log",
+            f"{libdir}/mgr/",
+            f"{libdir}/mgr*/",
```
pmoravec, line 68: The `/mgr/` is dedicatedly mentioned due to sizelimit? Since that `copy_spec` would be already contained in `mgr*/` otherwise.

sos/report/plugins/ceph_mgr.py
```diff
@@ -61,44 +101,33 @@ def setup(self):
...
-            "/var/lib/ceph/*/*/*keyring*",
-            "/var/lib/ceph/osd",
-            "/var/lib/ceph/mon",
-            # Excludes temporary ceph-osd mount location like
-            # /var/lib/ceph/tmp/mnt.XXXX from sos collection.
-            "/var/lib/ceph/tmp/*mnt*",
-            "/etc/ceph/*bindpass*",
-        ])
+        self.add_cmd_output([
+            f"ceph daemon {m} {cmd}" for m in self.get_socks() for cmd in cmds]
```
pmoravec, line 105: Not knowing `ceph` internals, the `container=..` removal from this command is intentional, right? (as we check the right sockets the container uses)?

sos/report/plugins/ceph_osd.py
```diff
@@ -6,37 +6,74 @@
...
+        ])
+
         # Only collect OSD specific files
         self.add_copy_spec([
-            "/run/ceph/ceph-osd*",
-            "/var/lib/ceph/osd/*/kv_backend",
-            "/var/log/ceph/ceph-osd*.log",
-            "/var/log/ceph/ceph-volume*.log",
+            f"{rundir}/ceph-osd*",
+            f"{libdir}/**/kv_backend",
```
pmoravec, line 66: The `**` is intentional as the `kv_backend` can be in any subdir, right?

sos/report/plugins/ceph_mon.py
```diff
@@ -6,28 +6,72 @@
...
     plugin_name = 'ceph_mon'
     profiles = ('storage', 'virt', 'container')
-    containers = ('ceph-(.*)?mon.*',)
-    files = ('/var/lib/ceph/mon/',)
+    # note: for RHCS 5 / Ceph v16 the containers serve as an enablement trigger
+    # but by default they are not capable of running various ceph commands in
+    # this plugin - the `ceph` binary is functional directly on the host
+    containers = ('ceph-(.*-)?mon.*',)
+    files = ('/var/lib/ceph/mon/', '/var/lib/ceph/*/mon*')
+    ceph_version = 0
```
pmoravec, line 41: Is this required, as the `get_ceph_version` method returns 0 as a safe fallback? (other plugins' changes dont have this)

sos/report/plugins/ceph_common.py
```diff
@@ -17,7 +17,7 @@ class Ceph_Common(Plugin, RedHatPlugin, UbuntuPlugin):
     plugin_name = 'ceph_common'
     profiles = ('storage', 'virt', 'container')
 
-    containers = ('ceph-(mon|rgw|osd).*',)
+    containers = ('ceph-(.*)?(mon|rgw|osd).*',)
```
pmoravec, line 20: Few other changes in other commits have `(.*-)` instead of just `(.*)`. Is that intentional?

sos/report/plugins/ceph_mds.py
```diff
@@ -7,17 +7,14 @@
...
 
 class CephMDS(Plugin, RedHatPlugin, UbuntuPlugin):
     short_desc = 'CEPH mds'
     plugin_name = 'ceph_mds'
     profiles = ('storage', 'virt', 'container')
-    containers = ('ceph-fs.*',)
-
-    def check_enabled(self):
-        return True if glob.glob('/var/lib/ceph/mds/*/*') else False
+    containers = ('ceph-(.*)?fs.*',)
```
pmoravec, line 16: Ditto `(.*-)`

sos/report/plugins/ceph_mgr.py
```diff
@@ -16,11 +15,8 @@ class CephMGR(Plugin, RedHatPlugin, UbuntuPlugin):
 
     plugin_name = 'ceph_mgr'
     profiles = ('storage', 'virt', 'container')
-
-    containers = ('ceph-mgr.*',)
-
-    def check_enabled(self):
-        return True if glob.glob('/var/lib/ceph/mgr/*/*') else False
+    files = ('/var/lib/ceph/mgr/',)
+    containers = ('ceph-(.*)?mgr.*',)
```
pmoravec: Ditto.

sos/report/plugins/ceph_common.py
nkshirsagar: if ceph container is named differently (not beginning with ceph-) would there be an issue gathering the ceph stuff from such a container?

sos/report/plugins/ceph_mgr.py
TurboTurtle: Nope, that's an oversight.

sos/report/plugins/ceph_mgr.py
TurboTurtle: Correct, containers are used for enablement but these commands run directly on the host.

sos/report/plugins/ceph_osd.py
TurboTurtle: Correct.

sos/report/plugins/ceph_mon.py
TurboTurtle: It squelches a code analysis alert to have this here. Not strictly required, though.

sos/report/plugins/ceph_common.py
TurboTurtle: Practically, with current naming schemes the `(.*)?` is sufficient, but `(.*-)?` is more exact. For consistency sake I'll sync all of these to be `(.*-)?`.

TurboTurtle: > if ceph container is named differently (not beginning with ceph-) would there be an issue gathering the ceph stuff from such a container?

It means we won't activate the plugin based solely on the presence of the container, and would rely on the other enablement triggers. For the `ceph_common` plugin specifically there are no other triggers, so this would result in that plugin not running.

pmoravec (approved):

sos/report/plugins/ceph_osd.py
```diff
@@ -6,40 +6,74 @@
...
+            f"{logdir}/ceph-osd*.log",
+            f"{logdir}/ceph-volume*.log",
         ])
```
nobuto-m, line 69: I've opened https://github.com/sosreport/sos/issues/3100

sos/report/plugins/ceph_mon.py
```diff
@@ -88,63 +135,96 @@ def setup(self):
...
+                    proc = procs.split()
+                    # Locate the '--id' value of the process
+                    if proc and proc[0].endswith("ceph-mon"):
+                        try:
+                            id_index = proc.index("--id")
+                            ceph_ids.append(proc[id_index + 1])
+                        except (IndexError, ValueError):
+                            self._log_warn('Unable to find ceph IDs')
+
+        # ceph version 16 is RHCS 5
```
nkshirsagar, line 183: I think this is very RH specific. Also, what happens for ceph version 15, octopus? The if condition in this function only covers version == 14, or >=16..

PR #3096, opened 2022-12-28
[collector] Prevent appending local host in strict_node_list mode

pmoravec:
The changes in respecting strict_node_list are three-fold:

1) Don't add local hostname among "list of nodes to collect from:"
2) Skip explicit adding of the primary node to client_list
3) Don't reduce_node_list (as it can purge away local host e.g. in case   of IP address node)

Resolves: #3096

---

### DIFF
--- a/sos/collector/__init__.py
+++ b/sos/collector/__init__.py
@@ -914,7 +914,8 @@ class SoSCollector(SoSComponent):
 
         self.ui_log.info('The following is a list of nodes to collect from:')
         if self.primary.connected and self.primary.hostname is not None:
-            if not (self.primary.local and self.opts.no_local):
+            if not ((self.primary.local and self.opts.no_local)
+                    or self.cluster.strict_node_list):
                 self.ui_log.info('\t%-*s' % (self.commons['hostlen'],
                                              self.primary.hostname))
 
@@ -1022,12 +1023,13 @@ class SoSCollector(SoSComponent):
         if applicable"""
         if (self.hostname in self.node_list and self.opts.no_local):
             self.node_list.remove(self.hostname)
-        for i in self.ip_addrs:
-            if i in self.node_list:
-                self.node_list.remove(i)
+        if not self.cluster.strict_node_list:
+            for i in self.ip_addrs:
+                if i in self.node_list:
+                    self.node_list.remove(i)
         # remove the primary node from the list, since we already have
         # an open session to it.
-        if self.primary is not None:
+        if self.primary is not None and not self.cluster.strict_node_list:
             for n in self.node_list:
                 if n == self.primary.hostname or n == self.opts.primary:
                     self.node_list.remove(n)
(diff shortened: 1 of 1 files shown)


pmoravec:
Particular reproducer: running `sos collect` from a `pacemaker` cluster node where nodes are named `node1` to `node3`. 

Then:
- https://github.com/sosreport/sos/blob/main/sos/collector/__init__.py#L1176-L1177 adds collecting `sos report` also from the primary node, which is `hostname -f` but not either `node*`
- https://github.com/sosreport/sos/blob/main/sos/collector/__init__.py#L918 prints this primary hostname among nodes
- on the other hand, when nodes would be provided as IP addresses, https://github.com/sosreport/sos/blob/main/sos/collector/__init__.py#L1025-L1027 would purge the current node from the node list - and basically nothing from the `reduce_host_list` is applicable in case of `strict_node_list`

Anyway, my gut feeling tells me this PR might break some other use case,so please review it carefully.

sos/collector/__init__.py
```diff
@@ -1089,7 +1090,8 @@ def get_nodes(self):
                     self.node_list.remove(node)
             if not self.cluster.strict_node_list:
                 self.node_list.append(self.hostname)
-        self.reduce_node_list()
+        if not self.cluster.strict_node_list:
+            self.reduce_node_list()
```
TurboTurtle, line 1094: I think we need to keep this. It would be an edge case, but if nodes are somehow duplicated, this prevents us from trying to connect and collect from the node twice.

TurboTurtle (commented):
Aside from the comment below related to your point 3, I think this is good to go.

sos/collector/__init__.py
pmoravec: Assume `pacemaker` cluster with nodes declared as IP addresses (for a reproducer, just have `/etc/corosync/corosync.conf` with one `ring0_addr:` pointing to your IP address and another `ring0_addr:` pointing to a different IP address - until you have `crm_mon` installed, then `sos collect --cluster-type pacemaker --batch -o qpid` will attempt to collect sosreport only from the other IP address but not from own).

Then, https://github.com/sosreport/sos/blob/main/sos/collector/__init__.py#L1029-L1031 will remove the IP address of the local host.

When having `ring0_addr: $(hostname -f)`, then https://github.com/sosreport/sos/blob/main/sos/collector/__init__.py#L1034-L1037 removes it from the node list, and the change from point 2. (`if self.primary.connected and not self.cluster.strict_node_list:`) will cause skipping sosreport from the local host.

Also not sure if similar scenario with hostname can't apply to https://github.com/sosreport/sos/blob/main/sos/collector/__init__.py#L1027-L1028 ? (I dont think so but unsure now)

While the rest of the `reduce_node_list` makes sense.

So I will update the PR to:
- call `reduce_node_list` unconditionally again
- within the method:
  -  call `for i in self.ip_addrs:` only when `not self.cluster.strict_node_list:`
  - enhance `if self.primary is not None:` by the same `and not self.cluster.strict_node_list:`

But still I feel there can be a use case where we keep/remove local host wrongly, as the conditions at https://github.com/sosreport/sos/blob/main/sos/collector/__init__.py#L1034 and https://github.com/sosreport/sos/blob/main/sos/collector/__init__.py#L1180 will differ?

TurboTurtle (approved):

Issue #3098, opened 2022-12-30
Missing log files for Ceph iSCSI gateway


nobuto-m:
In Ceph iSCSI gateway usage, there are some additional logs or config files for analyzing issues on top of what sosreport collects as of today for Ceph.
https://docs.ceph.com/en/latest/rbd/iscsi-overview/

Would be nice if the following files are collected for such a use case.

- /etc/ceph/iscsi/ceph.conf <- already collected as a part of ceph_common plugin
- /etc/ceph/iscsi-gateway.cfg <- already collected as a part of ceph_common plugin
- /etc/tcmu/tcmu.conf <- missing
- /var/log/ceph/ceph-client.iscsi.log <- missing
- /var/log/rbd-target-api/rbd-target-api.log <- missing
- /var/log/rbd-target-gw/rbd-target-gw.log <- missing
- /var/log/tcmu-runner.log <- missing

PR #3099, opened 2022-12-30
[ceph] add iSCSI gateway

nobuto-m:
Along with existing Ceph plugins, Ceph iSCSI gateway use case requires additional config and log files to be analyzed.
https://docs.ceph.com/en/latest/rbd/iscsi-overview/

Closes: #3098
Related-Bug: https://bugs.launchpad.net/ubuntu/+source/sosreport/+bug/2000672
Signed-off-by: Nobuto Murata <nobuto.murata@canonical.com>

---

### DIFF
--- a/sos/report/plugins/ceph_iscsi.py
+++ b/sos/report/plugins/ceph_iscsi.py
@@ -1,0 +1,36 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, RedHatPlugin, UbuntuPlugin
+
+
+class CephISCSI(Plugin, RedHatPlugin, UbuntuPlugin):
+
+    short_desc = "CEPH iSCSI"
+
+    plugin_name = "ceph_iscsi"
+    profiles = ("storage", "virt", "container")
+    packages = ("ceph-iscsi",)
+    services = ("rbd-target-api", "rbd-target-gw")
+    containers = ("rbd-target-api.*", "rbd-target-gw.*")
+
+    def setup(self):
+        self.add_copy_spec([
+            "/etc/tcmu/tcmu.conf",
+            "/var/log/**/ceph-client.*.log",
+            "/var/log/**/rbd-target-api.log",
+            "/var/log/**/rbd-target-gw.log",
+            "/var/log/tcmu-runner.log"
+        ])
+
+        self.add_cmd_output([
+            "gwcli info",
+            "gwcli ls"
+        ])
+
+# vim: set et ts=4 sw=4 :


sos/report/plugins/ceph_iscsi.py
```diff
@@ -0,0 +1,50 @@
...
+                "/var/lib/ceph/*keyring*",
+                "/var/lib/ceph/*/*keyring*",
+                "/var/lib/ceph/*/*/*keyring*",
+                "/var/lib/ceph/osd",
+                "/var/lib/ceph/mon",
+                # Excludes temporary ceph-osd mount location like
+                # /var/lib/ceph/tmp/mnt.XXXX from sos collection.
+                "/var/lib/ceph/tmp/*mnt*",
+                "/etc/ceph/*bindpass*",
+            ]
```
pmoravec, line 46: This is redundant here, as no such file is collected by this plugin.

pmoravec:
As not knowing `ceph` and esp. its nuances between Ubuntu and RedHat distros, I am asking for a review some more knowledgable people to check:

- do the `packages = ("ceph-iscsi",)` and `services = ("rbd-target-api", "rbd-target-gw")` enablement condition match on `RedHat` distro?
- logs and commands execution exist there as well?
- does it deserve to have a separate plugin for this - or is there some overlap with other `ceph_*` plugins?

Thanks anybody from @jcastill , @asm0deuz, @nkshirsagar for such review.

sos/report/plugins/ceph_iscsi.py
```diff
@@ -0,0 +1,50 @@
...
+    def setup(self):
+        self.add_copy_spec(
+            [
+                "/etc/tcmu/tcmu.conf",
+                "/var/log/ceph/ceph-client.iscsi.log",
+                "/var/log/rbd-target-api/rbd-target-api.log",
+                "/var/log/rbd-target-gw/rbd-target-gw.log",
+                "/var/log/tcmu-runner.log",
+            ]
+        )
```
TurboTurtle, line 30: Minor nitpick - sos style is to have the list brackets in-line with the parens of the calling method, thus reducing the indention level by one. E.G.:

```python
self.add_copy_spec([
    '/some/path',
    '/some/other/path'
])
```

sos/report/plugins/ceph_iscsi.py
```diff
@@ -0,0 +1,50 @@
...
+    profiles = ("storage", "virt")
+
+    packages = ("ceph-iscsi",)
+    services = ("rbd-target-api", "rbd-target-gw")
+
+    def setup(self):
+        self.add_copy_spec(
+            [
+                "/etc/tcmu/tcmu.conf",
+                "/var/log/ceph/ceph-client.iscsi.log",
```
TurboTurtle, line 25: In updating some of the other ceph plugins, I've found that the `/var/log` locations have changed in newer versions (16+) of ceph. Most include the fsid, e.g. `/var/log/ceph/` to `/var/log/ceph/fsid/`.

Can we double check to make sure we're able to collect this log file from both newer and older versions?

sos/report/plugins/ceph_iscsi.py
nobuto-m: I'd happy to remove the section to be simpler. I assumed that it's an extra harness across all ceph plugins since radosgw plugin is in the same situation.
https://github.com/sosreport/sos/blob/089839231aa54261f274f9e963f2d07d3dc7de86/sos/report/plugins/ceph_rgw.py#L28-L39

nobuto-m:
>     * do the `packages = ("ceph-iscsi",)` and `services = ("rbd-target-api", "rbd-target-gw")` enablement condition match on `RedHat` distro?

According to the manual installation guide for RHEL/CentOS, the package name and the two service names are common between Ubuntu and RHEL.
https://docs.ceph.com/en/latest/rbd/iscsi-target-cli/

>     * logs and commands execution exist there as well?

`gwcli` command for sure as in the doc above. And let me double check the log location. I doubt it contains fsid because ceph-iscsi is rbd client side basically not Ceph cluster side.

sos/report/plugins/ceph_iscsi.py
pmoravec: I see. Until @TurboTurtle opposes, let leave the section as is, then.

TurboTurtle: We can afford to drop it in other plugins if it's superfluous - since none of the paths added by this particular plugin are in this forbidden list, we can drop this call as well.

nobuto-m: Done, dropped.

sos/report/plugins/ceph_iscsi.py
nobuto-m: Done. It was from black; the format has been updated.

pmoravec (approved):

sos/report/plugins/ceph_iscsi.py
nobuto-m: I dug into this but I'm not sure I'm following. Indeed, there is a particular path conversion with fsid in bind mount with `cephadm`, but from a process point of view the path is just `/var/log/ceph`, e.g. `/var/log/ceph/ceph-osd.0.log`. So not sure if that's the case for non-cephadm deployments.

```
docker inspect ceph-894b217c-8b2c-11ed-829c-930768227c30-iscsi-iscsi-cephadm-onylll | jq '.[].Mounts | .[] | select(.Destination=="/var/log")'
{
  "Type": "bind",
  "Source": "/var/log/ceph/894b217c-8b2c-11ed-829c-930768227c30",
  "Destination": "/var/log",
  "Mode": "z",
  "RW": true,
  "Propagation": "rprivate"
}
```

TurboTurtle: If the collections needs to happen in a container, then you need to specify the container in the `add_copy_spec()`/`add_cmd_output()` calls with the `container` parameter (either the container name or the ID). 

Having the containers listed in the `containers` tuple just enables the plugin to run. As written, these collections would still happen directly on the host - so to copy them from the host you would need to use `/var/log/ceph/*/ceph-client.iscsi.log` or similar.

TurboTurtle (approved):
Ack.

Merge pending confirmation on if the collections are meant to be from the container or the host. If the host location is volume-mounted into the container, slight preference from me to copy from the host but it is not a required pattern.

nobuto-m:
> Merge pending confirmation on if the collections are meant to be from the container or the host. If the host location is volume-mounted into the container, slight preference from me to copy from the host but it is not a required pattern.

The intention was to detect containers and collect logs on the host instead of from inside a container.

I wasn't aware https://github.com/sosreport/sos/pull/3092 was coming so assumed that the existing code base didn't expect `cephadm` based deployment yet but package based and ceph-ansible container based deployments.

So this PR can be merged for package based and ceph-ansible based deployments. For accommodating both existing deployments and cephadm, I'd like to wait for a conclusion in https://github.com/sosreport/sos/issues/3100 so Ceph iSCSI plugin can be updated as well later.

TurboTurtle:
Per the discussion in #3100 - let's update this to use `**` and merge it at that point.

sos/report/plugins/ceph_iscsi.py
```diff
@@ -0,0 +1,36 @@
...
+    plugin_name = "ceph_iscsi"
+    profiles = ("storage", "virt", "container")
+    packages = ("ceph-iscsi",)
+    services = ("rbd-target-api", "rbd-target-gw")
+    containers = ("rbd-target-api.*", "rbd-target-gw.*")
+
+    def setup(self):
+        self.add_copy_spec([
+            "/etc/tcmu/tcmu.conf",
+            "/var/log/ceph/ceph-client.iscsi.log",
```
TurboTurtle, line 25: Cycling back around on this. I think we just need this to be `/var/log/ceph/**/ceph-client.iscsi.log` and this one will be good to merge.

pmoravec (approved):

PR #3102, opened 2023-01-05
Fix docstring for copyspecs param in add_copy_spec

swalladge:
The string or strings are treated as globs, not regular expressions.

---

### DIFF
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -1631,9 +1631,9 @@ class Plugin():
 
     def add_copy_spec(self, copyspecs, sizelimit=None, maxage=None,
                       tailit=True, pred=None, tags=[], container=None):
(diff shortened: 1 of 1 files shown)


pmoravec (approved):

TurboTurtle (commented):
Ack. However, please see the [Contribution Guidelines](https://github.com/sosreport/sos/wiki/Contribution-Guidelines) for formatting of the commit title.

TurboTurtle:
`[tag]` style is what we use, e.g. in this case `[Plugin] Fix docstring....`

It may seem pedantic, but it helps a lot when going through the git history and being able to pick out groups of commits like this.

Also, just noticed - we'll want to replace the use of "regex" in the first line of the docstring as well.

TurboTurtle (approved):

PR #3116, opened 2023-01-18
WIP: [microshift] Add new plugin

TurboTurtle:
Adds a new `microshift` plugin that serves to collect information from microshift clusters. Microshift is an edge-focused derivative of OpenShift Container Platform.

The collections are nearly identical to the Openshift plugin, however Microshift does not ship the oauth bits that enable the `oc login` and `oc whoami` commands that the Openshift plugin relies on (and with those commands missing, the plugin would never collect API information for Microshift deployments).

Create a new plugin to differentiate collections between deployment types (via the changed plugin collection directory), but simply build on top of the Openshift plugin rather than duplicating the entire `setup()` method. Override the method used to determine if we can collect from the API, as Microshift deployments do allow system root to make these collections.

---

### DIFF
--- a/sos/report/plugins/microshift.py
+++ b/sos/report/plugins/microshift.py
@@ -1,0 +1,68 @@
+# Copyright 2023 Red Hat, Inc. Jake Hunsaker <jhunsake@redhat.com>
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins.openshift import Openshift
+from sos.utilities import is_executable
+
+
+class Microshift(Openshift):
+    """
+    For Microshift users, this plugin does not need to perform any oauth steps
+    of any kind. Users should see the `microshift` plugin enable and NOT the
+    `openshift` plugin during a normal execution of `sos report`. All plugin
+    options are otherwise the same, however users should note that the `token`
+    option is ignored by the Microshift plugin variant.
+    """
+
+    short_desc = 'Microshift Container Platform'
+    plugin_name = 'microshift'
+    packages = ('microshift',)
+    services = ('microshift',)
+    profiles = ()
+    master_localhost_kubeconfig = ''  # TODO: is there a default kubeconfig?
+
+    def _check_oc_function(self):
+        """Microshift does not have the oauth bits that enable the ``oc login``
+        or ``oc whoami`` commands. As such, the ``oc`` commands are otherwise
+        simply available for use. In this case, don't block API collections
+        as the Openshift plugin would when these non-existent commands fail.
+        """
+        # is there an `oc` command we can run that's not `oc whoami` safely
+        # that we can use with microshift?
+        return is_executable('oc')
+
+    def get_global_resource_list(self):
+        return []  # TODO: define non-namespaced resources to collect
+
+    def get_namespaced_resource_list(self):
+        """
+        Microshift does not support all the same resources as OCP does, so we
+        need to define a shorter list here that only specifies the resources
+        we care about in a microshift environment.
+        """
+        return [
+            'configmaps',
+            'daemonsets',
+            'deployments',
+            'events',
+            'ingresscontrollers',
+            'ingresses',
+            'pods',
+            'pvc',
+            'resourcequotas',
+            'routes',
+            'secrets',
+            'services',
+            'statefulsets',
+        ]
+
+    def setup(self):
+        super(Microshift, self).setup()
+        # TODO: add microshift-specific collections here that should not be
+        # included in the parent openshift plugin
--- a/sos/report/plugins/openshift.py
+++ b/sos/report/plugins/openshift.py
@@ -32,7 +32,8 @@ class Openshift(Plugin, RedHatPlugin):
     `only-namespaces` options.
 
     It is expected to need to perform an `oc login` command in order for this
-    plugin to be able to correctly capture information, as system root is not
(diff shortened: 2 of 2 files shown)


TurboTurtle:
@pacevedom - here's the draft plugin so far for microshift. As you can see we're simply inheriting the openshift plugin and making minor changes as needed by microshift deployment differences.

The main questions are:

- What resources do we need to collect here? I don't know all the differences between what microshift and ocp support or consider relevant for diagnostics. I have pods and deployments there mainly as a placeholder for the moment.
- Is there a default kubeconfig we can leverage, or does `oc get` and similar commands just work out of the gate for system root?
- Are there any additional microshift-specific collections that we need to make that aren't relevant for OCP?


There was some talk about the `crio` plugin on the enhancement doc, and I'll mention this over there as well, but one thing to note is that the `openshift` plugin (and so, currently, this microshift plugin as well) _also_ collects pod logs, that might be a needless duplication? We could look at also creating a `preset` for microshift like we do for OCP - simply put, presets in sos are groups of options we apply by default if we detect we're running in a certain environment. E.G if we're running on OCP, we set higher timeouts, disable the cgroups plugin (long history there), and configure some `networking` plugin options - https://github.com/sosreport/sos/blob/main/sos/presets/redhat/__init__.py#L39-#L49

TurboTurtle:
Should be noted - I have not yet run this myself on microshift, as I am not aware of how to get my hands on a test environment.

sos/report/plugins/microshift.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+
+    short_desc = 'Microshift Container Platform'
+    plugin_name = 'microshift'
+    packages = ('microshift',)
+    services = ('microshift',)
+    profiles = ()
+    master_localhost_kubeconfig = ''  # TODO: is there a default kubeconfig?
+
+    def _check_oc_logged_in(self):
```
dhellmann, line 29: This function could verify that the required kubeconfig file exists.

sos/report/plugins/microshift.py
```diff
@@ -0,0 +1,52 @@
...
+    def get_resource_list(self):
+        """
+        Microshift does not support all the same resources as OCP does, so we
+        need to define a shorter list here that only specifies the resources
+        we care about in a microshift environment.
+        """
+        return [
+            'pods',
+            'deployments',
+            # TODO: build this out with the microshift team
```
dhellmann, line 46:
```suggestion
            'configmaps',
            'daemonsets',
            'deployments',
            'events', # not sure about this one?
            'ingresscontrollers',
            'ingresses',
            'pods',
            'pvc',
            'resourcequotas',
            'routes',
            'secrets',
            'services',
            'statefulsets',
```

We should get the team to verify this is complete.

sos/report/plugins/microshift.py
```diff
@@ -0,0 +1,52 @@
...
+        ]
+
+    def setup(self):
+        super(Microshift, self).setup()
+        # TODO: add microshift-specific collections here that should not be
+        # included in the parent openshift plugin
```
dhellmann, line 52: There's also some stuff in the parent class that we wouldn't want for MicroShift. We may need a common base class with OCP and MicroShift extensions of it.

For example, there's not a kubelet log on MicroShift. We would want the message about the time it takes to run oc to say "MicroShift" not "OpenShift". We'll want to look at some of the other details, too.

sos/report/plugins/microshift.py
TurboTurtle: The parent class does so - it's probably obscured by the diff output, but we check to see if the kubeconfig file exists, and if it does we run all our `oc` commands with the `KUBECONFIG` env var set to that location.

This function override is to simply bypass the `oc login` logic. If `oc whoami` won't work under microshift, we'd need a separate method override to run some `oc` command to verify that we're able to do so successfully (so _what_ command that is doesn't actually matter as much).

sos/report/plugins/microshift.py
TurboTurtle: Ack on the messaging.

If we specify a collection that doesn't exist on the host, we simply skip it and move on - we won't see an empty `/etc/kubernetes` directory in the archive for example.

sos/report/plugins/microshift.py
TurboTurtle: Oh, sorry - misread your note. Ack, that should be done here.

TurboTurtle: Realized I had the wrong method overridden here for what I was intending. The path verification is indeed done in the parent class, before calling `_check_oc_function()` which is what I had meant to originally override.

In `openshift` that is where we do `oc whoami`, and if that fails we try to do the `oc login`. As a placeholder, I'm just checking that we have `oc` available to us - if there's an actual `oc` command we can safely run for microshift then let's use that instead.

sos/report/plugins/microshift.py
TurboTurtle: Included this for now. I also split out the "global" (non-namespaced, e.g. nodes) resources similar to this for us to define for microshift. Same deal here, I'm not aware of what the differences are.

sos/report/plugins/microshift.py
dhellmann: MicroShift has a kubeadmin kubconfig on disk already, and there's no login. So we could test that the credentials work for listing something simple without an error (maybe like listing namespaces?) but there's no separate login step.

jrthms: I left a comment in https://github.com/openshift/enhancements/pull/1318/files regarding kube config file location and use of -k openshift.kubeconfig=/home/microshift/.kube/config. I'm not sure if that would be the default or the /var/lib/microshift/resources/kubeadmin/kubeconfig file.

TurboTurtle:
Closing in favor of #3119

PR #3117, opened 2023-01-19
[policies] set case_id before prompting for upload credentials

pmoravec:
Since prompting for upload username/password relies on case_id for some policies, let fully set case_id before that.

Resolves: #3117

---

### DIFF
--- a/sos/policies/distros/__init__.py
+++ b/sos/policies/distros/__init__.py
@@ -286,6 +286,7 @@ class LinuxPolicy(Policy):
         self.upload_password = cmdline_opts.upload_pass
         self.upload_archive_name = ''
 
+        # set or query for case id
         if not cmdline_opts.batch and not \
                 cmdline_opts.quiet:
             try:
@@ -296,6 +297,16 @@ class LinuxPolicy(Policy):
                         _("Optionally, please enter the case id that you are "
                           "generating this report for [%s]: ") % caseid
                     )
+            except KeyboardInterrupt:
+                raise
+        if cmdline_opts.case_id:
+            self.case_id = cmdline_opts.case_id
+
+        # set or query for upload credentials; this needs to be done after
+        # setting case id, as below methods might rely on detection of it
+        if not cmdline_opts.batch and not \
+                cmdline_opts.quiet:
+            try:
                 # Policies will need to handle the prompts for user information
                 if cmdline_opts.upload and self.get_upload_url():
                     self.prompt_for_upload_user()
@@ -303,9 +314,6 @@ class LinuxPolicy(Policy):
                 self.ui_log.info('')
             except KeyboardInterrupt:
                 raise
-
-        if cmdline_opts.case_id:
-            self.case_id = cmdline_opts.case_id
 
         return
 
--- a/sos/policies/distros/redhat.py
+++ b/sos/policies/distros/redhat.py
@@ -266,11 +266,19 @@ support representative.
         if self.commons['cmdlineopts'].upload_url:
             super(RHELPolicy, self).prompt_for_upload_user()
             return
-        if self.case_id and not self.get_upload_user():
-            self.upload_user = input(_(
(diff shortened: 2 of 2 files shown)


pmoravec:
Reproducer of the current problem: on `RedHat` distro:

```
sos report -o qpid --upload
...
(provide case id when prompted)
..
	/var/tmp/sosreport-pmoravec-rhel8-02732787-2023-01-18-wwyaytq.tar.xz

 Size	4.10KiB
 Owner	root
 sha256	7176170d965de5c3192b3f13ba55c1bfb4da215d4cebed9a8281c956e07bb108

Attempting upload to Red Hat Customer Portal
Upload to Red Hat Customer Portal failed. Trying sftp://sftp.access.redhat.com
User 'OBFUSCATED'used for anonymous upload. Please inform your support engineer so they may retrieve the data.
Uploaded archive successfully
```

And no attempt to prompt upload credentials was made.

The cause is `RedHat` distro prompts only when `self.case_id` is set (https://github.com/sosreport/sos/blob/main/sos/policies/distros/redhat.py#L269), which setting we must move higher in the code.

Relevant to: https://bugzilla.redhat.com/show_bug.cgi?id=2161778

pmoravec:
Cc @mhradile for a potential review, as there are multiple upload scenarios around.

TurboTurtle (approved):
Ack.

pmoravec:
From a testing perspective: the only difference in behaviour happens when

```
        if not cmdline_opts.batch and not \
                cmdline_opts.quiet:
```

is `True`. Here, all three variants of behaviour:
- `case_id` provided in cmdline or in config
- `case_id` provided interactively
- `case_id` not provided at all (and we need to fail over to SFTP *with* prompting for credentials (blank=anonymous) - THIS is just being improved by the additional commit

The new commit alters `sos` behaviour just and only in scenario (`case_id` not provided, `upload_url` is not set, but we are already asking for credentials - so fail over from RHCP to SFTP but still query for credentials), so no regression is expected.

Can @mhradile confirm?

sos/policies/distros/redhat.py
```diff
@@ -266,11 +266,19 @@ def prompt_for_upload_user(self):
...
-            )
+        if not self.get_upload_user():
+            if self.case_id:
+                self.upload_user = input(_(
+                    "Enter your Red Hat Customer Portal username for "
+                    "uploading [empty for anonymous SFTP]: ")
+                )
+            else:   # no case id provided => failover to SFTP
+                self.upload_url = RH_SFTP_HOST
+                print(_("No case id provided, uploading to SFTP"))
```
TurboTurtle, line 277: With #3118 now merged, let's move this to a logged message.

sos/policies/distros/__init__.py
```diff
@@ -296,6 +297,17 @@ def pre_work(self):
                         _("Optionally, please enter the case id that you are "
                           "generating this report for [%s]: ") % caseid
                     )
+            except KeyboardInterrupt:
+                self._print()
```
TurboTurtle, line 301: Do we need a blank print here? Also, `_print()` is now removed with #3118.

TurboTurtle (approved):

mhradile:
> is `True`. Here, all three variants of behaviour:
> 
>     * `case_id` provided in cmdline or in config
> 
>     * `case_id` provided interactively
> 
>     * `case_id` not provided at all (and we need to fail over to SFTP _with_ prompting for credentials (blank=anonymous) - THIS is just being improved by the additional commit
> 
> 
> The new commit alters `sos` behaviour just and only in scenario (`case_id` not provided, `upload_url` is not set, but we are already asking for credentials - so fail over from RHCP to SFTP but still query for credentials), so no regression is expected.
> 
> Can @mhradile confirm?

I did test (manually) all the mentioned scenarios (with the two commits in this PR applied) and it behaves just as you are mentioning. In the last case (no case_id given) the credentials are now also asked for before failing over to anonymous upload.

PR #3119, opened 2023-01-19
Add MicroShift plugin

pacevedom:
Add MicroShift initial plugin.

Does not share code with OpenShift plugin because of little similarity except helper functions. Preferred to keep it separated as MicroShift is far apart from OCP in terms of similarity and might evolve differently.

---

### DIFF
--- a/sos/report/plugins/microshift.py
+++ b/sos/report/plugins/microshift.py
@@ -1,0 +1,167 @@
+# Copyright 2023 Red Hat, Inc. Pablo Acevedo <pacevedo@redhat.com>
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, RedHatPlugin, PluginOpt
+import re
+
+
+class Microshift(Plugin, RedHatPlugin):
+    """This is the plugin for MicroShift 4.X. Even though it shares some of
+    the OpenShift components, its IoT/Edge target makes the product nimble and
+    light, thus requiring different a approach when operating it.
+
+    When enabled, this plugin will collect cluster information (such as
+    systemd service logs, configuration, versions, etc.)and also inspect API
+    resources in certain namespaces. The namespaces to scan are kube.* and
+    openshift.*. Other namespaces may be collected by making use of the
+    ``only-namespaces`` or ``add-namespaces`` options.
+    """
+
+    short_desc = 'Microshift'
+    plugin_name = 'microshift'
+    plugin_timeout = 900
+    packages = ('microshift', 'microshift-selinux', 'microshift-networking')
+    services = (plugin_name,)
+    profiles = (plugin_name,)
+    localhost_kubeconfig = '/var/lib/microshift/resources/kubeadmin/kubeconfig'
+
+    option_list = [
+        PluginOpt('kubeconfig', default=localhost_kubeconfig, val_type=str,
+                  desc='Path to a locally available kubeconfig file'),
+        PluginOpt('only-namespaces', default='', val_type=str,
+                  desc='colon-delimited list of namespaces to collect from'),
+        PluginOpt('add-namespaces', default='', val_type=str,
+                  desc=('colon-delimited list of namespaces to add to the '
+                        'default collection list'))
+    ]
+
+    def _setup_namespace_regexes(self):
+        """Combine a set of regexes for collection with any namespaces passed
+        to sos via the -k openshift.add-namespaces option. Note that this does
+        allow for end users to specify namespace regexes of their own.
+        """
+        if self.get_option('only-namespaces'):
+            return [n for n in self.get_option('only-namespaces').split(':')]
+
+        collect_regexes = [
+            r'^openshift\-.+$',
+            r'^kube\-.+$'
+        ]
+
+        if self.get_option('add-namespaces'):
+            for nsp in self.get_option('add-namespaces').split(':'):
+                collect_regexes.append(r'^%s$' % nsp)
+
+        return collect_regexes
+
+    def _reduce_namespace_list(self, nsps):
+        """Reduce the namespace listing returned to just the ones we want to
+        collect from. By default, as requested by OCP support personnel, this
+        must include all 'openshift' prefixed namespaces
+
+            :param nsps list:            Namespace names from oc output
+        """
+
+        def _match_namespace(namespace):
+            """Match a particular namespace for inclusion (or not) in the
+            collection phases
+
+                :param namespace str:   The name of a namespace
+            """
+
+            for regex in self.collect_regexes:
+                if re.match(regex, namespace):
+                    return True
+            return False
+
+        self.collect_regexes = self._setup_namespace_regexes()
+
+        return list(set([n for n in nsps if _match_namespace(n)]))
+
+    def _get_namespaces(self):
+        res = self.exec_cmd(
+            'microshift get namespaces'
+            ' -o custom-columns=NAME:.metadata.name'
+            ' --no-headers')
+        if res['status'] == 0:
+            return self._reduce_namespace_list(res['output'].split('\n'))
+        return []
+
+    def _get_cluster_resources(self):
+        """Get cluster-level (non-namespaced) resources to collect
+        """
+        global_resources = [
+            'apiservices',
+            'certificatesigningrequests',
+            'clusterrolebindings',
+            'clusterroles',
+            'componentstatuses',
+            'csidrivers',
+            'csinodes',
+            'customresourcedefinitions',
+            'flowschemas',
+            'ingressclasses',
+            'logicalvolumes',
+            'mutatingwebhookconfigurations',
+            'nodes',
+            'persistentvolumes',
+            'priorityclasses',
+            'prioritylevelconfigurations',
+            'rangeallocations',
+            'runtimeclasses',
+            'securitycontextconstraints',
+            'selfsubjectaccessreviews',
+            'selfsubjectrulesreviews',
+            'storageclasses',
+            'subjectaccessreviews',
+            'tokenreviews',
+            'validatingwebhookconfigurations',
+            'volumeattachments'
+        ]
+
+        _filtered_resources = []
+
+        for resource in global_resources:
+            res = self.exec_cmd(
+                "oc get --kubeconfig %s %s" % (
+                    self.get_option('kubeconfig'), resource),
+                timeout=Microshift.plugin_timeout)
+            if res['status'] == 0:
+                _filtered_resources.append(resource)
+        return _filtered_resources
+
+    def setup(self):
+        """The setup() phase of this plugin will first gather system
+        information and then iterate through all default namespaces, and/or
+        those specified via the `add-namespaces` and `only-namespaces` plugin
+        options. Both of these options accept shell-style regexes.
+
+        Output format for this function is based on `oc adm inspect` command,
+        which is used to retrieve all API resources from the cluster.
+        """
+        self.add_forbidden_path('/var/lib/microshift')
+
+        _cluster_resources_to_collect = ",".join(
+            self._get_cluster_resources())
+        _namespaces_to_collect = " ".join(
+            ['ns/%s' % n for n in self._get_namespaces()])
+
+        if self.is_service_running(Microshift.plugin_name):
+            _subdir = self.get_cmd_output_path(make=False)
+            _kubeconfig = self.get_option('kubeconfig')
+            self.add_cmd_output(
+                'oc adm inspect --kubeconfig %s --dest-dir %s %s' % (
+                    _kubeconfig, _subdir, _cluster_resources_to_collect),
+                suggest_filename='inspect_cluster_resources.log',
+                timeout=Microshift.plugin_timeout)
+            self.add_cmd_output(
+                'oc adm inspect --kubeconfig %s --dest-dir %s %s' % (
+                    _kubeconfig, _subdir, _namespaces_to_collect),
+                suggest_filename='inspect_namespaces.log',
+                timeout=Microshift.plugin_timeout)
--- a/sos/report/plugins/microshift_ovn.py
+++ b/sos/report/plugins/microshift_ovn.py
@@ -1,0 +1,47 @@
+# Copyright (C) 2023 Pablo Acevedo <pacevedo@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, RedHatPlugin
+
+
+class MicroshiftOVN(Plugin, RedHatPlugin):
+    """This plugin is used to collect MicroShift 4.x OVN logs.
+    """
+    short_desc = 'MicroShift OVN'
+    plugin_name = "microshift_ovn"
+    plugin_timeout = 300
+    containers = ('ovnkube-node', 'ovnkube-master',)
+    packages = ('microshift-networking',)
+    profiles = ('microshift',)
+
+    def setup(self):
+        self.add_copy_spec([
+            '/etc/openvswitch/conf.db',
+            '/etc/openvswitch/default.conf',
+            '/etc/openvswitch/system-id.conf'])
+
+        _ovs_cmd = 'ovs-appctl -t /var/run/ovn/'
+        _subcmds = [
+            'coverage/show',
+            'memory/show',
+            'ovsdb-server/sync-status'
+        ]
+        for file, db in [('ovnnb_db.ctl', 'OVN_Northbound'),
+                         ('ovnsb_db.ctl', 'OVN_Southbound')]:
+            self.add_cmd_output(
+                [f"{_ovs_cmd}{file} {cmd}" for cmd in _subcmds],
+                timeout=MicroshiftOVN.plugin_timeout)
+            self.add_cmd_output(
+                f"{_ovs_cmd}{file} ovsdb-server/get-db-storage-status {db}",
+                timeout=MicroshiftOVN.plugin_timeout)
+
+        self.add_cmd_output(
+            f'{_ovs_cmd}ovn-controller.*.ctl ct-zone-list',
+            timeout=MicroshiftOVN.plugin_timeout)


TurboTurtle:
This duplicates a lot of the openshift plugin, which is why #3116 directly builds on top of it. I get that there is separation in terms of use cases and deployments between OCP and microshift, but from strictly a data collection standpoint there seems to be more similarities than differences.

The main differences appears to come down to:

1. The resources collected. Looking at the global and namespaced resources there is a lot shared but there's also some unique ones for each. This is why #3116 separated those list definitions out between the two plugins - the collection mechanism remains the same underneath it all. In fact, I wonder if some of the resources added here wouldn't _also_ be useful for `openshift` collections.
2. The service/journal collected and a couple `microshift` command outputs. The former is already handled in the case of a missing service, and both the additions can be handled in an overridden `setup()` that invokes `super()` beforehand.


Since both microshift and OCP would rely on `oc` to generate these collections, I'm hesitant to not build from the base `openshift` plugin (or perhaps splitting out a separate base plugin that each builds off of independently?) from a maintainability standpoint. In either the case of something with `oc` changing or if new resources are added that both OCP and microshift want to collect, keeping everything copacetic is much easier if we're directly inheriting one from the other.

pacevedom:
> This duplicates a lot of the openshift plugin, which is why #3116 directly builds on top of it. I get that there is separation in terms of use cases and deployments between OCP and microshift, but from strictly a data collection standpoint there seems to be more similarities than differences.
> 
> The main differences appears to come down to:
> 
> 1. The resources collected. Looking at the global and namespaced resources there is a lot shared but there's also some unique ones for each. This is why [WIP: [microshift] Add new plugin #3116](https://github.com/sosreport/sos/pull/3116) separated those list definitions out between the two plugins - the collection mechanism remains the same underneath it all. In fact, I wonder if some of the resources added here wouldn't _also_ be useful for `openshift` collections.
> 2. The service/journal collected and a couple `microshift` command outputs. The former is already handled in the case of a missing service, and both the additions can be handled in an overridden `setup()` that invokes `super()` beforehand.

What do you think about extracting common functions to another base class and have OCP and MicroShift plugins inherit from it? In the end many (or almost all) functions can be parameterized and serve both platforms.

> 
> Since both microshift and OCP would rely on `oc` to generate these collections, I'm hesitant to not build from the base `openshift` plugin (or perhaps splitting out a separate base plugin that each builds off of independently?) from a maintainability standpoint. In either the case of something with `oc` changing or if new resources are added that both OCP and microshift want to collect, keeping everything copacetic is much easier if we're directly inheriting one from the other.

Agree, but what concerns me is the different evolution path for both platforms so as to rely on top of OCP. Having the MicroShift plugin inherit and then redefine retrieval information functions only to have the auxiliary ones looks complex to me, so I would rather go with the separate base plugin. Let me work out a prototype.

sos/report/plugins/microshift.py
```diff
@@ -0,0 +1,173 @@
...
+    openshift.*. Other namespaces may be collected by making use of the
+    ``only-namespaces`` or ``add-namespaces`` options.
+    """
+
+    short_desc = 'Microshift'
+    plugin_name = 'microshift'
+    plugin_timeout = 900
+    packages = (plugin_name,)
+    services = (plugin_name,)
+    profiles = (plugin_name,)
```
pmoravec, line 30: Plugin profile is sort of name of a set of similar plugins like OpenStack, services, storage or similar. Why to introduce a new profile here?

FYI list of profiles and numbers of plugins assigned to them:

```
      1 "apache"
      1 "cifs"
      1 "mrg"
      1 "perl"
      3 "gpu"
      3 "java"
      3 "sap"
      4 "mail"
      4 "memory"
      4 "nfs"
      4 "openstack_undercloud"
      5 "openstack_compute"
      6 "debug"
      6 "performance"
      7 "kernel"
      7 "openshift"
      8 "packagemanager"
      9 "desktop"
     11 "boot"
     11 "webserver"
     13 "security"
     15 "cluster"
     15 "identity"
     21 "container"
     23 "openstack_controller"
     27 "sysmgmt"
     29 "openstack"
     31 "virt"
     34 "services"
     35 "network"
     59 "hardware"
     59 "storage"
    111 "system"
```

sos/report/plugins/microshift.py
```diff
@@ -0,0 +1,173 @@
...
+            res = self.exec_cmd("microshift get %s" % resource)
+            if res['status'] == 0:
+                _filtered_resources.append(resource)
+
+        _subdir = self.get_cmd_output_path(make=False)
+        self.exec_cmd(
+            'microshift inspect --kubeconfig %s --dest-dir %s %s' % (
+                self.get_option('kubeconfig'),
+                _subdir,
+                ",".join(_filtered_resources)))
```
pmoravec, line 137: You are not interested in the command output, but in the file it automatically generates, did I get it right?

sos/report/plugins/microshift.py
```diff
@@ -0,0 +1,173 @@
...
+        those specified via the `add-namespaces` and `only-namespaces` plugin
+        options. Both of these options accept shell-style regexes.
+
+        Output format for this function is based on `oc adm inspect` command,
+        which is used to retrieve all API resources from the cluster.
+        """
+
+        self.add_forbidden_path('/var/lib/microshift')
+        self.add_journal('microshift')
+        self.add_service_status('microshift')
```
TurboTurtle, line 163: Since `microshift` is defined in the `services` attr, these are unnecessary as these collections are automatically added for any service defined for a plugin.

sos/report/plugins/microshift.py
TurboTurtle, line 137: Calling `exec_cmd()` to generate a file in the plugin directory is fine, however there is now the paradigm in plugin execution of the `collect()` method being used for actual collections, and leaving `setup()` to _only_ define what those collections should be.

So, in this case `setup()` calling `_collect_cluster_resources()` should simply compile the list of `_filtered_resources`, and then in this plugin's `collect()` method would be something like what you have here, e.g.:

```python
    def collect(self):
        _subdir = self.get_cmd_output_path(make=False)
        _konfig = self.get_option('kubeconfig')
        self.exec_cmd(
            f"microshift inspect --kubeconfig {_konfig} --dest-dir {_subdir} "
            f"{','.join(self._filtered_resources)}"
        )

        # then the other exec_cmd() call to inspect can follow here
```

Alternatively, there is also the possibility of running this in `setup()` and getting _both_ this file and the output of the command (if for some reason this collection fails):

```python
self.add_cmd_output(
        f"microshift inspect --kubeconfig {_konfig} --dest-dir {_subdir} "
            f"{','.join(self._filtered_resources)}",
        suggest_filename='some_shorthand_name'
)

`suggest_filename` might be useful here as the standard file naming convention for `add_cmd_output()` would not be especially intuitive/helpful to users.

On that note, how does `microshift inspect` name the resulting file? I suspect we'd want this to be standardized so that automation can be used despite what `_filtered_resources` turns out to be.

File: sos/report/plugins/microshift.py [hunk 4]
```diff
@@ -0,0 +1,173 @@
...
+        """Collect namespaced resources (including pod logs) from designated
+        namespaces
+        """
+        _namespaces = ['ns/%s' % n for n in self._get_namespaces()]
+        _subdir = self.get_cmd_output_path(make=False)
+        res = self.exec_cmd(
+            'microshift inspect --kubeconfig %s --dest-dir %s %s' % (
+                self.get_option('kubeconfig'),
+                _subdir,
+                " ".join(_namespaces)))
```
TurboTurtle, line 149: Same here as above - either call `exec_cmd()` during `collect()`, or change this to `add_cmd_output()` with a particular `suggest_filename` so that we get both the resulting output file from `microshift inspect` and any potential error output from the command.

TurboTurtle (commented):
Overall looks fine, a couple comments below.


I do wonder if we still shouldn't have some common base plugin from which both `openshift` and `microshift` are built. If nothing else, for the namespaces handling as well as the cluster-level resources.

sos/report/plugins/microshift.py
TurboTurtle: That's the perfect use case for a preset. We can auto-load a specific configuration for sos report when `microshift` is present. No need for a specific config file or wrapper script or similar.

Users will just execute `sos report`, and behind the scenes the preset will set whatever options you want for it. Users can still override individual options from the preset via the CLI if needed.

sos/report/plugins/microshift.py
TurboTurtle: Ack.

One thing we may want to look at down the line is including these files in `sos_reports/manifest.json`. Currently, none of these files would be included in there since sos didn't directly create them. It's not a blocker for us including this plugin as the manifest is primarily for automation like Insights (which can now use tags in the manifest to locate specific files rather than relying on globs or matching files with changing command options), but just something to keep in mind.

sos/report/plugins/microshift.py
pacevedom: I am exploring the presets, will add it to the PR when its ready.

sos/report/plugins/microshift.py
```diff
@@ -0,0 +1,173 @@
...
+                if re.match(regex, namespace):
+                    return True
+            return False
+
+        self.collect_regexes = self._setup_namespace_regexes()
+
+        return list(set([n for n in nsps if _match_namespace(n)]))
+    
+    def _get_namespaces(self):
+        res = self.exec_cmd('microshift get namespaces -o custom-columns=NAME:.metadata.name --no-headers')
```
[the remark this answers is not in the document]
pacevedom: Done

sos/report/plugins/microshift.py
pacevedom: Finally I went with `setup` instead. `collect` is called by `_collect_manual`, which specifies in the [description](https://github.com/sosreport/sos/blob/main/sos/report/plugins/__init__.py#L3120-L3123) that it is intended for any collections outside command output and/or existing files, which is what the plugin is doing here.

sos/report/plugins/microshift.py
pacevedom: changed it to `add_cmd_output`

sos/report/plugins/microshift.py
pacevedom: Done

TurboTurtle (approved):
LGTM.

We can either take this now and do the preset in a separate PR, or wait for the preset in this one. Whichever you'd prefer.

pmoravec (approved):

sos/report/plugins/microshift.py
```diff
@@ -0,0 +1,162 @@
...
+    def setup(self):
+        """The setup() phase of this plugin will first gather system
+        information and then iterate through all default namespaces, and/or
+        those specified via the `add-namespaces` and `only-namespaces` plugin
+        options. Both of these options accept shell-style regexes.
+
+        Output format for this function is based on `oc adm inspect` command,
+        which is used to retrieve all API resources from the cluster.
+        """
+        self.add_forbidden_path('/var/lib/microshift')
```
dhellmann, line 145: If we end up moving etcd's database out of /var/lib/microshift, will we need to include that location as forbidden, too?

See https://github.com/openshift/microshift/pull/1178
cc @dusk125

sos/report/plugins/microshift.py
```diff
@@ -0,0 +1,162 @@
...
+        _cluster_resources_to_collect = ",".join(
+            self._get_cluster_resources())
+        _namespaces_to_collect = " ".join(
+            ['ns/%s' % n for n in self._get_namespaces()])
+
+        if self.is_service_running(Microshift.plugin_name):
+            _subdir = self.get_cmd_output_path(make=False)
+            _kubeconfig = self.get_option('kubeconfig')
+            self.add_cmd_output(
+                'microshift inspect --kubeconfig %s --dest-dir %s %s' % (
```
dhellmann, line 156: I'm a bit worried that creating `microshift inspect` as an API forces us down a specific implementation path that will be difficult to change in the future. What's the best way to maintain some flexibility there?

Should the command be invoked with its full path, instead of relying on $PATH?

pacevedom:
Turns out we will be using `oc` instead of `microshift`, will update the PR.

sos/report/plugins/microshift.py
pacevedom: I would certainly do it, etcd contains everything and we are extracting the important bits with the `oc adm inspect` commands.

sos/report/plugins/microshift_ovn.py
```diff
@@ -0,0 +1,42 @@
...
+            'ovs-appctl -t /var/run/ovn/ovnnb_db.ctl memory/show',
+            'ovs-appctl -t /var/run/ovn/ovnsb_db.ctl memory/show',
+            'ovs-appctl -t /var/run/ovn/ovnnb_db.ctl ' +
+            'ovsdb-server/get-db-storage-status OVN_Northbound',
+            'ovs-appctl -t /var/run/ovn/ovnsb_db.ctl ' +
+            'ovsdb-server/get-db-storage-status OVN_Southbound',
+            'ovs-appctl -t /var/run/ovn/ovnnb_db.ctl ' +
+            'ovsdb-server/sync-status',
+            'ovs-appctl -t /var/run/ovn/ovnsb_db.ctl ' +
+            'ovsdb-server/sync-status'])
```
TurboTurtle, line 39: The defacto sos-style for this would be to define a list of subcmds and then use list comprehension like below so that it's easier to read and extend.

```python
ovs_cmd = 'ovs-appctl -t /var/run/ovn/'
subcmds = [
    'coverage/show',
    'memory/show',
    [...]
]

for ovn_db in ['ovnnb_db.ctl', 'ovnsb_db.ctl']:
    self.add_cmd_output([
        f"{ovs_cmd}{ovn_db} {cmd}" for cmd in subcmds
    ])
```

As a bonus, it helps with the PEP8 requirements.

pacevedom: Done.

sos/report/plugins/microshift.py
```diff
@@ -0,0 +1,162 @@
...
+            'subjectaccessreviews',
+            'tokenreviews',
+            'validatingwebhookconfigurations',
+            'volumeattachments'
+        ]
+
+        _filtered_resources = []
+
+        for resource in global_resources:
+            res = self.exec_cmd("oc get --kubeconfig %s %s" % (self.get_option('kubeconfig'), resource))
```
TurboTurtle, line 131: flake8 is failing here due to line length, need to break this up over multiple lines.

sos/report/plugins/microshift_ovn.py
```diff
@@ -0,0 +1,42 @@
...
+        _subcmds = [
+            'coverage/show',
+            'memory/show',
+            'ovsdb-server/sync-status'
+        ]
+        for file, db in [
+            ('ovnnb_db.ctl', 'OVN_Northbound'),
+            ('ovnsb_db.ctl', 'OVN_Southbound')]:
+            self.add_cmd_output(
+                [f"{_ovs_cmd}{file} {cmd}" for cmd in _subcmds])
```
TurboTurtle, line 37: flake8 is failing here due to the indention on line 35. Can either indent one more level, or do something like below. Up to you.

```python
for file, db in [('ovnnb_db.ctl', 'OVN_Northbound'),
                 ('ovnsb_db.ctl', 'OVN_Southbound')]:
```

TurboTurtle (commented):
Code looks fine, minor style changes highlighted by flake8 noted below.

sos/report/plugins/microshift_ovn.py
```diff
@@ -0,0 +1,42 @@
...
+
+from sos.report.plugins import Plugin, RedHatPlugin
+
+
+class MicroShiftOVN(Plugin, RedHatPlugin):
+    """This plugin is used to collect MicroShift 4.x OVN logs.
+    """
+    short_desc = 'MicroShift OVN'
+    plugin_name = "microshift_ovn"
+    profiles = ('microshift',)
```
pmoravec, line 19: Should not be there some enabling condition like `packages=..` or `services=..` or `containers=..` ? Currently the plugin is enabled on either system, every time.

sos/report/plugins/microshift.py
pacevedom: Done

sos/report/plugins/microshift_ovn.py
pacevedom: Done

sos/report/plugins/microshift_ovn.py
pacevedom: You are right, it was a leftover. Fixed now.

pmoravec (approved):
The older commit lacks DCO / Signed-off and https://github.com/sosreport/sos/pull/3119/files#diff-90b35793e0817e92a2b08600da680e52e95f51fbfff2adae8c8fbb017edc2319R20 would be great to have comma after the only one item of the list (as a code sanity/clarity nit-pick), but ACK as is.

Thanks for the PR.

pacevedom:
> The older commit lacks DCO / Signed-off and https://github.com/sosreport/sos/pull/3119/files#diff-90b35793e0817e92a2b08600da680e52e95f51fbfff2adae8c8fbb017edc2319R20 would be great to have comma after the only one item of the list (as a code sanity/clarity nit-pick), but ACK as is.
> 
> Thanks for the PR.

Both have been fixed.
Thanks!

TurboTurtle (approved):
LGTM.

Just thought of one last question - for the `oc adm inspect` command, how long is it expected to take generally? By default we have a 300 second timeout, and I just want to make sure that that is sufficient for most use cases. 

If not, or if we just want to increase it to be on the safe side, `add_cmd_output()` accepts a `timeout` parameter for a timeout in seconds.

Otherwise I think we're good to merge this.

pacevedom:
@TurboTurtle I added the timeouts for safety, but the plugin should be done in a few seconds only.

TurboTurtle (approved):

pacevedom:
@TurboTurtle @pmoravec this is ready to be merged now, as the enhancement is finished: https://github.com/openshift/enhancements/pull/1318

sos/report/plugins/microshift.py
vwalek: @pacevedom maybe we can rework that to collect at least the certificates and the kubeconfigs from that directory. It would help the situation where certificates might be expired.

pmoravec (approved):

sos/report/plugins/microshift.py
pacevedom: Arent the certificates and kubeconfigs sensitive data? Leaking them could lead to security issues.
If certs expire then that should be visible by other means, such as logs and other errors (for example, `sos report` would be incomplete, as it uses `oc`).

pmoravec: Very good (potential) point. From sos devel perspective, we dont know the content of data being collected so we can't assess or evaluate potential sensitivity of the data or a need to obfuscate them or even not collect some of them at all. We must rely on some domain experts here for a review.

Generally speaking, just certificates or just CAs *can* (but dont need to) be safe to be collected if they are publicly available or if they cant be used for some exploit. *Sometimes* it can be beneficial to collect them, sometimes those files are redundant for any troubleshooting. So please evaluate the certs to be collected from both security and usefullness point of view.

Certificate keys or passphrases should not be collected any time as they are in fact sort of passwords.

TurboTurtle (approved):
I'm going to merge this for now - if (scrubbed) certs are helpful, we can add those in a follow up commit. Cert-scrubbing is standard practice for plugins, and we just need to provide a command/file path (or regex) for our postprocessing to automatically scrub found certificates.

sos/report/plugins/microshift.py
vwalek: > Arent the certificates and kubeconfigs sensitive data? Leaking them could lead to security issues. If certs expire then that should be visible by other means, such as logs and other errors (for example, `sos report` would be incomplete, as it uses `oc`).

No, certificates are not sensitive as they are considered public. The kubeconfig in other hand is sensitive, but we can strip the sensitive data from the file. But yeah, thinking of it, kubeconfig is not really valuable, just the ca values in the server section. 

The point is that we just need to make sure that they are not expired, so we can run `openssl` with `-dates` to just list the expiry.

Unfortunately, sometimes the expired certificates are not visible in the logs. Most common errors in the logs are "signed by unknown authority" and "tls handshake error", which in both cases it is not clear if it is expired or not. Just that there is a problem.

PR #3120, opened 2023-01-20
[test] Add stage 2 test for ipv6 obfuscation

mhradile:
Relates: #3060

Signed-off-by: Miroslav Hradilek <mhradile@redhat.com>

---

### DIFF
--- a/tests/cleaner_tests/ipv6_test/ipv6.py
+++ b/tests/cleaner_tests/ipv6_test/ipv6.py
@@ -1,0 +1,23 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, IndependentPlugin
+
+
+class IPv6(Plugin, IndependentPlugin):
+    """Collect arbitrary file containing crafted ipv6 adresses to test ipv6
+    obfuscation.
+    """
+
+    plugin_name = 'ipv6'
+    short_desc = 'fake plugin to test ipv6 obfuscation'
+
+    def setup(self):
+        self.add_copy_spec([
+            '/tmp/sos-test-ipv6.txt',
+        ])
--- a/tests/cleaner_tests/ipv6_test/ipv6_test.py
+++ b/tests/cleaner_tests/ipv6_test/ipv6_test.py
@@ -1,0 +1,38 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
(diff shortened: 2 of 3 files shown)


mhradile:
Forgot to mention that the test is failing because I don't know whether the exception is desirable or not and whether I should assert it or not in the test.

TurboTurtle:
> Forgot to mention that the test is failing because I don't know whether the exception is desirable or not and whether I should assert it or not in the test.

That's a good question.

We're hitting an exception, because the entire address is identical hextets. So https://github.com/sosreport/sos/blob/main/sos/cleaner/mappings/ipv6_map.py#L267 actually produces `['', ':', '']` which is not expected, so we fallthru to https://github.com/sosreport/sos/blob/main/sos/cleaner/mappings/ipv6_map.py#L270 which produces the whole original address as a single item, which is not expected or handled.

I don't _think_ an IPv6 address that is comprised of _all_ identical hextets is valid according to the protocol, but I'm not sure. Which of course brings the thought - if it's not a valid IPv6 address, we shouldn't be obfuscating it in any way.

So here I think the test should verify that `BAD_IP` remains unchanged. The next question is how/if we handle this condition in code, or do we accept the logged error in `sos.log`: `2023-01-20 12:17:46,299 DEBUG: [cleaner:IPv6 Parser] failed to parse line: not enough values to unpack (expected 2, got 1)`.

TurboTurtle:
Of course, we can also make that logged error more helpful. Like "cannot obfuscate string {string} as IPv6 addr" or what have you.

pmoravec:
A few tools I checked (not only regexp based), they claimed it is a valid IPv6 address. Though a corner case, for sure.

Generally, big +1 to log `cannot obfuscate string {string} as IPv6 addr` as an exception handling, to let user to know "hey, sos detected a string as possible IPv6 address but failed to obfuscate it - are you ok with it..?". This would be a minimum fix/improvement I would like to see.

I will check the code if it is easy to make this IP address obfuscated, anyway.

mhradile:
I'm ok with it working either way as long as it is defined what it is supposed to act like. Either ignore invalid addresses or obfuscate invalid addresses for safety.

My approach would be to rather obfuscate more and have the rules simpler (replace everything that looks like ipv6) rather than with more difficult logic (validating) but I realize that this comes with the downside of replacing valid information mistaken for ipv6 address.

Either way, I believe we should avoid crashing on invalid ipv6 like strings if we encounter such situation.

Let me know how I should change the test or if i should propose a different one.

pmoravec:
Why the "split" method in https://github.com/sosreport/sos/blob/main/sos/cleaner/mappings/ipv6_map.py#L267, if we are interested in suffix? When simple:

```
                _n = self.network_addr.rstrip(':')
                _host = addr.compressed[len(_n):]
```

(i.e. remove network prefix without trailing ':' from the compressed address) instead the try-except block ? What corner case do I miss?

TurboTurtle:
That's certainly another way to get the host bits, and it seems to also cover this scenario by default.

pmoravec:
The failing (correct) test case should pass, once the #3121 is merged.

Good catch!

pmoravec:
@mhradile , could you please rebase your local branch to ensure the tests will succeed now?

pmoravec (approved):

PR #3125, opened 2023-01-31
[tags] Add insights specific tags

ryan-blakley:
* Added missing tags for plugins to match insights spec names, so they can be queried by insights when loading an archive. This will aid in spec loading via the manifest instead of defining the specs individually as we do now.

Signed-off-by: Ryan Blakley <rblakley@redhat.com>

---

### DIFF
--- a/sos/report/plugins/apache.py
+++ b/sos/report/plugins/apache.py
@@ -38,10 +38,10 @@ class Apache(Plugin):
     def setup(self):
         # collect list of installed modules and verify config syntax.
         self.add_cmd_output([
-            "apachectl -M",
             "apachectl -S",
             "apachectl -t"
         ], cmd_as_tag=True)
+        self.add_cmd_output("apachectl -M", tags="httpd_M")
 
         # Other plugins collect these files;
         # do not collect them here to avoid collisions in the archive paths.
@@ -81,10 +81,10 @@ class RedHatApache(Apache, RedHatPlugin):
     def setup(self):
 
         self.add_file_tags({
-            ".*/access_log": 'httpd_access_log',
-            ".*/error_log": 'httpd_error_log',
-            ".*/ssl_access_log": 'httpd_ssl_access_log',
-            ".*/ssl_error_log": 'httpd_ssl_error_log'
+            "/var/log/httpd/access_log": 'httpd_access_log',
+            "/var/log/httpd/error_log": 'httpd_error_log',
+            "/var/log/httpd/ssl_access_log": 'httpd_ssl_access_log',
+            "/var/log/httpd/ssl_error_log": 'httpd_ssl_error_log'
         })
 
         super(RedHatApache, self).setup()
--- a/sos/report/plugins/auditd.py
+++ b/sos/report/plugins/auditd.py
@@ -27,11 +27,12 @@ class Auditd(Plugin, IndependentPlugin):
             "/etc/audit/plugins.d/",
             "/etc/audisp/",
         ])
-        self.add_cmd_output([
-            "ausearch --input-logs -m avc,user_avc,fanotify -ts today",
-            "auditctl -s",
-            "auditctl -l"
-        ])
+
+        self.add_cmd_output(
+            "ausearch --input-logs -m avc,user_avc,fanotify -ts today"
+        )
+        self.add_cmd_output("auditctl -l", tags="auditctl_rules")
+        self.add_cmd_output("auditctl -s", tags="auditctl_status")
 
         config_file = "/etc/audit/auditd.conf"
         log_file = "/var/log/audit/audit.log"
--- a/sos/report/plugins/foreman_installer.py
+++ b/sos/report/plugins/foreman_installer.py
@@ -88,10 +88,10 @@ class RedHatForemanInstaller(ForemanInstaller, RedHatPlugin):
     def setup(self):
 
         self.add_file_tags({
-            '/var/log/foreman-installer/satellite.log.*':
-                ['insights_satellite_log' 'satellite_installer_log'],
-            '/var/log/foreman-installer/capsule.log.*':
-                ['insights_capsule_log' 'capsule_installer_log'],
+            '/var/log/foreman-installer/satellite.log':
+                ['foreman_satellite_log' 'satellite_installer_log'],
+            '/var/log/foreman-installer/capsule.log':
+                ['capsule_log' 'capsule_installer_log'],
         })
 
         super(RedHatForemanInstaller, self).setup()
--- a/sos/report/plugins/lvm2.py
+++ b/sos/report/plugins/lvm2.py
@@ -63,7 +63,7 @@ class Lvm2(Plugin, IndependentPlugin):
 
         self.add_cmd_output(
             "vgdisplay -vv %s" % lvm_opts_foreign,
-            root_symlink="vgdisplay"
+            root_symlink="vgdisplay", tags="vgdisplay"
         )
 
         pvs_cols = 'pv_mda_free,pv_mda_size,pv_mda_count,pv_mda_used_count'
@@ -72,12 +72,16 @@ class Lvm2(Plugin, IndependentPlugin):
         vgs_cols = vgs_cols + ',' + 'vg_tags,systemid'
         lvs_cols = ('lv_tags,devices,lv_kernel_read_ahead,lv_read_ahead,'
                     'stripes,stripesize')
-        self.add_cmd_output([
-            "vgscan -vvv %s" % lvm_opts,
-            "pvscan -v %s" % lvm_opts,
+        self.add_cmd_output("lvs -a -o +%s %s" % (lvs_cols, lvm_opts_foreign),
+                            tags="lvs_headings")
+        self.add_cmd_output(
             "pvs -a -v -o +%s %s" % (pvs_cols, lvm_opts_foreign),
-            "vgs -v -o +%s %s" % (vgs_cols, lvm_opts_foreign),
-            "lvs -a -o +%s %s" % (lvs_cols, lvm_opts_foreign)
+            tags="pvs_headings")
+        self.add_cmd_output("vgs -v -o +%s %s" % (vgs_cols, lvm_opts_foreign),
+                            tags="vgs_headings")
+        self.add_cmd_output([
+            "pvscan -v %s" % lvm_opts,
+            "vgscan -vvv %s" % lvm_opts
         ])
 
         self.add_copy_spec("/etc/lvm")
--- a/sos/report/plugins/networking.py
+++ b/sos/report/plugins/networking.py
@@ -242,6 +242,24 @@ class Networking(Plugin):
                         ns_cmd_prefix + "ethtool -S %(dev)s"
                     ], devices=_devs['ethernet'], priority=50, subdir=_subdir)
 
+        self.add_cmd_tags({
+            "ethtool [^-].*": "ethtool",
+            "ethtool -S.*": "ethtool_S",
+            "ethtool -T.*": "ethtool_T",
+            "ethtool -a.*": "ethtool_a",
+            "ethtool -c.*": "ethtool_c",
+            "ethtool -g.*": "ethtool_g",
+            "ethtool -i.*": "ethtool_i",
+            "ethtool -k.*": "ethtool_k",
+            "ip -d address": "ip_addr",
+            "ip -s -s neigh show": "ip_neigh_show",
+            "ip route show table all": "iproute_show_table_all",
+            "ip -s -d link": "ip_s_link",
+            "netstat.*-neopa": "netstat",
+            "netstat.*-agn": "netstat_agn",
+            "netstat -s": "netstat_s"
+        })
+
 
 class RedHatNetworking(Networking, RedHatPlugin):
     trace_host = "rhn.redhat.com"
--- a/sos/report/plugins/pulp.py
+++ b/sos/report/plugins/pulp.py
@@ -131,10 +131,13 @@ class Pulp(Plugin, RedHatPlugin):
         self.add_cmd_output(prun, suggest_filename="pulp-running_tasks")
         self.add_cmd_output(csizes, suggest_filename="mongo-collection_sizes")
         self.add_cmd_output(dbstats, suggest_filename="mongo-db_stats")
-        self.add_cmd_output([
-            "qpid-stat -%s --ssl-certificate=%s -b amqps://localhost:5671" %
-            (opt, self.messaging_cert_file) for opt in "quc"
-        ])
+
+        for opt in "quc":
+            self.add_cmd_output(
+                f"qpid-stat -{opt} --ssl-certificate="
+                f"{self.messaging_cert_file} -b amqps://localhost:5671",
+                tags=f"qpid_stat_{opt}")
+
         self.add_cmd_output(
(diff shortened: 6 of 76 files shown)


sos/report/plugins/__init__.py
```diff
@@ -2982,6 +2982,10 @@ def add_journal(self, units=None, boot=None, since=None, until=None,
         self._add_cmd_output(cmd=journal_cmd, timeout=timeout,
                              sizelimit=log_size, pred=pred, tags=tags,
                              priority=priority)
+        self.add_cmd_tags({
+            "journalctl --no-pager": "journal_all",
+            "journalctl --no-pager --boot": "journal_since_boot"
+        })
```
TurboTurtle, line 2988: These should already be applied:

```
{
                            "command": "journalctl",
                            "parameters": [
                                "--no-pager",
                                "",
                                "--boot",
                                ""
                            ],
                            "exec": "journalctl --no-pager  --boot ",
                            "filepath": "sos_commands/logs/journalctl_--no-pager_--boot",
                            "truncated": false,
                            "return_code": 0,
                            "priority": 10,
                            "start_time": 1675189264.6049192,
                            "end_time": 1675189265.145264,
                            "run_time": 0.5403447151184082,
                            "tags": [
                                "journal_since_boot"
                            ]
                        },
```

and

```
{
                            "command": "journalctl",
                            "parameters": [
                                "--no-pager",
                                ""
                            ],
                            "exec": "journalctl --no-pager ",
                            "filepath": "sos_commands/logs/journalctl_--no-pager",
                            "truncated": false,
                            "return_code": 0,
                            "priority": 100,
                            "start_time": 1675189266.1241925,
                            "end_time": 1675189267.3110363,
                            "run_time": 1.1868438720703125,
                            "tags": [
                                "journal_full"
                            ]
                        }
```

sos/report/plugins/apache.py
```diff
@@ -81,10 +81,10 @@ class RedHatApache(Apache, RedHatPlugin):
...
 
         self.add_file_tags({
-            ".*/access_log": 'httpd_access_log',
-            ".*/error_log": 'httpd_error_log',
-            ".*/ssl_access_log": 'httpd_ssl_access_log',
-            ".*/ssl_error_log": 'httpd_ssl_error_log'
+            "/var/log/httpd/access_log": 'httpd_access_log',
+            "/var/log/httpd/error_log": 'httpd_error_log',
+            "/var/log/httpd/ssl_access_log": 'httpd_ssl_access_log',
+            "/var/log/httpd/ssl_error_log": 'httpd_ssl_error_log'
```
TurboTurtle, line 87: Ack

sos/report/plugins/auditd.py
```diff
@@ -19,19 +19,20 @@ class Auditd(Plugin, IndependentPlugin):
     packages = ('audit',)
 
     def setup(self):
+        self.add_copy_spec("/etc/audit/auditd.conf", tags="auditd_conf")
         self.add_copy_spec([
-            "/etc/audit/auditd.conf",
```
TurboTurtle, line 23: Should already be tagged as such:

```
{
                            "specification": "/etc/audit/auditd.conf",
                            "files_copied": [
                                "etc/audit/auditd.conf"
                            ],
                            "tags": [
                                "auditd_conf"
                            ]
                        },
```

sos/report/plugins/auditd.py
```diff
@@ -19,19 +19,20 @@ class Auditd(Plugin, IndependentPlugin):
...
-            "ausearch --input-logs -m avc,user_avc,fanotify -ts today",
-            "auditctl -s",
-            "auditctl -l"
-        ])
+
+        self.add_cmd_output(
+            "ausearch --input-logs -m avc,user_avc,fanotify -ts today"
+        )
+        self.add_cmd_output("auditctl -l", tags="auditctl_rules")
+        self.add_cmd_output("auditctl -s", tags="auditctl_status")
```
TurboTurtle, line 35: Ack

sos/report/plugins/foreman_installer.py
```diff
@@ -88,10 +88,10 @@ class RedHatForemanInstaller(ForemanInstaller, RedHatPlugin):
     def setup(self):
 
         self.add_file_tags({
-            '/var/log/foreman-installer/satellite.log.*':
-                ['insights_satellite_log' 'satellite_installer_log'],
-            '/var/log/foreman-installer/capsule.log.*':
-                ['insights_capsule_log' 'capsule_installer_log'],
```
TurboTurtle, line 94: Would it be better to make this `(.*)?` so that we also tag rotated logs, or is that not relevant?

sos/report/plugins/lvm2.py
```diff
@@ -72,12 +72,16 @@ def setup(self):
         vgs_cols = vgs_cols + ',' + 'vg_tags,systemid'
         lvs_cols = ('lv_tags,devices,lv_kernel_read_ahead,lv_read_ahead,'
                     'stripes,stripesize')
+        self.add_cmd_output("lvs -a -o +%s %s" % (lvs_cols, lvm_opts_foreign),
+                            tags="lvs")
+        self.add_cmd_output(
+            "pvs -a -v -o +%s %s" % (pvs_cols, lvm_opts_foreign),
+            tags="pvs")
+        self.add_cmd_output("vgs -v -o +%s %s" % (vgs_cols, lvm_opts_foreign),
+                            tags="vgs")
```
TurboTurtle, line 81: Could we make these tags a bit more specific? I realize this would require _also_ updating the insights specs, but I'm hesitant here for the case of future-proofing where if a new collection is added it could be confusing to have a generic "lvs" tag for example not be intuitively linked to either.

sos/report/plugins/networking.py
```diff
@@ -242,6 +242,24 @@ def setup(self):
...
+            "ethtool -i.*": "ethtool_i",
+            "ethtool -k.*": "ethtool_k",
+            "ip -d address": "ip_addr",
+            "ip -s -s neigh show": "ip_neigh_show",
+            "ip route show table all": "iproute_show_table_all",
+            "ip -s -d link": "ip_s_link",
+            "netstat.*-neopa": "netstat",
+            "netstat.*-agn": "netstat_agn",
+            "netstat -s": "netstat_s"
+        })
```
TurboTurtle, line 261: Ack

sos/report/plugins/systemd.py
```diff
@@ -24,36 +24,36 @@ class Systemd(Plugin, IndependentPlugin):
...
             "systemd-delta",
             "systemd-analyze",
             "systemd-analyze blame",
             "systemd-analyze dump",
-            "systemd-inhibit --list",
-            "journalctl --list-boots",
-            "ls -lR /lib/systemd"
-        ])
+            "systemd-inhibit --list"
+        ], cmd_as_tag=True)
```
TurboTurtle, line 56: Ack

sos/report/plugins/tomcat.py
```diff
@@ -38,6 +38,12 @@ def setup(self):
         else:
             self.add_copy_spec("/var/log/tomcat*/*")
 
+        self.add_file_tags({
+            "/etc/tomcat*/web.xml": "tomcat_web_xml",
+            "/var/log/tomcat*/catalina.out": "catalina_out",
+            "/var/log/tomcat*/catalina*.log": "catalina_server_log"
```
TurboTurtle, line 44: These would need `tomcat.*/` or `tomcat(.*)?/`

TurboTurtle (commented):
There's a number of changes that are for files like "tag `/etc/foo.conf` as `foo_conf`", which should already happen automatically. Unless there's an environment you're not seeing that in?

Other than that, I think this looks fine. I like the preference of `add_(cmd|file)_tags()` and `cmd_as_tag` over splitting out new `add_$thing()` calls, but there's a lot of areas where we simply can't avoid it, but it doesn't seem to hurt readability too much.

As we align tags<>specs more, I wonder if we can come up with a more automated method for these tags?

sos/report/plugins/__init__.py
ryan-blakley: Hmm let me look into that again, the bottom one is off it's journal_full but the spec name is journal_all.

sos/report/plugins/auditd.py
ryan-blakley: Oh are files tag automatically now?

sos/report/plugins/foreman_installer.py
ryan-blakley: So the current spec is only the first file, it isn't setup for glob collection of all the files. So I removed the .* for any tag that had that at the end but insights didn't collect via glob all the files.

sos/report/plugins/__init__.py
TurboTurtle: Ah! I guess I went cross-eyed there.

If it's `journal_all`, then I'd say we should just make the `tags` argument for that one a list, since it's already broken out into its own `add_journal()` call.

sos/report/plugins/auditd.py
TurboTurtle: Yeah, but it's limited to the file name, not the whole path. So the above becomes `auditd_conf` and not `audit_auditd_conf`.

So, for the changes that make directory names part of the tag, keep those.

sos/report/plugins/foreman_installer.py
TurboTurtle: Ack, no need to tag rotated files then if the tag won't be useful on them.

sos/report/plugins/lvm2.py
ryan-blakley: Hmm let me see if it's possible to change the spec name, it appears to be a sos specific spec and not an insights collection spec name.

sos/report/plugins/auditd.py
ryan-blakley: Ah okay thanks I'll work on correcting those.

sos/report/plugins/lvm2.py
ryan-blakley: @TurboTurtle Would lvs_sos, vgs_sos, pvs_sos be a good enough tag or would you like more description in the tag/spec name?

TurboTurtle: Yeah, that's totally fine.

sos/report/plugins/__init__.py
ryan-blakley: Updated and removed the journal_since_boot and fixed the journal_all.

sos/report/plugins/lvm2.py
ryan-blakley: Updated the names, and put in a PR to rename the specs in core.

sos/report/plugins/tomcat.py
ryan-blakley: Fixed.

sos/report/plugins/__init__.py
```diff
@@ -2982,6 +2982,9 @@ def add_journal(self, units=None, boot=None, since=None, until=None,
         self._add_cmd_output(cmd=journal_cmd, timeout=timeout,
                              sizelimit=log_size, pred=pred, tags=tags,
                              priority=priority)
+        self.add_cmd_tags({
+            "journalctl --no-pager $": "journal_all",
+        })
```
TurboTurtle, line 2987: We can just add the new tag directly:

```diff
@@ -63,7 +63,8 @@ class Logs(Plugin, IndependentPlugin):
         journal = any([self.path_exists(self.path_join(p, "log/journal/"))
                        for p in ["/var", "/run"]])
         if journal and self.is_service("systemd-journald"):
-            self.add_journal(since=since, tags='journal_full', priority=100)
+            self.add_journal(since=since, tags=['journal_full', 'journal_all'],
+                             priority=100)
             self.add_journal(boot="this", since=since,
                              tags='journal_since_boot')
             self.add_journal(boot="last", since=since,
```

TurboTurtle (commented):
One note on the journal, rest of it LGTM.

sos/report/plugins/__init__.py
ryan-blakley: Updated

sos/report/plugins/pulp.py
```diff
@@ -131,10 +131,18 @@ def setup(self):
...
+            self.messaging_cert_file,
+            tags="qpid_stat_c")
+        self.add_cmd_output(
+            "qpid-stat -q --ssl-certificate=%s -b amqps://localhost:5671" %
+            self.messaging_cert_file,
+            tags="qpid_stat_q")
+        self.add_cmd_output(
+            "qpid-stat -u --ssl-certificate=%s -b amqps://localhost:5671" %
+            self.messaging_cert_file,
+            tags="qpid_stat_u")
```
pmoravec, line 145: This can be de-duplicated:

    for opt in "quc":
         self.add_cmd_output(
            f"qpid-stat -{opt} --ssl-certificate={self.messaging_cert_file} -b amqps://localhost:5671",
            tags=f"qpid_stat_{opt}")

(just I think `pep8` would dislike the huge line)

sos/report/plugins/systemd.py
```diff
@@ -24,36 +24,36 @@ class Systemd(Plugin, IndependentPlugin):
...
-            # status --all mostly seems to cover the others.
+            "journalctl --list-boots",
+            "ls -lR /lib/systemd",
+            "systemctl list-dependencies",
+            "systemctl list-jobs",
+            "systemctl list-machines",
             "systemctl list-units",
-            "systemctl list-units --failed",
             "systemctl list-units --all",
-            "systemctl list-unit-files",
```
pmoravec, line 43: We *stop* collecting `"systemctl list-unit-files"`, is that intentional?

sos/report/plugins/virsh.py
```diff
@@ -46,6 +45,9 @@ def setup(self):
         for subcmd in subcmds:
             self.add_cmd_output('%s %s' % (cmd, subcmd), foreground=True)
 
+        self.add_cmd_output("%s list --all" % cmd,
```
pmoravec, line 48: Please add `foreground=True` parameter like before (I *think* the cmd can hang otherwise..?).

sos/report/plugins/systemd.py
ryan-blakley: No it's below, I originally brought all the commands out into separate add_cmd_output calls. When I found out just adding cmd_as_tag works perfectly. I added them all back in alphabetical order, it's an ocd thing I do sometimes. I can undo all the alphabetical ordering if you think it's not needed.

sos/report/plugins/pulp.py
ryan-blakley: Ahh that's a good point, let me work on that change.

sos/report/plugins/systemd.py
pmoravec: Ah, I see. So then there is just a typo :)

-            "systemctl list-unit-files",
+            "systemctl list-units-files",

Whereas `systemctl list-unit-files` does work but `systemctl list-units-files` doesnt.

Ordering does not matter at all :)

ryan-blakley: Good catch, I corrected it.

sos/report/plugins/virsh.py
ryan-blakley: Updated

pmoravec (approved):

PR #3130, opened 2023-02-06
[ovn_central] Remove duplicate commands

fernandoroyosanchez:
When a list of command are sent to run on a container the output is
printed to file on sos_commands/<plugin> folder, but also a symbolic
link is created from the sos_container/<container> file to the
sos_commands/<plugin> file. When a command is duplicated, we get an
exception on the plugin and the rest of commands are not executed.

This patch removes some hardcoded commands, and clean the list from
duplicated commands before sending to the container.

---

### DIFF
--- a/sos/report/plugins/ovn_central.py
+++ b/sos/report/plugins/ovn_central.py
@@ -131,10 +131,6 @@ class OVNCentral(Plugin):
             'ovn-nbctl --no-leader-only show',
             'ovn-nbctl --no-leader-only get-ssl',
             'ovn-nbctl --no-leader-only get-connection',
-            'ovn-nbctl --no-leader-only list loadbalancer',
-            'ovn-nbctl --no-leader-only list Load_Balancer',
-            'ovn-nbctl --no-leader-only list ACL',
-            'ovn-nbctl --no-leader-only list Logical_Switch_Port',
         ]
 
         sbctl_cmds = [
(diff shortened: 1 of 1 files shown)


sos/report/plugins/ovn_central.py
```diff
@@ -131,10 +131,6 @@ def setup(self):
             'ovn-nbctl --no-leader-only show',
             'ovn-nbctl --no-leader-only get-ssl',
             'ovn-nbctl --no-leader-only get-connection',
-            'ovn-nbctl --no-leader-only list loadbalancer',
-            'ovn-nbctl --no-leader-only list Load_Balancer',
-            'ovn-nbctl --no-leader-only list ACL',
-            'ovn-nbctl --no-leader-only list Logical_Switch_Port',
```
pmoravec, line 137: Those commands are not required at all, right?

sos/report/plugins/ovn_central.py
```diff
@@ -161,8 +157,9 @@ def setup(self):
         cmds += sbctl_cmds
 
         # If OVN is containerized, we need to run the above commands inside
-        # the container.
-
+        # the container. Removing duplicates (in case there are) to avoid
+        # failing on collecting output to file on container running commands
+        cmds = list(set(cmds))
```
pmoravec, line 162: ACK to this as a workaround. While `sos` should be clever enough to prevent raising the exception you wrote about. Could you please open a new issue with an artificial reproducer "run sos in container with plugin that collects two identical commands, that raises exception X", to let us fix that later on?

sos/report/plugins/ovn_central.py
fernandoroyosanchez: No, they are table names that appear in the schema analysis a few lines below. They were added a long time ago, when there was no such iteration over the schema table names.

sos/report/plugins/ovn_central.py
fernandoroyosanchez: done! https://github.com/sosreport/sos/issues/3131

TurboTurtle (approved):
Ack here, following the clarification of the command removals being table names.

pmoravec (approved):

PR #3134, opened 2023-02-14
[mutter] Add a new mutter plugin

nielsdg:
Mutter provides useful information about the monitors it knows about, which can be useful when debugging issues related to resolutions, used connectors, etc.

Signed-off-by: Niels De Graef <ndegraef@redhat.com>

---

### DIFF
--- a/sos/report/plugins/mutter.py
+++ b/sos/report/plugins/mutter.py
@@ -1,0 +1,42 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+import os
+
+from pathlib import Path
+from sos.report.plugins import Plugin, IndependentPlugin
+
+
+class Mutter(Plugin, IndependentPlugin):
+
+    short_desc = 'GNOME X11 window manager and Wayland compositor'
+
+    plugin_name = 'mutter'
+    profiles = ('desktop',)
+    packages = ('mutter',)
+
+    def setup(self):
+        # Get the monitors.xml file that persists the monitor setup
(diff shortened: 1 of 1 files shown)


nielsdg:
Converting to draft, since I forgot this runs as root, so we need to figure out a way to get to the session of a specific user

sos/report/plugins/mutter.py
```diff
@@ -0,0 +1,42 @@
...
+
+    short_desc = 'GNOME X11 window manager and Wayland compositor'
+
+    plugin_name = 'mutter'
+    profiles = ('desktop',)
+    packages = ('mutter',)
+
+    def setup(self):
+        # Get the monitors.xml file that persists the monitor setup
+        monitors_xml = os.path.join(Path.home(), '.config', 'monitors.xml')
```
jadahl, line 25: Could fetch `/etc/xdg/monitors.xml` too while at it.

sos/report/plugins/mutter.py
```diff
@@ -0,0 +1,42 @@
...
+from sos.report.plugins import Plugin, IndependentPlugin
+
+
+class Mutter(Plugin, IndependentPlugin):
+
+    short_desc = 'GNOME X11 window manager and Wayland compositor'
+
+    plugin_name = 'mutter'
+    profiles = ('desktop',)
+    packages = ('mutter',)
```
arif-ali, line 21: you may want to add `mutter-common` for ubuntu based OSs

TurboTurtle:
> Converting to draft, since I forgot this runs as root, so we need to figure out a way to get to the session of a specific user

This would likely require a change in `sos_get_command_output()`. We should be able to do this via a `preexec_fn` there, similar to (or rather, in extension of) how we handle chrooting. I don't _think_ there are any concerns about allow a root-only utility to selectively de-escalate itself for this kind of collection, but I'd like to dig into that a little bit further.

arif-ali:
We have a new attributes that can be used to grab as a user now, maybe this can help to improve this PR?

PR #3146, opened 2023-02-24
[ssh/control_persist] Add option to set ControlPath directory

utopiabound:
The directory currently defaults to something under tmp-dir.
Allow this to be set since this is limited to 100 characters total.

---

### DIFF
--- a/sos/collector/__init__.py
+++ b/sos/collector/__init__.py
@@ -121,6 +121,7 @@ class SoSCollector(SoSComponent):
         'ssh_key': '',
         'ssh_port': 22,
         'ssh_user': 'root',
+        'ssh_control_path_dir': '',
         'timeout': 600,
         'transport': 'auto',
         'verify': False,
@@ -405,6 +406,9 @@ class SoSCollector(SoSComponent):
                                  help='Specify a sos preset to use')
         collect_grp.add_argument('--ssh-user',
                                  help='Specify an SSH user. Default root')
+        collect_grp.add_argument('--ssh-control-path-dir',
+                                 help='Override the SSH ControlPath directory.'
+                                      ' Default is under tmp-dir')
         collect_grp.add_argument('--timeout', type=int, required=False,
                                  help='Timeout for sosreport on each node.')
         collect_grp.add_argument('--transport', default='auto', type=str,
--- a/sos/collector/transports/control_persist.py
+++ b/sos/collector/transports/control_persist.py
@@ -95,8 +95,9 @@ class SSHControlPersist(RemoteTransport):
                            "Please update your OpenSSH installation.")
             raise
         self.log_info('Opening SSH session to create control socket')
-        self.control_path = ("%s/.sos-collector-%s" % (self.tmpdir,
-                                                       self.address))
+        self.control_path = ("%s/.sos-collector-%s" %
+                             (self.opts.ssh_control_path_dir or self.tmpdir,
+                              self.address))
         self.ssh_cmd = ''
         connected = False
         ssh_key = ''


pmoravec (approved):

TurboTurtle (commented):
I don't see the use case for a separate option, but perhaps I'm missing something.

> Allow this to be set since this is limited to 100 characters total.

Under what condition? There's no mention of this in the ssh man pages.


We place the controlpath inside the `--tmpdir` because we completely own the tmpdir we're using. Even if we fail the socket cleanup for some reason, we still are able to ensure removing the socket (and thus the open connection) when we cleanup the tmpdir. Moving it outside the tmpdir seems counter-productive, and breaks the SSH recommendation of placing it in a private directory.

utopiabound:
The SSH man page doesn't say it, but control_path gets appended with a control file and then it calls `openssh/misc.c::unix_listener()` which puts the path into `sockaddr_un.sun_path` which is only 108 chars.

TurboTurtle:
Ok, however what actual issue are you seeing as a result of this? Are socket names being truncated if they are over 100 characters? Are connections failing?

Beyond that, I'm still very hesitant to separate this out from `--tmp-dir`. If for nothing else, users can simply use `--tmp-dir` today to get around this issue. Alternatively, we could change the formatting of the socket name.

utopiabound:
> Ok, however what actual issue are you seeing as a result of this? Are socket names being truncated if they are over 100 characters? Are connections failing?
> 

It actually truncates the "wrong" way", It takes the TRAILING 100 characters of the path, (which could be VERY bad in some cases).

> Beyond that, I'm still very hesitant to separate this out from `--tmp-dir`. If for nothing else, users can simply use `--tmp-dir` today to get around this issue. Alternatively, we could change the formatting of the socket name.

If `tmp-dir` is deep enough in `/var/` you can run into this issue, which is where I'm running into this issue.  The path depth is long, and given by an upstream tool, so I can't change it.

TurboTurtle:
> > Ok, however what actual issue are you seeing as a result of this? Are socket names being truncated if they are over 100 characters? Are connections failing?
> 
> It actually truncates the "wrong" way", It takes the TRAILING 100 characters of the path, (which could be VERY bad in some cases).
> 

That strikes me more of an issue for OpenSSH to address. We're leveraging their CLI to say "create a control socket here", and if there are conditions where that doesn't happen - that's a bug within that project.

> > Beyond that, I'm still very hesitant to separate this out from `--tmp-dir`. If for nothing else, users can simply use `--tmp-dir` today to get around this issue. Alternatively, we could change the formatting of the socket name.
> 
> If `tmp-dir` is deep enough in `/var/` you can run into this issue, which is where I'm running into this issue. The path depth is long, and given by an upstream tool, so I can't change it.

What tool is that? Our default tmp dir path lengths are around 18-24 characters depending on the distribution. That leaves 76-82 characters for an FQDN, which would be impressive to overshoot. Providing a way to potentially shorten that seems like a half-measure to me (see above - this wouldn't actually _fix_ the problem at hand), especially if there's a different tool leverage sos that has different defaults.

utopiabound:
> What tool is that? Our default tmp dir path lengths are around 18-24 characters depending on the distribution. That leaves 76-82 characters for an FQDN, which would be impressive to overshoot. Providing a way to potentially shorten that seems like a half-measure to me (see above - this wouldn't actually _fix_ the problem at hand), especially if there's a different tool leverage sos that has different defaults.

The `tmp-dir` is `/var/log/ab-log-agent/clog/<iso-timestap>/` which is 47 + "sos.XXXXXXXX/" (13) + "/.sos-collector-" (16) + rbuf (see `openssh/mux.c::muxserver_listen()`) (16) which only leaves 16 bytes for hostname.

TurboTurtle:
At the end of the day, the fix lies with the [OpenSSH](https://www.openssh.com/report.html) project. Anything we do here is at best a band-aid that tries to side-step the problem, without actually fixing it.

Whichever tool is using `/var/log/ab-log-agent` (is it Apache related?) as a tmp-dir, should consider instead using the default tmpdir and moving the archive as needed afterwards.

pmoravec:
The limitation comes from Linux, rather than from OpenSSH - all BSD-based operating systems seem to have the limitation - see e.g. https://unix.stackexchange.com/questions/367008/why-is-socket-path-length-limited-to-a-hundred-chars/367012#367012 or https://stackoverflow.com/questions/34829600/why-is-the-maximal-path-length-allowed-for-unix-sockets-on-linux-108, or `man 7 unix` directly.

I see two possible workarounds (apart of accepting the PR which would react to the Linux limitation and pre-assume the user is responsible for proper permissions and control of the control-path directory):
- use shorter directory path by "the tool"
- have a symlink like `/var/tmp/<something-short>` to `/var/log/ab-log-agent/clog/<iso-timestap>/` which should save sufficient number of characters, I expect


Or really accept the PR which I am inclined to.

PR #3147, opened 2023-02-27
[presets] drop --verify from RHOCP preset, call "ostree fsck" under plugin specific option

pmoravec:
---

### DIFF
--- a/sos/report/plugins/ostree.py
+++ b/sos/report/plugins/ostree.py
@@ -6,7 +6,7 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.report.plugins import Plugin, IndependentPlugin
+from sos.report.plugins import Plugin, IndependentPlugin, PluginOpt
 
 
 class OSTree(Plugin, IndependentPlugin):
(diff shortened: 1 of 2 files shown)


sos/report/plugins/ostree.py
```diff
@@ -17,6 +17,9 @@ class OSTree(Plugin, IndependentPlugin):
     profiles = ('system', 'sysmgmt', 'packagemanager')
     files = ('/ostree',)
     services = ('ostree-finalize-staged', 'ostree-boot-complete')
+    option_list = [
+        PluginOpt('fsck', default=False, desc='collect ostree fsck')
```
TurboTurtle, line 21: Unittests are failing since `PluginOpt` is not imported in this plugin.

TurboTurtle (commented):
Ack to the option change. Note on the test suite below.

sos/report/plugins/ostree.py
pmoravec: I was blind, thanks.

TurboTurtle (approved):
Ack

PR #3148, opened 2023-02-27
[report,plugin] Control journal size separate from log-size

TurboTurtle:
Historically, journal sizes have been limited to the *higher* of 100MB or `--log-size`. While this had the benefit of potentially controlling both logs and journals with the same option, it was not immediately intuitive to end users and downright prevented collecting less than 100MB of journals.

Address this by separating journal size limiting from `--log-size` by adding a new `--journal-size` option (default 100). This will allow users to individually control journal sizes without any "gotcha" scenarios with relation to general log size limiting.

---

### DIFF
--- a/tests/report_tests/plugin_tests/logs.py
+++ b/tests/report_tests/plugin_tests/logs.py
@@ -30,7 +30,7 @@ class LogsPluginTest(StageOneReportTest):
         self.assertFileGlobInArchive('/var/log/journal/*')
 
 
-class LogsSizeLimitTest(StageTwoReportTest):
+class JournalSizeLimitTest(StageTwoReportTest):
     """Test that journal size limiting is working and is independent of
     --log-size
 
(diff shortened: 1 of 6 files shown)


pmoravec (approved):

tests/report_tests/plugin_tests/logs.py
```diff
@@ -77,7 +82,7 @@ def test_journal_tailed_and_linked(self):
         assert os.path.islink(journ), "Journal in sos_commands/logs is not a symlink"
 
     def test_string_not_in_manifest(self):
-        # we don't want truncated collections appearing in the strings section
+        # we don't want truncated journals appearing in the strings section
         # of the manifest for the plugin
         manifest = self.get_plugin_manifest('logs')
-        self.assertFalse(manifest['strings'])
+        self.assertFalse(any('journalctl' in _ms for _ms in manifest['strings']))
```
pmoravec, line 88: This does work well even if manifest does not contain `strings` section..? (I need to test that).

TurboTurtle: `strings` will always be a section in the manifest, but this does work even if it is unpopulated.

TurboTurtle:
Latest push is to update the version check to next sos release of 4.5.2

pmoravec (approved):

TurboTurtle:
Dropped the string test as mentioned in #3154 - will merge after CI finishes.

PR #3149, opened 2023-02-28
[component] Failsafe for not adding double handlers into logging

jjansky1:
Setting names for all handlers and verifying before adding it to Logger
if the handler is not present already.
That should solve the issue when double handlers are present
due to used preset with set verbosity.

Resolves: #3149

---


sos/component.py
```diff
@@ -384,14 +388,40 @@ def _setup_logging(self):
...
+        if self.soslog is not None:
+            loggers.append(self.soslog)
+        if self.ui_log is not None:
+            loggers.append(self.ui_log)
+        for logger in loggers:
+            for handler in logger.handlers:
+                if handler.name == name:
+                    result = True
+                    break
+        return result
```
TurboTurtle, line 412: This method seems redundant for what is going on in `add_handler()` in the `else` condition.

Since we're already passing the logger to `add_handler()`, we should just inspect that logger directly.

TurboTurtle (commented):
What's the condition where we can double add handlers? Would we not be better off fixing that case instead?

sos/component.py
```diff
@@ -384,14 +388,40 @@ def _setup_logging(self):
...
+
+    def add_handler(self, logger, handler):
+        if not self.handler_present(handler.name):
+            logger.addHandler(handler)
+        else:
+            for h in logger.handlers:
+                if h.name == handler.name:
+                    result = h
```
pmoravec, line 420: Why not directly `logger.removeHandler(h)` here?

sos/component.py
```diff
@@ -384,14 +388,40 @@ def _setup_logging(self):
...
+        if hasattr(self, 'soslog'):
+            loggers.append(self.soslog)
+        if hasattr(self, 'ui_log'):
+            loggers.append(self.ui_log)
+        for logger in loggers:
+            for handler in logger.handlers:
+                if handler.name == name:
+                    result = True
+                    break
+        return result
```
pmoravec, line 412: As we are iterating over 2 loggers / names, isn't it simplier to call

    if name in ['soslog', 'ui_log']:
        return hasattr(self, name)
    return False

or even simplier:

    return name in ['soslog', 'ui_log'] and hasattr(self, name)

?

pmoravec:
> What's the condition where we can double add handlers? Would we not be better off fixing that case instead?

Reproducer:

    # python3 bin/sosreport --add-preset sostestpreset -vvv -o host --batch
    ..
    # python3 bin/sosreport --preset sostestpreset --build
    ..
	    /var/tmp/sosreport-myrhel8-2023-02-28-nywvgzz

	    /var/tmp/sosreport-myrhel8-2023-02-28-nywvgzz

    # python3 bin/sosreport --preset sostestpreset
    ..
	    /var/tmp/sosreport-myrhel8-2023-02-28-frlndcj.tar.xz
    
    
	    /var/tmp/sosreport-myrhel8-2023-02-28-frlndcj.tar.xz
    
     Size	6.43KiB
     Size	6.43KiB
     Owner	root
     Owner	root
     sha256	7de15e22da3ab5919eb26da97132bff09f2743e234dfd13345a82644417ac38d
     sha256	7de15e22da3ab5919eb26da97132bff09f2743e234dfd13345a82644417ac38d
    
    # 

It is a direct consequence of https://github.com/sosreport/sos/pull/2304 and the change the trailing "here is the archive" logs moved from `print` to `logger`.

sos/component.py
jjansky1: above will return string instance, but we need logging.Handler instance, but i can shorten it a bit

jjansky1:
> What's the condition where we can double add handlers? Would we not be better off fixing that case instead?

As pavel wrote this may happen when preset with verbosity (-vvv) is used as below
```
# python3 bin/sosreport --add-preset sostestpreset -vvv -o host --batch
..
# python3 bin/sosreport --preset sostestpreset --build
```

In that case some handlers are registered twice which result in double messages as below, i was not able to find root cause and because of that i proposed this failsafe mechanism to ensure there will be only 1 handler with specific name at the time.

```
 Size	6.43KiB
 Size	6.43KiB
 Owner	root
 Owner	root
 sha256	7de15e22da3ab5919eb26da97132bff09f2743e234dfd13345a82644417ac38d
 sha256	7de15e22da3ab5919eb26da97132bff09f2743e234dfd13345a82644417ac38d
```

pmoravec:
> It's from here - https://github.com/sosreport/sos/blob/main/sos/report/__init__.py#L174
> 
> We need to not call `_setup_logging()` there and instead simply use `setLevel()` on the loggers to set the appropriate logging level.

Ah, nice! *Something* like this seems to work (for this use case):

- in https://github.com/sosreport/sos/blob/main/sos/component.py#L348-L368, make `flog` and `console` class objects (`self.flog` / `self.console`)
- in https://github.com/sosreport/sos/blob/main/sos/report/__init__.py#L174 , do call https://github.com/sosreport/sos/blob/main/sos/component.py#L360-L368

pmoravec:
That shall work but it adds too much redundant branching to the code. What about:

    diff -rup a/sos/component.py b/sos/component.py
    --- a/sos/component.py
    +++ b/sos/component.py
    @@ -337,6 +337,17 @@ class SoSComponent():
             ui_console.setLevel(logging.INFO)
             self.ui_log.addHandler(ui_console)
     
    +    def set_loggers_verbosity(self, verbosity):
    +        if verbosity:
    +            if self.flog:
    +                self.flog.setLevel(logging.DEBUG)
    +            if self.opts.verbosity > 1:
    +                self.console.setLevel(logging.DEBUG)
    +            else:
    +                self.console.setLevel(logging.WARNING)
    +        else:
    +            self.console.setLevel(logging.WARNING)
    +
         def _setup_logging(self):
             """Creates the log handler that shall be used by all components and any
             and all related bits to those components that need to log either to the
    @@ -345,28 +356,20 @@ class SoSComponent():
             # main soslog
             self.soslog = logging.getLogger('sos')
             self.soslog.setLevel(logging.DEBUG)
    -        flog = None
    +        self.flog = None
             if not self.check_listing_options():
                 self.sos_log_file = self.get_temp_file()
    -            flog = logging.StreamHandler(self.sos_log_file)
    -            flog.setFormatter(logging.Formatter(
    +            self.flog = logging.StreamHandler(self.sos_log_file)
    +            self.flog.setFormatter(logging.Formatter(
                     '%(asctime)s %(levelname)s: %(message)s'))
    -            flog.setLevel(logging.INFO)
    -            self.soslog.addHandler(flog)
    +            self.flog.setLevel(logging.INFO)
    +            self.soslog.addHandler(self.flog)
     
             if not self.opts.quiet:
    -            console = logging.StreamHandler(sys.stdout)
    -            console.setFormatter(logging.Formatter('%(message)s'))
    -            if self.opts.verbosity:
    -                if flog:
    -                    flog.setLevel(logging.DEBUG)
    -                if self.opts.verbosity > 1:
    -                    console.setLevel(logging.DEBUG)
    -                else:
    -                    console.setLevel(logging.WARNING)
    -            else:
    -                console.setLevel(logging.WARNING)
    -            self.soslog.addHandler(console)
    +            self.console = logging.StreamHandler(sys.stdout)
    +            self.console.setFormatter(logging.Formatter('%(message)s'))
    +            self.set_loggers_verbosity(self.opts.verbosity)
    +            self.soslog.addHandler(self.console)
             # still log ERROR level message to console, but only setup this handler
             # when --quiet is used, as otherwise we'll double log
             else:
    diff -rup a/sos/report/__init__.py b/sos/report/__init__.py
    --- a/sos/report/__init__.py
    +++ b/sos/report/__init__.py
    @@ -172,7 +172,7 @@ class SoSReport(SoSComponent):
                 print('\nWARNING: It is not recommended to set verbosity via the '
                       'preset as it might have\nunforseen consequences for your '
                       'report logs.\n')
    -            self._setup_logging()
    +            self.set_loggers_verbosity(self.preset.opts.verbosity)
     
             self._set_directories()
     
I.e.:

- add method `set_loggers_verbosity` that sets verbosity of the only two loggers sensitive to the `-v` option(s)
- set these two loggers as class instances to be repeatedly referenced
- call the method when preset contains verbosity flag

Since we know `_setup_logging` has been called (as a must action: we must setup logging at a very early stage any time), both `self.console` and `self.flog` objects should be initialized (but I can enhance the new method by a safety check `if self.console:`)

TurboTurtle:
Closing in favor of #3171

Issue #3070, opened 2022-11-15
[sudo] BINDPW directive is not hidden by sosreport


miztake:
The sudo plugin tries to hide the "bindpw" directive with ********.

https://github.com/sosreport/sos/blob/2556e64b36a1f312e331589dc77146e35c6941e8/sos/report/plugins/sudo.py#L24-L26

However, while the "bindpw" directive is hidden, the "BINDPW" directive is not.

sudoers.ldap(5) says below:
  Configuration options are listed below in upper case but are parsed in a case-independent manner.

In other words, sosreport must hide not only "bindpw" directive, but also "BINDPW" directive.

TurboTurtle:
I'm a little surprised we're not doing case insensitive matches in `do_file_sub()` - but perhaps this is intentional going back to the early days?

If not, then I think passing `re.I` to the `re.subn()` call in `do_file_sub()` might be best.

brozs:
Would this work?
[commit 98288373f6](https://github.com/sosreport/sos/commit/98288373f6ee30e20b7b4af6529f1b1bf258212c)

I'm ready to create the PR.

TurboTurtle:
I was referring more to here  for passing the `re.I` flag - https://github.com/sosreport/sos/blob/main/sos/report/plugins/__init__.py#L1287

That way we do this globally for all plugins, not just the sudo plugin. I _think_ it should be fine to do that, but it'd warrant deeper review with any PR.

pmoravec:
You mean to add to the https://github.com/sosreport/sos/blob/main/sos/report/plugins/__init__.py#L1287 `flags=re.I` unconditionally? That would extend existing scrubbing - maybe without a side effect, hard to say, but I would be careful.

Or adding an option `flags` to the `do_file_sub` method?

TurboTurtle:
The former - make our `do_file_sub()` calls case insensitive from the start. Agreed, it's hard to say if there wouldn't be a negative side effect today, but I think it's worth looking into.


We can, of course, change only the sudo plugin with the proposal above in the meantime while we look into this.

pmoravec:
I checked where we call the method and on a first sight, no evident issue. Rather opposite, we can truncate case-sensitive twins in:

- https://github.com/sosreport/sos/blob/ec9f5676082d7b63c8a2b46ad3780e28f4786cca/sos/report/plugins/monit.py#L45-L69
[...]

PR #3156, opened 2023-03-05
[report] Ignore case when scrubbing via do_file_sub

pmoravec:
File content substitutions done via do_file_sub will newly work case insensitive. This simplifies detection and description of sensitive content, while a risk to obfuscate a regular content is pretty low.

The change is followed by simplification of do_file_sub and do_path_regex_sub calls in individual plugins, and also making them more generic: bfuscated string should not contain explicit case-sensitive key, but \1 rather, to prevent unintentional change of case of the key.

Closes: #3070
Resolves: #3156

---

### DIFF
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -1284,7 +1284,8 @@ class Plugin():
             content = readable.read()
             if not isinstance(content, str):
                 content = content.decode('utf8', 'ignore')
-            result, replacements = re.subn(regexp, subst, content)
+            result, replacements = re.subn(regexp, subst, content,
+                                           flags=re.IGNORECASE)
             if replacements:
                 self.archive.add_string(result, srcpath)
             else:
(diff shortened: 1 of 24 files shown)


pmoravec:
The more eyes review the changes, the better. Esp. requesting a review from @ryan-blakley / @lyarwood for OpenStack changes:

- I fixed a broken scrubbing in https://github.com/sosreport/sos/blob/main/sos/report/plugins/openstack_keystone.py#L110 that evidently lacks `%s` substitution - is the `protect_keys` list the right one? (it is broken since very beginning, it never worked)
- I unified case-insensitively the list in https://github.com/sosreport/sos/blob/main/sos/report/plugins/openstack_instack.py#L116-L137, I run it via a script (that e.g. found `UNDERCLOUD_TUSKAR_PASSWORD` is mentioned there twice, now :) ), so the new list should be correct, but still it is worth to review in detail

lyarwood:
> The more eyes review the changes, the better. Esp. requesting a review from @ryan-blakley / @lyarwood for OpenStack changes:

I've not worked on OpenStack for a while now sorry.

TurboTurtle (approved):
LGTM

PR #3167, opened 2023-03-17
[spec] migrated to SPDX license

sandrobonazzola:
Migrated License field to SPDX format as defined in
https://docs.fedoraproject.org/en-US/legal/update-existing-packages/

---

### DIFF
--- a/sos.spec
+++ b/sos.spec
@@ -4,7 +4,7 @@ Version: 4.5.1
 Release: 1%{?dist}
 Group: Applications/System
 Source0: https://github.com/sosreport/sos/archive/%{name}-%{version}.tar.gz
-License: GPLv2+
+License: GPL-2.0-or-later
 BuildRoot: %{_tmppath}/%{name}-%{version}-%{release}-buildroot
 BuildArch: noarch
 Url: https://github.com/sosreport/sos/


sos.spec
```diff
@@ -1,10 +1,10 @@
 Summary: A set of tools to gather troubleshooting information from a system
 Name: sos
 Version: 4.5.1
-Release: 1%{?dist}
+Release: 2%{?dist}
 Group: Applications/System
 Source0: https://github.com/sosreport/sos/archive/%{name}-%{version}.tar.gz
-License: GPLv2+
+License: GPL-2.0-or-later
```
TurboTurtle, line 7: Ack to this change, but I don't think we need to cut a new release just for this.

sos.spec
```diff
@@ -63,6 +63,7 @@ rm -rf ${RPM_BUILD_ROOT}/usr/config/
 %changelog
 * Wed Mar 01 2023 Jake Hunsaker <jhunsake@redhat.com> = 4.5.1
 - New upstream release
+- migrated to SPDX license
```
TurboTurtle, line 66: Ok, I see some automation is looking for this.

We can make sure to include this in the upcoming 4.5.2 release (March 31) changelog, but we shouldn't add this to the 4.5.1 notes as it wasn't part of that release.

PR #3180, opened 2023-03-31
[tmpfiles] ignoring cleanup on /var/tmp/sosreport*

jjansky1:
tmpfiles relocated to separated folder, added tmpfilesd-sos-rh.conf which ignoring tmpfiles clean on /var/tmp/sosreport*.

This is introduced due to wrong selinux context on some collected files in temporal /var/tmp/sosreport which generate avc denials from tmpfiles or even repeated denials when --build option is used.

Related: RHBZ#1664651
Resolves: #3180

---


jjansky1:
Buids fails because source tar.gz do not yet contain file tmpfiles/tmpfilesd-sos-rh.conf

pmoravec (approved):
Cool, thanks!

sos.spec
```diff
@@ -53,6 +55,7 @@ rm -rf ${RPM_BUILD_ROOT}/usr/config/
 %dir /etc/sos/presets.d
 %dir /etc/sos/extras.d
 %dir /etc/sos/groups.d
+/etc/tmpfiles.d/%{name}.conf
```
TurboTurtle, line 58: I think this needs to be `%{_tmpfilesdir}/%{name}.conf`

TurboTurtle (commented):
Looking over https://fedoraproject.org/wiki/Packaging:Tmpfiles.d, I don't immediately see what is causing the `stat` failure on the source file. Small note below as well.

TurboTurtle (approved):

PR #3187, opened 2023-04-05
[policies] Add CBL-Mariner policy

eric-desrochers:
---

### DIFF
--- a/sos/policies/distros/mariner.py
+++ b/sos/policies/distros/mariner.py
@@ -1,0 +1,45 @@
+# Copyright (C) Eric Desrochers <edesrochers@microsoft.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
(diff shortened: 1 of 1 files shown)


sos/policies/distros/mariner.py
```diff
@@ -0,0 +1,43 @@
+# Copyright (C) Red Hat, Inc. 2019
```
pmoravec, line 1: I guess this is worth amending after copy&paste ;-)

pmoravec:
Apart of the three Flake syntax issues - what would be the purpose of sole policy? Until you add some Plugin class for it, it makes no sense..?

eric-desrochers:
> Apart of the three Flake syntax issues - what would be the purpose of sole policy? Until you add some Plugin class for it, it makes no sense..?

The intent for now is to have a customize message instead of the the generic one when sos report is run.

pmoravec:
ACK to the intent / policy itself. As I checked we do have several similar ones already.

Please fix the Flake errors and I guess also the Copyright comment.

pmoravec (approved):

TurboTurtle (approved):
Ack. My only question would be is there an upload target MSFT customers/support would want to use for this policy?

eric-desrochers:
> Ack. My only question would be is there an upload target MSFT customers/support would want to use for this policy?

Not yet.

PR #3192, opened 2023-04-10
[juju] Add juju integration to sos collect.

jneo8:
Add a new cluster profile and transport called "juju" for `sos collect`. Both the profile and transport are intended to be used on juju managed environments which assumes that `juju` is installed on the machine where `sos collect` is called, and that the juju user has superuser privilege to the current controller.

When using the "juju" cluster profile, the sos reports will be collected from all the applications within the current model by default. If necessary, one can filter the nodes by models / applications / units / machines with cluster options. For example `-c "juju.models=sos" -c "juju.apps=a,b,c"`. Moreover, transport will also be dynamically changed to "juju" when cluster type is juju.

If not using "juju" cluster profile, one can still choose to use the "juju" transport by specifying --transport option. However, not that the --nodes option will be expected to be a comma separated machine IDs , **not** IP addr, since `juju ssh` identifies the ssh target by machine ID. For example, `sos collect --nodes 0,1,2`

Closes: #2668

---


sos/collector/clusters/juju.py
```diff
@@ -0,0 +1,174 @@
...
+        format_option = "--format json"
+        status_cmd = f"{self.cmd} status {model_option} {format_option}"
+        juju_status = None
+        res = self.exec_primary_cmd(status_cmd)
+        if res["status"] == 0:
+            juju_status = json.loads(
+                self._cleanup_juju_output((res["output"]))
+            )
+        else:
+            raise Exception(f"{status_cmd} did not return usable output")
```
pmoravec, line 120: Improvement idea: Isn't something like

    f"'{status_cmd}' returned error: {res['status']}"

more informing error message?

sos/collector/clusters/juju.py
```diff
@@ -0,0 +1,174 @@
...
+
+        if not models:
+            models = [""]  # use current model by default
+
+        nodes = set()
+
+        for model in models:
+            model_info = self._get_model_info(model)
+            if not any(filters.values()):
+                pass
```
pmoravec, line 161: "if not any(filters.values())" then this method returns `[]` - does it make sense to move this test before the `for model in models` loop to skip calling `_get_model_info` / whole cycle?

sos/collector/clusters/juju.py
```diff
@@ -0,0 +1,174 @@
...
+            else:
+                for key, resource in filters.items():
+                    if key in ["apps"]:
```
pmoravec, line 164: `key in ["apps"]` is equivalent to `key == "apps"`, no?

(it took me a while to understand you want to apply different filter method for different key, which is the reason for such longer for cycle)

sos/collector/exceptions.py
```diff
@@ -112,6 +112,14 @@ def __init__(self):
         super(SaltStackMasterUnsupportedException, self).__init__(message)
 
 
+class JujuNotInstalledException(Exception):
+    """Raised when juju is not installed locally"""
+
+    def __init__(self):
+        message = 'Juju is not installed, please ensure you have install juju.'
```
pmoravec, line 119: `s/have install/have installed/` typo ;-)

pmoravec:
All the files lack trailing

    # vim: set et ts=4 sw=4 :

though I am not certain how much we enforce that "comment".. (most of current files do contain it, but definitely not all of them)?

sos/collector/transports/juju.py
```diff
@@ -0,0 +1,81 @@
...
+
+    name = "juju_ssh"
+
+    def _check_juju_installed(self):
+        cmd = "juju version"
+        try:
+            subprocess.check_output(cmd, stderr=subprocess.STDOUT, shell=True)
+        except subprocess.CalledProcessError:
+            self.log_error("Failed to check `juju` version")
+            raise JujuNotInstalledException
```
pmoravec, line 44: `exec_primary_cmd` can't be used as it can't detect `juju` is not installed, am I right?

sos/collector/transports/juju.py
```diff
@@ -0,0 +1,81 @@
...
+        return True
+
+    def _chmod(self, fname):
+        cmd = f"{self.remote_exec} chmod o+r {fname}"
+        try:
+            subprocess.check_output(cmd, stderr=subprocess.STDOUT, shell=True)
+        except subprocess.CalledProcessError:
+            self.log_error(f"Failed to make {fname} world-readable")
+            raise
```
pmoravec, line 53: Same here (`exec_primary_cmd`)..?

sos/collector/transports/juju.py
```diff
@@ -0,0 +1,81 @@
...
+    @property
+    def connected(self):
+        return self._connected
+
+    @property
+    def remote_exec(self):
+        model, unit = self.address.split(":")
+        model_option = f"-m {model}" if model else ""
+        target_option = unit
+        option = f"{model_option} {target_option}"
```
pmoravec, line 72: Can't this be merged to 1 command..?

sos/collector/transports/juju.py
```diff
@@ -0,0 +1,81 @@
...
+        return f"juju ssh {option} sudo"
+
+    def _retrieve_file(self, fname, dest):
+        self._chmod(fname)  # juju scp need the archive to be world-readable
```
pmoravec, line 76: nitpick: `s/need/needs/`

sos/collector/clusters/juju.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def _get_model_info(self, model_name):
+        juju_status = self._execute_juju_status(model_name)
+
+        index = collections.defaultdict(dict)
+        self._add_principals_to_index(index, juju_status, model_name)
+        self._add_subordinates_to_index(index, juju_status, model_name)
+        self._add_machines_to_index(index, juju_status, model_name)
+
+        return index
```
pmoravec, line 66: It might be beneficial to have a comment explaining purpose of the method, optionally a short example of "juju status" output and expected `index` dictionary - or at least a pointer to `tests/unittests/juju/data/juju_output_sos.json` :)

tests/unittests/juju/data/juju_output_sos.json
```diff
@@ -0,0 +1 @@
...since":"07 Apr 2023 14:41:10+08:00"},"persistent":false,"attachments":{"units":{"ubuntu/2":{"machine":"3","location":"/srv/data","life":"alive"}}}}},"filesystems":{"0/0":{"provider-id":"0/0","storage":"files/0","Attachments":{"machines":{"0":{"mount-point":"/srv/data","read-only":false,"life":"alive"}},"units":{"ubuntu/0":{"machine":"0","location":"/srv/data","life":"alive"}}},"pool":"rootfs","size":12464,"life":"alive","status":{"current":"attached","since":"06 Apr 2023 11:34:01+08:00"}},"2/1":{"provider-id":"2/1","storage":"files/1","Attachments":{"machines":{"2":{"mount-point":"/srv/data","read-only":false,"life":"alive"}},"units":{"ubuntu/1":{"machine":"2","location":"/srv/data","life":"alive"}}},"pool":"rootfs","size":10062,"life":"alive","status":{"current":"attached","since":"07 Apr 2023 14:41:11+08:00"}},"3/2":{"provider-id":"3/2","storage":"files/2","Attachments":{"machines":{"3":{"mount-point":"/srv/data","read-only":false,"life":"alive"}},"units":{"ubuntu/2":{"machine":"3","location":"/srv/data","life":"alive"}}},"pool":"rootfs","size":10113,"life":"alive","status":{"current":"attached","since":"07 Apr 2023 14:41:10+08:00"}}}},"controller":{"timestamp":"17:04:14+08:00"}}
```
pmoravec, line 1: I guess no data here are sensitive at all (e.g. IP addresses starting `10.*`)?

If it is expected a human can be working with the file (i.e. search for a value there to match with a test failure) and if human format isnt an issue for further parsing, then `json_reformat`-like output would be much more readable. 

I am leaving a decision on the author if it makes sense here (and in the second `json`).

sos/collector/transports/juju.py
jneo8: Yes, we require juju for remote connection. Sorry I'm not quite understand the question here. Can you elaborate more what you though?

jneo8:
> > Hi @TurboTurtle, would it be possible for you to kindly review our PR? Thanks!
> 
> I'm out of the country on PTO thru the end of next week. I have it on my short list to review this when I'm back in the saddle on the 24th. @pmoravec has some good points so far, and I'll make it a priority for me to review this on Monday the 24th so that this can potentially still make our next release.

I'm so sorry to bother your vacation, please have nice days.

tests/unittests/juju/data/juju_output_sos.json
jneo8: The data are not sensitive.

sos/collector/exceptions.py
jneo8: updated

sos/collector/clusters/juju.py
jneo8: Refactored this part

jneo8:
> All the files lack trailing
> 
> ```
> # vim: set et ts=4 sw=4 :
> ```
> 
> though I am not certain how much we enforce that "comment".. (most of current files do contain it, but definitely not all of them)?

updated.

sos/collector/transports/juju.py
jneo8: If the file permission is wrong then the `juju scp` will not work.

pmoravec:
Thanks for prompt reactions. Minor comment before I will do another round of code review: please squash the two commits into one (not necessarily now), as refactoring just created code makes little sense. (it is fine to have PR/branch with multiple commits. For example - not mandatory to follow this example - to have one commit for the feature itself and another for adding tests as these commits/changes are relatively separate)

jneo8:
Hi @pmoravec thanks for your quick reply. I did the squash.

pmoravec:
Tests still failing on the `bin/sos collect --list-options`: when using python3.6:

    # bin/sos collect --list-options
    Could not initialize 'collect': Type Dict cannot be instantiated; use dict() instead
    # 

(but manually assigning e.g. `apps: Dict[str, List[str]] = {}` does work there, so why that error..?)

When using python 3.9:

    # bin/sos collect --list-options

    The following clusters are supported by this installation
    Use the short name with --cluster-type or cluster options (-c)

      File "/root/sos-main/bin/sos", line 22, in <module>
        sos.execute()
      File "/root/sos-main/sos/__init__.py", line 193, in execute
        self._component.execute()
      File "/root/sos-main/sos/collector/__init__.py", line 1169, in execute
        self.list_options()
      File "/root/sos-main/sos/collector/__init__.py", line 630, in list_options
        self.clusters[cluster].cluster_name))
    AttributeError: 'Index' object has no attribute 'cluster_name'
    #

sos/collector/clusters/juju.py
```diff
@@ -0,0 +1,209 @@
...
+# See the LICENSE file in the source distribution for further information.
+
+import json
+import re
+
+from sos.collector.clusters import Cluster
+
+
+def _parse_option_string(strings=None):
+    """Parse commad separated string."""
```
TurboTurtle, line 18: Nitpick: typo. Should be `comma`, correct?

sos/collector/clusters/juju.py
```diff
@@ -0,0 +1,209 @@
...
+
+        Here are couple helper functions to parse the juju principals units,
+        subordinate units and machines.
+        """
+        juju_status = self._execute_juju_status(model_name)
+
+        index = _get_index(model_name=model_name)
+        index.add_principals(juju_status)
+        index.add_subordinates(juju_status)
+        index.add_machines(juju_status)
```
TurboTurtle, line 139: `juju_status` can be `None` - but these methods all assume the output was successfully processed by `json.loads()`, which would cause an exception. Granted, this exception _would_ currently be caught by the existing code flow, but perhaps we should provide a more intuitive message here than the generic `TypeError` we'd currently get. 

Alternatively, if `juju_status` remains `None` as part of `_execute_juju_status`, we should abort right then and there.

sos/collector/transports/juju.py
```diff
@@ -0,0 +1,83 @@
...
+    @property
+    def connected(self):
+        return self._connected
+
+    @property
+    def remote_exec(self):
+        model, target_option = self.address.split(":")
+        model_option = f"-m {model}" if model else ""
+        option = f"{model_option} {target_option}"
+        return f"juju ssh {option} sudo"
```
TurboTurtle, line 72: Has this been tested as a non-root connecting user? Our current flows will automatically leverage `sudo` where needed, so I think this could result in a command of `juju ssh sudo sudo <cmd>...`.

TurboTurtle (commented):
Looks good for the most part, only real note of substance is about `juju_status` handling as part of `_get_model_info()`

TurboTurtle:
> I'm so sorry to bother your vacation, please have nice days.

No worries at all :)

pmoravec (approved):
(approving up to Jake's comments)

sos/collector/transports/juju.py
jneo8: One part I am not very sure, when will sos leverage `sudo`?
Because we didn't encounter this situation in our testing. Usually the user of juju cli is not root, and the remote juju agent will be root.

TurboTurtle: `sudo` is used if we're connecting as a non-root user and `need_root=True` is passed to `run_command()` for nodes.

On a second review, I think we're skirting the issue here since `ssh_user` is defaulting to `root`, and this transport is just ignoring that option.

I think I'm ok with baking the use of `sudo` directly into command execution, but it's something we'll want to be mindful of once this gets released to end users.

jneo8: Sorry I gave wrong information in my previous comment.

Since the default ssh_user for juju is ubuntu, we need `--ssh-user ubuntu` to successful run collect. I want to provide this option as default so user won't need to provide it every time but I couldn't find a way to add it.

So one simple way is we force add sudo in `remote_exec`, which is fine in juju because the user have nopasswd sudo permission. And we cheat sos the user is root. Which is not a pretty way.

For example:
- if I run with `--ssh-user` `sos collect --cluster-type juju --no-local -c "juju.apps=ubuntu" --ssh-user ubuntu`
- The command run on remote machine is `juju ssh  0 sudo '/bin/bash -c '"'"'sudo -S /usr/bin/sosreport --batch  --chroot auto'"'"''`
- Which is still work to collect.

I think we can discuss is there any more clean way.

sos/collector/clusters/juju.py
jneo8: updated

sos/collector/clusters/juju.py
jneo8: updated, add error handling to `json.loads`

sos/collector/transports/juju.py
TurboTurtle: Hmm - ok, I understand better now.

Right now, we rely solely on the `ssh_user` value to detect if we need to leverage sudo (or, `su -c` if we instead want to directly leverage `root` when `--become` is used). [This is set here in commons](https://github.com/sosreport/sos/blob/main/sos/collector/__init__.py#L535). This is mainly due to the fact that we use the same user across all connected nodes, so running something akin to `whoami` on every node could be seen as superfluous since we allegedly already have the information we need.

A wrapper that automatically assumes its own specific user as well throws that off a bit, but it should be fairly straight forward to handle. We have a couple options, I think:

1. Do a `whoami` or similar check on each node during node `SosNode` initialization. Not a huge fan of this due to the reason above, but it would be easy and serviceable.
2. Within `JujuSSH`, add logic to `remote_exec` to handle the `need_root` parameter for the calling `run_command()`. This is technically possible, but likely very fragile and generally hacky.
3. Accept the default `need_sudo` setting from commons, but allow a node to override this based on the loaded transport. Basically we could add something like `default_user` to transports (default `''` or `None` or what have you) and if it's set, we align that like so:

```python
class SosNode():

def __init__(self, address, commons, password=None, local_sudo=None,
             load_facts=True):
    [...]
    self._transport = self._load_remote_transport(commons)
    if self._transport.default_user:
        self.need_sudo = self._transport.default_user != 'root'
```

I haven't tested the above, but something like that should suffice. I like this approach more because it allows an easy entry point to control this for current and future transports, not just specifically juju.

jneo8: This is works for juju. PR updated.

sos/collector/clusters/juju.py
```diff
@@ -0,0 +1,214 @@
...
+    def _execute_juju_status(self, model_name):
+        model_option = f"-m {model_name}" if model_name else ""
+        format_option = "--format json"
+        status_cmd = f"{self.cmd} status {model_option} {format_option}"
+        res = self.exec_primary_cmd(status_cmd)
+        if not res["status"] == 0:
+            raise Exception(f"'{status_cmd}' returned error: {res['status']}")
+        juju_json_output = self._cleanup_juju_output((res["output"]))
+
+        juju_status = None
```
pmoravec, line 152: I *think* this line is redundant (either the command in `try` succeeds and assigns a value to the variable, or we raise an exception and no return value matters?). But let be sure..

sos/collector/clusters/juju.py
```diff
@@ -0,0 +1,214 @@
...
+                    self.units[unit] = [node]
+                    self.machines[machine] = [node]
+                    nodes.append(node)
+
+                self.apps[app] = nodes
+
+        def add_subordinates(self, juju_status):
+            """Add subordinates to index.
+
+            Since subordinates does not have units they need to be
```
pmoravec, line 66: A nitpick: typo "subordinates does not have" :)

sos/collector/sosnode.py
```diff
@@ -75,6 +77,10 @@ def __init__(self, address, commons, password=None, local_sudo=None,
         self.soslog = logging.getLogger('sos')
         self.ui_log = logging.getLogger('sos_ui')
         self._transport = self._load_remote_transport(commons)
+        # Overwrite need_sudo if transports default_user
+        # is set and is not root.
+        if self._transport.default_user:
+            self.need_sudo = self._transport.default_user != 'root'
```
pmoravec, line 83: We might split this change and in [sos/collector/transports/__init__.py](https://github.com/sosreport/sos/pull/3192/files#diff-073bf5ff1b0f5ea4cb94d7440664fac5e05e496601ece19838dead00e0912859) into a separate commit, having this change as a generic change & preparation for the juju transport - but I am OK with this.

sos/collector/clusters/juju.py
```diff
@@ -0,0 +1,214 @@
...
+        return index
+
+    def _execute_juju_status(self, model_name):
+        model_option = f"-m {model_name}" if model_name else ""
+        format_option = "--format json"
+        status_cmd = f"{self.cmd} status {model_option} {format_option}"
+        res = self.exec_primary_cmd(status_cmd)
+        if not res["status"] == 0:
+            raise Exception(f"'{status_cmd}' returned error: {res['status']}")
+        juju_json_output = self._cleanup_juju_output((res["output"]))
```
pmoravec, line 150: Redundant pair of parentheses or does the `(res["output"])` have a semantic reason? (I expect the command does work well, just curious..)

pmoravec (approved):
ACK even to the current version of PR, technically nothing is wrong.

tests/unittests/juju/juju_transports_test.py
```diff
@@ -0,0 +1,85 @@
...
+    def test_check_juju_installed_true(self, mock_subprocess_check_output):
+        """Return True if juju is installed."""
+        result = self.juju_ssh._check_juju_installed()
+        assert result
+
+    @patch("sos.collector.transports.juju.subprocess.check_output")
+    def test_chmod(self, mock_subprocess_check_output):
+        self.juju_ssh._chmod(fname="file_abc")
+        mock_subprocess_check_output.assert_called_with(
+            f"{self.juju_ssh.remote_exec} chmod o+r file_abc",
```
pmoravec, line 55: Ah this needs to be fixed - as you changed `_chmod` method by adding `sudo`, this fails on:

    [stdout] AssertionError: Expected call: check_output('juju ssh -m model_abc unit_abc chmod o+r file_abc', shell=True, stderr=-2)
    [stdout] Actual call: check_output('juju ssh -m model_abc unit_abc sudo chmod o+r file_abc', shell=True, stderr=-2)

tests/unittests/juju/juju_transports_test.py
```diff
@@ -0,0 +1,85 @@
...
+        "sos.collector.transports.juju.JujuSSH._check_juju_installed",
+        return_value=True,
+    )
+    def test_connect(self, mock_result):
+        self.juju_ssh.connect(password=None)
+        assert self.juju_ssh.connected
+
+    def test_remote_exec(self):
+        assert (
+            self.juju_ssh.remote_exec == "juju ssh -m model_abc unit_abc sudo"
```
pmoravec, line 70: .. and the missing `sudo` is also here.

TurboTurtle (commented):
Ack, up to the unittest failures highlight by Pavel. Once those are resolved, I think we're good to merge this.

TurboTurtle (approved):
LGTM

PR #3194, opened 2023-04-12
[zfs] add collection of files from /proc/spl

rafalop:
Add collection of files under /proc/spl which contain useful statistics and data for debugging and analysing zfs.

---


pmoravec:
I guess there is no special file (like, say, `/dev/zero` or `/sys/kernel/debug/tracing/trace_pipe`) whose collection would be endless..?

rafalop:
Not to my knowledge, but to be safe (eg. in case such a special file is added in future) I will adjust and use explicit filenames.

pmoravec (approved):

PR #3196, opened 2023-04-18
[ultrapath] new-plugin to collect data of Huawei UltraPath multipathing solution

Nitn-Yewale:
---

### DIFF
--- a/sos/report/plugins/ultrapath.py
+++ b/sos/report/plugins/ultrapath.py
@@ -1,0 +1,40 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, RedHatPlugin
+
+
+class UltraPath(Plugin, RedHatPlugin):
+
+    short_desc = 'HUAWEI UltraPath'
+
+    plugin_name = 'ultrapath'
+    profiles = ('storage', 'hardware')
+    packages = ('UltraPath',)
+    kernel_mods = ('nxup', 'nxupext_a')
+
+    def setup(self):
+        """ Huawei UltraPath specific information - commands
+        """
+        self.add_cmd_output([
+            "upadm show version",
+            "upadm show connectarray",
+            "upadm show option",
+            "upadm show upconfig",
+            "upadm show diskarray",
+            "upadmin show vlun",
+        ])
+
+        result = self.collect_cmd_output('upadm show path')
+        if result['status'] == 0:
+            for line in result['output'].splitlines():
+                if line.startswith("Array ID :"):
+                    self.add_cmd_output("upadm show lun array=%s" %
+                                        line.split(':')[1].strip())
+
+# vim: set et ts=4 sw=4 :


Nitn-Yewale:
The commands mentioned in the plugin are from third party utility and hence are not tested however have been copied from a valid documentation - https://support.huawei.com/enterprise/en/doc/EDOC1100287084

Need help with reviewing the commands and also to check whether any more commands are needed or if existing commands needs to be removed.

It would be helpful to add this plugin as the data collected would help understanding the storage layout and mappings in OS created by ultrapath there by helping in troubleshooting.

nkshirsagar:
Could this plugin also be applicable for debian/ubuntu?

Nitn-Yewale:
Hi Nikhil,

https://support.huawei.com/enterprise/en/doc/EDOC1100287084 document says the commands are for Linux. Commands look valid as per the doc however I did not had access to the Ultrapath package to test these commands in any distro.

I would also need help with testing commands related to Ultrapath-Nvme - https://support.huawei.com/enterprise/en/doc/EDOC1100287085/ac1d9d45/logging-in-to-the-cli

We see there is command `#upadmin_plus` which provides a shell to run the commands. And we have command such as ` # upadmin_plus show version ` to perform the same task as with shell version of the commands.

As such, could we add commands such as `#upadmin_plus show path` for nvme arrays. OR above mentioned `upadm` command would collect the data that is being collected with `#upadmin_plus show path`

Reference :

https://support.huawei.com/enterprise/en/doc/EDOC1100287085/3d19b81e/querying-physical-path-information

Thank you,

sos/report/plugins/ultrapath.py
```diff
@@ -0,0 +1,40 @@
...
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, RedHatPlugin
+
+
+class UltraPath(Plugin, RedHatPlugin):
```
pmoravec, line 12: If / as also Ubuntu should run this plugin, I suggest replacing `RedHatPlugin` by `IndependentPlugin` (also in `import`, to let the plugin available on any distribution.

pmoravec:
> Hi Nikhil,
> 
> https://support.huawei.com/enterprise/en/doc/EDOC1100287084 document says the commands are for Linux. Commands look valid as per the doc however I did not had access to the Ultrapath package to test these commands in any distro.
> 
> I would also need help with testing commands related to Ultrapath-Nvme - https://support.huawei.com/enterprise/en/doc/EDOC1100287085/ac1d9d45/logging-in-to-the-cli
> 
> We see there is command `#upadmin_plus` which provides a shell to run the commands. And we have command such as `# upadmin_plus show version` to perform the same task as with shell version of the commands.

Is that (for) some other component of Ultrapath that should be in some different plugin? I.e. should we collect the `ultrapath` data on some set of systems by default, while these data on (usually similar but sometimes different) systems?

And what particular commands should be collected from the `upadmin_plus` suite?

sos/report/plugins/ultrapath.py
Nitn-Yewale: ack, will make these changes.. thank you.

Nitn-Yewale:
From the documentation it seems `upadmin_plus` is specifically for collecting NVMe devices data  and for NVMe over fabric data where as upadm seem to be  collecting scsi / FC/ iscsi data.

`upadmin_plus` gives the shell where the commands are run as shown in the document - https://support.huawei.com/enterprise/en/doc/EDOC1100287085/ac1d9d45/logging-in-to-the-cli

We do see command such as `upadmin_plus show vlun disk=xxx type=all` but not all commands similar to upadm are shown.

If we could run commands similar to upadm for `upadmin_plus`  we could collect the output from within this same plugin  as all ultrapath commands would be available at one place.

Thank you,

pmoravec:
> From the documentation it seems `upadmin_plus` is specifically for collecting NVMe devices data and for NVMe over fabric data where as upadm seem to be collecting scsi / FC/ iscsi data.

That sounds to me the use cases of "collect `upadmin_plus` data" and "collect `upadm` data" are different and there should be two separate plugins for them?

What is your opinion, @TurboTurtle (once you are back)?

Nitn-Yewale:
https://github.com/sosreport/sos/pull/3199 new pull request with the changes.

Thank you,

pmoravec:
Please replace the `RedHatPlugin` by `IndependentPlugin` and squash the two commits into one, then ACK. Thanks!

PR #3201, opened 2023-04-22
[scsi-block-nvme] move per device data to

Nitn-Yewale:
subdirectories for storage commands

---

### DIFF
--- a/sos/report/plugins/block.py
+++ b/sos/report/plugins/block.py
@@ -19,6 +19,9 @@ class Block(Plugin, IndependentPlugin):
     files = ('/sys/block',)
 
     def setup(self):
+        _subdir1 = "udevadm_block"
+        _subdir2 = "fdisk_parted"
+
         self.add_forbidden_path("/sys/block/*/queue/iosched")
 
         self.add_file_tags({
@@ -50,14 +53,24 @@ class Block(Plugin, IndependentPlugin):
             "/sys/block/loop*/loop/",
         ])
 
-        cmds = [
-            "parted -s %(dev)s unit s print",
+        udevadm_cmds = [
             "udevadm info %(dev)s",
             "udevadm info -a %(dev)s"
         ]
-        self.add_device_cmd(cmds, devices='block', blacklist='ram.*')
+
+        parted_cmds = [
+            "parted -s %(dev)s unit s print"
+        ]
+
+        self.add_device_cmd(udevadm_cmds, devices='block', blacklist='ram.*',
+                            subdir=_subdir1)
+
+        self.add_device_cmd(parted_cmds, devices='block', blacklist='ram.*',
+                            subdir=_subdir2)
+
         self.add_device_cmd("fdisk -l %(dev)s", blacklist="ram.*",
-                            devices="block", tags="fdisk_l_sos")
+                            devices="block", tags="fdisk_l_sos",
+                                    subdir=_subdir2)
 
         lsblk = self.collect_cmd_output("lsblk -f -a -l")
         # for LUKS devices, collect cryptsetup luksDump
--- a/sos/report/plugins/nvme.py
+++ b/sos/report/plugins/nvme.py
@@ -26,24 +26,50 @@ class Nvme(Plugin, IndependentPlugin):
     kernel_mods = ('nvme', 'nvme_core')
 
     def setup(self):
+        _subdir1 = "nvme_smartctl"
+        _subdir2 = "nvme_fw_smart_error_logs"
+
         self.add_copy_spec("/etc/nvme/*")
         self.add_cmd_output([
             "nvme list",
             "nvme list-subsys",
         ])
 
-        cmds = [
+        smartctl_cmds = [
             "smartctl --all %(dev)s",
-            "smartctl --all %(dev)s -j",
-            "nvme list-ns %(dev)s",
+            "smartctl --all %(dev)s -j"
+        ]
+
+        nvme_logs_cmds = [
             "nvme fw-log %(dev)s",
+            "nvme smart-log %(dev)s",
+            "nvme error-log %(dev)s"
+        ]
+
+        nvme_commands = [
+            "nvme list-ns %(dev)s",
             "nvme list-ctrl %(dev)s",
             "nvme id-ctrl -H %(dev)s",
             "nvme id-ns -H %(dev)s",
-            "nvme smart-log %(dev)s",
-            "nvme error-log %(dev)s",
             "nvme show-regs %(dev)s"
         ]
-        self.add_device_cmd(cmds, devices='block', whitelist='nvme.*')
+
+        subdirs = [
+            "nvme_list-ns",
+            "nvme_list-ctrl",
+            "nvme_id-ctrl_-H",
+            "nvme_id-ns_-H",
+            "nvme_show-regs"
+        ]
+
+        self.add_device_cmd(smartctl_cmds, devices='block', whitelist='nvme.*',
+                            subdir=_subdir1)
+
+        self.add_device_cmd(nvme_logs_cmds, devices='block',
+                            whitelist='nvme.*', subdir=_subdir2)
+
+        for cmd, dir in zip(nvme_commands, subdirs):
+            self.add_device_cmd(cmd, whitelist='nvme.*',
+                                devices='block', subdir=dir)
 
 # vim: set et ts=4 sw=4 :
--- a/sos/report/plugins/scsi.py
+++ b/sos/report/plugins/scsi.py
@@ -29,6 +29,9 @@ class Scsi(Plugin, IndependentPlugin):
     profiles = ('storage', 'hardware')
 
     def setup(self):
+        _subdir1 = "udevadm_scsi"
+        _subdir2 = "sg_persists_inq"
+
         self.add_copy_spec([
             "/proc/scsi",
             "/etc/stinit.def",
@@ -61,14 +64,18 @@ class Scsi(Plugin, IndependentPlugin):
             "lsscsi -L"
         ])
 
-        scsi_hosts = glob("/sys/class/scsi_host/*")
-        self.add_device_cmd("udevadm info -a %(dev)s", devices=scsi_hosts)
-
-        self.add_device_cmd([
+        scsi_persist_inq_cmds = [
             "sg_persist --in -k -d %(dev)s",
             "sg_persist --in -r -d %(dev)s",
             "sg_persist --in -s -d %(dev)s",
             "sg_inq %(dev)s"
-        ], devices='block', whitelist=['sd.*'])
+        ]
+
+        scsi_hosts = glob("/sys/class/scsi_host/*")
+        self.add_device_cmd("udevadm info -a %(dev)s",
+                            devices=scsi_hosts, subdir=_subdir1)
+
+        self.add_device_cmd(scsi_persist_inq_cmds, devices='block',
+                            whitelist=['sd.*'], subdir=_subdir2)
 
 # vim: set et ts=4 sw=4 :


Nitn-Yewale:
Through this patch, I would like to propose moving data collected for "per device" to sub directories. With 100s of devices, single directory gets cluttered and it takes time to search the appropriate commands. With this patch, tried naming the sub directories based on command output they collect OR as logically as possible. Data collected looks more neat and commands are easy to use.
With all the data in a single directory, we need to search for the generic commands. At times there are chances that we overlook the generic commands.

Thank you,

sos/report/plugins/block.py
```diff
@@ -19,6 +19,8 @@ class Block(Plugin, IndependentPlugin):
     files = ('/sys/block',)
 
     def setup(self):
+        _subdir1 = "udevadm_block"
+        _subdir2 = "fdisk_parted"
```
pmoravec, line 23: It is better to name the variables by their usage, e.g. `fdisk_dir` or similar.

sos/report/plugins/block.py
```diff
@@ -50,14 +52,21 @@ def setup(self):
...
-            "parted -s %(dev)s unit s print",
+        udevadm_cmds = [
             "udevadm info %(dev)s",
             "udevadm info -a %(dev)s"
         ]
-        self.add_device_cmd(cmds, devices='block', blacklist='ram.*')
+
+        parted_cmds = [
+            "parted -s %(dev)s unit s print"
+        ]
```
pmoravec, line 62: If the list contains just one item, we can directly use the command `"parted -s .."` in the `add_device_cmd`, like we do for `fdisk -l ..` few lines below, no?

sos/report/plugins/nvme.py
```diff
@@ -26,6 +26,7 @@ class Nvme(Plugin, IndependentPlugin):
     kernel_mods = ('nvme', 'nvme_core')
 
     def setup(self):
+        _subdir = "smart_ns_ctrl_fw_error_regs"
```
pmoravec, line 29: The subdirectory name is human-unreadable (at least for me) - if I would see that name in a directory structure, I won't guess what it contains. Maybe if we want to store outputs of per block devices commands, name the directory `devices` or `block_devices`? See other usages of that:  https://github.com/sosreport/sos/search?q=%22subdir%3D%22

sos/report/plugins/scsi.py
```diff
@@ -29,6 +29,8 @@ class Scsi(Plugin, IndependentPlugin):
     profiles = ('storage', 'hardware')
 
     def setup(self):
+        _subdir1 = "udevadm_scsi"
+        _subdir2 = "sg_persists_inq"
```
pmoravec, line 33: Likewise above, variable names and value names should be more intuitive.

pmoravec:
In general, the changes make sense. Maybe some other plugins iterating over devices deserve the same?

Not sure if some tools like Insights dont rely on some particular files location we break by this change (but such important files should have a tag already).

Nitn-Yewale:
> In general, the changes make sense. Maybe some other plugins iterating over devices deserve the same?
> 

Thank you for the review. I shall make the mentioned changes.

Yes, there are other plugins as well which collect per device data in the same directory and separating it would look the more understandable. One such plugin is network.

I can try and check for other plugins.

> Not sure if some tools like Insights dont rely on some particular files location we break by this change (but such important files should have a tag already).

sos/report/plugins/nvme.py
TurboTurtle: Yeah, I'm personally caught up on the subdirectory structure.

I don't mind adding subdirs here in general, but the question to me is do we make these per-command subdirs or per-device? An analogous use case in sos currently is containers throughout several plugins, or perhaps moreso the different resources that OpenShift supports. The current theme has been to make subdirs based on resources (in this case, devices) rather than commands.

sos/report/plugins/block.py
```diff
@@ -19,6 +19,9 @@ class Block(Plugin, IndependentPlugin):
     files = ('/sys/block',)
 
     def setup(self):
+        _subdir1 = "udevadm_info_block"
+        _subdir2 = "fdisk_and_parted"
```
pmoravec, line 23: For code readiness, could be the variable names be more descriptive, like e.g. `udevadm_dir` or `fdisk_dir`?

sos/report/plugins/block.py
```diff
@@ -50,14 +53,19 @@ def setup(self):
...
             "udevadm info %(dev)s",
             "udevadm info -a %(dev)s"
         ]
-        self.add_device_cmd(cmds, devices='block', blacklist='ram.*')
+
+        self.add_device_cmd(udevadm_cmds, devices='block', blacklist='ram.*',
+                            subdir=_subdir1)
+        self.add_device_cmd("parted -s %(dev)s unit s print",
+                            blacklist="ram.*", devices="block",
+                            tags="parted_s_sos", subdir=_subdir2)
```
pmoravec, line 65: My first thought was "why the tag has redundant `_sos` substring, it is evident the tag belongs to `sos` - but we already had `fdisk_l_sos` tag so I am OK,

sos/report/plugins/nvme.py
```diff
@@ -26,24 +26,44 @@ class Nvme(Plugin, IndependentPlugin):
     kernel_mods = ('nvme', 'nvme_core')
 
     def setup(self):
+        _subdir1 = "nvme_smartctl"
+        _subdir2 = "nvme_fw_smart_error_logs"
+        _subdir3 = "nvme_list-ns"
+        _subdir4 = "nvme_list-ctrl"
+        _subdir5 = "nvme_id-ctrl_-H"
+        _subdir6 = "nvme_id-ns_-H"
+        _subdir7 = "nvme_show-regs"
```
pmoravec, line 35: Likewise above, would it be possible to name variables more understandably? (i.e. `list_ctrl_dir`)?

sos/report/plugins/scsi.py
```diff
@@ -29,6 +29,10 @@ class Scsi(Plugin, IndependentPlugin):
     profiles = ('storage', 'hardware')
 
     def setup(self):
+        _subdir1 = "udevadm_info_scsihosts"
+        _subdir2 = "sg_inq_devices"
+        _subdir3 = "sg_persist_devices"
+
```
pmoravec, line 35: Ditto variable names.

sos/report/plugins/nvme.py
```diff
@@ -26,24 +26,44 @@ class Nvme(Plugin, IndependentPlugin):
...
+        self.add_device_cmd("nvme list-ns %(dev)s", whitelist='nvme.*',
+                            devices='block', subdir=_subdir3)
+        self.add_device_cmd("nvme list-ctrl %(dev)s", whitelist='nvme.*',
+                            devices='block', subdir=_subdir4)
+        self.add_device_cmd("nvme id-ctrl -H %(dev)s", whitelist='nvme.*',
+                            devices='block', subdir=_subdir5)
+        self.add_device_cmd("nvme id-ns -H %(dev)s", whitelist='nvme.*',
+                            devices='block', subdir=_subdir6)
+        self.add_device_cmd("nvme show-regs %(dev)s", whitelist='nvme.*',
+                            devices='block', subdir=_subdir7)
```
pmoravec, line 67: An option that *can* lead to more concise and readable code:

    for cmd, subdir in [(smartctl_cmds, _subdir1),
                                      (nvme_logs_cmds, _subdir2),
                                      ..
                                      ("nvme show-regs %(dev)s", _subdir7)]:
        self.add_device_cmd(cmd, whitelist='nvme.*', devices='block', subdir=subdir)

It is a matter of subjective opinion what is better here, imho, and I am OK with either approach - leaving on the author to decide.

sos/report/plugins/block.py
Nitn-Yewale: agree. I will make the necessary changes.

sos/report/plugins/block.py
Nitn-Yewale: >> My first thought was "why the tag has redundant _sos substring, it is evident the tag belongs to sos - but we already had fdisk_l_sos tag so I am OK,
>> For code readiness, could be the variable names be more descriptive, like e.g. udevadm_dir or fdisk_dir?
>> Ditto variable names.

Not added _dir for nvme or other commands due to similar reasons as we know that it's a directory. OR shall I change it similar to  `list_ctrl_dir`. Actual command is "nvme list-ctrl" hence there is nvme and its not because of the plugin. Could you please consider giving another thought on the same.

sos/report/plugins/scsi.py
Nitn-Yewale: >> My first thought was "why the tag has redundant _sos substring, it is evident the tag belongs to sos - but we already had fdisk_l_sos tag so I am OK,
>> For code readiness, could be the variable names be more descriptive, like e.g. udevadm_dir or fdisk_dir?
>> Ditto variable names.

Not added _dir for nvme or other commands due to similar reasons as we know that it's a directory. OR shall I change it similar to  list_ctrl_dir. Actual command is "nvme list-ctrl" hence there is nvme and its not because of the plugin. Could you please consider giving another thought on the same.

sos/report/plugins/nvme.py
Nitn-Yewale: Apologies for the delay. I have made the necessary changes and shall send the modified path by tomorrow.
Thank you.

Nitn-Yewale: I have pushed changes to https://github.com/Nitn-Yewale/sos. Could you please check now. Please let me know if any action item is on me.

pmoravec:
I don't see my recommendations (i.e. directory variable names) followed at all..? (doing so, you will also get rid of the Flake errors https://github.com/sosreport/sos/pull/3201/checks?check_run_id=17389330583 ;-) )

Instead of having a merge commit for rebasing your fork, please do the following:
- have on your branch just one commit with your changes - squash your commits into one since I dont see functional difference between work done in https://github.com/sosreport/sos/commit/61dd00557a636bbcaf6cee79c212bc53224ca6f6 and https://github.com/sosreport/sos/commit/7e186cc176f5bd950120696f1c8cf447edca601a
- rebase your fork / branch to current sos main branch, e.g. by following https://stackoverflow.com/questions/7929369/how-to-rebase-local-branch-onto-remote-master

Nitn-Yewale:
> I don't see my recommendations (i.e. directory variable names) followed at all..? (doing so, you will also get rid of the Flake errors https://github.com/sosreport/sos/pull/3201/checks?check_run_id=17389330583 ;-) )
> 
> Instead of having a merge commit for rebasing your fork, please do the following:
> 
> * have on your branch just one commit with your changes - squash your commits into one since I dont see functional difference between work done in [61dd005](https://github.com/sosreport/sos/commit/61dd00557a636bbcaf6cee79c212bc53224ca6f6) and [7e186cc](https://github.com/sosreport/sos/commit/7e186cc176f5bd950120696f1c8cf447edca601a)
> * rebase your fork / branch to current sos main branch, e.g. by following https://stackoverflow.com/questions/7929369/how-to-rebase-local-branch-onto-remote-master

Hello @pmoravec , thank you for all the guidance so far.

Could you please check now.

Regards,
Nitin Yewale

Nitn-Yewale:
> This is in a much better state with the condensed dirs, however I still am not a fan of the variable naming being so opaque.
> 
> For example, in `scsi`, using `udevadm_dir` instead of `_subdir1` is much more informative and obvious what we're doing to anyone looking at this in the future.
> 
> I still have some concern around making the directories per-command instead of per-device, but I'm willing to defer here on that and see how things evolve going forward.


Thank you @pmoravec  @TurboTurtle

PR #3203, opened 2023-04-24
[iscsi] Scrub password_in strings

jcastill:
Scrub strings in password_in in /etc/iscsi/iscsid.conf and the files within the 
directory /var/lib/iscsi/nodes/, i.e. /var/lib/iscsi/nodes/*/*/default and others.

Fixes: RHBZ#2187859

---


sos/report/plugins/iscsi.py
```diff
@@ -41,11 +41,18 @@ def postproc(self):
         # to
         #
         # node.session.auth.password = ********
-        nodesessionpwd = r"(node\.session\.auth\.password\s+=\s+)(\S+)"
-        discoverypwd = r"(discovery\.sendtargets\.auth\.password\s+=\s+)(\S+)"
+        nodesessionpwd = r"(node\.session\.auth\.password(_in)?\s+=\s+)(\S+)"
+        discoverypwd = (
+                r"(discovery\.sendtargets\.auth\.password(_in)!"
```
TurboTurtle, line 46: Is the `!` actually required here? It seems somewhat out of place to me at first glance, but I also am not fully aware of the context of what we're targeting here.

TurboTurtle (commented):
Ack, once the question about the `!` is answered.

sos/report/plugins/iscsi.py
pmoravec: That really seems like a typo, the format of either `node.session` and `discovery.sendtargets` is the same.

jcastill: Yes, was a typo while I was trying to fix the line that had more than 80 chars, sorry! I'll force push the fix soon.

pmoravec (approved):

TurboTurtle (approved):

Issue #3186, opened 2023-04-05
[tests] restrict foreman tests to x86_64 architecture only


pmoravec:
`Foreman` can run on `x86_64` architecture only (not speaking about architectures of systems it can provision, just where `foreman` can run *on*). So our testing should restrict `tests/product_tests/foreman` to that arch only.

This is not directly an issue in `sos` upstream where we run tests against given `foreman` instances (on the right arch), but running `sos` avocado tests on a random system on another arch causes very strange failures of `foreman` tests - let skip that.

Is there a way to skip whole directory / run all tests from a directory conditionally? Or whole class conditionally? The best I found is

    from avocado.utils.cpu import get_arch

and

    @skipIf(get_arch() != 'x86_64', f"foreman needs arch x86_64, not {get_arch()}")

added before any class method in `tests/product_tests/foreman/foreman_tests.py` - but then we stillr edundantly call parent class and its tests (e.g. `tests/product_tests/foreman/foreman_tests.py:ForemanBasicTest.test_archive_created`). And adding the `@skipIf` just before `class` definition has no effect.

PR #3210, opened 2023-05-01
[tests] Allow tests to restrict execution architecture

TurboTurtle:
Adds a new `arch` class attr that allows test classes to specify which architecture(s) they are allowed/designed to execute on. If the test is attempted to be executed on an architecture that is not in the specified list, we will skip the test before any setup or sos executions are done.

By default, the `arch` class attr is an empty list, which implies that the test can run on _any_ architecture. Tests should specify `arch` as a list even for single-arch tests.

Included in this change is restricting the Foreman tests to x86_64 only, as that is the only architecture on which that product is supported.

Closes: #3186

---


pmoravec (approved):
Nice.

PR #3215, opened 2023-05-02
[redhat] Add debugging output to failed SFTP uploads

TurboTurtle:
2-patch set first allows `ui_log` to print debug level messages if verbosity is set high enough, and second adds debugging output to failed SFTP uploads when using the RHEL policy.

---


TurboTurtle:
@mhradile PTAL, let me know if this is sufficient in your mind. We're printing the response status code and the response content. I don't think we need any headers or the like, but maybe there's something else I'm missing?

pmoravec (approved):
ACK.

Issue #3213, opened 2023-05-02
[juju] cluster collection doesn't work in 2 scenarios


arif-ali:
based on my testing, the following 2 scenarios will not work

1. if there is an application deployed, but no unit is consuming it, then should be ignored, and move on
2. subordinate charms will not have an allocated unit, so the `units` key will not exist

Proposing the following solution

```diff
diff --git a/sos/collector/clusters/juju.py b/sos/collector/clusters/juju.py
index d783bd76..3b596f4c 100644
--- a/sos/collector/clusters/juju.py
+++ b/sos/collector/clusters/juju.py
@@ -70,6 +70,10 @@ def _get_index(model_name):
                 subordinate_to = app_info.get("subordinate-to", [])
                 for parent in subordinate_to:
                     self.apps[app].extend(self.apps[parent])
+                    if not self.apps[parent]:
+                        continue
+                    if 'units' not in juju_status["applications"][parent]:
+                        continue
                     units = juju_status["applications"][parent]["units"]
                     for unit, unit_info in units.items():
                         node = f"{self.model_name}:{unit_info['machine']}"
```

\cc @jneo8

TurboTurtle:
Code-wise the changes look fine. I defer to yourself and @jneo8 as to applicability and fitness for `juju`.

jneo8:
Hi @arif-ali, the fix seems ok, can you provide the juju status json file so I can add unit-test to it?

jneo8:
I think it's because the unit status is not health, it's a broken deployment case.
We can add some error handling to it.

PR #3216, opened 2023-05-03
[juju] Fix cluster collection doesn't work in 2 scenarios

jneo8:
Fix below two scenarios:
1. If subordinate's parent is missing.
2. If subordinate's parent's units is missing.

Closes: #3213

---

### DIFF
--- a/sos/collector/clusters/juju.py
+++ b/sos/collector/clusters/juju.py
@@ -8,6 +8,7 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
+import logging
 import json
 import re
 
@@ -45,6 +46,7 @@ def _get_index(model_name):
             self.apps = {}
             self.units = {}
             self.machines = {}
+            self.ui_log = logging.getLogger("sos")
 
         def add_principals(self, juju_status):
             """Adds principal units to index."""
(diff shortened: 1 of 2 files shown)


sos/collector/clusters/juju.py
```diff
@@ -45,6 +46,7 @@ def __init__(self, model_name):
             self.apps = {}
             self.units = {}
             self.machines = {}
+            self.ui_log = logging.getLogger("sos_ui")
```
pmoravec, line 49: Isn't it better to use `self.log_warn` directly, to log the message both to console and logfile? (https://github.com/sosreport/sos/blob/main/sos/collector/clusters/__init__.py#L221-L223).

tests/unittests/juju/juju_cluster_tests.py
```diff
@@ -291,4 +292,29 @@ def test_subordinates(self, mock_exec_primary_cmd):
...
+    def test_subordinate_parent_miss_units(self):
+        """Fix if subordinate's parent is missing units."""
+        model = "sos"
+        index = _get_index(model_name=model)
+
+        juju_status = json.loads(get_juju_output(model=model))
+        juju_status["applications"]["ubuntu"].pop("units")
+
+        index.add_principals(juju_status)
+        index.add_subordinates(juju_status)
```
pmoravec, line 306: Good to have tests for it. Just my first thought was "and what do we technically test here, isn't an `assert*` missing here?". Maybe a comment like "ensure these commands won't fail even when no `units` are present" would be helpfull.

(the same applies to the 2nd method)

arif-ali (approved):
Functionally, this is now working and tested

sos/collector/clusters/juju.py
jneo8: If I want to use that, I need to pass the log function into `_get_index`.

tests/unittests/juju/juju_cluster_tests.py
jneo8: Ack

sos/collector/clusters/juju.py
TurboTurtle: In that case, I'd say use `getLogger('sos')` instead of `sos_ui`, as this will write to both the file log and to console.

TurboTurtle (approved):
LGTM, I restarted the CI job which failed due to scheduling, not anything related to this PR.

pmoravec (approved):

Issue #3214, opened 2023-05-02
[collect] sos collect does not detect snap


arif-ali:
Testing sos collect, now that juju is supported

deb works without any problems

however, if the sosreport deb is not installed, but the snap is, then it does not detect the sosreport snap

```
Connecting to nodes...
[juju-586a2b-5-lxd-2
Connection to 10.0.1.142 closed.:_load_sos_info] sos is not installed on this node
[juju-586a2b-3-lxd-2
Connection to 10.0.1.132 closed.:_load_sos_info] sos is not installed on this node
[juju-586a2b-4-lxd-2
Connection to 10.0.1.120 closed.:_load_sos_info] sos is not installed on this node
No nodes connected. Aborting...
```

Maybe for any distro policies they need to have multiple package manager capabilities, and be able to detect their primary ones  as  well as snap and flatpak

jneo8:
I think the question should be: *How do we know the sos bin path before we access remote?*
Imaging you have two ubuntu nodes but only one is using snap. I am not sure detect where is the binary before remote execution is a nice idea.

Also see here: https://github.com/sosreport/sos/commit/31102aff1a9a1a1338f075220edc8bc8f08db593

TurboTurtle:
> I think the question should be: _How do we know the sos bin path before we access remote?_

We load a policy for each connected `SosNode`. The policy defines where to look for `sos`.

>_ Imaging you have two ubuntu nodes but only one is using snap. I am not sure detect where is the binary before remote execution is a nice idea.
> 
> Also see here: [31102af](https://github.com/sosreport/sos/commit/31102aff1a9a1a1338f075220edc8bc8f08db593)

What we would need here is:

1)  a `PackageManager` for snaps, like we have for `dpkg` and `rpm`.
2a) One option: a "multi" `PackageManager` that allows policies who use it to leverage multiple individual package managers at the same time, then hook the Ubuntu policy up to this. There'd still be some details to work out like order of priority for individual package managers here
2b) Another option: Logic added directly to the Ubuntu policy that lets us dynamically choose between `dpkg` and `snap`.

2a is more flexible and likely has use cases outside of this immediate case. Implementation is more complex however. 2b is likely more straight forward, though only applicable (currently) to Ubuntu.

arif-ali:
> What we would need here is:
> 
>     1. a `PackageManager` for snaps, like we have for `dpkg` and `rpm`.
> 
> 
[...]

PR #3219, opened 2023-05-03
[package_managers] Add MultiPackageManager support for policies

TurboTurtle:
Adds a new `MultiPackageManager` class to handle systems that have multiple distinct package managers on them.

This initial implementation requires policies that will leverage this to define a `primary` package manager, and at least one fallback package manager. Policies will need to pass the uninstantiated class object to the initializing call to `MultiPackageManager`, for example:

    self.pkgman = MultiPackageManager(primary=RpmPackageManager,
                                      fallbacks=[DpkgPackageManager])

When a call to a `PackageManager` method is made to `MultiPackageManager`, the call is first directed to the primary manager. If that response is empty, None, etc... then the call is forwarded to each of the fallback managers in the order they were defined, until either a non-empty response is returned or we run out of fallbacks to try.

In order to allow for some introspection of which defined manager provides a non-empty response for package queries, a new `pkg_manager` entry in the contents of the `packages` property has been added.

Closes: #2237
Related: #3214

---

### DIFF
--- a/sos/policies/package_managers/__init__.py
+++ b/sos/policies/package_managers/__init__.py
@@ -154,12 +154,30 @@ class PackageManager():
         except Exception:
             return None
 
+    def _parse_pkg_list(self, pkg_list):
+        """
+        Using the output of `query_command`, build the _packages dict.
+
+        This should be overridden by distinct package managers and be a
+        generator for _generate_pkg_list which will insert the packages into
+        the _packages dict.
+
+        This method should yield a tuple of name, version, release for each
+        package parsed. If the package manager or distribution does not use a
+        release field, set it to None.
+
+        :param pkg_list: The output of the result of `query_command`
+        :type pkg_list:  ``str``
+        """
+        raise NotImplementedError
+
     def _generate_pkg_list(self):
         """Generates a dictionary of packages for internal use by the package
         manager in the format::
 
             {'package_name': {'name': 'package_name',
                               'version': 'major.minor.version',
+                              'release': 'package release' or None,
                               'pkg_manager': 'package manager name'}}
 
         """
@@ -167,21 +185,13 @@ class PackageManager():
             cmd = self.query_command
             pkg_list = self.exec_cmd(cmd, timeout=30, chroot=self.chroot)
 
-            for pkg in pkg_list.splitlines():
-                if '|' not in pkg:
-                    continue
-                elif pkg.count("|") == 1:
-                    name, version = pkg.split("|")
-                    release = None
-                elif pkg.count("|") == 2:
-                    name, version, release = pkg.split("|")
-                self._packages[name] = {
-                    'name': name,
-                    'version': version.split("."),
+            for pkg in self._parse_pkg_list(pkg_list):
+                self._packages[pkg[0]] = {
+                    'name': pkg[0],
+                    'version': pkg[1].split('.'),
+                    'release': pkg[2],
                     'pkg_manager': self.manager_name
                 }
-                release = release if release else None
-                self._packages[name]['release'] = release
 
     def pkg_version(self, pkg):
         """Returns the entry in self.packages for pkg if it exists
--- a/sos/policies/package_managers/__init__.py
+++ b/sos/policies/package_managers/__init__.py
@@ -53,18 +53,10 @@ class PackageManager():
     chroot = None
     files = None
 
-    def __init__(self, chroot=None, query_command=None, verify_command=None,
-                 verify_filter=None, files_command=None,
-                 query_path_command=None, remote_exec=None):
+    def __init__(self, chroot=None, remote_exec=None):
         self._packages = {}
         self.files = []
         self.remote_exec = remote_exec
-
-        self.query_command = query_command or self.query_command
-        self.verify_command = verify_command or self.verify_command
-        self.verify_filter = verify_filter or self.verify_filter
-        self.files_command = files_command or self.files_command
-        self.query_path_command = query_path_command or self.query_path_command
 
         if chroot:
             self.chroot = chroot
@@ -74,6 +66,10 @@ class PackageManager():
         if not self._packages:
             self._generate_pkg_list()
         return self._packages
+
+    @property
+    def manager_name(self):
+        return self.__class__.__name__.lower().split('package')[0]
 
     def exec_cmd(self, command, timeout=30, need_root=False, env=None,
                  get_pty=False, chroot=None):
@@ -163,7 +159,8 @@ class PackageManager():
         manager in the format::
 
             {'package_name': {'name': 'package_name',
-                              'version': 'major.minor.version'}}
+                              'version': 'major.minor.version',
+                              'pkg_manager': 'package manager name'}}
 
         """
         if self.query_command:
@@ -180,7 +177,8 @@ class PackageManager():
                     name, version, release = pkg.split("|")
                 self._packages[name] = {
                     'name': name,
-                    'version': version.split(".")
+                    'version': version.split("."),
+                    'pkg_manager': self.manager_name
                 }
                 release = release if release else None
                 self._packages[name]['release'] = release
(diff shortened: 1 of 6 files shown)


TurboTurtle:
@arif-ali I've worked out what I think is a functional abstraction for allowing a policy to define multiple package managers. I haven't tested it with your initial snap manager though, so if you could PTAL and see if it blows up on a system that _actually_ has multiple package managers in place (instead of my local mocked setup for that), it'd help a lot.

arif-ali:
@TurboTurtle this works for me; however, the caveat is that I did not use the sos_bin_path

```diff
diff --git a/sos/collector/sosnode.py b/sos/collector/sosnode.py
index a167c263..b76c19a7 100644
--- a/sos/collector/sosnode.py
+++ b/sos/collector/sosnode.py
@@ -661,7 +661,7 @@ class SosNode():
 
         sos_cmd = sos_cmd.replace(
             'sosreport',
-            os.path.join(self.host.sos_bin_path, self.sos_bin)
+            self.sos_bin
         )
 
         for opt in self.sos_options:
```

The ubuntu.py looked like the one below

```diff
diff --git a/sos/policies/distros/ubuntu.py b/sos/policies/distros/ubuntu.py
index a6bf7260..b5af254d 100644
--- a/sos/policies/distros/ubuntu.py
+++ b/sos/policies/distros/ubuntu.py
@@ -9,6 +9,10 @@
 from sos.report.plugins import UbuntuPlugin
 from sos.policies.distros.debian import DebianPolicy
 
+from sos.policies.package_managers.snap import SnapPackageManager
+from sos.policies.package_managers.dpkg import DpkgPackageManager
+from sos.policies.package_managers import MultiPackageManager
+
 import os
 
 
@@ -31,6 +35,12 @@ class UbuntuPolicy(DebianPolicy):
         super(UbuntuPolicy, self).__init__(sysroot=sysroot, init=init,
                                            probe_runtime=probe_runtime,
                                            remote_exec=remote_exec)
+
+        self.package_manager = MultiPackageManager(primary=DpkgPackageManager,
+                                                  fallbacks=[SnapPackageManager],
+                                                  chroot=self.sysroot,
+                                                  remote_exec=remote_exec)
+
         self.valid_subclasses += [UbuntuPlugin]
 
     @classmethod
```

TurboTurtle:
> @TurboTurtle this works for me; however, the caveat is that I did not use the sos_bin_path
> 

The `sos_bin_path` right now is defaulting to the stock setting in the base `LinuxPolicy`, that being `/usr/bin`. The Ubuntu policy would need some additional logic in `__init__()` to set the appropriate path if `sos` is installed via snap, e.g.:

```python
    def __init__(self, sysroot=None, init=None, probe_runtime=True,
                 remote_exec=None):
        [...]
        if self.package_manager.pkg_by_name('sosreport')['pkg_manager'] == 'snap':
            self.sos_bin_path = '/some/snap/path'

```

Or something similar.

arif-ali:
> The `sos_bin_path` right now is defaulting to the stock setting in the base `LinuxPolicy`, that being `/usr/bin`. The Ubuntu policy would need some additional logic in `__init__()` to set the appropriate path if `sos` is installed via snap, e.g.:
> 
> ```python
>     def __init__(self, sysroot=None, init=None, probe_runtime=True,
>                  remote_exec=None):
>         [...]
>         if self.package_manager.pkg_by_name('sosreport')['pkg_manager'] == 'snap':
>             self.sos_bin_path = '/some/snap/path'
> ```
> 
> Or something similar.

Yeah, that would work, will need a new PR for the ubuntu distro policy addition, as well as adding the pkg_manager item to the snap  once this and the snap.py PRs are merged. Seems lie we have a few inter-dependancies :)

Now, I need this PR to go in, then I'll add mine to add

TurboTurtle:
Ok. I've rebased and made a change to make future packagemanagers easier to implement, and updated the existing ones (plus the new `snap` manager) to use the new design which in short, has `_generate_pkg_list()` now rely on a new `_parse_pkg_list()` generator method that each implementation needs to define. This all now ensures that all package entries in a package manager are consistently formatted, which is extremely helpful for the new "multi" manager.


I've also added some unit tests for the rpm and dpkg managers. I am not personally sure of if snaps are used out of the box for Ubuntu systems, or if there's a very reliable way to determine if there are any snaps installed on a freshly spun VM of a given Ubuntu version, so I've left that one out for the time being.

TurboTurtle:
All that to say, @arif-ali - I believe this is ready for use, just need one more validation check from you and then we can make the changes to the Ubuntu policy to define `sos_bin_path` based on if it is installed via deb or snap.

sos/policies/package_managers/__init__.py
```diff
@@ -158,32 +154,43 @@ def pkg_by_name(self, name):
...
+        raise NotImplementedError
+
     def _generate_pkg_list(self):
         """Generates a dictionary of packages for internal use by the package
         manager in the format::
 
             {'package_name': {'name': 'package_name',
-                              'version': 'major.minor.version'}}
+                              'version': 'major.minor.version',
+                              'pkg_manager': 'package manager name'}}
```
arif-ali, line 180: should we add the `release:`  so that we know that is part of the dict?

It though me off earlier today, as didn't realise that was part of the dict

TurboTurtle:
There we go, had to find a package commonly named and installed by default on all our testing images. `kernel` didn't work because f38 changes to `kernel-core`.

arif-ali:
> I've also added some unit tests for the rpm and dpkg managers. I am not personally sure of if snaps are used out of the box for Ubuntu systems, or if there's a very reliable way to determine if there are any snaps installed on a freshly spun VM of a given Ubuntu version, so I've left that one out for the time being.

```
arif@arif-laptop02:~$ lxc launch ubuntu:jammy jammy01
Creating jammy01
Starting jammy01                            
arif@arif-laptop02:~ lxc exec jammy01 bash

root@jammy01:~# snap list
Name    Version        Rev    Tracking       Publisher   Notes
core20  20230308       1852   latest/stable  canonical✓  base
lxd     5.0.2-838e1b2  24322  5.0/stable/…   canonical✓  -
snapd   2.59.1         18933  latest/stable  canonical✓  snapd

```

Above output from `jammy` (22.04). So, we could check for `lxd` package. This is also valid for `focal` (20.04). This, however doesn't work for `bionic` (18.04), but I am not too worried about that, as standard support for that ceases at the end of May 2023.

But, I guess we need to ensure, that the VMs/instances in GCP would have a similar items. We could also check for `snapd` too. I don't trust to check for `core20`, but the default core package could change over time.

arif-ali (approved):
good for me after testing with additions to ubuntu distro policy

arif-ali:
this solved the problem

```diff
diff --git a/sos/policies/package_managers/dpkg.py b/sos/policies/package_managers/dpkg.py
index 117c29dd..435e7ed1 100644
--- a/sos/policies/package_managers/dpkg.py
+++ b/sos/policies/package_managers/dpkg.py
@@ -15,7 +15,7 @@ class DpkgPackageManager(PackageManager):
     """Subclass for dpkg-based distrubitons
     """
 
-    query_command = "dpkg-query -W -f='${Package}|${Version}\\n'"
+    query_command = "dpkg-query -W -f='${Package}|${Version}|${Status}\\n'"
     query_path_command = "dpkg -S"
     verify_command = "dpkg --verify"
     verify_filter = ""
@@ -24,7 +24,9 @@ class DpkgPackageManager(PackageManager):
         for pkg in pkg_list.splitlines():
             if '|' not in pkg:
                 continue
-            name, version = pkg.split('|')
+            name, version, status = pkg.split('|')
+            if "deinstall" in status:
+                continue
             yield (name, version, None)
 
 # vim: set et ts=4 sw=4 :
```

TurboTurtle:
I've pulled in the change you suggested and marked you on the commit for it, great catch there!

arif-ali (approved):
re-pulled the changes, re-ran the sos collect, no issues.

next up, implement the `MultiPackageManager` in `ubuntu.py` :)

Issue #3220, opened 2023-05-05
[networking] route call returns only IPv4 routing table


mkowalski:
In the networking plugin we have currently `route -n` call which does not return IPv6 table in dual-stack systems. As a result we are getting a file with only IPv4 table. This is confusing and could be easily fixed by e.g. `route -n -4 -6` call.

PR #3221, opened 2023-05-05
[networking] Collect IPv4 and IPv6 routes

mkowalski:
With this PR we are deprecating `route` command in favour of `ip route`.
We are also creating a single root symlink `ip_route` that will contain
aggregated IPv4 and IPv6 routes.

Closes: #3220

---

### DIFF
--- a/sos/report/plugins/networking.py
+++ b/sos/report/plugins/networking.py
@@ -82,7 +82,8 @@ class Networking(Plugin):
 
         self.add_cmd_output("ip -o addr", root_symlink="ip_addr",
                             tags='ip_addr')
-        self.add_cmd_output("route -n", root_symlink="route", tags='route')
+        self.add_cmd_output("ip route show table all", root_symlink="ip_route",
+                            tags=['ip_route', 'iproute_show_table_all'])
         self.add_cmd_output("plotnetcfg")
 
         self.add_cmd_output("netstat %s -neopa" % self.ns_wide,
@@ -93,7 +94,6 @@ class Networking(Plugin):
             "netstat -s",
             "netstat %s -agn" % self.ns_wide,
             "networkctl status -a",
-            "ip route show table all",
             "ip -6 route show table all",
             "ip -d route show cache",
             "ip -d -6 route show cache",
@@ -253,7 +253,6 @@ class Networking(Plugin):
             "ethtool -k.*": "ethtool_k",
             "ip -d address": "ip_addr",
             "ip -s -s neigh show": "ip_neigh_show",
-            "ip route show table all": "iproute_show_table_all",
             "ip -s -d link": "ip_s_link",
             "netstat.*-neopa": "netstat",
             "netstat.*-agn": "netstat_agn",


TurboTurtle:
I'm thinking we should actually make this a separate collection for `route -n -6` - similar to how we have split out the `ip route show` commands:

```
            "ip route show table all",
            "ip -6 route show table all",
            "ip -d route show cache",
            "ip -d -6 route show cache",
```

mkowalski:
I was thinking about this but then I look at other stuff and I see e.g.

```
self.add_cmd_output("ip -o addr", root_symlink="ip_addr",
                            tags='ip_addr')
```

so that would mean for IP addresses we collect them all together and for routes we split it. If we wanted to follow the pattern, we could then do separately

```
ip -4 -o addr
ip -6 -o addr
route -n -4
route -n -6
```

What do you think?

TurboTurtle:
Part of my hesitance for combining `route` output is automation - the IPv4 output there is formatted a certain way, and automation looking at sos reports is expecting that IPv4 style. The IPv6 information is grafted onto that output at the end, in a (slightly) different format. Automation that is looking at the `route` symlink will likely trip up on the different format of new output, and we generally try not to be disruptive especially for long-standing collections.

In the case of `ip -o addr`, both IPv4 and IPv6 content is formatted approximately the same, and is continuous - it's not simply a chaining of outputs. We wouldn't have the same concerns as above.

mkowalski:
Oh right this is a very valid reason. Let me create a separate call to get the v6 routes then. Thanks!

sos/report/plugins/networking.py
```diff
@@ -82,7 +82,10 @@ def setup(self):
 
         self.add_cmd_output("ip -o addr", root_symlink="ip_addr",
                             tags='ip_addr')
-        self.add_cmd_output("route -n", root_symlink="route", tags='route')
+        self.add_cmd_output("route -n -4", root_symlink="route",
+                            tags='route')
```
TurboTurtle, line 86: Ack to making this explicitly `-n -4`. For the tags, let's make that `tags=['route', 'route_v4']` as well. We keep current automation happy, and provide a mechanism to be more exact in the future.

sos/report/plugins/networking.py
```diff
@@ -82,7 +82,10 @@ def setup(self):
...
+        self.add_cmd_output("route -n -6", root_symlink="route_v6",
+                            tags='route_v6')
```
TurboTurtle, line 88: Hmm, not sure about dropping another root symlink with this. If anything, I've been mulling over reducing the number of root symlinks we create.

@pmoravec your thoughts?

mkowalski: If we don't want to create root symlink, I'm not sure there is point at all in pushing this PR. `route` is deprecated for quite some time now in favour of `ip route` and we run somewhere else `ip route show table all` already - so we do have the required data in the sosreport, it's just that the structure is inconsistent at the moment.

If it anyhow happened that someone runs sosreport on a system with `route` but without `ip route` I am not sure this system is recent enough to be even plugged into IPv6/dual-stack networking, so we may be also doing useless thing...

TurboTurtle: The root symlinks were traditionally for the most referenced parts of sos collections. Over time, as our collections have grown in both scope and complexity, we've shied away from dropping more symlinks in the archive root, in an effort to keep things consumable at a glance.

Removal of root symlinks is something we tend to avoid mainly because of just how long it has been "common" knowledge that you can untar an sos report, cd into the directory and immediately get certain networking information or packaging information without having to hunt it down in the subdirectories.

At the same time, while not necessarily in conjunction, we've added better accounting for the collections in a report and through `sos_reports/manifest.json` we provide an easy way for automation to pick out specific collections via tags so that those systems aren't reliant on keeping up with the exact command construction for every version of sos they might consume reports from.

If `route` is deprecated, I'd pause before adding a new collection for it just for the sake of having a new root symlink added. Instead, I'd look at the applicability of maintaining those collections and symlinks going forward. If `ip route` is the "correct" way to get this information, what do you think about dropping the current `route` collection entirely? Is `ip route show table all` sufficient to cover all the same information? We could perhaps drop the `route` collection and replace the symlink with `ip_route` pointing to this command, which seems to include IPv6 information as well, if there is significant concern over removal of the root symlink.

mkowalski: I think it's all about compatibility now... The output format of `route` and `ip route` is entirely different, so back to your initial point -- if there is someone automating sosreport's output and parsing the output, we would brake them.

Knowledge-wise, we are not losing anything by dropping the current `route` collection. Supportability-wise, we may break some people (even though they should follow the deprecation notice from years ago and adjust). I'm not sure there is a silver bullet here. For consistency we should either extend what we collect via `route` or remove it completely. My personal view would be to drop `route` because it's outdated anyway, but again, it's all about backwards-compatibility.

For reference, https://wiki.linuxfoundation.org/networking/net-tools

pmoravec: Having one `ip_route` root symlink and dropping `route ..` commands entirely makes the most sense to me - until some support or bigger group of users rely on it (though even then I would have this as a goal and warn those users to prepare for the change, in this case).

sos/report/plugins/networking.py
```diff
@@ -82,7 +82,8 @@ def setup(self):
 
         self.add_cmd_output("ip -o addr", root_symlink="ip_addr",
                             tags='ip_addr')
-        self.add_cmd_output("route -n", root_symlink="route", tags='route')
+        self.add_cmd_output("ip route show table all", root_symlink="ip_route",
+                            tags='ip_route')
```
pmoravec, line 86: We add one additional tag to this command output (https://github.com/sosreport/sos/blob/9206c4614a5844340b96b2416b19a5f7d2f91402/sos/report/plugins/networking.py#L256); both tags are properly added in a generated `manifest.json`. We might merge the tags association but I think it is redundant.

pmoravec (approved):

sos/report/plugins/networking.py
mkowalski: Just changed so that we define all tags in a single place

PR #3223, opened 2023-05-08
[powerpc] collect invscout logs

sourabhjains:
Collect microcode and Vital Product Data (vpd) files generated by invscout command.

The execution of invscout commands precedes the beginning of log file collection.

invscout log files are mostly in KBs and old logs files are deleted under /var/adm/invscout at the start of every execution of invscout command.

---


sos/report/plugins/powerpc.py
```diff
@@ -59,6 +59,8 @@ def setup(self):
             ])
 
         if ispSeries:
+            self.collect_cmd_output("invscout")
+            self.collect_cmd_output("invscout -v")
```
TurboTurtle, line 63: `collect_cmd_output()` is for if you want the output of the command returned so you can do something else with it as well as store the output in the archive.

If all you need here is to have the output saved in the archive, use `add_cmd_output()` instead.

sourabhjains: Hello @TurboTurtle 

I observed that, if the above commands are executed using ```add_cmd_output``` then
the logs files generated by them in ```/var/adm/invscout``` may not get capture with
with ```add_copy_spec```.

Is there a way in ```add_cmd_output``` through which I can enforce the execute of certain commands precedes the log files collection using ```add_copy_spec```.

Sourabh

sourabhjains:
One of the CI failed with the below error message:

``` Failed to start an instance: INTERNAL: API response null(0): An error has occurred. executor rejected```

Hope this is nothing to do with this patch.

- Sourabh

File: sos/report/plugins/powerpc.py (hunk 1 above)
Comment (TurboTurtle) [reply sos/report/plugins/powerpc.py:63]: `collect_cmd_output()` shouldn't be collecting any additional log files created by the command either.

In any case, there are generally two approaches for collections like this that also generate their own files.

1.  You can determine the plugin's directory path via `get_cmd_output_path()`. For example, if `invscout` supports an option for specifying where the log gets written, say `--output-dir`, then you could do something like this:

```python
_dir = self.get_cmd_output_path(make=False)
self.add_cmd_output(f"invscout --output-dir={_dir}")
```

2. If that's not possible, then use the `runat` parameter for `add_cmd_output()`. This will cause sos to run the given command from the specified directory, so that the logfile will simply be written to that directory. E.G.:

```python
self.add_cmd_output('invscout', runat=self.get_cmd_output_path(make=False))
```

The `storcli` plugin uses this method for a very similar use case.

TurboTurtle:
> One of the CI failed with the below error message:
> 
> ` Failed to start an instance: INTERNAL: API response null(0): An error has occurred. executor rejected`
> 
> Hope this is nothing to do with this patch.
> 
> * Sourabh


Yep, that's a CirrusCI error, nothing to do with your changes. On a re-push to your PR branch with any changes, the CI will get kicked off again automatically.

sos/report/plugins/powerpc.py
sourabhjains: invscout doesn't have an option to specify output directory so will try second approach.

Thanks for the info.

Sourabh

TurboTurtle: Well, that's unfortunate.

At the end of the day though, that really does only leave us with forcing the execution during `setup()`, and then specifying the newly created files.

I'm okay with having this execute this way (using `collect_cmd_output()` but not leveraging the returned output) if we add a comment explaining why this is the case.

@pmoravec any concerns?

pmoravec: ACK. The other way would be *not* collecting the logfiles (directly), but calling sort of:

    self.add_cmd_output(["invscout", "invscout -v"])  # with default priority 10
    self.add_cmd_output([f"cat {f}" for f in glob.glob("/var/adm/invscout/")], priority=11)

to collect the logfiles by `cat`ing them after `invscout` is called. But this is impractical as e.g. sos archive would keep the data under `sos_commands/powerpc` directory.

So +1 for the `collect_cmd_output()` approach with a comment.

sourabhjains: @TurboTurtle @pmoravec Thank you for taking the approach into consideration.

I have included comments that clarify the reason for executing the ```invscout``` command in the setup function. Please review the updated patch.

pmoravec (approved):

Issue #2525, opened 2021-05-06
[ubuntu][openstack_*] Openstack services need to explicitly include logs from /var/log/apache2


lathiat:
The apache plugin was recently changed to specifically exclude collecting logs for known openstack services, such as /var/log/{httpd,apache2}/nova*.log on any platform in #1950

The relevant paths for RedHat in /var/log/httpd were added to the various OpenStack plugins but they also need Ubuntu paths under /var/log/apache2. Need to confirm which services are affected and add them.

arif-ali:
circling back on this issue

i think the problem here is https://github.com/sosreport/sos/blob/main/sos/report/plugins/apache.py#L60

We are asking it to ignore some folders, but this won't work for ubuntu, as the apachepkg is wrong variable to use. The directory here should be `apache2` and not `apache`. i think this needs a bit of work in the same way RH stuff has been pushed over to the likes of `openstack_nova.py`

I think, quite a bit of work is needed here to bring it on-par imho

whereas RH stuff puts the logs in like `/var/log/http/<openstack_service>/`, ubuntu charming has `/var/log/apache2/<openstack-service>_*`, so could be a bit tricky

arif-ali:
Just realised, it's actually worse, it's not just ignoring folders in `/var/log/apache2` with those names, but files too. Just tested this on a `nova-cloud-controller` unit, and just realised we don't currently collect the following files

* `nova-api-os-compute_access.log`
* `nova-api-os-compute_error.log`
* `nova_api_access.log`
* `nova_api_error.log`
* `nova_meta_access.log`
* `nova_meta_error.log`

PR #3227, opened 2023-05-10
[openstack_*][ubuntu] Update paths and collections

arif-ali:
Update the apache plugin to only collect the defaults and update the openstack bits to collect specifics for their service

Closes: #2525

---

### DIFF
--- a/sos/report/plugins/openstack_nova.py
+++ b/sos/report/plugins/openstack_nova.py
@@ -92,6 +92,7 @@ class OpenStackNova(Plugin):
         if self.get_option("all_logs"):
             self.add_copy_spec([
                 "/var/log/nova/",
+                "/var/log/{}*/nova*".format(self.apachepkg),
             ])
         else:
             novadir = '/var/log/nova/'
@@ -106,6 +107,9 @@ class OpenStackNova(Plugin):
             ]
             for novalog in novalogs:
                 self.add_copy_spec(self.path_join(novadir, novalog))
+            self.add_copy_spec([
+                "/var/log/{}*/nova*.log".format(self.apachepkg),
+            ])
 
         pp = ['', '_libvirt', '_metadata', '_placement']
         sp = [
@@ -162,6 +166,7 @@ class OpenStackNova(Plugin):
 
 class DebianNova(OpenStackNova, DebianPlugin, UbuntuPlugin):
 
+    apachepkg = "apache2"
     nova = False
     packages = (
         'nova-api-ec2',
@@ -183,8 +188,8 @@ class DebianNova(OpenStackNova, DebianPlugin, UbuntuPlugin):
         'nova-volume',
         'novnc',
         'python-nova',
-        'python-novaclient',
-        'python-novnc'
+        'python-novnc',
+        'python3-nova',
     )
     service_name = "nova-api.service"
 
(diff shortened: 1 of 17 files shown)


arif-ali:
@TurboTurtle from the RH perspective do you know if manila has logs in `/var/log/httpd`?

arif-ali:
I am opening this up for review, for targeting 4.5.5; as it would be good to get feedback from other canonical folks

TurboTurtle (approved):
LGTM

sos/report/plugins/openstack_nova.py
```diff
@@ -92,6 +92,8 @@ def setup(self):
         if self.get_option("all_logs"):
             self.add_copy_spec([
                 "/var/log/nova/",
+                "/var/log/{}*/nova*".format(self.apachepkg),
+                "/var/log/{}*/placement*".format(self.apachepkg),
```
pmoravec, line 96: We collect the same `copy_spec` also in `openstack_placement` (same applies to `"/var/log/{}*/placement*.log".format(self.apachepkg),` few lines below). Two plugins can't collect the same `copy_spec` (I guess you know the reason).

Are we sure the two plugins wont be enabled concurrently any time? Or shouldnt we collect the `placement` logs just in `openstack_placement` plugin?

pmoravec:
ACK up to the one concern.

sos/report/plugins/openstack_nova.py
arif-ali: On the Ubuntu platforms and how it is deployed, it is typically on 2 separate hosts. But, I brought these in from the RedHat section primarily.

But, it is technically possible of that happening where you have both the nova and placement on the same host.

I will re-evaluate this later

pmoravec (approved):

arif-ali:
@TurboTurtle @pmoravec one thing that really threw me he on the apache2 plugin was the fact that line 48 mentions `subdir`, this in-fact is a wrong thing to label. This will exclude files or dirs under the `/var/log/<apachepkg>/`

I think, that should be renamed or refactored, maybe in a different PR

TurboTurtle:
> @TurboTurtle @pmoravec one thing that really threw me here on the `apache2` plugin was the fact that line 48 mentions `subdir`, this in-fact is a wrong thing to label. This will exclude files or dirs under the `/var/log/<apachepkg>/`
> 
> I think, that should be renamed or refactored, maybe in a different PR

That's the intent though...so that the `apache` plugin doesn't collect them, as these should be collected by the `openstack_*` plugins. The forbidden paths here are plugin-specific, if you're seeing something else then we have a deeper problem in `Plugin`.

arif-ali:
> > @TurboTurtle @pmoravec one thing that really threw me here on the `apache2` plugin was the fact that line 48 mentions `subdir`, this in-fact is a wrong thing to label. This will exclude files or dirs under the `/var/log/<apachepkg>/`
> > I think, that should be renamed or refactored, maybe in a different PR
> 
> That's the intent though...so that the `apache` plugin doesn't collect them, as these should be collected by the `openstack_*` plugins. The forbidden paths here are plugin-specific, if you're seeing something else then we have a deeper problem in `Plugin`.

I agree totally

I think it was the naming of `subdir` that threw me off, because it may not be a sub directory, it could be a file with that name too

arif-ali:
yeah, I've had a +1 from a colleague already, and this will already provide something that we are not getting. Would be awesome to get through, which will give enough testing before 4.5.6 when we will potentially cutting our next `deb` release

PR #3228, opened 2023-05-10
[rhc] New plugin for RHC

jcastill:
RHC is a client tool and daemon that connects the system to 
Red Hat hosted services enabling system and
subscription management. This plugin captures
configuration files and the output of 'rhc status'.

Closes: RHBZ#2196649

---

### DIFF
--- a/sos/report/plugins/rhc.py
+++ b/sos/report/plugins/rhc.py
@@ -1,0 +1,49 @@
+# Copyright (C) 2023 Red Hat, Inc., Jose Castillo <jcastillo@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, RedHatPlugin
+
+
+class Rhc(Plugin, RedHatPlugin):
+
+    """
+    RHC is a client tool and daemon that connects the system
+    to Red Hat hosted services enabling system and
+    subscription management. This plugin captures
+    configuration files and the output of 'rhc status'.
+    """
+    short_desc = 'Red Hat client for remote host configured services'
+
+    plugin_name = "rhc"
+    packages = ("rhc", )
+
+    def setup(self):
+        self.add_copy_spec([
+            "/etc/rhc/*",
+        ])
+
+        self.add_cmd_output([
+            "rhc status",
+        ])
+
+    def postproc(self):
+        # hide workers/foreman_rh_cloud.toml FORWARDER_PASSWORD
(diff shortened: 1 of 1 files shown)


sos/report/plugins/rhc.py
```diff
@@ -0,0 +1,33 @@
...
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, RedHatPlugin
+
+
+class Rhc(Plugin, RedHatPlugin):
+
+    """
+    This plugin collects configuration of the RHC package.
+    """
```
TurboTurtle, line 18: Since this is what gets printed when using `sos help report.plugins.rhc`, let's flesh this out a bit further - what you have in the commit message is actually great for using in the docstring.

sos/report/plugins/rhc.py
```diff
@@ -0,0 +1,33 @@
...
+    short_desc = 'Red Hat client for remote host configured services'
+
+    plugin_name = "rhc"
+    packages = ("rhc", )
+
+    def setup(self):
+        self.add_copy_spec([
+            "/etc/rhc/*",
```
TurboTurtle, line 26: No secrets we need to obfuscate in any of the files here?

jcastill: I haven't found references to secrets in the documentation. There's an entry for user and password in the command line tool, but nothing for /etc/rhc/*. I've asked to the original bug reporter.

pmoravec: Per https://github.com/theforeman/foreman-operations-collection/blob/develop/roles/cloud_connector/templates/foreman_rh_cloud.toml.j2 , anything behind `FORWARDER_PASSWORD=` is a password. Maybe not quote marks, but we can obfuscate also them. (and maybe I would use a safe variant with optional blank spaces around `=`, just to be sure).

sos/report/plugins/rhc.py
```diff
@@ -0,0 +1,49 @@
...
+        #
+        # "FORWARDER_PASSWORD=F0rW4rd3RPassW0rD"
+        #
+        # to
+        #
+        # "FORWARDER_PASSWORD= ********
+
+        self.do_path_regex_sub("/etc/rhc/workers/foreman_rh_cloud.toml",
+                               r"(FORWARDER_PASSWORD\s*=\s*)(.+)(\"\,)",
+                               r"\1 ********\3")
```
pmoravec, line 48: A very minor nitpick: replacement string without the extra space (`r"\1********\3"`), to prevent confusion sos adds a space here.

pmoravec (approved):

TurboTurtle (commented):
Looks good, aside from the inserted space Pavel highlighted. Let's get that cleared up and then this is ready I think.

jcastill:
Ack, pushing again without the white space

pmoravec (approved):

TurboTurtle (approved):

PR #3234, opened 2023-05-13
[lxd] Add logs from the snap

arif-ali:
---

### DIFF
--- a/sos/report/plugins/lxd.py
+++ b/sos/report/plugins/lxd.py
@@ -44,6 +44,23 @@ class LXD(Plugin, UbuntuPlugin):
         snap_list = self.exec_cmd('snap list lxd')
         if snap_list["status"] == 0:
             self.add_cmd_output("lxd.buginfo", pred=lxd_pred)
+
+            self.add_copy_spec([
+                '/var/snap/lxd/common/config',
+                '/var/snap/lxd/common/global-conf',
+                '/var/snap/lxd/common/lxc/local.conf',
+                '/var/snap/lxd/common/lxd/logs/*/*.conf',
+            ])
+
+            if not self.get_option("all_logs"):
+                self.add_copy_spec([
+                    '/var/snap/lxd/common/lxd/logs/*.log',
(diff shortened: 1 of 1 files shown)


pmoravec (approved):
ACK to the change itself, please follow DCO and have in the commit message "Closes: #3234" or similar.

sos/report/plugins/lxd.py
```diff
@@ -44,6 +44,17 @@ def setup(self):
         snap_list = self.exec_cmd('snap list lxd')
         if snap_list["status"] == 0:
             self.add_cmd_output("lxd.buginfo", pred=lxd_pred)
+
+            if not self.get_option("all_logs"):
+                self.add_copy_spec([
+                    '/var/snap/lxd/common/lxd/logs/*.log',
+                    '/var/snap/lxd/common/lxd/logs/*/*.log',
+                    '/var/snap/lxd/common/lxd/logs/*/*.conf',
```
pmoravec, line 52: Hm... when re-reading the particular paths to collect: are not config files important enough to be collected every time? They are often older than logfiles so when logs are over size limit, sos report would collect newest logs but not configs (when --all-lags is *not* used).

Does not it deserve to mention configs as an extra `copy_path` in the `else` branch? (or what do config files do in a `log` directory..? :-o)

arif-ali: The .conf files would be collected with the below line in `else` as seen in https://paste.ubuntu.com/p/n3PYq46XBS/

happy to separate it out, if that is preferred

pmoravec: If you know the whole `/var/snap/lxd/common/lxd/logs/**` in `else` branch will always be below default size limit, no need to separate. If you are unsure and need to collect configs while optionally truncate older logs, let separate the configs.

arif-ali: I think I'll separate it, I think that makes more sense

pmoravec (approved):

TurboTurtle (approved):

PR #3237, opened 2023-05-14
[plugins] collect strings before commands

pmoravec:
When plugin timeouts during commands execution/collection, ensure we have collected sos_strings for tailed files.

Resolves: #3237

---


pmoravec:
Reproducer of the underlying problem:

1) Ensure a plugin would collect a tail of a file over size limit, e.g. in `logs` plugin collecting `/var/log/messages`, generate sufficient file content:

    base64  < /dev/urandom | head -c 26M >> /var/log/messages

2) Add to the plugin:

    self.add_cmd_output('sleep 10') 

3) Call:

    sos report -o logs --plugin-timeout=9 --batch --build

4) Check the `var/log/messages in the report's directory is a broken symlink to sos_strings.


Credits for the reproducer and idea of fix:  Ameya Patil (rhbz#2203141).

TurboTurtle:
Ack to the quick fix, however I think we should take some time to review the overlapping use case of `_collect_tailed_files()` and `_collect_strings()` - i.e. can we reduce this separation at all?

PR #3238, opened 2023-05-16
[zvm] Don't require 'vmcp' and 'cpint' kernel modules

rmetrich:
These modules are not present anymore, at least on RHEL8. The modules are not needed to execute 'vmcp' commands.

---

### DIFF
--- a/sos/report/plugins/zvm.py
+++ b/sos/report/plugins/zvm.py
@@ -6,7 +6,7 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.report.plugins import Plugin, IndependentPlugin, SoSPredicate
+from sos.report.plugins import Plugin, IndependentPlugin
 from sos.utilities import is_executable
 
 
@@ -17,9 +17,6 @@ class ZVM(Plugin, IndependentPlugin):
     commands = ('vmcp', 'hcp')
 
     def setup(self):
-
-        zvm_pred = SoSPredicate(self, kmods=['vmcp', 'cpint'])
-        self.set_cmd_predicate(zvm_pred)
 
         self.vm_cmd = None
         for cmd in self.commands:


TurboTurtle (commented):
The plugin is currently written to be distro-agnostic, so I'd like to either get a confirmation that these kernel modules are not used on other distributions or we'd need to split up the plugin to be distribution sensitive again.

Alternatively, if RHEL (and derivatives) are the only supported OS(es) for Z environments, then we can take this change as is (and ideally should replace `IndependentPlugin` with `RedHatPlugin`).

I'd just like to be 100% sure before merging.

pmoravec:
If I understood https://www.ibm.com/docs/en/linux-on-systems?topic=c-vmcp properly, then `vmcp` kmod is available on (some) RHEL8.6, still.

@rmetrich , you say there are systems with no `vmcp` or `cpinit` kmod (even available), where either `vmcp` or `hcp` command works?

If so, maybe an idea of improvement of `kmod` predicate that would help here: have a list of all kmods available on the system, and apply the predicate to *intersection* of the list and `kmods` list (i.e. apply the `kmods` list to *available* kmods only).

rmetrich:
> If I understood https://www.ibm.com/docs/en/linux-on-systems?topic=c-vmcp properly, then `vmcp` kmod is available on (some) RHEL8.6, still.
> 
> @rmetrich , you say there are systems with no `vmcp` or `cpinit` kmod (even available), where either `vmcp` or `hcp` command works?
> 
> If so, maybe an idea of improvement of `kmod` predicate that would help here: have a list of all kmods available on the system, and apply the predicate to _intersection_ of the list and `kmods` list (i.e. apply the `kmods` list to _available_ kmods only).

I don't know, all I did is install a s390x KVM and tried to execute "modprobe vmcp" and there was no such module.
Then commenting out the predicate, I could see "vmcp" commands were just working fine anyway.

pmoravec:
Hi @BryanQuigley and @TurboTurtle , as you are authors of `zvm` plugin: is the predicate at https://github.com/sosreport/sos/blob/main/sos/report/plugins/zvm.py#L21-L22 still valid? I am unable to find `vmcp` or `cpint` kmods on either `s390x` system I checked (of several RHEL versions) - despite I have `vmcp` command available .

BryanQuigley:
I do not have access to s390x systems anymore so I can't speak to this, sorry!

TurboTurtle:
I also no longer have access to an s390x box.

If we're not seeing the kmod on current/previous versions of supported OSes, then I think we're safe to drop the predicate, no objections here.

pmoravec (approved):

TurboTurtle (approved):

arif-ali:
> EDIT: @arif-ali is there s390x support on Ubuntu? Do we need to check for these kmods there, or alternatively should we move this off of `IndependentPlugin`?

We do have s390x pacages, so based on that we have the support there. Although, I am not sure how I can get access to these boxes for testing

PR #3239, opened 2023-05-16
powerpc: collect RMC status logs

sourabhjains:
Resource Monitoring and Control (RMC) responsible for establishing management connection between HMC (Hardware Management Console) and Logical partition (LPAR).

Add commands to collect the status/info of RMC service, node, and other system resource. Additionally collect RSCT node ID file.

---

### DIFF
--- a/sos/report/plugins/powerpc.py
+++ b/sos/report/plugins/powerpc.py
@@ -39,7 +39,8 @@ class PowerPC(Plugin, IndependentPlugin):
                 "/proc/swaps",
                 "/proc/version",
                 "/dev/nvram",
-                "/var/lib/lsvpd/"
+                "/var/lib/lsvpd/",
+                "/etc/ct_node_id"
             ])
             self.add_cmd_output([
                 "ppc64_cpu --info",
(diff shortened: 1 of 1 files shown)


sos/report/plugins/powerpc.py
```diff
@@ -87,7 +92,8 @@ def setup(self):
                 "lsslot",
                 "amsstat"
             ])
-            self.add_service_status("hcn-init")
+            self.add_service_status(["hcn-init",
+                                     "ctrmc.service"])
```
TurboTurtle, line 96: Minor nitpick - sos style is to indent like so:

```python
self.add_service_status([
    'hcn-init',
    'ctrmc'
])
```

Also note that you should not need to include the `.service` suffix for this call.

TurboTurtle (commented):
Looks good for the most part, minor note below.

sos/report/plugins/powerpc.py
sourabhjains: Thanks for the review @TurboTurtle 

Let me update the patch.

Do you recommend checking sos changes with any python linter to avoid such mistakes in future?

sourabhjains: Thanks for the input @TurboTurtle

PR #3256, opened 2023-05-27
[cxl] Add a new Compute Express Link (CXL) plugin

sscargal:
Compute Express Link (CXL) is an industry-supported Cache-Coherent Interconnect for Processors, Memory Expansion, and Accelerators.

The CXL Consortium (https://www.computeexpresslink.org/) is an open industry standard group formed to develop technical specifications that facilitate breakthrough performance for emerging usage models while supporting an open ecosystem for data center accelerators and other high-speed enhancements.

There are currently three specifications (v1.1, 2.0, and v3.0) publicly available. CPUs are available from Intel and AMD that support CXL v1.1. Hardware is emerging that supports up to CXL v3.0.

This plugin uses open-source utilities from
https://github.com/pmem/ndctl that introduce new and updated commands to support CXL devices in the Linux Kernel.

---

### DIFF
--- a/sos/report/plugins/cxl.py
+++ b/sos/report/plugins/cxl.py
@@ -1,0 +1,46 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, IndependentPlugin
+
+
+class cxl(Plugin, IndependentPlugin):
+    """This plugin collects data from Compute Express Link (CXL) devices
+    """
+
+    short_desc = 'Compute Express Link (CXL)'
+    plugin_name = 'cxl'
+    profiles = ('storage', 'hardware', 'memory')
+    # Utilities can be installed by package or self compiled
+    packages = ('cxl-cli', 'daxctl')
+    commands = ('cxl', 'daxctl')
+
+    def setup(self):
+        """ Use the daxctl-list(1) command to collect disabled, devices,
+        mapping, and region information
+
+        Output is JSON formatted
+        """
+        self.add_cmd_output([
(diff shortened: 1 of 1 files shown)


sos/report/plugins/cxl.py
```diff
@@ -0,0 +1,109 @@
...
+    profiles = ('storage', 'hardware', 'memory')
+    # Utilities can be installed by package or self compiled
+    packages = ('cxl-cli', 'daxctl')
+    commands = ('cxl', 'daxctl')
+
+    def setup(self):
+        """ Use the daxctl-list(1) command to collect:
+        -i, --idle
+           Include idle (not enabled / zero-sized) devices in the listing
+
```
TurboTurtle, line 27: I don't think we need to carry `--help` output in comments in the plugin, especially for options we don't end up using.

sos/report/plugins/cxl.py
```diff
@@ -0,0 +1,109 @@
...
+            "daxctl version",
+            "daxctl list",
+            "daxctl list -iDRM"
+        ])
+
+        """ Use the cxl-list(1) command to collect data about
+        all CXL devices.
+
+        -M, --memdevs
+           Include CXL memory devices in the listing
```
TurboTurtle, line 49: Ditto.

TurboTurtle (commented):
Ack to the plugin code. Don't need to carry `--help` output though.

sos/report/plugins/cxl.py
sscargal: Thanks for the feedback. The args will be used in the next revision of the plugin as the OS and Hardware availability mature. I can remove them now and re-add them in the near future or keep them. Let me know.

sscargal: Help info was removed and replaced with a simpler message.

sos/report/plugins/cxl.py
sscargal: Help info was removed and replaced with a simpler message.

TurboTurtle (approved):

pmoravec (approved):

PR #3257, opened 2023-05-28
[pm] Update plugins to use package_manager

arif-ali:
Remove `check_enabled()` for the maas plugin, as that is now not required with `SnapPackageManager` now enabled for the `UbuntuPlugin`

Use package manager to detrmine is package is a snap rather than `snap list`

---

### DIFF
--- a/sos/report/plugins/lxd.py
+++ b/sos/report/plugins/lxd.py
@@ -17,7 +17,7 @@ class LXD(Plugin, UbuntuPlugin):
     plugin_name = 'lxd'
     profiles = ('container',)
     packages = ('lxd',)
-    commands = ('lxd',)
+    commands = ('lxc', 'lxd',)
 
     def setup(self):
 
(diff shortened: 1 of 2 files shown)


sos/report/plugins/maas.py
```diff
@@ -56,18 +56,9 @@ def _remote_api_login(self):
...
-        if super().check_enabled():
-            # deb-based MAAS and existing triggers
-            return True
-        # Do we have the snap installed?
-        return self._is_snap_installed()
-
     def setup(self):
-        self._is_snap = self._is_snap_installed()
+        self._is_snap = self.policy.package_manager.pkg_by_name(
+            'maas')['pkg_manager'] == 'snap'
```
arif-ali, line 61: seems like the function here for `pkg_by_name('maas')` is coming back with `Null`, and hence causing the avocoda tests to fail.

Maybe, I better option is to re-instate the `_is_snap_installed()` function below

```python
    def _is_snap_installed(self):
        maas_pkg = self.policy.package_manager.pkg_by_name('maas')
        if maas_pkg:
        return False
```

sos/report/plugins/lxd.py
```diff
@@ -41,8 +41,8 @@ def setup(self):
         lxd_pred = SoSPredicate(self, kmods=lxd_kmods,
                                 required={'kmods': 'all'})
 
-        snap_list = self.exec_cmd('snap list lxd')
-        if snap_list["status"] == 0:
+        if self.policy.package_manager.pkg_by_name(
+                'lxd')['pkg_manager'] == 'snap':
```
pmoravec, line 45: Theoretically, if one *manually* enables this plugin on a system without `lxd` package installed, we would get `TypeError: 'NoneType' object is not subscriptable` here (am I right?). Not sure how feasible this option is / if we should catch that in the code.

TurboTurtle:
Good catch, ack to handling `lxd` in the same way as `maas`, then this PR looks good to merge.

PR #3259, opened 2023-05-29
[specfile] align with Fedora spec file

sandrobonazzola:
---


sos.spec
```diff
@@ -2,21 +2,23 @@ Summary: A set of tools to gather troubleshooting information from a system
...
-BuildRoot: %{_tmppath}/%{name}-%{version}-%{release}-buildroot
 BuildArch: noarch
-Url: https://github.com/sosreport/sos/
+Url: https://github.com/sosreport/sos
 BuildRequires: python3-devel
 BuildRequires: python3-setuptools
-BuildRequires: gettext
 Requires: python3-rpm
+Requires: tar
+Requires: xz
```
pmoravec, line 13: We removed those dependencies in https://github.com/sosreport/sos/commit/d852669f6e0bcf61cdb561021da7bee3ef1cc1ed - see rationale there.

(Why) is required to bring back, please?

sos.spec
```diff
@@ -2,21 +2,23 @@ Summary: A set of tools to gather troubleshooting information from a system
...
 Requires: python3-pexpect
 Requires: python3-setuptools
 Recommends: python3-magic
+Requires: python3-requests
```
pmoravec, line 17: `python3-requests` is mandatory just for uploading to a SFTP server. Where Fedora sosreports are uploaded to / what protocol is used (even in Red Hat downstream, we use `Recommends` - do you really need the hard dependency here)?

I am ok with it, just I am asking as I am unsure if the requirement (as is) is such demanded.

sos.spec
```diff
@@ -2,21 +2,23 @@ Summary: A set of tools to gather troubleshooting information from a system
...
 Recommends: python3-pyyaml
 Obsoletes: sos-collector <= 1.9
+# For the _tmpfilesdir macro.
+BuildRequires: systemd
```
pmoravec, line 21: Good point.

sos.spec
sandrobonazzola: sorry, leftover from rebases within Fedora, removing there as well.

sos.spec
sandrobonazzola: moving to recommends as well

pmoravec (approved):

sos.spec
```diff
@@ -2,21 +2,22 @@ Summary: A set of tools to gather troubleshooting information from a system
 Name: sos
 Version: 4.5.4
 Release: 1%{?dist}
-Group: Applications/System
-Source0: https://github.com/sosreport/sos/archive/%{name}-%{version}.tar.gz
+Source0: https://github.com/sosreport/sos/archive/%{version}.tar.gz
```
pmoravec, line 5: This is unfortunate but this breaks rpm build from checkout, see e.g. https://github.com/sosreport/sos/pull/3259/checks?check_run_id=13833368826 .

curl-ing https://github.com/sosreport/sos/archive/sos4.5.4.tar.gz really does not work since it is redirected to https://codeload.github.com/sosreport/sos/tar.gz/sos-4.5.4 which does not exist. Contrary to:

    $ curl -vv https://github.com/sosreport/sos/archive/4.5.4.tar.gz 2>&1 | grep location
    < location: https://codeload.github.com/sosreport/sos/tar.gz/refs/tags/4.5.4
    $ curl https://codeload.github.com/sosreport/sos/tar.gz/refs/tags/4.5.4
    .. (this works)
    $

But the stored data is a tarball with content in `sos-4.5.4`. While `Source0` assumes the content to be in `%{version}` dir instead.

We can do some hack in `%py3_install`(?) but.. no idea how, atm.

pmoravec (changes requested):
The `Source0` breaks building from checkout.

sos.spec
sandrobonazzola: ok, I'll revert that line and keep it "downstream" change only.

sos.spec
```diff
@@ -2,21 +2,22 @@ Summary: A set of tools to gather troubleshooting information from a system
 Name: sos
 Version: 4.5.4
 Release: 1%{?dist}
-Group: Applications/System
 Source0: https://github.com/sosreport/sos/archive/%{name}-%{version}.tar.gz
```
pmoravec, line 5: @TurboTurtle , since we refer to nonexisting URL here, does it make sense to point here to `https://github.com/sosreport/sos/archive/%{version}.tar.gz` and update https://github.com/sosreport/sos/blob/main/.cirrus.yml#L107 accordingly?

PR #3260, opened 2023-05-29
[specfile] new fedora guidelines

sandrobonazzola:
Adapt the specfile to new fedora packaging guidelines.

- use pyproject macros (see also https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#_example_spec_file )
- use bin instead of sbin (see also [rhbz#2210708](https://bugzilla.redhat.com/show_bug.cgi?id=2210708)

Based on top of PR #3259

---

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -106,6 +106,7 @@ rpm_build_task:
         python3 setup.py sdist
         cp dist/sos*.tar.gz /rpmbuild/SOURCES/
         rpmbuild -bs sos.spec
+        dnf -y builddep /rpmbuild/SRPMS/sos*src.rpm
         rpmbuild -bb sos.spec
     # Retrieving the built rpm in later tasks requires knowing the exact name
     # of the file. To avoid having to juggle version numbers here, rename it
--- a/sos.spec
+++ b/sos.spec
@@ -28,11 +28,25 @@ support technicians and developers.
 %prep
 %setup -qn %{name}-%{version}
 
+%if 0%{?fedora} >= 39
+%generate_buildrequires
+%pyproject_buildrequires
+%endif
+
 %build
+%if 0%{?fedora} >= 39
+%pyproject_wheel
+%else
 %py3_build
+%endif
 
 %install
+%if 0%{?fedora} >= 39
+%pyproject_install
+%pyproject_save_files sos
+%else
 %py3_install '--install-scripts=%{_sbindir}'
+%endif
 
 install -d -m 755 %{buildroot}%{_sysconfdir}/%{name}
 install -d -m 700 %{buildroot}%{_sysconfdir}/%{name}/cleaner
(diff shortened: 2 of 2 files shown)


TurboTurtle:
I'm assuming we want to use this one over #3159?

Beyond that - I'm taking this to mean we need to install `python3-wheel` on the builder VMs?
```
error: Failed build dependencies:
	python3dist(wheel) is needed by sos-4.5.4-1.fc37.noarch
```

If so, that's fine - it'd just need to be added to the setup script for cirrus - here: https://github.com/sosreport/sos/blob/main/.cirrus.yml#L103

sandrobonazzola:
This one is rebased on top of #3259 ; the reason I pushed it as separate PR is because it introduces a breaking change moving the scripts from `sbin` to `bin` (on Fedora) and I wanted to be sure you're ok with the change.
I'll add the missing package into `cirrus.yml`

pmoravec (approved):

pmoravec:
Sigh, I should review failed tests prior ACKing a PR :)

CentOS Stream 9 does not contain `python3-wheel` and Fedora build fails on:

    error: Failed build dependencies:
	    python3dist(pexpect) is needed by sos-4.5.4-1.fc38.noarch
    	    python3dist(pyyaml) is needed by sos-4.5.4-1.fc38.noarch

that I dont understand why it can happen..?

TurboTurtle:
> This one is rebased on top of #3259 ; the reason I pushed it as separate PR is because it introduces a breaking change moving the scripts from `sbin` to `bin` (on Fedora) and I wanted to be sure you're ok with the change. I'll add the missing package into `cirrus.yml`


I suspect most/all downstreams roll their own spec files, and in that regard Fedora is our "closest" downstream, so I'm fine with the stock spec file to be mainly geared towards Fedora guidelines, as long as it can still be used for ad-hoc builds on non-Fedora rpm boxes.

In other words: totally cool to change to `bin`.

hroncok:
Is it intentional that RHEL 10 will not use this?

sandrobonazzola:
> Is it intentional that RHEL 10 will not use this?

It wasn't my intention, but I'm not the RHEL package maintainer.

hroncok:
I've noticed the package is built the new way in Fedora 39+ and the old way in Fedora ELN. I've tracked the origin of this conditional here, so I thought I'd ask. (I don't really care that much, IMHO using the new Python guidelines conditionally kinda eliminates their benefit anyway.)

sandrobonazzola:
> I've noticed the package is built the new way in Fedora 39+ and the old way in Fedora ELN. I've tracked the origin of this conditional here, so I thought I'd ask. (I don't really care that much, IMHO using the new Python guidelines conditionally kinda eliminates their benefit anyway.)

The condition on fedora >= 39 is because it failed to build on fedora 38 with the new macros. And as RHEL 9 and 8 are using completely different spec file I didn't care checking eln/el9 for this. I'll take a look at next  rebase

PR #3262, opened 2023-05-31
[cleaner] Separate process of preparing obfuscations

TurboTurtle:
This patchset is the first that should set us on a path to providing multi-process `--clean` support in a way that is not fragile or particularly disadvantageous for single-archive cleaning.

In order to facilitate multi-process down the road, we need to separate some tight bindings between parsers and mappings that exist today. The first of which, is how the mappings get prepared from archives before we do the actual obfuscation work.

This set introduces the use of `SoSPrepper`s, which will be used to pull select data out of archives and use it to prepare the parsers and mappings. This was previously done via `prep_files` dicts for archive abstractions and via additional methods in parsers and/or mappings to handle specific data from specific prep_file contents.

By separating this out, we gain flexibility to allow the individual Prepper entities to grok data as needed, rather than always relying on the existing regex parsing or special-case handling in mappings. This also allows us to provide a better framework for specific use cases in the future - e.g. we could create cluster-specific Preppers that ensure that we obfuscate cluster hosts together, so that their obfuscation numbering is sequential.

The next steps will include moving compiled regex handling directly to parsers, and not binding the mappings directly to the parsers.

Related: RH: SUPDEV-135

---

### DIFF
--- a/sos/cleaner/mappings/hostname_map.py
+++ b/sos/cleaner/mappings/hostname_map.py
@@ -41,7 +41,7 @@ class SoSHostnameMap(SoSMap):
     ]
 
     strip_exts = ('.yaml', '.yml', '.crt', '.key', '.pem', '.log', '.repo',
-                  '.rules')
+                  '.rules', '.conf', '.cfg')
 
     host_count = 0
     domain_count = 0
@@ -79,10 +79,6 @@ class SoSHostnameMap(SoSMap):
                         )
                         self._domains[_domain_to_inject] = _ob_domain
         self.set_initial_counts()
-
-    def load_domains_from_options(self, domains):
-        for domain in domains:
-            self.sanitize_domain(domain.split('.'))
 
     def get_regex_result(self, item):
         """Override the base get_regex_result() to provide a regex that, if
--- a/sos/cleaner/preppers/__init__.py
+++ b/sos/cleaner/preppers/__init__.py
@@ -47,7 +47,7 @@ class SoSPrepper():
     name = 'Undefined'
     priority = 100
 
-    def __init__(self):
+    def __init__(self, options):
         self.regex_items = {
             'hostname': set(),
             'ip': set(),
@@ -56,6 +56,7 @@ class SoSPrepper():
             'mac': set(),
             'username': set()
         }
+        self.opts = options
         self.soslog = logging.getLogger('sos')
         self.ui_log = logging.getLogger('sos_ui')
 
--- a/sos/cleaner/preppers/__init__.py
+++ b/sos/cleaner/preppers/__init__.py
@@ -1,0 +1,125 @@
+# Copyright 2023 Red Hat, Inc. Jake Hunsaker <jhunsake@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+import logging
+
+
+class SoSPrepper():
+    """
+    A prepper is a way to prepare loaded mappings with selected items within
+    an sos report prior to beginning the full obfuscation routine.
+
+    This was previously handled directly within archives, however this is a bit
+    cumbersome and doesn't allow for all the flexibility we could use in this
+    effort.
+
+    Preppers are separated from parsers but will leverage them in order to feed
+    parser-matched strings from files highlighted by a Prepper() to the
+    appropriate mapping for initial obfuscation.
+
+    Preppers may specify their own priority in order to influence the order in
+    which mappings are prepped. Further, Preppers have two ways to prepare
+    the maps - either by generating a list of filenames or via directly pulling
+    content out of select files without the assistance of a parser. A lower
+    priority value means the prepper should run sooner than those with higher
+    values.
+
+    For the former approach, `Prepper._get_$parser_file_list()` should be used
+    and should yield filenames that exist in target archives. For the latter,
+    the `Prepper._get_items_for_$map()` should be used.
+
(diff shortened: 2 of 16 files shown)


sos/cleaner/preppers/__init__.py
```diff
@@ -0,0 +1,126 @@
...
+    Preppers are separated from parsers but will leverage them in order to feed
+    parser-matched strings from files highlighted by a Prepper() to the
+    appropriate mapping for initial obfuscation.
+
+    Preppers may specify their own priority in order to influence the order in
+    which mappings are prepped. Further, Preppers have two ways to prepare
+    the maps - either by generating a list of filenames or via directly pulling
+    content out of select files without the assistance of a parser. A lower
+    priority value means the prepper should run sooner than those with higher
+    values.
```
pmoravec, line 32: Nitpick comment: swap 2nd and 3rd sentence (as priority levels follows the 1st sentence, while "two ways to prepare the maps" is followe by the next paragraph.

pmoravec:
Nice separation (and also unification of "prepping" methods like `s/load_hostname_into_map/mapping.add`).

tests/unittests/cleaner_tests.py
```diff
@@ -274,3 +279,35 @@ def test_ad_username(self):
...
+        self.prepper = SoSPrepper()
+        self.archive = SoSReportArchive(
+            archive_path='tests/test_data/sosreport-cleanertest-2021-08-03-qpkxdid.tar.xz',
+            tmpdir='/tmp'
+        )
+        self.host_prepper = HostnamePrepper()
+        self.ipv4_prepper = IPPrepper()
+
+    def test_parser_method_translation(self):
+        self.assertEqual([], self.prepper.get_parser_file_list('hostname', None))
```
pmoravec, line 300: Isn't it simpler to merge `test_ipv4_prepper_parser_files` into here? I.e. to have one method with a few asserts "a parser X has file_list Y"? And _maybe_ one assert per each one parser (if our focus here is to have coverage of parsers rather than coverage of `get_parser_file_list` translation "only")?

tests/unittests/cleaner_tests.py
```diff
@@ -274,3 +279,35 @@ def test_ad_username(self):
...
+
+    def test_mapping_method_translation(self):
+        self.assertEqual([], self.prepper.get_items_for_map('foobar', None))
```
pmoravec, line 303: Same comment like to `test_parser_method_translation`.

tests/unittests/cleaner_tests.py
```diff
@@ -274,3 +279,35 @@ def test_ad_username(self):
...
+
+    def test_hostname_prepper_map_items(self):
+        self.assertEqual(['cleanertest'], self.host_prepper.get_items_for_map('hostname', self.archive))
+
+    def test_ipv4_prepper_parser_files(self):
+        self.assertEqual(['sos_commands/networking/ip_-o_addr'], self.ipv4_prepper.get_parser_file_list('ip', self.archive))
+
+    def test_ipv4_prepper_invalid_parser_files(self):
+        self.assertEqual([], self.ipv4_prepper.get_parser_file_list('foobar', self.archive))
```
pmoravec, line 312: What is the purpose of this test? AFAIK `get_parser_file_list can be called only for non-disabled existing parsers - is the aim to test the `if hasattr(self, _check):` ? (maybe then have the analogy test for `get_items_for_map` as well..?)

sos/cleaner/mappings/hostname_map.py
```diff
@@ -80,10 +80,6 @@ def load_domains_from_map(self):
                         self._domains[_domain_to_inject] = _ob_domain
         self.set_initial_counts()
 
-    def load_domains_from_options(self, domains):
-        for domain in domains:
-            self.sanitize_domain(domain.split('.'))
-
```
pmoravec, line 86: How is this achieved in the new code, please..?

pmoravec:
Just the few minor issues / nitpicks found. But some testing is blocked by https://github.com/sosreport/sos/issues/3271 .

tests/unittests/cleaner_tests.py
TurboTurtle: This also tests the method name translation done inside `Prepper`. This means we call `get_foobar_file_list()` which doesn't exist, but here we want to simply return an empty list rather than generating a fatal exception for a non-existing method.

sos/cleaner/mappings/hostname_map.py
TurboTurtle: https://github.com/sosreport/sos/pull/3262/files#diff-affa4239dfe2658d3429ab817a6242bcab87ecb4e0959036e87b20ae2133c6e1R60-R61

sos/cleaner/preppers/mac.py
```diff
@@ -0,0 +1,28 @@
...
+
+from sos.cleaner.preppers import SoSPrepper
+
+
+class MacPrepper(SoSPrepper):
+    """
+    Prepper for sourcing the host's MAC address in order to prep the mapping.
+    """
+
+    name = 'mac'
```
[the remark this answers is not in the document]
TurboTurtle: We could add a test like the plugin compliance check, I guess? But I wonder if we should endeavor to make such a compliance test for all applicable areas if we're going to extend it outside of plugins? So not just preppers, parsers, and mappings - but also collect clusters, transports, policies, etc...

sos/cleaner/mappings/hostname_map.py
pmoravec: Ah I see, now it is more straighforward.

sos/cleaner/preppers/mac.py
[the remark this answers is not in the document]
pmoravec: Yeah, and each such test is more work than gain, I think.

pmoravec (approved):

PR #3274, opened 2023-06-15
[kernel] verify also kernel-redhat on RedHatPolicy

pmoravec:
RedHat distributions can deliver kernel in RPM named either 'kernel' or 'kernel-redhat', in which case we must verify both package names.

See e.g. https://fedoramagazine.org/changes-eln-kernel-rpm-nvr/ for the background.

It can be bit controversial to customize an `IndependentPlugin` subclass and not using `RedHatPlugin`. But having `class RedHatKernel(Plugin, RedHatPlugin)` would need to replace:

    - class Kernel(Plugin, IndependentPlugin):
    + class Kernel(Plugin, UbuntuPlugin, OpenEulerPlugin, CosPlugin, ..):

to list all other distros-plugins, *and maintain the list*. Since `Kernel` plugin being a subclass of `IndependentPlugin` would be honoured also on `redhat` distro/policy.

From a few options (like the "list all `*Plugin`s") I found neither as optimal and the chosen one as the least intruding.

Resolves: #3274

---

### DIFF
--- a/sos/report/plugins/kernel.py
+++ b/sos/report/plugins/kernel.py
@@ -7,6 +7,7 @@
 # See the LICENSE file in the source distribution for further information.
 
 from sos.report.plugins import Plugin, IndependentPlugin, PluginOpt
+from sos.policies.distros.redhat import RedHatPolicy
 import glob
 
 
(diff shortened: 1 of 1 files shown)


TurboTurtle (approved):
LGTM. Agreed it's not an ideal way to to do this, but it is the best/cleanest one we have available.

PR #3282, opened 2023-06-21
[vault] Add vault plugin

arif-ali:
---

### DIFF
--- a/sos/report/plugins/vault.py
+++ b/sos/report/plugins/vault.py
@@ -1,0 +1,55 @@
+# Copyright (C) 2023 Canonical Ltd., Arif Ali <arif.ali@canonical.com>
+#
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, UbuntuPlugin
+
+
+class Vault(Plugin, UbuntuPlugin):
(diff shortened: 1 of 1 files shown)


sos/report/plugins/vault.py
```diff
@@ -0,0 +1,60 @@
...
+    profiles = ('sysmgmt', 'security')
+    package = ('vault',)
+
+    def setup(self):
+
+        vault_cfg = "/var/snap/vault/common/vault.hcl"
+
+        self.add_service_status('vault')
+
+        self.add_journal('vault')
```
TurboTurtle, line 33: If you use the `services = ('vault',)` class attr these collections become implicit.

sos/report/plugins/vault.py
```diff
@@ -0,0 +1,60 @@
...
+                    if not line:
+                        continue
+                    words = line.split('=')
+                    if words[0].strip() == 'api_addr':
+                        api_addr = words[1].strip('\" ')
+        except IOError as error:
+            self._log_error('Could not open conf file %s: %s' %
+                            (vault_cfg, error))
+
+        if api_addr:
```
TurboTurtle, line 49: There's an edge case here that `api_addr` could be undefined if the file is missing that line (or is malformed), or we hit an error reading it.

sos/report/plugins/vault.py
```diff
@@ -0,0 +1,60 @@
...
+            self.add_cmd_output("vault status", env={'VAULT_ADDR': api_addr})
+
+    def postproc(self):
+        super(Vault, self).postproc()
```
TurboTurtle, line 53: This is unnecessary - the base `postproc()` simply uses `pass`.

TurboTurtle (approved):

PR #3287, opened 2023-06-23
[cos]Fixed issue by adding valid plugin class and change CosLogs as a hierarchy of Independent logs

vteratipally:
Fixed issue by adding valid plugin class 
change CosLogs as a hierarchy of Independent logs

Signed-off-by: Varsha Teratipally [teratipally@google.com](mailto:teratipally@google.com)

---

### DIFF
--- a/sos/policies/distros/cos.py
+++ b/sos/policies/distros/cos.py
@@ -41,7 +41,7 @@ class CosPolicy(LinuxPolicy):
         super(CosPolicy, self).__init__(sysroot=sysroot, init=init,
                                         probe_runtime=probe_runtime,
                                         remote_exec=remote_exec)
-        self.valid_subclasses += [CosPolicy]
+        self.valid_subclasses += [CosPlugin]
 
     @classmethod
     def check(cls, remote=''):
--- a/sos/report/plugins/logs.py
+++ b/sos/report/plugins/logs.py
@@ -118,7 +118,7 @@ class IndependentLogs(LogsBase, IndependentPlugin):
     profiles = ('system', 'hardware', 'storage')
 
 
-class CosLogs(LogsBase, CosPlugin):
+class CosLogs(IndependentLogs, CosPlugin):
     option_list = [
         PluginOpt(name="log_days", default=3,
                   desc="the number of days logs to collect")


sos/policies/distros/cos.py
```diff
@@ -41,7 +41,7 @@ def __init__(self, sysroot=None, init=None, probe_runtime=True,
         super(CosPolicy, self).__init__(sysroot=sysroot, init=init,
                                         probe_runtime=probe_runtime,
                                         remote_exec=remote_exec)
-        self.valid_subclasses += [CosPolicy]
+        self.valid_subclasses += [CosPlugin]
```
arif-ali, line 44: curious, the `valid_subclasses` already has `CosPlugin` already at line 36

sos/report/plugins/logs.py
```diff
@@ -118,7 +118,7 @@ class IndependentLogs(LogsBase, IndependentPlugin):
     profiles = ('system', 'hardware', 'storage')
 
 
-class CosLogs(LogsBase, CosPlugin):
+class CosLogs(IndependentLogs, CosPlugin):
```
arif-ali, line 121: Does this mean that you don't want to be collecting all the logs defined in `LogsBase` as above?

arif-ali:
can you also rebase your commit to the `main` branch that solves the CI for the snap build, and hence the rest of the checks

sos/policies/distros/cos.py
TurboTurtle: Line 36 needs to be removed, as it gets overwritten by the `super()` in `__init__()` from the base `Policy` class.

sos/report/plugins/logs.py
TurboTurtle: This will make it so that `CosLogs` has the `IndependentPlugin` tagging class assigned to it, which will enable it for _all_ distros, which is precisely why we made the change to split out `LogsBase`.

The change to `valid_subclasses` is correct (and can be made "more" correct by removing line 36), but this change here is not going to get us where we need to be.

arif-ali:
@vteratipally are you still wanting to get this though?

PR #3288, opened 2023-06-23
[tests] Make options_tests faster by calling them on a few plugins only

pmoravec:
options_tests.py tests -n option which means tens of plugins are redundantly collected. Testing -o option is effectively the same while the sosreport run needs a half time only.

The functionality of -n option is tested elsewhere, thus the change does not shrink test coverage.

Resolves: #3288

---

### DIFF
--- a/tests/report_tests/options_tests/options_tests.py
+++ b/tests/report_tests/options_tests/options_tests.py
@@ -22,8 +22,8 @@ class OptionsFromConfigTest(StageTwoReportTest):
     def test_case_id_from_config(self):
         self.assertTrue('8675309' in self.archive)
 
-    def test_plugins_skipped_from_config(self):
-        self.assertPluginNotIncluded(['networking', 'logs'])
+    def test_plugins_only_from_config(self):
+        self.assertOnlyPluginsIncluded(['host', 'kernel'])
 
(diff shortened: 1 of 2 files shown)


tests/report_tests/options_tests/options_tests.py
```diff
@@ -41,5 +41,5 @@ def test_plugopts_actually_set(self):
 
     def test_effective_options_logged_correctly(self):
         self.assertSosLogContains(
-            "effective options now: --batch --case-id 8675309 --plugopts kernel.with-timer=on,kernel.trace=yes --skip-plugins networking,logs"
+            "effective options now: --batch --case-id 8675309 --only-plugins host,kernel --plugopts kernel.with-timer=on,kernel.trace=yes"
```
pmoravec, line 44: The `effective options` in fact do contain `--plugopts kernel.with-timer=on,kernel.trace=yes,networking.traceroute=yes`, even the option for disabled `networking` plugin. That is a tiny bug of resolving options that has no effect to the data collected (so imho not worth fixing it). Just this assert string is made in the way to work well even if the tiny bug would be fixed.

TurboTurtle (approved):
ack

Issue #3278, opened 2023-06-20
[grafana] Collect data for snap installation


arif-ali:
Files that should be collected

```
/var/snap/grafana/current/conf/grafana.ini
```

`postproc()` should look out for the following parameters for obfuscations

```
password = ...
secret_key = ...
client_secret = ...
```

For logs:

```
/var/snap/grafana/common/data/log/grafana.log
```

`--all-logs` should give us:

```
/var/snap/grafana/common/data/log/grafana.log*
```

The command for `grafana-cli` should be `grafana.grafana-cli` to collect any information

PR #3289, opened 2023-06-27
[grafana] conf and logs from snap added

GairePravesh:
files added:
    /var/snap/grafana/common/data/log/grafana.log*
    /var/snap/grafana/current/conf/grafana.ini

cmds added:
    snap info grafana
    grafana.grafana-cli plugins ls
    grafana.grafana-cli plugins list-remote

Closes: #3278

---


sos/report/plugins/grafana.py
```diff
@@ -19,29 +19,65 @@ class Grafana(Plugin, IndependentPlugin):
...
+                regexp,
+                r"\1*********"
+            )
+        else:
+            regexp = r"((?m)^\s*(%s)\s*=\s*)(.*)" % "|".join(protect_keys)
+            self.do_path_regex_sub(
+                "/etc/grafana/grafana.ini",
+                regexp,
+                r"\1*********"
+            )
```
pmoravec, line 83: If the only change is in filename, maybe it is easier to have:

    inifile = `"/var/snap/grafana/current/conf/grafana.ini"` if self_is_snap else "/etc/grafana/grafana.ini"
    regexp = r"((?m)^\s*(%s)\s*=\s*)(.*)" % "|".join(protect_keys)
    self.do_path_regex_sub(
        inifile,
        regexp,
        r"\1*********"
    )

sos/report/plugins/grafana.py
```diff
@@ -19,29 +19,65 @@ class Grafana(Plugin, IndependentPlugin):
...
+                "/etc/sysconfig/grafana-server",
+            ])
 
     def postproc(self):
         protect_keys = [
-            "admin_password", "secret_key"
+            "admin_password",
+            "secret_key",
+            "password",
+            "client_secret"
```
pmoravec, line 68: +1, that helps also to non-snap instances.

arif-ali (commented):
@GairePravesh Thanks for your contribution

probably nits more than anything

*  What about creating a variable called `grafana_cli`, and based if it is snap or standard, and then use that variable to do the `grafana-cli` commands. This way, if more commands are needed we add 1 line rather than duplication
* We could probably do the same with both the log path and config path, as they would be similar

i.e.

```python
if self._is_snap:
    grafana_cli = "grafana.grafana-cli"
else:
    grafana_cli = "grafana-cli"

self.add_cmd_output([
    "%s plugin ls" % grafana_cli,
    "%s plugin list remote" % grafana_cli,
    "%s -v" % grafana_cli,
])
```

pmoravec:
Generally ACK, just a though: since the "snap instance or not" call flows differ a lot, it would maybe make more sense to have `SnapPlugin` class likewise we have `RedHatPlugin` or `DebianPlugin`? Though the `Plugin` hierarchy is likely distribution-based, not package-manager-based. But maybe if some other plugins might benefit from this in future, it should be easy to add `SnapPlugin` (likewise https://github.com/sosreport/sos/blob/main/sos/policies/distros/ubuntu.py#L49)?

arif-ali:
> Generally ACK, just a though: since the "snap instance or not" call flows differ a lot, it would maybe make more sense to have `SnapPlugin` class likewise we have `RedHatPlugin` or `DebianPlugin`? Though the `Plugin` hierarchy is likely distribution-based, not package-manager-based. But maybe if some other plugins might benefit from this in future, it should be easy to add `SnapPlugin` (likewise https://github.com/sosreport/sos/blob/main/sos/policies/distros/ubuntu.py#L49)?

I do like the idea, I will have a look at this in the background, and see how that looks with refactoring a few of the other plugins that has this already, like `maas` and `lxd`

TurboTurtle (commented):
> ```python
> if self._is_snap:
>     grafana_cli = "grafana.grafana-cli"
> else:
>     grafana_cli = "grafana-cli"
> 
> self.add_cmd_output([
>     "%s plugin ls" % grafana_cli,
>     "%s plugin list remote" % grafana_cli,
>     "%s -v" % grafana_cli,
> ])
> ```

This, but please note that f-strings are preferred per the contributions guidelines.

The intent of the changes looks good, just simplifying the flow as suggested above would be my 2 cents.

As far as a `SnapPlugin` goes, I'm not sure how well we'd be able to have that in tandem with distro-based enablement without things getting real fragile.

Perhaps instead we could add a helper method like `self.is_pkg_snap('$pkg')`  (and accompanying `is_pkg_rpm()`, `is_pkg_deb()` etc...) to better support these checks within the current scope of enablement desgns?

arif-ali (changes requested):

sos/report/plugins/grafana.py
```diff
@@ -19,29 +19,53 @@ class Grafana(Plugin, IndependentPlugin):
...
+        self.add_cmd_output(list(filter(None, add_cmds)))
+
+        log_file_pattern = "*.log*" if self.get_option("all_logs") else "*.log"
 
         self.add_copy_spec([
-            "/etc/grafana/",
-            "/etc/sysconfig/grafana-server",
+            log_path + log_file_pattern,
+            config_path,
+            "/etc/sysconfig/grafana-server" if not self._is_snap else None
```
TurboTurtle, line 54: `add_copy_spec()` also handles non-existing paths gracefully, so we can just have it listed here and it will be ignored for snap installations.

sos/report/plugins/grafana.py
```diff
@@ -19,29 +19,53 @@ class Grafana(Plugin, IndependentPlugin):
...
-            "grafana-cli plugins list-remote",
-            "grafana-cli -v",
-            "grafana-server -v",
-        ])
+        add_cmds = [
+            f'{grafana_cli} plugins ls',
+            f'{grafana_cli} plugins list-remote',
+            'snap info grafana' if self._is_snap else None,
+            f'{grafana_cli} -v' if not self._is_snap else None,
+            'grafana-server -v' if not self._is_snap else None
```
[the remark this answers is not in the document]
TurboTurtle: Likewise we don't need to gate the `snap info` command, we can just tuck an `add_cmd_output()` under the conditional on line 30.

arif-ali:
This is being closed in favour of #3410, Pravesh will not longer be contributing

PR #3292, opened 2023-06-28
[clean] Respect permissions of sanitised files

pmoravec:
When copying files we applied a substitution in, we must replace just original file content (`shutil.copyfile`) and not also its stat data (`shutil.copy`).

Resolves: #3292

---

### DIFF
--- a/sos/cleaner/__init__.py
+++ b/sos/cleaner/__init__.py
@@ -778,7 +778,7 @@ third party.
                                        % (short_name, err), caller=arc_name)
             tfile.seek(0)
             if subs:
-                shutil.copy(tfile.name, filename)
+                shutil.copyfile(tfile.name, filename)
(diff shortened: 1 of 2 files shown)


pmoravec:
I was thinking about it, but thought "would that test be usefull / why we would cause here a regression?" - but if also you have raised the concern about adding a test, I will add one (once I find proper `assert` method, probably comparing stat objects..?).

pmoravec:
I added a test case, but it does pass even *without* the codechange; whenever `sos` is run within `avocado`, the collected (sanitised) file has proper permissions. I dont understand why..

TurboTurtle (approved):
The test looks right, but the behavior from avocado is odd. I'm of the mind to accept this now, and investigate the test suite separately as manual inspection shows this working as intended.

PR #3293, opened 2023-06-29
[ceph] Add a new profile for Ceph plugins

jcastill:
This patch helps simplify the capture of data
from ceph plugins. After the original code was
divided in 8 different plugins, it may be a bit
confusing which plugins to use when running
sos and attempting to capture only ceph data.
By selecting the profile, sos will activate only
these ceph_* plugins that apply to a specific
ceph node.

---


pmoravec (approved):
good idea

TurboTurtle (approved):

PR #3296, opened 2023-07-03
[policies] Add Red Flag OS policy.

gogv:
---

### DIFF
--- a/sos/policies/distros/redflag.py
+++ b/sos/policies/distros/redflag.py
@@ -1,0 +1,48 @@
+# Copyright (c) Yang Qing <yangqing@redflag-os.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.policies.distros.redhat import RedHatPolicy, OS_RELEASE
+import os
+
+
+class RedFlagPolicy(RedHatPolicy):
+    distro = "Red Flag Asianux Server Linux"
+    vendor = "Red Flag Software"
+    vendor_urls = [('Distribution Website', 'http://www.redflag-os.com')]
+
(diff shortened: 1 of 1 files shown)


arif-ali (approved):

arif-ali (approved):

gogv:
When will this merge into the main line?

sos/policies/distros/redflag.py
```diff
@@ -0,0 +1,48 @@
...
+# See the LICENSE file in the source distribution for further information.
+
+from sos.policies.distros.redhat import RedHatPolicy, OS_RELEASE
+import os
+
+
+class RedFlagPolicy(RedHatPolicy):
+    distro = "Red Flag Asianux Server Linux"
+    vendor = "Red Flag Software"
+    vendor_urls = [('Distribution Website', 'http://www.redflag-os.com')]
```
pmoravec, line 18: Is there some reliable (https) web page? And ideally with information in English? Relying on translators, it is difficult to browse via the (unsecured, not reliable) web page.

Also, is the OS project alive? Per https://en.wikipedia.org/wiki/Red_Flag_Linux , the company developing the OS closed down their office (but Red Flag Desktop 10 was released 2y ago, per [2] at wikipedia..?)

TurboTurtle:
I'm on the fence on this. There seems to have been a trend developing lately around adding a policy and disappearing.

The idea behind policies is first and foremost to do plugin enablement control, so that we can have multiple forms of plugins where needed. The second is to provide a way to refine user experience; through branded messages and automatic configuration of `--upload` options. It also stands that if a distribution has a policy, there will be at least some level (for some amount of time) of ongoing interaction with the project.

As this is a RHEL clone, there's no difference at all in plugin management - there's no plugins called out that need to behave differently. Further, there's no definition for `--upload`.

I'm personally not a fan of loading up policies for the sole sake of branding when there's no actual change in plugin behavior or management, no integrations for potential consumer support, and no indication on ongoing involvement with the project.

But I'll wait to hear other viewpoints from the maintainer's team.

gogv:
May or may not understand.

PR #3298, opened 2023-07-03
[ssh] Add User .ssh Config File Option

dtzhou2:
Resolves issue SUPDEV-137.
Adds a plugin option for the ssh module, defining whether it will or will not collect .ssh config files per user.
Default for new option is True

Signed-off-by: Daniel Zhou <dzhou@redhat.com>

---

### DIFF
--- a/sos/report/plugins/ssh.py
+++ b/sos/report/plugins/ssh.py
@@ -8,7 +8,7 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.report.plugins import Plugin, IndependentPlugin
+from sos.report.plugins import Plugin, IndependentPlugin, PluginOpt
 
 
 class Ssh(Plugin, IndependentPlugin):
@@ -17,6 +17,12 @@ class Ssh(Plugin, IndependentPlugin):
 
     plugin_name = 'ssh'
     profiles = ('services', 'security', 'system', 'identity')
+
+    option_list = [
+        PluginOpt('userconfs', default=True, val_type=str,
+                  desc=('Changes whether module will '
+                        'collect user .ssh configs'))
+    ]
 
     def setup(self):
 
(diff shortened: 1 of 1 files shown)


arif-ali:
is it worth adding the if statement before the function call for `self.user_ssh_files_permissions()` instead in the function itself?

```python
if self.get_option('userconfs'):
    self.user_ssh_files_permissions()
```

The function is not collecting anything else anyhow, but still doing some work; so I think this should be a better way imho

sos/report/plugins/ssh.py
```diff
@@ -18,6 +18,12 @@ class Ssh(Plugin, IndependentPlugin):
     plugin_name = 'ssh'
     profiles = ('services', 'security', 'system', 'identity')
 
+    option_list = [
+        PluginOpt('userconfs', default='True', val_type=str,
```
TurboTurtle, line 22: `default` should be a boolean here, not a string. We do string conversion to bools from the CLI, but internally we should still be using the correct native type.

dtzhou2: fixed issue!

dtzhou2:
> is it worth adding the if statement before the function call for `self.user_ssh_files_permissions()` instead in the function itself?
> 
> ```python
> if self.get_option('userconfs'):
>     self.user_ssh_files_permissions()
> ```
> 
> The function is not collecting anything else anyhow, but still doing some work; so I think this should be a better way imho

yea I agree with that change

TurboTurtle:
Ack to the change, but it looks like your commit message got a bit mangled in any fixup work. Let's set that back to what you had with the original `[ssh] Add User .ssh Config File Option` content (you can do this with `git commit --amend`)

arif-ali (approved):

TurboTurtle:
Looks like the commit message got murky again in a follow up. To be clear - the commit title should be what is now on the first line (`[ssh] Add User .ssh Config File Option`). I know this is nitpicky, but this conformance really does help when trawling the git history.

dtzhou2:
Apologies for that. Don't know why my git commits keep on getting messed up titles. Just fixed

TurboTurtle (approved):

Issue #3018, opened 2022-08-31
Cleaner can attempt to create too long filenames


pmoravec:
Reproducer (yet artificial):

    mkdir /etc/vdsm/
    date > /etc/vdsm/foo.bar.foo.bar.foo.bar.foo.bar.foo.bar.foo.bar.foo.bar.foo.bar.conf
    sos report -o vdsm --clean --keywords foo,bar --batch --build

It "silently" fails to obfuscate the filename, as the archive contains `etc/vdsm/foo.bar.foo.bar.foo.bar.foo.bar.foo.bar.foo.bar.foo.bar.foo.bar.conf`. (Only) when running cleaner with `-vv`:

    [cleaner] Unable to parse file etc/vdsm/foo.bar.foo.bar.foo.bar.foo.bar.foo.bar.foo.bar.foo.bar.foo.bar.conf: [Errno 36] File name too long: '/var/tmp/sos.grvw1_ev/sosreport-pmoravec-rhel8-2022-08-31-qsxmplu/etc/vdsm/foo.bar.foo.bar.foo.bar.foo.bar.foo.bar.foo.bar.foo.bar.foo.bar.conf' -> '/var/tmp/sos.grvw1_ev/sosreport-pmoravec-rhel8-2022-08-31-qsxmplu/etc/vdsm/obfuscatedword0.obfuscatedword1.obfuscatedword0.obfuscatedword1.obfuscatedword0.obfuscatedword1.obfuscatedword0.obfuscatedword1.obfuscatedword0.obfuscatedword1.obfuscatedword0.obfuscatedword1.obfuscatedword0.obfuscatedword1.obfuscatedword0.obfuscatedword1.conf'

If nothing else, the "Unable to parse file" log should be an `error` and not `debug` (https://github.com/sosreport/sos/blob/main/sos/cleaner/__init__.py#L637) to notify user that some obfuscation failed. Yet better approach is to detect the long filename ahead and shorten it (likewise we do in https://github.com/sosreport/sos/blob/main/sos/report/plugins/__init__.py#L42 for `add_cmd_output` archive filenames) - though this might have some small negative side effects..?

PR #3299, opened 2023-07-03
[cleaner] Raise error for long filenames

dtzhou2:
Partially resolves issue #3018 
This fix just adds a debug error message stating that the filename was too long to obfuscate. 
Does not terminate the cleaner or change the filename, thinking of more complete fix now.

Signed-off-by: Daniel Zhou <dzhou@redhat.com>

---

### DIFF
--- a/sos/cleaner/__init__.py
+++ b/sos/cleaner/__init__.py
@@ -752,6 +752,7 @@ third party.
         if not filename:
             # the requested file doesn't exist in the archive
             return
+        self.length_check(filename)
         subs = 0
         if not short_name:
             short_name = filename.split('/')[-1]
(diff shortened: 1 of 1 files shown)


TurboTurtle (commented):
Logging this is fine and helpful, but there are 2 separate concerns here I have.

1. This doesn't actually log when a failure occurs. It preemptively does so, and still continues on with the failing code.
2. Related to that, this should ideally capture the failing code (e.g. in a try block) to gracefully fail out of the obfuscation attempt and _then_ log the message.

arif-ali:
@dtzhou2 you still working on this PR at all?

Issue #3307, opened 2023-07-18
Error when collecting sosreport from live environment: `Could not enumerate network devices: [Errno 2] No such file or directory: '/mnt/sys/class/net'`


pjmattingly:
Scenario: Writing a KB on a possible scenario where a customer has a system that does not boot normally, but can be booted into a live environment. For troubleshooting it is desirable to collect information from the host system, while in the live environment.

The issue was observed when the live environment was booted, the host root partition was mounted on `/mnt` and sos report was used in following forms:

`sudo sos report -a --all-logs --sysroot=/mnt --chroot=always --estimate-only`
`sudo sos report -a --all-logs --sysroot=/mnt --estimate-only`
`sudo sos report -a --all-logs --sysroot=/mnt --chroot=always`
`sudo sos report -a --all-logs --sysroot=/mnt`

Which resulted in the error:

`Could not enumerate network devices: [Errno 2] No such file or directory: '/mnt/sys/class/net'`

Looking more closely at the `/mnt/sys/class/net` directory, it was found that the `/mnt/sys` directory was empty. This is to be expected when booting from a live environment.

Expected behaviour: I would expect that sosreport might error on finding `/mnt/sys` empty, and exit. Or continue with some other workaround; For example, falling back to using the live-system kernel. But when executing the above commands `sos` hangs and will not exit unless killed.

Additional information:

Live environment:
```
uname -a
Linux ubuntu-server 5.15.0-60-generic #66-Ubuntu SMP Fri Jan 20 14:29:49 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
```
```
df -h
Filesystem      Size  Used Avail Use% Mounted on
tmpfs           393M  1.3M  392M   1% /run
/dev/sr0        1.9G  1.9G     0 100% /cdrom
/cow            2.0G  162M  1.8G   9% /
overlay         321M  321M     0 100% /media/filesystem
tmpfs           2.0G     0  2.0G   0% /dev/shm
tmpfs           5.0M     0  5.0M   0% /run/lock
tmpfs           2.0G  4.7M  2.0G   1% /tmp
/dev/vda2        25G  6.8G   17G  30% /mnt
tmpfs           393M  4.0K  393M   1% /run/user/1001
```

Host environment:
```
uname -a
Linux kbtestmachine 5.15.0-76-generic #83-Ubuntu SMP Thu Jun 15 19:16:32 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
```

[...]

PR #3313, opened 2023-07-19
[report] Don't use sysroot for network device enumeration

TurboTurtle:
If sos is being used in a live environment to diagnose an issue, using sysroot can cause the network device enumeration via /sys/class/net crawling to fail. This will be the case for systems that do not use `nmcli`.

When in a live environment, network devices will not be under `/$sysroot/sys/class/net` but the "regular" path for the booted environment. Similarly, if sos is being run in a container that is properly configured, network devices will appear under `/sys/class/net` and not (necessarily) under the sysroot path that mounts the host's filesystem.

As such, disregard a configured sysroot when enumerating network devices by crawling `/sys/class/net`, and trap any exceptions that may percolate up from this in edge case environments.

Closes: #3307

---

### DIFF
--- a/sos/report/__init__.py
+++ b/sos/report/__init__.py
@@ -612,10 +612,18 @@ class SoSReport(SoSComponent):
         filt_devs = ['bonding_masters']
         _eth_devs = []
         if not namespace:
-            _eth_devs = [
-                dev for dev in listdir('/sys/class/net', self.opts.sysroot)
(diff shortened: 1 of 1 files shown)


pmoravec (approved):

TurboTurtle:
I updated the commit message to be `Related` rather than `Closes`, since that issue has multiple fronts.

PR #3315, opened 2023-07-21
[openshift_ovn] Collect db files for ovn interconnect environment

pperiyasamy:
This updates openshift_ovn plugin to collect ovn db files and logs when it is running with interconnect configuration.

---

### DIFF
--- a/sos/report/plugins/openshift_ovn.py
+++ b/sos/report/plugins/openshift_ovn.py
@@ -27,7 +27,13 @@ class OpenshiftOVN(Plugin, RedHatPlugin):
             "/var/log/openvswitch/libreswan.log",
             "/var/log/openvswitch/ovs-monitor-ipsec.log"
         ])
+        # Collect ovn interconnect specific files if exists.
+        self.add_copy_spec([
+            "/var/lib/ovn-ic/etc/ovnnb_db.db",
+            "/var/lib/ovn-ic/etc/ovnsb_db.db"
+        ])
 
+        # The ovn cluster/status is not valid anymore for interconnect setup.
         self.add_cmd_output([
             'ovn-appctl -t /var/run/ovn/ovnnb_db.ctl ' +
             'cluster/status OVN_Northbound',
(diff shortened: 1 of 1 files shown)


sos/report/plugins/openshift_ovn.py
```diff
@@ -27,6 +27,12 @@ def setup(self):
             "/var/log/openvswitch/libreswan.log",
             "/var/log/openvswitch/ovs-monitor-ipsec.log"
         ])
+        # Collect ovn interconnect specific files if exists.
+        self.add_copy_spec([
+            "/var/lib/ovn-ic/etc/ovnnb_db.db",
+            "/var/lib/ovn-ic/etc/ovnsb_db.db",
+            "/var/log/ovn-ic/acl-audit-log.log"
```
npinaeva, line 34: why did you add this file? looks like we didn't collect it before

sos/report/plugins/openshift_ovn.py
```diff
@@ -27,6 +27,12 @@ def setup(self):
...
+        ])
```
npinaeva, line 35: I think `/var/run/ovn` folder also changed

pperiyasamy: yes, but it is still mounted as /var/run/ovn/ into ovnkube-node container, so i guess we're fine there.

sos/report/plugins/openshift_ovn.py
pperiyasamy: oh! yes, removed it.

arif-ali (approved):

sos/report/plugins/openshift_ovn.py
npinaeva: the problem is sos report runs on the host, not in a container. You can test these commands e.g. if you `chroot /host` from the debug pod

npinaeva: I guess we need to add collection for ovnkube-controller container?

pperiyasamy: Yes @npinaeva , the container commands don't work anymore due to directory path mismatch. now collecting ovs `ct-zone-list` directly on the host which is working fine.

TurboTurtle (approved):

PR #3325, opened 2023-08-04
[coredump] New plugin to capture coredump info

jcastill:
This plugin captures information about core dumps
on these systems where it's active and has replaced abrt.

Closes: #3325

---

### DIFF
--- a/sos/report/plugins/coredump.py
+++ b/sos/report/plugins/coredump.py
@@ -1,0 +1,43 @@
+# Copyright (C) 2023 Red Hat, Inc., Jose Castillo <jcastillo@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, IndependentPlugin, PluginOpt
+
+
+class Coredump(Plugin, IndependentPlugin):
+
+    short_desc = 'Retrieve coredump information'
+
+    plugin_name = "coredump"
+    profiles = ('system', 'debug')
+    packages = ('systemd-udev', 'systemd-coredump')
+
+    option_list = [
+        PluginOpt("detailed", default=False,
+                  desc="collect detailed information for every report")
+    ]
+
+    def setup(self):
+        self.add_copy_spec([
+            "/etc/systemd/coredump.conf",
+            "/etc/systemd/coredump.conf.d/",
+            "/run/systemd/coredump.conf.d/",
+            "/usr/lib/systemd/coredump.conf.d/"
+        ])
+
+        self.add_cmd_output("coredumpctl dump")
+
+        coredump_list = self.collect_cmd_output("coredumpctl list")
+        if self.get_option("detailed") and coredump_list['status'] == 0:
+            for line in coredump_list["output"].splitlines()[1:]:
+                self.add_cmd_output("coredumpctl info "
+                                    f"{line.split()[4]}")
+
+# vim: set et ts=4 sw=4 :


sos/report/plugins/coredump.py
```diff
@@ -0,0 +1,44 @@
...
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, RedHatPlugin, PluginOpt
+
+
+class Coredump(Plugin, RedHatPlugin):
```
nkshirsagar, line 14: Maybe add ubuntu too, since coredumpctl is available in ubuntu from the systemd-coredump package

sos/report/plugins/coredump.py
```diff
@@ -0,0 +1,44 @@
...
+
+    short_desc = 'Retrieve core dumps information'
```
nkshirsagar, line 16: nit: 'Retrieve coredump information'

sos/report/plugins/coredump.py
```diff
@@ -0,0 +1,44 @@
...
+
+    plugin_name = "coredump"
+    profiles = ('system', 'debug')
+    packages = ('systemd-udev',)
```
nkshirsagar, line 20: systemd-coredump package for ubuntu

sos/report/plugins/coredump.py
```diff
@@ -0,0 +1,44 @@
...
+            "/run/systemd/coredump.conf.d/",
+            "/usr/lib/systemd/coredump.conf.d/"
+        ])
+
+        self.add_cmd_output("coredumpctl dump")
+
+        coredump_list = self.collect_cmd_output("coredumpctl list")
+        if self.get_option("detailed") and coredump_list['status'] == 0:
+            for line in coredump_list["output"].splitlines():
+                if not line.startswith("TIME"):
```
nkshirsagar, line 40: is there a meaningful else to this "if"? What if it does start with "TIME" but it might still be useful getting stuff like info on whether the coredump can be provided for more analysis if needed, i.e if its still around?

           COREFILE
               Information whether the coredump was stored, and whether it is still accessible: "none" means the core was not stored, "-" means that it
               was not available (for example because the process was not terminated by a signal), "present" means that the core file is accessible by the
               current user, "journal" means that the core was stored in the "journal", "truncated" is the same as one of the previous two, but the core
               was too large and was not stored in its entirety, "error" means that the core file cannot be accessed, most likely because of insufficient
               permissions, and "missing" means that the core was stored in a file, but this file has since been removed.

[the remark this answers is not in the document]
TurboTurtle: If it's just to skip the first line, then this should be

```python
if self.get_option("detailed") and coredump_list['status'] == 0:
    for line in coredump_list["output"].splitlines()[1:]:
```

sos/report/plugins/coredump.py
jcastill: ack, will add Ubuntu as well

sos/report/plugins/coredump.py
jcastill: ack

sos/report/plugins/coredump.py
jcastill: I should add DebianPlugin as well, I imagine, right?

nkshirsagar: Yes I think that would be a good idea.

pmoravec: Worth to stay with `[|RedHat|Ubuntu|Debian]Plugin`, or directly with `[|Independent]Plugin` (which contains bit more)?

jcastill: I was thinking about using this approach, that we use in other plugins:

`class RedHatCoredump(Coredump, RedHatPlugin):

    packages = ('systemd-udev',)

    def setup(self):
        super(RedHatCoredump, self).setup()


class DebianCoredump(Coredump, DebianPlugin, UbuntuPlugin):

    packages = ('systemd-coredump',)

    def setup(self):
        super(DebianCoredump, self).setup()
`

Because the packages are different. But if 'packages' is 'any' instead of 'all', then I can simplify it with Independent plugin and list all packages in the same line.

TurboTurtle: `packages` is indeed 'any'. `IndependentPlugin` makes sense here.

arif-ali (approved):
I've just tested this on my laptop, and enabled the detailed option and works as expected

pmoravec (approved):

sos/report/plugins/coredump.py
```diff
@@ -0,0 +1,43 @@
...
+            "/usr/lib/systemd/coredump.conf.d/"
+        ])
+
+        self.add_cmd_output("coredumpctl dump")
+
+        coredump_list = self.collect_cmd_output("coredumpctl list")
+        if self.get_option("detailed") and coredump_list['status'] == 0:
+            for line in coredump_list["output"].splitlines()[1:]:
+                self.add_cmd_output("coredumpctl info '%s'"
+                                    % line.split()[4])
```
TurboTurtle, line 41: f-strings, please.

arif-ali (approved):

TurboTurtle (approved):

PR #3329, opened 2023-08-07
[slurm] Add slurm plugin

arif-ali:
Slurm is a workload manager in the HPC space, this is a start on this, and there may be further additions in the future

Resolves: #3329

---

### DIFF
--- a/sos/report/plugins/slurm.py
+++ b/sos/report/plugins/slurm.py
@@ -1,0 +1,128 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, UbuntuPlugin, RedHatPlugin
+from sos.utilities import is_executable
+
+
+class Slurm(Plugin, UbuntuPlugin, RedHatPlugin):
+
+    short_desc = "Slurm Workload Manager"
+
+    plugin_name = 'slurm'
+    profiles = ('hpc',)
+    packages = (
+        # Ubuntu
+        'slurm-wlm',
+        'slurmd',
+        'slurmdbd',
+        'slurmctld',
+        # EL
+        'slurm',
+        'slurm-slurmctld',
+        'slurm-slurmd',
+        'slurm-slurmdbd',
+    )
+    services = (
+        'slurmd',
+        'slurmdbd',
+        'slurmctld',
+    )
+
+    def setup(self):
+        """ Slurm Workload Manager
+        """
+
+        self.add_copy_spec('/etc/slurm/*.conf')
+
+        if is_executable('sinfo'):
+            self.add_cmd_output([
+                'sinfo --all --list-reasons --long',
+                'sinfo --all --long',
+            ])
+
+        if is_executable('squeue'):
+            self.add_cmd_output([
+                'squeue --all --long',
+            ])
+
+        scontrol_cmds = [
+            'aliases',
+            'assoc_mgr',
+            'bbstat',
+            'burstBuffer',
+            'config',
+            'daemons',
+            'dwstat',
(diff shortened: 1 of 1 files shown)


sos/report/plugins/slurm.py
```diff
@@ -0,0 +1,44 @@
...
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, UbuntuPlugin, RedHatPlugin
+
+
+class Slurm(Plugin, UbuntuPlugin, RedHatPlugin):
+
```
pmoravec, line 13: Please add `short_desc` (the reason for tests failing :), in final PR version ).

sos/report/plugins/slurm.py
```diff
@@ -0,0 +1,44 @@
...
+        'slurmd',
+        'slurmdbd',
+        'slurmctld',
+    )
+
+    def setup(self):
+        """ Slurm Workload Manager
+        """
+
+        self.add_copy_spec('/etc/slurm/')
```
pmoravec, line 38: Can't be there some credentials? Reading https://slurm.schedmd.com/slurm.conf.html, `AccountingStoragePass` or `JobCompPass` seems storing passwords..?

TurboTurtle (approved):

sos/report/plugins/slurm.py
```diff
@@ -0,0 +1,128 @@
...
+
+        self.add_copy_spec('/etc/slurm/*.conf')
+
+        if is_executable('sinfo'):
+            self.add_cmd_output([
+                'sinfo --all --list-reasons --long',
+                'sinfo --all --long',
+            ])
+
+        if is_executable('squeue'):
```
pmoravec, line 49: I think this test might not be needed. If a command isn't executable, `sos` won't store output of it (https://github.com/sosreport/sos/blob/main/sos/utilities.py#L261-L270) - but will add a record to `manifest.json`.

But definitely it is fine to use the test here.

sos/report/plugins/slurm.py
```diff
@@ -0,0 +1,128 @@
...
+            'licenses',
+            'node',
+            'partition',
+            'reservation',
+            'slurmd',
+            'step',
+            'topology',
+        ]
+
+        if is_executable('scontrol'):
```
pmoravec, line 74: Yeah, here the test makes a lot of sense :) (test executable once, save many "command not found" deadends ahead). :)

pmoravec (approved):

arif-ali:
I tested it without the test, and the file was created, so hence added the tests.

```
# more sosreport-slurm-comp01-2023-08-08-ddtiwod/sos_commands/slurm/squeue_--all_--long
timeout: failed to run command ‘squeue’: No such file or directory
```

I could easily split the plugin out to multiple plugins, one for `slurmd`, `slurmctld` and `slurmdbd`, but pointless imho. It was easier to add everything in one rather than multiple plugins.

`slurmd` nodes (on their own) will never have the commands, but logs would be there, the commands are only on hosts that the server service is running, `slurm` or EL hosts, and `slurm-client` for Ubuntu.

Alternative method would have been to test for the above 2 packages for the command availability

pmoravec:
> I tested it without the test, and the file was created, so hence added the tests.
> 
> ```
> # more sosreport-slurm-comp01-2023-08-08-ddtiwod/sos_commands/slurm/squeue_--all_--long
> timeout: failed to run command ‘squeue’: No such file or directory
> ```
> 
> I could easily split the plugin out to multiple plugins, one for `slurmd`, `slurmctld` and `slurmdbd`, but pointless imho. It was easier to add everything in one rather than multiple plugins.
> 
> `slurmd` nodes (on their own) will never have the commands, but logs would be there, the commands are only on hosts that the server service is running, `slurm` or EL hosts, and `slurm-client` for Ubuntu.
> 
> Alternative method would have been to test for the above 2 packages for the command availability

Ah I see, then the test makes a lot of sense on either place.

If there is more stuff to collect common than different, I would stay with one plugin.

PR #3335, opened 2023-08-11
[firewall tables] Fix code to capture NAT table

jcastill:
The code used to check the file /proc/net/ip_tables_names and if it didn't exist or was empty, would load
two tables by default - mangle, and filter. The logic was missing the nat table, and so it was not captured in certain scenarios.
The change in this PR simplifies the code by not
checking /proc/net/ip_tables_names anymore, and
looking directly at the output of nf tables list.

Related: RHBZ#2228642

---


pmoravec:
The change will resolvecases when `/proc/net/ip_tables_names` is missing. But can't be there counter-examples where `nft list` is smaller than content of the file? (I really dont know)

I.e. why not 1) try `/proc/net/ip_tables_names`, if it fails then 2) iterate over `nft_list`, 3) if that fails, failover to the default 2 tables..?

jcastill:
That was one of the things I considered originally, the other being adding 'nat' in default_ip_tables and get on with it.
From what I've seen so far, it looks like /proc/net/ip_tables_names is dependent on old iptables, and as we move out of that into nftables, relying on that file may not be the best way. 
But I'm happy to just add the 'nat' table into default_ip_tables if that's the better approach.

TurboTurtle:
> But I'm happy to just add the 'nat' table into default_ip_tables if that's the better approach.

I'm thinking that may be the best solution at the moment.

pmoravec (approved):
ACK. That table sounds too generic to collect it every time. The worst is we will call a command against not existing table, sometimes.

TurboTurtle (approved):
Ack. Not sure why the CI was cancelled, restarting. Will merge before version cut once the CI passes (no reason it won't).

PR #3336, opened 2023-08-11
[openshift_ovn] Collect additional ovnkube node logs

pperiyasamy:
With Interconnect support in latest OVN-Kubernetes, ovnkube-nodes logs grew large. This commit adds the ability to collect those additional logs.

---

### DIFF
--- a/sos/report/plugins/openshift_ovn.py
+++ b/sos/report/plugins/openshift_ovn.py
@@ -20,19 +20,28 @@ class OpenshiftOVN(Plugin, RedHatPlugin):
     profiles = ('openshift',)
 
     def setup(self):
+        all_logs = self.get_option("all_logs")
+
         self.add_copy_spec([
             "/var/lib/ovn/etc/ovnnb_db.db",
             "/var/lib/ovn/etc/ovnsb_db.db",
-            "/var/lib/openvswitch/etc/keys",
-            "/var/log/openvswitch/libreswan.log",
-            "/var/log/openvswitch/ovs-monitor-ipsec.log"
-        ])
-        # Collect ovn interconnect specific files if exists.
+            "/var/lib/openvswitch/etc/keys"
+        ], sizelimit=300)
+
+        # Collect ovn interconnect specific db files if exists.
         self.add_copy_spec([
             "/var/lib/ovn-ic/etc/ovnnb_db.db",
-            "/var/lib/ovn-ic/etc/ovnsb_db.db",
-            "/var/lib/ovn-ic/etc/libovsdb*log*"
-        ])
+            "/var/lib/ovn-ic/etc/ovnsb_db.db"
+        ], sizelimit=300)
+
+        # Collect libovsdb logs in case of ovn interconnect setup.
+        if not all_logs:
+            self.add_copy_spec([
+                "/var/lib/ovn-ic/etc/libovsdb.log",
+                "/var/lib/ovn-ic/etc/libovsdb*log.gz"
+            ], sizelimit=100)
+        else:
+            self.add_copy_spec("/var/lib/ovn-ic/etc/libovsdb*log*")
 
         # The ovn cluster/status is not valid anymore for interconnect setup.
         self.add_cmd_output([
--- a/sos/report/plugins/openshift_ovn.py
+++ b/sos/report/plugins/openshift_ovn.py
@@ -30,7 +30,8 @@ class OpenshiftOVN(Plugin, RedHatPlugin):
         # Collect ovn interconnect specific files if exists.
         self.add_copy_spec([
             "/var/lib/ovn-ic/etc/ovnnb_db.db",
-            "/var/lib/ovn-ic/etc/ovnsb_db.db"
+            "/var/lib/ovn-ic/etc/ovnsb_db.db",
+            "/var/lib/ovn-ic/etc/libovsdb*log*"
         ])
 
         # The ovn cluster/status is not valid anymore for interconnect setup.


sos/report/plugins/openshift_ovn.py
```diff
@@ -32,6 +32,10 @@ def setup(self):
             "/var/lib/ovn-ic/etc/ovnnb_db.db",
             "/var/lib/ovn-ic/etc/ovnsb_db.db"
         ])
+        # Collect additional ovnkube logs for interconnect setup.
+        self.add_copy_spec([
+            "/var/lib/ovn-ic/etc/*log*"
+        ])
```
TurboTurtle, line 38: This can be a one-line addition to the `add_copy_spec()` call above, since we're working in the same directory. Also, a_c_s accepts single strings as well as lists if this were to be kept separate.

sos/report/plugins/openshift_ovn.py
```diff
@@ -32,6 +32,10 @@ def setup(self):
             "/var/lib/ovn-ic/etc/ovnnb_db.db",
             "/var/lib/ovn-ic/etc/ovnsb_db.db"
         ])
+        # Collect additional ovnkube logs for interconnect setup.
+        self.add_copy_spec([
+            "/var/lib/ovn-ic/etc/*log*"
```
TurboTurtle, line 37: I'd also like to know what kind of logs are here, capturing *every* log file by default may not be ideal if there are other components logging here or if there is significant log rotation happening.

flavio-fernandes: hi @TurboTurtle reasonable questions. With the ovn0ic feature, we opted for having the libovsdb logs separate from the regular ovnkube controller. General details on this feature are here: 
More info on the logs are here: https://issues.redhat.com/browse/SDN-3507
CNO PR: https://github.com/openshift/cluster-network-operator/pull/1938
The logs are rotated via lumberjack and have the following values: 100Mb (uncompressed) and 5 rotated and compressed. With that, the worst case scenario will mean that the the SOS will have less than 500Mb.
How about we restrict the log files to use "libovsdb*log*" wildcard?
Here is an example output of what the files look like:
```
❯ POD=$(oc get pod -n openshift-ovn-kubernetes -l app=ovnkube-node -o jsonpath='{.items[0].metadata.name}')
ocn exec -it $POD -c ovnkube-controller -- bash -c "ls -lah /var/log/ovnkube/libovsdb*log*"
-rw-r-----. 1 root root  41K Aug 14 19:45 /var/log/ovnkube/libovsdb-2023-08-14T19-45-34.678.log.gz
-rw-r-----. 1 root root 660K Aug 14 23:38 /var/log/ovnkube/libovsdb.log
```

arif-ali: What about collecting `/var/log/ovnkube/*log*` when `--all-logs` is specified, and `/var/log/ovnkube/*.log` otherwise. This follows the same kind of scenario for many of the plugins. This ensures that the size of sosreport is not as big if `--all-logs` is not specified

npinaeva: I would say we want to have these files collected by default (since communicating the right flags that should be used is not that easy and may result in losing precious data), and we know this plugin is only active for openshift nodes, so this change shouldn't affect the majority of setups, wdyt?

sos/report/plugins/openshift_ovn.py
pperiyasamy: OK @TurboTurtle , moved this log path into existing list.

sos/report/plugins/openshift_ovn.py
```diff
@@ -20,18 +20,30 @@ class OpenshiftOVN(Plugin, RedHatPlugin):
...
-        ])
+        ], sizelimit=300)
+
+        if not all_logs:
+            self.add_copy_spec([
+                "/var/log/openvswitch/libreswan.log",
+                "/var/log/openvswitch/ovs-monitor-ipsec.log"
+            ])
+        else:
+            self.add_copy_spec("/var/log/openvswitch")
```
arif-ali, line 43: This folder is also collected in the `openvswitch` plugin, so you could have a conflict of collections. It may be better to target the logname and asterisk. or better yet, if the `openvswitch` plugin does get activated on an openshift node, that these items could be added to that plugin?

sos/report/plugins/openshift_ovn.py
```diff
@@ -20,18 +20,30 @@ class OpenshiftOVN(Plugin, RedHatPlugin):
...
+
+        # Collect libovsdb logs for ovn interconnect setup.
+        self.add_copy_spec("/var/lib/ovn-ic/etc/libovsdb*log*", sizelimit=600)
```
arif-ali, line 46: imo, I'm still keen to use all_log for all collections and `.log` if that has not been asked for. But, happy for other people to chime in here

sos/report/plugins/openshift_ovn.py
TurboTurtle: The prevailing approach in sos plugins is to capture `*.log` normally, and get `*log*` when either `--all-logs` or some plugin-specific option is used.

The `sizelimit` parameter here applies per item (in a list) handed to `add_copy_spec`, not the entire call or the individual matches for a regex (meaning a `sizelimit` setting of 600 is a bit heavy handed). So, another option would be having one entry for `*.log` and another for `*.log.gz` and setting sizelimit to 100 meaning by default we'd capture the active log and up to 100 MB of rotated logs. If something truly needs _all_ the rotated logs, then that's where `--all-logs` would come into play.

Would that be a reasonable middle ground?

pperiyasamy: @npinaeva @TurboTurtle @arif-ali  , Thanks! Based on your concerns and suggestions, updated the log collection with commit b098433, Is that ok now ?

sos/report/plugins/openshift_ovn.py
pperiyasamy: yes, this is mainly to collect ipsec related logs on `ovn`/`openvswitch` setup. 
now collecting only ipsec logs from ovn plugin when --all-logs option is not set.

sos/report/plugins/openshift_ovn.py
```diff
@@ -20,18 +20,32 @@ class OpenshiftOVN(Plugin, RedHatPlugin):
...
+        # Collect ipsec logs for openvswitch and additionally libovsdb
+        # logs in case of ovn interconnect setup.
+        if not all_logs:
+            self.add_copy_spec([
+                "/var/log/openvswitch/libreswan.log",
+                "/var/log/openvswitch/ovs-monitor-ipsec.log"
+            ])
+            self.add_copy_spec(["/var/lib/ovn-ic/etc/libovsdb.log",
+                                "/var/lib/ovn-ic/etc/libovsdb*log.gz"],
+                               sizelimit=100)
```
TurboTurtle, line 46: Nitpick on style guidelines, this should be
```python
self.add_copy_spec([
    "/var/lib/ovn-ic/etc/libovsdb.log",
    "/var/lib/ovn-ic/etc/libovsdb*log.gz"
], sizelimit=100)
```

pperiyasamy: done.

arif-ali (approved):

sos/report/plugins/openshift_ovn.py
```diff
@@ -20,18 +20,33 @@ class OpenshiftOVN(Plugin, RedHatPlugin):
...
             "/var/lib/ovn-ic/etc/ovnsb_db.db"
-        ])
+        ], sizelimit=300)
+
+        # Collect ipsec logs for openvswitch and additionally libovsdb
+        # logs in case of ovn interconnect setup.
+        if not all_logs:
+            self.add_copy_spec([
+                "/var/log/openvswitch/libreswan.log",
+                "/var/log/openvswitch/ovs-monitor-ipsec.log"
```
npinaeva, line 42: why was it moved here? I would say we need these logs even when all_logs is set?

arif-ali: `/var/log/openvswitch` is already being captured by the `openvswitch` plugin. Would that not be activated?

pmoravec: I have seen several such feedbacks that a PR wants to collect a copypath in *another* plugin; maybe does it make sense to have a CI test for it? Though it can be tricky to have it working properly (e.g. due to multiline a_s_c, asterisks in paths, paths with variables etc)?

npinaeva: I think it should be collected for openvswitch plugin, but I don't think that happens for some reason, may be useful to check/debug that

pperiyasamy: ok @npinaeva , I will check.

pperiyasamy: @npinaeva This works as expected as per openvswitch plugin logic [here](https://github.com/sosreport/sos/blob/main/sos/report/plugins/openvswitch.py#L66-L71) on an ovn cluster, in fact we can remove collecting ipsec logs here as openvswitch plugin already taking care of collecting `*.log` when` --all-logs` not specified.

npinaeva: nice, thanks for checking! then we can just remove these files

pperiyasamy: removed collecting ipsec logs from `/var/log/openvswitch` directory.

npinaeva:
/lgtm
thanks all for the nice discussion!

PR #3340, opened 2023-08-23
[maas] Update MAAS plugin to collect logs for version 3.5+

brigaccess:
In MAAS 3.5, multiple services are introduced, and supervisor is replaced by Pebble in snap. This PR updates the `maas` plugin to collect the logs for new services.

---

### DIFF
--- a/sos/report/plugins/maas.py
+++ b/sos/report/plugins/maas.py
@@ -28,8 +28,14 @@ class Maas(Plugin, UbuntuPlugin):
         'maas-rackd',
         'maas-regiond',
         'maas-syslog',
-        # For the snap:
+        # MAAS 3.5 deb:
+        'maas-temporal',
+        'maas-apiserver',
+        'maas-agent',
+        # For the pre-3.5 snap:
         'snap.maas.supervisor',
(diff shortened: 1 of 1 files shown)


sos/report/plugins/maas.py
```diff
@@ -62,13 +68,25 @@ def _is_snap_installed(self):
             return maas_pkg['pkg_manager'] == 'snap'
         return False
 
+    def check_enabled(self):
+        return super().check_enabled() or self.is_service("snap.maas.pebble")
```
arif-ali, line 72: if the `maas` snap is there, then the `packages` tuple will cover this already, so we shouldn't need the extra check here. The `packages` tuple checks or both debs and snaps for Ubuntu platforms

arif-ali (changes requested):

sos/report/plugins/maas.py
brigaccess: Awesome, thanks! I added a commit with a fix, should I squash it?

arif-ali: yes please, squash the commits

TurboTurtle (approved):

arif-ali (approved):

PR #3352, opened 2023-09-07
[greenboot] seperate  logs to a standalone plugin.

eslutsky:
---

### DIFF
--- a/sos/report/plugins/greenboot.py
+++ b/sos/report/plugins/greenboot.py
@@ -1,0 +1,26 @@
+# Copyright 2023 Red Hat, Inc. Evgeny Slutsky <eslutsky@redhat.com>
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, RedHatPlugin
+
+
+class Greenboot(Plugin, RedHatPlugin):
+    """The greenboot plugin collects systemd service logs and configuration.
+    """
+
+    short_desc = 'Greenboot'
+    plugin_name = 'greenboot'
+    services = (plugin_name, 'greenboot-healthcheck',
+                'greenboot-task-runner', 'redboot-task-runner',)
+    profiles = ('system',)
+
+    def setup(self):
+        self.add_copy_spec([
+            "/etc/greenboot/greenboot.conf",
+        ])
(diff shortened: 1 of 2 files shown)


sos/report/plugins/greenboot.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class Greenboot(Plugin, RedHatPlugin):
+    """The greenboot plugin collects systemd service logs and configuration.
+    """
+
+    short_desc = 'Greenboot'
+    plugin_name = 'greenboot'
+    plugin_timeout = 300
+    packages = ('microshift-greenboot')
```
arif-ali, line 20: you need to add a comma after `'microshift-greenboot'`, as this needs to be a tuple

arif-ali (changes requested):

arif-ali (approved):

sos/report/plugins/greenboot.py
```diff
@@ -0,0 +1,28 @@
...
+from sos.report.plugins import Plugin, RedHatPlugin
+
+
+class Greenboot(Plugin, RedHatPlugin):
+    """The greenboot plugin collects systemd service logs and configuration.
+    """
+
+    short_desc = 'Greenboot'
+    plugin_name = 'greenboot'
+    plugin_timeout = 300
```
pmoravec, line 19: This is the default value (cf. https://github.com/sosreport/sos/blob/main/sos/report/plugins/__init__.py#L534 and https://github.com/sosreport/sos/blob/main/sos/utilities.py#L47), so a redundant setting..?

pmoravec:
Commit message speaks about journal logs, but the change would collect just a config file - do I miss something..?

bmr-cymru:
Commit message for a plugin wants to be something like `[greenboot] separate logs to a standalone plugin`.

sos/report/plugins/greenboot.py
```diff
@@ -0,0 +1,28 @@
...
+    packages = ('microshift-greenboot',)
```
dhellmann, line 20: We probably don't want to tie this to MicroShift.

eslutsky: @dhellmann , greenboot is distributed as `microshift-greenboot` rpm, are we going to change that also?

TurboTurtle (commented):
LGTM, however this needs to be rebased on current `main` as there is now a merge conflict based upon another change to the `microshift` plugin (minor change to the services tuple).

Once that's done, this should be good. Note 4.6.1 is scheduled to close tomorrow.

eslutsky:
> LGTM, however this needs to be rebased on current `main` as there is now a merge conflict based upon another change to the `microshift` plugin (minor change to the services tuple).
> 
> Once that's done, this should be good. Note 4.6.1 is scheduled to close tomorrow.

ACK and done .

sos/report/plugins/greenboot.py
dhellmann: The `microshift-greenboot` RPM contains the scripts for MicroShift to integrate with greenboot, but the actual system software is in the `greenboot` RPM.

arif-ali (approved):

TurboTurtle (approved):

PR #3353, opened 2023-09-08
[foreman] Call gem list outside scl

pmoravec:
Since scl is RHEL7-only related and foreman on RHEL7 is not further supported, let fully remove from scl wrapper.

Resolves: #3353

---


TurboTurtle (approved):
Ack. At this point we should wholly remove `add_cmd_output_scl()` as well, right?

arif-ali (approved):

PR #3357, opened 2023-09-15
[microshift] add kube-apiserver log paths to microshift plugin

copejon:
---

Closes [ushift-1631](https://issues.redhat.com/browse/USHIFT-1631)

### DIFF
--- a/sos/report/plugins/microshift.py
+++ b/sos/report/plugins/microshift.py
@@ -151,6 +151,7 @@ class Microshift(Plugin, RedHatPlugin):
         if self.path_exists('/var/lib/microshift-backups'):
             self.add_copy_spec(['/var/lib/microshift-backups/*/version',
                                 '/var/lib/microshift-backups/*.json'])
+        self.add_copy_spec(['/var/log/kube-apiserver/*.log'])
 
         self.add_cmd_output([
             'microshift version',


sos/report/plugins/microshift.py
```diff
@@ -151,6 +151,8 @@ def setup(self):
         if self.path_exists('/var/lib/microshift-backups'):
             self.add_copy_spec(['/var/lib/microshift-backups/*/version',
                                 '/var/lib/microshift-backups/*.json'])
+        if self.path_exists('/var/log/kube-apiserver'):
```
pmoravec, line 154: I do feel this check as a bit redundant; when this directory is not present, `add_copy_spec` will detect no file to collect either way - just bit slower than via this explicit check (but if the dir exists, we do an extra check).

Usually, we dont have these checks, while we allowed it in this plugin without an objection, as it seems.

ACK for the current version as well as for plain unconditional `add_copy_spec` one.

copejon: Thanks for the feedback! I've removed the redundant check.

pmoravec (approved):

TurboTurtle (commented):
The addition is fine, but we don't allow fixup commits for merges. Please squash this into a single commit (you'll need to repush the branch with --force), and then we'll be good to merge this.

pmoravec (approved):

TurboTurtle (approved):

PR #3364, opened 2023-09-27
[filesys] skip filesystems picked up by other plugins

utopiabound:
Fix forbidding panfs - add_forbidden_path() needs to be called prior to add_copy_spec() for it to be useful.
Skip /proc/fs/FS for FS picked up by other plugins

---

### DIFF
--- a/sos/report/plugins/filesys.py
+++ b/sos/report/plugins/filesys.py
@@ -56,7 +56,21 @@ class Filesys(Plugin, DebianPlugin, UbuntuPlugin, CosPlugin):
             "lslocks"
         ])
 
-        self.add_forbidden_path('/proc/fs/panfs')
+        self.add_forbidden_path([
+            # cifs plugin
+            '/proc/fs/cifs',
+            # lustre plugin
+            '/proc/fs/ldiskfs',
+            '/proc/fs/lustre',
+            # nfs plugin
+            '/proc/fs/nfsd',
+            '/proc/fs/nfsfs',
+            # panfs (from Panasas company) provides statistics which can be
+            # very large (100s of GB)
+            '/proc/fs/panfs',
+            # xfs plugin
+            '/proc/fs/xfs'
+        ])
 
         if self.get_option('lsof'):
             self.add_cmd_output("lsof -b +M -n -l -P", root_symlink="lsof",


sos/report/plugins/filesys.py
```diff
@@ -35,6 +35,21 @@ class Filesys(Plugin, DebianPlugin, UbuntuPlugin, CosPlugin):
     ]
 
     def setup(self):
+        self.add_forbidden_path([
+            # cifs plugin
+            '/proc/fs/cifs',
```
pmoravec, line 40: We collect just some stuff from that dir, but I guess other stuff is not important (even for cifs plugin itself, so why important for filesys plugin?)

sos/report/plugins/filesys.py
```diff
@@ -35,6 +35,21 @@ class Filesys(Plugin, DebianPlugin, UbuntuPlugin, CosPlugin):
...
+            # lustre plugin
+            '/proc/fs/ldiskfs',
+            # nfs plugin
+            '/proc/fs/nfsd',
+            '/proc/fs/nfsfs',
```
pmoravec, line 45: Ditto, but again, since nfs doesnt care about other files than `/proc/fs/nfsfs/servers` and `/proc/fs/nfsfs/volumes`, no need to collect them here either.

TurboTurtle (changes requested):
Ack to the code change - we'll just want to make sure that consumers know that `-o filesys` may no longer capture certain /proc paths they might expect (but as @pmoravec said, this is more of a current bug than a real regression).

That said, please squash your fixup commit into the base commit and then re-push (you'll need to use --force).

utopiabound:
> Ack to the code change - we'll just want to make sure that consumers know that `-o filesys` may no longer capture certain /proc paths they might expect (but as @pmoravec said, this is more of a current bug than a real regression).
> 
> That said, please squash your fixup commit into the base commit and then re-push (you'll need to use --force).

Done.  I've moved the `add_forbidden_path` back to where it was.  The comment on the plugin already says that it doesn't collect filesystem information for filesystems with their own plugin.

pmoravec (approved):

TurboTurtle (approved):

PR #3373, opened 2023-10-06
[lxd][runtime] Add LXD Runtime

arif-ali:
---

### DIFF
--- a/sos/policies/runtimes/lxd.py
+++ b/sos/policies/runtimes/lxd.py
@@ -1,0 +1,149 @@
+# Copyright (C) 2023 Canonical Ltd., Arif Ali <arif.ali@canonical.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+import json
+
+from sos.policies.runtimes import ContainerRuntime
+from sos.utilities import sos_get_command_output
+from sos.utilities import is_executable
+
+
+class LxdContainerRuntime(ContainerRuntime):
+    """Runtime class to use for systems running LXD"""
+
+    name = 'lxd'
+    binary = 'lxc'
+
+    def check_is_active(self):
+        # the daemon must be running
+        if (is_executable('lxc', self.policy.sysroot) and
+                (self.policy.init_system.is_running('lxd') or
+                 self.policy.init_system.is_running('snap.lxd.daemon'))):
+            self.active = True
+            return True
+        return False
+
+    def get_containers(self, get_all=False):
+        """Get a list of containers present on the system.
+
+        :param get_all: If set, include stopped containers as well
+        :type get_all: ``bool``
+        """
+        containers = []
+
+        _cmd = f"{self.binary} list --format json"
+        if self.active:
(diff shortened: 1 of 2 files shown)


sos/policies/runtimes/lxd.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def get_containers(self, get_all=False):
+        """Get a list of containers present on the system.
+
+        :param get_all: If set, include stopped containers as well
+        :type get_all: ``bool``
+        """
+        containers = []
+
+        _cmd = "%s list --format json" % (self.binary)
```
TurboTurtle, line 41: Nitpick - the rest of the runtime uses f-strings (yay!), so let's use one here to for consistency, and for getting away from string substitution.

sos/policies/runtimes/lxd.py
```diff
@@ -0,0 +1,148 @@
...
+    binary = 'lxc'
+
+    def check_is_active(self, sysroot=None):
+        # the daemon must be running
+        if (is_executable('lxc', sysroot) and
+                (self.policy.init_system.is_running('lxd') or
+                 self.policy.init_system.is_running('snap.lxd.daemon'))):
+            self.active = True
+            return True
+        return False
```
TurboTurtle, line 31: Use `self.policy.sysroot` here, as the only place this is called from Policy doesn't pass a sysroot. On reviewing this, we should make a similar change to the `docker` runtime (in a different PR, no need to add that here).

sos/policies/runtimes/lxd.py
```diff
@@ -0,0 +1,148 @@
...
+            )
+            if out['status'] == 0:
+                out_json = json.loads(out['output'])
+                for profile in out_json:
+                    if profile['name'] == 'default':
+                        stg_pool = profile['devices']['root']['pool']
+                        break
+
+            out = sos_get_command_output(
+                f"{self.binary} storage volume list {stg_pool} --format json",
```
TurboTurtle, line 100: We should have a default value for `stg_pool`, or raise an exception after the check if we don't find one. Like CodeQL mentions, there's a possibility here for `stg_pool` to be undefined when we make this call to `sos_get_command_output()`

TurboTurtle (commented):
Looks mostly solid so far, just a couple minor comments. I'll get a PR up for the `docker` change shortly to cover that side of things.

sos/policies/runtimes/lxd.py
pmoravec: Indeed, our CI tests even detected it :) https://cirrus-ci.com/task/5971844470145024

sos/policies/runtimes/lxd.py
arif-ali: nice catch, I originally wrote this 5 months ago, so converted all to f-string, missed this one ;)

arif-ali:
I'm not really happy with the way `get_volumes()` function is expecting the return, as `lxc` doesn't necessarily give ids, but names of the volumes. So, at the moment this is giving a list of names of volumes in the `default` storage pool or the pool from the `default` profile. This is going to be as good as it goes for the time being.

If this would have a refactor later, we could do with pool + name/id tuple, then that would work better for `lxd`, but that would mean that `docker`/`podman` ones would need to change too. Different concepts of volumes/storage, need to work on understanding how this is used in the wider `sos` aspect

TurboTurtle (approved):

pmoravec (approved):

PR #3375, opened 2023-10-06
[docker] Fix runtime check_is_active sysroot consideration

TurboTurtle:
A runtime's `check_is_active()` does not get passed a sysroot at any point, so the docker runtime abstraction should just use the policy sysroot directly.

Additionally, `check_can_copy()` should be simplified to simply return the already-determined active state of the runtime.

---

### DIFF
--- a/sos/policies/runtimes/docker.py
+++ b/sos/policies/runtimes/docker.py
@@ -18,9 +18,9 @@ class DockerContainerRuntime(ContainerRuntime):
     name = 'docker'
     binary = 'docker'
 
-    def check_is_active(self, sysroot=None):
+    def check_is_active(self):
         # the daemon must be running
-        if (is_executable('docker', sysroot) and
+        if (is_executable('docker', self.policy.sysroot) and
                 (self.policy.init_system.is_running('docker') or
                  self.policy.init_system.is_running('snap.docker.dockerd'))):
             self.active = True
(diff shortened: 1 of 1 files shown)


arif-ali (approved):

sos/policies/runtimes/docker.py
```diff
@@ -18,16 +18,16 @@ class DockerContainerRuntime(ContainerRuntime):
...
+        if (is_executable('docker', self.policy.sysroot) and
                 (self.policy.init_system.is_running('docker') or
                  self.policy.init_system.is_running('snap.docker.dockerd'))):
             self.active = True
             return True
         return False
 
     def check_can_copy(self):
-        return self.check_is_active(sysroot=self.policy.sysroot)
+        return self.active
```
pmoravec, line 31: `self.active = False` until `check_is_active` is called (applies both for `ContainerRuntime` and `DockerContainerRuntime`). Can't we get wrong `False` return here just because we haven't called `check_is_active`?

TurboTurtle: In the current flow, no, because a runtime is only loaded if `check_is_active()` returns `True` when the policy is loaded. If this call returns False during policy instantiation, we drop the runtime entirely.

If someone were to separately try to leverage a runtime here, theoretically yes they could end up calling this before `check_is_acitve()`, but that's not something we support.

pmoravec (approved):

PR #3382, opened 2023-10-12
[snap] Add more items to collect

arif-ali:
Resolves: SET-313

---

### DIFF
--- a/sos/report/plugins/snap.py
+++ b/sos/report/plugins/snap.py
@@ -9,6 +9,8 @@
 
 from sos.report.plugins import Plugin, IndependentPlugin
 
+import re
+
 
 class Snap(Plugin, IndependentPlugin):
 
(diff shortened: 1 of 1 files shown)


sos/report/plugins/snap.py
```diff
@@ -23,8 +24,42 @@ def setup(self):
...
+        all_pkgs = self.policy.package_manager.packages
+
+        for pkg_name in all_pkgs:
+            pkg = self.policy.package_manager.pkg_by_name(pkg_name)
+            if pkg['pkg_manager'] == 'snap':
+                self.add_cmd_output(f"snap connections {pkg['name']}")
+
         self.add_cmd_output("snap debug connectivity", timeout=10)
 
+        snap_changes = sos_get_command_output('snap changes')
```
TurboTurtle, line 50: Use `Plugin.exec_cmd()` here, rather than directly calling the utilities method.

sos/report/plugins/snap.py
```diff
@@ -23,8 +24,42 @@ def setup(self):
...
+
+        if snap_changes['status'] == 0:
+            output = snap_changes['output']
+
+            for line in output.splitlines():
+                if line == "":
+                    continue
+                change = line.split()
+                if change[0] == "ID":
+                    continue
```
TurboTurtle, line 60: Is this a header field, or something that can appear anywhere? If the former, let's just use `output.splitlines()[1:]` (or whatever number of header lines we need to skip).

TurboTurtle (approved):
LGTM

arif-ali:
just added another section, thought we would need a plugin for each of the `gadget` snaps, but that is only really defined in the `Notes` field, and no easy trigger. So added to be consistent with our teams debug collection script

sos/report/plugins/snap.py
```diff
@@ -23,8 +25,56 @@ def setup(self):
         self.add_cmd_output("snap list --all", root_symlink="installed-snaps")
         self.add_cmd_output([
             "snap --version",
-            "snap changes"
```
TurboTurtle, line 26: Just noticed this was being removed and re-referenced later. Is the output directly from `snap changes` still desired? If so, change the `exec_cmd()` below for it to `collect_cmd()`, as the latter will both save the output to the archive (like `add_cmd_output()`) _and_ return it (like `exec_cmd()`).

TurboTurtle (approved):
New section looks good to me, just a minor note that I overlooked earlier. If `snap changes` is indeed no longer wanted as saved output then this all looks fine to me at this point.

sos/report/plugins/snap.py
arif-ali: ah, interesting, good to know :)

I'll get that modified

pmoravec (approved):

Issue #3126, opened 2023-02-02
[testing][snap] improve testing for snaps


arif-ali:
As we were prepping for testing of the 4.5.0 snap, found an issue where `sos` doesn't work on ubuntu 18.04. This therefore means the cirrus CI test for 18.04 hasn't actually worked, and has given a false positive.

We should have seen the following error in the testing
```
Traceback (most recent call last):
  File "/snap/sosreport/70/bin/sos", line 21, in <module>
    sos = SoS(sys.argv[1:])
  File "/snap/sosreport/70/lib/python3.10/site-packages/sos/__init__.py", line 72, in __init__
    import sos.collector
  File "/snap/sosreport/70/lib/python3.10/site-packages/sos/collector/__init__.py", line 29, in <module>
    from sos.collector.sosnode import SosNode
  File "/snap/sosreport/70/lib/python3.10/site-packages/sos/collector/sosnode.py", line 17, in <module>
    from pkg_resources import parse_version
  File "/snap/sosreport/70/lib/python3.10/site-packages/pkg_resources/__init__.py", line 82, in <module>
    from pkg_resources.extern import platformdirs
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 658, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 571, in module_from_spec
  File "/snap/sosreport/70/lib/python3.10/site-packages/pkg_resources/extern/__init__.py", line 52, in create_module
    return self.load_module(spec.name)
  File "/snap/sosreport/70/lib/python3.10/site-packages/pkg_resources/extern/__init__.py", line 37, in load_module
    __import__(extant)
  File "/snap/sosreport/70/lib/python3.10/site-packages/pkg_resources/_vendor/platformdirs/__init__.py", line 5
    from __future__ import annotations
    ^
SyntaxError: future feature annotations is not defined
```

Internal testing now, and adding the below diff has resolved the above error.

```diff
diff --git a/snap/snapcraft.yaml b/snap/snapcraft.yaml
index 2fe65699..fe1b860c 100644
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -9,7 +9,7 @@ base: core22
 confinement: classic
 adopt-info: sos
 environment:
[...]

PR #3383, opened 2023-10-13
packaging.version related changes: chatch the proper exception and switch the priority

hroncok:
I am sending this as a draft as I won't be able to finish the checklist right away. Please do let me know if this seems OK to you and if so, I can amend the magical phrase to commit messages and rewrap them at 72 characters.

---

### DIFF
--- a/sos/utilities.py
+++ b/sos/utilities.py
@@ -23,9 +23,9 @@ from contextlib import closing
 from collections import deque
 
 try:
-    from pkg_resources import parse_version
-except ImportError:
     from packaging.version import parse as parse_version
+except ImportError:
+    from pkg_resources import parse_version
 
 # try loading magic>=0.4.20 which implements detect_from_filename method
 magic_mod = False
--- a/sos/utilities.py
+++ b/sos/utilities.py
@@ -24,7 +24,7 @@ from collections import deque
 
 try:
     from pkg_resources import parse_version as version_parse
-except SyntaxError:
+except ImportError:
     from packaging.version import parse as version_parse
 
 # try loading magic>=0.4.20 which implements detect_from_filename method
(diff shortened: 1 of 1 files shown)


arif-ali:
This was discussed in #3126, and the preference of using `pkg_resources`, as it causes cross repo dependency in RHEL

hroncok:
I actually want to get rid of setuptools from BaseOS in RHEL 10+.

arif-ali:
> I actually want to get rid of setuptools from BaseOS in RHEL 10+.

OK, we at @canonical will need to be aware of this change, so that the relevant changes can be made too for Ubuntu OSs too. We have `pythnon3-packaging` already available in `main` repo, which is typically the requirement for dependencies, going back to 20.04. So, from our perspective, I don't think it will cause an issue

pmoravec:
Generally ACK, just unsure about the flake failure (https://github.com/sosreport/sos/pull/3383/checks?check_run_id=17680606514):

    sos/utilities.py:28:5: F401 'pkg_resources.parse_version' imported but unused

Technically the error is right, practically `utilities` imports a Python class in some way, and that is used later at other places (https://github.com/search?q=repo%3Asosreport%2Fsos%20parse_version&type=code).

`Utilities` should be loaded prior any call of `parse_version` either way, but.. still I think it would be safer to keep the original `utilities`' method `parse_version` (calling *some* imported method of same name)..?

hroncok:
I've added a commit to please the linter int he way I consider "proper".  If you do not like that, I can remove the two commits an leave the function definition as it was, because that's now what I came here to do anyway.

hroncok:
Rebased, hopefully, the commit snow adhere to the rules.

pmoravec (approved):

TurboTurtle (approved):
Ack, though I should note that we were indeed seeing `SyntaxError` previously.

For future contributions, please note that we prefer `[tags]` in the commit title, as it makes groking through the git history much easier, but this change I'm not going to be a stickler about it.

PR #3389, opened 2023-10-18
[pulpcore] Scrub AUTH_LDAP_BIND_PASSWORD value

pmoravec:
Likewise in #3379, scrub the password also in pulpcore plugin.

Resolves: #3389

---

### DIFF
--- a/sos/report/plugins/pulpcore.py
+++ b/sos/report/plugins/pulpcore.py
@@ -144,29 +144,18 @@ class PulpCore(Plugin, IndependentPlugin):
         return _dbcmd % (self.dbhost, self.dbport, self.dbname, quote(query))
 
     def postproc(self):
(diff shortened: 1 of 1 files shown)


jcastill (approved):

pmoravec:
Both `pulp` and `pulpcore` plugins collect the same file and command output which is not ideal. Gladly both plugins can be enabled by default only when having both `pulp-2` and `pulp-3` installed (quite rare situations) and even in that case, since the file should fit in any logsize limit, collecting it by two plugins should not harm.

The ideal solution will probably be merging plugins together - but that has some cons as well we can discuss later.

pmoravec:
Hi @TurboTurtle , could you please also review and ideally merge today?

TurboTurtle (approved):

Issue #3390, opened 2023-10-18
Required support for data collections within the sosreport


mohitkumarrh:
We need the data from sosreport which is required to build Tier-3 rules in RedHat Insights. There are few LS command data we which is currently not being collected in upstream SOS project. I request you to add the data for the below mentioned command.

```
1. /bin/ls -lanR /etc/rhsm
2. /bin/ls -lanR /sys/kernel
3. /bin/ls -lanR /var/lib/sss
``` 

Reference blocker MR for GSS Tier-3 repo: https://gitlab.cee.redhat.com/support-insights/gss-rules/-/merge_requests/4086

pmoravec:
I guess you want to collect that info just from systems having `redhat-access-insights` or `insights-client` package installed (i.e. within `insights` plugin, in sos language)?

SELinux context is not required (`ls -lanRZ`)?

`/bin/ls -lanR /sys/kernel` can be bigger (usually in MBs) - is there some upper limit (default is 25MB) or do you need whole output with no limit all the time?


(I removed the internal RH URL from issue description)

TurboTurtle:
No problem to add the collections, my first impulse is to add them to their respective plugins, but Pavel makes a good point that putting them in the `insights` plugin might be more appropriate if the only use case is automation.

mohitkumarrh:
@pmoravec - Yes what we want is, if the system is having `redhat-access-insights` or `insights-client` package installed the required data get collected.

SELinux context is not required

`/bin/ls -lanR /sys/kernel` for this we can use the default(25MB) upper limit. we don't need the whole output with no limit all the time.

Thanks for removing the internal RH URL from issue description

@TurboTurtle - Thanks for the insights. yes I agree with you and Pavel for putting them in `insights` plugin.

I'm going to use these data to create a common parser for ls, which will be used in Tier-3 rule for XE Insights and will be used by Red Hat engineers internally.

pmoravec:
I raised https://github.com/sosreport/sos/pull/3391 for this. Let me know if I should add tags so you can identify location of the files by them (see e.g. https://github.com/sosreport/sos/blob/main/sos/report/plugins/filesys.py#L49-L52 and check in generated `sos_reports/manifest.json` for : 

            "tag_summary": {
                "df__al": [
                    "sos_commands/filesys/df_-al_-x_autofs"
                ],
                "mount": [
                    "sos_commands/filesys/mount_-l"
                ]
            }

mohitkumarrh:
[...]

PR #3391, opened 2023-10-20
[insights] collect recursive listing of three dirs

pmoravec:
Resolves: #3391
Closes: #3390

---

### DIFF
--- a/sos/report/plugins/insights.py
+++ b/sos/report/plugins/insights.py
@@ -50,6 +50,9 @@ class RedHatInsights(Plugin, RedHatPlugin):
             timeout=30
         )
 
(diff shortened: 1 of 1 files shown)


jcastill (approved):

TurboTurtle (approved):
This looks fine, two thoughts though.

1. `cmd_as_tag` generates something like `_bin_ls_[...]` here, which is a bit odd to me, but perhaps it is fine since most tag consumers would be automation.
2. We've seen a handful of these types of collection requests recently, would it be desirable to add a wrapper to standardize these? E.G. `self.add_dir_listing(['dir1', 'dir2'])`, which collects `ls` with a standard set of options and a standard tag syntax?

Neither of these are blockers, point 2 is just food for thought at the moment, and if Insights is fine with the `cmd_as_tag` style, then all good.

PR #3392, opened 2023-10-24
[kernelrt] Update options format for tuna

jcastill:
In newer versions of tuna, the option -P has been
substituted for 'show_threads'.
Test if the new option is present in 'tuna --help' and if so, use it.

Closes: RH SUPDEV-150

---

### DIFF
--- a/sos/report/plugins/kernelrt.py
+++ b/sos/report/plugins/kernelrt.py
@@ -9,7 +9,7 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.report.plugins import Plugin, RedHatPlugin
+from sos.report.plugins import Plugin, RedHatPlugin, SoSPredicate
 
 
 class KernelRT(Plugin, RedHatPlugin):
@@ -36,6 +36,12 @@ class KernelRT(Plugin, RedHatPlugin):
         # note: rhbz#1059685 'tuna - NameError: global name 'cgroups' is not
         # defined this command throws an exception on versions prior to
         # 0.10.4-5.
-        self.add_cmd_output('tuna -CP')
+        co = {'cmd': 'tuna --help', 'output': '-P'}
+        option_present = self.test_predicate(
+            self, pred=SoSPredicate(self, cmd_outputs=co)
+        )
+        self.add_cmd_output(
+            f"tuna {'-CP' if option_present else 'show_threads -C'}"
+        )
 
 # vim: set et ts=4 sw=4 :


jcastill:
The change in the option names and format happened in RHEL 9, in 0.18-7. Unfortunately using --version only shows 0.18, so the only way to make sure the option is there is to mimic what we did in the grub2 plugin, i.e. get the output of --help and see if the new option name is listed.
This should not affect Ubuntu since there we have up to 0.15-1, if I'm not mistaken.

pmoravec (approved):

sos/report/plugins/kernelrt.py
```diff
@@ -36,6 +36,12 @@ def setup(self):
         # note: rhbz#1059685 'tuna - NameError: global name 'cgroups' is not
         # defined this command throws an exception on versions prior to
         # 0.10.4-5.
-        self.add_cmd_output('tuna -CP')
+        tuna_cmd = 'tuna'
+        co = {'cmd': '%s --help' % tuna_cmd, 'output': 'show_threads'}
+        if self.test_predicate(self, pred=SoSPredicate(self, cmd_outputs=co)):
+            tuna_cmd += ' show_threads -C'
+        else:
+            tuna_cmd += ' -CP'
```
TurboTurtle, line 44: Ack to the intent, but we're trying to move off of legacy string interpolation methods and to f-strings globally. We could do this and button it up a bit with something like:

```python
co = {'cmd': 'tuna --help', 'output': 'show_threads'}
self.add_cmd_output(
    f"tuna {'show_threads -C' if bool(SosPredicate(self, cmd_outputs=co)) else 'CP'}"
)
```

jcastill: Ack, this looks really elegant. I'll make the change and push again.

pmoravec (approved):

sos/report/plugins/kernelrt.py
jcastill: I moved the predicate test outside of the f-string because when I tried to break the curly braces and couldn't find a clean way. As always, any suggestion is welcome!

sos/report/plugins/kernelrt.py
```diff
@@ -36,6 +36,12 @@ def setup(self):
...
         # defined this command throws an exception on versions prior to
         # 0.10.4-5.
-        self.add_cmd_output('tuna -CP')
+        co = {'cmd': 'tuna --help', 'output': '-P'}
+        option_present = self.test_predicate(
+                self, pred=SoSPredicate(self, cmd_outputs=co)
+                )
+        self.add_cmd_output(
+                f"tuna {'-CP' if option_present else 'show_threads -C'}"
+                )
```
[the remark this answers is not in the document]
jcastill: Ack, done!

TurboTurtle (approved):

PR #3397, opened 2023-10-27
[salt] Update salt plugins to gather more data

m-czernek:
I'd like to modify the `salt` and `saltmaster` plugins to:

- Add support for the `venv-salt-minion` package
- Gather more useful data, such as:
  - Systemctl status information
  - Version reports
  - Running job reports
  - Grains/pillars

and others.

Related downstream issue: https://github.com/SUSE/spacewalk/issues/22700

---

### DIFF
--- a/sos/report/plugins/salt.py
+++ b/sos/report/plugins/salt.py
@@ -5,6 +5,7 @@
 # version 2 of the GNU General Public License.
 #
 # See the LICENSE file in the source distribution for further information.
+import re
 
 from sos.report.plugins import Plugin, IndependentPlugin
 
@@ -16,7 +17,7 @@ class Salt(Plugin, IndependentPlugin):
     plugin_name = 'salt'
     profiles = ('sysmgmt',)
 
-    packages = ('salt', 'salt-minion', 'salt-common',)
+    packages = ('salt', 'salt-minion', 'venv-salt-minion', 'salt-common',)
 
     def setup(self):
         all_logs = self.get_option("all_logs")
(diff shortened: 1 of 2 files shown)


pmoravec (approved):

sos/report/plugins/salt.py
```diff
@@ -26,12 +27,38 @@ def setup(self):
...
+
+        self.add_copy_spec(["/etc/salt",
+                            "/etc/venv-salt-minion/",
+                            "/usr/local/etc/salt"])
+        self.add_forbidden_path(["/etc/salt/pki/*/*.pem",
+                                 "/etc/venv-salt-minion/pki/*/*.pem",
+                                 "/usr/local/etc/salt/pki/*/*.pem"])
+
+        self.add_copy_spec(["/etc/machine-id",
+                            "/var/lib/dbus/machine-id"])
```
TurboTurtle, line 41: Style nit - sos style is to indent from the caller for list elements and close the bracket on the ending line, e.g.:

```python
self.add_copy_spec([
    '/path/one',
    '/path/two',
])
```

sos/report/plugins/saltmaster.py
```diff
@@ -26,7 +29,38 @@ def setup(self):
 
         self.add_copy_spec("/etc/salt")
         self.add_forbidden_path("/etc/salt/pki/*/*.pem")
-        self.add_cmd_output("salt-key --list all")
+
+        self.add_copy_spec(["/etc/machine-id",
+                            "/var/lib/dbus/machine-id"])
```
TurboTurtle, line 34: Ditto here.

sos/report/plugins/saltmaster.py
```diff
@@ -26,7 +29,38 @@ def setup(self):
...
+            "salt-run manage.list_not_state --out=yaml",
+            "salt-run manage.joined --out=yaml"
+            ], timeout=30)
+
+    def add_pillar_roots(self):
+        cfgs = glob.glob("/etc/salt/master.d/*conf")
+        main_cfg = "/etc/salt/master"
+
+        if os.path.exists(main_cfg):
+            cfgs.append(main_cfg)
```
TurboTurtle, line 53: There is a wrapper in `Plugin` for this - `Plugin.path_exists()` which automatically handles a non `/` sysroot, in the event we're running in e.g. a container or recovery environment.

TurboTurtle (commented):
A couple notes, but nothing major to change before this looks good to merge.

sos/report/plugins/salt.py
m-czernek: Unified all lists with this indentation style.

sos/report/plugins/saltmaster.py
m-czernek: Fixed, nice catch!

m-czernek:
Folks, I believe the CI failed because of some external factor (e.g. squashing and force pushing my commits, or something similar?); is there a way to re-run it? (I don't think I have the privileges to do so?)

pmoravec:
> Folks, I believe the CI failed because of some external factor (e.g. squashing and force pushing my commits, or something similar?); is there a way to re-run it? (I don't think I have the privileges to do so?)

Indeed, once upon a time (a week?), there is some infra issue causing jobs are not spawned properly. I restarted them now.

sos/report/plugins/salt.py
```diff
@@ -26,12 +27,46 @@ def setup(self):
...
+        ])
+        self.add_forbidden_path([
+            "/etc/salt/pki/*/*.pem",
+            "/etc/venv-salt-minion/pki/*/*.pem",
+            "/usr/local/etc/salt/pki/*/*.pem",
+        ])
+
+        self.add_copy_spec([
+            "/etc/machine-id",
+            "/var/lib/dbus/machine-id",
```
pmoravec, line 48: We collect it in `dbus` plugin already, while our aim is different plugins collect different data.

Is there some reason to keep this here? Or reversely, does it make sense if `dbus` plugin collects also `/etc/machine-id`? (not that `sos` collecting `/var/lib/dbus/machine-id` - which is usually a symlink to `/etc/machine-id` - does collect also the symlink destination, so there is no explicit need to collect it - until the symlink is not correct..?)

sos/report/plugins/saltmaster.py
```diff
@@ -26,7 +28,40 @@ def setup(self):
 
         self.add_copy_spec("/etc/salt")
         self.add_forbidden_path("/etc/salt/pki/*/*.pem")
-        self.add_cmd_output("salt-key --list all")
+
+        self.add_copy_spec([
+            "/etc/machine-id",
+            "/var/lib/dbus/machine-id",
+        ])
```
pmoravec, line 35: Ditto above; we should have just one plugin collecting same file(s).

sos/report/plugins/salt.py
m-czernek: Ah my bad, I was testing this with the `-k` option to execute only the salt plugins. 

Re `/etc/machine-id`, I found it in the documentation as the primary location to the alternative `/var/lib/dbus/machine-id`  location, so I simply included both. I see that the `dbus` plugin already collects the file as you say, so I removed it from this PR.

TurboTurtle (approved):

pmoravec (approved):

PR #3398, opened 2023-10-27
[build] Actually indicate this requires packaging

hroncok:
This is a followup for 56904018045fdbdad666e55f0a850bc365809e50

---

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,5 +3,5 @@ coverage>=4.0.3
 Sphinx>=1.3.5
 pexpect>=4.0.0
 pyyaml
-setuptools
+packaging
 
--- a/sos.spec
+++ b/sos.spec
@@ -10,7 +10,11 @@ BuildRequires: python3-devel
 BuildRequires: python3-setuptools
 Requires: python3-rpm
 Requires: python3-pexpect
+%if 0%{?rhel} && 0%{?rhel} < 10
 Requires: python3-setuptools
+%else
+Requires: python3-packaging
+%endif
 Recommends: python3-magic
 # Mandatory just for uploading to a SFTP server:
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -3,5 +3,5 @@ coverage>=4.0.3
 Sphinx>=1.3.5
 pexpect>=4.0.0
 pyyaml
-setuptools
+packaging
```
hroncok, line 6: Note: I don't know what is this file used for. Perhaps it needs setuptools listed, as a build dependency, perhaps not.

sos.spec
```diff
@@ -10,7 +10,7 @@ BuildRequires: python3-devel
 BuildRequires: python3-setuptools
 Requires: python3-rpm
 Requires: python3-pexpect
-Requires: python3-setuptools
+Requires: python3-packaging
```
hroncok, line 13: Note that in RHEL 9+ and Fedora, this line is redundant, as the dependency is generated, but so is the one for python3-pexpect above, so I decided to keep it consistent.

pmoravec: Sadly CentOS 8 and RHEL8 does not seem to have `python3-packaging` package, until I overlooked something..?

hroncok: Not in BaseOS at least. I had no idea you do rebases in RHEL 8. This can be conditionalized, but only in spec, not in the setup.py package metadata.

pmoravec: We do sos rebases in RHEL8. Please set this conditional in spec file, if nothing else to let centos-stream-8 CI tests to succeed.

pmoravec (approved):
Sounds great, thanks!

TurboTurtle (approved):

Issue #3399, opened 2023-10-27
sos collect in Juju environment


106106:
Hi sos team,

I'm trying to use sos collect in a juju environment. I see there is this [pull request](https://github.com/sosreport/sos/pull/3192) and this [issue](https://github.com/sosreport/sos/issues/2668) so I believe the integration with Juju is implemented? But I'm struggling to get it working.

I'm using:
```
apt-cache policy sosreport
sosreport:
  Installed: 4.5.6-0ubuntu1~22.04.2
  Candidate: 4.5.6-0ubuntu1~22.04.2
  Version table:
 *** 4.5.6-0ubuntu1~22.04.2 500
        500 http://archive.ubuntu.com/ubuntu jammy-updates/main amd64 Packages
        100 /var/lib/dpkg/status
     4.3-1ubuntu2.1 500
        500 http://security.ubuntu.com/ubuntu jammy-security/main amd64 Packages
     4.3-1ubuntu2 500
        500 http://archive.ubuntu.com/ubuntu jammy/main amd64 Packages
```

I have a Juju deployed environment. For example, and have tried something like:

```
multipass launch -n test-environment -c 4 -m 16G -d 50G
multipass shell test-environment
ssh-keygen
lxd init --auto
lxc network set lxdbr0 ipv6.address none
sudo snap install juju --channel 3.1/stable
mkdir -p ~/.local/share
juju bootstrap localhost controller-1
juju add-model model-1
juju ssh-keys
juju deploy vault
watch -c juju status --color
juju ssh vault/0
exit
sos collect --cluster-type=juju --batch -c juju.apps=vault --nopasswd-sudo 
sos collect --cluster-type=juju --batch -c juju.apps=vault --nopasswd-sudo --ssh-user ubuntu
sos collect --cluster-type=juju --batch -c juju.units=vault/0 --nopasswd-sudo --ssh-user ubuntu
sos collect --cluster-type=juju --batch -c juju.machines=0 --nopasswd-sudo --ssh-user ubuntu
```
and 
```
juju ssh vault/0
```
authenticates me into the unit. But from the same host (that I used `juju ssh vault/0` from) I can't seem to figure out how to use sos collect. I've tried the following:
```
sudo sos collect --cluster-type=juju --batch -c juju.apps=vault
```
and other options/combinations but I keep getting the following:
```
[...]
sos-collector ASSUMES that SSH keys are installed on all nodes unless the
--password option is provided.

[...]

PR #3400, opened 2023-10-30
[juju][collect] Fix new juju collection

arif-ali:
New version of juju uses colorisation, and therefore juju status and json.loads doesn't load the juju status correctly. By using --no-color based on the version of juju this should this particular use-case

Resolves: #3399
Resolves: SET-339

---

### DIFF
--- a/sos/collector/clusters/juju.py
+++ b/sos/collector/clusters/juju.py
@@ -13,6 +13,8 @@ import json
 import re
 
 from sos.collector.clusters import Cluster
+from sos.utilities import parse_version
+from sos.utilities import sos_get_command_output
 
 
 def _parse_option_string(strings=None):
@@ -156,9 +158,16 @@ class juju(Cluster):
 
         return index
 
+    def _get_juju_version(self):
+        """Grab the version of juju"""
+        res = sos_get_command_output("juju version")
+        return res['output'].split("-")[0]
+
     def _execute_juju_status(self, model_name):
         model_option = f"-m {model_name}" if model_name else ""
         format_option = "--format json"
+        if parse_version(self._get_juju_version()) > parse_version("3"):
+            format_option += " --no-color"
         status_cmd = f"{self.cmd} status {model_option} {format_option}"
         res = self.exec_primary_cmd(status_cmd)
         if not res["status"] == 0:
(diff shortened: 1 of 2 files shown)


TurboTurtle (approved):
LGTM.

We're trying to push f-strings, but here it'd be a bit awkward without re-writing the entire command slicing, so we'll put this one on the back burner for another time.

As an aside, I can't imagine we're the only ones calling `juju` so it may be worth working with the juju devs to automatically detect when to use `--no-color` from the `juju` command for situations like this (e.g. when not called from a shell).

sos/collector/clusters/juju.py
```diff
@@ -156,9 +158,17 @@ def _get_model_info(self, model_name):
 
         return index
 
+    def _get_juju_version(self):
+        """Grab the version of juju"""
+        cmd = "juju version"
+        res = sos_get_command_output(cmd)
```
pmoravec, line 164: Why not single

    res = sos_get_command_output("juju version")

?

(or maybe even merge it with `return`, though that might sacrifice code readability too much)

arif-ali:
> As an aside, I can't imagine we're the only ones calling `juju` so it may be worth working with the juju devs to automatically detect when to use `--no-color` from the `juju` command for situations like this (e.g. when not called from a shell).

Actually, I didn't think too much of it, and I think that is valid point.

So, I have opened a bug on our end.

I think, at this point, I'm on the opinion that I would put this as a draft/wip, until we are 100%, and if it does get fixed in `juju`, then there's no point on this PR.

TurboTurtle:
IIRC, that is due to the default usage of `ssh` and some behavior of that we found during early testing. I'm open to making that toggle-able at the transport in use, which would allow `juju` to avoid it (and at the very least provide us the opportunity to re-test `ssh`).

On that note, it's weird that we're setting this in a `Cluster` profile, and not having it fully determined in the transport.

I think the best approach here is to remove the `pty` determination from cluster, let the transports determine it via a (to be added) property, with current behavior as the default and an override for `juju`.

I'm happy to move bits around on this (though likely in a few days as day job has me tied up for the week), but if you'd like to handle it within this or another PR, that's fine too, just let me know.

PR #3402, opened 2023-11-02
[pacemaker] Replace deprecated "pcs property list" command

yinoguchi:
This patch replaces "pcs property list" with "pcs property config" because "pcs property list" is deprecated and will be removed:

```
# pcs property list --all 
Deprecation Warning: This command is deprecated and will be removed. Please use 'pcs property config' instead.
Cluster Properties:
...
```
This warning message was introduced in pcs v0.10.9:
https://github.com/ClusterLabs/pcs/commit/332f56ade4773bf22e2ee6c8c91962506b844135#diff-4c58cd856f4e8b23a9d99d961c357e224a9f6314542d50add71a04a2b0b6980cR96-R104

---

### DIFF
--- a/sos/report/plugins/pacemaker.py
+++ b/sos/report/plugins/pacemaker.py
@@ -8,6 +8,7 @@
 
 from sos.report.plugins import (Plugin, RedHatPlugin, DebianPlugin,
                                 UbuntuPlugin, PluginOpt)
+from sos.utilities import parse_version
 from datetime import datetime, timedelta
 import re
 
@@ -42,13 +43,23 @@ class Pacemaker(Plugin):
         ])
 
     def setup_pcs(self):
+        pcs_pkg = self.policy.package_manager.pkg_by_name('pcs')
+        if pcs_pkg is None:
+            return
+
         self.add_copy_spec("/var/log/pcsd/pcsd.log")
         self.add_cmd_output([
             "pcs stonith sbd status --full",
             "pcs stonith sbd watchdog list",
             "pcs stonith history show",
-            "pcs property list --all"
         ])
+
+        pcs_version = '.'.join(pcs_pkg['version'])
+        if parse_version(pcs_version) > parse_version('0.10.8'):
+            self.add_cmd_output("pcs property config --all")
+        else:
+            self.add_cmd_output("pcs property list --all")
+
         self.add_cmd_output("pcs config", tags="pcs_config")
         self.add_cmd_output("pcs quorum status", tags="pcs_quorum_status")
         self.add_cmd_output("pcs status --full", tags="pcs_status")


arif-ali (approved):

pmoravec (commented):
There theoretically might be old versions of `pcs` deployed (i.e. RHEL8.4 ships `0.10.8` which should have `pcs property list` only).

@sbradley7777 , do you know if it could be a problem?

jesusserranost:
@pmoravec - This will be definitely a problem unless there is some kind of internal check which runs either `pcs property list --all` or `pcs property config --all`, depending on the currently installed version of the `pcs` package. I can confirm there are still clusters deployed with older versions which are still officially supported (as of today), for which this change will break the functionality. I just tested this on a RHEL 8.2 lab and the new command just prints the "Usage help syntax", as expected:

```
[root@fastvm-rhel-8-2-201 ~]# pcs property config --all | head -5

Usage: pcs property ...
    [list|show [<property> | --all | --defaults]] | [--all | --defaults]
        List property settings (default: lists configured properties).
        If --defaults is specified will show all property defaults, if --all
```

pmoravec:
There are a few options:
- collect both outputs (once duplicated or invalid command)
- collect one output depending on `pcs` version - we have a check for it, see https://github.com/sosreport/sos/blob/main/sos/collector/clusters/pacemaker.py#L14 and https://github.com/sosreport/sos/blob/main/sos/collector/clusters/pacemaker.py#L66 applied to something like https://github.com/sosreport/sos/blob/main/sos/report/plugins/networking.py#L270 .

I would personally go for 2nd option: just have a check and use appropriate option per it.

yinoguchi:
Thank you for the review. 
As suggested, I have made modifications to check `pcs` version and change the command to be executed accordingly.

sos/report/plugins/pacemaker.py
```diff
@@ -42,13 +43,21 @@ def setup_crm_shell(self):
...
+
         self.add_copy_spec("/var/log/pcsd/pcsd.log")
         self.add_cmd_output([
             "pcs stonith sbd status --full",
             "pcs stonith sbd watchdog list",
             "pcs stonith history show",
-            "pcs property list --all"
         ])
+
+        pcs_version = '.'.join(pcs_pkg['version'])
```
pmoravec, line 55: Try running `sos report -o pcaemaker` on a system *without* `pcs` package installed - you get

    TypeError: 'NoneType' object is not subscriptable

here (CI tests fail on this)

yinoguchi: Thanks again. How about this? I had added a return if the `pcs` package does not exist. All commands within the `setup_pcs()` depend on the `pcs` package, so I believe this should be fine. 
Of course, if there are any problems with this, I will make corrections as per your suggestion.

yinoguchi: Oh, sorry.
```
flake8 sos
sos/report/plugins/pacemaker.py:48:12: E111 indentation is not a multiple of 4
```
I will fix this and push it again.

pmoravec (approved):
What an elegant solution I have overlooked :)

TurboTurtle (approved):

PR #3405, opened 2023-11-03
[collect] Refactor `get_pty` functionality

TurboTurtle:
The `get_pty` parameter for remote executed commands was both a bit of a misnomer and applied too broadly.

Refactor this to `use_shell` to be more obvious about what the intent behind the option is, and default all transports to `False`, so that by default we do not wrap any commands in a bash shell.

This may be overriden on a per-transport basis via the ned `_need_shell` property within transport subclasses. Further, this facility has been expanded to be allowed on a per-command basis from `SoSNode.run_command()` and wherever that is linked.

Related: #3399
Related: #3400

---

### DIFF
--- a/sos/collector/clusters/__init__.py
+++ b/sos/collector/clusters/__init__.py
@@ -294,7 +294,8 @@ class Cluster():
         """
         return node.address == self.primary.address
 
-    def exec_primary_cmd(self, cmd, need_root=False, timeout=180):
+    def exec_primary_cmd(self, cmd, need_root=False, timeout=180,
+                         use_shell='auto'):
         """Used to retrieve command output from a (primary) node in a cluster
 
         :param cmd: The command to run
@@ -306,12 +307,14 @@ class Cluster():
         :param timeout:  Amount of time to allow cmd to run in seconds
         :type timeout: ``int``
 
+        :param use_shell:   Does the command required execution within a shell?
+        :type use_shell:    ``auto`` or ``bool``
+
         :returns: The output and status of `cmd`
         :rtype: ``dict``
         """
-        pty = self.primary.local is False
-        res = self.primary.run_command(cmd, get_pty=pty, need_root=need_root,
-                                       timeout=timeout)
+        res = self.primary.run_command(cmd, need_root=need_root,
+                                       use_shell=use_shell, timeout=timeout)
         if res['output']:
             res['output'] = res['output'].replace('Password:', '')
         return res
--- a/sos/collector/sosnode.py
+++ b/sos/collector/sosnode.py
@@ -443,20 +443,25 @@ class SosNode():
             return False
         return self.host.package_manager.pkg_by_name(pkg) is not None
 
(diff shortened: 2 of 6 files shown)


TurboTurtle:
@arif-ali I believe this should fix the pty issue we discussed in #3400 and in turn I think resolves #3399, though I don't have a `juju` setup available to me to test this with.

In my testing, I found I did not have any issues with the default `ssh` transport collections, so I'm leaving it set to `False` for everything, but this approach at least allows for a much easier toggle for any future transports we may add that might need it for whatever reason (and as the commit message notes, allows per-command toggling too).

pmoravec (approved):

arif-ali:
> @arif-ali I believe this should fix the pty issue we discussed in #3400 and in turn I think resolves #3399, though I don't have a `juju` setup available to me to test this with.
> 
> In my testing, I found I did not have any issues with the default `ssh` transport collections, so I'm leaving it set to `False` for everything, but this approach at least allows for a much easier toggle for any future transports we may add that might need it for whatever reason (and as the commit message notes, allows per-command toggling too).

I have been trying to test this week, but have not been successful, as I am still getting coloured output whether I use `use_shell=False` or `use_shell=True`. Added a few print statements in the juju.py clusters file, and we have the same issue.

I am curious, if we need a different option that needs to be set here :shrug:

TurboTurtle:
If you run with `-vvv`, are you seeing [this line](https://github.com/sosreport/sos/pull/3405/files#diff-073bf5ff1b0f5ea4cb94d7440664fac5e05e496601ece19838dead00e0912859R234) get logged for the `juju` commands?

arif-ali:
[Here's a pastebin](https://paste.ubuntu.com/p/gyH828HFTX/) with use_shell set to False and then True.

Both of the json outputs on my screen are coloured

TurboTurtle:
Ah, ok - I see what's going on here now. This is coming from `pexpect` - notably because we are using `spawn()` and not `pexpect.popen_spawn.PopenSpawn`. The former ultimately leverages `ptyprocess` to execute the command, where as the latter uses `subprocess`.

It's not immediately clear to me though is using `PopenSpawn` would resolve this for us, and my own local testing is being interrupted by the fact that changing out the `PopenSpawn` breaks our output parsing and it'd likely be a decent effort to fix that in every place before even being able to test if this actually does anything for our immediate issue.

With that being the case, I think we'll have to modify the `juju` transport to use `--no-color` to fix that transport going forward. That said, I _think_ this change here is still relevant as I remember some cases in my initial development of `sos collect` where we absolutely needed to wrap remote commands in a bash shell, and that kind of functionality should be controlled by the transport as this change makes happen.

So, TL;DR:

- pexpect is forcing a pty allocation
- not an easy swap out to avoid that
- merge this change
- follow up with a PR for `juju` to use `--no-color`


Sound like a plan?

arif-ali:
@TurboTurtle sounds good to me, I'll just make my PR available for review, and go from there. Will do a rebase to this one, just to make sure all is good before requesting approval

Issue #3404, opened 2023-11-03
[flatpak] Missing flatpak versions break sos execution


TurboTurtle:
If `flatpak list` contains flatpaks without version numbers, e.g.:

```
$ sudo flatpak list --columns=name,version
Name                                                   Version
Spotify                                                1.1.55.498.gf9a83c60
Freedesktop Platform                                   19.08.8
Freedesktop Platform                                   20.08.13
default
Mesa                                                   21.1.1
ffmpeg-full
openh264
openh264                                               2.1.0
Fractal                                                4.4
GNOME Application Platform version 3.38
Adwaita theme
```

`sos` execution breaks when a policy tries to load the flatpak package manager abstraction:

```
  File "/home/jake/projects/sos/sos/policies/package_managers/__init__.py", line 67, in packages
    self._generate_pkg_list()
  File "/home/jake/projects/sos/sos/policies/package_managers/__init__.py", line 188, in _generate_pkg_list
    for pkg in self._parse_pkg_list(pkg_list):
  File "/home/jake/projects/sos/sos/policies/package_managers/flatpak.py", line 27, in _parse_pkg_list
    name, version = pkg[0], pkg[1]
                            ~~~^^^
IndexError: list index out of range
```

The `flatpak` manager does not have any current error handling associated with it:

```python
    def _parse_pkg_list(self, pkg_list):
        for line in pkg_list.splitlines():
            pkg = line.split()
            name, version = pkg[0], pkg[1]
            yield (name, version, None)

# vim: set et ts=4 sw=4 :
```

This is a straight forward fix from the python perspective - we need to wrap this in a try block, but the question becomes what do we do with version-less flatpaks?

If we don't include the flatpak in the package manager's package list, we could theoretically end up in a situation where a user expects some plugin to do something based on a flatpak being present and we don't do that $something. Alternatively, the package manager abstraction expects _some_ kind of version string here. Passing `None` will just break things further down the line. Should we set some bogus version like `0.0`? to ensure package checks pass, but in turn version checks will almost certainly fail?

Looking for feedback on this one. I'm personally of the mind that if there is no version associated with it, we should _not_ consider that flatpak to be present for our purposes, but I also am not familiar with flatpaks at all so I don't know how common this actually is.

jcastill:
[...]

PR #3406, opened 2023-11-03
[flatpak] Capture details about branch

jcastill:
Gather details about branch, so it complements
the version number (which can be empty sometimes).

Fixes: #3404

---


sos/policies/package_managers/flatpak.py
```diff
@@ -15,16 +15,20 @@ class FlatpakPackageManager(PackageManager):
...
     files_command = ''
     verify_command = ''
     verify_filter = ''
 
     def _parse_pkg_list(self, pkg_list):
-        for line in pkg_list.splitlines():
-            pkg = line.split()
-            name, version = pkg[0], pkg[1]
-            yield (name, version, None)
+        for line in pkg_list.splitlines()[1:]:
```
pmoravec, line 25: Why the `[1:]`? The `flatpak list` does not start with a header, the very first line already contains a package - at least in my case..

sos/policies/package_managers/flatpak.py
```diff
@@ -15,16 +15,20 @@ class FlatpakPackageManager(PackageManager):
...
+            pkg = line.split("\t")
+            name, branch = pkg[0], pkg[2]
+            if not pkg[1]:
+                version = ''
+            else:
+                version = pkg[1]
+            yield (name, version, branch)
```
pmoravec, line 32: Why not simple `yield (pkg[0], pkg[1], pkg[2])` ?  (maybe worth checking `len(pkg)>2`, just in case of truncated output (or that is over-paranoic..?))

sos/policies/package_managers/flatpak.py
jcastill: It does in my case, but if I recall correctly, the output is different when called from sos, and there's no header there. I've fixed this in the next version.

sos/policies/package_managers/flatpak.py
jcastill: Checking the length of pkg and then yielding either pkg[1] or None seems better than just yielding pkg[1] regardless of it being there or not:

            if len(pkg) > 2:
                yield (pkg[0], pkg[1], pkg[2])
            else:
                yield (pkg[0], None, pkg[2])

Unless yielding pkg[1] when doesn't exist changes it to None on the yield, which then may simply the code a lot. But I saw in the other sources in package_managers that we explicitly set to None.

pmoravec: Ah right, I am not sure now if the `split` returns `None` or empty string. But can we pass `None` as version? Does not we expect a string value later on?

jcastill: Later on I'm not sure, but if that's the case, we can fail back to "" instead of None. Let me see if I can test with packages with empty value and see what happens with None.

pmoravec: OK, `split` returns empty string and not `None` (which is in fact obvious, it returns list of strings):

    >>> "name\tvers\tbranch".split("\t")
    ['name', 'vers', 'branch']
    >>> "name\t\tbranch".split("\t")
    ['name', '', 'branch']
    >>> 

And as far as I checked, we assume strings there.

jcastill: And I just checked what we get with yield (pkg[0], pkg[1], pkg[2]) without checking pkg lenght, from _generate_pkg_list, and it is an empty string. So just that line should be enough.

TurboTurtle (approved):

PR #3407, opened 2023-11-06
[mellanox_firmware] Add Mellanox firmware plugin

naserdean:
This patch reports the output of the following Mellanox firmware commands:
mlxdump
mstconfig
mstdump
mstflint
mlxreg
mlxlink
mlxcables
mst

Additionally, if the user allows system changes, we will atempt to start the mst device, add the cables and stop the device in the end.

---

### DIFF
--- a/sos/report/plugins/mellanox_firmware.py
+++ b/sos/report/plugins/mellanox_firmware.py
@@ -1,0 +1,141 @@
+# Copyright (C) 2023 Nvidia Corporation, Alin Serdean <aserdean@nvidia.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, IndependentPlugin
+import os
+import time
+
+
+class MellanoxFirmware(Plugin, IndependentPlugin):
+
+    short_desc = 'Nvidia(Mellanox) firmware tools output'
+
+    plugin_name = "mellanox_firmware"
+    profiles = ('hardware', 'system')
+    packages = ('mst', 'mstflint')
+
+    MLNX_STRING = "Mellanox Technologies"
+
+    def check_enabled(self):
+        """
+        Checks if this plugin should be executed at all.
+        We will only enable the plugin if there is a
+        Mellanox Technologies network adapter
+        """
+        lspci = self.exec_cmd("lspci -D -d 15b3::0200")
+        return lspci['status'] == 0 and self.MLNX_STRING in lspci['output']
+
+    def collect(self):
+        if not self.get_option('allow_system_changes'):
+            self._log_info("Skipping mst/mlx cable commands as system changes"
+                           "would be made. Use --allow-system-changes to"
+                           "enable this collection.")
+            return
+
+        """
+        Run only if mft package is installed.
+        flint is available from the mft package.
+        """
+        co = self.exec_cmd('flint --version')
+        if co['status'] != 0:
+            return
+
+        co = self.collect_cmd_output('mst start')
+        if co['status'] != 0:
+            return
+
+        self.collect_cmd_output('mst cable add')
+        self.collect_cmd_output("mst status -v", timeout=10)
+        self.collect_cmd_output("mlxcables", timeout=10)
+        co = os.listdir("/dev/mst")
+        mlxcables = []
+        for device in co:
+            if 'cable' in device:
+                mlxcables.append(device)
+        for mlxcable in mlxcables:
+            self.collect_cmd_output(f"mlxcables -d {mlxcable} --DDM",
+                                    timeout=10)
+            self.collect_cmd_output(f"mlxcables -d {mlxcable} --dump",
+                                    timeout=10)
+        self.collect_cmd_output("mst stop", changes=True)
+
+    def setup(self):
+        # Get all devices which have the vendor Mellanox Technologies
+        devices = []
+        device_list = self.collect_cmd_output('lspci -D -d 15b3::0200')
+        """
+        Will return a string of the following format:
+        0000:08:00.0 Ethernet controller: Mellanox Technologies MT2892
+        Family
+        """
+        if device_list['status'] != 0:
+            # bail out if there no Mellanox PCI devices
+            return
+
+        for line in device_list["output"].splitlines():
+            """
+            Should return 0000:08:00.0
+            from the following string
+            0000:08:00.0 Ethernet controller: Mellanox Technologies MT2892
+            Family
+            """
+            devices.append(line[0:8]+'00.0')
+
+        devices = set(devices)
+
+        """
+        # Mft package is present if OFED is installed
+        # mstflint package is part of the distro and can be installed.
+        """
+        commands = []
+
+        # mft package is installed if flint command is available
+        co = self.exec_cmd('flint --version')
+        if co['status'] != 0:
+            """
+            mstflint package commands
+            the commands do not support position independent arguments
+            """
+            commands = [
+                ["mstconfig -d ", " -e q"],
+                ["mstflint -d ", " dc"],
+                ["mstflint -d ", " q"],
+                ["mstreg -d ", " --reg_name ROCE_ACCL --get"],
+                ["mstlink -d ", ""],
+            ]
+        else:
+            """
+            mft package commands
+            the commands do not support position independent arguments
+            """
+            commands = [
+                ["mlxdump -d ", " pcie_uc --all"],
+                ["mstconfig -d ", " -e q"],
+                ["flint -d ", " dc"],
+                ["flint -d ", " q"],
+                ["mlxreg -d ", " --reg_name ROCE_ACCL --get"],
+                ["mlxlink -d ", ""],
+                ["fwtrace -d ", " -i all --tracer_mode FIFO"],
+            ]
+        for device in devices:
+            for command in commands:
+                self.add_cmd_output(f"{command[0]} {device} "
+                                    f"{command[1]}", timeout=30)
+
+            """
+            Dump the output of the mstdump command three times
+            waiting for one second. This output is useful to check
+            if certain registers changed
+            """
+            for i in range(3):
+                self.add_cmd_output(f"mstdump {device}")
+                time.sleep(1)
+
+# vim: set et ts=4 sw=4 :


sos/report/plugins/mellanox_firmware.py
```diff
@@ -0,0 +1,126 @@
...
+import time
+
+
+class MellanoxFirmware(Plugin, IndependentPlugin):
+
+    short_desc = 'Nvidia(Mellanox) firmware tools output'
+
+    plugin_name = "mellanox_firmware"
+    profiles = ('hardware', 'system')
+
```
TurboTurtle, line 22: As written, this will enable the plugin by default on every system that runs sos. I don't think that's intended, especially given the predicates below.

Adding at least a `packages` tuple here, and/or a `files`, `commands`, or `kernel_mods` will restrict the plugin to only running on systems that match those enablement triggers.

sos/report/plugins/mellanox_firmware.py
```diff
@@ -0,0 +1,126 @@
...
+            # from the following string
+            # 0000:08:00.0 Ethernet controller: Mellanox Technologies MT2892
+            # Family
+            devices.append(line[0:8]+'00.0')
+
+        devices = set(devices)
+
+        # Collect info from certain commands, only if mft package is present
+        # Mft package is present if OFED is installed
+        mft_pred = SoSPredicate(self, packages=['mft'])
```
TurboTurtle, line 45: Setting the `packages` tuple to include `mft` will accomplish this a bit better.

sos/report/plugins/mellanox_firmware.py
```diff
@@ -0,0 +1,126 @@
...
+
+        for device in devices:
+            self.add_cmd_output("mlxdump -d %s pcie_uc --all" % device,
+                                pred=mft_pred)
+            self.add_cmd_output("mstconfig -d %s -e q" % device,
+                                pred=mft_pred)
+            self.add_cmd_output("flint -d %s dc" % device,
+                                pred=mft_pred)
+            self.add_cmd_output("flint -d %s q" % device,
+                                pred=mft_pred)
```
TurboTurtle, line 74: Generally speaking it is fine for sos plugins to call `add_cmd_output()` without a predicate for commands that don't exist on a system - we handle those gracefully and don't pollute the plugin directory with empty output.

This is particularly why I say using the `packages` tuple for the plugin is likely a better option here.

sos/report/plugins/mellanox_firmware.py
```diff
@@ -0,0 +1,126 @@
...
+            mlxcables = []
+            for device in co:
+                if 'cable' in device:
+                    mlxcables.append(device)
+            for mlxcable in mlxcables:
+                self.add_cmd_output("mlxcables -d %s --DDM" % mlxcable,
+                                    timeout=10)
+                self.add_cmd_output("mlxcables -d %s --dump" % mlxcable,
+                                    timeout=10)
+            self.add_cmd_output("mst stop", changes=True)
```
TurboTurtle, line 124: I'd need some more information/context on what is happening here. `--allow-system-changes` is generally meant to gate kernel mods being loaded or the like. We don't want sos making configuration changes even if this option is used, as that is out of scope for the intended use of both sos and the option.

TurboTurtle (commented):
Comments below about the gating approach used - I think we can clean that up a bit, but overall I think the design of the plugin is acceptable once that is done and we have more details on what the bits gated by `--allow-system-changes` is wanting to do.

In addition to the comments below, please note that this project is attempting to transition/convert to f-strings for string substitution/interpolation, so let's look at replacing the legacy substitutions before merge.

sos/report/plugins/mellanox_firmware.py
arif-ali: In my previous role, I was using Mellanox IB regularly, and one of the things we had to do to grab extra debug data was to start `mst` to grab the extra debugs, and then stop it. This service is not started by default, and used for debugging from what I can remember. For me, this is a +1

sos/report/plugins/mellanox_firmware.py
arif-ali: I guess we also would advise f-strings here?

sos/report/plugins/mellanox_firmware.py
naserdean: Sorry, I missed `commands` option...
I added it as part of the incremental.

Regarding packages, `mft`, `mstflint` or both could be installed on the system...
I'm unsure how I could describe this as part of the `packages` tuple.

sos/report/plugins/mellanox_firmware.py
naserdean: Unfortunately as stated above I can't limit just on `mft` or `mstflint` package.
I am unsure how can I achieve the above in a better way.

sos/report/plugins/mellanox_firmware.py
naserdean: Makes sense. 
That means I could safely get rid of the predicates: `mft_mlxfw_pred` and `mstflint_mlxfw_pred` since they further add constraints to `mft_pred` and `mstflint_pred`?

I also switched to f-strings. Hopefully I did not miss any lines.

sos/report/plugins/mellanox_firmware.py
naserdean: To add a bit more context, maybe I should add it as a comment, running `mst start` will load two kernel modules:
`mst_pci`
`mst_pciconf`
which will allow the capture, as @arif-ali described above, additional debug information.
There are some checks in order to validate the pci configuration after the two kernel modules are started.
running `mst stop` will unload those kernel modules.

TurboTurtle: Hmm, ok. I'm onboard with that, however I think this is a better use case for the (relatively new) `collect()` approach instead of `setup()`.

From a technical perspective, when you do an `add_cmd_output()` within `setup()`, you're adding those commands to a list that gets executed later in our run after all plugins have finished their setup. Once those collections are done we move into a "manual collections" phase which invokes each plugin's `collect()` method if defined. Usually this is used to "manually" generate or compile data and then write that to a plugin-generated file via `Plugin.collection_file`, however I'm thinking that we'd prefer to have a service start/stop controlled within this phase. @pmoravec feel free to chime in if you disagree and think we should keep this in `setup()`.

Practically, what that means here is that instead of doing collections with `add_cmd_output()`, you would use `collect_cmd_output()` to execute it right then and there and save it to the archive, while using `exec_cmd()` for anything that needs to be run but you don't care about saving the output. E.G. perhaps something like this:

```python
def collect(self):
    if not self.get_option('allow_system_changes'):
        return
    self.exec_cmd('mst start')
    self.collect_cmd_output('mst status -v')
    [... etc you get the idea ...]
    self.exec_cmd('mst stop')
```

sos/report/plugins/mellanox_firmware.py
```diff
@@ -0,0 +1,128 @@
...
+import time
+
+
+class MellanoxFirmware(Plugin, IndependentPlugin):
+
+    short_desc = 'Nvidia(Mellanox) firmware tools output'
+
+    plugin_name = "mellanox_firmware"
+    profiles = ('hardware', 'system')
+    commands = ('lspci -D -d 15b3::0200',)
```
TurboTurtle, line 22: This is not what `commands` is for. This is used as an enablement trigger if the specified binary exists and is executable. E.G. `commands = ('mstconfig',)`  would enable this plugin only if that command exists (assuming no other enablement triggers are set).

sos/report/plugins/mellanox_firmware.py
TurboTurtle: The enablement triggers, like `packages`, operates on an `any` approach - so as long as at least one of the listed packages is installed, it will enable the plugin.

sos/report/plugins/mellanox_firmware.py
```diff
@@ -0,0 +1,128 @@
...
+        co = {'cmd': 'which flint', 'output': '/usr/bin/flint'}
+        mstflint_pred = SoSPredicate(self, packages=['mstflint'],
+                                     cmd_outputs=co,
+                                     required={'cmd_outputs': 'none'})
+        mstflint_mlxfw_pred = SoSPredicate(self,
+                                           packages=['mstflint'],
+                                           kmods=['mlx5_core', 'mlxfw'],
+                                           cmd_outputs=co,
+                                           required={'kmods': 'all',
+                                                     'cmd_outputs': 'none'})
```
TurboTurtle, line 65: The way this is currently written, we will end up running `which flint` for every command that uses the predicate.

I think a better approach here would be to  put all the commands that share the same predicate in a list in a single `add_cmd_output()`. The other option would be with a `Plugin.test_predicate()` as a top-level conditional, rather than passing the predicate to each `add_cmd_output()` call.

sos/report/plugins/mellanox_firmware.py
TurboTurtle: You could potentially run a check at the beginning of `setup()` which determines which package is installed, and if both are installed then simply choose one as a default, and build a set of commands based on that result.

naserdean:
@TurboTurtle thank you for the review and suggestions, I will try to revisit this PR tomorrow

sos/report/plugins/mellanox_firmware.py
naserdean: Just to make sure we are on the same page.  I do use it as an enablement trigger, enable plugin if and only if Mellanox network adapters are available.
I'm unsure if that's reasonable or not.

naserdean:
I revisited the approach taken on how to mine for the packages and when to enable the plugin.

Thanks again for the guidance and reviews @TurboTurtle

sos/report/plugins/mellanox_firmware.py
```diff
@@ -0,0 +1,140 @@
...
+    def check_enabled(self):
+        """
+        Checks if this plugin should be executed at all.
+        We will only enable the plugin if there is a
+        Mellanox Technologies network adapter
+        """
+        lspci = self.exec_cmd("lspci -D -d 15b3::0200")
+        if lspci['status'] != 0:
+            return False
+        return self.MLNX_STRING in lspci['output']
```
pmoravec, line 35: Can't this be shortened to simple:

    return lspci['status'] == 0 and self.MLNX_STRING in lspci['output']

?

sos/report/plugins/mellanox_firmware.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def collect(self):
+        if not self.get_option('allow_system_changes'):
+            return
```
pmoravec, line 39: Maybe worth printing a log like:

    self._log_info(f"Skipping to collect plugin {self.plugin_name} as it requires --allow-system-changes option.")

?

sos/report/plugins/mellanox_firmware.py
```diff
@@ -0,0 +1,140 @@
...
+            the commands do not support position independent arguments
+            """
+            commands = [
+                ["mstconfig -d ", " -e q"],
+                ["mstflint -d ", " dc"],
+                ["mstflint -d ", " q"],
+                ["mstreg -d ", " --reg_name ROCE_ACCL --get"],
+                ["mstlink -d ", ""],
+            ]
+        else:
```
pmoravec, line 112: This is confusing / contradicting with "if `flint --version` returns nonzero status, return from `collect` method".

sos/report/plugins/mellanox_firmware.py
naserdean: Makes sense. Thank you for pointing it out.

sos/report/plugins/mellanox_firmware.py
```diff
@@ -0,0 +1,140 @@
...
+        Checks if this plugin should be executed at all.
+        We will only enable the plugin if there is a
+        Mellanox Technologies network adapter
+        """
+        lspci = self.exec_cmd("lspci -D -d 15b3::0200")
+        if lspci['status'] != 0:
+            return False
+        return self.MLNX_STRING in lspci['output']
+
+    def collect(self):
```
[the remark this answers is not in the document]
naserdean: It was suggested here: https://github.com/sosreport/sos/pull/3407#discussion_r1396383691

sos/report/plugins/mellanox_firmware.py
naserdean: In collect we bail out if we had a nonzero status because mft is not installed.

This else branch is on zero status and we run the some commands since mft is installed.

Maybe it would be less confusing if check [here](https://github.com/sosreport/sos/pull/3407/files#diff-b3d320e633b5fb0c0545076c63fe766dd33650c2b7b6c911533b55194a8de55dR100) for zero status?

sos/report/plugins/mellanox_firmware.py
```diff
@@ -0,0 +1,141 @@
...
+        Mellanox Technologies network adapter
+        """
+        lspci = self.exec_cmd("lspci -D -d 15b3::0200")
+        return lspci['status'] == 0 and self.MLNX_STRING in lspci['output']
+
+    def collect(self):
+        if not self.get_option('allow_system_changes'):
+            self._log_info("Skipping to collect plugin "
+                           f"{self.plugin_name} as it requires "
+                           "--allow-system-changes option.")
```
TurboTurtle, line 39: Let's reword this, as it makes it sound like the plugin does nothing without this option.

Perhaps something like `Skipping mst/mlx cable commands as system changes would be made. Use --allow-system-changes to enable this collection`?

sos/report/plugins/mellanox_firmware.py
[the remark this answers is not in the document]
TurboTurtle: Yeah, since we're modifying system state, I think it best if we do this during `collect()`, and that requires using `collect_cmd_output()` for the just-in-time writing to the plugin dir.

sos/report/plugins/mellanox_firmware.py
```diff
@@ -0,0 +1,141 @@
...
+
+        """
+        Run only if mft package is installed.
+        flint is available from the mft package.
+        """
+        co = self.exec_cmd('flint --version')
+        if co['status'] != 0:
+            return
+
+        co = self.collect_cmd_output('mst start')
```
TurboTurtle, line 50: Is there expected output from this and the `mst stop` command?

sos/report/plugins/mellanox_firmware.py
```diff
@@ -0,0 +1,141 @@
...
+                ["flint -d ", " dc"],
+                ["flint -d ", " q"],
+                ["mlxreg -d ", " --reg_name ROCE_ACCL --get"],
+                ["mlxlink -d ", ""],
+                ["fwtrace -d ", " -i all --tracer_mode FIFO"],
+            ]
+        for device in devices:
+            for command in commands:
+                self.add_cmd_output(f"{command[0]} {device} "
+                                    f"{command[1]}", timeout=30)
```
TurboTurtle, line 130: I'm trying to think of a cleaner way to structure this, but not entirely sure doing so buys us much. Let me think on this a bit, or if @pmoravec has some ideas, but if I/you/we can't come up with an easier-to-grok in a few days then I'm fine with this.

sos/report/plugins/mellanox_firmware.py
naserdean: Thank you for the suggestion. I folded in the changes.

sos/report/plugins/mellanox_firmware.py
naserdean: Ack. To be fair this looks a bit better than before...
Because the commands accept the device parameter only in that order, it makes things harder than it should be.

sos/report/plugins/mellanox_firmware.py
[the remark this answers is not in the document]
pmoravec: Ah I see, it makes sense (and I should read whole PR discussion before reviewing).

Maybe then it is worth having a comment abou the unusual approach in the code?

sos/report/plugins/mellanox_firmware.py
```diff
@@ -0,0 +1,141 @@
...
+        devices = []
+        device_list = self.collect_cmd_output('lspci -D -d 15b3::0200')
+        """
+        Will return a string of the following format:
+        0000:08:00.0 Ethernet controller: Mellanox Technologies MT2892
+        Family
+        """
+        if device_list['status'] != 0:
+            # bail out if there no Mellanox PCI devices
+            return
```
pmoravec, line 80: This is required over the `check_enabled` method in cases when user manually enables the plugin, ACK.

sos/report/plugins/mellanox_firmware.py
pmoravec: Ah I understand, depending on the (non)zero status, you want to (not) collect different stuff (and in different methods, very strictly saying).

sos/report/plugins/mellanox_firmware.py
pmoravec: Yeah I spent a while during my first review here, but havent figured out a better code.

pmoravec (approved):
Generally ACK, maybe it is worth commenting the usage of `collect` method.

TurboTurtle (approved):

PR #3409, opened 2023-11-07
[debian] Packaging and testing for debian pkgs

arif-ali:
Initial start of building and testing the deb package

This will improve the overall testing of sos over the Ubuntu releases, and also helps with the basis of then adding further stagetwo testing in the future. This will help to pick up issues automatically rather than the manual testing.

---


arif-ali:
@pmoravec @TurboTurtle can you guys retry the CI on the failed one please.

If all these pass now, this should be ready for review

debian/changelog
TurboTurtle, line 1: I don't think we need to carry this wholesale changelog, as these wouldn't have been built in the repo. The changelog in our carried `.spec` file for RPM builds was built up over the course of builds/releases done within the repository.

We should ideally be starting fresh here, imo.

debian/copyright
```diff
@@ -0,0 +1,51 @@
+Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: sosreport
+Upstream-Contact: Bryn M. Reeves <bmr@redhat.com>
```
TurboTurtle, line 3: This should probably be me, if anyone specific. Feel free to put my personal email here (see GH profile).

debian/copyright
```diff
@@ -0,0 +1,51 @@
...
+	   2007-2013 Red Hat, Inc.
+License: GPL-2+
+
+Files: sos/report/plugins/kernelrt.py
+Copyright: 2012 Red Hat, Inc.
+License: GPL-2
+
+Files: debian/*
+Copyright: 2013 Adam Stokes <adam.stokes@ubuntu.com>
+License: GPL-2+
```
TurboTurtle, line 17: Hmm.

The copyright assignment is generally done on a per-file basis within sos, and I don't want to leave a space for confusion here. I know, I know, GPL and all that, but this is one of those areas I have the mindset to be overly pedantic.

Is there a reason for specifically calling out `kernelrt.py` here?

TurboTurtle (commented):
Happy to see this come into upstream!

Couple notes below, but nothing major.

With regards to the GCE image changes, I'm fine with making those but I'd like to see it called out in the commit message for tracking purposes; as I would normally not expect to see CI image updates in a "start doing builds for this type of package manager" commit.

debian/copyright
arif-ali: updated

debian/copyright
arif-ali: Good points, we just grabbed the stuff that was coming from the original packaging, and hasn't been reviewed for a while.

I have updated these now, to be more generic

debian/changelog
nkshirsagar: So it's fine if the changelog resets to fresh..the idea though, to keep the changelog in sync with what the debian release (at least Mantic) was using, was so that in the future it may have been possible to use the same built and tested package to push to the repositories for ubuntu/debian, which would have saved us time and effort redoing the builds for the ubuntu mantic releases. If the changelog starts afresh, we lose that possible time saving option. 

In any case, it may be moot, looking into the specifics of the process of the debian packages builds for ubuntu, so it's not a great loss if we reset the changelog to start afresh.

arif-ali: The changelog has been updated as per our discussions here, any issues give me a shout

arif-ali:
After discussions, this is ready for review. Thanks for all the input from @TurboTurtle and @nkshirsagar

TurboTurtle (approved):
LGTM.

I'll note that if it turns out that the changelog is the only thing breaking the  streamlining of downstreams consuming these built packages, then no objection from me to unblock that.

.cirrus.yml
```diff
@@ -182,7 +220,9 @@ report_stageone_task:
             dnf -y install python3-pip ethtool
             dnf -y install ./sos_${BUILD_NAME}.rpm
         fi
-        pip3 install avocado-framework==94.0
+        PIP_EXTRA=""
+        [[ $(pip3 install --help | grep break-system) ]] && PIP_EXTRA="--break-system-packages"
```
pmoravec, line 224: This is redundant to add on line 133 since Ubuntu's `pip` always has `--break-system-packages` option, I guess?

arif-ali: I have tested all this, and it's not the case. 18.04, 20.04 and 22.04 don't have this option, and therefore would error out. This is now an option only in the new python3, which is in 23.10 that we're doing the test for the deb package. I haven't fully tested which pip3 version this was introduced

pmoravec: I meant it in opposite way :)  Does it make sense to add the `PIP_EXTRA` trick to line 133 as well?

arif-ali: Ah, I get it

We are only doing the deb testing on the latest Ubuntu release, as that is where we'll see problems. So in this scenario we don't need it as we'll always need that option there with newer releases.

I'm sure we'll see the same problem when rhel 10 and newer Fedora comes out too

pmoravec (approved):

PR #3413, opened 2023-11-10
[ebpf] Check for bpftool presence before running plugin

jcastill:
The plugin was running even when the bpftool was
not present, throwing an exception when it tried to parse a json output:

INFO: [plugin:ebpf] Could not parse bpftool prog list as
JSON: Expecting value: line 1 column 1 (char 0)

It now checks if the program is present before running any command at all.

Related: RH SUPDEV-151

---

### DIFF
--- a/sos/report/plugins/ebpf.py
+++ b/sos/report/plugins/ebpf.py
@@ -15,6 +15,7 @@ class Ebpf(Plugin, IndependentPlugin):
     short_desc = 'eBPF tool'
     plugin_name = 'ebpf'
     profiles = ('system', 'kernel', 'network')
+    commands = ('bpftool',)
 
     option_list = [
         PluginOpt("namespaces", default=None, val_type=int,


sos/report/plugins/ebpf.py
```diff
@@ -15,6 +15,7 @@ class Ebpf(Plugin, IndependentPlugin):
     short_desc = 'eBPF tool'
     plugin_name = 'ebpf'
     profiles = ('system', 'kernel', 'network')
+    commands = ('bpftool',)
```
pmoravec, line 18: `commands` mean the plugin will be enabled any time some command from the tuple is present. So the line will be a new independent way how the plugin will be enabled by default.

If your aim is to enable the plugin automatically *only* when the binary is present, I think you need to test it in overwritten `check_enabled` method..?

TurboTurtle:
Sorry this one definitely slipped through the cracks over the last few months.

> commands mean the plugin will be enabled any time some command from the tuple is present. So the line will be a new independent way how the plugin will be enabled by default.

If `commands` is the _only_ enablement trigger defined, then it will only enable the plugin when the command is present. The `commands` check resolving `False` if `bpftool` is not installed will cause `_check_plugin_triggers()` to return `False`. If no triggers are defined, like what is currently written, then it enables all the time which is not desired here.

@jcastill sorry for the churn here, but the original change was correct - just add the `commands` tuple. The `check_enabled()` override is unnecessary.

pmoravec:
Good point, my mistake. I hadn't realized the `commands` is the first enablement trigger to be added.

jcastill:
@TurboTurtle @pmoravec it's ok :)  I'll revert this change and push again in a bit.

pmoravec (approved):

arif-ali (approved):

TurboTurtle (approved):

PR #3414, opened 2023-11-13
Add support for RHOSP 18 Openstack EDPM deployment

rebtoor:
EDPM (External DataPlane Management) will be the new way to deploy compute nodes startin from RHOSP 18 which will replace TripleO/Director.

- Added `openstack_edpm` plugin
- Added custom path for config and logs in frr/libvirt/logrotate/openstack_neutron/openstack_nova/ovn_host modules
- Tagged libvirt/openstack_neutron/openstack_nova/ovn_host modules as `openstack_edpm_compute`

---

### DIFF
--- a/sos/report/plugins/libvirt.py
+++ b/sos/report/plugins/libvirt.py
@@ -15,7 +15,7 @@ class Libvirt(Plugin, IndependentPlugin):
     short_desc = 'libvirt virtualization API'
 
     plugin_name = 'libvirt'
-    profiles = ('system', 'virt')
+    profiles = ('system', 'virt', 'openstack_edpm')
 
     def setup(self):
         libvirt_keytab = "/etc/libvirt/krb5.tab"
@@ -44,6 +44,8 @@ class Libvirt(Plugin, IndependentPlugin):
             "/etc/libvirt/virtlockd.conf",
             "/var/lib/libvirt/dnsmasq/*",
             "/var/lib/libvirt/qemu/snapshot/*/*.xml",
+            "/var/lib/openstack/config/libvirt",
+            "/var/lib/openstack/containers/libvirt*.json",
         ])
 
         if not self.get_option("all_logs"):
@@ -58,10 +60,15 @@ class Libvirt(Plugin, IndependentPlugin):
                 "/var/log/containers/libvirt/lxc/*.log",
                 "/var/log/containers/libvirt/swtpm/libvirt/qemu/*.log",
                 "/var/log/containers/libvirt/uml/*.log",
+                "/var/log/containers/qemu/*.log",
+                "/var/log/containers/libvirt/*.log",
             ])
         else:
-            self.add_copy_spec("/var/log/libvirt")
-            self.add_copy_spec("/var/log/containers/libvirt")
+            self.add_copy_spec([
+                "/var/log/libvirt",
+                "/var/log/containers/qemu/",
+                "/var/log/containers/libvirt/",
+            ])
 
         if self.path_exists(self.path_join(libvirt_keytab)):
             self.add_cmd_output("klist -ket %s" % libvirt_keytab)
--- a/sos/report/plugins/openstack_edpm.py
+++ b/sos/report/plugins/openstack_edpm.py
@@ -1,0 +1,38 @@
+# Copyright (C) 2023 Red Hat, Inc., Roberto Alfieri <ralfieri@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, RedHatPlugin
+
+
+class OpenStackEDPM(Plugin, RedHatPlugin):
+
+    short_desc = 'Installation information from OpenStack EDPM deployment'
+
+    plugin_name = 'openstack_edpm'
+    profiles = ('openstack', 'openstack_edpm')
+    services = ('edpm-container-shutdown')
+
+    def setup(self):
+        # Notes: recursion is max 2 for edpm-config
+        # Those directories are present on all OpenStack nodes
+        self.edpm_log_paths = [
+            '/var/lib/edpm-config/'
+        ]
+        self.add_copy_spec(self.edpm_log_paths)
+
+    def postproc(self):
+        # Ensures we do not leak passwords from the edpm related locations
+        # Other locations don't have sensitive data.
+        regexp = r'(".*(key|password|pass|secret|database_connection))' \
+                 r'([":\s]+)(.*[^"])([",]+)'
+        for path in self.edpm_log_paths:
+            self.do_path_regex_sub(path, regexp, r'\1\3*********\5')
+
+# vim: set et ts=4 sw=4 :
--- a/sos/report/plugins/openstack_nova.py
+++ b/sos/report/plugins/openstack_nova.py
@@ -22,7 +22,8 @@ class OpenStackNova(Plugin):
 
     short_desc = 'OpenStack Nova'
     plugin_name = "openstack_nova"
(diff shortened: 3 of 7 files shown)


sos/report/plugins/libvirt.py
```diff
@@ -6,16 +6,16 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.report.plugins import Plugin, IndependentPlugin
+from sos.report.plugins import Plugin, RedHatPlugin
```
arif-ali, line 9: NAK, we need this for UbuntuPlugin too, as we also need this from a Ubuntu perspective. Maybe, worth adding UbuntuLibvirt as a function

arif-ali (changes requested):

sos/report/plugins/libvirt.py
rebtoor: ack!

sos/report/plugins/libvirt.py
```diff
@@ -6,16 +6,16 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.report.plugins import Plugin, IndependentPlugin
+from sos.report.plugins import Plugin, DebianPlugin, RedHatPlugin, UbuntuPlugin
 import glob
 
 
-class Libvirt(Plugin, IndependentPlugin):
+class Libvirt(Plugin, DebianPlugin, RedHatPlugin, UbuntuPlugin):
```
arif-ali, line 13: You don't need `RedHatPlugin` here, as the functions in this class will get triggered via `RedHatLibvirt` as the `RedHatPlugin` is being used there

arif-ali (changes requested):

sos/report/plugins/libvirt.py
rebtoor: ack, i took it as example another [plugin](https://github.com/sosreport/sos/blob/main/sos/report/plugins/samba.py), but I guess you're right.

sos/report/plugins/openstack_nova.py
```diff
@@ -215,15 +216,46 @@ def setup(self):
...
+        def postproc(self):
+            protect_keys = [
+                "ldap_dns_password", "neutron_admin_password",
+                "rabbit_password", "qpid_password", "powervm_mgr_passwd",
+                "virtual_power_host_pass", "xenapi_connection_password",
+                "password", "host_password", "vnc_password",
+                "admin_password", "connection_password",
+                "memcache_secret_key", "s3_secret_key",
+                "metadata_proxy_shared_secret", "fixed_key", "transport_url",
+                "rbd_secret_uuid",
```
arif-ali, line 248: you could add the `rbd_secret_uuid` above instead. As it's a regex, it should just work

arif-ali (changes requested):

sos/report/plugins/libvirt.py
```diff
@@ -89,4 +89,25 @@ def postproc(self):
...
+        if not self.get_option("all_logs"):
+            self.add_copy_spec([
+                "/var/log/containers/qemu/*.log",
+                "/var/log/containers/libvirt/*.log",
+            ])
+        else:
+            self.add_copy_spec([
+                "/var/log/containers/qemu/",
+                "/var/log/containers/libvirt/",
+            ])
```
TurboTurtle, line 111: I don't think we need to split this out.

The `/var/lib/openstack` paths should, arguably, be collected by one of the openstack plugins. However, assuming we keep the collection in the `libvirt` plugin, we can simply add these to the existing collection lists. If they don't exist on a particular system, we skip them gracefully.

Further, the `/var/log/containers` paths fit nicely in with what is already called in the existing `IndependentPlugin` structure.

sos/report/plugins/openstack_edpm.py
```diff
@@ -0,0 +1,37 @@
...
+from sos.report.plugins import Plugin, RedHatPlugin
+
+
+class OpenStackEDPM(Plugin, RedHatPlugin):
+
+    short_desc = 'Installation information from OpenStack EDPM deployment'
+
+    plugin_name = 'openstack_edpm'
+    profiles = ('openstack', 'openstack_edpm')
+
```
TurboTurtle, line 20: We should have at least one of `packages`, `files`, `commands`, `containers`, or `services` defined here to gate plugin enablement. Otherwise this will be enabled on every `sos report` execution.

sos/report/plugins/libvirt.py
rebtoor: I thought it was redundant to create an `openstack_libvirt` plugin since i would basically be copying most of this one, but i agree that we can avoid to split the logic. Let me fix it.

TurboTurtle (approved):
LGTM

fmount (commented):
lgtm, thanks @rebtoor

sos/report/plugins/openstack_nova.py
```diff
@@ -215,15 +216,46 @@ def setup(self):
...
             self.add_copy_spec([
                 "/var/log/httpd/placement*.log",
+                "/var/log/containers/nova/*.log"
             ])
 
+        self.add_forbidden_path([
+            "/var/lib/openstack/config/nova/ssh-privatekey"
+        ])
+
+        def postproc(self):
```
TurboTurtle, line 239: Ah, my bad here. Overlooked this, but the code scanning is flagging this because `postproc` is overindented and is part of `setup()` as written, where instead it should be just another method with the class.

TurboTurtle (changes requested):

sos/report/plugins/openstack_nova.py
rebtoor: Ack!

TurboTurtle (approved):

sos/report/plugins/openstack_nova.py
```diff
@@ -215,15 +216,46 @@ def setup(self):
...
+    def postproc(self):
+        protect_keys = [
+            "ldap_dns_password", "neutron_admin_password",
+            "rabbit_password", "qpid_password", "powervm_mgr_passwd",
+            "virtual_power_host_pass", "xenapi_connection_password",
+            "password", "host_password", "vnc_password",
+            "admin_password", "connection_password",
+            "memcache_secret_key", "s3_secret_key",
+            "metadata_proxy_shared_secret", "fixed_key", "transport_url",
+            "rbd_secret_uuid",
```
arif-ali, line 248: could we just not add `rbd_secret_uuid` to the main `postproc()` definition? that way it would work for both Ubuntu and RedHat? Otherwise the whole `protect_keys` list looks identical

rebtoor: sure! that makes a lot of sense.

Thanks @arif-ali

arif-ali (approved):

TurboTurtle:
@rebtoor this looks fine now, but it needs a rebase on current `main` due to a recent libvirt change. If you could do that and re-push this branch (you'll need to `--force` push it), then we can merge this.

rebtoor:
> @rebtoor this looks fine now, but it needs a rebase on current `main` due to a recent libvirt change. If you could do that and re-push this branch (you'll need to `--force` push it), then we can merge this.

Ack! I'll do it ASAP.

Thanks a lot!

rebtoor:
@TurboTurtle is there something I can do to unlock packit builds?

fmount:
LGTM, thanks @rebtoor

fmount:
@TurboTurtle is there an existing policy to get new tags/releases and builds? I'm wondering if we have a timeline to get this patch into a package (mostly commenting for CentOS).

pmoravec:
> @TurboTurtle is there an existing policy to get new tags/releases and builds? I'm wondering if we have a timeline to get this patch into a package (mostly commenting for CentOS).

A new sos upstream version (typically 4.Y) is released every 6 months, see https://github.com/sosreport/sos/releases .

Beyond that, tags (like 4.5.6) were created every 4 weeks to allow such frequent release policy from Red Hat. We / from Red Hat have paused this in past few months due to several independent reasons, but plan to start tagging upstream the same way soon.

PR #3417, opened 2023-11-16
[policies] Add policy for TencentOS

ZoeDong:
---

### DIFF
--- a/sos/policies/distros/tencentos.py
+++ b/sos/policies/distros/tencentos.py
@@ -1,0 +1,45 @@
+# Copyright (c) 2023 Tencent., ZoeDong <zoedong@tencent.com>
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.policies.distros.redhat import RedHatPolicy, OS_RELEASE
+import os
+
+
+class TencentOSPolicy(RedHatPolicy):
+    distro = "TencentOS Server"
+    vendor = "TencentOS"
+    vendor_urls = [
+        ('Distribution Website',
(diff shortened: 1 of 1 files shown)


jcastill (approved):

arif-ali (approved):

TurboTurtle:
I'm going to raise the [same concerns I have raised previously](https://github.com/sosreport/sos/pull/3296#issuecomment-1669616400) about the recent trend of adding policy support for RHEL clones and then nothing else being done with them, and the distro in question seemingly going silent not too long after.

arif-ali:
> I'm going to raise the [same concerns I have raised previously](https://github.com/sosreport/sos/pull/3296#issuecomment-1669616400) about the recent trend of adding policy support for RHEL clones and then nothing else being done with them, and the distro in question seemingly going silent not too long after.

Ah, yes, we discussed on IRC on this too; most of the policies are extending `RedHatPolicy` other then `ubuntu`, `suse` and `debian`. and they have limited input other than branding, like you suggested in the above comment.

Does that mean, we want to de-brand, and get rid of some of the policies?

TurboTurtle:
> Does that mean, we want to de-brand, and get rid of some of the policies?

Not so much that, but moreso that I think we should be a bit more particular about what policies get added - what's already in the tree is already there however.

So, for instance if a new distro wants to add an sos policy and has a new/different package manager (or combination of package managers) then that makes sense to me. Or if there's a new distro with professional support offerings. Or one that is trying to carve out a name in the OSS community. Or a distro needs specific changes to collections for certain plugins (i.e. a new tagging class is needed). Etc, etc..

What I don't want to encourage though is collecting a bunch of policies that are all clones of $some_other_distro and the only the difference is the branding and the website printed in our disclaimer. Especially if the plugins have no difference between these distros (e.g. no need for different tagging classes), or if we'll never see any further contributions to upstream from them because there's only a single entity using such an in-house distribution.

ZoeDong:
My original intention is to provide basic distro information when using sos. We will consider providing more support. Before that, I will close this PR. Thanks for your reply.

PR #3418, opened 2023-11-16
[canonical-livepatch-onprem] Add Canonical Livepatch on-prem server support

adam-vest:

### DIFF
--- a/sos/report/plugins/canonical_livepatch_onprem.py
+++ b/sos/report/plugins/canonical_livepatch_onprem.py
@@ -1,0 +1,42 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, UbuntuPlugin
+
+
+class CanonicaLivepatchOnprem(Plugin, UbuntuPlugin):
+
+    short_desc = 'Canonical Livepatch Onprem Service'
+
+    plugin_name = 'canonical_livepatch_onprem'
+    profiles = ('services',)
+    services = ("livepatch-server",)
+
+    def setup(self):
+        self.add_copy_spec([
+            "/etc/livepatchd.yaml",
+        ])
+
+    def postproc(self):
+        onprem_conf = "/etc/livepatchd.yaml"
+        protect_keys = [
+            "username",
+            "password",
+            "token",
+            "connection_string",
+        ]
+
+        # Redact simple yaml style "key: value".
+        keys_regex = r"(^(-|\s)*(%s)\s*:\s*)(.*)" % "|".join(protect_keys)
+        sub_regex = r"\1*********"
+        self.do_path_regex_sub(onprem_conf, keys_regex, sub_regex)
+
+        # Redact conf
+        self.do_file_private_sub(onprem_conf)
+
+# vim: set et ts=4 sw=4 :


sos/report/plugins/canonical_livepatch_onprem.py
```diff
@@ -0,0 +1,29 @@
...
+
+    short_desc = 'Canonical Livepatch Onprem Service'
+
+    plugin_name = 'canonical_livepatch_onprem'
+    profiles = ('services',)
+    files = ('/usr/local/bin/livepatchd',)
+
+    def setup(self):
+        self.add_copy_spec([
+            "/etc/livepatchd.yaml",
```
jcastill, line 22: lgtm, just a question: is there an api token or a user/pass we should obfuscate in this file?

arif-ali:
if you added `services = ("livepatch-server")` for the trigger at the top instead of `files`, then both the service status and the journal would be added by default, and then you will not need the 2 status checks and `files`

sos/report/plugins/canonical_livepatch_onprem.py
adam-vest: There are both, actually - user credentials and a token that authenticates against upstream livepatch.. User password is already encoded, but perhaps not a bad idea to obfuscate. Is that something we'd do in the plugin directly, or elsewhere in sos?

jcastill: You can do it directly in the plugin itself. Here's an example of obfuscation in a yaml file:

https://github.com/sosreport/sos/blob/main/sos/report/plugins/juju.py#L65

adam-vest:
> if you added `services = ("livepatch-server")` for the trigger at the top instead of `files`, then both the service status and the journal would be added by default, and then you will not need the 2 status checks and `files`

Hey @arif-ali , I tried doing the following as (I think you) suggested:

```
    services = ("livepatch-server")

    def setup(self):
        self.add_copy_spec([
            "/etc/livepatchd.yaml",
        ])
```

...but I didn't see the journal or service status in the resultant sos report? Normally these were appearing in `sos_commands/canonical_livepatch_server`. Is this new as of 4.6.0? I'm just testing against the version in apt, which is `4.5.6`.

sos/report/plugins/canonical_livepatch_onprem.py
adam-vest: Awesome, thanks for the link! I have some obfuscation code in now that looks like it's having the intended effect :slightly_smiling_face:

TurboTurtle:
> > if you added `services = ("livepatch-server")` for the trigger at the top instead of `files`, then both the service status and the journal would be added by default, and then you will not need the 2 status checks and `files`
> 
> Hey @arif-ali , I tried doing the following as (I think you) suggested:
> 
> ```
>     services = ("livepatch-server")
> 
>     def setup(self):
>         self.add_copy_spec([
>             "/etc/livepatchd.yaml",
>         ])
> ```
> 
> ...but I didn't see the journal or service status in the resultant sos report? Normally these were appearing in `sos_commands/canonical_livepatch_server`. Is this new as of 4.6.0? I'm just testing against the version in apt, which is `4.5.6`.

You're missing a comma in `services` to make it a tuple, rather than a string. `services = ('livepatch-server',)` The implicit collection for `services` was definitely in 4.5, but I forget exactly when we added it.

adam-vest:
> > > if you added `services = ("livepatch-server")` for the trigger at the top instead of `files`, then both the service status and the journal would be added by default, and then you will not need the 2 status checks and `files`
> > 
> > 
> > Hey @arif-ali , I tried doing the following as (I think you) suggested:
> > ```
> >     services = ("livepatch-server")
> > 
> >     def setup(self):
> >         self.add_copy_spec([
> >             "/etc/livepatchd.yaml",
> >         ])
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ...but I didn't see the journal or service status in the resultant sos report? Normally these were appearing in `sos_commands/canonical_livepatch_server`. Is this new as of 4.6.0? I'm just testing against the version in apt, which is `4.5.6`.
> 
> You're missing a comma in `services` to make it a tuple, rather than a string. `services = ('livepatch-server',)` The implicit collection for `services` was definitely in 4.5, but I forget exactly when we added it.

Awesome, got it! Thanks for the feedback - pushed an update to implement these things. Looking forward to any more feedback!

TurboTurtle (approved):
LGTM

arif-ali (approved):

PR #3420, opened 2023-11-18
[deb][pkg] Update control for downstream build

arif-ali:
* python3-setuptools is still required for building the package
* python3-yaml is required for build for tests with the conformance_tests
* Add the same ignores as dowstream build, so that we don't have to keep changing these. Once the avocado is packaged, we can take this out
* Update image for the deb creation, so that it can be minimal
* Fix lxd runtime, so that if the root device is not available, it will not look at the storage pools

Downstream build machines are much more minimal compared to gce instances, so we require more dependencies to be added.

---

### DIFF
--- a/sos/policies/runtimes/lxd.py
+++ b/sos/policies/runtimes/lxd.py
@@ -94,7 +94,8 @@ class LxdContainerRuntime(ContainerRuntime):
             if out['status'] == 0:
                 out_json = json.loads(out['output'])
                 for profile in out_json:
-                    if profile['name'] == 'default':
+                    if (profile['name'] == 'default' and
+                            'root' in profile['devices']):
                         stg_pool = profile['devices']['root']['pool']
                         break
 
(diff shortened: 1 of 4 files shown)


sos/policies/runtimes/lxd.py
```diff
@@ -94,7 +94,8 @@ def get_volumes(self):
             if out['status'] == 0:
                 out_json = json.loads(out['output'])
                 for profile in out_json:
-                    if profile['name'] == 'default':
+                    if (profile['name'] == 'default' and
+                            'root' in profile['devices']):
```
arif-ali, line 98: This had to be added as part of the fix of bringing the build of the deb package build on a minimal OS installation. But this also fixes an issue if the LXD was never initialised and lxd package is actually installed.

TurboTurtle (approved):

pmoravec (approved):

Issue #3430, opened 2023-11-30
Adding preset leaves its temp directory


pmoravec:
Reproducer:

    rm -rf /var/tmp/sos*
    sos report --add-preset mypreset2 --batch -o qpid
    ls -l /var/tmp/ | grep sos

It shows a temporary directory like `sos.sedy69wb` with two tmp files for `sos.log` and `ui.log`. We should remove the directory at the end..

PR #3431, opened 2023-11-30
[report] Clean tmpdir when adding or deleting preset

pmoravec:
When adding or deleting preset (and also when detecting no plugin leaves enabled), sos.report.execute must cleanup its working dir.

Since we don't use return value of the execute method, we can skip the return command.

Resolves: #3430
Closes: #3431

---

### DIFF
--- a/sos/report/__init__.py
+++ b/sos/report/__init__.py
@@ -1809,12 +1809,14 @@ class SoSReport(SoSComponent):
                 self.list_presets()
                 raise SystemExit
             if self.opts.add_preset:
-                return self.add_preset(self.opts.add_preset)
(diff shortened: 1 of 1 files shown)


pmoravec:
We theoretically might remove returning True/False values from `add_preset` and `del_preset` methods as we dont use them anywhere. But it makes sense to return the method there and passing meaningfull (though unused ATM) value can make sense (if we would decide to re-use it).

jcastill (approved):
lgtm

arif-ali (approved):

TurboTurtle (approved):

PR #3434, opened 2023-12-01
[arcconf] add getconfig and GETLOGS commands for all available controllers

sourabhjains:
Currently, the arcconf getconfig and GETLOGS commands are added with the assumption that the system has only one controller with ID 1. It is possible that a system can have multiple controllers, so commands need to be added to extract information for all available controllers.

The arcconf utility does not have a dedicated command to get the list of controllers available in the system. Therefore, parse the 'arcconf list' output to obtain the controller ID for all controllers in the system, and add getconfig and GETLOGS commands for each available controllers.

---

### DIFF
--- a/sos/report/plugins/arcconf.py
+++ b/sos/report/plugins/arcconf.py
@@ -10,6 +10,8 @@
 # This sosreport plugin is meant for sas adapters.
 # This plugin logs inforamtion on each adapter it finds.
 
+import re
+
 from sos.report.plugins import Plugin, IndependentPlugin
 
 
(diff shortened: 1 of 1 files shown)


arif-ali (changes requested):
a couple of minor notes, lgtm otherwise

sos/report/plugins/arcconf.py
```diff
@@ -21,11 +21,30 @@ class arcconf(Plugin, IndependentPlugin):
...
-            "arcconf GETLOGS 1 UART"
-        ])
+        # Get the list of available adapters
+        listarcconf = self.collect_cmd_output("arcconf list")
+
+        # Parse the 'arcconf list' output and extract controller IDs.
+        # For each Controller ID found in 'arcconf list', add commands
+        # for getconfig and GETLOGS
+        if listarcconf['status'] == 0:
+            for line in listarcconf['output'].splitlines():
```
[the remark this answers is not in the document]
sourabhjains: Yes I do. I will add it as a comment.

Thanks for the quick response.

sos/report/plugins/arcconf.py
```diff
@@ -21,11 +21,41 @@ class arcconf(Plugin, IndependentPlugin):
...
+                words = line.split()
+
+                # Skip lines with fewer than two words
+                if len(words) < 2:
+                    continue
+
+                # Line with "Controller XX: *" has Controller ID
+                if words[0] != "Controller" or not words[1].endswith(":"):
+                    continue
```
TurboTurtle, line 51: If we use a regex and matching groups here, we can simplify the line parsing and automatically get the `controller_id` as well.

E.G. (off the top of my head, may need some refining):

```python
try:
    controller_id = re.match(r"(^Controller (\d))", line).groups()[1]
    if controller_id:
        self.add_cmd_output([...])
except Exception:
    continue
```

TurboTurtle (commented):
We can clean up the slice-and-dice of the list output some, but overall ack to the intent.

sos/report/plugins/arcconf.py
```diff
@@ -21,11 +23,34 @@ class arcconf(Plugin, IndependentPlugin):
...
+        if listarcconf['status'] == 0:
+            for line in listarcconf['output'].splitlines():
+                try:
+                    match = re.match(r"^[\s]*Controller (\d)+", line).group(0)
+                    controller_id = match.split()[1]
+                    if controller_id:
+                        # Add new commands with Controller ID
+                        self.add_cmd_output([
+                            f"arcconf getconfig {controller_id}",
+                            f"arcconf GETLOGS UART {controller_id}",
```
TurboTurtle, line 52: The previous invocation of this was `arcconf GETLOGS $id UART` - does the position of the id not matter, or was it wrong previously?

I'm fine with whatever works, just something I noticed and want to double check since I'm not personally familiar with `arcconf`.

TurboTurtle (commented):
Looks good to me overall, one question below.

sos/report/plugins/arcconf.py
sourabhjains: Crucial observation. It does matter actually.

I have some special binary where it does not but in general it does.

Thanks @TurboTurtle , I will send updated patch.

sourabhjains:
@TurboTurtle @arif-ali Updates to the patch are still in progress. I will send the updated patch soon.

sourabhjains:
@arif-ali @TurboTurtle Please review the updated patch.

arif-ali (approved):

TurboTurtle (approved):

sourabhjains:
Thanks for accepting the patch.

PR #3443, opened 2023-12-13
[nvidia] collect nvidia-persistenced service status

sourabhjains:
Having nvidia-persistenced service status helps debugging issues with NVIDIA kernel driver behaviour with persistence mode on.

suggested-by: Borislav Stoymirski <borislav.stoymirski@bg.ibm.com>
Tested-by: Borislav Stoymirski <borislav.stoymirski@bg.ibm.com>

---

### DIFF
--- a/sos/report/plugins/nvidia.py
+++ b/sos/report/plugins/nvidia.py
@@ -30,6 +30,7 @@ class Nvidia(Plugin, IndependentPlugin):
             'nvlink -e'
         ]
 
+        self.add_service_status("nvidia-persistenced")
         self.add_cmd_output(["nvidia-smi %s" % cmd for cmd in subcmds])
 
         query = ('gpu_name,gpu_bus_id,vbios_version,temperature.gpu,'


TurboTurtle (approved):
The change is fine on its own, however I should point out that if we instead have the `services` tuple contain this, we not only get the service status but also the journal (so we can remove the explicit call to `add_journal()`) but also another way to enable the plugin.

Just an option - if we don't want this plugin to be enabled if the `nvidia-persistenced` service is running but `nvidia-smi` is not present (is that even possible?), then no problems from me for adding this commit as-is.

pmoravec (approved):

sourabhjains:
> The change is fine on its own, however I should point out that if we instead have the `services` tuple contain this, we not only get the service status but also the journal (so we can remove the explicit call to `add_journal()`) but also another way to enable the plugin.

> 
> Just an option - if we don't want this plugin to be enabled if the `nvidia-persistenced` service is running but `nvidia-smi` is not present (is that even possible?), then no problems from me for adding this commit as-is.

I am fine with the `nvidia-persistenced` service status being captured only when the `nvidia-smi` command is present.

Thanks for the review.

PR #3445, opened 2023-12-14
[maas] fix working directory for rsyslog configs on snap install

adam-vest:
---

### DIFF
--- a/sos/report/plugins/maas.py
+++ b/sos/report/plugins/maas.py
@@ -98,7 +98,7 @@ class Maas(Plugin, UbuntuPlugin):
                 "/var/snap/maas/current/supervisord",
                 "/var/snap/maas/current/preseeds",
                 "/var/snap/maas/current/proxy",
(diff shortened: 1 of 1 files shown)


arif-ali (approved):

TurboTurtle:
Sanity question - was this _always_ wrong, or has it recently changed? Is there a possible (valid) use case for both and we should try to collect each location, knowing that only one is likely to exist on any given host?

adam-vest:
Heh, I was wondering if this question would pop up. I'm honestly not sure if it's always been wrong, but I checked against all versions that are shipped in the snap store (which goes back to 2.7), and all of those have the directory named `syslog`, and not `rsyslog`. 3.0 is also the oldest version that's supported currently, so while it's certainly possible that this is a legacy naming convention, it would have been on versions that are no longer officially available or supported, going back at least 3 years.

TurboTurtle:
Works for me as-is then.

TurboTurtle (approved):

PR #3446, opened 2023-12-15
[redhat] Update support-tool images version

jcastill:
Update support-tools images version.

Related: RH: RHEL-19587

---

### DIFF
--- a/sos/policies/distros/redhat.py
+++ b/sos/policies/distros/redhat.py
@@ -447,7 +447,7 @@ support representative.
 
     containerzed = True
     container_runtime = 'docker'
-    container_image = 'registry.access.redhat.com/rhel7/support-tools'
+    container_image = 'registry.access.redhat.com/rhel9/support-tools'
     sos_path_strip = '/host'
     container_version_command = 'rpm -q sos'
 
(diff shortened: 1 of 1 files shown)


sos/policies/distros/redhat.py
```diff
@@ -447,7 +447,7 @@ class RedHatAtomicPolicy(RHELPolicy):
 
     containerzed = True
     container_runtime = 'docker'
-    container_image = 'registry.access.redhat.com/rhel7/support-tools'
+    container_image = 'registry.access.redhat.com/rhel9/support-tools'
```
pmoravec, line 450: This is for RHEL Atomic, which afaik runs on RHEL7 only. And running RHEL9 containers on RHEL7 is not supported.

How far does it make sense to keep this RHEL7-only related class / policy, if RHEL7 does not contain `python3` by default and hence we don't plan any `sos-4.*` release there..? There can be some corner case use cases, maybe..?

TurboTurtle:
We can likely drop Atomic support at this point.

As far as the support-tools image version is concerned, using RHEL 8 was a deliberate choice because RHEL 9 containers on RHEL 8 (which is what CoreOS was (is?) built off of) is unsupported, whereas RHEL 8 containers on RHEL 9 is perfectly fine.

jcastill:
Ok, then I'll close this PR and open one for dropping Atomic support. Thank you both!

PR #3447, opened 2023-12-16
[report|collector|policies] Add an S3 protocol upload option

### LINKED ISSUE #3112: --upload-protocol for s3
I'm interested in extending the functions for `--upload-protocol` such that collector and report would both have available:

```
              {{module}}_grp.add_argument("--upload-protocol", default='auto',
                                 choices=['auto', 'https', 'ftp', 'sftp', 's3'],
```

Would an s3 protocol upload option be a welcome addition? If so I will create a branch and a draft PR for further discussion.

Thanks

TrevorBenson:
Closes #3112

---

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -37,6 +37,7 @@ parts:
       - wheel
       - python_magic
       - packaging
+      - boto3==1.26.155
 
 apps:
   sos:
--- a/sos/collector/__init__.py
+++ b/sos/collector/__init__.py
@@ -134,7 +134,13 @@ class SoSCollector(SoSComponent):
         'upload_pass': None,
         'upload_method': 'auto',
         'upload_no_ssl_verify': False,
-        'upload_protocol': 'auto'
+        'upload_protocol': 'auto',
+        'upload_s3_endpoint': None,
+        'upload_s3_region': None,
+        'upload_s3_bucket': None,
+        'upload_s3_access_key': None,
+        'upload_s3_secret_key': None,
+        'upload_s3_object_prefix': None
     }
 
     def __init__(self, parser, parsed_args, cmdline_args):
@@ -440,8 +446,21 @@ class SoSCollector(SoSComponent):
                                  action='store_true',
                                  help="Disable SSL verification for upload url"
                                  )
+        collect_grp.add_argument("--upload-s3-endpoint", default=None,
+                                 help="Endpoint to upload to for S3 bucket")
+        collect_grp.add_argument("--upload-s3-region", default=None,
+                                 help="Region for the S3 bucket")
+        collect_grp.add_argument("--upload-s3-bucket", default=None,
+                                 help="Name of the S3 bucket to upload to")
+        collect_grp.add_argument("--upload-s3-access-key", default=None,
+                                 help="Access key for the S3 bucket")
+        collect_grp.add_argument("--upload-s3-secret-key", default=None,
+                                 help="Secret key for the S3 bucket")
+        collect_grp.add_argument("--upload-s3-object-prefix", default=None,
+                                 help="Prefix for the S3 object/key")
         collect_grp.add_argument("--upload-protocol", default='auto',
-                                 choices=['auto', 'https', 'ftp', 'sftp'],
+                                 choices=['auto', 'https', 'ftp', 'sftp',
+                                          's3'],
                                  help="Manually specify the upload protocol")
 
         # Group the cleaner options together
@@ -1270,7 +1289,8 @@ this utility or remote systems that it connects to.
             msg = 'No sosreports were collected, nothing to archive...'
             self.exit(msg, 1)
 
-        if self.opts.upload or self.opts.upload_url:
+        if (self.opts.upload and self.policy.get_upload_url()) or \
+                self.opts.upload_s3_endpoint:
             try:
                 self.policy.upload_archive(arc_name)
                 self.ui_log.info("Uploaded archive successfully")
--- a/sos/policies/distros/__init__.py
+++ b/sos/policies/distros/__init__.py
@@ -32,6 +32,12 @@ try:
 except ImportError:
     REQUESTS_LOADED = False
 
+try:
+    import boto3
+    BOTO3_LOADED = True
+except ImportError:
+    BOTO3_LOADED = False
+
 # Container environment variables for detecting if we're in a container
 ENV_CONTAINER = 'container'
 ENV_HOST_SYSROOT = 'HOST'
@@ -56,11 +62,23 @@ class LinuxPolicy(Policy):
     _upload_user = None
     _upload_password = None
     _upload_method = None
+    _upload_s3_endpoint = 'https://s3.amazonaws.com'
+    _upload_s3_bucket = None
+    _upload_s3_access_key = None
+    _upload_s3_secret_key = None
+    _upload_s3_region = None
+    _upload_s3_object_prefix = ''
     default_container_runtime = 'docker'
     _preferred_hash_name = None
     upload_url = None
     upload_user = None
     upload_password = None
+    upload_s3_endpoint = None
+    upload_s3_bucket = None
+    upload_s3_access_key = None
+    upload_s3_secret_key = None
+    upload_s3_region = None
+    upload_s3_object_prefix = None
     # collector-focused class attrs
     containerized = False
     container_image = None
@@ -292,6 +310,13 @@ class LinuxPolicy(Policy):
         self.upload_password = cmdline_opts.upload_pass
         self.upload_archive_name = ''
 
+        self.upload_s3_endpoint = cmdline_opts.upload_s3_endpoint
+        self.upload_s3_region = cmdline_opts.upload_s3_region
+        self.upload_s3_access_key = cmdline_opts.upload_s3_access_key
+        self.upload_s3_bucket = cmdline_opts.upload_s3_bucket
+        self.upload_s3_object_prefix = cmdline_opts.upload_s3_object_prefix
+        self.upload_s3_secret_key = cmdline_opts.upload_s3_secret_key
+
         # set or query for case id
         if not cmdline_opts.batch and not \
                 cmdline_opts.quiet:
@@ -314,9 +339,15 @@ class LinuxPolicy(Policy):
                 cmdline_opts.quiet:
             try:
                 # Policies will need to handle the prompts for user information
-                if cmdline_opts.upload and self.get_upload_url():
+                if cmdline_opts.upload and self.get_upload_url() and \
+                        not cmdline_opts.upload_protocol == 's3':
                     self.prompt_for_upload_user()
                     self.prompt_for_upload_password()
+                elif cmdline_opts.upload_protocol == 's3':
+                    self.prompt_for_upload_s3_bucket()
+                    self.prompt_for_upload_s3_endpoint()
+                    self.prompt_for_upload_s3_access_key()
+                    self.prompt_for_upload_s3_secret_key()
                 self.ui_log.info('')
             except KeyboardInterrupt:
                 raise
@@ -353,6 +384,58 @@ class LinuxPolicy(Policy):
             self.soslog.info('Set niceness of report to 19')
         except Exception as err:
             self.soslog.error(f"Error setting report niceness to 19: {err}")
+
+    def prompt_for_upload_s3_access_key(self):
+        """Should be overridden by policies to determine if an access key needs
+        to be provided for upload or not
+        """
+        if not self.get_upload_s3_access_key():
+
+            msg = (
+                "Please provide the upload access key for bucket"
+                f" {self.get_upload_s3_bucket()} via endpoint"
+                f" {self.get_upload_s3_endpoint()}: "
+            )
+            self.upload_s3_access_key = input(_(msg))
+
+    def prompt_for_upload_s3_secret_key(self):
+        """Should be overridden by policies to determine if a secret key needs
+        to be provided for upload or not
+        """
+        if not self.get_upload_s3_secret_key():
+            msg = (
+                "Please provide the upload secret key for bucket"
+                f" {self.get_upload_s3_bucket()} via endpoint"
+                f" {self.get_upload_s3_endpoint()}: "
+            )
+            self.upload_s3_secret_key = getpass(msg)
+
+    def prompt_for_upload_s3_bucket(self):
+        """Should be overridden by policies to determine if a bucket needs to
+        be provided for upload or not
+        """
+        if not self.upload_s3_bucket:
+            if self.upload_url and self.upload_url.startswith('s3://'):
+                self.upload_s3_bucket = self.upload_url[5:]
+            else:
+                user_input = input(_("Please provide the upload bucket: "))
+                self.upload_s3_bucket = user_input.strip('/')
+        return self.upload_s3_bucket
+
+    def prompt_for_upload_s3_endpoint(self):
+        """Should be overridden by policies to determine if an endpoint needs
+        to be provided for upload or not
+        """
+        default_endpoint = self._upload_s3_endpoint
+        if not self.upload_s3_endpoint:
+            msg = (
+                "Please provide the upload endpoint for bucket"
+                f" {self.get_upload_s3_bucket()}"
+                f" (default: {default_endpoint}): "
+            )
+            user_input = input(_(msg))
+            self.upload_s3_endpoint = user_input or default_endpoint
+        return self.upload_s3_endpoint
 
     def prompt_for_upload_user(self):
         """Should be overridden by policies to determine if a user needs to
@@ -438,7 +521,8 @@ class LinuxPolicy(Policy):
         prots = {
             'ftp': self.upload_ftp,
             'sftp': self.upload_sftp,
-            'https': self.upload_https
+            'https': self.upload_https,
+            's3': self.upload_s3
         }
         if self.commons['cmdlineopts'].upload_protocol in prots.keys():
             return prots[self.commons['cmdlineopts'].upload_protocol]
@@ -468,6 +552,73 @@ class LinuxPolicy(Policy):
 
         return requests.auth.HTTPBasicAuth(user, password)
 
+    def get_upload_s3_access_key(self):
+        """Helper function to determine if we should use the policy default
+        upload access key or one provided by the user
+
+        :returns: The access_key to use for upload
+        :rtype: ``str``
+        """
+        return (os.getenv('SOSUPLOADS3ACCESSKEY', None) or
+                self.upload_s3_access_key or
+                self._upload_s3_access_key)
+
+    def get_upload_s3_endpoint(self):
+        """Helper function to determine if we should use the policy default
+        upload endpoint or one provided by the user
+
+        :returns: The S3 Endpoint to use for upload
+        :rtype: ``str``
+        """
+        if not self.upload_s3_endpoint:
+            self.prompt_for_upload_s3_endpoint()
+        return self.upload_s3_endpoint
+
+    def get_upload_s3_region(self):
+        """Helper function to determine if we should use the policy default
+        upload region or one provided by the user
+
+        :returns: The S3 region to use for upload
+        :rtype: ``str``
+        """
+        return self.upload_s3_region or self._upload_s3_region
+
+    def get_upload_s3_bucket(self):
+        """Helper function to determine if we should use the policy default
+        upload bucket or one provided by the user
+
+        :returns: The S3 bucket to use for upload
+        :rtype: ``str``
+        """
+        if self.upload_url and self.upload_url.startswith('s3://'):
+            bucket_and_prefix = self.upload_url[5:].split('/', 1)
+            self.upload_s3_bucket = bucket_and_prefix[0]
+            if len(bucket_and_prefix) > 1:
+                self.upload_s3_object_prefix = bucket_and_prefix[1]
+        if not self.upload_s3_bucket:
+            self.prompt_for_upload_s3_bucket()
+        return self.upload_s3_bucket or self._upload_s3_bucket
+
+    def get_upload_s3_object_prefix(self):
+        """Helper function to determine if we should use the policy default
+        upload object prefix or one provided by the user
+
+        :returns: The S3 object prefix to use for upload
+        :rtype: ``str``
+        """
+        return self.upload_s3_object_prefix or self._upload_s3_object_prefix
+
+    def get_upload_s3_secret_key(self):
+        """Helper function to determine if we should use the policy default
+        upload secret key or one provided by the user
+
+        :returns: The S3 secret key to use for upload
+        :rtype: ``str``
+        """
+        return (os.getenv('SOSUPLOADS3SECRETKEY', None) or
+                self.upload_s3_secret_key or
+                self._upload_s3_secret_key)
+
     def get_upload_url(self):
         """Helper function to determine if we should use the policy default
         upload url or one provided by the user
@@ -475,6 +626,14 @@ class LinuxPolicy(Policy):
         :returns: The URL to use for upload
         :rtype: ``str``
         """
+        if not self.upload_url and (
+            self.upload_s3_bucket and
+            self.upload_s3_access_key and
+            self.upload_s3_secret_key
+        ):
+            bucket = self.get_upload_s3_bucket()
+            prefix = self.get_upload_s3_object_prefix()
+            self._upload_url = f"s3://{bucket}/{prefix}"
         return self.upload_url or self._upload_url
 
     def get_upload_url_string(self):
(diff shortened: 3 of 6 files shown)


TrevorBenson:
@TurboTurtle The x86_64 RPM build succeeds, however, I don't seem to be able to access repo metadata for installation and testing:
```
$ sudo dnf install -y sos-4.6.0-1.20231216003429832355.pr3447.82.g9e05d8c7.fc40.noarch
Copr repo for sosreport-sos-3447 owned by packit                                                                                                                                                              454  B/s | 341  B     00:00    
Errors during downloading metadata for repository 'copr:copr.fedorainfracloud.org:packit:sosreport-sos-3447':
  - Status code: 404 for https://download.copr.fedorainfracloud.org/results/packit/sosreport-sos-3447/fedora-39-x86_64/repodata/repomd.xml (IP: 2600:9000:25f2:ca00:4:bbc1:1840:93a1)
Error: Failed to download metadata for repo 'copr:copr.fedorainfracloud.org:packit:sosreport-sos-3447': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
Ignoring repositories: copr:copr.fedorainfracloud.org:packit:sosreport-sos-3447
Last metadata expiration check: 2:22:12 ago on Fri 15 Dec 2023 02:21:32 PM PST.
No match for argument: sos-4.6.0-1.20231216003429832355.pr3447.82.g9e05d8c7.fc40.noarch
Error: Unable to find a match: sos-4.6.0-1.20231216003429832355.pr3447.82.g9e05d8c7.fc40.noarch
```

pmoravec:
The `copr` build is not there since we haven't bumped Fedora versions since F39 release: we ask for such builds for F37, F38 and rawhide now.

I opened https://github.com/sosreport/sos/pull/3448 to fix that.

pmoravec:
NACK to the plain extension of install requirements. We should keep the hard requirements as small as possible, to allow installation of `sos` to as wide range of systems as possible.

E.g. here, `python3-boto3` is not in RHEL (8 or 9), just in EPEL. So this change would prevent installing `sos` on RHEL :)

Rather make the requirement and import conditional, like:

https://github.com/sosreport/sos/blob/main/sos/policies/distros/__init__.py#L29-L33

arif-ali:
The deb build is failing, as it does some unit tests as part of the build (That's how deb packaging works).

So, we will need to add `python3-boto3`, so that in `debian/control` now has the following change

```
Depends: ${python3:Depends}, ${misc:Depends}, python3-pexpect, python3-magic, python3-packaging, python3-boto3
```

For the snap (as that will then fail too), can you add `boto3` under `python-packages` in `snap/snapcraft.yaml`

@dnegreira FYI, does that sound reasonable for the `control` file, or do we want to avoid doing a Depends on a `universe` package?

TrevorBenson:
@arif-ali Would adding a **Recommends** and placing it there be more appropriate in case Universe is not enabled?

TrevorBenson:
> The `copr` build is not there since we haven't bumped Fedora versions since F39 release: we ask for such builds for F37, F38 and rawhide now.
> 
> I opened #3448 to fix that.

For now, I'll spin up a new F38 VM until the F39 issues are resolved.

arif-ali:
> @arif-ali Would adding a **Recommends** and placing it there be more appropriate in case Universe is not enabled?

could be an option, but want a +1 from my other fellow colleagues who take on the Ubuntu packaging. But like @pmoravec mentioned, we ought to have a condition of the `boto3` being available. Considering that `boto3` is in universe in Ubuntu and only in EPEL for EL 8 and EL9.

From the Ubuntu perspective, we expect this to be available for 20.04, 22.04, 23.04, 23.10 and the upcoming 24.04. All of these have `boto3` in universe

But happy on the snap side, as that would be the only way to enable it, and it builds on `core22` which is a 22.04 base and will ship as standard, and we ship all the libs as required and comes via `pip3`.

TrevorBenson:
> > The `copr` build is not there since we haven't bumped Fedora versions since F39 release: we ask for such builds for F37, F38 and rawhide now.
> > I opened #3448 to fix that.
> 
> For now, I'll spin up a new F38 VM until the F39 issues are resolved.

@pmoravec The instructions from **packit-as-a-service** don't seem to work as described for F38 either:

```
$ sudo dnf install -y dnf-plugins-core
Fedora 38 - x86_64                              3.7 MB/s |  83 MB     00:22    
Fedora 38 openh264 (From Cisco) - x86_64        1.0 kB/s | 2.5 kB     00:02    
Fedora Modular 38 - x86_64                      777 kB/s | 2.8 MB     00:03    
Fedora 38 - x86_64 - Updates                    2.7 MB/s |  36 MB     00:13    
Fedora Modular 38 - x86_64 - Updates            463 kB/s | 2.1 MB     00:04    
Package dnf-plugins-core-4.4.0-1.fc38.noarch is already installed.
Dependencies resolved.
Nothing to do.
Complete!

$ sudo dnf copr enable packit/sosreport-sos-3447
Enabling a Copr repository. Please note that this repository is not part
of the main distribution, and quality may vary.

The Fedora Project does not exercise any power over the contents of
this repository beyond the rules outlined in the Copr FAQ at
<https://docs.pagure.org/copr.copr/user_documentation.html#what-i-can-build-in-copr>,
and packages are not held to any quality or security level.

Please do not file bug reports about these packages in Fedora
Bugzilla. In case of problems, contact the owner of this repository.

Do you really want to enable copr.fedorainfracloud.org/packit/sosreport-sos-3447? [y/N]: y
Error: It wasn't possible to enable this project.
Repository 'fedora-38-x86_64' does not exist in project 'packit/sosreport-sos-3447'.
Available repositories: 'fedora-rawhide-aarch64', 'fedora-rawhide-ppc64le', 'fedora-rawhide-x86_64', 'fedora-rawhide-s390x'

If you want to enable a non-default repository, use the following command:
  'dnf copr enable packit/sosreport-sos-3447 <repository>'
But note that the installed repo file will likely need a manual modification.
```

Since it was identical behavior to F39 I enabled `fedora-rawhide-x86_64` as the only  x86_64 based repo suggested. I haven't used copr from packit in at least a year, should it automatically find the `fedora-38-x86_64`, or should rawhide have the repository metadata for F38?


```
$ sudo dnf copr enable packit/sosreport-sos-3447 fedora-rawhide-x86_64
Enabling a Copr repository. Please note that this repository is not part
of the main distribution, and quality may vary.

The Fedora Project does not exercise any power over the contents of
this repository beyond the rules outlined in the Copr FAQ at
<https://docs.pagure.org/copr.copr/user_documentation.html#what-i-can-build-in-copr>,
and packages are not held to any quality or security level.

Please do not file bug reports about these packages in Fedora
Bugzilla. In case of problems, contact the owner of this repository.

Do you really want to enable copr.fedorainfracloud.org/packit/sosreport-sos-3447? [y/N]: y
Repository successfully enabled.

$ dnf list sos --showduplicates
Copr repo for sosreport-sos-3447 owned by packit                                                           187  B/s | 341  B     00:01    
Errors during downloading metadata for repository 'copr:copr.fedorainfracloud.org:packit:sosreport-sos-3447':
  - Status code: 404 for https://download.copr.fedorainfracloud.org/results/packit/sosreport-sos-3447/fedora-38-x86_64/repodata/repomd.xml (IP: 108.138.85.40)
Error: Failed to download metadata for repo 'copr:copr.fedorainfracloud.org:packit:sosreport-sos-3447': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
Fedora Modular 38 - x86_64 - Updates                                                                       417 kB/s | 2.1 MB     00:05    
Ignoring repositories: copr:copr.fedorainfracloud.org:packit:sosreport-sos-3447
Last metadata expiration check: 0:00:01 ago on Sat 16 Dec 2023 09:21:59 PM EST.
Installed Packages
sos.noarch                                                      4.5.1-1.fc38                                                      @anaconda
Available Packages
sos.noarch                                                      4.5.1-1.fc38                                                      fedora   
```

sos/policies/distros/__init__.py
```diff
@@ -762,6 +878,63 @@ def upload_ftp(self, url=None, directory=None, user=None, password=None):
...
+        if not prefix:
+            prefix = self.get_upload_s3_object_prefix()
+
+        if not access_key:
+            access_key = self.get_upload_s3_access_key()
+
+        if not secret_key:
+            secret_key = self.get_upload_s3_secret_key()
+
+        s3_client = boto3.client('s3', endpoint_url=endpoint,
```
arif-ali, line 926: You probably want to have a condition here with `BOTO3_LOADED`. As if the import hasn't work above, this is going to throw an error. Especially if we're going to shift the boto3 package in Recommends or something like that.

So, I would throw an exception, and then catch it below, and suggest that for this work we need `python3-boto` installed, and then we can let the admin/user decide

arif-ali (changes requested):

pmoravec:
I am trying to understand the `copr` failure (recent copr builds really dont have it..?). Meantime you can use https://cirrus-ci.com/task/6006850617868288 that has `sos_fedora-38.rpm`

dnegreira:
Hi @arif-ali and @TrevorBenson,

> @dnegreira FYI, does that sound reasonable for the control file, or do we want to avoid doing a Depends on a universe package, see below?
> @arif-ali Would adding a **Recommends** and placing it there be more appropriate in case Universe is not enabled?

Unfortunately, both recommendations are not allowed as sosreport is in Ubuntu's main repositories.

The condition to load boto3, if it is available, LGTM as a way forward.

TrevorBenson:
I got a BAD key ID on F38, I originally suspected it was related to [this](https://discussion.fedoraproject.org/t/third-party-rpms-with-an-invalid-signing-key-might-cause-errors-during-package-operations/80077). However I didn't get any key issues on F39 when installing. 

I didn't check if F39 had an additional change and it now uses something only compatible with >=F39. Just letting you know in case you want to drop F38 since it does not appear to be testable.

TrevorBenson:
I think this may be ready to convert from draft to review. I've tested `sos report` with various options and option combinations with (and without):
* **--upload-protocol**
* **--upload-url**
* **--upload-s3-region**
* **--upload-s3-bucket**
* **--upload-s3-endpoint**
* **--upload-s3-access-key**
* **--upload-s3-secret-key**

For my F39 workstation/laptop I am having some issues with testing `sos collect` but I have an el8-based lab I should be able to try it on and believe I have previously used the collector `--transport saltstack` with. So believe I should be able to run some additional tests to confirm it works for collections as well. 

I do however welcome anyone with interest in uploads to s3 buckets to give a Packit RPM a try in case I overlooked anything by performing each test manually.

TrevorBenson:
## Enterprise Linux
### Repositories
@pmoravec Commit 437fe69 extends the `.packit.yaml` to include centos as I suspect CentOS Stream will be the intended repository used for testing on Red Hat 8/9 and downstream recompiles. As the subject of #3448 is specific to Fedora should I open a new PR specific to adding CentOS, or would you prefer to include these changes in the existing PR? 

### Testing
The 4.6.0-5 current el8 RPM was newer than the 4.6.0-1 from COPR. I suspect this may be an artifact due to Red Hat backporting fixes. I couldn't seem to target it specifically as a downgrade either:
```
dnf downgrade 4.6.0-1.20231223172713522168.pr3447.87.g0887f9f5.el8
Last metadata expiration check: 0:13:50 ago on Sat 23 Dec 2023 01:46:10 AM PST.
No package 4.6.0-1.20231223172713522168.pr3447.87.g0887f9f5.el8 available.
Error: No packages marked for downgrade.
```

In any case, I was able to avoid removing the dependencies and get it installed via:

```
rpm -e --nodeps sos
dnf install --disablerepo=baseos sos
```

However, given the simplicity of the Packit suggestion I'm not sure this is the intended way to install on EL distros w/ COPR, or just the quick workaround I came up with. Did I overlook an option to Packit that would have based the RPM off the current upstream NEVRA when generating the RPM for that COPR repository?

-----

I'll clean up this PR by dropping cherry-picks, or commits moving to a new PR, before converting this from Draft to Ready once I know your preference.

### 

Thanks

TrevorBenson:
@arif-ali Do actions/workflows build a Debian PKG for testing? I've tested Fedora and Enterprise Linux with success. I have not tested on Ubuntu distro's though since there were no suggestions similar to Packit.

arif-ali:
> @arif-ali Do actions/workflows build a Debian PKG for testing? I've tested Fedora and Enterprise Linux with success. I have not tested on Ubuntu distro's though since there were no suggestions similar to Packit.

Yes it does, below is the deb artifact from the cirrus builds and was built on Ubuntu 23.10 (mantic), which you could try

https://api.cirrus-ci.com/v1/artifact/task/5584901588647936/packages/sos_cirrus.deb

Alternatively, a snap is also created, and below is the link to the snap that as created as part of the build

https://api.cirrus-ci.com/v1/artifact/task/6710801495490560/packages/sosreport_test_amd64.snap

sos/collector/__init__.py
```diff
@@ -134,7 +134,13 @@ class SoSCollector(SoSComponent):
...
         'upload_method': 'auto',
         'upload_no_ssl_verify': False,
-        'upload_protocol': 'auto'
+        'upload_protocol': 'auto',
+        'upload_s3_endpoint': None,
+        'upload_s3_region': None,
+        'upload_s3_bucket': None,
+        'upload_s3_access_key': None,
+        'upload_s3_secret_key': None,
+        'upload_s3_object_prefix': None
```
TurboTurtle, line 143: Out of curiosity, are *all* these options *required* for an S3 upload? I've only messed with AWS a little bit myself over the years, and I seem to remember some tools using a json file with...some? all? of these items in it and then just having the tool reference the file. Is that a standard design (and should we support that here?) or am I imagining things?

sos/policies/distros/__init__.py
```diff
@@ -354,6 +385,54 @@ def _configure_low_priority(self):
...
 
+    def prompt_for_upload_s3_access_key(self):
+        """Should be overridden by policies to determine if an access key needs
+        to be provided for upload or not
+        """
+        if not self.get_upload_s3_access_key():
+
+            msg = ("Please provide the upload access key for bucket %s via "
+                   "endpoint %s: " % (self.get_upload_s3_bucket(),
+                                      self.get_upload_s3_endpoint()))
```
TurboTurtle, line 396: Please convert to f-strings, we're trying to make f-strings the standard in the project.

sos/policies/distros/__init__.py
```diff
@@ -354,6 +385,54 @@ def _configure_low_priority(self):
...
+            self.upload_s3_access_key = input(_(msg))
+
+    def prompt_for_upload_s3_secret_key(self):
+        """Should be overridden by policies to determine if a secret key needs
+        to be provided for upload or not
+        """
+        if not self.get_upload_s3_secret_key():
+            msg = ("Please provide the upload secret key for bucket %s via "
+                   "endpoint %s: " % (self.get_upload_s3_bucket(),
+                                      self.get_upload_s3_endpoint()))
```
TurboTurtle, line 406: Ditto.

sos/policies/distros/__init__.py
```diff
@@ -354,6 +385,54 @@ def _configure_low_priority(self):
...
+        return self.upload_s3_bucket
+
+    def prompt_for_upload_s3_endpoint(self):
+        """Should be overridden by policies to determine if an endpoint needs
+        to be provided for upload or not
+        """
+        default_endpoint = self._upload_s3_endpoint
+        if not self.upload_s3_endpoint:
+            msg = "Please provide the upload endpoint for bucket %s " \
+                  "(default: %s): " % (
```
TurboTurtle, line 428: Ditto.

sos/policies/distros/__init__.py
```diff
@@ -762,6 +917,71 @@ def upload_ftp(self, url=None, directory=None, user=None, password=None):
...
+        :type secret_key: str
+
+        :returns: True if upload is successful
+        :rtype: bool
+
+        :raises: Exception if upload is unsuccessful
+        """
+        if not BOTO3_LOADED:
+            raise Exception("Unable to upload due to missing python boto3 "
+                            "library")
```
TurboTurtle, line 949: Ack.

sos/policies/distros/__init__.py
```diff
@@ -762,6 +917,71 @@ def upload_ftp(self, url=None, directory=None, user=None, password=None):
...
+                                 aws_access_key_id=access_key,
+                                 aws_secret_access_key=secret_key)
+
+        try:
+            key = prefix + self.upload_archive_name.split('/')[-1]
+            s3_client.upload_file(self.upload_archive_name,
+                                  bucket, key)
+            return True
+        except Exception as e:
+            raise Exception("Failed to upload to S3: %s" % str(e))
```
TurboTurtle, line 983: f-string here, please.

TurboTurtle (commented):
Overall it looks good from a code perspective, some minor notes below.

I'm kind of torn on the `Recommends` - I fully get the intent (and certainly not having it as a hard `Requires`), but afaik _most_ rpm-based distros default to "always install recommends packages" and I wonder if pulling in boto3 as, effectively, a default is the right choice.

TrevorBenson:
> Overall it looks good from a code perspective, some minor notes below.
> 
> I'm kind of torn on the `Recommends` - I fully get the intent (and certainly not having it as a hard `Requires`), but afaik _most_ rpm-based distros default to "always install recommends packages" and I wonder if pulling in boto3 as, effectively, a default is the right choice.

I'll make whatever change is suggested. However, isn't the benefit of using **Recommends** instead of **Requires** that the end user could bypass this package altogether when using `--setopt=install_weak_deps=False` and avoid potential installation failures in the event the python3-boto RPM is not available? While the latter should be rare, unless repositories are being recreated without complete package sets, I have seen this and deal with it almost daily.

TurboTurtle:
> > Overall it looks good from a code perspective, some minor notes below.
> > I'm kind of torn on the `Recommends` - I fully get the intent (and certainly not having it as a hard `Requires`), but afaik _most_ rpm-based distros default to "always install recommends packages" and I wonder if pulling in boto3 as, effectively, a default is the right choice.
> 
> I'll make whatever change is suggested. However, isn't the benefit of using **Recommends** instead of **Requires** that the end user could bypass this package altogether when using `--setopt=install_weak_deps=False` and avoid potential installation failures in the event the RPM is not available? While the latter should be rare, unless repositories are being recreated without complete package sets, I have seen this and deal with it almost daily.


Oh I'm almost certainly being too paranoid about this :).

My new job has just happened to push package set security (i.e. "as little as possible") to the front of my mind at this point in time, that's all.

Taking a bit of a step back from that, you're right. Anyone who actually cares about this will either know about weak deps or have already disabled them. No issue from me on the spec file anymore.

sos/collector/__init__.py
TrevorBenson: > Out of curiosity, are all these options required for an S3 upload? 

Kind of.

## Required
* **upload_s3_bucket** required in any s3 upload.
* **upload_s3_access_key** required in any s3 upload.
* **upload_s3_secret_key** required in any s3 upload.

## Optional
* **upload_s3_endpoint** default is **s3.amazonaws.com**.
  * Could be removed if only supporting Amazon-hosted buckets, blocking all third-party s3 providers.
* **upload_s3_region** defaults to **us-east-1**
  * This default is usually acceptable for all providers, as even third-party providers tend to implement the API with **us-east-1** as the default region.
  * However, best practice is to provide the correct region for a bucket. This avoids first going to **us-east-1** to determine the correct region. Also allows avoiding upload failures during a regional or service outage in **us-east-1** where it cannot be used to determine the correct region.

## Specific to SOS
* **upload_s3_object_prefix** is a convention created by me for sos so that the filename being uploaded can be combined with an object prefix to define the key (including prefixes) to upload (URL path in standard terminology).
  * As there is no direct access to the awscli to specify the key including its prefixes (i.e. full path to the object inside the bucket) **and** because s3 does not honor normal protocol URL conventions. 
  * A standard URL`proto://<host.domain.tld>/path/of/url` vs. an S3 url `proto://<bucket>/object/prefix/key`
    * e.g. `https://www.example.com/sosreports/2024/01/sosreport-myhost-2024-01-08-askljf.tar.xz` is instead split into pieces where the host is `--endpoint-url www.example.com` and the rest of the URL is `s3://sosreports/2024/01/sosreport-myhost-2024-01-08-askljf.tar.xz`.

    
Since the API has specific differences for how the URL is defined I did not want to hack something together from `--upload-url` which would be unique to sos and not used in awscli or any other s3 client implementations I've seen.

---

> I've only messed with AWS a little bit myself over the years, and I seem to remember some tools using a json file with...some? all? of these items in it and then just having the tool reference the file. Is that a standard design (and should we support that here?) or am I imagining things?

When using **awscli** I've used the `~/.aws/credentials` and the `~/.aws/config` for SSO. Both configuration files are in INI syntax styles. I have also seen custom scripts read from a JSON file, but I do not believe that is a standard design from a review of the [boto3 credentials docs](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html). 

That said I did not test for values coming from either configuration file at all. I presumed they would likely not be setup for sreport submissions and therefore put options for each potential argument required.

----

sos/policies/distros/__init__.py
TrevorBenson: Cool, I prefer them anyway, but was maintaining any existing styles I observed from other plugins since I forgot which contribution page I thought I saw a suggestion of moving to f-strings.

sos/policies/distros/__init__.py
TrevorBenson: I presume this is intended as "acknowledge" for acceptance of a requested/suggested change.

sos/collector/__init__.py
TrevorBenson: Leaving this conversation as unresolved just in case my response drives additional suggestions.

TurboTurtle: Works for me. If we find that we need or want to add support for getting these values from a json file we can add that later - and in the meantime we can just direct users to putting these values in `sos.conf` if truly desired.

TrevorBenson: Sounds good. If we add a JSON file later I'll also add the `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY` and `AWS_DEFAULT_REGION` environment variables as well as an `--upload-s3-profile` similar to `aws --profile` to override the environment vars by leveraging the profiles in `~/.aws/config` or `~/.aws/credentials`.

TrevorBenson:
> > @arif-ali Do actions/workflows build a Debian PKG for testing? I've tested Fedora and Enterprise Linux with success. I have not tested on Ubuntu distro's though since there were no suggestions similar to Packit.
> 
> Yes it does, below is the deb artifact from the cirrus builds and was built on Ubuntu 23.10 (mantic), which you could try
> 
> https://api.cirrus-ci.com/v1/artifact/task/5584901588647936/packages/sos_cirrus.deb
> 
> Alternatively, a snap is also created, and below is the link to the snap that as created as part of the build
> 
> https://api.cirrus-ci.com/v1/artifact/task/6710801495490560/packages/sosreport_test_amd64.snap

Seems I forgot to report back on this the other week. 

@arif-ali  I tested both the deb package and the snap (using `--classic --dangerous`). As long as the system had the **python3-boto3** package installed the s3 upload worked as intended.

TrevorBenson:
The force push was for the rebase on #3463 so that a5b9539 could be added.

TrevorBenson:
@TurboTurtle @arif-ali 
Does the **rpm Build From Checkout - centos-stream-8** failure (due to timeout) need to be addressed? I don't see a workflow dispatch or an option to rerun failed tests.

If not I believe this PR is ready for final reviews.

TurboTurtle:
I restarted the job - since this particular build is done via Cirrus only Pavel, Arif, Bryn, and myself have permissions to restart it (since the permission handling on non GH-enterprise orgs like us is very coarse - you have to be an "owner" to do it here).

The timeout is definitely not related to the changes here, but it does block the other cirrus jobs from running, so I'll watch this and restart it again if needed.

TurboTurtle (commented):
Code looks good to me, just a stickler point from our Contribution Guidelines. The commit messages should be prefixed with a relevant tag, e.g. `[upload]` or somesuch (it makes trawling the git history _much_ easier as time goes on). Also the second commit should be squashed into the first. The S3 obfuscation commit is perfectly fine being a separate commit in the chain.

TrevorBenson:
> Code looks good to me, just a stickler point from our Contribution Guidelines. The commit messages should be prefixed with a relevant tag, e.g. `[upload]` or somesuch (it makes trawling the git history _much_ easier as time goes on). Also the second commit should be squashed into the first. The S3 obfuscation commit is perfectly fine being a separate commit in the chain.

Added the context `[upload]` and squashed.

TurboTurtle (approved):

pmoravec:
I just dont get the `ubuntu-18.4` CI test error:

    Could not initialize 'report': future feature annotations is not defined (__init__.py, line 5)

We dont use `annotations` from `future` class, esp. we dont import that on line 5. Logs from the test are very concise..

arif-ali:
> I just dont get the `ubuntu-18.4` CI test error:
> 
> ```
> Could not initialize 'report': future feature annotations is not defined (__init__.py, line 5)
> ```
> 
> We dont use `annotations` from `future` class, esp. we dont import that on line 5. Logs from the test are very concise..

I'll need to check this locally, and see what is going on.

Ultimately, the test is being done, once the snap is installed and the snap is based on 22.04's python, and doesn't use the python on 18.04. I will get back on this one, and report back on why this is not working

arif-ali:
On that, below is the error we get on this particular patch, it's not obvious where the issue is, and remember debugging something very similar early last year, I will continue to debug, and get back on the resolution

```
root@bionic-sos-snap-test:~/sos# sos collect --list-options --debug
Could not initialize 'collect': future feature annotations is not defined (__init__.py, line 5)
  File "/snap/sosreport/x1/bin/sos", line 21, in <module>
    sos = SoS(sys.argv[1:])
  File "/root/sos/sos/__init__.py", line 124, in __init__
    self._init_component()
  File "/root/sos/sos/__init__.py", line 189, in _init_component
    raise err
  File "/root/sos/sos/__init__.py", line 184, in _init_component
    self._component = _to_load(self.parser, self.args, self.cmdline)
  File "/root/sos/sos/collector/__init__.py", line 141, in __init__
    super(SoSCollector, self).__init__(parser, parsed_args, cmdline_args)
  File "/root/sos/sos/component.py", line 90, in __init__
    self.load_local_policy()
  File "/root/sos/sos/component.py", line 132, in load_local_policy
    probe_runtime=self.load_probe)
  File "/root/sos/sos/policies/__init__.py", line 36, in load
    import sos.policies.distros
  File "/root/sos/sos/policies/distros/__init__.py", line 30, in <module>
    import requests
  File "/usr/lib/python3/dist-packages/requests/__init__.py", line 43, in <module>
    import urllib3
  File "/snap/sosreport/x1/lib/python3.10/site-packages/urllib3/__init__.py", line 5
    from __future__ import annotations
    ^
SyntaxError: future feature annotations is not defined
```

snap/snapcraft.yaml
```diff
@@ -37,6 +37,7 @@ parts:
       - wheel
       - python_magic
       - packaging
+      - boto3
```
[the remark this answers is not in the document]
TrevorBenson: > @TrevorBenson lets change this to give it a specific version, I think the new version of the `boto3` may be incompatible with bionic, and hence causing issue. I rebuilt the snap with the new version, and didn't have this problem
> 
> ```
> - boto3==1.26.155
> ```

@TurboTurtle @pmoravec In the past, for Enterprise Linux systems when not also dealing with Debian packages, I have relied on Red Hat backporting to avoid breaking changes. At least for RHBAs and RHSAs, and fixed issues during an RHEA (rarely) when they had a breaking change. Is that how it's handled for sos or should I pin the boto3 version for RPMs as well to keep a similar logic on both distros?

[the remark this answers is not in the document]
arif-ali: The difference here is that in this particular scenario it is installing these `python_packages` from pip.

If this was a deb package, then it wouldn't hav been a problem as it would have grabbed it from it's distro packages and would be compatible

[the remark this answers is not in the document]
TrevorBenson: @arif-ali added commit 794619f to pin the version for the snap. I'll squash it once I hear back from the others if a similar change is wanted for the RPM.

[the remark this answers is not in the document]
dnegreira: @TrevorBenson thanks for pinning the  `boto` version on snap, I opened #3481 so that we follow-up on this later on.

arif-ali (approved):

PR #3450, opened 2023-12-18
[openstack] Add collection of commands

arif-ali:
Similar to how we do in openstack_nova, collect commands from cinder and collect all the detals of the glance images.

Resolves: SET-337

---

### DIFF
--- a/sos/report/plugins/openstack_cinder.py
+++ b/sos/report/plugins/openstack_cinder.py
@@ -13,6 +13,7 @@
 # See the LICENSE file in the source distribution for further information.
 
 from sos.report.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+import os
 
 
 class OpenStackCinder(Plugin):
@@ -48,6 +49,66 @@ class OpenStackCinder(Plugin):
                 "cinder-manage " + cinder_config + " db version",
                 suggest_filename="cinder_db_version"
             )
+            self.add_cmd_output(
+                f"cinder-manage {cinder_config} backup list"
+            )
+            self.add_cmd_output(
+                f"cinder-manage {cinder_config} config list"
+            )
+            self.add_cmd_output(
+                f"cinder-manage {cinder_config} host list"
+            )
+            self.add_cmd_output(
+                f"cinder-status {cinder_config} upgrade check"
+            )
+
+            vars_all = [p in os.environ for p in [
+                        'OS_USERNAME', 'OS_PASSWORD']]
+
+            vars_any = [p in os.environ for p in [
+                        'OS_TENANT_NAME', 'OS_PROJECT_NAME']]
+
+            if not (all(vars_all) and any(vars_any)):
+                self.soslog.warning("Not all environment variables set. "
+                                    "Source the environment file for the user "
+                                    "intended to connect to the OpenStack "
+                                    "environment.")
+            else:
+                list_cmds = [
+                    "backend pool",
+                    "group type",
+                    "message",
+                    "qos",
+                    "service",
+                    "type",
+                ]
+
+                for cmd in list_cmds:
+                    self.add_cmd_output(f"openstack volume {cmd} list")
+
+                list_cmds_projects = [
+                    "backup",
+                    "group",
+                    "group snapshot",
+                    "snapshot",
+                    "transfer request",
+                    "",
+                ]
+
+                for cmd in list_cmds_projects:
+                    self.add_cmd_output(
+                        f"openstack volume {cmd} list --all-projects"
+                    )
+
+                # get details for each volume
+                cmd = "openstack volume list -f value --all-projects"
+                res = self.exec_cmd(cmd)
+                if res['status'] == 0:
+                    cinder_volumes = res['output']
+                    for volume in cinder_volumes.splitlines():
+                        volume = volume.split()[0]
+                        cmd = f"openstack volume show {volume}"
+                        self.add_cmd_output(cmd)
 
         self.add_forbidden_path('/etc/cinder/volumes')
         self.add_copy_spec([
--- a/sos/report/plugins/openstack_glance.py
+++ b/sos/report/plugins/openstack_glance.py
@@ -70,7 +70,16 @@ class OpenStackGlance(Plugin):
                                     "intended to connect to the OpenStack "
                                     "environment.")
             else:
-                self.add_cmd_output("openstack image list --long")
+                res = self.collect_cmd_output(
+                    "openstack image list --long"
+                )
+
+                if res['status'] == 0:
+                    glance_images = res['output']
+                    for image in glance_images.splitlines()[3:-1]:
+                        image = image.split()[1]
+                        cmd = f"openstack image show {image}"
+                        self.add_cmd_output(cmd)
 
         self.add_file_tags({
             "/etc/glance/glance-api.conf": "glance_api_conf",
@@ -117,7 +126,7 @@ class DebianGlance(OpenStackGlance, DebianPlugin, UbuntuPlugin):
         'python-glance',
         'python3-glance',
     )
-    service_name = 'glance-api.service'
+    service_name = 'apache2.service'
 
 
 class RedHatGlance(OpenStackGlance, RedHatPlugin):
--- a/sos/report/plugins/openstack_heat.py
+++ b/sos/report/plugins/openstack_heat.py
@@ -52,7 +52,21 @@ class OpenStackHeat(Plugin):
                                     "intended to connect to the OpenStack "
                                     "environment.")
             else:
-                self.add_cmd_output("openstack stack list")
+                self.add_cmd_output("openstack stack list --all-projects "
+                                    "--nested")
+
+                res = self.collect_cmd_output(
+                    "openstack stack list --all-projects"
+                )
+
+                if res['status'] == 0:
+                    heat_stacks = res['output']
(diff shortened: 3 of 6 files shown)


sos/report/plugins/openstack_cinder.py
```diff
@@ -48,6 +49,71 @@ def setup(self):
...
+                    "snapshot",
+                    "transfer request",
+                ]
+
+                for cmd in list_cmds_projects:
+                    self.add_cmd_output(
+                        f"openstack volume {cmd} list --all-projects"
+                    )
+
+                self.add_cmd_output("openstack volume list --all-projects")
```
pmoravec, line 106: You might add it to the previous `for cmd in ..` loop - bit more concise code, bit less readable, I guess.

Leaving on you the decision, both is good for me.

sos/report/plugins/openstack_cinder.py
```diff
@@ -48,6 +49,71 @@ def setup(self):
...
+
+                # get details for each volume
+                cmd = "openstack volume list -f value --all-projects"
+                cinder_volumes = self.exec_cmd(cmd)['output']
```
pmoravec, line 110: How reliable the execution is? Isn't it safer to test for `'status'` first? Well, the output is always empty for failed command (or truncated stdout for timeouted use case)..

sos/report/plugins/openstack_glance.py
```diff
@@ -70,7 +70,18 @@ def setup(self):
                                     "intended to connect to the OpenStack "
                                     "environment.")
             else:
-                self.add_cmd_output("openstack image list --long")
+                self.add_cmd_output(
+                    "openstack image list --long --all-projects"
+                )
+
+                cmd = "openstack image list -f value --all-projects"
+                glance_images = self.exec_cmd(cmd)['output']
```
pmoravec, line 78: Same like in `cinder_volumes`. Again leaving on author's decision how safe it is..

sos/report/plugins/openstack_nova.py
```diff
@@ -71,17 +71,16 @@ def setup(self):
                                     "intended to connect to the OpenStack "
                                     "environment.")
             else:
-                self.add_cmd_output("nova service-list")
+                self.add_cmd_output("openstack compute service list")
                 self.add_cmd_output("openstack flavor list --long")
-                self.add_cmd_output("nova network-list")
-                self.add_cmd_output("nova list --all-tenants")
-                self.add_cmd_output("nova agent-list")
```
pmoravec, line 78: The `nova` commands are deprecated? Shall we keep them for some time for legacy reasons, or are they gone for a while?

pmoravec:
Cc @rebtoor or @lyarwood for a review, please.

sos/report/plugins/openstack_cinder.py
arif-ali: makes sense. I was kind of following the nova way, but happy to amend, to make it better :)

sos/report/plugins/openstack_nova.py
arif-ali: The `nova` commands are not deprecated, and still exist, some sub commands no longer exist, and more of the features are being added to the `openstack` command. So makes sense to move as many over. At least that was my thought process here

rebtoor: I agree with @arif-ali . I guess that the only command deprecated there is `nova network-list`

rebtoor (commented):
lgtm for me

sos/report/plugins/openstack_cinder.py
```diff
@@ -48,6 +49,72 @@ def setup(self):
...
+                        'OS_USERNAME', 'OS_PASSWORD']]
+
+            vars_any = [p in os.environ for p in [
+                        'OS_TENANT_NAME', 'OS_PROJECT_NAME']]
+
+            if not (all(vars_all) and any(vars_any)):
+                self.soslog.warning("Not all environment variables set. "
+                                    "Source the environment file for the user "
+                                    "intended to connect to the OpenStack "
+                                    "environment.")
```
pmoravec, line 79: Maybe add an information this execution is skipping some commands execution? Either explicitly, or e.g. by appending something like ".. to get further data collected".

But I am OK with the current text, it does contain an explicit request.

pmoravec (approved):

sos/report/plugins/openstack_cinder.py
```diff
@@ -48,6 +49,72 @@ def setup(self):
...
+                cmd = "openstack volume list -f value --all-projects"
+                res = self.exec_cmd(cmd)
+                if res['status'] == 0:
+                    cinder_volumes = res['output']
+                    for volume in cinder_volumes.splitlines():
+                        volume = volume.split()[0]
+                        cmd = f"openstack volume show {volume}"
+                        self.add_cmd_output(
+                            cmd,
+                            suggest_filename=f"volume-{volume}.log")
```
TurboTurtle, line 117: Why override the command naming here?

sos/report/plugins/openstack_glance.py
```diff
@@ -70,7 +70,20 @@ def setup(self):
...
+                cmd = "openstack image list -f value --all-projects"
+                res = self.exec_cmd(cmd)
+                if res['status'] == 0:
+                    glance_images = res['output']
+                    for image in glance_images.splitlines():
+                        image = image.split()[0]
+                        cmd = f"openstack image show {image}"
+                        self.add_cmd_output(
+                            cmd,
+                            suggest_filename=f"image-{image}.log")
```
TurboTurtle, line 86: Ditto, why override the command naming?

sos/report/plugins/openstack_heat.py
```diff
@@ -52,7 +52,24 @@ def setup(self):
...
+                    for stack in heat_stacks.splitlines():
+                        stack = stack.split()[0]
+                        cmd = f"openstack stack show {stack}"
+                        self.add_cmd_output(
+                            cmd,
+                            suggest_filename=f"stack-{stack}.log")
+                        cmd = f"openstack stack resource list {stack} -n 10"
+                        self.add_cmd_output(
+                            cmd,
+                            suggest_filename=f"stack-resources-{stack}.log")
```
TurboTurtle, line 72: Ditto. The command naming seems clear and human friendly enough, not sure why we're casting these as log files instead of the command output.

sos/report/plugins/openstack_neutron.py
```diff
@@ -55,14 +55,29 @@ def setup(self):
...
+
+                new_cmd = f"openstack {cmd} list -f value"
+                res = self.exec_cmd(new_cmd)
+                if res['status'] == 0:
+                    neutron_items = res['output']
+                    for item in neutron_items.splitlines():
+                        item = item.split()[0]
+                        show_cmd = f"openstack {cmd} show {item}"
+                        fname = f"{cmd.replace(' ', '_')}-{item}.log"
+                        self.add_cmd_output(show_cmd, suggest_filename=fname)
```
TurboTurtle, line 80: Again, same here.

sos/report/plugins/openstack_nova.py
TurboTurtle: Ack, if openstack is moving these commands over time and the idea is that at _some_ point the `nova` commands will be removed, it's good to get ahead of the game.

sos/report/plugins/openstack_placement.py
```diff
@@ -39,6 +40,38 @@ def setup(self):
...
+                            "inventory",
+                            "trait",
+                            "aggregate",
+                        ]
+                        for sub_cmd in sub_cmds:
+                            cmd = (f"openstack resource provider {sub_cmd} "
+                                   " list {provider}")
+                            file = (f"resource_provider_{sub_cmd}-{provider}"
+                                    ".log")
+                            self.add_cmd_output(cmd, suggest_filename=file)
```
TurboTurtle, line 73: Same, I don't see the need for renaming the command output to a log file.

This can also be neatly buttoned up as

```python
self.add_cmd_output([
    f"openstack resource provider {sub_cmd} list {provider}"
    for sub_cmd in sub_cmds
])
```

TurboTurtle (commented):
The code change looks ok overall, but I don't get why we're leaning on `suggest_filename` so much here, when the command syntax is fairly human-friendly to begin with. Not least of which this renaming implies we're capturing a log file from the host, rather than executing and saving command output which could obfuscate any debugging efforts in the future.

sos/report/plugins/openstack_glance.py
```diff
@@ -70,7 +70,18 @@ def setup(self):
                                     "intended to connect to the OpenStack "
                                     "environment.")
             else:
-                self.add_cmd_output("openstack image list --long")
+                self.add_cmd_output(
+                    "openstack image list --long --all-projects"
+                )
```
pmoravec, line 75: We are calling very similar command in `exec_cmd` few lines below. Can't we get the list of images from this output, if we would call here `self.collect_cmd_output("openstack image list --long --all-projects")`? Or is parsing that output worse than calling an extra similar command once again, like you do now?

arif-ali: let me see what I can do, the less commands we run the better for performance. I was in 2 minds before as well. But, now that you've questioned it as well, I'll try re-factoring it, and see where I can get this to

TurboTurtle (approved):

pmoravec (approved):

PR #3451, opened 2023-12-18
[atomic] Remove atomic policy

jcastill:
RHEL Atomic Host is RHEL7 only, and we since
we don't ship sos 4.* in RHEL 7, we can drop
this policy.

---


pmoravec:
Generally ACK, but isn't also `AtomicHost` class plugin (whole `sos/report/plugins/atomichost.py`) obsolete..?

jcastill:
I thought about removing it as well, but then I doubted because we've never removed a plugin as far as I know, and I didn't know if there's a corner case where we could still get this plugin to work. But if you think there's no such case, I'll update the PR removing it as well.

pmoravec:
Per https://access.redhat.com/solutions/4813941 , Atomic Host is EOL and on RHEL7 only.

@TurboTurtle, any thought about removing plugin?

TurboTurtle:
I'll defer to you on removal of the plugin, but given the history of Atomic Host I don't have any concerns about removing it personally.

TurboTurtle (approved):

pmoravec (approved):

Issue #3207, opened 2023-04-27
support for pebble service manager


dosaboy:
We have some deployments that use https://github.com/canonical/pebble to manage services and it would be helpful if sosreport contained related information in a similar way to the systemd plugin.

TurboTurtle:
No problem about adding a plugin - we just need a short list of what config files and/or command collections you'd like to see in it. Also if there is any helpful information that can be programmatically collated from either of those sources, we now have support for that in plugins as well.

One clarification for me - is pebble an init system in the same vein systemd is? If so, would it make sense to add a runtime abstraction for it so that plugins can be enabled based on the services pebble manages on systems that use it?

nkshirsagar:
@TurboTurtle I'd be interested in working on this so can you assign me to this issue please? I'm not able to assign myself to it maybe a permissions issue

TurboTurtle:
> @TurboTurtle I'd be interested in working on this so can you assign me to this issue please? I'm not able to assign myself to it maybe a permissions issue

Done.

PR #3453, opened 2023-12-22
[pebble][init] Add pebble service manager

arif-ali:
Closes: #3207
Resolves: SET-86

---

### DIFF
--- a/sos/policies/init_systems/pebble.py
+++ b/sos/policies/init_systems/pebble.py
@@ -1,0 +1,53 @@
+# Copyright (C) 2024 Canonical Ltd., Arif Ali <arif.ali@canonical.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.policies.init_systems import InitSystem
+from sos.utilities import shell_out
+
+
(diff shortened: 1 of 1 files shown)


sos/policies/init_systems/pebble.py
```diff
@@ -0,0 +1,53 @@
...
+            list_cmd='services',
+            query_cmd='pebble services',
+            chroot=chroot
+        )
+        self.load_all_services()
+
+    def parse_query(self, output):
+        for line in output.splitlines()[1:]:
+            if line.split()[2] == 'active':
+                return line.split()[2]
```
pmoravec, line 30: Why not simply `return 'active'`, then?

sos/policies/init_systems/pebble.py
```diff
@@ -0,0 +1,53 @@
...
+        for line in svcs:
+            try:
+                name = line.split()[0]
+                config = line.split()[1]
+                self.services[name] = {
+                    'name': name,
+                    'config': config,
+                }
+            except IndexError:
+                pass
```
TurboTurtle, line 44: To keep CodeQL happy, let's add a comment here explaining either how we got to this exception (header/foooter output?) or some similar note to future-us.

TurboTurtle (commented):
Looks good on its own (just the CodeQL note) - where do we anticipate this being used? Do we need to link up any existing policies to it now or is this prep work for something coming later?

sos/policies/init_systems/pebble.py
arif-ali: sure, it was a copy-paste from the systemd one. So, I think we ought to fix tat one too ;)

PR #3454, opened 2023-12-24
[snap] Add snapcraft for strictmode

arif-ali:
With Ubuntu Desktop Core to be available in April, a new sosreport may be required for debugging annd grabbing data of these systems. This is an inital start to get this implemented and automated in publishing for usage.

Related: SET-380

---

- [x] Does tsnap/snapcraft_strictmode.yamlhe subject start with **[plugin_name]** if submitting a plugin patch or a **[section_name]** if part of the core sosreport code?

### DIFF
--- a/snap/snapcraft_strictmode.yaml
+++ b/snap/snapcraft_strictmode.yaml
@@ -1,0 +1,69 @@
+name: sosreport
+summary: Sos is an extensible, portable, support data collection tool
+description: |
+  Sos is an extensible, portable, support data collection tool
+  primarily aimed at Linux distributions and other UNIX-like operating
+  systems.
+grade: stable
+base: core22
+confinement: strict
+adopt-info: sos
+license: GPL-2.0-or-later
+environment:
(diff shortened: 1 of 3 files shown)


snap/snapcraft_strictmode.yaml
```diff
@@ -0,0 +1,69 @@
...
+      - python3-magic
+      - python3-packaging
+      - python3-pexpect
+  cleanup:
+    after: [sos]
+    plugin: nil
+    build-snaps: [core22]
+    override-prime: |
+      set -eux
+      for snap in "core22"; do  # List all content-snaps you're using here
```
pmoravec, line 54: You plan to extend the list of snaps in future, right? As currently the for cycle with 1 iteration makes a little sense..

arif-ali: good point, it was a copy and paste from somewhere else. I will solve that, once I've figured out why the stageone tasks are not going through, and solve these.

PR #3456, opened 2023-12-24
[collector] Cluster profile for Saltstack

TrevorBenson:
Creates a cluster transport for Saltstack. Options include
- **minion_id_unresolvable**
  - Causes the cluster profile to obtain the fqdn from the minions grains.
  - Use when a custom minion_id is defined and/or the server hostname is not resolvable to the collector, but the fqdn grain is.
  - When the salt-minion service is offline the minion_id will be included in the nodes list, even when unresolvable.
- **compound** targeting
- **glob** targeting
- **grain** targeting
- **nodegroup** targeting
- **pillar** targeting
- **subnet** targeting

EDIT: Description updated to match final implementation.

---

### DIFF
--- a/sos/collector/clusters/saltstack.py
+++ b/sos/collector/clusters/saltstack.py
@@ -1,0 +1,82 @@
+# Copyright Red Hat 2022, Trevor Benson <trevor.benson@gmail.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+import json
+from shlex import quote
+from sos.collector.clusters import Cluster
+
+
+class saltstack(Cluster):
+    """
+    The saltstack cluster profile is intended to be used on saltstack
+    clusters (Salt Project).
+    """
+
+    cluster_name = "Saltstack"
+    packages = ("salt-master",)
+    sos_plugins = ["saltmaster"]
+    strict_node_list = True
+    option_list = [
+        ("compound", "", "Filter node list to those matching compound"),
+        ("glob", "", "Filter node list to those matching glob pattern"),
+        ("grain", "", "Filter node list to those with matching grain"),
+        ("minion_id_unresolvable", False, "Returns the FQDN grain of each"
+         " minion in the node list when the minion ID is not a hostname."),
+        ("nodegroup", "", "Filter node list to those matching nodegroup"),
+        ("pillar", "", "Filter node list to those with matching pillar"),
+        ("subnet", "", "Filter node list to those in subnet"),
+    ]
+    targeted = False
+
+    node_cmd = "salt-run --out=pprint manage.status"
+
+    def _parse_manage_status(self, output: str) -> list:
+        nodes = []
+        salt_json_output = json.loads(output.replace("'", '"'))
+        for _, value in salt_json_output.items():
+            nodes.extend(value)
+        return nodes
+
+    def _get_hostnames_from_grain(self, manage_status: dict) -> list:
+        hostnames = []
+        fqdn_cmd = "salt --out=newline_values_only {minion} grains.get fqdn"
+        for status, minions in manage_status.items():
+            if status == "down":
+                self.log_warn(f"Node(s) {minions} are status down.")
+                hostnames.extend(minions)
+            else:
+                for minion in minions:
+                    node_cmd = fqdn_cmd.format(minion=minion)
+                    hostnames.append(
+                        self.exec_primary_cmd(node_cmd)["output"].strip()
+                    )
+        return hostnames
+
+    def _get_nodes(self) -> list:
+        res = self.exec_primary_cmd(self.node_cmd)
+        if res["status"] != 0:
+            raise Exception("Node enumeration did not return usable output")
+        if self.get_option("minion_id_unresolvable"):
+            status = json.loads(res["output"].replace("'", '"'))
+            return self._get_hostnames_from_grain(status)
+        return self._parse_manage_status(res["output"])
+
+    def get_nodes(self):
+        # Default to all online nodes
+        for option in self.option_list:
+            if option[0] != "minion_id_unresolvable":
+                opt = self.get_option(option[0])
+                if opt:
+                    self.node_cmd += f" tgt={quote(opt)} tgt_type={option[0]}"
+                    break
+        return self._get_nodes()
+
+
+# vim: set et ts=4 sw=4 :


sos/collector/clusters/saltstack.py
```diff
@@ -0,0 +1,107 @@
...
+        ("grain", "", "Filter node list to those with matching grain"),
+        ("hostnames", False, "When minion isn't resolvable for control_persist"
+         " this will return the fqdn of the minion. This option is not"
+         " compatible with option all_nodes or transport saltstack"),
+        ("list", "", "Filter node list to those matching list"),
+        ("nodegroup", "", "Filter node list to those matching nodegroup"),
+        ("pillar", "", "Filter node list to those with matching pillar"),
+        ("regex", "", "Filter node list to those matching regex"),
+        ("subnet", "", "Filter node list to those in subnet"),
+    ]
```
TurboTurtle, line 42: A few thoughts here...

- The default behavior of a cluster profile should be to return all nodes in the cluster, and from there various options should allow reducing/filtering it.
- The `list`, `glob`, and `regex` options are redundant with the current behavior of `--nodes`.
- For `hostnames`, I'm confused. If we can't resolve the host, why return an FQDN that needs to be resolvable in order to be useful?

sos/collector/clusters/saltstack.py
```diff
@@ -0,0 +1,107 @@
...
+    """
+    The saltstack cluster profile is intended to be used on saltstack
+    clusters (Salt Project).
+    """
+
+    cluster_name = "Saltstack"
+    packages = ("salt-master",)
+    sos_plugins = ["saltmaster"]
+    cmd = "salt-run --out=pprint manage.up"
+    hostname_cmd = "salt --out=newline_values_only {minion} grains.get fqdn"
```
TurboTurtle, line 26: Since this is only used in one place, this should be used within that method rather than class-wide.

sos/collector/clusters/saltstack.py
```diff
@@ -0,0 +1,107 @@
...
+        if res["status"] != 0:
+            raise Exception("Node enumeration did not return usable output")
+        salt_json_output = json.loads(res["output"].replace("'", '"'))
+        for _, value in salt_json_output.items():
+            nodes.extend(value)
+        if self.get_option("hostnames"):
+            print("Hostname option not implemented for all_nodes, ignoring.")
+        return nodes
+
+    def get_nodes(self):
```
TurboTurtle, line 74: Is `tgt_type` exclusive within `salt-run` to a single entry? I think we can clean this up a bit, also considering the point above about defaulting to returning all nodes without needing to use an option for it:

```python
for option in self.option_list:
    opt = self.get_option(option[0])
    if opt:
        self.cmd += f" tgt={quote(opt)} tgt_type={option[0]}"
        break  # or don't if we can stack tgt and tgt_type in this salt-run command
return self._get_up_nodes()
```

sos/collector/clusters/saltstack.py
```diff
@@ -0,0 +1,107 @@
...
+    def _get_all_nodes(self) -> list:
+        nodes = []
+        res = self.exec_primary_cmd(self.cmd)
+        if res["status"] != 0:
+            raise Exception("Node enumeration did not return usable output")
+        salt_json_output = json.loads(res["output"].replace("'", '"'))
+        for _, value in salt_json_output.items():
+            nodes.extend(value)
+        if self.get_option("hostnames"):
+            print("Hostname option not implemented for all_nodes, ignoring.")
```
TurboTurtle, line 71: Use the logger here, not `print()`.

sos/collector/clusters/saltstack.py
```diff
@@ -0,0 +1,107 @@
...
+         " compatible with option all_nodes or transport saltstack"),
+        ("list", "", "Filter node list to those matching list"),
+        ("nodegroup", "", "Filter node list to those matching nodegroup"),
+        ("pillar", "", "Filter node list to those with matching pillar"),
+        ("regex", "", "Filter node list to those matching regex"),
+        ("subnet", "", "Filter node list to those in subnet"),
+    ]
+    targeted = False
+
+    def _get_up_nodes(self) -> list:
```
TurboTurtle, line 45: What does "up" mean here in the context of a Saltstack cluster? Literally just on? Reachable on the network? Functioning within the cluster?

I ask because I want to make sure that we're able to get hostnames of nodes that are problematic and thus interesting to look into those sos reports. If a salt node is having issues and we want to collect a report from it and other nodes, I would prefer that we return that node as part of the list of nodes we're collecting from, and then rely on the other code flows to report we could not collect a report from it for XYZ reason.

TurboTurtle:
Also, one note. This is a cluster _profile_ whereas this PR's title is for a _transport_, which already exists.

sos/collector/clusters/saltstack.py
TrevorBenson: > * The default behavior of a cluster profile should be to return all nodes in the cluster, and from there various options should allow reducing/filtering it.

I'll flip the behavior then and make it return every minion, even the offline ones, and make the option for "up_nodes" which are online. This leads to salt errors when a targeted minion does not respond, which is why the default behavior started off as only online nodes.

> * The list, glob, and regex options are redundant with the current behavior of --nodes.

For what transports? The saltstack transport currently only works with `--nodes` in a list format. Admittedly this is probably due to only coding it to use a list format when I wrote the transport. I don't recall noticing the  RemoteTransport having any specific handoff of nodes as one type vs the other and I did not write a parser to determine it in the saltstack transport.

I do see that `--help` mentions nodes can be a regex, but I'm not having any luck with testing that against **control_persist** transport. **Test Environment**: 6 hosts named **store-1** through **store-6**. GLOBS: Using `--nodes store-*`, `--nodes store-\*`, `--nodes 'store-*'`, and REGEX: `--nodes 'store-(1|2)'` where none of these return any nodes in the list to collect from, except the localhost. Unless I'm missing the requirements for it? These exist in my `/etc/hosts`, but are not defined inside `~/.ssh/config` if that is related somehow

> * For hostnames, I'm confused. If we can't resolve the host, why return an FQDN that needs to be resolvable in order to be useful?

Saltstack works via minion ID's unless the administrator makes the minion ID match the hosts shortname or fqdn. This is sometimes the case but with no requirement it often is not. Advanced targeting allows using grains, which then could target based on a hostname value.

I presumed a collector cluster type of saltstack would mostly likely used with the collector transport of saltstack. Therefore I returned the minion ID since that would be what saltstack definitely uses. If the administrator does not use the FQDN as the minion ID, then using it as the default returned value would not work for saltstack transport, but would for others. 

I realize during this conversation I have access to the `self.opts.transport` attribute and could abstract this hostnames tranport option entirely. I will probably adjust this to use if/else `if self.opts.transport == "saltstack"`. **If** the transport is saltstack return the nodes as minion IDs. **Else** parse each node (by minion ID) for it's `fqdn` "grain" and append it to the list of nodes, returning only resolvable (presumably) hostnames.

sos/collector/clusters/saltstack.py
TrevorBenson: Responsive to the cluster (i.e. the salt-minion service is running, minion keys are accepted on the master, etc.). It does not denote the host is offline, because the service could simply be down. If the node is down however we will NOT be able to return its hostname because that is a grain which has to be queried via the minion, through the salt-minion service, which is not responsive on down nodes.

TrevorBenson:
@TurboTurtle I've reworked the logic a little and added a new commit for review, if it looks good will fixup into the original commit to keep the merge clean.


### Options
* Removed options **all_nodes** and **hostnames**
* Added option **minion_id_resolvable**, defaults to False.

### Logic Changes
* If transport is not saltstack and option minion_id_resolvable is False 
  * Returns:
    * The fqdn for up nodes (via grains)
    * The minion ID for down nodes (no grains available).
* If transport is not saltstack and minion_id_resolvable is True
  * Returns: 
    * The minion ID for all nodes
  * Presumes/Requires the minion ID is an fqdn or a uniquely resolvable shortname
* If transport is saltstack or option minion_id_resolvable is True
  * Returns
    * The minion ID for all nodes.

In each case the **node_list** in `./sos_reports/manifest.json` of the collection contains all nodes of the cluster. Either all minion ID, all hostnames, or a combination of both when some nodes are down and **minion_id_resolvable** is not True. A logger warning is issued when attempting to get the fqdn grain and down nodes exist, hopefully providing the user some context about the mix of identity types in the **node_list**.

### Errors Observed

Similar to the control_persist transport error:
```
[store-2:__init__] Unable to open remote session: Timeout expires while trying to connect
```
The saltstack transport errors slightly differently for nodes that are down:
```
[store-2:read_file] Exception while reading /etc/os-release: Timeout expired executing salt --out json --static --no-color store-2 cmd.shell  'cat /etc/os-release'
```

Except for one additional error via `sys.stderr.write()` in `redhat.py`:
```
Could not obtain installed package list[root@EL8_supervisor ~]#
```
It prints once per down node a collection is attempted for. I suspect something is off in the `sos/collector/transport/saltstack.py`, but have not isolated the change to eliminate the attempt to parse a package list yet.

sos/collector/clusters/saltstack.py
TrevorBenson: If no minion ID is explicitly defined, it will default to the systems hostname. Either a fqdn or shortname depending on how the administrator configures the host.

My view on minion id being a hostname being uncommon may very well be skewed due to only seeing around 5-10% clusters I've worked on.  It could easily be the opposite is more common and that a minion ID is almost never explicitly defined and at some form of hostname is the minion id.

sos/collector/clusters/saltstack.py
```diff
@@ -0,0 +1,88 @@
...
+        ("regex", "", "Filter node list to those matching regex"),
+        ("subnet", "", "Filter node list to those in subnet"),
+    ]
+    targeted = False
+
+    def _parse_manage_status(self, output: str) -> list:
+        nodes = []
+        salt_json_output = json.loads(output.replace("'", '"'))
+        for _, value in salt_json_output.items():
+
```
pmoravec, line 46: Why this empty line?

Also, why simple

    nodes = list(salt_json_output.values())

can't work - is there some gotcha I dont see?

sos/collector/clusters/saltstack.py
```diff
@@ -0,0 +1,88 @@
...
+class saltstack(Cluster):
+    """
+    The saltstack cluster profile is intended to be used on saltstack
+    clusters (Salt Project).
+    """
+
+    cluster_name = "Saltstack"
+    packages = ("salt-master",)
+    sos_plugins = ["saltmaster"]
+    cmd = "salt-run --out=pprint manage.status"
```
pmoravec, line 25: Two suggestions, leaving on author to decide:

- move this variable assignment below `option_list`; i.e. declare variables common to all `Cluster` classes first, then declare the `Saltstack` specific ones
- rename the option to e.g. `node_cmd` or `node_list_cmd`; `cmd` is too generic and bit confusing with (non-class instance) `cmd` variable `cmd = fqdn_cmd.format(minion=minion)`.

sos/collector/clusters/saltstack.py
```diff
@@ -0,0 +1,88 @@
...
+    strict_node_list = True
+    option_list = [
+        ("compound", "", "Filter node list to those matching compound"),
+        ("glob", "", "Filter node list to those matching glob pattern"),
+        ("grain", "", "Filter node list to those with matching grain"),
+        ("list", "", "Filter node list to those matching list"),
+        ("minion_id_resolvable", False, "Returns minion IDs in the node list"
+         " for transports other than saltstck. Otherwise only nodes which are"
```
pmoravec, line 33: Typo: saltstck -> saltstack ;-)

pmoravec:
ACK in general, just few minor points.

A curious question to

    class saltstack(Cluster):

why do we use lower letters for children of `Cluster` classes (`satellite`, `ceph`, `pacemaker`, `juju`,..), while almost any other class name starts with a capital letter?

sos/collector/clusters/saltstack.py
TrevorBenson: I didn't test list of lists for validity. 

### Nodes based on the current code
```
['store-2', 'store-3', 'store-4', 'store-5', 'store-0', 'store-1', 'supervisor']
```

### Nodes based on the suggested change 

#### When two keys exist, one for up and one for down nodes:
```
[['store-2', 'store-3', 'store-4', 'store-5'], ['store-0', 'store-1', 'supervisor']]
```

#### When only one key exists as all nodes are up (or down):
```
[['store-0', 'store-1', 'supervisor', 'store-2', 'store-3', 'store-4', 'store-5']]
```


If a list of lists is valid then I'll simplify and remove the for loop.

pmoravec: Ah I see, thanks for explanation.

pmoravec (approved):

sos/collector/clusters/saltstack.py
TurboTurtle: To clarify, this is how `--nodes` operates:

First, we get all the nodes from the cluster that the profile decides to return. From there, any regex entries for `nodes` acts as a filter - so using `--nodes store-(1|2)` would only allow the nodes `store-1` and `store-2` **if** the cluster profile returned them. Of important note here is that this uses python regex, not shell-style globs. So `store-*` would not return what you're probably looking for, but `store-.*` would.

Finally, any non-regex entries for `--nodes` that look like regular hostnames are added on top of any nodes returned by the cluster profile.

To be clear, you cannot use a regex to **add** nodes to the list, since we have no way of validating what that list should expand to - we only use regexes as a filter to reduce the node list.

I've done some brief testing locally and `--nodes` is working as described as far as I can tell, however if there's some condition that may be influencing that, let's investigate and fix that.

I could maybe be persuaded about the `glob` option here, but `list` and `regex` are absolutely redundant of the base `--nodes` option, and I think we should err on the side of a consistent user experience between cluster profiles for these options.

sos/collector/clusters/saltstack.py
TurboTurtle: I think it'd be preferable to try and return *something* for a "down" node that might still be reachable, since ostensibly we're trying to collect sos reports for a problematic condition on at least one host.

If what the cluster returns is not resolvable due to how the end user has configured things, then I don't think there's much we can be expected to do in that case - and we'll just handle the "invalid" hostnames like we do for truly down nodes - a "failed to connect" error which I would expect most end users to be able to piece together "oh, this minion ID isn't directly resolvable and that's why...".

TurboTurtle (commented):
Just noticed the review request on me - I replied in some of the comment threads but didn't make it an actual review >.>;

Anyway, see my comments here:

https://github.com/sosreport/sos/pull/3456#discussion_r1439076177
https://github.com/sosreport/sos/pull/3456#discussion_r1439076495

Once those are addressed I think we'll be good.

sos/collector/clusters/saltstack.py
TrevorBenson: @TurboTurtle 

Your testing of control_persist using a regex "like" `store-(1|2)` worked? As I mentioned this did **not** work in my test when you first mentioned this.

In my test the only node was the localhost doing the collect command:
* Using it bare `--nodes store-(1|2)` causes `-bash: syntax error near unexpected token `('`.
* Using it quoted `--nodes 'store-(1|2)'` results in finding ONLY the localhost:
```
	EL8_supervisor

Press ENTER to continue with these nodes, or press CTRL-C to quit

Connecting to nodes...
Collection would only gather from localhost due to failure to either enumerate or connect to cluster nodes. Assuming single collection from localhost is not desired.
Aborting...
```
Maybe this is a combination effect of using both a `--cluster-type X` and `--nodes` together, but neither the documentation or your earlier comment made it clear that the regex for `--nodes` is only viable when a cluster profile returns `get_nodes()`, but not when `--nodes` is used without a cluster profile?

TurboTurtle: From `man sos-collect`:

--nodes NODES
      Provide  a comma-delimited list of nodes to collect sosreports from, or a **regex string to be used to
      compare _discovered_ node names against.** If using a regex, only nodes matching the regex will be  used
       - i.e. it can be used as a whitelist but not a blacklist.


If the cluster profile returns `store-1`, `store-2`, `store-3`, `store-4` for example, and you use `--nodes='store-(1|2)'` the list would be reduced to just `store-1` and `store-2`. If the cluster profile does **not** return `store-1` or `store-2`, this use of `--nodes` would cause **no** nodes to be returned.

TurboTurtle: Also note, that you generally need to quote regex strings (not just for python/sos, but any tool that has you pass these) as otherwise your shell will try to interpret the characters for its own purposes.

sos/collector/clusters/saltstack.py
TrevorBenson: The last commit persists this behavior to ensure all nodes are part of a manifest. Regardless of up/down all nodes are present in the `manifest.json` list of nodes. The end user can work out whether one minion is not resolvable, or the node is down, etc.

TrevorBenson:
> Just noticed the review request on me - I replied in some of the comment threads but didn't make it an actual review >.>;
> 
> Anyway, see my comments here:
> 
> [#3456 (comment)](https://github.com/sosreport/sos/pull/3456#discussion_r1439076177) [#3456 (comment)](https://github.com/sosreport/sos/pull/3456#discussion_r1439076495)
> 
> Once those are addressed I think we'll be good.

I've resolved the thread about ensuring the down nodes are part of the collection/manifest so they can be diagnosed when missing from the collection. I've left the other question open just to confirm your preference of removal even though they do not conflict, and it would remove 2 native targeting methods from the list and require the user to understand how to implement them via `--nodes`.


---


Separately, I've worked up a branch where the final commit reverses the **minion_id_resolvable** to **minion_id_unresolvable**. It otherwise applies all prior suggestions, but since it makes the behavior identical for all transports and requires the user to set the flag when minion IDs are unresolvable I thought it might be preferred. Here is the [alternate branch URL](https://github.com/TrevorBenson/sos/tree/sos-trevorbenson-collector-cluster-profile-saltstack-default-minion-hostname) in case anyone wants to check the final commit.

@TurboTurtle @pmoravec I'll leave it up to you two whether I cherry-pick include the final commit into this branch after applying any/all other suggestions.

Thanks

pmoravec:
The change in the alternate branch makes sense to me - esp to have uniform behaviour independent on transport.

TurboTurtle:
Alternative branch makes sense to me as well. That plus using `--nodes` for the list/regex node filtering bit would satisfy me for merge. Carrying `glob` is fine.

sos/collector/clusters/saltstack.py
TrevorBenson: :+1: My oversight, I only checked the --help output originally which mentioned support for regex, but not that its a filter against the cluster profile.

TrevorBenson: > Also note, that you generally need to quote regex strings (not just for python/sos, but any tool that has you pass these) as otherwise your shell will try to interpret the characters for its own purposes.

Yep, the first bare usage I didnt expect to work, however I wanted to be thorough in testing what your comment was showing `--nodes store=(1|2)` to be sure I did not overlook something unexpected.

TrevorBenson: @TurboTurtle Whats the final call on the options? Remove them and force the end user to use `--nodes` if they want to use list/regex filters or leave them in place because they do not prevent the end user from using `--nodes` but would still be accessible to end users more familiar with saltstack administration?

Thanks in advance.

TrevorBenson:
> Alternative branch makes sense to me as well. That plus using `--nodes` for the list/regex node filtering bit would satisfy me for merge. Carrying `glob` is fine.

I missed this comment earlier before I requested input in the conversation thread. I'll remove the direct profile options from both in a commit later today so they are only accessible via `--nodes`.

When a PR with suggestions gets approved, should they be a fixup, pushing everything into a single commit/message, or are any unwanted commit messages for each PR trimmed before producing the release notes?

Thanks

sos/collector/clusters/saltstack.py
```diff
@@ -0,0 +1,82 @@
...
+        for status, minions in manage_status.items():
+            if status == "down":
+                self.log_warn(f"Node(s) {minions} are status down.")
+                hostnames.extend(minions)
+            else:
+                for minion in minions:
+                    node_cmd = fqdn_cmd.format(minion=minion)
+                    hostnames.append(
+                        self.exec_primary_cmd(node_cmd)["output"].strip()
+                    )
```
pmoravec, line 59: If the `salt --out=..` command would fail due to whatever reason, the `hostnames` list will get empty string. That empty string seems to be vanished from the node_list somewhere, which is good - ACK to this.

pmoravec (approved):

TurboTurtle:
> When a PR with suggestions gets approved, should they be a fixup, pushing everything into a single commit/message, or are any unwanted commit messages for each PR trimmed before producing the release notes?
> 

We ask that fixup commits be squashed before merging. If a PR has multiple commits that make different, but related, changes that's fine (e.g. adding a new subcommand is usually multiple commits in a single PR as a result of building out the functionality).

TrevorBenson:
> > When a PR with suggestions gets approved, should they be a fixup, pushing everything into a single commit/message, or are any unwanted commit messages for each PR trimmed before producing the release notes?
> 
> We ask that fixup commits be squashed before merging. If a PR has multiple commits that make different, but related, changes that's fine (e.g. adding a new subcommand is usually multiple commits in a single PR as a result of building out the functionality).

Cool. Once you've approved I'll go ahead and squash (instead of fixup) in preparation for the merge, presuming you didn't mean the "squash and merge" via the PR button.

TurboTurtle (approved):
Ack, pending commit squashing.

TrevorBenson:
> Ack, pending commit squashing.

Doing this now. Just to confirm, by squash vs. fixup you mean to squash the 5 together and leave all 5 comments? Or did you mean squash all 5 together and remove the last 4 comments/sign-offs (essentially a rebase using "fixup" style so only 1 commit message remains)?

TurboTurtle:
The latter. There should be 1 commit, with whatever message you'd like to describe it with.

Just to be clear, this isn't to say all PRs need to be a single commit. It's certainly acceptable to have a PR have multiple related commits, we just want to have fixup commits like "apply review suggestions" or "fixed some typos" squashed into their relevant parent commits.

TrevorBenson:
> The latter. There should be 1 commit, with whatever message you'd like to describe it with.
> 
> Just to be clear, this isn't to say all PRs need to be a single commit. It's certainly acceptable to have a PR have multiple related commits, we just want to have fixup commits like "apply review suggestions" or "fixed some typos" squashed into their relevant parent commits.

Gotcha! I think we just got on different tracks with the term "fixup" originally. While it's often used to refer to a commit to fix the prior issue or suggestion, I was actually meaning the rebase fixup, which natively does the squash but drops all the comments automatically except for the commit being squashed (fixup'd) into. :+1:

TrevorBenson:
Commits squashed/fixup'd.

PR #3467, opened 2024-01-10
[tests] Drop obsolete TestCase.assertEquals()

martinpitt:
`unittest.TestCase.assertEquals()` was deprecated in Python 3 [1], and finally dropped in Python 3.12. This now causes the unit tests to fail [2].

[1] https://docs.python.org/3.5/library/unittest.html#deprecated-aliases
[2] https://bugs.debian.org/1058214

---

### DIFF
--- a/tests/report_tests/timeout/timeout_tests.py
+++ b/tests/report_tests/timeout/timeout_tests.py
@@ -20,7 +20,7 @@ class PluginTimeoutTest(StageTwoReportTest):
 
     def test_correct_plugin_timeout(self):
         man = self.get_plugin_manifest('timeout_test')
-        self.assertEquals(man['timeout'], 10)
+        self.assertEqual(man['timeout'], 10)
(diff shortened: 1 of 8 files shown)


martinpitt:
I was wondering why you didn't notice that long ago already -- but it seems neither "breakage test python latest" nor the RPM builds don't actually run the unit tests. I suggest to enable them in both places.

pmoravec (approved):

TurboTurtle (approved):

TurboTurtle:
> I was wondering why you didn't notice that long ago already -- but it seems neither "breakage test python latest" nor the RPM builds don't actually run the unit tests. I suggest to enable them in both places.

The unit tests are run in our "Stage One" jobs. The "breakage" test is literally just "do we fall flat on our face when execting with newer python versions?"

In any event, LGTM.

PR #3470, opened 2024-01-11
[convert2rhel] Capture information about pre-conversion

jcastill:
Capture the txt and json files generated when analyzing the system for pre-conversion to RHEL.

Related: RH: RHEL-21331

---

### DIFF
--- a/sos/report/plugins/convert2rhel.py
+++ b/sos/report/plugins/convert2rhel.py
@@ -23,6 +23,7 @@ class convert2rhel(Plugin, RedHatPlugin):
             "/var/log/convert2rhel/convert2rhel.log",
             "/var/log/convert2rhel/archive/convert2rhel-*.log",
             "/var/log/convert2rhel/rpm_va.log",
+            "/var/log/convert2rhel/convert2rhel-pre-conversion.*",
         ])
 
 


sos/report/plugins/convert2rhel.py
```diff
@@ -23,6 +23,7 @@ def setup(self):
             "/var/log/convert2rhel/convert2rhel.log",
             "/var/log/convert2rhel/archive/convert2rhel-*.log",
             "/var/log/convert2rhel/rpm_va.log",
+            "/var/log/convert2rhel/convert2rhel-pre-conversion.*"
```
arif-ali, line 26: minor nit, is it worth adding a comma at the end of the line?

arif-ali (approved):

sos/report/plugins/convert2rhel.py
pmoravec: .. to prevent potential future error when one adds a new copy spec (well in fact some CI test error, only, if I am right).

Good point, I like prevention!

arif-ali: it just means that when we add a new line, we're only changing 1 line, and not 2 lines in the same way here ;)

TurboTurtle (approved):
Ack. I've restarted the CI jobs.

PR #3474, opened 2024-01-12
[rhui] Added the rhui-installer directory to rhui sos plugin

taftsanders:
Fixes: https://issues.redhat.com/browse/RHEL-21490

---

  - Please advise what ^ this means.

### DIFF
--- a/sos/report/plugins/rhui.py
+++ b/sos/report/plugins/rhui.py
@@ -28,6 +28,7 @@ class Rhui(Plugin, RedHatPlugin):
             "/var/cache/rhui/*",
             "/root/.rhui/*",
             "/var/log/rhui/*",
+            "/var/log/rhui-installer/*",
         ])
         # skip collecting certificate keys
         self.add_forbidden_path("/etc/pki/rhui/**/*.key")


taftsanders:
@pmoravec please review and advise for proper PR formatting.

This was done with collaboration from jvlcek@redhat.com

jcastill:
@taftsanders the DCO test is failing because the signed off by is missing. You'll need to use 'git commit -s' to sign it

jcastill (approved):

sos/report/plugins/rhui.py
```diff
@@ -27,7 +27,7 @@ def setup(self):
             "/var/log/rhui-subscription-sync.log",
             "/var/cache/rhui/*",
             "/root/.rhui/*",
-            "/var/log/rhui/*",
+            "/var/log/rhui*/*",
```
pmoravec, line 30: I would prefer explicitly listing individual directories. First, to ensure `sos` is collecting just what we need (and not e.g. "rhui-backup" directory a user created years ago). Second, to ensure the default limit 25MB of logfiles is applied to `rhui-installer` and to `rhui` directories *independently*.

Assume a case when `rhui` directory has >25MB logs. Then `rhui-installer` logs wont be collected (with the current PR), since when `sos` must truncate logs from one copy_spec, it truncates the oldest logs - what will be those from install time.

So I expect


    +            "/var/log/rhui-installer/*",

should be the patch.

taftsanders: @pmoravec made the change to the original line and added the new line for the `/var/log/rhui-installer/*` directory.

taftsanders:
@jcastill DCO still appears to fail even with the signed off by section in the new commit. Is it failing because the old commit is still present?

jcastill:
@taftsanders it's probably because the old commit is still there without the signature. Squash the last one on the previous one, then amend the signature, and push again.

TurboTurtle (approved):
Ack.

PR #3480, opened 2024-01-17
[testing] Add testing via tox.ini

arif-ali:
Consolidate testing, so that things can be run from CLI, CI and various other means in a consistent way by using tox

---

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -57,7 +57,10 @@ flake8_task:
     name: "Flake8 linting test"
     container:
         image: alpine/flake8:latest
-    flake_script: flake8 sos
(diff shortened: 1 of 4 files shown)


arif-ali:
Just an initial draft to try and getting testing via `tox`, so that anyone picking up any bugs/fixes can see how to run tests. This may still require some work, but an initial start to get things potentially moving

arif-ali:
This isn't going to work for us (as much as I wanted it to), as this creates a venv, and with some rigorous testing, was not able to get it to use the system site packages.

So, using `tox` won't work for us :(. If I have time later in the year, I might revisit this

PR #3483, opened 2024-01-18
[containerd] obtain CNI config files

### LINKED ISSUE #3482: Obtain CNI files for containerd
Containerd uses the CNI configuration present in the defined folders by the configuration

```
   [plugins."io.containerd.grpc.v1.cri".cni]
      conf_dir = "/etc/cni/net.d
```

It will be very useful to obtain the cni configurations present on the folder for debugging networking related problems 

https://github.com/sosreport/sos/blob/b94ced8370824bd62f3c7573ae33fcb96c5da531/sos/report/plugins/containerd.py#L12-L28

aojea:
containerd uses CNI to configure the container's network.

The CNI uses different configuration files that by default are installed in `/etc/cni/net.d`, these files are important to understand how the network is configured.

Closes #3482
---

### DIFF
--- a/sos/report/plugins/containerd.py
+++ b/sos/report/plugins/containerd.py
@@ -19,6 +19,7 @@ class Containerd(Plugin, RedHatPlugin, UbuntuPlugin, CosPlugin):
     def setup(self):
         self.add_copy_spec([
             "/etc/containerd/",
+            "/etc/cni/net.d/",
         ])
 
         self.add_cmd_output('containerd config dump')


sos/report/plugins/containerd.py
```diff
@@ -19,6 +19,7 @@ class Containerd(Plugin, RedHatPlugin, UbuntuPlugin, CosPlugin):
     def setup(self):
         self.add_copy_spec([
             "/etc/containerd/",
+            "/etc/cni/net.d/",
```
jcastill, line 22: Just to be on the safe side - we don't have any data to obfuscate in the config files in that directory, right?

aojea: only ips and network  parameters
```
 cat /etc/cni/net.d/10-kindnet.conflist

{
        "cniVersion": "0.3.1",
        "name": "kindnet",
        "plugins": [
        {
                "type": "ptp",
                "ipMasq": false,
                "ipam": {
                        "type": "host-local",
                        "dataDir": "/run/cni-ipam-state",
                        "routes": [
                               { "dst": "::/0" }
                        ],
                        "ranges": [
                                [ { "subnet": "fd00:10:244::/64" } ]
                        ]
                }
                ,
                "mtu": 1500
        },
        {
                "type": "portmap",
                "capabilities": {
                        "portMappings": true
                }
        }
        ]
}

```

the spec does not have credentials or something like this https://github.com/containernetworking/cni/blob/main/SPEC.md#example-configuration

jcastill: Perfect then, thank you!

jcastill (approved):

pmoravec (approved):

TurboTurtle (approved):

PR #3489, opened 2024-01-22
[aide] Add Aide support

adam-vest:
---

### DIFF
--- a/sos/report/plugins/aide.py
+++ b/sos/report/plugins/aide.py
@@ -1,0 +1,56 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+
+
+class Aide(Plugin):
+
+    short_desc = 'Advanced Intrusion Detection Environment'
+
+    plugin_name = "aide"
+    profiles = ('system', 'security')
+
+    packages = ('aide',)
+    conf_file = "/etc/aide/aide.conf"
+
+    def setup(self):
+        self.add_cmd_output(f"aide -c {self.conf_file} --config-check")
+
+        if self.get_option("all_logs"):
+            self.add_copy_spec([
+                "/var/log/aide/",
+            ])
+        else:
+            self.add_copy_spec([
+                "/var/log/aide/aide.log"
+            ])
+
+
+class RedHatAide(Aide, RedHatPlugin):
+    conf_file = "/etc/aide.conf"
+
+    def setup(self):
+        super(RedHatAide, self).setup()
+        self.add_copy_spec([
+            "/etc/aide.conf",
+            "/etc/logrotate.d/aide"
+        ])
+
+
+class DebianAide(Aide, DebianPlugin, UbuntuPlugin):
+    conf_file = "/etc/aide/aide.conf"
+
+    def setup(self):
+        super(DebianAide, self).setup()
+        self.add_copy_spec([
+            "/etc/aide/",
+            "/etc/default/aide"
+        ])
+
+# vim: et ts=4 sw=4


sos/report/plugins/aide.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+class RedHatAide(Aide, RedHatPlugin):
+    def setup(self):
+        super(RedHatAide, self).setup()
+        self.add_copy_spec([
+            "/etc/aide.conf",
+            "/etc/logrotate.d/aide"
+        ])
+        self.add_cmd_output("aide -c /etc/aide.conf --config-check")
```
jcastill, line 39: This will generate an empty output and empty file if there are no inconsistencies in the database, right?

jcastill (approved):
Looks good in RHEL

sos/report/plugins/aide.py
adam-vest: Hey! It's only doing a linting of the config file, not anything with the database (that's `--check, -C`), but you are right that if the config file has no syntax errors then the output and resultant file will be empty.

arif-ali:
It would be neater for the `conf_file` to be defined in the respective classes, and then call the command to collect in the main cleass? wdyt?

```python
self.add_cmd_output(f"aide -c {self.conf_file} --config-check")
```

adam-vest:
Hey Arif! Do you mean like:
```
class Aide(Plugin):
...
    def setup(self):
        self.add_cmd_output(f"aide -c {self.conf_file} --config-check")
...
class RedHatAide(Aide, RedHatPlugin):
    conf_file = "/etc/aide.conf"
...
class DebianAide(Aide, DebianPlugin, UbuntuPlugin):
    conf_file = "/etc/aide/aide.conf"
```
I can see a certain logic behind going that route. It's one more line of code, but I think I do agree that it can be neater split up that way.

arif-ali:
@adam-vest yeah, something like that.

This also means that if we add more command collections in the future, we just add/update it to the common class, and it can then manage both scenarios

sos/report/plugins/aide.py
```diff
@@ -0,0 +1,54 @@
...
+        super(RedHatAide, self).setup()
+        self.add_copy_spec([
+            "/etc/aide.conf",
+            "/etc/logrotate.d/aide"
+        ])
+
+
+class DebianAide(Aide, DebianPlugin, UbuntuPlugin):
+    conf_file = "/etc/aide/aide.conf"
+
```
TurboTurtle, line 46: flake8 linting check is failing here with a whitespace.

TurboTurtle (commented):
LGTM after the flake8 alert is resolved. My only other comment would have also been to set `conf_file` as a class attr. Glad that's already been done.

sos/report/plugins/aide.py
adam-vest: Gah :man_facepalming: missed that tiny little squiggly in vscode. Fixed!

adam-vest:
Not sure why the tests failed, but I'm guessing it has something to do with this line:
```
self.add_cmd_output(f"aide -c {self.conf_file} --config-check")
```

I ran that on a Bionic machine and it seemed to work fine? It looks like the standard way that other plugins might do this would be like:
```
self.add_cmd_output("aide -c %s --config-check" % self.conf_file)
```

...which I've confirmed also works fine. Would that satisfy whatever check failed?

arif-ali:
@adam-vest I re-ran the test, and it went through without any problems, I think it was a one-off. I checked the previous runs too, and it had gone through, and was unrelated to your change.

Final thing is, if you got squash your commits then that would be awesome

adam-vest:
Ugh, I'm making it worse. idk how to "squash" these commits properly. I'll try to fix whatever I just did.

adam-vest:
OK...I think/hope that's all cleaned up now...sorry for all the trouble!

TurboTurtle:
I'm not entirely sure what, but something seems to have gone sideways as the second commit pulls in changes that are already in `main`.

Luckily this PR allows maintainer edits, so I should be able to fix this up on my end and force-push over your branch, and then merge this. I've not tried a force-push on a fork branch before though, so hopefully that works. If not I'll write out the steps to fix this so you can do them locally. Give me a little bit to finish up some work stuff and I'll get to straightening this out later today.

TurboTurtle:
Ok, looks like pushing to your fork worked. :tada: 

I think everything looks fine now, we'll let the CI run and then just need an ack from @arif-ali or @pmoravec.

adam-vest:
Good gravy...thanks for taking care of that @TurboTurtle , and sorry again heh

TurboTurtle:
No worries, at the end of the day sometimes it boils down to Git Is Weird(tm).

sos/report/plugins/aide.py
```diff
@@ -0,0 +1,56 @@
...
+
+class RedHatAide(Aide, RedHatPlugin):
+    conf_file = "/etc/aide.conf"
+
+    def setup(self):
+        super(RedHatAide, self).setup()
+        self.add_copy_spec([
+            "/etc/aide.conf",
+            "/etc/logrotate.d/aide"
```
pmoravec, line 42: We might add `add_forbidden_path` for `logrotate` plugin for this file to prevent both plugins to collect this file, but.. assuming this file won't ever rise to MBs in size, there is no technical need.

arif-ali (approved):

PR #3490, opened 2024-01-23
[ssh] Disable option 'userconfs' by default

jcastill:
Having this option enabled by default
can cause problems when users have their
home directories on automount. Leave the
option disabled by default so if needed
we can enable it in the command line.

Related: RHEL-22389

---

### DIFF
--- a/sos/report/plugins/ssh.py
+++ b/sos/report/plugins/ssh.py
@@ -19,7 +19,7 @@ class Ssh(Plugin, IndependentPlugin):
     profiles = ('services', 'security', 'system', 'identity')
 
     option_list = [
-        PluginOpt('userconfs', default=True, val_type=str,
+        PluginOpt('userconfs', default=False, val_type=str,
                   desc=('Changes whether module will '
                         'collect user .ssh configs'))
     ]


jcastill:
> Personally, changing a default behaviour can give unexpected behaviour, especially if we're expecting it to be there. My hunch here is that we leave this as is. As it typically helps to debug wrong permissions in `.ssh`, and it is very rare (at least in my experience) to find an environment that we're supporting that has automount.
> 

On the other hand, I have seen a couple of cases where this has happened to customers in the last month. So there's that.  We already have an unexpected behaviour because of having this enabled by default.

> If we could tackle this in a different way, I would rather go in that direction.
> 

I'm open to any suggestions.

jcastill:
> We could change the default of the plugin opt to our standard `auto/on/off` approach, and if `auto`, then we perform a check to see if autofs is in use for homedirs (perhaps easier said than done? I'm just spitballing here), and if it is we don't perform the collection. This allows `off/False` and `on/True` to be used by end users as desired.

I think this may work! Checking if home dirs are on autofs may be tricky, but I think I have an idea about how to do it.

pmoravec:
Just to clarify:

- the option should have `on`/`auto`/`off` values, with auto as the default
- value `on` means current behaviour (collect dirs of all users)
- value `off` means no dirs of users are collected
- value `auto` means just dirs not on automount will be collected

Let me know if I understood it wrongly.

TurboTurtle:
> * value `auto` means just dirs not on automount will be collected
> 

Not exactly. `auto` means "check if we're using autofs for home dirs, if we are behave as if this option is set to `False`, else (if we are not using autofs and only have locally mounted user homedirs) behave as if it is set to `True`".

arif-ali:
Any further thoughts on this, maybe we can merge this as is, and then look at the option of autofs issue and collecting this info?

jcastill:
This should be solved by the new approach in #3807 , so I'll close it.

PR #3491, opened 2024-01-23
[component] Check loggers attributes before referencing them

pmoravec:
When using own preset, loggers verbosity is set even before they are configured. In such a case, we must guard accessing them properly.

Resolves: #3491
Relevant: RHEL-22395

---

### DIFF
--- a/sos/component.py
+++ b/sos/component.py
@@ -402,15 +402,13 @@ class SoSComponent():
         self.ui_log.addHandler(ui_console)
 
     def set_loggers_verbosity(self, verbosity):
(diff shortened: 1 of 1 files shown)


pmoravec:
Reproducer: have some preset in `/etc/sos/presets.d/` and use it:

    # sos report --preset sostestpreset --batch
    Could not initialize 'report': 'SoSReport' object has no attribute 'flog'
    #

arif-ali (approved):

TurboTurtle (approved):

PR #3495, opened 2024-01-26
[vectordev] Add new vectordev plugin

dnegreira:
---

### DIFF
--- a/sos/report/plugins/vectordev.py
+++ b/sos/report/plugins/vectordev.py
@@ -1,0 +1,42 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, IndependentPlugin
+
+
+class VectorDev(Plugin, IndependentPlugin):
+
+    short_desc = 'A tool for building observability pipelines'
+
+    plugin_name = "vectordev"
+    profiles = ('observability',)
+    files = ('/etc/vector/',)
+
+    def setup(self):
+        self.add_copy_spec([
+            "/etc/vector/"
+        ])
+
+    def postproc(self):
+
+        vector_config_path = "/etc/vector/*"
+        protect_keys = [
+            "auth.password",
+            "auth.token",
+            "tls.key_pass",
+        ]
+
+        # Redact yaml and ini style "key (:|=) value".
+        keys_regex = r"(^\s*(%s)\s*(:|=)\s*)(.*)" % "|".join(protect_keys)
+        sub_regex = r"\1*********"
+        self.do_path_regex_sub(vector_config_path, keys_regex, sub_regex)
+        # Redact certificates
+        self.do_file_private_sub(vector_config_path)
+
+
+# vim: et ts=4 sw=4


sos/report/plugins/vectordev.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class Vectordev(Plugin):
+
+    short_desc = 'A lightweight, ultra-fast tool for building observability pipelines'
+
+    plugin_name = "vectordev"
+    profiles = ('observability',)
+
+    conf_file = "/etc/vector/vector.yaml"
```
jcastill, line 19: Did you define this variable to be used in the future? I don't see it used anywhere yet. Perhaps you wanted ot use it in add_copy_spec() below?

sos/report/plugins/vectordev.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def setup(self):
+        self.add_copy_spec([
+            "/etc/vector/"
+        ])
+
+
+class UbuntuVectordev(Vectordev, UbuntuPlugin):
+    conf_file = "/etc/vector/aggregator/vector.yaml"
```
jcastill, line 28: Same as above. Is this necessary if you are capturing the whole /etc/vector?

jcastill (commented):
Two very easy

sos/report/plugins/vectordev.py
dnegreira: Er, I confused `conf_file` with `files` :facepalm: 

Good point, should I rewrite this as an `IndependentPlugin` ?

For the time being we just need to capture /etc/vector/ config files and nothing else, we can always change it in the future for distro specific. WDYT?

dnegreira: > Another question - are there any containers or packages we should watch for? Like we do with microshift, i.e.:

Not for our specific use case, as we only get a binary and create the configuration files on `/etc/vector/`

sos/report/plugins/vectordev.py
```diff
@@ -0,0 +1,30 @@
...
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, UbuntuPlugin
+
+
+class Vectordev(Plugin):
```
pmoravec, line 12: What the `dev` stands for? vector.**dev** URL of the project? I would prefer either class name `Vector` or `VectorDev` - until "vectordev" is treated as one English word (totally non-English native speaker here, so I can't judge).

sos/report/plugins/vectordev.py
pmoravec: Re `IndependentPlugin`: per https://vector.dev/download/ , multiple OSes are supported, so this makes sense.

sos/report/plugins/vectordev.py
```diff
@@ -0,0 +1,30 @@
...
+    short_desc = 'A lightweight, ultra-fast tool for building observability pipelines'
+
+    plugin_name = "vectordev"
+    profiles = ('observability',)
+
+    conf_file = "/etc/vector/vector.yaml"
+
+    def setup(self):
+        self.add_copy_spec([
+            "/etc/vector/"
```
pmoravec, line 23: No credentials or secret keys there, I assume..?

sos/report/plugins/vectordev.py
dnegreira: I didn't want to call it Vector since it sounds -too- generic, as the word can be applied to a myriad of things.
Fixed on 5bd2d563d586b4fc9d02f2bcca49cb81b7b493cf

sos/report/plugins/vectordev.py
dnegreira: Not on our use case, as we don´t  require authentication to our sinks, but the configuration of `vector` allows a bunch of different authentications to destinations, each with their own set of different credential names and settings.
I'm not sure how to deal with this for the time being, but any comment is welcome.

TurboTurtle: Looking at the docs, it seems like these are done via INI configs? If so, I think we can take a reasonable stab at a generic `postproc` for that format, and use some globs for hitting the typical keys like `pass*`, `key`, `token`, etc...

It won't get everything most likely, but it should at least provide _some_ measure of confidence. We'll have to rely on bug reports from there.

sos/report/plugins/vectordev.py
```diff
@@ -0,0 +1,24 @@
...
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin
+
+
+class VectorDev(Plugin):
```
pmoravec, line 12: I think you must use `IndependentPlugin` instead of the base class `Plugin`, to allow the plugin on any distro.

dnegreira: yep, fixed now.

sos/report/plugins/vectordev.py
dnegreira: I captured a couple of keys that seemed more common like: 
"auth.password",
"auth.token",
"tls.key_pass",

and as vector allows both .init and .yaml style configs I adapted the regepx to capture both

sos/report/plugins/vectordev.py
```diff
@@ -0,0 +1,42 @@
+i# This file is part of the sos project: https://github.com/sosreport/sos
```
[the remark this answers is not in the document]
dnegreira: :man_facepalming:

arif-ali (approved):

pmoravec (approved):

Issue #3403, opened 2023-11-03
[cleaner] Skip obfuscation of too short (sub)strings


pmoravec:
We should make cleaner bit more clever to prevent literal obfuscation of some use cases like:

- too short strings (`map`, `sos`, `vir`, `os`,..): obfuscating those is misleading (the strings appear on many places where they should not be scrubbed), sometimes not helping (`host3_commands` directory is technically obfuscated, but any human understands what `host3` hides), often redundant (as the strings are so short that their informative value we try to obfuscate is practically none)
- substrings of other strings; e.g. `vir` should not be "hidden" in `environment` filename

I suggest - but am open to other opinions - to:
- don't treat strings of len<4 (or just len<3?) at all - *just* for parsers: hostname(+domainname), usernames (what about user defined keywords? imho let user to decide what to scrub)
- obfuscate just whole strings, not substrings; again this applies just to the hostname + username parsers, probably also to keywords one. But what should be boundaries of a string? For a secret item `secret`, should we apply the mapping to `secretary`, `my_secret`, `your:secret`,..?

Any thoughts to the length of "dont scrub words"? Or to what parsers to relax (in either case)? Or what boundaries of substrings to use?

(technically, the change should be here: https://github.com/sosreport/sos/blob/main/sos/cleaner/mappings/__init__.py#L85 and also https://github.com/sosreport/sos/tree/main/sos/cleaner/parsers/hostname_parser.py#L21)

pmoravec:
Until there is an objection, I will create a PR that will - just for hosts+domains+usernames :

- skip obfuscating all strings of length<4
- obfuscate whole strings only ; "string" means alphanumeric only letters here (so for obfuscating `secret`, `secretary` won't be scrubbed, while `my_secret` and `your:secret` will be hidden)

Let me know if you have some other opinion within 2 weeks.

PR #3496, opened 2024-01-28
[cleaner] skip shortstrings and substrings

pmoravec:
PR for https://github.com/sosreport/sos/issues/3403 . For relevant mappings (hostname, username, partly keyword), skip obfuscation of strings <4 length and skip obfuscation of substrings inside words.

Reproducer to play with:

    echo "repo" >> /etc/cron.deny
    echo "map" >> /etc/cron.deny
    echo "private" >> /etc/cron.deny
    rm -rf /etc/sos/cleaner/default_mapping
    sos report -o cron --batch --build --clean
    ..
    A mapping of obfuscated elements is available at
    	    /var/tmp/sosobfuscateduser2rt-pmoravec-rhel8-2024-01-28-ditdybr-obfuscateduser1_obfuscateduser0

while it should be:

        /var/tmp/sosreport-pmoravec-rhel8-2024-01-28-ditdybr-obfuscateduser0_map

---

### DIFF
--- a/sos/cleaner/parsers/__init__.py
+++ b/sos/cleaner/parsers/__init__.py
@@ -151,7 +151,8 @@ class SoSCleanerParser():
         if self.compile_regexes:
             for item, reg in self.mapping.compiled_regexes:
                 if reg.search(string_data):
-                    string_data = reg.sub(self.mapping.get(item), string_data)
+                    string_data = reg.sub(self.mapping.get(item.lower()),
+                                          string_data)
         else:
             for k, ob in sorted(self.mapping.dataset.items(), reverse=True,
                                 key=lambda x: len(x[0])):
--- a/sos/cleaner/mappings/__init__.py
+++ b/sos/cleaner/mappings/__init__.py
@@ -26,6 +26,7 @@ class SoSMap():
     skip_keys = []
     compile_regexes = True
     ignore_short_items = False
+    match_full_words_only = False
 
     def __init__(self):
         self.dataset = {}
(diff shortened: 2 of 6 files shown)


pmoravec:
A few TODO points:

- `pep8` fails on `W605 invalid escape sequence '\W'`  (and `\A` and `\g`). I dont see an elegant way how to bypass the error
- I should add some tests (advices *what* particular to test are welcomed)

sos/cleaner/parsers/__init__.py
```diff
@@ -151,7 +154,10 @@ def parse_string_for_keys(self, string_data):
         if self.compile_regexes:
             for item, reg in self.mapping.compiled_regexes:
                 if reg.search(string_data):
-                    string_data = reg.sub(self.mapping.get(item), string_data)
+                    repl = self.mapping.get(item.lower())
+                    if self.mapping.match_full_words_only:
+                        repl = f'\g<BEGIN>{repl}\g<END>'
```
jcastill, line 159: Would it work if you make it like this?

rf'\g<BEGIN>{repl}\g<END>'

pmoravec:
Good advice, I didnt know I can combine `r` and `f` in string prefixes!

Now just some tests are missing / pending on me.

pmoravec:
The test fails to obfuscate two secrets separated by space, like:

    testuser1 a testuser1 b testuser1 testuser1 c testuser1 testuser1 testuser1 testuser1 testuser1

is obfuscated to:

    obfuscateduser2 a obfuscateduser2 b obfuscateduser2 testuser1 c obfuscateduser2 testuser1 obfuscateduser2 testuser1 obfuscateduser2

But why..?

jcastill:
Looks like it's equalling "testuser1 c" to "testuser1 testuser1" so the second element in the regex is not obfuscated, right? What about obfuscating in your string (testuser1)+ instead of just testuser? Could that cover it?

TurboTurtle (approved):
We could potentially make this configurable for how short of string we skip, but for now I think this looks good.

pmoravec:
Having that configurable makes sense, as lower prio. I have a TODO here to write an avocado test - I hope I write it before 4.7.0 cut but if not, no big deal.

pmoravec:
> Having that configurable makes sense, as lower prio. I have a TODO here to write an avocado test - I hope I write it before 4.7.0 cut but if not, no big deal.

Why I was procrastinating to write such trivial unit cases? :) TODO->DONE.

Issue #3497, opened 2024-01-31
add /run/mount/utab


karelzak:
It would be nice to also have the file /run/mount/utab in the mount-related files collection. The file is libmount private (and optional) but may provide important information as it stores userspace-specific mount data. Thanks :-)

jcastill:
That's a very good idea. I'll send a PR for this.

TurboTurtle:
Closing this out as #3498 is merged and addresses this.

PR #3498, opened 2024-01-31
[filesys] Capture /run/mount/utab

jcastill:
The file is libmount private (and optional)
but may provide important information as it stores userspace-specific mount data

Related: #3497

---


arif-ali (approved):

karelzak:
utab (and /etc/fstab) is readable for everyone, so using fstab to store any password is a pretty bad idea ...

The utab file is used mostly for mount x-* options (it means not all mount options go there), and mount(8) does not use it for sensitive data. The another user is mount.nfs, but it also does not use it for sensitive data.

So, from my point of view, it's safe.

TurboTurtle (approved):

PR #3499, opened 2024-02-01
[networkmanager] Add "NetworkManager --print-config" command

hiroyukiyasuharajp:
There are some configuration files which could affect NetworkManager's behaviors. "NetworkManager --print-config" is useful to verify the current configuration at a glance.

---

### DIFF
--- a/sos/report/plugins/networkmanager.py
+++ b/sos/report/plugins/networkmanager.py
@@ -31,6 +31,8 @@ class NetworkManager(Plugin, RedHatPlugin, UbuntuPlugin):
 
         self.add_journal(units="NetworkManager")
 
+        self.add_cmd_output("NetworkManager --print-config")
+
         # There are some incompatible changes in nmcli since
         # the release of NetworkManager >= 0.9.9. In addition,
         # NetworkManager >= 0.9.9 will use the long names of


sos/report/plugins/networkmanager.py
```diff
@@ -19,6 +20,10 @@ class NetworkManager(Plugin, RedHatPlugin, UbuntuPlugin):
     packages = ('NetworkManager', 'network-manager')
 
     def setup(self):
+        nm_pkg = self.policy.package_manager.pkg_by_name('NetworkManager')
+        if nm_pkg is None:
+            return
+
```
arif-ali, line 26: This section is not required. The plugin will only come into use if the items in the `packages` tuple

sos/report/plugins/networkmanager.py
```diff
@@ -31,6 +36,10 @@ def setup(self):
 
         self.add_journal(units="NetworkManager")
 
+        nm_ver = '.'.join(nm_pkg['version'])
+        if parse_version(nm_ver) >= parse_version('1.2'):
```
arif-ali, line 40: You don't really need to check for the version either. If the command doesn't work, then that's ok, it will not create the file

arif-ali (changes requested):

sos/report/plugins/networkmanager.py
hiroyukiyasuharajp: Understood. I removed the section.

sos/report/plugins/networkmanager.py
hiroyukiyasuharajp: Understood, too. I removed the version check, thanks.

arif-ali (approved):

TurboTurtle (approved):

Issue #3500, opened 2024-02-01
[ubuntu][microk8s] Extend kubernetes options to microk8s plugin


phvalguima:
Hi,

I am currently using microk8s with SOS report as part of my testing. However, I have no options to automatically capture logs at the same level of details as `kubernetes` plugin.

In particular, I am interested in capturing pod-level logs, like `kubernetes.podlogs=true`, and information about all namespaces.

I see two alternatives:
1) Turning `microk8s` a subclass of `UbuntuKubernetes`: in this case, all options in Kubernetes plugin would be automatically available
2) Adding `/var/snap/microk8s/current/credentials/client.config` to `UbuntuKubernetes` `files` tuple

I am more inclined to doing option (1), as I also want to have some microk8s-related information, besides containers only.

phvalguima:
@dnegreira hi, I am tagging you in this issue as I see you've committed the first version of microk8s plugin

dnegreira:
Hey @phvalguima - We should extend the kubernetes plugins to gather the log pods from microk8s if something is missing there. The intent of microk8s plugin is to gather all the micork8s bits like microk8s config, microk8s status.

LMK if you need help with this.

PR #3501, opened 2024-02-01
[Kubernetes] Extend plugin to support microk8s

phvalguima:
Currently, we are not supporting microk8s with the
same options as provided by kubernetes. This PR
extends the UbuntuKubernetes to get information
from microk8s.

Following the review, it also extends microk8s plugin
to add the credentials file in the copy spec and
runs the postproc() routine.

Closes: https://github.com/sosreport/sos/issues/3500

---

- [X ] Are any related Issues or existing PRs [properly referenced](https://docs.github.com/en/issues/tracking-your-work-with-issues/creating-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) via a Closes (Issue) or Resolved (PR) line?

### DIFF
--- a/sos/report/plugins/kubernetes.py
+++ b/sos/report/plugins/kubernetes.py
@@ -176,12 +176,12 @@ class Kubernetes(Plugin):
         # like "pass", "pwd", "key" or "token"
         env_regexp = r'(?P<var>{\s*"name":\s*[^,]*' \
             r'(pass|pwd|key|token|cred|PASS|PWD|KEY)[^,]*,\s*"value":)[^}]*'
-        self.do_cmd_output_sub('kubectl', env_regexp,
+        self.do_cmd_output_sub(self.kube_cmd, env_regexp,
                                r'\g<var> "********"')
 
         # Next, we need to handle the private keys and certs in some
         # output that is not hit by the previous iteration.
-        self.do_cmd_private_sub('kubectl')
+        self.do_cmd_private_sub(self.kube_cmd)
 
 
 class RedHatKubernetes(Kubernetes, RedHatPlugin):
@@ -211,7 +211,8 @@ class UbuntuKubernetes(Kubernetes, UbuntuPlugin, DebianPlugin):
     packages = ('kubernetes',)
     files = (
         '/root/cdk/cdk_addons_kubectl_config',
-        '/etc/kubernetes/admin.conf'
+        '/etc/kubernetes/admin.conf',
+        '/var/snap/microk8s/current/credentials/client.config',
     )
 
     services = (
@@ -227,6 +228,9 @@ class UbuntuKubernetes(Kubernetes, UbuntuPlugin, DebianPlugin):
 
         for svc in self.services:
             self.add_journal(units=svc)
+
+        if self.is_installed('microk8s'):
+            self.kube_cmd = 'microk8s kubectl'
 
         super(UbuntuKubernetes, self).setup()
 
--- a/sos/report/plugins/microk8s.py
+++ b/sos/report/plugins/microk8s.py
@@ -40,6 +40,9 @@ class Microk8s(Plugin, UbuntuPlugin):
             'status',
             'version'
         ]
+        self.add_copy_spec(
+            "/var/snap/microk8s/current/credentials/client.config"
+        )
 
         self.add_cmd_output([
             f"microk8s {subcmd}" for subcmd in microk8s_subcmds


sos/report/plugins/microk8s.py
```diff
@@ -28,6 +29,22 @@ class Microk8s(Plugin, UbuntuPlugin):
     packages = ('microk8s',)
 
     microk8s_cmd = "microk8s"
+    kube_cmd = "microk8s kubectl"
+
+    files = (
+        "/var/snap/microk8s/current/credentials/client.config"
```
arif-ali, line 35: in order for this tuple to work, you need to add a comma at the end

However, this tuple is about triggering the plugin, and not about collecting file config. I would think having the `microk8s` in the `packages` tuple should suffice.

If we need to collect the config file, then we need to add that to the `setup()` function and add

```python
self.add_copy_spec("/var/snap/microk8s/current/credentials/client.config")
```

also noting that we may need to add a `postproc()` function if we need to mask any private information like passwords and private kets from that config file

sos/report/plugins/microk8s.py
```diff
@@ -44,6 +61,7 @@ def setup(self):
         self.add_cmd_output([
             f"microk8s {subcmd}" for subcmd in microk8s_subcmds
         ])
+        super(UbuntuKubernetes, self).setup()
```
arif-ali, line 64: I've not seen this being done before within `sos` to call out a class outside. I guess if it works, then great. But would be good to get review from @TurboTurtle or @pmoravec on this

arif-ali (changes requested):

sos/report/plugins/microk8s.py
phvalguima: Hi @arif-ali thaks! Missed the comma indeed :D

re:add_copy_spec
Do I need it? Just asking because we are not doing it in `sos/report/plugins/kubernetes.py::UbuntuKubernetes`

arif-ali: The UbuntuKubernetes is there for standard kubernetes gathering.

With this plugin dedicated for `microk8s`, it may be useful to gather that file. Is it useful to gather that file, if so, then we can add it as a `add_copy_spec`, ensuring any obfuscation is done wrt any secrets

phvalguima: Added the copy_spec

sos/report/plugins/microk8s.py
TurboTurtle: Hmmmm. Yeah, that's not something we've done and also not something that `Plugin` is designed with the idea of doing.

I think what our typical design would be in this case would be to have a check within the base `kubernetes` class for the installation of microk8s, and if so add the microk8s specific bits there, allowing the rest of the plugin to focus on the shared content between the two. E.G.:

```python
class kubernetes(Plugin)

self.kube_cmd = 'kubectl'

if self.is_installed('microk8s'):
    self.kube_cmd = 'microk8s kubectl'
    self.add_copy_spec(
            "/var/snap/microk8s/current/credentials/client.config"
        )
    microk8s_subcmds = [<whatever isn't shared with kuberentes>]

    self.add_cmd_output([
        f"microk8s {subcmd}" for subcmd in microk8s_subcmds
    ])

[...the rest of the existing kubernetes plugin...]
```

I think this wasn't considered when the `microk8s` plugin was first written due to a significant lack of certain kubernetes bits that we're present in microk8s. If that's changed, I think the above is probably the correct choice (folding microk8s into kubernetes). If that _isn't_ the case, then I'd ask what we gain by calling `super(UbuntuKubernetes, self).setup()` here, since we'd either be getting not-a-lot more collection, or a whole ton of error output, depending on how `microk8s kubectl` behaves when trying to retrieve a resource that doesn't exist in that project.

arif-ali: I think that makes sense, that we just added the `microk8s` bits 

I was already looking at the Kubeflow stuff internally, and actually we can already utilise the kubernetes plugin for his ,and having the microk8s bits within the kubernetes plugin would actually make it easier.

Something on the line of https://paste.ubuntu.com/p/gcrqHkNHxN/ would potentially work I think

dnegreira: Sorry for the late replies :)

I think this all makes sense, the intent of microk8s is to gather the microk8s bits (microk8s status, microk8s config, etc...)
For the kubernetes pod logs we should indeed add the bits in the kubernetes plugin to gather the logs

bin/sos
```diff
@@ -10,6 +10,7 @@
 import sys
 import os
 
+
```
arif-ali, line 13: blank like? can we get this removed, as this is unrelated to the change

sos/report/plugins/microk8s.py
```diff
@@ -48,5 +51,6 @@ def setup(self):
     def postproc(self):
         rsub = r'(certificate-authority-data:|token:)\s.*'
         self.do_cmd_output_sub("microk8s", rsub, r'\1 "**********"')
+        super(Microk8s, self).postproc()
```
arif-ali, line 54: We don't need this line, I think this was a typo from my diff

sos/report/plugins/kubernetes.py
```diff
@@ -212,6 +212,7 @@ class UbuntuKubernetes(Kubernetes, UbuntuPlugin, DebianPlugin):
     files = (
         '/root/cdk/cdk_addons_kubectl_config',
         '/etc/kubernetes/admin.conf'
```
arif-ali, line 214: comma at the end please, I missed that in my diff earlier

arif-ali (changes requested):
a few  comments below.

Then can you squash your commits please.

Thanks for your work on this

sos/report/plugins/kubernetes.py
```diff
@@ -228,6 +229,9 @@ def setup(self):
         for svc in self.services:
             self.add_journal(units=svc)
 
+        if self.is_installed('microk8s'):
+            self.kube_cmd = 'microk8s kubectl'
```
arif-ali, line 233: Just testing, and realised the `_kconf` files never got used, and I think, we out to have this if statement before the for loop above. This will ensure that the following command would run

```
microk8s kubectl --kubeconfig=/var/snap/microk8s/current/credentials/client.config
```

Would that make sense? or does `microk8s kubectl` already take care of this?

phvalguima: Hi @arif-ali, `microk8s kubectl` already calls kubectl with admin-level access. I think we do not actually need the client.config at all.

```
$ sudo snap install --classic --channel=latest/edge microk8s
microk8s (edge) v1.29.1 from Canonical✓ installed
$ sudo microk8s kubectl get pods -A
NAMESPACE     NAME                                     READY   STATUS     RESTARTS   AGE
kube-system   calico-node-tmkg7                        0/1     Init:0/2   0          1s
kube-system   calico-kube-controllers-77bd7c5b-98x9p   0/1     Pending    0          1s
kube-system   coredns-864597b5fd-4pq88                 0/1     Pending    0          1s
```

arif-ali (approved):

pmoravec (approved):

TurboTurtle (approved):

PR #3502, opened 2024-02-01
Separating Ext2/3/4 from filesys and using lsblk -rpo instead /proc/mounts

lherbolt:
[xfs,filesys,ext] Get list of devices by lsblk -lpo NAME,FSTYPE.

It can happen the FS of interest is not actually mounted and then we have no info about it. This is revorks of how we should get list of devices. It adds a new __init__.py fuction to decrease code duplication get_dev_by_fstype(self,fstype). It is using lsblk instead of /proc/mounts. Last change is separation of Ext2/3/4 info into separate module from filesys.

---

### DIFF
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -1616,6 +1616,19 @@ class Plugin():
                 tags.extend(val)
         return tags
 
+    def get_devices_by_fstype(self, fstype):
+        """Get list of devices based on fstype and returns empty array in case
+        of failure.
+        """
+        dev = []
+        all_devs = self.exec_cmd("lsblk -rpo NAME,FSTYPE")
+        if (all_devs['status'] == 0 and not all_devs['truncated']):
+            if all_devs['status'] == 0:
+                for line in all_devs['output'].splitlines():
+                    if (fstype in line) and (line.split()[0] not in dev):
+                        dev.append(line.split()[0])
+        return dev
+
     def generate_copyspec_tags(self):
         """After file collections have completed, retroactively generate
         manifest entries to apply tags to files copied by generic copyspecs
--- a/sos/report/plugins/ext.py
+++ b/sos/report/plugins/ext.py
@@ -1,0 +1,57 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, IndependentPlugin, PluginOpt
+
+
+class Ext(Plugin, IndependentPlugin):
+    """This plugin collects information on mounted Ext2/3/4 filessystems on
+    the local system.
+
+    Users should expect `dumpe2fs -h` or `dumpe2fs` collections by this
+    plugin for each Ext2/3/4 filesystem that is recognized by lsblk.
+    """
+
+    short_desc = 'Ext2/3/4 filesystem'
+
+    plugin_name = 'ext'
+    profiles = ('storage',)
+    files = ('/sys/fs/ext4/', '/proc/fs/ext4/', '/proc/fs/jbd2/')
+
+    option_list = [
+        PluginOpt('dumpe2fs', default=False, desc='dump filesystem info'),
+        PluginOpt('frag', default=False,
+                  desc='collect filesystem fragmentation status')
+    ]
+
+    def setup(self):
+        dumpe2fs_opts = '-h'
+        if self.get_option('dumpe2fs'):
+            dumpe2fs_opts = ''
+        allfs = self.get_devices_by_fstype('ext')
+        if allfs:
+            for fs in allfs:
+                self.add_cmd_output(f"dumpe2fs {dumpe2fs_opts} {fs}",
+                                    tags="dumpe2fs_h")
+
+                if self.get_option('frag'):
+                    self.add_cmd_output(f"e2freefrag {fs}", priority=100)
+
+        else:
+            mounts = '/proc/mounts'
+            ext_fs_regex = r"^(/dev/\S+).+ext[234]\s+"
+            for dev in self.do_regex_find_all(ext_fs_regex, mounts):
+                self.add_cmd_output(f"dumpe2fs {dumpe2fs_opts} {fs}",
+                                    tags="dumpe2fs_h")
+
+            if self.get_option('frag'):
+                self.add_cmd_output(f"e2freefrag {fs}", priority=100)
+
+        self.add_copy_spec(self.files)
+
+# vim: set et ts=4 sw=4 :
--- a/sos/report/plugins/filesys.py
+++ b/sos/report/plugins/filesys.py
@@ -29,9 +29,6 @@ class Filesys(Plugin, DebianPlugin, UbuntuPlugin, CosPlugin):
     option_list = [
         PluginOpt('lsof', default=False,
                   desc='collect information on all open files'),
-        PluginOpt('dumpe2fs', default=False, desc='dump filesystem info'),
-        PluginOpt('frag', default=False,
-                  desc='collect filesystem fragmentation status')
     ]
 
     def setup(self):
@@ -75,18 +72,36 @@ class Filesys(Plugin, DebianPlugin, UbuntuPlugin, CosPlugin):
         if self.get_option('lsof'):
             self.add_cmd_output("lsof -b +M -n -l -P", root_symlink="lsof",
                                 priority=50)
+        option_list = [
+            PluginOpt('dumpe2fs', default=False, desc='dump filesystem info'),
+            PluginOpt('frag', default=False,
+                      desc='collect filesystem fragmentation status')
+        ]
 
+    def setup(self):
         dumpe2fs_opts = '-h'
         if self.get_option('dumpe2fs'):
             dumpe2fs_opts = ''
-        mounts = '/proc/mounts'
-        ext_fs_regex = r"^(/dev/\S+).+ext[234]\s+"
-        for dev in self.do_regex_find_all(ext_fs_regex, mounts):
-            self.add_cmd_output("dumpe2fs %s %s" % (dumpe2fs_opts, dev),
-                                tags="dumpe2fs_h")
+        allfs = self.get_devices_by_fstype('ext')
+        if allfs:
+            for fs in allfs:
+                self.add_cmd_output(f"dumpe2fs {dumpe2fs_opts} {fs}",
+                                    tags="dumpe2fs_h")
+
+                if self.get_option('frag'):
+                    self.add_cmd_output(f"e2freefrag {fs}", priority=100)
+
+        else:
+            mounts = '/proc/mounts'
+            ext_fs_regex = r"^(/dev/\S+).+ext[234]\s+"
+            for dev in self.do_regex_find_all(ext_fs_regex, mounts):
+                self.add_cmd_output(f"dumpe2fs {dumpe2fs_opts} {fs}",
+                                    tags="dumpe2fs_h")
 
             if self.get_option('frag'):
-                self.add_cmd_output("e2freefrag %s" % (dev), priority=100)
+                self.add_cmd_output(f"e2freefrag {fs}", priority=100)
+
+        self.add_copy_spec(self.files)
 
     def postproc(self):
         self.do_file_sub(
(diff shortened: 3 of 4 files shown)


sos/report/plugins/ext.py
```diff
@@ -0,0 +1,51 @@
...
+    the local system.
+
+    Users should expect `dumpe2fs -h` or `dumpe2fs` collections by this
+    plugin for each Ext2/3/4 filesystem that is recognized by lsblk.
+    """
+
+    short_desc = 'Ext2/3/4 filesystem'
+
+    plugin_name = 'ext4'
+    profiles = ('storage')
```
arif-ali, line 23: This needs to be a tuple, so please update this to have a comma at the end, like the example below. This is the reason of the first deb creation failure in CI

```python
profiles = ('storage',)
```

sos/report/plugins/ext.py
```diff
@@ -0,0 +1,51 @@
...
+    files = ('/sys/fs/ext4/', '/proc/fs/ext4/', '/proc/fs/jbd2/')
+
+    def setup(self):
+        dumpe2fs_opts = '-h'
+        if self.get_option('dumpe2fs'):
```
arif-ali, line 28: The options are only available it they have been defined, similar to the `filesys.py`

arif-ali (changes requested):
With this refactor, we are now collecting potentially the same set of items from 2 separate plugins, and we are typically not keen on that.

We'll also need to double check to see if any automation that our downstream have that interrogates files collected in `filesys` in order for them to be removed from there.

I am sure other core reviewers may have other things on this

jcastill:
> Last change is separation of Ext2/3/4 info into separate module from filesys.

I'm not sure if this should go in the same commit or a separated one. My feeling is that it should be a second commit, but lets hear other's opinions.

We are still using /proc/mounts in filesys.py:

```
       dumpe2fs_opts = '-h'
        if self.get_option('dumpe2fs'):
            dumpe2fs_opts = ''
        mounts = '/proc/mounts'
        ext_fs_regex = r"^(/dev/\S+).+ext[234]\s+"
        for dev in self.do_regex_find_all(ext_fs_regex, mounts):
            self.add_cmd_output("dumpe2fs %s %s" % (dumpe2fs_opts, dev),
                                tags="dumpe2fs_h")
```

So it may be worth removing that code completely from filesys now, since you'll be getting it via this new plugin.

lherbolt:
> With this refactor, we are now collecting potentially the same set of items from 2 separate plugins, and we are typically
not keen on that.

Yeah sorry about that forgot to push filesys where the Ext2/3/4 are removed. I will update the pull request.

jcastill:
> We'll also need to double check to see if any automation that our downstream have that interrogates files collected in filesys in order for them to be removed from there.

Same on our side, but I see a lot of benefits on these changes, specially separating ext from filesys like xfs is -that is, with my storage/filesystem engineer hat, I can find it really really useful.

We need to find a way to make sure these changes are properly propagated to users using automation. Some kind of policy for coexistence and maybe duplication for some time, in terms of minor releases. Other options, but I imagine these are less agreeable, are:
- Create a link in filesys to the dumpe2fs outputs in ext.
- If filesys is called with dumpe2fs, make sure the option is enabled in the ext module.

I don't think we've ever done these kind of things in sos, I guess to avoid interlinking plugins.

sos/report/plugins/xfs.py
```diff
@@ -25,18 +25,23 @@ class Xfs(Plugin, IndependentPlugin):
...
+            for fs in allfs:
+                self.add_cmd_output("xfs_info %s" % fs,
                                     tags="xfs_info")
-                self.add_cmd_output("xfs_admin -l -u %s" % (parts[0]))
-
-        self.add_copy_spec([
-            '/proc/fs/xfs',
-            '/sys/fs/xfs'
-        ])
+                self.add_cmd_output("xfs_admin -l -u %s" % fs)
```
arif-ali, line 33: migrate this to f-strings

arif-ali (changes requested):
some other changes that came into mind, based on contribution guidelines

sos/report/plugins/__init__.py
```diff
@@ -1616,6 +1616,19 @@ def get_tags_for_file(self, fname):
...
+        of failure.
+        """
+        dev = []
+        all_devs = self.exec_cmd("lsblk -rpo NAME,FSTYPE")
+        if (all_devs['status'] == 0 and not all_devs['truncated']):
+            if all_devs['status'] == 0:
+                for line in all_devs['output'].splitlines():
+                    if (fstype in line) and (line.split()[0] not in dev):
+                        dev.append(line.split()[0])
+        return dev
```
TurboTurtle, line 1630: I'm torn about having this move from `Plugin` to `SoSReport`, and have it run alongside our device introspection, like for [block devs](https://github.com/sosreport/sos/blob/085ae37e0075a79e0fa16d93d011bc0b438dd0dc/sos/report/__init__.py#L505). Granted, these aren't _hardware devices_, but it's in the same vein as what we do for storage and network interfaces.

That way we run this once, when sos first initializes, and then in `Plugin` this function would instead simply filter through an already-made list/dict/whatever for those matching the requested `fstype`.

sos/report/plugins/ext.py
```diff
@@ -0,0 +1,57 @@
...
+                  desc='collect filesystem fragmentation status')
+    ]
+
+    def setup(self):
+        dumpe2fs_opts = '-h'
+        if self.get_option('dumpe2fs'):
+            dumpe2fs_opts = ''
+        allfs = self.get_devices_by_fstype('ext')
+        if allfs:
+            for fs in allfs:
```
TurboTurtle, line 38: The conditional is superfluous, if `allfs` is an empty list, python just continues on, skipping whatever is in the for-loop.

sos/report/plugins/filesys.py
```diff
@@ -75,18 +72,36 @@ def setup(self):
...
         if self.get_option('dumpe2fs'):
             dumpe2fs_opts = ''
-        mounts = '/proc/mounts'
-        ext_fs_regex = r"^(/dev/\S+).+ext[234]\s+"
-        for dev in self.do_regex_find_all(ext_fs_regex, mounts):
-            self.add_cmd_output("dumpe2fs %s %s" % (dumpe2fs_opts, dev),
-                                tags="dumpe2fs_h")
+        allfs = self.get_devices_by_fstype('ext')
+        if allfs:
+            for fs in allfs:
```
TurboTurtle, line 87: Same here, we can drop the conditional and unnest the entire block by one level here.

sos/report/plugins/xfs.py
```diff
@@ -25,18 +25,23 @@ class Xfs(Plugin, IndependentPlugin):
...
     def setup(self):
-        mounts = '/proc/mounts'
-        ext_fs_regex = r"^(/dev/.+).+xfs\s+"
-        for dev in zip(self.do_regex_find_all(ext_fs_regex, mounts)):
-            for e in dev:
-                parts = e.split(' ')
-                self.add_cmd_output("xfs_info %s" % (parts[1]),
+        allfs = self.get_devices_by_fstype('xfs')
+        if allfs:
+            for fs in allfs:
```
TurboTurtle, line 30: Ditto here.

TurboTurtle (commented):
Mostly a couple minor points below, however I think it warrants fleshing out moving the discovery of filesystems to `SoSReport`, as part of our `_get_hardware_devices()` bits, so that we generate this information _once_, and have plugins (and potentially any other part of sos) reference that collection.

TurboTurtle:
> > We'll also need to double check to see if any automation that our downstream have that interrogates files collected in filesys in order for them to be removed from there.
> 
> Same on our side, but I see a lot of benefits on these changes, specially separating ext from filesys like xfs is -that is, with my storage/filesystem engineer hat, I can find it really really useful.
> 
> We need to find a way to make sure these changes are properly propagated to users using automation. Some kind of policy for coexistence and maybe duplication for some time, in terms of minor releases. Other options, but I imagine these are less agreeable, are:
> 
> * Create a link in filesys to the dumpe2fs outputs in ext.
> * If filesys is called with dumpe2fs, make sure the option is enabled in the ext module.
> 
> I don't think we've ever done these kind of things in sos, I guess to avoid interlinking plugins.


Automation should be using `sos_reports/manifest.json` to find the files they're after. It's the entire point of the manifest - that things move and automation needs a source of truth for finding where stuff is written. Even better, would be the explicit use of the tags for the collections - every command gets an automated set of tags but we can add specific ones as well that can be tailored to what an automation pipeline wants to leverage.

This isn't really that different from other plugins we've broken up in the past, like `general`. Stuff moves, this just happens to be a particularly old plugin that is getting re-worked. Automation uses `manifest.json`, users get messaging. We'll include this in the change notes for 4.7.0 and from there it's up to downstreams to inform their own users.

sos/report/plugins/filesys.py
```diff
@@ -75,18 +72,36 @@ def setup(self):
         if self.get_option('lsof'):
             self.add_cmd_output("lsof -b +M -n -l -P", root_symlink="lsof",
                                 priority=50)
+        option_list = [
+            PluginOpt('dumpe2fs', default=False, desc='dump filesystem info'),
+            PluginOpt('frag', default=False,
+                      desc='collect filesystem fragmentation status')
+        ]
 
+    def setup(self):
```
TurboTurtle, line 81: Oh, missed this on my first review. This seems like a copy/paste error. The options_list shouldn't go within `setup()`, and we've got a duplicate definition of the `setup()` method, which would break several things. If I'm reading everything right on the github webui, I think you just need to remove these lines entirely.

sos/report/plugins/ext.py
lherbolt: Ok, but how to prevent not to collect  info from /proc/mount as fail safe? 
Something like this?
`
        for fs in allfs:
            self.add_cmd_output(f"xfs_info {fs}", tags="xfs_info")
                self.add_cmd_output(f"xfs_admin -l -u {fs}")

        if not allfs:
            mounts = '/proc/mounts'
            ext_fs_regex = r"^(/dev/.+).+xfs\s+"
            for dev in zip(self.do_regex_find_all(ext_fs_regex, mounts)):
                for e in dev:
                    parts = e.split()
                    self.add_cmd_output(f"xfs_info {parts[1]}",
                                        tags="xfs_info")
                    self.add_cmd_output(f"xfs_admin -l -u {parts[0]}")
`

sos/report/plugins/__init__.py
lherbolt: That's a good point. I will into it, the main idea here is to reuse the info, not just for FS but also md.py can benefit from it as the FSTYPE in limited to FS as per `blkid -k`

sos/report/plugins/ext.py
lherbolt: Are you suggesting something like:
```
    def setup(self):
        allfs = self.get_devices_by_fstype('xfs')
        for fs in allfs:
            self.add_cmd_output(f"xfs_info {fs}", tags="xfs_info")
                self.add_cmd_output(f"xfs_admin -l -u {fs}")

        if not allfs:
            mounts = '/proc/mounts'
            ext_fs_regex = r"^(/dev/.+).+xfs\s+"
            for dev in zip(self.do_regex_find_all(ext_fs_regex, mounts)):
                for e in dev:
                    parts = e.split()
                    self.add_cmd_output(f"xfs_info {parts[1]}",
                                        tags="xfs_info")
                    self.add_cmd_output(f"xfs_admin -l -u {parts[0]}")

        self.add_copy_spec(self.files)
```

sos/report/plugins/filesys.py
lherbolt: Yeah copy&paste error

sos/report/plugins/ext.py
TurboTurtle: Yep, that's correct. We could even button it up a little more by `for fs in self.get_devices_by_fstype('xfs'):` provided there is nothing else using `allfs` later on.

sos/report/plugins/__init__.py
TurboTurtle: Sorry I should have been clearer - my intent was to suggest that this be at the same level as `storage`, `network`, or similar. So plugins would access it via something like `for fs in self.devices['filesystems']['xfs']: ...blah...`.

Again, this is a bit of a misnomer placing it under `devices`, so if you prefer we can also make a separate attribute to hold these, so that plugins could do something like `for fs in self.filesystems['xfs']:`

lherbolt:
I will finalize this one after ACK/NACK/CHANGE of #3538

arif-ali:
@lherbolt any updated on this one, would be cool to finalise this, now that #3538 has been merged

lherbolt:
uups went out of my radar, will look at it asap

PR #3509, opened 2024-02-07
[kubernetes] Capture json output for nodes

### LINKED ISSUE #3484: can't collect KUB stats, plugin can't be used
Can't figure how to make kub plugin correctly used. Does not appear in documentation, but present in plugin directory 

Help please

```
root@k8s-master:~/sos# grep -v ^# sos.conf
[global]
verbose = 3
batch = yes

[report]
enable-plugins = kubernetes

[collect]

[clean]

[plugin_options]

root@k8s-master:~/sos# kubectl get no
NAME          STATUS   ROLES           AGE   VERSION
k8s-master    Ready    control-plane   14h   v1.28.1
k8s-worker1   Ready    <none>          14h   v1.28.1
k8s-worker2   Ready    <none>          14h   v1.28.1

root@k8s-master:~/sos# sos report --enable-plugins kubernetes -vvv --batch 

WARNING: Failed to load 'magic' module version >= 0.4.20 which sos aims to use for detecting binary files. A less effective method will be used. It is recommended to install proper python3-magic package with the module.

sosreport (version 4.6.1)

set sysroot to '/' (default)
Network devices not enumerated by nmcli. Will attempt to manually compile list of devices.
[sos.report:setup] executing 'sos report --enable-plugins kubernetes -vvv --batch'
[sos.report:setup] using 'none' preset defaults ()
[sos.report:setup] effective options now: --batch --enable-plugins kubernetes --tmp-dir /tmp -vvv
a non-existing plugin (kubernetes) was specified in the command line.
root@k8s-master:~/sos# 
```

jcastill:
json outputs are easier to parse than the default
output, so lets capture these.

Resolves: #3484

Co-authored-by: Hervé CHIBOIS

---

### DIFF
--- a/sos/report/plugins/kubernetes.py
+++ b/sos/report/plugins/kubernetes.py
@@ -112,7 +112,7 @@ class Kubernetes(Plugin):
                     continue
                 node = words[0]
                 self.add_cmd_output(
-                    "%s describe node %s" % (self.kube_cmd, node),
+                    f"{self.kube_cmd} {kube_get_cmd} node {node}",
                     subdir='nodes'
                 )
 


pmoravec (commented):
As we are changing the formatted string, could we move to f-string already..?

arif-ali:
so we are using `kube_get_cmd` only in 2 places for json output the rest is standard output? If we're going to change it for one, shouldn't we change it for all?

Personally, if I am looking at it, I would ideally look at the standard output, and then potentially if parsing through `jq` or other tools is required, I would then look at the json. Would it make sense to have both human-readable files as well as the json output?

jcastill:
That's what I was thinking, to gather both. In this case we don't lose any data and users and scripts can process both without having to transform what we currently capture. Only thing I'd modify perhaps is to add the jsons in a specific directory but that may break things now. So, both 'normal' and 'json' in the same directories where they are captured?

arif-ali:
@jcastill any updates on this one?

jcastill:
@arif-ali I'll check what else I need to add and push a new version if needed.

arif-ali:
@jcastill you're right, that commit does indeed collect the json output for nodes.

I'll go ahead and close this

PR #3515, opened 2024-02-09
[ocp] Fix backtrace in ocp collector

jcastill:
While runnng sos collector in an ocp cluster
we got the following trace:

  File "/usr/sbin/sos", line 22, in <module>
    sos.execute()
  File "/usr/lib/python3.6/site-packages/sos/__init__.py",
        line 193, in execute
    self._component.execute()
  File "/usr/lib/python3.6/site-packages/sos/collector/__init__.py",
        line 1175, in execute
    self.prep()
  File "/usr/lib/python3.6/site-packages/sos/collector/__init__.py",
        line 896, in prep
    self.cluster.setup()
  File "/usr/lib/python3.6/site-packages/sos/collector/clusters/ocp.py",
        line 132, in setup
    out = self.exec_primary_cmd(self.fmt_oc_cmd("auth can-i '*' '*'"))
  File "/usr/lib/python3.6/site-packages/sos/collector/clusters/ocp.py",
        line 104, in fmt_oc_cmd
    return "%s %s" % (self.oc_cmd, cmd)
  File "/usr/lib/python3.6/site-packages/sos/collector/clusters/ocp.py",
        line 81, in oc_cmd
    'which oc', chroot=self.primary.host.sysroot
TypeError: run_command() got an unexpected keyword argument 'chroot'

This patch changes the unused option 'chroot' that was undefined in run_command() 
to the option need_root that was in use in other commands in ocp.py.

Related: RH: RHEL-24351

---


arif-ali (approved):

alosadagrande:
verified the fix in my lab environment, lgtm

sos/collector/clusters/ocp.py
```diff
@@ -78,7 +78,7 @@ def oc_cmd(self):
             self._oc_cmd = 'oc'
             if self.primary.host.in_container():
                 _oc_path = self.primary.run_command(
-                    'which oc', chroot=self.primary.host.sysroot
+                    'which oc', need_root=True
```
TurboTurtle, line 81: These are very much not the same thing, and this is not the correct fix.

`chroot` means "run this command in a chroot, so that we aren't in `'/`'". The conditional before this makes it obvious that we're trying to make sure that we run in the host's mounted filesystem, not inside the container.

`need_root` simply makes sure we're executing as the root user, or otherwise we use `sudo.

TurboTurtle (changes requested):

sos/collector/clusters/ocp.py
jcastill: They are not the same thing, but run_command() doesn't provide a 'chroot' argument. It does provide a 'need_root' argument, and it's used in other places in ocp.py, like in line #349. Should we instead rewrite run_command() to accept a chroot argument?

pmoravec (approved):

sos/collector/clusters/ocp.py
TurboTurtle: The correct way is to look in the mounted root filesystem inside the container, which would be passed by the `HOST` or similar env var and be known to sos. This is what the `chroot` is intended for.

The current change does not actually fix that binary discovery. It simply runs `which oc`. The `needs_root` here is meaningless as we'd already be root in the container.

jcastill: Ok, so I'll remove that part and resend the PR with only the fix that solves the backtrace.

jcastill:
I removed `need_root=True` from the commit, and the fix is now just removing the option that triggered the call trace.

arif-ali (approved):

pmoravec:
Is propr `oc` detected, then? Could @alosadagrande test it, please? The `chroot` "scope" of the `oc` binary search is really not needed?

jcastill:
> Is propr `oc` detected, then? Could @alosadagrande test it, please? The `chroot` "scope" of the `oc` binary search is really not needed?

The default of 'chroot' is False iIrc, so Alberto's test should tell us more for sure

pmoravec:
> > Is propr `oc` detected, then? Could @alosadagrande test it, please? The `chroot` "scope" of the `oc` binary search is really not needed?
> 
> The default of 'chroot' is False iIrc, so Alberto's test should tell us more for sure

Kindly requesting @alosadagrande for a test..

alosadagrande:
I do not have the environment anymore. I'll try to ask for a new one, but it can take a couple of weeks

arif-ali:
@alosadagrande @jcastill any update on this one

jcastill:
The only blocker in this one is to test the latest fix. I'll try to set up an environment in the lab and check it soon.

pmoravec:
Any progress here, @jcastill ..?

jcastill:
@pmoravec same as before, the reporter still didn't have a chance to test properly. I'll ping again.

PR #3517, opened 2024-02-09
[ocp] Add an option to specify API URL

jcastill:
If no API url is specified, it will use
localhost always. In some cases, specially
when using Ansible, we'll want to specify
another API URL.

Related: RH: RHEL-24523

---

### DIFF
--- a/sos/collector/clusters/ocp.py
+++ b/sos/collector/clusters/ocp.py
@@ -69,7 +69,8 @@ class ocp(Cluster):
         ('role', 'master', 'Colon delimited list of roles to filter on'),
         ('kubeconfig', '', 'Path to the kubeconfig file'),
         ('token', '', 'Service account token to use for oc authorization'),
-        ('with-api', False, 'Collect OCP API data from a master node')
+        ('with-api', False, 'Collect OCP API data from a master node'),
+        ('api-url', '', 'Alternate API URL of an external control-plane'),
     ]
 
     @property
@@ -109,8 +110,9 @@ class ocp(Cluster):
         token
         """
         _res = self.exec_primary_cmd(
-            self.fmt_oc_cmd("login --insecure-skip-tls-verify=True --token=%s"
-                            % self.token)
+            self.fmt_oc_cmd("login --insecure-skip-tls-verify=True "
+                            f"--token={self.token} "
+                            f"{self.get_option('api-url')}")
         )
         return _res['status'] == 0
 


arif-ali (approved):

alosadagrande:
I was able to verify the change in my lab environment, so lgtm

sos/collector/clusters/ocp.py
```diff
@@ -108,8 +109,9 @@ def _attempt_oc_login(self):
         token
         """
         _res = self.exec_primary_cmd(
-            self.fmt_oc_cmd("login --insecure-skip-tls-verify=True --token=%s"
-                            % self.token)
+            self.fmt_oc_cmd("login --insecure-skip-tls-verify=True "
+                            f"--token={self.token} "
+                            f"{self.get_option('api-url')}")
```
TurboTurtle, line 114: This doesn't make sense to inject when the default value of `api-url` is used. We'd either break things, or introduce ambiguity into the login command, by passing a spurious `False` value at the end of the string.

TurboTurtle (changes requested):

sos/collector/clusters/ocp.py
jcastill: @alosadagrande checked and it still works when not specifying it. But, would it be cleaner to move from False to '' ?

alosadagrande:
> I also don't see the use case. This login always occurs on the OCP master, and only there - it is wrapped in `exec_primary_cmd()`. (EDIT: or, rather, the "master" in that you have `oc` configured for that OCP environment.)
> 
> Why would you be pointing `sos collect` at an ansible host to serve as an OCP master?

We are working on a use case where I want to run sos collect from Ansible AP (formerly Ansible Tower) installed on an OCP hub cluster (using RHACM). The hub cluster manages the provisioning and lifecycle of multiple OCP clusters.  The idea is the user has a fleet of OCP clusters managed from a central hub running RHACM. 

Then, if they have any issues on any of the managed clusters, the operators without having to connect to the affected cluster can label them using policies or a gitops approach and a sos report is automatically executed on the labeled cluster.  It can be done also manually, by providing the proper OCP cluster name to the Ansible playbook. Therefore, the workflow is the following:

* A sos collect Pod is executed on a OCP hub cluster.
* Since the hub cluster manages a fleet of clusters it has the credentials to access the managed clusters via OCP API.
* A sos report using the ocp plugin is executed on a managed cluster or clusters.

I've been working with AAP & ACM && ansible stolostron collection and I am able to run a (containerized) sos collect in the hub cluster with --cluster-type=ocp and passing the proper --cluster-option ocp.token credentials. However I am missing the managed cluster API URL to connect to. [https://github.com/sosreport/sos/blob/dea531fd6f8d73360c12e1b82953238eb904281e/sos/collector/clusters/ocp.py#L106]

sos/collector/clusters/ocp.py
TurboTurtle: For a project that has a _ton_ of external calls, it is imperative that sos be as exact and semantically correct as possible.

TurboTurtle:
Hmm, so you aren't utilizing, necessarily, a given kubeconfig at any given time for `oc` to be directed to the right cluster? That's somewhat strange to me, but I'll defer on that under the premise that Ansible does Ansible things, which may not always be the human thing to do.

So long as the default behavior remains that the login occurs against `localhost`, this seems reasonable to me.

sos/collector/clusters/ocp.py
```diff
@@ -69,7 +69,8 @@ class ocp(Cluster):
         ('role', 'master', 'Colon delimited list of roles to filter on'),
         ('kubeconfig', '', 'Path to the kubeconfig file'),
         ('token', '', 'Service account token to use for oc authorization'),
-        ('with-api', False, 'Collect OCP API data from a master node')
+        ('with-api', False, 'Collect OCP API data from a master node'),
+        ('api-url', False, 'Collect OCP API data from a master node'),
```
pmoravec, line 73: If `api-url` is meant to be string, shouldn't it have `''` as default instead of `False`?

sos/collector/clusters/ocp.py
pmoravec: How it could work with default `False`? Per my testing, `collector`'s `get_option` should return `bool` typed `False`, hence `fmt_oc_cmd` should get as an argument a string like

    login --insecure-skip-tls-verify=True --token=..... False

pmoravec (approved):
LGTM

sos/collector/clusters/ocp.py
```diff
@@ -69,7 +69,8 @@ class ocp(Cluster):
         ('role', 'master', 'Colon delimited list of roles to filter on'),
         ('kubeconfig', '', 'Path to the kubeconfig file'),
         ('token', '', 'Service account token to use for oc authorization'),
-        ('with-api', False, 'Collect OCP API data from a master node')
+        ('with-api', False, 'Collect OCP API data from a master node'),
+        ('api-url', '', 'Alternate API URL of an external control-plane'),
```
alosadagrande, line 73: I think there is an extra comma in api-url definition.

jcastill: If you mean at the end of line 73, I've been asked by sos maintainers in the past to add it to avoid changing one extra line on subsequent changes

sos/collector/clusters/ocp.py
pmoravec: As the default was changed to empty string, this comment is resolved.

pmoravec:
As @TurboTurtle's concerns were answered and the PR got 2 ACKs, I will merge it soon.

PR #3520, opened 2024-02-12
[cleaner] Add option to skip cleaning files

pmoravec:
A new option --skip-clean-files allows cleaner to skip cleaning files where the user is certain no sensitive information is present.

The option supports globs / wildcards.

Relevant: #3469
Closes: #3520

---

### DIFF
--- a/sos/cleaner/__init__.py
+++ b/sos/cleaner/__init__.py
@@ -15,6 +15,7 @@ import os
 import shutil
 import sos.cleaner.preppers
 import tempfile
+import fnmatch
 
 from concurrent.futures import ThreadPoolExecutor
 from datetime import datetime
@@ -81,6 +82,7 @@ class SoSCleaner(SoSComponent):
         'archive_type': 'auto',
         'domains': [],
         'disable_parsers': [],
+        'skip_clean_files': [],
         'jobs': 4,
         'keywords': [],
         'keyword_file': None,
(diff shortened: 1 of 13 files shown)


pmoravec:
I need to add an avocado test before merging.

sos/cleaner/__init__.py
```diff
@@ -262,6 +265,10 @@ def add_parser_options(cls, parser):
                                default=[], dest='disable_parsers',
                                help=('Disable specific parsers, so that those '
                                      'elements are not obfuscated'))
+        clean_grp.add_argument('--skip-clean-files', action='extend',
+                               default=[], dest='skip_clean_files',
+                               help=('List of files to skip/ignore during '
+                                     'cleaning. Asterisks are supported.'))
```
TurboTurtle, line 271: s/Asterisks/Globs in my opinion, as it is more precise.

TurboTurtle (commented):
Code LGTM, minor notes below.

pmoravec:
I was about to write a test for this when I discovered a problem: we can't easily tell cleaner not to obfuscate symbolic links.

Check:

    sos report --clean -o filesys --keywords=sysfs --batch --build --skip-cleaning-files etc/mtab

Then, https://github.com/sosreport/sos/blob/main/sos/cleaner/__init__.py#L758 causes we skip the decision "shall we obfuscate the file or not?" for a symlink.

We can't easily add "`else:`" branch as the target file of the symlink can be already obfuscated.

I was thinking about this solution: anywere before https://github.com/sosreport/sos/blob/main/sos/cleaner/__init__.py#L676 , traverse the `archive.get_file_list()` for symbolic links. If either symlink matches a parser's `skip_patterns`, add the symlink's `[os.path.realpath](http://docs.python.org/library/os.path.html#os.path.realpath)` target to the parser's `skip_patterns`.

It is bit ugly and lengthy, but I dont see an easier way..

Anyway, that is an independent issue I can deal afterwards.

pmoravec:
Avocado test added, symlinks issue raised as https://github.com/sosreport/sos/issues/3563 .

TurboTurtle (approved):

PR #3522, opened 2024-02-13
[infinidat] Add infinidat plugin

al3jandrosg:
Adding a plugint to collect:

- infinidat storage logs
- some info that could be useful for debugging purposes

Resolves: SET-454

---

- [ x ] Is the commit message split over multiple lines and hard-wrapped at 72 characters?
- [ x ] Is the subject and message clear and concise?
- [ x ] Does the subject start with **[plugin_name]** if submitting a plugin patch or a **[section_name]** if part of the core sosreport code?
- [ x ] Does the commit contain a **Signed-off-by: First Lastname <email@example.com>**?

### DIFF
--- a/sos/report/plugins/infinidat.py
+++ b/sos/report/plugins/infinidat.py
@@ -1,0 +1,44 @@
+# Copyright (C) 2024 Alejandro Santoyo <alejandro.santoyo@canonical.com>
+#
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, IndependentPlugin
+
+
+class InfinidatStorage(Plugin, IndependentPlugin):
+
+    short_desc = 'Infinidat Storage plugin'
+    plugin_name = 'infinidat'
+    profiles = ('storage',)
+    packages = ('host-power-tools',)
+
+    def setup(self):
+        # Get infinidat logs
+        if not self.get_option("all_logs"):
+            self.add_copy_spec([
+                "/var/log/infinihost.latest*.log",
+                "/var/log/infinihost.usage*.log",
+            ])
+        else:
+            self.add_copy_spec([
+                "/var/log/infinihost*.log",
+                "/var/log/buildout.*.log",
+            ])
+
+        # Get info from the infinidat boxes, etc.
+        self.add_cmd_output([
+            "infinihost volume list",
+            "infinihost connectivity list",
+            "infinihost system list",
+            "infinihost pool list",
+            "infinihost snapshot list",
+            "infinihost --version"
+        ])
+
+# vim: set et ts=4 sw=4 :


sos/report/plugins/infinidat.py
```diff
@@ -0,0 +1,37 @@
...
+
+import platform
+from sos.report.plugins import Plugin, IndependentPlugin
+    
+    
+class InfinidatStoragePlugin(Plugin, IndependentPlugin):
+    
+    short_desc = 'Infinidat Storage plugin'
+    plugin_name = 'infinidat'
+    profiles = ('storage')
```
arif-ali, line 20: needs t o be a tuple, so add a comma after `'storage'`

sos/report/plugins/infinidat.py
```diff
@@ -0,0 +1,37 @@
...
+    packages = ('host-power-tools',)
+
+    def setup(self):
+        # Get infinidat logs
+        self.add_copy_spec("/var/log/infinihost*.log")
```
arif-ali, line 25: We should collect one log normally, and can then collect all logs with `--all-logs`

arif-ali (changes requested):
Please my comments below, also double check the `flake8` tests please.

There's a CodeQL message on `import platform`, which is not being used.

sos/report/plugins/infinidat.py
```diff
@@ -0,0 +1,37 @@
...
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+
+import platform
+from sos.report.plugins import Plugin, IndependentPlugin
+    
+    
+class InfinidatStoragePlugin(Plugin, IndependentPlugin):
```
TurboTurtle, line 16: Pedantic nit: we don't generally suffix our plugin class names with 'Plugin', so this would usually just be `InfinidatStorage` or simply `Infinidat`.

sos/report/plugins/infinidat.py
```diff
@@ -0,0 +1,37 @@
...
+        # Get infinidat logs
+        self.add_copy_spec("/var/log/infinihost*.log")
+        
+        # Get info from the infinidat boxes, etc.  
+        self.add_cmd_output([
+            "infinihost volume list",
+            "infinihost connectivity list",
+            "infinihost system list",
+            "infinihost pool list",
+            "infinihost snapshot list",
```
TurboTurtle, line 33: Is there any inspection we would want to do on the output of these list commands? E.G. if `infinihost volume list` returns one or more volumes, are there any further commands we should run for each of the volumes reported?

sos/report/plugins/infinidat.py
al3jandrosg: Fixed, thanks

sos/report/plugins/infinidat.py
al3jandrosg: Fixed, thanks

sos/report/plugins/infinidat.py
al3jandrosg: Fixed, thanks

sos/report/plugins/infinidat.py
al3jandrosg: I've checked and there are no available/useful commands to follow up on the above (nothing like for example `infinihost volume show <volume>`). Hence, I believe these should suffice.

arif-ali (approved):

TurboTurtle (approved):

PR #3528, opened 2024-02-16
[flatpak] Prevent flatpak commands to mount gvfs

pmoravec:
Running various flatpak commands can start gvfs mount, which is undesirable outcome of sos. Calling the commands with GVFS_REMOTE_VOLUME_MONITOR_IGNORE=1 env.variable prevents that behaviour.

Relevant: RHEL-14328
Closes: #3528

---

### DIFF
--- a/sos/report/plugins/flatpak.py
+++ b/sos/report/plugins/flatpak.py
@@ -19,6 +19,7 @@ class Flatpak(Plugin, IndependentPlugin):
     packages = ("flatpak",)
 
     def setup(self):
+        env = {"GVFS_REMOTE_VOLUME_MONITOR_IGNORE": "1"}
         self.add_cmd_output([
             "flatpak --version",
             "flatpak --default-arch",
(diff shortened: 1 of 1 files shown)


jcastill (approved):
lgtm

TurboTurtle (approved):

sos/report/plugins/flatpak.py
```diff
@@ -19,6 +19,7 @@ class Flatpak(Plugin, IndependentPlugin):
     packages = ("flatpak",)
 
     def setup(self):
+        env = {"GVFS_REMOTE_VOLUME_MONITOR_IGNORE": 1}
```
TurboTurtle, line 22: This looked ok at first blush, but the smoke test is failing with `TypeError: expected str, bytes or os.PathLike object, not int`.

I'm figuring we'll need to quote that integer here.

TurboTurtle (changes requested):

sos/report/plugins/flatpak.py
pmoravec: Indeed, `Popen` requires `env` as a `dict` from `str` to `str` only. One still learns  new things.. :)

Fixed now.

arif-ali (approved):

TurboTurtle (approved):

PR #3531, opened 2024-02-20
doc: fix github url

makelinux:
---

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -17,7 +17,7 @@ To clone the current main (development) branch run:
 
 .. code::
 
-    git clone git://github.com/sosreport/sos.git
+    git clone git@github.com:sosreport/sos.git
 
 Reporting bugs
 ^^^^^^^^^^^^^^


jcastill (approved):
lgtm

docs/index.rst
```diff
@@ -17,7 +17,7 @@ To clone the current main (development) branch run:
 
 .. code::
 
-    git clone git://github.com/sosreport/sos.git
+    git clone git@github.com:sosreport/sos.git
```
arif-ali, line 20: would it be better to have the http endpoint, that way people don't have to specifically setup ssh? Especially if people don't have GH setup?

arif-ali (approved):
looks ok to me, just a comment

TurboTurtle (approved):

docs/index.rst
TurboTurtle: IIRC doesn't GH have plans in play to drop non-authenticated http? Meaning someone would have to do some setup/configuration anyways (and GH strongly encourages the use of ssh)?

PR #3534, opened 2024-02-21
[redhat|policy] Check for archive size before upload

jcastill:
The Red Hat Customer Portal has a max limit for
single http requests of 1Gb. From sos side, we never checked this and so customers had to wait for the whole upload to be attempted before receiving an error from the portal.
This patch attempts to stop the upload before it starts, informing customers of such limit, and switching to secure FTP where there's no limit.

Related: RHEL-22732

---

### DIFF
--- a/sos/policies/distros/redhat.py
+++ b/sos/policies/distros/redhat.py
@@ -22,7 +22,7 @@ from sos.policies.distros import LinuxPolicy, ENV_HOST_SYSROOT
 from sos.policies.package_managers.rpm import RpmPackageManager
 from sos.policies.package_managers.flatpak import FlatpakPackageManager
 from sos.policies.package_managers import MultiPackageManager
-from sos.utilities import bold
+from sos.utilities import bold, convert_bytes
 from sos import _sos as _
 
 try:
@@ -232,6 +232,8 @@ support representative.
     _upload_url = RH_SFTP_HOST
     _upload_method = 'post'
     _device_token = None
+    # Max size for an http single request is 1Gb
+    _max_size_request = 1073741824
 
     def __init__(self, sysroot=None, init=None, probe_runtime=True,
                  remote_exec=None):
@@ -429,16 +431,31 @@ support representative.
             return super().upload_sftp(user=_user, password=_token)
         raise Exception("Could not retrieve valid or anonymous credentials")
 
+    def check_file_too_big(self, archive):
+        size = os.path.getsize(archive)
+        # Lets check if the size is bigger than the limit.
+        # There's really no need to transform the size to Gb,
+        # so we don't need to call any size converter implemented
+        # in tools.py
+        if (size >= self._max_size_request):
+            self.ui_log.warning(
+                _("Size of archive is bigger than Red Hat Customer Portal "
+                  "limit for uploads of "
+                  f"{convert_bytes(self._max_size_request)} "
+                  " via sos http upload. \n")
+                  )
+            return RH_SFTP_HOST
+        else:
+            return RH_API_HOST
+
     def upload_archive(self, archive):
         """Override the base upload_archive to provide for automatic failover
         from RHCP failures to the public RH dropbox
         """
         try:
-            if self.upload_url and self.upload_url.startswith(RH_API_HOST) and\
-                    (not self.get_upload_user() or
-                     not self.get_upload_password()):
-                self.upload_url = RH_SFTP_HOST
-            uploaded = super().upload_archive(archive)
+            if self.get_upload_url().startswith(RH_API_HOST):
+                self.upload_url = self.check_file_too_big(archive)
+            uploaded = super(RHELPolicy, self).upload_archive(archive)
         except Exception as e:
             uploaded = False
             if not self.upload_url.startswith(RH_API_HOST):


jcastill:
I tried to add the check as early as possible, before uploading, but making sure it was a RH only check, so I chose to use the upload_archive() in redhat.py. Another option was to add it to the 'if' just below where it is, returning a boolean, but I thought that could add more complexity to that if.

sos/policies/distros/redhat.py
```diff
@@ -232,6 +232,9 @@ class RHELPolicy(RedHatPolicy):
     _upload_url = RH_SFTP_HOST
     _upload_method = 'post'
     _device_token = None
+    # Max size for an http single request is 1Gb
+    _max_size_request = 1073741824
```
pmoravec, line 236: This is the byte-precise granted size Red Hat Portal allows, right? (rather, one byte less due to `size >= self._max_size_request`)

I just want to prevent sporadic cases when a border-line-sized sosreport passes the check but fails to be uploaded - rather go the other way, failover to SFTP even for sizes slightly below the limit.

sos/policies/distros/redhat.py
```diff
@@ -430,11 +433,26 @@ def upload_sftp(self):
...
         raise Exception("Could not retrieve valid or anonymous credentials")
 
+    def check_file_too_big(self, archive):
+        size = os.path.getsize(archive)
+        # Lets check if the size is bigger than the limit.
+        # There's really no need to transform the size to Gb,
+        # so we don't need to call any size converter implemented
+        # in tools.py
+        if len(str(size)) >= 10 and (size >= self._max_size_request):
+            self.ui_log.error(
```
pmoravec, line 443: IMHO warning is the right verbosity level, as we alter the default (or user-defined) behaviour (hence warn) but no real error happens (rather we prevent one).

sos/policies/distros/redhat.py
```diff
@@ -430,11 +433,26 @@ def upload_sftp(self):
                                                        password=_token)
         raise Exception("Could not retrieve valid or anonymous credentials")
 
+    def check_file_too_big(self, archive):
+        size = os.path.getsize(archive)
+        # Lets check if the size is bigger than the limit.
+        # There's really no need to transform the size to Gb,
+        # so we don't need to call any size converter implemented
+        # in tools.py
+        if len(str(size)) >= 10 and (size >= self._max_size_request):
```
pmoravec, line 442: Why the first condition `len(str(size)) >= 10`? That implies from the second, no?

TurboTurtle, line 442: _What_?

The raw size comparison is the only thing needed.

sos/policies/distros/redhat.py
```diff
@@ -430,11 +433,26 @@ def upload_sftp(self):
...
+            self.ui_log.error(
+                _("Size of archive is bigger than Red Hat Customer Portal "
+                  f"limit for uploads of {self._max_size_human_readable} via "
+                  "sos http upload. \n")
+                  )
```
TurboTurtle, line 447: It seems strange to stick `1Gb` in a variable for it only to be used once in an f-string, rather than just making the warning message contain that static value.

sos/policies/distros/redhat.py
```diff
@@ -430,11 +433,26 @@ def upload_sftp(self):
...
+            self.upload_url = RH_SFTP_HOST
```
TurboTurtle, line 448: If we're changing the upload target on users, we should tell them as much. The current message makes it sound like we're not going to do any upload because of the size limit.

Alternatively, this should raise an exception to be caught by the try block this is called from.

sos/policies/distros/redhat.py
```diff
@@ -430,11 +433,26 @@ def upload_sftp(self):
...
+
     def upload_archive(self, archive):
         """Override the base upload_archive to provide for automatic failover
         from RHCP failures to the public RH dropbox
         """
         try:
+            self.check_file_too_big(archive)
```
TurboTurtle, line 455: This should be gated by checking that we are in fact trying to upload to the RH customer portal - the default is for the policy to set the target, but it's entirely possible for it to not be `RH_API_HOST`.

TurboTurtle (changes requested):

sos/policies/distros/redhat.py
jcastill: Yes, that's right. I checked with a bit more than 1Gb but can check the exact size as well in case we need to change the value.

sos/policies/distros/redhat.py
jcastill: Ack, you are right. We are failing over to SFTP so a warning is probably the best way to go.

sos/policies/distros/redhat.py
jcastill: yes, that's a leftover from the first implementation when I was transforming the size to "1GB" and overcomplicating things. I'll remove it.

sos/policies/distros/redhat.py
jcastill: Ack, but it will be good to know about the scenarios where the target is not RH_API_HOST when going through sos/policies/distros/redhat.py

sos/policies/distros/redhat.py
jcastill: Sure. I'll change that.

sos/policies/distros/redhat.py
jcastill: The output with this change is:

```

Size of archive is bigger than Red Hat Customer Portal limit for uploads of 1Gb via sos http upload. 

```

The difference is that before, users will get an error after attempting to upload the whole file, and then a switch to SFTP with the "Attempting upload to Red Hat Secure FTP" message.

Do I still need to add yet another "we are failing over to SFTP" to the code above?

pmoravec: I think this logging is sufficient.

TurboTurtle: That's fair. I had the order of logging for the upload target reversed in my head.

sos/policies/distros/redhat.py
jcastill: Fixed in next version.

sos/policies/distros/redhat.py
jcastill: Fixed

jcastill:
I think I've addressed all the notes, but apologies if I forgot anything! This is how the output looks right now:

```

	/var/tmp/sosreport-localhost-03093474-2024-04-24-wbdtdaf.tar.xz

 Size	1.41GiB
 Owner	root
 sha256	0d7bebe2a81a92f3b8849fa807a233e0eadaf8c99717a25e1f92480d6a492ed0

Size of archive is bigger than Red Hat Customer Portal limit for uploads of 1.0G  via sos http upload. 

```
We present GiB in "Size" and I print G in the warning. It may be worth changing the value I print so it's consistent.

sos/policies/distros/redhat.py
```diff
@@ -430,12 +432,30 @@ def upload_sftp(self):
...
         raise Exception("Could not retrieve valid or anonymous credentials")
 
+    def check_file_too_big(self, archive):
+        size = os.path.getsize(archive)
+        # Lets check if the size is bigger than the limit.
+        # There's really no need to transform the size to Gb,
+        # so we don't need to call any size converter implemented
+        # in tools.py
+        if (size >= self._max_size_request) and\
+                self.get_upload_url().startswith(RH_API_HOST):
```
TurboTurtle, line 442: The check against the upload URL should be done prior to the call to `check_file_too_big()`, otherwise we are doing the size comparison needlessly.

TurboTurtle (commented):
This looks mostly good to me now, just would like to have the upload URL check moved  to be the gating logic on the size check as noted (so that we don't perform the size check if a user has specified a different `--upload-url`.

jcastill:
I changed this a bit from the previous iteration. I considered two options. First one - have the check for the RH URL and the file size in the same _if_:

```
		try:
			if self.get_upload_url().startswith(RH_API_HOST) and\
			   self.check_file_too_big(archive):
				self.upload_url = RH_SFTP_HOST
```

Second option - modify the function to return upload URL, and assign it to _self.upload_url_ after gatekeeping, like this:

```
		try:
			if self.get_upload_url().startswith(RH_API_HOST):
				self.upload_url = self.check_file_too_big(archive)
```

I've chosen the second one to do the size check after the URL check, and not at the same time. And in both cases I considered, I removed the check for user and password because after the change to token auth these we always _None_.  
Let me know if I missed anything or if you think there's a better way to do this.

TurboTurtle (approved):
LGTM

sos/policies/distros/redhat.py
```diff
@@ -429,16 +431,31 @@ def upload_sftp(self):  # pylint: disable=too-many-branches
...
+        # There's really no need to transform the size to Gb,
+        # so we don't need to call any size converter implemented
+        # in tools.py
+        if (size >= self._max_size_request):
+            self.ui_log.warning(
+                _("Size of archive is bigger than Red Hat Customer Portal "
+                  "limit for uploads of "
+                  f"{convert_bytes(self._max_size_request)} "
+                  " via sos http upload. \n")
+                  )
```
pmoravec, line 446: Is it worth mentioning here "File will be uploaded to SFTP server instead"?

pmoravec (approved):
ACK with current change, optionally worth to extend the warning message.

jcastill:
> ACK with current change, optionally worth to extend the warning message.

It happens "naturally" in the code flow. This is the result:

```

Size of archive is bigger than Red Hat Customer Portal limit for uploads of 1.0G  via sos http upload. 

```

That's why I chose not to mention it again.
Another thing - I wondered about using 'bold' in the message but I'm not sure if that would be too much - do we have a preferred way to present certain important messages to the user?

pmoravec:
Ah right, I forgot the context / the subsequent message. Then the current log is sufficient, I would say.

I think no bold text is needed here, it is just a warning message, not even an error :)

PR #3536, opened 2024-02-23
Replacing all string formatting with f-strings.

TrevorBenson:
---

This is mostly for checks, and tests and to allow a first review by maintainers for further discussion in #3472.


sos/collector/transports/oc.py
```diff
@@ -71,91 +68,67 @@ def get_node_pod_config(self):
...
-                                "name": "run",
-                                "mountPath": "/run"
-                            },
-                            {
-                                "name": "varlog",
-                                "mountPath": "/var/log"
-                            },
+                            {"name": "host", "mountPath": "/host"},
+                            {"name": "run", "mountPath": "/run"},
+                            {"name": "varlog", "mountPath": "/var/log"},
```
TurboTurtle, line 108: Same kinda thing here. I get that these dict entries/definitions can fit in a single line, but having one-line-per-key is more human-friendly, at least to me. Happy to hear dissenting opinions.

sos/collector/transports/oc.py
```diff
@@ -71,91 +68,67 @@ def get_node_pod_config(self):
...
                         "stdinOnce": True,
-                        "tty": True
+                        "tty": True,
                     }
                 ],
-                "imagePullPolicy":
-                    "Always" if self.opts.force_pull_image else "IfNotPresent",
+                "imagePullPolicy": "Always"
+                if self.opts.force_pull_image
+                else "IfNotPresent",
```
TurboTurtle, line 125: Same here, this (lack of) nesting makes this less readable.

sos/policies/runtimes/crio.py
```diff
@@ -86,7 +88,10 @@ def fmt_container_cmd(self, container, cmd, quotecmd):
         else:
             quoted_cmd = cmd
         container_id = self.get_container_by_name(container)
-        return "%s %s %s" % (self.run_cmd, container_id,
-                             quoted_cmd) if container_id is not None else ''
+        return (
+            f"{self.run_cmd} {container_id} {quoted_cmd}"
+            if container_id is not None
+            else ''
+        )
```
TurboTurtle, line 95: Ditto.

sos/report/__init__.py
```diff
@@ -469,14 +469,15 @@ def _check_container_runtime(self):
             elif crun not in self.policy.runtimes.keys():
                 valid = ', '.join(p for p in self.policy.runtimes.keys()
                                   if p != 'default')
-                raise Exception("Cannot use container runtime '%s': no such "
-                                "runtime detected. Available runtimes: %s"
-                                % (crun, valid))
+                raise Exception(
+                    f"Cannot use container runtime{crun}%s': no such "
```
TurboTurtle, line 473: Left over `%s`?

sos/report/__init__.py
```diff
@@ -924,16 +924,18 @@ def _set_tunables(self):
...
                         except Exception as err:
                             self.soslog.error(err)
                             self._exit(1)
                     del opts[plugname]
             for plugname in opts.keys():
-                self.soslog.error('WARNING: unable to set option for disabled '
-                                  'or non-existing plugin (%s).' % (plugname))
+                self.soslog.error(
+                    'WARNING: unable to set option for disabled '
+                    'or non-existing plugin (%s).' % (plugname)
```
TurboTurtle, line 937: Indention looks better, but not actually replacing with an f-string.

sos/report/__init__.py
```diff
@@ -1058,8 +1061,12 @@ def list_plugins(self):
...
                     tmpopt = 0
 
-                self.ui_log.info(" %-25s %-15s %s" % (
-                    opt.plugin + "." + opt.name, tmpopt, opt.desc))
+                self.ui_log.info(
+                    (
+                        " %-25s %-15s %s"
+                        % (f"{opt.plugin}.{opt.name}", tmpopt, opt.desc)
+                    )
+                )
```
TurboTurtle, line 1069: Same here - got the "inner" substitution, but we can replace the spacing as well.

sos/report/__init__.py
```diff
@@ -1397,11 +1404,19 @@ def ui_progress(self, status_line):
...
+                    f"{name}={val}"
+                    for (name, val) in [
+                        (name, f'{os.environ.get(name)}')
+                        for name in self.env_vars
+                        if os.environ.get(name) is not None
+                    ]
+                ]
+            )
+            + '\n'
+        )
```
TurboTurtle, line 1419: Conversely to the other nesting changes, this one got a bit blown out, I think we can button this up a bit most likely.

sos/report/__init__.py
TrevorBenson: Multiline strings with string format, and a few other "complex" methods of string formatting get overlooked by the tool used to implement the f-strings. I forgot about this until after restoring nested lists and doing a search for any remaining `%s` in the repo. 

I've got another commit underway that I started on last night. There are around 200, out of almost 300, lines remaining to review before I push.

sos/report/__init__.py
TrevorBenson: Yep, these should be resolved after the next commit/push addressing every `%s` remaining individually. There are a rare few that might get a `# pylint: disable=`, I'll create a list of any so they can be discussed individually, as part of the PR review.

TrevorBenson:
> A lot of this looks good. Have some comments that follow; going solely by comments-to-code ratio it's quite a low potential miss rate though - happy to see that semi-automated conversions are mostly reliable.
> 
> I'll be honest I only got through about half of the files at this on this first look before I started to go cross-eyed. Will take another look at this tomorrow or so.

Yeah, mostly. Black didn't do the f-string conversion as I used another tool which I used for static analysis, and is generally very good for replacing formatting and concatenation with f-string. At least for the low-hanging fruit. Black was only for the E501 line length fixes since the first tool is great, but sometimes fails to achieve the intended line lengths. 

I manually went through the diff of every file after to restore contribution style guidelines, mostly for nested lists but a handful of other idiosyncrasies that stood out on quick pass. I'll add your comments to my checklist and look for additional unintended changes or style breakage.

I'll mark this as Draft again until I do the manual review and look back through for the unfixed string formatting. Most look to be multiline strings or other variations the first tool skipped, I had that underway last night, but was trying to make it a single commit with the final changes, instead of many little commits.

sos/collector/transports/oc.py
```diff
@@ -71,91 +68,67 @@ def get_node_pod_config(self):
...
-                            "/bin/bash"
-                        ],
-                        "env": [
-                            {
-                                "name": "HOST",
-                                "value": "/host"
-                            }
-                        ],
+                        if not self.opts.image
+                        else self.opts.image,
```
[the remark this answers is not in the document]
TrevorBenson: This is actually not back, and was something i was very unamused by as the tool should have (and in a year + of using it) have never seen it apply changes this wide when limited explicitely to string formatting replacement. I had a comment working up in the discussion regarding, but decided it was better to just continue doing the manual file by file diff review then mention my displeasure at finding it :smirk: 

Black however was literally a line by line edit in the pipeline. If flake8 pointed out line 151, this line was extracted from the file into a temporary file. The level of indentation of line 151 was used to determine a qty of prepending lines, like:
```
if True: 
    if True:
```
so black did not error about indentation. Formatting applied via `black --preview -l 79` to only this temp file . After deleting `if True` lines the entire temp file was used to replace line 151 in the originating file. Black couldn't actually touch any other part of the file. 

I used sourcery to handle f-strings, and didn't use a similar pipeline and technique for the f-string replacement, I just set the rules to only do f-string. I've never observed this type of leakage in sourcery previously, which was the shocking part after a year of some extensive testing and use.

sos/collector/transports/oc.py
TrevorBenson: Yep, the un-posted comment from last night mentioned this entire file (or any other file where non f-string change occurred) would be rolled back to its original format unless people wanted them to stay. I was fairly displeased to find  these were touched at all when reviewing the diffs, as they did not contain f-string changes at all. It appears as if a data structure had an fstring change to it, and within around 10 lines inside the data-structure the sourcery might adjust the format, it went ahead as if allowing f-string rules meant within +/-10 lines of the fstring adjustment other format changes were allowed.

sos/report/__init__.py
```diff
@@ -436,7 +436,7 @@ def display_help(cls, section):
         ssec.add_text(helpln)
 
     def print_header(self):
-        print("\n%s\n" % _("sosreport (version %s)" % (__version__,)))
+        print("\n%s\n" % _(f"sosreport (version {__version__})"))
```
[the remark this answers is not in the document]
TrevorBenson: I believe so. Just not fixed 100% by sourcery so falls to my manual pass.

TrevorBenson:
I went through each file manually and reviewed its existing styles/syntax. I went with a previous line/PR wins methodology. Meaning if it had a specific style, even if that diverged from the style guide or earlier comments, it wins out as being reviewed and approved previously. This should keep the changes as minimal as possible to the existing code, and keep the file diffs to a minimum number of changed or new lines

There are of course a few cases where this changed. 
* If it was a single string and it was shortened enough to fit on a single line after removing % string formatting, then it may have been shortened.
  * Nested lists that fit on a single line should still maintain the style guide preference.
* In rare instances a new variable was created.
  * This should be limited to only regex becoming a f-string regex. The variable was introduced to maintain the readability of the regex statement as well as keep them from breaking 80 characters and either requiring the regex to be spread over multiple lines or have a `# flake8: noqa` applied to avoid flake8 check failures. I felt this was the best overall solution since regex are inherently complex strings I didn't want f-string implementation to increase it and reduce their readability.
* This only applies to `./sos` currently. I had a full update for `./tests` but it led to ~5 failures. It wasn't immediately obvious where/how it was breaking anything when investigating. I rebased and dropped those commits for the time being.

Anything that still has a % string formatting is up for discussion, but if left in place it either:
1. Was an elegant or complex implementation and seemed worse to replace it without discussion.
2. Required more investigation to better understand the implementation and to find an adequate/acceptable replacement.

TrevorBenson:
I restored the f-strings for the `./tests` path after diagnosing the test breakage to be part of rpm.py. I thought the nested f-string formatting should have been equivalent to the original nested % string formatting. However, I must have overlooked something.

I'll count this as one of the cases that is now up for discussion. Flake8 returns no issues on the `./sos` path and Pylint C0209 has zero results for both `./sos` and `./tests` paths.

@TurboTurtle I believe this is ready for discussion on the build failures since some RPMs appear to build ok, but not packit. I'll go ahead and mark it as ready for review.

TurboTurtle:
> @TurboTurtle @pmoravec Any suggestions on how I should diagnose the packit builds failing while non packit RPM builds appear successful?

The packit builds are passing. The cirrus jobs are failing due to avocado not being able to tell that the tests are....well, tests. This typically happens when there is something within the test class that prevents python from being able to import the class. Nothing immediately jumps out at me in the diffs, so I'd need to dig into the updated test code a bit more to figure out what's going on here.

TrevorBenson:
> > @TurboTurtle @pmoravec Any suggestions on how I should diagnose the packit builds failing while non packit RPM builds appear successful?
> 
> The packit builds are passing. The cirrus jobs are failing due to avocado not being able to tell that the tests are....well, tests. This typically happens when there is something within the test class that prevents python from being able to import the class. Nothing immediately jumps out at me in the diffs, so I'd need to dig into the updated test code a bit more to figure out what's going on here.

I must have had an old check open via cirrus when I found the error for packit? If they are building fine then I'll review the tests again. Everything was done by hand/manually. However, I did that with the rpm plugin as well and there was something that expected the current format I did not account for there as well.

PR #3538, opened 2024-02-27
Add new section fstype section under hardware devices.

lherbolt:
As discussed in the [#3502](https://github.com/sosreport/sos/pull/3502#pullrequestreview-1859497147) adding new section under _get_hardware_devices(). Please review it.
---

### DIFF
--- a/sos/report/__init__.py
+++ b/sos/report/__init__.py
@@ -445,7 +445,8 @@ class SoSReport(SoSComponent):
                 'fibre': self._get_fibre_devs()
             },
             'network': self._get_network_devs(),
-            'namespaced_network': self._get_network_namespace_devices()
+            'namespaced_network': self._get_network_namespace_devices(),
+            'fstype': self._get_devices_by_fstype()
         }
 
     def _check_container_runtime(self):
@@ -679,6 +680,21 @@ class SoSReport(SoSComponent):
                     continue
                 out_ns.append(line.partition(' ')[0])
         return out_ns
+
+    def _get_devices_by_fstype(self):
+        _dev_fstypes = {}
+        _devs = sos_get_command_output("lsblk -nrpo FSTYPE,NAME")
+        if _devs['status'] != 0:
+            return _dev_fstypes
+        for line in (_devs['output'].splitlines()):
+            helper = line.strip().split()
+            if len(helper) == 1:
+                helper.insert(0, 'unknown')
+            if "ext" in helper[0]:
+                helper[0] = 'ext4'
+            _dev_fstypes.setdefault(helper[0], [])
+            _dev_fstypes[helper[0]].append(helper[1])
+        return _dev_fstypes
 
     def get_commons(self):
         return {


jcastill (approved):
lgtm
just tested it as well and no issues with lots of fs and empty first FSTYPE, nicely dealt with

sos/report/__init__.py
```diff
@@ -680,6 +681,21 @@ def _get_network_namespaces(self):
                 out_ns.append(line.partition(' ')[0])
         return out_ns
 
+    def _get_devices_by_fstype(self):
+        _dev_fstypes = {}
+        _devs = sos_get_command_output("lsblk -nrpo FSTYPE,NAME")['output']
```
pmoravec, line 686: We usually test `sos_get_command_output` to `status` (return code of the command), like 9 lines above - but when return code is nonzero, the output shall be empty - what is handled properly further. ACK.

sos/report/__init__.py
```diff
@@ -680,6 +681,21 @@ def _get_network_namespaces(self):
...
+        for line in (_devs.splitlines()):
+            helper = line.strip().split()
+            if len(helper) == 1:
+                helper.insert(0, 'unknown')
+            if "ext" in helper[0]:
+                helper[0] = 'ext2/3/4'
+            if helper[0] not in _dev_fstypes:
+                _dev_fstypes[helper[0]] = [helper[1]]
+            else:
+                _dev_fstypes[helper[0]].append(helper[1])
```
pmoravec, line 696: My brain does not work on full speed at 10:30pm, but can't this be elegantly resolved by either

`_dev_fstypes[helper[0]] = `_dev_fstypes.get(helper[0], []).append(helper[1])`

or using `setdefault` method?

pmoravec:
In general, I am not against maintaining centrally a next type of devices. Keeping a dict per fstype - I dont see much use case for the "per FS type" in particular (which does not mean there is a legitimate use case for it).

lherbolt:
> In general, I am not against maintaining centrally a next type of devices. Keeping a dict per fstype - I dont see much use case for the "per FS type" in particular (which does not mean there is a legitimate use case for it).

The fstype will not just enumerate the fstypes but also linux_raid_member or LVM2_member. The complete list can be obtained from `blkid -k`, not only the filesystems plugins could use this, but the md and lvm plugin as well and probably some others too. The fstype might be bit misleading but I kept the name to match the lsblk output.

sos/report/__init__.py
lherbolt: check for status added

sos/report/__init__.py
```diff
@@ -445,7 +445,8 @@ def _get_hardware_devices(self):
                 'fibre': self._get_fibre_devs()
             },
             'network': self._get_network_devs(),
-            'namespaced_network': self._get_network_namespace_devices()
+            'namespaced_network': self._get_network_namespace_devices(),
+            'fstype': self._get_devices_by_fstype()
```
TurboTurtle, line 449: Nit: I think naming this `filesystem(s)` is a bit more in line with naming for things plugin authors might access.

sos/report/__init__.py
```diff
@@ -680,6 +681,21 @@ def _get_network_namespaces(self):
...
+        _dev_fstypes = {}
+        _devs = sos_get_command_output("lsblk -nrpo FSTYPE,NAME")
+        if _devs['status'] != 0:
+            return _dev_fstypes
+        for line in (_devs['output'].splitlines()):
+            helper = line.strip().split()
+            if len(helper) == 1:
+                helper.insert(0, 'unknown')
+            if "ext" in helper[0]:
+                helper[0] = 'ext2/3/4'
```
TurboTurtle, line 694: Can we not be more specific here? If not, I think we should just put `ext` rather than slashed version possibilities.

TurboTurtle (commented):
Minor comments, code overall looks good.

sos/report/__init__.py
lherbolt: ACK will change it to ext there is no conflict with other blkid types 

```
blkid -k | grep ext
xfs_external_log
ext4dev
ext4
ext3
ext2
```

sos/report/__init__.py
lherbolt: While I have no issue giving it name filesystems, it might be even more misleading than fstypes. The blkid (libblkid) provides also other types such as lvm2, crypto, many RAIDs, swap, drdb, bcache ...  full list available via blkid -k.

TurboTurtle: That's a fair point. `fstypes` is fine by me then, as I mentioned just a nitpick anyways.

sos/report/__init__.py
lherbolt: Left it as ext4

TurboTurtle (approved):

PR #3540, opened 2024-02-28
[plugins] Fix Pylint and PEP8 issues - part3

pponnuvel:
Continuation of #3530.

Likewise excludes [C0114](https://pylint.readthedocs.io/en/latest/user_guide/messages/convention/missing-module-docstring.html), [C0115](https://pylint.readthedocs.io/en/latest/user_guide/messages/convention/missing-class-docstring.html), and [C0209](https://pylint.pycqa.org/en/latest/user_guide/messages/convention/consider-using-f-string.html).

---

### DIFF
--- a/sos/report/plugins/navicli.py
+++ b/sos/report/plugins/navicli.py
@@ -38,31 +38,31 @@ class Navicli(Plugin, RedHatPlugin):
             "/etc/Navisphere/log/[a-z]*"
         ])
 
-    def get_navicli_SP_info(self, SP_address):
+    def get_navicli_sp_info(self, sp_address):
         """ EMC Navisphere Host Agent NAVICLI specific
         information - CLARiiON - commands
         """
         self.add_cmd_output([
-            "navicli -h %s getall" % SP_address,
-            "navicli -h %s getsptime -spa" % SP_address,
-            "navicli -h %s getsptime -spb" % SP_address,
-            "navicli -h %s getlog" % SP_address,
-            "navicli -h %s getdisk" % SP_address,
-            "navicli -h %s getcache" % SP_address,
-            "navicli -h %s getlun" % SP_address,
+            "navicli -h %s getall" % sp_address,
+            "navicli -h %s getsptime -spa" % sp_address,
+            "navicli -h %s getsptime -spb" % sp_address,
+            "navicli -h %s getlog" % sp_address,
+            "navicli -h %s getdisk" % sp_address,
+            "navicli -h %s getcache" % sp_address,
+            "navicli -h %s getlun" % sp_address,
             "navicli -h %s getlun -rg -type -default -owner -crus "
-            "-capacity" % SP_address,
-            "navicli -h %s lunmapinfo" % SP_address,
-            "navicli -h %s getcrus" % SP_address,
-            "navicli -h %s port -list -all" % SP_address,
-            "navicli -h %s storagegroup -list" % SP_address,
-            "navicli -h %s spportspeed -get" % SP_address
+            "-capacity" % sp_address,
+            "navicli -h %s lunmapinfo" % sp_address,
+            "navicli -h %s getcrus" % sp_address,
+            "navicli -h %s port -list -all" % sp_address,
+            "navicli -h %s storagegroup -list" % sp_address,
+            "navicli -h %s spportspeed -get" % sp_address
         ])
 
     def setup(self):
         self.get_navicli_config()
-        for ip in set(self.get_option("ipaddrs").split()):
-            if self.exec_cmd("navicli -h %s getsptime" % (ip))['status'] == 0:
-                self.get_navicli_SP_info(ip)
+        for addr in set(self.get_option("ipaddrs").split()):
+            if self.exec_cmd(f"navicli -h {addr} getsptime")['status'] == 0:
+                self.get_navicli_sp_info(addr)
 
 # vim: set et ts=4 sw=4 :
--- a/sos/report/plugins/networking.py
+++ b/sos/report/plugins/networking.py
@@ -45,11 +45,7 @@ class Networking(Plugin):
     ethtool_shortopts = "acdgiklmPST"
 
     def setup(self):
-        super(Networking, self).setup()
-        for opt in self.ethtool_shortopts:
-            self.add_cmd_tags({
-                'ethtool -%s .*' % opt: 'ethool_%s' % opt
-            })
+        super().setup()
 
         self.add_file_tags({
             '/proc/net/bonding/bond.*': 'bond',
@@ -57,20 +53,20 @@ class Networking(Plugin):
         })
 
         self.add_copy_spec([
-            "/proc/net/",
-            "/etc/nsswitch.conf",
-            "/etc/yp.conf",
+            "/etc/dnsmasq*",
+            "/etc/host*",
             "/etc/inetd.conf",
+            "/etc/iproute2",
+            "/etc/network*",
+            "/etc/nsswitch.conf",
+            "/etc/resolv.conf",
             "/etc/xinetd.conf",
             "/etc/xinetd.d",
-            "/etc/host*",
-            "/etc/resolv.conf",
-            "/etc/network*",
-            "/etc/dnsmasq*",
+            "/etc/yp.conf",
+            "/proc/net/",
             "/sys/class/net/*/device/numa_node",
             "/sys/class/net/*/flags",
             "/sys/class/net/*/statistics/",
-            "/etc/iproute2"
         ])
 
         self.add_forbidden_path([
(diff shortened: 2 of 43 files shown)


sos/report/plugins/networking.py
```diff
@@ -246,14 +242,6 @@ def setup(self):
                     ], devices=_devs['ethernet'], priority=50, subdir=_subdir)
 
         self.add_cmd_tags({
-            "ethtool [^-].*": "ethtool",
```
pponnuvel, line 249: These tags were already done from lines 49-51.

sos/report/plugins/networking.py
```diff
@@ -136,6 +136,11 @@ def setup(self):
         macsec_pred = SoSPredicate(self, kmods=['macsec'])
         self.add_cmd_output(ip_macsec_show_cmd, pred=macsec_pred, changes=True)
 
+        self.collect_ss_ip_ethtool_info()
+        self.collect_bridge_info()
+
```
pponnuvel, line 141: Large function split.

sos/report/plugins/networking.py
```diff
@@ -295,23 +294,19 @@ def setup(self):
...
             "/etc/resolv.conf",
-            "/run/netplan/*.yaml",
-            "/etc/netplan/*.yaml",
+            "/etc/resolvconf",
             "/lib/netplan/*.yaml",
+            "/run/netplan/*.yaml",
             "/run/systemd/network"
         ])
 
-        if self.get_option("traceroute"):
```
pponnuvel, line 311: Already collected in the super class. This class just sets a different `trace_host`.

sos/report/plugins/npm.py
```diff
@@ -57,7 +57,4 @@ class NpmViaNodeJS(Npm):
     # also in Fedora 24+ it is just nodejs, no npm package
     packages = ('nodejs', )
 
-# TODO: in RHEL npm and nodejs is available via software collections
```
pponnuvel, line 60: This was originally added in 2016. 

If this still needs to be tracked, I can open a new issue and add a comment here.

sos/report/plugins/openstack_instack.py
```diff
@@ -61,19 +61,7 @@ def setup(self):
         vars_any = [p in os.environ for p in [
                     'OS_TENANT_NAME', 'OS_PROJECT_NAME']]
 
-        # capture yaml files to define overrides
-        uc_config = configparser.ConfigParser()
-        try:
-            uc_config.read(UNDERCLOUD_CONF_PATH)
-            override_opts = ['hieradata_override', 'net_config_override']
-            for opt in override_opts:
-                p = uc_config.get(opt)
```
pponnuvel, line 70: I believe this part never worked.

`configparser.get` requires 2 arguments: section and option - here it passes just the option.

sos/report/plugins/openstack_instack.py
```diff
@@ -112,6 +100,21 @@ def setup(self):
...
                                 "tripleo-ui-logs tripleo-ui.logs --file -")
 
+    def capture_undercloud_yamls(self):
+        """ capture yaml files to define overrides """
+        uc_config = configparser.ConfigParser()
+        try:
+            uc_config.read(UNDERCLOUD_CONF_PATH)
+            override_opts = ['hieradata_override', 'net_config_override']
+            for opt in override_opts:
+                path = uc_config.get('DEFAULT', opt)
```
pponnuvel, line 110: These two "override" options are available under `DEFAULT` section as https://access.redhat.com/documentation/en-us/red_hat_openstack_platform/10/html/director_installation_and_usage/chap-installing_the_undercloud

So I pass the section argument to `DEFAULT`.

sos/report/plugins/origin.py
```diff
@@ -107,7 +107,8 @@ def setup(self):
             if self.is_static_etcd():
                 self.add_cmd_output("%s etcd etcd" % static_pod_logs_cmd)
 
-            # TODO: some thoughts about information that might also be useful
+            # Possible enhancements:
```
pponnuvel, line 110: This doesn't feel like a TODO. So reworded it.

If this does need to be tracked, I can open a new issue and link here.

sos/report/plugins/openvswitch.py
```diff
@@ -32,33 +33,10 @@ def setup(self):
...
-            "/etc/fpm.env",
-            "/etc/6WINDGate/fp.config",
-            "/etc/6WINDGate/fpnsdk.config",
-            "/etc/dms.d/fp-dms.conf",
-            "/etc/dms.d/fpmd-dms.conf",
-            "/etc/dms.d/fpsd-dms.conf",
-            "/etc/fast-path.env",
-            "/etc/fps-fp.env",
-        ]
+        self.check_dpdk = (dpdk_enabled["status"] == 0 and
```
pponnuvel, line 36: This was a rather large function and had to refactor it a bit to solve too-many-locals, too-many-statements, etc.

TurboTurtle:
Marking this for later review since it is a bit large.

sos/report/plugins/networking.py
```diff
@@ -57,20 +57,20 @@ def setup(self):
...
 
         self.add_copy_spec([
-            "/proc/net/",
-            "/etc/nsswitch.conf",
-            "/etc/yp.conf",
+            "/etc/dnsmasq*",
+            "/etc/host*",
             "/etc/inetd.conf",
+            "/etc/iproute2",
+            "/etc/network*",
```
pponnuvel, line 64: Just sorted the list - no change.

sos/report/plugins/navicli.py
```diff
@@ -38,31 +38,31 @@ def get_navicli_config(self):
...
+            "navicli -h %s spportspeed -get" % sp_address
         ])
 
     def setup(self):
         self.get_navicli_config()
-        for ip in set(self.get_option("ipaddrs").split()):
-            if self.exec_cmd("navicli -h %s getsptime" % (ip))['status'] == 0:
-                self.get_navicli_SP_info(ip)
+        for addr in set(self.get_option("ipaddrs").split()):
+            if self.exec_cmd("navicli -h {addr} getsptime")['status'] == 0:
```
TurboTurtle, line 65: Missing leading `f` to make this an f-string.

sos/report/plugins/networking.py
```diff
@@ -262,6 +250,17 @@ def setup(self):
...
+    def collect_bridge_info(self):
+        """ Collect information about bridges (some data already collected via
+        "ip .." commands)
+        """
+        self.add_cmd_output([
+            "bridge -s -s -d link show",
+            "bridge -s -s -d -t fdb show",
+            "bridge -s -s -d -t mdb show",
+            "bridge -d vlan show"
+        ])
```
TurboTurtle, line 262: Why move this to a dedicated function instead of leaving it in `setup()`? It's a single call with no gating logic.

sos/report/plugins/networking.py
```diff
@@ -273,11 +272,11 @@ def setup(self):
             # major version
             if int(netstat_pkg['version'][0]) < 2:
                 self.ns_wide = "-T"
-        except Exception:
+        except Exception:  # pylint: disable=broad-except
```
TurboTurtle, line 275: Ack.

sos/report/plugins/openvswitch.py
TurboTurtle, line 1: I think this looks good, but with such a large change I'd like to see a validation test done, since I don't believe the test suite has any explicit checks on this plugin currently.

sos/report/plugins/ovirt.py
```diff
@@ -67,17 +67,17 @@ def setup(self):
...
                 self.soslog.error('Unable to get ovirt-engine pid')
                 self.add_alert('Unable to get ovirt-engine pid')
             for pid in engine_pids:
                 try:
                     # backtrace written to '/var/log/ovirt-engine/console.log
                     os.kill(pid, signal.SIGQUIT)
-                except OSError as e:
-                    self.soslog.error('Unable to send signal to %d' % pid, e)
+                except OSError:
+                    self.soslog.error('Unable to send signal to {pid}')
```
TurboTurtle, line 80: Missing leading `f` for f-strings.

sos/report/plugins/networking.py
pponnuvel: Was retrying to split/refactor a bit to fix too-many-{statements, locals, branches} warnings. You're right gating logic - aim was to split and thought "bridge" info could be one such independent part.

sos/report/plugins/navicli.py
pponnuvel: Fixed

sos/report/plugins/ovirt.py
pponnuvel: Fixed

sos/report/plugins/openvswitch.py
pponnuvel: Totally reasonable - I'll try to do some runs with openvswitch setup.

pponnuvel:
[openvswitch-sosreports.zip](https://github.com/sosreport/sos/files/14480803/openvswitch-sosreports.zip)

I've attached a zip that contains 2 sosreports (one with 4.5.6 and one with this branch) collected for openvswitch plugin. A quick check shows no issues. Please take a look as well. Thanks @TurboTurtle

sos/report/plugins/openvswitch.py
pponnuvel: https://github.com/sosreport/sos/pull/3540#issuecomment-1976380741 Attached sosreports

sos/report/plugins/networking.py
pmoravec: Good point!

For the sake of code "structure" or clarity, does it make sense to *move* the lines 49-51 here, like:

        for opt in self.ethtool_shortopts:
            self.add_cmd_tags({
                'ethtool -%s .*' % opt: 'ethool_%s' % opt
            })
        self.add_cmd_tags({
            "ip -d address": "ip_addr",
            ..

? To have all tags on one place.

pmoravec: Well, the

    "ethtool [^-].*": "ethtool",

should remain there - it is not covered by the loop on lines 49-51 and we call `ethtool %(dev)s` on line 144.

pponnuvel: Good catch!
`"ethtool [^-].*": "ethtool",` is back now.

pponnuvel: Also moved the `add_cmd_tags` together but moved into a new function on account of too-many-locals.

TurboTurtle (approved):

PR #3541, opened 2024-02-29
[openstack_keystone] Obfuscate OIDC client secret

jcastill:
Obfuscate OIDC client secret inside
/var/lib/config-data/puppet-generated/keystone/
	etc/httpd/conf.d/10-keystone_wsgi.conf 

The secret looks like this:
 
 OIDCClientSecret "Password"

And after obfuscation, it will look like this:

OIDCClientSecret *********

Related: RH: RHEL-26720

---

### DIFF
--- a/sos/report/plugins/openstack_keystone.py
+++ b/sos/report/plugins/openstack_keystone.py
@@ -91,17 +91,22 @@ class OpenStackKeystone(Plugin):
             self.var_puppet_gen + "/etc/keystone/*",
             regexp, subst
         )
+        self.do_path_regex_sub(
+            self.var_puppet_gen + "/etc/httpd/conf.d/",
+            regexp, subst
+        )
 
     def postproc(self):
         protect_keys = [
             "password", "qpid_password", "rabbit_password", "ssl_key_password",
             "ldap_dns_password", "neutron_admin_password", "host_password",
-            "admin_password", "admin_token", "ca_password", "transport_url"
+            "admin_password", "admin_token", "ca_password", "transport_url",
+            "OIDCClientSecret",
         ]
         connection_keys = ["connection"]
 
         self.apply_regex_sub(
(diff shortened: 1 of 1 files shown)


sos/report/plugins/openstack_keystone.py
```diff
@@ -106,6 +112,10 @@ def postproc(self):
             "|".join(connection_keys),
             r"\1*********\6"
         )
+        self.apply_regex_sub(
+            r"(^\s*(%s)\s*)(.*)" % "|".join(client_secrets),
+            r"\1*********"
+        )
```
pmoravec, line 118: The secret can be in either `/etc/keystone` config or in `httpd` config, such that you call `apply_regex_sub` and not `do_path_regex_sub`, right?

jcastill: That was my understanding from the report and the doc, but let me doublecheck.

TurboTurtle, line 118: Could we not add `OIDCClientSecret` to the existing list of keys, and make the existing regex just have the `=` optional? E.G regex in line 107 becomes `r"(^\s*(%s)\s*(=\s*)?)(.*)"`, that way we can expand the keys being obfuscated without adding a call for each key?

TurboTurtle (commented):
Looks ok, but wondering if we can just expand the existing call. If that expansion causes unintended obfuscation then it is fine as is.

jcastill:
With @TurboTurtle 's suggestion it works perfectly. I can:
- Move OIDCClientSecret to the list 'protect_keys' and run the new substitution in line 107.
- Leave OIDCClientSecret in its own list but apply the new substitution in line 107. 

Which one is simpler and more readable? perhaps the first one?

pmoravec (approved):

TurboTurtle (approved):

Issue #3544, opened 2024-03-01
[Ubuntu] sos collect is throwing a deprecation warning when verifying the sos package version


dnegreira:
There is currently a deprecation warning happening when we run sos collect on Ubuntu:

```
connecting to nodes...
/usr/lib/python3/dist-packages/pkg_resources/init.py:116: PkgResourcesDeprecationWarning: 4.5.6-0ubuntu1~22.04.2 is an invalid version and will not be supported in a future release
  warnings.warn(
/usr/lib/python3/dist-packages/pkg_resources/init.py:116: PkgResourcesDeprecationWarning: 4.5.6-0ubuntu1~22.04.2 is an invalid version and will not be supported in a future release
  warnings.warn(
```

This is due to the pep440 specification not being honored, as in Ubuntu, the packages can have the tilde `~` and plus signal `+` in their [version string](https://github.com/canonical/ubuntu-maintainers-handbook/blob/main/VersionStrings.md).

There are [several fixes](https://bugs.launchpad.net/ubuntu/+source/distro-info/+bug/1991606) being done across a lot of packages in Ubuntu, and they [mostly consist of modifying the output of the package version](https://launchpadlibrarian.net/663090440/distro-info_1.1ubuntu0.1.debdiff) to comply with [python PEP440 specifications](https://packaging.python.org/en/latest/specifications/version-specifiers/#version-specifiers).

I aim to patch this in `sos collect` so that we can fix this issue on later versions of `sos` packaged into Ubuntu, and provide a patch to our `sos 4.7.0` packages as we are currently working on addressing some minor issues before we can push the new version into Ubuntu.

What I found so far is that we will need to patch [_format_version](https://github.com/sosreport/sos/blob/9090280863581f2828fac42e03fea52a4ba42d82/sos/collector/sosnode.py#L415) on soscollect to accomodate the `~` and `+`, and maybe add a new variable under the already existing `_ver` to properly format the output of `self.sos_info['version']`.

I will push a patch later requesting comments and a review, but my approach will be to do something similar to what we do in Ubuntu, and substitute the `~` and `+` by a `.`.

PR #3545, opened 2024-03-02
[collect] refactor _format_version

dnegreira:
Ensure that we format the package versions to the pep440 standard. This is necessary as we are using `parse_version()` from the `packaging` package to verify that the sos collector node version is equal or bigger than the versions we are collecting from the nodes. If we pass the wrong format to `parse_version()`, we are unable to do the comparison and unable to run `sos report` on those nodes.
This addresses the issue of the Ubuntu packaging version naming with `+` and `~` signals as well as generic versioning schemes.

Resolves: #3544 
---

### DIFF
--- a/sos/collector/sosnode.py
+++ b/sos/collector/sosnode.py
@@ -412,25 +412,29 @@ class SosNode():
         :returns:   True if installed version is at least ``ver``, else False
         :rtype:     ``bool``
         """
-        def _format_version(ver):
-            # format the version we're checking to a standard form of X.Y.Z-R
+        def _format_version_to_pep440(ver):
+            """ Convert the version into a PEP440 compliant version scheme."""
+            public_version_re = re.compile(
+                    r"^([0-9][0-9.]*(?:(?:a|b|rc|.post|.dev)[0-9]+)*)\+?"
+                    )
             try:
-                _fver = ver.split('-')[0]
-                _rel = ''
-                if '-' in ver:
-                    _rel = '-' + ver.split('-')[-1].split('.')[0]
-                if len(_fver.split('.')) == 2:
-                    _fver += '.0'
-
-                return _fver + _rel
+                _, public, local = public_version_re.split(ver, maxsplit=1)
+                if not local:
+                    return ver
+                sanitized_local = re.sub("[+~]+", ".", local).strip("-")
+                pep440_version = f"{public}+{sanitized_local}"
+                return pep440_version
             except Exception as err:
-                self.log_debug("Unable to format '%s': %s" % (ver, err))
+                self.log_debug(f"Unable to format {ver} to pep440 format: "
+                               f"{err}")
                 return ver
 
-        _ver = _format_version(ver)
+        _ver = _format_version_to_pep440(ver)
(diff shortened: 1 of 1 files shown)


dnegreira:
I have decided to refactor the whole _format_version method as it was not compatible with the pep440 standard.
I deviated of what I have initially writen on #3544, and I am not just substituting the `~` and `+` characters for a `.`, but making the format compatible with pep440.

Since I am re-writing / deprecating a whole method, I am happy to hear any kind of feedback, especially from @TurboTurtle as he has written this function in particular.

From my quick investigation with CentOS8, the package policy for `rpm` will always be compatible with the format since the version returned is the major version IE, `4.X`. But appreciate some testing from anyone who can do it from RedHat or any other distro.

Talking about Testing, I think some unit testing would be -super- useful in this case and I am willing to spend some cycles on doing it, either on this PR or on a new PR, but I would need a bit of handholding as I have never really done anything with avocado, and setting up the basic infrastructure so that we can mock and do all the neat things with collect.

sos/collector/sosnode.py
```diff
@@ -412,25 +412,29 @@ def check_sos_version(self, ver):
...
-                    _rel = '-' + ver.split('-')[-1].split('.')[0]
-                if len(_fver.split('.')) == 2:
-                    _fver += '.0'
-
-                return _fver + _rel
+                _, public, local = public_version_re.split(ver, maxsplit=1)
+                if not local:
+                    return ver
+                sanitized_local = re.sub("[+~]+", ".", local).strip("-")
+                pep440_version = "%s+%s" % (public, sanitized_local)
```
TurboTurtle, line 425: f-string here please.

sos/collector/sosnode.py
```diff
@@ -412,25 +412,29 @@ def check_sos_version(self, ver):
...
+                return pep440_version
             except Exception as err:
-                self.log_debug("Unable to format '%s': %s" % (ver, err))
+                self.log_debug("Unable to format '%s' to pep440 format: %s"
+                               % (ver, err))
```
TurboTurtle, line 429: Here too.

TurboTurtle (commented):
Overall looks good to me, minor comments below.

sos/collector/sosnode.py
dnegreira: Whooops, gotta remind myself of this now :)

fixed on: [fceb5f3](https://github.com/sosreport/sos/commit/fceb5f393bbeb5c9da1fa63877a2f8d2294f223f)

sos/collector/sosnode.py
dnegreira: fixed on: [fceb5f3](https://github.com/sosreport/sos/commit/fceb5f393bbeb5c9da1fa63877a2f8d2294f223f)

TurboTurtle (approved):

pmoravec (approved):
It is strange to see versions like `4.6.0-5` to be formatted to `4.6.0+5` but it really follows the [PEP specification](https://peps.python.org/pep-0440/#local-version-identifiers).

PR #3553, opened 2024-03-05
[AAP RECEPTOR] Add a new AAP receptor plugin.

rbertol:
Adding the file 'aap_receptor.py' for the sos report collects 
the files used for troubleshooting issues at
Ansible Automation Platform Controller/Execution Nodes

Related: RH AAP-19657

---

### DIFF
--- a/sos/report/plugins/aap_receptor.py
+++ b/sos/report/plugins/aap_receptor.py
@@ -1,0 +1,53 @@
+# Copyright (c) 2024 Rudnei Bertol Jr <rudnei@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+import glob
+from sos.report.plugins import Plugin, RedHatPlugin
+
+
+class AAPreceptorPlugin(Plugin, RedHatPlugin):
+    short_desc = 'AAP receptor plugin'
+    plugin_name = 'aap_receptor'
+    profiles = ('sysmgmt', 'ansible')
+    packages = ('receptor', 'receptorctl')
+    services = ('receptor',)
+
+    def setup(self):
+        self.add_copy_spec([
+            "/etc/receptor",
+            "/var/lib/receptor",
+        ])
+
+        if self.get_option("all_logs"):
+            self.add_copy_spec([
+                "/var/log/receptor"
+            ])
+        else:
+            self.add_copy_spec([
+                "/var/log/receptor/*.log"
+            ])
+
+        self.add_forbidden_path([
+            "/etc/receptor/tls",
+            "/etc/receptor/*key.pem"
+        ])
+
+        self.add_cmd_output([
+            "ls -llZ /etc/receptor",
+            "ls -llZ /var/run/receptor",
+            "ls -llZ /var/run/awx-receptor"
+        ])
+
+        for s in glob.glob('/var/run/*receptor/*.sock'):
+            self.add_cmd_output(f"receptorctl --socket {s} status",
+                                suggest_filename="receptorctl_status")
+            break
+
+# vim: set et ts=4 sw=4 :


sos/report/plugins/aap_receptor.py
```diff
@@ -0,0 +1,53 @@
...
+        self.add_cmd_output([
+            "ls -ll /etc/receptor",
+            "ls -ll /var/run/receptor",
+            "ls -ll /var/run/awx-receptor"
+        ])
+
+        for s in glob.glob('/var/run/*receptor/*.sock'):
+            receptorctl = []
+            receptorctl.append(f"receptorctl --socket {s} status")
+        self.add_cmd_output(receptorctl, suggest_filename="receptorctl_status")
```
TurboTurtle, line 51: The `suggest_filename` here will give non-deterministic behavior when used against a list, and we generally don't want to use it as a way to mask what is being passed to a command unless what is being dropped either 1) makes things significantly more human-readable and/or 2) does not actually impact the understanding of what is being collected.

Also, resetting `receptorctl` inside the for loop would ensure that this list is only ever be 1 item long. We can also be a bit cleaner in the list generation here:

```python
self.add_cmd_output([
    f"receptorctl --socket {path} status" for path in
    glob.glob('/var/run/*receptor/*.sock')
])
```

sos/report/plugins/aap_receptor.py
```diff
@@ -0,0 +1,53 @@
...
+        self.add_forbidden_path([
+            "/etc/receptor/tls",
+            "/etc/receptor/*key.pem"
+        ])
+
+        self.add_cmd_output([
+            "ls -ll /etc/receptor",
+            "ls -ll /var/run/receptor",
+            "ls -ll /var/run/awx-receptor"
+        ])
```
TurboTurtle, line 46: Typically for `ls` collections we'll use `-alhR`. Are we not interested in recursive listing here? Do we need to consider `-Z` for selinux labeling? (On that point, note to self: we really need to look at if we want a wrapper for dir listing collections).

sos/report/plugins/aap_receptor.py
rbertol: Regarding the name of the file, the suggested name is just omitting the socket path, which is huge. But on receptor, it will be just one socket, not sure why the initial plugin had this for condition. I am working on this change.

sos/report/plugins/aap_receptor.py
rbertol: I liked the idea for the selinux, it will help us in case of mislabeling. Working on this change as well.

sos/report/plugins/aap_receptor.py
pmoravec: Also please ensure the code works fine when no receptor socket exists. Running it on a cirrus testing system, we got an [exception](https://cirrus-ci.com/task/5100827342274560?logs=main#L205):

      File "/tmp/cirrus-ci-build/sos/report/__init__.py", line 1230, in setup
        plug.setup()
      File "/tmp/cirrus-ci-build/sos/report/plugins/aap_receptor.py", line 51, in setup
        self.add_cmd_output(receptorctl, suggest_filename="receptorctl_status")
    UnboundLocalError: local variable 'receptorctl' referenced before assignment

as `glob.glob('/var/run/*receptor/*.sock') = []`.

sos/report/plugins/aap_receptor.py
```diff
@@ -0,0 +1,53 @@
+# Copyright (c) 2023 Rudnei Bertol Jr <rudnei@redhat.com>
```
pmoravec, line 1: A nitpick: `s/2023/2024` :)

sos/report/plugins/aap_receptor.py
```diff
@@ -0,0 +1,53 @@
...
+        self.add_cmd_output([
+            "ls -llZ /etc/receptor",
+            "ls -llZ /var/run/receptor",
+            "ls -llZ /var/run/awx-receptor"
+        ])
+
+        for s in glob.glob('/var/run/*receptor/*.sock'):
+            self.add_cmd_output(f"receptorctl --socket {s} status",
+                                suggest_filename="receptorctl_status")
+            break
```
pmoravec, line 51: Adding some context: it is expected there is just one socket file and we should collect status for it. Just in case there might be no or multiple sockets, we need the `for ..` and `break`.

pmoravec (approved):

TurboTurtle (approved):

PR #3554, opened 2024-03-05
[pacemaker] Use pep440 formatted version for pcs version

### LINKED ISSUE #3548: parse_version fails on Ubuntu
As noted in https://github.com/sosreport/sos/issues/3544, there are several packages on Ubuntu that do not conform to PEP440.

```
  File "/home/ponnuvel/myGit/sos/sos/report/__init__.py", line 1230, in setup
    plug.setup()
  File "/home/ponnuvel/myGit/sos/sos/report/plugins/pacemaker.py", line 151, in setup
    self.setup_pcs()
  File "/home/ponnuvel/myGit/sos/sos/report/plugins/pacemaker.py", line 61, in setup_pcs
    if parse_version(pcs_version) > parse_version('0.10.8'):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/packaging/version.py", line 52, in parse
    return Version(version)
           ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/packaging/version.py", line 197, in __init__
    raise InvalidVersion(f"Invalid version: '{version}'")
packaging.version.InvalidVersion: Invalid version: '0.11.4-1ubuntu3'
```

This is from pacemaker plugin. Probably needs to be fixed in collector/clusters/pacemaker.py too.

pponnuvel:
There are couple of instances (both on pacemaker) of `parse_version` being used to compare the package
versions. In cases, notably on Ubuntu, where the version comform to PEP440, this fails. So we now convert those to PEP440 format before comparing.

Fixes #3548.

---

### DIFF
--- a/sos/collector/clusters/juju.py
+++ b/sos/collector/clusters/juju.py
@@ -13,7 +13,7 @@ import json
 import re
 
 from sos.collector.clusters import Cluster
-from sos.utilities import parse_version
+from sos.utilities import sos_parse_version
 from sos.utilities import sos_get_command_output
 
 
@@ -161,12 +161,13 @@ class juju(Cluster):
     def _get_juju_version(self):
         """Grab the version of juju"""
         res = sos_get_command_output("juju version")
-        return res['output'].split("-")[0]
+        return res['output']
 
     def _execute_juju_status(self, model_name):
         model_option = f"-m {model_name}" if model_name else ""
         format_option = "--format json"
-        if parse_version(self._get_juju_version()) > parse_version("3"):
+        juju_version = self._get_juju_version()
+        if sos_parse_version(juju_version) > sos_parse_version("3"):
             format_option += " --no-color"
         status_cmd = f"{self.cmd} status {model_option} {format_option}"
         res = self.exec_primary_cmd(status_cmd)
--- a/sos/collector/clusters/pacemaker.py
+++ b/sos/collector/clusters/pacemaker.py
@@ -11,7 +11,7 @@
 import re
 
 from sos.collector.clusters import Cluster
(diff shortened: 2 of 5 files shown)


TurboTurtle (commented):
I like `format_version_to_pep440()` being moved to utilities, however I'm now thinking that we should just have `parse_version()` call that internally as part of it's flow, so that we don't need to "manually" do that formatting when we do comparisons like this.

sos/collector/clusters/juju.py
```diff
@@ -161,12 +161,13 @@ def _get_model_info(self, model_name):
     def _get_juju_version(self):
         """Grab the version of juju"""
         res = sos_get_command_output("juju version")
-        return res['output'].split("-")[0]
+        return res['output']
```
pponnuvel, line 164: I think this split was done to workaround `parse_version` as the juju version would look like `2.9.46-ubuntu-amd64`.

pponnuvel:
> I like `format_version_to_pep440()` being moved to utilities, however I'm now thinking that we should just have `parse_version()` call that internally as part of it's flow, so that we don't need to "manually" do that formatting when we do comparisons like this.

Yeah, makes sense. Still left an option to use "plain" `parse_version` if needed.

dnegreira:
@pponnuvel is the PR ready to be tested? I would like to give it a quick spin on my environments.

dnegreira:
Tested on a juju 2.9 cloud and all looks fine, thanks for the help!

dnegreira (approved):

TurboTurtle (approved):

pmoravec (approved):

PR #3557, opened 2024-03-06
[proxmox] Add new plugin for Proxmox Virtualization Environment

### LINKED ISSUE #3519: implement PROXMOX / PVE plugin
Hi team,

Running sos on a proxmox/pve host does not produce all expected logs (VM inventory, storage pools, SDN .. )

PVE is 100% api driven and gathering all those info is quite simple

I started to ask one of my dev guy to gather all commands (mostly with `pvesh`)

My question here is : how may we help you ?
- give you all the commands + directories to copy to sos output ?
- start from a plugin template and modify it ? ==> which one 
- start from scratch in python ?

Thx

PS : we will do the same with Proxmox Backup Server / PBS after 
PS2 : all my "recent" activity around sos is to help the opensource OGrEE Datacenter Digital Twin project which I started 4 years ago. (https://ogree.ditrit.io) (I've been using sos since years for linux hosts for that twin, with some other collect tools that run on other platforms (vmware,windows, ibm AIX... we could discuss about that elsewhere.

Hervé

sohzaz:
Resolves #3519 

---

### DIFF
--- a/sos/report/plugins/proxmox.py
+++ b/sos/report/plugins/proxmox.py
@@ -1,0 +1,101 @@
+# Copyright (C) 2024 ORNESS/Ditrit Drien Breton <drien.breton@orness.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+import json
+
+from sos.report.plugins import Plugin, DebianPlugin, PluginOpt
+
+
+class Proxmox(Plugin, DebianPlugin):
+    """
+    This plugin will capture information about the system's
+    Proxmox Virtualization Environment.
+
+    It will collect information about the cluster, nodes, pools and
+    storage from the Proxmox API.
+    """
+    short_desc = 'Proxmox cluster information'
+
+    plugin_name = 'proxmox'
+    packages = ('proxmox-ve',)
+
+    option_list = [
+        PluginOpt('output_formats',
+                  desc='List of output formats to use '
+                       'for the commands separated by ":".',
+                  default='text',
+                  ),
+    ]
+
+    def setup(self):
+        output_formats = self.get_option('output_formats').split(':')
+
+        commands = [
+            'cluster/resources',
+            'cluster/config/nodes',
+            'cluster/options',
+            'nodes/:id/status',
+            'nodes/:id/storage',
+            'nodes/:id/network',
+            'pools',
+            'storage',
+            'storage/:sid',
+            'cluster/ceph/status',
+            'cluster/ceph/metadata',
+            'cluster/ceph/flags',
+        ]
+
+        cmd_paths = []
+
+        for command in commands:
+            cmd_paths.extend(self.build_cmd_paths(command))
+
+        self.add_cmd_output([
+            f"pvesh get {cmd} {'--noborder' if format == 'text' else ''} "
+            f"--output-format {format}"
+            for cmd in cmd_paths
+            for format in output_formats
+        ])
+
+    def build_cmd_paths(self, base_path):
+        """
+        Build command paths,
+        replacing dynamic attributes with data from the API
+
+        :param base_path: The base path to build from
+        :type base_path: str
+
+        :return: A list of paths
+        """
+        parts = base_path.split('/')
+        paths = []
+
+        id_index = next((i for i, part in enumerate(parts)
+                         if part.startswith(':')), None)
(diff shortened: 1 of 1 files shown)


sos/report/plugins/proxmox.py
```diff
@@ -0,0 +1,114 @@
...
+    """
+    short_desc = 'Proxmox cluster information'
+
+    plugin_name = 'proxmox'
+    packages = ('proxmox-ve',)
+
+    option_list = [
+        PluginOpt('output_formats',
+                  desc='List of output formats to use '
+                       'for the commands separated by ":".',
```
pmoravec, line 32: We usually use comma as items separator (e.g. `sos report -o kernel,logs` to enable only those two plugins). I am not against a different separator here, just noting that this is unusual (and depends on users' preferences what is more natural for them).

TurboTurtle: We use commas for top-level options, we are _unable_ to use them in plugin options due to how argparse parses them vs. our expectation of slicing multiple different plugin options at once. Colon delimiters in plugin options is the normal path.

E.G. `sos report -o kernel,logs,proxmox` is fine, as is `sos report -o proxmox,logs -k proxmox.output_formats=json,logs.all=on`, but `sos report -o proxmox,logs -k proxmox.output_format=json,text` would not get parsed properly.

sos/report/plugins/proxmox.py
```diff
@@ -0,0 +1,114 @@
...
+                  default='text',
+                  ),
+        PluginOpt('cmd_override_file',
+                  desc='File containing the commands '
+                       'to run instead of the default ones.',
+                  default=None,
+                  val_type=[None, Path, str],
+                  ),
```
TurboTurtle, line 40: Hard nack.

We have the `sos_extras` plugin to handle extra collections that aren't in plugins, and it serves entirely as a stop-gap measure before those commands get contributed and released.

sos/report/plugins/proxmox.py
```diff
@@ -0,0 +1,114 @@
...
+                for line in f:
+                    cmd_paths.extend(self.build_cmd_paths(line.strip()))
+        else:
+            for command in commands:
+                cmd_paths.extend(self.build_cmd_paths(command))
+
+        for cmd_path in cmd_paths:
+            for output_format in output_formats:
+                cmd = f'pvesh get {cmd_path} --output-format {output_format}'
+                file_name = cmd.replace("/", ".").replace(" ", "_")
```
TurboTurtle, line 75: We do this internally in our command-to-file name mangling, so there is no need to do this yourself or to use `suggest_filename` here for it.

sos/report/plugins/proxmox.py
```diff
@@ -0,0 +1,114 @@
...
+                self.add_cmd_output(
+                    f'{cmd} --noborder',
+                    suggest_filename=file_name
+                )
```
TurboTurtle, line 79: With the above filename mangling in mind, we can button this whole block up like so, with proper PEP8 adherence of course:

```python
self.add_cmd_output([
    f"pvesh get {cmd} --output-format {format}" for cmd in cmd_paths for format in output_formats
])
```

sos/report/plugins/proxmox.py
```diff
@@ -0,0 +1,114 @@
...
+                for child in children:
+                    paths.extend(
+                        self.build_cmd_paths(
+                            f"{path}/{child['name']}/{trailing_path}"
+                        )
+                    )
+
+        else:
+            paths.append('/'.join(parts))
+        return paths
```
TurboTurtle, line 114: Nit: we end plugins with vim tags, e.g.:

```
        return paths

# vim: set et ts=4 sw=4 :
```

sos/report/plugins/proxmox.py
```diff
@@ -0,0 +1,114 @@
...
+            'cluster/ceph/status'
+            'cluster/ceph/metadata'
+            'cluster/ceph/flags'
+        ]
+
+        cmd_paths = []
+
+        override_file = self.get_option('cmd_override_file')
+        if override_file is not None:
+            with open(override_file, 'r') as f:
```
pmoravec, line 65: Could you please add `, encoding='UTF-8'`, likewise we are improving in all plugins (cf. e.g. https://github.com/sosreport/sos/pull/3546/files#diff-ded1fd003cd94bd33e0094c2e211a655c38c53a382badd42b0784a6bd0e37f83R136)?

pmoravec (approved):
ACK, just please squash all commits into one.

Thanks!

TurboTurtle (approved):
LGTM.

The Cirrus task timed out, so I restarted it. Assuming that and the re-started stage two jobs pass we'll get this merged shortly.

Issue #3559, opened 2024-03-09
Drop any SCL related code?


pmoravec:
sos report supports collecting data from Software Collections. That is roughly speaking RHEL7 specifics which we dont support in sos upstream for years. So we shall drop any `*scl` methods or `SCLPlugin` class completely, I think.

Since the SCL stuff is bundled into multiple plugins (foreman, postgres, redis, at least), we should review those plugins if we dont cause any regression by dropping the SCL methods / class(es).

TurboTurtle:
That's what I couldn't remember, if there was any bleed over into RHEL 8 or not.

I'm 100% onboard with dropping SCL wholesale. I'll work on that this weekend/earlier this coming week.

TurboTurtle:
Actually, this looks more straight forward than I was thinking...I'll have something posted shortly.

PR #3560, opened 2024-03-09
[SCLPlugin] Drop SCLPlugin support

TurboTurtle:
SCL is not used in RHEL past RHEL 7, and sos has not directly supported RHEL 7 for some years now due to the switch to python3 that came along with sos-4.0.

Drop SCLPlugin, as it will never be used in modern sos. Cleanup any references to it in existing plugins as well.

Resolves: #3559

---

### DIFF
--- a/sos/report/plugins/postgresql.py
+++ b/sos/report/plugins/postgresql.py
@@ -14,8 +14,8 @@
 
 import os
 
-from sos.report.plugins import (Plugin, UbuntuPlugin, DebianPlugin, SCLPlugin,
-                                PluginOpt)
+from sos.report.plugins import (Plugin, UbuntuPlugin, DebianPlugin,
+                                RedHatPlugin, PluginOpt)
 from sos.utilities import find
 
 
@@ -45,7 +45,7 @@ class PostgreSQL(Plugin):
                   desc='database server listening port')
     ]
 
-    def do_pg_dump(self, scl=None, filename="pgdump.tar"):
+    def do_pg_dump(self, filename="pgdump.tar"):
         if self.get_option("dbname"):
             if self.get_option("password") or "PGPASSWORD" in os.environ:
                 # We're only modifying this for ourself and our children so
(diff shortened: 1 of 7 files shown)


TurboTurtle:
The one oddity to me from this is the `redis` plugin. It was originally _just_ an SCL plugin, but redis is definitely used more widely than that.

I linked this up to `IndependentPlugin` and stripped the SCL bits, but I'm not sure how useful that makes the plugin for modern `redis` deployments, since it seems we've never actually had a functional plugin for it outside of SCL.

jcastill:
Do we need to remove the following references as well?

https://github.com/sosreport/sos/blob/main/sos/policies/distros/redhat.py#L46

`    default_scl_prefix = '/opt/rh'`

https://github.com/sosreport/sos/blob/main/sos/policies/__init__.py#L119

`    default_scl_prefix = ""`

https://github.com/sosreport/sos/blob/main/sos/policies/__init__.py#L269

```
    def get_default_scl_prefix(self):
        return self.default_scl_prefix
```

Sorry if these are not related or if I missed something and it doesn't apply.

pmoravec:
+1 for dropping also the `*scl_prefix` stuff.

sos/report/plugins/postgresql.py
```diff
@@ -85,34 +83,14 @@ def setup(self):
...
 
 
-class RedHatPostgreSQL(PostgreSQL, SCLPlugin):
-
-    packages = (
-        'postgresql',
-        'rh-postgresql95-postgresql-server',
-        'rh-postgresql10-postgresql-server',
-        'rh-postgresql12-postgresql-server',
-    )
```
pmoravec, line 95: We should not introduce any kind of regression by dropping `postgres-server` from `packages` here. Since `postgresql-server` package has a dependency on `postgresql` package already. ACK.

pmoravec:
Apart of the `*scl_prefix` bits, ACK. I can confirm `redis` plugin (updated per this PR) collects something relevant (well, all we need) on Satellite6 so some use case is covered well. For generic plugin purposes, I can't find what might be missing or wrong there, so ACK to the plugin changes also.

TurboTurtle:
Good eye on the prefixes. Dropped and re-pushed.

pmoravec (approved):

PR #3561, opened 2024-03-10
[aap_controller] Implementing AAP Automation Controller Plugin

lucas-benedito:
Adding the file 'aap_controller.py' for the sos report collects
The files used for troubleshooting issues at
Ansible Automation Platform Automation Controller

Related: RH AAP-18409

Signed-off-by: Lucas Benedito <lbenedit@redhat.com>

---

### DIFF
--- a/sos/report/plugins/aap_controller.py
+++ b/sos/report/plugins/aap_controller.py
@@ -1,0 +1,88 @@
+# Copyright (c) 2024 Pavel Moravec <pmoravec@redhat.com>
+# Copyright (c) 2024 Lucas Benedito <lbenedit@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, RedHatPlugin
+
+
+class AAPControllerPlugin(Plugin, RedHatPlugin):
+    short_desc = 'AAP Automation Controller plugin'
+    plugin_name = 'aap_controller'
+    profiles = ('sysmgmt', 'ansible',)
+
+    packages = ('automation-controller-venv-tower',
+                'automation-controller-server',
+                'automation-controller-ui',
+                'automation-controller')
+    commands = ('awx-manage',)
+    services = ('automation-controller',)
+
+    def setup(self):
+        self.add_copy_spec([
+            "/etc/tower/",
+            "/etc/supervisord.conf",
+            "/etc/supervisord.d/*",
+            "/var/log/tower",
+            "/var/log/nginx/automationcontroller.access.log*",
+            "/var/log/nginx/automationcontroller.error.log*",
+            "/var/log/supervisor",
+            "/var/log/unattended-upgrades",
+        ])
+
+        self.add_forbidden_path([
+            "/etc/tower/SECRET_KEY",
+            "/etc/tower/*.key",
+            "/etc/tower/*.cert",
+            "/var/log/tower/profile",
+        ])
+
+        self.add_cmd_output([
+            "awx-manage --version",
+            "awx-manage list_instances",
+            "awx-manage run_dispatcher --status",
+            "awx-manage run_callback_receiver --status",
+            "awx-manage check_license --data",
+            "awx-manage run_wsbroadcast --status",
+            "awx-manage run_wsrelay --status",
+            "supervisorctl status",
+            "/var/lib/awx/venv/awx/bin/pip freeze",
+            "/var/lib/awx/venv/awx/bin/pip freeze -l",
+            "/var/lib/awx/venv/ansible/bin/pip freeze",
+            "/var/lib/awx/venv/ansible/bin/pip freeze -l",
+            "tree -d /var/lib/awx",
+            "ls -ll /var/lib/awx",
+            "ls -ll /var/lib/awx/venv",
+            "ls -ll /etc/tower",
+            "umask -p",
+        ])
+
+    def postproc(self):
+        # remove database password
+        jreg = r"(\s*\'PASSWORD\'\s*:(\s))(?:\"){1,}(.+)(?:\"){1,}"
+        repl = r"\1********"
+        self.do_path_regex_sub("/etc/tower/conf.d/postgres.py", jreg, repl)
+
+        # remove email password
+        jreg = r"(EMAIL_HOST_PASSWORD\s*=)\'(.+)\'"
+        repl = r"\1********"
+        self.do_path_regex_sub("/etc/tower/settings.py", jreg, repl)
+
+        # remove email password (if customized)
+        jreg = r"(EMAIL_HOST_PASSWORD\s*=)\'(.+)\'"
+        repl = r"\1********"
+        self.do_path_regex_sub("/etc/tower/conf.d/custom.py", jreg, repl)
+
+        # remove websocket secret
+        jreg = r"(BROADCAST_WEBSOCKET_SECRET\s*=\s*)\"(.+)\""
+        repl = r"\1********"
+        self.do_path_regex_sub("/etc/tower/conf.d/channels.py", jreg, repl)
+
+
+# vim: set et ts=4 sw=4 :


sos/report/plugins/aap_controller.py
```diff
@@ -0,0 +1,96 @@
+# Copyright (c) 2023 Pavel Moravec <pmoravec@redhat.com>
```
TurboTurtle, line 1: Copy error?

sos/report/plugins/aap_controller.py
```diff
@@ -0,0 +1,96 @@
...
+    short_desc = 'AAP Automation Controller plugin'
+    plugin_name = 'aap_controller'
+    profiles = ('sysmgmt', 'ansible')
+
+    packages = ('automation-controller-venv-tower',
+                'automation-controller-server',
+                'automation-controller-ui',
+                'automation-controller')
+    commands = ('awx-manage')
+    services = ('automation-controller')
```
TurboTurtle, line 25: These need to be tuples.

sos/report/plugins/aap_controller.py
```diff
@@ -0,0 +1,96 @@
...
+            "/var/log/supervisor",
+            "/var/log/unattended-upgrades",
+        ])
+
+        self.add_forbidden_path([
+            "/etc/tower/SECRET_KEY",
+            "/etc/tower/tower.key",
+            "/etc/tower/awx.key",
+            "/etc/tower/tower.cert",
+            "/etc/tower/awx.cert",
```
TurboTurtle, line 45: Should these not be expanded to something like `/etc/tower/*.key` and `/etc/tower/*.cert`? I don't know if it's possible for other filenames to be used here, but if so I imagine we don't want to copy any of them.

sos/report/plugins/aap_controller.py
```diff
@@ -0,0 +1,96 @@
...
+            "/var/lib/awx/venv/awx/bin/pip freeze -l"
+            "/var/lib/awx/venv/ansible/bin/pip freeze"
+            "/var/lib/awx/venv/ansible/bin/pip freeze -l"
+            "tree -d /var/lib/awx"
+            "ls -ll /var/lib/awx"
+            "ls -ll /var/lib/awx/venv"
+            "ls -ll /etc/tower",
+            "ls -ll /var/run/awx-receptor"
+            "ls -ll /etc/receptor",
+            "receptorctl --socket /var/run/awx-receptor/receptor.sock status"
```
TurboTurtle, line 70: Aren't these collected by the plugin proposed in #3553?

sos/report/plugins/aap_controller.py
```diff
@@ -0,0 +1,96 @@
...
+
+        self.add_forbidden_path([
+            "/etc/tower/SECRET_KEY",
+            "/etc/tower/tower.key",
+            "/etc/tower/awx.key",
+            "/etc/tower/tower.cert",
+            "/etc/tower/awx.cert",
+            "/var/log/tower/profile",
+            "/etc/receptor/tls/ca/*.key",
+            "/etc/receptor/tls/*.key",
```
TurboTurtle, line 48: Aren't these also collected by the proposed receptor plugin in #3553?

jcastill:
Thanks for the new plugin @lucas-benedito ! A couple of things:
- Could you squash all four commits into one?
- Could you sign the resulting commit?

sos/report/plugins/aap_controller.py
pmoravec: I can leave with it.. :)

In fact, the plugin was internally prepared in a collaborative manner by few fellows, also with my contribution. I would vote for putting here my and Lucas's name, but I am ok with either option.

sos/report/plugins/aap_controller.py
pmoravec: I *think* - in general - that all keys (private part of cert) should be forbidden, while there can be some reason to copy a cert file (public part, yet internally for the deployment) in some cases. Don't know how this applies to the specific keys / certs in `/etc/tower`.

sos/report/plugins/aap_controller.py
lucas-benedito: @pmoravec++ made the initial draft and i just adjusted some points

sos/report/plugins/aap_controller.py
lucas-benedito: Ack, will remove the receptor references. thanks

sos/report/plugins/aap_controller.py
lucas-benedito: Ack, will remove the receptor references. thanks

sos/report/plugins/aap_controller.py
lucas-benedito: Changed it as suggested.

sos/report/plugins/aap_controller.py
lucas-benedito: done

sos/report/plugins/aap_controller.py
```diff
@@ -0,0 +1,88 @@
...
+    services = ('automation-controller',)
+
+    def setup(self):
+        self.add_copy_spec([
+            "/etc/tower/",
+            "/etc/supervisord.conf",
+            "/etc/supervisord.d/*",
+            "/var/log/tower",
+            "/var/log/nginx/automationcontroller.access.log*",
+            "/var/log/nginx/automationcontroller.error.log*",
```
pmoravec, line 34: These file patterns should be forbidden in `nginx` plugin (did I write a comment about it here or to a different PR? I cant find it now..)

pmoravec (approved):
LGTM, thanks!

Just please amend the commit to follow DCO (signed-off and ideally also Closes: #3561).

TurboTurtle (approved):

sos/report/plugins/aap_controller.py
```diff
@@ -0,0 +1,88 @@
...
+            "/var/log/tower/profile",
+        ])
+
+        self.add_cmd_output([
+            "awx-manage --version"
+            "awx-manage list_instances"
+            "awx-manage run_dispatcher --status"
+            "awx-manage run_callback_receiver --status"
+            "awx-manage check_license --data"
+            "awx-manage run_wsbroadcast --status"
```
TurboTurtle, line 52: Missing commas throughout the list

TurboTurtle (changes requested):
Comments above, from codeQL scan. Good sanity check here!

jskacel:
@lucas-benedito can we add also `podman images` and `podman ps`? Would be helpful I believe to see that. Maybe `podman info` as well.

arif-ali:
Curious on this, and I could be wrong here.

Isn't Ansible Tower/AWX something that can be used anywhere else. It seems like these are the main things we are collecting here. i.e. AWX specifically could be potentially be used on other distros, so we could make the name of the plugin more generic so that it can be used in other distros if needed in the future. I know ubuntu doesn't have packages yet, but could happen

TurboTurtle:
> @lucas-benedito can we add also `podman images` and `podman ps`? Would be helpful I believe to see that. Maybe `podman info` as well.

These are already captured by the `podman` plugin.

TurboTurtle:
> > > @lucas-benedito can we add also `podman images` and `podman ps`? Would be helpful I believe to see that. Maybe `podman info` as well.
> > 
> > 
> > These are already captured by the `podman` plugin.
> 
> not for awx user where all processes and images are running for AAP

You'd need to use the `containers_common` plugin for that, e.g. `sos report -k containers_common.rootlessusers=awx`, or we could hook that up to a preset for AAP as well.

For reference, everything in sos is collected by root. All collections from this plugin would be as root, and plugins _generally_ don't have the ability to change their execution context. Even the above plugin uses `machinectl` to run in the context of the specified user(s), but it is still being executed as root.

lucas-benedito:
> Curious on this, and I could be wrong here.
> 
> Isn't Ansible Tower/AWX something that can be used anywhere else. It seems like these are the main things we are collecting here. i.e. AWX specifically could be potentially be used on other distros, so we could make the name of the plugin more generic so that it can be used in other distros if needed in the future. I know ubuntu doesn't have packages yet, but could happen

Based on the information we have, there are no plans for making the RPMs available for platforms other than RHEL.

sos/report/plugins/aap_controller.py
```diff
@@ -0,0 +1,88 @@
...
+            "/var/lib/awx/venv/ansible/bin/pip freeze"
+            "/var/lib/awx/venv/ansible/bin/pip freeze -l"
+            "tree -d /var/lib/awx"
+            "ls -ll /var/lib/awx"
+            "ls -ll /var/lib/awx/venv"
+            "ls -ll /etc/tower",
+            "umask -p",
+        ])
+
+        def postproc(self):
```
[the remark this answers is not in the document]
lucas-benedito: Done.

pmoravec (approved):

lucas-benedito:
> > > > @lucas-benedito can we add also `podman images` and `podman ps`? Would be helpful I believe to see that. Maybe `podman info` as well.
> > > 
> > > 
> > > These are already captured by the `podman` plugin.
> > 
> > 
> > not for awx user where all processes and images are running for AAP
> 
> You'd need to use the `containers_common` plugin for that, e.g. `sos report -k containers_common.rootlessusers=awx`, or we could hook that up to a preset for AAP as well.
> 
> For reference, everything in sos is collected by root. All collections from this plugin would be as root, and plugins _generally_ don't have the ability to change their execution context. Even the above plugin uses `machinectl` to run in the context of the specified user(s), but it is still being executed as root.

@TurboTurtle what would be the best approach to hook it up to AAP to reduce the need for end-user interaction?

TurboTurtle:
> @TurboTurtle what would be the best approach to hook it up to AAP to reduce the need for end-user interaction?

Just the collection of the `containers_common` plugin for the `awx` user? A preset would be best. If you're looking for non-root globally, that's not something we support - non-root report generation was dropped with sos-4.0. `sos collect` (controlled parallel collections over multiple nodes at once) can execute as non-root to facilitate the collections, but it still needs to become root on the nodes it is generating reports on.

A preset would allow us to specify certain conditions that make us confident we are on an AAP node, and then tack on a pre-defined set of options for any execution of `sos report`. Usually this is by the presence of certain packages, so in this case if we highlight one or more AAP-specific packages we can create a preset that effective says "when these packages are present on the host, add `-k containers_common.rootlessusers=awx`  (and anything else you'd want the be standard use) to any options the user passes"

TurboTurtle (approved):

PR #3564, opened 2024-03-11
[convert2rhel] Update copy spec to include * report

r0x0d:
convert2rhel now generates a post-conversion report as well as the pre-conversion report. We want to be able to collect both reports during sosreport execution. To make sure we are not missing them (or any other future report we may introduce), the copy spec string was changed to use a `*` instead of the names of the reports.

---


jcastill (approved):

jcastill:
Is there anything inside /var/log/convert2rhel/ that we shouldn't collect? I'm just wondering if we could just reduce the capture to everything in that directory. 
Also, no passwords or secrets need to be scrubbed from pre/post conversion, right?

TurboTurtle (approved):

arif-ali (approved):

r0x0d:
> Is there anything inside /var/log/convert2rhel/ that we shouldn't collect? I'm just wondering if we could just reduce the capture to everything in that directory. Also, no passwords or secrets need to be scrubbed from pre/post conversion, right?

Hi, @jcastill! 

> Is there anything inside /var/log/convert2rhel/ that we shouldn't collect? 

I don't believe so... This is a directory managed by the convert2rhel tool, so it is possible that anything that will live inside there will be generated by us. 

> I'm just wondering if we could just reduce the capture to everything in that directory

We could, definitely, but for now, the items listed in that list is pretty much everything we generate. I don't believe we will generate more files to that directory in the near future. 

>  Also, no passwords or secrets need to be scrubbed from pre/post conversion, right?

That's right. We take care of that in the tool to not slip any of those secrets in the files.

PR #3565, opened 2024-03-11
[redhat] Enhance OCP preset

pmoravec:
Support engineers always call sos report with

-k crio.all=on -k crio.logs=on  -k podman.all=on -k podman.logs=on

options on OCP so let make it part of the preset by default.

Relevant: RHEL-28547
Closes: #3565

(it is even in OCP upstream doc: https://docs.openshift.com/container-platform/4.14/support/gathering-cluster-data.html)

---

### DIFF
--- a/sos/presets/redhat/__init__.py
+++ b/sos/presets/redhat/__init__.py
@@ -40,10 +40,14 @@ RHOCP_OPTS = SoSOptions(
     skip_plugins=['cgroups'], container_runtime='crio', no_report=True,
     log_size=100,
     plugopts=[
+        'crio.all=on',
+        'crio.logs=on',
         'crio.timeout=600',
         'networking.timeout=600',
         'networking.ethtool_namespaces=False',
-        'networking.namespaces=200'
+        'networking.namespaces=200',
+        'podman.all=on',
+        'podman.logs=on',
     ])
 
 RH_CFME = "cfme"


jcastill (approved):

TurboTurtle:
I, for years, had tried to educate against this. I simply do not understand the persistence with which this is pursued.

Among other things, this will result in duplicated log retrievals and the retrieval of long-dead container's stdout.

pmoravec:
> I, for years, had tried to educate against this. I simply do not understand the persistence with which this is pursued.
> 
> Among other things, this will result in duplicated log retrievals and the retrieval of long-dead container's stdout.

If that is the will of the OCP support community..?

If there would be some particular arguments against this change, I am ok to close it without merging. As an OCP noob, I can't evaluate the justification for that tuning (to be default).

jcastill:
@pmoravec maybe we can ask engineering about that. I see lots of documents and some KCS articles specifying these options, but they don't break down why they need each and everyone of them. If what @TurboTurtle mentions still applies, then engineering should change these docs?

pmoravec:
I will ask for some broader feedback among our OCP support engineers what is their opinion or preference. If there is some argument that @TurboTurtle would like to highlight (beyond the above written ones), please let me know within few days.

TurboTurtle:
> I will ask for some broader feedback among our OCP support engineers what is their opinion or preference. If there is some argument that @TurboTurtle would like to highlight (beyond the above written ones), please let me know within few days.

The original recommendation traces all the way back to the very first release of OCP 4, which used docker under the hood. There was a misunderstanding of what the `docker.all` plugin option did; that being that it caused commands like `docker ps` to include the `-a` option, which wasn't the case. I tried explaining this to several OCP engineers, but it didn't seem to land, the recommendation kept appearing in KCS written by sbr-shift, whereas the sbr-containers folks who _actually_ supported docker, yours included, very seldom ever used that option.

Fast forward a few versions of OCP and the advent of crio and podman, this recommendation is still in place because, and I quote from a particular shift engineer, "well we did it for docker so we should do it here too." I'm not one to ever endorse the "we've always done it this way" mentality, but beyond that the use is based in a fundamental misunderstanding of what the option does.

Going forward again through some months/years, I had some luck in explaining what this option actually does. In particular, during meetings about customers and some internal RH folks complaining about how big OCP sos report are and, to a lesser extent, how long they can take to run. While not solely responsible for the large sizes, this combination of options is _directly_ responsible for a good deal of that. `crio` will report on `podman` containers, which means from the start we're duplicating data collection which is something we have always tried to avoid. On top of that, this combination of options means we're collecting the stdout of _every_ container that has _ever_ run on the host and has yet to be garbage collected. On that note, there have been at times where the garbage collection wasn't completely cleaning up what it was intended to clean up, namely long-dead containers, which only exacerbated the issue.

Tracking this down, most OCP engineers agreed that they don't need every container's stdout, nor the stdout of every _dead_ version of every container that has since been relaunched. The exception to this was one particular engineer who insisted, without evidence of a single case where it helped, and went so far as to track the KCS that had this option recommendation and would revert any change made to it. This KCS was then what was referenced for the official docs, and the doc writer would have naturally assumed that the KCS was correct.

So there is the long and short of it. The use of these options is rooted in a fundamental misunderstanding of the  options, and a desire for behavior which is not evidenced to be helpful.

andreaskaris:
Correct me if I'm wrong, but without the crio.* options, the sos plugin does not gather the full /var/log/containers and /var/log/pods logs. I don't know what the criteria is, but the logs are pretty limited.  It also doesn't gather the logs of all running containers, either .. there's only a small subset of container logs that's gathered
I might be wrong, but IIRC without the above options, the sos plugin also does not gather some specific namespace related options.
So we aren't only talking about logs of already shut down containers, but also about data of running containers.

I am not the engineer who pushed for the inclusion of the above parameters, however I personally advocate for the inclusion of the parameters because:
* it's in our doc, and the KCS, and so that's what I personally request customers and partners to use (granted, we could also modify our doc and KCS and make it clear that the parameters are optional)
* whenever I do *not* request data with the aforementioned parameters, the sosreport to me feels (and is) incomplete. When I TS issues in OCP, I am interested in the container logs most of the time, because they allow me to correlate issues between containers and the system log. The journal just doesn't contain anything useful in CoreOS in a lot of cases because so much is actually going on inside the containers
* gathering logs from dead containers is actually a good thing in this context, because `oc adm must-gather` only gathers logs from live containers
* Data and bandwidth are cheap, human time is not. and sometimes you get a single shot at reproducing an issue. I'd prefer gathering 9 out of 10 times too much data, instead of having too few data in 1 out of 10 cases

TurboTurtle:
@andreaskaris  No, these options do _not_ affect anything to do with the collection of `/var/log/containers` or `/var/log/pods`. The `crio` plugin does not even perform those specific collections at all.

The `crio.all` option is used here - https://github.com/sosreport/sos/blob/main/sos/report/plugins/crio.py#L71 - and only there. It causes the command that generates the list of containers to collect `crictl inspect` and crictl logs` (if `crio.logs` is also used` to include dead containers. Not that it does not directly affect the copy of the output of `crictl ps` that gets stored in the sos report archive.

As for those two directories, the first is copied in whole by the `container_log` plugin, not `crio`, `podman`, or any other container-focused plugin. `/var/log/pods` is not explicitly captured by any plugin today that I see. If it is getting captured, please provide `sos_reports/manifest.json` from such an archive so we can see what is collecting it. It may be an out-of-tree plugin, for example.

That said, we can (separately to this PR) increase the default collection size for `container_log` (or perhaps we should rethink this plugin? A bunch of other plugins try to do specific collections under that directory) and add collection of `/var/log/pods` somewhere to an existing plugin.

andreaskaris:
I took a sosreport on one of my systems and compared it to the sosreport that I received from the customer. The customer sosreport indeed did not have crictl inspects of the containers, however, I do now believe this to be an issue with this specific sosreport, as the customer even ran it with `"cmdline": "report -k crio.all=on -k crio.logs=on -k podman.all=on -k podman.logs=on"`.  It's actually the exact opposite to what I suspected, because the crio plugin hit a timeout: `"timeout_hit": true,` and I now believe that this is the cause of the issue in that specific instance :-(  So it makes sense to close this PR and the Jira issue - I'm not on the same page with you, adding these commands (by default) will cause more issues than it will solve.

> As for those two directories, the first is copied in whole by the container_log plugin, not crio, podman, or any other container-focused plugin. /var/log/pods is not explicitly captured by any plugin today that I see. If it is getting captured, please provide sos_reports/manifest.json from such an archive so we can see what is collecting it. It may be an out-of-tree plugin, for example.

/var/log/containers contains symlinks to /var/log/pods in OCP:
```
[root@sno10 /]# find /var/log/containers/ -type f
[root@sno10 /]# find /var/log/containers/ -type l | wc -l
3929
[root@sno10 /]# find /var/log/pods -type l
[root@sno10 /]# find /var/log/pods -type f | wc -l
4120
[root@sno10 /]# ls -l /var/log/containers/ | tail -1
lrwxrwxrwx. 1 root root 134 Feb 15 17:05 version-4.14.7-5f8qw-jjvmz_openshift-cluster-version_rename-to-final-location-07bcf4f0a2891f35596fbf7038c90f0f8f9baf6f86d1ed3dc7cbc8d02e0f7adf.log -> /var/log/pods/openshift-cluster-version_version-4.14.7-5f8qw-jjvmz_8ee626ac-4eb5-4306-801b-473882b1bf4d/rename-to-final-location/0.log
```

For whatever reason, the `containers` plugin only gathers a subset of the container logs:
```
[root@sno10 /]# ls /host/var/log/containers/*.log | wc -l
3929
```

```
                "container_log": {
                    "start_time": "2024-03-19 11:27:53.300257",
                    "end_time": "2024-03-19 11:27:53.364087",
                    "run_time": "0:00:00.063830",
                    "setup_start": "2024-03-19 11:27:42.249017",
                    "setup_end": "2024-03-19 11:27:42.447763",
                    "setup_time": "0:00:00.198746",
                    "timeout": 300,
                    "timeout_hit": false,
                    "command_timeout": 300,
                    "commands": [],
                    "files": [
                        {
                            "specification": "/host/var/log/containers/*.log",
                            "files_copied": [
                                "host/var/log/containers/cluster-autoscaler-operator-865c5d468c-kpsmn_openshift-machine-api_cluster-autoscaler-operator-e96fc4b4b916ad8064d942ed772de10d73acf5909904cbe0d861d8fc81dc6362.log",
                                "host/var/log/containers/nmstate-handler-blcfm_openshift-nmstate_nmstate-handler-f5ebed34c9fad822b5cc31d60d7346636454bdd28acf3f00135aaceda2013883.log",
                                "host/var/log/containers/sriov-network-config-daemon-7trml_openshift-sriov-network-operator_sriov-network-config-daemon-08d5f94aa40ea9b86e1602e2cf87f1f276d47c91940fcfc64d66a4218881affb.log",
                                "host/var/log/containers/insights-operator-54876c4bff-92frt_openshift-insights_insights-operator-83cf0013695d5e409dd348c9c332a3f4ce6230a28923c0f95cfd14be4d626d4b.log",
                                "host/var/log/containers/kube-apiserver-sno10.workload.bos2.lab_openshift-kube-apiserver_kube-apiserver-24665343b6cd0d7110f579314ff0743fe0b0b447eb9772c345eb018165c31d36.log",
                                "host/var/log/containers/network-resources-injector-7lvb4_openshift-sriov-network-operator_webhook-server-c47990ccb5a48c6031198dfc8e3237c514de54ceb4535a0423f1f327f73b35ec.log",
                                "host/var/log/containers/ovnkube-node-gx9cx_openshift-ovn-kubernetes_ovn-acl-logging-502b8e6a6571285e240c4d49367dadce22545f25fc4fa98613027ba409fa49b9.log"
                            ],
                            "tags": []
                        }
                    ],
                    "strings": {
                        "host_var_log_containers_ovnkube-node-gx9cx_openshift-ovn-kubernetes_ovn-acl-logging-502b8e6a6571285e240c4d49367dadce22545f25fc4fa98613027ba409fa49b9_log_tailed": {
                            "path": "sos_strings/container_log/host.var.log.containers.ovnkube-node-gx9cx_openshift-ovn-kubernetes_ovn-acl-logging-502b8e6a6571285e240c4d49367dadce22545f25fc4fa98613027ba409fa49b9.log.tailed",
                            "tags": []
                        }
                    },
                    "containers": {},
                    "collections": []
                },
```

```

**Comment by andreaskaris** [comment]
o.k., so the "missing" `/var/log/container` logs are a result of the size limit being reached:
```
[root@sno10 sosreport-sno10-2024-03-19-lznjkiu]# less sos_logs/sos.log  | grep 'over size limit' | head
2024-03-19 16:50:23,209 INFO: [plugin:container_log] File '/host/var/log/containers/ovnkube-node-gx9cx_openshift-ovn-kubernetes_ovnkube-controller-dfc52c16431973fa7a4ec6883689d2cab20d182a95d2a9007c7ce89ad2cee221.log' is over size limit, will instead tail the file during collection phase.
2024-03-19 16:50:23,209 INFO: [plugin:container_log] skipping '/host/var/log/containers/ovnkube-control-plane-667fd4b5d8-zl2h8_openshift-ovn-kubernetes_ovnkube-cluster-manager-085b31c1fc569f6a11d000735ac79717ba8312949aeb2331dc9f1953af485d22.log' over size limit
2024-03-19 16:50:23,209 INFO: [plugin:container_log] skipping '/host/var/log/containers/sriov-network-config-daemon-7trml_openshift-sriov-network-operator_sriov-network-config-daemon-08d5f94aa40ea9b86e1602e2cf87f1f276d47c91940fcfc64d66a4218881affb.log' over size limit
2024-03-19 16:50:23,209 INFO: [plugin:container_log] skipping '/host/var/log/containers/speaker-csgc7_metallb-system_frr-7b3246c8ea9c8f6b1b2c3ff05cec0b668486114f569d977f54faf32958b681e8.log' over size limit
2024-03-19 16:50:23,209 INFO: [plugin:container_log] skipping '/host/var/log/containers/insights-operator-54876c4bff-92frt_openshift-insights_insights-operator-83cf0013695d5e409dd348c9c332a3f4ce6230a28923c0f95cfd14be4d626d4b.log' over size limit
2024-03-19 16:50:23,209 INFO: [plugin:container_log] skipping '/host/var/log/containers/cluster-autoscaler-operator-865c5d468c-kpsmn_openshift-machine-api_cluster-autoscaler-operator-e96fc4b4b916ad8064d942ed772de10d73acf5909904cbe0d861d8fc81dc6362.log' over size limit
2024-03-19 16:50:23,209 INFO: [plugin:container_log] skipping '/host/var/log/containers/network-resources-injector-7lvb4_openshift-sriov-network-operator_webhook-server-c47990ccb5a48c6031198dfc8e3237c514de54ceb4535a0423f1f327f73b35ec.log' over size limit
2024-03-19 16:50:23,209 INFO: [plugin:container_log] skipping '/host/var/log/containers/ovnkube-node-gx9cx_openshift-ovn-kubernetes_ovn-acl-logging-502b8e6a6571285e240c4d49367dadce22545f25fc4fa98613027ba409fa49b9.log' over size limit
2024-03-19 16:50:23,209 INFO: [plugin:container_log] skipping '/host/var/log/containers/kube-apiserver-sno10.workload.bos2.lab_openshift-kube-apiserver_kube-apiserver-24665343b6cd0d7110f579314ff0743fe0b0b447eb9772c345eb018165c31d36.log' over size limit
2024-03-19 16:50:23,209 INFO: [plugin:container_log] skipping '/host/var/log/containers/ovnkube-node-gx9cx_openshift-ovn-kubernetes_ovn-controller-e8f5b0af0cf33be6b78c0fab79edc4f283d1a44cbeb724021f0d28a336761606.log' over size limit
```

Due to the default of 100 MB:
```
less sos_logs/sos.log
(...)
2024-03-19 16:50:21,633 INFO: [sos.report:setup] effective options now: --container-runtime crio --log-size 100  (...)
```

That makes for smaller sosreports, but from my tests, for pretty undeterministic collection of container logs ... different runs of `sos report` yielded different containers logs.

The total log size can be influenced with `sos report --log-size 1000` (in this case for example to 1 TB):
```
less sos_logs/sos.log
(...)
2024-03-19 17:04:02,783 INFO: [sos.report:setup] effective options now: --container-runtime crio --log-size 1000  (...)
```

TurboTurtle:
We can definitely raise the log size used by `container_logs`, and if you use `--all-logs` that will cause the plugin to collect _all_ container logs regardless of size, but does have the knock-on effect of removing size limiting for all plugins.

One note, the default `log-size` for sos is 25MB, not 100 - so I'm not sure where you're seeing that get set from off the top of my head (aside from a manual setting in `/etc/sos/sos.conf`).

The genesis of `container_logs` was, I believe, OpenStack not OpenShift - and those infra containers typically are much smaller in number and produce less output than OCP containers, which is probably why we haven't yet had a complaint about lack of logs for the plugin; again though we can certainly raise the plugin-specific size limiting. The collection of log files is based on modification time - the newest files/logs are collected first, then the older ones, up until the log size limit gets hit as you've seen. In practice, this means that you'd see collections for the containers that most recently produced log output.

TurboTurtle:
Oh, right, I forgot we did that in the preset. Yeah, we typically don't like to blindly set `--all-logs`, which is why we would've made that change.

As for the timeout options...can you provide the beginning of the problematic `sos.log`, the first full 10 lines or so? That should capture the command as executed, the OCP loading, and the effective options we merged into. We might need to see a run with `-v` but hopefully something pops up.

On that note actually, what version of sos is your customer running? We've had some bugs previously around plugin timeout handling, so I wonder if they are still running one of those older versions?

pmoravec:
Closing asnot needed. We will followup with some ideas how to improve collection of container logs in (much) less intrusive and efficient way than the proposed options.

PR #3572, opened 2024-03-16
[global] Remove empty except warnings highlighted by CodeQL

TurboTurtle:
This commit is a first pass at addressing a number of the `Empty Except` warnings flagged by CodeQL. These are for try/except blocks whose except statement simply contains a `pass`, and to address these we add a log message in almost all cases. Where we don't log, there is an added comment for context as to why we're passing without further action.

---

### DIFF
--- a/sos/report/__init__.py
+++ b/sos/report/__init__.py
@@ -922,6 +922,7 @@ class SoSReport(SoSComponent):
                         try:
                             val = int(val)
                         except ValueError:
+                            # not a number to convert back to int from argparse
                             pass
 
                 try:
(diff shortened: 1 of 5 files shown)


TurboTurtle:
There will still be a few empty except alerts from CodeQL present after this, mostly in plugins, so this is just the first pass at them since I had a little time to kill this weekend.

sos/report/__init__.py
```diff
@@ -922,6 +922,7 @@ def _set_tunables(self):
                         try:
                             val = int(val)
                         except ValueError:
+                            # not a number to convert back to int from argparse
```
pmoravec, line 925: Maybe we can print here some log as well, for potential diagnosis "why my argument has been ignored"?

pmoravec (approved):
ACK, the one "maybe" comment is just a suggestion, approving the PR regardless of that.

arif-ali (approved):

sos/report/__init__.py
TurboTurtle: I don't _think_ that's something to be concerned about here, but would already be logged later on by existing checks. This is the attempt to see "did the user pass something that got auto converted to a string, but we should cast to an int? If not, no actual concern because it should really be a string".

PR #3573, opened 2024-03-18
[manpages] Add note about deprecation of options for RHEL

jcastill:
Add a note to the entries for --upload-user and --upload-pass clarifying that these options have been deprecated in RHEL systems.

---

### DIFF
--- a/man/en/sos-report.1
+++ b/man/en/sos-report.1
@@ -430,6 +430,7 @@ variable. If this variable is set, then no username prompt will occur and --batc
 may be used provided all other required values (case number, upload password)
 are provided.
 
+This option is ignored when uploading to the Red Customer Portal or Red Hat Secure FTP server in favour of web token authentication.
 .TP
 .B \-\-upload-pass PASS
 Specify the password to use for authentication with the destination server.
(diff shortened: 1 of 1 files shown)


arif-ali (approved):

man/en/sos-report.1
```diff
@@ -430,6 +430,7 @@ variable. If this variable is set, then no username prompt will occur and --batc
 may be used provided all other required values (case number, upload password)
 are provided.
 
+This option has been deprecated in RHEL systems in favour of web token authentication.
```
TurboTurtle, line 433: "This option has been deprecated when uploading to the Red Hat Customer Portal in favour of web token authentication."

The option works on RHEL, and the upload target isn't always going to be the RHCP, that's just the default.

pmoravec: It was deprecated also when uploading to Red Hat SFTP server, where either user must use web token authentication, or semi-anonymous random user. Just a nitpick to the right phrase / comment here)

jcastill: Yes, it's basically everywhere now except for anonymous ftp. I can make it more general perhaps, mentioning SFTP and customer portal. Something like:

"This option has been deprecated when uploading to the Red Hat Customer Portal or Red Hat Secure FTP in favour of web token authentication."

Does that look better?

pmoravec: Sounds good to me.

TurboTurtle: Thinking on it more, 'deprecated' may not actually be applicable. We're not getting rid of the option, and these targets simply don't use them anymore.

"This option is ignored when uploading to the Red Customer Portal or Red Hat Secure FTP server in favour of web token authentication", perhaps?

jcastill: Yes, that sounds much better. I'll push the change in a bit.

TurboTurtle (approved):

pmoravec (approved):

arif-ali (approved):

PR #3577, opened 2024-03-21
[kubernetes] k8s improvements for Charmed Kubernetes

shundezhang:
Add more resources to output when kubernetes.all option is true. Output json format if kubernetes.all option is true. Export pod logs to an individual file for each container, so the plugin can export logs if there are multiple containers in a pod.

---

### DIFF
--- a/sos/report/plugins/kubernetes.py
+++ b/sos/report/plugins/kubernetes.py
@@ -13,6 +13,7 @@ from fnmatch import translate
 import re
 from sos.report.plugins import (Plugin, RedHatPlugin, DebianPlugin,
                                 UbuntuPlugin, PluginOpt)
+import json
 
 
 class Kubernetes(Plugin):
@@ -21,6 +22,39 @@ class Kubernetes(Plugin):
 
     plugin_name = "kubernetes"
     profiles = ('container',)
+    plugin_timeout = 1200
+
+    config_files = [
+        "/etc/kubernetes",
+        "/run/flannel",
+    ]
+    resources = [
+        'events',
+        'deployments',
+        'ingresses',
+        'pods',
+        'pvc',
+        'services',
+        'daemonsets',
+        'replicasets',
+        'endpoints',
+        'statefulsets',
+        'configmaps',
+        'serviceaccounts',
+        'secrets',
+        'jobs',
+        'cronjobs',
+        'clusterroles',
+        'clusterrolebindings'
+    ]
+
+    # these are not namespaced, must pull separately.
+    global_resources = [
+        'sc',
+        'pv',
+        'roles',
+        'rolebindings'
+    ]
 
     option_list = [
         PluginOpt('all', default=False,
@@ -34,26 +68,13 @@ class Kubernetes(Plugin):
     ]
 
     kube_cmd = "kubectl"
-    resources = [
-        'deployments',
-        'ingresses',
-        'limitranges',
-        'pods',
-        'policies',
-        'pvc',
-        'rc',
-        'resourcequotas',
-        'routes',
-        'services'
-    ]
 
     def check_is_master(self):
         """ Check if this is the master node """
         return any(self.path_exists(f) for f in self.files)
 
     def setup(self):
-        self.add_copy_spec("/etc/kubernetes")
-        self.add_copy_spec("/run/flannel")
+        self.add_copy_spec(self.config_files)
 
         self.add_env_var([
             'KUBECONFIG',
@@ -83,21 +104,26 @@ class Kubernetes(Plugin):
             return
 
         for subcmd in ['version', 'config view']:
-            self.add_cmd_output('%s %s' % (self.kube_cmd, subcmd))
+            self.add_cmd_output(
+                f'{self.kube_cmd} {subcmd}',
+                subdir='cluster-info'
+            )
 
-        # these are not namespaced, must pull separately.
-        global_resources = [
-            'namespaces',
-            'projects',
-            'pvs'
-        ]
-        self.add_cmd_output([
-            "%s get %s" % (self.kube_cmd, res) for res in global_resources
-        ])
+        if self.get_option('all'):
+            self.add_cmd_output([
+                f"{self.kube_cmd} get -o json {res}"
+                for res in self.global_resources
+            ], subdir='cluster-info')
+        else:
+            self.add_cmd_output([
+                f"{self.kube_cmd} get {res}"
+                for res in self.global_resources
+            ], subdir='cluster-info')
 
         # Get detailed node information
-        nodes = self.collect_cmd_output("%s get nodes" % self.kube_cmd)
-        if nodes['status'] == 0:
+        nodes = self.collect_cmd_output(f"{self.kube_cmd} get nodes",
+                                        subdir='cluster-info')
+        if nodes['status'] == 0 and self.get_option('describe'):
             for line in nodes['output'].splitlines()[1:]:
                 # find first word in the line and ignore empty+blank lines
                 words = line.split()
@@ -105,12 +131,19 @@ class Kubernetes(Plugin):
                     continue
                 node = words[0]
                 self.add_cmd_output(
-                    "%s describe node %s" % (self.kube_cmd, node),
-                    subdir='nodes'
+                    f"{self.kube_cmd} describe node {node}",
+                    subdir='cluster-info'
                 )
 
+        self.add_cmd_output([
+            f"{self.kube_cmd} get -o json nodes",
+        ], subdir='cluster-info')
+
         # Also collect master metrics
-        self.add_cmd_output("%s get --raw /metrics" % self.kube_cmd)
+        self.add_cmd_output(
+            f"{self.kube_cmd} get --raw /metrics",
+            subdir='cluster-info'
+        )
 
         # CNV is not part of the base installation, but can be added
         if self.is_installed('kubevirt-virtctl'):
@@ -123,7 +156,8 @@ class Kubernetes(Plugin):
     def collect_per_resource_details(self):
         """ Collect details about each resource in all namespaces """
         # get all namespaces in use
-        kns = self.collect_cmd_output('%s get namespaces' % self.kube_cmd)
+        kns = self.collect_cmd_output(f'{self.kube_cmd} get namespaces',
+                                      subdir='cluster-info')
         # namespace is the 1st word on line, until the line has spaces only
         kn_output = kns['output'].splitlines()[1:]
         knsps = [n.split()[0] for n in kn_output if n and len(n.split())]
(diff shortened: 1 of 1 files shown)


sos/report/plugins/kubernetes.py
```diff
@@ -85,15 +96,16 @@ def setup(self):
...
-        ])
+        if self.get_option('all'):
+            self.add_cmd_output([
+                "%s get -o json %s" % (self.kube_cmd, res)
+                for res in self.global_resources
+            ])
+        else:
+            self.add_cmd_output([
+                "%s get %s" % (self.kube_cmd, res)
+                for res in self.global_resources
```
TurboTurtle, line 107: f-strings instead of string substitution, please.

sos/report/plugins/kubernetes.py
```diff
@@ -154,19 +168,28 @@ def collect_per_resource_details(self):
...
+                                self.add_cmd_output('%s %s logs %s -c %s' % (
+                                    self.kube_cmd, knsp,
+                                    pod["metadata"]["name"],
+                                    cont["name"]), subdir='pods')
+                        if "initContainers" in pod["spec"]:
+                            for cont in pod["spec"]["initContainers"]:
+                                self.add_cmd_output('%s %s logs %s -c %s' % (
+                                    self.kube_cmd, knsp,
+                                    pod["metadata"]["name"],
+                                    cont["name"]), subdir='pods')
```
TurboTurtle, line 192: Same, f-strings please.

sos/report/plugins/kubernetes.py
```diff
@@ -209,12 +232,21 @@ def setup(self):
...
+        self.global_resources.extend([
+            'projects',
+            'pvs'
+        ])
         super().setup()
 
 
 class UbuntuKubernetes(Kubernetes, UbuntuPlugin, DebianPlugin):
 
-    packages = ('kubernetes',)
```
TurboTurtle, line 217: I think we need to keep this, don't we?

sos/report/plugins/kubernetes.py
```diff
@@ -238,6 +270,26 @@ def setup(self):
...
+                'endpoints',
+                'statefulsets',
+                'configmaps',
+                'serviceaccounts',
+                'secrets',
+                'jobs',
+                'cronjobs',
+                'clusterroles',
+                'clusterrolebindings'
+            ])
```
TurboTurtle, line 292: Are these actually specific to the Ubuntu packaging for kubernetes? If not, they should be in the base class. I have no real opinion on gating them behind the `all` option, but am curious if these aren't usually desired for troubleshooting?

TurboTurtle (commented):
Please use f-strings where you are doing string interpolation, highlighted in a few areas but in general if you are making changes or adding net-new collections, then we want f-strings to be used going forward.

Other comments in-line.

sos/report/plugins/kubernetes.py
shundezhang: this is changed

sos/report/plugins/kubernetes.py
shundezhang: this is changed.

sos/report/plugins/kubernetes.py
shundezhang: There is no kubernetes package in Canonical k8s, which is deployed with snap.
I've put it back anyway.
Let me know if I need to remove it.

sos/report/plugins/kubernetes.py
shundezhang: I've moved them to the base class.
They are less common/important and can save time to dump less data.
We can move them around based on feedback.

sos/report/plugins/kubernetes.py
dnegreira: We also discover if it is installed as a snap with this method, so this should still be here.

sos/report/plugins/kubernetes.py
```diff
@@ -34,26 +56,32 @@ class Kubernetes(Plugin):
...
         return any(self.path_exists(f) for f in self.files)
 
     def setup(self):
-        self.add_copy_spec("/etc/kubernetes")
-        self.add_copy_spec("/run/flannel")
+        self.plugin_timeout = 1200
+        for config_file in self.config_files:
+            self.add_copy_spec(config_file)
+
+        if self.get_option('all'):
```
dnegreira, line 69: I would not put this under the all, lets gather all the info by default as this is quite useful for us to have.

sos/report/plugins/kubernetes.py
```diff
@@ -105,12 +134,16 @@ def setup(self):
                     continue
                 node = words[0]
                 self.add_cmd_output(
-                    "%s describe node %s" % (self.kube_cmd, node),
+                    f"{self.kube_cmd} describe node {node}",
                     subdir='nodes'
                 )
+        if self.get_option('all'):
```
dnegreira, line 140: same here, I would not gate this behind all.

sos/report/plugins/kubernetes.py
```diff
@@ -34,26 +56,32 @@ class Kubernetes(Plugin):
...
     def check_is_master(self):
         """ Check if this is the master node """
         return any(self.path_exists(f) for f in self.files)
 
     def setup(self):
-        self.add_copy_spec("/etc/kubernetes")
-        self.add_copy_spec("/run/flannel")
+        self.plugin_timeout = 1200
+        for config_file in self.config_files:
+            self.add_copy_spec(config_file)
```
arif-ali, line 67: you could use the `config_files` list here directly with `add_copy_spec`, no need for the for loop imo

arif-ali (changes requested):
and as per Davids comments too

sos/report/plugins/kubernetes.py
```diff
@@ -34,26 +67,14 @@ class Kubernetes(Plugin):
...
-    ]
 
     def check_is_master(self):
         """ Check if this is the master node """
         return any(self.path_exists(f) for f in self.files)
 
     def setup(self):
-        self.add_copy_spec("/etc/kubernetes")
-        self.add_copy_spec("/run/flannel")
+        self.plugin_timeout = 1200
```
TurboTurtle, line 76: `plugin_timeout` needs to be in the class definition, alongside `packages` or `services`, etc...

sos/report/plugins/kubernetes.py
```diff
@@ -123,57 +149,71 @@ def setup(self):
...
                     reg = (translate(self.get_option('podlogs-filter')) if
                            self.get_option('podlogs-filter') else None)
-                    for pod in pods:
-                        if reg and not re.match(reg, pod):
+                    for pod in pods["items"]:
+                        if reg and not re.match(reg, pod["metadata"]["name"]):
                             continue
-                        self.add_cmd_output('%s logs %s' % (k_cmd, pod),
-                                            subdir='pods')
+                        _subdir = f'pods/{pod["metadata"]["namespace"]}'
```
TurboTurtle, line 191: Just a thought here, should we make the subdir path `namespaces/{namespace}/pods`? So that other namespace-specific items can be grouped under the namespace dir and then have another subdir for the resource type?

We'd want to also update the `all` collection for resources to this new subdir layout here, if we do this.

TurboTurtle:
If we could get a before and after example sos report, since this change is on the larger size, that'd be great! Just want to make sure we're not inadvertently regressing somewhere since we don't (yet) have CI tests for kubernetes.

sos/report/plugins/kubernetes.py
shundezhang: I added the namespace subdir for pod logs since there will be too many files in pods/ if the k8s cluster has a lot of pods. It would be good to have subdirs so they can be separated.
At the moment the structure is grouped by resource type.
We can certainly change to grouped by namespace, just like what `kubectl cluster-info dump` do.
I can make the change if another developer gives another +1.

sos/report/plugins/kubernetes.py
shundezhang: Yup. This is changed.

sos/report/plugins/kubernetes.py
shundezhang: This is changed too.

sos/report/plugins/kubernetes.py
shundezhang: Yes I've changed the code accordingly.

shundezhang:
> If we could get a before and after example sos report, since this change is on the larger size, that'd be great! Just want to make sure we're not inadvertently regressing somewhere since we don't (yet) have CI tests for kubernetes.

This is an example of what the new code outputs now: https://pastebin.ubuntu.com/p/hPBmwSbxKh/
There is not much difference with the existing code except it has more resources: https://pastebin.ubuntu.com/p/dCBVq8FRnf/

sos/report/plugins/kubernetes.py
arif-ali: @shundezhang i think it makes sense do do what @TurboTurtle suggested

shundezhang: I have made the change to dir structure.
Now everything that is queried from k8s APIs, e.g. using kubectl command, is placed under cluster-info/
At the same level there are outputs from systemctl/journalctl commands, and maybe other system commands too.
Under cluster-info/ subdir, there are outputs for global resource and subdirs for each namespace.
Inside each namespace dir, a dir for each resource type holds outputs for kubectl describe outputs, and also there is a dir for pod logs. Json output files are placed under this namespace dir since there is only one file for each resource.
Examples:
1. kubernetes.all=on kubernetes.podlogs=on kubernetes.describe=on
https://paste.ubuntu.com/p/DCrSK92TYF/
2. kubernetes.all=off kubernetes.podlogs=off kubernetes.describe=off
https://paste.ubuntu.com/p/24fzFQ3kXQ/
3. kubernetes.describe=on kubernetes.podlogs=on
https://paste.ubuntu.com/p/mjS66wqT2W/
4. kubernetes.all=on kubernetes.podlogs=on
https://paste.ubuntu.com/p/C95sBQFjYJ/

TurboTurtle: We should place the namespaces under a dir that makes it obvious what the subdirs are. Given that there is other content under `cluster-info` such as the global resources it's not immediately obvious what, for example, the `default` directory actually _is_.

The OpenShift plugin puts these under `namespaces`, e.g. the dir layout would be `sos_commands/openshift/namespaces/$namespace` - I think we should mimic that here. The global resources being under `cluster-info` is fine though, and it makes more sense than putting that stuff in the top-level plugin directory.

shundezhang: Hi @TurboTurtle thank you for the feedback. The structure here is similar to the output of command `kubectl cluster-info dump --output-directory='./cluster-info' -A`. An example can be found here: https://paste.ubuntu.com/p/xXHRc9zfwP/
It doesn't have too many layers of dirs.
This structure should be easy to understand for a k8s user as it is the default output of kubectl command.
To make it more obvious, I can add `namespace_` to the beginning of those subdirs so they look like:
```
namespace_default\...
namespace_kube-system\...
```
What do you think?
Also could you please share a full output from openshift plugin to compare?

TurboTurtle: Oh, ok. I didn't realize that was how it was dropped by default. In that case, no issue from me with the current state of how the output gets written to the archive.

TurboTurtle (approved):

arif-ali (approved):

PR #3580, opened 2024-03-21
[kernel] Copy the kernel config

aadhar-agarwal:
Copy the kernel config - `/boot/config-{kernel-version}`

---

### DIFF
--- a/sos/report/plugins/kernel.py
+++ b/sos/report/plugins/kernel.py
@@ -111,7 +111,7 @@ class Kernel(Plugin, IndependentPlugin):
             "/proc/buddyinfo",
             "/proc/slabinfo",
             "/proc/zoneinfo",
-            "/lib/modules/%s/modules.dep" % self.policy.kernel_version(),
+            f"/lib/modules/{self.policy.kernel_version()}/modules.dep",
             "/etc/conf.modules",
             "/etc/modules.conf",
             "/etc/modprobe.conf",
@@ -136,7 +136,8 @@ class Kernel(Plugin, IndependentPlugin):
             "/sys/kernel/debug/extfrag/extfrag_index",
             clocksource_path + "available_clocksource",
             clocksource_path + "current_clocksource",
-            "/proc/pressure/"
+            "/proc/pressure/",
+            f"/boot/config-{self.policy.kernel_version()}"
         ])
 
         if self.get_option("with-timer"):


sos/report/plugins/kernel.py
```diff
@@ -147,4 +148,12 @@ def setup(self):
         if not self.get_option("trace"):
             self.add_forbidden_path("/sys/kernel/debug/tracing/trace")
 
+
+class AzureKernel(Kernel, AzurePlugin):
```
TurboTurtle, line 152: The way the tagging resolution works out, we'd have to change `IndependentPlugin` to the specific `RedHatPlugin`, `UbuntuPlugin`, etc... tags if we break out one for just Azure Linux here.

That said......how have we not been collecting this generally for so long? @pmoravec @arif-ali should this become a general collection? I'm surprised that we don't do this today already, but I have to assume that no one has been asking for it until now...somehow?

I think we'd probably want this to go into the regular collection, and keep `Kernel` as a strictly `IndependentPlugin`.

aadhar-agarwal: Ah I see, the `IndependentPlugin` are for plugins that are distribution agnostic?

Happy to move this to the regular collection (the kernel setup function) above and keep `Kernel` as a strictly `IndependentPlugin`

arif-ali: Agreed, that could be useful for Ubuntu Plugin as well, so agree to add this above in the main class

aadhar-agarwal: @TurboTurtle @arif-ali, I added `/boot/config-{kernel-version}` to the main class above and removed the AzureKernel specific class

sos/report/plugins/kernel.py
```diff
@@ -136,7 +136,8 @@ def setup(self):
             "/sys/kernel/debug/extfrag/extfrag_index",
             clocksource_path + "available_clocksource",
             clocksource_path + "current_clocksource",
-            "/proc/pressure/"
+            "/proc/pressure/",
+            "/boot/config-%s" % self.policy.kernel_version()
```
TurboTurtle, line 140: Please use an f-string here as we are converting the project to use those globally.

aadhar-agarwal: For sure! Can I also change [line 114](https://github.com/sosreport/sos/pull/3580/files#diff-622aa52be019af33ea587ea315ab1be1333530c98581290b57c6e44e493bae2cR114) to use an f string as well?

arif-ali: > For sure! Can I also change [line 114](https://github.com/sosreport/sos/pull/3580/files#diff-622aa52be019af33ea587ea315ab1be1333530c98581290b57c6e44e493bae2cR114) to use an f string as well?

you sure can :)

aadhar-agarwal: Updated to use f-strings! :)

TurboTurtle (approved):

pmoravec (approved):

arif-ali (approved):

Issue #3297, opened 2023-07-03
Add Openstack Sunbeam plugin


dosaboy:
If sosreport is run on a host running a [sunbeam](https://governance.openstack.org/tc/reference/projects/sunbeam.html) snap it ccould collect information about a sunbeam deployment both from the snap itself and the kubernetes cluster sunbeam components are deployed on.

TODO: add more information about what and how we will collect data.

PR #3586, opened 2024-04-04
[sunbeam] Add new plugin for sunbeam

arif-ali:
Refactor ovn_host and openswitch plugins to utilise the commands from
sunbeam

Closes: #3297
Resolves: #3586
Related: SET-87

---

### DIFF
--- a/sos/report/plugins/ovn_host.py
+++ b/sos/report/plugins/ovn_host.py
@@ -23,6 +23,7 @@ class OVNHost(Plugin):
         '/usr/local/var/run/openvswitch',
         '/run/openvswitch',
     ]
+    ovs_cmd_pre = ""
 
     def setup(self):
         if os.environ.get('OVS_RUNDIR'):
@@ -39,9 +40,9 @@ class OVNHost(Plugin):
         self.add_copy_spec('/etc/sysconfig/ovn-controller')
 
         self.add_cmd_output([
-            'ovs-ofctl -O OpenFlow13 dump-flows br-int',
-            'ovs-vsctl list-br',
-            'ovs-vsctl list Open_vSwitch',
+            f'{self.ovs_cmd_pre}ovs-ofctl -O OpenFlow13 dump-flows br-int',
+            f'{self.ovs_cmd_pre}ovs-vsctl list-br',
+            f'{self.ovs_cmd_pre}ovs-vsctl list Open_vSwitch',
         ])
 
         self.add_journal(units="ovn-controller")
@@ -75,3 +76,29 @@ class RedHatOVNHost(OVNHost, RedHatPlugin):
 class DebianOVNHost(OVNHost, DebianPlugin, UbuntuPlugin):
 
     packages = ('ovn-host', )
+
+    sunbeam_common_dir = '/var/snap/openstack-hypervisor/common'
+
+    pid_paths = [
+        f'{sunbeam_common_dir}/run/ovn',
+    ]
+
+    def setup(self):
+
+        if self.is_installed('openstack-hypervisor'):
+            self.ovs_cmd_pre = "openstack-hypervisor."
+
+            self.add_copy_spec([
+                f'{self.sunbeam_common_dir}/lib/ovn-metadata-proxy/*.conf',
+            ])
+
+            if self.get_option("all_logs"):
+                self.add_copy_spec([
+                    f"{self.sunbeam_common_dir}/var/log/ovn/",
+                ])
+            else:
+                self.add_copy_spec([
+                    f"{self.sunbeam_common_dir}/var/log/ovn/*.log",
+                ])
+
+        super().setup()
--- a/sos/report/plugins/sunbeam_hypervisor.py
+++ b/sos/report/plugins/sunbeam_hypervisor.py
@@ -1,0 +1,80 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, UbuntuPlugin
+
+
+class SunbeamHypervisor(Plugin, UbuntuPlugin):
+
+    short_desc = "Sunbeam Hypervisor"
+
+    plugin_name = "sunbeam_hypervisor"
+    profiles = ('cloud',)
+    packages = ('openstack-hypervisor',)
+
+    common_dir = '/var/snap/openstack-hypervisor/common'
+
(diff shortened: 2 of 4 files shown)


arif-ali:
Created this PR to get more collaboration, so hence in draft. Will sort out any CI/linting over time as we add/remove things from these 2 plugins

pmoravec:
I guess no secrets can appear in the config files being collected?

No issue/point otherwise.

arif-ali:
> I guess no secrets can appear in the config files being collected?
> 
> No issue/point otherwise.

There will be, like I mentioned in my prior comment (and hence draft), this is a start on this and there may be further things to be added. the `postproc()` bit is going to be key here.

We will also need to somehow handle duplicate related items as well, like openvswitch, ovn and libvirt, as these would not be packged in the same way for those plugins to be triggered

sos/report/plugins/sunbeam-hypervisor.py
```diff
@@ -0,0 +1,51 @@
...
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, UbuntuPlugin
+
+
+class SunbeamHypervisor(Plugin, UbuntuPlugin):
+
+    short_desc = "Sunbeam Hypervisor snap"
+
+    plugin_name = "sunbeam-hypervisor"
```
TurboTurtle, line 16: Plugin names use `_` not dashes by convention.

TurboTurtle:
> We will also need to somehow handle duplicate related items as well, like openvswitch, ovn and libvirt, as these would not be packged in the same way for those plugins to be triggered

Such as what items? If they are packaged differently so that they don't enable the other plugins, the collections here wouldn't really be duplicates. Otherwise we can add logic to those plugins to enable based on snap installation.

arif-ali:
> > We will also need to somehow handle duplicate related items as well, like openvswitch, ovn and libvirt, as these would not be packged in the same way for those plugins to be triggered
> 
> Such as what items? If they are packaged differently so that they don't enable the other plugins, the collections here wouldn't really be duplicates. Otherwise we can add logic to those plugins to enable based on snap installation.

I have put my ideas into the latest commit (I know it has flake8 issues, I'll sort them). I think something like this should be feasible, let me know your thoughts

sos/report/plugins/sunbeam_hypervisor.py
```diff
@@ -0,0 +1,114 @@
...
+
+    plugin_name = "sunbeam_hypervisor"
+    profiles = ('cloud',)
+    packages = ('openstack-hypervisor',)
+
+    common_dir = '/var/snap/openstack-hypervisor/common'
+
+    def setup(self):
+
+        self.add_service_status([
```
dnegreira, line 24: Can we have 'snap.openstack-hypervisor.*' instead? Things might change name/get added/removed, would be nice to capture all of it if possible.

arif-ali: yup, makes sense, I'll amend that

sos/report/plugins/sunbeam_hypervisor.py
```diff
@@ -0,0 +1,114 @@
...
+        match_exp = r"(\s*passwd=\s*')([^']*)('.*)"
+        libvirt_path_exps = [
+            fr"{self.common_dir}/etc/libvirt/qemu/.*\.xml",
+            fr"{self.common_dir}/etc/libvirt/.*\.conf"
+        ]
+        for path_exp in libvirt_path_exps:
+            self.do_path_regex_sub(path_exp, match_exp, r"\1******\3")
+
+        # nova/neutron bits
+        protect_keys = [
```
TurboTurtle, line 66: I think we could collapse this a bit, and get future coverage "for free", by doing something like:

```python
protect_keys = [
    '.*_key',
    '.*_pass(word)?'
]
```

Or otherwise build these keys into the regex.

TurboTurtle:
> I have put my ideas into the latest commit (I know it has flake8 issues, I'll sort them). I think something like this should be feasible, let me know your thoughts

I'm onboard with it. If this becomes a recurring pattern, we might want to look at building out some `Plugin` method(s) that automatically handle situations like this, but as is this is fine by me.

sos/report/plugins/ovn_host.py
```diff
@@ -75,3 +76,20 @@ def setup(self):
...
+    pid_paths = [
+        '/var/snap/openstack-hypervisor/common/run/ovn',
+    ]
+
+    def setup(self):
+
+        if self.is_installed('openstack-hypervisor'):
+            self.ovs_cmd_pre = "openstack-hypervisor."
+
+            common_path = "/var/snap/openstack-hypervisor/common"
```
pmoravec, line 88: Why not directly `ovn_log_path = "/var/snap/openstack-hypervisor/common/var/log/ovn"`? Or you will use `common_path` else-where, right?

pmoravec:
> > I have put my ideas into the latest commit (I know it has flake8 issues, I'll sort them). I think something like this should be feasible, let me know your thoughts
> 
> I'm onboard with it. If this becomes a recurring pattern, we might want to look at building out some `Plugin` method(s) that automatically handle situations like this, but as is this is fine by me.

You mean `Plugin` class automatically resolve situations when multiple plugins collect the same file (optionally with different sizelimit, optionally truncated)? I am in! I think there are just a few scenarios we would need to handle and having that covered would prevent a (small but annoying) burden.

TurboTurtle:
> You mean `Plugin` class automatically resolve situations when multiple plugins collect the same file (optionally with different sizelimit, optionally truncated)? I am in! I think there are just a few scenarios we would need to handle and having that covered would prevent a (small but annoying) burden.

I was referring to @arif-ali's design in this plugin for conditionally setting the various commands being run based on snap installation.

sos/report/plugins/sunbeam_hypervisor.py
```diff
@@ -0,0 +1,80 @@
...
+            f'{self.common_dir}/lib/nova/instances/*/console.log',
+            f'{self.common_dir}/cache/libvirt/qemu/capabilities/*.xml',
+        ])
+
+        self.add_forbidden_path([
+            f'{self.common_dir}/etc/ssl/',
+            f'{self.common_dir}/etc/libvirt/secrets',
+            f'{self.common_dir}/etc/libvirt/passwd.db',
+            f'{self.common_dir}/etc/libvirt/krb5.tab',
+            f'{self.common_dir}/var/log/ovn/',
```
dnegreira, line 41: Why is this forbidden?

dnegreira: NVM I see we are gathering it on the ovn_host so we don't  want to duplicate

pmoravec (approved):

TurboTurtle (approved):

PR #3587, opened 2024-04-05
[cirrus] Run tests on latest daily builds for ubuntu

arif-ali:
This PR is to ensure that we test against the upcoming releases of ubuntu, so that we don't encounter issues at release time, and we can solve these before releases

---

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -21,6 +21,7 @@ env:
     FEDORA_PROJECT: "fedora-cloud"
     SOS_PROJECT: "sos-devel-jobs"
     UBUNTU_PROJECT: "ubuntu-os-cloud"
+    UBUNTU_DEVEL_PROJECT: "ubuntu-os-cloud-devel"
 
     # Images exist on GCP already
     CENTOS_9_IMAGE_NAME: "centos-stream-9-v20230809"
@@ -28,12 +29,14 @@ env:
     DEBIAN_IMAGE_NAME: "debian-11-bullseye-v20230809"
     FEDORA_IMAGE_NAME: "fedora-cloud-base-gcp-38-1-6-x86-64"
     FEDORA_PRIOR_IMAGE_NAME: "fedora-cloud-base-gcp-37-1-7-x86-64"
+
     UBUNTU_DEB_IMAGE_NAME: "ubuntu-minimal-2310-mantic-amd64-v20231030"
     UBUNTU_LATEST_IMAGE_NAME: "ubuntu-2310-mantic-amd64-v20231031"
     UBUNTU_IMAGE_NAME: "ubuntu-2204-jammy-v20231030"
     UBUNTU_PRIOR_IMAGE_NAME: "ubuntu-2004-focal-v20231101"
     UBUNTU_PRIOR2_IMAGE_NAME: "ubuntu-1804-bionic-v20230605"
     UBUNTU_SNAP_IMAGE_NAME: "ubuntu-2204-jammy-v20231030"
+    UBUNTU_DEVEL_FAMILY_NAME: "ubuntu-2404-lts-amd64"
 
     # Curl-command prefix for downloading task artifacts, simply add the
     # the url-encoded task name, artifact name, and path as a suffix.
@@ -226,13 +229,39 @@ report_stageone_task:
         pip3 install avocado-framework==94.0 ${PIP_EXTRA}
     # run the unittests separately as they require a different PYTHONPATH in
     # order for the imports to work properly under avocado
-    unittest_script: PYTHONPATH=. avocado run tests/unittests/
-    main_script: PYTHONPATH=tests/ avocado run -p TESTLOCAL=true --test-runner=runner -t stageone tests/{cleaner,collect,report,vendor}_tests
+    unittest_script: &unit_test |
+        PYTHONPATH=. avocado run tests/unittests/
+    main_script: &stageone_test
+        PYTHONPATH=tests/ avocado run -p TESTLOCAL=true --test-runner=runner -t stageone tests/{cleaner,collect,report,vendor}_tests
     on_failure:
         fail_script: &faillogs |
             ls -d /var/tmp/avocado* /root/avocado* 2> /dev/null | xargs tar cf sos-fail-logs.tar
         log_artifacts: &logs
             path: "sos-fail-logs.tar"
+
+report_stageone_daily_task:
+    alias: "stageone_daily_report"
+    name: "Report Stage One - ${UBUNTU_DEVEL_FAMILY_NAME}"
+    allow_failures: true
+    depends_on:
+        - snap_build
+    gce_instance:
+        image_project: ${UBUNTU_DEVEL_PROJECT}
+        image_family: ${UBUNTU_DEVEL_FAMILY_NAME}
+        type: e2-medium
+    environment:
+      PKG: "snap"
+    setup_script: *setup
+    avocado_install_script: &avocado_setup |
+        pip3 uninstall -y avocado-framework --break-system-packages
+        git clone --single-branch -b 92lts https://github.com/avocado-framework/avocado.git
+        cd avocado
+        make install
+    unittest_script: *unit_test
+    main_script: *stageone_test
+    on_failure:
+        fail_script: *faillogs
+        log_artifacts: *logs
 
 # IFF the stage one tests all pass, then run stage two for latest distros
 report_stagetwo_task:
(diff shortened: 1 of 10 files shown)


TurboTurtle (commented):
LGTM overall.

Only one potential, albeit very small, concern. With `allow_failures` if the test fails it reports as neutral, which is easy to overlook in the same vein as overlooking skipped tests (like the Foreman tests when the changes don't trigger a Foreman test).

From https://cirrus-ci.org/guide/writing-tasks/#failure-toleration there is also a `skip_notifications` but it's not clear to me how that's actually different.

Its been a while since I've looked deeply into the GH framework for these jobs, but is there a way to produce a warning(-esque) status that doesn't fail the entire set of checks but stands out a bit more than a neutral response?

arif-ali:
Yeah, the documentation does seem a bit weird

I agree, I would prefer to be notified of the issues, but not fail the whole build. I'll have a go at changing and adding skip_notifications, and see if that helps

arif-ali:
so `skip_notification` is worse, as we don't get anything on GH, and we have to go directly to Cirrus.

I don't want to cause failures due to daily builds of alpha/beta images, but gives us an idea if it could be a python issue or an OS issue, and we can ignore as required.

I'll revert to `allow_failures=true` for the time being until there is a better way to do it

tests/cleaner_tests/basic_function_tests/report_with_mask.py
```diff
@@ -119,7 +119,7 @@ def test_binary_removed(self):
         self.assertFileNotCollected('var/log/binary_test.tar.xz')
 
     def test_binaries_removed_reported(self):
-        self.assertOutputContains('\[removed .* unprocessable files\]')
+        self.assertOutputContains(r'\[removed .* unprocessable files\]')
```
dnegreira, line 122: I have which might be a better idea, so tagging @TurboTurtle and @pmoravec to gather their thoughts.

On the `assertOutputContains`, `assertOutputNotContains`, `assertSosLogContains` maybe we could convert the input to a raw string and pass the raw string to the regexp method?
This way we could always use a regular string as an input to those functions rather than remembering to pass a raw string.

Example on `assertOutputContains`:
```
content_raw = fr"{content}"
found = re.search(fr"(.*)?{content_raw}(.*)?", self.cmd_output.stdout + self.cmd_output.stderr)
```
Hope it makes sense.

pmoravec: Eah that makes sense to me.

TurboTurtle: Yep, makes sense here.

.cirrus.yml
```diff
@@ -226,14 +229,40 @@ report_stageone_task:
...
+        image_family: ${UBUNTU_DEVEL_FAMILY_NAME}
+        type: e2-medium
+    environment:
+      PKG: "snap"
+    setup_script: *setup
+    avocado_install_script: &avocado_setup |
+        pip3 uninstall -y avocado-framework --break-system-packages
+        git clone --single-branch -b 92lts https://github.com/avocado-framework/avocado.git
+        cd avocado
+        make install
```
arif-ali, line 259: doing this to ensure to get the latest from the `92lts` branch so that it has the support for python3.12. Once they publish to pypi, we can then grab the 92.X lts.

thoughts?

arif-ali (commented):
So, seems like `allow_failures=true` will give green if it passes, but would stay neutral if it fails, and appears at the top of the checks. So, in essence this can still help us in the future.

tests/cleaner_tests/basic_function_tests/report_with_mask.py
arif-ali: this doesn't work, we still need the change to add the `r` before the string in the function, otherwise we get loads of these messages

```
tests/cleaner_tests/basic_function_tests/report_with_mask.py:122: SyntaxWarning: invalid escape sequence '\['
  self.assertOutputContains('\[removed .* unprocessable files\]'
```

below is the diff I had

```diff
diff --git a/tests/cleaner_tests/basic_function_tests/report_with_mask.py b/tests/cleaner_tests/basic_function_tests/report_with_mask.py
index bb72cdde..3c860ba5 100644
--- a/tests/cleaner_tests/basic_function_tests/report_with_mask.py
+++ b/tests/cleaner_tests/basic_function_tests/report_with_mask.py
@@ -119,7 +119,7 @@ class DefaultRemoveBinaryFilesTest(StageTwoReportTest):
         self.assertFileNotCollected('var/log/binary_test.tar.xz')
 
     def test_binaries_removed_reported(self):
-        self.assertOutputContains(r'\[removed .* unprocessable files\]')
+        self.assertOutputContains('\[removed .* unprocessable files\]')
 
 
 class KeepBinaryFilesTest(StageTwoReportTest):
diff --git a/tests/sos_tests.py b/tests/sos_tests.py
index 1fb31875..fd833183 100644
--- a/tests/sos_tests.py
+++ b/tests/sos_tests.py
@@ -333,7 +333,8 @@ class BaseSoSTest(Test):
         :param content:  The string that should not be in stdout
         :type content:  ``str``
         """
-        found = re.search(r"(.*)?%s(.*)?" % content, self.cmd_output.stdout + self.cmd_output.stderr)
+        content_raw = fr"{content}"
+        found = re.search(fr"(.*)?{content}(.*)?", self.cmd_output.stdout + self.cmd_output.stderr)
         assert found, "Content string '%s' not in output" % content
 
     def assertOutputNotContains(self, content):
@@ -342,7 +343,8 @@ class BaseSoSTest(Test):
         :param content:  The string that should not be in stdout
         :type content:  ``str``
         """
-        found = re.search(r"(.*)?%s(.*)?" % content, self.cmd_output.stdout + self.cmd_output.stderr)
+        content_raw = fr"{content}"
+        found = re.search(fr"(.*)?{content}(.*)?", self.cmd_output.stdout + self.cmd_output.stderr)
         assert not found, "String '%s' present in stdout" % content
 
```

.cirrus.yml
dnegreira: Seems good as a workaround for the time being until we wait for the new 92.X LTS is released into pypi.

tests/cleaner_tests/basic_function_tests/report_with_mask.py
dnegreira: You need to remove the escapes that the error complaints about like `\[` and `\(` as they are not valid escape sequences anymore, I think there are also more warnings for different characters and they need to be removed as well.

dnegreira: After some testing and talking with @arif-ali , seems that I was wrong on this :) one -needs- to escape the \[ and \( (and others..) in order for the regexp to work properly.

The alternative for us is to sanitize the input on all the functions, similar to my comment, instead of converting to raw we would have to verify if there are any non-escaped characters, escape them, and input that into the `re.` functions. But this is already some extra work and I think we should proceed and just convert those strings to raw where necessary, as well as doing this for all of the future string matching that we need.

dnegreira:
LGTM :+1:

dnegreira (approved):

arif-ali:
@TurboTurtle @pmoravec we are keen for this one to land, so that this can be added to our patch-set for our release. So any reviews on this would be greatly appreciated

.cirrus.yml
pmoravec: Maybe worth opening an sos issue to keep track this as a kind of TODO point?

pmoravec (approved):

.cirrus.yml
arif-ali: I agree, this will be tracked in #3588, That issue will not close till we have a permanent good solution

TurboTurtle (approved):
LGTM, my only concern was the LTS branch and tracking a permanent fix, but that's already been addressed above.

arif-ali:
> LGTM, my only concern was the LTS branch and tracking a permanent fix, but that's already been addressed above.

Out of curiosity, any reason why we are tracking 94.0, or was that the version available at the time? That is, if we shifted to 92.X would there be a problem for all the tests?

TurboTurtle:
> > LGTM, my only concern was the LTS branch and tracking a permanent fix, but that's already been addressed above.
> 
> Out of curiosity, any reason why we are tracking 94.0, or was that the version available at the time? That is, if we shifted to 92.X would there be a problem for all the tests?

Short answer is yes, it will largely (but not completely) break design of "one sos execution per test class instead of once per test class method". Longer answer I'd need to trawl back through a bunch of history to find, and the details of which are probably more buried in my old RH email that I no longer have access to.


I wasn't aware until just now that a new LTS (103.0) is available, so ideally we should look at moving to that release. The biggest change is that it drops `runner` in favor of `nrunner` to run the tests, with the aim of being highly parallelized but that actually breaks our tests as it tries to execute the checks before the report run is finished, so we need to either find a way around that or limit the parallelism to single thread. There will probably be other things that break as well.

PR #3591, opened 2024-04-09
[testing] Add most testing via tox.ini

arif-ali:
Consolidate testing, so that things can be run from CLI, CI and various other means in a consistent way by using tox.
    
* Updated all the flake8 tests in the tests folder.
* Added pylint as a neutral test so that we can work on this in stages and have some collaboration on what we test and don't
* The tox tests for unit, stageone, stagetwo testing makes is easier for users to know how to run tests, and not have to do things manually
* Using tox for CI doesn't make sense, as that will create virtual envs and will disregard system/snap based python modules so may not work
    
---


tox.ini
```diff
@@ -0,0 +1,67 @@
...
+commands =
+    avocado run -p TESTLOCAL=true --test-runner=runner -t foreman tests/product_tests/foreman/
+
+[testenv:nosetests]
+basepython = python3
+deps =
+    {[testenv]deps}
+    nose3
+commands =
+    nosetests -v --with-coverage --cover-package=sos tests/unittests --cover-html
```
TurboTurtle, line 54: We haven't used `nose` in a while, any particular reason behind bringing it back?

TurboTurtle (commented):
Mostly tracks, I think. One question follows.

tox.ini
arif-ali: Ah, I picked that one up from the tests that is run while the deb package is being built.

We don't have avocado in our repost yet, and this is some thing @dnegreira is going to correct in this cycle. Once that is done we can move to use avocado natively.

But this does work

TurboTurtle (approved):

tests/cleaner_tests/existing_archive.py
```diff
@@ -26,34 +27,52 @@ class ExistingArchiveCleanTest(StageTwoReportTest):
...
-        with open(os.path.join(self.tmpdir, '%s-obfuscation.log' % ARCHIVE), 'r') as log:
+        with open(os.path.join(
+                self.tmpdir,
+                f'{ARCHIVE}-obfuscation.log'), 'r') as log:
             for line in log:
                 if "Loaded %s" % ARCHIVE in line:
-                    assert 'as type sos report archive' in line, "Incorrect archive type detected: %s" % line
+                    assert \
+                        'as type sos report archive' in line, \
+                        "Incorrect archive type detected: %s" % line
```
pmoravec, line 42: Does it matter we dont use f-strings here?

tests/collect_tests/help_output_tests.py
```diff
@@ -88,17 +98,26 @@ def test_cluster_profiles_shown(self):
         # this has the secondary effect of enforcing a stylistic requirement
         # where at least one profile must match the pyfile name
         for clus in clusters:
-            assert clus in _profs, "Cluster '%s' not displayed in --list-options" % clus
+            assert \
+                clus in _profs, \
+                "Cluster '{clus}' not displayed in --list-options"
```
pmoravec, line 103: missing leading 'f' to denote f-string

tests/sos_tests.py
```diff
@@ -600,11 +635,15 @@ def assertPluginIncluded(self, plugin):
         :type plugin:  `` str``
         """
         if not self.manifest:
-            self.error("No manifest found, cannot check for %s execution" % plugin)
+            self.error(
+                "No manifest found, cannot check for {plugin} execution"
```
pmoravec, line 639: Same missing leading 'f'.

tests/sos_tests.py
```diff
@@ -614,11 +653,15 @@ def assertPluginNotIncluded(self, plugin):
...
```
pmoravec, line 657: Ditto.

pmoravec (approved):
ACK up to the three f-strings.

arif-ali:
good catches, thanks, fixed those. Found a few that I didn't convert to f-strings on some of my changes, so did those too.

Thanks for all the feedback

pmoravec (approved):

PR #3595, opened 2024-04-15
[ssh] Collect sshd_config.d directory

pmoravec:
Collect sshd config files in its dedicated sshd_config.d directory, as well as config files included from them.

Resolves: #3595

---

### DIFF
--- a/sos/report/plugins/ssh.py
+++ b/sos/report/plugins/ssh.py
@@ -8,6 +8,7 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
+from glob import glob
 from sos.report.plugins import Plugin, IndependentPlugin, PluginOpt
 
 
@@ -33,7 +34,8 @@ class Ssh(Plugin, IndependentPlugin):
 
         sshcfgs = [
             "/etc/ssh/ssh_config",
-            "/etc/ssh/sshd_config"
+            "/etc/ssh/sshd_config",
+            "/etc/ssh/sshd_config.d/*",
             ]
 
         # Include main config files
@@ -49,7 +51,12 @@ class Ssh(Plugin, IndependentPlugin):
         """ Include subconfig files """
         # Read configs for any includes and copy those
         try:
-            for sshcfg in sshcfgs:
+            cfgfiles = [
+                f for files in [
+                    glob(copyspec, recursive=True) for copyspec in sshcfgs
+                ] for f in files
+            ]
+            for sshcfg in cfgfiles:
                 tag = sshcfg.split('/')[-1]
                 with open(self.path_join(sshcfg), 'r',
                           encoding='UTF-8') as cfgfile:


pmoravec:
The implementation (of traversing the directory within `included_configs`) is not ideal: it skips hidden files (until we support Python 3.11+ that [supports](https://docs.python.org/3/library/glob.html#glob.glob) `include_hidden` parameter). I failed to find any more elegant solution covering also that, though.

arif-ali (approved):

sos/report/plugins/ssh.py
```diff
@@ -33,7 +34,8 @@ def setup(self):
 
         sshcfgs = [
             "/etc/ssh/ssh_config",
-            "/etc/ssh/sshd_config"
+            "/etc/ssh/sshd_config",
+            "/etc/ssh/sshd_config.d/*",
             ]
```
jcastill, line 39: I ran an "old" sos (4.6) and it seems that we capture the .d directories already:
```
# ls /var/tmp/sosreport-localhost-2024-04-15-rxpnjai/etc/ssh/
ssh_config  ssh_config.d  sshd_config  sshd_config.d
```

And in the logs:

```
2024-04-15 09:58:56,424 INFO: [plugin:ssh] collecting path '/etc/ssh/sshd_config.d/50-redhat.conf'
2024-04-15 09:58:56,424 DEBUG: [plugin:ssh] copying path '/etc/ssh/sshd_config.d/50-redhat.conf' to archive:'/etc/ssh/sshd_config.d/50-redhat.conf'
2024-04-15 09:58:56,425 DEBUG: [archive:TarFileArchive] Making leading paths for /etc/ssh/sshd_config.d
```

Could be that when we run through sshcfgs below we take these as the beginning of the strings to check?

pmoravec: Interesting.. Well, the `50-redhat.conf` is collected due to:

    Include /etc/ssh/sshd_config.d/*.conf

in `/etc/ssh/sshd_config`. Anyway I tested the file is honoured by `sshd` even without the `Include` line (and e.g. `https://manpages.debian.org/unstable/openssh-server/sshd_config.5.en.html` does suggest so as well).

Due to that "default include", it makes sense to dollect the dir (or just `*.conf` files..?), I think.

TurboTurtle (approved):

sos/report/plugins/ssh.py
```diff
@@ -49,7 +51,12 @@ def included_configs(self, sshcfgs):
         """ Include subconfig files """
         # Read configs for any includes and copy those
         try:
-            for sshcfg in sshcfgs:
+            cfgfiles = [
+                f for files in [
+                    glob(copyspec, recursive=True) for copyspec in sshcfgs
+                ] for f in files
+            ]
```
TurboTurtle, line 58: I think this could be cleaned up slightly, but it does track as it is.

LGTM either way.

PR #3597, opened 2024-04-15
[pylint] pylint and flake8 updates

arif-ali:
pylint for the following rules applied

* C0411: wrong-import-order
* R0912: too-many-branches
* R0914: too-many-locals
* R1725: super-with-arguments
* E1101: no-member

---

### DIFF
--- a/sos/collector/sosnode.py
+++ b/sos/collector/sosnode.py
@@ -8,6 +8,8 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
+# pylint: disable=too-many-branches
+
 import fnmatch
 import inspect
 import logging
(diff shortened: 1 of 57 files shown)


sos/collector/sosnode.py
```diff
@@ -838,36 +840,25 @@ def remove_file(self, path):
...
-        if self.sos_path:
-            if self.need_sudo or self.opts.become_root:
-                try:
-                    self.make_archive_readable(self.sos_path)
-                except Exception:
-                    self.log_error('Failed to make archive readable')
-                    return False
-            self.log_info(f'Retrieving sos report from {self.address}')
-            self.ui_msg('Retrieving sos report...')
+        if self.need_sudo or self.opts.become_root:
```
arif-ali, line 843: The reason why this was refactored was, where `retrieve_sosreport()` function is being used it already checks for `self.sos_path`, so the code under the else seemed redundant

arif-ali:
`pylint` now looks great with the 2 PRs for f-strings done. Based on the output of `pylint` here, #3604 should resolve the rest.

This way, we can then enable it, so that future PRs we'll get errors for f-strings too.

sos/collector/transports/__init__.py
```diff
@@ -322,7 +323,7 @@ def _send_pexpect_password(self, index, result):
         :type result:       ``pexpect.spawn``
         """
         if index == 2:
-            if not self.opts.sudo_pw and not self.opt.nopasswd_sudo:
+            if not self.opts.sudo_pw and not self.opts.nopasswd_sudo:
```
pmoravec, line 326: Nice catch.

sos/policies/distros/redhat.py
```diff
@@ -624,7 +623,8 @@ def check(cls, remote=''):
 
     def fedora_version(self):
         pkg = self.pkg_by_name("fedora-release") or \
-            self.all_pkgs_by_name_regex("fedora-release-.*")[-1]
+            self.package_manager.all_pkgs_by_name_regex(
+                "fedora-release-.*")[-1]
```
pmoravec, line 627: How this could even work now..? good catch.

pmoravec (approved):

sos/collector/transports/__init__.py
arif-ali: I didn't catch it, `pylint` did

sos/policies/distros/redhat.py
arif-ali: I didn't catch it, `pylint` did

PR #3599, opened 2024-04-15
[kdump] Create AzureKDump class

aadhar-agarwal:
This change collects kdump information for Azure Linux.

With this change, we will check the 'path' variable in /etc/kdump.conf to check where information is being dumped.

If get_vm_core is set to true, collect the latest vm core created
in the last 24 hours that is <= 2GB

---

### DIFF
--- a/sos/report/plugins/kdump.py
+++ b/sos/report/plugins/kdump.py
@@ -8,7 +8,7 @@
 
 import platform
 from sos.report.plugins import Plugin, PluginOpt, RedHatPlugin, DebianPlugin, \
-    UbuntuPlugin, CosPlugin
+    UbuntuPlugin, CosPlugin, AzurePlugin
 
 
 class KDump(Plugin):
@@ -123,5 +123,50 @@ class CosKDump(KDump, CosPlugin):
         self.add_cmd_output('ls -alRh /var/kdump*')
         if self.get_option("collect-kdumps"):
             self.add_copy_spec(["/var/kdump-*"])
+
+
+class AzureKDump(KDump, AzurePlugin):
+
+    files = ('/etc/kdump.conf',)
+    packages = ('kexec-tools',)
+
+    option_list = [
+        PluginOpt("get_vm_core", default=False, val_type=bool,
+                  desc="collect vm core")
+    ]
+
+    def read_kdump_conffile(self):
+        """ Parse /etc/kdump file """
+        path = "/var/crash"
+
+        kdump = '/etc/kdump.conf'
+        with open(kdump, 'r', encoding='UTF-8') as file:
+            for line in file:
+                if line.startswith("path"):
+                    path = line.split()[1]
+
+        return path
+
+    def setup(self):
+        super().setup()
+
+        self.add_copy_spec([
+            "/etc/kdump.conf",
+            "/usr/lib/udev/rules.d/*kexec.rules"
+        ])
+
+        try:
+            path = self.read_kdump_conffile()
+        except Exception:  # pylint: disable=broad-except
+            # set no filesystem and default path
+            path = "/var/crash"
+
+        self.add_cmd_output(f"ls -alhR {path}")
+        self.add_copy_spec(f"{path}/*/vmcore-dmesg.txt")
+        self.add_copy_spec(f"{path}/*/kexec-dmesg.log")
+
+        # collect the latest vmcore created in the last 24hrs <= 2GB
+        if self.get_option("get_vm_core"):
+            self.add_copy_spec(f"{path}/*/vmcore", sizelimit=2048, maxage=24)
 
 # vim: set et ts=4 sw=4 :


sos/report/plugins/kdump.py
```diff
@@ -124,4 +124,42 @@ def setup(self):
...
+    packages = ('kexec-tools',)
+
+    def read_kdump_conffile(self):
+        """ Parse /etc/kdump file """
+        path = "/var/crash"
+
+        kdump = '/etc/kdump.conf'
+        with open(kdump, 'r', encoding='UTF-8') as file:
+            for line in file:
+                if line.startswith("path"):
```
jcastill, line 140: Does this mean that azure kexec/kdump can only dump the vmcores into already mounted filesystems?

sos/report/plugins/kdump.py
```diff
@@ -124,4 +124,42 @@ def setup(self):
...
+        ])
+
+        try:
+            path = self.read_kdump_conffile()
+        except Exception:  # pylint: disable=broad-except
+            # set no filesystem and default path
+            path = "/var/crash"
+
+        if self.get_option("all_logs"):
+            self.add_copy_spec(f"{path}/**")
```
jcastill, line 160: This could potentially capture the vmcore themselves, not just the dmesg files. Capturing everything in that directory could create huge sos reports, and I'm not sure what's the benefit to capture old vmcores.

aadhar-agarwal: We do want to capture the vmcore files. However, it could create really huge sos reports. I took a look at it again and created a `get_vm_core` plugin opt that is set to false by default. If this value is set to true, then we will collect the vmcore only if its size is <= 2GB and has been created in the last 24 hours. What do you think about this approach? :)

sos/report/plugins/kdump.py
aadhar-agarwal: It dumps the vmcores into the rootfs by default; we wanted to start with this approach for now (by only reading the path variable in the /etc/kdump.conf file)

sos/report/plugins/kdump.py
jcastill: I like it. In support I remember having the issue of having to ask for both sos report and to upload the vmcores. If we can do this  in one step, we can have all the data we need really easily.

sos/report/plugins/kdump.py
```diff
@@ -124,4 +124,49 @@ def setup(self):
...
+            # set no filesystem and default path
+            path = "/var/crash"
+
+        self.add_cmd_output(f"ls -alhR {path}")
+        self.add_copy_spec(f"{path}/*/vmcore-dmesg.txt")
+        self.add_copy_spec(f"{path}/*/kexec-dmesg.log")
+
+        # collect the latest vmcore created in the last 24hrs <= 2GB
+        if self.get_option("get_vm_core"):
+            self.add_copy_spec(f"{path}/*/vmcore", sizelimit=2048, maxage=24)
```
TurboTurtle, line 170: The use of `sizelimit` here may be undesirable. With `add_copy_spec()`, this parameter tells sos to copy the last X MB of the file if it is over the given size, which would result in an unusable binary blob in the archive taking up 2GB of space.

What we would want here is to either capture a usable vmcore, or nothing at all, so rather than the use of `sizelimit`, we should get a list of potential vmcore(s) to copy, check their size and age (e.g. `os.stat()`), and only then pass the path to an `add_copy_spec()` call.

`maxage` makes sense here.

sos/report/plugins/kdump.py
```diff
@@ -124,4 +124,49 @@ def setup(self):
...
             self.add_copy_spec(["/var/kdump-*"])
 
+
+class AzureKDump(KDump, AzurePlugin):
+
+    files = ('/etc/kdump.conf',)
+    packages = ('kexec-tools',)
+
+    option_list = [
+        PluginOpt("get_vm_core", default=False, val_type=bool,
```
TurboTurtle, line 134: @pmoravec @arif-ali did we ever come up with a decision on dashes-vs-underscores for plugin options? We have a small mix today, whereas our top-level options are all dashes. I remember bringing this up on IRC a long time ago, but can't recall if we ever made a choice?

sos/report/plugins/kdump.py
aadhar-agarwal: This is a great point. When I was testing actually, I saw the "File is over size limit and is bianry" log - So, I think the last X MB is not copied?

https://github.com/sosreport/sos/blob/main/sos/report/plugins/__init__.py#L1854C1-L1860C37

Please correct me if I am wrong. Based on this, I think the whole binary is copied or nothing at all is copied :)

TurboTurtle: Oh, duh. We did change that default a while back. I swear some days I don't know where my head is at.

aadhar-agarwal: No worries at all :) 

Glad we're on the same page!

sos/report/plugins/kdump.py
aadhar-agarwal: Agreed!

sos/report/plugins/kdump.py
arif-ali: I don't recall on any discussions, maybe it was before I got active on the project

TurboTurtle: Yeah, the main reason is consistency for user experience. If we didn't come up with anything back then, we can start a new discussion around it and not hold up this PR.

TurboTurtle (approved):

aadhar-agarwal:
@jcastill Could I get another review on this when you get some time? Thank you!

jcastill:
@aadhar-agarwal done! But you need another maintaner to add their review, mine is worth nothing.
@pmoravec @arif-ali can you check this when you have time?

arif-ali:
I'm curious, the AzurePlugin has a lot of similar items for collection to RedHatPlugin, would it be worth extending the `AzureKdump` on top of `RedhatKdump`?

pmoravec:
> I'm curious, the AzurePlugin has a lot of similar items for collection to RedHatPlugin, would it be worth extending the `AzureKdump` on top of `RedhatKdump`?

That makes sense to me. Or maybe even have one common plugin for both distros?

The diffs are
- `get_vm_core` option in Azure (can be beneficial for RedHat as well, I guess)
- RedHat parses more config files from `/etc/kdump.conf` (does it harm for Azure..? I doubt)
- `lsinitrd` is conditionally collected in RedHat
- a few copy_specs differ (union of them does not harm)

Either inheritance or common plugin sounds good to me. Current approach "a lot of code is shared, some bits differ" is not ideal but acceptable to me.

jcastill:
I thought the same the key here is that they are not using the fs mount option, so their_kdump_conffile is much simpler than the rh one, and I'm not sure if we could/should make that function inside the class check what distribution is running and return different vmcore targets depending on that.
That said, I need to review the RedHat side of this plugin because there may be some things we can improve post RHEL 9 - so we can have this PR merged and have Azure collect their kdump info now, and then I can take a look and see if Azure and RH ones can be merged into one class, if that's ok with you all.

arif-ali:
> I thought the same the key here is that they are not using the fs mount option, so their_kdump_conffile is much simpler than the rh one, and I'm not sure if we could/should make that function inside the class check what distribution is running and return different vmcore targets depending on that. That said, I need to review the RedHat side of this plugin because there may be some things we can improve post RHEL 9 - so we can have this PR merged and have Azure collect their kdump info now, and then I can take a look and see if Azure and RH ones can be merged into one class, if that's ok with you all.

makes good sense, let's create a new issue out of this, so that it is not forgotten; then we can circle back, and make improvements

arif-ali (approved):

PR #3600, opened 2024-04-16
**[AAP_Receptor]** Change receptorctl status output to json

thom-at-redhat:
---


arif-ali (changes requested):
In a lot of the plugins we have, we have both the normal output and json output. Would it make sense to have both here?

Also, can you have a look at the DCO, and update your commit with Signed-off-by please based on the contribution guidelines

thom-at-redhat:
> In a lot of the plugins we have, we have both the normal output and json output. Would it make sense to have both here?
> 
> Also, can you have a look at the DCO, and update your commit with Signed-off-by please based on the contribution guidelines

Will do

arif-ali:
you didn't need to close the PR, as you can update the same branch

PR #3609, opened 2024-04-20
[Global] Resolve empty except alerts from CodeQL

TurboTurtle:
This commit resolves the remaining 'Empty except' alerts from CodeQL scanning.

Most of these involve logging something during exception handling when we were previously simply `pass`ing. Others, either when logging is unavailable or it makes no sense to log something for, are resolved by placing comments within the except block as the CodeQL alerts should be suppressed if the 'empty except' (`pass`ing) has an explanatory note.

---

### DIFF
--- a/sos/collector/clusters/ocp.py
+++ b/sos/collector/clusters/ocp.py
@@ -226,6 +226,8 @@ class ocp(Cluster):
                 try:
                     idx[state] = statline.index(state.upper())
                 except Exception:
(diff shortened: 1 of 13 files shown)


arif-ali (approved):

arif-ali:
interestingly the `msr` module is now being loaded by a plugin for 24.04 (`noble`), we need to investigate what that is. Will open a new issue

pmoravec (approved):

PR #3612, opened 2024-04-22
[kdump] Capture more commands for RH kdump

jcastill:
This change enables the RH plugin to display the content of /var/crash via the command 'ls', and adds a new option to capture the latest vmcores generated
in the last 24 hours, smaller than 2GB, as it's
done in the AzureDump class.

---

### DIFF
--- a/sos/report/plugins/kdump.py
+++ b/sos/report/plugins/kdump.py
@@ -42,6 +42,11 @@ class RedHatKDump(KDump, RedHatPlugin):
     files = ('/etc/kdump.conf',)
     packages = ('kexec-tools',)
 
+    option_list = [
+        PluginOpt("get-vm-core", default=False, val_type=bool,
+                  desc="collect vm core")
+    ]
+
     def fstab_parse_fs(self, device):
         """ Parse /etc/fstab file """
         fstab = self.path_join('/etc/fstab')
@@ -77,6 +82,7 @@ class RedHatKDump(KDump, RedHatPlugin):
         self.add_copy_spec([
             "/etc/kdump.conf",
             "/etc/udev/rules.d/*kexec.rules",
+            "/usr/lib/udev/rules.d/*kexec.rules",
             "/var/crash/*/kexec-dmesg.log",
             "/var/log/kdump.log"
         ])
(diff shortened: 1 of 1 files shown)


jcastill:
Opening this as a draft for now to discuss the possibility of merging AzureKDump and RedHatKDump (and any other changes we can make in general). With this PR as it is now, the only difference, unless I've missed something, is in the function read_kdump_conffile. In RH case we can dump vmcores to a directory, or specify a filesystem that will be mounted, by adding the fs at the beginning of the line. With Azure, it seems that only path is allowed. So what would be the best/recommended option to deal with having everything the same between these two classes except that specific function?

pmoravec (approved):

sos/report/plugins/kdump.py
```diff
@@ -42,6 +42,11 @@ class RedHatKDump(KDump, RedHatPlugin):
     files = ('/etc/kdump.conf',)
     packages = ('kexec-tools',)
 
+    option_list = [
+        PluginOpt("get_vm_core", default=False, val_type=bool,
```
arif-ali, line 46: can we change this to use hyphens, as per the global changes that were made a few days ago

sos/report/plugins/kdump.py
```diff
@@ -88,9 +93,14 @@ def setup(self):
             # set no filesystem and default path
             path = "/var/crash"
 
+        self.add_cmd_output(f"ls -alhR {path}")
         self.add_copy_spec(f"{path}/*/vmcore-dmesg.txt")
         self.add_copy_spec(f"{path}/*/kexec-dmesg.log")
 
+        # collect the latest vmcore created in the last 24hrs <= 2GB
+        if self.get_option("get_vm_core"):
```
arif-ali, line 101: same here

arif-ali (changes requested):

sos/report/plugins/kdump.py
jcastill: Done

sos/report/plugins/kdump.py
jcastill: And done

jcastill:
I've changed the option in the Azure plugin as well, because it was strange to have two different formats for the option. We haven't shipped the Azure kdump changes made some weeks ago yet, right? So no user has seen the option and we have still the room to change it.
That said, I was thinking on moving the option from get-vm-core to get-vmcore. It reads more naturally (we never asked for a vm-core in support, it was always vmcore) and the file itself is not hyphenated. What do you think?

I'm still working on merging Azure and RH plugins because they are almost the same, but that will have to be another PR.

arif-ali (approved):
I think the Azure bit was already changed by Jake

nevertheless, I think the Azure Linux folks have their own way of incorporating stuff, at least based on hat I saw a few months back, when they were trying to get 4.6.1 in. So, it could be that, when they added it here, they may of updated their RPMs

TurboTurtle (approved):

PR #3617, opened 2024-04-23
[neutron] Don't collect ml2 certs

arif-ali:
---

### DIFF
--- a/sos/report/plugins/openstack_neutron.py
+++ b/sos/report/plugins/openstack_neutron.py
@@ -44,6 +44,38 @@ class OpenStackNeutron(Plugin):
         self.add_forbidden_path("/var/lib/neutron/lock")
         self.add_cmd_output("ls -laZR /var/lib/neutron/lock")
 
+        if self.path_exists(self.var_puppet_gen):
+            ml2_pre = self.var_puppet_gen
+        else:
+            ml2_pre = ""
+
+        ml2_conf_file = f"{ml2_pre}/etc/neutron/plugins/ml2/ml2_conf.ini"
+
+        ml2_certs = []
+
+        ml2_cert_keys = [
+            'ovn_nb_private_key',
+            'ovn_nb_certificate',
+            'ovn_nb_ca_cert',
+            'ovn_sb_private_key',
+            'ovn_sb_certificate',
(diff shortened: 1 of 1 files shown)


sos/report/plugins/openstack_neutron.py
```diff
@@ -44,6 +44,13 @@ def setup(self):
         self.add_forbidden_path("/var/lib/neutron/lock")
         self.add_cmd_output("ls -laZR /var/lib/neutron/lock")
 
+        self.add_forbidden_path([
+            "/etc/neutron/plugins/ml2/cert_host",
+            "/etc/neutron/plugins/ml2/key_host",
+            f"{self.var_puppet_gen}/etc/neutron/plugins/ml2/cert_host",
+            f"{self.var_puppet_gen}/etc/neutron/plugins/ml2/key_host",
```
arif-ali, line 51: It would be good to check if this also exists on RH systems?

/cc @pmoravec @jcastill

arif-ali: @jcastill based my lab env that I found this, the `ml2_conf.ini` has a few parameters that show this. I think I could probably add the `neutron-api-plugin-ovn.crt` too.

I might improve this further to look for those in the config file, so that this can be more dynamic

```
ovn_nb_private_key = /etc/neutron/plugins/ml2/key_host
ovn_nb_certificate = /etc/neutron/plugins/ml2/cert_host
ovn_nb_ca_cert = /etc/neutron/plugins/ml2/neutron-api-plugin-ovn.crt
ovn_sb_private_key = /etc/neutron/plugins/ml2/key_host
ovn_sb_certificate = /etc/neutron/plugins/ml2/cert_host
ovn_sb_ca_cert = /etc/neutron/plugins/ml2/neutron-api-plugin-ovn.crt
```

jcastill: @arif-ali I think checking these entries in the config file and using them is a fantastic idea

arif-ali: The final solution here now is much longer, but this would work for any case now 🤞🏽

TurboTurtle (approved):

PR #3618, opened 2024-04-24
[flatpak] Predicate two cmds by gvfs service

pmoravec:
Two commands "flatpak --installations" and "flatpak --print-updated-env" can enable user's gvfs-daemon.service even with
GVFS_REMOTE_VOLUME_MONITOR_IGNORE=1 set, so let put them under a predicate.

Resolves: #3618
Relevant: RHEL-14328

---


arif-ali (approved):

pmoravec:
The "Report Stage One - centos-stream-8" job (repeatedly?) fails, in the middle of avocado testing, that is weird.. I am re-running it once again.

sos/report/plugins/flatpak.py
```diff
@@ -20,19 +20,27 @@ class Flatpak(Plugin, IndependentPlugin):
 
     def setup(self):
         env = {"GVFS_REMOTE_VOLUME_MONITOR_IGNORE": "1"}
+        gvfs_pred = SoSPredicate(
+            self, cmd_outputs={
+                'cmd': 'systemctl --user status gvfs-daemon.service',
+                'output': '(running)'
+            }
```
TurboTurtle, line 27: `SosPredicate()` supports a ` service` parameter to check for running services. Is that not sufficient here?

TurboTurtle (commented):
Comment below, but also shouldn't this instead be considered a bug to report to flatpak?

sos/report/plugins/flatpak.py
pmoravec: I was thinking about it, but I think the `--user` option is crucial difference here. Let me re-test it..

pmoravec:
> Comment below, but also shouldn't this instead be considered a bug to report to flatpak?

Yes, the RHEL-14328 is bug against flatpak.

sos/report/plugins/flatpak.py
pmoravec: As we get services list via `systemctl list-unit-files --type=service` and that output does not contain users' services, `InitSystem` is not aware of the service despite it is running. So `service` parameter is useless here /o\.

TurboTurtle: Ah, ok, good observation. I guess that makes the question, should we expand our profiling for `SystemdInit` to include user services, or just go forward with this as written and keep our SystemD abstraction in the dark about user services?

TurboTurtle:
> > Comment below, but also shouldn't this instead be considered a bug to report to flatpak?
> 
> Yes, the RHEL-14328 is bug against flatpak.

My question was more of the flavor "does it make sense for sos to make this change, when this is really a bug in flatpak that will (hopefully) be addressed and then render the predicate unnecessary?"

Alternatively, do we need to version-gate this predicate?

PR #3622, opened 2024-04-25
[tests] Add avocado tests for openstack plugins

arif-ali:
Inititally doing for the most common openstack services

---

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -290,7 +290,7 @@ report_stagetwo_task:
             dnf -y install python3-pexpect
         fi
     main_script: &stagetwo_test |
-        PYTHONPATH=tests/ avocado run -p TESTLOCAL=true --max-parallel-tasks=1 -t stagetwo tests/{cleaner,collect,report,vendor}_tests
+        PYTHONPATH=tests/:. avocado run -p TESTLOCAL=true --max-parallel-tasks=1 -t stagetwo tests/{cleaner,collect,report,vendor}_tests
     on_failure:
         fail_script: *faillogs
         log_artifacts: *logs
--- a/tests/report_tests/plugin_tests/openstack/openstack.py
+++ b/tests/report_tests/plugin_tests/openstack/openstack.py
@@ -1,0 +1,156 @@
+# Copyright (C) 2024 Canonical Ltd., Arif Ali <arif.ali@canonical.com>
+#
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+import json
+
+from sos_tests import StageTwoReportTest, ubuntu_only
+from sos.utilities import shell_out
+
+from avocado.utils import distro
+from avocado.utils.software_manager import distro_packages
+
+
+class OpenstackConfScrubbedTest(StageTwoReportTest):
+    """Ensure that the various openstack confs are picked up and properly
+    scrubbed
+
+    :avocado: tags=stagetwo
+    """
+
+    sos_cmd = ('-o openstack_cinder,openstack_keystone,openstack_glance,'
+               'openstack_heat,openstack_neutron,openstack_nova')
+    files = [
+        ('cinder.conf', '/etc/cinder/cinder.conf'),
+        ('keystone.domain1.conf',
+         '/etc/keystone/domains/keystone.domain1.conf'),
+        ('keystone.conf', '/etc/keystone/keystone.conf'),
+        ('glance-api.conf', '/etc/glance/glance-api.conf'),
+        ('heat.conf', '/etc/heat/heat.conf'),
+        ('neutron.conf', '/etc/neutron/neutron.conf'),
+        ('nova.conf', '/etc/nova/nova.conf'),
+        ('heat.conf', '/etc/heat/heat.conf'),
+    ]
+
+    packages = {
+        'rhel': ['openstack-cinder', 'openstack-glance',
+                 'openstack-heat-common', 'openstack-keystone',
+                 'openstack-nova'],
+        'Ubuntu': ['cinder-common', 'glance-common', 'heat-common',
+                   'keystone-common', 'nova-common']
+    }
+
+    openstack_only = True
+
+    def setup_mocked_packages(self):
+        this_distro = distro.detect()
+        if this_distro.name == "centos" or this_distro.name == "centos-stream":
+            shell_out("dnf config-manager --enable crb")
+            shell_out("dnf config-manager --enable powertools")
+            pkg_to_install = "centos-release-openstack-yoga"
+            installed = distro_packages.install_distro_packages(
+                {this_distro.name: [pkg_to_install]})
+            if not installed:
+                raise Exception(
+                    "Unable to install requested packages "
+                    f"{pkg_to_install}")
+            self._write_file_to_tmpdir(
+                'mocked_osp_package', f'["{pkg_to_install}"]')
+        super().setup_mocked_packages()
+
+    def teardown_mocked_packages(self):
+        this_distro = distro.detect()
+        if this_distro.name == "centos" or this_distro.name == "centos-stream":
+            pkgs = self.read_file_from_tmpdir('mocked_osp_package')
+            if not pkgs:
+                return
+            pkgs = json.loads(pkgs)
+            for pkg in pkgs:
+                self.sm.remove(pkg)
+        super().teardown_mocked_packages()
+
+    def test_cinder_conf_collected_and_scrubbed(self):
+        self.assertFileCollected('/etc/cinder/cinder.conf')
+
+        values_to_mask = [
+            'cmB4zBYq3VWFMNqNKFLcqS5Zq8ystLLsTd5BFLbCtX67qShnhgHFxxRFjkhbY54x',
+            'wPhFqY69x94YVJc7STrVH3CfsFrrcZPYw8NS2pjhzqyzw7wrL2VnTmN58c5XTnfV',
+            'Ck3r7zf6B6PscfjWhhj2zJdy8SNXYd59',
+        ]
+        for value in values_to_mask:
+            self.assertFileNotHasContent('/etc/cinder/cinder.conf', value)
+
+    def test_glance_conf_collected_and_scrubbed(self):
+        self.assertFileCollected('/etc/glance/glance-api.conf')
+
+        values_to_mask = [
+            '4LzVZZLsPTrKw7c49ZdxqmJ8mkcWkSSnPJzgYyZWPMkrVzyB62CyCzzKVRnBS4Mz',
+            'cbVtRMqPznc4M7XKVG73yPn5fBX4NXYSC4bMb7PGYTL5RkWXnc8ZsZ9hrJy2hynf',
+            'rM5hSHsw8JSV2mJVkcszMBxtVsHbG4MW',
+        ]
+
+        for value in values_to_mask:
+            self.assertFileNotHasContent('/etc/glance/glance-api.conf', value)
+
+    def test_heat_conf_collected_and_scrubbed(self):
+        self.assertFileCollected('/etc/heat/heat.conf')
+
+        values_to_mask = [
+            'TH4BRqtVNt4WhWkrqmKZX4Kpf4pr7cCT',
+            'sXNkqpmgprNs69SdF7jXM9JP9Nfc7m7n',
+            'zy5sPgTkbwF6qL6Gc4wngyxYFwB3r77HjRJtXWbVP4Cf4psmfJmqw2MZqZ8jxxFC',
+            'Jbf8JkTW2PXYC6hH5YC8XzmjYrLkcT6wJSZzccstNcnnFNNsKXz54PVd7NhKT45j',
+        ]
+
+        for value in values_to_mask:
+            self.assertFileNotHasContent('/etc/heat/heat.conf', value)
+
+    def test_keystone_conf_collected_and_scrubbed(self):
+        self.assertFileCollected('/etc/keystone/keystone.conf')
+        self.assertFileCollected('/etc/keystone/keystone.policy.yaml')
+
+        self.assertFileNotHasContent(
+            '/etc/keystone/keystone.conf', '2zx9jZZtxdn4grG3xcMV4PwgGwY7X7fP')
+
+    def test_keystone_ldap_conf_scrubbed(self):
+        self.assertFileNotHasContent(
+            '/etc/keystone/domains/keystone.domain1.conf', 'crapper')
+
+    @ubuntu_only
+    def test_neutron_ml2_certs_not_collected(self):
+        self.assertFileNotCollected('/etc/neutron/plugins/ml2/cert_host')
+        self.assertFileNotCollected('/etc/neutron/plugins/ml2/key_host')
+        self.assertFileNotCollected(
+            '/etc/neutron/plugins/ml2/neutron-api-plugin-ovn.crt')
+
+    def test_neutron_conf_collected_and_scrubbed(self):
+        self.assertFileCollected('/etc/neutron/neutron.conf')
+        self.assertFileCollected('/etc/neutron/plugins/ml2/ml2_conf.ini')
+
+        values_to_mask = [
+            'KwGKnxV4tSVMpyhyz8Pj8gXnncrfyKr7nYFJhN7t8m6V3jL8jHKqLchgc6fjLMJ6',
+            '2VX9YzhHm7gMsr7RFLx8Ypckpq5fCshsNM8CfMYBXWZcCPKcwcjXb9rFtzWcBgM4',
+            '8ZMkYGjw5bsYwc6cx72b9ZVBrCKM5JMB',
+        ]
+
+        for value in values_to_mask:
+            self.assertFileNotHasContent('/etc/neutron/neutron.conf', value)
+
+    def test_nova_conf_collected_and_scrubbed(self):
+        self.assertFileCollected('/etc/nova/nova.conf')
+
+        values_to_mask = [
+            'JpdsBfXHTwFMmTHsP8WY2gTRpb4ZrLfN6VmT2666TtZfjYcGKMnrJd7pPRxqtNfs',
+            'WCBG8sqmmpKPYBbpVNSjfsScpmh55W2jJgPnXYb7zM2Rr469g2LL4x642R67dFyX',
+            'Vrn7SkRMHL8FzjNg2NJjwF57t7hJfhjj',
+        ]
+
+        for value in values_to_mask:
+            self.assertFileNotHasContent('/etc/nova/nova.conf', value)
+
+# vim: set et ts=4 sw=4 :
--- a/tests/sos_tests.py
+++ b/tests/sos_tests.py
@@ -32,7 +32,9 @@ SOS_TEST_DATA_DIR = os.path.realpath(os.path.join(SOS_TEST_DIR, 'test_data'))
 SOS_TEST_BIN = os.path.realpath(os.path.join(SOS_TEST_DIR, '../bin/sos'))
 
 RH_DIST = ['rhel', 'centos', 'fedora', 'centos-stream']
+CENTOS_DIST = ['centos', 'centos-stream']
 UBUNTU_DIST = ['Ubuntu', 'debian']
+OPENSTACK_DIST = ['centos', 'centos-stream', 'Ubuntu']
 
 
 def skipIf(cond, message=None):
@@ -51,6 +53,13 @@ def redhat_only(tst):
     def wrapper(func):
         if distro.detect().name not in RH_DIST:
             raise TestSkipError('Not running on a Red Hat distro')
+    return wrapper
+
+
+def openstack_only(tst):
+    def wrapper(func):
+        if distro.detect().name not in OPENSTACK_DIST:
+            raise TestSkipError('Not running on a OpenStack supported distro')
     return wrapper
 
 
@@ -77,6 +86,7 @@ class BaseSoSTest(Test):
     sos_timeout = 600
     redhat_only = False
     ubuntu_only = False
+    openstack_only = False
     end_of_test_case = False
     arch = []
 
@@ -242,6 +252,10 @@ class BaseSoSTest(Test):
         elif self.ubuntu_only:
             if self.local_distro not in UBUNTU_DIST:
                 raise TestSkipError("Not running on a Ubuntu or Debian distro")
+        elif self.openstack_only:
+            if self.local_distro not in OPENSTACK_DIST:
+                raise TestSkipError("Not running on a OpenStack supported "
+                                    "distro")
 
     def check_arch_for_enablement(self):
         """
@@ -847,8 +861,8 @@ class StageTwoReportTest(BaseSoSReportTest):
     def teardown_mocking(self):
         """Undo any and all mocked setup that we did for tests
         """
-        self.teardown_mocked_packages()
         self.teardown_mocked_files()
+        self.teardown_mocked_packages()
         self.teardown_mocked_plugins()
 
     def setup_mocking(self):
--- a/sos/report/plugins/openstack_heat.py
+++ b/sos/report/plugins/openstack_heat.py
@@ -119,7 +119,7 @@ class OpenStackHeat(Plugin):
         protect_keys = [
             "admin_password", "memcache_secret_key", "password",
             "qpid_password", "rabbit_password", "stack_domain_admin_password",
-            "transport_url"
+            "transport_url", "auth_encryption_key"
         ]
         connection_keys = ["connection"]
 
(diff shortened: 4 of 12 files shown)


arif-ali:
mock teardown is potentially removing `/etc/keystone/domains`, and hence causing the issue 🤷🏽, I'll continue to debug

arif-ali:
> mock teardown is potentially removing `/etc/keystone/domains`, and hence causing the issue 🤷🏽, I'll continue to debug

ordering of the files. if `/etc/keystone` is deleted, and then it tries to delete `/etc/keystone/domains`, hence the complain. Reordering the list solves this

tests/report_tests/plugin_tests/openstack/openstack.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+class OpenstackPluginTest(StageOneReportTest):
+    """Basic sanity check to make sure common config files are collected
+
+    :avocado: tags=stageone
+    """
+
+    sos_cmd = ('-o openstack_cinder,openstack_keystone,openstack_glance,'
+               'openstack_neutron,openstack_nova')
```
TurboTurtle, line 19: Would it be better to mark this as a StageTwo test and have it temporarily install a set of openstack packages? When I was first looking at OSP testing at RH, I was looking at doing a base install like we do for Foreman (i.e. the openstack tests would become their own entity under product_tests and report a new CI job) but there were some hurdles to that on single node that I'm not fully recallling at the moment.

In lieu of that, can we temporarily install some openstack packages to get some more confidence in the test? As it is, won't all of these tests pass simply due to them being missing on the test VMs?

arif-ali: I hear you, and was thinking the same, at one point. 

One of the main purposes to add these as is, is to automate some the testing we at Canonical were doing at deb release time, so that we can assert them and ensure we are collecting them all automatically and spend less time collecting the data.

Happy to amend them to StageTwo, if it helps to do it here for RH OSP and Canonical OS

arif-ali: I this kind of points out the discussion point from #3611

If we are doing this in stagetwo, the file will exist, hence we should have a hard assert function. This way we know 100% that we are collecting it, wdyt?

TurboTurtle: Yeah, that's why I said I'd be onboard with a hard assert for `assertFileCollected()` and a renaming of the current method so that we have both "we absolutely 100% have this file" and "we are validating the behavior of `add_copy_spec()`'.

I'd say let's add the temporary package install for these openstack packages (I'm assuming here that the initial files get laid down by the packages, and not by some user-executed configuration script) and mark this as StageTwo.

arif-ali: Cool, no worries, I'll get this re factored and tested

arif-ali:
@TurboTurtle the ubuntu bits was actually quite easy, but centos was a real pain, as you have to install extra package for the openstack repo, and then enable crb repo for glance

I've put my ideas (it has issues), but wanted to see if everyone is on-board with the way the stagetwo is being done here

arif-ali:
we should teardown the files before packages. If we tear down package before files, then the file will not exist, and fail.

Also ensure we collect `/etc/keystone/keystone.policy.yaml`

This should now work for ubuntu distros, will need extra expertise for the RH bits

arif-ali:
While testing on centos9, I was able to overcome the issues represented here, and the package `dnf-plugins-core` was getting installed. Below is the output from my local env that I created today to try and test this, and seems to work. Could it be that `dnf-plugins-core` is not available in the GCE image, and somehow we have to enable it. Or, alternatively enable the CRB repo via a different mechanism

```
❱❱❱ ./do_sos_avocado_rh.sh 
Error: Failed creating project "sos-testing-rh": Failed adding database record: This "projects" entry already exists
Creating sos-avocado01-centos9
Starting sos-avocado01-centos9              
/home/arif/gitRepos/sos_testing
Checking VM status of sos-avocado01-centos9 .........done
Checking cloud-init status of sos-avocado01-centos9 .................done    
Cleaning up any old .tox environments ...
Running Unit Tests ...
Running Stage One Tests ...
Running Stage Two Tests ...
GLOB sdist-make: /root/sos/setup.py
stagetwo_tests create: /root/sos/.tox/stagetwo_tests
stagetwo_tests installdeps: -r/root/sos/requirements.txt, -r/root/sos/test-requirements.txt, python_magic
stagetwo_tests inst: /root/sos/.tox/dist/sos-4.7.1.zip
stagetwo_tests installed: alabaster==0.7.16,attrs==20.3.0,avocado-framework==103.0,Babel==2.9.1,chardet==4.0.0,cloud-init==23.4,configobj==5.0.6,coverage==7.5.1,distlib==0.3.8,distro==1.5.0,docutils==0.21.2,filelock==3.14.0,gpg==1.15.1,idna==2.10,imagesize==1.4.1,importlib_metadata==7.1.0,Jinja2==3.1.3,jsonpatch==1.21,jsonpointer==2.0,jsonschema==3.2.0,libcomps==0.1.18,MarkupSafe==2.1.5,netifaces==0.10.6,oauthlib==3.1.1,packaging==24.0,pexpect==4.9.0,platformdirs==4.2.1,pluggy==0.13.1,prettytable==0.7.2,ptyprocess==0.7.0,py==1.11.0,pycodestyle==2.11.1,Pygments==2.18.0,pyrsistent==0.17.3,pyserial==3.4,PySocks==1.7.1,python-magic==0.4.27,pytz==2021.1,PyYAML==5.4.1,requests==2.25.1,rpm==4.16.1.3,selinux==3.6,sepolicy==3.6,setools==4.4.4,six==1.15.0,snowballstemmer==2.2.0,sos @ file:///root/sos/.tox/dist/sos-4.7.1.zip#sha256=687c773075bc51146876c960903afdf7477183ea026227bb5b761a374df4615a,Sphinx==7.3.7,sphinxcontrib-applehelp==1.0.8,sphinxcontrib-devhelp==1.0.6,sphinxcontrib-htmlhelp==2.0.5,sphinxcontrib-jsmath==1.0.1,sphinxcontrib-qthelp==1.0.7,sphinxcontrib-serializinghtml==1.1.10,tomli==2.0.1,tox==2.5.0,urllib3==1.26.5,virtualenv==20.26.1,zipp==3.18.1
stagetwo_tests runtests: PYTHONHASHSEED='938793694'
stagetwo_tests runtests: commands[0] | avocado run -p TESTLOCAL=true --max-parallel-tasks=1 -t stagetwo tests/report_tests/plugin_tests/openstack
JOB ID     : da0119fdb0b3c6e8d15aec153f0200911368c807
JOB LOG    : /root/avocado/job-results/job-2024-05-05T19.08-da0119f/job.log
 (1/9) tests/report_tests/plugin_tests/openstack/openstack.py:OpenstackConfScrubbedTest.test_cinder_conf_collected_and_scrubbed: STARTED
 (1/9) tests/report_tests/plugin_tests/openstack/openstack.py:OpenstackConfScrubbedTest.test_cinder_conf_collected_and_scrubbed:  PASS (451.55 s)
 (2/9) tests/report_tests/plugin_tests/openstack/openstack.py:OpenstackConfScrubbedTest.test_glance_conf_collected_and_scrubbed: STARTED
 (2/9) tests/report_tests/plugin_tests/openstack/openstack.py:OpenstackConfScrubbedTest.test_glance_conf_collected_and_scrubbed:  PASS (0.08 s)
 (3/9) tests/report_tests/plugin_tests/openstack/openstack.py:OpenstackConfScrubbedTest.test_heat_conf_collected: STARTED
 (3/9) tests/report_tests/plugin_tests/openstack/openstack.py:OpenstackConfScrubbedTest.test_heat_conf_collected:  PASS (0.11 s)
 (4/9) tests/report_tests/plugin_tests/openstack/openstack.py:OpenstackConfScrubbedTest.test_keystone_conf_collected_and_scrubbed: STARTED
 (4/9) tests/report_tests/plugin_tests/openstack/openstack.py:OpenstackConfScrubbedTest.test_keystone_conf_collected_and_scrubbed:  PASS (0.10 s)
 (5/9) tests/report_tests/plugin_tests/openstack/openstack.py:OpenstackConfScrubbedTest.test_keystone_ldap_conf_scrubbed: STARTED
 (5/9) tests/report_tests/plugin_tests/openstack/openstack.py:OpenstackConfScrubbedTest.test_keystone_ldap_conf_scrubbed:  PASS (0.10 s)
 (6/9) tests/report_tests/plugin_tests/openstack/openstack.py:OpenstackConfScrubbedTest.test_neutron_ml2_certs_not_collected: STARTED
 (6/9) tests/report_tests/plugin_tests/openstack/openstack.py:OpenstackConfScrubbedTest.test_neutron_ml2_certs_not_collected:  SKIP: Not running on a Ubuntu or Debian distro
 (7/9) tests/report_tests/plugin_tests/openstack/openstack.py:OpenstackConfScrubbedTest.test_neutron_conf_collected_and_scrubbed: STARTED
 (7/9) tests/report_tests/plugin_tests/openstack/openstack.py:OpenstackConfScrubbedTest.test_neutron_conf_collected_and_scrubbed:  PASS (0.10 s)
 (8/9) tests/report_tests/plugin_tests/openstack/openstack.py:OpenstackConfScrubbedTest.test_nova_conf_collected_and_scrubbed: STARTED
 (8/9) tests/report_tests/plugin_tests/openstack/openstack.py:OpenstackConfScrubbedTest.test_nova_conf_collected_and_scrubbed:  PASS (0.09 s)
 (9/9) tests/report_tests/plugin_tests/openstack/openstack.py:OpenstackConfScrubbedTest.test_archive_created: STARTED
 (9/9) tests/report_tests/plugin_tests/openstack/openstack.py:OpenstackConfScrubbedTest.test_archive_created:  PASS (131.63 s)
RESULTS    : PASS 8 | ERROR 0 | FAIL 0 | SKIP 1 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB TIME   : 596.12 s
___________________________________________________________________________________________________________________________________ summary ____________________________________________________________________________________________________________________________________
  stagetwo_tests: commands succeeded
  congratulations :)
```

arif-ali:
Using the same image in Cirrus, did a quick deploy in my GCE account, and I don't have the same issue as the CI

```
❱❱❱ gcloud compute ssh --zone "us-central1-a" --project "x-alcove-359617" instance-20240505-192203                                                                                                  
Warning: Permanently added 'compute.3568923530342589063' (ED25519) to the list of known hosts.
X11 forwarding request failed
[arif@instance-20240505-192203 ~]$ sudo -i 
[root@instance-20240505-192203 ~]# dnf install dnf-plugins-core
CentOS Stream 9 - BaseOS                                                                                                                                                                                                                        7.4 MB/s | 8.0 MB     00:01    
CentOS Stream 9 - AppStream                                                                                                                                                                                                                      12 MB/s |  19 MB     00:01    
CentOS Stream 9 - Extras packages                                                                                                                                                                                                                39 kB/s |  16 kB     00:00    
Google Compute Engine                                                                                                                                                                                                                            24 kB/s | 7.1 kB     00:00    
Google Cloud SDK                                                                                                                                                                                                                                 45 MB/s | 125 MB     00:02    
Package dnf-plugins-core-4.3.0-9.el9.noarch is already installed.
Dependencies resolved.
================================================================================================================================================================================================================================================================================
 Package                                                                       Architecture                                                Version                                                            Repository                                                   Size
================================================================================================================================================================================================================================================================================
Upgrading:
 dnf-plugins-core                                                              noarch                                                      4.3.0-13.el9                                                       baseos                                                       37 k
 python3-dnf-plugins-core                                                      noarch                                                      4.3.0-13.el9                                                       baseos                                                      264 k

Transaction Summary
================================================================================================================================================================================================================================================================================
Upgrade  2 Packages

Total download size: 301 k
Is this ok [y/N]: y
Downloading Packages:
(1/2): dnf-plugins-core-4.3.0-13.el9.noarch.rpm                                                                                                                                                                                                 106 kB/s |  37 kB     00:00    
(2/2): python3-dnf-plugins-core-4.3.0-13.el9.noarch.rpm                                                                                                                                                                                         523 kB/s | 264 kB     00:00    
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Total                                                                                                                                                                                                                                           391 kB/s | 301 kB     00:00     
Running transaction check
Transaction check succeeded.
Running transaction test
Transaction test succeeded.
Running transaction
  Preparing        :                                                                                                                                                                                                                                                        1/1 
  Upgrading        : python3-dnf-plugins-core-4.3.0-13.el9.noarch                                                                                                                                                                                                           1/4 
  Upgrading        : dnf-plugins-core-4.3.0-13.el9.noarch                                                                                                                                                                                                                   2/4 
  Cleanup          : dnf-plugins-core-4.3.0-9.el9.noarch                                                                                                                                                                                                                    3/4 
  Cleanup          : python3-dnf-plugins-core-4.3.0-9.el9.noarch                                                                                                                                                                                                            4/4 
  Running scriptlet: python3-dnf-plugins-core-4.3.0-9.el9.noarch                                                                                                                                                                                                            4/4 
  Verifying        : dnf-plugins-core-4.3.0-13.el9.noarch                                                                                                                                                                                                                   1/4 
  Verifying        : dnf-plugins-core-4.3.0-9.el9.noarch                                                                                                                                                                                                                    2/4 
  Verifying        : python3-dnf-plugins-core-4.3.0-13.el9.noarch                                                                                                                                                                                                           3/4 
  Verifying        : python3-dnf-plugins-core-4.3.0-9.el9.noarch                                                                                                                                                                                                            4/4 

Upgraded:
  dnf-plugins-core-4.3.0-13.el9.noarch                                                                                               python3-dnf-plugins-core-4.3.0-13.el9.noarch                                                                                              

Complete!
[root@instance-20240505-192203 ~]#
```

.cirrus.yml
```diff
@@ -301,7 +301,7 @@ report_stagetwo_task:
             dnf -y install python3-pexpect
         fi
     main_script: &stagetwo_test |
-        PYTHONPATH=tests/ avocado run -p TESTLOCAL=true --max-parallel-tasks=1 -t stagetwo tests/{cleaner,collect,report,vendor}_tests
+        PYTHONPATH=tests/:. avocado run -p TESTLOCAL=true --max-parallel-tasks=1 -t stagetwo tests/{cleaner,collect,report,vendor}_tests
```
pmoravec, line 304: Why this change is required, please?

arif-ali: We need `sos` to get the `shell_out` function. It was one of the errors in a previous push

tests/report_tests/plugin_tests/openstack/openstack.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def test_keystone_conf_collected_and_scrubbed(self):
+        self.assertFileCollected('/etc/keystone/keystone.conf')
+        self.assertFileCollected('/etc/keystone/keystone.policy.yaml')
+
+        self.assertFileNotHasContent(
+            '/etc/keystone/keystone.conf', '2zx9jZZtxdn4grG3xcMV4PwgGwY7X7fP')
+
+    def test_keystone_ldap_conf_scrubbed(self):
+
```
pmoravec, line 126: Nitpick: redundant empty line?

arif-ali: Yup, there's probably a few of these, will clean up once I have a full set of tests passing

tests/report_tests/plugin_tests/openstack/openstack.py
```diff
@@ -0,0 +1,162 @@
...
+                        "dnf-plugins-core")
+            shell_out("dnf config-manager --enable crb")
+
+    def setup_mocked_packages(self):
+        this_distro = distro.detect()
+        if this_distro.name == "centos" or this_distro.name == "centos-stream":
+            if this_distro.version == "9":
+                self.osp_release = "yoga"
+            if this_distro.version == "8":
+                self.osp_release = "victoria"
```
pmoravec, line 66: Knowing OSP a little only, but why do we test on unmaintained releases? (https://releases.openstack.org/)

pmoravec:
> After some debugging last night, I was able to get centos stream 9 to work.
> 
> Centos Stream 8 has python 3.6 and avocado-framework 103.0 isn't supported and hence the `distro.detect().version` doesn't work. Using the `<` to install avocado may not have been the right choice. It installs 100.0 for both ubuntu 18.04 and centos stream 8.
> 
> Need to work out a different way to test centos stream 8 for this stagetwo test

I wanted to comment out on this topic (why the centos-8 fails to install OSP packages) but I see you are ahead of me here :).

tests/report_tests/plugin_tests/openstack/openstack.py
arif-ali: Thanks for the link, just made it consistent to be yoga, and enable the right repos for 8/9 and got it through, I think

arif-ali:
@MrStupnikov thanks for your constructive and detailed feedback, it's very much appreciative

* On the heat one, I made a typo, and was going to fix it, and get the scrub test in, hence I made this a draft just before you made the comments.
* This is a first stab at doing this, and surely there would be issues. I am taking your feedback on-board and will have another look at the best way of doing this; if not, I will abandon this PR totally.
* Not all keys are in all configuration files, we need to find where and when these are used, and update the relevant test configs. The plugins have evolved over time, and these keys have also been added over time, so to find every single one of these and which files they occur will take some time.
* A lot of the tests defined here, us at Canonical were doing manually to ensure that these files are being masked correctly, and also being collected as part of our downstream release process. The main thing was to get this automated so that we don't have to do these things manually.

Now that I have overcome the challenges of getting both centos and ubuntu in the same set of tests, I can start adding all the items that are brought forward to me, as well as some that I will be going through personally, and will get these through. Your feedback will be extremely valuable as I go through this process.

tests/report_tests/plugin_tests/openstack/openstack.py
```diff
@@ -0,0 +1,145 @@
...
+            pkgs = json.loads(pkgs)
+            for pkg in pkgs:
+                self.sm.remove(pkg)
+        super().teardown_mocked_packages()
+
+    def test_cinder_conf_collected_and_scrubbed(self):
+        self.assertFileCollected('/etc/cinder/cinder.conf')
+
+        keys_to_mask = [
+            'cmB4zBYq3VWFMNqNKFLcqS5Zq8ystLLsTd5BFLbCtX67qShnhgHFxxRFjkhbY54x',
```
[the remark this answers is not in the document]
arif-ali: I think this is how we are doing other tests with masks as well. We may need to have an overhaul of the tests to do it key based instead of value based. It easier to search for the value instead of the key.

tests/report_tests/plugin_tests/openstack/openstack.py
```diff
@@ -0,0 +1,145 @@
...
+            'wPhFqY69x94YVJc7STrVH3CfsFrrcZPYw8NS2pjhzqyzw7wrL2VnTmN58c5XTnfV',
+            'Ck3r7zf6B6PscfjWhhj2zJdy8SNXYd59',
+        ]
+        for key in keys_to_mask:
+            self.assertFileNotHasContent(
+                '/etc/cinder/cinder.conf', key)
+
+    def test_glance_conf_collected_and_scrubbed(self):
```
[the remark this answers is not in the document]
arif-ali: This is an initial start to get the tests in. I am intending to evolve this moving forward. Having these is better than not having any tests at all. So, for now I am only doing `cinder.conf` and the primary configs at first, and will evolve these over time.

arif-ali:
@MrStupnikov I've made some modifications to the tests, and actually found one key that wasn't already being obfuscated in `heat.conf`. I think you would agree with me here, that having a set of tests is better than having no tests, and we can evolve these moving forwards

arif-ali:
@TurboTurtle @pmoravec @MrStupnikov any further feedback. I know all the keys are not there, or not all files, but that would be a gradual thing that I will do eventually, getting the first bits to help move this forward

pmoravec (approved):
ACK, just having a curious question.

tests/sos_tests.py
```diff
@@ -847,8 +861,8 @@ def tearDown(self):
     def teardown_mocking(self):
         """Undo any and all mocked setup that we did for tests
         """
-        self.teardown_mocked_packages()
         self.teardown_mocked_files()
+        self.teardown_mocked_packages()
```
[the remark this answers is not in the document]
arif-ali: When you mock both files and packages, and the packages have the same files; the `tear_down_packages()` will delete the file that was mocked, and hence the `teardown_mocked_files()` couldn't tear down and wouldn't be able to find the file. It was one of the issues I had faced while I was working on this particular PR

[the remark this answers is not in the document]
pmoravec: Makes sense; thanks for explanation!

TurboTurtle (commented):
The code LGTM, but I'm still not clear on why we're not doing an actual install in our CI like we do for Foreman. We're going through the effort of installing some release package, but still dropping manual files on the test host.

This seems a little bit of a conflict here - if we're installing a release package, why don't we execute a "normal" setup so that we don't _have_ to drop the manual files, and are instead testing against what a real, actual install would look like after an installation (so that our test system would automatically reflect changes that go into the project)?

arif-ali:
> The code LGTM, but I'm still not clear on why we're not doing an actual install in our CI like we do for Foreman. We're going through the effort of installing some release package, but still dropping manual files on the test host.
> 
> This seems a little bit of a conflict here - if we're installing a release package, why don't we execute a "normal" setup so that we don't _have_ to drop the manual files, and are instead testing against what a real, actual install would look like after an installation (so that our test system would automatically reflect changes that go into the project)?



> The code LGTM, but I'm still not clear on why we're not doing an actual install in our CI like we do for Foreman. We're going through the effort of installing some release package, but still dropping manual files on the test host.
> 
> This seems a little bit of a conflict here - if we're installing a release package, why don't we execute a "normal" setup so that we don't _have_ to drop the manual files, and are instead testing against what a real, actual install would look like after an installation (so that our test system would automatically reflect changes that go into the project)?

Are you suggesting that we do an all-in-one openstack deployment on a heavy machine/VM in GCE, and that has all the components and then do the tests against that?

TurboTurtle:
> > The code LGTM, but I'm still not clear on why we're not doing an actual install in our CI like we do for Foreman. We're going through the effort of installing some release package, but still dropping manual files on the test host.
> > This seems a little bit of a conflict here - if we're installing a release package, why don't we execute a "normal" setup so that we don't _have_ to drop the manual files, and are instead testing against what a real, actual install would look like after an installation (so that our test system would automatically reflect changes that go into the project)?
> 
> Are you suggesting that we do an all-in-one openstack deployment on a heavy machine/VM in GCE, and that has all the components and then do the tests against that?

I guess? We do a similar thing with Foreman, where we take a larger GCE instance and do a scripted setup.

Granted, OSP has a heavier footprint than Foreman so I'm not sure if that's feasible as a test bed.

arif-ali:
> I guess? We do a similar thing with Foreman, where we take a larger GCE instance and do a scripted setup.
> 
> Granted, OSP has a heavier footprint than Foreman so I'm not sure if that's feasible as a test bed.

OK, makes sense, I have an idea to potentially use `devstack`, leave it with me, and we do it that way

pmoravec:
> Granted, OSP has a heavier footprint than Foreman so I'm not sure if that's feasible as a test bed.

Good point. I am not against having OSP instances, but as a kind of assesment, we should answer below questions:

How big such instance can be (compared to Foreman one)?

How much it would increase cost of testing / payment for GCE (just rough estimate, if possible)?

How much it would slow down the whole testing (probably not much) - esp. relevant to "what all changes in PR can trigger testing on the OSP instance" question (cf. https://github.com/sosreport/sos/blob/main/.cirrus.yml#L317 as an example)?

arif-ali:
> > Granted, OSP has a heavier footprint than Foreman so I'm not sure if that's feasible as a test bed.
> 
> Good point. I am not against having OSP instances, but as a kind of assesment, we should answer below questions:
> 
> How big such instance can be (compared to Foreman one)?
> 
> How much it would increase cost of testing / payment for GCE (just rough estimate, if possible)?
> 
> How much it would slow down the whole testing (probably not much) - esp. relevant to "what all changes in PR can trigger testing on the OSP instance" question (cf. https://github.com/sosreport/sos/blob/main/.cirrus.yml#L317 as an example)?

I've closed this particular issue, as the new method will look completely different.

I tried `devstack` last night; and we need at least 8G RAM machine, otherwise we'll get oom-killer kicking in, and devstack not completing. The whole installation bit takes 30 mins. This is where I got to.

I've tried running sosreport on that, as is, it won't trigger as the packages wouldn't have installed. The config files are in the right place, so we can assert files collected and check the obfuscation.

Alternatively, I can try `packstack`, which I haven't used for 10 years, so would be a learning cure.

Ubuntu doesn't have an all-in-one package per say; and even then it would be sunbeam; and that would require a high mem VM in GCE and would take some time to automate; but that would not replicate the same thing based on Canonical OpenStack right now.

The only other way to do this is, is to manually deploy everything with all the configs using a script; and that would require a significant development time.

If this is the preferred route, then we keep this PR closed; and I will work on an alternative, which may take a few months to get done

pmoravec:
Afaik we mostly use 4GB machines (`e2-medium`) and 8GB one for foreman tests (`e2-standard-2`). So having `e2-standard-2` - or `e2-highmem-2` (16GB RAM) when required - is feasible.

Manual deply would require a lot of.. well, manual work, better to use either approach from this PR or an image one. Either that sounds good to me.

PR #3623, opened 2024-04-25
[pulp] Including settings.local.py file to be collected and parsed.

rbertol:
This change, allows the sosreport to collect the
settings.local.py file, when customer has custom changes out of the default settings.py file.

---

### DIFF
--- a/sos/report/plugins/pulp.py
+++ b/sos/report/plugins/pulp.py
@@ -72,6 +72,7 @@ class Pulp(Plugin, RedHatPlugin):
         self.add_copy_spec([
             "/etc/pulp/*.conf",
             "/etc/pulp/settings.py",
+            "/etc/pulp/settings.local.py",
             "/etc/pulp/server/plugins.conf.d/",
             "/etc/default/pulp*",
             "/var/log/httpd/pulp-http.log*",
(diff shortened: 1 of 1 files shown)


arif-ali (approved):

sos/report/plugins/pulp.py
```diff
@@ -72,6 +72,7 @@ def setup(self):
         self.add_copy_spec([
             "/etc/pulp/*.conf",
             "/etc/pulp/settings.py",
+            "/etc/pulp/settings.local.py",
```
TurboTurtle, line 75: Do we need to add this to the `postproc()` handling of the "regular" `settings.py` file?

https://github.com/sosreport/sos/blob/f580ee0719b298d3358fc628a7baa690373a5b31/sos/report/plugins/pulp.py#L182

rbertol: By default, customer should not include passwords. But will not warm redact it, in case of any sensitive info.

rbertol:
Also, removed the settings.loca.py from the file section, as this file will not be always present on every installation.

arif-ali (changes requested):
Can you fix the flake8 issue please

arif-ali (approved):

TurboTurtle (approved):

PR #3626, opened 2024-04-27
[kubernetes] scrub collected files

champtar:
We were only scrubbing command outputs.

Code copied from openshift plugin.

---

### DIFF
--- a/sos/report/plugins/kubernetes.py
+++ b/sos/report/plugins/kubernetes.py
@@ -247,6 +247,21 @@ class Kubernetes(Plugin):
         # output that is not hit by the previous iteration.
         self.do_cmd_private_sub(self.kube_cmd)
 
+        pathregex = fr'^({"|".join(self.config_files)})'
+        self.do_file_private_sub(pathregex)
+
+        # clear the certificate data from files that does not have
+        # the certificate banners that the _private_sub() methods look for
+        _fields = [
+            'client-certificate-data',
+            'client-key-data',
+            'certificate-authority-data',
+        ]
+
+        regex = fr'(\s*({"|".join(_fields)}):)(.*)'
+
+        self.do_path_regex_sub(pathregex, regex, r'\1 *******')
+
 
 class RedHatKubernetes(Kubernetes, RedHatPlugin):
 


sos/report/plugins/kubernetes.py
```diff
@@ -247,6 +247,24 @@ def postproc(self):
         # output that is not hit by the previous iteration.
         self.do_cmd_private_sub(self.kube_cmd)
 
+        self.do_file_private_sub('/*')
```
TurboTurtle, line 250: Absolutely not.

sos/report/plugins/kubernetes.py
```diff
@@ -247,6 +247,24 @@ def postproc(self):
...
+            '.*.crt',
+            'client-certificate-data',
+            'client-key-data',
+            'certificate-authority-data',
+            '.*.key',
+            'token',
+            '.*token.*.value'  # don't blind match `.*token.*` and lose names
+        ]
+
+        regex = r'(\s*(%s):)(.*)' % '|'.join(_fields)
```
TurboTurtle, line 264: Please make this an f-string instead of using string substitution.

sos/report/plugins/kubernetes.py
```diff
@@ -247,6 +247,24 @@ def postproc(self):
...
+
+        self.do_path_regex_sub('/*', regex, r'\1 *******')
```
TurboTurtle, line 266: Post processing needs to be done very intentionally on the files the obfuscation is built around.

TurboTurtle (changes requested):

sos/report/plugins/kubernetes.py
champtar: This only scrub files copied by this plugin, listing all the RHEL/Ubuntu current paths only open the chance to miss some new paths in the future.
Also if we don't scrub everything, a change in k8s packaging/configuration might make it so that we include a non scrubbed file without any sos changes.

BTW I don't see why `do_file_private_sub` is not done on all files by default, with an option to opt out.

sos/report/plugins/kubernetes.py
champtar: done

sos/report/plugins/kubernetes.py
champtar: now limited to `config_files`, scrubbing  kubeconfig style

sos/report/plugins/kubernetes.py
champtar: now limited to `config_files`

champtar:
closing in favor of #3630

PR #3630, opened 2024-04-29
[kubernetes] plugin improvements

### LINKED ISSUE #3628: [RFE] collect kubernetes informations for kubeadm installs
Currently `RedHatKubernetes` is specific to OpenShift (only older versions ?)
I would like to have it run for classic kubeadm installs on top of RHEL family OS (also Ubuntu as a bonus),
but I don't have access to OpenShift to move the OpenShift bits in openshift.py and to make sure either openshift plugin runs or kubernetes plugin runs, but not both (a bit of a waste to collect kubelet logs twice)

To collect kubeadm installs we would need
```
files = (
    "/etc/kubernetes/admin.conf",
)
services = (
    "kubelet",
)
```

Would something like
```
def check_enabled(self):
    return not self.path_exists('/etc/origin/') and super().check_enabled()
```
be ok ?

champtar:
Rework kubernetes plugin to make it collect in more cases
This PR can be split in multiple ones if prefered

Resolves #3628 
Supersedes #3626 

---

### DIFF
--- a/sos/report/plugins/kubernetes.py
+++ b/sos/report/plugins/kubernetes.py
@@ -12,8 +12,28 @@
 from fnmatch import translate
 import re
 import json
+import os
 from sos.report.plugins import (Plugin, RedHatPlugin, DebianPlugin,
                                 UbuntuPlugin, PluginOpt)
+from sos.utilities import is_executable
+
+
+KUBE_PACKAGES = (
+    'kubelet',
+    'kubernetes',
+)
+
+KUBE_SVCS = (
+    'kubelet',
+    'kube-apiserver',
+    'kube-proxy',
+    'kube-scheduler',
+    'kube-controller-manager',
+)
+
+KUBECONFIGS = (
+    '/etc/kubernetes/admin.conf',
+)
 
 
 class Kubernetes(Plugin):
@@ -45,7 +65,9 @@ class Kubernetes(Plugin):
         'jobs',
         'cronjobs',
         'clusterroles',
-        'clusterrolebindings'
+        'clusterrolebindings',
+        'limitranges',
+        'resourcequotas',
     ]
 
     # these are not namespaced, must pull separately.
@@ -53,7 +75,7 @@ class Kubernetes(Plugin):
         'sc',
         'pv',
         'roles',
-        'rolebindings'
+        'rolebindings',
     ]
 
     option_list = [
@@ -69,6 +91,14 @@ class Kubernetes(Plugin):
 
     kube_cmd = "kubectl"
 
+    def set_kubeconfig(self):
+        if os.environ.get('KUBECONFIG'):
+            return
+        for _kconf in self.files:
+            if self.path_exists(_kconf):
+                self.kube_cmd += f" --kubeconfig={_kconf}"
+                break
+
     def check_is_master(self):
         """ Check if this is the master node """
         return any(self.path_exists(f) for f in self.files)
@@ -80,24 +110,8 @@ class Kubernetes(Plugin):
             'KUBECONFIG',
             'KUBERNETES_HTTP_PROXY',
             'KUBERNETES_HTTPS_PROXY',
-            'KUBERNETES_NO_PROXY'
+            'KUBERNETES_NO_PROXY',
         ])
-
-        svcs = [
-            'kubelet',
-            'kube-apiserver',
-            'kube-proxy',
-            'kube-scheduler',
-            'kube-controller-manager',
-            'snap.kubelet.daemon',
-            'snap.kube-apiserver.daemon',
-            'snap.kube-proxy.daemon',
-            'snap.kube-scheduler.daemon',
-            'snap.kube-controller-manager.daemon'
-        ]
-
-        for svc in svcs:
-            self.add_journal(units=svc)
 
         # We can only grab kubectl output from the master
         if not self.check_is_master():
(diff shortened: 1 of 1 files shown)


arif-ali:
I'll have a look at testing this for before and after, as we don't want to lose anything, especially from our perspective from a few scenarios such as `microk8s` and `sunbeam`

In the meantime can you squash your commits based on our contribution guidelines

champtar:
@arif-ali I can't find any mention of squashing in https://github.com/sosreport/sos/wiki/Contribution-Guidelines

arif-ali:
> @arif-ali I can't find any mention of squashing in https://github.com/sosreport/sos/wiki/Contribution-Guidelines

you're right, it's not there, we ought to as that is what we ask all our contributors to do

below is the common theme on this in our repo

> we don't allow fixup commits when merging (keeps the git history cleaner and much easier to read).

In the mean time, I will have a look at updating the guidelines

champtar:
I'll squash, but to be clear there are no fixup commits, I'm just implementing the changes in small increments, making review easier IMO

sos/report/plugins/kubernetes.py
```diff
@@ -247,62 +264,76 @@ def postproc(self):
...
+        self.do_file_private_sub(pathexp_config_files)
+
+        # clear base64 encoded PEM from kubeconfigs files
+        regexp = r'LS0tLS1CRUdJ[A-Za-z0-9+/=]+'
+        subst = '***** SCRUBBED BASE64 PEM *****'
+
+        self.do_path_regex_sub(pathexp_config_files, regexp, subst)
+
+        pathexp_files = fr'^({"|".join(self.files)})'
+        self.do_path_regex_sub(pathexp_files, regexp, subst)
```
pmoravec, line 277: Can§t we merge this into one `do_path_regexp_sub`? Or would it be too confused?

sos/report/plugins/kubernetes.py
```diff
@@ -247,62 +264,76 @@ def postproc(self):
...
+    files = KUBECONFIGS + (
         '/etc/origin/master/admin.kubeconfig',
         '/etc/origin/node/pods/master-config.yaml',
     )
 
-    kube_cmd = "kubectl"
+    services = KUBE_SVCS
+
+    def check_enabled(self):
+        if self.is_installed("openshift-hyperkube"):
```
pmoravec, line 295: Why this package presence shoud disable this plugin? (I guess there is a rationale that I dont see / know)

sos/report/plugins/kubernetes.py
```diff
@@ -247,62 +264,76 @@ def postproc(self):
...
 
     def setup(self):
-        # Rather than loading the config file, use the OCP command directly
-        # that wraps kubectl, so we don't have to manually account for any
-        # other changes the `oc` binary may implement
-        if self.path_exists('/etc/origin/master/admin.kubeconfig'):
+        self.set_kubeconfig()
+
+        # if present, use `oc` command and add some OCP specific ressources
+        if is_executable('oc'):
```
pmoravec, line 303: What if `set_cubectl` sets `self.kube_cmd = "kubectl --kubeconfig={_kconf}"` and now we reset that option? Is it expected (i.e. `oc` does not honour `--kubeconfig`) or shouldnt we call `self.set_kubeconfig()` after this if-block?

sos/report/plugins/kubernetes.py
champtar: For what it's worth the commit title is `[kubernetes] ensure we are not enabled if Openshift is`
The idea is that we match on a lot of things that might also be present on OpenShift installs, but running both plugins is wasteful
I'll add a comment

sos/report/plugins/kubernetes.py
champtar: it's on purpose, I don't want to break the old behavior for old openshift installs.
as I don't have any openshift installs to look at how files are organized I just play it safe here

sos/report/plugins/kubernetes.py
champtar: done

sos/report/plugins/kubernetes.py
champtar: done

champtar:
@arif-ali commits are now squashed

champtar:
@arif-ali any bandwidth to do the tests you wanted to do ?

arif-ali:
> @arif-ali any bandwidth to do the tests you wanted to do ?

I was due to do them today, but other things came in the way. As soon as I've done them will update here

sos/report/plugins/kubernetes.py
```diff
@@ -46,6 +66,8 @@ class Kubernetes(Plugin):
         'cronjobs',
         'clusterroles',
         'clusterrolebindings'
```
arif-ali, line 68: There's a comma missing here

sos/report/plugins/kubernetes.py
```diff
@@ -83,21 +113,9 @@ def setup(self):
...
-            'snap.kube-apiserver.daemon',
-            'snap.kube-proxy.daemon',
-            'snap.kube-scheduler.daemon',
-            'snap.kube-controller-manager.daemon'
-        ]
-
-        for svc in svcs:
-            self.add_journal(units=svc)
+        for svc in self.services:
+            if self.is_service_enabled(svc):
```
arif-ali, line 117: This is a change from before. Now if the service is not enabled, or for some reason someone has accidentally disabled the service, but we may still have journal logs, I think we should not have this check.

arif-ali (changes requested):
one change required, and a comment maybe others will chime in on that

sos/report/plugins/kubernetes.py
champtar: would this be ok ? (that's what I really wanted actually) It's a bit confusing to have empty log files for services that just don't exists
```suggestion
            if self.is_service(svc):
```

sos/report/plugins/kubernetes.py
champtar: fixed

sos/report/plugins/kubernetes.py
arif-ali: I know it's big **if**, what if the package is removed, and we still have the journals there to consume? 🤔

champtar: The plugin might not even start to collect if enough stuff are deleted, and recent logs will be collected by the logs plugin.
I'll follow whatever you prefer, but this is a weird corner case to account for IMO

arif-ali (approved):

champtar:
@pmoravec any other concerns ?

pmoravec (approved):

champtar:
@TurboTurtle when you get a chance please review

sos/report/plugins/kubernetes.py
```diff
@@ -80,24 +110,12 @@ def setup(self):
...
-            'snap.kube-proxy.daemon',
-            'snap.kube-scheduler.daemon',
-            'snap.kube-controller-manager.daemon'
-        ]
-
-        for svc in svcs:
-            self.add_journal(units=svc)
+        for svc in self.services:
+            if self.is_service(svc):
+                self.add_journal(units=svc)
```
TurboTurtle, line 118: This is redundant - we do this automatically in `Plugin.add_default_collections()`

TurboTurtle (commented):
Looks good to me, pending the redundant journal collection.

I _think_ we should be good to remove the OCP bits from this plugin entirely, as I do not believe RH supports any of those versions anymore, but I'm good with leaving that for another PR later on.

sos/report/plugins/kubernetes.py
champtar: removed

champtar:
I'll do a follow up PR to remove the OCP bits, but would love to have this PR in the upcoming release, so I can roll it sooner rather than later

TurboTurtle (approved):

TurboTurtle:
Next upstream release is in August, yes, which will be `4.8.0`.

Red Hat used to have a faster release cadence, then didn't, but then picked it up again with `4.7.1`, but we've not seen another z-stream release since then. @pmoravec?

PR #3633, opened 2024-05-01
[aap_gateway] Implementing AAP Gateway Plugin 

lucas-benedito:
Adding the plugin 'aap_gateway.py' for the sos report to collect
the files used for troubleshooting issues at 
Ansible Automation Platform Gateway

Related: RH AAP-20033

---

### DIFF
--- a/sos/report/plugins/aap_gateway.py
+++ b/sos/report/plugins/aap_gateway.py
@@ -1,0 +1,53 @@
+# Copyright (c) 2024 Lucas Benedito <lbenedit@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, RedHatPlugin
+
+
+class AAPGatewayPlugin(Plugin, RedHatPlugin):
+    short_desc = 'AAP Gateway plugin'
+    plugin_name = 'aap_gateway'
+    profiles = ('sysmgmt', 'ansible',)
+
+    packages = ('automation-gateway',
+                'automation-gateway-config')
+    commands = ('aap-gateway-manage',)
+    services = ('automation-gateway',)
+
+    def setup(self):
+        self.add_copy_spec([
+            "/var/log/supervisor",
+            "/etc/ansible-automation-platform",
+            "/etc/supervisord.d/",
+            "/var/log/ansible-automation-platform/gateway/",
+        ])
+
+        self.add_forbidden_path([
+            "/etc/ansible-automation-platform/gateway/SECRET_KEY",
+            "/etc/ansible-automation-platform/gateway/*.key",
+            "/etc/ansible-automation-platform/gateway/*.cert",
+        ])
+
+        self.add_cmd_output([
+            "aap-gateway-manage list_services",
+            "ls -ll /etc/ansible-automation-platform/",
+            "ls -ll /etc/ansible-automation-platform/gateway/",
+        ])
+
+    def postproc(self):
+        # remove database password
+        jreg = r"(DATABASE_PASSWORD)(\s*)(=|:)(\s*)(.*)"
+        repl = r"\1\2\3\4********"
+        self.do_path_regex_sub(
+            "/etc/ansible-automation-platform/gateway/settings.py",
+            jreg,
+            repl)
+
+# vim: set et ts=4 sw=4 :


sos/report/plugins/aap_gateway.py
```diff
@@ -0,0 +1,56 @@
...
+
+    packages = ('automation-gateway',
+                'automation-gateway-config')
+    commands = ('aap-gateway-manage',)
+    services = ('automation-gateway',)
+
+    def setup(self):
+        self.add_copy_spec([
+            "/var/log/nginx/access.log*",
+            "/var/log/nginx/error.log*",
```
arif-ali, line 27: The `nginx` stuff is already being collected by the `nginx` plugin, if that would be triggered as well, then you could have 2 plugins collect the same thing

sos/report/plugins/aap_gateway.py
```diff
@@ -0,0 +1,56 @@
...
+            "/var/log/supervisor",
+            "/etc/ansible-automation-platform",
```
arif-ali, line 29: We're collecting this folder, but we already  the `AAPEDAControllerPlugin` collecting some of this as well, would there not be conflict here?

maybe, in the same way have `/etc/ansible-automation-platform/gateway`? if that is the main thing that is relevant for this plugin

sos/report/plugins/aap_gateway.py
pmoravec: Indeed, we should avoid that duplicate collection. Since `ngix` plugin collects just non-logrotated files (https://github.com/sosreport/sos/blob/main/sos/report/plugins/nginx.py#L33-L34), I think the best way is to enhance `ngix` plugin to collect `access.log*` and `error.log*`?

sos/report/plugins/aap_gateway.py
```diff
@@ -0,0 +1,56 @@
...
+            "/etc/ansible-automation-platform/gateway/SECRET_KEY",
+            "/etc/ansible-automation-platform/gateway/*.key",
+            "/etc/ansible-automation-platform/gateway/*.cert",
+        ])
+
+        self.add_cmd_output([
+            "aap-gateway-manage list_services",
+            "ls -ll /etc/ansible-automation-platform/",
+            "ls -ll /etc/ansible-automation-platform/gateway/",
+            "umask -p",
```
pmoravec, line 44: We collect it in AAP_Controller already (https://github.com/sosreport/sos/blob/0a876d47196bf9738b4d14509b593583f59c7e3a/sos/report/plugins/aap_controller.py#L63). It might have sense to collect same command output by two plugins (less rationale against that that colecting same files), but still.. does not it make sense to have that collected on one place?

It might be beneficial to have it collected more "generically", but I cant find a proper plugin fir it. Maybe `host`?

@TurboTurtle , @arif-ali , any idea where to collect `umask -p`?

arif-ali: I don't see the `umask -p` command, but I do `umask -S`, could this be RH specific? i would lean towards `filesys` maybe?

sos/report/plugins/aap_gateway.py
lucas-benedito: The configuration made here will collect rotated logs. I am not sure if for other teams these logs are as relevant as they are for us, being the reason I did not suggest the change on the main nginx plugin.

sos/report/plugins/aap_gateway.py
lucas-benedito: Based on the packages, commands and services, the aap_gateway and aap_controller would not be executed in the same host. Those components only run on different nodes.
With that said, I can remove the umask from both plugins if we think there is a better place for it.

sos/report/plugins/aap_gateway.py
arif-ali: I don't think, this is about teams or anything like that. This could be useful for other people as well, and all the `nginx` stuff should go into the `nginx` plugin. If we have an environment that is not `aap_gateway` but the system has `nginx`, then it would be useful for that scenario. So +1 on what @pmoravec said above

sos/report/plugins/aap_gateway.py
lucas-benedito: Based on the packages, commands, and services, the aap_gateway and aap_controller would not be executed in the same host. Those components only run on different nodes. With that, there wouldn't be conflicts
For the Controller plugin this path rarely contains data, while for the Gateway plugin, this is the primary place where things are logged.

sos/report/plugins/aap_gateway.py
pmoravec: `filesys` seems a good plugin for it. +1 for removing `umask -p` from both plugins and adding it there.

Since Ubuntu does not know `-p` flag and that flag just prepends `umask ` word (such that the output string is a valid command to set umask) and that it does not bring any additional value, let add plain `umask` there.

sos/report/plugins/aap_gateway.py
pmoravec: Sounds good to me, then.

arif-ali: ack, sounds good to me

sos/report/plugins/aap_gateway.py
pmoravec: Thanks! I think `nginx` plugin will also much rather benefit from the change (of additionaly collecting logrotated logs) than harm from it (more data on slower sosreports done).

arif-ali (approved):

pmoravec (approved):

PR #3635, opened 2024-05-06
[policies] Add AlmaLinux policy

eabdullin1:
AlmaLinux OS is an Enterprise Linux distro, binary compatible with RHEL. This commit adds a AlmaLinux policy, which inherits the RedHatPolicy base.

Signed-off-by: Eduard Abdullin <eabdullin@almalinux.org>

---

### DIFF
--- a/sos/policies/distros/almalinux.py
+++ b/sos/policies/distros/almalinux.py
@@ -1,0 +1,46 @@
+# Copyright (C) Eduard Abdullin <eabdullin@almalinux.org>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+import os
+from sos.policies.distros.redhat import RedHatPolicy, OS_RELEASE
+
+
+class AlmaLinuxPolicy(RedHatPolicy):
+    distro = "AlmaLinux"
+    vendor = "AlmaLinux OS Foundation"
+    vendor_urls = [
+        ('Distribution Website', 'https://www.almalinux.org/'),
+        ('Commercial Support', 'https://tuxcare.com/linux-support-services/')
+    ]
+
+    def __init__(self, sysroot=None, init=None, probe_runtime=True,
+                 remote_exec=None):
+        super().__init__(sysroot=sysroot, init=init,
+                         probe_runtime=probe_runtime,
+                         remote_exec=remote_exec)
+
+    @classmethod
+    def check(cls, remote=''):
+        if remote:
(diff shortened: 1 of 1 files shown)


pmoravec:
We can surely add a new policy, but the change as is will not alter `sos` behaviour at all.

Do you plan to add some child `Plugin` classes that will alter `sos` behaviour from `RedHatPolicy`?

eabdullin1:
> We can surely add a new policy, but the change as is will not alter `sos` behaviour at all.
> 
> Do you plan to add some child `Plugin` classes that will alter `sos` behaviour from `RedHatPolicy`?

No. We don't plan to add child classes.

TurboTurtle:
Does Alma plan on adding upload support to this policy at some point, so that users can specify for the `--upload` option and have the generated report automatically uploaded to a technical support case?

We typically like to see so kind of use-case behind policy additions; either plugins that have different behavior under different distros, upload support, a different package manager, some clustering mechanism for `sos collect` that is used with the new policy/distro, etc...

TurboTurtle:
Drive by thought: we've long been wanting to directly test RHEL in our CI here but we run into the issue with cloud providers who by default charge per-instance for starting RHEL (which is doubly awkward given that RH is who sponsors our GCE costs). Does using Alma in CI make sense as a suitable stand in?

When I left RH there were some nebulous talks about how to get our GCE project setup for a "bring your own subscription" posture ("Cloud Access" for AWS, I forget what it's called for GCP) that would allow us to use official RHEL images without getting billed for the RHEL sub through GCE but I don't know if those continued at all?

jcastill:
One thing to bear in mind - Alma is picking up architectures and hardware that we are dropping/deprecating in RHEL, so we could end up with a situation where some tool or command does not exist in RHEL anymore but may be useful to capture for Alma. Would it make sense in this scenario to have an Alma policy? or would it be captured nonetheless?

About having RHEL images in GCP, if it's ok with you I'll do some research internally and see what's the current situation. If I cannot find a solution then we can check Alma images.

arif-ali:
Another thought: we also have a Rocky policy already, and that could be an alternative as well and they also have cloud images on GCE so that's another route we could take for testing

pmoravec:
> Another thought: we also have a Rocky policy already, and that could be an alternative as well and they also have cloud images on GCE so that's another route we could take for testing

What would be the difference from using CentOS and Rocky images / what would be the benefit of testing sos on Rocky as well (or instead of CentOS?)? (I think we shall raise a separate discussion thread on :what images are worth to test sos on")

TurboTurtle (approved):
The policy LGTM as is, however @eabdullin1 I'd be interested to know if the case upload support I mentioned is of any interest/value to Alma.

With regards to the testing conversation, I'll open a new GH discussion for it so that we don't derail this PR.

eabdullin1:
No at the moment

PR #3636, opened 2024-05-07
[Plugins, Options] Standardized PluginOpt names to use dashes

TurboTurtle:
This commit standardizes the use of dashes/hypens (`-`) for the names of PluginOpts. This is done to conform with argparse formatting so that the formating of options names is consistent across sos global options, component options, and now plugin options.

For example, the `networking` plugin's `namespace_pattern` option is now `namespace-pattern` and should now be specified via the `--plugin-option networking.namespace-pattern=foo` syntax. Similarly, developers will now need to use `Plugin.get_option('namespace-pattern')` in order to reference it.

Note that the use of global options within plugins (for example `--all-logs`) is unchanged, meaning that users will continue to use `--all-logs` and due to how these options get stored, developers would still use `Plugin.get_option('all_logs')`.

---

### DIFF
--- a/sos/presets/redhat/__init__.py
+++ b/sos/presets/redhat/__init__.py
@@ -29,7 +29,7 @@ RHOSP = "rhosp"
 RHOSP_DESC = "Red Hat OpenStack Platform"
 RHOSP_OPTS = SoSOptions(plugopts=[
                              'process.lsof=off',
-                             'networking.ethtool_namespaces=False',
+                             'networking.ethtool-namespaces=False',
                              'networking.namespaces=200'])
 
 RHOCP = "ocp"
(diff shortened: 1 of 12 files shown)


TurboTurtle:
As mentioned, this is done purely to create a consistent CLI experience for users, and I landed on converting to dashes from underscores rather than the other way around simply because our "top level" CLI options that are handled by argparse are exclusively dashes.

Happy to hear any arguments for moving the other way, though that would be a much larger shift.

The handling of global options that affect plugins (e.g. `--all-logs`) that get converted to underscores internally by necessity of the Python language when trying to get their value is the schism here for developers. We _could_ add a conversion in `get_option()` that allows one to use `get_option('all-logs')` but since it hasn't been an issue for devels up to this point, I didn't go forward with that (again, this change is mainly focused on end user usage).

arif-ali (approved):
I'm on par with that, make all the things consistent

TurboTurtle:
Note to self: after merging, update the contributions guideline to include a line about this.

pmoravec (approved):
Ack, though we shall (how?) warn users about the plugin opts syntax change.

Cc @mhradile for awareness.

PR #3644, opened 2024-05-16
[pylint] Refactor pylint configuration

arif-ali:
Also fix tox issues with posargs and further fine tuning

---

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -1,0 +1,97 @@
+[MAIN]
+# Use multiple processes to speed up Pylint. Specifying 0 will auto-detect the
+# number of processors available to use.
+jobs=0
+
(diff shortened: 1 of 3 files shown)


arif-ali:
The refactor of the `pylint` is done in this way to ensure that we enable
everything and only disable what we think we should disable. This was after some
discussions with one of our Engineering leads and was their main advice.

The disable section of the `pylintrc` is ordered in a way that the less amount
of work to accomplish is near the bottom. This also helps, if anyone wants to
contribute, and hence can work on a `pylint` module one at a time. I have grouped
the modules in 2, the top, that I think we should disable anyway and the bottom
section that we ought to fix. We may decide that some of these we can ignore.

I am hoping this will help with improved code and consistency across the board.

TurboTurtle (approved):

PR #3645, opened 2024-05-17
[ceph_mds] Capture alternative mds file path

jcastill:
Newer versions of Ceph place information in
/var/lib/ceph/<fsid>/mds* instead of
/var/lib/ceph/mds/*, so we need to capture
the new path as well.

Related: RH: SUPDEV-154

---

### DIFF
--- a/sos/report/plugins/ceph_mds.py
+++ b/sos/report/plugins/ceph_mds.py
@@ -16,7 +16,8 @@ class CephMDS(Plugin, RedHatPlugin, UbuntuPlugin):
     plugin_name = 'ceph_mds'
     profiles = ('storage', 'virt', 'container', 'ceph')
     containers = ('ceph-(.*-)?fs.*',)
-    files = ('/var/lib/ceph/mds/*', '/var/snap/microceph/common/data/mds/*')
+    files = ('/var/lib/ceph/mds/*', '/var/lib/ceph/*/mds.*',
+             '/var/snap/microceph/common/data/mds/*')
 
     def setup(self):
         all_logs = self.get_option("all_logs")
@@ -48,6 +49,7 @@ class CephMDS(Plugin, RedHatPlugin, UbuntuPlugin):
             self.add_copy_spec([
                 "/var/lib/ceph/bootstrap-mds/",
                 "/var/lib/ceph/mds/",
+                "/var/lib/ceph/*/mds.*",
                 "/run/ceph/ceph-mds*",
             ])
 


pmoravec (approved):

arif-ali:
@jcastill is it not worth adding to add_copy_spec below as well?

jcastill:
> @jcastill is it not worth adding to add_copy_spec below as well?

oh wow, I completely missed that. I'll add it and push again, thank you @arif-ali !

arif-ali:
> > @jcastill is it not worth adding to add_copy_spec below as well?
> 
> oh wow, I completely missed that. I'll add it and push again, thank you @arif-ali !

alternatively, we could just call `super().setup()`  and remove the copy_spec for those specific files, as that already this. I'm not sure why we're not already doing that, when the `Plugin` class can do this for us already. Something to think about on other plugins too

pmoravec:
> > > @jcastill is it not worth adding to add_copy_spec below as well?
> > 
> > 
> > oh wow, I completely missed that. I'll add it and push again, thank you @arif-ali !
> 
> alternatively, we could just call `super().setup()` at the end and remove the `copy_spec` for those specific files, as that already does this. I'm not sure why we're not already doing that, when the `Plugin` class can do this for us already. Something to think about on other plugins too

I was under the impression this is done automatically..? It imho makes sense to collect `files` by default, no?

arif-ali:
> > alternatively, we could just call `super().setup()` at the end and remove the `copy_spec` for those specific files, as that already does this. I'm not sure why we're not already doing that, when the `Plugin` class can do this for us already. Something to think about on other plugins too
> 
> I was under the impression this is done automatically..? It imho makes sense to collect `files` by default, no?

That was my original impression too, but the only place in the `Plugin` class we are is in the `setup()` function. Maybe we need some improvements here. I did a quick search for `self.files` in `__init__py`, we only find one scenario where we are collecting from that dict, and that is in the `setup()`, but because we always override it, we don't collect anything from the dict

arif-ali:
@pmoravec you have proved me wrong, just tested the same for the cron plugin, which only has `/etc/crontab` in `files` but nowhere else. So, this looks good to me :)

arif-ali (approved):

pmoravec:
> @pmoravec you have proved me wrong, just tested the same for the cron plugin, which only has `/etc/crontab` in `files` but nowhere else. So, this looks good to me :)

But we `add_copy_spec` the glob `/etc/cron*` - when I comment it out, `/etc/crontab` stops to be collected.

[add_default_collections](https://github.com/sosreport/sos/blob/main/sos/report/plugins/__init__.py#L3275) method does (all?) such automatic collection from plugins triggers.

Since we touched a topic we are not certain about proper behaviour, I kicked off discussion on it in https://github.com/orgs/sosreport/discussions/3647 .

pmoravec (approved):

arif-ali:
> > @pmoravec you have proved me wrong, just tested the same for the cron plugin, which only has `/etc/crontab` in `files` but nowhere else. So, this looks good to me :)
> 
> But we `add_copy_spec` the glob `/etc/cron*` - when I comment it out, `/etc/crontab` stops to be collected.

Dunno why, but I overlooked that bit  :), good catch

arif-ali (approved):

PR #3646, opened 2024-05-19
[sunbeam] Update plugin to collet more details

arif-ali:
* Add juju related info for sunbeam
* Add logs from the sunbeam snap home user

Resolves: SET-682

---

### DIFF
--- a/sos/report/plugins/sunbeam.py
+++ b/sos/report/plugins/sunbeam.py
@@ -1,3 +1,5 @@
+# Copyright (C) 2024 Canonical Ltd., Arif Ali <arif.ali@canonical.com>
+#
 # This file is part of the sos project: https://github.com/sosreport/sos
 #
 # This copyrighted material is made available to anyone wishing to use,
@@ -6,7 +8,9 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.report.plugins import Plugin, UbuntuPlugin
+import json
+import pwd
+from sos.report.plugins import Plugin, UbuntuPlugin, PluginOpt
 
 
 class Sunbeam(Plugin, UbuntuPlugin):
@@ -18,6 +22,13 @@ class Sunbeam(Plugin, UbuntuPlugin):
     packages = ('openstack',)
 
     common_dir = '/var/snap/openstack/common'
+
+    option_list = [
+        PluginOpt('sunbeam-user', default='ubuntu', val_type=str,
+                  desc='The user used for sunbeam installation'),
+        PluginOpt('juju-allow-login', default=False, val_type=bool,
+                  desc='Allow sos to login to juju'),
+    ]
 
     def setup(self):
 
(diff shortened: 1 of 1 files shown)


pmoravec:
Sounds good to me (from a pure sos side, not knowing either command or file to be collected).

arif-ali:
I think this does a beg a question, should we improve some of the commands to be able collect commands as a user, so we're not doing blanket `su - <user>` for every command we want to collect.

The problem also I have this which I discovered last night, is that 2 of these commands will also collect certs within the command itself. If the command could be different, do I put the same login the the `postproc()`, it just seems like duplicating code. I'll have a think on how best to do it, but others might have an idea already

This is here for collaboration with my team at the moment, and do some improvements.

pmoravec:
I think some support for "running under user" within `Plugin` class makes a lot of sense. Three other plugins (`aap_eda`, `saphana` and `sapnw`) will utilize from it as well.

sos/report/plugins/sunbeam.py
```diff
@@ -37,10 +48,90 @@ def setup(self):
...
+
+            self.add_copy_spec([
+                f"{sb_snap_homedir}/*.log",
+                f"{sb_snap_homedir}/etc/*/*.log",
+                f"{sb_snap_homedir}/logs/*.log",
+            ])
+
+            if self.get_option("juju-allow-login"):
+                self.exec_cmd(
+                    f'su - {sunbeam_user} -c "sunbeam utils juju-login"')
```
arif-ali, line 65: This did not work with `runas` argument as this command has some interactive related items, and the only way to get this working was via `su`

TurboTurtle (approved):

sos/report/plugins/sunbeam.py
```diff
@@ -37,10 +48,90 @@ def setup(self):
             'sunbeam cluster list --format yaml',
         ])
 
+        sunbeam_user = self.get_option("sunbeam-user")
+        user_pwd = pwd.getpwnam(sunbeam_user)
```
pmoravec, line 52: If the user does not exist, `getpwnam` raises an exception that will end up in `sos_logs/sunbeam-plugin-errors.txt`:

    >>> pwd.getpwnam('no-user')
      File "<stdin>", line 1, in <module>
    KeyError: "getpwnam(): name not found: 'no-user'"
    >>> 


Technically it doesnt matter much since even catching the exception will cause the plugin wont collect anything (due to `if user_pwd`), just that scenario looks bit non-professional.

Leaving on @arif-ali to decide if it is worth fixing it - I dont want to block this "form of error reporting" point from merging.

pmoravec (approved):

sos/report/plugins/sunbeam.py
arif-ali: good catch, I agree it's looks non-professional if that error file is created; I will fix and re-upload. Thanks for catching this

pmoravec (approved):

TurboTurtle (approved):

PR #3648, opened 2024-05-20
[global] Add the ability to run command as user

arif-ali:
---

### DIFF
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -2003,7 +2003,10 @@ class Plugin():
                 return
         if self.test_predicate(cmd=True, pred=pred):
             self.collect_cmds.append(soscmd)
-            self._log_info(f"added cmd output '{soscmd.cmd}'")
+            user = ""
+            if hasattr(soscmd, "runas"):
+                user = f", as the {soscmd.runas} user"
+            self._log_info(f"added cmd output '{soscmd.cmd}'{user}")
         else:
             self.log_skipped_cmd(soscmd.cmd, pred, changes=soscmd.changes)
 
@@ -2013,7 +2016,7 @@ class Plugin():
                        sizelimit=None, pred=None, subdir=None,
                        changes=False, foreground=False, tags=[],
                        priority=10, cmd_as_tag=False, container=None,
-                       to_file=False):
+                       to_file=False, runas=None):
         """Run a program or a list of programs and collect the output
 
         Output will be limited to `sizelimit`, collecting the last X amount
@@ -2086,6 +2089,9 @@ class Plugin():
         :param to_file: Should command output be written directly to a new
                         file rather than stored in memory?
         :type to_file:  ``bool``
+
+        :param runas: Run the `cmd` as the `runas` user
+        :type runas: ``str``
         """
         if isinstance(cmds, str):
             cmds = [cmds]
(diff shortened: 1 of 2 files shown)


arif-ali:
An initial idea to discuss

Once we migrate to python3.9, we can use the `user` argument for `Popen` and not have to do the `setuid()`. I get the results with that.

Once gotcha I've had so far, if a command is run, to establish some kind of connection, to then run commands afterwards, that doesn't seem to work.

So, any thoughts on this would be great, or if someone else has a better way/idea, happy to abandon this PR

TurboTurtle:
> Once we migrate to python3.9, we can use the `user` argument for `Popen` and not have to do the `setuid()`.

My immediate thought was to hold this implementation until the next minor version, but if it requires 3.9 then there's not much use to that, given that we've planned to go to 3.8 instead, with 3.9 coming in early/mid 2025 (per #3533).

I haven't yet played around with this, but something that comes to mind is, does the `os.setuid()` and `os.setgid()` calls not persist after the first invocation? Just thinking out loud here, wouldn't we need to reset to root after each non-root command in order to avoid running all following commands as the non-root user?

The [docs](https://docs.python.org/3/library/os.html#os.setuid) mention this sets the current _process'_ id, so I imagine this is not thread-safe?

arif-ali:
> I haven't yet played around with this, but something that comes to mind is do the `os.setuid()` and `os.setgid()` calls not persist after the first invocation? Just thinking out loud here, wouldn't we need to reset to root after each non-root command in order to avoid running all following commands as the non-root user?
> 
> The [docs](https://docs.python.org/3/library/os.html#os.setuid) mention this sets the current _process'_ id, so I imagine this is not thread-safe?

based on my testing on the updated PR I have for #3646, I ran through a full sos, and I only saw the commands that it needed to run as the user, actually do it. My understanding was that the `setuid()` for that process, i.e. the `Popen` being used, so hence didn't think we would need to change others.

Happy to hold off for others to do similar testing, and see how it goes. I've got an update on my system for docstrings, so can incorporate anything else that is required here.

TurboTurtle:
Oh! I didn't realize that was in `_child_prep_fn()` at first. Ok, that makes sense to me now.

arif-ali:
There is one gotcha with this at the moment especially with the other PR I am working with

one of the commands only works with `su -u <user> -c "command"` but doesn't with the `runas=user`, so there must be some subtleties on what will and won't work. I've tried a few different ideas with `Popen` but haven't got there, if anyone has any ideas on what I can try, then that would be awesome.

Ultimately for that plugin, I only need to run those as `exec_cmd`, and one-offs, so I am not precious about it, especially if the command collection actually works without the extra `su -u <user>`, especially it looks so much more cleaner too :)

arif-ali:
So, the issue I have with the current implementation is a corner case because of the way the command for `sunbeam` interacts. Spent a lot of time outside of `sos` and bare python code to see if there was a way to get that specific command to work, but to no avail.

So, this should be good to go, if we can get some reviews.

Ideally would like to get this, and once this is merged need to work on the sunbeam to be ready for the next z-release

pmoravec (approved):
ACK.

Maybe it makes sense to log the user somewhere, e.g. in `collecting output of` log in https://github.com/sosreport/sos/blob/main/sos/report/plugins/__init__.py#L3082, but it might be too detailed information already..? I dont know.. Leaving on @arif-ali to decide.

TurboTurtle:
> Maybe it makes sense to log the user somewhere

This was my thought as well. Code looks good, but I think we should explicitly highlight when we're collecting as non-root, since we run as root.

arif-ali:
After reviewing the logs, I found 2 places it was useful to add. Hopefully this fits with what you guys were anticipating

arif-ali:
tried re-running the centos-stream-8 rpm build, maybe an infra issue 🤷🏽

arif-ali:
Oh, CentOS Stream 8 is now EOL, maybe that's why the CI is not working today? Do we need to update that before we do new CI tests now, as all CI's will fail now

/cc @pmoravec  @jcastill

jcastill:
We can probably remove the references to CentOS Stream 8 now. Let me open a PR so we can discuss this separately and then if/when it's accepted, we can  retry this one.

pmoravec (approved):

TurboTurtle (approved):

PR #3650, opened 2024-05-28
[openstack_edpm] Add additional paths

slagle:
These paths are also used by EDPM and should be collected by the plugin.

Signed-off-by: James Slagle <jslagle@redhat.com>

---

### DIFF
--- a/sos/report/plugins/openstack_edpm.py
+++ b/sos/report/plugins/openstack_edpm.py
@@ -24,7 +24,9 @@ class OpenStackEDPM(Plugin, RedHatPlugin):
         # Notes: recursion is max 2 for edpm-config
         # Those directories are present on all OpenStack nodes
         self.edpm_log_paths = [
-            '/var/lib/edpm-config/'
+            '/etc/os-net-config/',
+            '/var/lib/config-data/',
+            '/var/lib/edpm-config/',
         ]
         self.add_copy_spec(self.edpm_log_paths)
 


sos/report/plugins/openstack_edpm.py
```diff
@@ -24,7 +24,10 @@ def setup(self):
         # Notes: recursion is max 2 for edpm-config
         # Those directories are present on all OpenStack nodes
         self.edpm_log_paths = [
-            '/var/lib/edpm-config/'
+            '/etc/os-net-config/'
```
arif-ali, line 27: missing comma at the end

sos/report/plugins/openstack_edpm.py
```diff
@@ -24,7 +24,10 @@ def setup(self):
...
+            '/var/lib/config-data/',
+            '/var/lib/edpm-config/',
+            '/var/log/containers/'
```
arif-ali, line 30: can we add a comma at the end here too

arif-ali (changes requested):

sos/report/plugins/openstack_edpm.py
slagle: Done.

sos/report/plugins/openstack_edpm.py
slagle: Done.

arif-ali (approved):

sos/report/plugins/openstack_edpm.py
```diff
@@ -24,7 +24,10 @@ def setup(self):
         # Notes: recursion is max 2 for edpm-config
         # Those directories are present on all OpenStack nodes
         self.edpm_log_paths = [
-            '/var/lib/edpm-config/'
+            '/etc/os-net-config/',
+            '/var/lib/config-data/',
+            '/var/lib/edpm-config/',
+            '/var/log/containers/',
```
TurboTurtle, line 30: Nack to `/var/log/containers` as a whole. There are _tons_ of components that drop logs there these days.

The `container_log` plugin capture this as needed, but if you need specific containers call them out explicitly in these paths, e.g. `/var/log/containers/$container/$log` which is what many of the other openstack plugins do.

slagle: dropped /var/log/containers

TurboTurtle (approved):

PR #3651, opened 2024-05-31
[charmed_postgresql] Add Charmed PostgreSQL support

acsgn:
Example output of the plugin can be seen at https://drive.google.com/drive/folders/1t1QOwzqwFTv140xvpaRpB5-kqRPOYth5?usp=sharing

---

### DIFF
--- a/sos/report/plugins/charmed_postgresql.py
+++ b/sos/report/plugins/charmed_postgresql.py
@@ -1,0 +1,191 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+import os
+import yaml
+
+from sos.report.plugins import Plugin, UbuntuPlugin
+
+SNAP_COMMON_PATH = "/var/snap/charmed-postgresql/common"
+SNAP_CURRENT_PATH = "/var/snap/charmed-postgresql/current"
+
+PATHS = {
+    "POSTGRESQL_CONF": SNAP_COMMON_PATH + "/var/lib/postgresql",
+    "POSTGRESQL_LOGS": SNAP_COMMON_PATH + "/var/log/postgresql",
+    "PATRONI_CONF": SNAP_CURRENT_PATH + "/etc/patroni",
+    "PATRONI_LOGS": SNAP_COMMON_PATH + "/var/log/patroni",
+    "PGBACKREST_CONF": SNAP_CURRENT_PATH + "/etc/pgbackrest",
+    "PGBACKREST_LOGS": SNAP_COMMON_PATH + "/var/log/pgbackrest",
+    "PGBOUNCER_CONF": SNAP_CURRENT_PATH + "/etc/pgbouncer",
+    "PGBOUNCER_LOGS": SNAP_COMMON_PATH + "/var/log/pgbouncer",
+}
+
+PATRONI_CONFIG_FILE = f"{PATHS['PATRONI_CONF']}/patroni.yaml"
+
+RUNAS = "snap_daemon"
+PSQL = "charmed-postgresql.psql"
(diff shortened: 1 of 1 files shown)


sos/report/plugins/charmed_postgresql.py
```diff
@@ -0,0 +1,216 @@
...
+
+class CharmedPostgreSQL(Plugin, UbuntuPlugin):
+
+    short_desc = "Charmed PostgreSQL"
+    plugin_name = "charmed_postgresql"
+    packages = ('charmed-postgresql',)
+
+    @property
+    def patroni_cluster_name(self) -> Optional[str]:
+        with open(f"{PATHS['PATRONI_CONF']}/patroni.yaml") as file:
```
arif-ali, line 41: We need to enclose this in a try/Except to catch if the file exists. Currently we get the following error with the AllPluginSmokeTest

```
  File "/tmp/cirrus-ci-build/sos/report/__init__.py", line 1251, in setup
    plug.setup()
  File "/tmp/cirrus-ci-build/sos/report/plugins/charmed_postgresql.py", line 140, in setup
    f"topology {self.patroni_cluster_name}"),
                ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/cirrus-ci-build/sos/report/plugins/charmed_postgresql.py", line 41, in patroni_cluster_name
    with open(f"{PATHS['PATRONI_CONF']}/patroni.yaml") as file:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/var/snap/charmed-postgresql/current/etc/patroni/patroni.yaml'
```

arif-ali (changes requested):
We probably want to make the same change with other places too

sos/report/plugins/charmed_postgresql.py
acsgn: Setup is redesigned. Now, we collect the files always. If the config file exists and structured correctly, additional commands run for better insight. Also, password is directly set on the process' environment so that it does not leak on verbose flag. This is the same with the regular PostgreSQL plugin.

arif-ali:
@acsgn can you rebase to `main` please as the centos-stream 8 was EOL'd and now the CI doesn't work on those. This was just merged, thanks

arif-ali:
looks good overall

a heads up, we have a new PR  (#3648) that we have been working on, that allows you to run a command as a user without using `su` or `runuser`. You could potentially rework those bits in this code, and utilise that feature once that is merged.

Furthermore, we prefer that any PR that comes through is squashed before we merge.

acsgn:
@arif-ali I believe we should move the chdir to below runas. Currently, Runas overrides the user's specification.
https://github.com/sosreport/sos/blob/91c7c5484beb7b8f4d05dea0cdf6b60d34200ca7/sos/utilities.py#L234

arif-ali:
> @arif-ali I believe we should move the chdir to under the runas. Currently, runas overrides the user's specification.
> 
> https://github.com/sosreport/sos/blob/91c7c5484beb7b8f4d05dea0cdf6b60d34200ca7/sos/utilities.py#L234-L239

good call and thanks for noticing that, I'll get that resolved in another PR

arif-ali (approved):
based on internal discussions and testing, this looks good now

sos/report/plugins/charmed_postgresql.py
```diff
@@ -0,0 +1,191 @@
...
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+import os
+import yaml
+
+from sos.report.plugins import Plugin, UbuntuPlugin
```
pmoravec, line 12: It seems `charmed-postgresql` is available on many distributions, per https://snapcraft.io/charmed-postgresql, but since "Users by distribution" show just Ubuntu, we can have just `UbuntuPlugin` until somebody wants otherwise.

pmoravec (approved):

PR #3652, opened 2024-06-03
[packit|cirrus] Remove centos-stream-8 references

jcastill:
Remove packit and cirrus references to centos-stream 8 that's EOL since May 31st 2024.

---

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -15,7 +15,6 @@ env:
     UBUNTU_PRIOR3_NAME: "ubuntu-18.04"
 
     CENTOS_9_NAME: "centos-stream-9"
-    CENTOS_8_NAME: "centos-stream-8"
 
     CENTOS_PROJECT: "centos-cloud"
     DEBIAN_PROJECT: "debian-cloud"
@@ -26,7 +25,6 @@ env:
 
     # Images exist on GCP already
     CENTOS_9_IMAGE_NAME: "centos-stream-9-v20240415"
-    CENTOS_8_IMAGE_NAME: "centos-stream-8-v20240415"
     DEBIAN_IMAGE_NAME: "debian-11-bullseye-v20230809"
     FEDORA_IMAGE_NAME: "fedora-cloud-base-gcp-38-1-6-x86-64"
     FEDORA_PRIOR_IMAGE_NAME: "fedora-cloud-base-gcp-37-1-7-x86-64"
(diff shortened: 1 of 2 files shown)


jcastill:
First time doing this, so hopefully I didn't miss anything. This is related to the errors that @arif-ali got in #3648 . CentOS Stream 8 was EOL'ed on May 31st 2024. 
@pmoravec , thoughts?

arif-ali:
Just thinking from the outside, and no experience with foreman at all.

Is it worth doing the tests for foreman on centos9 stream instead and the versions being removed, or does it not make sense?

jcastill:
Good point. I thought it was related to CentOS 8 only, but I haven't found any info regarding that yet. Hopefully Pavel should be able to shed some light

pmoravec:
Foreman is supported on CentOS Stream 8 and 9 (https://www.theforeman.org/manuals/3.10/quickstart_guide.html) while the downstream product Satellite is supported on RHEL8 only (so far). For those "legacy" reasons, I would suggest to:

- still keep CentOS Stream 8 just for foreman testing, for some time
- have foreman tested also on CentOS 9
- bump the foreman version from 3.3/3.5/3.7 to e.g. 3.9/3.10 (worth doing in either case as foreman evolved)

Like:

        matrix:
            - env:
                <<: *centos8
                FOREMAN_VER: "3.9"
            - env:
                <<: *centos9
                FOREMAN_VER: "3.10"
            - env:
                PROJECT: ${DEBIAN_PROJECT}
                VM_IMAGE_NAME: ${DEBIAN_IMAGE_NAME}
                BUILD_NAME: ${DEBIAN_NAME}
                FOREMAN_VER: "3.10"

jcastill:
Thank you @pmoravec . I'll rewrite this and push again

arif-ali:
you're still going to have issues with the yum repos right? and will need to point to vault.centos.org? We could do the same for the other CI bits, if we're going to keep it for foreman

pmoravec:
Indeed, the EOS of CentOS Stream 8 breaks a lot (for us). We are figuring out how to proceed with some testing for foreman.

As any PR will fail on the CI tests for CentOS Stream 8 now, I would suggest merging this PR as is as a kind of "hotfix" to prevent false negatives. And open a new issue where we will propose some more stable plan.

Does this make sense or do you see a better approach?

arif-ali:
I think that makes sense, let's create a new GH issue, and then we can track it, and resolve it long-term, if there is a resoluton.

I think, what you could do, update the setup section, detect that if it is centos 8, and then update the dnf/yum repos to point to vault? I think that could work

arif-ali (approved):
based on previous comment, happy for this to go through, but lets create a new issue.

however, if it's a quick fix/turnaround, then we can continue in this PR

pmoravec (approved):

PR #3653, opened 2024-06-04
[apt] Obfuscate passwords for configs/sources

arif-ali:
The configuration of apt was not being obfuscated if any of the URLs had a password; one example was the proxy. There is also a new format for the deb sources, so ensuring that would be obfuscated in the same way. Add some tests to ensure that this is always being obfuscated.

---

### DIFF
--- a/sos/report/plugins/apt.py
+++ b/sos/report/plugins/apt.py
@@ -47,16 +47,20 @@ class Apt(Plugin, DebianPlugin, UbuntuPlugin):
 
     def postproc(self):
         super().postproc()
-        self.do_file_sub(
-            "/etc/apt/sources.list",
-            r"(deb\shttp(s)?://)\S+:\S+(@.*)",
-            r"\1******:******\3"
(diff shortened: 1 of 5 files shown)


sos/report/plugins/apt.py
```diff
@@ -47,15 +47,27 @@ def setup(self):
...
+        )
+
+        self.do_file_sub(
+            "/etc/apt/apt.conf",
+            common_regex, common_replace
+        )
+
+        self.do_path_regex_sub(
+            "/etc/apt/apt.conf.d/",
+            common_regex, common_replace
```
pmoravec, line 70: If you apply same obfuscation four times, maybe you can shorten it to

    for p in ["/etc/apt/sources.list", ..]:
        self.do_file_sub(p, common_regex, common_replace)

?

pmoravec (approved):

TurboTurtle (approved):

PR #3654, opened 2024-06-06
[memcached] Collect from memcached using the tool

arif-ali:
---

### DIFF
--- a/sos/report/plugins/memcached.py
+++ b/sos/report/plugins/memcached.py
@@ -19,6 +19,22 @@ class Memcached(Plugin):
     profiles = ('webserver',)
     packages = ('memcached',)
 
+    mm_tool = "memcached-tool"
+
+    def setup(self):
+        host = "127.0.0.1:11211"
+
+        subcmds = [
+            "display",
+            "dump",
+            "keys",
+            "stats",
+            "settings",
+        ]
+
+        self.add_cmd_output([
+            f"{self.mm_tool} {host} {subcmd}" for subcmd in subcmds])
+
 
 class RedHatMemcached(Memcached, RedHatPlugin):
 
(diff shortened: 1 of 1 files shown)


TurboTurtle (approved):
LGTM, though it seems strange to me that `memcached-tool` wouldn't be in $PATH when included as a packaged executable (whereas if you were installing from source, totally understandable).

arif-ali:
> LGTM, though it seems strange to me that `memcached-tool` wouldn't be in $PATH when included as a packaged executable (whereas if you were installing from source, totally understandable).

It's in the path for rocky8/rocky9 (which I just tested) which I am presume will be the case for RH based systems but it is not for ubuntu. So, happy to change that for the `RedHatPlugin` if you prefer

TurboTurtle:
> > LGTM, though it seems strange to me that `memcached-tool` wouldn't be in $PATH when included as a packaged executable (whereas if you were installing from source, totally understandable).
> 
> It's in the path for rocky8/rocky9 (which I just tested) which I am presume will be the case for RH based systems but it is not for ubuntu. So, happy to change that for the `RedHatPlugin` if you prefer

Didn't need to change, was just making an observation that I am surprised that the Ubuntu packaging wouldn't drop it in $PATH.

Either way though, this looks good to me.

arif-ali:
> 
> Didn't need to change, was just making an observation that I am surprised that the Ubuntu packaging wouldn't drop it in $PATH.
> 
> Either way though, this looks good to me.

It looked more elegant, and less code.

Reading up on it, apparently it can cause a pause on the cache gathering/setting if this tool is run, so that could be the reason for not adding it to the $PATH. Maybe a warning is at hand here, wdyt?

TurboTurtle:
If it can cause an interruption to normal operation I think we'd ideally gate it behind a plugin option, but there isn't much else this plugin collects and I doubt that the general use case for support folks would be well served without this info. Maybe a warning would suffice, let's dig in a little bit.

Is there any indication on how likely a pause or how long the pause could be (full duration of the command, or is it very brief at the beginning of execution), or for that matter is it _any_ time this command is run, or only for specific sub commands like `dump`?

arif-ali:
Reading up on this if we run `star sizes` then there could be issues as per https://github.com/memcached/memcached/wiki/Commands#stats-sizes. As we're not passing that command, we should be ok.

So I think, this should be good to go

PR #3657, opened 2024-06-08
[kernel] Print human-readable timestamps with dmesg command in kernel

NitinYewale:
plugin

---

- [X ] Are all passwords or private data gathered by this PR [obfuscated](https://github.com/sosreport/sos/wiki/How-to-Write-a-Plugin#how-to-prevent-collecting-passwords)?

### DIFF
--- a/sos/report/plugins/kernel.py
+++ b/sos/report/plugins/kernel.py
@@ -75,7 +75,7 @@ class Kernel(Plugin, IndependentPlugin):
             self.add_cmd_output(f"find {' '.join(extra_mod_paths)} -ls")
 
         self.add_cmd_output([
-            "dmesg",
+            "dmesg -T",
             "dkms status"
         ], cmd_as_tag=True)
         self.add_cmd_output("sysctl -a", tags="sysctl")


NitinYewale:
There are other placed where we could add "-T" to `dmesg` command.. example :

gcp.py:47:        dmesg = self.exec_cmd("dmesg")
xen.py:77:                "xm dmesg",

Please suggest if these needs to be changed.

TurboTurtle (approved):
Makes sense to me.

The flake8 job failure seems unrelated to this change and appears to come from `tox`.

TurboTurtle:
> There are other placed where we could add "-T" to `dmesg` command.. example :
> 
> gcp.py:47: dmesg = self.exec_cmd("dmesg") xen.py:77: "xm dmesg",
> 
> Please suggest if these needs to be changed.

The `gcp` plugin usage is around finding if the GCP specific entry in dmesg is present to enable the plugin (come to think of it, is there a better approach here?), so timestamps wouldn't matter there (we also don't hold on to that output).

pmoravec (approved):

NitinYewale:
Thank you @TurboTurtle @pmoravec

bmr-cymru:
Bear in mind the warning from the `dmesg` man page:

```
       -T, --ctime
           Print human-readable timestamps.

           Be aware that the timestamp could be inaccurate! The time source used for the logs is not updated
           after system SUSPEND/RESUME. Timestamps are adjusted according to current delta between
           boottime and monotonic clocks, this works only for messages printed after last resume.
```

PR #3664, opened 2024-06-10
[Plugin] Standardize directory listings with new method

TurboTurtle:
This commit adds a new `add_dir_listing()` wrapper method to standardize the collection of directory content listings across plugins. By default this will use the `ls` command, and will use `tree` if the `tree` parameter is set to `True`.

This wrapper is not intended to replace directory listing that are handlded by other components with their own `ls` functionality (e.g. `npm ls` in the `npm` plugin). Additionally this does not replace the direct calling of `ls` when used with `Plugin.exec_cmd()` to take some action based on the output of the `ls` command.

---

### DIFF
--- a/sos/report/plugins/aap_controller.py
+++ b/sos/report/plugins/aap_controller.py
@@ -56,12 +56,15 @@ class AAPControllerPlugin(Plugin, RedHatPlugin):
             "/var/lib/awx/venv/awx/bin/pip freeze -l",
             "/var/lib/awx/venv/ansible/bin/pip freeze",
             "/var/lib/awx/venv/ansible/bin/pip freeze -l",
-            "tree -d /var/lib/awx",
-            "ls -ll /var/lib/awx",
-            "ls -ll /var/lib/awx/venv",
-            "ls -ll /etc/tower",
             "umask -p",
         ])
+
+        self.add_dir_listing([
+            '/var/lib/awx',
+            '/var/lib/awx/venv',
+            '/etc/tower',
+        ])
+        self.add_dir_listing('/var/lib/awx', tree=True)
 
     def postproc(self):
         # remove database password
--- a/sos/report/plugins/aap_eda.py
+++ b/sos/report/plugins/aap_eda.py
@@ -35,11 +35,11 @@ class AAPEDAControllerPlugin(Plugin, RedHatPlugin):
             "/etc/ansible-automation-platform/eda/server.key",
         ])
 
-        self.add_cmd_output([
-            "aap-eda-manage --version",  # eda version
-            "ls -alhR /etc/ansible-automation-platform/",
-            "ls -alhR /var/log/ansible-automation-platform/",
-        ])
+        self.add_cmd_output("aap-eda-manage --version")
+        self.add_dir_listing([
+            "/etc/ansible-automation-platform/",
+            "/var/log/ansible-automation-platform/",
+        ], recursive=True)
 
         self.add_cmd_output("su - eda -c 'export'",
                             suggest_filename="eda_export")
--- a/sos/report/plugins/aap_gateway.py
+++ b/sos/report/plugins/aap_gateway.py
@@ -35,11 +35,9 @@ class AAPGatewayPlugin(Plugin, RedHatPlugin):
             "/etc/ansible-automation-platform/gateway/*.cert",
         ])
 
-        self.add_cmd_output([
-            "aap-gateway-manage list_services",
-            "ls -ll /etc/ansible-automation-platform/",
-            "ls -ll /etc/ansible-automation-platform/gateway/",
-        ])
+        self.add_cmd_output("aap-gateway-manage list_services")
+        self.add_dir_listing("/etc/ansible-automation-platform/",
+                             recursive=True)
 
     def postproc(self):
         # remove database password
--- a/sos/report/plugins/aap_hub.py
+++ b/sos/report/plugins/aap_hub.py
@@ -29,9 +29,9 @@ class AAPAutomationHub(Plugin, RedHatPlugin):
 
         ])
 
-        self.add_cmd_output([
-            "ls -alhR /etc/ansible-automation-platform/",
-            "ls -alhR /var/log/ansible-automation-platform/",
-        ])
+        self.add_dir_listing([
+            "/etc/ansible-automation-platform/",
+            "/var/log/ansible-automation-platform/",
+        ], recursive=True)
 
 # vim: set et ts=4 sw=4 :
--- a/sos/report/plugins/apparmor.py
+++ b/sos/report/plugins/apparmor.py
@@ -29,10 +29,11 @@ class Apparmor(Plugin, UbuntuPlugin):
             "/etc/apparmor.d/abstractions"
         ])
 
-        self.add_cmd_output([
-            "apparmor_status",
-            "ls -alh /etc/apparmor.d/abstractions",
-            "ls -alh /etc/apparmor.d/libvirt",
+        self.add_cmd_output("apparmor_status")
+
+        self.add_dir_listing([
+            '/etc/apparmor.d/abstractions',
+            '/etc/apparmor.d/libvirt'
         ])
 
 # vim: set et ts=4 sw=4 :
--- a/sos/report/plugins/ds.py
+++ b/sos/report/plugins/ds.py
@@ -77,7 +77,7 @@ class DirectoryServer(Plugin, RedHatPlugin):
                 "/opt/redhat-ds/slapd-*/logs"
             ])
 
-        self.add_cmd_output("ls -l /var/lib/dirsrv/slapd-*/db/*")
+        self.add_dir_listing("/var/lib/dirsrv/slapd-*/db/*")
 
     def postproc(self):
         # Example for scrubbing rootpw hash
--- a/sos/report/plugins/filesys.py
+++ b/sos/report/plugins/filesys.py
@@ -108,6 +108,6 @@ class RedHatFilesys(Filesys, RedHatPlugin):
 
     def setup(self):
         super().setup()
-        self.add_cmd_output("ls -ltradZ /tmp")
+        self.add_cmd_output('ls -ldZ /tmp')
 
 # vim: set et ts=4 sw=4 :
--- a/sos/report/plugins/grub2.py
+++ b/sos/report/plugins/grub2.py
@@ -39,7 +39,6 @@ class Grub2(Plugin, IndependentPlugin):
             "/etc/grub2-efi.cfg"
         ])
 
-        self.add_cmd_output("ls -lanR /boot", tags="ls_boot")
         # call grub2-mkconfig with GRUB_DISABLE_OS_PROBER=true to prevent
         # possible unwanted loading of some kernel modules
         # further, check if the command supports --no-grubenv-update option
--- a/sos/report/plugins/insights.py
+++ b/sos/report/plugins/insights.py
@@ -52,8 +52,8 @@ class RedHatInsights(Plugin, RedHatPlugin):
             timeout=30
         )
 
-        for _dir in ["/etc/rhsm", "/sys/kernel", "/var/lib/sss"]:
-            self.add_cmd_output(f"/bin/ls -lanR {_dir}", cmd_as_tag=True)
+        self.add_dir_listing(["/etc/rhsm", "/sys/kernel", "/var/lib/sss"],
+                             recursive=True)
 
     def postproc(self):
         for conf in self.config_and_status:
--- a/sos/report/plugins/libraries.py
+++ b/sos/report/plugins/libraries.py
@@ -44,7 +44,9 @@ class Libraries(Plugin, IndependentPlugin):
                 dirs.add(fqlib[1].rsplit('/', 1)[0])
 
             if dirs:
-                self.add_cmd_output(f"ls -lanH {' '.join(dirs)}",
-                                    suggest_filename="ld_so_cache")
+                self.add_dir_listing(
+                    f"{' '.join(dirs)}",
+                    suggest_filename='ld_so_cache'
+                )
 
 # vim: set et ts=4 sw=4 :
--- a/sos/report/plugins/ovn_central.py
+++ b/sos/report/plugins/ovn_central.py
@@ -209,8 +209,8 @@ class OVNCentral(Plugin):
                 dbfilepath = self.path_join(path, dbfile)
                 if self.path_exists(dbfilepath):
                     self.add_copy_spec(dbfilepath)
-                    self.add_cmd_output(
-                        f"ls -lan {dbfilepath}", foreground=True)
+                    self.add_dir_listing(dbfilepath)
+
             if ovs_dbdir:
                 self.add_copy_spec(self.path_join(ovs_dbdir, dbfile))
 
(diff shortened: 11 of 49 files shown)


sos/report/plugins/vdsm.py
```diff
@@ -98,11 +98,12 @@ def setup(self):
...
             ])
-        self.add_cmd_output([
-            "ls -ldZ /etc/vdsm",
-            "su vdsm -s /bin/sh -c 'tree -l /rhev/data-center'",
-            "su vdsm -s /bin/sh -c 'ls -lR /rhev/data-center'"
-        ])
+
+        self.add_dir_listing(
+            ['/etc/vdsm', '/rhev/data-center'],
+            runas='vdsm'
```
arif-ali, line 104: do we not need tree here for `/rhev/data-center`

TurboTurtle: Oh, good catch!

sos/report/plugins/aap_gateway.py
```diff
@@ -35,10 +35,10 @@ def setup(self):
...
         ])
 
-        self.add_cmd_output([
-            "aap-gateway-manage list_services",
-            "ls -ll /etc/ansible-automation-platform/",
-            "ls -ll /etc/ansible-automation-platform/gateway/",
+        self.add_cmd_output("aap-gateway-manage list_services")
+        self.add_dir_listing([
+            "/etc/ansible-automation-platform/",
+            "/etc/ansible-automation-platform/gateway/",
```
arif-ali, line 41: We could remove this now, if we're doing directory listing recursively by default

sos/report/plugins/grub2.py
```diff
@@ -39,7 +39,7 @@ def setup(self):
             "/etc/grub2-efi.cfg"
         ])
 
-        self.add_cmd_output("ls -lanR /boot", tags="ls_boot")
+        self.add_dir_listing('/boot', tags=['ls_boot'])
```
arif-ali, line 42: just thinking, we have the listing of `/boot` here and the boot plugin, is it worth removing it from one of the plugins?

arif-ali (commented):
One nit, there is some inconsistency where we have used `-R` previously and hence stuff with the new default or `recursive=True`. But we've had newer updates here now where some places where we didn't have `-R` before we are now doing recursive and some we aren't.

Happy though, if this was all intended

sos/report/plugins/aap_gateway.py
TurboTurtle: Good eye!

sos/report/plugins/grub2.py
TurboTurtle: Yeah, I thought about dropping it but instead had this commit focus on changing over to the new method, rather than consolidating collections. It's a simple enough change though, so I think we'd be fine dropping it from `grub2`.

sos/report/plugins/aap_controller.py
```diff
@@ -56,13 +56,16 @@ def setup(self):
...
-            "ls -ll /var/lib/awx/venv",
-            "ls -ll /etc/tower",
             "umask -p",
         ])
 
+        self.add_dir_listing([
+            '/var/lib/awx',
+            '/var/lib/awx/venv',
+            '/etc/tower',
+        ])
```
pmoravec, line 66: Missing `recursive=False`?

(should `recursive` have default value `True` or `False`? I would expect `False` since that follows default `ls` behaviour and I would expect it as a human-default thinking of "list directory" as well; also since the patch replaces 38 instances of recursive listings and 32 non-recursive ones, there is no need to have such strong preference to the "recursive is far more used default")

sos/report/plugins/apparmor.py
```diff
@@ -29,10 +29,11 @@ def setup(self):
             "/etc/apparmor.d/abstractions"
         ])
 
-        self.add_cmd_output([
-            "apparmor_status",
-            "ls -alh /etc/apparmor.d/abstractions",
-            "ls -alh /etc/apparmor.d/libvirt",
+        self.add_cmd_output("apparmor_status")
+
+        self.add_dir_listing([
```
pmoravec, line 34: Ditto `recursive=False`

sos/report/plugins/aap_controller.py
arif-ali: > (should `recursive` have default value `True` or `False`? I would expect `False` since that follows default `ls` behaviour and I would expect it as a human-default thinking of "list directory" as well; also since the patch replaces 38 instances of recursive listings and 32 non-recursive ones, there is no need to have such strong preference to the "recursive is far more used default")

When I was first reviewing the PR I was thinking the same, and missed the `recursive=True` as the default in the function definition, and had put in ~10 comments before abandoning it. I would agree with @pmoravec here, I think having `recursive=False` should be default and we use `recursive=True` where needed

TurboTurtle:
I've swapped the default to `recursive=False` and re-aligned things as needed. I believe I hit all of them, but a second pair of eyes to keep me honest is appreciated.

sos/report/plugins/aap_eda.py
```diff
@@ -35,10 +35,10 @@ def setup(self):
...
         ])
 
-        self.add_cmd_output([
-            "aap-eda-manage --version",  # eda version
-            "ls -alhR /etc/ansible-automation-platform/",
-            "ls -alhR /var/log/ansible-automation-platform/",
+        self.add_cmd_output("aap-eda-manage --version")
+        self.add_dir_listing([
+            "/etc/ansible-automation-platform/",
+            "/var/log/ansible-automation-platform/",
```
arif-ali, line 41: `recursive=true` here

sos/report/plugins/aap_hub.py
```diff
@@ -29,9 +29,9 @@ def setup(self):
 
         ])
 
-        self.add_cmd_output([
-            "ls -alhR /etc/ansible-automation-platform/",
-            "ls -alhR /var/log/ansible-automation-platform/",
+        self.add_dir_listing([
+            "/etc/ansible-automation-platform/",
+            "/var/log/ansible-automation-platform/",
```
arif-ali, line 34: `recursive=true` here

sos/report/plugins/ds.py
```diff
@@ -77,7 +77,7 @@ def setup(self):
                 "/opt/redhat-ds/slapd-*/logs"
             ])
 
-        self.add_cmd_output("ls -l /var/lib/dirsrv/slapd-*/db/*")
+        self.add_dir_listing("/var/lib/dirsrv/slapd-*/db/*", recursive=False)
```
arif-ali, line 80: remove recursive=False now that is the default

sos/report/plugins/filesys.py
```diff
@@ -108,6 +108,6 @@ class RedHatFilesys(Filesys, RedHatPlugin):
 
     def setup(self):
         super().setup()
-        self.add_cmd_output("ls -ltradZ /tmp")
+        self.add_dir_listing('/tmp', recursive=False)
```
arif-ali, line 111: remove recursive=False now that is the default

sos/report/plugins/ovn_central.py
```diff
@@ -209,8 +209,8 @@ def setup(self):
                 dbfilepath = self.path_join(path, dbfile)
                 if self.path_exists(dbfilepath):
                     self.add_copy_spec(dbfilepath)
-                    self.add_cmd_output(
-                        f"ls -lan {dbfilepath}", foreground=True)
+                    self.add_dir_listing(dbfilepath, recursive=False)
```
arif-ali, line 212: remove recursive=False now that is the default

sos/report/plugins/grub2.py
```diff
@@ -39,7 +39,7 @@ def setup(self):
             "/etc/grub2-efi.cfg"
         ])
 
-        self.add_cmd_output("ls -lanR /boot", tags="ls_boot")
+        self.add_dir_listing('/boot', tags=['ls_boot'], recursive=True)
```
arif-ali, line 42: do we still want to drop this duplicate?

TurboTurtle:
@arif-ali - updated and I dropped the duplicate collection from `grub2`.

sos/report/plugins/filesys.py
```diff
@@ -108,6 +108,6 @@ class RedHatFilesys(Filesys, RedHatPlugin):
 
     def setup(self):
         super().setup()
-        self.add_cmd_output("ls -ltradZ /tmp")
+        self.add_dir_listing('/tmp')
```
arif-ali, line 111: what this did before, and what it will give now are 2 distinct things, not sure how we best want to do that

sos/report/plugins/vdsm.py
```diff
@@ -98,11 +98,14 @@ def setup(self):
...
                 for name in files
             ])
-        self.add_cmd_output([
-            "ls -ldZ /etc/vdsm",
-            "su vdsm -s /bin/sh -c 'tree -l /rhev/data-center'",
-            "su vdsm -s /bin/sh -c 'ls -lR /rhev/data-center'"
-        ])
+
+        self.add_dir_listing(
+            ['/etc/vdsm', '/rhev/data-center'],
```
arif-ali, line 103: same here, the `-ldZ` gives only the details of the directory, and no listing

sos/report/plugins/ssh.py
```diff
@@ -84,6 +84,6 @@ def user_ssh_files_permissions(self):
         for user in users_data:
             home_dir = self.path_join(user.pw_dir, '.ssh')
             if self.path_isdir(home_dir):
```
arif-ali, line 86: is it worth removing this check now, as you're checking if the path exists in `add_dir_listing()`

arif-ali (commented):
Hopefully last review, missed a couple where `-d` is being used, where it only gives the information of that dir, and doesn't actually list the contents of the directory. It does give the same info at the top, with `.` for the directory, instead of the directory name.

TurboTurtle:
> Hopefully last review, missed a couple where -d is being used, where it only gives the information of that dir, and doesn't actually list the contents of the directory. It does give the same info at the top, with . for the directory, instead of the directory name.

I think we can accept that difference. For the vdsm ones, RHV is an EOL product that won't be getting updated further, and I wasn't aware of any automation around those collections when I left RH, and I seriously doubt anything has been added for it since.

For `/tmp`, I think falling back on the `.` listing should be fine, but I'll defer to @pmoravec on that one.

> is it worth removing this check now, as you're checking if the path exists in add_dir_listing()

Oh, good eye. I'll drop that check in the next update, pending Pavel's thoughts on `/tmp`.

sos/report/plugins/aap_gateway.py
```diff
@@ -35,11 +35,9 @@ def setup(self):
...
         ])
 
-        self.add_cmd_output([
-            "aap-gateway-manage list_services",
-            "ls -ll /etc/ansible-automation-platform/",
-            "ls -ll /etc/ansible-automation-platform/gateway/",
-        ])
+        self.add_cmd_output("aap-gateway-manage list_services")
+        self.add_dir_listing("/etc/ansible-automation-platform/",
+                             recursive=True)
```
pmoravec, line 40: We will additionally collect subdirs of `/etc/ansible-automation-platform/gateway/` but I doubt that will cause an issue (like, speculating, some fiel in nested dir has some secret name we should not collect - but I doubt that would ever happen). ACK

pmoravec:
I see one skipped placefor the new method:

    sos/report/plugins/openstack_neutron.py:        self.add_cmd_output("ls -laZR /var/lib/neutron/lock")

sos/report/plugins/insights.py
```diff
@@ -52,8 +52,8 @@ def setup(self):
             timeout=30
         )
 
-        for _dir in ["/etc/rhsm", "/sys/kernel", "/var/lib/sss"]:
-            self.add_cmd_output(f"/bin/ls -lanR {_dir}", cmd_as_tag=True)
+        self.add_dir_listing(["/etc/rhsm", "/sys/kernel", "/var/lib/sss"],
+                             recursive=True)
```
pmoravec, line 56: We stop adding tags here. Also not sure if `-n` is important or not (`like -l, but list numeric user and group IDs`).

I am checking this internally..

sos/report/plugins/libraries.py
```diff
@@ -44,7 +44,9 @@ def setup(self):
                 dirs.add(fqlib[1].rsplit('/', 1)[0])
 
             if dirs:
-                self.add_cmd_output(f"ls -lanH {' '.join(dirs)}",
-                                    suggest_filename="ld_so_cache")
+                self.add_dir_listing(
+                    f"{' '.join(dirs)}",
+                    suggest_filename='ld_so_cache'
+                )
```
pmoravec, line 50: We will stop printing UIDs/GIDs here (dropping `-n` option), no idea if / how much it can matter. I expect no big deal - can @rmetrich as author of https://github.com/sosreport/sos/commit/ff74942afb5b1f0d2b840204325f709fa0896452 confirm?

sos/report/plugins/tftpserver.py
```diff
@@ -21,8 +21,6 @@ class TftpServer(Plugin, IndependentPlugin):
     packages = ('tftp-server',)
 
     def setup(self):
-        self.add_cmd_output("ls -lanR /tftpboot")
-        self.add_cmd_output('ls -lanR /srv/tftp')
-
+        self.add_dir_listing(['/tftpboot', '/sr/tftp'], recursive=True)
```
pmoravec, line 24: Also removing `-n` option, again no idea if it matters or not. I am ok with removal, if somebody will miss it, we can add it back.

rmetrich:
No need for UID indeed.

On Saturday, June 22, 2024, Pavel Moravec ***@***.***> wrote:
> @pmoravec commented on this pull request.
>
> ________________________________
>
> In sos/report/plugins/tftpserver.py:
>
>> @@ -21,8 +21,6 @@ class TftpServer(Plugin, IndependentPlugin):
> packages = ('tftp-server',)
>
> def setup(self):
> - self.add_cmd_output("ls -lanR /tftpboot")
> - self.add_cmd_output('ls -lanR /srv/tftp')
> -
> + self.add_dir_listing(['/tftpboot', '/sr/tftp'], recursive=True)
>
> Also removing -n option, again no idea if it matters or not. I am ok with
removal, if somebody will miss it, we can add it back.
>
> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> You are receiving this because you were mentioned.<
https://ci3.googleusercontent.com/meips/ADKq_NYRNj593dWT-krSyFYtzfnYZKe-fvAhVHIXOIEOJ-RTFsNGfa4R6X0ECQb0MvAMktdU2_-EGvgQQCEB4gI-mONCYkdxeCkRPVx0X_Iy7_LJl_-uB3_OIAdYY1piXdplGfnVgtZIKMrQAkRjn3I0IR6AlHhOqXotskNNv0Gb-HpD4WpsV5op0LXN5Q0EuGGoYpHJrdN-Y6vMwlqDrb9omt-wuC8bv_cAPfSeofW41bHsNllh8zdYVrQxMnB3eWFeOQ=s0-d-e1-ft#https://github.com/notifications/beacon/AAI4C42GBNSRF7DRYOS6XBLZIV6HFA5CNFSM6AAAAABJC7342WWGG33NNVSW45C7OR4XAZNRKB2WY3CSMVYXKZLTORJGK5TJMV32UY3PNVWWK3TUL5UWJTT7F52HY.gif>Message
ID: ***@***.***>

-- 
Renaud
Out of working hours
Sent from my phone, sorry for the typos...

sos/report/plugins/filesys.py
TurboTurtle: Ack, will move this one back to `ls` directly. `a` also is meaningless with `d`, so this will become `ls -ldZ /tmp`.

arif-ali (approved):

TurboTurtle:
@pmoravec any other remaining concerns?

pmoravec (approved):
ACK, sorry for a delay.

Issue #3671, opened 2024-06-12
sos.conf options dont match cmdline options


pmoravec:
Example:

    sos report --batch --build -o networking --clean --skip-cleaning-files="*"

should basically disable cleaner (artificial example). But having in `/etc/sos/sos.conf`:

    [clean]
    skip-cleaning-files=*

is ignored. The "right" option there equivalent to `--skip-cleaning-files="*"` is:

    [clean]
    skip-clean-files=*

Not sure if just this option is affected, or some others as well.

TurboTurtle:
It's because of this:

```python
        clean_grp.add_argument('--skip-cleaning-files', '--skip-masking-files',
                               action='extend', default=[],
                               dest='skip_clean_files',  <====
                               help=('List of files to skip/ignore during '
                                     'cleaning. Globs are supported.'))
```

So we just need to update the internal variable name, I think. Will have something up shortly after some quick tests.

jcastill:
And this as well, right?

 arg_defaults = {
        'archive_type': 'auto',
        'domains': [],
        'disable_parsers': [],
        'skip_clean_files': [], <-- 
        'jobs': 4,
        'keywords': [],
        'keyword_file': None,
        'map_file': '/etc/sos/cleaner/default_mapping',
        'no_update': False,
        'keep_binary_files': False,
        'target': '',
        'usernames': []
    }


In sos/cleaner/__init__.py .

TurboTurtle:
Yes, it's in a number of places. Also, I've found a bug in handling skipping files where if we don't have _any_ parsing to do (e.g. the `*` value, or anything that would cause all parsers to skip the file) we still go through the process of creating a temp file and reading/scanning/writing stuff out. It's definitely an edge case primarily highlighted by this synthetic scenario but it's worth addressing.

PR #3672, opened 2024-06-12
[clean,options] Correct skip-clean-files in sos.conf and skipping logic

TurboTurtle:
First, fix the issue raised in #3671 for the `--skip-cleaning-files` option, which would need to be written to `sos.conf` as `skip-clean-files` due to the internal variable naming.

Second, noticed as part of the reproducer for this issue when `--skip-cleaning-files` would be seti, that we were inadvertantly still going through the motions of creating a temporary file and scanning the file even if we weren't intending to do any obfuscations to it. Correct this by bailing out after determining there are no parsers for which we want to do an obfuscation. Apply the same update to symlink paths so that we don't create broken symlinks.

Resolves: #3671

---

### DIFF
--- a/sos/cleaner/__init__.py
+++ b/sos/cleaner/__init__.py
@@ -84,7 +84,7 @@ class SoSCleaner(SoSComponent):
         'archive_type': 'auto',
         'domains': [],
         'disable_parsers': [],
-        'skip_clean_files': [],
+        'skip_cleaning_files': [],
         'jobs': 4,
         'keywords': [],
         'keyword_file': None,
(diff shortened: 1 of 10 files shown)


sos/cleaner/__init__.py
```diff
@@ -769,15 +769,21 @@ def obfuscate_file(self, filename, short_name=None, arc_name=None):
...
                 _p for _p in self.parsers if not
                 any(
                     _skip.match(short_name) for _skip in _p.skip_patterns
                 )
             ]
+            if not _parsers:
+                self.log_debug(
+                    f"Skipping obfuscation of {short_name or filename} due to "
+                    f"matching file skip pattern"
+                )
```
pmoravec, line 782: Great! I missed log like this (and got confusing log `Obfuscating <filename>` instead) when discovering the reproducer for the issue.

pmoravec (approved):
Sorry for leaving this bias in code from my original feature.

I checked code for other options that can be affected by the same and found:

    sos/collector/__init__.py:        collect_grp.add_argument('-b', '--become', action='store_true',
    sos/collector/__init__.py-                                 dest='become_root',

    sos/collector/__init__.py:        collect_grp.add_argument('-c', '--cluster-option',
    sos/collector/__init__.py-                                 dest='cluster_options', action='append',

    sos/report/__init__.py:        report_grp.add_argument("-p", "--profile", "--profiles",
    sos/report/__init__.py-                                action="extend", dest="profiles", type=str,

    sos/report/__init__.py:        report_grp.add_argument("-k", "--plugin-option", "--plugopts",
    sos/report/__init__.py-                                dest="plugopts", type=str,

The latest is such an old option that users have adopted the difference rather than being confused (I think), so I would keep the `--plugin-option` vs. `plugopts=` difference.

The rest "diffs" are worth fixing the same way, I expect. I will raise a PR for them once this is one merged.

PR #3674, opened 2024-06-14
[curtin] New plugin curtin

arif-ali:
---

### DIFF
--- a/sos/report/plugins/curtin.py
+++ b/sos/report/plugins/curtin.py
@@ -1,0 +1,57 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, IndependentPlugin
+
+
+class Curtin(Plugin, IndependentPlugin):
+
+    short_desc = 'Curt Installer'
+
+    plugin_name = 'curtin'
+    profiles = ('boot',)
+    files = ('/root/curtin-install-cfg.yaml', )
+
+    def setup(self):
+        self.add_copy_spec([
+            '/root/curtin-install.log',
+            '/root/curtin-install-cfg.yaml',
+            '/etc/default/grub.d/50-curtin-settings.cfg',
+            '/etc/apt/apt.conf.d/90curtin-aptproxy',
+            '/etc/apt/sources.list.curtin.old',
+            '/etc/cloud/cloud.cfg.d/50-curtin-networking.cfg',
+            '/etc/cloud/cloud.cfg.d/curtin-preserve-sources.cfg',
+        ])
+
+    def postproc(self):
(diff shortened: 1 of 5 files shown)


tests/report_tests/plugin_tests/curtin/curtin-install.log
TurboTurtle, line 1: I'm entirely unfamiliar with curtin, so forgive me if this is way off base.

Can we not perform a curtin-based install as a separate CI job, and then reference a "native" install log? I'd feel more comfortable being able to validate against a recently-generated log (which would also allow us to see potential new things to obfuscate) over a bundled copy of one.

tests/report_tests/plugin_tests/curtin/curtin.py
```diff
@@ -0,0 +1,56 @@
...
+    files = [
+        ('curtin-install-cfg.yaml', '/root/curtin-install-cfg.yaml'),
+        ('curtin-install.log', '/root/curtin-install.log'),
+    ]
+
+    keys_obfuscate = [
+        'yaYmBSUuArGt3JYwHU',
+        'WCCtp2JewdWkHEDAW6',
+        'Bm2k3ZdGxFPUK7aZUpTbkRmKQuACKURR',
+    ]
```
TurboTurtle, line 48: I know we have similar stance in other tests, but this got me thinking - should we not be checking for the actual scrubbed value instead of (or as well as?) the static strings not being present? I.E. we ensure that `token_key` for example is literally written as `*********`?

sos/report/plugins/curtin.py
```diff
@@ -0,0 +1,57 @@
...
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, IndependentPlugin
+
+
+class Curtin(Plugin, IndependentPlugin):
+
+    short_desc = 'GRUB bootloader'
```
TurboTurtle, line 14: Maybe something to distinguish it from the `grub2` plugin a bit more when looking at the `--list-plugins` output?

tests/report_tests/plugin_tests/curtin/curtin-install.log
arif-ali: There is no package per say that is installed by default.  The package is typically installed in the diskless image that MAAS provides for the purpose of bootstrapping the node.

Once rebooted, curtin is not installed on the target disk, but you still have the relevant logs and configs to look through for any troubleshooting purposes.

This plugin was created so that we can troubleshoot, and the key triggers to see that curtin was used is the yaml file as per the `files` tuple

sos/report/plugins/curtin.py
arif-ali: oops, total typo from our side; will fix in next iteration

tests/report_tests/plugin_tests/curtin/curtin.py
arif-ali: I've updated the tests, hopefully this would now test for most scenarios, as well as looking at the keys too

pmoravec (approved):

TurboTurtle (approved):

PR #3678, opened 2024-06-17
[container_log] Get rotated logs

### LINKED ISSUE #3677: There are no plugin to gater K8s/OCP rotated logs
On Kubernetes/OpenShift nodes when the logs are rotated there is no way to get them in an sos report.

It is true than in most cases there is a centralized log solution, but in some other it is convenient to be able to get these logs.

All rotated logs are present under `/var/log/pods` which cannot be retrieved any way:

- Not using `container_logs` plugin, since it only gathers the logs symlinked in `/var/log/containers`
- Not using the `openshift` plugin, as it is only able to get the current logs using `oc logs`.

Akrog:
In Kubernetes/OpenShift rotated logs have no symbolic link in /var/log/containers and they cannot be retrieved using `oc logs` either, so there is no way to get these rotated logs in a SOS report.

This patch proposes a new functionality for the `container_log` plugin to retrieve rotated logs for the containers.

Since there may be lots of rotated logs we also add the `maxage` option.

The plugin will not only download the `/var/log/pods` files but also the `/var/log/container` symlinks.

Closes #3677
Signed-off-by: Gorka Eguileor <geguileo@redhat.com>

---

### DIFF
--- a/sos/report/plugins/container_log.py
+++ b/sos/report/plugins/container_log.py
@@ -9,7 +9,7 @@
 # See the LICENSE file in the source distribution for further information.
 
 import os
-from sos.report.plugins import Plugin, IndependentPlugin
+from sos.report.plugins import Plugin, IndependentPlugin, PluginOpt
 
 
 class ContainerLog(Plugin, IndependentPlugin):
@@ -17,18 +17,29 @@ class ContainerLog(Plugin, IndependentPlugin):
     short_desc = 'All logs under /var/log/containers'
     plugin_name = 'container_log'
     logdir = '/var/log/containers/'
+    poddir = '/var/log/pods/'
+    rotated_dirs = [poddir + '*/*.log.*', poddir + '*/*/*.log.*']
     files = (logdir, )
+
+    option_list = [
+        PluginOpt('rotated', default=False, val_type=bool,
+                  desc='also get rotated logs from /var/log/pods'),
+    ]
 
     def setup(self):
         if self.get_option('all_logs'):
             self.add_copy_spec(self.logdir)
+            if self.get_option('rotated'):
+                self.add_copy_spec(self.rotated_dirs)
         else:
-            self.collect_subdirs()
+            self.collect_subdirs(self.logdir, '*.log')
+            if self.get_option('rotated'):
+                self.collect_subdirs(self.poddir, '*.log.*')
 
-    def collect_subdirs(self, root=logdir):
+    def collect_subdirs(self, root, glob):
         """Collect *.log files from subdirs of passed root path
         """
         for dir_name, _, _ in os.walk(root):
-            self.add_copy_spec(self.path_join(dir_name, '*.log'))
+            self.add_copy_spec(self.path_join(dir_name, glob))
 
 # vim: set et ts=4 sw=4 :


gibizer (commented):
Thanks @Akrog

fmount:
+1 on this change, thank you @Akrog

TurboTurtle (changes requested):
I don't like the idea of adding yet another "get logs for containers" plugin from the start. On top of that though, this should be an addition to the `openshift` and `kubernetes` plugins as a simple `add_copy_spec()` using the existing `maxage` parameter.

If the rotated logs are written to disk, there's no reason the `openshift` plugin can't just collect those outside of the `oc logs` command.

Akrog:
> I don't like the idea of adding yet another "get logs for containers" plugin from the start. On top of that though, this should be an addition to the `openshift` and `kubernetes` plugins as a simple `add_copy_spec()` using the existing `maxage` parameter.
> 
> If the rotated logs are written to disk, there's no reason the `openshift` plugin can't just collect those outside of the `oc logs` command.

Makes sense not adding yet another plugin.

Would it be ok to add it as an option to the `container_log` plugin instead?

I'm saying it because that plugin is already getting `/var/log/containers` which in many cases are just symlinks to `/var/log/pods`.

bogdando:
well done, thanks a lot!

TurboTurtle:
> > I don't like the idea of adding yet another "get logs for containers" plugin from the start. On top of that though, this should be an addition to the `openshift` and `kubernetes` plugins as a simple `add_copy_spec()` using the existing `maxage` parameter.
> > If the rotated logs are written to disk, there's no reason the `openshift` plugin can't just collect those outside of the `oc logs` command.
> 
> Would it be ok to add it as an option to the `container_log` plugin instead?
> 
> I'm saying it because that plugin is already getting `/var/log/containers` which in many cases are just symlinks to `/var/log/pods`.


Yeah, that makes sense.

jcastill:
@Akrog can you squash these two commits into one please?

Akrog:
@jcastill Done.  I usually try to do 1 thing on each patch, but each project is different.  :-)

pmoravec (approved):

sos/report/plugins/container_log.py
```diff
@@ -9,26 +9,51 @@
...
+        PluginOpt('maxage', default=None, val_type=int,
+                  desc='gather only logs with `mtime` not older than this many'
+                  ' hours')
+    ]
+
     def setup(self):
+        maxage = self.get_option('maxage')
+        sizelimit = None
+        # Remove size limit from containers logs when getting rotated, since
+        # they are just symlinks to the pods logs
```
TurboTurtle, line 36: Our sizelimiting takes the link's destination into account when checking sizes for collections, so I'm not sure what the implication is here? We definitely do not want to have a wide open "collect all logs regardless of sizes" default posture here. We've had that bite us in the rear in the past several times - no one likes several GB sos reports.

sos/report/plugins/container_log.py
```diff
@@ -9,26 +9,51 @@
...
+            self.collect_subdirs(sizelimit=sizelimit, maxage=maxage)
 
-    def collect_subdirs(self, root=logdir):
+    def collect_subdirs(self, root=logdir, sizelimit=None, maxage=None):
         """Collect *.log files from subdirs of passed root path
         """
         for dir_name, _, _ in os.walk(root):
-            self.add_copy_spec(self.path_join(dir_name, '*.log'))
+            self.add_copy_spec(self.path_join(dir_name, '*.log'),
+                               sizelimit=sizelimit,
```
TurboTurtle, line 56: Yeah, not ignoring sizelimit entirely as a default behavior.

sos/report/plugins/container_log.py
```diff
@@ -9,26 +9,51 @@
...
+    poddir = '/var/log/pods/'
     files = (logdir, )
 
+    option_list = [
+        PluginOpt('rotated', default=False, val_type=bool,
+                  desc='get all logs from /var/log/pods regardless of size, '
+                  'which will include all rotated logs'),
+        PluginOpt('maxage', default=None, val_type=int,
+                  desc='gather only logs with `mtime` not older than this many'
+                  ' hours')
```
TurboTurtle, line 29: I think if we do anything here, we should add a `maxage` (or `since`) -esque options to all plugins by default, like we do for `timeout` and `postproc`.

sos/report/plugins/container_log.py
Akrog: I tried, and failed, to explain it in the PR description.

The idea is that if you set the `rotated` option but you don't change the limits at all you are going to either get the exact same content as without the option (because like you said getting the symlinks checks the actual destination files) or we are going to get logs from the pods that don't have a symlink (not necessarily rotated logs), in both cases it seems wrong.

Another option is for the `rotated` option to ignore all the `.log` files from the pods directory and only get rotated logs instead.

This later options sounds reasonable to me and in line with the option name.  I'll update the PR.

sos/report/plugins/container_log.py
Akrog: Adding that global feature in this PR doesn't feel right, so I'll submit another PR for that one.

Akrog: Does this PR look reasonable? https://github.com/sosreport/sos/pull/3681

TurboTurtle (approved):

sos/report/plugins/container_log.py
```diff
@@ -9,26 +9,35 @@
...
     plugin_name = 'container_log'
     logdir = '/var/log/containers/'
+    poddir = '/var/log/pods/'
+    rotated_dirs = [poddir + '*/*.log.*', poddir + '*/*/*.log.*']
     files = (logdir, )
 
+    option_list = [
+        PluginOpt('rotated', default=False, val_type=bool,
+                  desc='also get rotated logs from /var/log/pods'),
+    ]
```
TurboTurtle, line 27: Just noticed this on a second pass, but I'm assuming this should be dropped with the intent being that the rotated logs are collected by `-k container_log.all-logs=true` once #3681 is merged?

Akrog: I'm stupid, you are right.
This PR is changing existing behavior and it's adding a plugin option that is not being used.
I'll make the rotated log gathering conditional to the `rotated` plugin option.

PR #3681, opened 2024-06-21
[plugins] maxage and all_logs on all plugins

Akrog:
This patch makes 2 options settable globally and per-plugin: `maxage` and `all_logs`.

Examples:

```
sos report --all-logs --maxage=1
sos report -k container_log.all-logs=on,container_log.maxage=1
sos report --all-logs --maxage=1 -k container_log.all-logs=off,container_log.maxage=2
```

This adds flexibility to what users want to retrieve.

---

### DIFF
--- a/sos/report/__init__.py
+++ b/sos/report/__init__.py
@@ -122,6 +122,7 @@ class SoSReport(SoSComponent):
         'cmd_timeout': TIMEOUT_DEFAULT,
         'profiles': [],
         'since': None,
+        'maxhours': None,
         'verify': False,
         'allow_system_changes': False,
         'usernames': [],
@@ -200,6 +201,12 @@ class SoSReport(SoSComponent):
                                 help="Escapes archived files older than date. "
                                      "This will also affect --all-logs. "
                                      "Format: YYYYMMDD[HHMMSS]")
+        report_grp.add_argument("--maxhours", action="store",
+                                dest="maxhours", default=None, type=int,
+                                help="Escapes archived files older (according "
+                                "to `mktime`) than this many hours. This will "
+                                "also affect --all-logs (sets the default for "
+                                "all plugins)")
         report_grp.add_argument("--build", action="store_true",
                                 dest="build", default=False,
                                 help="preserve the temporary directory and do "
@@ -904,6 +911,12 @@ class SoSReport(SoSComponent):
                         opt.value = True
 
     def _set_tunables(self):
+        # Set plugin option's defaults to the global argument values
+        for pluginname, plugin in self.loaded_plugins:
+            for optname in plugin.options:
+                if hasattr(self.opts, optname):
+                    plugin.options[optname].value = getattr(self.opts, optname)
+
         if self.opts.plugopts:
             opts = {}
             for opt in self.opts.plugopts:
@@ -1063,12 +1076,13 @@ class SoSReport(SoSComponent):
                         val = TIMEOUT_DEFAULT
                 if opt.name == 'postproc':
                     val = not self.opts.no_postproc
-                self.ui_log.info(f"{opt.param_name:<25} {val:<15} {opt.desc}")
+                self.ui_log.info(
+                    f"{opt.param_name:<25} {val!s:<15} {opt.desc}")
             self.ui_log.info("")
 
             self.ui_log.info(_("The following plugin options are available:"))
             for opt in self.all_options:
-                if opt.name in ('timeout', 'postproc', 'cmd-timeout'):
+                if opt.name in _defaults:
                     if opt.value == opt.default:
                         continue
                 # format option value based on its type (int or bool)
--- a/sos/report/__init__.py
+++ b/sos/report/__init__.py
@@ -940,13 +940,17 @@ class SoSReport(SoSComponent):
 
             for plugname, plug in self.loaded_plugins:
                 if plugname in opts:
-                    for opt in opts[plugname]:
-                        if opt not in plug.options:
-                            self.soslog.error(f'no such option "{opt}" for '
+                    # Map from parameter names to internal option names
+                    plug_param_map = {o.param_name: o.name
+                                      for o in plug.options.values()}
+                    for param in opts[plugname]:
+                        opt = plug_param_map.get(param)
+                        if not opt:
+                            self.soslog.error(f'no such option "{param}" for '
                                               f'plugin ({plugname})')
                             self._exit(1)
                         try:
-                            plug.options[opt].set_value(opts[plugname][opt])
+                            plug.options[opt].set_value(opts[plugname][param])
                             self.soslog.debug(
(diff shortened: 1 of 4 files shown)


sos/report/plugins/__init__.py
```diff
@@ -1674,6 +1685,9 @@ def add_copy_spec(self, copyspecs, sizelimit=None, maxage=None,
         if self.get_option('since'):
             since = self.get_option('since')
 
+        # User's plugin option argument takes precedence over plugin passed arg
+        maxage = self.get_option('maxage') or maxage
+
```
pmoravec, line 1690: Shall not we unify the code here? Both manipulations perform the same at the end, the unification would improve clarity of code.

sos/report/__init__.py
```diff
@@ -200,6 +201,12 @@ def add_parser_options(cls, parser):
                                 help="Escapes archived files older than date. "
                                      "This will also affect --all-logs. "
                                      "Format: YYYYMMDD[HHMMSS]")
+        report_grp.add_argument("--maxage", action="store",
+                                dest="maxage", default=None, type=int,
+                                help="Escapes archived files older (according "
+                                "to `mktime`) than this many hours. This will "
+                                "also affect --all-logs (sets the default for "
+                                "all plugins)")
```
pmoravec, line 209: Would it be clear for users the option takes values in *hours*? Esp. if `--since` is primarily in days (default expected format is a day, optionally time)? Should not we rename it to, I dont know, `--maxhours`? (that sounds badly to me, I like the `--maxage` more).

I can easily be wrong here. I really dont know what users' perception would be here.

Akrog: I like the suggested change to `maxhours`, as it has the benefit of including the unit in the name and not force people to look at the description to figure it out.

I'll make the change.

pmoravec:
I meant that

            since = None
            if self.get_option('since'):
                since = self.get_option('since')
            
            # User's plugin option argument takes precedence over plugin passed arg
            maxhours = self.get_option('maxhours') or maxhours

applies same logical change to either `since` or `maxhours` option. For the sake of code coherence, we should use same construction in either assignment, so let have there either:

            since = self.get_option('since') or None
            
            # User's plugin option argument takes precedence over plugin passed arg
            maxhours = self.get_option('maxhours') or maxhours

or:

            since = None
            if self.get_option('since'):
                since = self.get_option('since')
            
            # User's plugin option argument takes precedence over plugin passed arg
            if self.get_option('maxhours'):
                maxhours = self.get_option('maxhours')

The first construction is more concise, the second more explicit. I like the first more but 2nd is fully OK for sure.

sos/report/__init__.py
```diff
@@ -904,6 +911,12 @@ def _set_all_options(self):
                         opt.value = True
 
     def _set_tunables(self):
+        # Set plugin option's defaults to the global argument values
+        for pluginname, plugin in self.loaded_plugins:
+            for optname in plugin.options:
+                if hasattr(self.opts, optname):
+                    plugin.options[optname].value = getattr(self.opts, optname)
```
TurboTurtle, line 918: I'm not seeing why this is needed? We have the defaults for the plugin options set when they're instantiated as a `PluginOpt`, and these shouldn't differ at this point.

TurboTurtle: Nevermind, I see it now. The other options to date have had specific handling to them for default vs. plugin-specific values due to the need to do some dancing with magic numbers. We don't need to do that here, so we're good.

TurboTurtle:
In a quick test, this change works for `maxhours`, but does not work for `all_logs` - the plugin only gets the global value, which is due to this in `Plugin`:

https://github.com/sosreport/sos/blob/7b0d5a4355e380e9d662754efe0eb9b786e9f217/sos/report/plugins/__init__.py#L1544-L1564

sos/report/plugins/__init__.py
```diff
@@ -585,7 +585,18 @@ def __init__(self, commons):
...
     def get_default_plugin_opts(self):
+        # The default of options that have a global argument counterpart with
+        # the same name must use the same default in both places.
         return {
+            'maxhours': PluginOpt(
+                'maxhours', default=None, val_type=int,
+                desc='Escapes archived files older (according to `mktime`) '
+                'than this many hours'
+            ),
+            'all_logs': PluginOpt(
```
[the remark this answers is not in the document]
Akrog: I'm new to this, but I assumed looking at the [kernel plugin `with-timer` definition](https://github.com/sosreport/sos/blob/e48a41dd6da6178cc04ce6a6d7685814195e4384/sos/report/plugins/kernel.py#L39) and how it is [accessed as `with-timer` and not `with_timer`](https://github.com/sosreport/sos/blob/e48a41dd6da6178cc04ce6a6d7685814195e4384/sos/report/plugins/kernel.py#L144) that there is no translation on the `PluginOpt`.

I looked a bit into the `get_option` and `PluginOpt` and didn't find any translation mechanism so I went with the solution that required fewer code changes.

But it is true that it would be weird to call it `--all-logs` only to have to use `all_logs` in the plugins.

I'll see if I can use `all-logs` everywhere.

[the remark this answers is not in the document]
Akrog: Done

Akrog:
> In a quick test, this change works for `maxhours`, but does not work for `all_logs` - the plugin only gets the global value, which is due to this in `Plugin`:
> 
> https://github.com/sosreport/sos/blob/7b0d5a4355e380e9d662754efe0eb9b786e9f217/sos/report/plugins/__init__.py#L1544-L1564

Done, I implemented it in what I consider the safest way, and it also adds a missing feature: different variable and user facing parameter name for plugin options.

Alternatively the `all-logs` command line argument could be changed to save the parameter in `all-logs` instead of `all_logs` and all code changed to access it with `getattr` instead of using `.all_logs`.

TurboTurtle:
Sorry for the silence here, cycling around on it now.

The `all-logs` vs `all_logs` convention is due to python variable identifiers not allowing hyphen characters. We aligned to using hyphens in the user-facing CLI because that's typically what users expect, and it is the prevailing syntax for argparse and what it does automatically for name conversions.

Internally, developers just need to know that hyphens become underscores when accessing the associated variables, which is the case for any project using argparse and directly accessing the parsed options.

So, we're good with the original intent of the PR - expanding all-logs and maxage/maxhours to all plugins on a potentially individual basis, but we don't need to mess around with internal references to option names.

pmoravec:
I am not sure of the sttaus of this PR - are we waiting for a change from author, or for a review? (I plan to review the current PR within a day or two, though it would be great if PR is rebased to current main - there are conflicts..)

pmoravec:
I like it. Maybe worth adding a (report_)test? (I can come up with something if preferred)

arif-ali:
@Akrog any update on this PR at all?

Issue #3683, opened 2024-06-21
[snaps] when using `--all-logs` some command outputs will not be collected


arif-ali:
Working with one of my engineering team members we discovered that snap command outputs will not be collected on the filesystem due to the strict confinement of snaps and hence not having permissions to the destination. A lot of snap related have been added over the past 6 months, and first time we've discovered this issue.

The output from STDOUT to files was introduced in PR #2729 to improve the memory management and speed of collection of the corresponding commands.

Some of the data around this is documented in Launchpad [1]

One possible workaround is that we could have a disable this feature with all logs if the command being run is a command from a snap, but I'll think of an idea on how we can achieve this moving forward.

[1] https://launchpad.net/bugs/2069941

TurboTurtle:
Just to clarify, you're saying this is happening when _sos_ is installed as a snap, yes? Not that the commands we are trying to collect are installed from snaps?

arif-ali:
So, this is when a package is a snap, and hence the command is being run from that snap package.

i.e. the `maas` plugin is triggered, the `maas status` command will run, but will not be able to get the output directly to the file due to snap confinement restrictions

PR #3684, opened 2024-06-22
[common] handle all_logs when snap package is installed

arif-ali:
Add a new flag to add_cmd_output to suggest a snap_cmd, so that sos
does not direct the command directly to a file.

Resolves: #3683

---

### DIFF
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -2007,8 +2007,11 @@ class Plugin():
             kwargs['priority'] = 10
         if 'changes' not in kwargs:
             kwargs['changes'] = False
-        if self.get_option('all_logs') or kwargs['sizelimit'] == 0:
+        if (not getattr(SoSCommand(**kwargs), "snap_cmd", False) and
+           (self.get_option('all_logs') or kwargs['sizelimit'] == 0)):
             kwargs['to_file'] = True
+        if "snap_cmd" in kwargs:
+            kwargs.pop("snap_cmd")
         soscmd = SoSCommand(**kwargs)
         self._log_debug("packed command: " + soscmd.__str__())
         for _skip_cmd in self.skip_commands:
@@ -2073,7 +2076,7 @@ class Plugin():
                        sizelimit=None, pred=None, subdir=None,
                        changes=False, foreground=False, tags=[],
                        priority=10, cmd_as_tag=False, container=None,
-                       to_file=False, runas=None):
+                       to_file=False, runas=None, snap_cmd=False):
         """Run a program or a list of programs and collect the output
 
         Output will be limited to `sizelimit`, collecting the last X amount
@@ -2149,6 +2152,9 @@ class Plugin():
 
         :param runas: Run the `cmd` as the `runas` user
         :type runas: ``str``
+
+        :param snap_cmd: Are the commands being run from a snap?
+        :type snap_cmd: ``bool``
         """
         if isinstance(cmds, str):
             cmds = [cmds]
@@ -2177,7 +2183,7 @@ class Plugin():
                                  changes=changes, foreground=foreground,
                                  priority=priority, cmd_as_tag=cmd_as_tag,
                                  to_file=to_file, container_cmd=container_cmd,
-                                 runas=runas)
+                                 runas=runas, snap_cmd=snap_cmd)
 
     def add_cmd_tags(self, tagdict):
         """Retroactively add tags to any commands that have been run by this
--- a/sos/report/plugins/grafana.py
+++ b/sos/report/plugins/grafana.py
@@ -25,7 +25,7 @@ class Grafana(Plugin, IndependentPlugin):
             log_path = "/var/snap/grafana/common/data/log/"
             config_path = "/var/snap/grafana/current/conf/grafana.ini"
 
-            self.add_cmd_output("snap info grafana")
+            self.add_cmd_output("snap info grafana", snap_cmd=True)
         else:
             grafana_cli = "grafana-cli"
             log_path = "/var/log/grafana/"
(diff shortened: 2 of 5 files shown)


arif-ali:
Putting this as an idea, but, maybe we can do something in check_enabled instead to simplify this

dnegreira:
I like the idea overall, but I think the logic could be improved a bit, left some comments on it

sos/report/plugins/__init__.py
```diff
@@ -999,6 +1001,18 @@ def is_installed(self, package_name):
...
+        """Is the package $package_name installed and is a snap?
+
+        :param package_name:    The name of the package to check
+        :type package_name:     ``str``
+
+        :returns: ``True`` id the package is installed, else ``False``
+        :rtype: ``bool``
+        """
+        pkg = self.policy.package_manager.pkg_by_name(package_name)
+        self.is_snap = (pkg and pkg['pkg_manager'] == 'snap')
```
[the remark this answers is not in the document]
arif-ali: yeah, I was pondering that idea before, and was trying different ideas. Agreed that we should mimic is_installed.

there could be a scenario where `self.policy.package_manager.pkg_by_name(package_name)` doesn't actually return anything, so testing for `pkg` is important, otherwise we'll get `keyErrror` with `pkg_manager` index not existing

arif-ali:
Put in 2 ideas

1. first commit still has some legs, and we standardise how we check for snaps, and just use one variable in plugins. Added option to disable `to_file` as well. The disabling is not ideal, as you want to ensure we don't use too much of the memory
1. Second commit removes this, and we export the command to `STDOUT` and grab it via `cat `, this works for snap confinement case as well, which I have tested locally.

if we decide the second option is better, then we could split this PR into 2, and have one for standardising `is_snap_installed()` and the other to resolve the confinement issue using `cat`

thoughts?

arif-ali:
alternatively, we can the change below (after the recent PR) but that defeats the purpose of the original commit of having commands outputting to file with `all_logs`

```
diff --git a/sos/report/plugins/__init__.py b/sos/report/plugins/__init__.py
index 74ade1fb..2009a2cf 100644
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -1990,7 +1990,8 @@ class Plugin():
             kwargs['priority'] = 10
         if 'changes' not in kwargs:
             kwargs['changes'] = False
-        if self.get_option('all_logs') or kwargs['sizelimit'] == 0:
+        if not self.is_snap and (self.get_option('all_logs') or
+                                 kwargs['sizelimit'] == 0):
             kwargs['to_file'] = True
         soscmd = SoSCommand(**kwargs)
         self._log_debug("packed command: " + soscmd.__str__())
```

sos/report/plugins/__init__.py
```diff
@@ -2007,8 +2007,11 @@ def _add_cmd_output(self, **kwargs):
             kwargs['priority'] = 10
         if 'changes' not in kwargs:
             kwargs['changes'] = False
-        if self.get_option('all_logs') or kwargs['sizelimit'] == 0:
+        if (not getattr(SoSCommand(**kwargs), "snap_cmd", False) and
+           (self.get_option('all_logs') or kwargs['sizelimit'] == 0)):
```
pmoravec, line 2011: So the impact of the `snap_cmd` option will be "dont store directly to file even when all_logs is set"? That is independent on snap packaging, so I suggest changing option name.

E.g. why not to use `to_file` directly? We would just have to change condition at https://github.com/sosreport/sos/blob/main/sos/report/plugins/__init__.py#L2010 .

sos/report/plugins/__init__.py
```diff
@@ -2007,8 +2007,11 @@ def _add_cmd_output(self, **kwargs):
...
             kwargs['to_file'] = True
+        if "snap_cmd" in kwargs:
+            kwargs.pop("snap_cmd")
```
pmoravec, line 2014: Why dropping the argument (and not drop any other)? I understand we wont use the argument further more - but I am juts curious why we dont the same for other arguments, then :)

arif-ali: we drop it because the argument is passed to the other functions, and `pylint` complains that the argument is not being used, so we drop it, once we use it

sos/report/plugins/__init__.py
arif-ali: we could use to_file directly to be set false for snap specific, but we can't afford to have `to_file` to be true if the command is being run is a snap command. We'd need a different test here for this then, if we drop the `snap_cmd` paramater. Ultimately `to_file` is False by default, so somehow we need to detect is the command being run is a snap or not, maybe check where the command is coming from, i.e. `/snap/bin`.

We don't want to do a blanket all commands to be not run to file, as journalctl wouldn't be snapped, and hence can write to file.

So, we could have for example

sunbeam cluster list <--snap
journalctl <-- not snap

now, the latter command, we can run with `to_file=True`, but the former we can't

sos/report/plugins/__init__.py
pmoravec: I see, makes sense.

sos/report/plugins/__init__.py
pmoravec: I see. One part of my concern was also to the parameter name itself, trying to keep `Plugin` independent on packaging implementation (same concernt I would raise against, say, `is_rpm` parameter name). But I cant figure out some generic enough name (`force_to_file`? too long..) that I would suggest to use.

So ACK from me.

pmoravec (approved):

TurboTurtle (approved):
As mentioned on IRC, I have some concerns around maintainability and the potential for `snap_cmd=True` to be marked for commands that aren't coming from a snap in a given installation - but for the state of sos today at least, that doesn't appear to be an actual potential issue.

Ack for the current PR, but we'll want to keep this in mind going forward so that we don't drown ourselves in a sort of "snap_cmd dance" for more generalized plugins.

PR #3685, opened 2024-06-22
[fail2ban] Add new plugin for fail2ban

slapcat:
New plugin for fail2ban daemon.

---

### DIFF
--- a/sos/report/plugins/fail2ban.py
+++ b/sos/report/plugins/fail2ban.py
@@ -1,0 +1,30 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, IndependentPlugin
+
+
+class Fail2Ban(Plugin, IndependentPlugin):
+
+    short_desc = "Fail2Ban daemon"
+
+    plugin_name = "fail2ban"
+    profiles = ('security',)
+    packages = ('fail2ban', 'fail2ban-server')
+    servicess = ('fail2ban',)
+
+    def setup(self):
+
+        self.add_copy_spec([
+            '/etc/fail2ban',
+        ])
+
+        self.add_cmd_output([
+            'fail2ban-client status',
+            'fail2ban-client banned',
+        ])


sos/report/plugins/fail2ban.py
```diff
@@ -0,0 +1,40 @@
...
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, DebianPlugin, UbuntuPlugin
+
+
+class Fail2Ban(Plugin, DebianPlugin, UbuntuPlugin):
```
TurboTurtle, line 12: We can make this `IndependentPlugin`  so that this plugin is enabled for RHEL-family distros as well. Taking a quick look, the collection patterns would be the same, just an additional package name would be needed in the `packages` tuple.

sos/report/plugins/fail2ban.py
```diff
@@ -0,0 +1,40 @@
...
+
+    short_desc = "Fail2Ban daemon"
+
+    plugin_name = "fail2ban"
+    profiles = ('security',)
+    packages = ('fail2ban',)
```
TurboTurtle, line 18: This package exists on RHEL, but so does `fail2ban-server` , which can be installed standalone.

sos/report/plugins/fail2ban.py
```diff
@@ -0,0 +1,40 @@
...
+
+    common_dir = '/etc/fail2ban'
+
+    def setup(self):
+
+        self.add_service_status('fail2ban')
```
TurboTurtle, line 24: Setting `services = ('fail2ban',)` in the constructor will grab this, as well as the journal for the service (and servers as another enablement trigger in case the packaging is different on some other distros).

sos/report/plugins/fail2ban.py
```diff
@@ -0,0 +1,40 @@
...
+            f'{self.common_dir}/jail.d',
+        ])
+
+        self.add_cmd_output([
+            'fail2ban-client status',
+            'fail2ban-client banned',
+        ])
+
+# vim: et ts=4 sw=4
+
```
TurboTurtle, line 40: The `flake8` linting tests are failing on this trailing newline.

TurboTurtle (commented):
Overall looks good, just a couple suggestions.

sos/report/plugins/fail2ban.py
slapcat: Thanks for the quick feedback @TurboTurtle ! I updated this part as requested.

sos/report/plugins/fail2ban.py
slapcat: Added.

sos/report/plugins/fail2ban.py
slapcat: Added the `services` line and removed `add_service_status`.

sos/report/plugins/fail2ban.py
slapcat: Removed.

arif-ali:
Just doubling check, any config files we are collecting here, any passwords that needs obfuscating?

Also the trail line is still there, in the recent push

slapcat:
> Just doubling check, any config files we are collecting here, any passwords that needs obfuscating?

There should not be any passwords, but the `fail2ban-client banned` output will contain IP addresses. I know is some circumstances these would be obfuscated, but I don't think it would make sense here since it would be valuable diagnostic data for troubleshooting fail2ban issues.

arif-ali:
overall, looks good to me, before we can merge this, we need the commits to be squashed, so that we have one commit for the PR. Once you do that we can move this one forward

arif-ali:
just coming back to this PR< is there any reason why we just don't do `/etc/fail2ban` and the reason for specifically pointing out those directories? As per @pmoravec issue raised in #3686 we have potential 2 identical items being collecting, and hence could have an issue here

slapcat:
Nothing sensitive should be in `/etc/fail2ban`, so it is okay to grab the entire directory. I've updated the code to reflect that and address the issue raised in https://github.com/sosreport/sos/issues/3686m

pmoravec (approved):

sos/report/plugins/fail2ban.py
```diff
@@ -0,0 +1,36 @@
...
+    common_dir = '/etc/fail2ban'
+
+    def setup(self):
+
+        self.add_copy_spec([
+            f'{self.common_dir}/fail2ban.*',
+            f'{self.common_dir}/jail.*',
+            f'{self.common_dir}/filter.d',
+            f'{self.common_dir}/action.d',
+            f'{self.common_dir}/jail.d',
```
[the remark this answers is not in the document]
arif-ali: that's already been removed; I think you've commented on an older commit :)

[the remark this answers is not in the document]
pmoravec: Eh, blame github WebUI :). I reviewed current code changes, clicked to review changes and Submit review (as usual). Since I had some unresolved comment on an outdated code change, WebUI repeated my comment, apparently..? /o\

ACK to the current PR, just to be sure :)

PR #3687, opened 2024-06-25
[maas] improve/refactor MAAS plugin

troyanov:
Changes:
- plugin_timeout = 1800 (because journal collection takes time)
- collect journal for all MAAS related services (by default since yesterday)
- mask passwords in (conf|yaml|yml|toml) files with `do_path_regex_sub`
- collect only recent syslog data for machines (if `--all-logs` was not provided) 

snap:
- collect all `.conf` and `.yaml` files
- exclude files containing secrets and keys under `/var/snap/maas/**`

deb:
- add missing service `maas-temporal-worker` introduced in MAAS 3.5
- collect `/var/lib/maas/temporal`
- use glob pattern to collect `.conf` files
- exclude secrets and keys under `/var/lib/maas/**` and `/etc/maas/**`
- remove `/var/log/upstart/maas-*` as it was removed in xenial
- remove `maas-region dumpdata` as it can expose sensitive information

---

### DIFF
--- a/sos/report/plugins/maas.py
+++ b/sos/report/plugins/maas.py
@@ -8,19 +8,29 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.report.plugins import Plugin, UbuntuPlugin, PluginOpt
+import os
+from sos.report.plugins import Plugin, UbuntuPlugin
 
 
-class Maas(Plugin, UbuntuPlugin):
+class MAAS(Plugin, UbuntuPlugin):
 
-    short_desc = 'Ubuntu Metal-As-A-Service'
+    short_desc = 'MAAS | Metal as a Service'
 
     plugin_name = 'maas'
+    plugin_timeout = 1800
     profiles = ('sysmgmt',)
-    packages = ('maas', 'maas-common')
 
-    services = (
-        # For the deb:
+    packages = (
+        'maas',
+        'maas-region-api',
+        'maas-region-controller',
+        'maas-rack-controller',
+        'maas-agent',
+    )
+
+    _services = (
+        'maas-agent',
+        'maas-apiserver',
         'maas-dhcpd',
         'maas-dhcpd6',
         'maas-http',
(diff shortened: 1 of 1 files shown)


sos/report/plugins/maas.py
```diff
@@ -11,44 +11,39 @@
...
-class Maas(Plugin, UbuntuPlugin):
+class MAAS(Plugin, UbuntuPlugin):
 
-    short_desc = 'Ubuntu Metal-As-A-Service'
+    short_desc = 'MAAS | Metal as a Service'
 
     plugin_name = 'maas'
+    plugin_timeout = 1800
     profiles = ('sysmgmt',)
-    packages = ('maas', 'maas-common')
```
arif-ali, line 20: I think its worthwhile keeping this here, I am hoping with the work we've been doing with the `--all-logs` issue, it will parse this and will improve the snap package detection in #3684

sos/report/plugins/maas.py
```diff
@@ -67,58 +62,59 @@ def _is_snap_installed(self):
             return maas_pkg['pkg_manager'] == 'snap'
         return False
 
+    def _snap_collect(self):
+        self.add_cmd_output([
+            'snap info maas',
+            'maas status'
```
arif-ali, line 68: minor nit, comma at the end

sos/report/plugins/maas.py
```diff
@@ -67,58 +62,59 @@ def _is_snap_installed(self):
...
+
+        self.add_copy_spec([
+            "/etc/maas/**/*.conf",
+            "/etc/maas/**/*.yaml",
+            "/etc/maas/preseeds",
+            "/var/lib/maas/**/*.conf",
+            "/var/lib/maas/dhcp",
+            "/var/lib/maas/rsyslog",
+            "/var/lib/maas/temporal",
+            "/var/log/maas",
```
arif-ali, line 110: with the improvements we've been discussing, is it worth splitting this apart so that we collect recent logs normally, and we collect all when we specify `--all-logs`

sos/report/plugins/maas.py
```diff
@@ -67,58 +62,59 @@ def _is_snap_installed(self):
...
+        ])
+
+        self.add_forbidden_path([
+            "/var/snap/maas/**/*.key",
+            "/var/snap/maas/**/*.pem",
+            "/var/snap/maas/**/secret",
+        ])
+
+        self.add_copy_spec([
+            "/var/snap/maas/common/log",
```
arif-ali, line 78: with the improvements we've been discussing, is it worth splitting this apart so that we collect recent logs normally, and we collect all when we specify `--all-logs`

arif-ali (commented):
some comments based on our discussions

sos/report/plugins/maas.py
troyanov: Thanks for the feedback! 

So we always want to copy recent `.log` files, but _rotated logs_ (e.g. `.log.1.gz`) only when `--all-logs` is provided?
 
```python
    self.add_copy_spec([
        "/var/log/maas/",
    ])
else:
    self.add_copy_spec([
        "/var/log/maas/*.log",
    ])
```

sos/report/plugins/maas.py
troyanov: Returned `packages` back with a more explicit list.

sos/report/plugins/maas.py
arif-ali: I don't have a deb implementation myself, but something like the following would work for the snap version

```
    self.add_copy_spec([
        "/var/snap/maas/common/log/",
    ])
else:
    self.add_copy_spec([
        "/var/snap/maas/common/log/*.log",
        "/var/snap/maas/common/log/*/*.log",
    ])
```

We may need to look at the `rsyslog` dir, on how best to accommodate that as that has folders already with date timestamps. Depending on the number of nodes in the environment, having all of the logs for a non `--all-logs` could be overwhelming. If we could somehow get the latest directory from that, then that would work in the `else` section. Thoughts? You're the MAAS expert, I let you decide :)

sos/report/plugins/maas.py
```diff
@@ -11,44 +11,47 @@
...
+    plugin_timeout = 1800
     profiles = ('sysmgmt',)
-    packages = ('maas', 'maas-common')
+
+    packages = (
+        'maas',
+        'maas-region-api',
+        'maas-region-controller',
+        'maas-rack-controller',
+        'maas-agent',
```
cpg1111, line 27: should we also include maas-netmon or does it make sense to leave it out since it will eventually be part of the agent?

arif-ali: The whole ethos here, it is a trigger. So, as long as one of these packages are there, then the plugin would be triggered.

troyanov: Right.
I decided to remove `maas-common` and have an explicit agent/rack/region so it is easier to understand when this plugin is triggered automatically.

sos/report/plugins/maas.py
troyanov: Added some logic to handle `rsyslog`. However it is applicable only to MAAS before 3.5 (as in 3.5 we keep everything in the journal)

sos/report/plugins/maas.py
troyanov: Updated to keep only latest `*.log` (not to include `*.log.1` and others)
Added logic to collect only recent syslog data for each machine (if `--all-logs` not provided)

sos/report/plugins/maas.py
arif-ali: ack, sounds good to me

sos/report/plugins/maas.py
```diff
@@ -8,133 +8,168 @@
...
+            ]
+
+            if not subdirs:
+                continue
+
+            recent.append(sorted(
+                subdirs, key=lambda f: os.stat(f).st_mtime, reverse=True)[0]
+            )
+
+        return recent
```
pmoravec, line 80: I think using `glob.glob` can be shorter here but since you need to apply some logic (take most recent file from each subdir only), the code sounds good to me as well.

sos/report/plugins/maas.py
```diff
@@ -8,133 +8,168 @@
...
 
-        if self._has_login_options():
-            if self._remote_api_login():
-                self.add_cmd_output(f"maas {self.get_option('profile-name')} "
-                                    "commissioning-results list")
-            else:
-                self._log_error(
-                    "Cannot login into MAAS remote API with provided creds.")
+    def setup(self):
+        for service in self._services:
```
troyanov, line 154: @arif-ali I decided not to use `services` because I don't want default behaviour as defined in https://github.com/sosreport/sos/blob/ca7300e7007c8b4faf7502ec3afb3a891267c2b2/sos/report/plugins/__init__.py#L3334-L3347

I would like to call `add_journal` with `since -1days` by default

pmoravec (approved):

sos/report/plugins/maas.py
troyanov: Thanks for the review!

I was looking for some built-in functions that I can use together with `glob.glob` but didn't find anything.
MAAS collects _machine logs_ using rsyslog and stores them in the following way:
```
/var/snap/maas/common/log/rsyslog$ tree
.
├── 32toledo
│   └── 2021-01-21
│       └── messages
├── adm21
│   └── 2021-03-03
│       └── messages
├── amco
│   ├── 2020-05-26
│   │   └── messages
│   ├── 2020-05-28
│   │   └── messages
│   └── 2020-08-06
│       └── messages
└── witty-shrimp
    └── 2021-10-05
        └── messages
```

I thought it would be great to collect just the most recent _machine log_ (e.g. for `amco` that will be `amco/2020-08-06`) as it will speed up `sos report -o maas` in most of the cases.

If `--all_logs` provided, we want to collect all the data for all the dates

troyanov:
I am going to try the new plugin on various MAAS deployments over the weekend and I will move PR into "Read for review" if there will be no issues.

arif-ali (approved):
lgtm

sos/report/plugins/maas.py
troyanov: Hm, I think for a better UX I should also check `self.get_option("since")` here and return matching directories

sos/report/plugins/maas.py
arif-ali: yeah, I think that's what was here before from the MAAS plugin

pmoravec (approved):

sos/report/plugins/maas.py
```diff
@@ -8,133 +8,177 @@
...
+        for host_dir in os.listdir(dir):
+            host_path = os.path.join(dir, host_dir)
+            if not os.path.isdir(host_path):
+                continue
+
+            subdirs = [
+                os.path.join(host_path, d)
+                for d in os.listdir(host_path)
+                if os.path.isdir(host_path)
+            ]
```
TurboTurtle, line 71: `Plugin` includes wrappers for these path methods, e.g. `Plugin.path_join()`, that are preferred because they will account for any `sysroot` handed to sos.

TurboTurtle (commented):
Looks good overall, just noting that we should prefer the path wrappers in `Plugin` so that we can account for `sysroot` changes.

sos/report/plugins/maas.py
troyanov: Oh, TIL. Thanks!

arif-ali (approved):

pmoravec (approved):

TurboTurtle (approved):

PR #3689, opened 2024-06-27
[common] Standardise snap package checks

arif-ali:
* Add function is_snap_installed, so that this can be used more widely
* This also reduces duplicate code in plugins.
* Add is_snap variable, that can be used if any of the packages is a snap; typically it's one of the other.

---

### DIFF
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -541,6 +541,7 @@ class Plugin():
     cmdtags = {}
     filetags = {}
     option_list = []
+    is_snap = False
 
     # Default predicates
     predicate = None
@@ -580,6 +581,11 @@ class Plugin():
         for opt in self.option_list:
             opt.plugin = self.name()
             self.options[opt.name] = opt
+
+        # Check if any of the packages tuple is a snap
+        self.is_snap = any(
+            self.is_snap_installed(pkg) for pkg in list(self.packages)
+        )
 
         # Initialise the default --dry-run predicate
         self.set_predicate(SoSPredicate(self))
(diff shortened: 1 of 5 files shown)


sos/report/plugins/__init__.py
```diff
@@ -999,6 +1005,18 @@ def is_installed(self, package_name):
...
+        """Is the snap package $package_name installed?
+
+        :param package_name:    The name of the package to check
+        :type package_name:     ``str``
+
+        :returns: ``True`` if the snap package is installed, else ``False``
+        :rtype: ``bool``
+        """
+        pkg = self.policy.package_manager.pkg_by_name(package_name)
+        return pkg and pkg['pkg_manager'] == 'snap'
```
TurboTurtle, line 1018: `pkg_by_name()` returns `None` for any named package that is not known to the package manager, the `return` does not guarantee a `bool` to be returned:

```python
>>> from sos.policies.package_managers.rpm import RpmPackageManage
>>> pm = RpmPackageManager()

>>> def is_snap_installed(package_name):
...     pkg = pm.pkg_by_name(package_name)
...     return pkg and pkg['pkg_manager'] == 'snap'
... 
>>> is_snap_installed('kernel')
False
>>> is_snap_installed('notapkg')
>>> 
>>> out = is_snap_installed('notapkg')
>>> type(out)
<class 'NoneType'>
>>> 
```


This doesn't affect the `any()` call in practice, but if we end up using this elsewhere, it could throw devels off.

arif-ali: okie, I'll get this refactored

TurboTurtle (approved):

PR #3690, opened 2024-06-28
[Krb5]  Additon of keytab Validation

Be-Rahul:
---

- [ X] Are all passwords or private data gathered by this PR [obfuscated](https://github.com/sosreport/sos/wiki/How-to-Write-a-Plugin#how-to-prevent-collecting-passwords)?

### DIFF
--- a/sos/report/plugins/krb5.py
+++ b/sos/report/plugins/krb5.py
@@ -8,6 +8,8 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
+import re
+import socket
 from sos.report.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
 
 
@@ -33,9 +35,37 @@ class Krb5(Plugin):
             f"{self.kdcdir}/kdc.conf",
             "/var/log/kadmind.log"
         ])
+        self.collect_kinit()
         self.add_copy_spec("/var/log/krb5kdc.log", tags="kerberos_kdc_log")
         self.add_cmd_output(f"klist -ket {self.kdcdir}/.k5*")
         self.add_cmd_output("klist -ket /etc/krb5.keytab")
+
+    def collect_kinit(self):
+        """
+        Collect the kinit command output for the system with id_provider "AD"
+        or "IPA" domains.
+
+        While integrating the Linux M/c with AD the realmd will create a
+        computer object on the AD side. The realmd and AD restrict the
+        Hostname/SPN to 15 Characters.
+        """
+
+        hostname = socket.getfqdn()
+        sssd_conf = "/etc/sssd/sssd.conf"
+        if self.path_isfile(sssd_conf):
+            with open(sssd_conf, 'r') as f:
+                for line in f:
+                    if re.match(r'\s*id_provider\s*=\s*ad',
+                                line, re.IGNORECASE):
+                        hostname = hostname.split('.')[0][:15].upper()
+                        self.add_cmd_output(f"KRB5_TRACE=/dev/stdout \
+                                            kinit -k '{hostname}$'")
+                        break
+                    if re.match(r'\s*id_provider\s*=\s*ipa',
+                                line, re.IGNORECASE):
+                        self.add_cmd_output(f"KRB5_TRACE=/dev/stdout \
+                                            kinit -k '{hostname}'")
+                        break
 
 
 class RedHatKrb5(Krb5, RedHatPlugin):


jcastill:
You'll need to sign the commit using 'git commit -s' when sending PRs

jcastill:
Try:

`git commit --amend -s
`
And push again

sos/report/plugins/krb5.py
```diff
@@ -36,6 +37,8 @@ def setup(self):
         self.add_copy_spec("/var/log/krb5kdc.log", tags="kerberos_kdc_log")
         self.add_cmd_output(f"klist -ket {self.kdcdir}/.k5*")
         self.add_cmd_output("klist -ket /etc/krb5.keytab")
+        hostname = socket.gethostname()[:11].upper()
+        self.add_cmd_output(f"kinit -k '{hostname}$@REDHAT.LOCAL'")
```
jcastill, line 41: I think it may be better to move this down to RedHatKrb5.

sos/report/plugins/krb5.py
```diff
@@ -9,6 +9,7 @@
 # See the LICENSE file in the source distribution for further information.
 
 from sos.report.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
+import socket
```
jcastill, line 12: Nitpic, but in other plugins, like 'ceph_common', 'pcp', and 'qpid_dispatch' we do this instead:

`from socket import gethostname`

So it may be good to do the same here.

jcastill:
We try to avoid having multiple commits per PR, so please squash them into one.

sos/report/plugins/krb5.py
arif-ali, line 12: can you add this above the previous import please as per pylint's error message

sos/report/plugins/krb5.py
arif-ali, line 41: The domain `REDHAT.LOCAL` has to be more generic than that please. You may need to  grab the domain from the config file, and use that here

arif-ali (changes requested):
some changes below

jcastill (changes requested):

sos/report/plugins/krb5.py
Be-Rahul: Sure @arif-ali

Be-Rahul: Did the suggested changes and move this above the sos.report.plugin import

Be-Rahul:
Hello @arif-ali , we are validating the keytab. A keytab file/table or SPN will get generated if we join the AD domain, with id_provider =ad or using iPA and this is wrt to all linux distribution. We are not focusing on other sssd or kerbero stuff.

Do you still want to use realm from kdc.conf, then please let me know.

arif-ali:
> Hello @arif-ali , we are validating the keytab. A keytab file/table or SPN will get generated if we join the AD domain, with id_provider =ad or using iPA and this is wrt to all linux distribution. We are not focusing on other sssd or kerbero stuff.
> 
> Do you still want to use realm from kdc.conf, then please let me know.

ack, sounds good to me

arif-ali (approved):

jcastill (approved):

sos/report/plugins/krb5.py
```diff
@@ -33,10 +35,32 @@ def setup(self):
...
+    def collect_kinit(self):
+        """
+        Collect the kinit command output for the system with id_provider "AD"
+        or "IPA" domains.
+        """
+
+        hostname = socket.gethostname()[:11].upper()
+        hostname_fqdn = socket.getfqdn()
+        sssd_conf = "/etc/sssd/sssd.conf"
+        if os.path.isfile(sssd_conf):
```
TurboTurtle, line 52: `Plugin.path_isfile()` is a wrapper for this which accounts for any change in our sysroot, please use that.

sos/report/plugins/krb5.py
```diff
@@ -33,10 +35,32 @@ def setup(self):
...
+            with open(sssd_conf, 'r') as f:
+                content = f.read()
+                if 'id_provider = ad' in content:
+                    self.add_cmd_output(f"kinit -k '{hostname}$'")
+                elif 'id_provider = ipa' in content:
+                    self.add_cmd_output(f"kinit -k 'host/{hostname_fqdn}'")
+                else:
+                    pass
+        else:
+            pass
```
TurboTurtle, line 62: This is superfluous.

sos/report/plugins/krb5.py
```diff
@@ -33,10 +35,32 @@ def setup(self):
...
+        sssd_conf = "/etc/sssd/sssd.conf"
+        if os.path.isfile(sssd_conf):
+            with open(sssd_conf, 'r') as f:
+                content = f.read()
+                if 'id_provider = ad' in content:
+                    self.add_cmd_output(f"kinit -k '{hostname}$'")
+                elif 'id_provider = ipa' in content:
+                    self.add_cmd_output(f"kinit -k 'host/{hostname_fqdn}'")
+                else:
+                    pass
```
TurboTurtle, line 60: We can button this up a bit better by doing a per-line regex iteration, rather than checking the entirety of the file at once.

sos/report/plugins/krb5.py
```diff
@@ -33,10 +35,32 @@ def setup(self):
...
         self.add_cmd_output(f"klist -ket {self.kdcdir}/.k5*")
         self.add_cmd_output("klist -ket /etc/krb5.keytab")
 
+    def collect_kinit(self):
+        """
+        Collect the kinit command output for the system with id_provider "AD"
+        or "IPA" domains.
+        """
+
+        hostname = socket.gethostname()[:11].upper()
```
TurboTurtle, line 49: Why the last 11 characters? I'm not super familiar with kerberos myself, but it seems odd to take a statically sized slice here to me.

TurboTurtle (changes requested):

sos/report/plugins/krb5.py
Be-Rahul: Removed and used break statements. If the string match execute the command and break

sos/report/plugins/krb5.py
Be-Rahul: made the changes in the code and used re.match with IGNORECASE to match with string and execute the command.

sos/report/plugins/krb5.py
arif-ali: @Be-Rahul please update this to read the following as per @TurboTurtle's comment

```python
        if self.path_isfile(sssd_conf):
```

sos/report/plugins/krb5.py
arif-ali: @Be-Rahul below is a simple thing that would cause problems of just using 11 characters, and hence you would not get the full hostname

```
ubuntu@juju-728f4e-13-lxd-3:~$ python3
Python 3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import socket
>>> socket.gethostname()
'juju-728f4e-13-lxd-3'
>>> socket.gethostname()[:11]
'juju-728f4e'
>>> socket.gethostname()[:11].upper()
'JUJU-728F4E'
```

It would be better to split on the `.` if it is FQDN , and if there isn't, just get the full hostname

sos/report/plugins/krb5.py
```diff
@@ -33,10 +36,32 @@ def setup(self):
...
+        hostname_fqdn = socket.getfqdn()
+        sssd_conf = "/etc/sssd/sssd.conf"
+        if os.path.isfile(sssd_conf):
+            with open(sssd_conf, 'r') as f:
+                for line in f:
+                    if re.match(r'\s*id_provider\s*=\s*ad',
+                                line, re.IGNORECASE):
+                        self.add_cmd_output(f"kinit -k '{hostname}$'")
+                        break
+                    elif re.match(r'\s*id_provider\s*=\s*ipa',
```
arif-ali, line 60: as per the `pylint` message, replace the `elif` with just `if`

arif-ali (changes requested):
before this is merged the commits need to be squashed and the Signed-by line needs to be in the commit message.

sos/report/plugins/krb5.py
Be-Rahul: Yes @arif-ali I missed that while doing the changes.

sos/report/plugins/krb5.py
Be-Rahul: Hello @arif-ali , I had discussion with IDM SME and there is limit of 15 characters in the command used for AD.
I used the above suggestion as.

for id_provider = ipa
hostname = socket.getfqdn()

for id_provider = ad
hostname = hostname.split('.')[0][:15].upper()

arif-ali (approved):

sos/report/plugins/krb5.py
```diff
@@ -33,10 +35,34 @@ def setup(self):
...
+        """
+
+        hostname = socket.getfqdn()
+        sssd_conf = "/etc/sssd/sssd.conf"
+        if self.path_isfile(sssd_conf):
+            with open(sssd_conf, 'r') as f:
+                for line in f:
+                    if re.match(r'\s*id_provider\s*=\s*ad',
+                                line, re.IGNORECASE):
+                        hostname = hostname.split('.')[0][:15].upper()
```
pmoravec, line 56: Could the limit of 15 chars be explained in a comment? Reading the code, it is hard to understand otherwise..

Be-Rahul:
> Thinking loudly and not knowing the topic much: `sos` should not alter the system it is running on. Doesn't `kinit` obtain Kerberos ticket and hence alter the system that way..?

This kinit -k 'HOSTNAME$' is not obtaining a ticket from Active Directory and LDAP server. We are validating the existing keytab SPN.
Note: This does not modifying anything

pmoravec (approved):
OK then, thanks for clarification.

sos/report/plugins/krb5.py
Be-Rahul: While integrating a Linux machine with AD the realmd  will create a computer object on the AD side in the name of the linux hostname and if the linux hostname is more than 15 char AD will not permit so the realmd will trim the hostname to 15 charc and on the basis of this trimmed hostname we will have the SPN.

Will add above explanation in the code

TurboTurtle (approved):

PR #3697, opened 2024-07-10
[process] Collect process stat files

rh-jhester:
The /proc/<PID> subdirectory for a process contains many files, most of which the report doesn't capture. One of those files is stat, which contains, among other pieces of information, the CPU on which the process is running. This information is not contained in any of the files that the report currently collects, and is important for troubleshooting, e.g., isolcpus or other scheduling issues.  The status file contains the field Cpus_allowed and Cpus_allowed_list, but this is not the same information. The underlying task_struct has separate members for the current CPU and the allowed CPU mask (list). The stat file is quite small and should be included in the report since it contains important information not contained elsewhere.

Signed-off-by: Jesse Hester <jhester@redhat.com>
---

- [ X] Are all passwords or private data gathered by this PR [obfuscated](https://github.com/sosreport/sos/wiki/How-to-Write-a-Plugin#how-to-prevent-collecting-passwords)?

### DIFF
--- a/sos/report/plugins/process.py
+++ b/sos/report/plugins/process.py
@@ -50,6 +50,7 @@ class Process(Plugin, IndependentPlugin):
 
         for proc in procs:
             self.add_copy_spec([
+                f"/proc/{proc}/stat",
                 f"/proc/{proc}/status",
                 f"/proc/{proc}/cpuset",
                 f"/proc/{proc}/oom_*",


rh-jhester:
After submitting this PR I noticed that the CPU that each process is running on is included in the output of the pidstat commands. We may want to include the stat files anyway--especially if we thought that the status files contained equivalent information--but the use case that prompted me to create the PR is technically invalidated.

arif-ali:
@rh-jhester is still something you're working on?

pmoravec:
@rh-jhester , what is the status, then? Is this PR still worth merging? Code change itself is trivially acknowledge-able, if the newly collected data would be beneficial (something I dont know but I understood your original business justification for the PR is gone).

PR #3698, opened 2024-07-11
[ceph_mds] Add required path parameter to `dump tree`

slapcat:
Current output of this command without a path is:

```
Invalid command: missing required parameter root(<string>)
dump tree <root> [<depth:int>] :  dump metadata cache for subtree
admin_socket: invalid command
```

---

### DIFF
--- a/sos/report/plugins/ceph_mds.py
+++ b/sos/report/plugins/ceph_mds.py
@@ -75,7 +75,7 @@ class CephMDS(Plugin, RedHatPlugin, UbuntuPlugin):
             "counter schema",
             "damage ls",
             "dump loads",
-            "dump tree",
+            "dump tree /",
             "dump_blocked_ops",
             "dump_historic_ops",
             "dump_historic_ops_by_duration",


jcastill (approved):
lgtm

sos/report/plugins/ceph_mds.py
```diff
@@ -75,7 +75,7 @@ def setup(self):
             "counter schema",
             "damage ls",
             "dump loads",
-            "dump tree",
+            "dump tree /",
```
arif-ali, line 78: curious question, would `dump tree` work on older ceph, and now it is `dump tree /`?

arif-ali (approved):

sos/report/plugins/ceph_mds.py
slapcat: I'm not sure. I noticed the issue when reviewing a sosreport from Quincy.

pmoravec (approved):

pmoravec:
Do you know, the command has never worked? Or it worked well for some old `ceph` version? I just want to prevent situation when some old, still supported `ceph` version works well without the directory parameter, AND it raises an error when providing the argument.

TurboTurtle (approved):

PR #3701, opened 2024-07-16
[selinux] Capture authselect config file

jcastill:
Capture authselect config file

Related: RHEL-48608

---

### DIFF
--- a/sos/report/plugins/pam.py
+++ b/sos/report/plugins/pam.py
@@ -27,7 +27,8 @@ class Pam(Plugin):
 
         self.add_copy_spec([
             "/etc/pam.d",
-            "/etc/security"
+            "/etc/security",
+            '/etc/authselect/authselect.conf',
         ])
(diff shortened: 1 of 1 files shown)


arif-ali (approved):

arif-ali:
We're doing `authselect current` in the pam plugin, would that not be related? Ie. we're collecting the config here, but doing the command in the `pam` plugin?

TurboTurtle:
> We're doing `authselect current` in the pam plugin, would that not be related? Ie. we're collecting the config here, but doing the command in the `pam` plugin?

My thoughts as well. We should keep `authselect` stuff together, and I think the correct place for that is `pam`.

jcastill:
You are absolutely right, sorry I missed that. I'll rework and push again.

pmoravec (approved):

TurboTurtle (approved):

PR #3711, opened 2024-07-17
[microcloud] Added new plugin support for MicroCloud.

alanbach:
This implements a new small plugin support for the MicroCloud snap.

### DIFF
--- a/sos/report/plugins/microcloud.py
+++ b/sos/report/plugins/microcloud.py
@@ -1,0 +1,44 @@
+# Copyright (C) 2024 Alan Baghumian <alan.baghumian@canonical.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 or later of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, UbuntuPlugin
+
+
+class MicroCloud(Plugin, UbuntuPlugin):
+    """The MicroCloud plugin collects the current status of the microcloud
+    snap.
+
+    It will collect journald logs as well as output from various microcloud
+    commands.
+    """
+
+    short_desc = 'MicroCloud Snap'
+    plugin_name = "microcloud"
+    profiles = ('container',)
+
+    packages = ('microcloud',)
+
+    def setup(self):
+        self.add_journal(units="snap.microcloud.*")
+
+        microcloud_subcmds = [
+            'cluster list',
+            '--version'
+        ]
+        self.add_copy_spec([
+            '/var/snap/microcloud/common/state/database/cluster.yaml',
+            '/var/snap/microcloud/common/state/database/info.yaml',
+        ])
+
+        self.add_cmd_output([
+            f"microcloud {subcmd}" for subcmd in microcloud_subcmds
+        ])
+
+# vim: set et ts=4 sw=4


sos/report/plugins/microcloud.py
```diff
@@ -0,0 +1,44 @@
...
+    commands.
+    """
+
+    short_desc = 'MicroCloud Snap'
+    plugin_name = "microcloud"
+    profiles = ('container',)
+
+    packages = ('microcloud',)
+
+    microcloud_cmd = "microcloud"
```
pmoravec, line 28: You don't use the variable either where.

sos/report/plugins/microcloud.py
```diff
@@ -0,0 +1,44 @@
...
+        self.add_journal(units="snap.microcloud.*")
+
+        microcloud_subcmds = [
+            'cluster list',
+            '--version'
+        ]
+        self.add_copy_spec([
+            '/var/snap/microcloud/common/state/database/cluster.yaml',
+            '/var/snap/microcloud/common/state/database/info.yaml',
+        ])
```
pmoravec, line 40: I guess there are no secrets there (like passwords to the database..?)?

sos/report/plugins/microcloud.py
```diff
@@ -0,0 +1,44 @@
...
+
+        self.add_cmd_output([
+            f"microcloud {subcmd}" for subcmd in microcloud_subcmds
+        ])
```
pmoravec, line 44: As you will respin the PR, could you please add here the `vim` config for indentation? The `# vim: set et ts=4 sw=4 :` line like e.g. https://github.com/sosreport/sos/blob/main/sos/report/plugins/abrt.py#L44 or at the botton of any other source code file.

arif-ali: > As you will respin the PR, could you please add here the `vim` config for indentation? The `# vim: set et ts=4 sw=4 :` line like e.g. https://github.com/sosreport/sos/blob/main/sos/report/plugins/abrt.py#L44 or at the botton of any other source code file.

we ought to add that to the Contribution Guidelines as well for future contributors

sos/report/plugins/microcloud.py
```diff
@@ -0,0 +1,44 @@
+# Copyright (C) 2024 Alan Baghumian <alan.baghumian@canonical.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
```
arif-ali, line 7: We're just changing the guidelines on licensing, so would you update this to `version 2 or later` please

sos/report/plugins/microcloud.py
```diff
@@ -0,0 +1,44 @@
...
+    short_desc = 'MicroCloud Snap'
+    plugin_name = "microcloud"
+    profiles = ('container',)
+
+    packages = ('microcloud',)
+
+    microcloud_cmd = "microcloud"
+
+    def setup(self):
+        self.add_journal(units="snap.microcloud.*")
```
arif-ali, line 31: do we need to check for service status as well?

sos/report/plugins/microcloud.py
alanbach: Yup, removing that last one, we do need the packages one thought

sos/report/plugins/microcloud.py
alanbach: Done! Coming.

sos/report/plugins/microcloud.py
alanbach: Done. Coming.

sos/report/plugins/microcloud.py
alanbach: Nah, the systemd module is already collecting it, so no need for a duplicated efforts

pmoravec (approved):

arif-ali (approved):

PR #3716, opened 2024-07-19
[policies] Fallback to /lib/modules/*/config

champtar:
When using rpm-ostree, /boot/config-{release} doesn't exists but RHEL/Fedora family also ship the kernel config file in /lib/modules/{release}/config, so try that as fallback.

---

### DIFF
--- a/sos/policies/distros/__init__.py
+++ b/sos/policies/distros/__init__.py
@@ -269,7 +269,19 @@ class LinuxPolicy(Policy):
             'dm_mod': 'CONFIG_BLK_DEV_DM'
         }
 
-        booted_config = self.join_sysroot(f"/boot/config-{release}")
+        kconfigs = (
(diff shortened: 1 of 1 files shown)


arif-ali:
curious, is it worth removing the try/except for the IOError noe, as we're now checking the file existing now, and we wouldn't even get to that section of the code if the file didn't exist

champtar:
> curious, is it worth removing the try/except for the IOError noe, as we're now checking the file existing now, and we wouldn't even get to that section of the code if the file didn't exist

We only check if it exists, maybe the /boot partition is on a disk with I/O errors, or messed up permissions, I think we do want to continue here. After all we use sos report when you have a problem :)

arif-ali (approved):

pmoravec (approved):

PR #3717, opened 2024-07-19
[vdo] Capture more information about vdo volumes

jcastill:
Capture more information about vdo volumes

Related: VDO-5797

---

### DIFF
--- a/sos/report/plugins/vdo.py
+++ b/sos/report/plugins/vdo.py
@@ -28,6 +28,32 @@ class Vdo(Plugin, RedHatPlugin):
         vdos = self.collect_cmd_output('vdo list --all')
         for vdo in vdos['output'].splitlines():
             self.add_cmd_output(f"vdo status -n {vdo}")
-        self.add_cmd_output('vdostats --human-readable')
+        self.add_cmd_output([
+            'vdostats --human-readable',
+            'vdostats --verbose',
+        ])
+        vdo_cols1 = ('vdo_slab_size,vdo_header_size,vdo_minimum_io_size,'
+                     'vdo_block_map_cache_size,vdo_block_map_era_length,'
+                     'vdo_write_policy,vdo_max_discard')
+
+        vdo_cols2 = ('vdo_ack_threads,vdo_bio_rotation,vdo_bio_threads,'
+                     'vdo_cpu_threads,vdo_hash_zone_threads,'
+                     'vdo_logical_threads,vdo_physical_threads')
+        vdo_cols3 = ('vdo_compression,vdo_deduplication,'
+                     'vdo_use_metadata_hints,vdo_use_sparse_index,'
+                     'vdo_index_state,vdo_index_memory_size')
+        self.add_cmd_output([f"lvs -a -o +{cols}"
+                             for cols in [vdo_cols1, vdo_cols2]])
+        lvm_vdos = self.collect_cmd_output(f"lvs -a -o +{vdo_cols3}")
+        if lvm_vdos['status'] == 0:
+            for vdo in lvm_vdos['output'].splitlines():
+                # we can find the pool and pool data maps in the output
+                # of lvs, in the column Volume type, marked as 'D'
+                lv, vg, lv_attr = vdo.split()[:3]
+                if lv_attr.startswith("D"):
+                    vdo_path = f"{vg}-{lv.strip('[]')}"
+                    self.add_cmd_output(
+                        f"vdodumpconfig /dev/mapper/{vdo_path}"
+                    )
 
 # vim set et ts=4 sw=4 :


jcastill:
In this case I moved the 'lvs' command to the VDO plugin because that will guarantee that the command is run when the conditions for the other VDO command captures are met. If I add it to the LVM plugin, I'll have to add extra logic to make sure VDO is in place, and I thought it was a more natural place to have it in the VDO than the LVM one. But I can change it if  the logic is not sound.

pmoravec (approved):

sos/report/plugins/vdo.py
```diff
@@ -28,6 +28,11 @@ def setup(self):
         vdos = self.collect_cmd_output('vdo list --all')
         for vdo in vdos['output'].splitlines():
             self.add_cmd_output(f"vdo status -n {vdo}")
-        self.add_cmd_output('vdostats --human-readable')
+        self.add_cmd_output([
+            'vdostats --human-readable',
+            'vdostats --verbose'
```
arif-ali, line 33: very minor nit, comma at the end

arif-ali (approved):

sos/report/plugins/vdo.py
jcastill: Fixed

jcastill:
Added the missing comma and another extra command. With this change now we should capture both non-LVM and LVM VDO devices

sos/report/plugins/vdo.py
```diff
@@ -28,6 +28,18 @@ def setup(self):
         vdos = self.collect_cmd_output('vdo list --all')
         for vdo in vdos['output'].splitlines():
             self.add_cmd_output(f"vdo status -n {vdo}")
-        self.add_cmd_output('vdostats --human-readable')
+        self.add_cmd_output([
+            'vdostats --human-readable',
+            'vdostats --verbose',
+        ])
+        vdo_cols = 'vdo_deduplication,vdo_index_state'
+        lvs_opts = '-a -o +'
```
pmoravec, line 36: Why the two extra variables, if you use either of them just once?

sos/report/plugins/vdo.py
```diff
@@ -28,6 +28,18 @@ def setup(self):
...
+        lvm_vdos = self.collect_cmd_output(f"lvs {lvs_opts}{vdo_cols}")
+        for vdo in lvm_vdos['output'].splitlines():
```
pmoravec, line 38: What if the `lvs` command would raise an error - is it safe to parse its output like this? (we often guard `'output'` under `if lvm_vdos['status'] == 0:` - but here I see this plugin lacks this check (cf. lines 28-29).

sos/report/plugins/vdo.py
jcastill: Two reasons:

- The LVS command was too long before I removed the headers option, so using these two variables I was within the limit, and
- The request has a very big list of columns and I'll need to split the output in at least two variables to make sure is readable. 

I'm still playing with the columns I need to display, so the name vdo_cols will probably change.

sos/report/plugins/vdo.py
jcastill: I'll do this, sorry I missed the check.

sos/report/plugins/vdo.py
pmoravec: OK, could be.. Bit redundant for the code, more readable for users.

jcastill: This is still a draft PR so I'll see how can I improve it

pmoravec (approved):

jcastill:
After having a chat with VDO engineering, we've reworked the column output to have three captures that group the outputs that make more sense, instead of one or two with too many columns that makes them unreadable.

sos/report/plugins/vdo.py
```diff
@@ -28,6 +28,32 @@ def setup(self):
...
+                     'vdo_index_state,vdo_index_memory_size')
+        self.add_cmd_output([f"lvs -a -o +{cols}"
+                             for cols in [vdo_cols1, vdo_cols2]])
+        lvm_vdos = self.collect_cmd_output(f"lvs -a -o +{vdo_cols3}")
+        if lvm_vdos['status'] == 0:
+            for vdo in lvm_vdos['output'].splitlines():
+                # we can find the pool and pool data maps in the output
+                # of lvs, in the column Volume type,
+                # marked as either 'd' or 'D'
+                if vdo.split()[2].startswith("D"):
```
pmoravec, line 53: will this catch 'd'? As the comment speaks about 'd' or 'D', while the test checks for 'D' only..?

sos/report/plugins/vdo.py
```diff
@@ -28,6 +28,32 @@ def setup(self):
...
+                    vdo_path = f"{vdo.split()[1]}-{vdo.split()[0].strip('[]')}"
```
pmoravec, line 54: Are we sure `vdo` will always contain at least 3 words..? (we do rely on such assumptions, sometimes. I dont know `vdo` at all to assess if we can safely rely here).

As we call `vdo.split()` three times in a row, isn't it faster and also better to understand something like:

     foo, bar, volume = vdo.split()[:3]

but with some more appropriate names?

sos/report/plugins/vdo.py
jcastill: I can see the text is confusing, so I'll change it to mention only capital D

sos/report/plugins/vdo.py
jcastill: The output of lvs should always be the same, because we rely on standard columns in the output, and we add the rest with the '-o+'. If the default column order change then we'll have to revisit other plugins as well. The three columns I rely on are:

`  LV             VG    Attr       `

But the device in /dev/ has the format VG-LV, and the VG is always surrounded (in this case) by [], so I have to split and then strip. I'll try to rewrite vdo_path using your approach.

pmoravec (approved):

PR #3718, opened 2024-07-19
[ceph] Expand Ceph RGW data collection

pponnuvel:
---

### DIFF
--- a/sos/report/plugins/ceph_rgw.py
+++ b/sos/report/plugins/ceph_rgw.py
@@ -6,6 +6,9 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
+import json
+
+from socket import gethostname
 from sos.report.plugins import Plugin, RedHatPlugin, UbuntuPlugin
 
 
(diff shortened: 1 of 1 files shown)


sos/report/plugins/ceph_rgw.py
```diff
@@ -21,6 +22,29 @@ class CephRGW(Plugin, RedHatPlugin, UbuntuPlugin):
...
+            'metadata sync status',
+            'period list',
+            'realm list',
+            'reshard list',
+            'sync error list',
+            'sync status',
+            'zone get --rgw-zone=default',
+            'zone list',
+            'zone placement list',
+            'zonegroup get --rgw-zonegroup=default',
```
arif-ali, line 43: I see here we're only looking at `rgw-zone=default` and `rgw-zonegroup=default`, is it worth looking at others if they exist?

pponnuvel: I think so. Usually default is only one. But there can be others too, so expanded.

sos/report/plugins/ceph_rgw.py
```diff
@@ -57,4 +81,52 @@ def setup(self):
...
+                self._log_error(f'Error while getting get rgw '
+                                f'zonegroup list: {err}')
+
+    def postproc(self):
+        """ Obfuscate secondary zone access keys """
+
+        rsub = r'("access_key":|"secret_key":)\s.*'
+
+        for _cmd in self.cmd_list:
+            self.do_cmd_output_sub(_cmd, rsub, r'\1 "**********"')
```
arif-ali, line 129: You don't actually need to know the list of commands to do this. You could just do `radosgw-admin` as a text to the `do_cmd_output_sub`, and it should sort itself out.

arif-ali (changes requested):

pmoravec (approved):

TurboTurtle (approved):

PR #3720, opened 2024-07-20
[pylint] Re-enable W0622: redefined-builtin

pponnuvel:
---

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -52,7 +52,7 @@ master_doc = 'index'
 
 # General information about the project.
 project = 'SoS'
-copyright = '2014, Bryn Reeves'
+project_copyright = '2014, Bryn Reeves'
 
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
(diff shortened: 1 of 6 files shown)


sos/__init__.py
```diff
@@ -34,12 +34,6 @@ def _default(msg):
 
 _sos = _default
 
-# py3 < 3.6 compat
-try:
-    ModuleNotFoundError
-except NameError:
-    ModuleNotFoundError = ImportError
-
```
arif-ali, line 42: We may need to leave this as is, and disable the check here while we're still supporting python 3.6 for both bionic and centos 8, till later in the year

pponnuvel: It's needed for < Python 3.6. It should be OK for 3.6 or later; I thought the minimum version is 3.6?

docs/conf.py
```diff
@@ -52,7 +52,7 @@
 
 # General information about the project.
 project = 'SoS'
-copyright = '2014, Bryn Reeves'
+project_copyright = '2014, Bryn Reeves'
```
pponnuvel, line 55: `project_copyright` is just an alias for `copyright` but using the former avoids the redefinition conflict with builtin `copyright`.

TurboTurtle (approved):

arif-ali (approved):

PR #3724, opened 2024-07-21
[pylint] re-enable no-else-return and inconsistent-return-statements

pponnuvel:
---

### DIFF
--- a/sos/collector/__init__.py
+++ b/sos/collector/__init__.py
@@ -619,18 +619,13 @@ class SoSCollector(SoSComponent):
                 msg = "Invalid option type for %s. Expected %s got %s"
                 self.exit(msg % (cli.name, default.opt_type, cli.opt_type), 1)
             return cli.value
-        else:
-            val = cli.value.lower()
-            if val not in ['true', 'on', 'yes', 'false', 'off', 'no']:
-                msg = ("Invalid value for %s. Accepted values are: 'true', "
-                       "'false', 'on', 'off', 'yes', 'no'.")
-                self.exit(msg % cli.name, 1)
-            else:
-                if val in ['true', 'on', 'yes']:
-                    return True
-                else:
-                    return False
-        self.exit(f"Unknown option type: {cli.opt_type}")
+
+        val = cli.value.lower()
+        if val not in ['true', 'on', 'yes', 'false', 'off', 'no']:
+            msg = ("Invalid value for %s. Accepted values are: 'true', "
+                   "'false', 'on', 'off', 'yes', 'no'.")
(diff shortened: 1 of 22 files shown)


sos/collector/__init__.py
```diff
@@ -619,18 +619,13 @@ def _validate_option(self, default, cli):
...
-                else:
-                    return False
-        self.exit(f"Unknown option type: {cli.opt_type}")
+
+        val = cli.value.lower()
+        if val not in ['true', 'on', 'yes', 'false', 'off', 'no']:
+            msg = ("Invalid value for %s. Accepted values are: 'true', "
+                   "'false', 'on', 'off', 'yes', 'no'.")
+            self.exit(msg % cli.name, 1)
+        return val in ['true', 'on', 'yes']
```
TurboTurtle, line 628: We need to account for non-matching/indeterminate option types here.

TurboTurtle (commented):
Looks mostly good to me, just the one note.

sos/collector/__init__.py
TurboTurtle: Actually, looking at this again, perhaps the unknown type is superfluous.

pponnuvel: yeah, the unknown value is already covered by the previous `if` clause. So the final `self.exit(f"Unknown option type: {cli.opt_type}")` was always an unreachable code.

pponnuvel:
Also addressed `inconsistent-return-statements` here (combined https://github.com/sosreport/sos/pull/3726 with this one).

sos/policies/distros/redhat.py
```diff
@@ -138,9 +138,8 @@ def check_usrmove(self, pkgs):
         """
         if 'filesystem' not in pkgs:
             return os.path.islink('/bin') and os.path.islink('/sbin')
-        else:
-            filesys_version = pkgs['filesystem']['version']
-            return True if filesys_version[0] == '3' else False
+        filesys_version = pkgs['filesystem']['version']
+        return True if filesys_version[0] == '3' else False
```
pmoravec, line 142: Isn't `return (filesys_version[0] == '3')` more concise and readable / understandable?

Anyway, that is a change different than fixinf the pylint issue, ACK to the current change either way.

pmoravec (approved):

sos/policies/distros/redhat.py
pponnuvel: yeah, fixed that now.

pmoravec (approved):

TurboTurtle (approved):

PR #3727, opened 2024-07-21
[pylint] Re-enable unnecessary-{negation,dunder-call,comprehension, lambda}

pponnuvel:
Fixes unnecessary-{negation,dunder-call,comprehension,lambda}.

---

### DIFF
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -1821,7 +1821,7 @@ class Plugin():
                 continue
 
             if since or maxage:
-                files = list(filter(lambda f: time_filter(f), files))
+                files = list(filter(time_filter, files))
 
             files.sort(key=getmtime, reverse=True)
             current_size = 0
@@ -2013,7 +2013,7 @@ class Plugin():
         if "snap_cmd" in kwargs:
             kwargs.pop("snap_cmd")
         soscmd = SoSCommand(**kwargs)
-        self._log_debug("packed command: " + soscmd.__str__())
+        self._log_debug(f"packed command: {str(soscmd)}")
         for _skip_cmd in self.skip_commands:
             # This probably seems weird to be doing filename matching on the
             # commands, however we want to remain consistent with our regex
@@ -3156,7 +3156,7 @@ class Plugin():
     def _collect_cmds(self):
         self.collect_cmds.sort(key=lambda x: x.priority)
         for soscmd in self.collect_cmds:
-            self._log_debug("unpacked command: " + soscmd.__str__())
+            self._log_debug(f"unpacked command: {str(soscmd)}")
             user = ""
             if getattr(soscmd, "runas", None) is not None:
                 user = f", as the {soscmd.runas} user"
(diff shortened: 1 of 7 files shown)


TurboTurtle (commented):
For the unecessary-pass changes, notably in `archive.py`, `policies/__init__.py`, and `clusters/__init__.py`, should we not instead raise e.g. `NotImplementedError` or something similar where appropriate? The `pass` here was originally intended to coincide with methods that should be overwritten by actual archives, policies, clusters, etc...

pponnuvel:
> For the unecessary-pass changes, notably in `archive.py`, `policies/__init__.py`, and `clusters/__init__.py`, should we not instead raise e.g. `NotImplementedError` or something similar where appropriate? The `pass` here was originally intended to coincide with methods that should be overwritten by actual archives, policies, clusters, etc...

Ack. I've replaced several of those with 'raise NotImplementedError'. Please take a look again.

sos/report/plugins/__init__.py
```diff
@@ -2011,7 +2011,7 @@ def _add_cmd_output(self, **kwargs):
         if self.get_option('all_logs') or kwargs['sizelimit'] == 0:
             kwargs['to_file'] = True
         soscmd = SoSCommand(**kwargs)
-        self._log_debug("packed command: " + soscmd.__str__())
+        self._log_debug("packed command: " + str(soscmd))
```
pmoravec, line 2014: We might move to f-strings here already, but that would require bit bigger change here, I think (since f-strings use `format()` method, not currently called `__str_` or `str()` one).

pmoravec (approved):

sos/report/plugins/__init__.py
pponnuvel: I think f-string should work here:
```
>>> s=1
>>> o=f"test: {str(s)}"
>>> print(o)
```

pponnuvel: Switched to f-strings for the two changes in this file.

sos/collector/clusters/__init__.py
TurboTurtle, line 1: I need to look a bit deeper at this, and ideally have some tests against some actual clusters. Some of these should be raising the `NotImplementedError`, but a few of these should probably just be `return None` for accepting defaults.

pmoravec (approved):

arif-ali (approved):

arif-ali (approved):

pmoravec (approved):

arif-ali (approved):

pponnuvel:
I was working on `abstract-method` in https://github.com/sosreport/sos/pull/3750. Realised `raise NotImplementedError` can trigger quite a lot of `abstract-method` (79 more if this PR goes in). 

I think it's best to handle `unnecessary-pass` and `abstract-method` together. So removed handling `unnecessary-pass` (except the empty `pass` from classes) from this PR. I think this can be merged as-is now.

Thanks, @TurboTurtle - good call on not merging this earlier :)

TurboTurtle (approved):

pmoravec (approved):

arif-ali (approved):

PR #3739, opened 2024-08-08
[redhat] Fix NoneType return when checking file size limit

jcastill:
The new fuction check_file_too_big() was returning None when the file size was bigger than the max limit for Red Hat Customer Portal uploads. By removing the f-string that converted the size limit via convert_bytes() the function returned RH_SFTP_HOST as expected.

---

### DIFF
--- a/sos/policies/distros/redhat.py
+++ b/sos/policies/distros/redhat.py
@@ -22,7 +22,7 @@ from sos.policies.distros import LinuxPolicy, ENV_HOST_SYSROOT
 from sos.policies.package_managers.rpm import RpmPackageManager
 from sos.policies.package_managers.flatpak import FlatpakPackageManager
 from sos.policies.package_managers import MultiPackageManager
-from sos.utilities import bold, convert_bytes
+from sos.utilities import bold
 from sos import _sos as _
 
 try:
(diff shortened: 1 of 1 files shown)


jcastill:
I don't really like having to hard code the limit like this, but we need this workaround while I investigate how to properly fix the issue.

pponnuvel:
@jcastill I think I've found the issue. 

I think commit d65e135c8 appears to have introduced a bug when converting formats to use f-strings. This apparently broke the `convert_bytes()` function. So I believe this isn't an issue with `check_file_too_big` but `convert_bytes`.

Can you check if https://github.com/sosreport/sos/pull/3741 fixes the issue?

pmoravec:
The behaviour is a regression from https://github.com/sosreport/sos/commit/d65e135c8d5dbf75a0dc05427e9a4f777bcbf383 . Having previous strings like `'%.1fT' % (fn / T)` instead of `f'{(fn / T):.1fT}'` works well.

What is wrong with that f-string? The nested exception is `ValueError: Invalid format specifier` (this one is caught at https://github.com/sosreport/sos/pull/3739/files#diff-9e337c4eedfbd3c072d5d88a69b34a2d859ed1c5f5cae9180a62a9a25f468ef8R452)

jcastill:
Thank you for finding out the issue @pponnuvel ! I'll close this now, since @pmoravec verified it fixed the issue here.

PR #3741, opened 2024-08-08
[common] Fix a bug in f-string usage

pponnuvel:
The commit d65e135c8 appears to have introduced a bug when converting formats to use f-strings. This apparently broke the `convert_bytes()` function.

---

### DIFF
--- a/sos/utilities.py
+++ b/sos/utilities.py
@@ -121,18 +121,13 @@ def fileobj(path_or_file, mode='r'):
         return closing(path_or_file)
 
 
-def convert_bytes(bytes_, K=1 << 10, M=1 << 20, G=1 << 30, T=1 << 40):
+def convert_bytes(num_bytes):
     """Converts a number of bytes to a shorter, more human friendly format"""
-    fn = float(bytes_)
-    if bytes_ >= T:
-        return f'{(fn / T):.1fT}'
-    if bytes_ >= G:
-        return f'{(fn / G):.1fG}'
-    if bytes_ >= M:
-        return f'{(fn / M):.1fM}'
-    if bytes_ >= K:
-        return f'{(fn / K):.1fK}'
-    return f'{bytes_}'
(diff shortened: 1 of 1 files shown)


pmoravec (approved):
I can confirm it prevents the NoneType exception in https://github.com/sosreport/sos/pull/3739 .

TurboTurtle:
I was thinking we could perhaps button this up a bit:

```python
def convert_bytes(num_bytes):
    sizes = {'T': 1 << 40, 'G': 1 << 30, 'M': 1 << 20, 'K': 1 << 10}
    for symbol, size in sizes.items():
        return f"{float(num_bytes) / size:.1f}{symbol}" if num_bytes > size
    return f"{num_bytes}"
```

pponnuvel:
> I was thinking we could perhaps button this up a bit:
> 
> ```python
> def convert_bytes(num_bytes):
>     sizes = {'T': 1 << 40, 'G': 1 << 30, 'M': 1 << 20, 'K': 1 << 10}
>     for symbol, size in sizes.items():
>         if num_bytes > size:
>             return f"{float(num_bytes) / size:.1f}{symbol}"
>     return f"{num_bytes}"
> ```

This should work just as well. I can tweak the PR to use this. But a question: the previous check uses `>=` but it's just `>`. Is that intentional?

pmoravec (approved):
Nice!

TurboTurtle:
> > I was thinking we could perhaps button this up a bit:
> > ```python
> > def convert_bytes(num_bytes):
> >     sizes = {'T': 1 << 40, 'G': 1 << 30, 'M': 1 << 20, 'K': 1 << 10}
> >     for symbol, size in sizes.items():
> >         if num_bytes > size:
> >             return f"{float(num_bytes) / size:.1f}{symbol}"
> >     return f"{num_bytes}"
> > ```
> 
> This should work just as well. I can tweak the PR to use this. But a question: the previous check uses `>=` but the proposed code uses just `>`. Is that intentional?

It was not, good eye.

TurboTurtle (approved):

TurboTurtle (approved):

PR #3747, opened 2024-08-09
[pylint] Fixes more pylint issues

pponnuvel:
The following are fixed in this change:

    W3101 - missing-timeout
    W0231 - super-init-not-called
    C0123 - unidiomatic-typecheck
    C0207 - use-maxsplit-arg
    E0611 - no-name-in-module
    E0702 - raising-bad-type
    E1101 - no-member
    E1135 - unsupported-membership-test

---

### DIFF
--- a/sos/policies/auth/__init__.py
+++ b/sos/policies/auth/__init__.py
@@ -17,6 +17,8 @@ except ImportError:
 import time
 from datetime import datetime, timedelta
 
+from sos.utilities import TIMEOUT_DEFAULT
+
 DEVICE_AUTH_CLIENT_ID = "sos-tools"
 GRANT_TYPE_DEVICE_CODE = "urn:ietf:params:oauth:grant-type:device_code"
 
@@ -67,7 +69,8 @@ class DeviceAuthorizationClass:
             res = requests.post(
                 self.client_identifier_url,
                 data=data,
-                headers=headers)
+                headers=headers,
+                timeout=TIMEOUT_DEFAULT)
             res.raise_for_status()
             response = res.json()
             self._user_code = response.get("user_code")
@@ -99,7 +102,8 @@ class DeviceAuthorizationClass:
             time.sleep(self._interval)
             try:
                 check_auth_completion = requests.post(self.token_endpoint,
-                                                      data=token_data)
+                                                      data=token_data,
+                                                      timeout=TIMEOUT_DEFAULT)
 
                 status_code = check_auth_completion.status_code
 
@@ -187,7 +191,8 @@ class DeviceAuthorizationClass:
                               refresh_token else refresh_token}
 
         refresh_token_res = requests.post(self.token_endpoint,
-                                          data=refresh_token_data)
+                                          data=refresh_token_data,
+                                          timeout=TIMEOUT_DEFAULT)
 
         if refresh_token_res.status_code == 200:
             self._set_token_data(refresh_token_res.json())
--- a/tests/sos_tests.py
+++ b/tests/sos_tests.py
@@ -389,7 +389,7 @@ class BaseSoSReportTest(BaseSoSTest):
                 self._manifest = json.loads(content)
             except Exception:
                 self._manifest = ''
-                self.warning('Could not load manifest for test')
+                self.log.warn('Could not load manifest for test')
         return self._manifest
 
     @property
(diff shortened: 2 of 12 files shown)


sos/policies/auth/__init__.py
```diff
@@ -67,7 +67,8 @@ def _request_device_code(self):
             res = requests.post(
                 self.client_identifier_url,
                 data=data,
-                headers=headers)
+                headers=headers,
+                timeout=300)
```
pponnuvel, line 71: I am unsure what the right timeout value can be; picked 300 to be a conservative value. Can increase/decrease depending on feedback.

arif-ali: We have a `TIMEOUT_DEFAULT = 300` in `sos/utilities.py`, maybe we can utilise that value? This is already set to 300

tests/vendor_tests/redhat/rhbz1928628.py
```diff
@@ -7,7 +7,8 @@
 # See the LICENSE file in the source distribution for further information.
 
 
-from report_tests.plugin_tests.networking import NetworkingPluginTest
+from tests.report_tests.plugin_tests.networking.networking import \
+    NetworkingPluginTest
```
arif-ali, line 11: When the tests are run, we already have `PYTHONPATH=tests`, so curious why we need this change?

sos/policies/auth/__init__.py
jcastill: 300 or TIMEOUT_DEFAULT should be enough as a starting point for RH auths

tests/vendor_tests/redhat/rhbz1928628.py
pponnuvel: This is to fix `E0401: Unable to import 'report_tests.plugin_tests.networking.networking' (import-error)`. If I set `PYTHONPATH=tests`, it still complains because of missing module `networking` (i.e. `tests` can be dropped but the extra `networking` is still needed).

Since this isn't currently raised in github CI, I guess I can drop this change altogether. But still wonder if this test is run at all as I can't see how it'd work without `networking` at the end.

sos/policies/auth/__init__.py
pponnuvel: ack, used TIMEOUT_DEFAULT from sos/utilities.py

tests/vendor_tests/redhat/rhbz1928628.py
arif-ali: cool, sounds good to me

tests/sos_tests.py
```diff
@@ -389,7 +389,7 @@ def manifest(self):
                 self._manifest = json.loads(content)
             except Exception:
                 self._manifest = ''
-                self.warning('Could not load manifest for test')
+                self.log.debug('Could not load manifest for test')
```
pmoravec, line 392: Why warning verbosity replaced by debug, please?

pponnuvel: It's to fix:
```
tests/sos_tests.py:392:16: E1101: Instance of 'BaseSoSReportTest' has no 'warning' member (no-member)
```

TurboTurtle: I believe that should be `warn` in that case, instead of `warning`.

pponnuvel: Doesn't work: https://github.com/sosreport/sos/pull/3747/checks?check_run_id=28716881665

pponnuvel: I don't see any `warn` or `warning` method anywhere in the class chain. I am also not sure where this message is *supposed* to go. So if `self.log.debug` isn't good, maybe just `print` is ok?

arif-ali: Try `self.log.warn`

pponnuvel: `self.log.warn` works.

arif-ali: basically, is coming from the Test class from avocado. If you look through that class; you'll see the following line

```
self.log.warn = self.log.warning = record_and_warn
```

So, you could have used `self.log.warn` or `self.log.warning`

pponnuvel: Thanks, that makes sense. Should be good now then.

arif-ali (approved):

TurboTurtle (approved):

PR #3748, opened 2024-08-11
[pylint] Fix unspecified-encoding

pponnuvel:
---

### DIFF
--- a/sos/utilities.py
+++ b/sos/utilities.py
@@ -113,7 +113,7 @@ def fileobj(path_or_file, mode='r'):
     """Returns a file-like object that can be used as a context manager"""
     if isinstance(path_or_file, str):
         try:
-            return open(path_or_file, mode)
+            return open(path_or_file, mode, encoding='utf-8')
         except IOError:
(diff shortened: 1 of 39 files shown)


sos/utilities.py
```diff
@@ -275,7 +275,8 @@ def _check_poller(proc):
         else:
             expanded_args.append(arg)
     if to_file:
-        _output = open(to_file, 'w')  # pylint: disable=consider-using-with
+        # pylint: disable=consider-using-with
```
TurboTurtle, line 278: This diable shouldn't be needed since we're now specifying the encoding, right?

TurboTurtle (commented):
Looks good up to my one note.

sos/utilities.py
pponnuvel: That's a different warning (`consider-using-with`).

TurboTurtle: Oh, duh. Sorry, went cross-eyed there I guess.

TurboTurtle (approved):

arif-ali (approved):

PR #3755, opened 2024-08-18
[pylint] Fix raise-missing-from

pponnuvel:
---

### DIFF
--- a/sos/options.py
+++ b/sos/options.py
@@ -7,8 +7,7 @@
 # See the LICENSE file in the source distribution for further information.
 
 from argparse import Action
-from configparser import (ConfigParser, ParsingError, Error,
-                          DuplicateOptionError)
+from configparser import ConfigParser
 
 
 def _is_seq(val):
@@ -169,10 +168,7 @@ class SoSOptions():
                     "analagous")
             return val
         if isinstance(self.arg_defaults[key], int):
-            try:
-                return int(val)
-            except ValueError:
-                raise Exception(f"Value of '{key}' in {conf} must be integer")
+            return int(val)
         return val
 
     def update_from_conf(self, config_file, component):
@@ -220,15 +216,8 @@ class SoSOptions():
 
         config = ConfigParser()
         try:
-            try:
-                with open(config_file, encoding='utf-8') as f:
-                    config.read_file(f, config_file)
-            except DuplicateOptionError as err:
-                raise Exception(f"Duplicate option '{err.option}' in section "
-                                f"'{err.section}' in file {config_file}")
-            except (ParsingError, Error):
-                raise Exception(f'Failed to parse configuration file '
-                                f'{config_file}')
+            with open(config_file, encoding='utf-8') as f:
+                config.read_file(f, config_file)
         except (OSError, IOError) as e:
             print(
                 f'WARNING: Unable to read configuration file {config_file} : '
--- a/sos/policies/distros/__init__.py
+++ b/sos/policies/distros/__init__.py
@@ -693,9 +693,9 @@ class LinuxPolicy(Policy):
         # via ssh python bindings commonly available among downstreams
         try:
             import pexpect
-        except ImportError:
+        except ImportError as err:
             raise Exception('SFTP upload requires python3-pexpect, which is '
-                            'not currently installed')
+                            'not currently installed') from err
 
         sftp_connected = False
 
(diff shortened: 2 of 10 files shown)


pponnuvel:
Generally removed the exception handling in form:
```
try:
    something
except X:
    raise Exception("X's description")
```

as throwing X otherwise would be fine.

sos/report/__init__.py
```diff
@@ -1257,7 +1257,7 @@ def setup(self):
                 plug.manifest.add_field('setup_end', end)
                 plug.manifest.add_field('setup_time', end - start)
             except KeyboardInterrupt:
-                raise KeyboardInterrupt
+                raise KeyboardInterrupt  # pylint: disable=raise-missing-from
```
pponnuvel, line 1260: Just a `raise` would be enough but that triggers `try-except-raise`.

Removing 
```
except KeyboardInterrupt:
                raise KeyboardInterrupt
```
altogether would mean `KeyboardInterrupt` would fall into 
```
except Exception:
                self.handle_exception(plugname, "setup")
```

which handles the exception differently.

sos/policies/distros/__init__.py
```diff
@@ -691,11 +691,7 @@ def upload_sftp(self, user=None, password=None):
         # soft dependency on python3-pexpect, which we need to use to control
         # sftp login since as of this writing we don't have a viable solution
         # via ssh python bindings commonly available among downstreams
-        try:
-            import pexpect
-        except ImportError:
-            raise Exception('SFTP upload requires python3-pexpect, which is '
-                            'not currently installed')
+        import pexpect
```
pmoravec, line 694: When `python3-pexpect` is not installed, this will lead to message

    Upload attempt failed: No module named 'pexpect'

that does not tell the user how to fix it. Could the original error/exception text be preserved?

pponnuvel: ack, I've put it back now. Thanks.

pponnuvel:
Not sure if something went wrong with the CI/test env. It's nearly 5 hours and it's still not done...

sos/options.py
```diff
@@ -220,15 +216,8 @@ def _update_from_section(section, config):
...
         try:
-            try:
-                with open(config_file, encoding='utf-8') as f:
-                    config.read_file(f, config_file)
-            except DuplicateOptionError as err:
-                raise Exception(f"Duplicate option '{err.option}' in section "
-                                f"'{err.section}' in file {config_file}")
-            except (ParsingError, Error):
-                raise Exception(f'Failed to parse configuration file '
-                                f'{config_file}')
```
TurboTurtle, line 231: What's the motivation behind removing these exception messages?

sos/policies/distros/__init__.py
```diff
@@ -900,31 +894,24 @@ def upload_ftp(self, url=None, directory=None, user=None, password=None):
                 raise Exception("connection failed, did you set a user and "
                                 "password?")
             session.cwd(directory)
-        except socket.timeout:
-            raise Exception(f"timeout hit while connecting to {url}")
-        except socket.gaierror:
-            raise Exception(f"unable to connect to {url}")
```
TurboTurtle, line 906: Same here, why are we dropping these handlers and the messaging that is more informative for end users?

TurboTurtle (commented):
Overall I'm onboard, but for a few specific removals I'm not sure what we gain.

sos/options.py
pponnuvel: I felt the catching and repeating the same exception message is just verbose for no additional info.

DuplicateSectionError:
```
  File "<stdin>", line 2, in <module>
  File "/usr/lib/python3.12/configparser.py", line 705, in read_file
    self._read(f, source)
  File "/usr/lib/python3.12/configparser.py", line 1049, in _read
    raise DuplicateSectionError(sectname, fpname,
configparser.DuplicateSectionError: While reading from './conf' [line 11]: section 'SectionTwo' already exists
```

ParsingError:
```
  File "<stdin>", line 2, in <module>
  File "/usr/lib/python3.12/configparser.py", line 705, in read_file
    self._read(f, source)
  File "/usr/lib/python3.12/configparser.py", line 1096, in _read
    raise e
configparser.ParsingError: Source contains parsing errors: './conf'
	[line  2]: 'name\n'
```

So the default exceptions provide equivalent info already.

sos/policies/distros/__init__.py
pponnuvel: The `url` is displayed by the default exceptions, so think that's a useful info - restored these two.

sos/options.py
TurboTurtle: I'd argue that `[line 11]: section 'SectionTwo' already exists` is vastly more helpful than `[line  2]: 'name\n'` as an error message.

TurboTurtle: Ack, fair point.

pmoravec (approved):

TurboTurtle (approved):

PR #3756, opened 2024-08-19
[redhat] Fix return of function check_file_too_big()

jcastill:
Fix the return of the size checks for Red Hat uploads to customer portal. Without this fix, uploads were failing silently.

---


pmoravec (approved):
ACK. Without this change, we set `self.upload_url` "too early and wrong" - original call flow set it later on via `get_upload_url` method, and to `RH_API_HOST + rh_case_api % self.case_id`.

TurboTurtle (approved):

PR #3758, opened 2024-08-22
[Plugin] Collect exact filesizes in add_dir_listing

pmoravec:
`add_dir_listing` method in #3664 hid exact filesizes in "`ls -l`" like outputs in favour of human readable output. That has two drawbacks:

1) It provides less precise information, while particular filesizes for some files can be essential to know. E.g. `3.1K` can mean `3075` or `3174` bytes and the difference can matter for some well known files with well known default sizes.
2) It makes it difficult to parse the size by various analytical tools

`sos` tool should be data collectio tool rather than interpreting them. So let replace human readable filesize by exact one.

Resolves: #3758

---

### DIFF
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -2058,9 +2058,9 @@ class Plugin():
         paths = [p for p in paths if self.path_exists(p)]
 
         if not tree:
-            options = f"alhZ{'R' if recursive else ''}"
+            options = f"alZ{'R' if recursive else ''}"
         else:
-            options = 'lhp'
+            options = 'lp'
 
         for path in paths:
             self.add_cmd_output(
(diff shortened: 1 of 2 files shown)


TurboTurtle:
Should this be a toggle? When I standardized the existing calls to `add_dir_listing()` there was no clear preference across plugins as to if `-h` was used or not in the "manual" `ls` collections.

I'm also fine with just dropping human sizes across the board, just a thought.

pmoravec:
Good point. Sometimes (e.g. https://github.com/sosreport/sos/pull/3664/files#diff-2e1b7ad69a0e01805218f011055c3ce7d123809531eed5ed157e004be5fe315bR175 or https://github.com/sosreport/sos/pull/3664/files#diff-b8035c7551e828c6ec017b8b2a986cce1884a2d87110a9edfa11bbdf4c40f561R57) the `ls` output is meant rather user-informative than precisely descriptive. But still I feel making the `-h` option toggle-able would "cost" us on code (execution) complexity more than users would gain. So until there would be a push back, I would stay with the `-h` drop.

I will just bump the PR to update the failing test.

TurboTurtle (approved):

arif-ali (approved):

PR #3763, opened 2024-08-28
[bird, networking] Carry forward PR 1541

TurboTurtle:
Two commits that first add the `bird` plugin and second add several `devlink` commands to the `networking` plugin, originally proposed in #1541, which seems to have been abandoned.

Commits have been updated to reflect modern sos/plugin design and current state of the networking plugin.

Closes #1541 

---

### DIFF
--- a/sos/report/plugins/bird.py
+++ b/sos/report/plugins/bird.py
@@ -1,0 +1,70 @@
+# Copyright (C) 2024 Jake Hunsaker <jacob.r.hunsaker@gmail.com>
+# Copyright (C) 2019 Alexander Petrovskiy <alexpe@mellanox.com>
+#
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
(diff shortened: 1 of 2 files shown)


jcastill (commented):
There could be some passwords stored in the config file. Inside bird.conf we have two examples:


#		password "hello" { algorithm hmac sha256; }; # Default is MD5
#	password "secret";	# Password used for MD5 authentication

First one is for protocol RIP, second one is for BGP. The format seems to be very similar if not the same, so it should be easy to obfuscate these lines.

Also I think the user can specify a log file instead of syslog, so it may be worth to check the config file for an entry with a pattern that starts like this:

log "filename"

TurboTurtle:
Interesting, the example config file I was looking at had no passwords but sure enough the upstream docs call that out.

Pushed a new set with an added `postproc()` and grabbing the log files called out in `bird.conf`.

pmoravec (approved):

jcastill (approved):

PR #3764, opened 2024-08-28
[Policy] Consolidate check() across policies

TurboTurtle:
There is a common theme for new policies to subclass an existing one and just update the `check()` method. In almost all instances, the `check()` method remains almost entirely the same, simply changing a release file name or string.

Reduce repeating ourselves by consolidating this into the base `LinuxPolicy.check()`, and leverage per-policy class attributes `os_release_file`, `os_release_name`, and `os_release_id`. The first of which will enable a policy if the specified file is found. Failing that, `os_release_name` is used to check the `NAME` field in `/etc/os-release`, and `os_release_id` is used to check the `ID` field in the same file if the name pattern does not match (or is not set).

Note that we do _not_ check the `ID_LIKE` field.

This allows a policy to be defined with as little as something like the following:

```
class MyPolicy(LinuxPolicy):
    distro = 'My Awesome Linux'
    vendor = 'Myself'
    os_release_file = '/etc/mylinux-release'
    os_release_name = 'MyLinux'
    os_release_id ='mynix'
```

Resolves: #1934

---

### DIFF
--- a/sos/policies/distros/debian.py
+++ b/sos/policies/distros/debian.py
@@ -6,17 +6,16 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-import os
-
 from sos.report.plugins import DebianPlugin
 from sos.policies.distros import LinuxPolicy
 from sos.policies.package_managers.dpkg import DpkgPackageManager
 
 
 class DebianPolicy(LinuxPolicy):
-    distro = "Debian"
     vendor = "the Debian project"
     vendor_urls = [('Community Website', 'https://www.debian.org/')]
+    os_release_name = 'Debian'
+    os_release_file = '/etc/debian_version'
     name_pattern = 'friendly'
     valid_subclasses = [DebianPlugin]
     PATH = "/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games" \
@@ -49,16 +48,6 @@ class DebianPolicy(LinuxPolicy):
         return {
             "xz": "xz-utils"
         }.get(binary, binary)
-
-    @classmethod
-    def check(cls, remote=''):
-        """This method checks to see if we are running on Debian.
-           It returns True or False."""
-
-        if remote:
-            return cls.distro in remote
-
-        return os.path.isfile('/etc/debian_version')
 
     def dist_version(self):
         try:
--- a/sos/policies/distros/redhat.py
+++ b/sos/policies/distros/redhat.py
@@ -18,7 +18,7 @@ from sos.report.plugins import RedHatPlugin
 from sos.presets.redhat import (RHEL_PRESETS, RHV, RHEL, CB, RHOSP,
                                 RHOCP, RH_CFME, RH_SATELLITE, AAPEDA,
                                 AAPCONTROLLER)
-from sos.policies.distros import LinuxPolicy, ENV_HOST_SYSROOT
+from sos.policies.distros import LinuxPolicy, ENV_HOST_SYSROOT, OS_RELEASE
 from sos.policies.package_managers.rpm import RpmPackageManager
 from sos.policies.package_managers.flatpak import FlatpakPackageManager
 from sos.policies.package_managers import MultiPackageManager
@@ -31,12 +31,10 @@ try:
 except ImportError:
     REQUESTS_LOADED = False
 
-OS_RELEASE = "/etc/os-release"
 RHEL_RELEASE_STR = "Red Hat Enterprise Linux"
 
 
 class RedHatPolicy(LinuxPolicy):
-    distro = "Red Hat"
     vendor = "Red Hat"
     vendor_urls = [
         ('Distribution Website', 'https://www.redhat.com/'),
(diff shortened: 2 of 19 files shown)


TurboTurtle:
I tested this on Fedora 39, RHEL 9, and Ubuntu 24.04 and the correct policy loaded on all of them. Additional sanity checks would be appreciated.

arif-ali:
I've created a deb package specifically for this in Launchpad, to test this end-to-end on 24.04, and work as expected.

The manifist.json shows the policy Ubuntu being used.
The `--upload` worked as expected.

I'll create a few more packages for prior releases next week, and do further testing.

sos/policies/distros/debian.py
```diff
@@ -17,6 +15,7 @@ class DebianPolicy(LinuxPolicy):
     distro = "Debian"
     vendor = "the Debian project"
     vendor_urls = [('Community Website', 'https://www.debian.org/')]
+    os_release_file = '/etc/debian_version'
```
pmoravec, line 18: Shouldn't we also set `os_release_name = 'Debian'`? Otherwise the variable is nested from `LinuxPolicy` with `''` value, and that can make troubles in `def _check_release(content):` method.

Anyway, see my more general question in PR.

pmoravec:
Why to distinguish `distro` and `os_release_name`, as we always check against both variables, and they are usually the same (I frown on the ambiguous Azure class :) )?

Isn't it easier to have `os_release_names` as a list (always), where the very first item plays the role of `distro`?

sos/policies/distros/debian.py
arif-ali: In this scenario, we're only checking for the existence of the file. If that exists then the debian policy is used. 

So, I think it makes sense, and we don't necessarily need the name.

Nevertheless, I can also check some debian variants too and check back here too

sos/policies/distros/redhat.py
```diff
@@ -240,34 +230,6 @@ def __init__(self, sysroot=None, init=None, probe_runtime=True,
...
-        if not os.path.exists(OS_RELEASE):
-            return False
-
-        with open(OS_RELEASE, "r", encoding='utf-8') as f:
-            for line in f:
-                if line.startswith("NAME"):
-                    (_, value) = line.split("=")
-                    value = value.strip("\"'")
-                    if value.startswith(cls.distro):
-                        return True
```
pmoravec, line 268: We are changing this to direct `value`s match, but that is imho correct. `NAME` should be `"Red Hat Enterprise Linux Server"`.

sos/policies/distros/suse.py
```diff
@@ -62,6 +62,7 @@ class OpenSuSEPolicy(SuSEPolicy):
     distro = "OpenSuSE"
     vendor = "SuSE"
     vendor_urls = [('Community Website', 'https://www.opensuse.org/')]
+    os_release_file = '/etc/SUSE-brand'
```
pmoravec, line 65: Same here wrt missing `os_release_name(s)`.

sos/policies/distros/debian.py
arif-ali: I think the lines below, that were added already tackle this in the same way as previously. As long as the file exists, then the Debian policy is activated.

```python
        # use the os-specific file primarily
        if os.path.isfile(cls.os_release_file):
            return True
```

Checking 4 releases of Debian, `sid`, `trixie`, `bookworm` and `bullseye`; they all have this file.

Alternatively, if we only want to use `/etc/os-release`, then we could have

```python
os_release_id="debian"
os_release_name="Debian GNU/Linux"
```

With the new package always going into `sid` and/or `trixie`, this will only really impact those OSs. The debian package is typically not back-ported to older releases. At least from what I can see based on my interaction these past 2 months

arif-ali:
OK, this doesn't work for debian.

Ultimately, checking through the policies, it will try to check all. As the Ubuntu policy inherits the variables and such from Debian, this now suggests that the policy being used is Ubutu.

We may need to add the following to the `UbuntuPolicy`, at least this worked for me.

```python
os_release_file = ''
```

sos/policies/distros/suse.py
arif-ali: This is valid, if we're only looking for the location of the file as is, and no `id`/`name`

arif-ali:
> I'll create a few more packages for prior releases next week, and do further testing.

Tested on `focal`, `jammy` and `noble`, works as expected and policy is Ubuntu in the manifest

TurboTurtle:
> Why to distinguish `distro` and `os_release_name`, as we always check against both variables, and they are usually the same (I frown on the ambiguous Azure class :) )?
> 
> Isn't it easier to have `os_release_names` as a list (always), where the very first item plays the role of `distro`?
> 
> Upon to that, ACK from me and good idea.

There are a few places where we, intentionally or not, use `distro` as a shorthand reference for a particular distribution. We could very well drop `distro` I think with little churn. Let me poke around a bit more and if everything checks out like I'm thinking it will, I'll push an updated branch. Likely tomorrow as I am currently sitting in an airport.

TurboTurtle:
I've updated the branch to entirely remove `distro`, and replace it with references to `os_release_name`. I thought about allowing this to be a list, but the implementation gets murky in my opinion. The only policy that has/had more than one name in this context was `AzureLinux`, which carried the CBL Mariner name since that policy was actually previously renamed from the Marine policy. In this case I simply reinstated a CBL policy whose only difference is the `os_release_name` value.

We can certainly revisit this if we start getting a need for multiple policy names, but given the existing precedent it didn't seem necessary.

TurboTurtle:
> OK, this doesn't work for debian.
> 
> Ultimately, checking through the policies, it will try to check all. As the Ubuntu policy inherits the variables and such from Debian, this now suggests that the policy being used is Ubutu.
> 
> We may need to add the following to the `UbuntuPolicy`, at least this worked for me.
> 
> ```python
> os_release_file = ''
> ```

Added in the latest push.

arif-ali:
Built updated packages in the [PPA](https://launchpad.net/~arif-ali/+archive/ubuntu/sos-pr-3764) again, will test in both Debian and Ubuntu this week sometime

arif-ali:
Tested this on both Ubuntu and Debian, and manifest.json now looks good to me

arif-ali (approved):

pmoravec (approved):

PR #3765, opened 2024-08-28
[sudo] Capture sudo and sudoers debug log files

jcastill:
Capture sudo and sudoers debug log files.

---

### DIFF
--- a/sos/report/plugins/sudo.py
+++ b/sos/report/plugins/sudo.py
@@ -21,6 +21,22 @@ class Sudo(Plugin, IndependentPlugin):
     def setup(self):
         self.add_copy_spec("/etc/sudo*")
 
+        config_file = "/etc/sudo.conf"
+        log_files = ['/var/log/sudo_debug', '/var/log/sudoers_debug']
+        try:
+            with open(config_file, 'r', encoding='UTF-8') as cfile:
+                for line in cfile:
+                    if line.startswith('Debug'):
+                        log_files.append(line.split()[2])
+        except IOError as error:
+            self._log_error(f'Could not open conf file {config_file}: '
+                            f'{error}')
+
+        if not self.get_option('all_logs'):
+            self.add_copy_spec(log_files)
+        else:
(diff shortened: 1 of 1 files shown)


sos/report/plugins/sudo.py
```diff
@@ -21,6 +21,29 @@ class Sudo(Plugin, IndependentPlugin):
...
+                            log_file_sudoers = words[2].strip()
+        except IOError as error:
+            self._log_error(f'Could not open conf file {config_file}: '
+                            f'{error}')
+
+        if not self.get_option("all_logs"):
+            self.add_copy_spec([log_file_sudo, log_file_sudoers])
+        else:
+            self.add_copy_spec([log_file_sudo+'*', log_file_sudoers+'*'])
+
```
TurboTurtle, line 46:
```python
log_files = ['/var/log/sudo_debug', '/var/log/sudoers_debug']
with open(config_file, 'r', encoding='UTF-8') as cfile:
    for line in cfile:
        if line.startswith('Debug'):
            log_files.append(line.split()[2])

if not self.get_option('all_logs'):
    self.add_copy_spec(log_files)
else:
    self.add_copy_spec([f"{log}*" for log in log_files]
```

We don't seem to actually care which file is for sudo or sudoers, as we are just passing the filepaths on regardless. With this we also get the other subsystems which are probably valuable if someone has gone through the effort of separating them out. And if the defaults don't exist, `add_copy_spec()` gracefully skips them.

With the above, we also don't spuriously split the line before we determine if we care about the line by it being started with the `Debug` keyword. Also, on my Fedora and RHEL boxes locally none of the config files use `=` as a delimiter, so not sure why we want to split on that?

jcastill: Nice, I'll make these changes. 
Regarding the `=` that was an oversight on my side when copying from another plugin. Fixed in the next force push.

pmoravec (approved):

arif-ali (approved):

TurboTurtle (approved):

PR #3768, opened 2024-09-01
[sos_extras] Allow obfuscation in sos_extras configs

pmoravec:
Resolves: #3768

---

### DIFF
--- a/sos/report/plugins/sos_extras.py
+++ b/sos/report/plugins/sos_extras.py
@@ -19,14 +19,21 @@ class SosExtras(Plugin, IndependentPlugin):
     it executes commands or collects files optionally with sizelimit. Expected
     content of a file:
     - empty lines or those starting with '#' are ignored
-    - add_copy_spec called to lines starting by ':', optionally followed by
-      sizelimit
+    - lines starting by ':' are treated as files to copy, optionally followed
+      by sizelimit
+    - lines starting by '*' are treated as obfuscation of a secret (postproc)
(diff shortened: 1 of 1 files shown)


pmoravec:
Limitations:
- different globs for files/cmds to collect (`/path/to/file.log*`) and to obfuscate in (`/path/to/file.log.*`) - this is user unfriendly
- no way to express space in either RE (to search or to replace)
- example line `*:/path/to/files(.*) pass:\\s*(.*) \1********` should have `\s` instead of `\\s` but that would raise `W605 invalid escape sequence '\s'` :(

Worth to discuss:
- particular syntax

TurboTurtle:
I'm torn on this.

On the one hand, it is useful functionality and I can see the practicality. On the other, we don't really want to encourage using sos_extras as an _alternative_ to plugin development. It's chiefly meant to serve as a function of "I need this right away while a plugin is being updated or developed which is either currently happening or I am about to open a request for".

pmoravec:
The request comes from a customer having some logs written to an unusual customized location that `sos` will never collect. But the logfile is important for troubleshooting. Like, e.g. `/var/log/httpd/access_log`  is moved to `/var/log/httpd/my_custom_access_log`.

The httpd access log is just ilustrative example. I will check if we can't (or even shouldn't) parse some config file that specifies such custom logfiles, and collect the custom logfiles automatically.

pmoravec:
I understand (and mostly agree). I will close the PR as going against the original intention of `sos_extras`.

PR #3769, opened 2024-09-02
[specfile] specify GPL v2 only

sandrobonazzola:
According to https://spdx.org/licenses/GPL-2.0.html "GPL-2.0" is deprecated and "GPL-2.0-only" must be used instead. See: https://spdx.org/licenses/GPL-2.0-only.html

---

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -8,7 +8,7 @@ grade: stable
 base: core24
 confinement: classic
 adopt-info: sos
-license: GPL-2.0
+license: GPL-2.0-only
 environment:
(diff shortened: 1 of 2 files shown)


arif-ali:
That was my bad, and good catch

Could you also update the license field in `snap/snapcraft.yaml` too, based on https://snapcraft.io/docs/snapcraft-yaml-reference#license we should be doing the same there

sandrobonazzola:
> That was my bad, and good catch
> 
> Could you also update the license field in `snap/snapcraft.yaml` too, based on https://snapcraft.io/docs/snapcraft-yaml-reference#license we should be doing the same there

Done

arif-ali (approved):

TurboTurtle (approved):

PR #3770, opened 2024-09-02
[specfile] drop python-rpm dependency

sandrobonazzola:
rpm module is not being used by sos.

---


arif-ali (approved):

TurboTurtle:
This makes me wonder if there'd be a benefit to using the rpm module in conjunction with the `RpmPackageManager` abstraction we use, but that requires a much more thorough investigation as it would likely require us to do something similar for dpkg at the least, but that then gives us even more packaging dependencies at the end of that day. Writing that out, sounds like it's not worth it.

Ack.

TurboTurtle (approved):

PR #3776, opened 2024-09-11
[leapp] improve plugin enablement and add more collected files

evgeni:
The `leapp-repository` package was renamed to `leapp-upgrade-elXtoY`
in [1], and while we could list all possible combinations here,
the main `leapp` package should be enough to trigger this plugin

While at it, also enable plugin when /var/lib/leapp exists, as
after an upgrade is performed, users often uninstall the leapp packages,
but it can still be beneficial for SOS to collect data about the fact
that this is a system that was upgraded (and data about the upgrade) and
/var/lib/leapp shows exactly that.
    
Also include /etc/migration-results and /var/log/leapp/answerfile in
collected files.
    
[1] https://github.com/oamg/leapp-repository/commit/924c398a003f8ada079f2819fa3305adda7dfcda
    
---

### DIFF
--- a/sos/report/plugins/leapp.py
+++ b/sos/report/plugins/leapp.py
@@ -16,10 +16,13 @@ class Leapp(Plugin, RedHatPlugin):
     short_desc = 'Leapp upgrade handling tool'
 
     plugin_name = 'leapp'
-    packages = ('leapp', 'leapp-repository')
+    packages = ('leapp',)
+    files = ('/var/lib/leapp',)
 
     def setup(self):
         self.add_copy_spec([
+            '/etc/migration-results',
+            '/var/log/leapp/answerfile',
             '/var/log/leapp/dnf-debugdata/',
             '/var/log/leapp/leapp-preupgrade.log',
             '/var/log/leapp/leapp-upgrade.log',


sos/report/plugins/leapp.py
```diff
@@ -16,10 +16,12 @@ class Leapp(Plugin, RedHatPlugin):
     short_desc = 'Leapp upgrade handling tool'
 
     plugin_name = 'leapp'
-    packages = ('leapp', 'leapp-repository')
+    packages = ('leapp',)
+    files = ('/var/lib/leapp',)
 
     def setup(self):
         self.add_copy_spec([
+            '/etc/migration-results',
```
pirat89, line 24: does sosreport check whether the file exists? so it's not failing when the file is not present yet. just checking to be sure.

I wuold possibly extende it right away to include also:
```
/var/log/leapp/answerfile
/var/log/leapp/dnf-plugin-data.txt
```

evgeni: it does not fail when files are absent, no :)

evgeni: `dnf-plugin-data.txt` is already here, just scroll down a bit.
I  can add answerfile, sure

arif-ali:
Hi @evgeni can you squash your commits please, before we can land the PR, TiA

Also, just making sure, does the answerfile, or any new files that are being collected have passwords that need obfuscating?

evgeni:
@arif-ali I can sure, just thought it's cleaner if they are separate as they change different aspects of the plugin

And no, to the best of my knowledge, the answers file should not contain any sensitive data, right @pirat89?

sos/report/plugins/leapp.py
```diff
@@ -16,10 +16,13 @@ class Leapp(Plugin, RedHatPlugin):
     short_desc = 'Leapp upgrade handling tool'
 
     plugin_name = 'leapp'
-    packages = ('leapp', 'leapp-repository')
+    packages = ('leapp',)
```
jcastill, line 19: In which version of RHEL did we stop distributing leapp-repository? 
Or in other words - will this stop the plugin executing in old RHELs where migration could potentially still happen?

evgeni: https://github.com/oamg/leapp-repository/commit/924c398a003f8ada079f2819fa3305adda7dfcda is in leapp-repository (SRPM name!) 0.15+, which is RHEL 7.9+
Pretty sure older RHEL 7.x is not supported anymore for leapp, and even if: the `leapp` package is present in any case (unless the user uninstalled it, ofc).

jcastill: Perfect, thank you for clarifying this!

evgeni:
squashed

arif-ali (approved):

TurboTurtle (approved):

PR #3777, opened 2024-09-12
[nvidia] Capture more nvidia commands

jcastill:
Capture commands related to nvidia container toolkit.

Related: RHEL-58172

---

### DIFF
--- a/sos/report/plugins/nvidia.py
+++ b/sos/report/plugins/nvidia.py
@@ -16,9 +16,13 @@ class Nvidia(Plugin, IndependentPlugin):
 
     short_desc = 'Nvidia GPU information'
     plugin_name = 'nvidia'
-    commands = ('nvidia-smi',)
+    commands = ('nvidia-smi', 'nvidia-ctk',)
+    services = ('nvidia-persistenced', 'nvidia-fabricmanager',
+                'nvidia-toolkit-firstboot')
 
     def setup(self):
+        self.add_copy_spec("/etc/cdi/nvidia.yaml")
+
         subcmds = [
             '--list-gpus',
             '-q -d PERFORMANCE',
@@ -29,9 +33,12 @@ class Nvidia(Plugin, IndependentPlugin):
             'nvlink -s',
             'nvlink -e'
         ]
-
-        self.add_service_status("nvidia-persistenced")
+        ctk_subcmds = [
+            'cdi list',
+            '--version',
+        ]
         self.add_cmd_output([f"nvidia-smi {cmd}" for cmd in subcmds])
+        self.add_cmd_output([f"nvidia-ctk {cmd}" for cmd in ctk_subcmds])
 
         query = ('gpu_name,gpu_bus_id,vbios_version,temperature.gpu,'
                  'utilization.gpu,memory.total,memory.free,memory.used,'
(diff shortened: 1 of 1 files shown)


pmoravec (approved):
Sounds good to me. I re-started timeouted tests that failed.

sos/report/plugins/nvidia.py
```diff
@@ -30,8 +30,17 @@ def setup(self):
...
 
+        ctk_subcmds = [
+            'cdi list',
+            '--version',
+        ]
+
+        self.add_copy_spec("/etc/cdi/nvidia.yaml")
         self.add_service_status("nvidia-persistenced")
+        self.add_service_status("nvidia-fabricmanager")
+        self.add_service_status("nvidia-toolkit-firstboot")
```
arif-ali, line 41: curious, would it good to add journal for all these too. i.e. if one only wants to enable the `nvidia` plugin and nothing else?

TurboTurtle: If these are added to the `services` tuple, it can be used for plugin enablement as well as automatically getting the journal and service status.

jcastill: Good idea. Shall I add the three services to the plugin enablement? We have its journal capture a bit different:

`self.add_journal(boot=0, identifier='nvidia-persistenced')`

But a quick test I did in a RHEL AI image showed that the nvidia-persistenced service is captured correctly. My only doubt is the 'boot=0' option.

pmoravec (approved):

sos/report/plugins/nvidia.py
```diff
@@ -44,4 +53,5 @@ def setup(self):
         )
         self.add_journal(boot=0, identifier='nvidia-persistenced')
 
+
```
arif-ali, line 56: nit, do we need the extra line here? not overly precious about it though

arif-ali (approved):
lgtm

sos/report/plugins/nvidia.py
jcastill: Good point, I think we can remove it. 
My only doubt then is what to do with this add_journal() call before it. The service is now in the services tuple, but the boot=0 is what makes me doubt if it's safe or not to remove it. Any thoughts on this?

jcastill:
Thank you for clarifying it Jake. I'll remove the extra capture and will push again.

sos/report/plugins/nvidia.py
```diff
@@ -30,8 +32,15 @@ def setup(self):
             'nvlink -e'
         ]
 
-        self.add_service_status("nvidia-persistenced")
+        ctk_subcmds = [
+            'cdi list',
+            '--version',
+        ]
+
+        self.add_copy_spec("/etc/cdi/nvidia.yaml")
```
pmoravec, line 40: Just a nitpick, for the sake of code/plugin readiness, I would move the `a_s_c` to the beginning of `setup(self)` - or at least outside of the logical block:

    subcmds = ..
    ctk_subcmds = ..
    # something completely else here..?
    add_cmd_output( .. subcmds ..)
    add_cmd_output( .. ctk_subcmds ..)

jcastill: Makes sense, I'll do that

pmoravec (approved):

arif-ali (approved):

TurboTurtle (approved):

Issue #3780, opened 2024-09-17
Plugins not added to sos.html


jpbruijstens:
When using sosreport version 4.5.6, all plugins are listed in sos.html as follows: \<h2 id="tunnel">Plugin \<em>tunnel\</em>\</h2>

When using sosreport version 4.8.0 no plugins are not listed in sos.html anymore.

jcastill:
Thank you for reporting this. Indeed there's a difference - this is what I have in 4.6.0:

```
<h2 id="alternatives">Plugin <em>alternatives</em></h2>
<p>Commands executed:</p><ul>
<li><a href="../sos_commands/alternatives/alternatives_--display_arptables">alternatives --display arptables</a></li>
<li><a href="../sos_commands/alternatives/alternatives_--display_cifs-idmap-plugin">alternatives --display cifs-idmap-plugin</a></li>
<li><a href="../sos_commands/alternatives/alternatives_--display_ebtables">alternatives --display ebtables</a></li>
<li><a href="../sos_commands/alternatives/alternatives_--display_iptables">alternatives --display iptables</a></li>
<li><a href="../sos_commands/alternatives/alternatives_--display_libnssckbi.so.x86_64">alternatives --display libnssckbi.so.x86_64</a></li>
<li><a href="../sos_commands/alternatives/alternatives_--display_man">alternatives --display man</a></li>
<li><a href="../sos_commands/alternatives/alternatives_--display_man.7.gz">alternatives --display man.7.gz</a></li>
<li><a href="../sos_commands/alternatives/alternatives_--display_nc">alternatives --display nc</a></li>
<li><a href="../sos_commands/alternatives/alternatives_--display_soelim">alternatives --display soelim</a></li>
<li><a href="../sos_commands/alternatives/alternatives_--list">alternatives --list</a></li>
<li><a href="../sos_commands/alternatives/alternatives_--version">alternatives --version</a></li>
</ul>
<hr/>
```

And this is what I have in 4.7.2:

```
alternatives
<p>Commands executed:</p><ul>
<li><a href="../sos_commands/alternatives/alternatives_--display_arptables">alternatives --display arptables</a></li>
<li><a href="../sos_commands/alternatives/alternatives_--display_cifs-idmap-plugin">alternatives --display cifs-idmap-plugin</a></li>
<li><a href="../sos_commands/alternatives/alternatives_--display_ebtables">alternatives --display ebtables</a></li>
[...]

PR #3781, opened 2024-09-17
[report] Fix html formatting for Plugin section

jcastill:
After moving from .format() to f-strings via
57d21346b, the formatting for Plugin sections was
lost. This patch attempts to fix this while still
using f-strings.

Related: #3780

---

### DIFF
--- a/sos/report/reporting.py
+++ b/sos/report/reporting.py
@@ -136,11 +136,9 @@ class PlainTextReport:
     ALERT = "  ! %s"
     NOTE = "  * %s"
     PLUGLISTHEADER = "Loaded Plugins:"
-    PLUGLISTITEM = "  {name}"
     PLUGLISTSEP = "\n"
     PLUGLISTMAXITEMS = 5
     PLUGLISTFOOTER = ""
-    PLUGINFORMAT = "{name}"
     PLUGDIVIDER = "=" * 72
 
     subsections = (
@@ -156,6 +154,12 @@ class PlainTextReport:
     def __init__(self, report_node):
         self.report_data = sorted(dict.items(report_node.data))
 
+    def plugListHeader(self, name):
+        return f"  {name}"
+
+    def pluginFormat(self, name):
+        return f"{name}"
+
     def unicode(self):
         self.line_buf = line_buf = []
 
@@ -168,7 +172,7 @@ class PlainTextReport:
         i = 0
         plugcount = len(self.report_data)
         for section_name, _ in self.report_data:
-            line += f"  {section_name}"
+            line += self.plugListHeader(section_name)
             i += 1
             if (i % self.PLUGLISTMAXITEMS == 0) and (i < plugcount):
                 line += self.PLUGLISTSEP
@@ -177,7 +181,7 @@ class PlainTextReport:
 
         for section_name, section_contents in self.report_data:
             line_buf.append(self.PLUGDIVIDER)
-            line_buf.append(f"{section_name}")
+            line_buf.append(self.pluginFormat(section_name))
             for type_, format_, header, footer in self.subsections:
                 self.process_subsection(section_contents, type_.ADDS_TO,
                                         header, format_, footer)
@@ -224,11 +228,9 @@ class HTMLReport(PlainTextReport):
     ALERT = "<li>%s</li>"
     NOTE = "<li>%s</li>"
     PLUGLISTHEADER = "<h3>Loaded Plugins:</h3><table><tr>"
-    PLUGLISTITEM = '<td><a href="#{name}">{name}</a></td>\n'
     PLUGLISTSEP = "</tr>\n<tr>"
     PLUGLISTMAXITEMS = 5
     PLUGLISTFOOTER = "</tr></table>"
-    PLUGINFORMAT = '<h2 id="{name}">Plugin <em>{name}</em></h2>'
     PLUGDIVIDER = "<hr/>\n"
 
     subsections = (
@@ -238,6 +240,12 @@ class HTMLReport(PlainTextReport):
         (Alert, ALERT,       "<p>Alerts:</p><ul>",            "</ul>"),
         (Note, NOTE,         "<p>Notes:</p><ul>",             "</ul>"),
     )
+
+    def plugListHeader(self, name):
+        return f'<td><a href="#{name}">{name}</a></td>\n'
+
+    def pluginFormat(self, name):
+        return f'<h2 id="{name}">Plugin <em>{name}</em></h2>'
 
 
 class JSONReport(PlainTextReport):


TurboTurtle (commented):
Two thoughts, no particular preference on either:

1) We are only using these in 2 places, we could simply do the f-string in-line instead of breaking it out into a function.  We haven't really made any _major_ changes to how the HTML reports are generated for in a very long time so I doubt we're going to be expanding the use of these any time soon.

2) If we keep it as a function, then we don't use camelCase in sos, and PEP8 [explicitly discourages it](https://peps.python.org/pep-0008/#function-and-variable-names)

jcastill:
> Two thoughts, no particular preference on either:
> 
>     1. We are only using these in 2 places, we could simply do the f-string in-line instead of breaking it out into a function.  We haven't really made any _major_ changes to how the HTML reports are generated for in a very long time so I doubt we're going to be expanding the use of these any time soon.
> 

Fair enough. The problem I tried to solve was for example with PLUGLISTITEM, we have a definition for plain text and another one for html, and while setting the variable via f-strings was a bit tricky. Do you know if there's a pythonic way to use f-strings with strings defined as variables like PLUGLISTITEM?
Another option, perhaps, is to set a is_html boolean var, and if that's true, then add the html code, and if it's false (i.e. coming from the plain text report) then just print the var.  Something like:

`str_plugin = [ f"{value}" if not is_html else f'<td><a href="#{value}">{value}</a></td>\n' ]`

>     2. If we keep it as a function, then we don't use camelCase in sos, and PEP8 [explicitly discourages it](https://peps.python.org/pep-0008/#function-and-variable-names)

Sorry, it seems that old habits die hard...

jcastill:
> Fair enough. The problem I tried to solve was for example with PLUGLISTITEM, we have a definition for plain text and another one for html, and while setting the variable via f-strings was a bit tricky. Do you know if there's a pythonic way to use f-strings with strings defined as variables like PLUGLISTITEM? Another option, perhaps, is to set a is_html boolean var, and if that's true, then add the html code, and if it's false (i.e. coming from the plain text report) then just print the var. Something like:
> 
> `str_plugin = [ f"{value}" if not is_html else f'<td><a href="#{value}">{value}</a></td>\n' ]`
> 

Replying to myself:  List comprehension won't be the best approach here, because we are concatenating the result. Perhaps an old fashioned if/then could be simpler.

pmoravec:
As we use the `%(name)s` syntax else-where in `reporting.py`, what about patch:

    --- a/sos/report/reporting.py
    +++ b/sos/report/reporting.py
    @@ -136,11 +136,11 @@ class PlainTextReport:
         ALERT = "  ! %s"
         NOTE = "  * %s"
         PLUGLISTHEADER = "Loaded Plugins:"
    -    PLUGLISTITEM = "  {name}"
    +    PLUGLISTITEM = "  %(name)s"
         PLUGLISTSEP = "\n"
         PLUGLISTMAXITEMS = 5
         PLUGLISTFOOTER = ""
    -    PLUGINFORMAT = "{name}"
    +    PLUGINFORMAT = "%(name)s"
         PLUGDIVIDER = "=" * 72
     
         subsections = (
    @@ -168,7 +168,7 @@ class PlainTextReport:
             i = 0
             plugcount = len(self.report_data)
             for section_name, _ in self.report_data:
    -            line += f"  {section_name}"
    +            line += self.PLUGLISTITEM % {'name': section_name }
                 i += 1
                 if (i % self.PLUGLISTMAXITEMS == 0) and (i < plugcount):
                     line += self.PLUGLISTSEP
    @@ -177,7 +177,7 @@ class PlainTextReport:
     
             for section_name, section_contents in self.report_data:
                 line_buf.append(self.PLUGDIVIDER)
    -            line_buf.append(f"{section_name}")
    +            line_buf.append(self.PLUGINFORMAT % {'name' : section_name})
                 for type_, format_, header, footer in self.subsections:
                     self.process_subsection(section_contents, type_.ADDS_TO,
                                             header, format_, footer)
    @@ -224,11 +224,11 @@ class HTMLReport(PlainTextReport):
         ALERT = "<li>%s</li>"
         NOTE = "<li>%s</li>"
         PLUGLISTHEADER = "<h3>Loaded Plugins:</h3><table><tr>"
    -    PLUGLISTITEM = '<td><a href="#{name}">{name}</a></td>\n'
    +    PLUGLISTITEM = '<td><a href="#%(name)s">%(name)s</a></td>\n'
         PLUGLISTSEP = "</tr>\n<tr>"
         PLUGLISTMAXITEMS = 5
         PLUGLISTFOOTER = "</tr></table>"
    -    PLUGINFORMAT = '<h2 id="{name}">Plugin <em>{name}</em></h2>'
    +    PLUGINFORMAT = '<h2 id="%(name)s">Plugin <em>%(name)s</em></h2>'
         PLUGDIVIDER = "<hr/>\n"
     
         subsections = (

pmoravec:
No strong preference. Just the %s-strings are quite obsolete (citation needed) and we might rather use `.format()` syntax..?

pmoravec:
Hi @jcastill , are you planning to update the PR..?

jcastill:
@pmoravec will push a new PR with your suggestion asap.

jcastill:
rpm build for Fedora 40 issue fixed by #4014

pmoravec:
> rpm build for Fedora 40 issue fixed by #4014

Could you please rebase your PR there?

jcastill:
> > rpm build for Fedora 40 issue fixed by #4014
> 
> Could you please rebase your PR there?

Ack, will do it this morning

sos/report/reporting.py
```diff
@@ -168,7 +168,7 @@ def unicode(self):
         i = 0
         plugcount = len(self.report_data)
         for section_name, _ in self.report_data:
-            line += f"  {section_name}"
+            line += self.PLUGLISTITEM % {'name': section_name}
```
pmoravec, line 171: This removes leading spaces - isnt it a problem?

sos/report/reporting.py
```diff
@@ -224,11 +224,11 @@ class HTMLReport(PlainTextReport):
     ALERT = "<li>%s</li>"
     NOTE = "<li>%s</li>"
     PLUGLISTHEADER = "<h3>Loaded Plugins:</h3><table><tr>"
-    PLUGLISTITEM = '<td><a href="#{name}">{name}</a></td>\n'
+    PLUGLISTITEM = '<td><a href="#%(name)s">%(name)s</a></td>\n'
```
pmoravec, line 227: Why not using f-string (here and for `PLUGINFORMAT`)?

sos/report/reporting.py
```diff
@@ -177,7 +177,7 @@ def unicode(self):
 
         for section_name, section_contents in self.report_data:
             line_buf.append(self.PLUGDIVIDER)
-            line_buf.append(f"{section_name}")
+            line_buf.append(self.PLUGINFORMAT % {'name': section_name})
```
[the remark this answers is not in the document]
jcastill: I used f-strings originally but then I used your approach as per https://github.com/sosreport/sos/pull/3781#issuecomment-2367421527

sos/report/reporting.py
jcastill: I didn't see an issue when I checked the result, but I'll add the spaces to keep the same format

jcastill: The space is added when the variable PLUGLISTITEM is defined, via `PLUGLISTITEM = "  %(name)s"`

pmoravec (approved):

arif-ali (approved):

PR #3782, opened 2024-09-20
[tests] Allow tests to restrict OS version

jcastill:
Add a new `versions` class attribute that allows test classes to specify which version(s) are allowed/designed to execute on. If the test is attempted to be executed on a version that's not specified, the test will be skipped and a message will be printed.

By default, the list of versions will be empty, which means that the test should run in all versions of the OS.

This patch contains a change to restrict the execution of Teamd to RHEL 8, because it was deprecated for RHEL 9 and later.

---

### DIFF
--- a/tests/report_tests/plugin_tests/teamd.py
+++ b/tests/report_tests/plugin_tests/teamd.py
@@ -25,6 +25,9 @@ class TeamdPluginTest(StageTwoReportTest):
     sos_cmd = '-o teamd'
     redhat_only = True
 
+    # teaming has been deprecated from RHEL 9
+    only_os_versions = ['8']
+
     def pre_sos_setup(self):
         # restart NetworkManager to account for the new package
         nmout = process.run('systemctl restart NetworkManager', timeout=30)
--- a/tests/sos_tests.py
+++ b/tests/sos_tests.py
@@ -34,6 +34,8 @@ SOS_TEST_BIN = os.path.realpath(os.path.join(SOS_TEST_DIR, '../bin/sos'))
 RH_DIST = ['rhel', 'centos', 'fedora', 'centos-stream']
 UBUNTU_DIST = ['Ubuntu', 'debian']
 
+_distro = distro.detect()
+
 
 def skipIf(cond, message=None):
     # pylint: disable=unused-argument
@@ -51,7 +53,7 @@ def skipIf(cond, message=None):
 # pylint: disable=unused-argument
 def redhat_only(tst):
     def wrapper(func):
-        if distro.detect().name not in RH_DIST:
+        if _distro.name not in RH_DIST:
             raise TestSkipError('Not running on a Red Hat distro')
     return wrapper
 
@@ -59,7 +61,7 @@ def redhat_only(tst):
 # pylint: disable=unused-argument
 def ubuntu_only(tst):
     def wrapper(func):
-        if distro.detect().name not in UBUNTU_DIST:
+        if _distro.name not in UBUNTU_DIST:
             raise TestSkipError('Not running on a Ubuntu or Debian distro')
     return wrapper
 
@@ -82,6 +84,7 @@ class BaseSoSTest(Test):
     ubuntu_only = False
     end_of_test_case = False
     arch = []
+    only_os_versions = []
 
     @property
     def klass_name(self):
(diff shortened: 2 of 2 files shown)


jcastill:
Result of running this PR against a RHEL 9 where teamd is deprecated:

```
 (19/55) tests/report_tests/plugin_tests/teamd.py:TeamdPluginTest.test_teamd_plugin_executed: STARTED
 (19/55) tests/report_tests/plugin_tests/teamd.py:TeamdPluginTest.test_teamd_plugin_executed: SKIP: Unsupported OS version 9 (supports: ['8'])
 (20/55) tests/report_tests/plugin_tests/teamd.py:TeamdPluginTest.test_team_dev_iteration: STARTED
 (20/55) tests/report_tests/plugin_tests/teamd.py:TeamdPluginTest.test_team_dev_iteration: SKIP: Unsupported OS version 9 (supports: ['8'])
 (21/55) tests/report_tests/plugin_tests/teamd.py:TeamdPluginTest.test_archive_created: STARTED
 (21/55) tests/report_tests/plugin_tests/teamd.py:TeamdPluginTest.test_archive_created: SKIP: Unsupported OS version 9 (supports: ['8'])
```

Does this look ok?

tests/sos_tests.py
```diff
@@ -261,6 +262,20 @@ def check_arch_for_enablement(self):
...
 
+    def check_os_version_for_enablement(self):
+        """
+        Check if the test case is meant only for a specific version or versions
+
+        Takes the `versions` class attribute, a list that specifies
+        the versions where the test applies. If the list is empty, assume all
+        versions of the OS are acceptable. Otherwise, raise a TestSkipError.
+        """
+        os_version = distro.detect().version
```
pmoravec, line 273: As the `distro.detect()` is called on four places and executed many times, does it deserve to set `self.distro = distro.detect()` once and refer to the variable?

pmoravec (approved):
ACK with the current content. The `distro.dectect()` is rather a suggestion.

tests/report_tests/plugin_tests/teamd.py
```diff
@@ -25,6 +25,9 @@ class TeamdPluginTest(StageTwoReportTest):
     sos_cmd = '-o teamd'
     redhat_only = True
 
+    # teaming has been deprecated from RHEL 9
+    versions = ['8']
```
pmoravec, line 29: Ah I realize: `teamd` is also on RHEL7 so we shall allow tests of the plugin on version 7 also. But we (sos upstream) dont further support so old Python version that is on RHEL7, so..?

jcastill: It's tricky, isn't it? I was thinking of having a min version and max version and use ranges, and that could have covered that case. So it's not just a case of RHEL7, rather how to deal with possible ranges - what's the most elegant and pythonic way to do. If I say that this test runs in RHEL 8, do we assume that that's the last version where it ran?
And what about Ubuntu? How do we cover their cases?
It may be simpler to be explicit only when something has been deprecated, so for something that will get deprecated in RHEL10, we explicitly write '8', '9'.

tests/sos_tests.py
jcastill: I think so. I'll wait for other's thoughts and then will push a fix including this change (unless you think a new PR is necessary, but I don't think so?)

tests/report_tests/plugin_tests/teamd.py
arif-ali: This was already passing here with centos Stream 9, or is the basis here that this is done in downstream testing?

jcastill: It's mainly to do with RHEL testing, but the way I implemented it should cover any other distro that have such issue

pmoravec: This raises more generic concern that we can imho ignore now: how to deal with requirement "test this on RHEL/CentOS 8 *or* Ubuntu XY" - the `versions` would have to be a distro->versions mapping then.

But since we need this feature for one distro-family only, no need for that (now).

jcastill: Would this work?
	versions = ['8', '9', '22.04']

Or do we rather prefer something like:
	versions = ['rhel8', 'rhel9', 'ubuntu22.04']

pmoravec: OK, my bad example since Ubuntu version cant be confused with RHEL version and most probably there isn't a pair of distros with similar version number that we would need to differentiate for the tests.

I would keep with the current "test version only (of either distro)" since that is sufficient for our current need. If we would require more in future, let open a new PR (that time).

TurboTurtle, line 29: Nit: `os_versions` or `only_os_versions` would be more descriptive here.

tests/sos_tests.py
TurboTurtle: Agreed, set it once, use it everywhere.

tests/report_tests/plugin_tests/teamd.py
jcastill: Yes, it sounds much better this way. I'll fix and push again, thank you!

pmoravec (approved):

Issue #3788, opened 2024-09-25
most plugins doesn't collect ALL_PROXY environment variable


pafernanr:
Multiple OS commands rely on `libcurl` hence in the environment variables `libcurl` understands. Most of the plugins are collecting `http_proxy` and similar env_vars. But `ALL_PROXY` is not being collected by the plugins which leads to miss important environment details.

Currently only `crio`, `docker` and `podman` are collecting this env_var.

It could also be great to include `[scheme]_proxy` into collected vars, but it would require rewrite `def collect_env_vars` to make some kind of `regex` evaluation.

- libcurl supported env_vars: https://curl.se/libcurl/c/libcurl-env.html

In addition to that, let me ask: (Probably this question was already answered, solved, or has security implications)
Is it better to `self.add_env_var(["ALL_PROXY"])` on all required plugins? Or worth to obtain "ALL" current shell environment variables in a dedicated plugin?

pafernanr:
Another approach could be initialize `self.env_vars` with a commonly used env_vars `report/__init__.py` and let plugins define only the specific required ones. 
E.g:
~~~
self.env_vars = set(["ALL_PROXY", "NO_PROXY", "FTP_PROXY", "HTTP_PROXY", "HTTPS_PROXY"])
~~~

Does it makes sense?

TurboTurtle:
Environment variables are handled a bit differently than most collections. The plugins aren't actually doing the collection themselves, but rather updating a global (well, report-level) list of environment variables to capture _after_ all plugins have finished their normal execution.

We can add these `*_PROXY` vars to a default set, though I believe we initially avoided doing so do to concerns around end users (read: customers of vendors) not wanting something like this collected by default. We can mitigate some related concerns around passing credentials in the variable, but the big item was "should this _always_ be collected?". At the time we said no, but that's by no means set in stone.

Similarly, when we initially added this there was the thought of "why don't we just collect every environment variable there is", and were met with a swift "not just no, but hell no" from various security folks.

pafernanr:
Sure, I fully agree  with that from a security point of view.

My concern is because, currently, a connection can be "proxyfied/not proxyfied" depending on env_vars that are not being collected. Some plugins seems to have noted this and include related env_vars, but not all of them. 

To mitigate this, I think ["ALL_PROXY", "NO_PROXY"] could be added on all plugins that are currently including `*_PROXY` on their added `env_vars`.

pmoravec:
[...]

PR #3789, opened 2024-09-25
[plugins] Obfuscate proxy credentials in env. settings

pmoravec:
Obfuscate credentials in env.variable settings for e.g. HTTP_PROXY or similar.

Inspired by: #3788
Resolves: #3789

---


pmoravec:
Trivial reproducer:

    echo HTTP_PROXY=http://foouser:barpassword@proxyurl:8080 >> /etc/environment
    sos report -o system --batch --build

I am not sure if whole `/etc/sysconfig` or `/etc/default` needs such scrubbing - various services can define the proxy there so it is imho better to cover all. BUT if either directory is password-less every time, I am happy to remove it from the `postproc`.

pmoravec:
OK, the `/etc/sysconfig` or `/etc/default` could also collect the env.variables, worth processing them as well.

sos/report/plugins/anaconda.py
```diff
@@ -24,21 +24,21 @@ class Anaconda(Plugin, RedHatPlugin):
...
         else:
-            paths = paths + [
+            self.copypaths = self.copypaths + [
                 "/var/log/anaconda.*",
                 "/root/install.log",
                 "/root/install.log.syslog"
             ]
 
-        self.add_copy_spec(paths)
+        self.add_copy_spec(self.copypaths)
```
pmoravec, line 41: The credentials can be in either anaconda log or in the `*ks.cfg` file - hence apply postproc to all copyspecs.

jcastill:
The pylint failure is solved already but checking out this pr doesn't show the commit, so I think you'll have to resync.

jcastill (approved):

arif-ali (approved):
looks good to me, it would be good to do the same in apt.py as well, but if you prefer, I can do a seperate PR for that later (as that is already obfuscating proxy)

pmoravec:
Good point wrt Apt plugin where I in fact copied the RE from - I just updated the PR accordingly.

TurboTurtle (approved):
This LGTM, however I should note that what this _does_ and what it _says_ it does are slightly different.

The obfuscation isn't actually looking for `*_PROXY` variables, just urls with username and passwords in them, which _so happen_ to be the convention for proxy authentication.

I have a slight preference to rename to `do_path_http_sub()` or something similar, but it is not a big item for me and I'm fine if you want to merge as is.

PR #3794, opened 2024-09-27
[system] List env_var names

pafernanr:
This code gets the environment Names (without values) and saves them to `environment_varnames`.

Inspired by: #3792
Resolves: #3792

Signed-off-by: Pablo Fernández Rodríguez <pafernan@redhat.com>
---

### DIFF
--- a/sos/report/plugins/system.py
+++ b/sos/report/plugins/system.py
@@ -6,6 +6,7 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
+import os
 from sos.report.plugins import Plugin, IndependentPlugin
 
 
@@ -39,6 +40,12 @@ class System(Plugin, IndependentPlugin):
             "ld.so --list-diagnostics",
             "ld.so --list-tunables"
         ])
+
+        var_names = list(os.environ.keys())
+        var_names.sort()
+        self.add_string_as_file('\n'.join(var_names),
+                                "environment_varnames",
+                                plug_dir=True)
 
     def postproc(self):
         self.do_paths_http_sub([


pmoravec (approved):

TurboTurtle:
Should this instead go into the top-level `environment` file? We'd have to pull it out of plugin code (easy enough to add it to the report-level env var collection method), but it feels odd having a schism between "the `environment` file has env vars in it, and `sos_commands/system/environment_varnames` has just the names, including those already in `environment`".

pafernanr:
That makes much sense, It can be very confusing to put the environment_vars in two different places. The logic can be easily moved to `collect_env_vars` and could be something similar to:
~~~
    def collect_env_vars(self):
        env = ""
        var_names = list(os.environ.keys())
        var_names.sort()
        for name in var_names:
            if self.env_vars and name in self.env_vars:
                env += f"{name}={os.environ.get(name)}\n"
            else:   
                env += f"{name}=ValueNotCollected\n"
        self.archive.add_string(env, 'environment')
~~~

- I think the point is: How to format the output and make it comprehensible; IMO it should let us know which values are "collected" or "not collected", and also show if collected values are "None". Example output:
~~~
BASH_FUNC_which%%=ValueNotCollected
DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/0/bus
HISTCONTROL=ValueNotCollected
HISTSIZE=ValueNotCollected
...
HTTP_PROXY=
CUSTOM_EMPTY_VAR=ValueNotCollected
...
~~~
** Note that HTTP_PROXY is a "collected empty var" while CUSTOM_EMPTY_VAR is a "uncollected empty var".

arif-ali (approved):

pmoravec (approved):

Issue #3796, opened 2024-09-30
Feature Request: Include container image manifests and layers in sosreport


tawfiqfm:
Currently, sosreport does not capture detailed information about container images running on a node. This can make it difficult to troubleshoot issues with containerized applications. Requesting feature to dump all manifests and layers for images on a Node in sosreport.

TurboTurtle:
Sounds reasonable - what environments/runtimes are you looking for?

`podman` I think this would be provided via `podman image inspect $image`, is that what you're looking for, or something else? `docker` would be the same I think? Would need to look into k8s or OCP if you're looking for those plugins as well or instead.

SergeyKanzhelev:
We are interested in containerd. But other runtimes may be interested in it as well. Some hierarchy of layers and all the manifests would be great to have in the report.

cc: @samuelkarp to advice on the best way to collect this for containerd

tawfiqfm:
Seems like this was closed by mistake. This hasn't been implemented by Containerd.

TurboTurtle:
Yep, mistakenly closed. I had updated the tag to `Related`, so not sure why this got closed out.

lutierigb:
I don't see an easy way to retrieve the manifest for an image with crictl(containerd). @samuelkarp any ideas?

Sergey and I agree that would be beneficial to be able to see all the layers for the images present on the node.

PR #3797, opened 2024-10-02
[docker,podman] Collect container image layers

TurboTurtle:
Adds collection of container image layers for the docker and podman plugins via the `history` and `tree` subcommands repsectively.

Resolves: #3796

---

### DIFF
--- a/sos/report/plugins/docker.py
+++ b/sos/report/plugins/docker.py
@@ -97,6 +97,11 @@ class Docker(Plugin, CosPlugin):
             insp = name if 'none' not in name else img_id
             self.add_cmd_output(f"docker inspect {insp}", subdir='images',
                                 tags="docker_image_inspect")
+            self.add_cmd_output(
+                f"docker image history {insp}",
+                subdir='images/history',
+                tags='docker_image_tree'
+            )
 
         for vol in volumes:
             self.add_cmd_output(f"docker volume inspect {vol}",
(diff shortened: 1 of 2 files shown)


jcastill (approved):
Tested with podman in Fedora and RHEL, works perfectly.

pmoravec (approved):

samuelkarp:
We don't need to hold this open (we can do it in another PR), but we're interested in the image layer information from containerd as well.

SergeyKanzhelev:
Maybe not resolve the ticket to wait for Containerd implementation? This PR says "Resolves: https://github.com/sosreport/sos/issues/3796" which will close the ticket once PR is merged

TurboTurtle:
Dropped the `resolves`, was figuring we could get feedback for containerd during review but if we need to push that collection and these separately that's fine too.

SergeyKanzhelev:
@samuelkarp does `ctr` supports something like this? What is the recommended way to collect this on Containerd?

samuelkarp:
We don't really have a high-level interface for this in `ctr`.  Probably closest would be something like:

* `ctr images list` - grab all refs (first column) and digests (second column)
* `ctr content get <digest>` - for each digest
* For any digest that was then either an index or manifest list, grab the subsequent manifest digests from there and `ctr content get <digest>` again for the actual manifest
* If we _also_ want the config, then `ctr content get <digest>` on the config digest.  But configs can have private information in them (environment variables and command line, mostly).

PR #3798, opened 2024-10-03
[tests] Scrub tests

jjansky1:
Adding stageone and stagetwo tests for [system] plugin.

Also adding tag scrub for only testing scrub of sensitive data.

And updating README with how to call scrub and stagetwo tests.

Related: #3788
Related: #3789
Resolves: #3798

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,8 +60,17 @@ against the `flake8` linter, the unit tests, and the stage one avocado test suit
 $ flake8 sos
 $ nosetests -v tests/unittests/
 
-# as root
-# PYTHONPATH=tests/ avocado run --test-runner=runner -t stageone tests/{cleaner,collect,report,vendor}_tests
+# sudo PYTHONPATH=tests/ avocado run --test-runner=runner -t stageone tests/{cleaner,collect,report,vendor}_tests
+```
+
+For further test run stagetwo tests
+```
+# sudo PYTHONPATH=tests/ avocado run --test-runner=runner -t stagetwo tests/{cleaner,collect,report,vendor}_tests
+```
+
+If you want to check basic scrub tests
+```
+# sudo PYTHONPATH=tests/ avocado run --test-runner=runner -t scrub tests/{cleaner,collect,report,vendor}_tests
 ```
 
 Note that the avocado test suite will generate and remove several reports over its
--- a/tests/report_tests/plugin_tests/apt/apt.py
+++ b/tests/report_tests/plugin_tests/apt/apt.py
@@ -15,7 +15,7 @@ class AptConfScrubTest(StageTwoReportTest):
     """Ensure that sources.list and apt conf are picked up and properly
     scrubbed
 
-    :avocado: tags=stagetwo
+    :avocado: tags=stagetwo,scrub
     """
 
     sos_cmd = '-o apt'
--- a/tests/report_tests/plugin_tests/curtin/curtin.py
+++ b/tests/report_tests/plugin_tests/curtin/curtin.py
@@ -12,7 +12,7 @@ from sos_tests import StageTwoReportTest
 class CurtinScrubbedTest(StageTwoReportTest):
     """Ensure that curtin conf is picked up and properly scrubbed
 
-    :avocado: tags=stagetwo
+    :avocado: tags=stagetwo,scrub
     """
 
     sos_cmd = '-o curtin'
--- a/tests/report_tests/plugin_tests/juju/juju.py
+++ b/tests/report_tests/plugin_tests/juju/juju.py
@@ -12,7 +12,7 @@ from sos_tests import StageTwoReportTest
 class JujuAgentConfScrubbedTest(StageTwoReportTest):
     """Ensure that agent conf is picked up and properly scrubbed
 
-    :avocado: tags=stagetwo
+    :avocado: tags=stagetwo,scrub
     """
 
     sos_cmd = '-o juju'
--- a/tests/report_tests/plugin_tests/networking/networking.py
+++ b/tests/report_tests/plugin_tests/networking/networking.py
@@ -46,7 +46,7 @@ class NetplanScrubTest(StageTwoReportTest):
     ensure that netplan configuration is collected and then the wifi password
     is scrubbed correctly
 
-    :avocado: tags=stagetwo
+    :avocado: tags=stagetwo,scrub
     """
 
     sos_cmd = '-o networking'
--- a/tests/report_tests/plugin_tests/sudo/sudo.py
+++ b/tests/report_tests/plugin_tests/sudo/sudo.py
@@ -10,7 +10,7 @@ from sos_tests import StageOneReportTest, StageTwoReportTest
 
 
 class SudoPluginTest(StageOneReportTest):
-    """Basic sanity check to make sure ocmmon config files are collected
+    """Basic sanity check to make sure common config files are collected
 
     :avocado: tags=stageone
     """
@@ -25,7 +25,7 @@ class SudoPluginTest(StageOneReportTest):
 class SudoLdapScrubbedTest(StageTwoReportTest):
     """Ensure that sudo conf is picked up and properly scrubbed
 
-    :avocado: tags=stagetwo
+    :avocado: tags=stagetwo,scrub
     """
 
     sos_cmd = '-o sudo'
--- a/tests/report_tests/plugin_tests/system/system.py
+++ b/tests/report_tests/plugin_tests/system/system.py
@@ -1,0 +1,90 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos_tests import StageOneReportTest, StageTwoReportTest, redhat_only
+
+
+class SystemPluginTest(StageOneReportTest):
+    """Basic sanity check to make sure common config files are collected
+
+    :avocado: tags=stageone
+    """
+
+    sos_cmd = '-o system'
+
+    def test_system_files_collected(self):
+        self.assertFileCollected("/proc/sys")
+        self.assertFileCollected("/etc/default")
+        self.assertFileCollected("/etc/environment")
+
+    @redhat_only
+    def test_system_files_collected_RH(self):
+        self.assertFileGlobInArchive("/etc/sysconfig")
+
+    def test_system_files_forbidden(self):
+        self.assertFileGlobNotInArchive("/proc/sys/net/ipv4/route/flush")
+        self.assertFileGlobNotInArchive("/proc/sys/net/ipv6/route/flush")
+        self.assertFileGlobNotInArchive("/proc/sys/net/ipv6/neigh/" +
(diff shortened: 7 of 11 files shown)


tests/test_data/system_test_data
```diff
@@ -0,0 +1,4 @@
+HTTP_PROXY=http://foouser:badpassword@proxyurl:8080
+HTTPS_PROXY=https://foouser:badpassword@proxyurl:8080
+http_proxy=http://foouser:badpassword@proxyurl:8080
+https_proxy=https://foouser:badpassword@proxyurl:8080
```
pmoravec, line 4: You test for `somesecretpassword` presence while having `badpassword` here..?

jcastill (approved):
From my limited knowledge of avocado tests, LGTM, just need to address Pavel's note about the password string

tests/report_tests/plugin_tests/systemd.py
```diff
@@ -0,0 +1,92 @@
...
+    """
+
+    sos_cmd = '-o systemd'
+
+    def test_systemd_files_collected(self):
+        self.assertFileCollected("/etc/systemd")
+        self.assertFileCollected("/lib/systemd/system")
+        self.assertFileCollected("/lib/systemd/user")
+        self.assertFileCollected("/etc/vconsole.conf")
+        self.assertFileCollected("/run/systemd/generator*")
```
pmoravec, line 25: Hm.. sadly `assertFileCollected` does not work for globs, it just executes:

        if fname.startswith(('sos_', '/sos_')) or os.path.exists(fname):
            self.assertFileExists(self.get_name_in_archive(fname))
        else:
            assert True

And while having e.g. `/run/systemd/generator` directory on my system, `os.path.exists("/run/systemd/generator*")` returns False - so no test is executed if either `/run/systemd/generator*` stuff is collected.

Worth expanding the method to cover globs?

tests/report_tests/plugin_tests/system.py
```diff
@@ -0,0 +1,70 @@
...
+class SystemScrubTest(StageTwoReportTest):
+    """Ensure that environment, default and sysconfig are picked up
+    and properly scrubbed
+
+    :avocado: tags=stagetwo,scrub
+    """
+    sos_cmd = '-o system'
+    files = [
+        ('../../../tests/test_data/system_test_data', '/etc/environment'),
+        ('../../../tests/test_data/system_test_data', '/etc/default/proxy'),
```
pmoravec, line 50: When I run `PYTHONPATH=tests/ avocado run -p TESTLOCAL=true --test-runner=runner -t stagetwo tests/report_tests` on my test rhel8 system, I get:

     (17/57) tests/report_tests/plugin_tests/system.py:SystemScrubTest.test_archive_created: ERROR: [Errno 2] No such file or directory: '/etc/default/proxy' (0.00 s)

Did I run the avocado test in a proper way..? (I am often uncertain on this..) FYI my system has no `/etc/default/proxy` file/dir.

jcastill: I cannot find that file/dir either in Fedora or RHEL 9. What package creates it?

tests/report_tests/plugin_tests/systemd.py
jcastill: +1 to expand the method to cover globs

tests/report_tests/plugin_tests/systemd.py
```diff
@@ -0,0 +1,91 @@
...
+    :avocado: tags=stageone
+    """
+
+    sos_cmd = '-o systemd'
+
+    def test_systemd_files_collected(self):
+        self.assertFileGlobInArchive("/etc/systemd")
+        self.assertFileGlobInArchive("/lib/systemd/system")
+        self.assertFileGlobInArchive("/lib/systemd/user")
+        self.assertFileGlobInArchive("/etc/vconsole.conf")
```
pmoravec, line 24: The `Glob` test is redundantly generic for a direct filename without a glob, but let it be..

pmoravec (approved):

tests/report_tests/plugin_tests/system.py
```diff
@@ -0,0 +1,78 @@
...
+        ('../../../tests/test_data/system_test_data', '/etc/default/proxy1'),
```
jcastill, line 54: Ubuntu tests are failing here:
```
 (070/116) tests/report_tests/plugin_tests/system.py:SystemScrubTest.test_archive_created: STARTED
 (070/116) tests/report_tests/plugin_tests/system.py:SystemScrubTest.test_archive_created:  ERROR: [Errno 2] No such file or directory: '/etc/sysconfig/proxy1' (0.02 s)
 (071/116) tests/report_tests/plugin_tests/systemd.py:SystemdScrubTest.test_systemd_files_collected: STARTED
```

Do we really need the 1 at the end of this file and /etc/sysconfig/proxy1 or is that a typo?

jjansky1: Issue is not with naming like proxy or proxy1, the scrub in code should obfuscate any file in /etc/sysconfig. 

Issue seems to be that ubuntu do not have /etc/sysconfig at all, due to that i have made separated tests for all + specific /etc/sysconfig for redhat + centos distribution. As you can see in last pipeline results of all stagetwo tests passed.

tests/report_tests/plugin_tests/system.py
```diff
@@ -0,0 +1,98 @@
...
+    and properly scrubbed
+
+    :avocado: tags=stagetwo,scrub
+    """
+    redhat_only = True
+    sos_cmd = '-o system'
+    files = [
+        ('../../../tests/test_data/system_test_data', '/etc/sysconfig/proxy'),
+        ('../../../tests/test_data/system_test_data', '/etc/sysconfig/proxy1'),
+    ]
```
TurboTurtle, line 83: The mocked file functionality assumes the files are co-located with the test pyfile because the general design is that each test (which has associated mocked data with it) will be in its own subdir.

Rather than maintaining relative paths like this, place the test file in a subdir under tests/report_tests/plugin_tests and put the test case and the mocked data there.

tests/report_tests/plugin_tests/systemd.py
```diff
@@ -0,0 +1,91 @@
...
+    and properly scrubbed
+
+    :avocado: tags=stagetwo,scrub
+    """
+    sos_cmd = '-o systemd'
+    files = [
+        ('../../../tests/test_data/system_test_data', '/etc/systemd/system'),
+        ('../../../tests/test_data/system_test_data', '/lib/systemd/system'),
+        ('../../../tests/test_data/system_test_data', '/run/systemd/system'),
+    ]
```
TurboTurtle, line 76: Same here.

tests/report_tests/plugin_tests/systemd.py
```diff
@@ -0,0 +1,91 @@
...
+    secrets_list = [
+        'foouser',
+        'somesecretpassword'
+    ]
+
+    def test_systemd_files_collected(self):
+        for file in self.files:
+            self.assertFileGlobInArchive(file[1])
```
TurboTurtle, line 84: I'm not sure why we're testing the collection of files we're mocking?

tests/report_tests/plugin_tests/systemd.py
```diff
@@ -0,0 +1,91 @@
...
+        self.assertFileGlobInArchive("systemctl_show-environment")
+        self.assertFileGlobInArchive("systemctl_show_service_--all")
+        self.assertFileGlobInArchive("systemctl_status_--all")
+        self.assertFileGlobInArchive("systemd-analyze")
+        self.assertFileGlobInArchive("systemd-analyze_blame")
+        self.assertFileGlobInArchive("systemd-analyze_dump")
+        self.assertFileGlobInArchive("systemd-analyze_plot.svg")
+        self.assertFileGlobInArchive("systemd-delta")
+        self.assertFileGlobInArchive("systemd-inhibit_--list")
+        self.assertFileGlobInArchive("timedatectl")
```
TurboTurtle, line 62: Should be using `assertFileCollected` for specific file names.

tests/report_tests/plugin_tests/system.py
```diff
@@ -0,0 +1,98 @@
...
+
+    :avocado: tags=stageone
+    """
+
+    sos_cmd = '-o system'
+
+    def test_system_files_collected(self):
+        self.assertFileGlobInArchive("/proc/sys")
+        self.assertFileGlobInArchive("/etc/default")
+        self.assertFileGlobInArchive("/etc/environment")
```
TurboTurtle, line 23: Same. Use `assertFileCollected` for specific paths, only use `assertFileGlobInArchive` if there are actual globs being checked.

tests/report_tests/plugin_tests/system.py
```diff
@@ -0,0 +1,98 @@
...
+        ('../../../tests/test_data/system_test_data', '/etc/default/proxy1'),
+    ]
+    secrets_list = [
+        'foouser',
+        'somesecretpassword'
+    ]
+
+    def test_system_files_collected(self):
+        for file in self.files:
+            self.assertFileGlobInArchive(file[1])
```
TurboTurtle, line 64: Again, not sure why we're testing for a collection that we're mocking to begin with.

TurboTurtle (commented):
Sorry this one sat for so long without another review. Notes below.

arif-ali:
Just thinking out loud here, we're not using the tag scrub anywhere. Is there a need for a new tag when we're already doing stageone and stagetwo tests and these are being encompassed by them?

pmoravec:
> Just thinking out loud here, we're not using the tag scrub anywhere. Is there a need for a new tag when we're already doing stageone and stagetwo tests and these are being encompassed by them?

Good point - why or where is this tag needed, in what use cases?

jjansky1:
You kidding right? cannot you next time make sense of PR and not watch code 9 time rewrite and now thinking about propose? done with upstream

jjansky1:
This PR should be about future, for example not all stagetwo do not need to be scrub and not all of scrub need to be stagetwo, they are now just case we make same class ad only automatic tests, honestly after 9 rewrite, month and now you askind about base or the PR.... decline it, i have not such kind of patience

arif-ali:
Hi Jan, With due respect, I've had similar PRs and others would have done so and have taken multiple re-writes and then merged (especially 1 which took me 4 months after several reviews), I can understand the frustration in the time taken; I feel it sometimes too, but we just need crack on and work together for the same goals.

Sometimes we all have to be patient with this, it's an open-source project, not everyone here dedicates their 100% time on this, and things can easily missed in the first pass, maybe on annual leave or other priorities..

 I hope that makes sense, and hope you understand

> This PR should be about future, for example not all stagetwo do not need to be scrub and not all of scrub need to be stagetwo, they are now just case we make same class and only automatic tests, honestly after 9 rewrite, month and now you asking about base or the PR.... decline it, i have not such kind of patience

I appreciate that, we're not using it in the CI here, and hence my question.

However, if there is an intention to do something internal within RH then I am 100% fully on-board with that, or just do scub tests here as a separate test (Although, it seems we are adding an extra tag on top).

I am fully on-board on adding more tests, the better testing we have, the better the product.

Thanks for your understanding

jcastill:
@arif-ali thank you so much for your input, as always. We've moved the conversation with Jan internally, and we'll close this PR as per his request.

Issue #3799, opened 2024-10-04
Feature request: collect cgroup information per process


samuelkarp:
`/proc/<pid>/cgroup` should expose the association between processes and cgroups.  It does not appear like this is [currently captured](https://github.com/sosreport/sos/blob/96e4d79231819b45ef62629206a7a3da31b3a1dd/sos/report/plugins/process.py#L52-L58).  The same information can be found through the cgroup hierarchy (`cgroup.procs` file for cgroup v2 and `tasks` file for v1), but that requires traversing the entire hierarchy to find.

jcastill:
Nice catch. We have a cgroups plugin where we can collect this, so it may be better to place it there. 
I'll send a quick PR so we can discuss some possible implementations with the maintainers.

TurboTurtle:
I believe this was satisfied by #3801. Feel free to reopen if not fully addressed.

PR #3801, opened 2024-10-05
[cgroups] Collect cgroup information per process

jcastill:
Collect cgroup information for each process

Related: #3799

---

### DIFF
--- a/sos/report/plugins/cgroups.py
+++ b/sos/report/plugins/cgroups.py
@@ -26,7 +26,8 @@ class Cgroups(Plugin, DebianPlugin, UbuntuPlugin, CosPlugin):
 
         self.add_copy_spec([
             "/proc/cgroups",
-            "/sys/fs/cgroup"
+            "/sys/fs/cgroup",
+            "/proc/[0-9]*/cgroup",
         ])
 
         self.add_cmd_output("systemd-cgls")


arif-ali (approved):

sos/report/plugins/cgroups.py
```diff
@@ -26,7 +26,8 @@ def setup(self):
 
         self.add_copy_spec([
             "/proc/cgroups",
-            "/sys/fs/cgroup"
+            "/sys/fs/cgroup",
+            "/proc/*/cgroup",
```
jcastill, line 30: I've tested this and it works, but this will go through all subdirs of /proc, and only collects 'cgroup' when it finds it. 
We do something similar but more elegantly in the 'process' plugin:

https://github.com/sosreport/sos/blob/96e4d79231819b45ef62629206a7a3da31b3a1dd/sos/report/plugins/process.py#L52-L58

So my question is, is my approach good enough, or should I use the one in the process plugin gathering the PID entries with:

`procs = [p for p in self.listdir("/proc") if re.match("[0-9]", p)]` 

Then iterating through them?

TurboTurtle: `"/proc/[0-9]*/cgroup`

pmoravec (approved):

samuelkarp (commented):
Thank you!

TurboTurtle (approved):

PR #3803, opened 2024-10-10
[juju] Add plugin option for Juju state reporting

MichaelThamm:
This change modifies the Juju plugin to optionally capture Juju state (controllers, models, applications, units) information. The capture assumes that `juju` is installed on the machine where `sos` is called, and that the juju user has superuser privilege to the current (or requested) controllers.

This option is disabled by default because, while the information can be very useful, the collection acts on the live Juju state.

Use the plugin with the feature:
* `sos report --only-plugin juju -k juju.juju-state=True`

Run with a specific user (who has access to the controllers specified). This defaults to `ubuntu`:
* `-k juju.juju-user=super_admin`

If necessary, you can filter by controllers or models with cluster options.
* `-k juju.controllers="controller_a controller_b"`
* `-k juju.models="controller_a:model_x controller_b:model_y"`

If nothing is supplied for either the controllers or models options, the report will include all state information for them respectively.

---

- [ ] Add tests to ensure results are scrubbed / functioning


sos/report/plugins/juju.py
```diff
@@ -53,6 +65,52 @@ def setup(self):
...
+
+        juju_user = self.get_option("juju-user")
+        try:
+            pwd.getpwnam(juju_user)
+        except KeyError:
+            # The user doesn't exist, this will skip the rest
+            self._log_warn(f'User "{juju_user}" does not exist, will not collect Juju information.')
+            return
+
+        user_context = f'sudo -u {juju_user}'
```
TurboTurtle, line 80: Use the `runas` option for `add_cmd_output()`, `collect_cmd_output()`, and `exec_cmd()` to run as non-root. We don't embed `sudo` into command collections.

sos/report/plugins/juju.py
```diff
@@ -53,6 +65,52 @@ def setup(self):
...
+            if desired_controllers and desired_controllers != {''}:
+                controllers = desired_controllers
+            else:
+                controllers = set(json.loads(controllers_json["output"])["controllers"].keys())
+        else:
+            controllers =  {}
+
+        if self.get_option('controllers') and self.get_option('models'):
+            self._log_warn('Options: controllers, models are mutually exclusive.')
+            return
```
arif-ali, line 95: is it worth moving this test straight after the get_option("juju-user"), this would mean we return any of these to items before actually checking for `juju controllers`?

arif-ali (changes requested):
As well as turboturtle comments, can we also look at the flake/pylint issues as well is any updated PR

sos/report/plugins/juju.py
```diff
@@ -121,10 +165,11 @@ def postproc(self):
...
         # Redact simple yaml style "key: value".
-        keys_regex = fr"(^\s*({'|'.join(protect_keys)})\s*:\s*)(.*)"
+        keys_regex = rf"(^\s*({'|'.join(protect_keys)})\s*:\s*)(.*)"
         sub_regex = r"\1*********"
         self.do_path_regex_sub(agents_path, keys_regex, sub_regex)
         # Redact certificates
         self.do_file_private_sub(agents_path)
 
+
 # vim: set et ts=4 sw=4 :
```
jcastill, line 175: I think you need to add a new line after the vim modeline to avoid a linting error.

jcastill (changes requested):

jcastill:
A couple of very small comments from me. 
The DCO is still failing, I think you are missing '-s' while running 'git commit'.
Last thing, please make sure you squash the two commits you have at the moment into one.

jcastill (approved):

jcastill:
@MichaelThamm remember to squash all the commits into one and sign it

arif-ali:
@MichaelThamm The IP addresses and endpoints, not so, as they would be typically handled by `--clean` or `sos clean`, if indeed these needed to be obfuscated.

In terms of the certs, you can get the example of the sunbeam.py plugin, where we just needs `self.do_cmd_private_sub('juju controllers')` in the `postproc()` function. This will automatically obfuscate any command that showed the cert.

sos/report/plugins/juju.py
```diff
@@ -33,26 +62,99 @@ def setup(self):
         self.add_copy_spec("/var/lib/juju/agents/*/agent.conf")
 
         # Get a directory listing of /var/log/juju and /var/lib/juju
-        self.add_dir_listing([
-            '/var/log/juju*',
-            '/var/lib/juju*'
-        ], recursive=True)
+        self.add_dir_listing(
+            ["/var/log/juju*", "/var/lib/juju*"], recursive=True
+        )
```
arif-ali, line 67: I don't think we need to change this here, and leave this as is. With our project, we prefer the format with `([` being together as it was before

sos/report/plugins/juju.py
```diff
@@ -63,10 +165,11 @@ def postproc(self):
         ]
 
         # Redact simple yaml style "key: value".
-        keys_regex = fr"(^\s*({'|'.join(protect_keys)})\s*:\s*)(.*)"
+        keys_regex = rf"(^\s*({'|'.join(protect_keys)})\s*:\s*)(.*)"
```
arif-ali, line 168: I don't think we need this change here

sos/report/plugins/juju.py
```diff
@@ -63,10 +165,11 @@ def postproc(self):
...
         sub_regex = r"\1*********"
         self.do_path_regex_sub(agents_path, keys_regex, sub_regex)
         # Redact certificates
         self.do_file_private_sub(agents_path)
 
+
```
arif-ali, line 174: nit: i don't think this extra line is required

sos/report/plugins/juju.py
```diff
@@ -33,26 +62,99 @@ def setup(self):
...
-                "/var/lib/juju",
-                "/var/lib/juju/**/.*",
-            ])
+            self.add_copy_spec(
+                [
+                    "/var/log/juju",
+                    "/var/lib/juju",
+                    "/var/lib/juju/**/.*",
+                ]
+            )
```
arif-ali, line 79: same as above

sos/report/plugins/juju.py
```diff
@@ -8,19 +8,48 @@
...
 
 class Juju(Plugin, UbuntuPlugin):
 
-    short_desc = 'Juju orchestration tool'
+    short_desc = "Juju orchestration tool"
 
-    plugin_name = 'juju'
-    profiles = ('virt', 'sysmgmt')
+    plugin_name = "juju"
+    profiles = ("virt", "sysmgmt",)
```
arif-ali, line 21: I don't think there is a need to change the quotes from single to double 🤷🏽

MichaelThamm:
@arif-ali 
> In terms of the certs, you can get the example of the sunbeam.py plugin, where we just needs `self.do_cmd_private_sub('juju controllers')` in the `postproc()` function. This will automatically obfuscate any command that showed the cert.

When I run: `self.do_cmd_private_sub('juju controllers --format=json')` I see this in the logs:

```
2024-10-15 10:13:54,360 DEBUG: [plugin:juju] substituting '-----SCRUBBED' for '----(?:-| )BEGIN.*?----(?:-| )END' in commands matching '*juju controllers --format=json*'
```

However, I do not see results in the `commands` dir of the sos report tar.xz file

This method does not have a filename parameter either. Not sure how to continue with this?

arif-ali:
> When I run: `self.do_cmd_private_sub('juju controllers --format=json')` I see this in the logs:
> 
> ```
> 2024-10-15 10:13:54,360 DEBUG: [plugin:juju] substituting '-----SCRUBBED' for '----(?:-| )BEGIN.*?----(?:-| )END' in commands matching '*juju controllers --format=json*'
> ```
> 
> However, I do not see results in the `commands` dir of the sos report tar.xz file
> 
> This method does not have a filename parameter either. Not sure how to continue with this?

You can just specify `juju controllers` in the function, and it will do a match. In the `sunbeam.py`, I did the same with doing a `self.collect_cmd_output("juju controllers --format json")`, and then used the `juju controllers` in the `self.do_cmd_private_sub()`

MichaelThamm:
I think the PR is almost ready for review, but I noticed we have [tests for the Juju plugin](https://github.com/sosreport/sos/blob/main/tests/report_tests/plugin_tests/juju/juju.py). Is there any need to update this since the test for scrubbing seems to be sufficient? I also tested the scrubbing for `juju controllers` manually and it passes.

sos/report/plugins/juju.py
```diff
@@ -53,6 +82,71 @@ def setup(self):
...
+                "Will not collect Juju information."
+            )
+            return
+
+        controllers_json = self.collect_cmd_output(
+            "juju controllers --format=json", runas=juju_user
+        )
+        if controllers_json["status"] == 0:
+            desired_controllers = set(
+                self.get_option("controllers").split(" ")
```
TurboTurtle, line 111: We generally have plugins use `:` as a delimiter in plugin options that can take a list (as the use of commas gets murky with argparse and how plugin options are....parsed.

sos/report/plugins/juju.py
```diff
@@ -53,6 +82,71 @@ def setup(self):
...
+                    json.loads(controllers_json["output"])[
+                        "controllers"
+                    ].keys()
+                )
+        else:
+            controllers = {}
+
+        # Specific models
+        if self.get_option("models"):
+            for model in self.get_option("models").split(" "):
```
TurboTurtle, line 128: Ditto, we prefer plugin options to use `:` as a delimiter. We may need to add this to the plugin guide, I'll check this shortly.

sos/report/plugins/juju.py
```diff
@@ -53,6 +82,71 @@ def setup(self):
...
+                command = f"juju status -m {model} --format=json"
+                self.collect_cmd_output(command, runas=juju_user)
```
TurboTurtle, line 130: `collect_cmd_output()` is designed to return the output for further processing, if you don't need to do anything with the output, use `add_cmd_output()` instead.

sos/report/plugins/juju.py
```diff
@@ -53,6 +82,71 @@ def setup(self):
...
+                )
+                if models_json["status"] == 0:
+                    models = json.loads(models_json["output"])["models"]
+                    for model in models:
+                        short_name = model["short-name"]
+                        command = (
+                            f"juju status -m {controller}:{short_name} "
+                            f"--format=json"
+                        )
+                        self.collect_cmd_output(command, runas=juju_user)
```
TurboTurtle, line 148: Same here, use `add_cmd_output()` if you don't need to do anything further with the output.

sos/report/plugins/juju.py
```diff
@@ -53,6 +82,71 @@ def setup(self):
...
+                "will not collect Juju information."
+            )
+            return
+
+        if self.get_option("controllers") and self.get_option("models"):
+            self._log_warn(
+                "Options: controllers, models are mutually exclusive. "
+                "Will not collect Juju information."
+            )
+            return
```
TurboTurtle, line 104: I'm assuming these are mutually exclusive within the Juju stack, but would we not anticipate some (admittedly niche) use case where an sos end user isn't sure if a particular host is using one or the other, and wants to pass e.g. `-k juju.controllers=foo,juju.models=bar` to get either without needing to know the specific host configuration? I'm mainly thinking of some sort of automation-driven sos collection here, but I could be way off with that idea.

sos/report/plugins/juju.py
MichaelThamm: The only problem with this is for the Juju model synatx it can be:

- The current controller's model -> `model_a`
- Or it can be a specific controller and model -> `controller_x:model_a`

For this reason I chose to split via empty space since models and controllers are not allowed to have spaces in their names according to Juju. Would this be a reasonable exception to the Sos plugin standard?

TurboTurtle: If it's an expectation that users would pass something like `-k juju.controllers="foo:bar"`, then yeah it's totally reasonable to use something else. We should just make note of that, ideally in both the plugin option description (for `sos report -l` output) and the docstring for the plugin (as the docstring gets used in `sos help` output, e.g. `sos help report.plugins.juju` - see the `kernel` plugin docstring for a good example).

sos/report/plugins/juju.py
MichaelThamm: According to the the Juju CLI:

```
-m, --model (= "")
    Model to operate in. Accepts [<controller name>:]<model name>|<model UUID>
```

There is no `controller` option and so I think it is reasonable to be opinionated and keep the 2 options mutually exclusive. As an added bonus, this simplifies the logic in the plugin.

_Note: I tested this with model UUID and it works fine._

sos/report/plugins/juju.py
MichaelThamm: I guess there is no reason why we couldn't use a colon delimiter for the controllers option. However, it is required for the models option (as I mentioned above) so I think it makes sense to be consistent within the scope of this plugin to keep the white-space delimiter rather than colon. What do you think?

MichaelThamm: I added a docstring and the PluginOpts have `description` args. Is this enough to show up in the `sos report -l` output? I tested this locally and I could not get my Juju plugin options to show up in this output and I have no idea why.

[Reference](https://github.com/sosreport/sos/pull/3803/files#diff-3b9a49f615b1355c9acdf50238eb773562b9f1903d0c9ecbdbe5ae1eb3b6100aR70)

jcastill: This is what I get when I run your changes:

```
# bin/sos report -l | grep juju
 juju                 Juju orchestration tool
 juju.juju-state           off             Include Juju state in the report
 juju.juju-user            ubuntu          Juju client user.
 juju.controllers                          Collect Juju state for specified controllers. Uses a             whitespace delimited list.
 juju.models                               Collect Juju state for specified models. Uses a whitespace             delimited list.

```

And this is what I get when running sos help:

```
Plugin Options
These options may be toggled or changed using '-k juju.option_name=$value'

    Option Name         Default                       Description         
    juju-state          False/Off                     Include Juju state in the report
    juju-user           ubuntu                        Juju client user.
    controllers         None/Unset                    Collect Juju state for specified controllers. Uses a
                                                      whitespace delimited list.
    models              None/Unset                    Collect Juju state for specified models. Uses a whitespace
                                                      delimited list.

```

jcastill: Regarding why the options are not showing up in the output locally, perhaps it's because the machine where you are running doesn't have the /var/log/juju file present? Since that's the activation condition for this plugin:

`files = ('/var/log/juju',)`

As far as I understand it, if it's not present then the plugin should show as inactive and the options won't be printed.

jcastill:
@MichaelThamm remember to squash the three commits into one, and to sign the commit (that is what's causing the DCO failure).

arif-ali (approved):

arif-ali:
You can ignore the `pylint` errors, seems like new tests have come through for `too-many-positional-arguments`

MichaelThamm:
Closing in favour of:
- https://github.com/sosreport/sos/pull/3844

PR #3804, opened 2024-10-11
[instructlab] Add new plugin for Instructlab

jcastill:
This change adds a new plugin that captures information from an installation of Instructlab. The capture works in two ways:
- It goes through all containers whose names contain Instructlab, and
- If specified, it will gather information from a user home directory.

In the first case, we gather different outputs from the command 'ilab', as well as directories containing training data, configuration files, chatlogs, taxonomies, and other data.
In the second case, we capture only certain directories. Users can also specify to gather .cache directory, where models and OCI directories can be found. This option is disabled by default because, while the information can be very useful, the directories are very big.

Related: RH: PLMCORE-10599, RHEL-54137, RHEL-58173

---

### DIFF
--- a/sos/report/plugins/instructlab.py
+++ b/sos/report/plugins/instructlab.py
@@ -1,0 +1,158 @@
+# Copyright (C) 2024 Red Hat, Inc., Jose Castillo <jcastillo@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+import pwd
+from sos.report.plugins import Plugin, IndependentPlugin, PluginOpt
+
+
+class Instructlab(Plugin, IndependentPlugin):
+    """
+    This plugin is used to capture information about
+    Instructlab installations.
+    InstructLab is an open source project for enhancing
+    large language models (LLMs) used in generative
+    artificial intelligence (gen AI) applications.
+    Instructlab can run either as a container, or directly
+    outside a container.
+    """
+
+    short_desc = 'Instructlab'
+    plugin_name = 'instructlab'
+    profiles = ('ai',)
+    containers = ('instructlab', 'ilab',)
+    commands = ('ilab',)
+
+    option_list = [
+        PluginOpt('ilab-user', default='cloud-user', val_type=str,
+                  desc='user that runs instructlab'),
+        PluginOpt('ilab-conf-dir', default='', val_type=str,
+                  desc='instructlab data directory'),
+        PluginOpt('get-cache', default=False,
+                  desc='Capture models and osci cached data')
+    ]
+
+    def setup(self):
+        cont_share_conf_path = "/usr/share/instructlab/config"
+        cont_opt_path = "/opt/app-root/src"
+        # .cache dir contains the models and oci directories
+        # which can be quite big. We'll gather this only if
+        # specifying it via command line option
+        cache_dir = ".cache/instructlab"
+        # .config is where the configuration yaml files can
+        # be found. We gather this always.
+        config_dir = ".config/instructlab"
+        # In the .local directory we can find datasets,
+        # chat logs, taxonomies, and other very useful data
+        # We gather this always.
+        local_share_dir = ".local/share/instructlab"
+
+        # container paths
+        cont_cache_path = self.path_join(cont_opt_path, cache_dir)
+        cont_config_path = self.path_join(cont_opt_path, config_dir)
+        cont_local_path = self.path_join(cont_opt_path, local_share_dir)
+
+        self.add_forbidden_path([
+            self.path_join(cont_local_path,
+                           "taxonomy/.git"),
+            self.path_join(cont_local_path, "taxonomy/.github"),
+            self.path_join(cont_opt_path,
+                           "src/.local/share/instructlab/taxonomy/.git"),
+            self.path_join(cont_opt_path,
+                           "src/.local/share/instructlab/taxonomy/.github"),
+        ])
+
+        subcmds = [
+            'taxonomy diff',
+            'taxonomy diff --taxonomy-base=empty',
+            'system info',
+            'model list',
+            'config show',
+        ]
+
+        data_dirs = [
+            'data',
+            'generated',
+            'taxonomy',
+            'taxonomy_data',
+            'chatlogs',
+            'checkpoints',
+            'datasets',
+            'internal',
+            'phased',
+        ]
+
+        # If containerized, run commands in containers
+        try:
+            ilab_con = self.get_all_containers_by_regex("instructlab*")[0][1]
+        except Exception:  # pylint: disable=broad-except
+            ilab_con = None
+
+        self.add_copy_spec(
+            [f"{cont_share_conf_path}/rhel_ai_config.yaml",
+             f"{cont_config_path}/config.yaml"],
+            container=ilab_con
+        )
+        self.add_copy_spec(
+            [self.path_join(cont_local_path, data_dir)
+             for data_dir in data_dirs],
+            container=ilab_con
+        )
+        self.add_cmd_output(
+            [f"ilab {sub}" for sub in subcmds],
+            container=ilab_con
+        )
+        self.add_dir_listing(
+            cont_cache_path,
+            recursive=True,
+            container=ilab_con
+        )
+        if self.get_option('get-cache'):
+            self.add_copy_spec(
+                f"{cont_cache_path}",
+                container=ilab_con
+            )
+        self.add_container_logs(list(self.containers))
+
+        ilab_user = self.get_option("ilab-user")
+        try:
+            user_pwd = pwd.getpwnam(ilab_user)
+        except KeyError:
+            # The user doesn't exist, this will skip the rest
+            self._log_warn(
+                f'User "{ilab_user}" does not exist, will not '
+                'collect Instructlab information. Use '
+                '`-k instructlab.ilab-user` option to define '
+                'the user to use to collect data for Instructlab')
+            return
+
+        if user_pwd:
+            ilab_dir = user_pwd.pw_dir
+            if self.get_option("ilab-conf-dir"):
+                ilab_dir = self.path_join(
+                    ilab_dir, self.get_option('ilab-conf-dir')
+                )
+            data_dirs_base = self.path_join(ilab_dir, local_share_dir)
+
+            self.add_copy_spec(self.path_join(ilab_dir, config_dir))
+            self.add_copy_spec([
+                self.path_join(data_dirs_base, data_dir)
+                for data_dir in data_dirs
+            ])
+            self.add_dir_listing(
+                self.path_join(ilab_dir, cache_dir),
+                recursive=True
+            )
+
+            if self.get_option("get-cache"):
+                self.add_copy_spec(
+                    self.path_join(ilab_dir, cache_dir)
+                )
+
+# vim: set et ts=4 sw=4 :


sos/report/plugins/instructlab.py
```diff
@@ -0,0 +1,130 @@
...
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import (Plugin, IndependentPlugin, PluginOpt)
+
+
+class Instructlab(Plugin, IndependentPlugin):
+
+    short_desc = 'Instructlab'
+    plugin_name = 'instructlab'
+    profiles = ('ai',)
```
jcastill, line 18: Not really happy about this, but 'artificial-intelligence' and similar names were too long. I can remove the profile name for now if there are no other suggestions and this one doesn't look good.

jcastill:
Looking forward to suggestions about cleaning up this a bit.
The last section, when a user can be specified, will eventually have the same commands as with the containers, but I had some issues running them through both 'runas' and 'su -'. I want to investigate these issues further (mainly, timeouts), but I want to have this plugin ready asap so our support guys can start using it.

sos/report/plugins/instructlab.py
```diff
@@ -0,0 +1,130 @@
...
+                self.add_copy_spec(
+                    [f"{cont_local_path}/{data_dir}"
+                     for data_dir in data_dirs],
+                    container=cont
+                    )
+                self.add_cmd_output(
+                    [f"ilab {sub}" for sub in subcmds],
+                    container=cont
+                )
+                self.add_cmd_output(f"ls -laR {cont_cache_path}",
```
arif-ali, line 102: can use the new `self.add_dir_listing()` now :)

sos/report/plugins/instructlab.py
```diff
@@ -0,0 +1,130 @@
...
+            ilab_dir = f'/home/{ilab_user}'
+            if self.get_option("ilab_conf_dir"):
+                ilab_dir = f'{ilab_dir}{self.get_option("ilab_conf_dir")}'
+            data_dirs_base = f'{ilab_dir}{local_share_dir}'
+
+            self.add_copy_spec(f"{ilab_dir}{config_dir}")
+            self.add_copy_spec([
+                f"{data_dirs_base}/{data_dir}" for data_dir in data_dirs
+            ])
+            self.add_cmd_output(f"ls -laR {ilab_dir}/{cache_dir}")
```
arif-ali, line 123: same here, can use `self.add_dir_listing()`

arif-ali (commented):
some initial comments

sos/report/plugins/instructlab.py
```diff
@@ -0,0 +1,130 @@
...
+            f"{cont_opt_path}/src/.local/share/instructlab/taxonomy/.git",
+            f"{cont_opt_path}/src/.local/share/instructlab/taxonomy/.github",
+            ])
+
+        subcmds = [
+            'taxonomy diff',
+            'taxonomy diff --taxonomy-base=empty',
+            'system info',
+            'model list',
+            'config show'
```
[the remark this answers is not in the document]
jcastill: Fixed!

sos/report/plugins/instructlab.py
jcastill: nice! Changing this then

sos/report/plugins/instructlab.py
jcastill: And changing this one as well

sos/report/plugins/instructlab.py
```diff
@@ -0,0 +1,132 @@
...
+            self.add_copy_spec(f"{ilab_dir}{config_dir}")
+            self.add_copy_spec([
+                f"{data_dirs_base}/{data_dir}" for data_dir in data_dirs
+            ])
+            self.add_dir_listing(f"{ilab_dir}/{cache_dir}",
+                                 recursive=True)
+
+            if self.get_option("get-cache"):
+                self.add_copy_spec(
+                    f'{ilab_dir}/{cache_dir}'
```
TurboTurtle, line 129: Nit:  be consistent with the use of single or double quotes for f-strings. I personally use double-quotes for any f-string substitutions though I'm not of the mind to impose that as a project style, but it is helpful to be consistent within a plugin.

sos/report/plugins/instructlab.py
```diff
@@ -0,0 +1,132 @@
...
+                container_names.append(_con[1])
+
+        if in_container:
+            for cont in container_names:
+                self.add_copy_spec(
+                    [f'{cont_share_conf_path}rhel_ai_config.yaml',
+                     f'{cont_config_path}/config.yaml'],
+                    container=cont)
+                self.add_copy_spec(
+                    [f"{cont_local_path}/{data_dir}"
```
TurboTurtle, line 94: It is harder to follow what paths we're working with when some of the variables include the trailing `/` in dirs and others do not. This should be consistent within a given plugin even if we don't have a formal style recommendation for it.

sos/report/plugins/instructlab.py
```diff
@@ -0,0 +1,132 @@
...
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import (Plugin, IndependentPlugin, PluginOpt)
```
TurboTurtle, line 11: Superfluous parenthesis when imports don't span multiple lines.

sos/report/plugins/instructlab.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+class Instructlab(Plugin, IndependentPlugin):
+
```
TurboTurtle, line 15: A docstring explaining what the plugin is for and a high-level explanation of collections and options should be added for use with `sos help`.

TurboTurtle (changes requested):

sos/report/plugins/instructlab.py
```diff
@@ -0,0 +1,132 @@
...
+            'internal',
+            'phased',
+        ]
+
+        in_container = False
+        container_names = []
+        _containers = self.get_containers()
+
+        for _con in _containers:
+            if _con[1].startswith('instructlab'):
```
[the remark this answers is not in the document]
jcastill: ack

sos/report/plugins/instructlab.py
```diff
@@ -0,0 +1,132 @@
...
+                in_container = True
+                container_names.append(_con[1])
+
+        if in_container:
```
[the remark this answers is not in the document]
jcastill: At least in version 1.1 of RHEL AI, instructlab can run both in a container and outside a container, in the same installation. This may change for 1.2 and later

sos/report/plugins/instructlab.py
```diff
@@ -0,0 +1,132 @@
...
+                    container=cont)
+                self.add_copy_spec(
+                    [f"{cont_local_path}/{data_dir}"
+                     for data_dir in data_dirs],
+                    container=cont
+                    )
+                self.add_cmd_output(
+                    [f"ilab {sub}" for sub in subcmds],
+                    container=cont
+                )
```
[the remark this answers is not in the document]
jcastill: ack

sos/report/plugins/instructlab.py
```diff
@@ -0,0 +1,132 @@
...
+                    self.add_copy_spec(
+                        f'{cont_cache_path}',
+                        container=cont
+                        )
+                self.add_container_logs(cont)
+
+        ilab_user = self.get_option("ilab_user") if \
+            self.get_option("ilab_user") else ''
+        if ilab_user:
+            ilab_dir = f'/home/{ilab_user}'
```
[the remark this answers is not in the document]
jcastill: In the case of RHEL AI it is... the default user will live in /home always.

sos/report/plugins/instructlab.py
jcastill: Yeah, I totally missed to change this while reviewing, sorry

sos/report/plugins/instructlab.py
jcastill: Same, missed arranging some consistency while reviewing. Will fix this as well in t he next push

sos/report/plugins/instructlab.py
jcastill: ack

sos/report/plugins/instructlab.py
jcastill: ack

sos/report/plugins/instructlab.py
```diff
@@ -0,0 +1,132 @@
...
+                if self.get_option("get-cache"):
+                    self.add_copy_spec(
+                        f'{cont_cache_path}',
+                        container=cont
+                        )
+                self.add_container_logs(cont)
+
+        ilab_user = self.get_option("ilab_user") if \
+            self.get_option("ilab_user") else ''
+        if ilab_user:
```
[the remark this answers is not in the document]
arif-ali: @TurboTurtle we are more than happy to go ahead with py38 as agreed for future developments

sos/report/plugins/instructlab.py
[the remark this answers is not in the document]
pmoravec: Can't one override the home directory (of one user or all)? Well, the bootc's kickstart will create the home directory *there*, but the KS template can be customized (and the docs even assumes the `/home/<user>` is not hardcoded, per sentence "The default location for the InstructLab data is in the home/<user> directory."). So I would vote for kind-of `~{ilab_user}` approach than the hardcoded path.

sos/report/plugins/instructlab.py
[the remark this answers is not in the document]
pmoravec: As stated in the [#3533](https://github.com/orgs/sosreport/discussions/3533#discussioncomment-8916412), ok for us as well.

sos/report/plugins/instructlab.py
pmoravec: I would be more concise here. https://instructal.ai advertises itself as "A new community-based approach to build truly open-source LLMs" and that seems legit for me (well, haveng the "a new (thing)" in sos source for years will become ridiculous, but.. at least not now).

sos/report/plugins/instructlab.py
```diff
@@ -0,0 +1,131 @@
...
+        PluginOpt('get-cache', default=False,
+                  desc='Capture models and osci cached data')
+    ]
+
+    def setup(self):
+        cont_share_conf_path = "/usr/share/instructlab/config"
+        cont_opt_path = "/opt/app-root/src"
+        # .cache dir contains the models and oci directories
+        # which can be quite big. We'll gather this only if
+        # specifying it via command line option
```
pmoravec, line 45: Collecting the model itself, even guarded by an option..? The default Granite model is 5GB-ish big, is it worth packing it to sosreport at all?

sos/report/plugins/instructlab.py
```diff
@@ -0,0 +1,131 @@
...
+        config_dir = "/.config/instructlab"
+        # In the .local directory we can find datasets,
+        # chat logs, taxonomies, and other very useful data
+        # We gather this always.
+        local_share_dir = "/.local/share/instructlab"
+
+        # container paths
+        cont_cache_path = f"{cont_opt_path}{cache_dir}"
+        cont_config_path = f"{cont_opt_path}{config_dir}"
+        cont_local_path = f"{cont_opt_path}{local_share_dir}"
```
pmoravec, line 58: I would prefer using `os.path.join` method to joining.. well.. paths :)

sos/report/plugins/instructlab.py
jcastill: I was concerned about the same. Default Granite is 4.1G, and the ilab options should give a good idea about what's installed, the diffs between base and changes, etc. But from engineering they told me that it could be useful in certain cases. That's why I decided not to get it by default.

sos/report/plugins/instructlab.py
jcastill: ack

sos/report/plugins/instructlab.py
jcastill: So you want me to remove everything and add the instructlab.ai definition?

pmoravec (approved):

sos/report/plugins/instructlab.py
jcastill: Ah sorry @pmoravec I meant the profile, 'ai'. If feels kind of weird that's so small, but I don't think there's another way to specify it. And yes, there are more plugins that should go into this profile, but I'm still waiting for the whole list. 
For the doctext I tried to copy what we have in our docs, because I thought it would be clearer, and then added a note that this can run both as container and as a non-container. Should I try to make it clearer?

sos/report/plugins/instructlab.py
```diff
@@ -0,0 +1,134 @@
...
+            'datasets',
+            'internal',
+            'phased',
+        ]
+
+        ilab_con = None
+        for con in self.containers:
+            if self.get_container_by_name(con):
+                ilab_con = con
+                break
```
TurboTurtle, line 92: This can be significantly more efficient and readable as a one-liner using the regex supporting container method.

sos/report/plugins/instructlab.py
```diff
@@ -0,0 +1,134 @@
...
+        self.add_copy_spec(
+            [f"{cont_local_path}/{data_dir}"
+             for data_dir in data_dirs],
+            container=ilab_con)
+        self.add_cmd_output(
+            [f"ilab {sub}" for sub in subcmds],
+            container=ilab_con)
+        self.add_dir_listing(cont_cache_path,
+                             recursive=True,
+                             container=ilab_con)
```
TurboTurtle, line 107: Same as the other paths, be consistent and use `Plugin.path_join()`.

sos-style still needs addressing.

sos/report/plugins/instructlab.py
```diff
@@ -0,0 +1,134 @@
...
+    """
+
+    short_desc = 'Instructlab'
+    plugin_name = 'instructlab'
+    profiles = ('ai',)
+    containers = ('instructlab', 'ilab',)
+    commands = ('ilab',)
+
+    option_list = [
+        PluginOpt('ilab_user', default='', val_type=str,
```
TurboTurtle, line 33: Is there not a default user to collect for?

TurboTurtle (changes requested):

sos/report/plugins/instructlab.py
```diff
@@ -0,0 +1,134 @@
...
+                  desc='user that runs instructlab'),
+        PluginOpt('ilab_conf_dir', default='', val_type=str,
+                  desc='instructlab data directory'),
```
jcastill, line 36: Moving this option and `ilab_user` to `ilab-conf-dir` and `ilab-user` in the next push

arif-ali (approved):

PR #3805, opened 2024-10-14
[Plugin] Check if a path exists inside a container in add_dir_listing()

jcastill:
Currently, add_dir_listing() checks if a path exists outside of a container, and when we call the command for a container, it fails to collect any listing. This patch adds a new function container_path_exists() that checks for the specified path in add_dir_listing() inside the container.

---

### DIFF
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -2075,7 +2075,11 @@ class Plugin():
         if isinstance(paths, str):
             paths = [paths]
 
-        paths = [p for p in paths if self.path_exists(p)]
+        if container:
+            paths = [p for p in paths if
+                     self.container_path_exists(p, container=container)]
+        else:
+            paths = [p for p in paths if self.path_exists(p)]
 
         if not tree:
             options = f"alZ{'R' if recursive else ''}"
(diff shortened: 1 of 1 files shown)


sos/report/plugins/__init__.py
```diff
@@ -3407,6 +3411,22 @@ def setup_verify(self):
...
+                        in a container
+        :type path:     ``str``
+
+        :param container: The container where to check for the path
+        :type container: ``str``
+
+        :returns:       True if the path exists in the container, else False
+        :rtype:         ``bool``
+        """
+        return self.exec_cmd(f"test -e {path}", container=container)
```
jcastill, line 3428: I'm sure there are better ways to do this, but I didn't find any way in our code already, so I had to come up with this 'dirty' trick. 
As always, open to suggestions to do this in a better way :)

pmoravec (commented):
ACK in general. I would expect there is more elegant solution than the ` self.exec_cmd(f"test -e {path}", container=container)` but I cant find any /o\.

arif-ali (approved):
I'm ok with this. I think as we have the container abstraction, there doesn't seem to  be an elegant way of doing it

pmoravec (approved):

PR #3807, opened 2024-10-14
[ssh] Stop collecting 'ls' information from non-local home dirs

jcastill:
Commands like 'ls -laZ' can take some time to run in remote filesystems, for example NFS. When you have hundreds of them, sos can take a long time to walk through it, and could be even worse if we use the option 'recursive'.
For ssh home dirs, we don't need the listing, so in this patch we detect the type of filesystem and run the command only on local ones.

Related: RHEL-22389, SUPDEV-156

---

### DIFF
--- a/sos/report/plugins/ssh.py
+++ b/sos/report/plugins/ssh.py
@@ -80,8 +80,25 @@ class Ssh(Plugin, IndependentPlugin):
         """
         users_data = pwd.getpwall()
 
+        fs_mount_info = {}
+        try:
+            with open('/proc/mounts', "r", encoding='UTF-8') as mounts_file:
+                for line in mounts_file:
+                    (fs_file, fs_vstype) = line.split()[1:3]
+                    fs_mount_info[fs_file] = fs_vstype
+        except Exception:
+            self._log_error("Couldn't read /proc/mounts")
+            return
+        non_local_fs = {'nfs', 'nfs4', 'autofs'}
         # Read the home paths of users in the system and check the ~/.ssh dirs
         for user in users_data:
+            if user.pw_dir in fs_mount_info and \
+                    fs_mount_info[user.pw_dir] in non_local_fs:
+                self._log_info(
+                        f"Skipping capture in {user.pw_dir}"
+                        " because it's a remote directory"
+                )
+                continue
             home_dir = self.path_join(user.pw_dir, '.ssh')
             self.add_dir_listing(home_dir)
 


jcastill:
This is a different approach to what I was trying to solve in #3490 and I opened a new PR to keep both approaches separated. 
In this case, we'll only skip these homedirs that are not mounted locally. In the other PR, I was dancing between disable the option completely or discarding all homedirs completely. 

I tried to find if we gather this info in other places, but I could only find the function _get_devices_by_fstype() that I think is slightly different. Would it make sense to create a _get_mounted_fstypes() or similar in sos/report/__init__.py and call that from the ssh (and other) plugins?

pmoravec:
The change makes sense; maybe worth having an option to keep current behaviour..? (no idea how useful it would be, though).

Will this make #3490 obsolete?

jcastill:
> The change makes sense; maybe worth having an option to keep current behaviour..? (no idea how useful it would be, though).
> 
I thought about that, but I keep being told that the reward for getting this info is not worth the slowness running sos that we get when going through NFS filesystems

> Will this make #3490 obsolete?

Yes, I'll keep open only the one that you all agree it's a better approach.

pmoravec (approved):

sos/report/plugins/ssh.py
```diff
@@ -80,8 +80,21 @@ def user_ssh_files_permissions(self):
...
+                    fs_mount_info[fs_file] = fs_vstype
+        except Exception:
+            self._log_error("Couldn't read /proc/mounts")
+            return
+        non_local_fs = {'nfs', 'nfs4', 'autofs'}
         # Read the home paths of users in the system and check the ~/.ssh dirs
         for user in users_data:
+            if user.pw_dir in fs_mount_info and \
+                    fs_mount_info[user.pw_dir] in non_local_fs:
+                continue
```
TurboTurtle, line 97: We should log something so that end users aren't left wondering why this wasn't collected when previously they may have been expecting it.

TurboTurtle (changes requested):

sos/report/plugins/ssh.py
pmoravec: What verbosity level? Info? (or debug? isnt it too hidden..?)

jcastill: With warn I get this:

```
 Setting up archive ...
 Setting up plugins ...
[plugin:ssh] Skipping capture in /home/jcastill because it's a remote directory

  Starting 1/1   ssh             [Running: ssh]
```

pmoravec: Yeah, and I dont know if that is not too "vocal". I think info verbosity is the best, but I am ok with any.

jcastill: You are right, in the specific scenario of tens or hundreds of NFS mounted homedirs, the output would be huge.

jcastill: Exactly. I'm pushing a change with info now

arif-ali (approved):

pmoravec (approved):

TurboTurtle (approved):

PR #3810, opened 2024-10-18
[coredump] Add sizelimit for coredump

jcastill:
While most coredumps could be relatively small,
in certain environments, like Ceph, these can be
as big as 5GB. This change adds a limit of 100MB
to the latest core collection.

Related: RHEL-62972

---

### DIFF
--- a/sos/report/plugins/coredump.py
+++ b/sos/report/plugins/coredump.py
@@ -32,7 +32,7 @@ class Coredump(Plugin, IndependentPlugin):
             "/usr/lib/systemd/coredump.conf.d/"
         ])
 
-        self.add_cmd_output("coredumpctl dump")
+        self.add_cmd_output("coredumpctl dump", sizelimit=100)
 
         coredump_list = self.collect_cmd_output("coredumpctl list")
         if self.get_option("detailed") and coredump_list['status'] == 0:


pmoravec:
This would mean we grab 100MB of incomplete / usefulness data for coredumps >100MB. Not great though no idea how to simply prevent that.

jcastill:
OK, I thought it was "grab the file only if max is 100MB". I'll add a check then, because incomplete core dumps are useless

jcastill:
@pmoravec some options:
a- Stop gathering the latest coredump.
b- Increase the limit for coredumps smaller than, lets say, 500M
c- Add an option to add_cmd_output for 'max_size' that only gathers commands less than that argument?
d- Add an option 'tailit' like in add_copy_spec, so if size is bigger than sizelimit and it's a binary, skip the collection and log it?
e- Other option I didn't think about?

TurboTurtle:
> OK, I thought it was "grab the file only if max is 100MB". I'll add a check then, because incomplete core dumps are useless

That's for binary files. This is a command capture.

> c- Add an option to add_cmd_output for 'max_size' that only gathers commands less than that argument?
> d- Add an option 'tailit' like in add_copy_spec, so if size is bigger than sizelimit and it's a binary, skip the collection and log it?

We would need to hold the entire command output in memory to determine its size, which is not desirable here. Any `sizelimit` that isn't `0` means we store it in memory, otherwise we write directly to a file to not potentially OOM systems where command output can be large.

TurboTurtle:
I see there's a lot we can do with the info output, and generally spruce up the collections at large. I'll throw something together and get something posted on my lunch.

arif-ali:
do we need this, if the re-work has been done?

jcastill:
I don't think we need this anymore, so I'll close it.

Issue #3819, opened 2024-10-23
[saltmaster] Failure when pillar_roots undefined in any config


TrevorBenson:
```
# cat sos_logs/saltmaster-plugin-errors.txt
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/sos/report/__init__.py", line 1231, in setup
    plug.setup()
  File "/usr/lib/python3.6/site-packages/sos/report/plugins/saltmaster.py", line 32, in setup
    self.add_pillar_roots()
  File "/usr/lib/python3.6/site-packages/sos/report/plugins/saltmaster.py", line 55, in add_pillar_roots
    yaml.safe_load(f).get("pillar_roots", {}).get("base", [])
AttributeError: 'NoneType' object has no attribute 'get'
```

PR #3820, opened 2024-10-23
[saltmaster] handle undefined pillar_roots

TrevorBenson:
AttributeError: 'NoneType' object has no attribute 'get'
Resolves: #3819

---

### DIFF
--- a/sos/report/plugins/saltmaster.py
+++ b/sos/report/plugins/saltmaster.py
@@ -52,10 +52,13 @@ class SaltMaster(Plugin, IndependentPlugin):
         all_pillar_roots = []
         for cfg in cfgs:
             with open(cfg, "r", encoding='UTF-8') as file:
-                cfg_pillar_roots = (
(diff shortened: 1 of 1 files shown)


TurboTurtle (approved):

pmoravec (approved):
Maybe we might add a debug log to the `except` block, to let user understand why something was not collected. But since I dont know how useful such logging would be, I am leaving it on the author to decide.

ACK from me to the PR as is now.

TrevorBenson:
> Maybe we might add a debug log to the `except` block, to let user understand why something was not collected. But since I dont know how useful such logging would be, I am leaving it on the author to decide.
> 
> ACK from me to the PR as is now.

As it is up to the admin how to organize configurations the files in `/etc/salt/master.d` have no requirement to each contain a `pillar_roots`. The plugin already collects the `/etc/salt` files for review. As such I don't see much benefit to debug logging about each file without it.

PR #3821, opened 2024-10-23
[foreman] Size limit production.log and ssl logs

TurboTurtle:
Adds a (high) default sizelimit to `production.log` and foreman `*ssl.log` collections to avoid potentially causing out of space issues on systems with large active logs.

---

### DIFF
--- a/sos/report/plugins/foreman.py
+++ b/sos/report/plugins/foreman.py
@@ -91,11 +91,10 @@ class Foreman(Plugin):
         _host_f = self.exec_cmd('hostname -f')['output']
         _host_f = _host_f.strip()
 
-        # Collect these completely everytime
         self.add_copy_spec([
             "/var/log/foreman/production.log",
             f"/var/log/{self.apachepkg}*/foreman-ssl_*_ssl.log"
-        ], sizelimit=0)
+        ], sizelimit=500)
 
         # Allow limiting these
         self.add_copy_spec([


TurboTurtle:
@pmoravec let's chat about this before merging anywhere.

I understand this log is very verbose and big, hence the current "collect it all" posture. However, I just ran into a situation where collecting an sos report on our (thankfully, lab) Satellite caused an out-of-space issue due to our production.log being over 1.5GB and having a relatively small /var/tmp partition per policy.

So, I'd like to add at least some kind of bounding here, but I get that it needs to be unusually (for sos) high for a default limit for it to be of any use to support engineers. 500 was a bit of an off the hip number, so happy to discuss raising or lowering it.

pmoravec:
Cc @pafernanr who knows more about max sizes of the logfile(s).

I hit some customers who disabled logrotation and had either of the two files really / ridiculously huge. So some limit makes snese. Let me do some checks internally, we should reply in a week or two.

pafernanr:
Hello,

I think I will be able to provide valuable stats for `production.log`. But I only have access to a "small" number of `foreman-ssl_access_ssl.log`. Let me sometime to get it and I will share the output with you.

Regards,

pafernanr:
Find below inittial stats for over 200 sos reports (included @pmoravec and other colleagues)
### access.log
Mb(>)   count
0       188
200     18
400     8
600     6
800     3
1000    1
1600    2
2000    1
2400    1
2600    1
3600    1
9400    1
### production.log
Mb(>)   count
0       202
200     17
400     6
600     1
800     2
1000    1
1200    1
2000    1
2600    1
9800    1

I'm still trying to get more accurate stats for production.log. Let's see what can be done.

TurboTurtle:
@pmoravec @pafernanr - any preferences based on that data? It looks like 500 would get the majority, but I am also open to raising it a bit further.

pafernanr:
Hello @TurboTurtle,

Setting it to 500 is a good choice IMO. The extra time/resource consumption should be asumible for those big-foreman-infrastructure users, and that file size should contain most than enough time range to look at any "recent" issue. 

@pmoravec ¿?

pmoravec:
+1 for 500, seems as the best compromise.

pmoravec (approved):

Issue #3688, opened 2024-06-26
[RFE] Set a UserAgent for --upload-protocol s3


TrevorBenson:
# Request
Set a custom UserAgent for S3 (boto3) so that sos report can be identified as the upload client for S3 buckets. This could be a string like `sos`, `sosreport`, or `sos-report` etc..

# Reason

Permits identifying sos report/collect as a specific UserAgent instead of generically reporting the same UseAgent any Python script using boto3 does. Custom policies can be defined on s3 buckets. One use of a policy could be a condition/requirement for the UserAgent string. Two potential use cases by using a custom UserAgent in a policy:
* Limit uploads even, when valid credentials are provided, to only work with the specified UserAgent.
* Allow a specific UserAgent to upload an archive to a bucket without requiring valid credentials (access/secret keypair).

TrevorBenson:
@pmoravec @TurboTurtle Is `sosreport` acceptable as the UserAgent label for uploads from both `sos report` and `sos collect`? Did I miss any UserAgent which might be preferred instead?

TIA

TurboTurtle:
Good question. A safe default may just be `sos`, to cover both use cases. Alternatively, we could just match it to the component e.g. `sos-report` and `sos-collect`, if it would for any reasonable be desirable to not glob everything together.

PR #3823, opened 2024-10-25
[sos] Set a User Agent for s3 based uploads

TrevorBenson:
Signed-off-by: Trevor Benson <trevor.benson@gmail.com>

---


jcastill (approved):
LGTM
Just a thought, and please excuse my ignorance if it's not applicable... but would it be better to add an option for the command line that allows to set up an s3_user_agent? So the user  could specify a user custom agent, use a default one (app/sos), or set None?

TrevorBenson:
> LGTM Just a thought, and please excuse my ignorance if it's not applicable... but would it be better to add an option for the command line that allows to set up an s3_user_agent? So the user could specify a user custom agent, use a default one (app/sos), or set None?

It is an option, but it wasn't discussed in the original issue #3688. It may be possible to use AWS_EXECUTION_ENV to achieve this, but I haven't tested it.

FWIW, I did not find examples that allow setting the user agent string via a flag in many s3 CLI applications. That is not to say they don't exist; Simply that I did not find them previously when looking.

TrevorBenson:
@arif-ali rpm-build:fedora-41-ppc64le & rpm-build:fedora-41-s390x still displayed as if they were running 2 weeks after your message to packit. I asked packit to rebuild failures again, but they do not seem to progress.

sos/policies/distros/__init__.py
```diff
@@ -997,10 +998,13 @@ def upload_s3(self, endpoint=None, region=None, bucket=None, prefix=None,
         if not secret_key:
             secret_key = self.get_upload_s3_secret_key()
 
+        boto3_config = BotocoreConfig(user_agent_extra='app/sos')
```
TrevorBenson, line 1001: I set it to `app/sos` because in a later version of boto3 (once Python 3.9 is the minimum supported version) the `user_agent_appid="sos"` would append "app/sos" to the user agent. This way in the future if or when sos migrates to using `user_agent_appid` the agent string for sos can remain consistent.

TurboTurtle (approved):

arif-ali (approved):

arif-ali:
> @arif-ali rpm-build:fedora-41-ppc64le & rpm-build:fedora-41-s390x still displayed as if they were running 2 weeks after your message to packit. I asked packit to rebuild failures again, but they do not seem to progress.

I don't thing it matters here much, as it's not based on your change, so all good imho. Maybe someone who knows pack will have better ideas

arif-ali:
imho, user agent is not something that is changed in applications, and it is usually a definition of which application came through. In my opinion, having the option to change would not make sense, but that's my opinion.

Issue #3827, opened 2024-10-31
[collect] Fails for the local node when using --nopasswd-sudo


TrevorBenson:
# Problem
Executing sos collect results in failures on the local node executing `sos collect via sudo`.

`sudo sos collect --nopasswd-sudo --ssh-key /etc/project/id_edd2519 --ssh-user project-user --nodes 172.31.100.121,172.31.100.85`

```
Beginning collection of sosreports from 2 nodes, collecting a maximum of 4 concurrently

system1  : Generating sos report...
system2        : Generating sos report...
system2        : Retrieving sos report...
system2        : Successfully collected sos report
system1  : Error running sos report: sudo attempt failed
[system1:sosreport] Error during sos execution: sudo attempt failed
```

When **project-user** executes `sudo sos report` directly on either system1 or system2 a valid sos archive is created in `/var/tmp`. No errors are observed ab out 

* **OS**: Enterprise Linux 8 (Rocky Linux 8.10 in this case)
* **Version of sos**: sos-4.7.2-2.el8_10.noarch

## Environment Details

The environment has 2 nodes hardened with no direct shell access to root. Executing `sos report`, or `sos collect` requires using `sudo`. The **project-user** account exists on both nodes and has sudoers permissions to run `sos`.

The public ssh key `/etc/project/id_edd2519.pub` is part of `/home/project-user/.ssh/authorized_keys` on both system1 and system2.

### Sudoers Configuration
```
$ sudo grep SOS /etc/sudoers.d/project
Cmnd_Alias SOS = /usr/sbin/sos, /usr/sbin/sosreport, /usr/bin/chown * /var/tmp/sosreport-*
project-user ALL=(root) EXEC:/home/project-user/project-installer, SOS
```

## Additional Details

Checking the sos_logs/sos.log it mentions something wrong with the account password even when using `--nopasswd-sudo` and having confirmed that sudo works perfectly fine without a password request to execute `sudo sos report` on either system:
```
2024-10-31 20:56:52,518 DEBUG: [sos_collector:__init__] Executing /usr/sbin/sos collect --nopasswd-sudo --debug --verbose --ssh-key /etc/project/pki/salt-bootstrap --ssh-user project-user --nodes 172.31.100.121,172.31.100.85 -o block
[...]

PR #3830, opened 2024-11-01
[collect] Fix exception while parsing sos help

jcastill:
While parsing the output of 'sos report -l', we
were attempting to split an empty line, and
getting the following exception:

[<ip address>:_regex_sos_help] Error parsing sos help: list index out of range

Related: #3827

---

### DIFF
--- a/sos/collector/sosnode.py
+++ b/sos/collector/sosnode.py
@@ -367,7 +367,8 @@ class SosNode():
             for line in result.splitlines():
                 if not is_list:
                     try:
-                        res.append(line.split()[0])
+                        if ls := line.split():
+                            res.append(ls[0])
                     except Exception as err:
                         self.log_debug(f"Error parsing sos help: {err}")
                 else:


sos/collector/sosnode.py
```diff
@@ -367,7 +367,8 @@ def _regex_sos_help(self, regex, sosinfo, is_list=False):
             for line in result.splitlines():
                 if not is_list:
                     try:
-                        res.append(line.split()[0])
+                        if line.split():
```
jcastill, line 370: We could do something else instead of calling split() twice, let me know if you have any other preference.

pmoravec: Yeah I am on torns here. Comparing performance on big text files:

    if line.split():
        print(line.split()[0])

took:

    real	0m5.488s
    user	0m5.090s
    sys	0m0.324s

while:

    ls = line.split()
    if ls:
        print(ls[0])

took:

    real	0m4.161s
    user	0m3.841s
    sys	0m0.252s

jcastill: Then lets go for the second option

pmoravec (approved):

arif-ali (approved):

arif-ali:
restarting the failed job

sos/collector/sosnode.py
```diff
@@ -367,7 +367,9 @@ def _regex_sos_help(self, regex, sosinfo, is_list=False):
             for line in result.splitlines():
                 if not is_list:
                     try:
-                        res.append(line.split()[0])
+                        ls = line.split()
+                        if ls:
```
TurboTurtle, line 371: If we're concerned about empty lines, and we're concerned about performance given the earlier conversation, then we should be applying the conditional before the `.split()`.

```python
for line in results.splitlines():
    if not is_list:
        if not line:
            continue
    [...]
```

jcastill: I tried this, and it covers all the empty lines, but there's a case where we print a single space and the exception gets printed. Playing with the output, this is what we get:

```
'a python.hashes             off             collect hashes for all python files'
'a rpm.rpmq                  on              query package information with rpm -q'
'a rpm.rpmva                 off             verify all packages'
'a rpm.rpmdb                 off             collect /var/lib/rpm'
'a selinux.fixfiles          off             collect incorrect file context labels'
'a services.servicestatus    off             collect status of all running services'
'a ssh.userconfs             on              Changes whether module will collect user .ssh configs'
''
' '
[localhost:_regex_sos_help] Error parsing sos help: list index out of range
```

That space is coming from here:

https://github.com/sosreport/sos/blob/main/sos/report/__init__.py#L1093

Any simple way to cover this case as well, without having to change the code in report? 
A walrus operator here would be more useful and cover all cases, but we are still waiting for  a decision on #3533 , right?

arif-ali: @jcastill walrus is good, we already discussed this :)

jcastill: Perfect, I'll push a new version soon

pmoravec (approved):

arif-ali (approved):

TurboTurtle (approved):

Issue #3839, opened 2024-11-15
[collect] Using batch flag with password blocks requesting the ssh password


TrevorBenson:
SOS Version: sos-4.7.2-3.el9.noarch

```
$ man sos-collect |grep -A1 -e '\-\-batch$' -e '\-\-password$'
       --batch
              Run in non-interactive mode. This will skip prompts for user input, with the exception of a prompt for the SSH password.
--
       --password
              Specifying this option will cause sos collect to prompt the user for an SSH password that will be used to connect to all nodes.
```

Using `--batch --password` leads to an **unsupported operand type(s) for +: 'bool' and 'str'** error:

```
[root@localhost ~]# sos collect --batch --password --ssh-user project-user --nodes 192.168.124.121

sos-collector (version 4.7.2)

This utility is used to collect sos reports from multiple nodes simultaneously.
Remote connections are made and/or maintained to those nodes via well-known
transport protocols such as SSH.

An archive of sos report tarballs collected from the nodes will be generated in
/var/tmp/sos.mwlhdiqu and may be provided to an appropriate support
representative.

The generated archive may contain data considered sensitive and its content
should be reviewed by the originating organization before being passed to any
third party.

No configuration changes will be made to the system running this utility or
remote systems that it connects to.


The following is a list of nodes to collect from:
	localhost
	192.168.124.121


Connecting to nodes...
[192.168.124.121:__init__] Unable to open remote session: unsupported operand type(s) for +: 'bool' and 'str'
Collection would only gather from localhost due to failure to either enumerate or connect to cluster nodes. Assuming single collection from localhost is not desired.
Aborting...
```

This was found while using a fix for nested sudo in PR #3838

jcastill:
I managed to reproduce and get a call trace:

```
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/sos/collector/sosnode.py", line 89, in __init__
    self._transport.connect(self._password)
  File "/usr/lib/python3.9/site-packages/sos/collector/transports/__init__.py", line 146, in connect
    if self._connect(password):
  File "/usr/lib/python3.9/site-packages/sos/collector/transports/control_persist.py", line 134, in _connect
    res.sendline(password)
  File "/usr/lib/python3.9/site-packages/pexpect/pty_spawn.py", line 578, in sendline
    return self.send(s + self.linesep)
TypeError: unsupported operand type(s) for +: 'bool' and 'str'
[...]

PR #3840, opened 2024-11-15
[collect] Fix exception when using collect with batch and password

jcastill:
Using --batch --password together in sos-collect leads to an error:

Unable to open remote session:
unsupported operand type(s) for +: 'bool' and 'str'

This PR removes the check for the batch option, enabling
the prompt for password as specified in the man page:

  --batch
	  Run in non-interactive mode.
	 This will skip prompts for user input,  with  the
	 exception  of  a prompt for the SSH password.

Related: #3839

---

### DIFF
--- a/man/en/sos-collect.1
+++ b/man/en/sos-collect.1
@@ -74,8 +74,7 @@ This does NOT enable all sos collect options.
 Become the root user on the remote node when connecting as a non-root user.
 .TP
 \fB\-\-batch\fR
-Run in non-interactive mode. This will skip prompts for user input, with the
-exception of a prompt for the SSH password.
+Run in non-interactive mode. This will skip prompts for user input.
 .TP
 \fB\-\-all\-logs\fR
 Report option. Collects all logs regardless of size. 
--- a/sos/collector/__init__.py
+++ b/sos/collector/__init__.py
@@ -1194,6 +1194,14 @@ this utility or remote systems that it connects to.
 
         self.intro()
 
+        if self.opts.batch and self.opts.password:
+            self.exit(
+                "\nsos-collector was called with incompatible options --batch "
+                "and --password.\nIf you need to use --password,"
+                " please omit batch mode.\n",
+                1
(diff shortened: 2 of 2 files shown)


jcastill:
The call trace we get when this error occurs is the following:

```
  File "/usr/lib/python3.9/site-packages/sos/collector/sosnode.py", line 89, in __init__
    self._transport.connect(self._password)
  File "/usr/lib/python3.9/site-packages/sos/collector/transports/__init__.py", line 146, in connect
    if self._connect(password):
  File "/usr/lib/python3.9/site-packages/sos/collector/transports/control_persist.py", line 134, in _connect
    res.sendline(password)
  File "/usr/lib/python3.9/site-packages/pexpect/pty_spawn.py", line 578, in sendline
    return self.send(s + self.linesep)
TypeError: unsupported operand type(s) for +: 'bool' and 'str'
```

But I thought it may be better to catch the combination of options earlier, in execute() in SoSCollector, and exit there. 
I'm not really sure about the actual message I wrote, so if anybody has any other suggestions please let me know.

jcastill:
New fix, that takes into account the fact that we can use --password with --batch in sos-collect

pmoravec:
As stated in [this comment](https://github.com/sosreport/sos/issues/3839#issuecomment-2483811798), I think the more expected `sos collect` behaviour is ignoring `--password` whenever `--batch` is used. But if somebody has a different opinion, no big deal.

arif-ali:
> As stated in [this comment](https://github.com/sosreport/sos/issues/3839#issuecomment-2483811798), I think the more expected `sos collect` behaviour is ignoring `--password` whenever `--batch` is used. But if somebody has a different opinion, no big deal.

I'd agree, the whole ethos of batch would be that no interaction is needed; at least that was my assumption too. So ignoring `--password` is valid, and maybe add a warning message instead that the option is not valid in this scenarios

arif-ali (approved):

PR #3842, opened 2024-11-18
[kubernetes] Add support for k8s snap

arif-ali:
Also, fix command for microk8s, and use variable

---

### DIFF
--- a/sos/report/plugins/kubernetes.py
+++ b/sos/report/plugins/kubernetes.py
@@ -105,6 +105,8 @@ class Kubernetes(Plugin):
     def setup(self):
         self.add_copy_spec(self.config_files)
 
+        self.add_forbidden_path("/etc/kubernetes/pki")
+
         self.add_env_var([
             'KUBECONFIG',
             'KUBERNETES_HTTP_PROXY',
@@ -294,7 +296,9 @@ class RedHatKubernetes(Kubernetes, RedHatPlugin):
 
 class UbuntuKubernetes(Kubernetes, UbuntuPlugin, DebianPlugin):
 
-    packages = KUBE_PACKAGES
+    packages = KUBE_PACKAGES + (
+        'k8s',
+    )
 
     files = KUBECONFIGS + (
         '/root/cdk/cdk_addons_kubectl_config',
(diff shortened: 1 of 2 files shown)


jcastill (approved):

sos/report/plugins/k8s.py
```diff
@@ -0,0 +1,75 @@
...
+
+        self.add_copy_spec([
+            "/etc/kubernetes/"
+            f"{k8s_common}/args",
+            f"{k8s_common}/var/lib/k8s-dqlite/info.yaml",
+            f"{k8s_common}/var/lib/k8s-dqlite/cluster.yaml",
+            f"{k8s_common}/var/lib/k8sd/state/truststore/k8s.yaml",
+            f"{k8s_common}/var/lib/k8sd/state/database/info.yaml",
+            f"{k8s_common}/var/lib/k8sd/state/database/cluster.yaml",
+            f"{k8s_common}/var/lib/k8sd/state/daemon.yaml",
```
pmoravec, line 48: The `yaml` files cant contain any credentials worth to `postproc`, right?

arif-ali: I have gone through these files maually twice :), and there isn't anything to obfuscate, except for `/var/snap/k8s/common/var/lib/k8sd/state/truststore/k8s.yaml`, which I've already added in `postproc()`

I am now getting all of this verified by our k8s SMEs to ensure we got everything

TurboTurtle:
Plugin looks fine. My only concern is around naming. We have `kubernetes` as well as `microk8s` already today. `k8s` is a well-known shorthand for `kubernetes`, so it may start being a little opaque what plugin collects what for "$defacto_standard_container_orchestration_platform"

arif-ali:
I've moved it now to the `kubernetes.py` plugin, hopefully that makes much more sense. Always good to think outside of the box, and this does make better sense.

This also made me realise that we were collecting a folder, and obfuscating the said folder. So, actually needed less code adding.

Maybe we can also move the `microk8s`  here, and have everything under the one umbrella, I'll leave that thought here, and see what others think

TurboTurtle (approved):

PR #3850, opened 2024-11-22
[cleaner] Skip obfuscation of substrings for hostnames properly

pmoravec:
Leftover from #3403 / #3496 where match_full_words_only is ignored for hostname mapping due to own get_regex_result method.

Resolves: #3593

---

### DIFF
--- a/sos/cleaner/mappings/hostname_map.py
+++ b/sos/cleaner/mappings/hostname_map.py
@@ -87,7 +87,7 @@ class SoSHostnameMap(SoSMap):
         """
         if '.' in item:
             item = item.replace('.', '(\\.|_)')
-        return re.compile(item, re.I)
+        return super().get_regex_result(item)
 
     def set_initial_counts(self):
         """Set the initial counter for host and domain obfuscation numbers
--- a/sos/cleaner/mappings/__init__.py
+++ b/sos/cleaner/mappings/__init__.py
@@ -99,7 +99,7 @@ class SoSMap():
         :rtype:         ``re.Pattern``
         """
         if self.match_full_words_only:
-            item = rf'(?=\b|_|-){re.escape(item)}(?=\b|_|-)'
+            item = rf'(?<![a-z0-9]){re.escape(item)}(?=\b|_|-)'
         else:
             item = re.escape(item)
         return re.compile(item, re.I)
--- a/tests/cleaner_tests/full_report/full_report_run.py
+++ b/tests/cleaner_tests/full_report/full_report_run.py
@@ -64,7 +64,10 @@ class FullCleanTest(StageTwoReportTest):
         host = self.sysinfo['pre']['networking']['hostname']
         short = host.split('.')[0]
         # much faster to just use grep here
-        content = self.grep_for_content(host) + self.grep_for_content(short)
+        host = rf'(?<![a-z0-9]){re.escape(host)}(?=\b|_|-)'
+        short = rf'(?<![a-z0-9]){re.escape(short)}(?=\b|_|-)'
+        content = self.grep_for_content(host, regexp=True) + \
+            self.grep_for_content(short, regexp=True)
         if not content:
             assert True
         else:


sos/cleaner/mappings/hostname_map.py
```diff
@@ -87,7 +87,7 @@ def get_regex_result(self, item):
         """
         if '.' in item:
             item = item.replace('.', '(\\.|_)')
-        return re.compile(item, re.I)
+        return super(SoSHostnameMap, self).get_regex_result(item)
```
jcastill, line 90: If you want to return that way, perhaps add:

`# pylint: disable=super-with-arguments`

To the end of the line?

arif-ali: or better, and then we don't need to disable it :)

```python
return super().get_regex_result(item)
```

jcastill (approved):

arif-ali:
I think, that makes sense to me, is it worth updating the tests to reflect this at all?

pmoravec:
> I think, that makes sense to me, is it worth updating the tests to reflect this at all?

I came to the same conclusion - that `someHOST-something`  does *not* contain `HOST` as a *hostname*. It contains it as a substring, but not as hostname.

So:
- the original reproducer https://github.com/sosreport/sos/pull/3593#issuecomment-2045819623 shows a valid problem where hostname `port` should not be obfuscated from `sosreport-host0` string.
- but the current patch is wrong, as it does not obfuscate `cleanertest` hostname from `lvm_cleanertest_2015` string

So I will come up with a better patch.

pmoravec:
This is tricky :) Let see what tests will return, but `item = rf'(?=\b|_|-){re.escape(item)}(?=\b|_|-)'` is simply wrong. We test positive lookahead - twice! - but dont test lookbehind. That makes a lot of confusion where `my-hostname_TEST` finds `hostname`, but `my_hostname_TEST` doesnt find it.

OK, let try positive lookbehind `(?<=\b|_|-)` - but Python denies it as "look-behind requires fixed-width pattern".

Let re-formulate it to negative lookbehind :) `(?<![A-Za-z0-9])` . Since alphanumeric characters (*not* `\w`, that's another trick) are exactly those forbidden before a full word match.

So additionally to the patch, we must:

```
diff --git a/sos/cleaner/mappings/__init__.py b/sos/cleaner/mappings/__init__.py
--- a/sos/cleaner/mappings/__init__.py
+++ b/sos/cleaner/mappings/__init__.py
@@ -99,7 +99,7 @@ class SoSMap():
         :rtype:         ``re.Pattern``
         """
         if self.match_full_words_only:
-            item = rf'(?=\b|_|-){re.escape(item)}(?=\b|_|-)'
+            item = rf'(?!<[A-Za-z0-9]){re.escape(item)}(?=\b|_|-)'
         else:
             item = re.escape(item)
         return re.compile(item, re.I)
```

pmoravec:
FYI I made a test script similar to https://github.com/sosreport/sos/pull/3766#issuecomment-2370595072 : 

```
import re

secret = "SECRET"

REG = rf'(?<![A-Za-z0-9]){re.escape(secret)}(?=\b|_|-)'

test_strings = (
        (True, 'test SECRET:something'),
        (True, 'test:SECRET something'),
        (True, 'test my_SECRET_TEST will be cleaned?'),
        (True, 'test my-SECRET_TEST will be cleaned?'),
        (False, 'test mySECRET-should stay'),
        (True, 'test SECRET-cleaned'),
        (True, 'SECRET-cleaned also'),
        (False, 'SECRETNOT-cleaned'),
        (True, 'SECRET_cleaned underscore'),
        (True, 'SECRET.cleaned as well'),
        (True, 'cleaned.SECRET also'),
        (False, 'mySECRET.notcleaned hostname'),
)

intro=f"Matching against {REG}"
print(intro)
print("-"*len(intro))
print()

for match, line in test_strings:
    search = re.search(REG, line)
    print(f"{line}\t:\t\t{search}")
    if (None!=search) != match:
        print(f"\t\tERROR {line} should {'' if match else 'not '}match, but re.search={search}")
```

The script tests if given word (`SECRET`) is matched as a full word in several lines by proper R.E..

arif-ali:
I've put this through regex101 (always my fallback these days) too ;), and works as expected

pmoravec:
Once I fixed the stupid typo, it should pass tests now :)

pmoravec:
The failing test is a false alarm: `sosreport-host0-2024-11-27-pyqgxyc/sos_commands/systemd/systemctl_status_--all` file contains log:

```
Nov 27 12:49:10 cirrus-task-5243713155235840Unit display-manager.service could not be found.
```

and the test finds the hostname `cirrus-task-5243713155235840` there. But hostname is full word matching.

OK, next round of PR will fix the test :)

arif-ali (approved):

pmoravec:
Since I think I touch a fragile&core part of cleaner (how explicitly we search for matches), and the PR fixes three bugs altogether, I would prefer yet another ACK from either @jcastill or @TurboTurtle .

(three bugs: each commit fixes one)

TurboTurtle:
I'll give this a run through this Saturday.

jcastill:
@pmoravec   I ran some very basic tests and the commits seem to work pretty well. I'll see if I have some time this weekend to run something more complex

TurboTurtle (approved):
LGTM

PR #3852, opened 2024-11-26
[collect] Handle custom node and inherited configs with collections

TrevorBenson:
---


sos/collector/transports/__init__.py
```diff
@@ -352,6 +352,37 @@ def _get_hostname(self):
...
+                    return True
+                self.log_info(f"File copy attempt {attempts} failed")
+            self.log_info("File copy failed after 5 attempts")
+            return False
+        except Exception as err:
+            self.log_error("Exception encountered during config copy attempt "
+                           f"{attempts} for {fname}: {err}")
+            raise err
+
+    def _copy_file_to_remote(self, fname, dest):
```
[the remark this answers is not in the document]
jcastill: Ah nice, you fixed this already via 24cdb35

sos/collector/transports/control_persist.py
```diff
@@ -193,6 +193,12 @@ def remote_exec(self):
                             f"{self.opts.ssh_user}@{self.address}")
         return self.ssh_cmd
 
+    def _copy_file_to_remote(self, fname, dest):
+        cmd = (f"/usr/bin/scp -oControlPath={self.control_path} "
+               f"{fname} {self.opts.ssh_user}@{self.address}:{dest}")
+        res = sos_get_command_output(cmd)
```
jcastill, line 199: sos_get_command_output() uses TIMEOUT_DEFAULT, which is 5 minutes. Since you run this for 5 attempts, we may be waiting for a long time. Would it be worth to reduce the timeout in this call, or the number of attempts in copy_file_to_remote()?

jcastill (changes requested):

sos/collector/transports/__init__.py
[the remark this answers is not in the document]
TrevorBenson: Yep,opened the Draft mostly to get flake8/pylint and do the rest before marking ready to review. For some reason my local flake8 is refusing to show me the line length warnings, even when I set `--max-line-length` manually.

Apologies about the extra noise and CI tests.

sos/collector/transports/control_persist.py
TrevorBenson: I'm thinking 10 seconds on control persist, and 15 seconds on juju, oc and saltstack, just to handle any extra latency that might be introduced or delays in retrieving jobs from message queues. I'm not dead set on the values if you think more or less time would be better.

Thoughts?

sos/collector/transports/__init__.py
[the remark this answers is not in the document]
TrevorBenson: @jcastill If you have a juju or oc environment can can test the packit RPM that would be great. I'm not familiar with juju, but if I had to I could setup an multinode oc for testing, but probably not until next week at the soonest.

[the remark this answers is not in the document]
arif-ali: @TrevorBenson `juju` is a Canonical product, so I can test it there, but only works on the old version.

So, I may be removing `juju` soon once the LTS subsides, as the new version of `juju` doesn't comply with the way `sos` allows us to collect stuff with `sudo` and `su`.

sos/collector/transports/control_persist.py
jcastill: These numbers sound good to me, I think it will reduce the wait a lot in extreme cases

arif-ali:
I've restarted the 2 failed jobs, not sure why they are failing, and it's not the first time, I'll take that away as a separate issue

sos/collector/transports/control_persist.py
TrevorBenson: I went with the above listed timeouts, and 3 total attempts/retries. So control persist should only have 30 seconds max before failing to copy a config, and oc, juju and saltstack 45 seconds max.

sos/collector/__init__.py
```diff
@@ -463,6 +465,15 @@ def add_parser_options(cls, parser):
                                  choices=['auto', 'https', 'ftp', 'sftp',
                                           's3'],
                                  help="Manually specify the upload protocol")
+        collect_grp.add_argument('--inherit-config-file', type=str,
+                                 default=None,
+                                 help='Path to a local config file to'
+                                 ' copy to the remote node and use with'
+                                 ' sos report')
```
TurboTurtle, line 472: My initial thought was that this would be a toggle, whereby if `True` we implicitly take the config-file used locally, so `sos.conf` by default or whatever gets set.

Is there a use case you have where a user would have config_file_A locally that is used for the `collect` and local `report` collections, but would have config_file_B _also_ local to the system that is running collect but should be used only for remote node collections?

sos/collector/sosnode.py
```diff
@@ -762,12 +764,35 @@ def execute_sos_command(self):
...
+                config_file_arg = f'--config-file={self.opts.node_config_file}'
+            elif self.opts.inherit_config_file:
+                if not self.local:
+                    remote_config = os.path.join(self.tmpdir,
+                                                 'remote_sos.conf')
+                    self.run_command(f"mkdir -pv {self.tmpdir}",
+                                     timeout=10)
+                    self._transport.copy_file_to_remote(
+                        self.opts.inherit_config_file,
+                        remote_config)
```
TurboTurtle, line 778: I'm not a fan of creating directories on the remote node, but we do need a place to put them and we wouldn't have a tempdir at this point.

What about placing them in `/tmp/`? We then are closer to our "no changes will be made" (which, writing that out, perhaps we should talk about that within this context?) stance, and we don't introduce the potential for an exploit that causes us to write to arbitrary locations if someone gets crafty enough.

sos/collector/sosnode.py
```diff
@@ -762,12 +764,35 @@ def execute_sos_command(self):
...
             res = self.run_command(self.sos_cmd,
                                    timeout=self.opts.timeout,
                                    use_shell=True,
                                    need_root=True,
                                    use_container=True,
                                    env=self.sos_env_vars)
+            if self.opts.inherit_config_file and not self.local:
+                self.run_command(f"rm {remote_config} ; rmdir {self.tmpdir}",
+                                 use_shell=True,
+                                 need_root=True)
```
TurboTurtle, line 795: Yeah, same here. Very much not a fan of doing `rm`s like this - the local sos installation on the remote node should take care of any cleanup. If we place the inherited config files in `/tmp`, then they would be both out of the way and cleaned up separately by long-standing behavior.

sos/collector/__init__.py
TrevorBenson: I suspect any difference is mostly my misinterpretation of the discussion, not intentional changes to your thought on implementation.

> Is there a use case you have where [...]

Possibly, but not one of my requirements. I plan to use the same config for the collector and the local/remote nodes.

---

I'll attempt to layout the resulting behavior from this change.

### Overview

Three arguments

1. --config-file: str
2. --node-config-file: str
3. --inherit-config-file: bool

### Logic

* `--inherit-config-file` is `False`
  * `--config-file` & `--node-config-file` unset
    * Current behavior for both **collector** and **report**.
  * `--config-file` set to`/etc/sos/custom1.conf`
    * **collector** uses `custom1.conf`
    * **report** uses `sos.conf`
  * `--node-config` set to `/etc/sos/custom2.conf`
    * **collector** uses default `/etc/sos/sos.conf`
    * **report** uses (pre-existing) `/etc/sos/custom2.conf` for the local and remote nodes .
* `--inherit-config-file` is `True`
  * `--config-file` & `--node-config-file` unset
    * **collector** uses `/etc/sos/sos.conf`
    * `/etc/sos/sos.conf` is copied to all nodes in a temp path.
    * **report** uses `/var/tmp/sos.XXXXX/sos.conf` (local could use default path)
  * `--config-file` set to`/etc/sos/custom3.conf`
    * **collector** uses `/etc/sos/custom3.conf`
    * `/etc/sos/custom3.conf` is copied to all nodes in a temp path.
    * **report** uses `/var/tmp/sos.XXXXX/custom3.conf` (local ...)

`--node-config-file` & `--inherit-config-file are an incompatible argument pair. Potentially also `--config-file` & `--node-config-file`.

Does this correctly describe what you were thinking?

sos/collector/sosnode.py
TrevorBenson: Works for me. I wasn't really happy with the way I implemented the temp directory handling anyway, I just didnt see a simple path to leverage **TempFileUtil** class without drastically changing collector/report interaction.

I'll make this use a unique name in `/tmp/` instead.

sos/collector/sosnode.py
TrevorBenson: I'll presume you mean leaving the inherited conf in the tmp directory and letting `systemd-tmpfiles-clean.timer` RHEL, or `tmpfiles.d` for Ubuntu, handle the cleanup on schedule?

It works for my use case, as long as we are comfortable that the file remains might (although probably uncommon) store `plugin_options` in the file for things like `mysql.dbpass`, `postgresql.password`, `skydive.password`?

sos/collector/__init__.py
TrevorBenson: I've got this logic and additional comments worked out. I will add the new commit for review, but can be dropped/reverted if this wasn't what you were thinking of.

arif-ali:
I've started building some packages for testing [1], so will do the testing soon

[1] https://launchpad.net/~arif-ali/+archive/ubuntu/sos-pr-3852

jcastill:
@TrevorBenson did you test the oc transport already? I'm asking because I'm setting up an environment to test some other stuff and I can use it to test this PR as well.

arif-ali (approved):
tested both the of the new config options, and works as expected from the `juju` perspective. Thanks for your work on this

EDIT: don't forget to squash your commits, as ideally it is preferred to have one commit per PR

sos/collector/transports/__init__.py
```diff
@@ -352,6 +352,37 @@ def _get_hostname(self):
...
+        """
+        attempts = 0
+        try:
+            while attempts < 3:
+                attempts += 1
+                ret = self._copy_file_to_remote(fname, dest)
+                if ret:
+                    return True
+                self.log_info(f"File copy attempt {attempts} failed")
+            self.log_info("File copy failed after 5 attempts")
```
pmoravec, line 375: `while attempts < 3` vs. `failed after 5 attempts` :)

sos/collector/sosnode.py
```diff
@@ -762,6 +764,22 @@ def execute_sos_command(self):
...
+                    self._transport.copy_file_to_remote(
+                        self.opts.config_file,
+                        remote_config)
+                    config_file_arg = f'--config-file={remote_config}'
+                else:
+                    config_file_arg = (
+                        f'--config-file={self.opts.config_file}')
+            self.sos_cmd = (
+                f"{self.sos_cmd} {config_file_arg}"
+                if config_file_arg else self.sos_cmd)
```
pmoravec, line 782: Isn't it easier just to have

```
if config_file_arg:
    self.sos_cmd = f"{self.sos_cmd} {config_file_arg}"
```

? But I am ok with the current code.

pmoravec:
Just the 3 / 5 retry attempts are worth to fix, otherwise ACK.

sos/collector/transports/control_persist.py
```diff
@@ -193,6 +193,13 @@ def remote_exec(self):
                             f"{self.opts.ssh_user}@{self.address}")
         return self.ssh_cmd
 
+    def _copy_file_to_remote(self, fname, dest):
+        cmd = (f"/usr/bin/scp -oControlPath={self.control_path} "
+               f"{fname} {self.opts.ssh_user}@{self.address}:{dest}")
+        res = sos_get_command_output(cmd,
+                                     timeout=10)
```
TurboTurtle, line 200: Nit: this can fit on one line

sos/collector/transports/juju.py
```diff
@@ -72,6 +72,14 @@ def remote_exec(self):
         option = f"{model_option} {target_option}"
         return f"juju ssh {option}"
 
+    def _copy_file_to_remote(self, fname, dest):
+        model, unit = self.address.split(":")
+        model_option = f"-m {model}" if model else ""
+        cmd = f"juju scp {model_option} -- {fname} {unit}:{dest}"
+        res = sos_get_command_output(cmd,
+                                     timeout=15)
```
TurboTurtle, line 80: Nit: same here

sos/collector/transports/local.py
```diff
@@ -41,6 +41,9 @@ def _retrieve_file(self, fname, dest):
     def _format_cmd_for_exec(self, cmd):
         return cmd
 
+    def _copy_file_to_remote(self, fname, dest):
+        pass
```
TurboTurtle, line 45: Hmm. I was initially going to say this should return `True`, but that might not track given that we may not be writing the the same location (and in this first implementation we already aren't).

This might need some attention later one depending on how much use this functionality sees, but for now I think it's fine.

TurboTurtle (commented):
Looks good to me overall. Just two super minor nits, and Pavel's comment on the 3 vs 5  attempts. FWIW, I think 3 attempts is sufficient.

TrevorBenson:
> @TrevorBenson did you test the oc transport already? I'm asking because I'm setting up an environment to test some other stuff and I can use it to test this PR as well.

Nope. not yet. I'd need to reconfigure some things or get a spare system so would take me at least a week before I could do testing via `oc`.

sos/collector/transports/control_persist.py
TrevorBenson: I'll change it, don't recall why its split since it doesn't seem long enough for flake8 to have complained about it. I may have removed a different parameter.

This does bring up a question though, is 3.6 still the default python to target? I don't recall where, maybe looking at CI tests or some code in the actual sos files, but something last week looked to be using a minimum version of Python 3.8.

If were not minimum 3.8 yet, and it wasn't just in the CI, I will track down what it was an open an issue.

sos/collector/transports/local.py
TrevorBenson: Unsure if by same location you mean to the `/tmp/sos.XXXX` path where report does all its temporary work? 

I most often follow the Google Style guide, with typing and return syntax (i.e. `def _copy_file_to_remote(self, fname: str, dest: str) -> bool:`. Since that isn't part of sos style I just leave it out which made it easy to forget about the consistent function return.

FWIW for control persist testing I saw no problem when it was pass, but now that you point it out I'd probably opt for the consistent return among all transports. Unless you see an issue with making it `return True`?

TrevorBenson:
> tested both the of the new config options, and works as expected from the `juju` perspective. Thanks for your work on this
> 
> EDIT: don't forget to squash your commits, as ideally it is preferred to have one commit per PR

No problem as I tend to squash/fixup everything, but sometimes I've seen the force push removing a line hide comments which appeared to come from a review of a specific commit vs. the entire file.

Once all reviewers are happy I'll get it into the single commit and force push. 👍🏼

sos/collector/sosnode.py
TrevorBenson: That is actually what I had at first. Then the python static analyzer I use complained at me. Of course flake8 didnt like the line length after 😏 

I'm going to use the if syntax and just ignore the analyzer's suggestions since it ends up on multiple lines anyway. 👍🏼

sos/collector/transports/control_persist.py
pmoravec: > This does bring up a question though, is 3.6 still the default python to target? I don't recall where, maybe looking at CI tests or some code in the actual sos files, but something last week looked to be using a minimum version of Python 3.8.

Python3.6 is not further the minimal version, 3.8 it is. We already started to use e.g. walrus operator.

TrevorBenson:
> LGTM. Please just squash at least the 2 review commits.

I've squashed into a single commit. 

My initial tests were with control persist and all worked as expected. Once the checks are complete I will load from the updated copr repo for the saltstack transport test.

I believe @arif-ali will test, or already has tested, with juju. Then @jcastill will test with oc.

arif-ali:
> I believe @arif-ali will test, or already has tested, with juju & @jcastill will test with oc.

Yup, I already tested, and reported above from the `juju` perspective

TrevorBenson:
FYI I moved the arguments below `--become` and `--case-id`, before `--cluster-type`. They felt to easy to overlook when placed after the `--upload` options.

PR #3855, opened 2024-11-26
[component] Apply plugopts from cmdline gently

pmoravec:
When plugopts are set both in a loaded preset and in cmdline, we should merge the options instead of blindly using cmdline ones.

This means preset plugopt options not present in cmdline are added, while preset plugopt options also present in cmdline are ignored.

Closes: #3855

---

### DIFF
--- a/sos/component.py
+++ b/sos/component.py
@@ -243,7 +243,20 @@ class SoSComponent():
                         setattr(opts, oopt, [x for x in getattr(opts, oopt)
                                 if x not in common])
 
-            if val != opts.arg_defaults[opt]:
+            # plugin options as a list should be concatenated, not overriden
+            # BUT if cmdline plugoption overrides same option in opts, we must
+            # drop the opts's value; since the items are in form
+            # 'apache.log=on', we must separate the *name* of each option
+            if opt == 'plugopts':
+                oplugopts = getattr(opts, opt)
+                valnames = [v.split('=')[0] for v in val]
+                ovalnames = [v.split('=')[0] for v in oplugopts]
+                for common in set(valnames) & set(ovalnames):
+                    cstring = f"{common}="
+                    oplugopts = [oopt for oopt in oplugopts
+                                 if not oopt.startswith(cstring)]
+                setattr(opts, opt, list(set(val) | set(oplugopts)))
+            elif val != opts.arg_defaults[opt]:
                 setattr(opts, opt, val)
 
         return opts


pmoravec:
Reproducer:

```
sos report -k qpid.port=5672 --preset satellite -o qpid,apache -vvv
..
[sos.report:setup] using 'satellite' preset defaults (--log-size 100 --plugopts apache.log=on)
[sos.report:setup] effective options now: --log-size 100 --only-plugins qpid,apache --plugopts qpid.port=5672 --preset satellite -vvv
```

(see the `apache.log=on` option is ignored)

A simple solution "let merge the option lists" has a counterexample:

```
sos report -k qpid.port=5672,apache.log=off --preset satellite -o qpid,apache -vvv
..
[sos.report:setup] using 'satellite' preset defaults (--log-size 100 --plugopts apache.log=on)
[sos.report:setup] effective options now: --log-size 100 --only-plugins qpid,podman --plugopts qpid.port=5672,apache.log=off,apache.log=on --preset satellite -vvv
```

where we cant guarantee what value of `apache.log` will be really used. Hence the "separate value names" dance.

TurboTurtle:
Early comment to say, let's definitely get a test case written for this.

pmoravec:
> Early comment to say, let's definitely get a test case written for this.

I realized we have no test for presets. So I will add some generic plus some covering this PR.

Before I dig into it - can I extend the current https://github.com/sosreport/sos/blob/main/tests/report_tests/options_tests/options_tests.py#L13 ? By changing it like:

```
    files = [('options_tests_sos.conf', '/etc/sos/sos.conf')]   # sos.conf newly has plugin-timeout = 150 and cmd-timeout = 20
    sos_cmd = '--preset minimal --journal-size 20 --cmd-timeout 60'

    ..

    def test_effective_options_logged_correctly(self):
        for option in [
            "--case-id 8675309",
            "--only-plugins host,kernel",
            "--plugopts kernel.with-timer=on,kernel.trace=yes",
            "--preset minimal",
            "--cmd-timeout 60",   # cmdline beats config and preset
            "--journal-size 20",     # cmdline beats preset
            "--log-size 10",            # preset setting is honored
            "--plugin-timeout 30", # preset beats config file - this is current behaviour, is it correct..?
        ]:
            self.assertSosLogContains(f"effective options now: .* {option}")
```

This test will ensure the right priority of options among preset, config file and cmdline - see comments.

Can I "squeeze" the preset tests into this (renamed) class?

Additionally, I will have a class (or two) for testing the `plugopts` - that will work on Red Hat distro only (as plugopts are set in RH presets only and I dont want to make testing over-complex by creating a preset first and then testing it).

Does it sound a good plan?

arif-ali:
I do plan to add presets from Canonical at some point, so if we can generalise it, that would help when I get round to it. Or, otherwise I can extend the work here once I get to it

pmoravec:
> I do plan to add presets from Canonical at some point, so if we can generalise it, that would help when I get round to it. Or, otherwise I can extend the work here once I get to it

Sounds good. So I will test the plugopts stuff via Red Hat preset for now.

Extending the current `OptionsFromConfigTest` to test also presets sounds reasonable? I am ok with it, to have all/most of the options priority testing within one class and one `sos report` execution. But if you would like to have more granular classes (more readable, but slower tests execution), let me know.

pmoravec:
OK, tests added, both generic for presets and for this specific use case.

pmoravec:
OK, it is still failing: why on ubuntu this does not work? `/etc/sos/presets.d/plugopts_preset` with content:

```
{
    "plugopts_preset": {
        "desc": "test preset for plugin options",
        "args": {
            "plugopts": [
                "crio.timeout=600",
                "networking.namespaces=200",
                "networking.ethtool-namespaces=False",
                "networking.timeout=600"
            ]
        }
    }
}
```

but `sos report --preset plugopts_preset` fails with `Unknown preset: 'plugopts_preset'` error..?

TurboTurtle (approved):

arif-ali (approved):
lgtm

PR #3869, opened 2024-12-11
[kubernetes] add kube-system-log-files option, collect some kubelet files

champtar:
2 small enhancements to the kubernetes plugin

I'm open to proposal for another name than `kube-system-log-files`:
```
The following plugin options are available:
 kubernetes.all            on              collect all namespace output separately
 kubernetes.describe       on              collect describe output of all resources
 kubernetes.podlogs        on              capture stdout/stderr logs from pods
 kubernetes.podlogs-filter                 only collect logs from pods matching this pattern
 kubernetes.kube-system-log-files off             copy some kube-system pod logs without using the API
```

---

### DIFF
--- a/sos/report/plugins/kubernetes.py
+++ b/sos/report/plugins/kubernetes.py
@@ -46,6 +46,12 @@ class Kubernetes(Plugin):
     config_files = [
         "/etc/kubernetes",
         "/run/flannel",
+        "/var/lib/kubelet/config.yaml",
+        "/var/lib/kubelet/kubeadm-flags.env",
+        "/var/lib/kubelet/*_manager_state",
+    ]
+    forbidden_paths = [
+        "/etc/kubernetes/pki",
     ]
     resources = [
         'events',
@@ -105,7 +111,7 @@ class Kubernetes(Plugin):
     def setup(self):
         self.add_copy_spec(self.config_files)
 
-        self.add_forbidden_path("/etc/kubernetes/pki")
+        self.add_forbidden_path(self.forbidden_paths)
 
         self.add_env_var([
             'KUBECONFIG',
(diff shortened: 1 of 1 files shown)


pmoravec:
No secrets in newly collected data? (the env file: cant it contain some password?)?

Re the option name: I would prefer something shorter due to the alignment, what about `kube-system-logs` (still too long) or simply `kubelogs`?

I dont know kubernetes, so I evaluate the PR just from `sos` perspective.

champtar:
> No secrets in newly collected data? (the env file: cant it contain some password?)?

On a kubeadm install, both files have chmod 644, and the parents folders have 755, so hopefully no secrets (I also looked at the docs but could have missed something)

> Re the option name: I would prefer something shorter due to the alignment, what about `kube-system-logs` (still too long) or simply `kubelogs`?

will change to `kubelogs`, I can't find something short and self descriptive, but the long description is displayed right next to it

> I dont know kubernetes, so I evaluate the PR just from `sos` perspective.

TurboTurtle (commented):
Code looks fine, pending the PEP8 failure for line 100 (just needs to be shortened).

Before I ack however - is there a reason we shouldn't just be collecting these regularly? I.E. not gated behind an option and just a default collection by the plugin?

champtar:
> Code looks fine, pending the PEP8 failure for line 100 (just needs to be shortened).
> 
> Before I ack however - is there a reason we shouldn't just be collecting these regularly? I.E. not gated behind an option and just a default collection by the plugin?

Default kube log rotation is 10MB per file and keep 5 files, we collect 4 containers,
so that's 200MB uncompressed. Also it's redundant with podlogs option when API is working.

TurboTurtle (approved):

Issue #3875, opened 2024-12-12
NetworkManager keyfiles are not being gathered from all locations


palonsoro:
Currently, the `networkmanager` sos plugin collects all the NetworkManager connection keyfiles from `/etc/NetworkManager/system-connections` as we can see [here](https://github.com/sosreport/sos/blob/add55ea9ef6c8a753b8646546c2da5ec07bb54cc/sos/report/plugins/networkmanager.py#L24).

However, that's not the only location for NetworkManager connection keyfiles. As we can see in `man 5 nm-settings-keyfile`, the apart from `/etc/NetworkManager/system-connections/`, we should be including `/usr/lib/NetworkManager/system-connections/` and `/run/NetworkManager/system-connections/`.

The latter is specially important in case that temporary in-memory connections are defined, like the default auto-generated ones.

jcastill:
Thank you for opening the issue @palonsoro . I imagine that we have to mask passwords and secrets from these locations as well, since the file formats will be the same in all of them, right?

palonsoro:
> Thank you for opening the issue @palonsoro . I imagine that we have to mask passwords and secrets from these locations as well, since the file formats will be the same in all of them, right?

Correct. Same data masking must be done on those files too. It is the same format (in case you are curious, it is documented in `man 5 nm-settings-keyfile` and `man 5 nm-settings`).

PR #3876, opened 2024-12-12
[NetworkManager] Get system-connections files from other locations

jcastill:
system-connections files can be found not only under /etc/NetworkManager, but also /usr/lib/NetworkManager and /run/NetworkManager, so we need to capture and post process these locations as well.

Resolves: #3875

---

### DIFF
--- a/sos/report/plugins/networkmanager.py
+++ b/sos/report/plugins/networkmanager.py
@@ -6,7 +6,6 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-import os
 from sos.report.plugins import Plugin, RedHatPlugin, UbuntuPlugin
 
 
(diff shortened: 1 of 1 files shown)


sos/report/plugins/networkmanager.py
```diff
@@ -110,16 +117,19 @@ def test_nm_status(version=1):
             "nmcli dev show .*": "nmcli_dev_show_sos"
         })
 
-    def postproc(self):
-        for _, _, files in os.walk(
-                "/etc/NetworkManager/system-connections"):
+    def postproc_system_connections(self, sc_path):
+        for _, _, files in os.walk(sc_path):
```
champtar, line 121: Why not use `do_path_regex_sub()` ?
https://github.com/sosreport/sos/blob/279ea13ce25ef5b60e9b3a120d6a2add8e962f16/sos/report/plugins/kubernetes.py#L291-L292

arif-ali: good catch, I think that would make sense, that would cover all files in the directory

sos/report/plugins/networkmanager.py
```diff
@@ -110,16 +116,18 @@ def test_nm_status(version=1):
             "nmcli dev show .*": "nmcli_dev_show_sos"
         })
 
+    def postproc_system_connections(self, sc_path):
```
arif-ali, line 119: curious question, the fact that we're doing the one thing, do we need a new function?

jcastill: Probably now is not needed anymore. I'll fix this and resend.

arif-ali (approved):

PR #3877, opened 2024-12-12
[ceph_mon] Fix capture of ceph json outputs

jcastill:
Some ceph command captured with json output are
storing stderr at the end of the capture, and this can cause problems while parsing these outputs.
This commits disables stderr in add_cmd_output()
only for these captures that attempt to get json
outputs.

Related:RHEL-71127

---


jcastill:
In theory we capture the same output both in plain text and json output, so can we live with stderr only for plain text? Unless there's an error capturing the json itself, but I imagine that in that case we could investigate the actual output we got and move from there.

arif-ali (approved):

TurboTurtle (approved):

PR #3878, opened 2024-12-12
[foreman] Obfuscate http_proxy passwords

pmoravec:
production.log can (unintentionally) contain http_proxy URL including credentials. Since the URL can appear on one logline twice, obfuscating the 2nd instance is more tricky than usual.

Also, obfuscate the same password in settings table.

Related: SAT-30137
Closes: #3878

---


pmoravec:
I dont like the "apply similar-but-not-identical regexp twice" when we are trying to replace one or two instances of the same pattern on a line. But:

1) the replacement is applied just once (really we cant apply multiple times on one line..?) to the first instance
2) if I call `do_paths_http_sub` twice in a row in attempt to replace the 2nd password, the RE matches the obfuscated `******:******` :(

If you come up with some more elegant solution, I will appreciate.

TurboTurtle (approved):

PR #3879, opened 2024-12-13
[utilities,plugins] speedup journal collection (v2)

champtar:
Instead of generating all the logs and tailing the last 100M, we get the first 100M of 'journalctl --reverse' that we then reverse again using our own implementation of tac.

To handle multiline logs we would need to use "tac -brs '^[^ ]'" that takes ~30s on 100M of logs when plain 'tac' takes ~0.3s. Our simple implementation in python takes 0.7s, and avoid an extra dependency.

On journalctl timeout we now get the most recents logs.

During collection logs are now buffered on disk, so we use 2xsizelimit. While running our tac we could actually truncate the source file to limit disk usage. Previously buffering was in RAM (also 2xsizelimit).

On my test server, logs plugin runtime goes from 34s to 9.5s.

Fixes https://github.com/sosreport/sos/issues/3615

---

### DIFF
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -3088,8 +3088,15 @@ class Plugin():
         if output:
             journal_cmd += output_opt % output
 
+        fname = journal_cmd
+        tac = False
+        if log_size > 0:
+            journal_cmd = f"{journal_cmd} --reverse"
+            tac = True
+
         self._log_debug(f"collecting journal: {journal_cmd}")
         self._add_cmd_output(cmd=journal_cmd, timeout=timeout,
+                             tac=tac, to_file=True, suggest_filename=fname,
                              sizelimit=log_size, pred=pred, tags=tags,
                              priority=priority)
 
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -2032,6 +2032,7 @@ class Plugin():
             kwargs['changes'] = False
         if (not getattr(SoSCommand(**kwargs), "snap_cmd", False) and
            (self.get_option('all_logs') or kwargs['sizelimit'] == 0)):
+            kwargs['sizelimit'] = 0
             kwargs['to_file'] = True
         if "snap_cmd" in kwargs:
             kwargs.pop("snap_cmd")
@@ -2372,7 +2373,7 @@ class Plugin():
                             binary=False, sizelimit=None, subdir=None,
                             changes=False, foreground=False, tags=[],
                             priority=10, cmd_as_tag=False, to_file=False,
-                            container_cmd=False, runas=None):
+                            tac=False, container_cmd=False, runas=None):
         """Execute a command and save the output to a file for inclusion in the
         report.
 
@@ -2400,6 +2401,7 @@ class Plugin():
             :param cmd_as_tag:          Format command string to tag
             :param to_file:             Write output directly to file instead
                                         of saving in memory
+            :param tac:                 Reverse log lines order
             :param runas:               Run the `cmd` as the `runas` user
 
         :returns:       dict containing status, output, and filename in the
@@ -2451,7 +2453,7 @@ class Plugin():
             cmd, timeout=timeout, stderr=stderr, chroot=root,
             chdir=runat, env=_env, binary=binary, sizelimit=sizelimit,
             poller=self.check_timeout, foreground=foreground,
-            to_file=out_file, runas=runas
+            to_file=out_file, tac=tac, runas=runas
         )
 
         end = time()
@@ -2489,7 +2491,7 @@ class Plugin():
                     result = sos_get_command_output(
                         cmd, timeout=timeout, chroot=False, chdir=runat,
                         env=env, binary=binary, sizelimit=sizelimit,
-                        poller=self.check_timeout, to_file=out_file
+                        poller=self.check_timeout, to_file=out_file, tac=tac,
                     )
                     run_time = time() - start
             self._log_debug(f"could not run '{cmd}': command not found")
@@ -2508,6 +2510,9 @@ class Plugin():
                            "truncated")
             linkfn = outfn
             outfn = outfn.replace('sos_commands', 'sos_strings') + '.tailed'
+            if out_file:
+                dest = self.archive.check_path(outfn, P_FILE, force=True)
+                os.rename(out_file, dest)
 
         if not to_file:
             if binary:
--- a/sos/utilities.py
+++ b/sos/utilities.py
@@ -540,6 +540,43 @@ class FakeReader():
         return self.process.poll() is None
 
 
+class HeadReader(threading.Thread):
+    """Used to 'head' the command output (f_src) to a given size
+    without deadlocking sos. Takes a sizelimit value in MB.
+    """
+
+    COPY_BUFSIZE = 1024*1024
+
+    def __init__(self, f_src, f_dst, sizelimit, binary):
+        super().__init__()
+        self.f_src = f_src
+        self.f_dst = f_dst
+        self.remaining = sizelimit * 1048576  # convert to bytes
+        self.binary = binary
+        self.running = True
+        self.start()
+
+    def run(self):
+        """Reads from the f_src (Popen stdout pipe) until we reach sizelimit.
+        once done, close f_src to signal the program that we are done.
+        """
+        while self.remaining > 0:
+            buf = self.f_src.read(min(self.remaining, self.COPY_BUFSIZE))
+            if not buf:
+                break
+            self.f_dst.write(buf)
+            self.remaining -= len(buf)
+        self.f_src.close()
+        self.running = False
+
+    def get_contents(self):
+        return '' if not self.binary else b''
+
+    @property
+    def is_full(self):
+        return self.remaining <= 0
+
+
 class TailReader(threading.Thread):
     """Used to tail the command output to a given size without deadlocking
     sos.
--- a/sos/utilities.py
+++ b/sos/utilities.py
@@ -215,7 +215,7 @@ def is_executable(command, sysroot=None):
 def sos_get_command_output(command, timeout=TIMEOUT_DEFAULT, stderr=False,
                            chroot=None, chdir=None, env=None, foreground=False,
                            binary=False, sizelimit=None, poller=None,
-                           to_file=False, runas=None):
+                           to_file=False, tac=False, runas=None):
     # pylint: disable=too-many-locals,too-many-branches
     """Execute a command and return a dictionary of status and output,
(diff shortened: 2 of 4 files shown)


pmoravec:
I can confirm VERY evident speedup on a tested RHEL8 system with:

```
# journalctl --disk-usage 
Archived and active journals take up 560.0M in the file system.
#
```

where current upstream executes `python3 bin/sos report -o logs --batch --build` in 18s and this PR shrinks it to 3s.

Some test is failing - can you @champtar review it or do you need some help? I might get to the failure later today  /tomorrow.

sos/report/plugins/__init__.py
```diff
@@ -3083,10 +3084,18 @@ def add_journal(self, units=None, boot=None, since=None, until=None,
         if output:
             journal_cmd += output_opt % output
 
+        fname = journal_cmd
+        tac = False
+        if log_size > 0 and is_executable("head"):
+            journal_cmd = f"sh -c '{journal_cmd} --reverse | " \
+                "head -c {log_size*1024*1024}'"
```
pmoravec, line 3091: Missing leading `f` as you use f-string :)

champtar:
I'm off until the 27th so no rush for the review, and I was thinking about even implementing 'head' in Python

pmoravec:
```
tests/report_tests/plugin_tests/defaults.py:40:            if cmd['exec'] == 'journalctl --no-pager  --unit cups':
```

needs to be updated to reflect changed command exec `"sh -c 'journalctl --no-pager  --unit cups --reverse | head -c 104857600'"` - I vote for `'journalctl --no-pager  --unit cups' in cmd['exec']` test.


The `2-tests/report_tests/plugin_tests/logs.py:JournalSizeLimitTest.test_journal_tailed_and_linked` failure: The PR breaks part of `tailit` functionality: when a command output is bigger than `--log-size` (or `--journal-size` here), sosreport looks like:

```
# ll /var/tmp/sosreport-pmoravec-rhel9-8675309-2024-12-18-dusvggk/sos_commands/logs/
total 24
-rw-r--r--. 1 root root    63 Dec 18 15:50 journalctl_--disk-usage
lrwxrwxrwx. 1 root root    51 Dec 18 15:50 journalctl_--no-pager -> ../../sos_strings/logs/journalctl_--no-pager.tailed
lrwxrwxrwx. 1 root root    58 Dec 18 15:50 journalctl_--no-pager_--boot -> ../../sos_strings/logs/journalctl_--no-pager_--boot.tailed
-rw-r--r--. 1 root root    82 Dec 18 15:50 journalctl_--no-pager_--boot_-1
-rw-r--r--. 1 root root 13355 Dec 18 15:50 ls_-alRh_.var.log
# ll /var/tmp/sosreport-pmoravec-rhel9-8675309-2024-12-18-dusvggk/sos_strings/logs/
total 24824
-rw-r--r--. 1 root root 10484923 Dec 18 15:50 journalctl_--no-pager_--boot.tailed
-rw-r--r--. 1 root root 10484923 Dec 18 15:50 journalctl_--no-pager.tailed
-rw-r--r--. 1 root root  4445393 Dec 18 15:50 var.log.messages-20241201.tailed
#
```

And `manifest.json` tracks that via `"truncated": true,`.

With this PR, I see:

```
# ll /var/tmp/sosreport-pmoravec-rhel9-8675309-2024-12-18-zagarel/sos_commands/logs/
total 20464
-rw-r--r--. 1 root root       63 Dec 18 15:50 journalctl_--disk-usage
-rw-r--r--. 1 root root 10460043 Dec 18 15:50 journalctl_--no-pager
-rw-r--r--. 1 root root 10460043 Dec 18 15:50 journalctl_--no-pager_--boot
-rw-r--r--. 1 root root       82 Dec 18 15:50 journalctl_--no-pager_--boot_-1
-rw-r--r--. 1 root root    22730 Dec 18 15:50 ls_-alZR_.var.log
#
```

That suggests the `journalctl` outputs are not truncated, which is misleading.

pmoravec:
Hi @arif-ali and @TurboTurtle , is this approach worth to try? My answer is "carefully yes".

Hi @champtar , do you need some help with updating the tests?

champtar:
@pmoravec cirrus-ci is broken right now, but this should be ready for review

sos/utilities.py
```diff
@@ -275,8 +277,15 @@ def _check_poller(proc):
...
-        _output = open(to_file, 'w', encoding='utf-8')
+        if sizelimit:
+            # going to use HeadReader
+            _output = PIPE
+        elif tac:
+            # no limit but we need an intermediate file
+            _output = tempfile.TemporaryFile(dir=os.path.dirname(to_file))
+        else:
+            # pylint: disable=consider-using-with
+            _output = open(to_file, 'wb')
```
pmoravec, line 288: (now nitpick) missing original `encoding='utf-8'`?

sos/utilities.py
```diff
@@ -301,19 +320,25 @@ def _check_poller(proc):
...
                             'truncated': reader.is_full}
-            if to_file:
-                _output.close()
 
             # wait for Popen to set the returncode
             while p.poll() is None:
                 pass
 
+            if to_file:
+                if tac:
```
pmoravec, line 337: Also nitpick now: why not merging into single `if to_file and tac:`?

pmoravec:
Thanks for the new push. CI jobs restarted, seems good. I review the code now, will test functionality later (today or tomorrow, I assume).

sos/utilities.py
```diff
@@ -496,8 +562,45 @@ def running(self):
...
+            buf = self.f_src.read(min(self.remaining, self.COPY_BUFSIZE))
+            if not buf:
+                break
+            self.f_dst.write(buf)
+            self.remaining -= len(buf)
+        self.f_src.close()
+        self.running = False
+
+    def get_contents(self):
+        return '' if not self.binary else b''
```
pmoravec, line 595: note rather to myself: we return empty string here, despite the method is called when we return status 124, 126 or 127..? (I need to review behaviour in these use cases).

sos/utilities.py
champtar: Using binary mode (`wb`), encoding doesn't make sense

sos/utilities.py
champtar: There was something else in the `if to_file:` at some point (will fix on the next push)

champtar:
Not in the PR, but I was thinking about making TailReader work with to_file at some point, as I think we don't use the 'output' that often

pmoravec:
So far the PR survived various my corner-case tests I invented. I shall review the code tomorrow..

sos/utilities.py
pmoravec: Can't the binary mode negatively affect writing usual (textual) output of other commands..? It does not seem so, even encoding is preserved. But.. writing text command output using binary write is weird. While I understand with `tac=True` it is a must.

sos/utilities.py
```diff
@@ -285,10 +294,20 @@
...
-                reader = AsyncReader(p.stdout, sizelimit, binary)
+            if to_file:
+                if sizelimit:
+                    if tac:
+                        _output = tempfile.TemporaryFile(
+                            dir=os.path.dirname(to_file)
+                        )
+                    else:
+                        # pylint: disable=consider-using-with
+                        _output = open(to_file, 'wb')
```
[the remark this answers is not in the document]
pmoravec: Ditto - why binary mode for `tac=False`?

sos/utilities.py
```diff
@@ -325,11 +350,52 @@ def _check_poller(proc):
...
+
+def tac_logs(f_src, f_dst, drop_last_log=False):
+    """Python implementation of the tac utility with support
+    for multiline logs (starting with space). It is intended
+    to reverse the output of 'journalctl --reverse'.
+    """
+    NEWLINE_B = b'\n'
+    NEWLINE_I = 10
+    SPACE_I = 32
+    f_src.flush()
```
pmoravec, line 369: Nice one!

pmoravec:
All up to one nitpick is good. I confirm great performance gain (and no visible performance penalty for journals having 100MB where the new approach additionally has to `tac` the data). Several corner cases checked, no issue found. Code seems good.

I like the idea, but would welcome another profound review of others as well - since we touch a frequently used code.

sos/utilities.py
champtar: Fixed

sos/utilities.py
champtar: Added a comment to make it explicit why it's mandatory to flush here

sos/utilities.py
champtar: I would reverse the question, why was it ever text mode ?
If the command returns binary, you don't want to use text mode.
I'm not even sure if it matters, as the command outputs directly to the file, I'm not sure text or binary has an impact here

pmoravec: Outputs of commands collected by `sos` are usually of text content. So it makes sense to treat it as text and not binary.

Anyway I think the discussion here is rather theoretical - we are writing whole content directly to a file and such approach should provide same behaviour for text and binary formats (as far as I checked the text/bin differences).

pmoravec:
I rerun the failed test - it should succeed, I rather smell some infra issue (and logs dont show any reason).

sos/utilities.py
champtar: Just tested, even for binary content it doesn't change anything:
```
import subprocess
f = open('test', 'wt')
subprocess.run(['printf', '\\0'], stdout=f)
subprocess.run(['printf', '\r\n'], stdout=f)
subprocess.run(['printf', '\r'], stdout=f)
```

```
$ hexdump -C test
00000000  00 0d 0a 0d                                       |....|
00000004
```

Looking with strace, we just `openat`
```
openat(AT_FDCWD</home/champtar>, "test", O_WRONLY|O_CREAT|O_TRUNC|O_CLOEXEC, 0666) = 7</home/champtar/test>
```

And then `dup2`
```
dup2(7</home/champtar/test>, 1</dev/pts/0<char 136:0>>) = 1</home/champtar/test>
```
The text vs binary only has an effect if we use the file with python function (read/write)

> Outputs of commands collected by sos are usually of text content. So it makes sense to treat it as text and not binary.

This function do support binary, so it's safer to handle everything as binary by default, as text is just a subset of binary (even if for now here it doesn't make any difference)

champtar:
Thanks for the review @pmoravec !

pmoravec (approved):
Maybe it makes sense to squash some commits, but I am ok also with this granularity.

Looking forward for either @arif-ali or @TurboTurtle review.

champtar:
@arif-ali or @TurboTurtle friendly ping

TurboTurtle (approved):

PR #3886, opened 2024-12-16
[collector] Saltstack transport remote_exec handle stderr like other transports

### LINKED ISSUE #3860: [collect] Saltstack transport fails to collect when from nodes without flatpak installed
```
[sostest-2:run_command] Running command cat /etc/os-release
[sostest-1:run_command] Running command rpm -qa --queryformat "%{NAME}|%{VERSION}|%{RELEASE}\n"
[sostest-2:run_command] Running command rpm -qa --queryformat "%{NAME}|%{VERSION}|%{RELEASE}\n"
[sostest-1:run_command] Running command flatpak list --columns=name,version,branch
[sostest-2:run_command] Running command flatpak list --columns=name,version,branch
No nodes connected. Aborting...
Could not obtain installed package list
[root@sostest-1 ~]# flatpak list
-bash: flatpak: command not found
```
By installing flatpak on a single node, instead of `No nodes connected. Aborting...` 1 node ends up connected and the collection continues to execute.

Found while testing from packit copr repository in sos-4.8.1.

TrevorBenson:
---

Closes #3860

### DIFF
--- a/sos/collector/transports/saltstack.py
+++ b/sos/collector/transports/saltstack.py
@@ -125,7 +125,7 @@ class SaltStackMaster(RemoteTransport):
     def remote_exec(self):
         """The remote execution command to use for this transport."""
         salt_args = "--out json --static --no-color"
-        return f"salt {salt_args} {self.address} cmd.shell "
+        return f"salt {salt_args} {self.address} cmd.exec_code sh "
 
     def _copy_file_to_remote(self, fname, dest):
         """Copy a file to the remote host using SaltStack Master


TurboTurtle:
Not knowing saltstack myself, can you give some context as to the implications of this change? At a high level I'm assuming this means we're wrapping everything inside a `bash` invocation, but what's the significance of the `cmd.exec_code` bit?

TrevorBenson:
> Not knowing saltstack myself, can you give some context as to the implications of this change? At a high level I'm assuming this means we're wrapping everything inside a `bash` invocation, but what's the significance of the `cmd.exec_code` bit?

Short answer, it returns only the exit code 127 for command not found, but not the string.

---

The long answer:

The cmd.shell with either `/bin/sh`, or `/bin/bash when `use_shell=True`, returned the exit code, but includes the stderr in the output the master receives. This in result appears as if it was stdout.

Adding some debug I examined `pkg_list` and found it containing the stderr (which control persist did not):

```
[server-1:run_command] Shell requested, command is now /bin/bash -c 'flatpak list --columns=name,version,branch'
[FlatpakPackageManager._parse_pkg_list] pkg_list: /bin/bash: flatpak: command not found
```

Applicable lines: https://github.com/sosreport/sos/blob/b7cdc235f03bcae4b069a1843515f0619274b467/sos/policies/package_managers/flatpak.py#L23-L28


Changing the module to `cmd.exec_code bash` the exit code 127 is still reported, but stderr no longer shows up in the output returned to the master from each minion, and pkg_list is empty when flatpak is not installed.

TrevorBenson:
Looking back at `run_command` and the logic from `use_shell` this should probably instead be `cmd.exec_code sh`, which should also fix the issue. Then the behavior should be more consistent with prior functionality and I presume other transports.

https://github.com/sosreport/sos/blob/b7cdc235f03bcae4b069a1843515f0619274b467/sos/collector/transports/__init__.py#L230-L233

TurboTurtle (approved):
Thanks for the explanation! Makes sense to me.

PR #3889, opened 2024-12-19
[oracle] Add initial implementation of oracle plugin

dwolstroRH:
Related: RHEL-59870, TYCHE-732

---

### DIFF
--- a/sos/report/plugins/oracle.py
+++ b/sos/report/plugins/oracle.py
@@ -1,0 +1,62 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+import os
+from sos.report.plugins import Plugin, RedHatPlugin, PluginOpt
+
+
+class Oracle(Plugin, RedHatPlugin):
+    short_desc = 'Oracle on Linux'
+
+    plugin_name = "oracle"
+    profiles = ('services',)
+    packages = ('oracle',)
+
+    pw_warn_text = " (password visible in process listings)"
+
+    option_list = [
+        PluginOpt('oratab', default='/etc/oratab', val_type=str,
+                  desc='location of the oratab file')
+    ]
+
+    def setup(self):
+        ora_tab = self.get_option('oratab')
+
+        if os.path.isfile(ora_tab):
+            self._log_warn("Found file oratab")
+            try:
+                with open(ora_tab, 'r', encoding='UTF-8') as ofile:
+                    # Since the oratab file can have multiple unique SID
+                    # entries that point to the same oracle home directory,
+                    # keep track of the unique paths so we don't gather
+                    # duplicate files
+                    path_list = []
+                    for line in ofile.read().splitlines():
+                        if not line.startswith('#') and \
+                           not line.isspace() and \
+                           len(line) != 0:
+                            words = line.split(':')
+                            # check if this path has already been collected
+                            if not words[1] in path_list:
+                                path_list.append(words[1])
+                                oracle_path = words[1]
+                                if not oracle_path.endswith('/'):
+                                    oracle_path = f"{oracle_path}/"
+                                self.add_copy_spec([
+                                    f"{oracle_path}dbs/init*.ora",
+                                    f"{oracle_path}dbs/spfile*.ora"
+                                ])
+
+            except IOError as ex:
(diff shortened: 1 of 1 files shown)


sos/report/plugins/oracle.py
```diff
@@ -0,0 +1,62 @@
...
+    option_list = [
+        PluginOpt('oratab', default='/etc/oratab', val_type=str,
+                  desc='location of the oratab file')
+    ]
+
+    def setup(self):
+        ora_tab = self.get_option('oratab')
+
+        if os.path.isfile(ora_tab):
+            self._log_warn("Found file oratab")
```
TurboTurtle, line 31: Why should we warn on this?

sos/report/plugins/oracle.py
```diff
@@ -0,0 +1,62 @@
...
+                            # check if this path has already been collected
+                            if not words[1] in path_list:
+                                path_list.append(words[1])
+                                oracle_path = words[1]
+                                if not oracle_path.endswith('/'):
+                                    oracle_path = f"{oracle_path}/"
+                                self.add_copy_spec([
+                                    f"{oracle_path}dbs/init*.ora",
+                                    f"{oracle_path}dbs/spfile*.ora"
+                                ])
```
TurboTurtle, line 53: I think we can button this up a little bit, and not have to manually track unique entries:

```python
path_list = set()

with open(ora_tab, 'r', encoding='UTF-8') as ofile:
    for line in ofile.read().splitlines():
        if line.startswith('#') or not line.strip():
            continue
        path_list.add(line.split(':')[1])

dbfiles = [
    'dbs/init*.ora',
    'dbs/spfile*.ora'
]

self.add_copy_spec([
    self.path_join(path, dbfile) for dbfile in dbfiles for path in path_list
])
```

sos/report/plugins/oracle.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class Oracle(Plugin, RedHatPlugin):
+    short_desc = 'Oracle on Linux'
+
+    plugin_name = "oracle"
+    profiles = ('services',)
+    packages = ('oracle',)
+
+    pw_warn_text = " (password visible in process listings)"
```
TurboTurtle, line 20: We don't seem to use this anywhere?

sos/report/plugins/oracle.py
```diff
@@ -0,0 +1,62 @@
...
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+import os
+from sos.report.plugins import Plugin, RedHatPlugin, PluginOpt
+
+
+class Oracle(Plugin, RedHatPlugin):
```
TurboTurtle, line 13: Is there a more specific component name to use here perhaps? I think there may be some ambiguity with this and Oracle Linux as a distribution, or the Oracle DB offering, etc...

Since the plugin is centered around `oratab`, would that make sense as a name? I'll admit I'm not familiar with what the component/service surrounding oratab is, so maybe that doesn't fit.

TurboTurtle (changes requested):

sos/report/plugins/oracle.py
dwolstroRH: Removed

sos/report/plugins/oracle.py
dwolstroRH: Updated per your suggestion, thank you

sos/report/plugins/oracle.py
dwolstroRH: Removed

sos/report/plugins/oracle.py
dwolstroRH: Renamed to oratab for the initial release..at a later date we may want to include more Oracle DB specific data.

sos/report/plugins/oratab.py
```diff
@@ -0,0 +1,57 @@
...
+
+        if os.path.isfile(ora_tab):
+            try:
+                with open(ora_tab, 'r', encoding='UTF-8') as ofile:
+                    # Since the oratab file can have multiple unique SID
+                    # entries that point to the same oracle home directory,
+                    # don't gather duplicate files
+                    path_list = set()
+
+                    with open(ora_tab, 'r', encoding='UTF-8') as ofile:
```
arif-ali, line 36: why are we opening the file again, and assigning to `ofile` again?

arif-ali (changes requested):

arif-ali (commented):
I don't have any oratab to check against, so from a plugin perspective all good

TurboTurtle (approved):

PR #3891, opened 2024-12-23
[redhat|policy] Provide a URL message for S3 bucket based uploads

### LINKED ISSUE #3890: Upload URL set to RH_SFTP_HOST on Red Hat when using upload protocol s3 w/o case id.
I was testing out sos collect and upload protocol s3 for release 4.8.1.

On the very first test the collection was uploaded to the expected S3 bucket when using `--config-file` 

`/etc/sos/custom.conf` contents:
```/etc/sos/custom.conf
[global]
batch = yes
all-logs = no
log-size = 1

[collect]
encrypt = yes
encrypt-key = sosreport.example.com
upload-protocol = s3
upload-s3-endpoint = https://sosreport.example.com
upload-s3-bucket = projectb-sosreports
upload-s3-access-key = <OBFUSCATED_ACCESS_KEY>
upload-s3-secret-key = <OBFUSCATED_SECRET_KEY>
```

I didn't include a case ID, the archive was uploaded to the expected prefix under a UUID, year and month (defined at the time of upload): `a75d5bb0-fd17-4422-8b1b-497310c74efd/2024/12/sos-collector-2024-12-20-yosca.tar.xz.gpg`

However, I noticed that the following messages were being provided at the end, even if the archive was being uploaded to an alternative protocol.

```
No case id provided, uploading to SFTP
No case id provided, uploading to SFTP
```

I don't believe that anything is being uploaded to SFTP as I have each collection within the defined bucket, but the RHELPolicy isn't considering this in its version of `get_upload_url`:

https://github.com/sosreport/sos/blob/a642018f3c733348b682e109fef09de37ed81e83/sos/policies/distros/redhat.py#L250-L261

TrevorBenson:
---

Closes #3890


jcastill (approved):
LGTM. 
Did a very quick test with non-S3 uploads and worked as expected. I'll test more thoroughly in a couple of days

sos/policies/distros/redhat.py
```diff
@@ -254,6 +254,14 @@ def get_upload_url(self):
             return self.commons['cmdlineopts'].upload_url
         if self.commons['cmdlineopts'].upload_protocol == 'sftp':
             return RH_SFTP_HOST
+        if self.commons['cmdlineopts'].upload_protocol == 's3':
+            endpoint = self.get_upload_s3_endpoint()
+            bucket = self.get_upload_s3_bucket()
+            if self.commons['cmdlineopts'].case_id:
+                rh_case_api = "/support/v1/cases/%s/attachments"
+                return f"{endpoint}/{bucket}" + rh_case_api % self.case_id
```
TurboTurtle, line 262: Let's make this all f-string formatting, mixing the two syntax is a little awkward.

TurboTurtle (commented):
The code LGTM, but I wonder if we wouldn't be better served if we instead change the logic that was generating the original problematic messaging in #3890. It would strike me as a little strange for the RH api path the be appended on to whatever endpoint/bucket path I'd otherwise be expecting.

TrevorBenson:
> The code LGTM, but I wonder if we wouldn't be better served if we instead change the logic that was generating the original problematic messaging in #3890.

Unless I overlooked something this function is what ends up generating both parts of the message: the missing case id; the upload to Red Hat Secure FTP.

> It would strike me as a little strange for the RH api path the be appended on to whatever endpoint/bucket path I'd otherwise be expecting.


Only this I think @jcastill mentioned RH potentially using an S3 bucket in the future. I presume he may be adding an RH_S3_HOST (aka an S3 Endpoint) in the work he is doing separately.

I want to be sure I understand your request. When both a case id and a bucket prefix are provided the resulting sos-collector tar archive will contain the case id, but the archive itself is only found inside the provided s3 object prefix? Or are you suggesting we drop the rh_case_api altogether and if the user provides a case id without providing an object prefix the tar file, with a case ID in its name, ends up in the root of the bucket?

Thanks in advance for any clarification.

sos/policies/distros/redhat.py
TrevorBenson: I'll adjust this accordingly once I have additional feedback to https://github.com/sosreport/sos/pull/3891#issuecomment-2561360210

TurboTurtle:
Let me answer this way, please let me know if I'm not making sense here.

As an end-user creating an sos report and uploading it to an S3 location that I've defined in my config file or in the `sos` command I'm running, I would expect:

- No message about SFTP, regardless of it a case id was provided or not
- No message about a Red Hat location
-  The sos report to end up in the endpoint/bucket I defined, and not nested under a path I did not specify.

On that last point that means I'd expect, with or without a case id, this path from #3890:

` s3://project-sosreports/ba59c727-bfd2-4808-8f5f-f62c63903c13/2024/12` or, potentially, `s3://project-sosreports/ba59c727-bfd2-4808-8f5f-f62c63903c13/2024/12/$case_id`

and **not**

`s3://project-sosreports/support/v1/cases/12345/attachments`


If RH starts using an S3 bucket, we'd address any handling of that within the Red Hat upload methods so that it doesn't impact a user uploading to a non-RH S3 location in any way. We really need to separate upload from policy, and I'll try and spend some time on that this week and next - but for the purposes of this PR I'm mainly just concerned about the above - predictable and expected locations within a user-specified S3 destination.

TrevorBenson:
Makes sense and clarifies the case of no prefix with a case id, which will result in the path: `s3://project-sosreports/$case_id`.

I'll make the adjustments and force push a new commit for review.

TrevorBenson:
I thought providing an `https://` based upload_url that includes the endpoint would be more informative. On further review I see why I used the `s3://` url instead:

https://github.com/sosreport/sos/blob/ace5715fb34dbfb057da04241203a74fdc42c3ed/sos/policies/distros/__init__.py#L547-L567

---

The upload_s3 method doesn't rely on upload_url currently, so get_upload_url returning a string w/ case_id appended shows the Attempting upload to correctly, but its dropped by the upload process.

While appending the case_id to the object prefix seemed a simple fix to this I found on a freshly installed rhel 8.9 system **get_upload_url()** method of **RHELPolicy** is called multiple times, or recursively. This lead to the case_id being appended more than once when appending it directly to the s3 object prefix instead of the upload_url. 

```
[RHELPolicy.get_upload_url] called
[RHELPolicy.get_upload_url] called
[RHELPolicy.get_upload_url] called
[RHELPolicy.get_upload_url] called
Attempting upload to https://sosreport.example.com/project-sosreports/ba59c727-bfd2-4808-8f5f-f62c63903c13/2024/12/12345/12345
```


I'm going to convert this PR to a draft while determining how best to implement the requests and if aligning s3 more closely to the other upload protocols logic is possible.

TrevorBenson:
@TurboTurtle What do you think of the most recent commit? The thought is to fix the issue, clarify which endpoint gets in the string output, but to leave off the case id as a prefix (directory) for now. I figure this can be handled if/when Red Hat decides to implement an s3 bucket upload target.

PR #3893, opened 2025-01-02
[aap_containerized] Fixes #3892 - added new AAP Containerized plugin

### LINKED ISSUE #3892: Introduce a new plugin to capture AAP Containerized logs
Now that we support AAP Containerized installation, there should be a plugin to capture these logs.

snagoor:
---

### DIFF
--- a/sos/report/plugins/aap_containerized.py
+++ b/sos/report/plugins/aap_containerized.py
@@ -1,0 +1,123 @@
+# Copyright (c) 2025 Nagoor Shaik <nshaik@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+import os
+from sos.report.plugins import Plugin, RedHatPlugin, PluginOpt
+
+
+class AAPContainerized(Plugin, RedHatPlugin):
+    """Collects details about AAP Containerized setup
+    under a user's home directory"""
+
+    short_desc = "AAP Containerized Setup"
+    plugin_name = "aap_containerized"
+    profiles = ("sysmgmt", "ansible",)
+    packages = ("podman",)
+
+    option_list = [
+        PluginOpt(
+            "username",
+            default="",
+            val_type=str,
+            desc="Username that was used to setup "
+            "AAP containerized installation"
+        ),
+        PluginOpt(
+            "directory",
+            default="",
+            val_type=str,
+            desc="Absolute path to AAP containers volume directory. "
+            "Defaults to 'aap' under provided user's home directory"
+        )
+    ]
+
+    def setup(self):
+        # Check if username is passed as argument
+        username = self.get_option("username")
+        if not username:
+            self._log_error("Username is mandatory to collect "
+                            "AAP containerized setup logs")
+            return
+
+        # Grab aap installation directory under user's home
+        if not self.get_option("directory"):
+            user_home_directory = os.path.expanduser(f"~{username}")
+            aap_directory_name = self.path_join(user_home_directory, "aap")
+        else:
+            aap_directory_name = self.get_option("directory")
+
+        # Don't collect cert and key files from the installation directory
+        if self.path_exists(aap_directory_name):
+            forbidden_paths = [
+                self.path_join(aap_directory_name, path)
+                for path in [
+                    "containers",
+                    "tls",
+                    "controller/etc/*.cert",
+                    "controller/etc/*.key",
+                    "eda/etc/*.cert",
+                    "eda/etc/*.key",
+                    "gateway/etc/*.cert",
+                    "gateway/etc/*.key",
+                    "hub/etc/*.cert",
+                    "hub/etc/*.key",
+                    "hub/etc/keys/*.pem",
+                    "postgresql/*.crt",
+                    "postgresql/*.key",
+                    "receptor/etc/*.crt",
+                    "receptor/etc/*.key",
+                    "receptor/etc/*.pem",
+                    "redis/*.crt",
+                    "redis/*.key",
+                ]
+            ]
+            self.add_forbidden_path(forbidden_paths)
+            self.add_copy_spec(aap_directory_name)
+        else:
+            self._log_error(f"Directory {aap_directory_name} does not exist "
+                            "or invalid absolute path provided")
+
+        # Gather output of following podman commands as user
+        podman_commands = [
+            (f"su - {username} -c 'podman info --debug'", "podman_info"),
+            (f"su - {username} -c 'podman ps -a --format json'",
+                "podman_ps_all_json"),
+        ]
+
+        for command, filename in podman_commands:
+            self.add_cmd_output(command, suggest_filename=filename)
+
+        # Collect AAP container names
+        aap_containers = self._get_aap_container_names(username)
+
+        # Copy podman container log files in plugin sub directory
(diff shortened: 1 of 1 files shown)


sos/report/plugins/aap_containerized.py
```diff
@@ -0,0 +1,117 @@
...
+                ]
+            ]
+            self.add_forbidden_path(forbidden_paths)
+            self.add_copy_spec(aap_directory_name)
+        else:
+            self._log_debug(f"Directory {aap_directory_name} does not exist")
+
+        # Gather output of following podman commands as user
+        podman_commands = [
+            (f"sudo -u {username} -i podman info --debug", "podman_info"),
```
jcastill, line 78: We have a 'runas=user' option now for add_cmd_output() that may be better than calling sudo here. 
Check the sunbeam plugin for examples of how to use it, so we don't have to call 'sudo' here.

sos/report/plugins/aap_containerized.py
```diff
@@ -0,0 +1,117 @@
...
+
+class AAPContainerizedPlugin(Plugin, RedHatPlugin):
+    """Collects details about AAP Containerized setup
+    under a user's home directory"""
+
+    short_desc = "AAP Containerized Setup Plugin"
+    plugin_name = "aap_containerized"
+    profiles = ("sysmgmt", "ansible",)
+    packages = ("podman",)
+
```
jcastill, line 24: Any container pattern or regex that we could watch for via "containers =" and could help with activation of this plugin?

sos/report/plugins/aap_containerized.py
snagoor: @jcastill thanks for the review, I initially tried using the `runas` argument. However, its failing with permission denied errors as below. So, I had to rely on the literal commands, I can switch to `su` instead of `sudo` as used in the sunbeam plugin https://github.com/sosreport/sos/blob/main/sos/report/plugins/sunbeam.py#L84

~~~
[root@host tmp]# cat sosreport-host-2025-01-02-zovejyg/sos_commands/aap_containerized/podman_ps_all_json
time="2025-01-02T13:57:37+05:30" level=error msg="mkdir /run/user/0/libpod: permission denied"

[root@host tmp]# cat sosreport-host-2025-01-02-zovejyg/sos_commands/aap_containerized/podman_info
time="2025-01-02T13:57:36+05:30" level=error msg="mkdir /run/user/0/libpod: permission denied"
~~~

jcastill: @snagoor I got a very similar issue while working on the instructlab plugin, and had to drop runas there as well until I can investigate what's going on (which I'll do next week). 
I think with this issue you could continue with sudo until we find the problem with runas, then move on to using it in the future.

sos/report/plugins/aap_containerized.py
snagoor: All of the AAP containers are launched under a specific non-root user that end-user wish. There is no certain pattern to it. However, for now I can list down all the containers but that would need to be maintained manually in case something changes in future.

~~~
# su - nshaik -c "podman ps --format {{.Names}}"
postgresql
redis-unix
redis-tcp
receptor
automation-controller-rsyslog
automation-controller-task
automation-controller-web
automation-eda-api
automation-eda-daphne
automation-eda-web
automation-eda-worker-1
automation-eda-worker-2
automation-eda-activation-worker-1
automation-eda-activation-worker-2
automation-eda-scheduler
automation-hub-api
automation-hub-content
automation-hub-web
automation-hub-worker-1
automation-hub-worker-2
automation-gateway-proxy
automation-gateway
~~~

TurboTurtle:
Before I go through with a full review, why does this need to be a separate plugin? We typically support containerized and non-containerized within the same plugin for the same component(s), and just dynamically provide a container to run command or file collections in when needed via the `container` parameter that is part of both `add_copy_spec()` and `add_cmd_output()`.


We do this so that we don't duplicate efforts, when the only difference between collections is if they need to be executed on the host or in a container. If the actual functioning of a component/whatever actually changes when inside a container, then that makes sense for a separate plugin.


We have several different AAP plugins today - so is there a reason we can't extend those to look for their relevant containers? It would seem awkward to me at first glance if we were to either 1) double up on the number of plugins just to account for containers or 2) had multiple plugins for host-based installs and a single "unified" plugin for containerized installs which this seems like it is trying to be.

sos/report/plugins/aap_containerized.py
```diff
@@ -0,0 +1,117 @@
...
+                    subdir="aap_containers_log"
+            )
+
+    def _get_aap_container_names(self, username):
+        try:
+            containers_output = subprocess.check_output(
+                ["su", "-", username, "-c",
+                 "podman ps -a --format {{.Names}}"],
+                text=True
+            )
```
TurboTurtle, line 105: Please use `Plugin.exec_cmd()` here instead of subprocess, as `exec_cmd()` will inherit our standard sos handling. It's used precisely for "I need output, but don't need to collect that output" like here.

TurboTurtle (commented):
Code LGTM aside from the one note around subprocess.

Still have the outstanding question of this single plugin vs. integrating containerized bits in the existing ones, but if we feel that a new single plugin makes the most sense then I'm fine with it.

sos/report/plugins/aap_containerized.py
snagoor: done

snagoor:
> Code LGTM aside from the one note around subprocess.
> 
> Still have the outstanding question of this single plugin vs. integrating containerized bits in the existing ones, but if we feel that a new single plugin makes the most sense then I'm fine with it.

@TurboTurtle thank you for the review.

There is a significant difference in terms of AAP containerized and RPM based architecture. That is why we have 2 different installers for the same. IMO this can't be clubbed into the existing plugin and increase its complexity in maintaining it in future, because the storage, logs and configs are totally different. 

Also, this Containerized setup is going to be future. So, having this plugin by itself is what I vouch for. 

I have now updated the changes as request, please have a look when you can.

TurboTurtle (approved):
LGTM, just the linting alert from flake8 on line 50.

Thanks for the details, makes sense to me to split this out going forward.

sos/report/plugins/aap_containerized.py
```diff
@@ -0,0 +1,123 @@
...
+    def setup(self):
+        # Check if username is passed as argument
+        username = self.get_option("username")
+        if not username:
+            self._log_error("Username is mandatory to collect "
+                            "AAP containerized setup logs")
+            return
+
+        # Grab aap installation directory under user's home
+        if not (self.get_option("directory")):
```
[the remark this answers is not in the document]
snagoor: updated and pushed new changes.

sos/report/plugins/aap_containerized.py
```diff
@@ -0,0 +1,123 @@
...
+        except Exception:
+            self._log_error("Error retrieving Podman containers")
+            return []
+
+    # Check and enable plugin on a AAP Containerized setup
+    def check_enabled(self):
+        ps = self.exec_cmd("ps aux --noheaders")
+        if "awx-manage" in ps["output"] and "aap-gateway" in ps["output"]:
+            return True
+        return False
```
pmoravec, line 123: Can't this be simple:

           return "awx-manage" in ps["output"] and "aap-gateway" in ps["output"]

?

Can't this provide false positive results when there would be a running process like, say, "grep awx-manage some.log"? Isn't `ps --noheaders axco command` and matching whole line a better check here, or is that already over-complicated..?

But I am OK with the current code.

snagoor: @pmoravec Thank you for the suggestion. Changes amended as requested.

pmoravec (approved):
Thanks!

PR #3907, opened 2025-01-16
[testing] Add debian to standard testing

arif-ali:
* Add new debian_only to replace current ubuntu_only.
* ubuntu_only now only tests true for only Ubuntu distros.
* Add extra packages to install for Debian ditsros, as these are not installed by default.

As per discussions in the maintainer call, wanted to start testing more on Debian, and this is a first start on testing debian regularly

---

### DIFF
--- a/tests/report_tests/plugin_tests/defaults.py
+++ b/tests/report_tests/plugin_tests/defaults.py
@@ -16,8 +16,11 @@ class DefaultCollectionsTest(StageTwoReportTest):
     :avocado: tags=stagetwo
     """
 
-    packages = {'rhel': 'cups',
-                'Ubuntu': 'cups'}
+    packages = {
+        'rhel': 'cups',
+        'Ubuntu': 'cups',
+        'debian': 'cups',
+    }
 
     sos_cmd = '-o cups'
 
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -6,7 +6,8 @@ env:
     FEDORA_NAME: "fedora-38"
     FEDORA_PRIOR_NAME: "fedora-37"
 
-    DEBIAN_NAME: "debian-11"
+    DEBIAN_12_NAME: "debian-12"
+    DEBIAN_11_NAME: "debian-11"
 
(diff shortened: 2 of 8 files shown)


arif-ali:
I'm trying to figure out what is wrong here. Tried to bring up the shell on Cirrus, and run the same command as the config, and all works. Maybe I am missing something, and I can't see it clearly and stared at it for a bit. A second head might see the issue if anyone wants to look at the PR

tests/report_tests/plugin_tests/defaults.py
```diff
@@ -16,8 +16,11 @@ class DefaultCollectionsTest(StageTwoReportTest):
     :avocado: tags=stagetwo
     """
 
-    packages = {'rhel': 'cups',
-                'Ubuntu': 'cups'}
+    packages = {
+        'rhel': 'cups',
+        'Ubuntu': 'cups',
```
pmoravec, line 21: I see sometimes `packages['Ubuntu']` and sometimes `packages['ubuntu']` declared. While distro checking is case sensitive, I think..?

This PR does not change the case sensitivity on either place, but maybe it reveals the problem behind it..?

arif-ali: Yeah, the [Report Stage One - ubuntu-24.10 - deb](https://github.com/sosreport/sos/pull/3907/checks?check_run_id=35723768189) is not even getting to that point, it's as if it's not even running the if statement or is coming out as false, and hence not completing the setup

pmoravec: Yeah, I dont understand the error. How it could ever attempt to run `apt -y install sosreport` if that package/file name (for `apt`) is never mentioned..?

TurboTurtle (approved):
LGTM, and the CI is passing. Was there something about this we wanted or needed to hold back on merge for 4.9, or are we good to bring it in?

arif-ali:
> LGTM, and the CI is passing. Was there something about this we wanted or needed to hold back on merge for 4.9, or are we good to bring it in?

no, not really, I was waiting for it to be reviewed by the team

PR #3910, opened 2025-01-22
[openshift] Capture ls long listing for /mnt

jmagrini:
This plugin is used to collect Openshift LSO details. This expands the 
ceph_osd plugin since storage nodes by default are not setup for 
ceph access. When gathering data from an OpenShift node when 
LSO is in use, we currently do not collect the symlink data location 
for LSO, which is always under /mnt. Many times this directory has 
pointers/symlinks to paths instead of devices.

---

- [ x] Are any related Issues or existing PRs [properly referenced](https://docs.github.com/en/issues/tracking-your-work-with-issues/creating-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) via a Closes (Issue) or Resolved (PR) line?

### DIFF
--- a/sos/report/plugins/openshift_lso.py
+++ b/sos/report/plugins/openshift_lso.py
@@ -1,0 +1,41 @@
+# openshift_lso.py
+# Copyright (C) 2007-2025 Red Hat, Inc., Jon Magrini <jmagrini@redhat.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+import os
+from sos.report.plugins import (Plugin, RedHatPlugin)
+
+class OpenshiftLSO(Plugin, RedHatPlugin):
+    """
+    This plugin is used to collect Openshift LSO details. This expands 
+    the ceph_osd plugin since storage nodes by default are not setup for
+    ceph access. When gathering data from an OpenShift node when LSO is in use, 
+    we currently do not collect the symlink data location for LSO, 
+    which is always under /mnt (basically two layers lower, 
+    but the names can change). This is useful in determining if LSO 
+    is setup correctly when also having a must-gather output.  
+    Many times the LSO directory has pointers to paths instead of devices. 
+    This can cause issues with lost access to the OCP backend storage 
+    used by OpenShift.
+    """
+
+    short_desc = 'Openshift LSO'
+    
+    plugin_name = "openshift_lso"
+    profiles = ('storage', 'openshift', 'ceph')    
+    # Each node runs the ceph-osd daemon, 
+    # which interacts with logical disks attached to the node.
+    files = ('/run/ceph/**/ceph-osd*')
+    
+    def setup(self): 
+        self.add_cmd_output([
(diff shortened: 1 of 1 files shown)


sos/report/plugins/openshift_lso.py
```diff
@@ -0,0 +1,39 @@
+# openshift_lso.py
```
TurboTurtle, line 1: Drop this top header line, please

sos/report/plugins/openshift_lso.py
```diff
@@ -0,0 +1,39 @@
...
+    plugin_name = "openshift_lso"
+    profiles = ('storage', 'openshift', 'ceph')
+    # Each node runs the ceph-osd daemon,
+    # which interacts with logical disks attached to the node.
+    files = '/run/ceph/**/ceph-osd*'
+    
+    def setup(self):
+        self.add_cmd_output([
+            'ls -lanR /mnt'
+        ])
```
TurboTurtle, line 39: We have a standardized wrapper for dir listings so that stuff those types of collections are consistent across plugins - `add_dir_listing(path)`.

TurboTurtle (commented):
Couple notes. I'm honestly surprised to find out we don't capture this listing already. I'm ok with the new plugin but I'm also wondering if it'd be useful to collect this by default for everyone and move it into the `filesys` plugin instead - thoughts @pmoravec @arif-ali ?

arif-ali:
> Couple notes. I'm honestly surprised to find out we don't capture this listing already. I'm ok with the new plugin but I'm also wondering if it'd be useful to collect this by default for everyone and move it into the `filesys` plugin instead - thoughts @pmoravec @arif-ali ?

I'm not keen to collect this by default, it's like collecting directory listing of `/`. You could have users actually mounting filesystem in `/mnt/<mount point>`, so if we automatically collect directory of `/mnt` for every sos, that we'll see the listing of all files. Now, if this is a big filesystem with lots of content, that could take a very long time to collect that data too.

As an example, I already do this on my laptop, it took 13 seconds and the file would have 863198 lines. 

So from my perspective is a -1 to include it by default in `filesys` plugin, unless there is a good case to collect this info

arif-ali:
@jmagrini please check the pylint and flake8 tests, and resolve those 2, when you resolve the above issues that have been mentioned.

So, this plugin will run whenever there is ceph-osd on a RedHat node, even if there is no OpenShift installed it will still give you the listing for `/mnt`, Personally, I don't get it, but maybe I don't know Openshift enough

pmoravec (approved):

pmoravec:
(approval up to the failing tests, ofc)

pmoravec (changes requested):
Please merge the four commits into one.

Ensure the final commit has DCO (`Signed-off-by: Jon Magrini <jmagrini@redhat.com>` in commit message) - older 2 commits lack this hence failing DCO test.

Fix the `flake8`/`pylint` formatting issues:

```
sos/report/plugins/openshift_lso.py:13:1: E302 expected 2 blank lines, found 1
sos/report/plugins/openshift_lso.py:26:1: W293 blank line contains whitespace
sos/report/plugins/openshift_lso.py:28:1: W293 blank line contains whitespace
sos/report/plugins/openshift_lso.py:34:1: W293 blank line contains whitespace
```

arif-ali:
@jmagrini any update on this PR at all, doesn't seem like much update is required to get this in

Issue #3374, opened 2023-10-06
[Feedback requested] Dropping redirectors and updating text output for sosreport


TurboTurtle:
It's been a few years since we released sos-4.0 and changed the binary to `sos`, and along with it clarified that the tool is `sos`, of which `report` is a function.

I'd like to propose dropping the `sosreport` and `sos-collector` redirectors in an upcoming release. I'd ideally say 4.7, but that's probably not feasible since we're in 4.6 currently and most downstreams, I'd assume, want a little more time to coordinate the change.

So how do we feel about 4.7 or 4.8 for no longer carrying these wrappers?

Secondly, along with this change I'd like to update the text output from `sosreport` to `sos (space) report` globally. We've got a most of the project using this for a bit now, but the former still in use in the initial banner text, as well as the most-likely-critical "Your sosreport has been generated" message. Note I am **_not_** proposing that the tarball name change in any way. 

I imagine there is a decent amount of automation looking for that string (in fact, we ourselves look for it during `sos collect` runs), so this is a change that needs to be well communicated.

I'd like both of these changes to go out at the same time, so if we need to wait until 4.8 for our downstreams to prepare for it, then so be it.

pmoravec:
Cc @jcastill and @mhradile at least.

AFAIK original  plans were to keep the shortcut/redirectors for the `sos-4.*` timeframe. While the planned change itself is sound, my gut feeling is users are not yet prepared, still thinking about "sosreport (tarball) is taken by sosreport (command)". We do have a warning text `Please note the 'sosreport' command has been deprecated` but it is hard to say how many people read the leading text (or realize its implication). People usually run `sosreport`, wait for the outcome and copy the generated tarball. So maybe we shall put such string (also) at the end of the stdout where it gets more attention as a heads up, now?

Also, what would be the justification for the redirectors drop - just "we have the legacy redirectors for a while"? Isn't better message "here we add another `sos` subcommand (e.g. `sos upload`) under the `sos` umbrella and it is the good time to drop the redirectors"?

mhradile:
+1 for keeping them. Our (old and nowhere near migrated) tests are
definitely not ready.

TurboTurtle:
> AFAIK original plans were to keep the shortcut/redirectors for the `sos-4.*` timeframe. 

[...]

PR #3915, opened 2025-01-29
[build] Drop legacy redirectors

TurboTurtle:
Following on from earlier work that added a deprecation message to the `sosreport` and `sos-collector` redirectors about the commands being dropped in sos-4.9, drop the redirectors and their manpage symlinks so that only `sos` is available as an entry point.

Closes: #3374

---


arif-ali:
It's probably worth making the following change too

```diff
diff --git a/snap/snapcraft.yaml b/snap/snapcraft.yaml
index 1c796aa3..ab865f34 100644
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -41,7 +41,3 @@ parts:
 apps:
   sos:
     command: bin/sos
-  sosreport:
-    command: bin/sos report
-  sos-collector:
-    command: bin/sos collector
```

Also, setup.py needs updating, maaybe

```diff
diff --git a/setup.py b/setup.py
index 9ad4bb75..4295b6b9 100644
--- a/setup.py
+++ b/setup.py
@@ -19,11 +19,10 @@ setup(
     maintainer_email='jhunsake@redhat.com',
     url='https://github.com/sosreport/sos',
     license="GPLv2+",
-    scripts=['bin/sos', 'bin/sosreport', 'bin/sos-collector'],
+    scripts=['bin/sos'],
     data_files=[
-        ('share/man/man1', ['man/en/sosreport.1', 'man/en/sos-report.1',
+        ('share/man/man1', ['man/en/sos-report.1', 'man/en/sos-clean.1',
                             'man/en/sos.1', 'man/en/sos-collect.1',
-                            'man/en/sos-collector.1', 'man/en/sos-clean.1',
                             'man/en/sos-mask.1', 'man/en/sos-help.1']),
         ('share/man/man5', ['man/en/sos.conf.5']),
         ('share/licenses/sos', ['LICENSE']),
```

pmoravec:
Maybe worth updating also:

- https://github.com/sosreport/sos/blob/main/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L6 (this is questionable, as whole github project is named `sosreport`)
- https://github.com/sosreport/sos/blob/main/README.md?plain=1#L113-L117 (rewrite it that `sosreport` was dropped? or remove that paragraph?)
- https://github.com/sosreport/sos/blob/main/sos/collector/__init__.py#L979 - this might need `if sos<4.0 then sosreport else sos` test (or do we support sos collect against sos-3 code..?)
- https://github.com/sosreport/sos/blob/main/sos/collector/__init__.py#L1004 - ditto
- https://github.com/sosreport/sos/blob/main/sos/collector/sosnode.py#L80 (and https://github.com/sosreport/sos/blob/main/sos/collector/sosnode.py#L333-L334) - if we support collector against sos-3, let's keep it

pmoravec (approved):
LGTM

arif-ali (approved):

arif-ali:
For upstream (i.e. here), I don't have a problem with us removing it, and I think we should be good to go.

From our perspective, and in hindsight, we should have renamed the Debian/Ubuntu package a long time ago for a seamless transition, but I had no control or knew about that. As I only recently taking maintainer-ship of the package recently. We are only making that transition now, and has been in the queue for 3 weeks in Debian.

At the moment the Debian and Ubuntu packages and any distros that evolve from these 2 have still been using `sosreport`. From the Canonical perspective, once the `sos` package arrives which is likely in the (25.10, **o** release) to keep transitional package that will carry the `sosreport` and `sos-collector` command into the next LTS that is 26.04. We are then likely to do a full drop of the transitional package on time for 26.10 and the future. What this means that when users install `sosreport`, they will get 2 packages `sosreport` and `sos`, they will be able to remove `sosreport` package and then the code of upstream will remain.

The snap will remain as a `sosreport` name, as we couldn't and wasn't allowed the `sos` name, and this will follow any upstream decision in removal as that is built as part of the GH workflow.

Ultimately, we will be adding the details in our Newsletter (once the package is released) to inform our users and customers of the change so that everyone is aware anyhow to use the new `sos` package and that the `sosreport` command should not be used.

So, tldr, this doesn't effect upstream, and we are making these adjustments to downstream due to the fact we were late to rename the package.

pmoravec:
OK, so let drop `sosreport` in upstream and deal with it in downstream only, per downstream-specific requirements.

I will merge the PR, then.

pmoravec:
Closed by https://github.com/sosreport/sos/commit/b13a3cc1bf0f4e33c536e24aa95f771763208c16 .

PR #3922, opened 2025-02-07
[python] Collect verbose output via pip

sayan3296:
The idea is to be able to see the path where the module is installed. So I would prefer to having 
```
# pip3.9 -v list installed
Package                         Version   Location                               Installer
------------------------------- --------- -------------------------------------- ---------
aiodns                          3.0.0     /usr/lib/python3.9/site-packages
aiofiles                        22.1.0    /usr/lib/python3.9/site-packages
aiohttp                         3.9.2     /usr/lib64/python3.9/site-packages
```

Over
```
# pip3.9 list installed
Package                         Version
------------------------------- ---------
aiodns                          3.0.0
aiofiles                        22.1.0
aiohttp                         3.9.2
```

to ease my troubleshooting of support-cases, where we suspect the module is being picked up from the wrong path. 

---

### DIFF
--- a/sos/report/plugins/python.py
+++ b/sos/report/plugins/python.py
@@ -40,7 +40,7 @@ class Python(Plugin):
             # pip: /usr/bin/pip2.7 /usr/bin/pip3.6
             # where we must skip the first word
             for pip in pips['output'].split()[1:]:
-                self.add_cmd_output(f"{pip} list installed")
+                self.add_cmd_output(f"{pip} -v list installed")
 
 
 class UbuntuPython(Python, DebianPlugin, UbuntuPlugin):


pmoravec (approved):

sayan3296:
One small thing here. I noticed that for Python 3.6 or before, the -v flag does not print the location of the module. It prints the output normally. 
For python 3.8 and later, the -v flag shows the location as expected. 

It's still better than nothing I guess.

TurboTurtle (approved):

TurboTurtle:
> One small thing here. I noticed that for Python 3.6 or before, the -v flag does not print the location of the module. It prints the output normally. For python 3.8 and later, the -v flag shows the location as expected.
> 
> It's still better than nothing I guess.

Nothing to be done about that, I guess. At least 3.6 will be phasing out more quickly now with RHEL and other LTS distros.

Issue #3918, opened 2025-02-04
Feature request: collect stack dump from containerd


samuelkarp:
containerd can produce a stack dump, which is helpful when troubleshooting issues that may turn out to be bugs in containerd.  The stack dump can be triggered by sending `SIGUSR1` to the main containerd daemon, and the dump is then both written to the journal (via containerd's normal logger) and to a file in `/tmp` named `containerd.<pid>.stacks.log` (with `<pid>` being the PID of containerd).

It would be helpful if the containerd plugin could trigger the stack dump and collect them.

TurboTurtle:
Certainly sounds reasonable. I can work something up this week most likely, unless you're wanting to do this yourself and opening this issue just for tracking.

samuelkarp:
I'm not sure I'll have bandwidth soon (and Python is not a language I'm super familiar with) so I'd certainly appreciate it if you made the change. Thanks!

TurboTurtle:
@samuelkarp I just pushed #3926 for this, I think it covers the request but please take a look when you have a moment and let us know if there is anything missing.

samuelkarp:
I think that looks like it would work.  To invoke a user would need to pass `--plugin-option containerd.stackdump`?  The only other bit is if we also wanted to have stack dumps from the shims (`containerd-shim-runc-v2`) which would just be written to the journal, but that seems a more rare use-case.

TurboTurtle:
> To invoke a user would need to pass --plugin-option containerd.stackdump?

Yup.

> The only other bit is if we also wanted to have stack dumps from the shims (containerd-shim-runc-v2) 

That was actually what prompted me to make the changes to `get_process_pids()`, so we weren't signalling those processes :upside_down_face:

samuelkarp:
> That was actually what prompted me to make the changes to get_process_pids(), so we weren't signalling those processes 🙃

That's still probably valuable; getting a stack dump from each shim would be fairly noisy and only useful in rare cases.  Maybe if we put that behind a second option that might be better than having everything behind `containerd.stackdump`?

samuelkarp:
@TurboTurtle Thank you!

PR #3926, opened 2025-02-11
[containerd] Add option to collect stack dump files

TurboTurtle:
Two patchset that:

First, updates `Plugin.get_process_pids()` with a new `exact` argument that, if `True`, will only return PIDs that exactly match the given string, which is necessary to capture the "root"/"main" `containerd` process(es) for the second patch.

Second, adds a `containerd.stackdump` option that, if enabled, will send `SIGUSR1` to the containerd process(es) to trigger writing stack dumps to disk, then marking them for collection.

Resolves: #3918 

---

### DIFF
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -16,8 +16,9 @@ import contextlib
 import os
 import glob
 import re
+import signal
 import stat
-from time import time
+from time import time, sleep
 import logging
 import fnmatch
 import errno
@@ -3569,6 +3570,34 @@ class Plugin():
             except IOError:
                 continue
         return pids
+
+    def signal_process_usr1(self, process):
+        """
+        Send a SIGUSR1 to the pid(s) associated with the specified process
+        name. Callers should be aware that a 1-second delay per signalled pid,
+        up to 5 seconds at most, is expected as to allow sufficient time for
+        the signalled process(es) to react to the received signal.
+
+        :param process: The name or regex pattern of the process(es) to signal
+        :type process:  ``str``
+
+        :returns:   A list of pids that were successfully signalled
+        :rtype:     ``list``
+        """
+        signalled = []
+        pids = self.get_process_pids(process)
+        for pid in pids:
+            try:
+                os.kill(int(pid), signal.SIGUSR1)
+                signalled.append(pid)
+            except Exception as err:
+                self._log_debug(
+                    f"Failed to signal pid {pid} for '{process}': {err}"
+                )
+        # allow a small grace period to allow the signalled pids to complete
+        # whatever they're doing in response to the signal
+        sleep(min(len(signalled), 5))
+        return signalled
 
     def get_network_namespaces(self, ns_pattern=None, ns_max=None):
         if ns_max is None and self.commons['cmdlineopts'].namespaces:
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -3551,7 +3551,7 @@ class Plugin():
     def get_process_pids(self, process):
         """Get a list of all PIDs that match a specified name
 
-        :param process:     The name of the process the get PIDs for
+        :param process:     The name or regex of the process the get PIDs for
         :type process:  ``str``
 
         :returns: A list of PIDs
@@ -3563,8 +3563,8 @@ class Plugin():
         for path in cmd_line_paths:
             try:
                 with open(path, 'r', encoding='utf-8') as f:
-                    cmd_line = f.read().strip()
-                    if process in cmd_line:
+                    cmd_line = f.read().strip('\x00')
+                    if re.match(process, cmd_line):
                         pids.append(path.split("/")[2])
             except IOError:
                 continue
--- a/sos/report/plugins/containerd.py
+++ b/sos/report/plugins/containerd.py
@@ -6,7 +6,7 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.report.plugins import Plugin, IndependentPlugin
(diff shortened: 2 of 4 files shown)


bmr-cymru:
This was ugly when Gluster wanted it and I still don't like the "send magic signals and await magic files" pattern, but this is now the third plugin in the tree to do something similar in different ways (`gluster`, `crio`, now `containerd`). It seems like there are opportunities to consolidate the mechanism for sending the signal (vs. open coding `kill`/`killall`/`os.kill()`) and then waiting for files to appear at a predictable path.

TurboTurtle:
Hmm, the linting errors seem to be 3.9 (and earlier) based. Locally, the syntax works as expected on 3.11. I'll fiddle with this later today.


> It seems like there are opportunities to consolidate the mechanism for sending the signal (vs. open coding kill/killall/os.kill()) and then waiting for files to appear at a predictable path.

Yeah, that's fair. I lean towards leveraging `os.kill()` over shelling out,  but happy to hear alternative before going forward with the consolidation work here.

Sidebar - is Gluster still a thing?

bmr-cymru:
>  linting errors seem to be 3.9 (and earlier) based. Locally, the syntax works as expected on 3.11

I'm seeing:

```python
./bin/sos report --batch
  File "/tmp/cirrus-ci-build/./bin/sos", line 21, in <module>
    sos = SoS(sys.argv[1:])
          ^^^^^^^^^^^^^^^^^
  File "/tmp/cirrus-ci-build/sos/__init__.py", line 54, in __init__
    import sos.report
  File "/tmp/cirrus-ci-build/sos/report/__init__.py", line 27, in <module>
    import sos.report.plugins
  File "/tmp/cirrus-ci-build/sos/report/plugins/__init__.py", line 3570
    reg = rf"{process if not exact else f"^{process}$"}"
                                          ^
SyntaxError: f-string: expecting '}'
```

With `python-3.13.1` on Fedora. I can't see how the nested f-string can not be ambiguous when it's using `"` in both the inner and outer f-string.

This works for me:

```python
    reg = rf"{process if not exact else f'^{process}$'}"
```

TurboTurtle:
Interesting, I'm actually on 3.13.1 as well (thought I was earlier) and I don't get any failure:

```
jake@terra:~/projects/sos$ python -V
Python 3.13.1
jake@terra:~/projects/sos$ sudo ./bin/sos report --batch -k containerd.stackdump

sos report (version 4.8.2)
[...]
Your sos report has been generated and saved in:
	/var/tmp/sosreport-terra-2025-02-12-kzradvj.tar.xz

 Size	172.89MiB
 Owner	root
 sha256	e9c3f650d4d47c6095a6402eaa32040e1d54c27ee79e2e85131172d33a6f2909

Please send this file to your support representative
```

Regardless, fix is straightforward. I'll transition this over to a Plugin method and get something pushed up today/tomorrow.

bmr-cymru:
> Interesting, I'm actually on 3.13.1 as well (thought I was earlier) and I don't get any failure:

My bad - I pasted the snippet into a terminal in an older VM by accident.

TurboTurtle:
Ok, round 2.

The update to `get_process_pids()` is simplified and just allows plugins to pass regex strings or regular strings at their choice, leaving `re` to handle both cases as it already does. No need to morph the process strings into a specific regex based on a new argument, just let the plugin know what's best.

Then, update `crio` and `gluster` - I was able to verify `crio` behavior, but I can't for gluster. I _believe_ what is there now is functionally the same/equivalent as existing behavior, which actually allowed us to drop an extra method.

Finally update `containerd` to align with the new standardized design.

arif-ali:
just a curious one, the gluster plugin was waiting till the file existed, rather than the 5 seconds, will this give enough time for it to come through. Or with the gluster not being used much this is not going to be much of a concern?

bmr-cymru:
> just a curious one, the gluster plugin was waiting till the file existed, rather than the 5 seconds, will this give enough time for it to come through. Or with the gluster not being used much this is not going to be much of a concern?

Since the file names are always predictable one option would be to have a combined "signal-and-wait" method that accepts a process and a path name or pattern, and then returns when the file is present (or raises an error after some reasonable timeout).

(it's nice to see the `wait_for_statedump()` ick finally disappear from gluster.py...)

TurboTurtle:
> just a curious one, the gluster plugin was waiting till the file existed, rather than the 5 seconds, will this give enough time for it to come through. Or with the gluster not being used much this is not going to be much of a concern?

Educated guess says it _should_ be enough, and the previous approach was probably a bit on the paranoid side of cautious. Practically speaking though, probably the latter point of not being much of a concern. Looking at the website, their last release was supposed to be a 12.0 version that was closed by their own automation in for being stale for 6 months back in 2023.

> Since the file names are always predictable one option would be to have a combined "signal-and-wait" method that accepts a process and a path name or pattern, and then returns when the file is present (or raises an error after some reasonable timeout).

I had thought about that, but given the above I didn't think it was warranted at the moment. If we start doing more stackdump(-esque) collections like this we can certainly whip something up like the parsing we do for `add_device_cmd()`.

arif-ali:
> > just a curious one, the gluster plugin was waiting till the file existed, rather than the 5 seconds, will this give enough time for it to come through. Or with the gluster not being used much this is not going to be much of a concern?
> 
> Educated guess says it _should_ be enough, and the previous approach was probably a bit on the paranoid side of cautious. Practically speaking though, probably the latter point of not being much of a concern. Looking at the website, their last release was supposed to be a 12.0 version that was closed by their own automation in for being stale for 6 months back in 2023.

I'm cool with that

arif-ali (approved):

PR #3944, opened 2025-02-26
[networking] List contents of netplan configuration dirs

MggMuggins:
If a user uses `.yml` for a netplan configuration file, netplan silently ignores it. I've engaged with upstream since that behavior isn't ideal, but it isn't likely that it will change soon.

This at least gives us better visibility without needing to worry about obfuscating the contents of other (possibly malformatted) files.

---

### DIFF
--- a/sos/report/plugins/networking.py
+++ b/sos/report/plugins/networking.py
@@ -349,6 +349,14 @@ class UbuntuNetworking(Networking, UbuntuPlugin, DebianPlugin):
             "/run/systemd/network"
         ])
 
+        # Netplan only consumes files with the `.yaml` extension (LP#1815734),
+        # so give visibility on other files that might be present.
+        self.add_dir_listing([
+            "/etc/netplan",
+            "/lib/netplan",
+            "/run/netplan",
+        ])
+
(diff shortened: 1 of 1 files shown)


sos/report/plugins/networking.py
```diff
@@ -349,6 +349,14 @@ def setup(self):
             "/run/systemd/network"
         ])
 
+        # Netplan only consumes files with the `.yaml` extension (LP#1815734),
+        # so give visibility on other files that might be present.
+        self.add_cmd_output([
+            "ls -la /etc/netplan",
+            "ls -la /lib/netplan",
+            "ls -la /run/netplan",
```
arif-ali, line 357: There's a new function that was added a couple of months ago, which may be better aligned here, something like the following could work here.

```python
self.add_dir_listing([
    "/etc/netplan",
    "/lib/netplan",
    "/run/netplan",
])
```

arif-ali (changes requested):

arif-ali (approved):

pmoravec (approved):

PR #3946, opened 2025-02-27
[memory] Collect all sysfs THP files

### LINKED ISSUE #3945: Need all of /sys/kernel/mm/transparent_hugepage to determine if THP is enabled
Need to collect all of `/sys/kernel/mm/transparent_hugepage` to determine if THP is enabled or not now. Changes to THP result in sysfs `enabled` file alone (and the kernel parameter) are now insuffiecient to know if THP is enabled or not. For example, `shmem_enabled` can be set to `[always]` while `enabled` can be set to `[never]` and THP will still be enabled but only for shmem segments. As such, to better know if and how THP is enabled, collect all files under `/sys/kernel/mm/transparent_hugepage/` rather than just `enabled`.

haithcockce:
Changes to THP result in sysfs `enabled` file alone (and the kernel parameter) are now insuffiecient to know if THP is enabled or not. For example, `shmem_enabled` can be set to `[always]` while `enabled` can be set to `[never]` and THP will still be enabled but only for shmem segments. As such, to better know if and how THP is enabled, collect all files under `/sys/kernel/mm/transparent_hugepage/` rather than just `enabled`.

Fixes #3945 

---

- [X] Are any related Issues or existing PRs [properly referenced](https://docs.github.com/en/issues/tracking-your-work-with-issues/creating-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) via a Closes (Issue) or Resolved (PR) line? _N/A_?
- [ ] Are all passwords or private data gathered by this PR [obfuscated](https://github.com/sosreport/sos/wiki/How-to-Write-a-Plugin#how-to-prevent-collecting-passwords)? _N/A_?

### DIFF
--- a/sos/report/plugins/memory.py
+++ b/sos/report/plugins/memory.py
@@ -26,7 +26,7 @@ class Memory(Plugin, IndependentPlugin):
             "/proc/pagetypeinfo",
             "/proc/vmallocinfo",
             "/sys/kernel/mm/ksm",
-            "/sys/kernel/mm/transparent_hugepage/enabled",
+            "/sys/kernel/mm/transparent_hugepage",
             "/sys/kernel/mm/hugepages",
             "/sys/kernel/mm/lru_gen/enabled",
             "/sys/kernel/mm/lru_gen/min_ttl_ms",


TurboTurtle (approved):
Been a while, Lambda! Thanks for the PR. LGTM, afaik there's nothing nested under here that we'd need to avoid like we do with some other `/sys/kernel` paths.

pmoravec (approved):

PR #3949, opened 2025-03-03
[helm] Created new sos report plugin for Helm

### LINKED ISSUE #3504: [plugin] Helm with helm list
I'd like to suggest adding a plugin for helm, specifically helm list to show which charts are installed. If you're using Kubernetes, it is highly likely you have helm and are using helm charts. Thanks!

bryanfraschetti:
This commit adds a new sos report plugin that performs some basic output
collection from several helm commands to view the installed
repositories, plugins, and charts.

Closes #3504

Signed-off-by: Bryan Fraschetti <bryan.fraschetti@canonical.com>

---

### DIFF
--- a/sos/report/plugins/helm.py
+++ b/sos/report/plugins/helm.py
@@ -1,0 +1,40 @@
+# Copyright (C) 2025 Canonical Ltd.,
+#                    Bryan Fraschetti <bryan.fraschetti@canonical.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, IndependentPlugin
+
+
+class Helm(Plugin, IndependentPlugin):
+    """The Helm plugin collects information about the currently installed
+    Helm charts, plugins, and repositories used in delpoyments
+    """
+
+    short_desc = 'The k8s templating and deployment manager'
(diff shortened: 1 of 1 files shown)


sos/report/plugins/helm.py
```diff
@@ -0,0 +1,38 @@
...
+        helm_subcmds = [
+            'repo list',
+            'plugin list',
+            'list -a',
+            'version',
+        ]
+
+        self.add_cmd_output([
+            f"{self.helm_cmd} {subcmd}" for subcmd in helm_subcmds
+        ])
```
TurboTurtle, line 38: Please add the following vim tags as the last line in the file:

```
# vim: set et ts=4 sw=4 :
```

sos/report/plugins/microk8s.py
```diff
@@ -34,6 +34,9 @@ def setup(self):
 
         microk8s_subcmds = [
             'addons repo list',
+            'helm list -a',
+            'helm repo list',
+            'helm plugin list',
```
TurboTurtle, line 39: While I understand the intent, we don't duplicate collections in `sos` - instead we aim to ensure the proper plugin is enabled by default in any environment that could reasonably want that to happen. In this case, that means ensuring that something in the `microk8s` environment is used as an enablement trigger in the new `helm` plugin (if it's possible to have `microk8s` installed without `helm` being present using vendor-support or commonly-documented installation methods).

TurboTurtle (changes requested):
The plugin LGTM on its own, just a small note about vim tags (note to self: take a peak at where we may be missing those across all plugins), and dropping the collections from `microk8s` given the context below.

TurboTurtle (approved):

arif-ali (approved):

PR #3950, opened 2025-03-04
[iscsi] Add /var/lib/iscsi/send_targets to the obfuscation list

jcastill:
Obfuscate discovery and node seesion passwords that can be found in other files inside /var/lib/iscsi directory.

Related: RHEL-81187

---

### DIFF
--- a/sos/report/plugins/iscsi.py
+++ b/sos/report/plugins/iscsi.py
@@ -50,9 +50,9 @@ class Iscsi(Plugin):
         self.do_path_regex_sub('/etc/iscsi/iscsid.conf', nodesessionpwd, repl)
         self.do_path_regex_sub('/etc/iscsi/iscsid.conf', discoverypwd, repl)
         self.do_path_regex_sub(
-                '/var/lib/iscsi/nodes/*/*/*', nodesessionpwd, repl)
+                '/var/lib/iscsi/', nodesessionpwd, repl)
         self.do_path_regex_sub(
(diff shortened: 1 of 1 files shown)


sos/report/plugins/iscsi.py
```diff
@@ -50,9 +50,9 @@ def postproc(self):
         self.do_path_regex_sub('/etc/iscsi/iscsid.conf', nodesessionpwd, repl)
         self.do_path_regex_sub('/etc/iscsi/iscsid.conf', discoverypwd, repl)
         self.do_path_regex_sub(
-                '/var/lib/iscsi/nodes/*/*/*', nodesessionpwd, repl)
+                '/var/lib/iscsi/', nodesessionpwd, repl)
```
pmoravec, line 53: This means much more files are obfuscated than required - is that intentional? Well I guess it can't harm, maybe it can prolong the execution redundantly..?

jcastill: It's intentional, to try to cover every file for the directories there. The passwords can be in the nodes/ directory or in send_targets/ at the same level, and passowrds can be in both of them. There may be other directories in /var/lib/iscsi/ so I thought it may be better to be on the safe side. 
Do you know if this kind of change causes lots of delays when obfuscating? Do you think it would be better to specify the two dirs that we know contain passwords instead?

pmoravec (approved):

arif-ali (approved):

PR #3951, opened 2025-03-04
[ceph_common] Obfuscate rgw password in ceph.conf

arif-ali:
---

### DIFF
--- a/sos/report/plugins/ceph_common.py
+++ b/sos/report/plugins/ceph_common.py
@@ -118,4 +118,11 @@ class CephCommon(Plugin, RedHatPlugin, UbuntuPlugin):
             "ceph -v",
         ])
 
+    def postproc(self):
+        protect_keys = [
+            "rgw keystone admin password",
+        ]
+        regex = fr"(^({'|'.join(protect_keys)})\s*=\s*)(.*)"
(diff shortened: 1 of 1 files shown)


sos/report/plugins/ceph_common.py
```diff
@@ -118,4 +118,11 @@ def setup(self):
             "ceph -v",
         ])
 
+    def postproc(self):
+        protect_keys = [
+            "rgw keystone admin password",
+        ]
+        regex = fr"(^({'|'.join(protect_keys)})\s*=\s*)(.*)"
```
pmoravec, line 125: I dont understand the '|'.join part. Since the list contains just one item, the `'|'.join(..)` will be pure `'rgw keystone admin password'` - is this a preparation for possible future keys to obfuscate, i.e. for protect_keys to be a multi-item list?

arif-ali: yeah, that was the idea, that if there is anything in the future, it's just a one-line change.

Can remove that, and make it just simple, and only obfuscate that key for the time being

pmoravec: No prob having it as is, thanks for explanation.

pmoravec (approved):

PR #3959, opened 2025-03-11
[cups] Expand captures in cups plugin

jcastill:
Add new command captures to the cups plugin.
Also, add a new option, 'userconfs', that when enabled is used to capture config files inside .cups in users' home directories.

Related: SUPDEV-171

---

### DIFF
--- a/sos/report/plugins/cups.py
+++ b/sos/report/plugins/cups.py
@@ -6,7 +6,9 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.report.plugins import Plugin, IndependentPlugin
+import pwd
+
+from sos.report.plugins import Plugin, IndependentPlugin, PluginOpt
 
 
 class Cups(Plugin, IndependentPlugin):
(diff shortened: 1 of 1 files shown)


sos/report/plugins/cups.py
```diff
@@ -39,4 +55,41 @@ def setup(self):
...
+
+    def get_user_configs(self):
+        """
+        Iterate over .cups folders in user homes to capture config files.
+        """
+        users_data = pwd.getpwall()
+        config_files = ([
+            "client.conf",
+            "lpoptions",
+        ])
```
TurboTurtle, line 69: The parens here are redundant.

sos/report/plugins/cups.py
```diff
@@ -39,4 +55,41 @@ def setup(self):
...
+            return
+        non_local_fs = {'nfs', 'nfs4', 'autofs'}
+        # Read the home paths of users in the system and
+        # config files from .cups
+        for user in users_data:
+            if user.pw_dir in fs_mount_info and \
+                    fs_mount_info[user.pw_dir] in non_local_fs:
+                self._log_info(
+                        f"Skipping capture in {user.pw_dir}"
+                        " because it's a remote directory"
```
TurboTurtle, line 87: Nit: over-indented

sos/report/plugins/cups.py
```diff
@@ -39,4 +55,41 @@ def setup(self):
...
+        fs_mount_info = {}
+        try:
+            with open('/proc/mounts', "r", encoding='UTF-8') as mounts_file:
+                for line in mounts_file:
+                    (fs_file, fs_vstype) = line.split()[1:3]
+                    fs_mount_info[fs_file] = fs_vstype
+        except Exception:
+            self._log_error("Couldn't read /proc/mounts")
+            return
+        non_local_fs = {'nfs', 'nfs4', 'autofs'}
```
TurboTurtle, line 79: Not blocking on this, but something for later: we may want to revisit the exclusion of `nfs` for this and the `ssh` plugin (or, realistically, any new method to standardize this exact scenario in other plugins). `autofs` we should _definitely_ be skipping, but for _already mounted_ `nfs` locations, we should be able to collect from there. Worth looking into how practical that is or isn't - I'm not remembering all the details that went into that decision originally, so we may have already gone down that route and still landed here.

pmoravec: Also, if the demand to iterate over home dirs except for those on autofs (or nfs?) would appear once again, we should consider implementing a helper function for it. To deduplucate the code..

pmoravec (approved):

sos/report/plugins/cups.py
jcastill: In the ssh plugin, the problem was running `ls` and causing huge delays in the best case, and hangs in the worst. Part of it was using the option `-Z` iIrc. But in general, if we have a system with lots of NFS mountpoints, and we need to go through them, we may end up in similar scenarios - specially if we run a recursive walk. Meanwhile we could end up not capturing useful data.

Some initial ideas about how to deal with this:
- We could add an argument to `add_dir_listing()` to disable listings of certain remote filesystems by default.
- Same for `add_copy_spec()`?
- Or would it be better to handle home directories in a new function?

Then we disable remote directories by default, and add a new option in the command line to explicitly enable nfs mount point captures, so if we really really need the data, support engineers can enable the capture, advising the end user about possible consequences (it may hang, it could be slow, etc)

I think there's a balance that we need to find here - the cases I checked in the past where there was a problem with sos, timeouts/hangs and NFS dirs involved at least 50 or more NFS mount points, so disabling captures by default could lead to missing data in the cases where there aren't a lot of remote mount points.

I can open a new PR with a very simple implementation and we move the discussion there if it's better.

arif-ali: Added to what you mention, if the NFS mount is the problem, and trying to capture the data from the said NFS mount point could hang, and timeout, and hence the sos would take a considerable amount of time to come through. So, I am on the viewpoint, that we should add it, if it is requested.

arif-ali (approved):

arif-ali:
just restarted the CI job as well, as it was failing on infra related issue

TurboTurtle (approved):

Issue #3853, opened 2024-11-26
[tests] 24.10 Stage Two tests failing for logs


arif-ali:
There's been a few occurances now, and have found that the test below failing, and on restart of the job, it works as normal

```
066-tests_report_tests_plugin_tests_logs.py_JournalSizeLimitTest.test_journal_tailed_and_linked
```

debug.log from avocodo

```
[stdlog] 2024-11-26 08:43:06,719 avocado.test test             L0313 INFO | INIT 1-tests/report_tests/plugin_tests/logs.py:JournalSizeLimitTest.test_journal_tailed_and_linked
[stdlog] 2024-11-26 08:43:06,721 avocado.test parameters       L0141 DEBUG| PARAMS (key=timeout, path=*, default=None) => None
[stdlog] 2024-11-26 08:43:06,721 avocado.test parameters       L0141 DEBUG| PARAMS (key=timeout_factor, path=*, default=1.0) => 1.0
[stdlog] 2024-11-26 08:43:06,721 avocado.test test             L0345 DEBUG| Test metadata:
[stdlog] 2024-11-26 08:43:06,721 avocado.test test             L0347 DEBUG|   filename: /tmp/cirrus-ci-build/tests/report_tests/plugin_tests/logs.py
[stdlog] 2024-11-26 08:43:06,721 avocado.test test             L0353 DEBUG|   teststmpdir: /var/tmp/avocado_30pafaq7
[stdlog] 2024-11-26 08:43:06,721 avocado.test test             L0354 DEBUG|   original timeout: None
[stdlog] 2024-11-26 08:43:06,721 avocado.test test             L0355 DEBUG|   timeout factor: 1.0
[stdlog] 2024-11-26 08:43:06,721 avocado.test test             L0356 DEBUG|   actual timeout: None
[stdlog] 2024-11-26 08:43:06,722 avocado.test test             L0548 INFO | START 1-tests/report_tests/plugin_tests/logs.py:JournalSizeLimitTest.test_journal_tailed_and_linked
[stdlog] 2024-11-26 08:43:06,724 avocado.test stacktrace       L0040 ERROR| 
[stdlog] 2024-11-26 08:43:06,725 avocado.test stacktrace       L0042 ERROR| Reproduced traceback from: /usr/local/lib/python3.12/dist-packages/avocado/core/test.py:638
[stdlog] 2024-11-26 08:43:06,726 avocado.test stacktrace       L0049 ERROR| Traceback (most recent call last):
[...]

PR #3962, opened 2025-03-13
[tests] Rework journal size calculation for tests

jcastill:
In 3853 it was discovered that size calculations via get_usage() were not corresponding to the expected size, and the tests were failing sometimes.
This commit tries to fix the problem by calling
'journalctl' directly and then calculating the size on the output.

Related: #3853

---

### DIFF
--- a/tests/report_tests/plugin_tests/logs.py
+++ b/tests/report_tests/plugin_tests/logs.py
@@ -41,7 +41,7 @@ class JournalSizeLimitTest(StageTwoReportTest):
     :avocado: tags=stagetwo
     """
 
-    sos_cmd = '-o logs --journal-size=20 --log-size=10'
+    sos_cmd = '-o logs --journal-size=10 --log-size=5'
     sos_timeout = 500
     packages = {
         'rhel': ['python3-systemd'],
@@ -54,7 +54,7 @@ class JournalSizeLimitTest(StageTwoReportTest):
         from systemd import journal  # pylint: disable=import-error
         _reader = journal.Reader()
         _size = _reader.get_usage() / 1024 / 1024
-        if _size > 20:
+        if _size > 30:
             return
         # write 20MB at a time to side-step rate/size limiting on some distros
         # write over 20MB to ensure we will actually size limit inside sos,


jcastill:
Starting this as a draft. This is a first attempt at solving the issue in 3853. I'm sure there are more elegant ways to do so, so please let me know your thoughts. 
I used _process.run_ from _avocado.utils_ like we do in the _teamd_ plugin test, I imagined that's the best way to get the output of `journalctl`.

pmoravec:
I see just bad answers here :) I dont want to undermine idea that I came up with. But I dont like running plain `journalctl` on a system that can have huge logs (the test will consume huge pile of memory, no?). But I dont have anything better at hand.. (apart an obvious idea "let generate 20M logs everytime" which has own gotchas as well (slow, spams logs on repeated runs).

jcastill:
> I see just bad options here :) I dont want to undermine idea that I came up with. But I dont like running plain `journalctl` on a system that can have huge logs (the test will consume huge pile of memory, no?).

Yes, that's one possibility. We have a timeout as well so we can control the capture somehow.

> But I dont have anything better at hand.. (apart an obvious idea "let generate 20M logs everytime" which has own gotchas as well (slow, spams logs on repeated runs).

We could modify the sizes on the actual command, so instead of:

-o logs --journal-size=20 --log-size=10

We reduce these to half of less than that. Unless 20M and 10M are there because of a specific scenario. 

And a third option could be to generate at least 20M anyway, without checking the current size. We have already a disclaimer in the test:

```
    Note: this test will insert over 100MB of garbage into the test system's
    journal
```

So the expectation is set already.

pmoravec:
> > I see just bad options here :) I dont want to undermine idea that I came up with. But I dont like running plain `journalctl` on a system that can have huge logs (the test will consume huge pile of memory, no?).
> 
> Yes, that's one possibility. We have a timeout as well so we can control the capture somehow.
> 
> > But I dont have anything better at hand.. (apart an obvious idea "let generate 20M logs everytime" which has own gotchas as well (slow, spams logs on repeated runs).
> 
> We could modify the sizes on the actual command, so instead of:
> 
> -o logs --journal-size=20 --log-size=10
> 
> We reduce these to half of less than that. Unless 20M and 10M are there because of a specific scenario.
> 
> And a third option could be to generate at least 20M anyway, without checking the current size.

When you repeatedly run same test on your system, you get journal logs redundantly bloated. Each option has its gotcha :)

Maybe a "compromise" in:
- decrease tested sizes from `--journal-size=20 --log-size=10` several times, e.g. to 10 / 5
- modify the current test to "if stored journal is less than 30 (3times what we need), then generate 10M"
?

The 3times more than what we need should be sufficiently high (really?) to prevent the phenomena I discovered, while lowering the test's sizes and having a upper limit ("less than 30MB") limits the journal bloating..?

jcastill:
> Maybe a "compromise" in:
> 
> * decrease tested sizes from `--journal-size=20 --log-size=10` several times, e.g. to 10 / 5
> * modify the current test to "if stored journal is less than 30 (3times what we need), then generate 10M"
>   ?
> 
> The 3times more than what we need should be sufficiently high (really?) to prevent the phenomena I discovered, while lowering the test's sizes and having a upper limit ("less than 30MB") limits the journal bloating..?

I think this could work. I'll modify this PR and send it again

pmoravec (approved):

arif-ali (approved):

pmoravec:
Closed via [23a7018](https://github.com/sosreport/sos/commit/23a7018473247a4355ae1a615fc644d1661a8dcb) .

PR #3968, opened 2025-03-17
[report] Don't lowercase values by default

arif-ali:
When a value is given it should be preserved rather than changing the case of the value

---

### DIFF
--- a/sos/report/__init__.py
+++ b/sos/report/__init__.py
@@ -916,17 +916,17 @@ class SoSReport(SoSComponent):
                     val = True
 
                 if isinstance(val, str):
-                    val = val.lower()
-                    if val in ["on", "enable", "enabled", "true", "yes"]:
+                    arg = val.lower()
+                    if arg in ["on", "enable", "enabled", "true", "yes"]:
                         val = True
-                    elif val in ["off", "disable", "disabled", "false", "no"]:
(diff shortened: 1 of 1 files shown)


sos/report/__init__.py
```diff
@@ -916,10 +916,11 @@ def _set_tunables(self):
                     val = True
 
                 if isinstance(val, str):
-                    val = val.lower()
-                    if val in ["on", "enable", "enabled", "true", "yes"]:
+                    val_lower = val.lower()
```
bmr-cymru, line 919: This code reads kinda confusingly now. The value you're preserving in `val` is only relevant if we take the exception branch from the attempted `int()` conversion. Mostly `val` is rewritten/type converted, either by coercing it to a boolean or parsing it to an integer.

It might be more readable to separate _the value we receive_ from _the value we set_, maybe calling them something like `arg` vs. `val`, or at least changing the comment on the pass branch in the `int()` conversion to be more explicit about the meaning of the branch?

bmr-cymru:
```
# not a number to convert back to int from argparse
```

Maybe something like "Value is a string argument to be used unmodified"?

bmr-cymru (approved):

TurboTurtle (approved):

PR #3969, opened 2025-03-17
[crio] Add `all` option to `crictl ps -v`

oarribas:
Add the `all` option to `crictl ps -v`, to identify OoM pods when the pod is no longer running.

If the pod is no longer running, the `crictl ps -v` will not contain data to identify OoM pods following [1].

Fixes: SUPDEV-176

---

[1] https://access.redhat.com/solutions/5571301

### DIFF
--- a/sos/report/plugins/crio.py
+++ b/sos/report/plugins/crio.py
@@ -61,7 +61,7 @@ class CRIO(Plugin, RedHatPlugin, UbuntuPlugin, CosPlugin):
             'pods',
             'ps',
             'ps -a',
-            'ps -v',
+            'ps -va',
             'stats',
             'version',
         ]


sos/report/plugins/crio.py
```diff
@@ -61,7 +61,7 @@ def setup(self):
             'pods',
             'ps',
             'ps -a',
-            'ps -v',
+            'ps -va',
```
pmoravec, line 64: Isn't collecting `ps -a` redundant, then? Or are *both* outputs valuable to collect and use, as the output format differs a lot?

pmoravec:
Please amend the commit for DCO / add `Signed-off-by: Author Name <authoremail@example.com>` - like line to the commit message.

oarribas:
> Please amend the commit for DCO / add `Signed-off-by: Author Name <authoremail@example.com>` - like line to the commit message.

@pmoravec , is it enough adding a line like that manually to the commit?

sos/report/plugins/crio.py
oarribas: The `ps -a` collects a "short" output for all containers (including the non-running ones), while `ps -va`  collects a verbose output (and the `a` added by this PR is to include also the non-running ones). I think both are useful.

pmoravec:
> > Please amend the commit for DCO / add `Signed-off-by: Author Name <authoremail@example.com>` - like line to the commit message.
> 
> @pmoravec , is it enough adding a line like that manually to the commit?

Yes, just run `git commit --amend`, append a `Signed-off-by` line to the commit message and `git push --force`.

arif-ali (approved):

pmoravec:
Closed via [cec0e3c](https://github.com/sosreport/sos/commit/cec0e3cd65c03381ddaa325b3ebed81ecad4b3db).

oarribas:
@pmoravec , so the changes from this PR are now merged via commit [cec0e3c](https://github.com/sosreport/sos/commit/cec0e3cd65c03381ddaa325b3ebed81ecad4b3db), right?

PR #3970, opened 2025-03-17
[charmed_mysql] Add Charmed MySQL support

vlad-apostol:
This plugin is intended to add support for Charmed MySQL deployments. 
An example of the output for the report can be found here: [Charmed MySQL example report](https://drive.google.com/drive/folders/1SyFB5T0JNM8ey-dMRRp4Ju8MGW7_dF2x?usp=sharing)

Please note: this PR requires PR #3968 to be merged before this is functional.
---

### DIFF
--- a/sos/report/plugins/charmed_mysql.py
+++ b/sos/report/plugins/charmed_mysql.py
@@ -1,0 +1,112 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.#
+
+import os
+
+from sos.report.plugins import Plugin, PluginOpt, UbuntuPlugin
+
+SNAP_PATH_COMMON = "/var/snap/charmed-mysql/common"
+SNAP_PATH_CURRENT = "/var/snap/charmed-mysql/current"
+
+PATHS = {
+    "MYSQL_CONF": f"{SNAP_PATH_CURRENT}/etc/mysql",
+    "MYSQL_LOGS": f"{SNAP_PATH_COMMON}/var/log/mysql",
+    "MYSQL_ROUTER_CONF": f"{SNAP_PATH_CURRENT}/etc/mysqlrouter",
+    "MYSQL_ROUTER_LOGS": f"{SNAP_PATH_COMMON}/var/log/mysqlrouter",
+    "MYSQL_SHELL_LOGS": f"{SNAP_PATH_COMMON}/var/log/mysqlsh",
+}
+
+
+class CharmedMySQL(Plugin, UbuntuPlugin):
+    """
+    The Charmed MySQL plugin is used to collect MySQL configuration and logs
+    from the Charmed MySQL snap package. It also collects MySQL Router
+    and MySQL Shell configuration and logs where available,
+    journal logs for the snap, and snap info.
+
+    If the `dumpdbs` option is set to `True`, the plugin will also try and
+    collect the names of the databases that the user has access to. The
+    `mysql` user is used by default, but that can be set using the `dbuser`
+    option. When using the `dumpdbs` option, you must then provide the
+    password for the user using the `dbpass` option.
+    """
+
+    short_desc = "Charmed MySQL"
+    plugin_name = "charmed_mysql"
+    packages = ("charmed-mysql",)
+
+    option_list = [
+        PluginOpt(
+            "dbuser", default="mysql", val_type=str,
+            desc="Username for database dump collection"
+        ),
+        PluginOpt(
+            "dbpass", default="", val_type=str,
+            desc="Password for database dump collection",
+        ),
+        PluginOpt(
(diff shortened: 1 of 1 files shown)


sos/report/plugins/charmed_mysql.py
```diff
@@ -0,0 +1,102 @@
...
+            # Check password is not already an environment variable
+            # and user did not supply a password
+            if "MYSQL_PWD" in os.environ and not db_pass:
+                self.soslog.info(
+                    "MYSQL_PWD present: Using MYSQL_PWD environment variable, "
+                    "user did not provide password."
+                )
+                db_pass = os.environ["MYSQL_PWD"]
+            elif not db_pass:  # Environment variable not set and no password
+                self.soslog.warning(dumpdbs_error)
```
arif-ali, line 88: nit; no need to create a variable for `dumpdbs_error`, just bring it down here, wdyt

arif-ali (changes requested):
Can you squash your commits, so that it's one commit please

sos/report/plugins/charmed_mysql.py
```diff
@@ -0,0 +1,102 @@
...
+                return
+
+            os.environ["MYSQL_PWD"] = db_pass
+
+            opts = f"-h 127.0.0.1 -u{db_user}"
+
+            # Record all available databases
+            self.add_cmd_output(f"mysql {opts} -e 'show databases;'")
```
TurboTurtle, line 96: Don't directly set the environment variable via `os`. Instead, pass the vars via the `env` parameter:

```python
mysql_env = {'MYSQL_PWD': db_pass}

self.add_cmd_output(
    f"mysql -h 127.0.0.1 -u{db_user} -e 'show databases;'",
    env=mysql_env
)

File: sos/report/plugins/charmed_mysql.py [hunk 3]
```diff
@@ -0,0 +1,102 @@
...
+
+    def postproc(self):
+
+        # Remove the password from the environment
+        if "MYSQL_PWD" in os.environ:
+            del os.environ["MYSQL_PWD"]
```
TurboTurtle, line 102: Won't be needed when using the `env` parameter.

sos/report/plugins/charmed_mysql.py
```diff
@@ -0,0 +1,102 @@
...
+    "MYSQL_CONF": f"{SNAP_PATH_CURRENT}/etc/mysql",
+    "MYSQL_LOGS": f"{SNAP_PATH_COMMON}/var/log/mysql",
+    "MYSQL_ROUTER_CONF": f"{SNAP_PATH_CURRENT}/etc/mysqlrouter",
+    "MYSQL_ROUTER_LOGS": f"{SNAP_PATH_COMMON}/var/log/mysqlrouter",
+    "MYSQL_SHELL_LOGS": f"{SNAP_PATH_COMMON}/var/log/mysqlsh",
+}
+
+
+class CharmedMySQL(Plugin, UbuntuPlugin):
+
```
TurboTurtle, line 26: Adding a docstring describing the plugin's behavior here is helpful, as that output is used when invoking `sos help report.plugins.charmed_mysql` so that users can get more info.

See the `filesys` or `kernel` plugins for good examples - we don't need everything that gets collected detailed, just a short overall description of behavior.

sos/report/plugins/charmed_mysql.py
vlad-apostol: I think that's sensible. I've gone ahead and remove the variable and added that in the log itself.

sos/report/plugins/charmed_mysql.py
vlad-apostol: Sounds good, I've gone ahead and done so

sos/report/plugins/charmed_mysql.py
vlad-apostol: Thanks for the resources, I've added a docstring that I think sounds sensible for the goal of the plugin.

vlad-apostol:
Commits have been squashed

sos/report/plugins/charmed_mysql.py
```diff
@@ -0,0 +1,109 @@
...
+                return
+
+            mysql_env = {"MYSQL_PWD": db_pass}
+
+            opts = f"-h 127.0.0.1 -u{db_user}"
+
+            # Record all available databases
+            self.add_cmd_output(f"mysql {opts} -e 'show databases;'",
+                                env=mysql_env
+                                )
```
arif-ali, line 109: Sorry, missed this in the previous review, the indentation and the follow is not consistent with style guide.

```
            self.add_cmd_output(
                f"mysql {opts} -e 'show databases;'",
                env=mysql_env
            )
```

It's possible that it may now be 80 characters now with the env on the same line

Also can you add the vim line at the bottom

```
# vim: set et ts=4 sw=4 :
```

arif-ali (changes requested):
one final nit from me

sos/report/plugins/charmed_mysql.py
vlad-apostol: I've changed that to follow the style guide and vim line has also been added

arif-ali (approved):

TurboTurtle (approved):
LGTM, thanks!

PR #3976, opened 2025-03-26
[charmed_mysql] Add Charmed MySQL K8s support

vlad-apostol:
This PR is a modification to the Charmed MySQL plugin to add support for the Kubernetes version of this charm. The main difference is what paths are constructed for configuration and logs.

There are also new MySQL queries to gather more data about the state of the database and juju deployment.

An example of the report can be found here: [Charmed MySQL K8s Example report](https://drive.google.com/drive/folders/1lnH0EYvAYWjqAsmMnnrkSDkyoMXpO8V1?usp=sharing)

---

### DIFF
--- a/sos/report/plugins/charmed_mysql.py
+++ b/sos/report/plugins/charmed_mysql.py
@@ -10,17 +10,6 @@ import os
 
 from sos.report.plugins import Plugin, PluginOpt, UbuntuPlugin
 
-SNAP_PATH_COMMON = "/var/snap/charmed-mysql/common"
-SNAP_PATH_CURRENT = "/var/snap/charmed-mysql/current"
-
-PATHS = {
-    "MYSQL_CONF": f"{SNAP_PATH_CURRENT}/etc/mysql",
-    "MYSQL_LOGS": f"{SNAP_PATH_COMMON}/var/log/mysql",
-    "MYSQL_ROUTER_CONF": f"{SNAP_PATH_CURRENT}/etc/mysqlrouter",
-    "MYSQL_ROUTER_LOGS": f"{SNAP_PATH_COMMON}/var/log/mysqlrouter",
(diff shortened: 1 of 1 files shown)


sos/report/plugins/charmed_mysql.py
```diff
@@ -109,4 +123,34 @@ def setup(self):
                 env=mysql_env
             )
 
+            # Get unit operations from MySQL to see for example
+            # if a unit is stuck on joining the cluster
+            self.add_cmd_output(
+                f"mysql {opts} -e "
+                "'select * from mysql.juju_units_operations;'",
+                env=mysql_env
+            )
```
TurboTurtle, line 132: We can button these up a bit, e.g.

```python
sqlcmd = f"mysql {opts} -e"
queries = [
    "'select * from mysql.juju_units_operations;'",
    "'show global status like \"%conne%\";'",
    ("'select * from performance_schema.replication_group_members "
     "order by MEMBER_HOST;' "),
    [...etc, you get the idea...]
]

self.add_cmd_output(
    [f"{sqlcmd} {query}" for query in queries],
    env=mysql_env
)
```

vlad-apostol: I like this, I'll push it up soon

vlad-apostol:
@TurboTurtle Unsure about the fedora failure in the CI, I am not convinced the problem is from the PR but maybe a transient error that may go away on a rerun?

TurboTurtle:
Yeah, doesn't look like it was PR related. I've restarted the job.

TurboTurtle (approved):

arif-ali (approved):
lgtm, as the plugin is specific for charmed mysql, and just tests for snap vs k8s, then this makes sense.

Issue #3975, opened 2025-03-26
[testing] autopkgtests failure for report_disabled_parsers.py:ReportDisabledParsersTest.test_local_ip_not_obfuscated


arif-ali:
In preparation of avocado landing in Debian and Ubuntu (which is due to land in a week or so), I have been doing some tests in our test environments, and we get the following issue

```
(007/276) tests/cleaner_tests/report_disabled_parsers.py:ReportDisabledParsersTest.test_local_ip_not_obfuscated:  FAIL: Content '127.0.1.1' does not appear in /var/tmp/avocado_4k2ol_orsos_tests.py.ReportDisabledParsersTest/sosreport-ReportDisabledParsersTest/sosreport-host0-2025-03-24-vlozhxk/ip_addr\n1: lo    inet 127.0.0.1/8 scope host lo\       valid_lft forever preferre... (13.42 s)
```

This issue has been created to track the issue so that it's not lost, and will form a fix soon.

Full Logs: https://autopkgtest.ubuntu.com/results/autopkgtest-plucky-arif-ali-sos-avocado/plucky/amd64/s/sos/20250324_192636_f5814@/log.gz#S5

arif-ali:
So, I've figured this out

When creating new instances we have the following whether it's rocky, or ubuntu 

Hosts file
```
[root@rocky9-01 ~]# cat /etc/hosts
127.0.1.1	rocky9-01
127.0.0.1   localhost localhost.localdomain localhost4 localhost4.localdomain4
::1         localhost localhost.localdomain localhost6 localhost6.localdomain6
```

Use the pythonic way to get IP
```
[root@rocky9-01 ~]# python3
Python 3.9.21 (main, Dec  5 2024, 00:00:00) 
[GCC 11.5.0 20240719 (Red Hat 11.5.0-2)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import socket
>>> print(socket.gethostbyname("rocky9-01"))
127.0.1.1
```

Check the current ip addresses
```
[root@rocky9-01 ~]# ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
[...]

PR #3981, opened 2025-04-03
[tests] Fix issue with 127.0.1.1 IP being detected

arif-ali:
Running stageone tests in some environments sees the hostname of the
machine to have an IP of 127.0.1.1 but the host is never set up in a
way that this address is configured on the host. We use socket.connect
to connect to a dummy address, which ensure to go via a default route
and therefore not track the loopbak address. Then grabbing the IP of
the socket give you a proper local IP address.

Closes: #3975

---

### DIFF
--- a/tests/sos_tests.py
+++ b/tests/sos_tests.py
@@ -136,7 +136,10 @@ class BaseSoSTest(Test):
 
         # get networking info
         hostname = socket.gethostname()
-        ip_addr = socket.gethostbyname(hostname)
+        s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+        # This doesn't send any data
+        s.connect(('10.255.255.255', 1))
+        ip_addr = s.getsockname()[0]
         sysinfo['networking'] = {}
         sysinfo['networking']['hostname'] = hostname
         sysinfo['networking']['ip_addr'] = ip_addr


pmoravec:
Isn't it better to change the `ip_addr = socket.gethostbyname(hostname)` assignment of `self.sysinfo['pre']['networking']['ip_addr']`, as it does not match `ip addr` output?

(.. which can have its caveats as well, so I am still OK-ish with the workaround kind of fix)

arif-ali:
@pmoravec thanks for your input, I was looking for a quick way to solve it, but after further investigation, and looking through stuff, I think I may have found a better approach

Interestingly that I get 2 different results with the following 2 different hostnames

```
>>> print(socket.gethostbyname(f"{socket.gethostname()}.local"))
10.201.177.101
```

```
>>> print(socket.gethostbyname(f"{socket.gethostname()}"))
127.0.1.1
```

Maybe we can check to see if it's just hostname without the domain, we can append `.local` otherwise we keep it as is. The first result is the right result we should be getting

pmoravec (approved):
Really nice fix!

(maybe worth having a comment in the code explaining the "why we dont like 127.* addresses? :-o" but.. well.. git blame is enough)

arif-ali:
Although I've tested this locally and all works where it wasn't before, need to still test on the Ubuntu servers.

So, please don't merge before that. Probably do further testing next week.

TurboTurtle (approved):

arif-ali:
So, I have spent some time on this over the weekend, and not had much luck.

This current proposal doesn't actually solve the problem, and we still see 127.0.1.1 as the ip address.

There are other ways to do this, but there are various pros/cons

1. Find the default gateway via the socket, and get the IP address that way, seen that been done elsewhere. I've tried using ` socket.gethostbyname_ex())`, but I think that will doe the same thing, except that it will bring multiple ip addresses
2. Use `netifaces`, but I think this has been long deprecated
3. Use `psutil`, but that then adds a new dependency for tests

For both `netifaces` and `psutil` options, we could get all IP addresses that the system sees, and do the tests against that too, so that could extra testing there (I have this mocked up locally).

Any other thoughts that anyone else has that other libraries that could give us a hand here? thoughts?

Latest logs: https://autopkgtest.ubuntu.com/results/autopkgtest-plucky-arif-ali-sos-avocado/plucky/amd64/s/sos/20250407_121309_876af@/log.gz

arif-ali:
This latest update works now on the Ubuntu packaging server with the autopkgtests, hopefully this should be cool here.

Referecnce: https://autopkgtest.ubuntu.com/results/autopkgtest-plucky-arif-ali-sos-avocado/plucky/amd64/s/sos/20250408_135857_ff144@/log.gz

TurboTurtle (approved):
LGTM, clever way to handle the edge cases we've seen here.

pmoravec:
Sorry for a delay from me here. I am finalizing a draft version of concurrent cleaner that took me so many iterations and changes and time. So I will get to this most probably on Friday.

.. which is still on time to get it merged before 4.9.1 I plan to tag on Tuesday 15th.

pmoravec (approved):
Nice!

pmoravec:
Fixed via https://github.com/sosreport/sos/commit/8c3288b971edd5f073d0788bd82ebae34861ad31.

PR #3982, opened 2025-04-03
[cirrus] Update Fedora testing images

TurboTurtle:
Update the Fedora images we use for testing, now that newer Fedora image builds, beyond the coreos images that we don't use, are again being published by the Fedora Project.

---

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -3,8 +3,8 @@
 # Main environment vars to set for all tasks
 env:
 
-    FEDORA_NAME: "fedora-38"
-    FEDORA_PRIOR_NAME: "fedora-37"
+    FEDORA_NAME: "fedora-41"
+    FEDORA_PRIOR_NAME: "fedora-40"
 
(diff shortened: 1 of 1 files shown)


TurboTurtle:
I don't think this is actually valid:

```
error: Failed build dependencies:
	python3dist(pexpect) is needed by sos-4.9.0-1.fc41.noarch
	python3dist(pyyaml) is needed by sos-4.9.0-1.fc41.noarch
```

These are not _build_ dependencies, but they are runtime dependencies. This was added via #3260 - @sandrobonazzola thoughts? I'm not very familiar with these macros, or why they might be treating these as build requirements. I can manually build the current version of `sos` locally without either of these packages being installed.

sandrobonazzola:
I think the macro is adding `python3dist(pexpect)` and  `python3dist(pyyaml)`
because of `setup.py`:

```python
install_requires=['pexpect', 'pyyaml', 'packaging;python_version>="3.11"'],
```

arif-ali (approved):

PR #3985, opened 2025-04-04
Fadump sysfs

sourabhjains:
---

### DIFF
--- a/sos/report/plugins/kdump.py
+++ b/sos/report/plugins/kdump.py
@@ -25,11 +25,9 @@ class KDump(Plugin):
             "/proc/sys/kernel/panic",
             "/proc/sys/kernel/panic_on_oops",
             "/sys/kernel/kexec_loaded",
+            "/sys/kernel/fadump",
             "/sys/kernel/fadump_enabled",
-            "/sys/kernel/fadump/enabled",
             "/sys/kernel/fadump_registered",
-            "/sys/kernel/fadump/registered",
-            "/sys/kernel/fadump/mem_reserved",
             "/sys/kernel/kexec_crash_size"
         ])
 


arif-ali:
I think, it still makes sense to collect both `/sys/kernel/fadump_enabled` and `/sys/kernel/fadump_registered`, as sos does get back-ported to older distros, and we need to have this compatible with older releases where the kernel change has not been implemented.

sourabhjains:
> I think, it still makes sense to collect both `/sys/kernel/fadump_enabled` and `/sys/kernel/fadump_registered`, as sos does get back-ported to older distros, and we need to have this compatible with older releases where the kernel change has not been implemented.

For RHEL, the kernel patch that makes /sys/kernel/fadump_enabled and /sys/kernel/fadump_registered
symlinks to their respective new locations is included in RHEL 8.3 and later versions, RHEL 9 and RHEL 10.

Given the availability of this kernel patch, do we still need to collect the old sysfs paths?

If you think we do, I will drop the second patch.

Please share your thoughts.

Thanks for the review @arif-ali 

- Sourabh Jain

arif-ali:
Thanks for the details, I will double check on the Ubuntu side, give me a couple of days, and I will verify that this has also been already implemented as well.

sourabhjains:
> @sourabhjains I have done a kernel tree walk across the supported distros and the kernel that has the commit is in 5.15, and this is the one for 22.04 (jammy). We still support 20.04 (focal)o, so taking entries out at the moment will not make sense, as we could potentially lose that from a supported distro.
> 
> I hope that makes sense

Yeah, let's drop the second patch for now.

I'll update the pull request and remove the second patch.

Thanks for your effort.

arif-ali (approved):

TurboTurtle (approved):

sourabhjains:
Thanks for accepting the patch.

PR #3987, opened 2025-04-09
[presets] Adjust OCP preset options

jcastill:
Change the options used in the OCP preset based
on current use by support engineers and
documentation for OCP 4. Basically:
- Enable crio options crio.all and crio.logs.
- Enable podman options podman.all and podman.logs.
- Explicitly enable the 'openshift' plugin to cover old sos installations where commit a379ab3d2a is not present.

Related: RHEL-86516

---

### DIFF
--- a/sos/presets/redhat/__init__.py
+++ b/sos/presets/redhat/__init__.py
@@ -36,9 +36,13 @@ RHOCP = "ocp"
 RHOCP_DESC = "OpenShift Container Platform by Red Hat"
 RHOCP_OPTS = SoSOptions(
     skip_plugins=['cgroups'], container_runtime='crio', no_report=True,
-    log_size=100,
+    log_size=100, enable_plugins=['openshift', 'openshift_ovn'],
     plugopts=[
         'crio.timeout=600',
+        'crio.all=on',
+        'crio.logs=on',
+        'podman.all=on',
+        'podman.logs=on',
         'networking.timeout=600',
         'networking.ethtool-namespaces=False',
         'networking.namespaces=200'


jcastill:
This is still a draft because we are discussing the size limit. One of the requests was to add --all-logs, but we are evaluating if we could just increase the capture from 100M without having to enable this option.

jcastill:
I'm not really sure that having explicitly enabled the 'openshift' plugin will be useful here at all. If you get this updated 'preset' definition, you'll have an updated sos anyway. Unless anyone can think of a scenario where this makes sense, I'll remove it in the next push.

oarribas:
@jcastill , as discussed in RHEL-86516, having explicitly enabled some plugins could prevent some bugs (like SUPDEV-161) to not enable the desired plugins for OCP.

pmoravec (approved):
If the "collect all logs" option for the 2 plugins is the will, let be so!

sos/presets/redhat/__init__.py
```diff
@@ -36,9 +36,13 @@
 RHOCP_DESC = "OpenShift Container Platform by Red Hat"
 RHOCP_OPTS = SoSOptions(
     skip_plugins=['cgroups'], container_runtime='crio', no_report=True,
-    log_size=100,
+    log_size=100, enable_plugins=['openshift','openshift_ovn'],
```
pmoravec, line 39: Nitpick: ad space after comma after 'openshift' (`flake8 complains).

jcastill: Apologies for missing it. I'll push a fix today

pmoravec (approved):

sos/presets/redhat/__init__.py
```diff
@@ -36,9 +36,13 @@
...
-    log_size=100,
+    log_size=100, enable_plugins=['openshift', 'openshift_ovn'],
     plugopts=[
         'crio.timeout=600',
+        'crio.all=on',
+        'crio.logs=on',
+        'podman.all=on',
+        'podman.logs=on',
         'networking.timeout=600',
         'networking.ethtool-namespaces=False',
```
jcastill, line 47: > would it be possible to also add networking.ethtool-namespaces=off (or equivalent) the plugopts?

@palonsoro It's disabled already. Maybe we should investigate why did the option run in the examples you mentioned.

arif-ali (approved):

Issue #3097, opened 2022-12-29
Run cleaner concurrently also inside one archive 


pmoravec:
`sos report --cleaner` cleans the generated report sequentially using one thread, what prolongs its already "naturally slow" execution. We should run it concurrently, similarly like `sos report` executes plugins.

I see two approaches here:
- run individual parsers (IP, MAC, username,..) concurrently, with a lock on each file they update. This is required to prevent concurrent modification of the same file by multiple parsers.
- or for each parser running sequentially, clean files in parallel

Currently, `cleaner` iterates over all files and runs individual parsers sequentially for one file after another. That is quite away of either above approach.

Also, `sos clean --jobs=4` declares concurrency on individual archives that can be cleaned in parallel. This additional concurrency (with assumed 4 workers as default) should be aligned with that one - we should not end up with 4*4 concurrent workers running in parallel.

Despite there are the above obstacles to the proposal, I think the justification for the improvement is solid: the use cases when `cleaner` is run on a sequential archive dominate and the improvement would speed them up significantly.

TurboTurtle:
I took a stab at moving our concurrency to process-based a while back and the direction I took there was to try and assign each archive to a specific process. There were a few challenges there - in no particular order:

- We need to keep the parsers sync'd across processes, so that obfuscations are kept consistent across both archives and files.
- Because of this, we need to use shared memory objects for holding those obfuscations. For multiple archives this is fine, but in the case of single archive cleaning it actually significantly harms performance. 
- Logging is performed via the archive abstractions - but logging streams are not able to be serialized and so the pickling that occurs when forking out to a new process fails. There was/is some ambiguity here as newer versions of python don't seem to have the same problem (the new process initializes without issue) but core python maintainers have stated that _shouldn't_ be happening.

That being said, I'm not of the opinion that we have to have parallelism at the archive level. If we instead move to file-level parallelism, that's fine - we just need to be able to report when we've finished the last file in an archive and when we've started a new archive's file list.

This also doesn't have to be with a specific library, but it _does_ have to be available for python-3.6 for our downstreams that use that version.

NikhilKakade-1:
@TurboTurtle @pmoravec , any new developments to fix it?

pmoravec:
> @TurboTurtle @pmoravec , any new developments to fix it?

Nothing I am aware of..

NikhilKakade-1:
@TurboTurtle  @pmoravec , any update?
[...]

PR #3988, opened 2025-04-10
[cleaner] Make cleaner concurrent inside one archive

pmoravec:
Draft version adding to the traditional sequential backend also sqlite3 and file based concurrent ones.

TL;DR: please review the idea, test it and comment or ack or nack the approach. Then I will fix the TODOs.

Let me explain various factors and reasons that affected the chosen implementation.

As very very first, I implemented both sqlite3 and files based approaches, see reasoning below. And left the original behaviour just for (performance) comparison (it must be run with `-j 1`).

First, mappings are very independent objects that do maintain their dataset on their own. Even adding an item to obfuscate (say, FQDN) means the dataset is updated a few times (by the host, FQDN and/or domain). I tried to respect this as I like that independency (and also changing it would require a lot of changes).

Also, the dataset is growing in time (as we discover more domains or IP networks) , which means some instances of lately discovered domain might not be obfuscated or that changing ordering of files to obfuscate result in *different* final mapping (also in the size of the final map). You can check it by yourself if you reorder the list from `get_file_list`. So running cleaner concurrently - which means dataset being populated non-deterministically - can end up in different final map. Dont be confused by this as I was.

I chose individual processes to sync over the pieces of information "we obfuscate item X as the first one, and item Y as the next". An option to exchange or sync on the whole mapping/datasets would mean 1) more data to sync and 2) altering the mapping classes "independent" behaviour.

This also means there are gaps in numbers and it is fine. The ordering number is the size of the dataset, which can be incremented more than by one when adding just one item. This is fine, as each and every process replays the same and adds the same stuff into its dataset. And also this allows a smart replay of the whole "how was the dataset created?" process at the end - see the `archive.load_parser_entries()` call.

Then, ProcessPoolExecutor has two limitations for us that I described directly in the code, plus one substantial one: passing the whole `SoSCleaner` class is not easily feasible. Doing so would prevent some code movement among classes and would be nicer to hook new code in, but it does not work easily. Trying that, I hit issues like "oh, cloning `SoSCleaner` into a child process means cloning there sos arguments parsing that overrides some __* method that blocks successful process spawn and iteration over a list". It is possible to fix or hack those traps in our code, but after iteratively doing that for five such traps, I gave up this rabbit hole journey.

Please, consider this as a draft only - see the number of TODO points, plus various methods need a better name. Anyway the code is functionally ready, works well(*) and scales fairly well.

(*) the only concern is that sqlite DB can lock itself. I did a few changes there, but still I seldomly hit a live-lock behaviour over locked DB on some artificial test cases (have 8 identical files each with 100 unique IP addresses inside a sosreport, and run cleaner with `-j 8` is my "favourite" one).

While I like the sqlite approach more (it looks so professional!), we cant choose it until we fix these live-locks / DB locks.
Gladly, the file-based approach works smoothly and provides comparable (or even slightly better) performance.

When speaking about performance, some benchmark tests are running. Results from 8cores RHEL10 beta, median time from 3 runs each:

a small, 11MB packed sosreport:
- current cleaner run 0m59.799s
- sqlite-based cleaner run 0m59.956s (j=1), 0m34.710s (j=2), 0m22.761s (j=4) and 0m14.137s (j=8)
- files-based cleaner run 0m59.760s (j=1), 0m34.594s (j=2), 0m22.650s (j=4) and 0m14.111s (j=8)

220MB packed sosreport:
- current cleaner run 42m2.282s
- sqlite-based cleaner run 42m4.665s (j=1), 23m17.864s (j=2), 19m17.219s (j=4) and 10m26.143s (j=8)
- files-based cleaner run 42m0.994s (j=1), 23m12.987s (j=2), 19m23.130s (j=4) and 10m26.608s (j=8)

350MB packed sosreport:
- current cleaner run 65m30.090s
- sqlite-based cleaner run 66m59.097s (j=1), 42m5.778s (j=2), 29m3.550s (j=4) and 25m45.191s (j=8)
- files-based cleaner run 66m51.753s (j=1), 41m39.086s (j=2), 28m21.102s (j=4) and 24m20.798s (j=8)

770MB packed sosreport:
- current cleaner run 794m0.238s - yes, over 12 hours
- sqlite-based cleaner run 794m47.511s (j=1), 436m9.563s (j=2), 219m25.159s (j=4) and 164m10.205s (j=8) - speeded up from 12h to 2.5h
- files-based cleaner run 793m45.589s (j=1), 432m25.081s (j=2), 218m22.004s (j=4) and 165m10.775s (j=8) - again speeded up from 12h to 2.5h

1GB packed sosreport:
- current cleaner run 66m17.037s
- sqlite-based cleaner run 66m14.740s (j=1), 48m7.181s (j=2), 26m5.231s (j=4) and 23m20.612s (j=8)
- files-based cleaner run 65m56.254s (j=1), 47m57.850s (j=2), 25m50.313s (j=4) and 23m19.893s (j=8)

Closes: #3097

---

### DIFF
--- a/sos/cleaner/__init__.py
+++ b/sos/cleaner/__init__.py
@@ -13,10 +13,9 @@ import json
 import logging
 import os
 import shutil
-import tempfile
 import fnmatch
 
-from concurrent.futures import ThreadPoolExecutor
+from concurrent.futures import ProcessPoolExecutor
 from datetime import datetime
 from pwd import getpwuid
 
@@ -35,8 +34,12 @@ from sos.cleaner.archives.sos import (SoSReportArchive, SoSReportDirectory,
                                       SoSCollectorDirectory)
 from sos.cleaner.archives.generic import DataDirArchive, TarballArchive
 from sos.cleaner.archives.insights import InsightsArchive
-from sos.utilities import (get_human_readable, import_module, ImporterHelper,
-                           file_is_binary)
+from sos.utilities import (get_human_readable, import_module, ImporterHelper)
+
+
+# an auxiliary method to kick off child processes over its instances
+def obfuscate_arc_files(arc, flist):
+    return arc.obfuscate_arc_files(flist)
 
 
 class SoSCleaner(SoSComponent):
@@ -62,7 +65,8 @@ class SoSCleaner(SoSComponent):
     In the case of IP addresses, support is for IPv4 and IPv6 - effort is made
     to keep network topology intact so that later analysis is as accurate and
     easily understandable as possible. If an IP address is encountered that we
-    cannot determine the netmask for, a random IP address is used instead.
+    cannot determine the netmask for, a private IP address from 172.17.0.0/22
+    range is used instead.
 
     For IPv6, note that IPv4-mapped addresses, e.g. ::ffff:10.11.12.13, are
     NOT supported currently, and will remain unobfuscated.
@@ -111,8 +115,9 @@ class SoSCleaner(SoSComponent):
             self.policy = hook_commons['policy']
             self.manifest = hook_commons['manifest']
             self.from_cmdline = False
+            # precede 'report -t' option above 'cleaner --jobs'
             if not hasattr(self.opts, 'jobs'):
-                self.opts.jobs = 4
+                self.opts.jobs = self.opts.threads
             self.opts.archive_type = 'auto'
             self.soslog = logging.getLogger('sos')
             self.ui_log = logging.getLogger('sos_ui')
@@ -129,14 +134,20 @@ class SoSCleaner(SoSComponent):
 
         self.cleaner_md = self.manifest.components.add_section('cleaner')
 
-        skip_cleaning_files = self.opts.skip_cleaning_files
+        cleaner_dir = os.path.dirname(self.opts.map_file) \
+            if self.opts.map_file else '/etc/sos/cleaner'
+        parser_args = [
+            self.cleaner_mapping,
+            cleaner_dir,
+            self.opts.skip_cleaning_files,
+        ]
         self.parsers = [
-            SoSHostnameParser(self.cleaner_mapping, skip_cleaning_files),
-            SoSIPParser(self.cleaner_mapping, skip_cleaning_files),
-            SoSIPv6Parser(self.cleaner_mapping, skip_cleaning_files),
-            SoSMacParser(self.cleaner_mapping, skip_cleaning_files),
-            SoSKeywordParser(self.cleaner_mapping, skip_cleaning_files),
-            SoSUsernameParser(self.cleaner_mapping, skip_cleaning_files)
+            SoSHostnameParser(*parser_args),
+            SoSIPParser(*parser_args),
+            SoSIPv6Parser(*parser_args),
+            SoSMacParser(*parser_args),
+            SoSKeywordParser(*parser_args),
+            SoSUsernameParser(*parser_args),
         ]
 
         for _parser in self.opts.disable_parsers:
@@ -308,14 +319,17 @@ third party.
             check_type = self.opts.archive_type.replace('-', '_')
             for archive in self.archive_types:
                 if archive.type_name == check_type:
-                    _arc = archive(self.opts.target, self.tmpdir)
+                    _arc = archive(self.opts.target, self.tmpdir,
+                                   self.opts.keep_binary_files)
         else:
             for arc in self.archive_types:
                 if arc.check_is_type(self.opts.target):
-                    _arc = arc(self.opts.target, self.tmpdir)
+                    _arc = arc(self.opts.target, self.tmpdir,
+                               self.opts.keep_binary_files)
                     break
         if not _arc:
             return
+        self.main_archive = _arc
         self.report_paths.append(_arc)
         if _arc.is_nested:
             self.report_paths.extend(_arc.get_nested_archives())
@@ -523,7 +537,7 @@ third party.
                 logfile.write(line)
 
         if archive:
-            self.obfuscate_file(log_name)
+            self.obfuscate_file(log_name, short_name="sos_logs/cleaner.log")
             self.archive.add_file(log_name, dest="sos_logs/cleaner.log")
 
     def get_new_checksum(self, archive_path):
@@ -554,7 +568,8 @@ third party.
         try:
             msg = (
                 f"Found {len(self.report_paths)} total reports to obfuscate, "
-                f"processing up to {self.opts.jobs} concurrently\n"
+                f"processing up to {self.opts.jobs} concurrently within one "
+                "archive\n"
             )
             self.ui_log.info(msg)
             if self.opts.keep_binary_files:
@@ -562,9 +577,9 @@ third party.
                     "WARNING: binary files that potentially contain sensitive "
                     "information will NOT be removed from the final archive\n"
                 )
-            pool = ThreadPoolExecutor(self.opts.jobs)
-            pool.map(self.obfuscate_report, self.report_paths, chunksize=1)
-            pool.shutdown(wait=True)
+            for report_path in self.report_paths:
+                self.ui_log.info(f"Obfuscating {report_path.archive_path}")
+                self.obfuscate_report(report_path)
             # finally, obfuscate the nested archive if one exists
             if self.nested_archive:
                 self._replace_obfuscated_archives()
@@ -635,6 +650,8 @@ third party.
 
             for ritem in prepper.regex_items[pname]:
                 _parser.mapping.add_regex_item(ritem)
+        # we must initialize stuff inside (cloned processes') archive - REALLY?
+        archive.set_parsers(self.parsers)
 
     def get_preppers(self):
         """
@@ -668,8 +685,9 @@ third party.
 
         Positional arguments:
 
-            :param report str:      Filepath to the directory or archive
+            :param archive str:      Filepath to the directory or archive
         """
+
         try:
             arc_md = self.cleaner_md.add_section(archive.archive_name)
             start_time = datetime.now()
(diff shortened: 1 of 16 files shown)


pmoravec:
How to get the sqlite DB locks:

In an unpacked sosreport:

    for i in $(seq 1 10); do
      for j in $(seq 1 10); do
        echo "${i}.${j}.${i}.${j}"
      done
    done > hunderd_of_IPs.0.txt
    
    for i in $(seq 1 8); do
      cp hunderd_of_IPs.0.txt hunderd_of_IPs.${i}.txt

And run `sos clean -j 8 --concurrency-backend sql <DIR>`  (`-j 4` should get the same outcome as well with smaller probability).

TurboTurtle (commented):
From a pure review standpoint, I like this approach and the code changes overall. We'd want to ensure that the `--jobs` option gets exposed to `report` for leveraging in-line cleaning, but that can come later.

I am unsure of the need to carry both `files` and `sqlite` as the concurrency mechanism though. Especially if the results are largely the same. I have a preference to only carry `files`, but am certainly open to hearing why we should prefer the `sqlite` approach (if we only carry one forward).

That said, on python 3.13 (my Fedora daily driver install), I get this error when trying to clean an archive:

```
sosreport-terra-2025-04-22-xicbmuo :               Beginning obfuscation...
Exception while processing sosreport-terra-2025-04-22-xicbmuo: cannot pickle 'BufferedReader' instances
No reports obfuscated, aborting...
```

and it is not immediately obvious to me where that `BufferedReader` is coming into play here but I suspect it is from the `ProcessPoolExecutor` instantiation.

What python version(s) are you able to get successful executions on?

pmoravec:
One more bug:

      File "/path/to/sos-pmoravec-concurrent-cleaner-TESTING/sos/cleaner/mappings/__init__.py", line 135, in add
        os.link(tmpfile.name, os.path.join(self.link_dir, f"{counter}"))
    OSError: [Errno 18] Invalid cross-device link: '/tmp/tmp9emqdshx' -> '/etc/sos/cleaner/cleaner_links/soshostnamemap/1'

Because `/tmp` is on a different lvolume/device than /etc. The tmpfile must be under `self.link_dir` (plus minus).

sos/cleaner/parsers/username_parser.py
```diff
@@ -26,8 +26,8 @@ class SoSUsernameParser(SoSCleanerParser):
     map_file_key = 'username_map'
     regex_patterns = []
 
-    def __init__(self, config, skip_cleaning_files=[]):
-        self.mapping = SoSUsernameMap()
+    def __init__(self, config, skip_cleaning_files=[], workdir='/tmp'):
```
pmoravec: The `workdir=`/tmp` is here due to https://github.com/sosreport/sos/blob/main/tests/unittests/cleaner_tests.py#L33-L38 where we need to pass some value.

Calls from the cleaner itself always use `os.path.dirname(self.opts.map_file)` so we just need some safe fallback for tests.

I am not sure if `/tmp` is a good dir for all distros, and if we should set it here or in the tests - let me know your thoughts.

arif-ali: it's ok from the Debian/Ubuntu side, as we use `/tmp` already for all our configs as a patch (although I'll be changing that in the policy instead)

jcastill:
I've run some quick tests, and the results are looking good. In RHEL 10 (but I'm testing 8 and 9 as well) here's an example, with 3000 IP addresses on dummy network devices and the same number of entries in /etc/hosts. Posting here two runs of tests but I've run around 20 with very similar results.

RHEL 10 with new code (i.e. the code in this PR):

Test 1:

```
real	20m31.553s
user	34m53.220s
sys	0m2.397s
```

Test 2:

```
real	21m38.153s
user	36m25.341s
sys	0m2.380s
```


RHEL 10 with current code (i.e. the one that exists in SoS at the moment):

Test 1:

```
real	32m37.409s
user	32m27.055s
sys	0m11.651s
```

Test 2:

```
real	31m57.659s
user	31m47.392s
sys	0m2.013s
```

I'll run the same tests in RHEL 8, 9, and Fedora 42 over the weekend. Also no errors or issues at all with the code proposed here.
@pmoravec did you get similar results in your tests?

sos/cleaner/mappings/__init__.py
```diff
@@ -28,11 +29,33 @@ class SoSMap():
     ignore_short_items = False
     match_full_words_only = False
 
-    def __init__(self):
-        self.dataset = {}
+    def __init__(self, workdir='/tmp'):
+        self.dataset = {}  # TODO: mention the child classes can update it as
+        # well as add_sanitised_item_to_dataset method, so the files dont need
```
jcastill: Perhaps it may be better to add this comment on top of self.dataset = {} instead of starting at the end of the line and then continuing below?

sos/cleaner/mappings/__init__.py
```diff
@@ -28,11 +29,33 @@ class SoSMap():
...
+        # Our approach guarantees same input is added in all processes in the
+        # same ordering - it can add a new network to the dataset.
+        # So dataset bumps by 2
         self._regexes_made = set()
         self.compiled_regexes = []
-        self.lock = Lock()
+        self.cname = self.__class__.__name__.lower()
+        # TODO: ensure the directory <workdir> does exist
+        # TODO: deal with permissions..? IMHO default ones are OK
+        self.workdir = workdir  # the default value is used just by avocado
```
jcastill: Same as above - it may be clearer to add the comment on top (maybe adding a line between the last # TODO and the one in this line)

sos/cleaner/parsers/username_parser.py
jcastill: When I run avocado tests in rhel, I get them placed in /var/tmp. I imagine it uses the _tmp_dir inside policy. 
Since @arif-ali has added now the same var in the debian policy via e4c50c2526f5f2a83a0302c5bb986ca14e00ecf9, maybe it would be better to use policy.get_tmp_dir() to get the right tmp directory depending on distro?

arif-ali: We had the config option specific in the sos.conf it was being placed in `/tmp` already. I only just moved it in the polict instead, which made more sense.

I'd agree, we ought to use the policy configuration (if we can), then you know it will be consistent based on those parameters rather than a static entry on the code.

pmoravec: So we would need to load Policy and its `get_tmp_dir()` as a fallback. Sadly, neither `SoSCleanerParser` or cleaner tests have direct access to Policy - I will enhance tests accordingly.

pmoravec:
The failing tests are due to missing rebase of my PR to have the Debian default workdir commit. Will do tomorrow.

tests/cleaner_tests/ipv6_test/ipv6_test.py
```diff
@@ -8,7 +8,7 @@
 
 from sos_tests import StageTwoReportTest
 
-MOCK_FILE = '/tmp/sos-test-ipv6.txt'
+MOCK_FILE = '/sos-test-ipv6.txt'
```
pmoravec: This change (and same in `ipv6_test.py` is tricky. Some previous test obfuscated `tmp` into `obfuscatedword1`. We replaced `/etc/sos/cleaner/default_mapping` by empty file, yet the cleaner's cache kept the mapping. So `self.assertFileCollected(MOCK_FILE)` fails as the filename is changed by cleaner in the tarball.

pmoravec:
Various tests are failing the same way I fixed one instance of that failure in https://github.com/sosreport/sos/pull/3988#discussion_r2155523518 :

- https://github.com/sosreport/sos/blob/main/tests/cleaner_tests/existing_archive.py#L127 adds `tmp` to a keyword mapper, which is stored in static `/etc/sos/cleaner/default_mapping` and also "transient" `[/var|]/tmp/sos_avocado_tests/cleaner_tests/soskeywordmap` directory.
- various tests remove `/etc/sos/cleaner/default_mapping`  but not the `cleaner_cache` dir, so they load `var -> obfuscatedword0` and `tmp -> obfuscatedword1` mapping
- so the tests replace `tmp` dir inside the archives they work on, causing errors like `/var/tmp/avocado_yhobmqaosos_tests.py.rhbz1950350/sosreport-rhbz1950350/sosreport-host1-2025-06-18-ryovfyo/var/log/clean_config_test.txt does not exist`

How to resolve it? A few ideas:

- restrict the https://github.com/sosreport/sos/blob/main/tests/cleaner_tests/existing_archive.py#L127 just for `avocado` keyword
- use `--tmp-dir /var/tmp/foo/bar/` there and test for `foo,bar,avocado` keywords (or something more clever but unique than `foo` and `bar`, ofc - simply something we wont see inside any tested archive)
- remove `[/var|]/tmp/sos_avocado_tests/cleaner_cache` within tests (either after the `existing_archive.py#L127` execution, or in `setUp` phase of any potentially affected test) - but such removal can step on other running tests toes that work over already built `cleaner_cache` dir
- put the auxiliary testing files directly to `/` instead of `/tmp` or `/var/*`, like I did for the `ipv6` case
- enhance `cleaner` code by some routine to safely remove `cleaner_cache` when nobody uses it (needs some locking mechanism, doable). And tests will call that removal routine whenever they purge away `default_mapping`

Any other idea? Any preferences? I think the 1st or 2nd option is less painful.

pmoravec:
After some offline discussion, the best is:

- short-term solution in change from `--keywords var,tmp,avocado` to `--tmp-dir /var/tmp/soscleanertest/banana --keywords avocado,banana,soscleanertest` in https://github.com/sosreport/sos/blob/main/tests/cleaner_tests/existing_archive.py#L127
- long-term solution to have some method to safely remove `cleaner_cache` when no cleaner is running, and use this method in tests setUp

Until there will be another idea, I will follow the short-term solution in next few days.

pmoravec:
IMHO that traceback can not happen in concurrent cleaner where I had to catch all exceptions to prevent child process termination (https://github.com/sosreport/sos/pull/3988/files#diff-64f7cecab583ab4d9d9a3d4d4d89fda30a4b0e5b8e9371cc8a156bdb626e2d0fR209-R211). In such a case, no sanitisation happens for the given line and parser.

But we should investigate also the root cause of infinite recursion - that sounds like generic problem that my code would rather hide than resolve.

pmoravec:
Hi mainly @TurboTurtle but also others - can you please do a tech review of the codechanges as well?

I would like to move this PR further / have it merged.

sos/cleaner/__init__.py
```diff
@@ -111,8 +115,10 @@ def __init__(self, parser=None, args=None, cmdline=None, in_place=False,
             self.policy = hook_commons['policy']
             self.manifest = hook_commons['manifest']
             self.from_cmdline = False
+            # precede 'report -t' option above 'cleaner --jobs' above default 4
             if not hasattr(self.opts, 'jobs'):
-                self.opts.jobs = 4
+                self.opts.jobs = self.opts.threads if hasattr(self.opts,
+                                                              'threads') else 4
```
TurboTurtle: We should always have _some_ `threads` value here, right? So if we're not setting jobs we can just set jobs equal to threads, and skip the `hasattr()`.

sos/cleaner/mappings/ip_map.py
```diff
@@ -162,13 +166,12 @@ def sanitize_ipaddr(self, addr):
...
-                _octets.append(random.randint(11, 99))
-            return f"{_octets[0]}.{_octets[1]}.{_octets[2]}.{_octets[3]}"
-
-        _addr = _gen_address()
+        # increment the counter and ignore *.0 and *.255 addresses
+        self._saddr_cnt += 1
+        while self._saddr_cnt % 256 in (0, 255):
+            self._saddr_cnt += 1
+        _addr = f"{self._saddr_cnt >> 24}.{(self._saddr_cnt >> 16) % 256}." \
+            f"{(self._saddr_cnt >> 8) % 256}.{self._saddr_cnt % 256}"
```
TurboTurtle: This makes sense overall, but it does take a moment to process it logically, at least for me. A short comment would I think help for future code reviews.

sos/cleaner/mappings/mac_map.py
```diff
@@ -20,8 +19,9 @@ class SoSMacMap(SoSMap):
     MAC addresses added to this map will be broken into two halves, vendor and
     device like how MAC addresses are normally crafted. For the vendor hextets,
     obfuscation will take the form of 53:4f:53, or 'SOS' in hex. The following
-    device hextets will be randomized, for example a MAC address of
-    '60:55:cb:4b:c9:27' may be obfuscated into '53:4f:53:79:ac:29' or similar
+    device hextets will be obfuscated by a series of suffixes starting from
+    zeroes. For example a MAC address of '60:55:cb:4b:c9:27' may be obfuscated
+    into '53:4f:53:00:00:1a' or similar.
```
TurboTurtle, line 24: ack to making this serialized for mac assignments.

pmoravec:
> I'm having some inconsistencies in performance. I've had some executions that run better than current `main`, and others that sit for 15+ minutes, using default values.
> 
> I've not been able to track down what's happening here as of yet.

Can you provide the sosreport you get stuck cleaner on, so I can investigate? If you run cleaner with `-vvv`, does it print to stdout `pid=12345: obfuscating /var/tmp/sosreport-foo...`?

pmoravec:
The new version contains the two Jake's comments applied.

TurboTurtle:
> Can you provide the sosreport you get stuck cleaner on, so I can investigate? If you run cleaner with `-vvv`, does it print to stdout `pid=12345: obfuscating /var/tmp/sosreport-foo...`?

Generally it would hang on the system journal collection, but I've not been able to reproduce the issue since then.

LGTM, would like to get any feedback from @arif-ali before merge, but also want to have this in before our release in August either way.

PR #3993, opened 2025-04-17
[tests] Add report output format tests

tiredPotato:
Adds a few tests to check the output format of "--list-presets" and "--list-plugins" related to the changes in https://github.com/sosreport/sos/pull/3762 and a fix to an existing test (which seemed to be just testing whether a line contains more than a single character instead of checking whether a description is missing).

---

### DIFF
--- a/tests/report_tests/help_output_tests.py
+++ b/tests/report_tests/help_output_tests.py
@@ -71,7 +71,7 @@ class ReportListPluginsTest(StageOneOutputTest):
         for plug in disabled_plugins:
             if not plug.strip():
                 continue
-            self.assertRegex(plug, r' ([\S ]){20} (inactive[ ]{6}) ([\S ])*')
+            self.assertRegex(plug, r' ([\S ]){30} (inactive[ ]{6}) ([\S ])*')
         for opt in options:
             if not opt.strip():
                 continue
@@ -79,7 +79,7 @@ class ReportListPluginsTest(StageOneOutputTest):
         for opt in plugin_options:
             if not opt.strip():
                 continue
-            self.assertRegex(opt, r' ([\S ]){30} ([\S ]{15}) ([\S ])*')
+            self.assertRegex(opt, r' ([\S ]){35} ([\S ]{15}) ([\S ])*')
 
 
 class ReportListPresetsTest(StageOneOutputTest):
--- a/sos/report/__init__.py
+++ b/sos/report/__init__.py
@@ -1041,7 +1041,7 @@ class SoSReport(SoSComponent):
                                "disabled:"))
             self.ui_log.info("")
             for (plugname, plugclass, reason) in self.skipped_plugins:
-                self.ui_log.info(f" {plugname:<20} {reason:<14} "
+                self.ui_log.info(f" {plugname:<30} {reason:<14} "
                                  f"{plugclass.get_description()}")
 
         self.ui_log.info("")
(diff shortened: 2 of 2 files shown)


tests/report_tests/help_output_tests.py
```diff
@@ -45,7 +45,62 @@ def test_no_missing_plugin_descriptions(self):
                          "The following plugins are currently disabled:",
                          self.cmd_output.stdout, re.S).group(1).splitlines()
         for ln in _out:
+            ln = ln.split()
```
pmoravec, line 48: Nice!

tests/report_tests/help_output_tests.py
```diff
@@ -45,7 +45,62 @@ def test_no_missing_plugin_descriptions(self):
...
+                         self.cmd_output.stdout, re.S)
+        enabled_plugins = _out.group(1).splitlines()
+        disabled_plugins = _out.group(2).splitlines()
+        options = _out.group(3).splitlines()
+        plugin_options = _out.group(4).splitlines()
+        for plug in enabled_plugins:
+            # Ignore empty lines
+            if not plug.strip():
+                continue
+            self.assertRegex(plug, r' ([\S ]){20} ([\S ])*')
```
pmoravec, line 70: Should we really test each line's *formatting*? On one side, such a test is too pedantic / nitpick one, on the other side it can reveal unexpected TUI issues.

And well, since we already had some formatting issues (#3762 or #3984), the test does make sense.

pmoravec (approved):

tests/report_tests/help_output_tests.py
```diff
@@ -45,7 +45,62 @@ def test_no_missing_plugin_descriptions(self):
...
+        for opt in options:
+            if not opt.strip():
+                continue
+            self.assertRegex(opt, r' ([\S ]){25} ([\d ]{15}) ([\S ])*')
+        for opt in plugin_options:
+            if not opt.strip():
+                continue
+            self.assertRegex(
+                opt,
+                r' ([\S ]){30} (on[ ]{13}|off[ ]{12}|[\d ]{15}) ([\S ])*'
```
pmoravec, line 84: The `(on[ ]{13}|off[ ]{12}|[\d ]{15})` does not match `mysql` or other string values of plugopts - see failing match on `' mysql.dbuser                   mysql           username for database dump collection'`.

Since plugin options default values can be arbitrary, I would simplify this test to "any string with optional trailing spaces of length (??) chars altogether".

tests/report_tests/help_output_tests.py
tiredPotato: Yes, I was debating on this with myself for really long time and I came to the same decision that testing each line makes sense since it already caught some formatting issues. 

For example some checks are failing because there is `canonical_livepatch_onprem` in disabled plugins that should be maximum of 20 chars which is obviously not (it has 26 chars). Other columns in that line are shifted by 6 chars because of that. 

Also not sure what to do with `containers_common.rootlessusers` in plugin options. We've already discussed that shift by 1 char is not so bad - however, relaxing the regex will mostly bypass what the test was originally meant to test. Any ideas?

pmoravec: I agree the test makes sense at it can prevent issues like those two. Relaxing it for the sake of formatting changes to pass tests would invalidate the purpose of the test - simply, the "shift by 1" code change must be (in future) followed by equivalent change in tests.

pmoravec:
The failing rpmbuild "tests" are due to missing https://github.com/sosreport/sos/commit/2780b0909fede8b4f725e37090aecae82f44ef6a .

pmoravec (approved):

pmoravec (approved):
Thanks!

PR #3999, opened 2025-04-29
[system] add system proxy check

reidliu41:
---

Based on the experience, the proxy would caused lots of issues(e.g cannot register, cannot dnf, curl...etc.), 
it would set in different conf(rhsm.conf, dnf.conf..) for different services, but the system setting(e.g profile, or simply export) also work, currently sos just collect variable name under system plugin, so it would better to collect proxy (e.g. https_proxy=xxx) individually, and make it easy check with `root_symlink`.
```
# ll
system-proxy-setting -> sos_commands/system/system_proxy_setting

e.g.
# cat system-proxy-setting
No system proxy settings found.

# cat system-proxy-setting
http_proxy=http://********@proxy.example-test-test.com:8080
https_proxy=http://********@proxy.example-test-test.com:8080

# cat system-proxy-setting
http_proxy=http://proxy.example-test-test.com:8080
https_proxy=http://proxy.example-test-test.com:8080
```

### DIFF
--- a/sos/report/plugins/crio.py
+++ b/sos/report/plugins/crio.py
@@ -38,13 +38,6 @@ class CRIO(Plugin, RedHatPlugin, UbuntuPlugin, CosPlugin):
             "/etc/sysconfig/crio-*"
         ])
 
-        self.add_env_var([
-            'HTTP_PROXY',
-            'HTTPS_PROXY',
-            'NO_PROXY',
-            'ALL_PROXY'
-        ])
-
         self.add_cmd_output([
             "crio config"
         ])
--- a/sos/report/__init__.py
+++ b/sos/report/__init__.py
@@ -27,7 +27,7 @@ from shutil import rmtree
 import sos.report.plugins
 from sos.utilities import (ImporterHelper, SoSTimeoutError, bold,
                            sos_get_command_output, TIMEOUT_DEFAULT, listdir,
-                           is_executable)
+                           is_executable, scrub_url_credential)
 
 from sos import _sos as _
 from sos import __version__
@@ -71,6 +71,12 @@ def _format_since(date):
 
 # valid modes for --chroot
 chroot_modes = ["auto", "always", "never"]
+
+# collect default env vars
+default_env_vars = {
+    'http_proxy', 'https_proxy', 'HTTP_PROXY', 'HTTPS_PROXY',
+    'NO_PROXY', 'ALL_PROXY'
+}
 
 
 class SoSReport(SoSComponent):
@@ -151,7 +157,7 @@ class SoSReport(SoSComponent):
         self.loaded_plugins = []
         self.skipped_plugins = []
         self.all_options = []
-        self.env_vars = set()
+        self.env_vars = default_env_vars
         self._args = args
         self.sysroot = "/"
         self.estimated_plugsizes = {}
@@ -1422,7 +1428,7 @@ class SoSReport(SoSComponent):
         if not self.env_vars:
             return
         env = '\n'.join([
-            f"{name}={val}" for (name, val) in
+            f"{name}={scrub_url_credential(val)}" for (name, val) in
             [(name, f'{os.environ.get(name)}') for name in self.env_vars if
              os.environ.get(name) is not None]
         ]) + '\n'
--- a/sos/report/plugins/docker.py
+++ b/sos/report/plugins/docker.py
@@ -35,10 +35,6 @@ class Docker(Plugin, CosPlugin):
         ])
 
         self.add_env_var([
-            'HTTP_PROXY',
-            'HTTPS_PROXY',
-            'NO_PROXY',
-            'ALL_PROXY',
             'DOCKER_BUILD_PROXY',
             'DOCKER_RUN_PROXY'
         ])
--- a/sos/report/plugins/foreman.py
+++ b/sos/report/plugins/foreman.py
@@ -304,14 +304,6 @@ class Foreman(Plugin):
                                         subdir='smart_proxies_features',
                                         timeout=10)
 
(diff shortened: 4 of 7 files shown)


pmoravec:
In general, +1 for the PR. Some minor nit-picks commented.

sos/report/plugins/system.py
```diff
@@ -47,6 +48,42 @@ def setup(self):
...
                                 plug_dir=True)
 
+        sys_proxy_content = self._check_system_proxy_setting()
+
+        self.add_cmd_output(f'echo "{sys_proxy_content}"',
+                            suggest_filename="system_proxy_setting",
+                            root_symlink="system-proxy-setting",
+                            tags="sys_proxy")
+
+    def _check_system_proxy_setting(self):
```
pmoravec, line 58: It's rather `_get` than `_check` of `_system_proxy_setting`, no?

sos/report/plugins/system.py
```diff
@@ -47,6 +48,42 @@ def setup(self):
                                 "environment_varnames",
                                 plug_dir=True)
 
+        sys_proxy_content = self._check_system_proxy_setting()
+
+        self.add_cmd_output(f'echo "{sys_proxy_content}"',
+                            suggest_filename="system_proxy_setting",
+                            root_symlink="system-proxy-setting",
+                            tags="sys_proxy")
```
pmoravec, line 56: Isnt `system_proxy` morintuitive?

sos/report/plugins/system.py
```diff
@@ -47,6 +48,42 @@ def setup(self):
...
+            var: os.environ.get(var)
+            for var in sys_proxy_env_vars
+            if os.environ.get(var)
+        }
+
+        def scrub_username_password(url):
+            """
+            Replace username:password@ with ********@ in proxy URL if present
+            """
+            return re.sub(r'//([^/@]+)@', '//********@', url)
```
pmoravec, line 75: We have obfuscation of proxy URL credentials on several places of code already. Isn't it worth unifying it via one auxiliary method, @TurboTurtle ?

sos/report/plugins/system.py
```diff
@@ -47,6 +48,42 @@ def setup(self):
...
+
+        if sys_proxy_settings:
+            sys_proxy_content = "\n".join(
+                f"{var}={scrub_username_password(value)}"
+                for var, value in sys_proxy_settings.items()
+            )
+        else:
+            sys_proxy_content = "No system proxy settings found."
```
pmoravec, line 83: No strong preference here either. Isnt empty output better? It's meaning is almost the same like with this text, but it allows easier tooling which can rely on "every line has a proxy setting, no comments"?

But some people can see an empty file as confusing..?

reidliu41: seems fine for empty and up to users. I see `/etc/environment` always empty.

pmoravec:
Thx. Please squash the two commits into one, otherwise good!

TurboTurtle:
This is already captured in the top-level `environment` file [0]. Plugins just need to specify the env var name - the various container plugins as well as the `foreman` plugin do this [1].

Previously we talked about always capturing these two specifically, and it was shot down. If we now want to change that, then we'd just need to add the call to a relevant plugin (either `system` is fine or perhaps `networking).

[0] https://github.com/sosreport/sos/blob/main/sos/report/__init__.py#L1421

[1] https://github.com/search?q=repo%3Asosreport%2Fsos%20HTTP_PROXY&type=code

reidliu41:
hi, thank you for your reply， just tested `add_env_var`,  found two things
- it will not handle the password issue
- most of them are duplicate, maybe it would be better to just write one and maintain？？https://github.com/search?q=repo%3Asosreport%2Fsos%20HTTP_PROXY&type=code
```
# cat environment
https_proxy=http://test:testpass@proxy.example.com:8080  <<<---
http_proxy=http://test:testpass@proxy.example.com:8080    <<<---
HTTP_PROXY=http://proxy.example.com:8080
HTTPS_PROXY=http://proxy.example.com:8080
DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/0/bus
PATH=/usr/sbin:/usr/bin:/root/bin:/usr/local/bin:/usr/local/sbin
XDG_RUNTIME_DIR=/run/user/0
XDG_SESSION_ID=11
TERM=xterm-256color
```

reidliu41:
and the another issue is that the system wide proxy  might impact the whole system, not just one or two plugins.

reidliu41:
like system can collect system level, plugin can collect their own...

sos/report/plugins/crio.py
```diff
@@ -38,13 +38,6 @@ def setup(self):
             "/etc/sysconfig/crio-*"
         ])
 
-        self.add_env_var([
-            'HTTP_PROXY',
-            'HTTPS_PROXY',
-            'NO_PROXY',
-            'ALL_PROXY'
-        ])
-
```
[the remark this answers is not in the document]
TurboTurtle: As a follow up here, we could take these specific vars out of plugin control altogether - just make them default collections by making them default elements in `Report.env_vars` and then putting the above mentioned scrubbing pass into `Report.collect_env_vars()`.

reidliu41:
thank you for you feedback, just updated it

reidliu41:
hi team, can someone help to take a look?

pmoravec (approved):
I like it!

We already have a similar method https://github.com/sosreport/sos/blob/main/sos/report/plugins/__init__.py#L1320 but I think it would be hard to combine/merge them to unify the URL credentials obfuscation..?

I am ok with the PR as it is, just an idea.

reidliu41:
thanks

arif-ali (approved):

PR #4002, opened 2025-05-06
[authd] Implement authd plugin

### LINKED ISSUE #3939: `authd`
We're starting to see support cases for Ubuntu [authd](https://github.com/ubuntu/authd) at Canonical Support; I don't see a plugin for it so I'm opening this issue to track it. I'd like to put together a first draft but I don't have an ETA. Please assign to me but feel free to reassign if somebody has time to work on this.

cc @arif-ali 🙂

MggMuggins:
Fixes #3939

The secrets included in the tests are fake; I generated all the components locally. They are syntactically equivalent to the real secrets I was able to find online.

I've also included a commit to fix the `{debian,ubuntu,redhat}_only` test decorators. My first iteration of the tests passed even though the test called an assert function that didn't exist. The decorators were introduced in f2e99b89d and looking at that change it makes sense how this was missed.

---

### DIFF
--- a/sos/report/__init__.py
+++ b/sos/report/__init__.py
@@ -453,7 +453,7 @@ class SoSReport(SoSComponent):
         ssec.add_text(helpln)
 
     def print_header(self):
-        print(f"\n{_(f'sos report (version {__version__})')}\n")
+        self.ui_log.info(f"\n{_(f'sos report (version {__version__})')}\n")
 
     def _get_hardware_devices(self):
         self.devices = {
--- a/sos/report/plugins/authd.py
+++ b/sos/report/plugins/authd.py
@@ -1,0 +1,74 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, UbuntuPlugin
+
+
+# See the docs: https://documentation.ubuntu.com/authd/
+class Authd(Plugin, UbuntuPlugin):
+
+    short_desc = 'Authd daemon & broker information'
+
+    plugin_name = 'authd'
+
+    apt_packages = (
+        'authd',
+    )
+
+    snap_packages = (
+        'authd-msentraid',
+        'authd-google',
+    )
+
+    packages = apt_packages + snap_packages
+
+    services = (
+        'authd',
+        'snap.authd-msentraid.authd-msentraid',
+        'snap.authd-google.authd-google',
+    )
+
+    def setup(self):
+        self.add_dir_listing([
+            "/etc/authd/brokers.d",
+        ])
+
+        self.add_copy_spec([
+            "/etc/authd/brokers.d/msentraid.conf",
+            "/etc/authd/brokers.d/google.conf",
+            "/var/snap/authd-google/current/broker.conf",
+            "/var/snap/authd-google/current/broker.conf.d/*",
+            "/var/snap/authd-msentraid/current/broker.conf",
+            "/var/snap/authd-msentraid/current/broker.conf.d/*",
+        ])
+
+        self.add_cmd_output([
+            f"apt-cache policy {' '.join(self.apt_packages)}",
+            f"snap list --all {' '.join(self.snap_packages)}",
+            "/usr/libexec/authd version",
+        ])
+
+    def postproc(self):
+        # Entra uses hex encoded IDs/secrets so just filter all hex data (with
+        # `-`) to be safe. These can be generated with uuidgen:
+        # $ uuidgen
+        # dd591ced-483e-4c47-beaf-ff46f68aab0a
+        self.do_path_regex_sub(
+            r".*",
+            r"[a-fA-F0-9-]{18,}",
+            r"******",
+        )
+
+        # Google's encoding is less clear, so we'll just filter out the values
+        # of the config fields (client_id and client_secret):
+        # client_secret = some.base64.stuff.with.domain
+        self.do_path_regex_sub(
+            r".*",
+            r"(.*_(id|secret)\s*=\s*)(.*)",
+            r"\1******",
+        )
--- a/tests/report_tests/plugin_tests/authd/entra_broker.conf
+++ b/tests/report_tests/plugin_tests/authd/entra_broker.conf
@@ -1,0 +1,4 @@
+[oidc]
+issuer = https://login.microsoftonline.com/a24144fa-4c27-4094-b21a-77c8fc56554a/v2.0
+client_id = 82e3e7f7-a1a3-4483-b782-9db24b8edc71
+force_provider_authentication = false
--- a/tests/report_tests/smoke_tests.py
+++ b/tests/report_tests/smoke_tests.py
@@ -10,7 +10,7 @@ import re
 
 
 from avocado.utils import process
-from sos_tests import StageOneReportTest, redhat_only, debian_only
+from sos_tests import StageOneReportTest, redhat_only, debian_only, ubuntu_only
 
 
 # These are the header strings in --list-plugins output
@@ -52,9 +52,6 @@ class AllPluginSmokeTest(StageOneReportTest):
 
         Make sure our warnings are displayed
         """
-        self.assertOutputContains('Not logged in to OCP API, and no login '
-                                  'token provided. Will not collect `oc` '
-                                  'commands')
         self.assertOutputContains('Source the environment file for the user '
                                   'intended to connect to the OpenStack '
                                   'environment.')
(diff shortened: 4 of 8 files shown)


sos/report/plugins/authd.py
```diff
@@ -0,0 +1,53 @@
...
+        ])
+
+    def postproc(self):
+        # Entra uses hex encoded IDs/secrets so just filter all hex data (with
+        # `-`) to be safe. These can be generated with uuidgen:
+        # $ uuidgen
+        # dd591ced-483e-4c47-beaf-ff46f68aab0a
+        self.do_path_regex_sub(
+            r".*",
+            r"[a-fA-F0-9-]{6,}",
```
pmoravec, line 42: This will match also words like `de-caf` or `caffee` or `baffed` etc. - are you ok with obfuscating some unnecessary strings (even substrings inside a word)?

sos/report/plugins/authd.py
```diff
@@ -0,0 +1,53 @@
...
+            r"******",
+        )
+
+        # Google's encoding is less clear, so we'll just filter out the values
+        # of the config fields (client_id and client_secret):
+        # client_secret = some.base64.stuff.with.domain
+        self.do_path_regex_sub(
+            r".*",
```
pmoravec, line 50: Ditto "really to all files?".

pmoravec:
Please squash the `[tests] Fix Debian smoke-tests ` commit to the other [tests] one.

    (209/277) tests/report_tests/smoke_tests.py:ExpectedDefaultPluginsTest.test_debian_default_plugins:  FAIL: Plugin 'apparmor' not recorded in manifest (0.05 s)

test is failing (maybe something more..?) and seems relevant to the change in tests.

sos/report/plugins/authd.py
MggMuggins: Upped the quantifier to 18, that should reduce the likelihood of a false positive.

sos/report/plugins/authd.py
```diff
@@ -0,0 +1,53 @@
...
+            "/var/snap/authd-msentraid/current/broker.conf.d/*",
+        ])
+
+    def postproc(self):
+        # Entra uses hex encoded IDs/secrets so just filter all hex data (with
+        # `-`) to be safe. These can be generated with uuidgen:
+        # $ uuidgen
+        # dd591ced-483e-4c47-beaf-ff46f68aab0a
+        self.do_path_regex_sub(
+            r".*",
```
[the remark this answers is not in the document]
MggMuggins: I think so, yes. The conf files in `/etc` probably don't require it, but I think I'd rather be safe and at least try to catch misconfiguration (user added their IDs to `/etc/authd` instead of `/var/snap/<broker>`). It's inconvenient to go back and ask for a file's contents again, but I prefer to do that than explain "oh, we collected your IDP secrets by mistake, you need to reissue those now".

If there were other files then it maybe makes sense to limit the scrubbing to `.*.conf`.

MggMuggins:
I feel a little foolish as I needed `ubuntu_only = True` in my test class instead of `@ubuntu_only` on my individual tests.

This is ready to go again.

sos/report/__init__.py
```diff
@@ -447,7 +447,7 @@ def display_help(cls, section):
         ssec.add_text(helpln)
 
     def print_header(self):
-        print(f"\n{_(f'sos report (version {__version__})')}\n")
+        self.ui_log.info(f"\n{_(f'sos report (version {__version__})')}\n")
```
pmoravec, line 450: We might use direct `print` by a purpose here as there is a call flow when `ui_log` might not be set up..?

Do you @TurboTurtle recall anything? If not I am ok with this.

tests/report_tests/smoke_tests.py
```diff
@@ -52,9 +52,6 @@ def test_expected_warnings_displayed(self):
 
         Make sure our warnings are displayed
         """
-        self.assertOutputContains('Not logged in to OCP API, and no login '
-                                  'token provided. Will not collect `oc` '
-                                  'commands')
```
pmoravec, line 57: Why this removal?

tests/report_tests/smoke_tests.py
```diff
@@ -97,20 +94,25 @@ def test_rhel_default_plugins(self):
         """Plugins expected to always run on a RHEL (-like) system
         """
         self.assertPluginIncluded([
-            'anaconda',
```
pmoravec, line 100: Interesting, I as under the impression this is a must.. But I can easily get e.g. a RHEL system where the plugin is not enabled.

tests/sos_tests.py
```diff
@@ -51,27 +51,27 @@ def wrapper(self, *args, **kwargs):
     return decorator
 
 
-# pylint: disable=unused-argument
 def redhat_only(tst):
-    def wrapper(func):
+    def wrapper(*args, **kwargs):
         if _distro.name not in RH_DIST:
             raise TestSkipError('Not running on a Red Hat distro')
+        tst(*args, *kwargs)
```
pmoravec, line 58: Aren't the tests run even without this explicit call of the method? (same applies to all three `*_only`).

pmoravec:
> * `anaconda` is no longer installed in Fedora 40+ by default (can someone confirm this?)

I think so, while I see `anaconda-core` package instead (but not sent on RHEL10 beta, e.g.), and same logs in `/var/log/anaconda` (on F40 and also 10 beta). The later is also a plugin trigger.

The removal in tests seems legit, but let me double-check it first.

tests/sos_tests.py
MggMuggins: No:
```python
def my_decorator(func):
    print("my_decorator() called")

    def wrapper(*args, **kwargs):
        print("wrapper() called")

    return wrapper


@my_decorator
def my_func():
    print("my_func() called")


my_func()
```
```
$ python3.10 decorators.py
my_decorator() called
wrapper() called
```

arif-ali: I think what was meant was the change of `*args, **kwargs` changes in the functions, do we need this change?

sos/report/plugins/authd.py
```diff
@@ -0,0 +1,53 @@
...
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, UbuntuPlugin
+
+
+# See the docs: https://documentation.ubuntu.com/authd/
+class Authd(Plugin, UbuntuPlugin):
```
arif-ali, line 13: For the `authd` plugin do we need to check for any service status, or journal logs that could be useful?

MggMuggins: Is it idomatic for sos to assume that version information, systemd unit status & logs will be collected by the `snap`, `apt`, and `systemd` plugins rather than the plugin for a specific service's configuration? Happy to add those things if needed.

The only information that I've not been able to view about authd while working cases has been the content of these config files. The [authd bug report template](https://github.com/ubuntu/authd-oidc-brokers/issues/new?template=bug_report.yml) has a reasonable collection of information to grab and I'm quite confident that with this plugin all of that will be included in a sosreport.

tests/sos_tests.py
MggMuggins: Yes; if this is used to annotate a class method:
```python
class MyTests(...):
    @ubuntu_only
    def test_smth(self):
        # ...
```
the `self` parameter will be missed & the tests will fail. I included `kwargs` for completeness, I don't think any functions currently take kwargs.

sos/report/plugins/authd.py
arif-ali: We also have to assume that if this plugin is ran standalone with `-o authd`, then it could be that this information is not collected, imo, we should assume that this is a standalone plugin, so if we need to collect those then let's get them added

tests/sos_tests.py
arif-ali: yeah, no arguments are being passed anywhere in the sos code for any of these decorators, hence probably the question of the change here

pmoravec: If I got it right, the change is not needed "per se" or for the sake of this PR, but it allows potential usage of arguments in decorators, right?

Could be..

pmoravec (approved):

sos/report/plugins/authd.py
MggMuggins: Added systemd services and an authd version check to the plugin; they automatically call [`add_service_status`](https://sos.readthedocs.io/en/main/plugins.html#sos.report.plugins.Plugin.add_service_status) and [`add_journal`](https://sos.readthedocs.io/en/main/plugins.html#sos.report.plugins.Plugin.add_journal) for those services.

arif-ali (approved):

tests/report_tests/plugin_tests/authd/entra_broker.conf
```diff
@@ -0,0 +1,4 @@
+[oidc]
+issuer = https://login.microsoftonline.com/a24144fa-4c27-4094-b21a-77c8fc56554a/v2.0
+client_id = 82e3e7f7-a1a3-4483-b782-9db24b8edc71
+force_provider_authentication = false
```
pmoravec, line 4: Those are fake credentials, right?

I would like to double-confirm it before I merge it.

Same applies to google_broker.conf .

MggMuggins: Yes, I generated all the binary bits locally based on the shape of some real ones.

pmoravec:
Closed via https://github.com/sosreport/sos/commit/273d3819098b2f62063a9845ddb4ba21079f58f7 .

PR #4013, opened 2025-05-13
[juju] Fix juju agent introspect commands

arif-ali:
By default the introspect commands are aliases in the bash login profile, so these won't be available via sos. This extraplotes the commands that are available and passes it to juju-introspect.

---

### DIFF
--- a/sos/report/plugins/juju.py
+++ b/sos/report/plugins/juju.py
@@ -10,6 +10,7 @@
 
 import pwd
 import json
+import re
 from sos.report.plugins import Plugin, UbuntuPlugin, PluginOpt
 
 
@@ -76,6 +77,8 @@ class Juju(Plugin, UbuntuPlugin):
         ),
     ]
 
+    agent_name = ""
+
     def setup(self):
         # Juju service names are not consistent through deployments,
         # so we need to use a wildcard to get the correct service names.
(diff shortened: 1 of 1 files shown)


arif-ali:
seems like fedora 40 went EoL, this now depends on #4014 being merged

sos/report/plugins/juju.py
```diff
@@ -83,18 +84,25 @@ def setup(self):
             self.add_journal(service)
             self.add_service_status(service)
 
-        self.add_cmd_output([
-            'juju_engine_report',
-            'juju_goroutines',
-            'juju_heap_profile',
-            'juju_leases',
```
pmoravec, line 90: You stop collecting this, is it intentional?

sos/report/plugins/juju.py
```diff
@@ -186,6 +194,15 @@ def setup(self):
...
                         self.add_cmd_output(command, runas=juju_user)
 
+    def _juju_agent(self, command):
+        agent_name = ""
+        for dir_name in self.listdir("/var/lib/juju/agents"):
+            if re.search('machine-*|controller-*|application-*', dir_name):
+                agent_name = dir_name
+                break
+
+        return f"juju-introspect --agent={agent_name} {command}"
```
pmoravec: Isnt it faster to store the `agent_name` in a variable and use it statically? Currently you call this method once for each `juju_agents_cmds` while you get a static info that can be fetched just once.

Or do I miss something?

arif-ali: yup, good catch, was trying to mimic a different version of the same thing and didn't really think about it too much. I'll get that updated

sos/report/plugins/juju.py
arif-ali: The juju_leases alias doesn't exist anymore, plus it was very complex to bring it here in sos.

As this portion didn't work, it doesn't make much of a difference to actually remove it.

TurboTurtle (approved):

PR #4020, opened 2025-05-21
[policy] Re-add logic to request case-id if not present

jcastill:
This commit re-adds the logic to ask for case-id if the user hasn't specified it in the command line
explicitly.

Related: RHEL-92071

---

### DIFF
--- a/sos/policies/distros/__init__.py
+++ b/sos/policies/distros/__init__.py
@@ -13,6 +13,7 @@
 import os
 import re
 
+from sos import _sos as _
 from sos.policies import Policy
 from sos.policies.init_systems import InitSystem
 from sos.policies.init_systems.systemd import SystemdInit
@@ -296,10 +297,21 @@ class LinuxPolicy(Policy):
         # this method will be called before the gathering begins
 
         cmdline_opts = self.commons['cmdlineopts']
+        caseid = cmdline_opts.case_id if cmdline_opts.case_id else ""
 
         if cmdline_opts.low_priority:
             self._configure_low_priority()
 
+        # set or query for case id
+        if not cmdline_opts.batch and not \
+                cmdline_opts.quiet:
+            if caseid:
+                self.commons['cmdlineopts'].case_id = caseid
+            else:
+                self.commons['cmdlineopts'].case_id = input(
+                    _("Optionally, please enter the case id that you are "
+                      "generating this report for [%s]: ") % caseid
+                )
         if cmdline_opts.case_id:
             self.case_id = cmdline_opts.case_id
 


sos/policies/distros/__init__.py
```diff
@@ -296,10 +297,21 @@ def pre_work(self):
...
+        # set or query for case id
+        if not cmdline_opts.batch and not \
+                cmdline_opts.quiet:
+            if caseid:
+                self.commons['cmdlineopts'].case_id = caseid
+            else:
+                self.commons['cmdlineopts'].case_id = input(
+                    _("Optionally, please enter the case id that you are "
+                      "generating this report for [%s]: ") % caseid
+                )
```
pmoravec, line 314: We have the same code in `prompt_for_case_id` method in https://github.com/sosreport/sos/blob/main/sos/upload/targets/redhat.py#L55-L72 . Shouldnt we have just one routine for that identical code block?

(wont this code block here make the `prompt_for_case_id` method redundant?)

pmoravec:
Interesting; somewhere between sos-4.8.2 and sos-4.9.0, we stopped asking for the case ID.

Is it worth having a test for this? That would need `expect` or something similar to interact with non-batched `sos report`, which can be tricky. So I am not sure..

TurboTurtle (changes requested):
As @pmoravec mentioned, this is a duplication of the code from `prompt_for_case_id()`, which was moved into `upload` from the base `distro/__init__.py`.

We shouldn't just copy code into multiple locations. The method needs to be removed from upload, and added back into the base policy class (as a whole method) so that case id is again prompted for across all relevant subcommands, not just upload.

sos/policies/distros/__init__.py
```diff
@@ -300,9 +301,26 @@ def pre_work(self):
...
+
+        if not cmdline_opts.batch and not \
+                cmdline_opts.quiet:
+            if not cmdline_opts.case_id:
+                cmdline_opts.case_id = input(
+                    _("Optionally, please enter the case id that you are "
+                      "generating this report for: ")
+                )
         if cmdline_opts.case_id:
             self.case_id = cmdline_opts.case_id
```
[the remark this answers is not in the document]
jcastill: Ack, checking this now

pmoravec (approved):

sos/upload/targets/redhat.py
```diff
@@ -100,9 +81,6 @@ def get_upload_url(self):
                 return self.commons['cmdlineopts'].upload_url
             if self.commons['cmdlineopts'].upload_protocol == 'sftp':
                 return self.RH_SFTP_HOST
-            if not self.commons['cmdlineopts'].case_id and not\
-               self.prompt_for_case_id():
-                return self.RH_SFTP_HOST
```
arif-ali, line 105: I am presuming this was removed on purpose and you don't need this if a case id is not there?

TurboTurtle: Also not sure why this is being removed.

TurboTurtle:
Not sure why the sftp fallback is being removed, and would like to hold merge until that is answered. Aside from that, looks ok.

sos/upload/targets/redhat.py
pmoravec: Indeed, quite basic use case `sos report -o host --batch --upload` diverts: it tries to upload to the customer portal while it should use SFTP (no case id).

Instead of 3 lines removal, there must be a change on 2nd line to:

    self.commons['policy'].prompt_for_case_id(self.commons['cmdlineopts']):

I had to overview this removal, sorry for ACKing it.

jcastill: The reason why I removed it was to run some tests and then I pushed the commit without re-adding it, sorry. 
Thank you @arif-ali ++ for spotting this :)
I've just pushed a new commit with the code back

pmoravec (approved):

arif-ali (approved):

PR #4031, opened 2025-05-28
[puppet] add 'puppet-agent', 'openvox-agent' and 'openvox-server' to packages

evgeni:
---


pmoravec (approved):

evgeni:
done

TurboTurtle (commented):
This would enable the plugin on puppet _clients_ now, when the current version in `main` does not, correct? Or does the `puppet-common` package exist on clients? I don't recall off the top of my head.

I don't see anything immediately that raises concerns for me for enabling this on clients, it's just a non-trivial change to the targets of this plugin and I want to make sure this is in fact what we want to do.

Code LGTM.

evgeni:
The old "puppet" and "puppet-common" names were present on clients too. At some point things got renamed to puppet-agent and puppetserver, but never completely updated here. So you could say this re-enables the plugin on clients. :)

TurboTurtle (approved):

PR #4036, opened 2025-06-03
[sunbeam] Add deployment commands

arif-ali:
* Update some commands that now need a sunbeam_user
* manifest show doesn't need --id any more

---


sos/report/plugins/sunbeam.py
```diff
@@ -57,7 +55,9 @@ def setup(self):
             manifests = yaml.safe_load(manifest_raw['output'])
             for manifest in manifests:
                 self.add_cmd_output(
-                    f'sunbeam manifest show --id {manifest["manifestid"]}')
+                    f'sunbeam manifest show {manifest["manifestid"]}',
```
gboutry: I don't think any sunbeam command should be run outside the context of the `sunbeam_user`. If it works, it's not by design.

arif-ali: Thankf for that, I'll get that updated on the next update on this PR

arif-ali: This should be updated in the updated PR now

pmoravec (approved):
From sos plugin perspective, ack. No idea what is being collected etc :) but that is not needed for such review.

TurboTurtle (approved):

TurboTurtle:
Restarted the stage two jobs. Will merge afterwards.

Issue #4039, opened 2025-06-04
[sunbeam_hypervisor] systemd journal logs does not provide logs


gboutry:
The sunbeam_hypervisor is using `nova-compute` unit to collect output for the openstack hypervisor snap's nova-compute process, which will return nothing, as it is defined on systemd as `snap.openstack-hypervisor.nova-compute`.

sos output:
```
-- No entries --
```

`snap logs -n all openstack-hypervisor` is a better candidate, because then you don't have to know which systemd unit are created by the snap, it will always return all of the systemd unit logs managed by the snap.

PR #4040, opened 2025-06-04
[sunbeam_hypervisor] Add services

arif-ali:
The services tuple does not like regex, so adding them individually

Closes: #4039

---

### DIFF
--- a/sos/report/plugins/sunbeam_hypervisor.py
+++ b/sos/report/plugins/sunbeam_hypervisor.py
@@ -17,13 +17,26 @@ class SunbeamHypervisor(Plugin, UbuntuPlugin):
     profiles = ('cloud',)
     packages = ('openstack-hypervisor',)
 
+    plugin_timeout = 600
+
+    services = (
+        'snap.openstack-hypervisor.ceilometer-compute-agent',
+        'snap.openstack-hypervisor.libvirt-exporter',
+        'snap.openstack-hypervisor.libvirtd',
+        'snap.openstack-hypervisor.masakari-instancemonitor',
+        'snap.openstack-hypervisor.neutron-ovn-metadata-agent',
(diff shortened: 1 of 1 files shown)


gboutry (commented):
If thats the case, isn't it better to `sudo snap logs -n all openstack-hypervisor`?


I don't think maintaining a list of units is very pertinent in the context of a snap, what do you think @arif-ali ?

sos/report/plugins/sunbeam_hypervisor.py
```diff
@@ -17,14 +17,27 @@ class SunbeamHypervisor(Plugin, UbuntuPlugin):
...
+        'snap.openstack-hypervisor.virtlogd',
+    )
+
     common_dir = '/var/snap/openstack-hypervisor/common'
 
     def setup(self):
 
-        self.add_service_status('snap.openstack-hypervisor.*')
-
-        self.add_journal('nova-compute')
```
pmoravec, line 26: Is this removal intentional? (could be, just asking for confirmation)

arif-ali: yes, as the service doesn't exist, and now under the `snap.openstack-hypervisor.nova-compute`

arif-ali:
> If thats the case, isn't it better to `sudo snap logs -n all openstack-hypervisor`?
> 
> I don't think maintaining a list of units is very pertinent in the context of a snap, what do you think @arif-ali ?

After some testing this morning in my lab env, it seems that running `sudo snap logs -n all openstack-hypervisor` can take a long time to run, and aggregates loads of logs. This means with the timeouts we have, not all logs would be collected. It makes sense to split these to ensure that we get all the logs.

We do this similarly in all the plugins we have in `sos`, so keeping this as is makes sense.

TurboTurtle (approved):
LGTM

PR #4041, opened 2025-06-09
[config] Update sos.conf with example for upload

jcastill:
Update /etc/sos/sos.conf example file with details for the upload subsystem

Related: RHEL-95358

---


sos.conf
```diff
@@ -30,6 +30,13 @@
 #domains = mydomain.com
 #no-update = true
 
+[upload]
+# Options that will apply to any `sos upload` run should be listed here.
+# Note that the option names *must* be the long-form name as seen in --help
+# output. Use a comma for list delimitations
+#case-id = 0000001
```
arif-ali: case-id can be there for report, collect and upload, is it worth adding this as example here. Is it worth adding a different option here, or just leave it with just the `upload-prootocol`

jcastill: ack, I can add the option to select target that is explicitly for upload and nothing else. Would that work better?

arif-ali (approved):
lgtm

PR #4042, opened 2025-06-10
[doc] Remove deprecated call to setup.py

jcastill:
When running 'python3 setup.py install' we get an alert about deprecation:

running install
/usr/lib/python3.12/site-packages/setuptools/_distutils/cmd.py:66: SetuptoolsDeprecationWarning: setup.py install is deprecated. !!

setup.py per se is not deprecated, but the way
it's invoked is. This commit provides an alternative way to install sos still using setup.py.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ $ sudo ./bin/sos report -l
 
 To install locally (as root):
 ```
-# python3 setup.py install
+# python3 -m pip install .
 ```
 
 


jcastill:
There were at least possible solutions/substitutions here:

1- When running the command as it is now in the README, we get this after the warning:
```
		********************************************************************************
		Please avoid running ``setup.py`` directly.
		Instead, use pypa/build, pypa/installer or other
		standards-based tools.

		See https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html for details.
		********************************************************************************
```

The blog post offers as an alternative `pip install`, in a table in the section "Summary".

2- There's also this doc that offers an alternative to `python3 setup.py install`:

https://packaging.python.org/en/latest/discussions/setup-py-deprecated/#setup-py-deprecated
https://packaging.python.org/en/latest/guides/modernize-setup-py-project/#modernize-setup-py-project  *

Here the command specified is:

`python -m pip install .`

I chose this one but can change it to the first one if you all think it's better. 


* In the second link it's mentioned that '_A pyproject.toml file is strongly recommended._' I've been experimenting with a pyproject.toml for sos just as a side thing, but if the community thinks it could be useful for us to have one, let me know and I can open a draft PR with the work I've done so far, and the issues I've encountered, so we can all work on this together.

arif-ali:
I had a go at converting the setup.py to pyproject.toml a while back, and had got stuck (I think it was abut the `data_files` option from setup.py that I couldn't replicate), and then didn't move it along and forgot about it. I would love to see us move to the new recommended approach (but that is my personal opinion)

jcastill:
Thank you both for your insights, it's really helpful. I noticed the same problems with man pages, and a similar one while translating/transforming from setup.py's `scripts` section to pyproject.toml `[project.scripts]`, but that could be because I still don't understand it fully. 

I'll open a discussion or an issue so we can put some ideas together and track some possible solutions.

jcastill:
Hey team, this is still relevant, right?

arif-ali:
@jcastill yeah, still looks relevant to me.

When I run the `python3 -m pip install` in a newer system, then I get the following error

```
root@sos-testing:~/sos# python3 -m pip install .
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.14/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
```

So had to install it using the following command instead

```
python3 -m pip install . --break-system-packages
```

Personally, I don't like this method of installing, I think it makes better sense that this would be for testing purposes only, and ideally people should use pre-pacakaged versions of `sos`

bmr-cymru:
In `boom-boot` and `snapm` we document `pip install` but advise users to treat this as an "expert"/"development"/"testing" option (we use it in CI).

There's a blog post explaining it in more detail here: https://snapshotmanager.github.io/blog/getting-snapm/

I don't think it would be unreasonable to document a similar limitation/caution for users wanting to install `sos` via `pip`.

jcastill:
> In `boom-boot` and `snapm` we document `pip install` but advise users to treat this as an "expert"/"development"/"testing" option (we use it in CI).
> 
> There's a blog post explaining it in more detail here: [snapshotmanager.github.io/blog/getting-snapm](https://snapshotmanager.github.io/blog/getting-snapm/)
> 
> I don't think it would be unreasonable to document a similar limitation/caution for users wanting to install `sos` via `pip`.

I'll see if I can adapt your approach. I like specially points 2 and 3 in your link, because it makes very clear the different ways to get the tool depending on the user intentions (i.e. "early access to new features & fixes") and I don't think we mention copr in our documentation at all.

arif-ali:
> > In `boom-boot` and `snapm` we document `pip install` but advise users to treat this as an "expert"/"development"/"testing" option (we use it in CI).
> > There's a blog post explaining it in more detail here: [snapshotmanager.github.io/blog/getting-snapm](https://snapshotmanager.github.io/blog/getting-snapm/)
> > I don't think it would be unreasonable to document a similar limitation/caution for users wanting to install `sos` via `pip`.
> 
> I'll see if I can adapt your approach. I like specially points 2 and 3 in your link, because it makes very clear the different ways to get the tool depending on the user intentions (i.e. "early access to new features & fixes") and I don't think we mention copr in our documentation at all.

If you're going to mention copr in your updaye, then we also have daily builds available if people want to test newly merged commits in Ubuntu. https://launchpad.net/~sosreport/+archive/ubuntu/daily for reference. These aren't necessarily daily builds per say, but when a new commit arrives in `main` branch, and based on timings (typically job runs every 4-6 hours), it will automatically build a new package if there is a new change.

PR #4046, opened 2025-06-11
[dqlite] Conduct Dqlite Collection in Consumers

bryanfraschetti:
Collect files relating to the configuration of the distributed cluster that underlies micro* applications like microceph, microovn, microcloud, microk8s, and lxd. Collect the directory listing of the dqlite database dir to check for corrupt intervals, and collect the expiry date of cluster certs. As well, inspect information pertinent to the cluster by querying against the dqlite db.

---

### DIFF
--- a/sos/report/plugins/ceph_common.py
+++ b/sos/report/plugins/ceph_common.py
@@ -5,7 +5,7 @@
 # version 2 of the GNU General Public License.
 #
 # See the LICENSE file in the source distribution for further information.
-
+import json
 from socket import gethostname
 from sos.report.plugins import Plugin, RedHatPlugin, UbuntuPlugin
 
@@ -103,9 +103,6 @@ class CephCommon(Plugin, RedHatPlugin, UbuntuPlugin):
                 'client config list',
                 'cluster config list',
                 'cluster list',
-                # exclude keyrings from the config db
-                'cluster sql \'select * from config where key NOT LIKE \
-                    \"%keyring%\"\'',
                 'disk list',
                 'log get-level',
                 'status',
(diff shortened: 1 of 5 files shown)


sos/report/plugins/ceph_common.py
```diff
@@ -117,6 +114,87 @@ def setup(self):
...
+                {
+                    "query": "SELECT * FROM services;",
+                    "suggested_file_suffix": "services",
+                },
+                {
+                    "query": (
+                        "SELECT id, name, expiry_date "
+                        "FROM core_token_records;"
+                    ),
+                    "suggested_file_suffix": "token_records",
```
MggMuggins, line 163: I'm not sure that we need the token records; these tend to be short lived (they expire after 3 hrs by default) so the table will almost always be empty.

MggMuggins (commented):
Thanks, this looks good, just one item.

I build a [test ppa](https://launchpad.net/~whershberger/+archive/ubuntu/sosreport) with this change and generated a report with `sudo sos report -o ceph_common,lxd,microcloud,microovn`; everything that should be there is there and I don't see any issues in the sos logs.

sos/report/plugins/ceph_common.py
bryanfraschetti: I agree but also it could be useful if a customer is having issues when generating and using their tokens. We could check if for some reason there aren't any valid entries in the db when we might be expecting entries, or confirm that it does exist and know to investigate something else

MggMuggins: I'm amenable to that argument; this is fine then :)

sos/report/plugins/ceph_common.py
```diff
@@ -117,6 +114,87 @@ def setup(self):
...
+                subdir="microceph",
+            )
+
+            self.add_copy_spec(
+                [
+                    f"{db_path}/info.yaml",
+                    f"{db_path}/cluster.yaml",
+                    f"{db_path}/../daemon.yaml",
+                ]
+            )
```
arif-ali: Just to follow previous standards, lets use the following structure instead, so that the brackets are together here

```python
            self.add_copy_spec([
                f"{db_path}/info.yaml",
                f"{db_path}/cluster.yaml",
                f"{db_path}/../daemon.yaml",
            ])
```

sos/report/plugins/microovn.py
```diff
@@ -38,3 +39,69 @@ def setup(self):
...
+            suggest_filename="ls_microovn_dqlite_dir",
+        )
+
+        self.add_copy_spec(
+            [
+                f"{db_path}/info.yaml",
+                f"{db_path}/cluster.yaml",
+                f"{db_path}/../daemon.yaml",
+            ]
+        )
```
arif-ali: same as above wrt indentation, and brackets

arif-ali (changes requested):
Based on discussions, this is what we agreed on

With @MggMuggins testing, I am good with the overall PR

Other than a couple minor points, this looks good.

Issue #4047, opened 2025-06-12
[sunbeam] Missing terraform logs


hemanthnakkina:
To collect terraform logs, the path should be `snap_home_dir/etc/*/*/*.log`, an extra /* missing in [1]
Can we also collect terraform vars set by sunbeam - `snap_home_dir/etc/*/*/terraform.tfvars.json`

[1] https://github.com/sosreport/sos/blob/820cfea36894e88368217e14729b2997ef35a9dc/sos/report/plugins/sunbeam.py#L97

arif-ali:
@hemanthnakkina thanks for that, the logs folder was a typo from myself, thanks for raising that one. New PR has been created, and hopefully that solves the issue.

PR #4048, opened 2025-06-12
[sunbeam] Collect terraform logs and variables

arif-ali:
Closes: #4047

---

### DIFF
--- a/sos/report/plugins/sunbeam.py
+++ b/sos/report/plugins/sunbeam.py
@@ -94,7 +94,8 @@ class Sunbeam(Plugin, UbuntuPlugin):
 
             self.add_copy_spec([
                 f"{sb_snap_homedir}/*.log",
-                f"{sb_snap_homedir}/etc/*/*.log",
+                f"{sb_snap_homedir}/etc/**/*.log",
+                f"{sb_snap_homedir}/etc/**/terraform.tfvars.json",
                 f"{sb_snap_homedir}/logs/*.log",
                 f"{sb_snap_homedir}/reports/*.yaml",
             ])


hemanthnakkina (commented):
Thanks for the quick fix

sos/report/plugins/sunbeam.py
```diff
@@ -94,7 +94,8 @@ def setup(self):
 
             self.add_copy_spec([
                 f"{sb_snap_homedir}/*.log",
-                f"{sb_snap_homedir}/etc/*/*.log",
+                f"{sb_snap_homedir}/etc/*/*/*.log",
+                f"{sb_snap_homedir}/etc/*/*/terraform.tfvars.json",
```
jcastill: I'm reading in some places, that some users may store the following two tf variables in `terraform.tfvars.json`:

```
"aws_access_key": "accesskey",
"aws_secret_key": "secretkey"
```

It seems it's discouraged, as it should be, and users could use instead environment variables. But just to be on the safe side, should we add these two possible entries to be obfuscated via postproc?

arif-ali: tbh, the variables could be anything, and depends on the terraform plans or the terraform providers. If you're not using the AWS terraform provider, then there is no need for these variables to be obfuscated.

sos/report/plugins/sunbeam.py
```diff
@@ -94,7 +94,8 @@ def setup(self):
 
             self.add_copy_spec([
                 f"{sb_snap_homedir}/*.log",
-                f"{sb_snap_homedir}/etc/*/*.log",
+                f"{sb_snap_homedir}/etc/*/*/*.log",
```
TurboTurtle: We could use `**` here, e.g. `etc/**/*.log`, which matches any number of subdirs.

sos/report/plugins/sunbeam.py
jcastill: ok!

TurboTurtle (approved):

PR #4056, opened 2025-06-14
[tftpserver] Add Directory Listing /var/lib/tftpboot

CharlieKrell:
Add /var/lib/tftpboot to the directory listing.
This is the default directory for boot files for tftp booting. 
Reference: /usr/share/doc/tftp-server/README.security

Signed-off-by: Charlie Krell <ckrell@redhat.com>

---

### DIFF
--- a/sos/report/plugins/tftpserver.py
+++ b/sos/report/plugins/tftpserver.py
@@ -24,7 +24,7 @@ class TftpServer(Plugin, IndependentPlugin):
         self.add_dir_listing([
             '/var/lib/tftpboot',
             '/tftpboot',
-            '/sr/tftp'
+            '/srv/tftp'
         ], recursive=True)
 
 # vim: set et ts=4 sw=4 :
--- a/sos/report/plugins/tftpserver.py
+++ b/sos/report/plugins/tftpserver.py
@@ -21,6 +21,10 @@ class TftpServer(Plugin, IndependentPlugin):
     packages = ('tftp-server',)
 
     def setup(self):
-        self.add_dir_listing(['/tftpboot', '/sr/tftp'], recursive=True)
+        self.add_dir_listing([
+            '/var/lib/tftpboot',
+            '/tftpboot',
+            '/sr/tftp'
+        ], recursive=True)
 
 # vim: set et ts=4 sw=4 :


sos/report/plugins/tftpserver.py
```diff
@@ -21,6 +21,10 @@ class TftpServer(Plugin, IndependentPlugin):
     packages = ('tftp-server',)
 
     def setup(self):
-        self.add_dir_listing(['/tftpboot', '/sr/tftp'], recursive=True)
```
bmr-cymru, line 24: Unrelated to your change, but while we're here - that `/sr/tftp` should surely be `/srv/tftp` (the [default location on SUSE](https://documentation.suse.com/sled/15-SP5/html/SLED-all/cha-deployment-prep-pxe.html) - would you mind adding an additional commit to your PR to put this right along with the `/var/lib/tftpboot` change?

bmr-cymru (commented):
The DCO check doesn't like your Signed-off-by since the committer/author is set to "ckrell@ckrell-thinkpadt14gen5.rmtusil.csb", rather than the expected "ckrell@redhat.com".

TurboTurtle:
Hrm. There's some, perhaps expected, overlap with the `pxe` plugin and this plugin. Neither of which have gotten attention for a while. The main difference being `tftpboot` simply lists the contents, where `pxe` is collecting them - but only on Debian.

Perhaps we should merge these two plugins?

sos/report/plugins/tftpserver.py
CharlieKrell: Oh thank you! I wasn't sure why the DCO was failing.  This is my first commit so I really appreciate the help.

And yes, I'll add another commit for the /srv/tftp part too and add it to the PR.

CharlieKrell:
Oh sorry I missed the signed-off in the second commit.  I'll amend

TurboTurtle (approved):

TurboTurtle:
This LGTM for the change as proposed. I'll make a note to merge the plugins (most likely into `tftpserver`) later this week/weekend.

arif-ali (approved):

PR #4057, opened 2025-06-17
[upload] Fix backtrace when cancelling upload

jcastill:
This commit fix a backtrace that appears when
the user cancels a file upload after authentication.

Related: RHEL-94968

---

### DIFF
--- a/sos/upload/__init__.py
+++ b/sos/upload/__init__.py
@@ -278,45 +278,50 @@ this utility.
             self.policy._configure_low_priority()
 
     def execute(self):
-        self.pre_work()
-        if self.from_cmdline:
-            self.intro()
-            self.archive = self.opts.upload_file
-            self.caseid = self.policy.prompt_for_case_id(
-                cmdline_opts=self.opts
-            )
-        cmdline_target = self.opts.upload_target
-        if cmdline_target and cmdline_target != 'local':
-            self.upload_target = self.upload_targets[cmdline_target]
-        else:
-            self.determine_upload_target()
-        if not self.upload_target:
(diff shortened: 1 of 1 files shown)


sos/upload/__init__.py
```diff
@@ -278,45 +278,50 @@ def pre_work(self):
...
+                            sys.exit(1)
+                    else:
+                        self.ui_log.error(_(f"{self.archive} is not a file."))
+                else:
+                    self.ui_log.error(_(f"File {self.archive} is empty."))
+            except Exception as e:
+                self.ui_log.error(_(f"Cannot upload {self.archive}: {e} "))
+        except KeyboardInterrupt:
+            self._exit("Exiting on user cancel", 130)
+
```
jcastill, line 325: I was thinking about adding a general exception after KeyboardInterrupt to capture any other uncatched exceptions. Any thoughts on this?

arif-ali: Might as well, of there is a chance for another issue to happen?

jcastill: I've tried to come up with possible ways to trigger other exceptions, but haven't found a way to get one yet. Rather, I thought I could add them generally after being explicit with the KeyboardInterrupt one, and if we capture one we can refine it later on (i.e. add a new explicit one before the general one).

jcastill:
@pmoravec @arif-ali shall I add a general exception catch then and force push? Or do you prefer if I leave it as it's now?

arif-ali (approved):
if you've tried other means, lets stick with what we have, and we move on

PR #4059, opened 2025-06-18
[saltmaster] Obfuscate pillar roots in any path

### LINKED ISSUE #4058: [saltmaster] can fail to obfuscate during postproc
1. When pillar roots are collected outside of the `/etc/salt/*` path.
2. When passwords, secrets, or keys are not nested inside a dictionary (are at a root level).
3. When an api or encryption key uses anything besides the bare word `key`.

TrevorBenson:
- Obfuscates root level passwords, secrets and keys
- Includes api and encryption keys, but keeps original negative lookbehind for any other "key"
- Obfuscates a pillar root in any collected path in addition to `/etc/salt`

Closes #4058

---

### DIFF
--- a/sos/report/plugins/saltmaster.py
+++ b/sos/report/plugins/saltmaster.py
@@ -20,6 +20,10 @@ class SaltMaster(Plugin, IndependentPlugin):
 
     packages = ('salt-master', 'salt-api',)
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.collected_pillar_roots = []
+
     def setup(self):
         if self.get_option("all_logs"):
             self.add_copy_spec("/var/log/salt")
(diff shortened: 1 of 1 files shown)


sos/report/plugins/saltmaster.py
```diff
@@ -20,6 +20,10 @@ class SaltMaster(Plugin, IndependentPlugin):
 
     packages = ('salt-master', 'salt-api',)
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.collected_pillar_roots = []
```
TurboTurtle, line 24: We don't need to override `__init__()` here for this, I don't think. We should be able to initialize this in `setup()`? If not, then make it as a class attr and we can avoid the `super()` invocation.

TrevorBenson: Confirmed with a test. Applied and pushed.

PR #4064, opened 2025-06-24
[aap_containerized] collect podman inspect logs

snagoor:
---

### DIFF
--- a/sos/report/plugins/aap_containerized.py
+++ b/sos/report/plugins/aap_containerized.py
@@ -97,14 +97,18 @@ class AAPContainerized(Plugin, RedHatPlugin):
         # Collect AAP container names
         aap_containers = self._get_aap_container_names(username)
 
-        # Copy podman container log files in plugin sub directory
-        # under aap_containers_log
+        # Copy podman container log and inspect files
+        # into their respective sub directories
         for container in aap_containers:
-            log_file = f"{container}.log"
             self.add_cmd_output(
                     f"su - {username} -c 'podman logs {container}'",
-                    suggest_filename=f"{log_file}",
-                    subdir="aap_containers_log"
+                    suggest_filename=f"{container}.log",
+                    subdir="aap_container_logs"
+            )
+            self.add_cmd_output(
+                    f"su - {username} -c 'podman inspect {container}'",
+                    suggest_filename=container,
+                    subdir="podman_inspect_logs"
             )
 
         if "automation-controller-task" in aap_containers:


sos/report/plugins/aap_containerized.py
```diff
@@ -97,14 +97,17 @@ def setup(self):
         # Collect AAP container names
         aap_containers = self._get_aap_container_names(username)
 
-        # Copy podman container log files in plugin sub directory
-        # under aap_containers_log
+        # Copy podman container log and inspect files into their respective subdirectories
```
arif-ali: please ensure that the characters on this line are no more than 79 characters

sos/report/plugins/aap_containerized.py
```diff
@@ -97,14 +97,17 @@ def setup(self):
...
         for container in aap_containers:
-            log_file = f"{container}.log"
             self.add_cmd_output(
                     f"su - {username} -c 'podman logs {container}'",
-                    suggest_filename=f"{log_file}",
-                    subdir="aap_containers_log"
+                    suggest_filename=f"{container}.log",
+                    subdir="aap_container_logs"
```
arif-ali, line 106: any reason why you're changing this? I'm no RH expert, but if the RH folks have any automation against the old filename or people are looking for the old name, then this shouldn't change

arif-ali (commented):
You are removing `podman logs` and adding `podman inspect`, should we not collect both?

I don't know enough about `podman`, this is where I need to be educated :)

snagoor:
> You are removing `podman logs` and adding `podman inspect`, should we not collect both?
> 
> I don't know enough about `podman`, this is where I need to be educated :)

If you observe the change we do collect both `podman logs` and `podman inspect` within the same for loop.

sos/report/plugins/aap_containerized.py
snagoor: @arif-ali this is just a place-holder directory. It never mattered and i incorrectly named initially as `aap_containers_log` instead of `aap_container_logs` which sounds more logical.

sos/report/plugins/aap_containerized.py
snagoor: ack

arif-ali (approved):
lgtm

arif-ali:
btw, have you tried the `runas` parameter, that could help without having to use `su - {username}` on all these commands

snagoor:
> btw, have you tried the `runas` parameter, that could help without having to use `su - {username}` on all these commands

I tried it initially, it never worked for me. Same was discussed here  https://github.com/sosreport/sos/pull/4030#discussion_r2114132989 @TurboTurtle said he will look into it later.

TurboTurtle (approved):

TurboTurtle:
> I tried it initially, it never worked for me. Same was discussed here [#4030 (comment)](https://github.com/sosreport/sos/pull/4030#discussion_r2114132989) @TurboTurtle said he will look into it later.

The issue there was using `runas` in conjunction with `container` - we don't currently introspect non-root container states, which is what throws that off. I think this _should_ work with just `runas`, but I won't block on that here since we'd ultimately want both of those working together, and doing a partial "fix" now would still need to be replaced.

I'm hoping to have some time in the next month or so to get non-root container introspection hooked up.

PR #4071, opened 2025-07-08
[report] Fix formatting of distros in plugins_overview

jcastill:
The formatting of the distros column was leaving the string 'Plu' when it found 'PluginOpt'. This fix
explicitly deals with both 'PluginOpt' and 'Plugin' strings and removes them.

---

### DIFF
--- a/plugins_overview.py
+++ b/plugins_overview.py
@@ -55,7 +55,10 @@ def add_all_items(method, dest, plugfd, wrapopen=r'\(', wrapclose=r'\)'):
                         # dirty hack to remove spaces and "Plugin"
                         if "Plugin" not in it:
                             continue
-                        it = it.strip(' ()')[0:-6]
+                        if "=" in it:
+                            it = re.sub(r"Plugin.*", "Plugin", it)
+                        plug_col_len = 9 if "PluginOpt" in it else 6
+                        it = it.strip(' ()')[:-plug_col_len]
                         if len(it):
                             dest.append(it)
         # list of specs separated by comma ..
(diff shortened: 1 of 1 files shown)


arif-ali:
I never used the plugins_overview in the past, so looking into it, there's probably other fixes we ought to do as well, as there are other issues too

If the sos.report.plugins import is the only line, and then there is a comment/variables before the class deceleration, then we'll see the whole comment.

- authd
- kubernetes
- charmed_postgresql
- openstacK_instack
- openstack_mistral
- origin
- validation_framework
- vdsm

Either we update these plugins so that we don't have that, or we fix the plugins_overview.py; Personally, I wpuld lean towards fixing the plugins_overview.py

thoughts?

jcastill:
I agree, I think fixing plugins_overview.py may be the way to go.

arif-ali:
I just found one more scenario

```
from sos.report.plugins import Plugin, RedHatPlugin, \
    DebianPlugin, UbuntuPlugin, CosPlugin
```

You then get the following in the distros list

```
"\    Debian"
```

I found 2 scenarios of this, apologies I didn't find this one earlier

jcastill:
No apologies needed at all @arif-ali , and thank you for taking the time to check this.
Somehow I didn't catch these either in the csv and json outputs. I'll work on this tomorrow.

jcastill:
@arif-ali I was thinking... it may be better to change the way we import in these two places (I've found kdump and release plugins, let me know if you found others) so it's consistent with the other plugins' imports of using parenthesis for multi-line imports?
I'll open  a new PR for that, but feel free to close it if you think it makes no sense.

PR #4076, opened 2025-07-09
[openshift_ovn] Allow specific container runtime for plugin command

dwolstroRH:
This Update allows a specific container runtime to be passed to add_cmd_output so commands that must use a specific runtime do so.

Related: SUPDEV-180

---

### DIFF
--- a/sos/report/plugins/openshift_ovn.py
+++ b/sos/report/plugins/openshift_ovn.py
@@ -1,5 +1,4 @@
 # Copyright (C) 2021 Nadia Pinaeva <npinaeva@redhat.com>
-
 # This file is part of the sos project: https://github.com/sosreport/sos
 #
 # This copyrighted material is made available to anyone wishing to use,
@@ -8,6 +7,7 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
+import glob
 from sos.report.plugins import Plugin, RedHatPlugin
 
 
@@ -21,6 +21,7 @@ class OpenshiftOVN(Plugin, RedHatPlugin):
     profiles = ('openshift',)
 
     def setup(self):
+
         all_logs = self.get_option("all_logs")
 
         self.add_copy_spec([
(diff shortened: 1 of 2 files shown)


sos/report/plugins/openshift_ovn.py
```diff
@@ -50,21 +51,32 @@ def setup(self):
...
-        self.add_cmd_output([
-            'ovs-appctl -t /var/run/ovn/ovn-controller.*.ctl ' +
-            'ct-zone-list'],
-            container='ovnkube-controller')
+            container='ovnkube-master',
+            runtime='crio')
+        # We need to determine the actual file name to send
+        # to the command
+        files = glob.glob("/var/run/ovn/ovn-controller.*.ctl")
+        for file in files:
```
TurboTurtle, line 59: Does this collection not behave correctly when directly specifying the glob? Is that why we're also changing this to a per-file iteration?

sos/report/plugins/openshift_ovn.py
```diff
@@ -50,21 +51,32 @@ def setup(self):
...
+            self.add_cmd_output([
+                'ovs-appctl -t ' + file +
+                ' ct-zone-list'],
+                container='ovnkube-node',
+                runtime='crio')
+            self.add_cmd_output([
+                'ovs-appctl -t ' + file +
+                ' ct-zone-list'],
+                container='ovnkube-controller',
+                runtime='crio')
```
TurboTurtle: `a_c_o()` takes a list or a string. With this change now doing single files at a time the list is superfluous.

Also, please convert these to f-strings. A while back we made an effort to convert everything we could to f-strings, but I guess we overlooked this.

TurboTurtle (approved):

sos/report/plugins/openshift_ovn.py
dwolstroRH: When this command was getting executed, the filename '/var/run/ovn/ovn-controller.*.ctl' parameter within the command string was not getting expanded. The '*' represents the current process id of the container, ie, '/var/run/ovn/ovn-controller.3127.ctl', which lead to a "File Not Found Error" condition when executing the command. I changed to find the filename before passing the command string into add_cmd_output.

sos/report/plugins/openshift_ovn.py
dwolstroRH: These are separate calls because they target different containers and the container parameter is not part of the list. These calls previously existed, I simply updated them to pass the file name correctly. I will update the 3 calls that I changed to use f-strings.

TurboTurtle (approved):
LGTM. If you could please rebase to current `main`, we'll get this merged.

Issue #4082, opened 2025-07-17
Upload fails


sbernhard:
This shouldn't be hard coded:
https://github.com/sosreport/sos/blame/main/sos/upload/__init__.py#L267

In case you have a own distro policy and a own upload target, but its not part of the list above, the upload will fail. 
Importing the sos uplod targets should work without hard-coded values in the init method.

pmoravec:
The code works well (maybe in an ugly way) for the upstream `sos` and its policies. Am I right you wanted to add some custom policy and that failed due to the description above?

Could the policy be added to upstream `sos` (and to the list https://github.com/sosreport/sos/blame/main/sos/upload/__init__.py#L267-L268)?

But I agree it would be better to load relevant policy(/ies) rather than load everything and then remove one by one. Are you able to propose a PR with the change?

sbernhard:
would it be possible to load only the sos/upload/target class and not the Policy? Maybe just to iterate throw the upload classes, add them to the supported list but load just afterwards - it should not be in the modules[] array.

PR #4083, opened 2025-07-17
[upload] Only load sos upload targets and not its policies

sbernhard:
Use importlib.import_module(modname) instead of python internal __import__ method

Closes: #4082


pmoravec (approved):
Nice, thanks!

TurboTurtle (approved):

TurboTurtle:
Build failure looks unrelated to the change, and other builds passed. LGTM.

PR #4087, opened 2025-07-21
[presets] Ensure that presets override config files

jcastill:
Per the man pages:

'In other words, config files will override defaults, presets override config files, and command line values override presets and config files.'

This commit is an attempt t on ensure that the
any option in a preset has precedence/overrides
an existing option in the config file.

Related: RHEL-104463

---

### DIFF
--- a/sos/options.py
+++ b/sos/options.py
@@ -28,7 +28,7 @@ def str_to_bool(val):
 
 class SoSOptions():
 
-    def _merge_opt(self, opt, src, is_default):
+    def _merge_opt(self, opt, src, is_default, prefer_new):
         def _unset(val):
             return (val == "" or val is None)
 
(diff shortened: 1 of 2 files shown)


jcastill:
I'm checking the too-many-boolean-expressions linting error, will push a new change as soon as possible

sos/options.py
```diff
@@ -41,7 +41,8 @@ def _unset(val):
             # - non-sequential variable keeps its default value
             if (_unset(oldvalue) and not _unset(newvalue)) or \
                is_default or \
-               ((opt not in self._nondefault) and (not _is_seq(newvalue))):
+               ((opt not in self._nondefault) and (not _is_seq(newvalue))) or \
+               prefer_new:
```
jcastill, line 46: Two possible options I can see here:

1. Group is_default and prefer_new before the if, or maybe the _unset() ones, and add them as a single option, or
2. Add # pylint: disable=too-many-boolean-expressions to the if.

pmoravec: I personally vote for 2nd option. 3 boolean operators are still easy to eval. 8 possible of True/False combinations isnt much yet :)

jcastill: Ack, I'll push a new version based on the second option

jcastill:
centos-stream-9 failure probably fixed via #4139

pmoravec:
> centos-stream-9 failure probably fixed via #4139

Good catch - could you pls. rebase this PR to `main` if it really makes the test green?

jcastill:
Done

sos/options.py
pmoravec, line 245: Please add here the new param.

Issue #3992, opened 2025-04-16
sos collect tgz/xz compression


palash3:
sos collect should not do a tgz or xz compression of the already compressed sos reports. It just adds to computation time and does none of the compression.

Atleast user should have option to make sos collect as a tar bundle.

pmoravec:
Typical content of a sos collector tarball:

    # tar tJvf /var/tmp/sos-collector-2025-04-23-xqqpb.tar.xz
    drwx------ root/root         0 2025-04-23 13:29 sos-collector-2025-04-23-xqqpb/sos_reports/
    -rw------- root/root      2727 2025-04-23 13:29 sos-collector-2025-04-23-xqqpb/sos_reports/manifest.json
    drwx------ root/root         0 2025-04-23 13:29 sos-collector-2025-04-23-xqqpb/sos_logs/
    -rw------- root/root      5109 2025-04-23 13:29 sos-collector-2025-04-23-xqqpb/sos_logs/sos.log
    -rw------- root/root      2494 2025-04-23 13:29 sos-collector-2025-04-23-xqqpb/sos_logs/ui.log
    drwx------ root/root         0 2025-04-23 13:29 sos-collector-2025-04-23-xqqpb/
    -rw------- root/root      6616 2025-04-23 13:29 sos-collector-2025-04-23-xqqpb/sosreport-pmoravec-rhel91-2025-04-23-satdudd.tar.xz
    -rw------- root/root      6648 2025-04-23 13:29 sos-collector-2025-04-23-xqqpb/sosreport-pmoravec-rhel92-2025-04-23-vpkhlav.tar.xz
    -rw------- root/root      6584 2025-04-23 13:29 sos-collector-2025-04-23-xqqpb/sosreport-pmoravec-rhel93-2025-04-23-vrkftje.tar.xz
    #

While most of the tarball is already compressed, a small amount of data is not compressed. But compressing it is more time pain than size gain, indeed.

Historically, `sos` is assumed to provide a `tar.xz` (or `.gz`) file as its outcome. Your request would break this but it has a lot of sense to me.

Are there some use cases when an individual sosreport within that collect is *not* compressed, @TurboTurtle ?

TurboTurtle:
Yeah, this is mainly done to maintain consistency in archive generation. Back when `collect` was first being developed we didn't note significant time deltas, which is why `archive` was never updated to allow for no compression when it was brought in as a component from the standalone tool.

[...]

PR #4088, opened 2025-07-23
[collect] Do not compress top-level collector archive

TurboTurtle:
Previously, when running `sos collect`, the resulting archive that contained sos report archives, would itself be compressed. This "top level" compression did not actually gain us much, but did have a time penalty to overall collect execution times.

This commit solves this by allowing a "none" value (as a string) to be used for the compression method when calling `Archive.finalize()`, which will skip compressing that archive. The actual sos reports that a collect execution captures are still compressed.

Note: this "none" option is not currently exposed on the CLI, and all sos reports will still be compressed.

Resolves: #3992

---


TurboTurtle:
Thinking on it more, since this isn't exposed via the CLI, just using `None` directly makes more sense. If we ever expose this via `-z`, we can either convert then or just coerce the CLI arg.

pmoravec (approved):

PR #4091, opened 2025-07-25
[networking] Capture /etc/services file

jcastill:
Capture /etc/services file

Related: RHEL-105407

---


arif-ali (approved):
lgtm, I would have thought that this file would be the same; would we expect this to change?

jcastill:
@arif-ali that's what I thought, that it never changed. But the guys in support told me that that's not the case, and some customers actually modify them for their own applications.

PR #4092, opened 2025-07-28
[upload] Enable sos uploads of files bigger than 1Gb

dwolstroRH:
Uploads of files greater than 1Gb switch to go to sftp. The code incorrectly tried to write to the home directory in sftp, but should switch into the /{user} directory instead so proper permissions exist.

Related: RHEL-52919

---

### DIFF
--- a/sos/upload/targets/__init__.py
+++ b/sos/upload/targets/__init__.py
@@ -465,7 +465,7 @@ class UploadTarget():
                 self.upload_password or
                 self._upload_password)
 
-    def upload_sftp(self, user=None, password=None):
+    def upload_sftp(self, user=None, password=None, user_dir=None):
         """Attempts to upload the archive to an SFTP location.
 
         Due to the lack of well maintained, secure, and generally widespread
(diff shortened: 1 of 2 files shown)


TurboTurtle:
I'm assuming this is for the Red Hat Support sftp failover? In that case, the change would need to be made in the redhat target, e.g. `sos/upload/targets/redhat.py`, as the current change is making this the setup for _all_ sftp uploads, which may not mimic RH's configuration.

sos/upload/targets/__init__.py
```diff
@@ -540,8 +540,9 @@ def upload_sftp(self, user=None, password=None):
             raise Exception("Unable to connect via SFTP to "
                             f"{self.get_upload_url_string()}")
 
+        # certain implementations require file to be put in the user dir
         put_cmd = (f'put {self.upload_archive_name} '
-                   f'{self._get_sftp_upload_name()}')
+                   f'{user_dir}/{self._get_sftp_upload_name()}')
```
TurboTurtle: The trailing `/` might throw off instances where no `user_dir` is set. E.G. `put sos.tar mysos.tar` would become `put sos.tar /mysos.tar`.

What about:
```python
put_cmd = f"put {self.upload_archive_name} {f'{user_dir}/' if user_dir else ''}{self._get_sftp_upload_name()}"
```

Then in the Red Hat class we'd have:

```python
return super().upload_sftp(user=_user, password=_token, user_dir=_user)
```

arif-ali (approved):

Issue #3919, opened 2025-02-04
Cleaner ignores hostnames in SSL certificates


pmoravec:
User story: Cleaner was applied to a sosreport. Since the sosreport was taken from a RedHat distro, `/etc/pki/consumer/cert.pem` was collected by `dnf` plugin. The certificate file contains plain FQDNs. Example from my system (`some.domain` was manually obfuscated by myself):

```
# openssl x509 -in /etc/pki/consumer/cert.pem -text
..
        Issuer: C=US, ST=North Carolina, L=Raleigh, O=Katello, OU=SomeOrgUnit, CN=pmoravec-sat615.some.domain.redhat.com
..
            X509v3 Subject Alternative Name: 
                DirName:/O=RedHat/CN=5b9d4751-2720-42fa-846f-7bdd62074ab5, DirName:/CN=pmoravec-rhel9.some.domain.redhat.com
..
#
```

The file is not binary, so cleaner does not remove it. But the hostnames are straighforwardly "encoded" within the file.

Can be there some solution, at all? I ruled out a few ideas already..

Imho the best we can do is cleaner raising a warning "hey, I leave untouched a certificate that may contain sensitive info".

pafernanr:
Not sure if this could fit: 
- Replace cert.pem with cert.pem-text using the `openssl x509 -in /etc/pki/consumer/cert.pem -text` output and obfuscate it.

This implies `*.pem` files become "unusable" for `openssl` commands, although I guess this is expected/desired.

pmoravec:
That sounds good to me, or at least I dont have a better idea. I think a good place for this would be around `keep_binary_files` test at https://github.com/sosreport/sos/blob/142ddb42c34b1daa70900f24c0ec0e1bf5ed43a0/sos/cleaner/__init__.py#L692. Sill I see some concerns:

- should we have a cmdline option to keep the certificate files and not malforming them, even in cleaner? Sometimes it is useful to get the cert file itself, not only its textual content.
- is there a reliable way to identify SSL certificates?
- what if cleaner fails to call `openssl` command or the tool isnt present..?

@TurboTurtle , does the idea from Pablo sound good to you?

TurboTurtle:
I think it's a reasonable approach for when users definitely want that content, however I hesitate to make it a default because there are certainly environments/organizations out there that would not want _any_ part of their certs discernable (whether or not that's a valid posture to take is tangent to what we're talking about here).

I like an approach where we have these options:

[...]

PR #4093, opened 2025-07-29
[cleaner] ignores hostnames in SSL certificates

pafernanr:
Refer to [3919](https://github.com/sosreport/sos/issues/3919)

- New `clean` argument added `--treat-certificates` with next options:
  - obfuscate: (Default) convert certificates to text and let  the method `obfuscate_report` do its job.
  - remove: Remove certificate files from the archive.
  - keep: Don't touch certificate files and keep them as is in the archive.

- Notes:
  - Certificate files are identified by extension (.csr, .pem) and header:
    -  r'-----BEGIN CERTIFICATE-----'
    -  r'-----BEGIN [A-Z]+ PRIVATE KEY-----'
  - If `openssl` is not available and `obfuscate` is selected. It will be replaced by `remove` after show a WARNING message.
  - Certificate Key files are always removed.

---


sos/cleaner/__init__.py
```diff
@@ -562,6 +571,31 @@ def obfuscate_report_paths(self):
...
+                    "WARNING: certificate files that potentially contain "
+                    "sensitive information will be KEPT in the final "
+                    "archive as is.\n"
+                )
+            elif self.opts.treat_certificates == "remove":
+                self.ui_log.warning(
+                    "WARNING: certificate files that potentially contain "
+                    "sensitive information will be REMOVED in the final "
+                    "archive.\n"
+                )
```
pmoravec: Just a thought / option to consider: do we really want a WARNING message (either one of those three) printed on every cleaner run?

sos/cleaner/__init__.py
```diff
@@ -696,6 +730,19 @@ def obfuscate_report(self, archive):  # pylint: disable=too-many-branches
                     # is used. In this case we want to make sure
                     # the cleaner doesn't try to clean a binary file
                     continue
+                is_certificate = file_is_certificate(fname)
+                if is_certificate:
+                    if is_certificate == "certificatekey":
```
pmoravec: In theory, key files should have been removed by report already. But surely they can remain also here, makes sense to remove them despite they dont usually contain any *cleaner*-sensitive data.

sos/cleaner/__init__.py
```diff
@@ -696,6 +730,19 @@ def obfuscate_report(self, archive):  # pylint: disable=too-many-branches
...
+                        # Always remove certificate Key files
+                        archive.remove_file(short_name)
+                        continue
+                    if self.opts.treat_certificates == "keep":
+                        continue
+                    if self.opts.treat_certificates == "remove":
+                        archive.remove_file(short_name)
```
pmoravec: Originally, I thought we should put some debug message here "removing certificate file" or similar.

But a better approach is updating `remove_file` method in `sos/cleaner/archives/__init__.py` to have extra argument `filetype=binary` and replace:

    self.log_info(f"Removing binary file '{fname}' from archive")

to:
    self.log_info(f"Removing {filetype} file '{fname}' from archive")

(and call the method from here with proper argument)

sos/cleaner/archives/__init__.py
```diff
@@ -150,6 +150,16 @@ def is_tarfile(self):
...
 
+    def certificate_to_text(self, fname):
+        """Convert a certificate to text. This is used when cleaner encounters
+        a certificate file and the option 'treat_certificates' is 'obfuscate'.
+        """
+        self.log_info(f"Converting certificate file '{fname}' to text")
+        sos_get_command_output(
+            f"openssl x509 -noout -text -in {str(fname)}",
+            to_file=f"{fname}.text")
+        shutil.move(f"{fname}.text", fname)
```
pmoravec: Just a thought / option to consider: should we keep original file's permissions and ownership? We are overwriting it here (which makes sense as well as we update the file content).

sos/cleaner/__init__.py
```diff
@@ -289,6 +290,14 @@ def add_parser_options(cls, parser):
         clean_grp.add_argument('--usernames', dest='usernames', default=[],
                                action='extend',
                                help='List of usernames to obfuscate')
+        clean_grp.add_argument('--treat-certificates', default='obfuscate',
+                               choices=['obfuscate', 'keep', 'remove'],
+                               dest='treat_certificates',
+                               help=('How to treat certificate files.'
```
pmoravec: nitpick: Extra dot at the end. Same applies on two equivalent places later on.

sos/utilities.py
```diff
@@ -132,6 +132,28 @@ def convert_bytes(num_bytes):
...
+    :returns:   The type of the certificate or ``None``
+    :rtype:     ``string`` or ``None``
+    """
+    if fname[-4:] in [".csr", ".pem"]:
+        with open(fname, 'r', encoding='utf-8') as f:
+            if re.search(r'-----BEGIN CERTIFICATE-----', f.read()):
+                return "certificate"
+            if re.search(r'-----BEGIN [A-Z]+ PRIVATE KEY-----', f.read()):
+                return "certificatekey"
+        return None
```
pmoravec, line 154: Just a thought / option to consider: isnt this return redundant?

pmoravec:
The foreman tests failing is some test setup or infra issue - I am checking it.

sos/utilities.py
pafernanr: Yes, its is :) was added just to make lint happy

sos/cleaner/archives/__init__.py
```diff
@@ -289,6 +305,16 @@ def is_tarfile(self):
         except Exception:
             return False
 
+    def certificate_to_text(self, fname):
+        """Convert a certificate to text. This is used when cleaner encounters
+        a certificate file and the option 'treat_certificates' is 'obfuscate'.
+        """
+        self.log_info(f"Converting certificate file '{fname}' to text")
+        sos_get_command_output(
+            f"openssl x509 -noout -text -in {str(fname)}",
```
pmoravec: I tested it on a CA chain, and sadly the collected output contains only the very first certificate.

E.g. try:

~~~
sos report -o subscription_manager --batch --build --clean
~~~

and compare collected `/etc/rhsm/ca/redhat-uep.pem` (one certificate in text) with the original file (3 original certs).

Sadly, I am not aware of an easy solution here. IMHO there is no *one* command that would do the job, and the best I am aware is either of procedures:

~~~
awk '/BEGIN CERTIFICATE/,/END CERTIFICATE/{ if(/BEGIN CERTIFICATE/){i++}; out="individual-"i".pem"; print >out}'
~~~

or:

~~~
csplit -z -f individual- /etc/rhsm/ca/redhat-uep.pem '/-----BEGIN CERTIFICATE-----/' '{*}'
~~~

.. and then calling `openssl` to all of them:

~~~
for cert in individual-*.pem; do
    openssl x509 -noout -text -in ${cert}
done
~~~

I expect we can replace the `awk` or `csplit` approach by some native python RE split, then calling `openssl` to all splited certs and then merging the outputs into one file..?

sos/utilities.py
```diff
@@ -132,6 +132,28 @@
...
+    certificate files content and instead, by default, will keep them as is.
+
+    :param fname:   The full path of the file to check
+    :type fname:    ``str``
+
+    :returns:   The type of the certificate or ``None``
+    :rtype:     ``string`` or ``None``
+    """
+    if fname[-4:] in [".csr", ".pem"]:
+        with open(fname, 'r', encoding='utf-8') as f:
```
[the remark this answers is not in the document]
pmoravec: I dont understand this warning. Same style is used else-where in the same file, e.g.:

~~~
def tail(filename, number_of_bytes):
    """Returns the last number_of_bytes of filename"""
    with open(filename, "rb") as f:
        if os.stat(filename).st_size > number_of_bytes:
            f.seek(-number_of_bytes, 2)
        return f.read()
~~~

sos/cleaner/archives/__init__.py
pafernanr: I think we can use next command instead of the current one:
~~~
openssl storeutl -noout -text -certs /tmp/BUNDLE_FILE.crt
~~~

sos/utilities.py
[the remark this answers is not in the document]
pafernanr: it has no sense, but I suspect move the penultimate return into the "with open" could make lint happy. 
~~~
    if fname[-4:] in [".csr", ".pem"]:
        with open(fname, 'r', encoding='utf-8') as f:
            if re.search(r'-----BEGIN CERTIFICATE-----', f.read()):
                return "certificate"
            if re.search(r'-----BEGIN [A-Z]+ PRIVATE KEY-----', f.read()):
                return "certificatekey"
            return None
    return None
~~~

sos/cleaner/archives/__init__.py
pmoravec: Nice! I hadnt known that option. Neither stackoverflow had :D.

sos/utilities.py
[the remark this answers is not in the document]
pmoravec: That sounds to me like a CodeQL mishap, do you agree @TurboTurtle and @arif-ali ?

pmoravec (approved):
ACK (up to the doubt about CodeQL warning).

Thanks Pablo for applying repetitive feedback! :)

sos/utilities.py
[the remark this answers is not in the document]
bmr-cymru: The CodeQL error is surely wrong, but this code looks off to me. The `file_is_certificate()` function is calling the `read()` method twice on the `File` object `f`: since `read()` without a size argument is greedy the first call will exhaust the file content (returning when EOF is reached). If the first `if re.search()` evaluates to `False` the second call is left with nothing to read and will always return the empty string.

"When size is omitted or negative, the entire contents of the file will be read and returned; it’s your problem if the file is twice as large as your machine’s memory."

This should either read the file once, and store the result in a variable for the two regex checks, or it should call `f.seek(0)` to rewind the file object if you really want to read the entire file twice.:

```python
>>> f = open("/tmp/foo", "r", encoding="utf-8")
>>> f.read()
'foo\nbar\nbaz\nquux\n'
>>> f.read()
''
>>> f.read()
''
>>> f.seek(0)
0
>>> f.read()
'foo\nbar\nbaz\nquux\n'
```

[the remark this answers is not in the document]
pafernanr: thanks for the explanation @bmr-cymru ++, you're right. so I modified it reading the content con a variable.

sos/cleaner/archives/__init__.py
pafernanr: :-) actually I found it  at https://serverfault.com/questions/590870/how-to-view-all-ssl-certificates-in-a-bundle

sos/cleaner/archives/__init__.py
TurboTurtle: My only hangup here is actually dropping the `.text` suffix. I think if we're converting the cert file, then we should make that obvious, and I do like the idea of having a `.text` (or `.txt` if you prefer) appended to the filename.

pmoravec: I agree here, this behaviour can be confusing ("does the pem file really contain just text content, or..?").

Maybe worth appending the new filename to the log `"Converting certificate ..` above?

Gladly we do the change in cleaner and not in report where we would have to update filelist of collected files - here just removing the old file should be enough.

pafernanr:
Hi all, last commit behaves as you requested. `.pem` files are renamed to `.pem.text` when using `sos report --clean --treat-certificates obfuscate`

sos/utilities.py
```diff
@@ -132,6 +132,29 @@ def convert_bytes(num_bytes):
...
+    This is especially helpful for `sos clean`, which cannot obfuscate
+    certificate files content and instead, by default, will keep them as is.
+
+    :param fname:   The full path of the file to check
+    :type fname:    ``str``
+
+    :returns:   The type of the certificate or ``None``
+    :rtype:     ``string`` or ``None``
+    """
+    if fname[-4:] in [".csr", ".pem"]:
```
pmoravec: Can we check against `.crt` suffix as well? E.g. various Satellite-related certificates (though afaik neither of them being collected by sos, in neither of plenty locations) have that suffix.

(tiny update: even sosreport from Satellite can contain `./etc/pki/pulp/qpid/ca.crt` until Qpid broker is deprecated in Sat)

pmoravec:
Apart of the `.crt` suffix, two nitpicks:

- Please merge the two commits
- The WARNING messages shall be wrapped to 80chars like e.g. `print_disclaimer` in `sos/cleaner/__init__.py`

sos/cleaner/__init__.py
```diff
@@ -300,6 +302,14 @@ def add_parser_options(cls, parser):
         clean_grp.add_argument('--usernames', dest='usernames', default=[],
                                action='extend',
                                help='List of usernames to obfuscate')
+        clean_grp.add_argument('--treat-certificates', default='obfuscate',
+                               choices=['obfuscate', 'keep', 'remove'],
+                               dest='treat_certificates',
+                               help=('How to treat certificate files '
+                                     '[.pem, .crt]. Defaults to "obfuscate" '
```
pmoravec: Great to have the .crt mentioned also in help.

TurboTurtle:
Friendly reminder ping to please rebase on current `main` and repush so we can get a sanity run of the test suite.

PR #4096, opened 2025-08-05
[plugin] Add a new option to capture spare files' size with the 'ls' command

jcastill:
Add a new option to capture sparse files' size when running ls.
Also, add this new option to the call in the 'logs' plugin.

---

### DIFF
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -2064,7 +2064,7 @@ class Plugin():
     def add_dir_listing(self, paths, tree=False, recursive=False, chroot=True,
                         env=None, sizelimit=None, pred=None, subdir=None,
                         tags=[], runas=None, container=None,
-                        suggest_filename=None):
+                        suggest_filename=None, extra_opts=None):
         """
         Used as a way to standardize our collections of directory listings,
         either as an output of `ls` or `tree` depending on if the `tree`
@@ -2093,7 +2093,8 @@ class Plugin():
             paths = [p for p in paths if self.path_exists(p)]
 
         if not tree:
-            options = f"alZ{'R' if recursive else ''}"
+            options = (f"alZ{'R' if recursive else ''}"
+                       f"{extra_opts if extra_opts else ''}")
         else:
             options = 'lp'
 
--- a/sos/report/plugins/logs.py
+++ b/sos/report/plugins/logs.py
@@ -56,7 +56,7 @@ class LogsBase(Plugin):
         ])
 
         self.add_cmd_output("journalctl --disk-usage")
-        self.add_dir_listing('/var/log', recursive=True)
+        self.add_dir_listing('/var/log', recursive=True, extra_opts='s')
 
         # collect journal logs if:
         # - there is some data present, either persistent or runtime only


sos/report/plugins/__init__.py
```diff
@@ -2064,7 +2064,7 @@ def _add_cmd_output(self, **kwargs):
     def add_dir_listing(self, paths, tree=False, recursive=False, chroot=True,
                         env=None, sizelimit=None, pred=None, subdir=None,
                         tags=[], runas=None, container=None,
-                        suggest_filename=None):
+                        suggest_filename=None, sparse=False):
```
jcastill: @pmoravec suggested adding an option extra_opts='' (or some better name) to `add_dir_listing` to have more generic way of customizing the method. And calling `options += extra_opts` inside the method. 

I think it's a good idea, othewise we'll find ourselves adding each option from ls as argument to add_dir_listing(). The only issue here is that some of these options are mutually exclusive, so we'll still need to control what gets added and how. My alternative suggestion to the code above is:

```
    def add_dir_listing(self, paths, tree=False, recursive=False, chroot=True,
                        env=None, sizelimit=None, pred=None, subdir=None,
                        tags=[], runas=None, container=None,
                        suggest_filename=None, **other_options):
```

And then:

```
        if not tree:
            options = (f"a{'s' if other_options.get('sparse') else 'l'}"
                       f"Z{'R' if recursive else ''}")
```

Thoughts about this alternative approach?

sos/report/plugins/logs.py
```diff
@@ -57,6 +57,7 @@ def setup(self):
 
         self.add_cmd_output("journalctl --disk-usage")
         self.add_dir_listing('/var/log', recursive=True)
+        self.add_dir_listing('/var/log', recursive=True, sparse=True)
```
jcastill: I thought about substituting the previous call, but that output is still very useful. So I decided to add a new option here and capture both outputs.

sos/report/plugins/__init__.py
```diff
@@ -2093,7 +2093,7 @@ def add_dir_listing(self, paths, tree=False, recursive=False, chroot=True,
             paths = [p for p in paths if self.path_exists(p)]
 
         if not tree:
-            options = f"alZ{'R' if recursive else ''}"
+            options = f"a{'s' if sparse else 'l'}Z{'R' if recursive else ''}"
```
pmoravec: Why dropping `-l` option when you want to use `-s`? These are not mutually exclusive:

```
# ls -alZR sparse-file 
-rw-r--r--. 1 root root unconfined_u:object_r:admin_home_t:s0 1073741824 Aug  4 19:31 sparse-file
# ls -asZR sparse-file 
0 unconfined_u:object_r:admin_home_t:s0 sparse-file
# ls -alsZR sparse-file 
0 -rw-r--r--. 1 root root unconfined_u:object_r:admin_home_t:s0 1073741824 Aug  4 19:31 sparse-file
#
```

sos/report/plugins/__init__.py
pmoravec: The only mutually exclusive options can be set by a programmer calling the `add_dir_listing` method with mutually exclusive options :). We can have some internal logic within the method to prevent that, but till we review any new or modified calls of `add_dir_listing` method, we can prevent the "mutually exclusive options: issue easily.

That is why I suggested (in an internal discussion) some generic option/argument, e.g. directly a string of `ls`/`tree` (concatenated extra) option(s). And for this use case, call `self.add_dir_listing('/var/log', recursive=True, extra_opts='s')`.

If such extension is not seen as being used in future, we can use `sparse` option instead. That one is much more understandable for code readiness (who recalls what `-s` option means?).

sos/report/plugins/logs.py
pmoravec: This (and any other point I made) is subjective and not definitive or strict. Calling almost the same command just to grab one another value for each item seems bit redundant - esp. if we can just extend the current `ls` call to grab all required data.

I see two valid reasons against such extension (here I assume `-s` does not replace `-l` - if that would be the case, let collect both outputs):
- some tooling relies on the current output
- users are used to the traditional output and would be confused by the leading int (see example output above) - I already got one feedback the traditional file should be kept as is

I feel neither the reason is strong enough (.. or the second is?) to collect almost duplicate data (though a small and quickly).

Again, that is just my preference, not much strong. If somebody else would aree with either particular implementation (`sparse` option, `-s` replacing `-l`, call extra `ls` instead of etending the current one), I am ok with it.

sos/report/plugins/__init__.py
jcastill: You are right, it prints both - I misread the man page, thought that it printed only one size. I'll push a new version

sos/report/plugins/logs.py
jcastill: I moved this to only one capture, adding the -s.

sos/report/plugins/__init__.py
jcastill: Ack, will use your approach then.

jcastill:
New version using @pmoravec's suggestion of 'extra_opts'.

arif-ali (approved):

TurboTurtle (approved):

PR #4102, opened 2025-08-14
[loki] Fetch loki logs

mmkay:
Plugin for Loki, a logs backend by Grafana. If run from a Loki container, it fetches Loki internal logs and config (both on paths from a regular container / standalone installation as well as the charmed version). If `-k loki.fetch-logs=true` is passed, it also calls Loki API to fetch logs. 

---

### DIFF
--- a/sos/report/plugins/loki.py
+++ b/sos/report/plugins/loki.py
@@ -1,0 +1,189 @@
+# Copyright (C) 2025 Canonical Ltd.,
+#                    Mateusz Kulewicz <mateusz.kulewicz@canonical.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from typing import Optional
+from datetime import datetime, timedelta, timezone
+from json import JSONDecodeError, dumps, loads
+from sos.report.plugins import Plugin, IndependentPlugin, PluginOpt
+
+
+class Loki(Plugin, IndependentPlugin):
+    """
+    Collects logs and configuration from Loki.
+
+    This plugin interacts with the Loki API to fetch logs based on specified
+    labels and provides options for pagination and label detection.
+    It also collects relevant configuration files and masks sensitive
+    information. It works with both charmed and non-charmed Loki.
+    To fetch internal Loki logs, run it from the Loki container.
+    You can also run it from another machine and fetch only logs from
+    Loki API, by providing the following parameters:
+        `-k loki.collect-logs=true -k loki.endpoint=LOKI_URL`
+
+    Usage:
+        sos report -o loki -k loki.collect-logs=true \
+            -k loki.labels=severity:charm -k loki.detect-labels=true \
+            -k loki.paginate=true -k loki.endpoint=LOKI_URL
+    """
+
+    short_desc = 'Loki service'
+    plugin_name = 'loki'
+    profiles = ('services', )
+    LOKI_QUERY_LIMIT = 5000
+    MAX_PAGINATION_ITERATIONS = 100
+
+    packages = ('loki', )
+
+    option_list = [
+        PluginOpt('collect-logs', default=False,
+                  desc='collect logs from Loki API'),
+        PluginOpt('detect-labels', default=False,
+                  desc=('fetch logs for all available labels. '
+                        'May result in multiple files with the same logs')),
+        PluginOpt('paginate', default=False,
+                  desc='fetch all available logs from Loki API.'),
+        PluginOpt('labels', default='', val_type=str,
+                  desc='colon-delimited list of labels to fetch logs from'),
+        PluginOpt('endpoint', default='http://localhost:3100', val_type=str,
+                  desc=('loki endpoint to fetch logs from. '
+                        'Defaults to http://localhost:3100.')),
+    ]
+
+    def query_command(self, endpoint, label, start: Optional[datetime],
+                      end: Optional[datetime]):
+        if not end:
+            end = datetime.now(timezone.utc)
+        if not start:
+            start = end - timedelta(days=1)
+
+        start_formatted = start.strftime("%Y-%m-%dT%H:%M:%S.%fZ")
+        end_formatted = end.strftime("%Y-%m-%dT%H:%M:%S.%fZ")
+
+        command = (
+            f"curl -G -s '{endpoint}/loki/api/v1/query_range' "
+            f"--data-urlencode 'query={{{label}=~\".+\"}}' "
+            f"--data-urlencode 'start={start_formatted}' "
+            f"--data-urlencode 'end={end_formatted}' "
+            f"--data-urlencode 'limit={Loki.LOKI_QUERY_LIMIT}' "
+        )
+        return command
+
+    def get_logs(self, endpoint, label, start: Optional[datetime],
+                 end: Optional[datetime]):
+        output = self.exec_cmd(self.query_command(endpoint, label, start, end))
+        try:
+            return loads(output["output"])
+        except JSONDecodeError:
+            # if an error is returned from Loki API, the output will be str
+            self._log_warn((f"An error was returned from Loki API on label "
+                            f"{label}. "
+                            f"Error message stored, not querying further."))
+            return output["output"]
+
+    def get_earliest_log_timestamp(self, logs):
+        log_streams = logs["data"]["result"]
+        # use now as a comparison
+        earliest_log = int(datetime.now().timestamp()*1_000_000_000)
+        for stream in log_streams:
+            for log in stream["values"]:
+                timestamp = int(log[0])
+                earliest_log = min(earliest_log, timestamp)
+        return earliest_log
+
+    def get_logs_for_label(self, endpoint, label, paginate):
+        logs = self.get_logs(endpoint, label, None, None)
+        with self.collection_file(f'{label}.log') as logfile:
+            logfile.write(dumps(logs, indent=2))
+        if isinstance(logs, str):
+            # don't paginate if error was returned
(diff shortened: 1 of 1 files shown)


sos/report/plugins/loki.py
```diff
@@ -0,0 +1,190 @@
...
+# See the LICENSE file in the source distribution for further information.
+
+from typing import Optional
+from datetime import datetime, timedelta, timezone
+import logging
+import json
+from sos.report.plugins import Plugin, IndependentPlugin, PluginOpt
+
+
+logger = logging.getLogger("sos")
```
arif-ali: This is not used in the code, so could be removed.

Alternatively, there is also `self._log_error`, `self._log_warn` and/or `self._log_debug` if any logging is required.

You could probably also remove the `import logging` above too

sos/report/plugins/loki.py
```diff
@@ -0,0 +1,190 @@
...
+                  desc='fetch all available logs from Loki API.'),
+        PluginOpt('labels', default='', val_type=str,
+                  desc='colon-delimited list of labels to fetch logs from'),
+        PluginOpt('endpoint', default='http://localhost:3100', val_type=str,
+                  desc=('loki endpoint to fetch logs from. '
+                        'Defaults to http://localhost:3100.')),
+    ]
+
+    def query_command(
+            self,
```
arif-ali: This would be easier to see, and follow convention we use

```python
    def query_command(self, endpoint, label, start: Optional[datetime]
                      end: Optional[datetime]):
```

sos/report/plugins/loki.py
```diff
@@ -0,0 +1,190 @@
...
+        command = (
+            f"curl -G -s '{endpoint}/loki/api/v1/query_range' "
+            f"--data-urlencode 'query={{{label}=~\".+\"}}' "
+            f"--data-urlencode 'start={start_formatted}' "
+            f"--data-urlencode 'end={end_formatted}' "
+            f"--data-urlencode 'limit={Loki.LOKI_QUERY_LIMIT}' "
+        )
+        return command
+
+    def get_logs(
```
arif-ali: same as above with following our convention

sos/report/plugins/loki.py
```diff
@@ -0,0 +1,190 @@
...
+            logfile.write(json.dumps(logs, indent=2))
+        if paginate:
+            earliest_log = self.get_earliest_log_timestamp(logs)
+            previous_earliest_log = int(
+                datetime.now().timestamp()*1_000_000_000
+            )
+            iterations_count = 0
+            while iterations_count < Loki.MAX_PAGINATION_ITERATIONS and \
+                    earliest_log < previous_earliest_log:
+                new_logs = self.get_logs(
```
arif-ali: we could probably do the same here, with as many arguments on the same line, and then go to the next after the 79 character limit

sos/report/plugins/loki.py
```diff
@@ -0,0 +1,190 @@
...
+                earliest_log = \
+                    self.get_earliest_log_timestamp(new_logs)
+                with self.collection_file(f'{label}.log.{iterations_count}') \
+                        as logfile:
+                    logfile.write(json.dumps(new_logs, indent=2))
+                # exit at most after 100 pages to avoid infinite loops
+                iterations_count += 1
+
+    def setup(self):
+        els_config_file = self.path_join(
```
arif-ali: this could be easily on the same line

sos/report/plugins/loki.py
```diff
@@ -0,0 +1,190 @@
...
+            "/etc/loki/*.yaml"
+        )
+        self.add_copy_spec(els_config_file)
+
+        # charms using cos-coordinated-workers have their config elsewhere
+        coordinated_workers_config_file = self.path_join(
```
arif-ali: same as above

sos/report/plugins/loki.py
```diff
@@ -0,0 +1,190 @@
...
+        self.do_path_regex_sub(
+            r"/etc/loki/*",
+            r"(access_key_id:)(.*)",
+            r"\1[********]"
+        )
+        self.do_path_regex_sub(
+            r"/etc/loki/*",
+            r"(secret_access_key:)(.*)",
+            r"\1[********]"
+        )
```
arif-ali: The above 2 calls, we could use regex for the matching, there a few examples of this in other plugins

sos/report/plugins/loki.py
```diff
@@ -0,0 +1,190 @@
...
+        self.do_path_regex_sub(
+            r"/etc/worker/*",
+            r"(access_key_id:)(.*)",
+            r"\1[********]"
+        )
+        self.do_path_regex_sub(
+            r"/etc/worker/*",
+            r"(secret_access_key:)(.*)",
+            r"\1[********]"
+        )
```
arif-ali: same about regex as above

sos/report/plugins/loki.py
```diff
@@ -0,0 +1,190 @@
...
+        self.add_copy_spec(els_config_file)
+
+        # charms using cos-coordinated-workers have their config elsewhere
+        coordinated_workers_config_file = self.path_join(
+            "/etc/worker/*.yaml"
+        )
+        self.add_copy_spec(coordinated_workers_config_file)
+
+        self.add_copy_spec("/var/log/loki/*")
+        self.add_cmd_output(["pebble logs loki -n 10000"])
```
arif-ali: If it is only one command, then it doesn't need to be a list

sos/report/plugins/loki.py
```diff
@@ -0,0 +1,190 @@
...
+
+        if self.get_option("collect-logs") is True:
```
arif-ali: You could just do

```python
        if self.get_option("collect-logs"):
```

sos/report/plugins/loki.py
```diff
@@ -0,0 +1,190 @@
...
+            endpoint = self.get_option("endpoint") or "http://localhost:3100"
+            labels = []
+            if self.get_option("labels"):
+                labels_option = self.get_option("labels")
+                if isinstance(labels_option, str) and labels_option:
+                    labels.extend(labels_option.split(":"))
+
+            if self.get_option("detect-labels") is True:
```
arif-ali: The same as above on the True 

```python
           if self.get_option("detect-labels"):
```

sos/report/plugins/loki.py
```diff
@@ -0,0 +1,190 @@
...
+        )
+        self.add_copy_spec(coordinated_workers_config_file)
+
+        self.add_copy_spec("/var/log/loki/*")
+        self.add_cmd_output(["pebble logs loki -n 10000"])
+
+        if self.get_option("collect-logs") is True:
+            endpoint = self.get_option("endpoint") or "http://localhost:3100"
+            labels = []
+            if self.get_option("labels"):
```
arif-ali: You could use a walrus operatot

```python
            if labels_option := self.get_option("labels"):
```

Then you can ditch the `labels_option = self.get_option("labels")` the line below

sos/report/plugins/loki.py
mmkay: That was a leftover. Removed now.

sos/report/plugins/loki.py
mmkay: Done in the similar way as in `curtin` as suggested.

sos/report/plugins/loki.py
mmkay: Updated the code here. As the inner expression was causing a rather odd ordering, I've extracted it into a variable.

sos/report/plugins/loki.py
```diff
@@ -0,0 +1,174 @@
...
+            if self.get_option("detect-labels"):
+                labels_cmd = self.collect_cmd_output(
+                    f"curl -G -s '{endpoint}/loki/api/v1/labels'"
+                )
+                labels_json = json.loads(labels_cmd["output"])
+                labels.extend(labels_json["data"])
+
+            for label in labels:
+                paginate = self.get_option("paginate")
+                self.get_logs_for_label(endpoint, label, paginate)
```
TurboTurtle: Since we're doing an actual collection here, this should go into `collect()`, rather than `setup()`. The former is almost entirely for the purpose of using `Plugin.collection_file()` - you can take a look at the `process` or `gcp` plugins for good examples.

This will let the collection be performed with all of our normal guardrails and error handling in place during collections, which is not necessarily present during setup. You could potentially keep the label determinations in `setup()` and then in `collect()` you'd just iterate over that list of labels.

mmkay: Thank you for the clarification with clear pointers! I've moved the use of `collection_file()` into the `collect()` method. I wasn't sure whether it makes sense to move the entire logic to `collect()` but it sounds like the practice is rather to store the regular logs through `setup()` so I've drawn the line at gathering the labels and local logs in `setup()`, then running fetching logs for each label in `collect()`.

I wasn't sure if it's a good practice for these plugins to pass data through `self` like I did with labels here: https://github.com/sosreport/sos/pull/4102/files#diff-46ed7c7304b6203c9944334b0bd9185e4d5fb4635dcc4cd1728885a350538231R156 - please let me know what you think.

Also found one more edge case while moving the code around so now we're able to handle Loki API errors. This might be useful for troubleshooting as well.

sos/report/plugins/loki.py
```diff
@@ -0,0 +1,190 @@
...
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from typing import Optional
+from datetime import datetime, timedelta, timezone
+import json
```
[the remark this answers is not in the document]
arif-ali: It's probably worth fixing this as well, as we're importing `json` module twice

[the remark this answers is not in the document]
mmkay: The double import is now fixed.

arif-ali (approved):
all good now, thanks for your work on this

PR #4105, opened 2025-08-17
[plugins] Fixes #4104 - Improved AAP plugins to capture more useful diagnostics

### LINKED ISSUE #4104: AAP sos report plugin improvements
AAP sos report plugins for gateway, controller, eda and hub are missing capturing some crucial information for troubleshooting.

snagoor:
- **[aap_containerized]** – Resolved an issue where aap_containerized was incorrectly enabled on RPM-based Private Automation Hub servers.

- **[aap_controller]** – Expanded the set of gathered command outputs and now conditionally collect run_wsbroadcast or run_wsrelay depending on the AWX release version.

- **[aap_eda]** – Collected service output details based on the installed EDA version. Starting from AAP 2.5, specific commands are used to obtain service status information.

- **[aap_gateway]** – Added additional command outputs for improved troubleshooting on Gateway servers.

- **[aap_hub]** – Centralized the collection of service information for PAH servers under a single location within the plugin directory.

### DIFF
--- a/sos/report/plugins/aap_controller.py
+++ b/sos/report/plugins/aap_controller.py
@@ -10,6 +10,7 @@
 # See the LICENSE file in the source distribution for further information.
 
 from sos.report.plugins import Plugin, RedHatPlugin
+from sos.utilities import sos_parse_version
 
 
 class AAPControllerPlugin(Plugin, RedHatPlugin):
@@ -22,7 +23,6 @@ class AAPControllerPlugin(Plugin, RedHatPlugin):
                 'automation-controller-ui',
                 'automation-controller')
     commands = ('awx-manage',)
-    services = ('automation-controller',)
 
     def setup(self):
         self.add_copy_spec([
@@ -44,7 +44,8 @@ class AAPControllerPlugin(Plugin, RedHatPlugin):
         ])
 
         self.add_cmd_output([
-            "awx-manage --version",
+            "automation-controller-service status",
+            "awx-manage showmigrations",
             "awx-manage list_instances",
             "awx-manage run_dispatcher --status",
             "awx-manage run_callback_receiver --status",
@@ -58,6 +59,17 @@ class AAPControllerPlugin(Plugin, RedHatPlugin):
             "/var/lib/awx/venv/ansible/bin/pip freeze -l",
             "umask -p",
         ])
+
+        # run_wsbroadcast is replaced with run_wsrelay in AAP 2.4 and above
+        awx_version = self.collect_cmd_output('awx-manage --version')
+        if awx_version['status'] == 0:
+            if (
+                sos_parse_version(awx_version['output']) >
+                sos_parse_version('4.4.99')
+            ):
+                self.add_cmd_output("awx-manage run_wsrelay --status")
+            else:
+                self.add_cmd_output("awx-manage run_wsbroadcast --status")
 
         self.add_dir_listing([
             '/var/lib/awx',
@@ -86,6 +98,5 @@ class AAPControllerPlugin(Plugin, RedHatPlugin):
         jreg = r"(BROADCAST_WEBSOCKET_SECRET\s*=\s*)\"(.+)\""
         repl = r"\1********"
         self.do_path_regex_sub("/etc/tower/conf.d/channels.py", jreg, repl)
-
 
 # vim: set et ts=4 sw=4 :
--- a/sos/report/plugins/aap_eda.py
+++ b/sos/report/plugins/aap_eda.py
@@ -1,4 +1,5 @@
 # Copyright (c) 2025 Rudnei Bertol Jr <rudnei@redhat.com>
+# Copyright (c) 2025 Nagoor Shaik <nshaik@redhat.com>
 
 # This file is part of the sos project: https://github.com/sosreport/sos
 #
@@ -9,6 +10,7 @@
 # See the LICENSE file in the source distribution for further information.
 
 from sos.report.plugins import Plugin, RedHatPlugin
+from sos.utilities import sos_parse_version
 
 
 class AAPEDAControllerPlugin(Plugin, RedHatPlugin):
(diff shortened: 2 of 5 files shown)


sos/report/plugins/aap_eda.py
```diff
@@ -40,14 +41,41 @@ def setup(self):
...
+
         self.add_dir_listing([
             "/etc/ansible-automation-platform/",
             "/var/log/ansible-automation-platform/",
         ], recursive=True)
 
-        self.add_cmd_output("su - eda -c 'export'",
-                            suggest_filename="eda_export")
+        self.add_cmd_output("su - eda -c 'env'",
+                            suggest_filename="eda_environment")
```
TurboTurtle, line 56: Why drop the `export` collection? Is it replaced by `env`? Was it ever correct? If it's a version dependent collection we typically carry both for some time - invalid collections due to command syntax are handled gracefully by sos.

sos/report/plugins/aap_eda.py
```diff
@@ -40,14 +41,41 @@ def setup(self):
...
+
+        pkg_name = 'automation-eda-controller'
+        pkg = self.policy.package_manager.pkg_by_name(f'{pkg_name}')
+        if pkg is not None:
+            eda_pkg_ver = '.'.join(pkg['version'])
+            # EDA version in 2.5 release starts with 1.1.0 version
+            if eda_pkg_ver > '1.0.99':
```
TurboTurtle: Straight comparison of version strings is unreliable. Please use `sos_parse_version` from `sos/utilities.py`.

sos/report/plugins/aap_controller.py
```diff
@@ -59,6 +59,13 @@ def setup(self):
             "umask -p",
         ])
 
+        # run_wsbroadcast is replaced with run_wsrelay in AAP 2.4 and above
+        awx_version = self.collect_cmd_output('awx-manage --version')
+        if awx_version['output'] > '4.4.99':
```
TurboTurtle: Same here, please use the provided utility for version parsing.

sos/report/plugins/aap_controller.py
```diff
@@ -22,7 +22,6 @@ class AAPControllerPlugin(Plugin, RedHatPlugin):
                 'automation-controller-ui',
                 'automation-controller')
     commands = ('awx-manage',)
-    services = ('automation-controller',)
```
TurboTurtle, line 25: Why are we dropping this?

TurboTurtle (commented):
Some highlights follow below.

In addition, please see our [Contributor Guildeines](https://github.com/sosreport/sos/wiki/Contribution-Guidelines) as it pertains to commit messages. "plugin improvements" is vague and without more detail in the commit message body, tracking things going forward becomes difficult.

sos/report/plugins/aap_controller.py
snagoor: We should collect the dedicated `automation-controller-service status` command in favour of this one https://github.com/sosreport/sos/pull/4105/files#diff-0e395eca00bd515ae8c4fa91be0f50a7d667dca95123e1978196837d4e3e1f14R46

sos/report/plugins/aap_eda.py
snagoor: Both captures the same information, `env` is more user-friendly to read. See below snip

`export` output 
~~~
declare -x DBUS_SESSION_BUS_ADDRESS="unix:path=/run/user/989/bus"
declare -x HISTCONTROL="ignoredups"
declare -x HISTSIZE="1000"
declare -x HOME="/var/lib/ansible-automation-platform/eda"
declare -x HOSTNAME="eda.example.com"
declare -x LANG="en_IN.UTF-8"
declare -x LESSOPEN="||/usr/bin/lesspipe.sh %s"
declare -x LOGNAME="eda"
~~~

`env` output
~~~
SHELL=/usr/bin/bash
HISTCONTROL=ignoredups
HISTSIZE=1000
HOSTNAME=eda.example.com
PWD=/var/lib/ansible-automation-platform/eda
LOGNAME=eda
HOME=/var/lib/ansible-automation-platform/eda
LANG=en_IN.UTF-8
~~~

sos/report/plugins/aap_controller.py
```diff
@@ -59,6 +60,13 @@ def setup(self):
             "umask -p",
         ])
 
+        # run_wsbroadcast is replaced with run_wsrelay in AAP 2.4 and above
+        awx_ver = self.collect_cmd_output('awx-manage --version')
+        if sos_parse_version(awx_ver['output']) > sos_parse_version('4.4.99'):
+            self.add_cmd_output("awx-manage run_wsrelay --status")
+        else:
+            self.add_cmd_output("awx-manage run_wsbroadcast --status")
```
pmoravec: Put this under `if awx_ver['status'] == 0:` test. Then the CI tests will pass :) - if one (accidentally) runs this plugin on a system with no `awx-manage` binary, `awx_ver` gets `{'status': 127, 'output': b'', 'truncated': False}` and the binary empty string causes an uncaught exception.

@TurboTurtle , wrt the `b''` output, does it make sense to modify https://github.com/sosreport/sos/blob/main/sos/utilities.py#L367:

    -    stdout = b""
    +    stdout = ""

to prevent such errors? We should guard such situations via the "status is zero?" test, but still..?

snagoor: applied the changes.

pmoravec (approved):

sos/report/plugins/aap_hub.py
```diff
@@ -17,6 +17,7 @@ class AAPAutomationHub(Plugin, RedHatPlugin):
     plugin_name = 'aap_hub'
     profiles = ('sysmgmt', 'ansible',)
     packages = ('automation-hub',)
+    services = ('nginx', 'redis',)
```
arif-ali: This means that any machine that has nginx or redis service, this plugin will be run also. Is it worth triggering based of files, like `/etc/ansible-automation-platform/  and/or `/var/log/nginx/automationhub.access.log*`

snagoor: @arif-ali you are absolutely right, i missed this one. I guess the `packages` part should be enough to auto-enable this on PAH server. 

I moved the service status capture to the bottom.

arif-ali: if both the redis and the nginx plugins will be triggered, is it worth duplicating the collection of the service here?

snagoor: @arif-ali i guess this plugin won't be enabled on any systems with just `nginx` or `redis` service running (i have removed the services part).

sos/report/plugins/aap_gateway.py
```diff
@@ -48,4 +56,18 @@ def postproc(self):
             jreg,
             repl)
 
+        # Mask PASSWORD from print_settings command
+        jreg = r'((["\']?PASSWORD["\']?\s*[:=]\s*)[rb]?["\'])(.*?)(["\'])'
```
pmoravec, line 60: PASSWORD is separated by `:=`while SECRET_KEY is separated by `=`, right? Just double-checking the difference in syntax.

pmoravec (approved):

sos/report/plugins/aap_gateway.py
snagoor: @pmoravec this is the sample data from the test lab https://gist.github.com/snagoor/d647c502e290a362adedf0c5f0cd3164

Here 2 key feilds that needs to be masked 1) under `DATABASES` > `PASSWORD` 2) `SECRET_KEY`

pmoravec: Ah I missed the `[]` there, so this r.e. matches both password occurrences, good.

pmoravec (approved):

TurboTurtle (approved):

PR #4108, opened 2025-08-19
[openstack_nova] Improve scrubbing

pmoravec:
Improve postproc obfuscation in two ways:
- apply postproc also to /var/lib/openstack/config/nova on RedHatNova
- obfuscate just password from transport_url, not the whole URL

Closes: #4108

---

### DIFF
--- a/sos/report/plugins/openstack_nova.py
+++ b/sos/report/plugins/openstack_nova.py
@@ -29,6 +29,7 @@ class OpenStackNova(Plugin):
     var_puppet_gen = "/var/lib/config-data/puppet-generated/nova"
     service_name = "openstack-nova-api.service"
     apachepkg = None
+    postproc_dirs = ["/etc/nova/",]
 
     def setup(self):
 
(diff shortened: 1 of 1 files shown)


pmoravec:
We can't directly use `do_paths_http_sub` method since the format of the `transport_url` is like:

    transport_url=rabbit://user:password@some.hostname:5671/?ssl=1

i.e. not starting with `http`.

sos/report/plugins/openstack_nova.py
```diff
@@ -214,6 +215,7 @@ class RedHatNova(OpenStackNova, RedHatPlugin):
     apachepkg = "httpd"
     nova = False
     packages = ('openstack-selinux',)
+    postproc_dirs = ["/etc/nova/", "/var/lib/openstack/config/nova"]
```
arif-ali, line 218: what about just adding to the list, if the common directories increase, then you don't have to change it here? I know it's unlikely, but something to think about

sos/report/plugins/openstack_nova.py
```diff
@@ -29,6 +29,7 @@ class OpenStackNova(Plugin):
     var_puppet_gen = "/var/lib/config-data/puppet-generated/nova"
     service_name = "openstack-nova-api.service"
     apachepkg = None
+    postproc_dirs = ["/etc/nova/",]
```
arif-ali, line 32: do we need to comma here, I think we only do that for tuples where it makes sense?

arif-ali (commented):
a couple of comments, otherwise looks good to me

sos/report/plugins/openstack_nova.py
pmoravec: Indeed, it is obsolete here for the sake of iterating through the list. I added it there just in case we expand the list, to prevent implicit concatenation of adjacent string items.

sos/report/plugins/openstack_nova.py
pmoravec: You mean always iterate in `apply_regex_sub` method over:

    for _dir in ["/etc/nova/", "/var/lib/openstack/config/nova"]:

? Otherwise I dont understand.

arif-ali: I meant

```python
postproc_dirs += ["/var/lib/openstack/config/nova"]
```

or

```python
postproc_dirs.extend(["/var/lib/openstack/config/nova"])
```

pmoravec:
ok I will enhance the `do_paths_http_sub` method by making protocol as a parameter.

PR #4113, opened 2025-08-26
[spyre] new plugin to collect IBM spyre card data

sourabhjains:
Spyre chip is IBM’s AI accelerator, designed to handle AI inferencing and workloads.

The Spyre plugin collects data about the Spyre card’s VFIO device node tree, configuration files, and more.

---

### DIFF
--- a/sos/report/plugins/spyre.py
+++ b/sos/report/plugins/spyre.py
@@ -1,0 +1,74 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+#
+# This plugin enables collection of logs for system with IBM Spyre card
+
+import re
+import pyudev
+
+from sos.report.plugins import Plugin, IndependentPlugin
+
+
+class Spyre(Plugin, IndependentPlugin):
+    """Spyre chip is IBM’s AI accelerator, designed to handle AI inferencing
+    and workloads.
+
+    The Spyre plugin collects data about the Spyre card’s VFIO device node
+    tree, configuration files, and more.
+    """
+
+    short_desc = 'IBM Spyre Accelerator Information'
+    plugin_name = 'spyre'
+    architectures = ('ppc.*',)
+
+    @staticmethod
+    def get_spyre_cards_bus_ids():
+        context = pyudev.Context()
+        spyre_cards_bus_ids = []
+        card_vendor_ids = ["0x1014"]    # IBM : 0x1014
+        card_device_ids = ["0x06a7", "0x06a8"]  # spyre card device IDs
+
+        for device in context.list_devices(subsystem='pci'):
+            vendor_id = device.attributes.get("vendor").decode("utf-8").strip()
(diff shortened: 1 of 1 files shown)


sos/report/plugins/spyre.py
```diff
@@ -0,0 +1,74 @@
...
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+#
+# This plugin enables collection of logs for system with IBM Spyre card
+
+import re
+import pyudev
```
pmoravec: You can't rely on presence of this library, and default CentOS Stream 9 (or e.g. Fedora 42) installation does not have it.

pmoravec (changes requested):
Please get rid of the `pyudev` library usage.

pmoravec:
`re` is fully ok since it is a native Python library. Let me review the code..

sos/report/plugins/spyre.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def setup(self):
+
+        lspci = self.exec_cmd("lspci -n")
+        if lspci['status'] != 0:
+            return
+
+        spyre_cards = self.get_ibm_spyre_devices(lspci['output'])
+
+        # Nothing to collect if spyre card is not found
```
pmoravec, line 80: Neither the `add_copy_spec` config files are worth to be collected in this case?

sos/report/plugins/spyre.py
```diff
@@ -0,0 +1,97 @@
...
+                vendor, device = ids.lower().split(":")
+            except ValueError:
+                continue
+
+            if vendor == "1014" and device in ("06a7", "06a8"):
+                if pci_addr.count(":") == 1:
+                    pci_addr = "0000:" + pci_addr
+                try:
+                    domain, bus, dev_func = pci_addr.split(":")
+                    device, function = dev_func.split(".")
```
pmoravec: You override `device` variable here, is that intentional? I guess the meaning is different and surely it does not harm at all..

sos/report/plugins/spyre.py
sourabhjains: No, actually I don't even want the spyre folder under sos_commands if Spyre cards are not present in the system. This is the reason I used exec_cmd instead of collect_cmd_output.

sos/report/plugins/spyre.py
sourabhjains: Yeah. It is harmless. But might created confusion when one reads the code. I will change that variable name to pci_device.

sos/report/plugins/spyre.py
pmoravec: OK, I understand it is intentional. Some programs / tools we collect data for can be in a semi-broken state such that their config files exist but in an invalid shape, causing nothing is running or mounted or any command execution fails - due to the broken config. Then it is important to collect the config files.

Defering on your knowledge of the subject and explicit answer here, I understand this is not the case here.

pmoravec (approved):

TurboTurtle (approved):
LGTM. CI failures are unrelated and we're looking into them separately with Cirrus.

sourabhjains:
Thanks for accepting the changes @TurboTurtle

PR #4120, opened 2025-09-04
[dnf] Add 'dnf updateinfo list cves' command

jcastill:
Add 'dnf updateinfo list cves' to the dnf plugin.

Related: RHEL-94859

---

### DIFF
--- a/sos/report/plugins/dnf.py
+++ b/sos/report/plugins/dnf.py
@@ -80,6 +80,7 @@ class DNFPlugin(Plugin, RedHatPlugin):
             "dnf list extras",
             "dnf updateinfo info security",
             "dnf updateinfo list --available",
+            "dnf updateinfo list cves",
             "package-cleanup --dupes",
             "package-cleanup --problems",
         ])


sos/report/plugins/dnf.py
```diff
@@ -80,6 +80,7 @@ def setup(self):
             "dnf list extras",
             "dnf updateinfo info security",
             "dnf updateinfo list --available",
+            "dnf updateinfo list cves",
```
pmoravec, line 83: What is the additional value to `dnf updateinfo list --available`?

The `--available` one already prints the errata ID so the details about CVEs and BZs/JIRAs from the `cves` output can be found on the errata page, when required. E.g. for line:

~~~
RHSA-2025:6966  Moderate/Sec.  kernel-uki-virt-5.14.0-570.12.1.el9_6.x86_64
~~~

the page https://access.redhat.com/errata/RHSA-2025:6966 lists the CVEs that the `cve` output provides.

Or do you focus on scenarios where the `cves` output would be malformed due to malformed metadata on the system..? I simply miss the use case when this output can be valuable to extra collect.

jcastill: You are right that `--available` already prints the errata ID, so for a single check that's useful - the support engineer can go to that web page and check everything. But if you are interested in more than one errata, the output of `list cves` saves a lot of time and gives you a better overall review of the system, neatly grouped and in a clearer way that '--available' could do. Or even if you are just interested in the cves for whatever reason and not in the erratas, this new command gives you everything you need.

pmoravec: Hm so this is a request to rather "postprocess" some collected data, than collect anything new. Similarly like we generate some xml output in `sar` plugin.

Here this command takes 3ish seconds to execute.

I dont know if it is worth of the cost..?

jcastill: No, the data about CVEs hasn't been collected at this point. Only this command will provide it, for all the packages installed in the system.
If you are concerned about these 3 seconds, we can gate it behind an option like we do with some of the rpm plugin captures.
If that's still not enough for you, let me know.

pmoravec: I see collecting this data on a border line where pros ~= cons. So I am deferring to other reviewers for their opinion.

jcastill: That's why I think gating it with an option (default off) could help here

gmcnealy: You are correct, the list cves and list --available commands are redundant, as long as we have the list cves option we dont need to add list --available. The point is to be able to show all errata that affects the system at the current time. In most cases there are hundreds of CVE and RHSA that apply to the system and referencing that many in the CVE database or on each individual CVE or RHSA page is impossible for engineers to do manually.

jcastill: Thank you for the feedback @gmcnealy . I'll close this PR then since we already have the output we need.

PR #4121, opened 2025-09-05
[rhc] Block the 'rhc' plugin starting the rhsm service

jcastill:
When the command 'rhc status' runs in the plugin rhc, it starts automatically the rhsm service. This
commit gates the command to make sure it's not
run when the rhssm service is not running.

Related: RHEL-112563

---

### DIFF
--- a/sos/report/plugins/rhc.py
+++ b/sos/report/plugins/rhc.py
@@ -8,7 +8,7 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.report.plugins import Plugin, RedHatPlugin
+from sos.report.plugins import Plugin, RedHatPlugin, SoSPredicate
 
 
 class Rhc(Plugin, RedHatPlugin):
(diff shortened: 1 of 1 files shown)


sos/report/plugins/rhc.py
```diff
@@ -30,9 +30,8 @@ def setup(self):
             "/var/log/rhc-worker-playbook",
         ])
 
-        self.add_cmd_output([
-            "rhc status",
-        ])
+        rhc_pred = SoSPredicate(self, services=["rhsm"])
```
jcastill: Do we prefer the predicate as a separated variable or should I just integrate it with the command capture directly?

pmoravec: Either is fine. If formatting allows, I prefer direct "integration" as that *usually* means no extra variable to be maintained, hence slightly faster code execution. But hell knows if Python does not detect the variable is one-timer and does not optimise the code the same way..

pmoravec (approved):

sos/report/plugins/rhc.py
bmr-cymru: My rule of thumb was "if it's used more than once, stash it in a variable, otherwise initialise it in-line". You're only using it in a single call so imho in-line is more appropriate here (but that's just my take).

jcastill: Thank you both. I'll integrate the predicate in-line then.

bmr-cymru (approved):

PR #4130, opened 2025-09-15
new plugin to collect some specific metadata from an AWS instance

stojedna:
---


liangxiao1:
The new plugin access the metadata via IMDSv1, it is required IMDSv2 support and try firstly.
Without this support, when launch the instance with IMDSv2 only,  the snippet does not work.
Here is the doc for your reference.
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html

sos/report/plugins/aws.py
```diff
@@ -0,0 +1,65 @@
...
+# See the LICENSE file in the source distribution for further information.
+
+import os
+import subprocess
+from sos.report.plugins import Plugin, IndependentPlugin
+
+class Aws(Plugin, IndependentPlugin):
+    short_desc = 'AWS EC2 instance metadata'
+    profiles = ('virt',)
+    commands = ('curl',)
```
pmoravec: The plugin is automatically enabled on any system, regardless it is an AWS client or not. Please specify some condition(s) when to enable the plugin. Like similar `Azure` plugin is enabled by `packages = ('WALinuxAgent',)`. See https://github.com/sosreport/sos/wiki/How-to-Write-a-Plugin#dependencies for more.

jcastill: maybe [amazon-ssm-agent](https://docs.aws.amazon.com/systems-manager/latest/userguide/ssm-agent-linux.html) ? I think this is both a package and a service that can be checked and could be roughly the equivalent of WALinuxAgent. But someone with more knowledge of AWS may be able to confirm

sos/report/plugins/aws.py
```diff
@@ -0,0 +1,65 @@
...
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+import os
```
jcastill: You can probably remove this import since you are not using it anymore

sos/report/plugins/aws.py
stojedna: At first, I thought about awscli, but it may or may not be installed. This is why I added a check in '/sys/...'
I'll check about the amazon-ssm-agent as @jcastill suggests.

sos/report/plugins/aws.py
stojedna: True that, thanks!

sos/report/plugins/aws.py
pmoravec: (btw I plan to enhance the doc by list of types of triggers (other than `files` and `packages`) to be more verbatim there)

stojedna: The 'amazon-ssm-agent' seems not to be mandatory, so it is not an option for that. Same for the 'amazon-cloudwatch-agent'. As far as I know, there is no AWS specific package that is installed by default in all instances.

liangxiao1: amazon-ssm-agent is popular on aws, but it is not in RHEL repository and installed by default in official RHEL AMIs.
I do not see any aws specific pkg(must have) for this purpose.
cloud-init may be a good candidate for it, it is not AWS specific pkg but can exclude some systems. _is_ec2() is also helpful.

jcastill: I can see this cloud-init file in an AWS machine:

`/run/cloud-init/cloud-id-aws`

Is this the one you are referring to @liangxiao1 ?
We could add multiple conditions and activate the plugin if any is present, so something like:

```
packages = ('amazon-ssm-agent',)
services = ('amazon-ssm-agent',)
files = ('/run/cloud-init/cloud-id-aws',)  # or any other path to this file
```
Could cover a good chunk of installations

liangxiao1: It does make sense if it can take files as condition. Thanks.

stojedna: I understand the need to add a hard condition to keep the sos execution as light as possible, but these conditions don't cover the entire spectrum, and there will be instances where the plugin won't be executed. This is why I added the '_is_ec2' part.

jcastill: @stojedna , in that case maybe the logic in the function _is_ec2() can be moved onto check_enabled(). The advantage of using check_enabled() is, as far as I know, that it will run as a condition checker and the plugin wont run if it returns False, vs how is it now with _is_ec2(), where the plugin will run every time (@pmoravec can you confirm this?)

bmr-cymru: In that case just implement a custom `check_enabled()` method in the plugin that returns a call to your `_is_ec2()` method. See plugins `gcp`, `mellanox_firmware`, `navicli` as examples of other plugins that do this (`git grep 'def check_enabled' sos/report/plugins/`).

stojedna: I've added the check_enabled. Thank you all for your help :)

pmoravec: Now, the line `commands = ('curl',)` is obsolete. Since `commands = ` specifies a list of binaries that - whenever a binary is present on the system - triggers automatic enabling of the plugin (it works likewise `packages = ..`).

Can you please remove this line?

sos/report/plugins/aws.py
```diff
@@ -0,0 +1,68 @@
...
+        if not self._is_ec2():
+            self.soslog.info("Not an EC2 instance; skipping AWS metadata collection")
+            return
+
+        # Using IMDSv2 if possible, if not, going IMDSv1
+        # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html
+
+        # Try to get an IMDSv2 token
+        token_url = 'http://169.254.169.254/latest/api/token'
+        token_cmd = ['curl', '-sS', '-X', 'PUT', '-H', 'X-aws-ec2-metadata-token-ttl-seconds: 21600', token_url]
```
pmoravec: I bet pylint will fail on `Line too long` (not only here).

Worth running `pycodestyle aws.py` or `pylint aws.py` to detect such formatting issues (our CI uses `pylint`).

sos/report/plugins/aws.py
```diff
@@ -0,0 +1,68 @@
...
+        try:
+            token = subprocess.check_output(token_cmd, text=True, timeout=1).strip()
+        except Exception:
+            token = ''
```
pmoravec, line 52: You can use `self.exec_cmd("curl -Ss -X PUT ......", timeout=1)` e.g. like https://github.com/sosreport/sos/blob/main/sos/report/plugins/xen.py#L24-L25, or potentially call `sos_get_command_output` from `sos/utilities.py` directly - we have auxiliary functions for calling extra commands :)

sos/report/plugins/aws.py
```diff
@@ -0,0 +1,68 @@
...
+        self.add_cmd_output(['curl', '-sS'] + token_header + [meta_url], suggest_filename='aws_metadata_instance-id.txt')
+
+        meta_url = 'http://169.254.169.254/latest/meta-data/instance-life-cycle'
+        self.add_cmd_output(['curl', '-sS'] + token_header + [meta_url], suggest_filename='aws_metadata_instance-life-cycle.txt')
+
+        meta_url = 'http://169.254.169.254/latest/meta-data/instance-type'
+        self.add_cmd_output(['curl', '-sS'] + token_header + [meta_url], suggest_filename='aws_metadata_instance-type.txt')
+
+        meta_url = 'http://169.254.169.254/latest/meta-data/placement/availability-zone-id'
+        self.add_cmd_output(['curl', '-sS'] + token_header + [meta_url], suggest_filename='aws_metadata_availability-zone-id.txt')
```
pmoravec: You might create a list of tuples `(meta_url, suggest_fname)` and iterate over it - it is better to maintain such a list, if there might potentially be changes there. Example (though bit complicated) is https://github.com/sosreport/sos/blob/main/sos/report/plugins/foreman.py#L280 (and L285 there).

stojedna: I did it this way intentionally to keep it simple.

sos/report/plugins/aws.py
```diff
@@ -0,0 +1,68 @@
...
+
+        # Those metadata entries do not include any sensitive information. There is no need to mask any data.
+
```
jcastill: can you add the modeline at the end of the file? I.e.:

`# vim: set et ts=4 sw=4 :
`

sos/report/plugins/aws.py
```diff
@@ -0,0 +1,68 @@
...
+from sos.report.plugins import Plugin, IndependentPlugin
+
+class Aws(Plugin, IndependentPlugin):
+    short_desc = 'AWS EC2 instance metadata'
+    profiles = ('virt',)
+    commands = ('curl',)
+
+    def _is_ec2(self):
+        try:
+            with open('/sys/devices/virtual/dmi/id/sys_vendor') as f:
```
jcastill: Can you add the encoding here after the file name? I.e.:

`encoding='UTF-8'`

sos/report/plugins/aws.py
```diff
@@ -0,0 +1,68 @@
...
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+import subprocess
+from sos.report.plugins import Plugin, IndependentPlugin
+
```
jcastill, line 12: Please add one extra empty line here

sos/report/plugins/aws.py
stojedna: Yeah, both tools report 'line too long', but if you check the code, those lines are not that long, and because they are in one line are more readable.

sos/report/plugins/aws.py
```diff
@@ -0,0 +1,69 @@
...
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, IndependentPlugin
+
+class Aws(Plugin, IndependentPlugin):
+    short_desc = 'AWS EC2 instance metadata'
```
arif-ali, line 16: please add

```python
plugin_name = "aws"
```

sos/report/plugins/aws.py
pmoravec: I understand your point, the lines are better readable in the current syntax rather than in the pylint-compliant multiline way (in this specific case). But:

1) a few suggestions I gave you would help you follow the pylint rules while having the code still pretty well readable.
2) We do (too ridiculously?) insist on the pylint formatting (up to some exceptions, grep for `# pylint`). In general, `pylint` code style recommendations are very sane and we follow that - until the above mentioned exceptions.

stojedna: Understood! I'll do the changes needed.

arif-ali:
The final RED is the DCO, so you need to sign your commit, you should easily be able to do that by running the following command

```
git commit -s --amend --no-edit
```

and then push the updated commit

sos/report/plugins/aws.py
```diff
@@ -0,0 +1,113 @@
...
+        # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html
+
+        # Try to get an IMDSv2 token
+        token_url = 'http://169.254.169.254/latest/api/token'
+        token_cmd = [
+            'curl',
+            '-sS',
+            '-X',
+            'PUT',
+            '-H',
```
arif-ali: what about combining the 5 lines into one?

sos/report/plugins/aws.py
```diff
@@ -0,0 +1,113 @@
...
+        # Issue each metadata request individually
+        meta_url = (
+            'http://169.254.169.254/latest/meta-data/'
+            'hostname'
+        )
+        self.add_cmd_output(
+            ['curl', '-sS'] + token_header + [meta_url],
+            suggest_filename='aws_metadata_hostname.txt'
+        )
+
```
arif-ali: It's what @pmoravec  mentioned, i think it would be cleaner if you had all the items in a list, and then loop through it, something similar to below, I think it would be easier to read. Also, if you need to add anything afterwards, it's a one line change

```python
endpoint = 'http://169.254.169.254/latest/meta-data'

actions = [
    `hostname`.
    'instance-id'.
    'instance-life-cycle',
    'instance-type',
]

for action in actions:
    self.add_cmd_output(
        f'curl -sS {token_header} {url}/{action}',
        suggest_filename=f'aws_metadata_{action}.txt'
    )
```

arif-ali (commented):
some suggestions to improve the flow and readability of the code

arif-ali (approved):
lgtm, thanks for your work on this PR and patience on the reviews, as we did have a few inter mingling ones, and apologies it took so many rounds

Issue #696, opened 2015-12-04
CVE-2015-7529 - predictable tmp files usage


bmr-cymru:
Mateusz Guzik reports that sos-3.x is vulnerable to symbolic link attacks since it uses predictable temporary file names in the configured `--tmp-dir` location. If this location is shared with other users (the default on most distributions) then it is possible for a malicious user to obtain content from the archive and in some circumstances to execute arbitrary commands with administrative privileges.

There are two recommended mitigations for these attacks:

* Use a private temporary directory (all versions, all kernels)

Instead of using the default system temporary directory (`/tmp`, `/var/tmp`) use a private directory that excludes non-administrative users from searching or reading the directory contents, e.g.:

```
# mkdir /var/tmp/sos
# chmod 700 /var/tmp/sos
# sosreport --tmp-dir /var/tmp/sos
```

Note that the `chmod` is not strictly necessary assuming the root user's umask is set appropriately.

* Enable kernel link protection (partial mitigation, requires kernel support)

Recent kernels support the `protected_symlinks` feature that can be used to mitigate this class of attack and this is enable by default by most distributions that support the feature:

```
   /proc/sys/fs/protected_symlinks (since Linux 3.6)
      When the value in this file is 0, no restrictions are placed on following
      symbolic links (i.e., this is the historical behavior before Linux 3.6). 
      When  the value in this file is 1, symbolic links are followed only in
      the following circumstances:

       *  the  filesystem UID of the process following the link matches the
          owner (UID) of the symbolic link (as described in credentials(7),
          a process's filesystem UID is normally the same as its effective UID);

      *  the link is not in a sticky world-writable directory; or

      *  the symbolic link and its parent directory have the same owner (UID)
       A system call that fails to follow a symbolic link because of the above
       restrictions returns the error EACCES in errno.

      The default value in this file is 0.  Setting the value to 1 avoids a
      longstanding class of security issues based on time-of-check,  time-of-use
      races  when accessing symbolic links.
```

Note that relying on the `protect_symlinks` feature does not entirely eliminate risk: an attacker could create a regular file that they own (in place of the output file sos expects to create) and use this to obtain archive content that would not normally be exposed to their user and group ID.

Users of sosreport are recommended to use an alternate `--tmp-dir` setting whenever possible.

A fix for this problem is currently in master and will be included in the next sos release. This may be backported to earlier releases carried by distributions. Concerned users should contact their distribution's normal support channels for information regarding updated packages to fix this flaw.

[...]

PR #4134, opened 2025-09-24
[charmed_mysql] refactor & support k8s env

Deezzir:
This PR introduces a proper K8S variant for the `charmed_mysql` plugin. The collection will use a `kubectl` if it detects a K8S environment and `mysql` pods or the `snap` to collect data.

---

### DIFF
--- a/sos/report/plugins/charmed_mysql.py
+++ b/sos/report/plugins/charmed_mysql.py
@@ -7,27 +7,67 @@
 # See the LICENSE file in the source distribution for further information.#
 
 import os
+import tempfile
 
 from sos.report.plugins import Plugin, PluginOpt, UbuntuPlugin
+from sos.utilities import is_executable
 
 
 class CharmedMySQL(Plugin, UbuntuPlugin):
     """
     The Charmed MySQL plugin is used to collect MySQL configuration and logs
-    from the Charmed MySQL snap package. It also collects MySQL Router
-    and MySQL Shell configuration and logs where available,
-    journal logs for the snap, and snap info.
+    from the Charmed MySQL snap package or K8s deployment.
+    It also collects MySQL Router and MySQL Shell configuration and logs
+    where available, journal logs for the snap, and snap info.
 
     If the `dumpdbs` option is set to `True`, the plugin will also try and
     collect the names of the databases that the user has access to. The
     `mysql` user is used by default, but that can be set using the `dbuser`
     option. When using the `dumpdbs` option, you must then provide the
-    password for the user using the `dbpass` option.
+    password for the user using the `dbpass` option or the `MYSQL_PWD`
+    environment variable.
     """
 
     short_desc = "Charmed MySQL"
     plugin_name = "charmed_mysql"
-    packages = ("charmed-mysql",)
+
+    mysql_queries = [
+        # Get databases user has access to
+        "show databases;",
+
+        # Get unit operations from MySQL to see, for example,
+        # if a unit is stuck on joining the cluster
+        "select * from mysql.juju_units_operations;",
+
+        # Get the cluster group replication status
+        ("select * from performance_schema.replication_group_members "
+            "order by MEMBER_HOST;"),
+
+        # Get connection stats
+        "show global status like \"%conne%\";",
+
+        # Get errors per client and host
+        # Useful for problens like an app disconnectting randomly
+        "select * from performance_schema.host_cache;",
+
+        # Get InnoDB status for any deadlocks, etc.
+        "show ENGINE InnoDB STATUS;",
+    ]
+
+    snap_package = "charmed-mysql"
+    snap_path_common = "/var/snap/charmed-mysql/common"
+    snap_path_current = "/var/snap/charmed-mysql/current"
+
+    kube_cmd = "kubectl"
+    selector = "app.kubernetes.io/name=mysql-k8s"
+
+    conf_paths = {
+        "MYSQL_CONF": "/etc/mysql",
+        "MYSQL_LOGS": "/var/log/mysql",
+        "MYSQL_ROUTER_CONF": "/etc/mysqlrouter",
+        "MYSQL_ROUTER_LOGS": "/var/log/mysqlrouter",
+        "MYSQL_SHELL_LOGS": "/var/log/mysqlsh",
+    }
 
     option_list = [
         PluginOpt(
@@ -42,103 +82,213 @@ class CharmedMySQL(Plugin, UbuntuPlugin):
             "dumpdbs", default=False, val_type=bool,
             desc="Get name of all databases"
         ),
+        PluginOpt(
+            "logs_since", default="48h", val_type=str,
+            desc="How far back to fetch logs with kubectl --since, K8s only"
+        ),
     ]
 
-    def setup(self):
-        # Set default paths for snap
-        snap_path_common = "/var/snap/charmed-mysql/common"
-        snap_path_current = "/var/snap/charmed-mysql/current"
+    def _get_db_credentials(self):
+        db_user = self.get_option("dbuser")
+        db_pass = self.get_option("dbpass")
 
-        # Check if deployment is in a Kubernetes cluster
-        k8s_deploy = any("KUBERNETES" in key for key in os.environ)
+        if "MYSQL_PWD" in os.environ and not db_pass:
+            self.soslog.info(
+                "MYSQL_PWD present: Using MYSQL_PWD environment variable, "
+                "user did not provide password."
+            )
+            db_pass = os.environ["MYSQL_PWD"]
+        elif not db_pass:
+            self.soslog.warning(
+                "dumpdbs_error: option is set, but username and password "
+                "are not provided"
+            )
+            return None, None
 
-        # If in a Kubernetes cluster, set the snap paths to "None" as
-        # MySQL configuration does not exist in snap paths
-        if k8s_deploy:
-            snap_path_common = ""
-            snap_path_current = ""
+        return db_user, db_pass
 
-        # Set the configuration paths
-        conf_paths = {
-            "MYSQL_CONF": f"{snap_path_current}/etc/mysql",
-            "MYSQL_LOGS": f"{snap_path_common}/var/log/mysql",
-            "MYSQL_ROUTER_CONF": f"{snap_path_current}/etc/mysqlrouter",
-            "MYSQL_ROUTER_LOGS": f"{snap_path_common}/var/log/mysqlrouter",
-            "MYSQL_SHELL_LOGS": f"{snap_path_common}/var/log/mysqlsh",
-        }
+    def _determine_namespaces(self):
+        namespaces = self.exec_cmd(
+            f"{self.kube_cmd} get pods -A -l {self.selector} "
+            "-o jsonpath='{.items[*].metadata.namespace}'"
+        )
+        if namespaces['status'] == 0:
+            return list(set(namespaces['output'].strip().split()))
+        return []
 
+    def _get_pod_names(self, namespace):
+        pods = self.exec_cmd(
+            f"{self.kube_cmd} -n {namespace} get pods -l {self.selector} "
+            "-o jsonpath='{.items[*].metadata.name}'"
+        )
+        if pods['status'] == 0:
+            return pods['output'].strip().split()
+        return []
+
+    def _collect_per_namespace(self, namespace, logs_since):
+        kube_cmd = f"{self.kube_cmd} -n {namespace}"
+
+        # Describe the resources
+        self.add_cmd_output([
+            f"{kube_cmd} get all -l {self.selector} -o wide",
+            f"{kube_cmd} describe pods -l {self.selector}",
+        ])
+
+        # Get pod logs
+        self.add_cmd_output(
+            f"{kube_cmd} logs -l {self.selector} --since={logs_since} "
+            "--all-containers=true --prefix --all-pods"
+        )
+
+        mysql_cont = "mysql"
+        pods = self._get_pod_names(namespace)
+
+        # Get the config and logs from each pod
+        dump_files_path = self.get_cmd_output_path()
+        for name, path in self.conf_paths.items():
+            for pod in pods:
+                os.makedirs(f"{dump_files_path}/{pod}/{name}", exist_ok=True)
+                copy_cmd = (
+                    f"{kube_cmd} cp -c {mysql_cont} {pod}:{path} "
+                    f"{dump_files_path}/{pod}/{name}"
+                )
+                self.exec_cmd(copy_cmd)
+
+        self.add_forbidden_path([
+            f"{dump_files_path}/*/MYSQL_CONF/*.pem",
+            f"{dump_files_path}/*/MYSQL_CONF/*.key",
+        ])
+
+        # If dumpdbs is set, then get all databases
+        if self.get_option("dumpdbs"):
+            db_user, db_pass = self._get_db_credentials()
+            if not db_user or not db_pass:
+                return
+
+            opts = f"-h 127.0.0.1 -u{db_user}"
+            sql_cmd = f"mysql {opts} -e"
+            queries = " ".join(
+                query.replace('\"', '\\\"')
+                for query in self.mysql_queries
+            )
+
+            with tempfile.TemporaryDirectory() as tmpdir:
+                pwd_file = "mysql_pwd"
+                pwd_path = tmpdir + "/" + pwd_file
+
+                with open(pwd_path, "w", encoding="utf8") as f:
+                    f.write(db_pass)
+
+                for pod in pods:
+                    mkdir_cmd = (
+                        f"{kube_cmd} exec -c {mysql_cont} {pod} -- "
+                        f"mkdir -p {tmpdir}"
+                    )
+                    self.exec_cmd(mkdir_cmd)
+
+                    copy_cmd = (
+                        f"{kube_cmd} cp -c {mysql_cont} "
+                        f"{pwd_path} {pod}:{pwd_path}"
+                    )
+                    self.exec_cmd(copy_cmd)
+
+                    queries_cmd = (
+                        f"{kube_cmd} exec -c {mysql_cont} {pod} -- "
+                        f"sh -lc 'MYSQL_PWD=$(cat {pwd_path}) "
+                        f"{sql_cmd} \"{queries}\"' "
+                        f"&& rm -rf {tmpdir}"
+                    )
+                    self.add_cmd_output(
+                        queries_cmd,
+                        suggest_filename=f"{pod}_dbs.txt",
+                    )
+
+    def _join_conf_path(self, base, *parts):
+        stripped_parts = [p.lstrip(os.path.sep) for p in parts]
+        return os.path.join(base, *stripped_parts)
+
+    def _process_k8s(self):
+        logs_since = self.get_option("logs_since")
+        namespaces = self._determine_namespaces()
+        for namespace in namespaces:
+            self._collect_per_namespace(namespace, logs_since)
+
+    def _process_snap(self):
         # Ignore private keys
         self.add_forbidden_path([
-            f"{conf_paths['MYSQL_CONF']}/*.pem",
-            f"{conf_paths['MYSQL_CONF']}/*.key",
+            self._join_conf_path(
+                self.snap_path_current,
+                self.conf_paths["MYSQL_CONF"],
+                "*.pem"
+            ),
+            self._join_conf_path(
+                self.snap_path_current,
+                self.conf_paths["MYSQL_CONF"],
+                "*.key"
+            ),
         ])
 
         # Include the files we want to get
         self.add_copy_spec([
-            conf_paths["MYSQL_CONF"],
-            conf_paths["MYSQL_LOGS"],
-            conf_paths["MYSQL_ROUTER_CONF"],
-            conf_paths["MYSQL_ROUTER_LOGS"],
-            conf_paths["MYSQL_SHELL_LOGS"],
+            self._join_conf_path(
+                self.snap_path_current,
+                self.conf_paths["MYSQL_CONF"]
+            ),
+            self._join_conf_path(
+                self.snap_path_common,
+                self.conf_paths["MYSQL_LOGS"]
+            ),
+            self._join_conf_path(
+                self.snap_path_current,
+                self.conf_paths["MYSQL_ROUTER_CONF"]
+            ),
+            self._join_conf_path(
+                self.snap_path_common,
+                self.conf_paths["MYSQL_ROUTER_LOGS"]
+            ),
+            self._join_conf_path(
+                self.snap_path_common,
+                self.conf_paths["MYSQL_SHELL_LOGS"]
+            ),
         ])
 
-        # Only applicable if not in a Kubernetes cluster
-        # as `snap info` and `journalctl` are not available
-        # in the Kubernetes container
-        if not k8s_deploy:
-            # Get snap logs
-            self.add_journal("snap.charmed-mysql.*")
+        # Get snap logs
+        self.add_journal("snap.charmed-mysql.*")
 
-            # Get snap info
-            self.add_cmd_output("snap info charmed-mysql")
+        # Get snap info
+        self.add_cmd_output("snap info charmed-mysql")
 
         # If dumpdbs is set, then get all databases
         if self.get_option("dumpdbs"):
-
-            db_user = self.get_option("dbuser")
-            db_pass = self.get_option("dbpass")
-
-            # Check password is not already an environment variable
-            # and user did not supply a password
-            if "MYSQL_PWD" in os.environ and not db_pass:
-                self.soslog.info(
-                    "MYSQL_PWD present: Using MYSQL_PWD environment variable, "
-                    "user did not provide password."
-                )
-                db_pass = os.environ["MYSQL_PWD"]
-            elif not db_pass:  # Environment variable not set and no password
-                self.soslog.warning(
-                    "dumpdbs_error: option is set, but username and password "
-                    "are not provided"
-                )
+            db_user, db_pass = self._get_db_credentials()
+            if not db_user or not db_pass:
                 return
 
             mysql_env = {"MYSQL_PWD": db_pass}
             opts = f"-h 127.0.0.1 -u{db_user}"
             sql_cmd = f"mysql {opts} -e"
-            queries = [
-                # Get databases user has access to
-                "'show databases;'",
-
-                # Get unit operations from MySQL to see, for example,
-                # if a unit is stuck on joining the cluster
-                "'select * from mysql.juju_units_operations;'",
-
-                # Get the cluster group replication status
-                ("'select * from performance_schema.replication_group_members "
-                    "order by MEMBER_HOST;'"),
-
-                # Get connection stats
-                "'show global status like \"%conne%\";'",
-
-                # Get errors per client and host
-                # Useful for problens like an app disconnectting randomly
-                "'select * from performance_schema.host_cache;'"
-            ]
 
             self.add_cmd_output(
-                [f"{sql_cmd} {query}" for query in queries],
-                env=mysql_env
+                [f"{sql_cmd} '{query}'" for query in self.mysql_queries],
+                env=mysql_env,
             )
+
+    def check_enabled(self):
+        # Check for snap package
+        if self.is_installed(self.snap_package):
+            return True
+
+        # Check for kubectl and pods
+        return (
+            is_executable(self.kube_cmd, self.sysroot) and
+            bool(self._determine_namespaces())
+        )
+
+    def setup(self):
+        if self.is_installed(self.snap_package):
+            self._process_snap()
+
+        if is_executable(self.kube_cmd, self.sysroot):
+            self._process_k8s()
 
 # vim: set et ts=4 sw=4 :


sos/report/plugins/charmed_mysql_k8s.py
```diff
@@ -0,0 +1,185 @@
...
+        dump_files_path = f"/tmp/{self.plugin_name}"
+        for name, path in self.mysql_conf_paths.items():
+            for pod in pods:
+                mkdir_cmd = f"mkdir -p {dump_files_path}/{pod}/{name}"
+                self.exec_cmd(mkdir_cmd)
+                copy_cmd = (
+                    f"{kube_cmd} cp -c {mysql_cont} {pod}:{path} "
+                    f"{dump_files_path}/{pod}/{name}"
+                )
+                self.exec_cmd(copy_cmd)
```
pmoravec: Can't you dump the data directly to the final location? Like https://github.com/sosreport/sos/blob/main/sos/report/plugins/subscription_manager.py#L84-L86 ?

sos/report/plugins/charmed_mysql.py
```diff
@@ -29,6 +29,40 @@ class CharmedMySQL(Plugin, UbuntuPlugin):
     plugin_name = "charmed_mysql"
     packages = ("charmed-mysql",)
 
+    mysql_queries = [
+        # Get databases user has access to
+        "'show databases;'",
```
pmoravec: I understand you just move this code around, so this is more nitpickyou can ignore.

Are the extra `'..'` quotation marks really needed? And if so, isnt it better to put them to the

    [f"{sql_cmd} {query}" for query in self.mysql_queries],

command list? I mean, maintaining this `mysql_queries` list can be error-prone with the extra quote marks..?

(but again, I leave on you to decide if it is worth adjusting this or not)

sos/report/plugins/charmed_mysql_k8s.py
```diff
@@ -0,0 +1,175 @@
...
+        pods = (
+            f"{kube_cmd} get pods -l {self.selector} "
+            "-o jsonpath='{{.items[*].metadata.name}}'"
+        )
+
+        # Get the config and logs from each pod
+        dump_files_path = f"/tmp/{self.plugin_name}"
+        for name, path in self.mysql_conf_paths.items():
+            command = (
+                f"for pod in $({pods}); do "
```
bmr-cymru: This doesn't look like it will work: you're using shell commands here - these are then executed via `self.add_cmd_output()` or `self.exec_cmd()`; both of these ultimately end up running the command via `utilities.sos_get_command_output()`, which explicitly disables `shell=` when calling `Popen`:

```python
237 def sos_get_command_output(command, timeout=TIMEOUT_DEFAULT, stderr=False,
238                            chroot=None, chdir=None, env=None, foreground=False,
239                            binary=False, sizelimit=None, poller=None,
240                            to_file=False, tac=False, runas=None):
241     # pylint: disable=too-many-locals,too-many-branches
242     """Execute a command and return a dictionary of status and output,
243     optionally changing root or current working directory before
244     executing command.
245     """
...
316     try:
317         with Popen(expanded_args, shell=False, stdout=_output, <<<<<
318                    stderr=STDOUT if stderr else PIPE,
319                    bufsize=-1, env=cmd_env, close_fds=True,
320                    preexec_fn=_child_prep_fn) as p:
321 
```

You cannot use shell notation when constructing commands to pass to these methods - we disabled `shell=` many years ago:

```
commit 46b6c3d39f923d19fa7fcfec96c1cf2d23c768be
Date:   Sun Apr 6 18:01:33 2014 +0100

    Call Popen with shell=False
    
    Fixes Issue #253.
    
```

There are a very small number of plugins that execute commands like `sh -c ...` or `bash -c ...` but please don't use this pattern - write the iterators in Python rather than shell. This is generally more readable and easier to debug and maintain.

sos/report/plugins/charmed_mysql_k8s.py
```diff
@@ -0,0 +1,175 @@
...
+                f"echo '{db_pass}' >> {pass_file} && "
+                f"for pod in $({pods}); do "
+                f"    {kube_cmd} cp -c {mysql_cont} "
+                f"{pass_file} $pod:{pass_file} || continue; "
+                f"    for query in {' '.join(self.mysql_queries)}; do "
+                f"        {kube_cmd} exec $pod -c {mysql_cont} -- sh -lc "
+                f"MYSQL_PWD=$(cat {pass_file}) '{sql_cmd} $query'; "
+                f"    done; "
+                f"done"
+            )
```
bmr-cymru: Same here - this shell construct needs to be refactored as Python. Also be aware that executing `echo {db_pass} >> {pass_file}` will cause the value of `db_pass` to be visible (even very briefly) in process listings, hence it can leak the credential to other processes on the same system. You also cannot use shell environment/variable notation (`MYSQL_PWD=`).

This will also likely lead to very unfriendly file names when collected with the `add_cmd_output()` defaults, since the default behaviour when `suggest_filename` is not given is to base the output file on the `cmd` passed to `add_cmd_output()`. This will mangle the exec args according to `sos.report.plugins._mangle_command()` and will generate a very odd and long string from that `command` string.

Deezzir: I think you're looking at the outdated commit, I reworked the logic [here](https://github.com/sosreport/sos/pull/4134/commits/f4f97accee2a6697aca44cd2790f3b14a96f0765)

sos/report/plugins/charmed_mysql_k8s.py
bmr-cymru: y, this would be better using `dump_files_path = self.get_cmd_output_path()` and then calling `os.makedirs()` to create the `{pod}/{name}` paths under this directory.

See e.g. [`pacemaker.py`](https://github.com/sosreport/sos/blob/1884dbd676bff8cab14797c575b605df08956d0b/sos/report/plugins/pacemaker.py#L117) which uses this to determine the path for `crm_report` output, or `git grep get_cmd_output_path sos/report/plugins` for other examples.

Deezzir: Addressed

Deezzir:
@bmr-cymru I've updated the PR description, I think it is better to keep the changes for both plugins in this PR, as I remove some dead K8s code from the `charmed_mysql` plugin

sos/report/plugins/charmed_mysql_k8s.py
Deezzir: I've tested the plugin with the changes and everything seems to be working as expected

bmr-cymru: That's great - but but there seem to be some outstanding issues there too. This is one reason why it's best to squash any fixup patches into the parent: it's easier and cleaner to review.

sos/report/plugins/charmed_mysql_k8s.py
```diff
@@ -0,0 +1,188 @@
...
+            opts = f"-h 127.0.0.1 -u{db_user}"
+            sql_cmd = f"mysql {opts} -e"
+            queries = " ".join(self.mysql_queries)
+
+            for pod in pods:
+                setup_cmds = [
+                    f"echo $MYSQL_PWD > {pass_file}; ",
+                    f"{kube_cmd} cp -c {mysql_cont} {pass_file} "
+                    f"{pod}:{pass_file}",
+                ]
```
bmr-cymru: This is still using shell notation in the first member of the `setup_cmds` list:

```python
    f"echo $MYSQL_PWD > {pass_file}; ",
```

Relies on the shell to interpolate `$MYSQL_PWD` which shouldn't work when passed to `self.exec_cmd()` (if it does then something is very broken). It also still leaks the value of `$MYSQL_PWD` in process listings.

You can see this with a simple test plugin:

```python
from sos.report.plugins import Plugin, IndependentPlugin


class Test(Plugin, IndependentPlugin):

    short_desc = 'test plugin'

    plugin_name = "test"

    def setup(self):
        my_env = {"MY_VAR": "quux"}
        self.exec_cmd("echo $MY_VAR > /tmp/foo", env=my_env)

# vim: set et ts=4 sw=4 :
```

After running this, if shell interpolation were supported, the file `/tmp/foo` should exist and contain the string `quux`:

```
# ls /tmp/foo
ls: cannot access '/tmp/foo': No such file or directory
# sos report -vvv --batch --build -o test

sos report (version 4.9.1)

set sysroot to '/' (default)
[sos.report:setup] executing 'sos report -vvv --batch --build -o test'
[sos.report:setup] using 'none' preset defaults ()
[sos.report:setup] effective options now: --batch --build --only-plugins test -vvv
This command will collect system configuration and diagnostic
information from this Fedora Linux system.

For more information on the Fedora Project visit:

        Community Website : https://fedoraproject.org/
        Community Forums  : https://discussion.fedoraproject.org/

The generated archive may contain data considered sensitive and its
content should be reviewed by the originating organization before being
passed to any third party.

No changes will be made to system configuration.



 Setting up archive ...
[archive:TarFileArchive] initialised empty FileCacheArchive at '/var/tmp/sos.deo04ip8/sosreport-f42-snapm-vm1-2025-09-25-feyujwp'
[archive:TarFileArchive] created directory at 'sos_commands' in FileCacheArchive '/var/tmp/sos.deo04ip8/sosreport-f42-snapm-vm1-2025-09-25-feyujwp'
[archive:TarFileArchive] created directory at 'sos_logs' in FileCacheArchive '/var/tmp/sos.deo04ip8/sosreport-f42-snapm-vm1-2025-09-25-feyujwp'
[archive:TarFileArchive] created directory at 'sos_reports' in FileCacheArchive '/var/tmp/sos.deo04ip8/sosreport-f42-snapm-vm1-2025-09-25-feyujwp'
 Setting up plugins ...

  Starting 1/1   test            [Running: test]
[plugin:test] collected plugin 'test' in 4.4345855712890625e-05

  Finished running plugins

[archive:TarFileArchive] added open file to FileCacheArchive '/var/tmp/sos.deo04ip8/sosreport-f42-snapm-vm1-2025-09-25-feyujwp'
[archive:TarFileArchive] added open file to FileCacheArchive '/var/tmp/sos.deo04ip8/sosreport-f42-snapm-vm1-2025-09-25-feyujwp'
[archive:TarFileArchive] added open file to FileCacheArchive '/var/tmp/sos.deo04ip8/sosreport-f42-snapm-vm1-2025-09-25-feyujwp'
[archive:TarFileArchive] added string at 'version.txt' to FileCacheArchive '/var/tmp/sos.deo04ip8/sosreport-f42-snapm-vm1-2025-09-25-feyujwp'
[archive:TarFileArchive] added open file to FileCacheArchive '/var/tmp/sos.deo04ip8/sosreport-f42-snapm-vm1-2025-09-25-feyujwp'
[archive:TarFileArchive] added open file to FileCacheArchive '/var/tmp/sos.deo04ip8/sosreport-f42-snapm-vm1-2025-09-25-feyujwp'
[archive:TarFileArchive] added string at 'sos_reports/manifest.json' to FileCacheArchive '/var/tmp/sos.deo04ip8/sosreport-f42-snapm-vm1-2025-09-25-feyujwp'
Your sos report build tree has been generated in:
	/var/tmp/sosreport-f42-snapm-vm1-2025-09-25-feyujwp

# cat /tmp/foo
cat: /tmp/foo: No such file or directory
```

It's even clearer if you pass the expression to `add_cmd_output()`:

```
[plugin:test] unpacked command: binary=False, changes=False, chroot=True, cmd=echo $MY_VAR > /tmp/foo, cmd_as_tag=False, container_cmd=None, env={'MY_VAR': 'quux'}, foreground=False, priority=10, root_symlink=None, runas=None, runat=None, sizelimit=25, stderr=True, subdir=None, suggest_filename=None, tags=[], timeout=None, to_file=False
[plugin:test] collecting output of 'echo $MY_VAR > /tmp/foo'
[plugin:test] collected output of 'echo' in 0.017114639282226562 (changes=False)
```

Results in:

```
# cat /var/tmp/sosreport-f42-snapm-vm1-2025-09-25-vunimel/sos_commands/test/echo_MY_VAR_.tmp.foo 
$MY_VAR > /tmp/foo
```

It also seems like the members of this list are expected to be concatenated together as a single shell command string? (semicolon and trailing whitespace)

That's not the behaviour you'll get from:

```python
    for cmd in setup_cmds:
        self.exec_cmd(cmd, env=mysql_env)
```

It will attempt to run:

* `echo $MYSQL_PWD > {pass_file}; `
* `{kube_cmd} cp -c {mysql_cont} {pass_file} {pod}:{pass_file}`

As two separate commands: that doesn't look like it's what you want - the compound command:

`echo $MYSQL_PWD > {pass_file}; {kube_cmd} cp -c {mysql_cont} {pass_file} {pod}:{pass_file}`

Right?

The `echo $MYSQL_PWD > {pass_file}` won't work anyway so might as well just have `setup_cmd = "{kube_cmd} cp -c {mysql_cont} {pass_file} {pod}:{pass_file}"` and do away with the loop.

As a general style point I would strongly encourage avoiding mixing trailing commas vs. implicit string concatenation in the same list/tuple - at least without a giant comment - it's very confusing to read, and easily broken by future changes. 

If passing the password via `{pass_file}` is required consider writing it first then calling `{kube_cmd} cp -c {mysql_cont} {pass_file} {pod}:{pass_file}` as a single command passed to `self.exec_cmd()`.

sos/report/plugins/charmed_mysql_k8s.py
```diff
@@ -0,0 +1,188 @@
...
+                for cmd in setup_cmds:
+                    self.exec_cmd(cmd, env=mysql_env)
+
+                queries_cmd = (
+                    f"{kube_cmd} exec -c {mysql_cont} {pod} -- sh -lc "
+                    f"'MYSQL_PWD=$(cat {pass_file}) {sql_cmd} \"{queries}\"'"
```
bmr-cymru: The path `{pass_file}` won't exist here due to the reliance on the shell redirection. You'll need to write it out from Python code, something like:

```python
>>> pass_file="/somewhere/safe"
>>> with open(pass_file, "w", encoding="utf8") as fp:
...     fp.write(os.environ["MYSQL_PWD"])
```

sos/report/plugins/charmed_mysql_k8s.py
```diff
@@ -0,0 +1,188 @@
...
+            f"{dump_files_path}/*/MYSQL_CONF/*.key",
+        ])
+        for pod in pods:
+            self.add_copy_spec(f"{dump_files_path}/{pod}/*")
+
+        # If dumpdbs is set, then get all databases
+        if self.get_option("dumpdbs"):
+            db_user = self.get_option("dbuser")
+            db_pass = self.get_option("dbpass")
+            pass_file = "/tmp/mysql_pwd"
```
bmr-cymru: I appreciate this is expected to run in relatively constrained k8s environments, but it's still not great practice to store a password in a world-readable/writable directory like `/tmp`. It's vulnerable to symlink hijacking and unwanted disclosure as-is. Consider putting this inside the archive (which is already protected by containing directory permissions - but then you'd need to clean it up/remove it after use), or better just use `tempfile.mkstemp()` / `with tempfile.TemporaryDirectory() as dir` to create a secure temporary path to store this in. Using the context manager style (`with ...`) has the benefit that you don't have to worry about cleaning up explicitly.

sos/report/plugins/charmed_mysql.py
Deezzir: Addressed

Deezzir:
@bmr-cymru the commits are now squashed into two separate commits - one for `charmed_mysql` and the second for `charmed_mysql_k8s`

bmr-cymru:
> @bmr-cymru the commits are now squashed into two separate commits - one for `charmed_mysql` and the second for `charmed_mysql_k8s`

Great, thanks!

sos/report/plugins/charmed_mysql_k8s.py
bmr-cymru: (and this isn't simply some academic concern - the project was bitten by this once before, [CVE-2015-7529](https://nvd.nist.gov/vuln/detail/CVE-2015-7529) / #696)

sos/report/plugins/charmed_mysql_k8s.py
Deezzir: But the file is read from inside the container, so it should be available after the `kubectl cp`

sos/report/plugins/charmed_mysql_k8s.py
Deezzir: Please review the latest patch, I tried to address your concerns

bmr-cymru: This looks good! It removes the potential for `/tmp` file races, eliminates the use of shell/environment variables in the command list and writes the password out before calling any of the kube commands. Thanks for working to add these improvements to the plugin.

sos/report/plugins/charmed_mysql_k8s.py
bmr-cymru: It's fine now that you aren't relying on `MYSQL_PWD` being expanded in the command list.

sos/report/plugins/charmed_mysql_k8s.py
```diff
@@ -0,0 +1,200 @@
...
+    The Charmed MySQL K8s plugin is used to collect MySQL configuration and
+    logs from the Charmed MySQL on Kubernets. It also collects MySQL Router
+    and MySQL Shell configuration and logs where available,
+    journal logs for the snap, and snap info.
+
+    If the `dumpdbs` option is set to `True`, the plugin will also try and
+    collect the names of the databases that the user has access to. The
+    `mysql` user is used by default, but that can be set using the `dbuser`
+    option. When using the `dumpdbs` option, you must then provide the
+    password for the user using the `dbpass` option.
```
bmr-cymru: Minor docstring nit: the plugin also supports obtaining the password from `MYSQL_PWD` if set, but it's not a big deal.

bmr-cymru (approved):

sos/report/plugins/charmed_mysql_k8s.py
```diff
@@ -0,0 +1,200 @@
...
+                    f"{dump_files_path}/{pod}/{name}"
+                )
+                self.exec_cmd(copy_cmd)
+
+        self.add_forbidden_path([
+            f"{dump_files_path}/*/MYSQL_CONF/*.pem",
+            f"{dump_files_path}/*/MYSQL_CONF/*.key",
+        ])
+        for pod in pods:
+            self.add_copy_spec(f"{dump_files_path}/{pod}/*")
```
pmoravec: This means you add the files copied via `{kube_cmd} cp -c` once *more*, no?

E.g. when I mimic this by:

```
outdir = self.get_cmd_output_path()
self.exec_cmd(f"cp /var/log/foo {outdir}")
```

then I get in generated sosreport `sos_commands/qpid/foo` file. But when I also call:

```
self.add_copy_spec(f"{outdir}/foo")
```

then I *additionally* get files:

```
var/tmp/sos.yv3nkdhg/sosreport-pmoravec-rhel9-2025-09-29-ylawifb/sos_commands/qpid/foo
sos_strings/qpid/var.tmp.sos.yv3nkdhg.sosreport-pmoravec-rhel9-2025-09-29-ylawifb.sos_commands.qpid.foo.tailed
```

(the `sos_strings` in my case because the `foo` file was >25MB).

Is this intentional..?

sos/report/plugins/charmed_mysql_k8s.py
```diff
@@ -0,0 +1,200 @@
...
+                for pod in pods:
+                    copy_cmd = (
+                        f"{kube_cmd} cp -c {mysql_cont} "
+                        f"{pwd_path} {pod}:{container_pwd_path}"
+                    )
+                    self.exec_cmd(copy_cmd)
+
+                    queries_cmd = (
+                        f"{kube_cmd} exec -c {mysql_cont} {pod} -- sh -lc "
+                        f"'MYSQL_PWD=$(cat {container_pwd_path}) "
```
pmoravec: Please do I get this right?

- plugin keeps the password in `db_pass` variable
- it writes the password to `pwd_path` on the host
- it copies the password file to the container
- you run command where you set env.variable per the container's file content
- password file in container is removed

Isnt there still the same security problem like @bmr-cymru pointed out, but now for the container's file? Its location is deterministic.

Deezzir: Do you think that creating a randomly named folder will resolve this as well?

sos/report/plugins/charmed_mysql_k8s.py
Deezzir: Removed it, thanks for pointing. Question, will `add_forbidden_path` work in this case?

sos/report/plugins/charmed_mysql_k8s.py
Deezzir: @pmoravec  Addressed. A temporary directory with the same name as the one within the context is created. It should be indeterministic now.

Deezzir: Also, afterward, I decided not to delete the created tempdir inside the container; I only removed the password file. I don't feel comfortable doing `rm—rf`.

sos/report/plugins/charmed_mysql_k8s.py
```diff
@@ -0,0 +1,203 @@
...
+    If the `dumpdbs` option is set to `True`, the plugin will also try and
+    collect the names of the databases that the user has access to. The
+    `mysql` user is used by default, but that can be set using the `dbuser`
+    option. When using the `dumpdbs` option, you must then provide the
+    password for the user using the `dbpass` option.
+    """
+
+    short_desc = "Charmed MySQL on Kubernetes"
+    plugin_name = "charmed_mysql_k8s"
+    packages = ("juju",)
```
arif-ali: The plugin will run on any machine that has juju installed, I think this is not ideal, as not all juju environments wlll have charmed_mysql_k8s. It would be better to have a different trigger point on when the plugin is activated. Otherwise this plugin would run on any environment where the juju client or snap is installed.

It may be better to implement a `check_enabled()` or another file that would be good to check, for then this plugin to run.

My hunch is that you probably need to check that `kubectl` command is available, and then also to check for something else. This is too generic at the moment for my liking.

pmoravec:
The code itself looks good to me up to the 2 points mentioned by Arif:

- `charmed_mysql_k8s` needs better enablement check(s)
- it really makes sense to merge the two plugins into one. Sorry I havent realized this earlier that so much code is common. Maybe the best approach is to have one file with two classes where `CharmedMySQLK8s` is a child one of `CharmedMySQL` with different `check_enabled` method and `setup`, but lots of variables (`mysql_queries`, `conf_paths`,..) and some code common?

I can live with the current plugins design, the preference among options can be subjective and we also have e.g. `aap_containerised` plugin (though the difference from `aap_*` ones is bigger). But I bit defer to Arif's opinion here as I know nothing about juju.

sos/report/plugins/charmed_mysql_k8s.py
Deezzir: I've addressed the generic trigger, now the plugin will be enabled if it detects the presence of the `mysql` pods

Deezzir:
@pmoravec @arif-ali I've tried to make the `charmed_mysql_k8s` inherit from the `charmed_mysql` plugin, but it does not seem to work. The child plugin is not being initialized/checked for triggers. I can still put them in one file and make some variables like `queries` and `conf_paths` shared. WDYT?

I've also addressed the generic trigger, now it should be better

sos/report/plugins/charmed_mysql_k8s.py
```diff
@@ -0,0 +1,206 @@
...
+    If the `dumpdbs` option is set to `True`, the plugin will also try and
+    collect the names of the databases that the user has access to. The
+    `mysql` user is used by default, but that can be set using the `dbuser`
+    option. When using the `dumpdbs` option, you must then provide the
+    password for the user using the `dbpass` option.
+    """
+
+    short_desc = "Charmed MySQL on Kubernetes"
+    plugin_name = "charmed_mysql_k8s"
+    packages = ("kubectl",)
```
arif-ali: This will detect if the kubectl package is installed, whether it's a snap or deb in this scenario. I don't think from a Canonical perspective we would have a package like this.

What you're probably looking for is `commands = ("kubectl",)`

There are many ways to look out for triggers, such as `files`, `packages`, `commands`, `kernel_mods` and `services`. We just gotta choose a combination of them, and any one being true will trigger

sos/report/plugins/charmed_mysql.py
```diff
@@ -8,10 +8,10 @@
 
 import os
 
-from sos.report.plugins import Plugin, PluginOpt, UbuntuPlugin
+from sos.report.plugins import DebianPlugin, Plugin, PluginOpt, UbuntuPlugin
 
 
-class CharmedMySQL(Plugin, UbuntuPlugin):
+class CharmedMySQL(Plugin, UbuntuPlugin, DebianPlugin):
```
arif-ali: Curious question, do we need debian here, as I am not sure charmed_mysql is being used/delivered on Debian systems?

arif-ali (changes requested):
> The child plugin is not being initialized/checked for triggers. I can still put them in one file and make some variables like queries and conf_paths shared. WDYT?

We should be able to do this in one file, like we have done in the past with other plugins. If it helps, we can work on this offline, and help you to get to the end result.

sos/report/plugins/charmed_mysql.py
Deezzir: You're correct, removed

sos/report/plugins/charmed_mysql_k8s.py
Deezzir: Thanks for pointing that, I didn't now that there is a commands variable

sos/report/plugins/charmed_mysql.py
```diff
@@ -42,103 +81,189 @@ class CharmedMySQL(Plugin, UbuntuPlugin):
...
+                "MYSQL_PWD present: Using MYSQL_PWD environment variable, "
+                "user did not provide password."
+            )
+            db_pass = os.environ["MYSQL_PWD"]
+        elif not db_pass:
+            self.soslog.warning(
+                "dumpdbs_error: option is set, but username and password "
+                "are not provided"
+            )
+            return None, None
```
pmoravec, line 106: Just a theoretical thought: could there be a deployment with username set but empty password? If so, we must return `db_user, None` here.

sos/report/plugins/charmed_mysql.py
```diff
@@ -42,103 +81,189 @@ class CharmedMySQL(Plugin, UbuntuPlugin):
...
+
+        self.add_forbidden_path([
+            f"{dump_files_path}/*/MYSQL_CONF/*.pem",
+            f"{dump_files_path}/*/MYSQL_CONF/*.key",
+        ])
+
+        # If dumpdbs is set, then get all databases
+        if self.get_option("dumpdbs"):
+            db_user, db_pass = self._get_db_credentials()
+            if not db_user or not db_pass:
```
pmoravec, line 165: Ditto if a user can have empty password. (but again, rather theoretical than practical comment, you can ignore it)

sos/report/plugins/charmed_mysql.py
```diff
@@ -42,103 +81,189 @@ class CharmedMySQL(Plugin, UbuntuPlugin):
...
+                        f"{kube_cmd} cp -c {mysql_cont} "
+                        f"{pwd_path} {pod}:{pwd_path}"
+                    )
+                    self.exec_cmd(copy_cmd)
+
+                    queries_cmd = (
+                        f"{kube_cmd} exec -c {mysql_cont} {pod} -- "
+                        f"sh -lc 'MYSQL_PWD=$(cat {pwd_path}) "
+                        f"{sql_cmd} \"{queries}\"' "
+                        f"&& rm {pwd_path}"
```
pmoravec: This will leave orphaned temp.directories inside the container (one dir for every plugin run), am I right? I am aware of https://github.com/sosreport/sos/pull/4134#discussion_r2396862633 just dont understand why.

sos/report/plugins/charmed_mysql.py
```diff
@@ -42,103 +81,189 @@ class CharmedMySQL(Plugin, UbuntuPlugin):
...
-            conf_paths["MYSQL_CONF"],
-            conf_paths["MYSQL_LOGS"],
-            conf_paths["MYSQL_ROUTER_CONF"],
-            conf_paths["MYSQL_ROUTER_LOGS"],
-            conf_paths["MYSQL_SHELL_LOGS"],
+            f"{self.snap_path_current}{self.conf_paths['MYSQL_CONF']}",
+            f"{self.snap_path_common}{self.conf_paths['MYSQL_LOGS']}",
+            f"{self.snap_path_current}{self.conf_paths['MYSQL_ROUTER_CONF']}",
+            f"{self.snap_path_common}{self.conf_paths['MYSQL_ROUTER_LOGS']}",
+            f"{self.snap_path_common}{self.conf_paths['MYSQL_SHELL_LOGS']}",
```
pmoravec: Rather a nitpick (or not, @TurboTurtle ?): while `self.conf_paths` entries start with leading `/`, isn't it safer to have os.path.join here (or at least `{self.snap_path_current}/{self.conf_paths[..]}`)? Or am I too paranoically overdoing it?

sos/report/plugins/charmed_mysql.py
```diff
@@ -42,103 +81,189 @@ class CharmedMySQL(Plugin, UbuntuPlugin):
...
+        if self.is_installed(self.snap_package):
+            return True
 
-                # Get connection stats
-                "'show global status like \"%conne%\";'",
+        # Check for kubectl and pods
+        if (
+            is_executable(self.kube_cmd, self.sysroot) and
+            bool(self._determine_namespaces())
+        ):
```
pmoravec: Can't you simply

```
        return (
            is_executable(self.kube_cmd, self.sysroot) and
            bool(self._determine_namespaces())
        )
```

?

sos/report/plugins/charmed_mysql.py
bmr-cymru: If the `self.conf_paths` entries begin with `os.path.sep` (`/`) then you need to trim that off before passing to `join()`, otherwise the joined path is treated as absolute and returned unmodified:

```python
>>> join("/myroot", "/some/other/path")
'/some/other/path'
```

sos/report/plugins/charmed_mysql.py
arif-ali: or can this be better done in the following way? with all the statements into one?

```python
        return (
            self.is_installed(self.snap_package) or
            (is_executable(self.kube_cmd, self.sysroot) and
            bool(self._determine_namespaces()))
        )
```

sos/report/plugins/charmed_mysql.py
arif-ali: @Deezzir can you also have a look at this too

Deezzir: What can we do here? As pointed by @bmr-cymru we cannot use `os.path.join` without escaping the leading `/`. I guess we can do a dedicated function for that. Wdyt? Or we leave it as is

bmr-cymru: I guess there are three ways to go here:

* Keep it as-is - maybe add a comment above the blocks 214-215 and 220-224 noting that this expects the paths to be in the form `/path/with/leading/slash`. I don't mind this too much tbh.
* Change the `conf_paths` table to use relative paths, and switch to `join($ROOT, self.conf_paths[...])` everywhere (so line 151 would just be `join("/", self.conf_path[...])`, this list would be `join(self.snap_path_*, self.conf_paths[...])`). This is more churn but consistent and obedient to preferred path construction style.
* Add a little `_join_conf_path()` helper to trim the leading `/` and glue the paths together in the list literals. This is less preferable imho.

What do others thing?

sos/report/plugins/charmed_mysql.py
Deezzir: Added `rm -rf`

sos/report/plugins/charmed_mysql.py
Deezzir: I like the third option, proactively added a `_join_conf_path` to safely join the paths, PTAL!

sos/report/plugins/charmed_mysql_k8s.py
Deezzir: Added the comment, thanks for pointing it out

PR #4146, opened 2025-10-14
[kdump] Collect dump on Debian

sourabhjains:
Collects dump files on Debian when the following conditions are met:

 - The dump file is not older than 24 hours.
 - The dump file size is not greater than 2 GB.
 - Plugin specific option kdump.get-vm-core=true is passed

Additionally, it collects the dmesg files, the kexec_cmd file, and the linux-image file present in the dump target configured in the kdump configuration file /etc/default/kdump-tools.

---

### DIFF
--- a/sos/report/plugins/kdump.py
+++ b/sos/report/plugins/kdump.py
@@ -114,6 +114,24 @@ class DebianKDump(KDump, DebianPlugin, UbuntuPlugin):
     files = ('/etc/default/kdump-tools',)
     packages = ('kdump-tools',)
 
+    option_list = [
+        PluginOpt("get-vm-core", default=False, val_type=bool,
+                  desc="collect memory dumps")
+    ]
+
+    def read_kdump_conffile(self):
+        """ Parse /etc/default/kdump-tools """
+        path = "/var/crash"
+
+        kdump = '/etc/default/kdump-tools'
+        with open(kdump, 'r', encoding='UTF-8') as file:
+            for line in file:
+                line = line.strip()
+                if line.startswith("KDUMP_COREDIR"):
+                    path = line.split('=')[1].strip('"')
+
+        return path
+
     def setup(self):
         super().setup()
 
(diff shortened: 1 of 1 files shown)


sos/report/plugins/kdump.py
```diff
@@ -114,6 +114,23 @@ class DebianKDump(KDump, DebianPlugin, UbuntuPlugin):
...
+    ]
+
+    def read_kdump_conffile(self):
+        """ Parse /etc/default/kdump-tools """
+        path = "/var/crash"
+
+        kdump = '/etc/default/kdump-tools'
+        with open(kdump, 'r', encoding='UTF-8') as file:
+            for line in file:
+                if line.startswith("KDUMP_COREDIR"):
```
vaibhav92, line 130: May want to check for spaces at the beginning of the line

sos/report/plugins/kdump.py
```diff
@@ -127,6 +144,22 @@ def setup(self):
...
 
+        try:
+            path = self.read_kdump_conffile()
+        except Exception:  # pylint: disable=broad-except
+            # set default path of coredir
+            path = "/var/crash"
+
+        self.add_dir_listing(path, recursive=True)
+        self.add_copy_spec(f"{path}/kexec_cmd")
+        self.add_copy_spec(f"{path}/kdump_lock")
```
vaibhav92: Why do we need to capture this file ? AFAIK this only indicates that a kdump is present in the directory

sos/report/plugins/kdump.py
sourabhjains: Good point about trimming spaces first. Thanks for the review.

sos/report/plugins/kdump.py
sourabhjains: @vaibhav92 `kexec_cmd` and `kdump_lock` might be needed to debug kdump issue itself.

sos/report/plugins/kdump.py
```diff
@@ -127,6 +144,22 @@ def setup(self):
...
+        self.add_copy_spec(f"{path}/*/dmesg*")
+        self.add_copy_spec(f"{path}/linux-image-*", sizelimit=2048, maxage=24)
```
vaibhav92, line 160: This looks like ubuntu specific kdump filename

sourabhjains: I thought that since the `kdump-tools` package is common to both Debian and Ubuntu, the above files would be available on Debian too. But even if these files are not available, they will be ignored.

sos/report/plugins/kdump.py
sourabhjains: Updated the patch to take care of the spaces at the start of the line.

sos/report/plugins/kdump.py
```diff
@@ -127,6 +145,22 @@ def setup(self):
...
```
pmoravec: Worth calling just one `add_copy_spec` for the four files/globs?

Please add DCO to your commit message (add there line like "Signed-off-by: Arif Ali <arif-ali@ubuntu.com>").

Otherwise ack for the PR.

sourabhjains: @pmoravec It is better to make one add_copy_spec call for kexec_cmd, kdump_lock and dmesg* file.

Ah I don't how I forgot the signoff. I will update the pr.

Thanks for the review.

pmoravec (approved):

PR #4149, opened 2025-10-24
[aap_*] fix passwords and secrets obfuscation

snagoor:
---

Passwords and SECRET_KEYS are not properly obfuscated

### DIFF
--- a/sos/report/plugins/aap_controller.py
+++ b/sos/report/plugins/aap_controller.py
@@ -50,8 +50,6 @@ class AAPControllerPlugin(Plugin, RedHatPlugin):
             "awx-manage run_dispatcher --status",
             "awx-manage run_callback_receiver --status",
             "awx-manage check_license --data",
-            "awx-manage run_wsbroadcast --status",
-            "awx-manage run_wsrelay --status",
             "supervisorctl status",
             "/var/lib/awx/venv/awx/bin/pip freeze",
             "/var/lib/awx/venv/awx/bin/pip freeze -l",
@@ -64,7 +62,7 @@ class AAPControllerPlugin(Plugin, RedHatPlugin):
         awx_version = self.collect_cmd_output('awx-manage --version')
         if awx_version['status'] == 0:
             if (
-                sos_parse_version(awx_version['output']) >
+                sos_parse_version(awx_version['output'].strip()) >
                 sos_parse_version('4.4.99')
             ):
                 self.add_cmd_output("awx-manage run_wsrelay --status")
@@ -80,7 +78,7 @@ class AAPControllerPlugin(Plugin, RedHatPlugin):
 
     def postproc(self):
         # remove database password
-        jreg = r"(\s*\'PASSWORD\'\s*:(\s))(?:\"){1,}(.+)(?:\"){1,}"
+        jreg = r"(\s*'PASSWORD'\s*:\s*)('.*')"
         repl = r"\1********"
         self.do_path_regex_sub("/etc/tower/conf.d/postgres.py", jreg, repl)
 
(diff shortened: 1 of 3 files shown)


sos/report/plugins/aap_controller.py
```diff
@@ -80,7 +78,7 @@ def setup(self):
 
     def postproc(self):
         # remove database password
-        jreg = r"(\s*\'PASSWORD\'\s*:(\s))(?:\"){1,}(.+)(?:\"){1,}"
+        jreg = r"(\s*'PASSWORD'\s*:\s*)('.*')"
```
pmoravec, line 81: The password value is always in single quote marks? Could you show some example (ideally that failed to catch now and will be obfuscated by your RE), please?

sos/report/plugins/aap_controller.py
```diff
@@ -50,8 +50,6 @@ def setup(self):
             "awx-manage run_dispatcher --status",
             "awx-manage run_callback_receiver --status",
             "awx-manage check_license --data",
-            "awx-manage run_wsbroadcast --status",
-            "awx-manage run_wsrelay --status",
```
pmoravec, line 54: Why to drop collection of the two commands? (at least it does not seem to be related to passwords obfuscation..?)

EDIT: from the `self.eda_pkg_ver` changes I understand the PR also covers some additional changes popped up in new AAP EDA version - worth mentioning it in the commit, then.

sos/report/plugins/aap_gateway.py
```diff
@@ -49,8 +49,8 @@ def setup(self):
 
     def postproc(self):
         # remove database password
-        jreg = r"(DATABASE_PASSWORD)(\s*)(=|:)(\s*)(.*)"
-        repl = r"\1\2\3\4********"
+        jreg = r"(\s*'PASSWORD'\s*:\s*)('.*')"
```
pmoravec, line 52: Ditto: password value is always in single quote marks?

sos/report/plugins/aap_eda.py
```diff
@@ -21,6 +21,12 @@ class AAPEDAControllerPlugin(Plugin, RedHatPlugin):
                 'automation-eda-controller-server')
 
     def setup(self):
+
+        pkg_name = 'automation-eda-controller'
+        pkg = self.policy.package_manager.pkg_by_name(f'{pkg_name}')
+        if pkg is not None:
+            self.eda_pkg_ver = '.'.join(pkg['version'])
```
pmoravec, line 28: You need to set `self.eda_pkg_ver` to some safe low value (zero?) if `pkg is None` - otherwise when one runs the plugin on a system without AAP EDA - like the cirrus job did - it will fail with backtrace:

```
  File "/tmp/cirrus-ci-build/sos/report/__init__.py", line 1260, in setup
    plug.setup()
  File "/tmp/cirrus-ci-build/sos/report/plugins/aap_eda.py", line 68, in setup
    if sos_parse_version(self.eda_pkg_ver) > sos_parse_version('1.0.99'):
AttributeError: 'AAPEDAControllerPlugin' object has no attribute 'eda_pkg_ver'
```

sos/report/plugins/aap_controller.py
snagoor: @pmoravec thanks for the review. Yes, the password is always in single quote. Here is an example configuration file for the same https://gist.github.com/snagoor/702feb1f403a2857c77a31d48d4079f3

sos/report/plugins/aap_gateway.py
snagoor: This is an example _settings.py_ on AAP Gateway nodes https://gist.github.com/snagoor/c27124ceb00f4bf073e90c65c4e3a2d8

sos/report/plugins/aap_controller.py
snagoor: Capturing data is now a part of a conditional check https://github.com/sosreport/sos/blob/main/sos/report/plugins/aap_controller.py#L63-L72. Cleaning them up is missed in the previous PR. 

Regarding _eda_pkg_ver_ I added the message in the plugin itself. I will also update the PR message about this change.

sos/report/plugins/aap_eda.py
snagoor: Thank you. Updated as suggested.

pmoravec, line 28: I think it is easier to add here `else` branch to set `self.eda_pkg_ver = '0.0.0'` and you can reference `self.eda_pkg_ver` directly, but I am ok with the current approach.

pmoravec (approved):

TurboTurtle (approved):

PR #4150, opened 2025-10-29
[aap_containerized] Add heuristic for detecting username

pmoravec:
Forgetting to provide mandatory username of the AAP containerized deployment leads to skipping plugin completely and lack of collected data.

Let add a heuristic "is there a sole user running some podman container?" to detect the user in such a case.

Closes: #4150

---

### DIFF
--- a/sos/report/plugins/aap_containerized.py
+++ b/sos/report/plugins/aap_containerized.py
@@ -42,9 +42,23 @@ class AAPContainerized(Plugin, RedHatPlugin):
         # Check if username is passed as argument
         username = self.get_option("username")
         if not username:
-            self._log_error("Username is mandatory to collect "
-                            "AAP containerized setup logs")
-            return
+            self._log_warn("AAP username is missing, use '-k "
+                           "aap_containerized.username=<user>' to set it")
+            ps = self.exec_cmd("ps aux")
+            if ps["status"] == 0:
+                podman_users = set()
+                for line in ps["output"].splitlines():
+                    if ("/usr/bin/podman" in line) and \
+                       ("/.local/share/containers/storage/" in line):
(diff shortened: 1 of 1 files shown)


sos/report/plugins/aap_containerized.py
```diff
@@ -42,9 +42,23 @@ def setup(self):
...
+                    if ("/usr/bin/podman" in line) and \
+                       ("/.local/share/containers/storage/" in line):
+                        user, _ = line.split(maxsplit=1)
+                        podman_users.add(user)
+                if len(podman_users) == 1:
+                    username = podman_users.pop()
+                    self._log_warn(f"AAP username detected as '{username}'")
+                else:
+                    self._log_error("No AAP username detected, "
+                                    "terminating plugin.")
```
TurboTurtle: Since we can hit this when there are either 0 or >1 `podman_users`, maybe this should be something like `"Unable to determine AAP username"`?

pmoravec: Why not, PR updated.

snagoor:
Thanks @pmoravec for the update. I appreciate the effort to improve usability, but I have a few concerns:

- The heuristic for detecting the AAP username via ps aux is unreliable in edge cases—especially when no containers are running or multiple users use Podman.

- AAP containerized deployments use a non-root user chosen by the end-user, so automatic detection isn't guaranteed to be accurate.

- The original plugin design intentionally required the username to be provided to avoid ambiguity, which I still believe is the safest approach.

- Even if we retain the heuristic, it should be optional, with fallback logic that avoids incorrectly capturing the data not relevant to AAP (another root-less user pods can also exist on the system, but its a corner case). 

So, my final word would be stick to the old design and log a warning to pass the **username** as its mandatory for this plugin.

PR #4158, opened 2025-11-07
[upload] Add proxy server details for sos uploads

dwolstroRH:
Related: RHEL-2356

---

### DIFF
--- a/sos/report/__init__.py
+++ b/sos/report/__init__.py
@@ -150,7 +150,12 @@ class SoSReport(SoSComponent):
         'upload_target': None,
         'add_preset': '',
         'del_preset': '',
-        'treat_certificates': 'obfuscate'
+        'treat_certificates': 'obfuscate',
+        'proxy_user': None,
+        'proxy_password': None,
+        'proxy_protocol': 'http',
+        'proxy_host': None,
+        'proxy_port': None
     }
 
     def __init__(self, parser, args, cmdline):
@@ -354,6 +359,29 @@ class SoSReport(SoSComponent):
                                 help="Add a new named command line preset")
         preset_grp.add_argument("--del-preset", type=str, action="store",
                                 help="Delete the named command line preset")
+
+        # Group the proxy options together
+        proxy_grp = parser.add_argument_group(
+            'HTTP/HTTPS Proxy Options',
+            'These options control the configuration of a '
+            'proxy server if used for uploads.'
+        )
+        proxy_grp.add_argument('--proxy-user',
+                               default=None,
+                               help='Specify the User for the proxy server')
+        proxy_grp.add_argument('--proxy-pass',
+                               default=None,
+                               help='Specify the Password for the proxy user')
+        proxy_grp.add_argument('--proxy-protocol', default='http',
+                               choices=['http', 'https'],
+                               help=('Specify the protocol for the proxy '
+                                     'server'))
+        proxy_grp.add_argument('--proxy-host',
+                               default=None,
+                               help='Specify the host for the proxy server')
+        proxy_grp.add_argument('--proxy-port',
+                               default=None,
+                               help='Specify the Port for the proxy server')
 
         # Group the cleaner options together
         cleaner_grp = parser.add_argument_group(
--- a/sos/upload/targets/__init__.py
+++ b/sos/upload/targets/__init__.py
@@ -11,6 +11,8 @@
 import os
 import re
 import logging
+import subprocess
+import sys
 
 from getpass import getpass
 from sos import _sos as _
@@ -27,6 +29,92 @@ try:
     BOTO3_LOADED = True
 except ImportError:
     BOTO3_LOADED = False
+
+
+def check_netcat_syntax(proxy_protocol, proxy_host, proxy_port,
+                        proxy_user, proxy_password, nc_executable='nc'):
+    """
+    Determines the correct authenticated HTTP proxy syntax for the local
+    Netcat/Ncat executable by running test commands. There are (at least)
+    two different versions of the proxy command, so we want to return
+    the correct syntax for the local netcat/ncat executable so the
+    proxy command successfully executes.
+
+    Args:
+        proxy_protocol (str): The protocol (http/https) of the proxy server
+        proxy_host (str): The host of the proxy server
+        proxy_port (str): The port to use for the proxy server
+        proxy_user (str): The user to use to access the proxy server
+        proxy_password (str): The password for authenticating the proxy user
+        nc_executable (str): The name of the netcat command (default is 'nc').
+
+    Returns:
+        tuple: (syntax_string, is_successful_test)
+               - syntax_string: The correct ProxyCommand format
+                                (e.g., 'nc -X connect -x ...')
+               - is_successful_test: Boolean indicating if a successful
+               syntax was found.
+    """
+
+    # Define a generic, non-resolvable target (like 127.0.0.1:1) and dummy
+    # credentials. The goal is to see if the command *parses* the syntax
+    # correctly before it fails to connect.
+    # A successful syntax check will result in an "unreachable" error,
+    # while a syntax failure yields an "invalid option" error (exit code 1).
+    dummy_proxy = "127.0.0.1:1"
+    dummy_auth = "user:pass"
+
+    # --- 1. Ncat/Nmap Syntax Check ---
+    # Command: nc --proxy-type http --proxy 127.0.0.1:1
+    #             --proxy-auth user:pass 127.0.0.1 80
+    ncat_syntax = (f"{nc_executable} --proxy-type http --proxy {dummy_proxy} "
+                   f"--proxy-auth {dummy_auth} %h %p")
+    try:
+        # Check if the command accepts the Ncat-style syntax flags
+        result = subprocess.run(ncat_syntax.split() + ['127.0.0.1', '80'],
+                                capture_output=True, timeout=5, check=False)
+        # Ncat/Nmap's version will usually exit with a code 0 or 1, but its
+        # *output* won't contain "invalid option." A successful parse will
+        # proceed to connect/fail.
+        if b"invalid option" not in result.stderr:
+            # If no syntax error is returned, this is the correct format.
+            return (
+                f"{nc_executable} "
+                f"--proxy-type {proxy_protocol} "
+                f"--proxy {proxy_host}:{proxy_port} "
+                f"--proxy-auth {proxy_user}:{proxy_password} %h %p",
+                True)
+
+    except (FileNotFoundError, subprocess.CalledProcessError,
+            subprocess.TimeoutExpired):
+        pass  # Continue to next check
+
+    # --- 2. GNU/OpenBSD Syntax Check (using -X connect) ---
+    # Command: nc -X connect -x 127.0.0.1:1 -P user:pass 127.0.0.1 80
+    gnu_syntax = (f"{nc_executable} -X connect -x {dummy_proxy} "
+                  f"-P {dummy_auth} %h %p")
+    try:
+        # Check if the command accepts the GNU/OpenBSD-style syntax flags
+        result = subprocess.run(gnu_syntax.split() + ['127.0.0.1', '80'],
+                                capture_output=True, timeout=5, check=False)
+
+        if (b"invalid option" not in result.stderr and
+                b"unknown option" not in result.stderr):
+            # If no syntax error is returned, this is the correct format.
+            return (f"{nc_executable} -X connect -x {proxy_host}:{proxy_port} "
+                    f"-P {proxy_user}:{proxy_password} %h %p",
+                    True)
+
+    except (FileNotFoundError, subprocess.CalledProcessError,
+            subprocess.TimeoutExpired):
+        pass  # Continue to next check
+
+    # --- 3. Fallback (If no standard syntax is found) ---
+    # If the system netcat doesn't support an authenticated proxy,
+    # you might need to use a dedicated tool like 'corkscrew' or 'connect'.
+    return ("Warning: Native authenticated proxy syntax not found for "
+            f"'{nc_executable}'. Consider installing ncat or corkscrew.",
+            False)
 
 
 class UploadTarget():
@@ -72,6 +160,13 @@ class UploadTarget():
     upload_s3_region = None
     upload_s3_object_prefix = None
     upload_target = None
+    proxy_host = None
+    proxy_port = None
+    proxy_user = None
+    proxy_password = None
+    proxy_protocol = 'http'
+    proxy_list = {}
+    use_proxy_server = False
 
     arg_defaults = {
         'upload_file': '',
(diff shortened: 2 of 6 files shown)


sos/report/__init__.py
```diff
@@ -150,7 +150,14 @@ class SoSReport(SoSComponent):
         'upload_target': None,
         'add_preset': '',
         'del_preset': '',
+<<<<<<< Updated upstream
```
jcastill: I believe this is one of the issues causing the pylint error. You just need to resolve the conflict with the stashed changes and force push again

sos/upload/targets/__init__.py
```diff
@@ -29,6 +30,81 @@
     BOTO3_LOADED = False
 
 
+def check_netcat_syntax(nc_executable='nc'):
+    """
+    Determines the correct authenticated HTTP proxy syntax for the local
+    Netcat/Ncat executable by running test commands.
+
```
jcastill, line 42: Could you add here (or perhaps as a comment in the gh PR interface) why do we need to determine the syntax? 
And what happens if we don't have this function at all?

jcastill (changes requested):

jcastill:
Make sure you add these options to any relevant man page (at least the upload and report ones).

sos/upload/targets/__init__.py
```diff
@@ -684,10 +799,20 @@ def upload_ftp(self, url=None, directory=None, user=None, password=None):
...
+                session.login(proxy_formatted_user, self.proxy_password,
+                              password)
+            else:
+                session = ftplib.FTP(url, user, password, timeout=15)
             if not session:
-                raise Exception("connection failed, did you set a user and "
-                                "password?")
+                raise Exception("connection failed, did you set a user"
+                                " and password?")
```
jcastill: Any reason for this specific change? Unless I'm mistaken, the original line was within col limit

sos/upload/targets/__init__.py
```diff
@@ -497,7 +595,22 @@ def upload_sftp(self, user=None, password=None, user_dir=None):
...
-        sftp_cmd = f"sftp -oStrictHostKeyChecking=no {user}@{sftp_url}"
+        if self.use_proxy_server:
+            proxy_template, success = check_netcat_syntax()
+            if success:
+                # Final ProxyCommand to use in your SFTP execution
+                proxy_cmd = (proxy_template.
+                             replace("PROXY_IP", self.get_proxy_host()).
+                             replace("PROXY_PORT", self.get_proxy_port()).
+                             replace("PROXY_USER", self.get_proxy_user()).
+                             replace("PROXY_PASS", self.get_proxy_password()))
```
jcastill: David, we are not leaking the proxy password here or in any of the other changes to, say, sos logs or similar, right?

sos/upload/targets/__init__.py
dwolstroRH: I will add in a more verbose comment to try to explain why we use this function and why (I think) it's needed. Basically, if we don't have this function, only certain versions of ncat/nc will work correctly with the proxy command. This function tries to figure out the correct syntax of the proxy command needed based on the installed version of ncat/nc

sos/upload/targets/__init__.py
dwolstroRH: I honestly have no idea why I updated that. I can change it back if you want me to.

sos/upload/targets/__init__.py
dwolstroRH: I did a recursive grep of my report directory for both my proxy user and my proxy password and did not find them.

sos/upload/targets/__init__.py
jcastill: No need, I just wasn't sure if I missed something. Doesn't look bad to me, so unless anyone else objects, this change can be kept

sos/upload/targets/__init__.py
jcastill: nice work, thank you!

sos/upload/targets/__init__.py
```diff
@@ -29,6 +30,81 @@
...
+                    True)
+
+    except (FileNotFoundError, subprocess.CalledProcessError,
+            subprocess.TimeoutExpired):
+        pass  # Continue to next check
+
+    # --- 3. Fallback (If no standard syntax is found) ---
+    # If the system netcat doesn't support an authenticated proxy,
+    # you might need to use a dedicated tool like 'corkscrew' or 'connect'.
+    return ("Warning: Native authenticated proxy syntax not found for 'nc'. "
```
pmoravec: Minor nit: replace `nc` by `{nc_executable}` (and have it as an f-string)?

Currently, you call the method with default `nc_executable='nc'` only. If you plan this argumet to have different values in future, let be prepared here.

sos/report/__init__.py
```diff
@@ -355,6 +359,25 @@ def add_parser_options(cls, parser):
...
+        proxy_grp.add_argument('--proxy-pass',
+                               default=None,
+                               help='Specify the Password for the proxy user')
+        proxy_grp.add_argument('--proxy-host',
+                               default=None,
+                               help='Specify the host for the proxy server')
+        proxy_grp.add_argument('--proxy-port',
+                               default=None,
+                               help='Specify the Port for the proxy server')
+
```
pmoravec, line 385: We need to enhance man pages accordingly.

When you will describe `--proxy-pass`, it is worth mentioning that using the option on cmdline reveals the password to anybody running `ps` at the same time.

sos/upload/targets/__init__.py
```diff
@@ -176,7 +269,12 @@ def pre_work(self, hook_commons):
                 self.prompt_for_upload_s3_endpoint()
                 self.prompt_for_upload_s3_access_key()
                 self.prompt_for_upload_s3_secret_key()
+            if self.get_proxy_host():
+                self.generate_proxy_list(use_prompts=True)
             self.ui_log.info('')
+        else:
+            if self.get_proxy_host():
+                self.generate_proxy_list(use_prompts=False)
```
pmoravec: Code de-duplication: isnt it better to call (after the `if not cmdline_opts.batch ..` block):

```
if self.get_proxy_host():
    self.generate_proxy_list(use_prompts=(not cmdline_opts.batch and not cmdline_opts.quiet))
```

?

Well, then the `log_info` of empty line would be entered before the prompt, not welcomed, I guess?

sos/upload/targets/__init__.py
```diff
@@ -777,4 +902,93 @@ def upload_s3(self, endpoint=None, region=None, bucket=None, prefix=None,
...
+                self.prompt_for_proxy_user()
+                self.prompt_for_proxy_password()
+                if self.get_proxy_user() and \
+                   self.get_proxy_password():
+                    proxy = (f"{self.get_proxy_user()}:"
+                             f"{self.get_proxy_password()}@")
+                else:
+                    self.ui_log.error(
+                        "Both proxy user and proxy password must be"
+                        " specified. Proxy server will be ignored."
```
pmoravec: What if I enter empty user and password? That should be allowed, no..?

sos/upload/targets/__init__.py
```diff
@@ -777,4 +902,93 @@ def upload_s3(self, endpoint=None, region=None, bucket=None, prefix=None,
...
+                                 f"{self.get_proxy_password()}@")
+                    else:
+                        self.ui_log.error(
+                            "Both proxy user and proxy password must be"
+                            " specified. Proxy server will be ignored."
+                        )
+                        return {}
+
+            if self.get_proxy_host() and self.get_proxy_port():
+                proxy = (f"{proxy}{self.get_proxy_host()}:"
```
pmoravec: So `proxy` stores just credentials so far, which is bit confusing. Worth adding a comment or having `proxy_creds` variable (up to this point)?

(also a nitpick, you can ignore this comment if you think the code is self-explanatory)

sos/upload/targets/__init__.py
```diff
@@ -777,4 +902,93 @@ def upload_s3(self, endpoint=None, region=None, bucket=None, prefix=None,
...
+                         f"{self.get_proxy_port()}")
+                proxy_list['http'] = f"http://{proxy}"
+                proxy_list['https'] = f"http://{proxy}"
```
pmoravec: Shouldnt be there `https://` instead of `http://`? (but I am not an expert on the syntax here..)

dwolstroRH: Great catch! thanks

sos/upload/targets/__init__.py
```diff
@@ -684,10 +799,20 @@ def upload_ftp(self, url=None, directory=None, user=None, password=None):
             directory = self.upload_directory or self._upload_directory
 
         try:
-            session = ftplib.FTP(url, user, password, timeout=15)
+            if self.use_proxy_server:
+                proxy_formatted_user = f"{user}@{url} {self.proxy_user}"
```
arif-ali: What if a user and password is not provided, and is actually not needed for the proxy. I have seen in at least one environment where a proxy user/password is not required to upload a sos.

sos/upload/targets/__init__.py
```diff
@@ -684,10 +799,20 @@ def upload_ftp(self, url=None, directory=None, user=None, password=None):
...
+
+                # Connect to the FTP proxy
+                session = ftplib.FTP()
+                session.connect(self.get_proxy_host(), self.get_proxy_port())
+
+                session.login(proxy_formatted_user, self.proxy_password,
+                              password)
```
arif-ali: same comment applies, what if it is anonymous FTP, but still need a proxy without username/password?

arif-ali (commented):
primarily around if we don't have a user/password, but just a proxy, we should be able to use the proxy on its own

sos/report/__init__.py
dwolstroRH: Thanks, fixed.

sos/upload/targets/__init__.py
```diff
@@ -497,7 +599,25 @@ def upload_sftp(self, user=None, password=None, user_dir=None):
...
+            if success:
+                # Final ProxyCommand to use in your SFTP execution
+                proxy_cmd = (proxy_template.
+                             replace("PROXY_PROTOCOL",
+                                     self.get_proxy_protocol()).
+                             replace("PROXY_IP", self.get_proxy_host()).
+                             replace("PROXY_PORT", self.get_proxy_port()).
+                             replace("PROXY_USER", self.get_proxy_user()).
+                             replace("PROXY_PASSWORD",
+                                     self.get_proxy_password()))
```
pmoravec: repeatedly replacing strings might be dangerous - what if a user is named `PROXY_IP` or a host named `PROXY_USER`? Isn't it safer to use `str.format()` method, like:

- `check_netcat_syntax` returns the `PROXY_*` strings in curly brackets
- `proxy_cmd = proxy_template.format(PROXY_PROTOCOL=self.get_proxy_protocol(), PROXY_IP=self.get_proxy_host(),....)

Also, I would rename `PROXY_IP` to `PROXY_HOST` to adhere to internal methods and variables.

(or am I too pedantic / paranoic wrt the `replace`/`format`, @TurboTurtle ?)

sos/upload/targets/__init__.py
pmoravec: +1, we should support that option.

sos/upload/targets/__init__.py
```diff
@@ -777,4 +909,103 @@ def upload_s3(self, endpoint=None, region=None, bucket=None, prefix=None,
         except Exception as e:
             raise Exception(f"Failed to upload to S3: {str(e)}") from e
 
+    def get_proxy_protocol(self):
+        return self.commons['cmdlineopts'].proxy_protocol \
+            if self.commons['cmdlineopts'].proxy_protocol else 'http'
```
pmoravec: Doesn't simple `getattr(self.commons['cmdlineopts'], 'proxy_protocol', 'http')` work well, and be more concise?

Ditto to other `get_proxy_*` methods.

sos/upload/targets/__init__.py
```diff
@@ -777,4 +909,103 @@ def upload_s3(self, endpoint=None, region=None, bucket=None, prefix=None,
...
+        return self.proxy_password
+
+    def prompt_for_proxy_user(self):
+        """Should be overridden by targets to determine if a proxy user
+        needs to be provided or not
+        """
+        if not self.get_proxy_user():
+            msg = (f"Please provide a proxy user (if required) for proxy "
+                   f"{self.get_proxy_host()}: ")
+            self.proxy_user = input(_(msg))
```
pmoravec, line 934: If I want to provide "no user" (see Arif's point for no user/pass), we shall convert empty response to `None`, maybe? (I havent tested it though).

sos/upload/targets/__init__.py
```diff
@@ -777,4 +909,103 @@ def upload_s3(self, endpoint=None, region=None, bucket=None, prefix=None,
...
+                self.prompt_for_proxy_user()
+                if self.get_proxy_user():
+                    self.prompt_for_proxy_password()
+                if self.get_proxy_user() and \
+                   self.get_proxy_password():
+                    proxy_creds = (f"{self.get_proxy_user()}:"
+                                   f"{self.get_proxy_password()}@")
+                else:
+                    if self.get_proxy_user() or \
+                       self.get_proxy_password():
```
pmoravec: Is it a valid use case a proxy allows user with no password? I dont know this.

If it is a valid use case, we prohibit it here.

sos/upload/targets/__init__.py
```diff
@@ -777,4 +909,103 @@ def upload_s3(self, endpoint=None, region=None, bucket=None, prefix=None,
...
+                        )
+                    return {}
+            else:
+                if self.get_proxy_user() and \
+                   self.get_proxy_password():
+                    proxy_creds = (f"{self.get_proxy_user()}:"
+                                   f"{self.get_proxy_password()}@")
+                else:
+                    if self.get_proxy_user() or \
+                       self.get_proxy_password():
```
pmoravec, line 985: Ditto the same use case (if it is valid).

sos/upload/targets/__init__.py
dwolstroRH: I have updated the code to allow non-authenticated proxy calls (no user/password)

sos/upload/targets/__init__.py
dwolstroRH: Modified per your recommendations. Thanks!

sos/upload/targets/__init__.py
dwolstroRH: Updated. Thanks. I'm still learning the intricacies of python coding.

sos/upload/targets/__init__.py
dwolstroRH: I am not sure. Everything that I've read has discouraged that setup profusely, but technically it can be done. Maybe I can put a warning message and prompt the user to confirm they want to execute without a password?

ie, "Warning! You have entered a proxy user without an accompanying password. It is highly recommended to avoid this usage. Do you wish to proceed? (y/N)"

pmoravec (approved):
Sounds good to me, yet I tested it in a limited manner only.

jcastill:
@mhradile @tiredPotato @dzalezak this PR is ready for RH QE - can we assign one of you to it?

mhradile:
> @mhradile @tiredPotato @dzalezak this PR is ready for RH QE - can we assign one of you to it?

Hello @jcastill , I stared but could not continue on this, would like to get back to it but currently can not. Feel free to assign it to me though.

PR #4159, opened 2025-11-10
[aap_gateway] Add feature flags collection with availability check

fao89:
# [aap_gateway] Add feature flags collection with availability check

Issue: [AAP-45877](https://issues.redhat.com/browse/AAP-45877)

## Summary

This PR enhances the AAP Gateway plugin to collect feature flags information when available, while maintaining backward compatibility with older AAP Gateway versions.

## Changes Made

- Added conditional collection of feature flags using `aap-gateway-manage feature_flags --list`
- Implemented availability check using `aap-gateway-manage help feature_flags` to ensure compatibility
- Added informational logging when feature flags command is not available
- Maintains backward compatibility with AAP Gateway versions < 2.6

## Testing

- ✅ Tested on AAP 2.6+ with feature flags command available
- ✅ Tested logic on AAP < 2.6 where command doesn't exist
- ✅ Verified exit codes and error handling behavior
- ✅ Confirmed feature flags data collection works correctly

## Contributor Guidelines Checklist

## Backward Compatibility

This change maintains full backward compatibility:
- **AAP Gateway >= 2.6**: Collects feature flags data
- **AAP Gateway < 2.6**: Gracefully skips collection with informational logging
- No breaking changes or additional dependencies

## Security Considerations

- No sensitive data is collected
- Feature flags information is configuration metadata only
- Existing password and secret obfuscation remains intact

### DIFF
--- a/sos/report/plugins/aap_gateway.py
+++ b/sos/report/plugins/aap_gateway.py
@@ -42,6 +42,7 @@ class AAPGatewayPlugin(Plugin, RedHatPlugin):
             "aap-gateway-manage authenticators",
             "aap-gateway-manage showmigrations",
             "aap-gateway-manage list_services",
+            "aap-gateway-manage feature_flags --list",
             "aap-gateway-manage --version",
         ])
         self.add_dir_listing("/etc/ansible-automation-platform/",


sos/report/plugins/aap_gateway.py
```diff
@@ -44,6 +44,16 @@ def setup(self):
...
+
+        # Check if the command exists before trying to run it
+        feature_flags_check = self.exec_cmd(
+            'aap-gateway-manage help feature_flags')
+        if feature_flags_check['status'] == 0:
+            self.add_cmd_output("aap-gateway-manage feature_flags --list")
+        else:
+            # Log that feature flags command is not available for debugging
+            self._log_info(
+                "aap-gateway-manage feature_flags command not available")
```
TurboTurtle: `add_cmd_output()` gracefully handles when a command doesn't exist.

In this case however, since it's a sub-command and the `aap-gateway-manage` command would presumably print some stderr and exit non-zero, we'd capture that stderr. Would it not be beneficial to use that stderr as an explicit signal to the reviewing person that `feature_flags` in unavailable on the host that generated the report?

fao89: this command is supposed to exist on AAP 2.6,
this is basically handling the cases where AAP < 2.6

TurboTurtle: Ok. My point is that we don't need to do presence checking of commands in `sos`. If the command doesn't exist, we log that and move on already. If it produces stderr (like a missing subcommand would), we capture that output so that a reviewing party knows that the attempt was made but the host state does not provide that information.

fao89: gotcha! I'll update it then

pmoravec:
The failed test timeouted, I am trying to re-run it.

TurboTurtle (approved):
LGTM

PR #4165, opened 2025-11-24
[subscription_manager] Add condition to run deprecated options

jcastill:
Some subscription_manager options have been deprecated in RHEL 10,
and when they are run thrown an error. This fix adds a check for OS version
and if it's not RHEL 10, runs the deprecated commands.

Related: RHEL-130523

---

### DIFF
--- a/sos/report/plugins/subscription_manager.py
+++ b/sos/report/plugins/subscription_manager.py
@@ -10,6 +10,7 @@ from configparser import NoOptionError, NoSectionError
 import glob
 from os import remove
 from sos.report.plugins import Plugin, RedHatPlugin
+from sos.utilities import sos_parse_version
 
 
 class SubscriptionManager(Plugin, RedHatPlugin):
@@ -67,13 +68,9 @@ class SubscriptionManager(Plugin, RedHatPlugin):
             "/var/log/rhsm/rhsmcertd.log"])
         self.add_cmd_output("subscription-manager identity",
                             tags="subscription_manager_id")
-        self.add_cmd_output("subscription-manager list --consumed",
-                            tags="subscription_manager_list_consumed")
         self.add_cmd_output("subscription-manager list --installed",
                             tags="subscription_manager_installed")
         self.add_cmd_output([
-            "subscription-manager list --available",
-            "subscription-manager list --all --available",
             "subscription-manager release --show",
             "subscription-manager release --list",
             "syspurpose show",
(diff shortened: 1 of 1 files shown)


sos/report/plugins/subscription_manager.py
```diff
@@ -67,20 +67,23 @@ def setup(self):
...
-            "subscription-manager list --all --available",
             "subscription-manager release --show",
             "subscription-manager release --list",
             "syspurpose show",
             "subscription-manager syspurpose --show",
             "subscription-manager status",
             "subscription-manager facts",
         ], cmd_as_tag=True)
+        # Deprecated options in RHEL 10, that throw errors when run
+        if self.policy.dist_version() < 10:
```
pmoravec: Isnt it better to test `subscription-manager` version instead? That was removed in https://github.com/candlepin/subscription-manager/commit/336f78bb400a12e1887bdda38edd036897a826b3, so a test of "version lower than" [subscription-manager-1.30.3-1](https://github.com/candlepin/subscription-manager/releases/tag/subscription-manager-1.30.3-1) would be a better match.

jcastill: Ack. I used distribution because I thought it was cleaner - the deprecated options don't exist at all in RHEL 10. 
I'll change the test to use package version instead

sos/report/plugins/subscription_manager.py
```diff
@@ -67,20 +68,25 @@ def setup(self):
...
             "subscription-manager release --show",
             "subscription-manager release --list",
             "syspurpose show",
             "subscription-manager syspurpose --show",
             "subscription-manager status",
             "subscription-manager facts",
         ], cmd_as_tag=True)
+        # Deprecated options in RHEL 10, that throw errors when run
+        smpkg = self.policy.package_manager.pkg_by_name('subscription-manager')
+        sm_version = '.'.join(smpkg['version'])
```
jcastill: I'm not sure if I should add here a quick check for smpkg being None. The plugin can be activated without 'subscription-manager' being present, so it could happen that there's no package anymore in the system

sos/report/plugins/subscription_manager.py
```diff
@@ -67,20 +68,25 @@ def setup(self):
...
+        if sos_parse_version(sm_version) < sos_parse_version('1.30.3-1'):
```
jcastill: Another option could be to check here first if smpkg exists, before going into comparing the versions.
@pmoravec thoughts on this?

jcastill:
I'm investigating the centos-stream-9 error

sos/report/plugins/subscription_manager.py
pmoravec: +1 to the check. Since when I run the plugin on Fedora with no sub-man, I get even 2 backtraces:

```
  File "/root/sos-main/sos/report/__init__.py", line 1270, in setup
    plug.setup()
    ~~~~~~~~~~^^
  File "/root/sos-main/sos/report/plugins/subscription_manager.py", line 83, in setup
    sm_version = '.'.join(smpkg['version'])
                          ~~~~~^^^^^^^^^^^
TypeError: 'NoneType' object is not subscriptable

  File "/root/sos-main/sos/report/__init__.py", line 1508, in postproc
    plug.postproc()
    ~~~~~~~~~~~~~^^
  File "/root/sos-main/sos/report/plugins/subscription_manager.py", line 135, in postproc
    if self._curl_cfg_fname:
       ^^^^^^^^^^^^^^^^^^^^
AttributeError: 'SubscriptionManager' object has no attribute '_curl_cfg_fname'
```

jcastill: Let me fix and push. My guess is that this should solve the issue with centos, but lets see

pmoravec (approved):
Sounds good! No uncaught exception on system without sub-man, commands collected on RHEL9 but not on RHEL10 (resp. on the relevant sub-man verison).

pmoravec (approved):
Works fine for me on either tested system. Thanks!

(worth reverting from Draft to normal PR to let me merge it?)

jcastill:
Moving to normal PR, thank you both!

PR #4167, opened 2025-11-25
[upload] "Generic Upload" directory issue

dwolstroRH:
When a "Generic Upload" is encountered (meaning a
	non-implemented distribution) and the upload directory
	is not specified, fail with a helpful error message (the
	upload would previously fail on permissions as it tries
	to write to the root directory of sftp).

Related: RHEL-124680

---

### DIFF
--- a/sos/upload/__init__.py
+++ b/sos/upload/__init__.py
@@ -204,6 +204,15 @@ this utility.
             self.upload_name = self.upload_target.name()
         self.ui_log.info(
                     f"Upload target set to {self.upload_name}")
+        if self.opts.upload_directory is None and \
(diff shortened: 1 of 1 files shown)


TurboTurtle:
I don't know that we want to impose this as an always-fail situation.

In the case of a configured `ChrootDirectory`, the session will automatically switch to the user's home dir after applying the chroot. This means that it is entirely a valid configuration to not specify a directory _and_ at the same time properly expect that the file being transferred end up in the ssh user's homedir instead of the rootdir of the sftp session.

This change would prevent an upload in such a configuration.

PR #4168, opened 2025-12-01
[aap_containerized] Collect same commands like other AAP plugins

pmoravec:
Let collect same command outputs from containerized deployments, likewise we do collect from RPM based deployments in all other aap_* plugins today.

Closes: #4168
Relevant-to: RHEL-129225

---

### DIFF
--- a/sos/report/plugins/aap_containerized.py
+++ b/sos/report/plugins/aap_containerized.py
@@ -125,18 +125,48 @@ class AAPContainerized(Plugin, RedHatPlugin):
                     subdir="podman_inspect_logs"
             )
 
-        if "automation-controller-task" in aap_containers:
-            container = "automation-controller-task"
-            podman_commands = [
-                (f"su - {username} -c 'podman exec -it {container} bash -c"
-                 " \"awx-manage check_license --data\"'",
-                 "awx-manage_check_license_--data"),
-                (f"su - {username} -c 'podman exec -it {container} bash -c"
-                 " \"awx-manage list_instances\"'",
-                 "awx-manage_list_instances"),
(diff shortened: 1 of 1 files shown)


sos/report/plugins/aap_containerized.py
```diff
@@ -125,18 +125,46 @@ def setup(self):
...
+                "receptorctl status",
+                "receptorclt work list",
+            ],
+        }
+        for pod, cmds in pod_cmds.items():
+            if pod in aap_containers:
+                for cmd in cmds:
+                    fname = self._mangle_command(cmd)
+                    self.add_cmd_output(f"su - {username} -c 'podman exec -it"
+                                        f"  {pod} bash -c \"{cmd}\"'",
```
TurboTurtle, line 167: We have probably discussed this in the past, but just for completeness sake, does `runas` not function here to avoid the embedded `su` calls?

pmoravec: I dont recall the original reason, but when I tried to use `runas=username` now, I am getting hung processes like:

```
aap       152503  0.0  0.0   3068  1664 pts/1    S    19:44   0:00 timeout 300s podman exec -it automation-controller-task bash -c awx-manage check_license --data
aap       152505  0.0  0.3 2023996 53964 pts/1   Sl   19:44   0:00 podman exec -it automation-controller-task bash -c awx-manage check_license --data
~~~

hung for minutes (while manual execution of the same takes a fraction of second).

pmoravec:
I spent some time trying to identify the cause but without a luck. So I added just the explanation to the code.

PR #4170, opened 2025-12-02
[drbd] Modernize and extend plugin

chrboe:
The current drbd plugin is aimed at older DRBD 8-based setups, and only collects some basic data.
Extend the plugin to collect more valuable data for real-world support use cases.
This version has DRBD 9 in mind, but stays backwards compatible to version 8.

---

### DIFF
--- a/sos/report/plugins/drbd.py
+++ b/sos/report/plugins/drbd.py
@@ -7,27 +7,74 @@
 # See the LICENSE file in the source distribution for further information.
 
 from sos.report.plugins import Plugin, RedHatPlugin, UbuntuPlugin
+from sos.utilities import find
 
 
-class DRDB(Plugin, RedHatPlugin, UbuntuPlugin):
+class DRBD(Plugin, RedHatPlugin, UbuntuPlugin):
 
     short_desc = 'Distributed Replicated Block Device (DRBD)'
 
     plugin_name = 'drbd'
     profiles = ('storage',)
-    packages = ('drbd*-utils',)
+    commands = ('drbdsetup',)
+
+    # In case of kernel bugs, drbdsetup may hang indefinitely.
+    # Set a shorter than default timeout.
+    cmd_timeout = 60
+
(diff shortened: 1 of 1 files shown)


sos/report/plugins/drbd.py
```diff
@@ -7,27 +7,78 @@
...
-            "drbdadm dump-xml",
-            "drbdsetup status all",
+            "drbdadm dump",
+            "drbdadm -d -vvv adjust all",
+            "drbdsetup status -vs all",
             "drbdsetup show all"
         ])
+        self.add_cmd_output([
+            # to identify potential processes stuck in D state
+            "sh -c 'ps aux | grep drbd'",
```
bmr-cymru: Please don't use the shell in callouts unless it's absolutely unavoidable - in this case there are plenty of existing `ps` grabs in the `process` plugin that can be trivially grepped for `drbd` processes from the resulting report and that provide broader `ps` option coverage. Plugins generally shouldn't grab generic stuff like `ps` "just in case" as it's redundant with the default data collected.

chrboe: Sounds reasonable, I've removed the call.

bmr-cymru (approved):

sos/report/plugins/drbd.py
```diff
@@ -7,27 +7,74 @@
...
+            "/proc/drbd",
+            "/sys/kernel/debug/drbd/*",
+            "/var/lib/drbd/*",
+            "/var/lib/drbd-support/*",
+            "/var/lib/linstor.d/*"
         ])
 
+    def postproc(self):
+        # Scrub nodehash from /var/lib/drbd-support/registration.json
+        nodehash_re = r'("nodehash"\s*:\s*")[a-zA-Z0-9]+"'
```
pmoravec, line 66: You *might* shorten the `[a-zA-Z0-9]` by simple and elegant `\w` but that will catch also `_`  you might not want to be matched. So / anyway ack with the current RE.

PR #4181, opened 2025-12-11
[stunnel] New plugin

ShreyasMahangade:
Stunnel is SSL based component which acts as Universal SSL Stunnel plugin will primarily collect the stunnel configuration files and chroot folder to check its existance which is needed for stunnel to work.

---

- [X ] Are all passwords or private data gathered by this PR [obfuscated](https://github.com/sosreport/sos/wiki/How-to-Write-a-Plugin#how-to-prevent-collecting-passwords)?

### DIFF
--- a/sos/report/plugins/stunnel.py
+++ b/sos/report/plugins/stunnel.py
@@ -1,0 +1,35 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, IndependentPlugin
+
+
+class Stunnel(Plugin, IndependentPlugin):
+
+    short_desc = 'Universal SSL Tunnel'
+    plugin_name = "stunnel"
+    profiles = ('network', 'security')
+
+    packages = ('stunnel',)
+
+    def setup(self):
+
+        self.add_service_status('stunnel*')
+        self.add_journal('stunnel*')
+
+        self.add_dir_listing(
+            '/var/run/stunnel',
+            tags=['ls_var_run_stunnel'],
+            recursive=True)
+
+        self.add_copy_spec([
+            "/etc/stunnel/*.conf",
+            "/var/log/stunnel.log"
+        ])
+
+# vim: set et ts=4 sw=4 :


sos/report/plugins/stunnel.py
```diff
@@ -0,0 +1,26 @@
...
+
+    short_desc = 'Universal SSL Tunnel'
+    plugin_name = "stunnel"
+    profiles = ('network', 'security')
+
+    packages = ('stunnel',)
+
+    def setup(self):
+
+        self.add_dir_listing('/var/run/stunnel', tags=['ls_var_run_stunnel'], recursive=True)
```
TurboTurtle: CI is failing on linting here as this is over 79 characters. sos-style is to indent one level from the calling line, with the closing paren inline, like so:

```python
self.add_dir_listing(
    '/var/run/stunnel',
    tags=['ls_var_run_stunnel'],
    recursive=True,
)
```

sos/report/plugins/stunnel.py
```diff
@@ -0,0 +1,26 @@
...
+
+        self.add_copy_spec(["/etc/stunnel/stunnel.conf", "/etc/stunnel/*.conf"])
```
TurboTurtle: These are redundant. You could either just go with `/etc/stunnel/*.conf` or grab the entire directory (if there's nothing we need to avoid there) with `/etc/stunnel/`.

The main reason you might see similar entries in other plugins would be around size limiting. Each list element gets `sizelimit` applied to it separately, so unless we can expect `stunnel.conf` to be over 25MB there's no need to separate it from the glob.

TurboTurtle (commented):
Looks good overall, just a couple minor comments below.

sos/report/plugins/stunnel.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class Stunnel(Plugin, IndependentPlugin):
+
+    short_desc = 'Universal SSL Tunnel'
+    plugin_name = "stunnel"
+    profiles = ('network', 'security')
+
+    packages = ('stunnel',)
+
```
jcastill, line 19: You may want to add the service here, like this:

`    services = ('stunnel',)
`

This way we'll capture details about the service, like its status and journal logs

sos/report/plugins/stunnel.py
```diff
@@ -0,0 +1,29 @@
...
+    def setup(self):
+
+        self.add_dir_listing(
+            '/var/run/stunnel', 
+            tags=['ls_var_run_stunnel'], 
+            recursive=True)
+
+        self.add_copy_spec(["/etc/stunnel/*.conf"])
+
```
jcastill, line 34: There may be a log file, '/var/log/stunnel.log', that could be useful to gather as well

sos/report/plugins/stunnel.py
```diff
@@ -0,0 +1,29 @@
...
+    plugin_name = "stunnel"
+    profiles = ('network', 'security')
+
+    packages = ('stunnel',)
+
+    def setup(self):
+
+        self.add_dir_listing(
+            '/var/run/stunnel', 
+            tags=['ls_var_run_stunnel'],
```
jcastill: You'll need to remove the whitespace at the end of this line and the one before to avoid the linting errors

sos/report/plugins/stunnel.py
ShreyasMahangade: I have done it differently.
I thought of including stunnel* because it will collect all stunnel.services including stunnel@other-config.service data as well.

ShreyasMahangade: Let me know if those wildcard commands have any dangers.. I will fallback to your suggestion.

sos/report/plugins/stunnel.py
```diff
@@ -0,0 +1,37 @@
...
+    profiles = ('network', 'security')
+
+    packages = ('stunnel',)
+
+    def setup(self):
+
+        self.add_cmd_output([
+            "systemctl status stunnel*",
+            "journalctl --no-pager --unit stunnel*"
+        ])
```
TurboTurtle: We actually have 2 wrappers for these. `Plugin.add_service_status()` and `Plugin.add_journal()`. The benefits of using these over `add_cmd_output()` are that we have standardized the collection patterns; for example with `add_journal()` you get more control over how the journal is collected, and it also uses a different size limit than typical collections.

With `add_service_status()` it (can) account for different init systems.

TurboTurtle (approved):

PR #4182, opened 2025-12-12
[gcp] Collecting RHUI details on GCP

pawelpbm:
- Enabling collection of RHUI connection on RHEL with Google repositories

---

### DIFF
--- a/sos/report/plugins/gcp.py
+++ b/sos/report/plugins/gcp.py
@@ -64,6 +64,15 @@ class GCP(Plugin, IndependentPlugin):
             except RuntimeError as err:
                 mfile.write(str(err))
 
+        # Collect zone.txt file from RHUI
+        curl_command = ('curl -v --connect-timeout 10'
+                        'https://rhui.googlecloud.com/zone.txt')
+
+        rhui_package = self.policy.package_manager.all_pkgs_by_name_regex(
+            r'google-rhui-client')
+        if rhui_package:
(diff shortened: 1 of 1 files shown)


sos/report/plugins/gcp.py
```diff
@@ -64,6 +64,14 @@ def collect(self):
             except RuntimeError as err:
                 mfile.write(str(err))
 
+        # Collect zone.txt file from RHUI
+        curl_command = 'curl -v https://rhui.googlecloud.com/zone.txt'
+
+        rhui_package = self.exec_cmd('rpm -qa google-rhui-client-*')
```
pmoravec: No need to run an extra command. You can use either `self.policy.package_manager.pkg_by_name('google-rhui-client')` (if that is the package name), or `self.policy.package_manager.all_pkgs_by_name_regex(r'google-rhui-client')` for a regexp search.

sos/report/plugins/gcp.py
```diff
@@ -64,6 +64,14 @@ def collect(self):
...
+        if (rhui_package['status'] == 0 and
+                'google-rhui-client' in rhui_package['output']):
+            self.collect_cmd_output(curl_command)
```
pmoravec, line 74: Worth adding some timeout? Default is 600s and in case of some routing issues, you probably dont want to wait 10 minutes (while I assume the `curl` command should terminate within seconds).

PR #4183, opened 2025-12-15
[plugin_aws]fix plugin is not loaded in aws xen instances

liangxiao1:
Plugin aws is not loaded in aws xen instances, update _is_ec2() to add bios_version check.
$ cat /sys/devices/virtual/dmi/id/sys_vendor
Xen
$ cat /sys/devices/virtual/dmi/id/bios_version
4.11.amazon

---

### DIFF
--- a/sos/report/plugins/aws.py
+++ b/sos/report/plugins/aws.py
@@ -19,12 +19,9 @@ class Aws(Plugin, IndependentPlugin):
     profiles = ('virt',)
 
     def _is_ec2(self):
-        try:
-            with open('/sys/devices/virtual/dmi/id/sys_vendor',
-                      encoding='utf-8') as f:
-                return 'Amazon' in f.read()
-        except FileNotFoundError:
-            return False
+        dmi_files = ("/sys/devices/virtual/dmi/id/sys_vendor",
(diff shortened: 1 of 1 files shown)


sos/report/plugins/aws.py
```diff
@@ -19,12 +19,12 @@ class Aws(Plugin, IndependentPlugin):
...
-                      encoding='utf-8') as f:
-                return 'Amazon' in f.read()
-        except FileNotFoundError:
-            return False
+        dmi_files = ("/sys/devices/virtual/dmi/id/sys_vendor",
+                     "/sys/devices/virtual/dmi/id/bios_version")
+        ret = self.file_grep(r"(Amazon|.*amazon)", *dmi_files)
+        if ret:
+            return True
+        return False
```
pmoravec: Isnt it simplier to have here `return ret != []` (or maybe merge it with the `self.file_grep` method since `ret` variable is not otherwise used)?

liangxiao1: Thanks for your suggestions, removed `ret` to make changes simpler.
Updated "ret != []" to bool(), because pylint complains about it.

Issue #4187, opened 2025-12-18
[logs] Log files not collected if using rsyslog with include()


oneoneonepig:
In plugin logs, the current implementation only parse log files specified in `$IncludeConfig`, but not `include()`.

https://github.com/sosreport/sos/blob/0e067966bcf793c247b834ec3df786eb1fb6e20e/sos/report/plugins/logs.py#L25-L29

[`include()`](https://www.rsyslog.com/doc/rainerscript/include.html) is supported since rsyslog [8.33.0](https://github.com/rsyslog/rsyslog/releases/tag/v8.33.0) (Feb 18, 2018) and it deprecates the older `$IncludeConfig` obsolete legacy format directive, it would be great if both are support.

Also, as a workaround, is there a way to specify custom log paths and pack them into the sos report?

TurboTurtle:
> [`include()`](https://www.rsyslog.com/doc/rainerscript/include.html) is supported since rsyslog [8.33.0](https://github.com/rsyslog/rsyslog/releases/tag/v8.33.0) (Feb 18, 2018) and it deprecates the older `$IncludeConfig` obsolete legacy format directive, it would be great if both are support.
>
ack. Shouldn't be too difficult to add support for that.
 
> Also, as a workaround, is there a way to specify custom log paths and pack them into the sos report?

[sos_extras](https://github.com/sosreport/sos/blob/main/sos/report/plugins/sos_extras.py) would probably be your best bet. Drop a file in `/etc/sos/extras.d/` matching the syntax specified in the docstring and it'll start grabbing those.

PR #4188, opened 2025-12-18
[logs] Add support for newer `include()` for rsyslog configs

TurboTurtle:
Adds support for parsing the newer-style `include()` directives within rsyslog config files. There is support for both file paths and limited support for echo'ing environment variables via the same `file` parameter, as described in the rsyslog documentation for `include()`.

Additionally, fixes an apparent bug where the `rsyslog` variable for the default `rsyslog.conf` location was missing the leading `/`.

Resolves: #4187

---

### DIFF
--- a/sos/report/plugins/logs.py
+++ b/sos/report/plugins/logs.py
@@ -7,6 +7,8 @@
 # See the LICENSE file in the source distribution for further information.
 
 import glob
+import os
+import re
 from sos.report.plugins import Plugin, PluginOpt, IndependentPlugin, CosPlugin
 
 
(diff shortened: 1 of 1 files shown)


oneoneonepig:
@TurboTurtle Thanks for the quick patch!

One question, does it handle [multiline configuration](https://www.rsyslog.com/doc/rainerscript/include.html)?

```
include(
   file="/path/to/include.conf"
   mode="optional"
)
```

```
include(
   file=`echo $ENV_VAR`
   mode="optional"
)
```

pmoravec (approved):

TurboTurtle:
> @TurboTurtle Thanks for the quick patch!
> 
> One question, does it handle [multiline configuration](https://www.rsyslog.com/doc/rainerscript/include.html)?
> 

Yep! :)

TurboTurtle:
I've updated the regex to support the single-line-with-multiple-options syntax, as you've suggested. I do think the exception you've noted is probably more on the artificial end of things.

sos/report/plugins/logs.py
```diff
@@ -18,13 +20,31 @@ class LogsBase(Plugin):
...
+                    string=content,
+                    flags=re.I | re.M
+                ):
+                    try:
+                        _ent = match[3]
+                        # best-effort to get env var provided files
+                        # this WILL break on anything other than basic echos
+                        # as shown in the rsyslog documentation
+                        if _ent.startswith('echo '):
+                            if envc := os.getenv(_ent.split()[1].strip(' $')):
```
pmoravec: Please add grave accent to `strip`:

```
strip(' $`')
```

, since for:

```
include(
   file=`echo $ENV_VAR`
   mode="optional"
)
```

the `_ent.split()[1]` is:

```
ENV_VAR`
```

with the extra trailing `.

TurboTurtle: Oh, nice catch.

pmoravec:
Merged via https://github.com/sosreport/sos/commit/67317e3edcf30b9cf64bc454347613af6c84a544.

PR #4193, opened 2026-01-04
[collect] Allow sudo alternatives for privilege escalation

TurboTurtle:
This commit allows for users to specify a different privilege escalation binary to use in place of `sudo`, for environments where that utility is not used.

We will perform a basic sanity test on any provided value by way of ensuring that the binary is in $PATH on the remote host.

Signed-off-by: Jake Hunsaker <jacob.r.hunsaker@gmail.com>

---

### DIFF
--- a/sos/collector/sosnode.py
+++ b/sos/collector/sosnode.py
@@ -95,6 +95,7 @@ class SosNode():
         # cluster primary but do not want a local report as we still need to do
         # package checks in that instance
         self.host = self.determine_host_policy()
+        self._sudo_binary = None
         self.hostname = self._transport.hostname
         if self.local and self.opts.no_local:
             load_facts = False
@@ -156,6 +157,21 @@ class SosNode():
                         _val = ln.split('=')
                         self._env_vars[_val[0]] = _val[1]
         return self._env_vars
+
+    @property
+    def sudo_binary(self):
+        if not self._sudo_binary:
+            _bin = self.opts.sudo_binary.split('/')[-1]
+            # verify the provided binary is at least in our PATH
+            if ret := self.run_command(f"command -v {_bin}"):
+                if not ret['status'] == 0:
+                    err = f"Privilege escalation command not in PATH: {_bin}"
+                    self.log_error(err)
+                    raise Exception(err)
+            if _bin == 'sudo':
+                _bin = 'sudo -S'
+            self._sudo_binary = _bin
+        return self._sudo_binary
 
     def set_node_manifest(self, manifest):
         """Set the manifest section that this node will write to
@@ -285,7 +301,7 @@ class SosNode():
         if self.opts.become_root:
             return f"su -c {quote(cmd)}"
         if self.need_sudo:
-            return f"sudo -S {cmd}"
+            return f"{self.sudo_binary} {cmd}"
         return cmd
 
     def _load_sos_info(self):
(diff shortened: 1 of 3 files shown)


sos/collector/sosnode.py
```diff
@@ -157,6 +158,21 @@ def env_vars(self):
                         self._env_vars[_val[0]] = _val[1]
         return self._env_vars
 
+    @property
+    def sudo_binary(self):
+        if not self._sudo_binary:
+            _bin = self.opts.sudo_binary.split('/')[-1]
+            # verify the provided binary is at least in our PATH
+            if ret := self.run_command(f"command -v {_bin}"):
```
pmoravec, line 166: Should not we test it on the remote system, and not locally? I.e. what if I have locally "just" `sudo`, but want to use, say, `doas`?

sos/collector/sosnode.py
```diff
@@ -157,6 +158,21 @@ def env_vars(self):
                         self._env_vars[_val[0]] = _val[1]
         return self._env_vars
 
+    @property
+    def sudo_binary(self):
+        if not self._sudo_binary:
+            _bin = self.opts.sudo_binary.split('/')[-1]
```
pmoravec, line 164: So you assume the option value can be specified with absolute path - but then you test if it is in `PATH`? That test would be redundant (worth locating the file on abs.location and its exec bit, then?)

Or did I get this wrong?

sos/collector/sosnode.py
TurboTurtle: This does run on the remote host.

sos/collector/sosnode.py
TurboTurtle: It can be passed as such, yes. However, we don't (or at least, I'm assuming we don't) want to allow arbitrary executables to be referenced here. By striping the absolute path and checking with `command`, we have a reliable way to ensure that the executable is at least in $PATH and not in some questionable (for some definition of that) location.

We could do some other validation earlier on perhaps, during option parsing. I don't have any strong feelings either way.

pmoravec: I mean when does it make sense to provide an absolute path, then? The current code will rule out `/some/weird/path/foosudo` that exists but is placed only on a weird location. But it will still allow running `/some/weird/path/doas` from that unusual location, as far as `doas` (probably from other location and probably some other binary) is in `PATH` - is this behaviour anticipated?

For me, this behaviour is kind of unexected, but I am not the target audience and see it as a nuance only. So if @TurboTurtle sees this as intentional, I am all good with it.

TurboTurtle: Assume a "proper" `doas` is at `/usr/sbin/doas`. A non-root user specifies `--sudo-binary /some/weird/path/doas` in their `sos` command.

In this scenario, the above code would run `command -v doas`, which would return truth-y, because of the "proper" binary in $PATH. Then, on command execution, we use `doas`, not the absolute path provided by the user, which would resolve to the "proper" `/usr/sbin/doas`.

This obviously does not protect against a polluted $PATH, but I'd say such a condition is outside of what we can reasonably be expected to handle.

pmoravec: Yeah and my question is if the user really wants to use the `/some/weird/path/doas` binary, or if they are OK the truth-y `/user/sbin/doas` will be used under the hood.

Anyway it is such a unusual corner case, that I am ok with the behaviour.

TurboTurtle:
`--xkcd-binary`  is tempting. ;)

I'm good with keeping it as `--sudo-binary`.

PR #4195, opened 2026-01-06
[report] Add UI message for post-processing

TurboTurtle:
On systems that generate larger sized sos reports, post-processing can take an appreciable amount of time. This isn't an issue in itself, however the delay between the `Finished running plugins` message and the `Creating compressed archive` message can lead to some ambiguity as to what `sos` is currently doing, or even if it has "hung".

Add a UI message to tell users that we are actually doing something between the end of `collect()` and compressing the archive.

---

### DIFF
--- a/sos/report/__init__.py
+++ b/sos/report/__init__.py
@@ -1880,9 +1880,10 @@ class SoSReport(SoSComponent):
             if not self.opts.no_report:
                 self.generate_reports()
             if not self.opts.no_postproc:
+                self.ui_log.info('Starting post-processing of collected data')
                 self.postproc()
             else:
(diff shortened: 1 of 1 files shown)


pmoravec (approved):
Great idea, this was indeed missing.

My only concern is the message is not much informative ("what postprocessing is sos report supposed to do?" is a legitimate user question). No better term in my mind, though (and "secrets obfuscation" doesn't sound good to me).

TurboTurtle:
I think for most purposes the generic "we're doing post-proc stuff" is fine, at least for now. If someone gets curious they'll probably go code diving, and if we need to do any kind of "we're doing post-proc for $plugin" like we do for execution, we can address it when it becomes an issue.

PR #4196, opened 2026-01-07
[PipeWire]: Add PipeWire Plugin

bryanfraschetti:
Create a new plugin for collection of PipeWire commands and configuration files

---

### DIFF
--- a/sos/report/plugins/pipewire.py
+++ b/sos/report/plugins/pipewire.py
@@ -23,7 +23,7 @@ class PipeWire(Plugin, IndependentPlugin):
     plugin_name = "pipewire"
     profiles = ('system', 'desktop', 'hardware')
 
-    packages = ('pipewire-bin', 'pipewire')
+    packages = ('pipewire-bin', 'pipewire-utils', 'pipewire')
 
     # PipeWire is scoped to user-session, but `sudo sos report` will cause
     # the commands to run in root context. If the logged in user is not root,


bmr-cymru (changes requested):
Generally looks good - one thing I did notice is that (at least on Fedora) when running `sos` as the root user (as usual), the `pw-cli`/`pw-dot` commands report:

```text
Error: "failed to connect: Host is down"
```

Since they're trying to connect to root's instance rather than the user who's actually logged in. It may be worth trying to figure that out and using `runas=` to execute the commands in that context.

sos/report/plugins/pipewire.py
```diff
@@ -0,0 +1,76 @@
...
+    """The PipeWire plugin collects information about the system's input
+    and output devices, their configuration, metadata, and map of logical
+    signal flow as managed by PipeWire
+    """
+
+    short_desc = 'A low-level multimedia framework for audio-visual playback'
+    plugin_name = "pipewire"
+    profiles = ('system', 'desktop', 'hardware')
+
+    packages = ('pipewire-bin', 'pipewire')
```
jcastill: In RHEL we also have pipewire-utils, and this package is the one that contains most of the binaries you are capturing below, so it may be worth adding it here

TurboTurtle (approved):
LGTM.

If we don't override the plugin option, we end up with `runas='root'` instead of the default `runas=None`, which will cause `sos_get_command_output()` to run the following snippet:

```python
        if runas:
            os.setgid(pwd.getpwnam(runas).pw_gid)
            os.setuid(pwd.getpwnam(runas).pw_uid)
            os.chdir(pwd.getpwnam(runas).pw_dir)
```

This shouldn't actually cause anything to change or happen, but thought this was an interesting edge case.

PR #4198, opened 2026-01-08
[charmed_opensearch] Add new plugin

Pepsiqqq:
Add a plugin for the VM charmed opensearch-operator:

GitHub: https://github.com/canonical/opensearch-operator

Collects:
- Opensearch config 
- Certificates paths 
- Journal for snap Opensearch
- Service status and info about service for snap Opensearch
- Logs from Opensearch
- Snap info about snap Opensearch

API endpoints from Opensearch:
1. /_cluster/settings
2. /_cluster/health
3. /_cluster/stats
4. /_cat/nodes
5. /_cat/indices
6. /_cat/shards
7. /_cat/aliases

---

### DIFF
--- a/sos/report/plugins/charmed_opensearch.py
+++ b/sos/report/plugins/charmed_opensearch.py
@@ -1,0 +1,146 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.#
+import os
+
+import yaml
+
+from sos.report.plugins import Plugin, PluginOpt, UbuntuPlugin
+
+
+class CharmedOpenSearch(Plugin, UbuntuPlugin):
+    short_desc = 'Charmed OpenSearch'
+    plugin_name = 'charmed_opensearch'
+    option_list = [
+        PluginOpt(
+            "user", default="admin", val_type=str,
+            desc="Username for opensearch, to check APIs"
+        ),
+        PluginOpt(
+            "password", default="", val_type=str,
+            desc="Password for opensearch, to check APIs",
+        ),
+        PluginOpt(
+            "all_journal", default=False, val_type=bool,
+            desc="Export all the journal entries",
+        )
+    ]
+
+    log_path = "/var/log/opensearch"
+    etc_path = "/etc/opensearch"
+    certificates_path = f"{etc_path}/certificates"
+    config_path = f"{etc_path}/opensearch.yml"
+    snap_current_path = "/var/snap/opensearch/current"
+    snap_common_path = "/var/snap/opensearch/common"
+    hostname = "localhost"
+    port = "9200"
+    packages = ('opensearch',)
+    services = ('jujud-machine-*.service',)
+
+    def setup(self):
+        opensearch_config_directory = (f"{self.snap_current_path}"
+                                       f"{self.etc_path}")
+        # FORBIDDEN
+        self.add_forbidden_path([
+            f"{opensearch_config_directory}/certificates",
+            f"{opensearch_config_directory}/opensearch.keystore",
+        ])
+
+        # CONFIGS
+        self.add_copy_spec(f"{opensearch_config_directory}/*")
+
+        # CERTIFICATES
+        self.add_dir_listing(
+            f"{self.snap_current_path}{self.certificates_path}",
+            suggest_filename="certificates_paths"
+        )
+
+        # JOURNAL
+        if self.get_option("all_journal"):
+            self.add_journal(units="snap.opensearch.*", allfields=True)
+        else:
+            self.add_journal(units="snap.opensearch.*", lines=1000,
+                             allfields=True)
+
+        # LOGS
+        if self.get_option("all_logs"):
+            self.add_copy_spec(f"{self.snap_common_path}{self.log_path}/*")
+        else:
+            self.add_copy_spec(f"{self.snap_common_path}{self.log_path}/*.log")
+
+        # SNAP
+        self.add_cmd_output(
+            "snap get opensearch -d",
+            suggest_filename="snap_get_opensearch"
+        )
+
+        # API
+        self.get_hostname_port(f"{self.snap_current_path}{self.config_path}")
+        base_url = f"https://{self.hostname}:{self.port}"
+        self.export_api(base_url)
+
+    def get_hostname_port(self, opensearch_config_file):
+        """ Get hostname and port number parsing as YAML """
+        try:
+            with open(opensearch_config_file, 'r', encoding='UTF-8') as f:
+                config = yaml.safe_load(f) or {}
+            if config.get('network.publish_host'):
+                self.hostname = str(config.get('network.publish_host'))
+            if config.get('http.port'):
+                self.port = str(config.get('http.port'))
+        except Exception as err:  # pylint: disable=broad-except
+            self._log_info(f"Failed to parse {opensearch_config_file}: {err}")
+
+    def export_api(self, base_url):
+        temp_env = os.environ.copy()
+        if not os.environ.get("OPENSEARCH_PWD"):
+            if not self.get_option('password'):
+                self.soslog.warning(
+                    "dump_error: password are not provided,"
+                    " skipping API dumps."
+                )
+                return
+            temp_env["OPENSEARCH_PWD"] = self.get_option('password')
+
+        if not os.environ.get("OPENSEARCH_USER"):
+            temp_env["OPENSEARCH_USER"] = self.get_option('user')
+
+        cluster_collect = [
+            "stats",
+            "settings",
+            "health",
+        ]
+        for collect in cluster_collect:
+            self.add_cmd_output(
+                f"sh -c 'curl -k -u $OPENSEARCH_USER:$OPENSEARCH_PWD "
+                f"-X GET {base_url}/_cluster/{collect}?pretty'",
+                suggest_filename=collect,
+                env=temp_env
+            )
+        cat_collect = [
+            "nodes?v",
+            "indices",
+            "shards",
+            "aliases",
+        ]
+        for collect in cat_collect:
+            self.add_cmd_output(
+                f"sh -c 'curl -k -u $OPENSEARCH_USER:$OPENSEARCH_PWD "
+                f"-X GET {base_url}/_cat/{collect}'",
+                suggest_filename=collect,
+                env=temp_env
+            )
+
+    def postproc(self):
+        # SCRUB PASSWORDS
+        opensearch_config_file = (f"{self.snap_current_path}"
+                                  f"{self.config_path}")
+        self.do_path_regex_sub(
+            f"{opensearch_config_file}",
+            r"(\s*plugins\.security\.ssl\..+password\s*:\s+).*",
+            r'\1"*********"',
+        )


sos/report/plugins/charmed_opensearch.py
```diff
@@ -0,0 +1,169 @@
...
+        )
+
+        # JOURNAL
+        self.add_journal(units="snap.opensearch.*", lines=1000, allfields=True)
+
+        # SERVICES
+        self.add_service_status("snap.opensearch.daemon",
+                                suggest_filename="service_status_opensearch")
+        self.add_copy_spec(
+            "/etc/systemd/system/"
```
pmoravec: Ditto from the other plugin: `systemd` plugin will aoutomatically collect this (until you disable `systemd` plugin).

sos/report/plugins/charmed_opensearch.py
```diff
@@ -0,0 +1,169 @@
...
+                                                'inet_diag'],
+                                   required={'kmods': 'all'})
+            res = self.exec_cmd("ss -tulpn", pred=ss_pred)
+            if not res['status'] == 0:
+                pofile.write(f"Unable to get ports list: {res['output']}")
+                return
+            lines = res['output'].splitlines()
+            filtered_ports = [
+                line for index, line in enumerate(lines)
+                if index == 0 or "9200" in line
```
pmoravec: Ditto with false positives when `9200` appears other than in port. Again, I guess it is ok? (or maybe grep for `:9200`?)

pmoravec: I realize: what if the port is not default? See your `get_hostname_port` method. (and ditto this in the other plugin).

sos/report/plugins/charmed_opensearch.py
```diff
@@ -0,0 +1,169 @@
...
+    log_path = "/var/log/opensearch"
+    etc_path = "/etc/opensearch"
+    certificates_path = f"{etc_path}/certificates"
+    config_path = f"{etc_path}/opensearch.yml"
+
+    snap_current_path = "/var/snap/opensearch/current"
+    snap_common_path = "/var/snap/opensearch/common"
+
+    def setup(self):
+        if not self.check_vm():
```
arif-ali: same comment on the other PR, wrt check_enabled() function.

In fact, if you just had packages = ("opensearch",), that could suffice, and this plugin would trigger if the opensearch snap is installed.

sos/report/plugins/charmed_opensearch.py
```diff
@@ -0,0 +1,169 @@
...
+        self.add_copy_spec(f"{opensearch_config_directory}/*")
+
+        # CERTIFICATES
+        self.add_dir_listing(
+            f"{self.snap_current_path}{self.certificates_path}",
+            suggest_filename="certificates_paths"
+        )
+
+        # JOURNAL
+        self.add_journal(units="snap.opensearch.*", lines=1000, allfields=True)
```
arif-ali: consider having an option for all_logs with lines=, and one with lines=

sos/report/plugins/charmed_opensearch.py
```diff
@@ -0,0 +1,198 @@
...
+    def export_api(self, base_url):
+        password = os.environ.get("OPENSEARCH_PWD",
+                                  self.get_option('password'))
+        if not password:
+            self.soslog.warning(
+                "dump_error: password are not provided, skipping API dumps."
+            )
+            return
+
+        with tempfile.NamedTemporaryFile(mode='w', delete=False) as temp_netrc:
```
pmoravec: Why not `delete=True` and you dont need to bother with `self.netrc_path` or its `os.remove`?

Pepsiqqq: The 'add_cmd_output' method does not execute the command when it is called. This means that the 'with' temporary file section closes and deletes the file before the 'curl' commands is executed. This means that curl cannot authorise OpenSearch, as the file containing the login information has been deleted. So I delete it in the postproc method.

sos/report/plugins/charmed_opensearch.py
```diff
@@ -0,0 +1,198 @@
...
+        base_url = f"https://{self.hostname}:{self.port}"
+        self.export_api(base_url)
+
+    def check_enabled(self):
+        return self.check_vm()
+
+    def check_vm(self):
+        return self.path_exists(self.snap_current_path)
+
+    def collect(self):
```
pmoravec: Both `PORTS` and `PROCESS TREE` will be "filtered duplicates" of data from other plugins (process and networking).

But I guess it is ok, it just slows down the plugin execution "redundantly", but if you / users of the plugin are ok with it, this is a tiny stuff.

sos/report/plugins/charmed_opensearch.py
```diff
@@ -0,0 +1,198 @@
...
+        with tempfile.NamedTemporaryFile(mode='w', delete=False) as temp_netrc:
+            temp_netrc.write(
+                f"machine {self.hostname}"
+                f" login {self.get_option('user')} password {password}\n")
+            self.netrc_path = temp_netrc.name
+            base_cmd = f"curl -k --netrc-file {self.netrc_path} -X GET"
+            self.add_cmd_output(
+                f"{base_cmd} '{base_url}/_cluster/settings?pretty'",
+                suggest_filename="settings"
+            )
```
pmoravec: You might have a list of tuples `(cmd, suggest_fname)` and iterate over them for `self.add_cmd_output`, likewise e.g. `foreman` plugin does at https://github.com/sosreport/sos/blob/main/sos/report/plugins/foreman.py#L287 .

Sometimes that approach improves maintainability, but sometimes it is worse for understanding the code.

I fully leave the decision on you, both have pros and cons, both is fully OK, imho.

sos/report/plugins/charmed_opensearch.py
pmoravec: Ah, I see, that makes sense. You can also call `collect_cmd_output` instead of `add_cmd_output` (the diff is the cmd is called immediatelly), but that would not gain any improvement, imho.

Maybe a comment in code why you postpone deletion of the tempfile can help?

Anyway, I am good wrt the PR, thanks!

pmoravec (approved):

sos/report/plugins/charmed_opensearch.py
```diff
@@ -0,0 +1,198 @@
...
+import os
+import re
+import tempfile
+
+from sos.report.plugins import Plugin, PluginOpt, UbuntuPlugin, SoSPredicate
+
+
+class CharmedOpenSearch(Plugin, UbuntuPlugin):
+    short_desc = 'Charmed OpenSearch'
+    plugin_name = 'charmed_opensearch'
```
pmoravec, line 17: This plugin is always enabled by default.

Is that intentional? SHould not be the plugin enabled only when some package or command or file or kernel module is present?

The same applies to `charmed_opensearch_dashboards` plugin.

Hint: run `sos help report.plugins.charmed_opensearch_dashboards` ;-)

arif-ali: @Pepsiqqq I made a similar comment a while back, this plugin will always run no matter what. Ideally you need to have a set of triggers set so that the plugin is only invoked when needed

sos/report/plugins/charmed_opensearch.py
```diff
@@ -0,0 +1,162 @@
...
+                    " skipping API dumps."
+                )
+                return
+            temp_env["OPENSEARCH_PWD"] = self.get_option('password')
+
+        if not os.environ.get("OPENSEARCH_USER"):
+            temp_env["OPENSEARCH_USER"] = self.get_option('user')
+
+        self.add_cmd_output(
+            f"sh -c 'curl -k -u $OPENSEARCH_USER:$OPENSEARCH_PWD "
```
arif-ali: Do we need this, `sh -c`, as the command will run as is, in the default shell of the system. Seems like this got added in the latest update

sos/report/plugins/charmed_opensearch.py
```diff
@@ -0,0 +1,162 @@
...
+                    "dump_error: password are not provided,"
+                    " skipping API dumps."
+                )
+                return
+            temp_env["OPENSEARCH_PWD"] = self.get_option('password')
+
+        if not os.environ.get("OPENSEARCH_USER"):
+            temp_env["OPENSEARCH_USER"] = self.get_option('user')
+
+        self.add_cmd_output(
```
arif-ali: Wouldn't it be nicer and neater to have a list of items you collect, and then parse through the list, something like below

```python

cluster_collect = [
    "settings",
    "health",
    "stats",
]

for collect in cluster_collect:

   self.add_cmd_output(
        f"curl -k -u $OPENSEARCH_USER:$OPENSEARCH_PWD "
        f"-X GET {base_url}/_cluster/{collect}?pretty",
        suggest_filename=collect,
        env=temp_env
    )

cat_collect = [
    "nodes",
    "indices",
    "shards",
    "aliases",
]

for collect in cat_collect:

   self.add_cmd_output(
        f"curl -k -u $OPENSEARCH_USER:$OPENSEARCH_PWD "
        f"-X GET {base_url}/_cat/{collect}?pretty",
        suggest_filename=collect,
        env=temp_env
    )
```

sos/report/plugins/charmed_opensearch.py
Pepsiqqq: Because SOS report uses Popen with shell=false to collect the command [link to Popen function](https://github.com/sosreport/sos/blob/main/sos/utilities.py#L340). If I do it without sh -c the resulting command will look like this: "curl -k -u : GET some-link". On either side of the colon, there should be credentials, but they will be empty. The environment variables will be passed to the command, but the command will be formed before that, so the password and username will not be included in the command.

arif-ali: ack, sounds good

arif-ali: as per our private discussions, is it worth trying to use the environment variables by escaping the $ in the curl command without the `sh -c`, and see if this would work. If this doesn't work, then we can go with the current version, and we can come back to the shell=false issue being the default thing for a future backlog item

Pepsiqqq: No, if I do it like 

            self.add_cmd_output(
                f"curl -k -u \$OPENSEARCH_USER:\$OPENSEARCH_PWD -X GET {base_url}/_cat/{collect}",
                suggest_filename=collect,
                env=temp_env
            )
it just returns `Unauthorized`

arif-ali: no worries, thanks for checking, I think I'm ok with it

jcastill:
@arif-ali I take it you added these two labels to have @pmoravec's review, right? 
Pavel, can you check the latest changes and see if we can merge this?

sos/report/plugins/charmed_opensearch.py
```diff
@@ -0,0 +1,146 @@
...
+        # SNAP
+        self.add_cmd_output(
+            "snap get opensearch -d",
+            suggest_filename="snap_get_opensearch"
+        )
+
+        # API
+        self.get_hostname_port(f"{self.snap_current_path}{self.config_path}")
+        base_url = f"https://{self.hostname}:{self.port}"
+        self.export_api(base_url)
```
pmoravec, line 84: nitpick: you might directly call ` self.export_api(f"https://{self.hostname}:{self.port}") to avoid one extra variable, but this makes the code more clear. I am ok with this.

sos/report/plugins/charmed_opensearch.py
```diff
@@ -0,0 +1,146 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.#
+import os
+
+import yaml
```
pmoravec, line 10: Sorry for a late notice, but afaik `yaml` is not a common library in Python (at least on Fedora/RHEL, while this plugin is for Ubuntu "only").

Is the library always present on Ubuntu/Debian? If not, won't this harm..? (I dont have such system at hand for a quick test, ATM).

arif-ali: The yaml import is actually being used by other plugins too like `sunbeam`, `mongodb`, `saltmaster.py` and `charmed_postgresql` as well as the `openstack` collector. In both Debian and Ubuntu packaging we already have `python3-yaml` as a dependency for this reason. So no issue from the Debian/Ubuntu side of things.

PR #4199, opened 2026-01-08
[charmed_opensearch_dashboards] Add new plugin

Pepsiqqq:
Add a plugin for the VM charmed opensearch-dashboards-operator:

GitHub: https://github.com/canonical/opensearch-dashboards-operator

Collects:

- Opensearch dashboards configs
- Logs from Opensearch dashboards
- Journal for snap Opensearch dashboards
- Service status and info about service for snap Opensearch dashboards
- Snap info about snap Opensearch dashboards
- Certificates paths 

API endpoints from Opensearch dashboards:

1. /api/status
2. /api/stats?extended=true
3. /api/saved_objects/_find?"type=dashboard&type=visualization&type=index-pattern&type=search&per_page=1&fields=id

---

### DIFF
--- a/sos/report/plugins/charmed_opensearch_dashboards.py
+++ b/sos/report/plugins/charmed_opensearch_dashboards.py
@@ -1,0 +1,143 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.#
+import os
+import yaml
+
+from sos.report.plugins import Plugin, PluginOpt, UbuntuPlugin
+
+
+class OpenSearchDashboards(Plugin, UbuntuPlugin):
+    short_desc = "Charmed OpenSearch Dashboards"
+    plugin_name = "charmed_opensearch_dashboards"
+    option_list = [
+        PluginOpt(
+            "user", default="admin", val_type=str,
+            desc="Username for opensearch, to check APIs"
+        ),
+        PluginOpt(
+            "password", default="", val_type=str,
+            desc="Password for opensearch, to check APIs",
+        ),
+        PluginOpt(
+            "all_journal", default=False, val_type=bool,
+            desc="Export all the journal entries",
+        )
+    ]
+
+    log_path = "/var/log/opensearch-dashboards"
+    etc_path = "/etc/opensearch-dashboards"
+    config_path = f"{etc_path}/opensearch_dashboards.yml"
+    certificates_path = f"{etc_path}/certificates"
+    node_config_path = f"{etc_path}/node.options"
+    snap_current_path = "/var/snap/opensearch-dashboards/current"
+    snap_common_path = "/var/snap/opensearch-dashboards/common"
+    hostname = "localhost"
+    port = "5601"
+    packages = ('opensearch-dashboards',)
+
+    def setup(self):
+        # CONFIGS
+        opensearch_dashboard_config_file = \
+            f"{self.snap_current_path}{self.config_path}"
+
+        self.add_copy_spec(opensearch_dashboard_config_file)
+        self.add_copy_spec(
+            f"{self.snap_current_path}{self.node_config_path}")
+
+        # LOGS
+        if self.get_option("all_logs"):
+            self.add_copy_spec(f"{self.snap_common_path}{self.log_path}/*")
+        else:
+            self.add_copy_spec(
+                f"{self.snap_common_path}{self.log_path}/*.log")
+
+        # JOURNAL
+        if self.get_option("all_journal"):
+            self.add_journal(units="snap.opensearch-dashboards.*",
+                             allfields=True)
+        else:
+            self.add_journal(units="snap.opensearch-dashboards.*", lines=1000,
+                             allfields=True)
+
+        # SNAP
+        self.add_cmd_output(
+            "snap get opensearch-dashboards -d",
+            suggest_filename="snap_get_opensearch_dashboard"
+        )
+
+        # CERTIFICATES
+        self.add_dir_listing(
+            f"{self.snap_current_path}{self.certificates_path}",
+            suggest_filename="certificates_paths"
+        )
+
+        # API
+        self.get_hostname_port(opensearch_dashboard_config_file)
+        base_url = f"https://{self.hostname}:{self.port}"
+        self.export_api(base_url)
+
+    def get_hostname_port(self, opensearch_config_file):
+        """ Get hostname and port number parsing as YAML """
+        try:
+            with open(opensearch_config_file, 'r', encoding='UTF-8') as f:
+                config = yaml.safe_load(f) or {}
+
+            if config.get('server.host'):
+                self.hostname = str(config.get('server.host'))
+
+            if config.get('server.port'):
+                self.port = str(config.get('server.port'))
+
+        except Exception as err:  # pylint: disable=broad-except
+            self._log_info(f"Failed to parse {opensearch_config_file}: {err}")
+
+    def export_api(self, base_url):
+        temp_env = os.environ.copy()
+        if not os.environ.get("OPENSEARCH_PWD"):
+            if not self.get_option('password'):
+                self.soslog.warning(
+                    "dump_error: password are not provided,"
+                    " skipping API dumps."
+                )
+                return
+            temp_env["OPENSEARCH_PWD"] = self.get_option('password')
+
+        if not os.environ.get("OPENSEARCH_USER"):
+            temp_env["OPENSEARCH_USER"] = self.get_option('user')
+
+        query = ("type=dashboard&"
+                 "type=visualization&"
+                 "type=index-pattern&"
+                 "type=search&per_page=1&"
+                 "fields=id")
+
+        self.add_cmd_output(
+            f"sh -c 'curl -k -u $OPENSEARCH_USER:$OPENSEARCH_PWD "
+            f"-X GET {base_url}/api/settings'",
+            suggest_filename="settings",
+            env=temp_env)
+        self.add_cmd_output(
+            f"sh -c 'curl -k -u $OPENSEARCH_USER:$OPENSEARCH_PWD "
+            f"-X GET {base_url}/api/stats?extended=true'",
+            suggest_filename="stats",
+            env=temp_env)
+        self.add_cmd_output(
+            f"sh -c 'curl -k -u $OPENSEARCH_USER:$OPENSEARCH_PWD "
+            f"-X GET {base_url}/api/saved_objects/_find?{query}'",
+            suggest_filename="saved_object_count",
+            env=temp_env)
+
+    def postproc(self):
+        # SCRUB PASSWORDS
+        opensearch_dashboard_config_file = (f"{self.snap_current_path}"
+                                            f"{self.config_path}")
+        self.do_path_regex_sub(
+            f"{opensearch_dashboard_config_file}",
+            r"(\s*opensearch\.password\s*:\s+).*",
+            r'\1"*********"',
+        )


jcastill:
The errors you are getting in ReportStageOne are related to this:

`(185/281) tests/report_tests/help_output_tests.py:ReportListPluginsTest.test_plugin_formatting:  FAIL: Regex didn't match: ' ([\\S ]){20} ([\\S ])*' not found in ' charmed_opensearch_dashboard Charmed OpenSearch Dashboard' (0.06 s)`

The problem is that when running that test, it's expecting that the first column will be 20 chars long, but the plugin name is longer than that:

`plugin_name = 'charmed_opensearch_dashboard'`

This is what we get when we run `sos report --list-plugins`:

```
...
 buildah              Buildah container and image builder
 cgroups              Control groups subsystem
 charmed_opensearch_dashboard Charmed OpenSearch Dashboard
 chrony               Chrony clock (for Network time protocol)
 cockpit              Cockpit Web Service
...
```

I'm not sure what could be a meaningful shorter string for this specific plugin, so it may be worth considering the increase of the column size and the test. I'll prepare a separate PR so we keep the discussion here related to the plugin itself.

sos/report/plugins/charmed_opensearch_dashboard.py
```diff
@@ -0,0 +1,170 @@
...
+            desc="Username for opensearch, to check APIs"
+        ),
+        PluginOpt(
+            "pass", default="", val_type=str,
+            desc="Password for opensearch, to check APIs",
+        ),
+        PluginOpt(
+            "all_log", default=True, val_type=bool,
+            desc="Export all logs",
+        )
```
TurboTurtle: Don't use a plugin option for this; gating "all logs" vs "smart default selection" is controlled by the global `--all-logs` option.

sos/report/plugins/charmed_opensearch_dashboard.py
```diff
@@ -0,0 +1,170 @@
...
+class OpenSearchDashboard(Plugin, UbuntuPlugin):
+    short_desc = 'Charmed OpenSearch Dashboard'
+    plugin_name = 'charmed_opensearch_dashboard'
+    option_list = [
+        PluginOpt(
+            "user", default="admin", val_type=str,
+            desc="Username for opensearch, to check APIs"
+        ),
+        PluginOpt(
+            "pass", default="", val_type=str,
```
TurboTurtle: For consistency with other options, please use the full word `password` here.

sos/report/plugins/charmed_opensearch_dashboard.py
```diff
@@ -0,0 +1,170 @@
...
+
+    user, password = None, None
+
+    def setup(self):
+        self.user, self.password = self.get_credentials()
+        if self.check_vm():
+            self.export_vm()
+
+    def check_vm(self):
+        return os.path.exists(self.snap_current_path)
```
TurboTurtle: Use `self.path_exists()` here, as this automatically accounts for differing `sysroot`.

sos/report/plugins/charmed_opensearch_dashboard.py
```diff
@@ -0,0 +1,170 @@
...
+            suggest_filename="service_status_opensearch_dashboards_daemon")
+        self.add_cmd_output(
+            "systemctl cat snap.opensearch-dashboards.exporter-daemon.service",
+            suggest_filename="service_cat_dashboard_exporter"
+        )
+        self.add_cmd_output(
+            "systemctl cat snap.opensearch-dashboards.opensearch-dashboards"
+            "-daemon.service",
+            suggest_filename="service_cat_dashboard_opensearch"
+        )
```
TurboTurtle: Rather than the `systemctl cat` output, can we not just collect the service unit file?

sos/report/plugins/charmed_opensearch_dashboard.py
```diff
@@ -0,0 +1,170 @@
...
+            "snap get opensearch-dashboards -d",
+            suggest_filename="snap_get_opensearch_dashboard"
+        )
+
+        # CERTIFICATES
+        # We don't export certificates, just paths
+        self.add_cmd_output(
+            f"ls -la {self.snap_current_path}{self.certificates_path}",
+            suggest_filename="certificates_paths"
+        )
```
TurboTurtle: We standardize our path listing collections via `self.add_dir_listing()`.

sos/report/plugins/charmed_opensearch_dashboard.py
```diff
@@ -0,0 +1,170 @@
...
+        self.add_cmd_output(
+            "sh -c \"ss -tulpn | awk 'NR==1 || /5601/'\"",
+            suggest_filename="listening_ports"
+        )
+
+        # PROCESS TREE
+        self.add_cmd_output(
+            "sh -c \"ps aux | awk 'NR==1 || /opensearch-dashboards/'\"",
+            suggest_filename="process_tree"
+        )
```
TurboTurtle: No embedded shell scripting in plugin collections. If you need to modify data, do so during the `collect()` phase with the `self.collection_file()` mechanism. See the `process`, `python`, or `unpackaged` plugins for some examples.

sos/report/plugins/charmed_opensearch_dashboard.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def postproc(self):
+        # SCRUB PASSWORDS
+        if self.check_vm():
+            opensearch_dashboard_config_file = (f"{self.snap_current_path}"
+                                                f"{self.config_path}")
+        else:
+            opensearch_dashboard_config_file = ""
+
+        self.do_path_regex_sub(
```
TurboTurtle: If we're meant to skip post-proc when `check_vm()` returns False, then move the `do_path_regex_sub()` under the conditional. That way we're not running a superfluous call.

sos/report/plugins/charmed_opensearch_dashboard.py
```diff
@@ -0,0 +1,170 @@
...
+            self._log_info(f"Failed to parse {opensearch_config_file}: {err}")
+        return hostname, port
+
+    def get_credentials(self):
+        user = self.get_option("user")
+        password = self.get_option("pass")
+        return user, password
+
+    def export_api(self, base_url):
+        base_cmd = f"curl -s -k -u {self.user}:{self.password} -X GET"
```
TurboTurtle: A couple thoughts here:

1. We're not doing any validation here, so what happens when a user doesn't pass a password in? Is the default config a passwordless `admin` account? Or do we expect the following calls to fail?
2. Since we're only using these in one spot, we can likely drop the method and just reference the options directly.

E.G.

```python
def export_api(self, base_url):
    if not (user := self.get_option('user')) or not (pass := self.get_option('password')):
        return  # no password, no way to hit api?
    base_cmd = f"curl -s -k -u {user}:{password} -X GET"
    [...]
```

sos/report/plugins/charmed_opensearch_dashboards.py
```diff
@@ -0,0 +1,170 @@
...
+        # API
+        host, port = self.get_hostname_port(
+            opensearch_dashboard_config_file)
+        base_url = f"http://{host}:{port}"
+        self.export_api(base_url)
+
+    def collect(self):
+        # PORTS
+        with self.collection_file('listening_ports') as pofile:
+            res = self.exec_cmd("ss -tulpn")
```
pmoravec: The tests are failing on newly loaded kernel modules:

```
[stdlog] 2026-01-12 14:29:20,845 avocado.test stacktrace       L0049 ERROR| avocado.core.exceptions.TestFail: Element counts were not equal:
[stdlog] 2026-01-12 14:29:20,845 avocado.test stacktrace       L0049 ERROR| First has 0, Second has 1:  'inet_diag'
[stdlog] 2026-01-12 14:29:20,845 avocado.test stacktrace       L0049 ERROR| First has 0, Second has 1:  'tcp_diag'
[stdlog] 2026-01-12 14:29:20,845 avocado.test stacktrace       L0049 ERROR| First has 0, Second has 1:  'udp_diag'
```

Quickly reviewing the plugin, I think the `ss` execution did it. See `ss_kmods` starting from https://github.com/sosreport/sos/blob/main/sos/report/plugins/networking.py#L42, expanded on  line 335 and below, and guarding the `ss ..` commands in https://github.com/sosreport/sos/blob/main/sos/report/plugins/networking.py#L199-L201 .

Warning: maybe some other command(s) can load the kmods, you might need to iteratively identify them.

sos/report/plugins/charmed_opensearch_dashboards.py
```diff
@@ -0,0 +1,173 @@
...
+        base_url = f"http://{host}:{port}"
+        self.export_api(base_url)
+
+    def collect(self):
+        # PORTS
+        with self.collection_file('listening_ports') as pofile:
+            ss_pred = SoSPredicate(self, kmods=['tcp_diag', 'udp_diag',
+                                                'inet_diag'],
+                                   required={'kmods': 'all'})
+            res = self.exec_cmd("ss -tulpn", pred=ss_pred)
```
pmoravec: \o/ just this helped, great!

sos/report/plugins/charmed_opensearch_dashboards.py
```diff
@@ -0,0 +1,173 @@
...
+        self.add_service_status(
+            "snap.opensearch-dashboards.opensearch-dashboards-daemon",
+            suggest_filename="service_status_opensearch_dashboards_daemon"
+        )
+        self.add_copy_spec([
+            "/etc/systemd/system/"
+            "snap.opensearch-dashboards.exporter-daemon.service",
+            "/etc/systemd/system/"
+            "snap.opensearch-dashboards.opensearch-dashboards-daemon.service",
+        ])
```
pmoravec: We copy whole `/etc/systemd` directory in `systemd` plugin. The generic "dont cross-copy-spec data by multiple plugins, sizelimit can bite you" wont much apply here. But if you dont plan disabling `systemd` during `sos report` collection, this `add_copy_spec` will be redundant.

sos/report/plugins/charmed_opensearch_dashboards.py
```diff
@@ -0,0 +1,173 @@
...
+            self._log_info(f"Failed to parse {opensearch_config_file}: {err}")
+        return hostname, port
+
+    def export_api(self, base_url):
+        if not (password := self.get_option('password')):
+            return
+        base_cmd = f"curl -s -k -u {self.get_option('user')}:{password} -X GET"
+
+        self.add_cmd_output(f"{base_cmd} '{base_url}/api/status'",
+                            suggest_filename="status")
```
pmoravec: The password will be exposed in several places in the sosreport archive (in `manifest.json`, `sos.log`, at least). I think this is inevitable (and I think we have it in few other places?), just be aware of it (r ensure if it is ok?).

sos/report/plugins/charmed_opensearch_dashboards.py
```diff
@@ -0,0 +1,173 @@
...
+                                                'inet_diag'],
+                                   required={'kmods': 'all'})
+            res = self.exec_cmd("ss -tulpn", pred=ss_pred)
+            if not res['status'] == 0:
+                pofile.write(f"Unable to get ports list: {res['output']}")
+                return
+            lines = res['output'].splitlines()
+            filtered_ports = [
+                line for index, line in enumerate(lines)
+                if index == 0 or "5601" in line
```
pmoravec: This will return "false positive" lines like:

```
tcp            LISTEN          0               4096                               [::1]:631                               [::]:*              users:(("cupsd",pid=5601,fd=7))                 
tcp            LISTEN          0               56015                               [::1]:631                               [::]:*              users:(("cupsd",pid=123,fd=7))                 
```

but I assume it doesnt matter?

sos/report/plugins/charmed_opensearch_dashboards.py
```diff
@@ -0,0 +1,173 @@
...
+        # LOGS
+        if self.get_option("all_logs"):
+            self.add_copy_spec(f"{self.snap_common_path}{self.log_path}/*")
+        else:
+            self.add_copy_spec(
+                f"{self.snap_common_path}{self.log_path}/*.log")
+
+        # JOURNAL
+        self.add_journal(units="snap.opensearch-dashboards.*", lines=1000,
+                         allfields=True)
```
arif-ali: is it worth having an all_logs option, where we don't specify the lines, so that we get more logs?

sos/report/plugins/charmed_opensearch_dashboards.py
```diff
@@ -0,0 +1,173 @@
...
+        # API
+        host, port = self.get_hostname_port(
+            opensearch_dashboard_config_file)
+        base_url = f"http://{host}:{port}"
+        self.export_api(base_url)
+
+    def collect(self):
+        # PORTS
+        with self.collection_file('listening_ports') as pofile:
+            ss_pred = SoSPredicate(self, kmods=['tcp_diag', 'udp_diag',
```
arif-ali: we need to ensure this are all the predicates, as we found in a previous issue with the networking plugin, in the end we needed to add a lot more predicates. I needed to test a lot to get to the final list. Just something to think about.

sos/report/plugins/charmed_opensearch_dashboards.py
```diff
@@ -0,0 +1,173 @@
...
+# See the LICENSE file in the source distribution for further information.#
+
+import re
+
+from sos.report.plugins import Plugin, PluginOpt, UbuntuPlugin, SoSPredicate
+
+
+class OpenSearchDashboards(Plugin, UbuntuPlugin):
+    short_desc = "Charmed OpenSearch Dashboards"
+    plugin_name = "charmed_opensearchui"
```
arif-ali: The plugin_name seems to differ from the name of the plugin itself, was that on purpose? When you list the plugins, it will show as `charmed_opensearchui`, but when you look through the code, then we'll see `charmed_opensearch_dashboards.py`

sos/report/plugins/charmed_opensearch_dashboards.py
```diff
@@ -0,0 +1,173 @@
...
+    etc_path = "/etc/opensearch-dashboards"
+    config_path = f"{etc_path}/opensearch_dashboards.yml"
+    certificates_path = f"{etc_path}/certificates"
+    node_config_path = f"{etc_path}/node.options"
+
+    snap_current_path = "/var/snap/opensearch-dashboards/current"
+    snap_common_path = "/var/snap/opensearch-dashboards/common"
+
+    def setup(self):
+        if not self.check_vm():
```
arif-ali: couldn't we just add, and expand the `check_enabled()` function, that's the purpose of that to see if this plugin should be enabled or not.

Then this tests doesn't need to be run here, and instead in the `check_enabled()` function

arif-ali: To add, you could also just ensure that the opensearch-dashboards snap is installed, and that can reduce the number of lines of code, and allow `sos` to do the triggering for you by adding `packages = ('opensearch-dashboards',) below the class definition.

sos/report/plugins/charmed_opensearch_dashboards.py
Pepsiqqq: If I add another predicate, the command breaks for some reason and returns nothing

sos/report/plugins/charmed_opensearch_dashboards.py
```diff
@@ -0,0 +1,183 @@
...
+import os
+import re
+import tempfile
+
+from sos.report.plugins import Plugin, PluginOpt, UbuntuPlugin, SoSPredicate
+
+
+class OpenSearchDashboards(Plugin, UbuntuPlugin):
+    short_desc = "Charmed OpenSearch Dashboards"
+    plugin_name = "charmed_opensearch_dashboards"
```
pmoravec, line 16: Ditto with "This plugin is always enabled by default."  (https://github.com/sosreport/sos/pull/4198/changes#r2740532023)

sos/report/plugins/charmed_opensearch_dashboards.py
```diff
@@ -0,0 +1,183 @@
...
+                                                'inet_diag'],
+                                   required={'kmods': 'all'})
+            res = self.exec_cmd("ss -tulpn", pred=ss_pred)
+            if not res['status'] == 0:
+                pofile.write(f"Unable to get ports list: {res['output']}")
+                return
+            lines = res['output'].splitlines()
+            filtered_ports = [
+                line for index, line in enumerate(lines)
+                if index == 0 or f":{self.port}" in line
```
pmoravec: Theoretically, you get "false positive matches" when a process listens on ports like 5601[0-9]. If it is OK for you, no prob for me :).

sos/report/plugins/charmed_opensearch_dashboards.py
TurboTurtle: I think maybe Arif was meaning all the kmods in that predicate. The `networking` plugin specifies a lot more kmods in order to gate `ss` commands, and that has been a thorny issue in the past.

At the moment we do only support one predicate per a_c_o call, but the predicate can check any number of supported elements.

sos/report/plugins/charmed_opensearch_dashboards.py
```diff
@@ -0,0 +1,183 @@
...
+                "dump_error: password are not provided, skipping API dumps."
+            )
+            return
+
+        with tempfile.NamedTemporaryFile(mode='w', delete=False) as temp_netrc:
+            temp_netrc.write(
+                f"machine {self.hostname}"
+                f" login {self.get_option('user')} password {password}\n")
+            self.netrc_path = temp_netrc.name
+            base_cmd = f"curl -k --netrc-file {self.netrc_path} -X GET"
```
TurboTurtle: I'm not a fan of this approach. If something causes sos to exit early, even if that's the user doing it manually, that means we have a credential file hanging around.

Another approach would be to save these strings as env vars in a dict, and then pass those vars directly in the a_c_o call, referencing those variables in the `curl` command.

pmoravec: > Another approach would be to save these strings as env vars in a dict, and then pass those vars directly in the a_c_o call, referencing those variables in the curl command.

Not sure how do you mean this, but can't that leak the credentials to `ps aux` output?

TurboTurtle: No. I mean, for example:

```python
my_env = {
    'FOO': 'baz',
    'BAR': 'zab'
}

self.add_cmd_output(
    'curl -u "${FOO}:${BAR}" [...]',
    env=my_env
)
```

The `subprocess.Popen()` call that a_c_o ultimately results in will inject the `my_env` variables into the command execution.

Pepsiqqq: I tried to do what you suggested in your example, but the environment variables were passed to "curl" without being expanded in the shell, resulting in empty strings instead of credentials and an 'Unauthorised' response from OpenSearch.
Here is what I tested with:
```
        password = os.environ.get("OPENSEARCH_PWD",
                                  self.get_option('password'))
        if not password:
            self.soslog.warning(
                "dump_error: password are not provided, skipping API dumps."
            )
            return
        my_env = {
            'OPENSEARCH_USER':self.get_option('user'),
            'OPENSEARCH_PWD':password,
        }

        self.add_cmd_output(
            f'curl -k -u ${{OPENSEARCH_USER}}:${{OPENSEARCH_PWD}} -X GET "{base_url}/_cluster/settings?pretty"',
            env=my_env,
            suggest_filename="settings"
        )
        self.add_cmd_output(
            f'curl -k -u "$OPENSEARCH_USER:$OPENSEARCH_PWD" -X GET {base_url}/_cluster/settings?pretty',
            env=my_env,
            suggest_filename="settings2"
        )
        self.add_cmd_output(
            f'curl -k -u "${{OPENSEARCH_USER}}:{{$OPENSEARCH_PWD}}" -X GET {base_url}/_cluster/settings?pretty',
            env=my_env,
            suggest_filename="settings3"
        )
```
Here is how it looks in the sos.log
```
2026-02-02 11:30:45,048 INFO: [plugin:charmed_opensearch] added cmd output 'curl -k -u ${OPENSEARCH_USER}:${OPENSEARCH_PWD} -X GET "https://10.79.169.121:9200/_cluster/settings?pretty"'
2026-02-02 11:30:45,048 INFO: [plugin:charmed_opensearch] added cmd output 'curl -k -u "$OPENSEARCH_USER:$OPENSEARCH_PWD" -X GET https://10.79.169.121:9200/_cluster/settings?pretty'
2026-02-02 11:30:45,048 INFO: [plugin:charmed_opensearch] added cmd output 'curl -k -u "${OPENSEARCH_USER}:{$OPENSEARCH_PWD}" -X GET https://10.79.169.121:9200/_cluster/settings?pretty'
```
And in manifest.json:
```
{
                            "command": "curl",
                            "parameters": [
                                "-k",
                                "-u",
                                "${OPENSEARCH_USER}:${OPENSEARCH_PWD}",
                                "-X",
                                "GET",
                                "\"https://10.79.169.121:9200/_cluster/settings?pretty\""
                            ],
                            "exec": "curl -k -u ${OPENSEARCH_USER}:${OPENSEARCH_PWD} -X GET \"https://10.79.169.121:9200/_cluster/settings?pretty\"",
                            "filepath": "sos_commands/charmed_opensearch/settings",
                            "truncated": false,
                            "return_code": 0,
                            "priority": 10,
                            "start_time": 1770031845.327699,
                            "end_time": 1770031845.4999578,
                            "run_time": 0.17225885391235352,
                            "tags": []
                        },
                        {
                            "command": "curl",
                            "parameters": [
                                "-k",
                                "-u",
                                "\"$OPENSEARCH_USER:$OPENSEARCH_PWD\"",
                                "-X",
                                "GET",
                                "https://10.79.169.121:9200/_cluster/settings?pretty"
                            ],
                            "exec": "curl -k -u \"$OPENSEARCH_USER:$OPENSEARCH_PWD\" -X GET https://10.79.169.121:9200/_cluster/settings?pretty",
                            "filepath": "sos_commands/charmed_opensearch/settings2",
                            "truncated": false,
                            "return_code": 0,
                            "priority": 10,
                            "start_time": 1770031845.5002406,
                            "end_time": 1770031845.6827571,
                            "run_time": 0.18251657485961914,
                            "tags": []
                        },
                        {
                            "command": "curl",
                            "parameters": [
                                "-k",
                                "-u",
                                "\"${OPENSEARCH_USER}:{$OPENSEARCH_PWD}\"",
                                "-X",
                                "GET",
                                "https://10.79.169.121:9200/_cluster/settings?pretty"
                            ],
                            "exec": "curl -k -u \"${OPENSEARCH_USER}:{$OPENSEARCH_PWD}\" -X GET https://10.79.169.121:9200/_cluster/settings?pretty",
                            "filepath": "sos_commands/charmed_opensearch/settings3",
                            "truncated": false,
                            "return_code": 0,
                            "priority": 10,
                            "start_time": 1770031845.6830783,
                            "end_time": 1770031845.865656,
                            "run_time": 0.18257761001586914,
                            "tags": []
                        }
```

pmoravec: And `ps` output from the time when the `curl` command is running?

I think when using `my_env` (resp. when calling even from cmdline `OPENSEARCH_USER=... OPENSEARCH_PWD=.. curl ..`), then no need to provide `curl -u` option - I dont know what is the precedence there.

One typo spotted that can affect your tests: `settings3` has "${{OPENSEARCH_USER}}:{{$OPENSEARCH_PWD}}" so password isnt expanded..?

Pepsiqqq: How can I capture 'ps' of curl command running?

We pass the credentials to curl in this way, as it will not use the environment variables automatically.

If you try the same thing in a shell, the result will be empty strings:
```
ubuntu@juju-5164ee-14:~$ OPENSEARCH_PWD=test1 OPENSEARCH_USER=test2 echo "curl -k -u $OPENSEARCH_USER:$OPENSEARCH_PWD -X GET https://10.79.169.121:9200/_cluster/settings?pretty"
curl -k -u : -X GET https://10.79.169.121:9200/_cluster/settings?pretty
```
I fixed the typo but result is same

sos/report/plugins/charmed_opensearch_dashboards.py
```diff
@@ -0,0 +1,143 @@
...
+            temp_env["OPENSEARCH_USER"] = self.get_option('user')
+
+        query = ("type=dashboard&"
+                 "type=visualization&"
+                 "type=index-pattern&"
+                 "type=search&per_page=1&"
+                 "fields=id")
+
+        self.add_cmd_output(
+            f"sh -c 'curl -k -u $OPENSEARCH_USER:$OPENSEARCH_PWD "
```
arif-ali, line 120: same here again base on the other PR, do we really need the `sh -c`

arif-ali (approved):

TurboTurtle (approved):

PR #4200, opened 2026-01-08
[archive] Add option for tar-only archive

jcastill:
Add new option --compression-type=none to
create a tarfile only, and avoid compressing the file.

Related: RHEL-139389, RHEL-89615

---

### DIFF
--- a/sos/__init__.py
+++ b/sos/__init__.py
@@ -148,7 +148,8 @@ class SoS():
 
         global_grp.add_argument('-z', '--compression-type',
                                 dest="compression_type",
-                                choices=['auto', 'gzip', 'xz'],
+                                choices=['auto', 'gzip', 'xz',
+                                         'none'],
                                 help="compression technology to use")
 
         # Group to make tarball encryption (via GPG/password) exclusive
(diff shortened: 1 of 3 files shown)


sos/__init__.py
```diff
@@ -148,7 +148,8 @@ def _add_common_options(self, parser):
 
         global_grp.add_argument('-z', '--compression-type',
                                 dest="compression_type",
-                                choices=['auto', 'gzip', 'xz'],
+                                choices=['auto', 'gzip', 'xz',
+                                         'none'],
```
jcastill, line 152: is 'none' clear enough? I thought about writing 'taronly' as option but I don't think it would be clearer

bmr-cymru: I like `none` here. I think it's clear and concise. You could maybe add a short passage to the man page description along the lines of "if the value \fBnone\fP is given the tar archive will be written directly with no compression".

pmoravec (approved):
Nice and concise!

+1 to add "none means tar" comment to man pages.

jcastill:
Done, and I've fixed a small typo in the man page as well

PR #4202, opened 2026-01-09
[report] Incrase column size for plugin list output

jcastill:
The current column size of 20 characters in the output of 'sos report --list-plugins' for 'active' plugins may not be enough in some cases. This commit increases the width to 30. The corresponding avocado test is modified as well to reflect this change.
The same happens with the column size in the
'plugin options' section, so it is increased to
40 characters, with the corresponding avocado test increasing as well.

Related: #4199

---


bmr-cymru:
If you use the `snapm`/`boom` report module it does all that & more for free - like JSON & shell friendly output, user specified columns etc.  ;-)  (just saying - it may not be a good fit here, but I've had it in mind for a while to look at splitting it out into a reusable library. I'd be happy to help with the integration if there was interest in pursuing this).

PR #4203, opened 2026-01-12
[coredump] add option to save crashing executable

tcasett:
Setting `coredump.save_executable` to a boolean True value will add a copy spec for the crashing processes executable.

Those crashing process executables will not always be available locally. This adds value for developers who are collecting and analyzing remote crash reports and need to import them into their debuggers.

Related: RHEL-139389, RHEL-89615

---

n/a
no changes

### DIFF
--- a/sos/report/plugins/coredump.py
+++ b/sos/report/plugins/coredump.py
@@ -49,6 +49,9 @@ class Coredump(Plugin, IndependentPlugin):
 
     option_list = [
         PluginOpt("dumps", default=3, desc="number of dump files to collect"),
+        PluginOpt("save_executable", default=False,
+                  desc="Add the crashed executable file in the sos report, "
+                  "useful when loading cores into debuggers"),
         PluginOpt("executable", default='',
                   desc=("only collect info and dump output for executables "
                         "matching this regex"))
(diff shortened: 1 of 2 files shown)


pmoravec (approved):

pmoravec:
> The copy spec call was bumping into the size limit option again. Tested adding an exe_size parameter to bypass the global log limit option. Worked out well. Will update PR w/ new changes

You can also use `sizelimit=0` to copy whole binary, see https://github.com/sosreport/sos/blob/main/sos/report/plugins/__init__.py#L1672 . But it is questionable if you really want to copy a binary of arbitrary big size..? (I guess one is interested in the parts that contain references from backtrace and/or objects definitions, but that can be present in arbitrary places in the binary, not only at its beginning, right?)

tcasett:
@pmoravec after much internal testing, and some forgetting to come back to this, I have updated this PR and it's ready for your review. We've captured literally hundreds of cores with this plugin across our infrastructure.

TurboTurtle (commented):
Ack to the plugin code.

On the plugin unittest, I have a couple comments:

1. Holy moly, this is a superbly annotated and thought out test case.
2. But...it does not run as part of our testing suite.

Our test suite inherits from the base sos test classes within `tests/sos_tests` which is what the entire implementation is built around. I know you have the notes in L50-57, but the point is the project uses avocado, and that's what we need the tests to leverage. Whatever anyone else does outside of this project is not something we build for.

I'd like to see the test refactored such that it runs as part of our avocado-scoped testing CI setup.

tcasett:
> Ack to the plugin code.
> 
> On the plugin unittest, I have a couple comments:
> 
>     1. Holy moly, this is a superbly annotated and thought out test case.
> 
>     2. But...it does not run as part of our testing suite.
> 
> 
> Our test suite inherits from the base sos test classes within `tests/sos_tests` which is what the entire implementation is built around. I know you have the notes in L50-57, but the point is the project uses avocado, and that's what we need the tests to leverage. Whatever anyone else does outside of this project is not something we build for.
> 
> I'd like to see the test refactored such that it runs as part of our avocado-scoped testing CI setup.

Righty! I'll see if I can switch it back. I had some import error that made me pivot back to this direction.

PR #4213, opened 2026-01-23
[aap_containerized] Carry forward postproc from other AAP plugins

pmoravec:
Secrets obfuscations from 2a46e99 commit must be reflected in containerized plugin.

Further, fix a typo in a regexp, to properly obfuscate:

```
EMAIL_HOST_PASSWORD = 'FAKESECRET!!!'
```

in (both) controller's settings.

Closes: #4213

---

### DIFF
--- a/sos/report/plugins/aap_containerized.py
+++ b/sos/report/plugins/aap_containerized.py
@@ -41,6 +41,7 @@ class AAPContainerized(Plugin, RedHatPlugin):
     def setup(self):
         # Check if username is passed as argument
         username = self.get_option("username")
+        self.aap_directory_name = self.get_option("directory")
         if not username:
             self._log_warn("AAP username is missing, use '-k "
                            "aap_containerized.username=<user>' to set it")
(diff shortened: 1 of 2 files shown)


sos/report/plugins/aap_containerized.py
```diff
@@ -63,14 +63,15 @@ def setup(self):
         # Grab aap installation directory under user's home
         if not self.get_option("directory"):
             user_home_directory = os.path.expanduser(f"~{username}")
-            aap_directory_name = self.path_join(user_home_directory, "aap")
+            self.aap_directory_name = self.path_join(user_home_directory,
+                                                     "aap")
```
jcastill, line 68: I think you need to define aap_directory_name before setup, perhaps around line 40, to avoid the call trace you get in the avocado tests. Is there some default value you could use there?

pmoravec: Interesting failure.

The directory has no default value, since it is dependant on name of user that installed AAP (it is e.g. `/home/aap/aap` or `/home/ansible/aap`). I will try `None` as default.

pmoravec:
Hi @TurboTurtle and @arif-ali , could you please review this such that we can merge it? It is in fact a regression that sos report newly collects some passwords.

Thanks.

sos/report/plugins/aap_containerized.py
```diff
@@ -200,6 +200,24 @@ def check_enabled(self):
         return False
 
     def postproc(self):
+        # remove controller email password
+        file_path = f"{self.aap_directory_name}/controller/etc/settings.py"
+        jreg = r"(EMAIL_HOST_PASSWORD\s*=\s*)\'(.+)\'"
```
TurboTurtle, line 205: Should we make this a little more generic? Are there other `*_PASSWORD` fields that could exist?

pmoravec: Nope, no other password. Still I would be hesitant to modify it to potentially cause some harm to either side. So I leave it as is.

PR #4214, opened 2026-01-26
[charmed_kafka] feat: add Charmed Kafka plugins

marcoppenheimer:
---

### DIFF
--- a/sos/report/plugins/charmed_cruise_control.py
+++ b/sos/report/plugins/charmed_cruise_control.py
@@ -1,0 +1,141 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.#
+
+import glob
+import re
+from datetime import datetime
+from functools import cached_property
+
+from sos.report.plugins import Plugin, UbuntuPlugin
+
+PATHS = {
+    "CONF": "/var/snap/charmed-kafka/current/etc/cruise-control",
+    "LOGS": "/var/snap/charmed-kafka/common/var/log/cruise-control",
+}
+
+DATE_FORMAT = "%Y-%m-%d-%H"
+
+
+class CharmedCruiseControl(Plugin, UbuntuPlugin):
+    short_desc = "Cruise Control (from Charmed Kafka)"
+    plugin_name = "charmed_cruise_control"
+    packages = ("charmed-kafka",)
+
+    @cached_property
+    def credentials_args(self) -> str:
+        try:
+            with open(
+                f"{PATHS['CONF']}/cruisecontrol.credentials",
+                encoding="utf-8",
+            ) as f:
+                content = f.read().strip()
+
+            if match := re.match(r"balancer: (?P<pwd>\w+),ADMIN", content):
+                pwd = match.group("pwd")
+                return f"-u balancer:{pwd}"
+
+            return ""
+        except FileNotFoundError:
+            return ""
+
+    def setup(self):
+        if not self.credentials_args:
+            # service not properly set-up, skip
+            return
+
+        # --- FILE EXCLUSIONS ---
+
+        all_logs = self.get_option("all_logs")
+        since = self.get_option("since")
+
+        for file in glob.glob(f"{PATHS['LOGS']}/*"):
+            date = re.search(
+                pattern=r"([0-9]{4}-[0-9]{2}-[0-9]{2}-[0-9]{2})", string=file
+            )
+
+            # include files without date, aka current files
+            if not date:
+                continue
+
+            file_dt = datetime.strptime(date.group(1), DATE_FORMAT)
+
+            if (
+                since
+                and not all_logs
+                and file_dt < datetime.strptime(str(since), DATE_FORMAT)
+            ):
+                # skip files outside given range
+                self.add_forbidden_path(file)
+
+        # hide keys/stores
+        self.add_forbidden_path([
+                f"{PATHS['CONF']}/*.pem",
+                f"{PATHS['CONF']}/*.key",
+                f"{PATHS['CONF']}/*.p12",
+                f"{PATHS['CONF']}/*.jks",
+        ])
+
+        # --- FILE INCLUSIONS ---
+
+        self.add_copy_spec(
+            [
+                f"{PATHS['CONF']}",
+                f"{PATHS['LOGS']}",
+            ]
+        )
+
+        # --- SNAP LOGS ---
+
+        if all_logs:
+            self.add_cmd_output(
+                "snap logs -n all charmed-kafka.cruise-control",
+                suggest_filename="snap_logs_charmed-kafka_cruise-control",
+            )
+        else:
+            self.add_cmd_output(
+                "snap logs -n 500 charmed-kafka.cruise-control",
+                suggest_filename="snap_logs_charmed-kafka_cruise-control",
+            )
+
+        # --- STATE, TASKS, PARTITION LOAD ---
+
+        endpoints = {
+            'cruise-control-state': 'state?super_verbose=true',
+            'cluster-state': 'kafka_cluster_state?verbose=true',
+            'partition_load': 'partition_load',
+            'user-tasks': 'user_tasks',
+        }
+
+        url = 'localhost:9090/kafkacruisecontrol'
+
+        for fname, api in endpoints.items():
+            self.add_cmd_output(
+                f"curl {self.credentials_args} {url}/{api}",
+                suggest_filename=fname,
+            )
+
+        # --- JMX METRICS ---
+
+        self.add_cmd_output(
+            "curl localhost:9102/metrics",
+            suggest_filename="jmx-metrics"
+        )
+
+    def postproc(self):
+        if not self.credentials_args:
+            # service not properly set-up, skip
+            return
+
+        # --- SCRUB PASSWORDS ---
+
+        for scrub_pattern in [r'(password=")[^"]*', r"(balancer: )[^,]*"]:
+            self.do_path_regex_sub(
+                f"{PATHS['CONF']}/*",
+                scrub_pattern,
+                r"\1*********",
+            )
--- a/sos/report/plugins/charmed_kafka.py
+++ b/sos/report/plugins/charmed_kafka.py
@@ -1,0 +1,222 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.#
+
+import glob
+import json
+import re
+from datetime import datetime
+from functools import cached_property
+from typing import Optional
+
+from sos.report.plugins import Plugin, UbuntuPlugin
+
+PATHS = {
+    "CONF": "/var/snap/charmed-kafka/current/etc/kafka",
+    "LOGS": "/var/snap/charmed-kafka/common/var/log/kafka",
+}
+
+DATE_FORMAT = "%Y-%m-%d-%H"
+
+
+class CharmedKafka(Plugin, UbuntuPlugin):
+    short_desc = "Charmed Kafka"
+    plugin_name = "charmed_kafka"
+    packages = ("charmed-kafka",)
+
+    @cached_property
+    def bootstrap_server(self) -> Optional[str]:
+        try:
+            lines = []
+            with open(
+                    f"{PATHS['CONF']}/client.properties",
+                    encoding="utf-8",
+            ) as f:
+                lines = f.readlines()
+
+            for line in lines:
+                if "bootstrap" in line:
+                    # ensure using internal address if
+                    # not set in client.properties
+                    return re.sub(
+                        r":(?!1)(\d+)", r":1\1",
+                        line.split("=")[1]
+                    )
+
+            return None
+        except FileNotFoundError:
+            return None
+
+    @cached_property
+    def default_bin_args(self) -> str:
+        if not self.bootstrap_server:
+            return ""
+
+        return (
+            f"--bootstrap-server {self.bootstrap_server}"
+            f" --command-config {PATHS['CONF']}/client.properties"
+        )
+
+    def setup(self):
+        if not self.bootstrap_server:
+            # service not properly set-up by the charm, skip
+            return
+
+        # --- FILE EXCLUSIONS ---
+
+        all_logs = self.get_option("all_logs")
+        since = self.get_option("since")
+
+        for file in glob.glob(f"{PATHS['LOGS']}/*"):
+            date = re.search(
+                pattern=r"([0-9]{4}-[0-9]{2}-[0-9]{2}-[0-9]{2})", string=file
+            )
+
+            # include files without date, aka current files
+            if not date:
+                continue
+
+            file_dt = datetime.strptime(date.group(1), DATE_FORMAT)
+
+            if (
+                since
+                and not all_logs
+                and file_dt < datetime.strptime(str(since), DATE_FORMAT)
+            ):
+                # skip files outside given range
+                self.add_forbidden_path(file)
+
+        # hide keys/stores
+        self.add_forbidden_path([
+                f"{PATHS['CONF']}/*.pem",
+                f"{PATHS['CONF']}/*.key",
+                f"{PATHS['CONF']}/*.p12",
+                f"{PATHS['CONF']}/*.jks",
+        ])
+
+        # --- FILE INCLUSIONS ---
+
+        self.add_copy_spec(
+            [
+                f"{PATHS['CONF']}",
+                f"{PATHS['LOGS']}",
+            ]
+        )
+
+        # --- SNAP LOGS ---
+
+        if all_logs:
+            self.add_cmd_output(
+                "snap logs -n all charmed-kafka.daemon",
+                suggest_filename="snap_logs_charmed-kafka_daemon",
+            )
+        else:
+            self.add_cmd_output(
+                "snap logs -n 500 all charmed-kafka.daemon",
+                suggest_filename="snap_logs_charmed-kafka_daemon",
+            )
+
+        # --- TOPICS ---
+
+        self.add_cmd_output(
+            f"charmed-kafka.topics --describe {self.default_bin_args}",
+            env={"KAFKA_OPTS": ""},
+            suggest_filename="kafka-topics",
+        )
+
+        # --- CONFIGS ---
+
+        for entity in ["topics", "clients", "users", "brokers", "ips"]:
+            self.add_cmd_output(
+                (
+                    "charmed-kafka.configs --describe --all "
+                    f"--entity-type {entity} {self.default_bin_args}"
+                ),
+                env={"KAFKA_OPTS": ""},
+                suggest_filename=f"kafka-configs-{entity}",
+            )
+
+        # --- ACLs ---
+
+        self.add_cmd_output(
+            f"charmed-kafka.acls --list {self.default_bin_args}",
+            env={"KAFKA_OPTS": ""},
+            suggest_filename="kafka-acls",
+        )
+
+        # --- JMX METRICS ---
+
+        self.add_cmd_output(
+            "curl localhost:9101/metrics",
+            suggest_filename="jmx-metrics"
+        )
+
+    def collect(self):
+
+        # --- LOG DIRS ---
+
+        log_dirs_output = self.exec_cmd(
+            f"charmed-kafka.log-dirs --describe {self.default_bin_args}",
+            env={"KAFKA_OPTS": ""},
+        )
+        log_dirs = {}
+
+        # output has leading non-json lines that need cleaning, e.g:
+
+        # Querying brokers for log directories information
+        # Received log directory information from brokers 100,101,102
+        # {"brokers":[{"broker":100,"logDirs":...
+
+        if log_dirs_output and log_dirs_output["status"] == 0:
+            for line in log_dirs_output["output"].splitlines():
+                try:
+                    log_dirs = json.loads(line)
+                    break
+                except json.JSONDecodeError:
+                    continue
+
+        with self.collection_file("kafka-log-dirs") as f:
+            f.write(json.dumps(log_dirs, indent=4))
+
+        # --- TRANSACTIONS ---
+
+        transactions_list = self.exec_cmd(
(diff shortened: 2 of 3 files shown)


sos/report/plugins/charmed_cruise_control.py
```diff
@@ -0,0 +1,141 @@
...
+            ]
+        )
+
+        # --- FILE INCLUSIONS ---
+
+        self.add_copy_spec(
+            [
+                f"{PATHS['CONF']}",
+                f"{PATHS['LOGS']}",
+                "/var/log/juju",
```
arif-ali: this is already collected by the juju plugin, so this line can be removed

sos/report/plugins/charmed_cruise_control.py
```diff
@@ -0,0 +1,141 @@
...
+            ]
+        )
+
+        # --- SNAP LOGS ---
+
+        self.add_cmd_output(
+            "snap logs charmed-kafka.cruise-control -n 100000",
+            suggest_filename="snap-logs",
+        )
```
arif-ali: would it be worth using the all_logs option to get all logs, and then have a default of 10000 lines?

sos/report/plugins/charmed_kafka.py
```diff
@@ -0,0 +1,196 @@
...
+            ]
+        )
+
+        # --- FILE INCLUSIONS ---
+
+        self.add_copy_spec(
+            [
+                f"{PATHS['CONF']}",
+                f"{PATHS['LOGS']}",
+                "/var/log/juju",
```
arif-ali: same again here, /var/log/juju is collected by the juju plugin, so can be removed here

sos/report/plugins/charmed_kafka.py
```diff
@@ -0,0 +1,196 @@
...
+            ]
+        )
+
+        # --- SNAP LOGS ---
+
+        self.add_cmd_output(
+            "snap logs charmed-kafka.daemon -n 100000",
+            suggest_filename="snap-logs",
+        )
```
arif-ali: same comment here around all_logs, and maybe have a conditional here for that. There are examples of this in other plugins if you wanted to look

sos/report/plugins/charmed_zookeeper.py
```diff
@@ -0,0 +1,198 @@
...
+            ]
+        )
+
+        # --- FILE INCLUSIONS ---
+
+        self.add_copy_spec(
+            [
+                f"{PATHS['CONF']}",
+                f"{PATHS['LOGS']}",
+                "/var/log/juju",
```
arif-ali: same as above the juju logs being in the juju plugin

sos/report/plugins/charmed_zookeeper.py
```diff
@@ -0,0 +1,198 @@
...
+            ]
+        )
+
+        # --- SNAP LOGS ---
+
+        self.add_cmd_output(
+            "snap logs charmed-zookeeper.daemon -n 100000",
+            suggest_filename="snap-logs",
+        )
```
arif-ali, line 124: same as above on all_logs option

sos/report/plugins/charmed_cruise_control.py
```diff
@@ -0,0 +1,122 @@
+import glob
```
TurboTurtle, line 9: Please add the sos header to the top of these plugins. You can see the header in any existing plugin. The copyright line is optional and can either be you, your company, or none at all.

sos/report/plugins/charmed_cruise_control.py
```diff
@@ -0,0 +1,122 @@
...
+            f"{PATHS['CONF']}/cruisecontrol.credentials",
+            encoding="utf-8",
+        ) as f:
+            content = f.read().strip()
+
+        if match := re.match(r"balancer: (?P<pwd>\w+),ADMIN", content):
+            pwd = match.group("pwd")
+            return f"-u balancer:{pwd}"
+
+        return ""
```
TurboTurtle: This would open and read the file every time the property is called, which is wasteful as it is highly unlikely the credentials are going to change during execution.

sos/report/plugins/charmed_cruise_control.py
```diff
@@ -0,0 +1,122 @@
...
+                and not all_logs
+                and file_dt < datetime.strptime(str(since), DATE_FORMAT)
+            ):
+                # skip files outside given range
+                self.add_forbidden_path(file)
+
+        # hide keys/stores
+        self.add_forbidden_path(
+            [
+                f"{PATHS['CONF']}/*.pem",
```
TurboTurtle: sos-style is to put the opening brackets on the same line as the parens, and then indent only one level, e.g.:

```python
self.add_forbidden_path([
    f"{PATHS['CONF']}/*.pem",
    f"{PATHS['CONF']}/*.key",
    [...]
])
```

sos/report/plugins/charmed_cruise_control.py
```diff
@@ -0,0 +1,122 @@
...
+                f"{PATHS['CONF']}/*.jks",
+            ]
+        )
+
+        # --- SNAP LOGS ---
+
+        if all_logs:
+            self.add_cmd_output(
+                "snap logs charmed-kafka.cruise-control",
+                suggest_filename="snap-logs",
```
TurboTurtle: `suggest_filename` is meant to make human-friendly versions of long/rambling/weird/ugly command strings - otherwise the preference is that command collections inform the end user of what was executed to generate the output. In these cases, the full command string is preferable.

sos/report/plugins/charmed_cruise_control.py
```diff
@@ -0,0 +1,122 @@
...
+            )
+        else:
+            self.add_cmd_output(
+                "snap logs charmed-kafka.cruise-control -n 10000",
+                suggest_filename="snap-logs",
+            )
+
+        # --- CRUISE CONTROL STATE ---
+
+        self.add_cmd_output(
```
TurboTurtle: These can all be cleaned up, and made to fit line lengths:

```python
endpoints = {
    'cruise-control-state': 'state?super_verbose=true',
    'cluster-state': 'kafka_cluster_state?verbose=true',
    'partition_load': 'partition_load',
    ...etc...
}

url = 'localhost:9090/kafkacruisecontrol'

for fname, api in endpoints.items():
    self.add_cmd_output(
        f"curl {self.credentials_args} {url}/{api}",
        suggest_filename=fname
    )
```

sos/report/plugins/charmed_kafka.py
```diff
@@ -0,0 +1,177 @@
...
+DATE_FORMAT = "%Y-%m-%d-%H"
+
+
+class CharmedKafka(Plugin, UbuntuPlugin):
+    short_desc = "Charmed Kafka"
+    plugin_name = "charmed_kafka"
+    packages = ("charmed-kafka",)
+
+    @property
+    def bootstrap_server(self) -> Optional[str]:
```
TurboTurtle: Same thing here. These are very unlikely to change during our execution, so we should be opening the file, parsing it, and saving the value once, instead of doing the whole operation every time.

sos/report/plugins/charmed_kafka.py
```diff
@@ -0,0 +1,177 @@
...
+
+        for line in lines:
+            if "bootstrap" in line:
+                # ensure using internal address if not set in client.properties
+                return re.sub(r":(?!1)(\d+)", r":1\1", line.split("=")[1])
+
+        return None
+
+    @property
+    def default_bin_args(self) -> str:
```
TurboTurtle: Ditto.

sos/report/plugins/charmed_kafka.py
```diff
@@ -0,0 +1,177 @@
...
+                since
+                and not all_logs
+                and file_dt < datetime.strptime(str(since), DATE_FORMAT)
+            ):
+                # skip files outside given range
+                self.add_forbidden_path(file)
+
+        # hide keys/stores
+        self.add_forbidden_path(
+            [
```
TurboTurtle: Same sos-style note as above.

sos/report/plugins/charmed_kafka.py
```diff
@@ -0,0 +1,177 @@
...
+        if log_dirs_output and log_dirs_output["status"] == 0:
+            for line in log_dirs_output["output"].splitlines():
+                try:
+                    log_dirs = json.loads(line)
+                    break
+                except json.JSONDecodeError:
+                    continue
+
+        with self.collection_file("kafka-log-dirs") as f:
+            f.write(json.dumps(log_dirs, indent=4))
```
TurboTurtle, line 183: Can you show an example of the output here that makes us want to do this instead of just collecting the whole `log-dirs` command output?

sos/report/plugins/charmed_kafka.py
```diff
@@ -0,0 +1,177 @@
...
+                    f"charmed-kafka.transactions {self.default_bin_args}",
+                    f"describe --transactional-id {transactional_id}",
+                ),
+            )
+
+            if transactions_describe and transactions_describe["status"] == 0:
+                transactions_outputs.append(transactions_describe["output"])
+
+        with self.collection_file("kafka-transactions") as f:
+            f.write("\n".join(transactions_outputs))
```
TurboTurtle, line 209: For this and likewise above, `collection_file()` collections are typically done as part of the `collect()` phase, not `setup()`. Reason being that `collect()` has more guardrails in place than `setup()` does.

sos/report/plugins/charmed_zookeeper.py
```diff
@@ -0,0 +1,179 @@
...
+TEMP_JAAS = "client.jaas"
+
+
+class CharmedZooKeeper(Plugin, UbuntuPlugin):
+    short_desc = "Charmed ZooKeeper"
+    plugin_name = "charmed_zookeeper"
+    packages = ("charmed-zookeeper",)
+
+    @property
+    def default_env(self) -> dict[str, str]:
```
TurboTurtle: There's no need for this to be a property.

sos/report/plugins/charmed_zookeeper.py
```diff
@@ -0,0 +1,179 @@
...
+        return {
+            "JAVA_HOME": PATHS["JRE"],
+            "CLIENT_JVMFLAGS": f"-Djava.security.auth.login.config={PATHS['CONF']}/{TEMP_JAAS}",  # noqa: E501 # pylint:disable=line-too-long
+            "SERVER_JVMFLAGS": "",
+        }
+
+    @property
+    def super_password(self) -> str:
```
TurboTurtle: Same as before; either cache the value somehow or set it once since this is unlikely to change during execution, so we don't want to be continually opening a file, reading it in, and then parsing it again, for the same info every time.

sos/report/plugins/charmed_zookeeper.py
```diff
@@ -0,0 +1,179 @@
...
+        with open(f"{PATHS['CONF']}/{TEMP_JAAS}", "w", encoding="utf-8") as f:
+            f.write(
+                f"""
+            Client {{
+                org.apache.zookeeper.server.auth.DigestLoginModule required
+                username="super"
+                password="{self.super_password}";
+            }};
+                """
+            )
```
TurboTurtle: Absolutely not. We don't make changes to the host system during an `sos` execution, and we definitely don't want to be writing passwords out. From a security point of view, what happens when someone starts an `sos report` run, then kills the process after this has run but before the `postproc()` fires for example?

sos/report/plugins/charmed_zookeeper.py
```diff
@@ -0,0 +1,179 @@
...
+                since
+                and not all_logs
+                and file_dt < datetime.strptime(str(since), DATE_FORMAT)
+            ):
+                # skip files outside given range
+                self.add_forbidden_path(file)
+
+        # hide keys/stores
+        self.add_forbidden_path(
+            [
```
TurboTurtle: Same sos-style note as in the other two plugins.

sos/report/plugins/charmed_zookeeper.py
```diff
@@ -0,0 +1,179 @@
...
+                f"{PATHS['CONF']}/{TEMP_JAAS}",
+            ]
+        )
+
+        # --- SNAP LOGS ---
+
+        if all_logs:
+            self.add_cmd_output(
+                "snap logs charmed-zookeeper.daemon",
+                suggest_filename="snap-logs",
```
TurboTurtle: Same filename note as the other two plugins.

sos/report/plugins/charmed_zookeeper.py
```diff
@@ -0,0 +1,179 @@
...
+                    str(self.get_option("date-to")), DATE_FORMAT
+                ):
+                    continue
+
+                valid_dt_transactions.append(transaction)
+
+            with self.collection_file(
+                f"zookeeper-transaction-{os.path.basename(file)}"
+            ) as f:
+                f.write("\n".join(valid_dt_transactions))
```
TurboTurtle, line 169: Same as the other plugin - doing this sort of thing and writing it out to a file in the archive is fine, but do it during `collect()` and not `setup()`

sos/report/plugins/charmed_cruise_control.py
marcoppenheimer: I don't see issue with the verbosity here, as having each step explicit helps with potential future post-processing for individual steps if needed.

sos/report/plugins/charmed_zookeeper.py
marcoppenheimer: Fixed, was a mistake anyway left in during dev. File is no longer created.

sos/report/plugins/charmed_zookeeper.py
```diff
@@ -0,0 +1,177 @@
...
+
+        # --- SNAPSHOT ---
+        files = glob.glob(f"{PATHS['DATA']}/version-2/*")
+        most_recent_file = sorted(files)[-1]
```
pmoravec: `IndexError: list index out of range` raised when `files` is empty.

pmoravec:
Here are the backtraces that caused the `AllPluginSmokeTest.test_no_exceptions_during_execution` test to fail:

```
  File "/tmp/cirrus-ci-build/sos/report/__init__.py", line 1270, in setup
    plug.setup()
  File "/tmp/cirrus-ci-build/sos/report/plugins/charmed_cruise_control.py", line 92, in setup
    f"curl {self.credentials_args} localhost:9090/kafkacruisecontrol/state?super_verbose=true",  # noqa: E501 # pylint:disable=line-too-long
  File "/usr/lib/python3.10/functools.py", line 981, in __get__
    val = self.func(instance)
  File "/tmp/cirrus-ci-build/sos/report/plugins/charmed_cruise_control.py", line 31, in credentials_args
    with open(
FileNotFoundError: [Errno 2] No such file or directory: '/var/snap/charmed-kafka/current/etc/cruise-control/cruisecontrol.credentials'
```

```
  File "/tmp/cirrus-ci-build/sos/report/__init__.py", line 1270, in setup
    plug.setup()
  File "/tmp/cirrus-ci-build/sos/report/plugins/charmed_kafka.py", line 103, in setup
    f"charmed-kafka.topics --describe {self.default_bin_args}",
  File "/usr/lib/python3.10/functools.py", line 981, in __get__
    val = self.func(instance)
  File "/tmp/cirrus-ci-build/sos/report/plugins/charmed_kafka.py", line 46, in default_bin_args
    if not self.bootstrap_server:
  File "/usr/lib/python3.10/functools.py", line 981, in __get__
    val = self.func(instance)
  File "/tmp/cirrus-ci-build/sos/report/plugins/charmed_kafka.py", line 34, in bootstrap_server
    with open(f"{PATHS['CONF']}/client.properties", encoding="utf-8") as f:
FileNotFoundError: [Errno 2] No such file or directory: '/var/snap/charmed-kafka/current/etc/kafka/client.properties'
```


Plus also:

```
[plugin:charmed_zookeeper] Error during plugin collections: list index out of range
```

that I commented directly in the code.

marcoppenheimer:
@pmoravec - Thank you! CI fixed I think, we good to go?

sos/report/plugins/charmed_cruise_control.py
```diff
@@ -0,0 +1,140 @@
...
+
+            file_dt = datetime.strptime(date.group(1), DATE_FORMAT)
+
+            if (
+                since
+                and not all_logs
+                and file_dt < datetime.strptime(str(since), DATE_FORMAT)
+            ):
+                # skip files outside given range
+                self.add_forbidden_path(file)
```
pmoravec, line 73: Doesn't `--since` option skip those files natively?

Well, there is a slight difference. `add_copy_path` would add files with `mtime` newer than "since" value, while filename can contain timestamp from another day. So if you want to be such precise, why not!

(same applies to other plugins as well)

sos/report/plugins/charmed_cruise_control.py
```diff
@@ -0,0 +1,140 @@
...
+
+        # hide keys/stores
+        self.add_forbidden_path([
```
pmoravec, line 76: You have several `add_forbidden_path` but no `add_copy_spec`. How do you copy files for all the three plugins?

sos/report/plugins/charmed_cruise_control.py
```diff
@@ -0,0 +1,140 @@
...
+        if all_logs:
+            self.add_cmd_output(
+                "snap logs charmed-kafka.cruise-control",
+                suggest_filename="snap_logs_charmed-kafka_cruise-control",
+            )
+        else:
+            self.add_cmd_output(
+                "snap logs charmed-kafka.cruise-control -n 10000",
+                suggest_filename="snap_logs_charmed-kafka_cruise-control",
+            )
```
pmoravec, line 103: Two points:

1) by default, `add_cmd_output` collects up to `--log-size` output (25MB by default). This limit does not suit you well and prefer the 10k lines, regardless of `--log-size`? Are you ok to collect exactly given number of logs, hardcoded?

2) Can't you shorten it to something like:

```
lines = "" if all_logs else " -n 10000"
self.add_cmd_output(f"snap logs charmed-kafka.cruise-control{lines}",
                                       suggest_filename="snap_logs_charmed-kafka_cruise-control")
```

(same applies to other plugins as well)

sos/report/plugins/charmed_cruise_control.py
marcoppenheimer: Good catch, got dropped by accident during review. Have added back.

sos/report/plugins/charmed_cruise_control.py
marcoppenheimer: I didn't know about the log-size, that's a better option, have went with it and dropped the `lines` part! Thanks.

sos/report/plugins/charmed_zookeeper.py
arif-ali: `snap logs` by default only captures 10 lines, is it intended that you only want 10 lines

```
 snap logs --help
Usage:
  snap logs [logs-OPTIONS] <service>...

The logs command fetches logs of the given services and displays them in
chronological order.

[logs command options]
          --abs-time   Display absolute times (in RFC 3339 format). Otherwise, display relative times up to 60 days, then YYYY-MM-DD.
      -n=              Show only the given number of lines, or 'all'. (default: 10)
      -f               Wait for new lines and print them as they come in.

[logs command arguments]
  <service>:           A service specification, which can be just a snap name (for all services in the snap), or <snap>.<app> for a single service.
```

What I meant with all_logs option was that we could do something similar to

```python
if self.getOption('all_logs'):
        self.add_cmd_output(
            "snap logs -n 50000 charmed-zookeeper.daemon",
            suggest_filename="snap_logs_charmed-zookeeper_daemon",
        )
else:
        self.add_cmd_output(
            "snap logs -n 500 charmed-zookeeper.daemon",
            suggest_filename="snap_logs_charmed-zookeeper_daemon",
        )
```

sos/report/plugins/charmed_kafka.py
arif-ali: same comment I made below

sos/report/plugins/charmed_cruise_control.py
arif-ali: again same comment as below

sos/report/plugins/charmed_cruise_control.py
```diff
@@ -0,0 +1,149 @@
...
+            self.add_cmd_output(
+                "snap logs -n 500 charmed-kafka.cruise-control",
+                suggest_filename="snap_logs_charmed-kafka_cruise-control",
+            )
+
+        # --- CRUISE CONTROL STATE ---
+
+        self.add_cmd_output(
+            f"curl {self.credentials_args} localhost:9090/kafkacruisecontrol/state?super_verbose=true",  # noqa: E501 # pylint:disable=line-too-long
+            "cruise-control-state",
```
pmoravec: Nitpick: adding `suggest_filename="cruise-control-state"` helps understanding the parameter. Same applies  for other add_cmd_outputs.

sos/report/plugins/charmed_cruise_control.py
pmoravec: Jake's point was there is an elegant way of keeping the lines under 80 chars. I support his position, though dont want to block the PR on this.

sos/report/plugins/charmed_kafka.py
```diff
@@ -0,0 +1,219 @@
...
+
+        self.add_cmd_output(
+            f"charmed-kafka.acls --list {self.default_bin_args}",
+            env={"KAFKA_OPTS": ""},
+            suggest_filename="kafka-acls",
+        )
+
+        # --- JMX METRICS ---
+
+        self.add_cmd_output("curl localhost:9101/metrics", "jmx-metrics")
```
pmoravec: ditto with `suggest_name` nitpick.

sos/report/plugins/charmed_kafka.py
pmoravec: I dont understand the json-reformatting here either..

pmoravec (commented):
I would appreciate some feedback to the Jake's comments that were silently marked as resolved. Up to that, I am OK with the plugins, thanks!

sos/report/plugins/charmed_kafka.py
marcoppenheimer: It removes the initial two header lines that are non-json, and formats the output with indentations. I've put an example output in a comment above this section.

sos/report/plugins/charmed_cruise_control.py
arif-ali: I will also support Jake's position here, we have many plugins that were done this way to fine tune.

If there is a need for post-processing in the future, I am sure an elegant way could be found

marcoppenheimer: No problem, strong consensus. Have updated with the recommendation provided :)

arif-ali (approved):
Thanks for the update, all good from me now

PR #4228, opened 2026-02-09
[cleaner] Better distribution of files among child cleaner processes

pmoravec:
Static split of input files often meant uneven distribution of load among the child processes. That caused just one process being busy at the end for many minutes, which ruins concurrency benefits.

Using multiprocessing.[Process|Queue] allow to:
1) order files per sizes, starting from biggest (with the assumption of
   longest cleanup time)
2) dynamic fetching of new file by a child process, only when it gets
   idle

Relevant: #4227
Closes: #4228

---

### DIFF
--- a/sos/cleaner/__init__.py
+++ b/sos/cleaner/__init__.py
@@ -14,8 +14,8 @@ import logging
 import os
 import shutil
 import fnmatch
+import multiprocessing
 
-from concurrent.futures import ProcessPoolExecutor
 from datetime import datetime
 from pwd import getpwuid
 
@@ -39,8 +39,26 @@ from sos.utilities import (get_human_readable, import_module,
 
 
 # an auxiliary method to kick off child processes over its instances
-def obfuscate_arc_files(arc, flist):
-    return arc.obfuscate_arc_files(flist)
+def _obfuscate_arc_files(arc, input_queue, output_queue):
+    while True:
+        try:
+            file = input_queue.get()
+        except (EOFError, OSError) as e:
+            print(f"Child process exception when reading input queue: '{e}'")
+            break
+        if file is None:  # Sentinel value to stop the process
+            try:
+                output_queue.put((
+                    arc.files_obfuscated_count,
+                    arc.total_sub_count,
+                    arc.removed_file_count))
+            except OSError as e:
+                print(
+                    f"Child process exception when writing to output queue: "
+                    f"'{e}'"
+                )
+            break
+        arc.obfuscate_arc_file(file)
 
 
 class SoSCleaner(SoSComponent):
@@ -726,6 +744,18 @@ third party.
             :param archive str:      Filepath to the directory or archive
         """
 
+        def _order_files_by_size(file_list, base_path):
+            """ Order files per their sizes, if they are provided relatively
+            to base_path
+            """
+            files_with_sizes = [
+                (f, os.path.getsize(os.path.join(base_path, f)))
+                for f in file_list
+            ]
+            files_sizes_sorted = sorted(files_with_sizes, key=lambda x: x[1],
+                                        reverse=True)
+            return [file for file, _ in files_sizes_sorted]
+
         try:
             arc_md = self.cleaner_md.add_section(archive.archive_name)
             start_time = datetime.now()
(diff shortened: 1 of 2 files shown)


pmoravec:
Performance comparison (on one small testing sosreport, with default 4 workers/jobs):

- original runtime: 45s
- improved runtime: 32s, hence improved by 1/3

For bigger sosreports, I assume smaller, yet evident improvement (I plan to share some stats once my performance tests complete).

sos/cleaner/archives/__init__.py
```diff
@@ -152,86 +152,83 @@ def obfuscate_line(self, line, parsers=None):
...
-                if (not self.keep_binary_files and
-                        self.should_remove_file(rel_name)):
-                    # We reach this case if the option --keep-binary-files
-                    # was not used, and the file is in a list to be removed
+    def obfuscate_arc_file(self, filename):
+        self.log_debug(f"    pid={os.getpid()}: obfuscating {filename}")
+        try:
+            rel_name = os.path.relpath(filename, start=self.extracted_path)
+            if self.should_skip_file(rel_name):
+                # TODO: add to some/all returns a debug?
```
jcastill: +1 to add debug info, it could be very useful. But perhaps, at least in this case, it could be added directly inside should_skip_file(), where it looks like we don't log the skip either

jcastill (approved):

sos/cleaner/archives/__init__.py
pmoravec: Hmm.. that method is an auxiliary one and those usually dont have logs.. but as it is called just from here and sometimes it can be beneficial to know the reason for the skip ("not a file or symlink, skipping" != "filename in skip list, skipping"), it makes sense.

sos/cleaner/__init__.py
```diff
@@ -735,39 +755,45 @@ def obfuscate_report(self, archive):  # pylint: disable=too-many-branches
...
+                processes.append(p)
+            # Distribute items to the input queue, after reordering them by
+            # size. Since files can be both relative and absolute paths,
+            # depending on input tarball or directory, we must pass the
+            # relative base_path to call os.path.getsize properly
+            for file in _order_files_by_size(
+                list(archive.get_files()),
+                os.path.dirname(os.path.abspath(archive.extracted_path))
+            ):
+                input_queue.put(file)
```
jcastill, line 794: I'm not familiar with the multiprocessing library yet, so I'm sorry if this is just noise.
From the way you are calling both put() and get(), we have the defaults Block=False and Timeout=None, if I'm reading the documentation correctly. These should be enough for the cleaner, but I was wondering what kind of exceptions do these two functions throw, and it seems that with the current values we won't need to catch queue.Full or queue.Empty, or TimeoutError. That said, would still be worth to add a try/except block to cover any other type of exception? And perhaps in the future we want to add blocking or a timeout, and so the exception handing block is already there.

pmoravec: I havent noticed any exception whenever I tested it. Python creates pipes under the hood during `multiprocessing.Process(..)` call, and the child processes communicate over them / wait there for almost immediate items to be added via `input_queue.put(file)` / `input_queue.put(None)`.

Is there some scenario (like parent process killed (I tested this), or some pipe damage? or..?) that can cause an exception? Well I can add a blank try/catch block (for queue.Full/Empty + Timeout?) just in case - does that make sense or it is over-safe?

sos/cleaner/archives/__init__.py
jcastill: Honestly, either way, inside the function or in the 'if's that you have here, I don't have a strong preference. But for sure these messages would be very useful

sos/cleaner/__init__.py
jcastill: if we use OS pipes, perhaps we can at least capture OSError, and then a generic exception that we could refine later on if necessary. But if it's over-safe then I'm happy with what you have at the moment

pmoravec: I understand capturing some (which one in particular?) exceptions can have sense here. I don't want to object that. I just don't know *if* any catch will be an improvement, or *what* particular exception caught are worth of it.

Let me think about it bit more. Or, does @TurboTurtle or @arif-ali has an opinion here..?

pmoravec: I agree no need of queue empty/full or timeout exception handling. What is worth to consider:

- `pickle.PicklingError` - if (de)serialization of the tuple with stats fails - I doubt this can happen..?
- `EOFError` and/or `OSError` when pipes broke, on either side: defensive logging on child side, nice to have; similarly on the parent side when collecting stats
- `TypeError`/`ValueError`/`AttributeError` - if wrong data passed over the pipe, that is hard to imagine, imho not worth catching

So it makes sense to me to log nicely the catching of `OSError` and `EOFError` "only", both in client and parent code.

I will update the code soon.

arif-ali (approved):
I've done some testing, all looks good to me

PR #4238, opened 2026-02-13
[foreman_installer] Collect satellite-clone data

pmoravec:
Collect satellite-clone related data, as a wrapper of foreman/Satellite installation with cloned data.

Closes: #4238

---

### DIFF
--- a/sos/report/plugins/foreman_installer.py
+++ b/sos/report/plugins/foreman_installer.py
@@ -95,6 +95,13 @@ class RedHatForemanInstaller(ForemanInstaller, RedHatPlugin):
                 ['capsule_log', 'capsule_installer_log'],
         })
 
+        self.add_copy_spec([
+            "/usr/share/satellite-clone/logs",
+            "/usr/share/satellite-clone/satellite-clone-vars.yml",
+            "/usr/share/satellite-clone/ansible.production.cfg",
+            "/var/log/satellite-clone/playbook.log",
+        ])
+
         super().setup()
 
 


pmoravec:
Hi @evgeni , could you please review if I haven't forgotten anything, or if the collected data cant contain some sensitive info (passwords or similar)? AFAIK no, but more eyes see more.

sos/report/plugins/foreman_installer.py
```diff
@@ -95,6 +95,13 @@ def setup(self):
                 ['capsule_log', 'capsule_installer_log'],
         })
 
+        self.add_copy_spec([
+            "/usr/share/satellite-clone/logs",
+            "/usr/share/satellite-clone/satellite-clone-vars.yml",
```
evgeni, line 111: this can contain sensitive data

sos/report/plugins/foreman_installer.py
```diff
@@ -95,6 +95,13 @@ def setup(self):
...
+            "/usr/share/satellite-clone/ansible.production.cfg",
+            "/var/log/satellite-clone/playbook.log",
```
evgeni, line 113: this can contain sensitive data

evgeni:
can this be an own collector? it's not really foreman related.

sos/report/plugins/foreman_installer.py
pmoravec: Can you show an example, please?

sos/report/plugins/foreman_installer.py
pmoravec: Ditto.

sos/report/plugins/foreman_installer.py
evgeni: https://github.com/RedHatSatellite/satellite-clone/blob/75531626365bcd9102c2eff00d2a663e844a81cd/satellite-clone-vars.sample.yml#L34-L45

sos/report/plugins/foreman_installer.py
evgeni: output of https://github.com/RedHatSatellite/satellite-clone/blob/75531626365bcd9102c2eff00d2a663e844a81cd/roles/satellite-clone/tasks/main.yml#L40

there might be others

TurboTurtle (approved):
ack to the code.

I personally wouldn't consider activation keys as sensitive data, but I'll defer to Pavel and Evgeni on that.

@pmoravec feel free to merge after deciding the above.

edit:
> I am not against putting this into a separate plugin, I see pros and cons for both solutions. Let see for others feedback (@TurboTurtle shall have some Satellite experience as well?).

Uh, no real opinion here. It'd be a small plugin indeed, and I wonder what the enablement trigger would be such that the same host wouldn't also _already_ trigger this plugin.

arif-ali:
@pmoravec does this require further iteration to obfuscate the secrets?

pmoravec:
> @pmoravec does this require further iteration to obfuscate the secrets?

Thanks for reminding me, I just added the scrubbing there.

PR #4250, opened 2026-02-25
[tests] Adjust cleaner test file content check

dzalezak:
---

### DIFF
--- a/tests/cleaner_tests/basic_function_tests/report_with_mask.py
+++ b/tests/cleaner_tests/basic_function_tests/report_with_mask.py
@@ -115,7 +115,10 @@ class ReportWithUserCustomisations(StageOneReportTest):
         self.assertFileNotHasContent('sos_commands/kernel/uname_-a', 'Linux')
 
     def test_skip_cleaning_single_file(self):
-        self.assertFileHasContent('proc/cmdline', 'BOOT_IMAGE')
+        try:
+            self.assertFileHasContent('proc/cmdline', 'BOOT_IMAGE')
+        except AssertionError:
+            self.assertFileHasContent('proc/cmdline', 'root')
 
     def test_skip_cleaning_glob_file(self):
         self.assertFileHasContent(


tests/cleaner_tests/basic_function_tests/report_with_mask.py
```diff
@@ -115,7 +115,10 @@ def test_keyword_obfuscated_in_file(self):
         self.assertFileNotHasContent('sos_commands/kernel/uname_-a', 'Linux')
 
     def test_skip_cleaning_single_file(self):
-        self.assertFileHasContent('proc/cmdline', 'BOOT_IMAGE')
+        try:
+            self.assertFileHasContent('proc/cmdline', 'BOOT_IMAGE')
+        except AssertionError:
+            self.assertFileHasContent('proc/cmdline', 'root')
```
pmoravec, line 121: I understand this handles the case when `/proc/cmdline` file does contain a line like `BOOT_IMAGE=(hd0,gpt1)/vmlinuz-4.18.0-553.el8_10.x86_64 ..`. Then I would rather suggest:

- a test "if `/proc/cmdline` doesnt contain the string, then ..". Assume a hypothetical situation when the file does contain the `BOOT_IMAGE` string, our code does obfuscate it (as a failure in `--keywords` feature) but our test does not detect it.
- add `root` to the `--keywords` list on line 94-95 to really ensure we do want to obfuscate the string, but not in this file in `--skip-cleaning-files`

Maybe also worth adding a comment to explain the try/except or if-then-else rationale? (or is it too obvious..?)

jcastill: I think adding it to lines 94-95 should be the best approach - in cmdline we can have either BOOT_IMAGE= or root=, depending on different things, so both strings are actually allowed. 
Either that or remove BOOT_IMAGE from these lines

dzalezak: Previously it checked only if there is `BOOT_IMAGE=` in `/proc/cmdline`. But I found case where there was not `BOOT_IMAGE=` in `/proc/cmdline` but only `root=` , so logic behind this was to first try to check if there is `BOOT_IMAGE=` and if not ( except AssertionError ) then check if there is at least `root=` .

dzalezak:
```
$ cat sosreport-host0-2026-02-05-zhiaamo/proc/cmdline 
root=/dev/mapper/rhel_ibm--ic922--02-root ro crashkernel=2G-4G:384M,4G-16G:512M,16G-64G:1G,64G-128G:2G,128G-:4G rd.lvm.lv=rhel_ibm-ic922-02/root rd.lvm.lv=rhel_ibm-ic922-02/swap
```

jcastill: Fair enough @dzalezak . My suggestion was because if both keywords have the same probability of appearing, it may be better to add them both to the list as @pmoravec said. But I defer to your expertise on the testing side

arif-ali: But you can also have both as well, at least that's what I can see on some of my systems, but for example rpis, that only have `root`. Most of the testing we do is on x86 architecture, so do we need this changed at all? We've not had testing issues so far from what I can remember

PR #1198, opened 2018-01-30
Archlinux support

chr0mag:
---
Please place an 'X' inside each '[]' to confirm you adhere to our [Contributor Guidelines](https://github.com/sosreport/sos/wiki/Contribution-Guidelines)

- [X] Is the commit message split over multiple lines and hard-wrapped at 72 characters?
- [X] Is the subject and message clear and concise?
- [X] Does the subject start with **[plugin_name]** if submitting a plugin patch or a **[section_name]** if part of the core sosreport code?
- [] Does the commit contain a **Signed-off-by: First Lastname <email@example.com>**?

PR #4273, opened 2026-03-17
Arch Linux policy

mawiegand:
Hey,

I’d like to see support for **Arch Linux**, and since there hasn’t been any progress on #1198 for a while, I’ve created a new pull request to introduce a policy for **Arch Linux**. I’d be thrilled if the policy made it into a release. I look forward to your feedback and/or any next steps I can take to make this happen!

Thanks! :slightly_smiling_face: 

Resolves #1198

---

### DIFF
--- a/sos/policies/distros/arch.py
+++ b/sos/policies/distros/arch.py
@@ -1,0 +1,51 @@
+# Copyright 2023 Marcel Wiegand <wiegand@linux.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import ArchPlugin
+from sos.policies.distros import LinuxPolicy
+from sos.policies.package_managers.pacman import PacmanPackageManager
+
+
+class ArchPolicy(LinuxPolicy):
+    distro = "Arch Linux"
+    vendor = "Arch Linux"
+    vendor_urls = [("Community Website", "https://www.archlinux.org/")]
+
+    def __init__(
+        self, sysroot=None, init=None, probe_runtime=True, remote_exec=None
+    ):
+        super().__init__(
+            sysroot=sysroot,
+            init=init,
+            probe_runtime=probe_runtime,
+            remote_exec=remote_exec,
+        )
+        self.package_manager = PacmanPackageManager(
+            chroot=self.sysroot, remote_exec=remote_exec
+        )
+        self.valid_subclasses += [ArchPlugin]
+
+    @classmethod
+    def check(cls, remote=""):
+        """This method checks to see if we are running on Arch Linux.
+        It returns True or False."""
+
+        if remote:
+            return cls.distro in remote
+
+        try:
+            with open("/etc/os-release", "r", encoding='utf-8') as f:
+                for line in f:
+                    if line.startswith("ID"):
+                        if "arch" in line:
+                            return True
+        except IOError:
+            return False
+        return False
--- a/sos/report/plugins/pacman.py
+++ b/sos/report/plugins/pacman.py
@@ -1,0 +1,40 @@
+# Copyright (C) 2023 Marcel Wiegand <wiegand@linux.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, ArchPlugin
+
+
+class Pacman(Plugin, ArchPlugin):
+
+    short_desc = "Package management utility for Arch Linux"
+
+    plugin_name = "pacman"
+    profiles = ("system", "sysmgmt", "packagemanager")
+
+    def setup(self):
+        self.add_copy_spec([
+            "/etc/pacman.conf",
+            "/etc/pacman.d",
+            "/var/log/pacman.log",
+        ])
+
+        self.add_cmd_output([
+            "pacman -Qi",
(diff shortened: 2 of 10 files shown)


sos/policies/distros/arch.py
```diff
@@ -0,0 +1,51 @@
...
+        )
+        self.package_manager = PacmanPackageManager(
+            chroot=self.sysroot, remote_exec=remote_exec
+        )
+        self.valid_subclasses += [ArchPlugin]
+
+    @classmethod
+    def check(cls, remote=""):
+        """This method checks to see if we are running on Arch Linux.
+        It returns True or False."""
```
TurboTurtle, line 38: No need to define this in the subclass, as it is handled in the base `LinuxPolicy`. Just set the `os_release_name` class attr.

sos/report/plugins/k3s.py
```diff
@@ -0,0 +1,22 @@
...
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import ArchPlugin
+from sos.report.plugins.kubernetes import Kubernetes
+
+
+class K3s(Kubernetes, ArchPlugin):
```
TurboTurtle: We don't have a pattern of subclassing `Plugin`s like this (into new plugin sources).

If there is truly nothing different about the collections except the `kube_cmd` and a single yaml file, I'd suggest we move this into the base `kubernetes` plugin. I'm not super familiar with `k3s` myself, but aren't there a reduced number of features with `k3s`? If so, we'd probably see a lot of collection failures when just directly subclassing the "normal" plugin, but I definitely get not wanting to re-write a plugin as large as that one just to strip out what may not actually be a lot of collection attempts.

@pmoravec or @arif-ali  happy to hear your thoughts here too.

sos/policies/distros/arch.py
```diff
@@ -0,0 +1,51 @@
+# Copyright 2023 Marcel Wiegand <wiegand@linux.com>
```
TurboTurtle, line 1: Wrong copyright year, unless these are coming from some downstream we're unaware of that has carried these since then?

TurboTurtle (commented):
Overall, pretty solid. Just a couple notes, but nothing major.

sos/policies/distros/arch.py
mawiegand: It’s actually been in my fork since then, so it should be correct. That said, I’m not too attached to that specific year.

TurboTurtle: Works for me then.

sos/policies/distros/arch.py
mawiegand: Thanks for the feedback. I did this to support `ID_LIKE` as well. See: https://www.man7.org/linux/man-pages/man5/os-release.5.html#OPTIONS

Or am I missing something, and is this also supported in the base `LinuxPolicy`? Or should I handle this differently?

sos/report/plugins/k3s.py
mawiegand: As far as I'm concerned, we're welcome to move this to `kubernetes`. I haven't noticed any issues with the collection process in the past. However, the plugin from my fork hasn't been used for quite some time. Alternatively, we could just drop it entirely, as far as I'm concerned. It doesn't have any direct connection to **Arch Linux**, and `k3s` is only available in the `AUR`.

pmoravec: +1 for moving it under Kubernetes.

sos/report/plugins/pacman.py
```diff
@@ -0,0 +1,32 @@
...
+
+    short_desc = "Package management utility for Arch Linux"
+
+    plugin_name = "pacman"
+    profiles = ("system", "sysmgmt", "packagemanager")
+
+    def setup(self):
+        self.add_copy_spec([
+            "/etc/pacman.conf",
+            "/etc/pacman.d",
```
pmoravec, line 24: Can't be there some secrets to obfuscate? Even credentials inside URL, like https://github.com/sosreport/sos/blob/main/sos/report/plugins/dnf.py#L156 ?

sos/report/plugins/k3s.py
mawiegand: Moved it to `kubernetes` as suggested.

sos/report/plugins/pacman.py
mawiegand: @pmoravec thanks for the feedback!

As far as I know, there shouldn't be any credentials in **Arch Linux** itself. However, a custom command like `curl` or `wget` could be used for the download, and that command might contain credentials. That's why I took up the suggestion and implemented a similar obfuscation measure as a safety precaution.

sos/policies/distros/arch.py
pmoravec: So then it is better to extend https://github.com/sosreport/sos/blob/main/sos/policies/distros/__init__.py#L118, rather than having an extra care for one distro..?

sos/report/plugins/k3s.py
mawiegand: @TurboTurtle can this be resolved?

sos/report/plugins/pacman.py
mawiegand: @pmoravec can this be resolved?

pmoravec:
LGTM with three comments:
- preferably follow my-our advice https://github.com/sosreport/sos/pull/4273#discussion_r2951557291
- preferably squash some commits, the granularity is too much, e.g. extending existing plugins by `ArchPlugin` can be done in 1-2 commit(s). (but I am also ok as it is now)
- pyint fails on https://github.com/mawiegand/sos/blob/73b9676bd176cd1c52a019adc4a82f92eaa35e72/sos/report/plugins/filesys.py#L108C1-L108C44 with `sos/report/plugins/filesys.py:108:0: R0901: Too many ancestors (8/7) (too-many-ancestors)`. Maybe can we mark the `Filesys` plugin as `IndependentPlugin`?

arif-ali:
@mawiegand are you still working on this PR?

arif-ali:
@mawiegand any update on this?

PR #4278, opened 2026-03-31
[charmed_mongodb] feat: Add Charmed MongoDB plugin

Gu1nness:
<---

### DIFF
--- a/sos/report/plugins/charmed_mongodb.py
+++ b/sos/report/plugins/charmed_mongodb.py
@@ -1,0 +1,455 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.#
+
+from enum import Enum
+from urllib.parse import quote_plus, urlencode
+import os
+import shutil
+from pathlib import Path
+import typing
+from typing import Dict, Optional, Tuple
+
+import yaml
+
+from sos.report.plugins import Plugin, PluginOpt, UbuntuPlugin
+from sos.utilities import is_executable
+
+DATE_FORMAT = "%Y-%m-%d-%H"
+
+
+class Substrate(Enum):
+    VM = "vm"
+    K8S = "k8s"
+
+
+Role = typing.Literal["replication", "shard", "config-server"]
+
+
+class CharmedMongoDB(Plugin, UbuntuPlugin):
+    """The Charmed MongoDB plugin is used to collect MongoDB configuration
+    and logs from the Charmed MongoDB snap package or K8s deployment.
+
+    If all_logs is set to True, it collects all logs by default.
+    The parameters `dbuser` and `dbpass` are used to dump database information,
+    replicaset status, shard status, etc. You can provide those parameters with
+    the environment variables `MONGODB_USER` and `MONGODB_PASSWORD`
+    """
+
+    short_desc = "Charmed MongoDB"
+    plugin_name = "charmed_mongodb"
+
+    # Triggers
+
+    packages = ("charmed-mongodb",)
+    containers = ("mongod",)
+
+    snap_package = "charmed-mongodb"
+    snap_path_common = "/var/snap/charmed-mongodb/common"
+    snap_path_current = "/var/snap/charmed-mongodb/current"
+
+    kube_cmd = "kubectl"
+    selector = "app.kubernetes.io/name=mongodb-k8s"
+
+    conf_paths = {
+        "MONGODB_CONF": "/etc/mongod",
+        "MONGODB_LOGS": "/var/log/mongodb",
+    }
+
+    option_list = [
+        PluginOpt(
+            name="dumpdbs",
+            default=False,
+            val_type=bool,
+            desc="Set to true to dump server information.",
+        ),
+        PluginOpt(
+            "dbuser",
+            default="",
+            val_type=str,
+            desc="Username for database dump collection",
+        ),
+        PluginOpt(
+            "dbpass",
+            default="",
+            val_type=str,
+            desc="Password for database dump collection",
+        ),
+    ]
+
+    regular_commands: Tuple[Tuple[str, str], ...] = (
+        ("EJSON.stringify(db.serverStatus())", "server_status.txt"),
+        ("EJSON.stringify(rs.status())", "replicaset_status.txt"),
+        ("EJSON.stringify(db.getUsers())", "db_users.txt"),
+        ("EJSON.stringify(db.getRoles())", "db_roles.txt"),
+        (
+            "EJSON.stringify(db.adminCommand({listDatabases: 1}))",
+            "db_databases.txt",
+        ),
+    )
+
+    config_server_commands: Tuple[Tuple[str, str], ...] = (
+        ("EJSON.stringify(sh.status())", "shard_cluster_status.txt"),
+        ("EJSON.stringify(sh.listShards())", "shard_shards.txt"),
+    )
+
+    def _join_conf_path(self, base: str, *parts: str):
+        stripped_parts = [p.lstrip(os.path.sep) for p in parts]
+        return self.path_join(base, *stripped_parts)
+
+    @staticmethod
+    def _match_role(role: str) -> Optional[Role]:
+        if role == "configsvr":
+            return "config-server"
+        if role == "shardsvr":
+            return "shard"
+        if role == "replication":
+            return "replication"
+        return None
+
+    def vm_role(self) -> Optional[Role]:
+        role: str = "replication"
+        conf_path = self._join_conf_path(
+            self.snap_path_current, self.conf_paths["MONGODB_CONF"]
+        )
+        try:
+            with open(
+                f"{conf_path}/mongod.conf",
+                encoding="utf-8",
+            ) as f:
+                data = yaml.safe_load(f)
+
+            if sharding_conf := data.get("sharding", {}):
+                role = sharding_conf.get("clusterRole", "")
+
+            return self._match_role(role)
+        except FileNotFoundError:
+            return None
+
+    def k8s_role(self, kube_cmd: str, cont: str, pod: str) -> Optional[Role]:
+        role: str = "replication"
+
+        # Cat the configuration file.
+        cat_conf_cmd = (
+            f"{kube_cmd} exec -c {cont} {pod} -- "
+            f"cat {self.conf_paths['MONGODB_CONF']}/mongod.conf"
+        )
+        result = self.exec_cmd(cat_conf_cmd)
+
+        if result.get("status") != 0:
+            return None
+
+        data = yaml.safe_load(result.get("output", ""))
+        if sharding_conf := data.get("sharding", {}):
+            role = sharding_conf.get("clusterRole", "")
+        return self._match_role(role)
+
+    def _get_db_credentials(self) -> Tuple[Optional[str], Optional[str]]:
+        db_user = self.get_option("dbuser")
+        db_pass = self.get_option("dbpass")
+
+        if not db_user:
+            if "MONGODB_USER" in os.environ:
+                self.soslog.info(
+                    "MONGODB_USER present: Using MONGODB_USER environment"
+                    "variable, user did not provide username."
+                )
+                db_user = os.environ["MONGODB_USER"]
+            else:
+                self.soslog.warning("error: Missing credentials (username)")
+                return None, None
+
+        if not db_pass:
+            if "MONGODB_PWD" in os.environ:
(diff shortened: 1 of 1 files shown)


sos/report/plugins/charmed_mongodb.py
```diff
@@ -0,0 +1,493 @@
...
+
+        if "MONGODB_USER" in os.environ and not db_user:
+            self.soslog.info(
+                "MONGODB_USER present: Using MONGODB_USER environment"
+                "variable, user did not provide username."
+            )
+            db_user = os.environ["MONGODB_USER"]
+        elif not db_user:
+            self.soslog.warning("error: Missing credentials (username)")
+            return None, None
```
pmoravec: isnt:

if not db_user:
    if "MONGODB_USER" in os.environ:
        self.soslog.info(..)
        db_user = os.environ["MONGODB_USER"]
    else:
        self.soslog.warning(..)
        returt None, None

shorter and better readable? Anyway I am ok with the current code.

sos/report/plugins/charmed_mongodb.py
```diff
@@ -0,0 +1,493 @@
...
+
+        if "MONGODB_PWD" in os.environ and not db_pass:
+            self.soslog.info(
+                "MONGODB_PWD present: Using MONGODB_PWD environment variable, "
+                "user did not provide password."
+            )
+            db_pass = os.environ["MONGODB_PWD"]
+        elif not db_pass:
+            self.soslog.warning("error: Missing credentials (password)")
+            return None, None
```
pmoravec: ditto like above

sos/report/plugins/charmed_mongodb.py
```diff
@@ -0,0 +1,493 @@
...
+        role = self.vm_role()
+        base_conf = self._join_conf_path(
+            self.snap_path_current, self.conf_paths["MONGODB_CONF"]
+        )
+        base_logs = self._join_conf_path(
+            self.snap_path_common, self.conf_paths["MONGODB_LOGS"]
+        )
+        if not role:
+            # The service is not properly set up by the charm, exiting.
+            return
```
pmoravec: worth moving just after role = .. assignment, to prevent ridiculous execution of lines above?

sos/report/plugins/charmed_mongodb.py
```diff
@@ -0,0 +1,493 @@
...
+        since = self.get_option("since")
+
+        for file in glob.glob(f"{base_logs}/*"):
+            date = re.search(
+                pattern=r"([0-9]{4}-[0-9]{2}-[0-9]{2}-[0-9]{2})", string=file
+            )
+            # include files without date, aka current files
+            if not date:
+                continue
+            file_dt = datetime.strptime(date.group(1), DATE_FORMAT)
```
pmoravec: Isn't it better to enhance `add_copy_spec` method by `since` option and let `a_s_c` itself to do this code (well, we compare by `mtime`, see https://github.com/sosreport/sos/blob/main/sos/report/plugins/__init__.py#L1783).

It will be slightly different behaviour, so I am ok with this. Just it is a possible enahncement of our API and reusing already existing code.

sos/report/plugins/charmed_mongodb.py
```diff
@@ -0,0 +1,493 @@
...
+            [
+                f"{base_conf}/*.pem",
+                f"{base_conf}/*.crt",
+                f"{base_conf}/keyFile",
+            ]
+        )
+        self.add_copy_spec([base_conf, base_logs])
+
+        lines = 500 if all_logs else None
+        self.add_journal("snap.charmed-mongodb.*", lines=lines)
```
pmoravec: Interesting limitation of "all logs" option, but I trust you know what you are setting here :)

sos/report/plugins/charmed_mongodb.py
```diff
@@ -0,0 +1,493 @@
...
+            self.exec_cmd(mkdir_cmd)
+
+            copy_cmd = f"{kube_cmd} cp -c {cont} {pwd_path} {pod}:{pwd_path}"
+            self.exec_cmd(copy_cmd)
+
+            query_cmd = (
+                f"{kube_cmd} exec -c {cont} {pod} -- "
+                f"sh -lc 'MONGODB_PWD=$(cat {pwd_path}) {mongod_cmd} {uri} "
+                f'--quiet --eval "{cmd}"'
+                f"&& rm -rf {tmpdir}"
```
pmoravec: great the password file is removed afterwards. I guess the `kubectl cp` respects permission and ownerships, such that the file will be visible by root only, in the container.

sos/report/plugins/charmed_mongodb.py
```diff
@@ -0,0 +1,493 @@
...
+
+                for file in glob.glob(f"{name_prefix}/{logs_path}/*"):
+                    date = re.search(
+                        pattern=r"([0-9]{4}-[0-9]{2}-[0-9]{2}-[0-9]{2})",
+                        string=file,
+                    )
+                    # include files without date, aka current files
+                    if not date:
+                        continue
+                    file_dt = datetime.strptime(date.group(1), DATE_FORMAT)
```
pmoravec: ditto suggestion wrt add_copy_spec enhancement

pmoravec:
Please squash the commits into one. generally ack otherwise.

sos/report/plugins/charmed_mongodb.py
Gu1nness: I agree it makes more sense :)

sos/report/plugins/charmed_mongodb.py
Gu1nness: This made me realize that there is an inconsistency it seems between the way I use since `since=2026-01-03` and the way it should be used `since=48h`.
It seems reasonable to keep the official definition.
I'll use the `maxage` parameter to handle that, but in case of `all_logs`, the `maxage` should then be "infinity".
I'll try to find a clean way to express that.

sos/report/plugins/charmed_mongodb.py
Gu1nness: I inverted the logic here, it should have been `None if all_logs else 500`.
Fixing that.

sos/report/plugins/charmed_mongodb.py
Gu1nness: My understanding is that it does yes

sos/report/plugins/charmed_mongodb.py
Gu1nness: This is done.

sos/report/plugins/charmed_mongodb.py
Gu1nness: Done

sos/report/plugins/charmed_mongodb.py
Gu1nness: Done

pmoravec (approved):

sos/report/plugins/charmed_mongodb.py
```diff
@@ -0,0 +1,472 @@
...
+            val_type=str,
+            desc="Username for database dump collection",
+        ),
+        PluginOpt(
+            "dbpass",
+            default="",
+            val_type=str,
+            desc="Password for database dump collection",
+        ),
+        PluginOpt(
```
arif-ali: There's no need for all-logs here. We have a global option already for --all-logs, and we should be using that instead.

sos/report/plugins/charmed_mongodb.py
```diff
@@ -0,0 +1,472 @@
...
+                f"sh -lc 'MONGODB_PWD=$(cat {pwd_path}) {mongod_cmd} {uri} "
+                f'--quiet --eval "{cmd}"'
+                f"&& rm -rf {tmpdir}"
+            )
+            self.add_cmd_output(
+                cmds=query_cmd,
+                suggest_filename=f"{pod}_{cmd_name}",
+            )
+
+    def _collect_per_namespace(
```
arif-ali: The kubernetes plugin already can collect details from namespaces and such, and that would be triggered if Canonical K8s, microk8s are being used. Is it worth adding all this again, as this would be duplicate data?

arif-ali (changes requested):
There is no trigger for this plugin, which means that this plugin will run on every Ubuntu system, and that doesn't make sense. Please used one of `packages`, `files`, `services` tuples that can help with triggering this plugin

sos/report/plugins/charmed_mongodb.py
Gu1nness: I think it might make sense it terms of separation of concerns (those are only the logs that matter for the charmed-mongodb which is deployed).
However I don't have a strong blocker for the log collection.
For the rest (specific logs of mongodb instance and dumbdbs, it does make sense to have it.
I can remove the first commands until it gets to collecting the mongodb specific logs.

Gu1nness:
@arif-ali Would you have time for a new review ? This one and also https://github.com/sosreport/sos/pull/4279

arif-ali (approved):

sos/report/plugins/charmed_mongodb.py
```diff
@@ -0,0 +1,449 @@
...
+        self.add_forbidden_path(
+            [
+                f"{dump_files_path}/*/mongod/*.pem",
+                f"{dump_files_path}/*/mongod/*.cert",
+                f"{dump_files_path}/*/mongod/keyFile",
+            ]
+        )
+
+        # METRICS
+        for pod in pods:
```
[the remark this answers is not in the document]
arif-ali: is this expected where the `pod` variable is not being used?

sos/report/plugins/charmed_mongodb.py
```diff
@@ -0,0 +1,449 @@
...
+        uri: str,
+        password: str,
+        cmd: str,
+        cmd_name: str,
+    ):
+        pwd_file = "_mongodb_pwd"
+        with tempfile.TemporaryDirectory() as tmpdir:
+            pwd_path = tmpdir + "/" + pwd_file
+            with open(pwd_path, mode="w", encoding="utf8") as f:
+                f.write(password)
```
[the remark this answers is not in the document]
arif-ali: as per the comment above, is there a way that we can avoid clear text passwords being written

Gu1nness:
@arif-ali Pushed the fix for the " Suspicious unused loop iteration variable".
The other one is a false positive and can be discarded.

arif-ali:
please resolve the flake8 and pylint tests, and rebase to main so that there is only one commit with only your changes

sos/report/plugins/charmed_mongodb.py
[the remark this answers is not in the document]
Gu1nness: I'm trying to see because we still need to transfer the password from the host to the pod.
Passing it in the CLI does not seem a better idea.

I'll have a look at other possibilities.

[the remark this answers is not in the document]
arif-ali: that would be great. other than this all looks good on this and the other PR

Gu1nness:
@arif-ali Done.
I also improved the copy of config and logs from the containers to the host so that it would not leave any residues in the `cmd_output_path`.
Same done on Mongos.

sos/report/plugins/charmed_mongodb.py
```diff
@@ -0,0 +1,453 @@
...
+    )
+
+    config_server_commands: Tuple[Tuple[str, str], ...] = (
+        ("EJSON.stringify(sh.status())", "shard_cluster_status.txt"),
+        ("EJSON.stringify(sh.listShards())", "shard_shards.txt"),
+    )
+
+    def _join_conf_path(self, base: str, *parts: str):
+        stripped_parts = [p.lstrip(os.path.sep) for p in parts]
+        return os.path.join(base, *stripped_parts)
```
arif-ali: you can use `self.path_join()` instead

sos/report/plugins/charmed_mongodb.py
```diff
@@ -0,0 +1,453 @@
...
+            substrate = Substrate.K8S
+
+        base_conf = self.conf_paths["MONGODB_CONF"]
+
+        if substrate == Substrate.VM:
+            base_conf = self._join_conf_path(self.snap_path_current, base_conf)
+        else:
+            base_conf = f"{self.get_cmd_output_path()}/pods/*/{base_conf}"
+            if os.path.isdir(f"{self.get_cmd_output_path()}/pods"):
+                shutil.rmtree(f"{self.get_cmd_output_path()}/pods")
```
arif-ali: What about using `ignore_errors=False` as an option to rmtree instead, then you don't need to check if that folder exists?

arif-ali (commented):
couple of comments below

sos/report/plugins/charmed_mongodb.py
Gu1nness: True, that's way better.

arif-ali (approved):

pmoravec (approved):

PR #4279, opened 2026-04-03
[charmed_mongos] feat: Add Charmed Mongos plugin

Gu1nness:
---

This is very similar to #4278 for the Mongos router that is used when using sharded mongodb.

### DIFF
--- a/sos/report/plugins/charmed_mongos.py
+++ b/sos/report/plugins/charmed_mongos.py
@@ -1,0 +1,387 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.#
+
+from enum import Enum
+import shutil
+from urllib.parse import quote, quote_plus, urlencode
+import os
+from pathlib import Path
+from typing import Dict, Optional, Tuple
+
+import yaml
+
+from sos.report.plugins import Plugin, PluginOpt, UbuntuPlugin
+from sos.utilities import is_executable
+
+DATE_FORMAT = "%Y-%m-%d-%H"
+
+
+class Substrate(Enum):
+    VM = "vm"
+    K8S = "k8s"
+
+
+class CharmedMongos(Plugin, UbuntuPlugin):
+    """The Charmed Mongos plugin is used to collect Mongos configuration
+    and logs from the Charmed Mongos snap package or K8s deployment.
+
+    If all_logs is set to True, it collects all logs by default.
+    The parameters `dbuser` and `dbpass` are used to dump database information,
+    replicaset status, shard status, etc. You can provide those parameters with
+    the environment variables `MONGOS_USER` and `MONGOS_PASSWORD`
+    """
+
+    short_desc = "Charmed Mongos"
+    plugin_name = "charmed_mongos"
+
+    # Triggers
+    packages = ("charmed-mongodb",)
+    containers = ("mongos",)
+
+    snap_package = "charmed-mongodb"
+    snap_path_common = "/var/snap/charmed-mongodb/common"
+    snap_path_current = "/var/snap/charmed-mongodb/current"
+
+    kube_cmd = "kubectl"
+    selector = "app.kubernetes.io/name=mongos-k8s"
+
+    conf_paths = {
+        "MONGODB_CONF": "/etc/mongod",
+        "MONGODB_LOGS": "/var/log/mongodb",
+    }
+
+    option_list = [
+        PluginOpt(
+            name="dumpdbs",
+            default=False,
+            val_type=bool,
+            desc="Set to true to dump server information.",
+        ),
+        PluginOpt(
+            "dbuser",
+            default="",
+            val_type=str,
+            desc="Username for database dump collection",
+        ),
+        PluginOpt(
+            "dbpass",
+            default="",
+            val_type=str,
+            desc="Password for database dump collection",
+        ),
+    ]
+
+    mongos_commands: Tuple[Tuple[str, str], ...] = (
+        ("EJSON.stringify(db.serverStatus())", "server_status.txt"),
+        ("EJSON.stringify(db.getUsers())", "db_users.txt"),
+        ("EJSON.stringify(db.getRoles())", "db_roles.txt"),
+        (
+            "EJSON.stringify(db.adminCommand({listDatabases: 1}))",
+            "db_databases.txt",
+        ),
+        ("EJSON.stringify(sh.status())", "shard_cluster_status.txt"),
+        ("EJSON.stringify(sh.listShards())", "shard_shards.txt"),
+    )
+
+    def _join_conf_path(self, base: str, *parts: str):
+        stripped_parts = [p.lstrip(os.path.sep) for p in parts]
+        return self.path_join(base, *stripped_parts)
+
+    def _get_db_credentials(self) -> Tuple[Optional[str], Optional[str]]:
+        db_user = self.get_option("dbuser")
+        db_pass = self.get_option("dbpass")
+
+        if not db_user:
+            if "MONGOS_USER" in os.environ:
+                self.soslog.info(
+                    "MONGOS_USER present: Using MONGOS_USER environment"
+                    "variable, user did not provide username."
+                )
+                db_user = os.environ["MONGOS_USER"]
+            else:
+                self.soslog.warning("error: Missing credentials (username)")
+                return None, None
+
+        if not db_pass:
+            if "MONGOS_PWD" in os.environ:
+                self.soslog.info(
+                    "MONGOS_PWD present: Using MONGOS_PWD environment "
+                    "variable, user did not provide password."
+                )
+                db_pass = os.environ["MONGOS_PWD"]
+            else:
+                self.soslog.warning("error: Missing credentials (password)")
+                return None, None
+
+        return db_user, db_pass
+
+    def vm_config_db(self) -> Optional[str]:
+        _conf_file = f"{self.conf_paths['MONGODB_CONF']}/mongos.conf"
+        conf_path = self._join_conf_path(self.snap_path_current, _conf_file)
+        try:
+            with open(conf_path, encoding="utf-8") as f:
+                data = yaml.safe_load(f)
+
+            return data.get("sharding", {}).get("configDB", None)
+        except FileNotFoundError:
+            return None
+
+    def k8s_config_db(
+        self, kube_cmd: str, cont: str, pod: str
+    ) -> Optional[str]:
+        # Cat the configuration file.
+        cat_conf_cmd = (
+            f"{kube_cmd} exec -c {cont} {pod} -- "
+            f"cat {self.conf_paths['MONGODB_CONF']}/mongos.conf"
+        )
+        result = self.exec_cmd(cat_conf_cmd)
+
+        if result.get("status") != 0:
+            return None
+
+        data = yaml.safe_load(result.get("output", ""))
+        return data.get("sharding", {}).get("configDB", None)
+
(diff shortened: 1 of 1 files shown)


arif-ali:
can you do a rebase instead, so that you only have your commit please

Gu1nness:
@arif-ali Done, thanks for the heads up.

sos/report/plugins/charmed_mongos.py
```diff
@@ -0,0 +1,385 @@
...
+            "EJSON.stringify(db.adminCommand({listDatabases: 1}))",
+            "db_databases.txt",
+        ),
+        ("EJSON.stringify(sh.status())", "shard_cluster_status.txt"),
+        ("EJSON.stringify(sh.listShards())", "shard_shards.txt"),
+    )
+
+    def _join_conf_path(self, base: str, *parts: str):
+        stripped_parts = [p.lstrip(os.path.sep) for p in parts]
+        return os.path.join(base, *stripped_parts)
```
arif-ali: same as the mongod plugin, use `self.path_join()` instead

sos/report/plugins/charmed_mongos.py
```diff
@@ -0,0 +1,385 @@
...
+        else:
+            substrate = Substrate.K8S
+
+        base_conf = self.conf_paths["MONGODB_CONF"]
+        if substrate == Substrate.VM:
+            base_conf = self._join_conf_path(self.snap_path_current, base_conf)
+        else:
+            base_conf = f"{self.get_cmd_output_path()}/pods/*/{base_conf}"
+            if os.path.isdir(f"{self.get_cmd_output_path()}/pods"):
+                shutil.rmtree(f"{self.get_cmd_output_path()}/pods")
```
arif-ali: what about using ignore_errors=True instead, and then there's no need to check for directory being there?

arif-ali (commented):
some comments, similar to mongod one

sos/report/plugins/charmed_mongos.py
Gu1nness: Done

sos/report/plugins/charmed_mongos.py
Gu1nness: Done

arif-ali (approved):

sos/report/plugins/charmed_mongos.py
```diff
@@ -0,0 +1,387 @@
...
+import os
+from pathlib import Path
+from typing import Dict, Optional, Tuple
+
+import yaml
+
+from sos.report.plugins import Plugin, PluginOpt, UbuntuPlugin
+from sos.utilities import is_executable
+
+DATE_FORMAT = "%Y-%m-%d-%H"
```
pmoravec: This is unused in the plugin.

sos/report/plugins/charmed_mongos.py
```diff
@@ -0,0 +1,387 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.#
```
pmoravec: Nitpick, trailing # :)

sos/report/plugins/charmed_mongos.py
```diff
@@ -0,0 +1,387 @@
...
+
+
+class CharmedMongos(Plugin, UbuntuPlugin):
+    """The Charmed Mongos plugin is used to collect Mongos configuration
+    and logs from the Charmed Mongos snap package or K8s deployment.
+
+    If all_logs is set to True, it collects all logs by default.
+    The parameters `dbuser` and `dbpass` are used to dump database information,
+    replicaset status, shard status, etc. You can provide those parameters with
+    the environment variables `MONGOS_USER` and `MONGOS_PASSWORD`
```
pmoravec: `s/MONGOS_PASSWORD/MONGOS_PWD`?

sos/report/plugins/charmed_mongos.py
```diff
@@ -0,0 +1,387 @@
...
+
+    def _get_db_credentials(self) -> Tuple[Optional[str], Optional[str]]:
+        db_user = self.get_option("dbuser")
+        db_pass = self.get_option("dbpass")
+
+        if not db_user:
+            if "MONGOS_USER" in os.environ:
+                self.soslog.info(
+                    "MONGOS_USER present: Using MONGOS_USER environment"
+                    "variable, user did not provide username."
```
pmoravec: Nitpick: Missing space between `environment` and `variable`.

sos/report/plugins/charmed_mongos.py
```diff
@@ -0,0 +1,387 @@
...
+        uri: str,
+        password: str,
+        cmd: str,
+        cmd_name: str,
+    ):
+        output_file = f"/tmp/eval_{cmd_name}"  # nosec: B108
+        # We first execute the command and write into a file
+        query_cmd = (
+            f"{kube_cmd} exec -c {cont} {pod} -- "
+            f'sh -lc \'export MONGOS_PWD="{password}"; {mongod_cmd} {uri} '
```
pmoravec: Not sure if we had this discussion here or somewhere else, but cant the `MONGOS_PWD=<real-password>` be visible in process listing?

sos/report/plugins/charmed_mongos.py
```diff
@@ -0,0 +1,387 @@
...
+        kube_cmd: str,
+        cont: str,
+        pod: str,
+        mongod_cmd: str,
+        uri: str,
+        password: str,
+        cmd: str,
+        cmd_name: str,
+    ):
+        output_file = f"/tmp/eval_{cmd_name}"  # nosec: B108
```
pmoravec: Cant concurrent sos executions overlap the file..? (or is that so so so corner case that we can ignore?)

sos/report/plugins/charmed_mongos.py
Gu1nness: We did not have this discussion indeed but I think it could be the case.
The thing is that I removed the use of a side-file because CodeQL was reporting that we were writing the password to a file. It feels like one way or another, we have to send the password from the host to the container.
If it is written to a file that is sent over, then it will appear on the file system for a few seconds.
If it is passed as an environment variable on the command executed on the pod, then it appears in process listing.
Sadly, `kubectl exec` does not provide a way to forward the local environment to the remote pod to execute the command.
I will try a final idea this morning and report here.

sos/report/plugins/charmed_mongos.py
Gu1nness: I think that it is extremely corner case: It would mean you're running two concurrent reports on the same deployment at the same time, and both are executing the `remote_exec` at the same moment.
However I think it's not very complex to ensure that they never overlap so I'll add that.

sos/report/plugins/charmed_mongos.py
Gu1nness: Fixed

sos/report/plugins/charmed_mongos.py
Gu1nness: Fixed

sos/report/plugins/charmed_mongos.py
Gu1nness: Fixed

sos/report/plugins/charmed_mongos.py
Gu1nness: Fixed

sos/report/plugins/charmed_mongos.py
Gu1nness: Fixed: Passing password as stdin to kubectl exec and using `--password` on mongosh to read password from stdin.

sos/report/plugins/charmed_mongos.py
Gu1nness: Fixed by appending a random string to the file name.

sos/report/plugins/charmed_mongos.py
```diff
@@ -198,6 +171,7 @@ def _process_snap(self):
                 f"{base_conf}/*.pem",
                 f"{base_conf}/*.crt",
                 f"{base_conf}/keyFile",
+                f"{base_conf}/mongod.conf",
```
pmoravec, line 174: The plugin will *not* collect `mongod.conf` file? I guess it is intentional but worth double-checking.

sos/report/plugins/charmed_mongos.py
```diff
@@ -306,15 +267,19 @@ def _remote_exec(
...
+        query_cmd = shlex.split(
+            f"{kube_cmd} exec -i -c {cont} {pod} -- "
+            f'sh -lc \'{mongod_cmd} {uri} '
+            f'--quiet --eval "{cmd}" --password > {output_file}\''
         )
 
-        self.exec_cmd(query_cmd)
+        # Pass the password as stdin that is forwarded to the
+        # container through kubectl exec
+        subprocess.run(query_cmd, input=password.encode(), check=False)
```
pmoravec, line 282: Nice! I think we should have this password passing as an option in our `utilities.py`, @TurboTurtle ?

sos/report/plugins/charmed_mongos.py
```diff
@@ -360,60 +325,37 @@ def _collect_per_namespace(self, ns: str, all_logs: bool):
                     f"{dump_files_path}/pods/{pod}/{conf_path}/*.pem",
                     f"{dump_files_path}/pods/{pod}/{conf_path}/*.crt",
                     f"{dump_files_path}/pods/{pod}/{conf_path}/keyFile",
+                    f"{dump_files_path}/pods/{pod}/{conf_path}/mongod.conf",
```
pmoravec, line 328: Ditto to ensure `mongod.conf` is not collected by purpose.

sos/report/plugins/charmed_mongos.py
pmoravec, line 393: In `Substrate.K8S` case, we first delete parent-parent dir of `base_conf` and then scrub passwords in the dir..?

sos/report/plugins/charmed_mongos.py
Gu1nness: It is intentional, because mongod is not used on Charmed Mongos, only  èmongos.conf`

sos/report/plugins/charmed_mongos.py
Gu1nness: I think it would be a good idea to be able to have an input to `exec_cmd` for this kind of situation yes

sos/report/plugins/charmed_mongos.py
Gu1nness: Answered above, yes it's on purpose :)

sos/report/plugins/charmed_mongos.py
Gu1nness: The path regex substitution is done in the output archive in the files directory, while the directory that is removed it the one that lives in the cmd output path if I'm not mistaken. So they are not the same paths unless I'm mistaken?
`do_path_regexp_sub` is for a set of files archived by sos, while for commands, it would be `do_cmd_output_sub`

sos/report/plugins/charmed_mongos.py
pmoravec: ACK, thx for explanation.

sos/report/plugins/charmed_mongos.py
pmoravec: Ah, indeed,you are right, my mistake.

pmoravec (approved):

sos/report/plugins/charmed_mongos.py
TurboTurtle: Yeah, agreed. Let's get this into `utilities.sos_get_command_output()` and then exposed via the various Plugin methods.

Sorry I missed this before merge.

pmoravec: I raised https://github.com/sosreport/sos/issues/4317 and planning to work on it (later, imho not urgent?).

PR #4286, opened 2026-04-10
[cleaner] Optimize file obfuscation with set-based lookup and pre-filters

gotoweb:
Reopening this PR as the previous PR #4281 was closed during the process of resolving conflicts and recreating the branch.

Relevant: #4227 

---

### DIFF
--- a/sos/cleaner/mappings/__init__.py
+++ b/sos/cleaner/mappings/__init__.py
@@ -28,6 +28,12 @@ class SoSMap():
     compile_regexes = True
     ignore_short_items = False
     match_full_words_only = False
+    # When True, use set-based token lookup instead of compiled_search regex
+    # for the pre-filter in _parse_line_with_compiled_regexes.
+    # Suitable for parsers with simple word items (username, keyword).
+    use_token_lookup = False
+
+    _token_split_re = re.compile(r'[^a-z0-9]+')
 
     def __init__(self, workdir):
         self.initializing = True
@@ -35,6 +41,9 @@ class SoSMap():
         self._regexes_made = set()
         self.compiled_regexes = []
         self.compiled_search = re.compile('(?!)')  # no match
+        self._simple_tokens = set()
+        self._complex_items = []
+        self._regex_dict = {}
         self.cname = self.__class__.__name__.lower()
         # workdir's default value '/tmp' is used just by avocado tests,
         # otherwise we override it to /etc/sos/cleaner (or map_file dir)
@@ -165,9 +174,17 @@ class SoSMap():
             # from scratch every time we add something like we would do if we
             # tracked/saved the item and the Pattern() object in a dict or in
             # the set above
-            self.compiled_regexes.append((item, self.get_regex_result(item)))
+            _reg = self.get_regex_result(item)
+            self.compiled_regexes.append((item, _reg))
             self.compiled_regexes.sort(key=lambda x: len(x[0]), reverse=True)
-            self.generate_compiled_regexes(only_search=True)
+            self._regex_dict[item] = _reg
+            if self.use_token_lookup:
+                if item.isalnum():
+                    self._simple_tokens.add(item)
+                else:
+                    self._complex_items.append(item)
+            else:
+                self.generate_compiled_regexes(only_search=True)
 
     def generate_compiled_regexes(self, only_search=False):
         keys = sorted(self._regexes_made, key=len, reverse=True)
@@ -175,11 +192,49 @@ class SoSMap():
             self.compiled_regexes = [
                 (item, self.get_regex_result(item)) for item in keys
             ]
-        pattern = "|".join([f'{self.get_regex_escape(k)}' for k in keys])
-        self.compiled_search = re.compile(
-            self.get_regex_fullword(pattern),
-            flags=re.I
-        )
+            self._regex_dict = dict(self.compiled_regexes)
+        if self.use_token_lookup:
+            self._simple_tokens = {
+                item for item in self._regexes_made if item.isalnum()
+            }
+            self._complex_items = [
+                item for item in self._regexes_made if not item.isalnum()
+            ]
+        else:
+            pattern = "|".join([f'{self.get_regex_escape(k)}' for k in keys])
+            self.compiled_search = re.compile(
+                self.get_regex_fullword(pattern),
+                flags=re.I
+            )
+
+    def get_matched_items(self, line):
+        """Return (item, regex) pairs for items that match in the line.
+
+        For token-lookup maps, returns only the specific items found via
+        set intersection / substring matching, rather than the full
+        compiled_regexes list.
+
+        For regex-based maps, falls back to returning all compiled_regexes
+        when compiled_search matches (original behavior).
+        """
+        if self.use_token_lookup:
+            line_lower = line.lower()
+            tokens = set(self._token_split_re.split(line_lower))
+            tokens.discard('')
+
+            result = []
+            for item in tokens & self._simple_tokens:
+                result.append((item, self._regex_dict[item]))
+            for item in self._complex_items:
+                if item in line_lower:
+                    result.append((item, self._regex_dict[item]))
+
+            if len(result) > 1:
+                result.sort(key=lambda x: len(x[0]), reverse=True)
+            return result
+        if self.compiled_search.search(line):
+            return self.compiled_regexes
+        return []
 
     def get_regex_escape(self, item):
         return rf'{re.escape(item)}'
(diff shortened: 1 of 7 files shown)


pmoravec:
> resolving conflicts and recreating the branch.

You can also rebase your branch to current upstream, there are multiple ways how to achieve that. Anyway, whatever works for me/us (including opening new PR) :). Reviewing the code now and running same perf.comparison (preliminary results are better than before).

sos/cleaner/mappings/__init__.py
```diff
@@ -165,21 +174,71 @@ def add_regex_item(self, item):
...
+                    self._simple_tokens.add(item)
+                else:
+                    self._complex_items.append(item)
+            else:
+                self.compiled_regexes.append((item, _reg))
+                self.compiled_regexes.sort(
+                    key=lambda x: len(x[0]), reverse=True)
+                self.generate_compiled_regexes(only_search=True)
 
     def generate_compiled_regexes(self, only_search=False):
```
pmoravec, line 190: With this change, the `generate_compiled_regexes` method does not further juts generate compiled regexes (not only `self.compiled_search = ..`) but it .. (generates proper structures for efficient pattern search?). I would appreciate change in the method name to prevent some confusion, but I lack a better name.

So until one gets a better name, let ignore this comment.

sos/cleaner/parsers/__init__.py
```diff
@@ -76,8 +76,11 @@ def generate_item_regexes(self):
         """
         if not self.compile_regexes:
             return
+        self.mapping.initializing = True
         for obitem in self.mapping.dataset:
             self.mapping.add_regex_item(obitem)
+        self.mapping.initializing = False
+        self.mapping.generate_compiled_regexes()
```
pmoravec, line 83: I see reasoning for this (generate proper structures for efficient pattern search, *after* we preload `/etc/hosts` of `lastlog` and similar), which makes me think if the other call at https://github.com/gotoweb/sos/blob/main/sos/cleaner/mappings/__init__.py#L47 isnt redundant..? At least for those parsers with `self.compile_regexes = True`..?

Currently the method `generate_compiled_regexes` is explicitly called twice, that is redundant..

sos/cleaner/parsers/__init__.py
```diff
@@ -110,14 +113,13 @@ def _parse_line_with_compiled_regexes(self, line):
...
-                    count += _count
-                    # break the cycle if no further search can apply
-                    if not self.mapping.compiled_search.search(line):
-                        break
+        for item, reg in self.mapping.get_matched_items(line):
+            if reg.search(line):
+                line, _count = reg.subn(self.mapping.get(item), line)
+                count += _count
+                # break the cycle if no further search can apply
+                if not self.mapping.compiled_search.search(line):
```
pmoravec: So *here* you need `self.mapping.compiled_search` even for `use_token_lookup = True`? 

This brings me to counterexample:

```
mkdir sosreport
echo "foo bar foo bar qux" > sosreport/test.txt
rm -rf /etc/sos/cleaner/* /var/tmp/sos*
python3 bin/sos clean sosreport --batch --keywords foo,bar,qux
cat /var/tmp/sosreport
```

Then I randomly get

```
obfuscatedword0 bar obfuscatedword0 bar qux
```

or

```
foo obfuscatedword1 foo obfuscatedword1 qux
```

depending what is the first pair in `for item, reg in self.mapping.get_matched_items(line):` loop.

This condition in `if not self.mapping.compiled_search.search(line):` must be conditional depending on `use_token_lookup` value.

sos/cleaner/parsers/ipv6_parser.py
```diff
@@ -18,6 +18,11 @@ class SoSIPv6Parser(SoSCleanerParser):
 
     name = 'IPv6 Parser'
     map_file_key = 'ipv6_map'
+    # Cheap pre-filter: "::" (compressed) or "hex:hex:" (full form).
+    # Every valid IPv6 address contains at least one of these substrings.
+    _quick_check = re.compile(
+        r'::|[0-9a-f]{1,4}:[0-9a-f]{1,4}:', re.I
+    )
```
pmoravec, line 25: Nice (as well as in `mac_parser`)!

sos/cleaner/parsers/__init__.py
gotoweb: Thanks for the review and the counterexample, It made the bug very clear.
As you suggested, I fixed this by skipping the early-exit for token-lookup parsers

https://github.com/gotoweb/sos/commit/7e762c0d06e439f864d3ef9d5ed430a7ad87e3a6

Token-lookup parsers don't maintain `compiled_search`, so the check was always `True` after the first substitution, causing the loop to break prematurely. With this guard, token-lookup parsers rely on `get_matched_items()` exhausting its results naturally.

I also added regression tests for multi-keyword and multi-username obfuscation to `tests/unittests/cleaner_tests.py`

sos/cleaner/parsers/__init__.py
pmoravec: Deferring this to @TurboTurtle , I think we can drop the https://github.com/gotoweb/sos/blob/main/sos/cleaner/mappings/__init__.py#L47 but not 100% sure. Having both calls adds a tiny (probably redundant) execution time.

sos/cleaner/parsers/__init__.py
```diff
@@ -110,14 +113,16 @@ def _parse_line_with_compiled_regexes(self, line):
...
-        if self.mapping.compiled_search.search(line):
-            for item, reg in self.mapping.compiled_regexes:
-                if reg.search(line):
-                    line, _count = reg.subn(self.mapping.get(item), line)
-                    count += _count
-                    # break the cycle if no further search can apply
-                    if not self.mapping.compiled_search.search(line):
-                        break
+        for item, reg in self.mapping.get_matched_items(line):
+            if reg.search(line):
```
pmoravec: Just a thought: for `use_token_lookup=True`, we know all `(item, reg)` pairs will be found in the line, so this search is redundant..?

Is it safe to have here

                if self.mapping.use_token_lookup or reg.search(line):

?

(apologize for such iterative feedback, I should raise this question in previous rounds already..)

sos/cleaner/parsers/__init__.py
```diff
@@ -159,7 +164,7 @@ def parse_string_for_keys(self, string_data):
         :rtype: ``str``
         """
         if self.compile_regexes:
-            for item, reg in self.mapping.compiled_regexes:
+            for item, reg in self.mapping.get_matched_items(string_data):
                 if reg.search(string_data):
```
pmoravec: Ditto for `self.mapping.use_token_lookup or ..` sub-condition.

pmoravec:
Nice improvement overall. Leaving one decision to @TurboTurtle plus one improvement on 2 places to the author, other than that I will ACK it.

A very rule of thumb performance gained will be 5-6% over all (per my benchmark; better results on systems e.g. with many users).

sos/cleaner/parsers/__init__.py
gotoweb: Thanks for the suggestion. You're right, for token-lookup maps, `get_matched_items()` already confirms presence via set intersection and substring match, so `reg.search()` is redundant here.

Applied if `self.mapping.use_token_lookup or reg.search(line): `in both `_parse_line_with_compiled_regexes()` and `parse_string_for_keys()`.

sos/cleaner/parsers/__init__.py
gotoweb: Done. same pattern applied here as well.

sos/cleaner/parsers/__init__.py
gotoweb: I investigated removing the `generate_compiled_regexes()` call from https://github.com/gotoweb/sos/blob/faster-combined/sos/cleaner/mappings/__init__.py#L56 , It cannot be safely removed.

During `load_entries()`, `add_regex_item()` runs with `initializing=True`, which only adds items to `_regexes_made` without building actual regex objects. The line 56 is what materializes those cached entries into usable regex structures.

Later `add_regex_item()` calls (with `initializing=False`) do build regexes incrementally for new items, but they don't cover items already in `_regexes_made` from the cache load. Removing line 56 leaves cached entries without regexes.

Confirmed by 3 unit test failures 
- `test_hostname_short_name_in_line`
- `test_keyword_parser_concurrently_changed`
- `test_ad_username`

pmoravec: Thanks for the effort made! It is bit sad we are forced to do some (semi)redundant calls as both calls do matter, but until one gets an idea how to make this more efficient, I am ok with it.

pmoravec (approved):
No further comments from my side. Thanks for the improvement and iterative feedback from my side.

Comparing performance on my test suite, I see 5-10% improvement (depends on multiple factors, the bigger sosreport or default mapping, the bigger (even relative) improvement). Nice!

Let's wait for a second approval ACK now..

pmoravec:
Kindly asking @arif-ali or @TurboTurtle for a review, this improves cleaner performance by 5-10%.

TurboTurtle (approved):
LGTM

PR #4290, opened 2026-04-13
[RFC] [podman] Collect podman Data for All Users

sahithiRavindranath:
Issue Description:
The current sos report only gathers podman data for the root user. It is unable to collect container data belonging to non‑root users because sos report runs with elevated (sudo) privileges.

Solution:
Added a new plugin option, allusers, to the podman plugin. When enabled, this option collects podman container data for all non‑root users which has container running on the system in addition to the root user.

---

### DIFF
--- a/sos/report/plugins/podman.py
+++ b/sos/report/plugins/podman.py
@@ -8,6 +8,8 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
+import shlex
+import subprocess
 from sos.report.plugins import Plugin, RedHatPlugin, UbuntuPlugin, PluginOpt
 
 
@@ -44,7 +46,9 @@ class Podman(Plugin, RedHatPlugin, UbuntuPlugin):
                     'container produces stdout/stderr output. Use cautiously '
                     'when also using the \'all\' option.')),
         PluginOpt('size', default=False,
-                  desc='collect image sizes for podman ps')
+                  desc='collect image sizes for podman ps'),
+        PluginOpt('allusers', default=False,
+                  desc='collect for all users, including non root users')
     ]
 
     def setup(self):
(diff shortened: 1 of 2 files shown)


sos/report/plugins/podman.py
```diff
@@ -68,57 +81,142 @@ def setup(self):
...
-        ], recursive=True)
+        # Iterate through each user to collect their podman data
+        for user in users:
+            if not user:
+                continue
+            command = "sudo -u "+ user
+            if user == 'root':
+                command = ""
+            else:
+                # Check if user has any containers before collecting data
```
pmoravec, line 92: Worth to move `command = "sudo -u "+ user` from line 88 here.

sos/report/plugins/podman.py
```diff
@@ -44,14 +44,27 @@ class Podman(Plugin, RedHatPlugin, UbuntuPlugin):
...
-            'podman images': 'podman_list_images',
-            'podman ps': 'podman_list_containers'
-        })
+        # Collect podman data for root user by default
+        users = ['root']
+        # If allusers option is enabled, collect for all non-root users too
+        if self.get_option('allusers'):
+            non_root_users = self.exec_cmd("lslogins -u -o USER --noheadings")
+            if non_root_users['status'] == 0:
+                for user in non_root_users['output'].splitlines()
```
pmoravec: Missing trailing `:`

sos/report/plugins/podman.py
```diff
@@ -68,57 +81,142 @@ def setup(self):
...
+            else:
+                # Check if user has any containers before collecting data
+                cmd = self.exec_cmd(f"{command} podman ps -aq")
+                if (cmd['status'] != 0
+                       or not cmd['output'].strip()):
+                    continue
+
+            self.add_cmd_tags({
+                f'{command} podman images': 'podman_list_images',
+                f'{command} podman ps': 'podman_list_containers'
```
pmoravec, line 100: Do we (and should we) support adding one tag for multiple command outputs, @TurboTurtle ? This is being tagged for each and every user, incl. `root`.

sos/report/plugins/podman.py
```diff
@@ -68,57 +81,142 @@ def setup(self):
...
+        # Iterate through each user to collect their podman data
+        for user in users:
+            if not user:
+                continue
+            command = "sudo -u "+ user
+            if user == 'root':
+                command = ""
+            else:
+                # Check if user has any containers before collecting data
+                cmd = self.exec_cmd(f"{command} podman ps -aq")
```
pmoravec, line 93: This fails if the `user` does not have permissions to access ${PWD} where the `root` is. E.g. trying this on a containerised AAP deployment where podman containers run under my `aap` user:

```
# who am I
root     pts/0        2026-04-13 12:03 (10.44.32.126)
# pwd
/root/sos-main-pr4290
# sudo -u aap podman ps -aq
cannot chdir to /root/sos-main-pr4290: Permission denied
# su - aap -c "podman ps -aq"
8a634b8461cb
.. list of podman container ids ..
#
```

(but I dont know if the `su - {user} -c ..` is the best approach here)

sos/report/plugins/podman.py
```diff
@@ -68,57 +81,142 @@ def setup(self):
...
-            '/etc/containers'
-        ], recursive=True)
+        # Iterate through each user to collect their podman data
+        for user in users:
+            if not user:
+                continue
+            if user == 'root':
+                command = ""
+            else:
+                command = "sudo -u "+ user
```
TurboTurtle, line 91: nack.

We do not permit embedded `sudo` calls in plugins. For any collection that needs to be done as a non-root user, you need to use the `runas` parameter to `add_cmd_output()`.  I believe we have a single exception to this today in `pulp`, but it is something we have worked to eliminate over the last few years.

sos/report/plugins/podman.py
```diff
@@ -68,57 +81,142 @@ def setup(self):
...
-        self.add_dir_listing([
-            '/etc/cni',
-            '/etc/containers'
-        ], recursive=True)
+        # Iterate through each user to collect their podman data
+        for user in users:
+            if not user:
+                continue
+            if user == 'root':
+                command = ""
```
TurboTurtle, line 89: Rather than nesting everything under a potentially large item loop like this, it'd be cleaner to instead break the collection into another method (or several, even) and call that method from the loop, with the user as an optional argument. The default here would be to run as `root`, and always be fired, with non-root iterations only firing if the `users` list above is populated at all (i.e. `root` should not be included).

This would be a cleaner approach and would also allow for easier future modification and any per-user validations we may want to do.

TurboTurtle (changes requested):

sos/report/plugins/podman.py
TurboTurtle: We do, and I believe we should. I'm not immediately opposed to adding the same tag that `root` gets to non-root users, but I'd suggest checking with at least the Insights team to see if this would throw off anything for them.

sos/report/plugins/podman.py
sahithiRavindranath: Fixed

sos/report/plugins/podman.py
sahithiRavindranath: I've removed all embedded sudo calls and now use the runas parameter instead, as requested by @TurboTurtle.

sos/report/plugins/podman.py
sahithiRavindranath: Fixed! Using the runas parameter instead of embedded sudo solves this - the sos framework handles working directory permissions properly. I've also added exception handling to gracefully skip users that can't be accessed

sos/report/plugins/podman.py
sahithiRavindranath: Fixed! All embedded sudo calls removed. Now using runas parameter throughout.

sos/report/plugins/podman.py
sahithiRavindranath: broke the loop into several helper functions  and calling the those function for root user always regardless if non-root_list is populated at all.

sos/report/plugins/podman.py
sahithiRavindranath: Current implementation uses the same tag for all users. Should I switch to unique tags per user (e.g., podman_list_images_root, podman_list_images_user1) or keep the shared tag approach?"

sahithiRavindranath:
I can see there are 1 test case failing on Debian and 2 on Ubuntu:

Debian 13
Link: https://github.com/sosreport/sos/pull/4290/checks?check_run_id=71792161692

Error: ModuleNotFoundError: No module named 'pkg_resources'

This is nothing to do with my changes I think. The module is needed for package build.

Ubuntu 25.10 DEB
Link: https://github.com/sosreport/sos/pull/4290/checks?check_run_id=71792157716

Results: PASS 254 | ERROR 0 | FAIL 1 | SKIP 26 | WARN 0 | INTERRUPT 0 | CANCEL 0

Job Time: 990.12 s

Failed Test: 152-tests/report_tests/encryption_tests.py:EncryptedCleanedReportTest.test_archive_gpg_encrypted

This is something to do with Encryption, which I think is not related to my patch.

Ubuntu 25.10 Snap
✅ 00:04 clone
❌ 29:55 setup
It is a setup failure.

@pmoravec  @TurboTurtle It seems the above test case failures are unrelated to the changes introduced by this pull request. Please confirm if that is not the case.

arif-ali:
@sahithiRavindranath if you rebase to main, the `pkg_resources` will go away. I have just restarted the 2 ubuntu jobs, as one timed out, and the other may have other issues.

sos/report/plugins/podman.py
```diff
@@ -68,57 +72,212 @@ def setup(self):
...
+                    users.append(user)
+        return users
+
+    def _user_has_containers(self, user):
+        """Check if user has any containers before collecting data"""
+        try:
+            cmd = self.exec_cmd("podman ps -aq", runas=user)
+            return cmd['status'] == 0 and cmd['output'].strip()
+        except subprocess.SubprocessError:
+            return False
```
TurboTurtle: We shouldn't ever be throwing a `subprocess` exception here via `exec_cmd()`, or at least we don't _intend_ to be doing so. Did you run into one? I'd rather fix that within `Plugin` (separately to this PR), than capture that here. The intent behind `exec_cmd()`, like anything else that culminates in a call to `sos_get_command_output()`, is that any exception results in a non-zero `status`  value rather than trickling back up to the call.

sos/utilities.py
```diff
@@ -298,7 +298,10 @@ def _check_poller(proc):
             'HOME': pwd_user.pw_dir,
             'LOGNAME': runas,
             'PWD': pwd_user.pw_dir,
-            'USER': runas
+            'USER': runas,
+            # XDG_RUNTIME_DIR is required for rootless podman to access
+            # user-specific runtime files and sockets
+            'XDG_RUNTIME_DIR': f"/run/user/{pwd_user.pw_uid}"
```
TurboTurtle, line 308: ack for this usecase, but I wonder if there are situation where we _don't_ want to be passing this by default? @pmoravec @arif-ali any concerns here?

sos/report/plugins/podman.py
```diff
@@ -8,6 +8,8 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
+import shlex
```
TurboTurtle: We haven't seen the need to `shlex.quote` the entities this is being added to previously, or at least none has been reported.

I'm curious what causes the need here? Not saying doing this is a problem or anything, just wondering.

TurboTurtle (commented):
Looking very solid, just a few minor touch points below.

sos/report/plugins/podman.py
sahithiRavindranath: Added as defensive coding for edge cases with special characters in container/image/volume names. However, since it hasn't been needed historically, I've removed it per your feedback. All `shlex.quote()` calls and the import have been removed.

sos/report/plugins/podman.py
sahithiRavindranath: Yes, encountered this during testing when switching to certain users:

subprocess.SubprocessError: Exception occurred in preexec_fn.
  File "/usr/lib64/python3.9/subprocess.py", line 1838, in _execute_child
    raise child_exception_type(err_msg)

The exception occurs at utilities.py line 346 during Popen() initialization when preexec_fn fails during user switching (permission issues).

I've removed the try-except from the plugin. Can the exception handling be added in sos_get_command_output() as below?

```diff
diff --git a/sos/utilities.py b/sos/utilities.py
index 2646ab22..4910acbe 100644
--- a/sos/utilities.py
+++ b/sos/utilities.py
@@ -10,7 +10,7 @@ import os
 import pwd
 import re
 import inspect
-from subprocess import Popen, PIPE, STDOUT
+from subprocess import Popen, PIPE, STDOUT, SubprocessError
 import logging
 import fnmatch
 import errno
@@ -407,6 +407,11 @@ def sos_get_command_output(command, timeout=TIMEOUT_DEFAULT, stderr=False,
         if e.errno == errno.ENOENT:
             return {'status': 127, 'output': "", 'truncated': ''}
         raise e
+    except SubprocessError:
+        # Handle exceptions in preexec_fn (e.g., user switching failures)
+        # Return 126 (command cannot execute)
+        return {'status': 126, 'output': "", 'truncated': ''}
     finally:
         if hasattr(_output, 'close'):
             _output.close()
```

sos/report/plugins/podman.py
```diff
@@ -68,57 +70,209 @@ def setup(self):
...
+        if result['status'] == 0:
+            for line in result['output'].splitlines():
+                user = line.strip()
+                if user and user != "root":
+                    users.append(user)
+        return users
+
+    def _user_has_containers(self, user):
+        """Check if user has any containers before collecting data"""
+        cmd = self.exec_cmd("podman ps -aq", runas=user)
```
pmoravec, line 116: This does not scale well when there are tens or hundreds of users (I hope `lslogins -u -o USER --noheadings` does not list AD / LDAP users..?). I think it is fine for now, but we might need to improve it in future.

sos/utilities.py
pmoravec: No concerns.

pmoravec (approved):

sos/report/plugins/podman.py
```diff
@@ -68,57 +70,209 @@ def setup(self):
...
-                                        tags='podman_list_networks')
+        # Always collect for root (None represents root user)
+        self._collect_for_user(None, '', subcmds)
+
+        # Only collect for non-root users if allusers option is enabled
+        if self.get_option('allusers'):
+            non_root_users = self._get_non_root_users()
+            for user in non_root_users:
+                if self._user_has_containers(user):
+                    self._collect_for_user(user, f'{user}/', subcmds)
```
TurboTurtle: One last change request here. Instead of putting user dirs in the top-level plugin directory, let's nest them, perhaps under a `users` dir or somesuch. Just to keep things a little more tidy.

TurboTurtle (commented):
Minor change I'd like to see, should have caught it earlier, my apologies for the review delay.

sos/report/plugins/podman.py
sahithiRavindranath: Ack. I'll optimize it in future iterations.

sos/report/plugins/podman.py
sahithiRavindranath: Fixed. User directories are now nested under users/

sahithiRavindranath:
rebased to main

sos/report/plugins/podman.py
sahithiRavindranath: @TurboTurtle Following up on the diff above. The SubprocessError handling would prevent preexec_fn exceptions by returning status 126 instead.  Thoughts on this approach?

sahithiRavindranath:
rebased to main  ---> no changes

sos/report/plugins/podman.py
sahithiRavindranath: @TurboTurtle Following up on the SubprocessError handling discussion. I've opened PR #4363 to address this separately.

Would appreciate your review when you have a chance.

sos/report/plugins/podman.py
sahithiRavindranath: @TurboTurtle  I've nested all user directories under users/ as requested. Gentle reminder for review when you get a chance. Thanks!

TurboTurtle (approved):

Issue #4272, opened 2026-03-12
Collect user groups details


pafernanr:
Hello,

I noticed user groups are not listed/collected by `sos report` I would expect groups to be collected, maybe something similar to `sos_commands/login/lslogins`. The most similar command I'm aware is `getent group`.

In addition could it be useful to also include all/other supported `getent` entities output? 
If yes, where would be the right place? existent host plugin? new getent plugin? 

I think almost all `getent` data is already collected. check:
~~~
for ent in $(getent --help | grep "Supported databases" -A1 | grep -v "Supported databases"); do echo "___$ent"; getent $ent; done
~~~

Regards,

pafernanr:
- Draft PR submited:
  - Some databases are not collected, most of them because they are already collected by other plugins.  
- About obfuscation. 
  - There is no need to parse the files, it is simpler to remove next ones.
    - getent_aliases
    - getent_group
  - Even easier: skip plugin execution. All collected data is susceptible for obfuscation IMO.

pmoravec:
>       * Even easier: skip plugin execution. All collected data is susceptible for obfuscation IMO.

But cleaner can run on a generated sosreport also. Rather we should treat both(?) files as sources of sensitive data to obfuscate in whole sosreport, likewise we do with lastlog and similar files in https://github.com/sosreport/sos/blob/main/sos/cleaner/preppers/usernames.py#L39 and below (maybe we would need to come up with a new branch like https://github.com/sosreport/sos/blob/main/sos/cleaner/preppers/usernames.py#L59).

PR #4291, opened 2026-04-14
[getent] New plugin draft

pafernanr:
Plugin draft PR for issue "Collect user groups details" #4272 

---

### DIFF
--- a/sos/report/plugins/getent.py
+++ b/sos/report/plugins/getent.py
@@ -1,0 +1,33 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import IndependentPlugin
+from sos.report.plugins import Plugin
+
+
+class Getent(Plugin, IndependentPlugin):
+
+    short_desc = 'query Name Service Switch (NSS) databases'
+
+    plugin_name = "getent"
+    profiles = ('system',)
+    verify_packages = ('glibc-common', 'libc-bin', 'glibc')
+
+    def setup(self):
+
+        # For reference. This databases are not collected.
+        # ["ahosts", "ahostsv4", "ahostsv6", "ethers", "gshadow", "hosts",
+        # "networks", "passwd", "services", "shadow"]
+
+        # Collected databases
+        databases = ["aliases", "group", "netgroup", "protocols", "rpc"]
+
+        for db in databases:
+            self.add_cmd_output(f"getent {db}")
+
+# vim: set et ts=4 sw=4 :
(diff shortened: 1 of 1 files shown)


sos/report/plugins/getent.py
```diff
@@ -0,0 +1,33 @@
...
+from sos.report.plugins import IndependentPlugin
+from sos.report.plugins import Plugin
+
+
+class Getent(Plugin, IndependentPlugin):
+
+    short_desc = 'query Name Service Switch (NSS) databases'
+
+    plugin_name = "getent"
+    profiles = ('system')
```
pmoravec, line 18: add comma to tell Python it is a tuple and not a str :)

```
>>> profiles = ('system')
>>> profiles.__class__
<class 'str'>
>>> profiles = ('system', )
>>> profiles.__class__
<class 'tuple'>
>>> 
```

The `conformance_tests.py` CI tests are failing on this.

sos/report/plugins/getent.py
```diff
@@ -0,0 +1,33 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import IndependentPlugin
+from sos.report.plugins import Plugin
```
arif-ali: you could do this instead

```python
from sos.report.plugins import IndependentPlugin, Plugin
```

sos/report/plugins/getent.py
```diff
@@ -0,0 +1,33 @@
...
+
+    plugin_name = "getent"
+    profiles = ('system',)
+    verify_packages = ('glibc-common', 'libc-bin', 'glibc')
+
+    def setup(self):
+
+        # For reference. This databases are not collected.
+        # ["ahosts", "ahostsv4", "ahostsv6", "ethers", "gshadow", "hosts",
+        # "networks", "passwd", "services", "shadow"]
```
arif-ali: we could collect the `passwd` as well I think, I don't see any reasons of not collecting it. It doesn't have sensitive data, so we should be ok I think

In-fact, if a system has sssd, and it has more users than what /etc/passwd has, then we can use sos clean against the users that `passwd` lists, which can actually help on another issue that I will be working on in the next couple of weeks.

arif-ali (commented):
a couple of comments

sos/report/plugins/getent.py
pafernanr: @arif-ali you're right. "passwd" db added

pmoravec (approved):

sos/report/plugins/getent.py
arif-ali: I didn't see this change :)

arif-ali (commented):
PR out of draft, so that it can be approved and merged if it is no longer WIP

pafernanr:
> I didn't see this change :)

@arif-ali Sorry I forgot that :sweat_smile:  Now it is fixed

arif-ali:
The PR is still in draft mode, so if it's no longer draft, mark it for review, and we can approve and merge

TurboTurtle (approved):

PR #4295, opened 2026-04-16
[auth] Add logic to store and retrieve auth tokens

jcastill:
This commit implements functionality to store and 
retrieve authorization tokens obtained remotely from a
distribution-defined authentication portal. This
enables users to reuse previously obtained tokens
that remain valid, eliminating the need to request a new token for each subsequent interaction.

Related: RHEL-167336, SUPDEV-186

---

### DIFF
--- a/sos/policies/auth/__init__.py
+++ b/sos/policies/auth/__init__.py
@@ -16,13 +16,15 @@ except ImportError:
     REQUESTS_LOADED = False
 import time
 from datetime import datetime, timedelta, timezone
+import os
+import json
 
-from sos.utilities import TIMEOUT_DEFAULT
+from sos.utilities import TIMEOUT_DEFAULT, path_join
 
 DEVICE_AUTH_CLIENT_ID = "sos-tools"
 GRANT_TYPE_DEVICE_CODE = "urn:ietf:params:oauth:grant-type:device_code"
-
-logger = logging.getLogger("sos")
+OIDC_TOKEN_FILE = ".sos-auth-oidc-token.json"
+DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
 
 
 class DeviceAuthorizationClass:
@@ -30,15 +32,86 @@ class DeviceAuthorizationClass:
     Device Authorization Class
     """
 
-    def __init__(self, client_identifier_url, token_endpoint):
+    def __init__(self, client_identifier_url, token_endpoint, token_dir):
 
         self._access_token = None
         self._access_expires_at = None
         self.__device_code = None
+        self._token_dir = f"{token_dir}/"
 
         self.client_identifier_url = client_identifier_url
         self.token_endpoint = token_endpoint
+        self.ui_log = logging.getLogger('sos_ui')
         self._use_device_code_grant()
+
+    def try_read_refresh_token(self, base_path):
+        """
+        Try to read locally stored refresh token
+
+        Args:
+            base_path (str): Local path where OIDC token is stored
+
+        Returns:
+            str: Returns OIDC refresh token information if found,
+            otherwise None
+        """
+        token_file_path = path_join(base_path, OIDC_TOKEN_FILE)
+        if not os.path.exists(token_file_path):
+            self.ui_log.error("Cannot find "
+                              f"{OIDC_TOKEN_FILE} file using {base_path} "
+                              "path, so we'll request a new token.")
+            return None
+
+        self.ui_log.info("Retrieving OIDC token information from local "
+                         f"{OIDC_TOKEN_FILE} file")
+        try:
+            with open(token_file_path, "r", encoding='utf-8') as fileobj:
+                read_in = fileobj.read()
+        except OSError as err:
+            self.ui_log.error("There was an exception while reading "
+                              f"the token file - {err}")
+            raise
+
+        if not read_in.strip():
+            self.ui_log.info(f"The {OIDC_TOKEN_FILE} file was empty.")
+            return None
+
+        try:
+            token_data = json.loads(read_in)
+        except json.JSONDecodeError:
+            self.ui_log.error("There was an issue decoding the JSON file "
+                              f"while loading {OIDC_TOKEN_FILE}")
+            return None
+
+        refresh_token = token_data.get("refresh_token")
+        dt_str = token_data.get("refresh_expires_at")
+        if not refresh_token or not dt_str:
+            self.ui_log.error(
+                "Locally cached offline token is missing required fields "
+                f"in {OIDC_TOKEN_FILE}. We will request a new token."
+            )
+            return None
+
+        try:
+            refresh_expires_at = datetime.strptime(dt_str, DATETIME_FORMAT)
+            refresh_expires_at = refresh_expires_at.replace(
+                tzinfo=timezone.utc
+            )
+        except ValueError:
+            self.ui_log.error(
+                f"Invalid refresh_expires_at in {OIDC_TOKEN_FILE}: "
+                f"{dt_str!r}"
+            )
+            return None
+
+        if refresh_expires_at - timedelta(seconds=300) <= \
+                datetime.now(timezone.utc):
+            self.ui_log.error("Locally cached offline token has expired "
+                              "or is incorrectly formatted. "
+                              "We will request a new token.")
+            return None
+
+        return refresh_token
 
     def _use_device_code_grant(self):
         """
@@ -46,13 +119,16 @@ class DeviceAuthorizationClass:
         store the tokens in an in-memory keyring.
 
         """
-
-        self._request_device_code()
-        print(
-            "Please visit the following URL to authenticate this"
-            f" device: {self._verification_uri_complete}"
-        )
-        self.poll_for_auth_completion()
+        stored_refresh_token = self.try_read_refresh_token(self._token_dir)
+        if not stored_refresh_token:
+            self._request_device_code()
+            self.ui_log.info(
+                "Please visit the following URL to authenticate this"
+                f" device: {self._verification_uri_complete}"
+            )
+            self.poll_for_auth_completion()
+        else:
+            self._use_refresh_token_grant(stored_refresh_token)
 
     def _request_device_code(self):
         """
@@ -108,7 +184,7 @@ class DeviceAuthorizationClass:
                 status_code = check_auth_completion.status_code
 
                 if status_code == 200:
-                    logger.info("The SSO authentication is successful")
+                    self.ui_log.info("The SSO authentication is successful")
                     self._set_token_data(check_auth_completion.json())
                 if status_code not in [200, 400]:
                     raise Exception(status_code, check_auth_completion.text)
@@ -117,13 +193,13 @@ class DeviceAuthorizationClass:
                         ("authorization_pending", "slow_down"):
                     raise Exception(status_code, check_auth_completion.text)
             except requests.exceptions.RequestException as e:
-                logger.error(f"Error was found while posting a request: {e}")
+                self.ui_log.error("An error was found while posting a request:"
+                                  f" {e}")
 
     def _set_token_data(self, token_data):
         """
         Set the class attributes as per the input token_data received.
-        In the future we will persist the token data in a local,
-        in-memory keyring, to avoid visting the browser frequently.
+
         :param token_data: Token data containing access_token, refresh_token
         and their expiry etc.
         """
@@ -137,6 +213,8 @@ class DeviceAuthorizationClass:
         else:
             self._refresh_expires_at = datetime.now(timezone.utc) + \
                 timedelta(seconds=self._refresh_expires_in)
+        self.persist_refresh_token(self._token_dir)
+        self.ui_log.info(f"Token data stored in: {self._token_dir}")
 
     def get_access_token(self):
         """
@@ -199,10 +277,11 @@ class DeviceAuthorizationClass:
 
         elif refresh_token_res.status_code == 400 and 'invalid' in\
                 refresh_token_res.json()['error']:
-            logger.warning("Problem while fetching the new tokens from refresh"
-                           f" token grant - {refresh_token_res.status_code} "
-                           f"{refresh_token_res.json()['error']}."
-                           " New Device code will be requested !")
+            self.ui_log.warning("Problem while fetching the new tokens "
+                                f"from refresh token grant "
+                                f"- {refresh_token_res.status_code} "
+                                f"{refresh_token_res.json()['error']}."
+                                " New Device code will be requested!")
             self._use_device_code_grant()
         else:
             raise Exception(
@@ -210,3 +289,51 @@ class DeviceAuthorizationClass:
                 "Refresh token grant for fetching tokens:"
                 f" Returned status code {refresh_token_res.status_code}"
                 f" and error {refresh_token_res.json()['error']}")
+
+    def request_new_device_code(self):
+        """
+        Initialize new Device Authorization Grant attempt by requesting
+          a new device code.
+        """
+        self._use_device_code_grant()
+
+    def persist_refresh_token(self, base_path):
+        """
+        Persist current refresh token to a local file
+        Args:
+            base_path (str): Local path to a directory where token
+            should be stored
+
+        Returns:
+            bool: True if refresh token was successfully persisted,
+            otherwise False
+        """
+        if self.is_refresh_token_valid():
+            # ToDo: While now we'll use the user's home directory
+            # in the future we may want to offer the possibility
+            # of specifying the directory (making sure the user knows
+            # the security implications of this decision)
+            if not os.path.exists(base_path):
+                os.mkdir(base_path)
+            token_data = {
+                "refresh_token": self._refresh_token,
+                "refresh_expires_at": self._refresh_expires_at.strftime(
+                    DATETIME_FORMAT
+                    )
+            }
+            token_file_path = path_join(base_path, OIDC_TOKEN_FILE)
+            fd = os.open(
+                token_file_path,
+                os.O_CREAT | os.O_WRONLY | os.O_TRUNC,
+                0o600,
+            )
+            with os.fdopen(fd, 'w', encoding='utf-8') as fileobj:
+                json.dump(token_data, fileobj)
+            os.chmod(token_file_path, 0o600)
+            self.ui_log.info("The new refresh token was successfully saved to"
+                             f" {token_file_path}")
+            return True
+        self.ui_log.error("Cannot save invalid refresh token")
+        return False
+
+# vim: set et ts=4 sw=4 :
--- a/sos/policies/auth/__init__.py
+++ b/sos/policies/auth/__init__.py
@@ -16,13 +16,15 @@ except ImportError:
     REQUESTS_LOADED = False
 import time
 from datetime import datetime, timedelta, timezone
+import os
+import json
 
-from sos.utilities import TIMEOUT_DEFAULT
+from sos.utilities import TIMEOUT_DEFAULT, path_join
 
 DEVICE_AUTH_CLIENT_ID = "sos-tools"
 GRANT_TYPE_DEVICE_CODE = "urn:ietf:params:oauth:grant-type:device_code"
-
-logger = logging.getLogger("sos")
+OIDC_TOKEN_FILE = ".sos-oidc-token.json"
+DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
 
 
 class DeviceAuthorizationClass:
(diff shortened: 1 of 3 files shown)


sos/policies/auth/__init__.py
```diff
@@ -16,43 +16,119 @@
...
+import json
 
-from sos.utilities import TIMEOUT_DEFAULT
+from sos.utilities import TIMEOUT_DEFAULT, path_join
 
 DEVICE_AUTH_CLIENT_ID = "sos-tools"
 GRANT_TYPE_DEVICE_CODE = "urn:ietf:params:oauth:grant-type:device_code"
-
-logger = logging.getLogger("sos")
+OIDC_TOKEN_FILE = ".sos-auth-oidc-token.json"
```
jcastill, line 26: I'm open for other names for the token. I chose this one to be consistent with other tools that use this approach, but it's really a mouthful. Could remove 'oidc' perhaps? Or add a distro-based string, like '.sos-auth-redhat.json'?

sos/upload/targets/redhat.py
```diff
@@ -111,7 +112,8 @@ def _upload_https_post(self, archive, verify=True):
         # longer than the token timeout
         RHELAuth = DeviceAuthorizationClass(
                 self.client_identifier_url,
-                self.token_endpoint
+                self.token_endpoint,
+                Path.home()
```
jcastill, line 116: I used Path.home() to be consistent with what we do in other places, like in component.py and the code in collector

jcastill, line 116: The use of the home directory was suggested internally by our sec team. Other options considered were /run/user, and $XDG_RUNTIME_DIR (but this environment variable is not always set).

sos/policies/auth/__init__.py
```diff
@@ -199,14 +277,63 @@ def _use_refresh_token_grant(self, refresh_token=None):
...
+            }
+            token_file_path = path_join(base_path, OIDC_TOKEN_FILE)
+            fd = os.open(
+                token_file_path,
+                os.O_CREAT | os.O_WRONLY | os.O_TRUNC,
+                0o600,
+            )
+            with os.fdopen(fd, 'w', encoding='utf-8') as fileobj:
+                json.dump(token_data, fileobj)
+            os.chmod(token_file_path, 0o600)
```
jcastill, line 332: Tried to make the token file permissions as restrictive as possible. Any suggestion or other layers on top of this to improve it is welcome!

jcastill:
@pmoravec @arif-ali I'd be very grateful if you can start the review of this PR. I'll move it out of draft after RH QE finish their review as well

sos/policies/auth/__init__.py
pmoravec: 2c of non-native English speaker: `auth` is redundant, `redhat` would make sense rather within redhat.py target, dont bloat the parent class / code.

sos/policies/auth/__init__.py
pmoravec: Isn't the `os.chmod` a dup of `os.open(..0o600)`?

sos/policies/auth/__init__.py
```diff
@@ -199,14 +277,63 @@ def _use_refresh_token_grant(self, refresh_token=None):
...
                 "Refresh token grant for fetching tokens:"
                 f" Returned status code {refresh_token_res.status_code}"
                 f" and error {refresh_token_res.json()['error']}")
+
+    def request_new_device_code(self):
+        """
+        Initialize new Device Authorization Grant attempt by requesting
+          a new device code.
+        """
+        self._use_device_code_grant()
```
pmoravec: A dead code / method?

sos/policies/auth/__init__.py
```diff
@@ -16,43 +16,119 @@
...
+            )
+            return None
+
+        if refresh_expires_at - timedelta(seconds=300) <= \
+                datetime.now(timezone.utc):
+            self.ui_log.error("Locally cached offline token has expired "
+                              "or is incorrectly formatted. "
+                              "We will request a new token.")
+            return None
+
```
pmoravec, line 117: Idea: in either case when returning `None`, does it make sense to delete the token file as it is not of further use and might contain sensitive data (from past)? Or is that too obscure?

sos/policies/auth/__init__.py
```diff
@@ -199,14 +277,63 @@ def _use_refresh_token_grant(self, refresh_token=None):
...
-            logger.warning("Problem while fetching the new tokens from refresh"
-                           f" token grant - {refresh_token_res.status_code} "
-                           f"{refresh_token_res.json()['error']}."
-                           " New Device code will be requested !")
+            self.ui_log.warning("Problem while fetching the new tokens "
+                                f"from refresh token grant "
+                                f"- {refresh_token_res.status_code} "
+                                f"{refresh_token_res.json()['error']}."
+                                " New Device code will be requested!")
             self._use_device_code_grant()
```
pmoravec: Adding line 131 (`_use_device_code_grant` calling `_use_refresh_token_grant` again) can trigger a recursion.

Am I right one loop of the recursion contains a new user activity in Web browser, approving the request? If so, this is fine. If no user activity within the loop, I would add some delay / sleep.

EDIT: I think it is sufficient to have invalid token and we enter the recursion without human/user touch. We should remove invalid token file, then.

sos/policies/auth/__init__.py
```diff
@@ -16,43 +16,119 @@
...
+        Args:
+            base_path (str): Local path where OIDC token is stored
+
+        Returns:
+            str: Returns OIDC refresh token information if found,
+            otherwise None
+        """
+        token_file_path = path_join(base_path, OIDC_TOKEN_FILE)
+        if not os.path.exists(token_file_path):
+            self.ui_log.error("Cannot find "
```
pmoravec: s/error/info/ since this is very usual scenario for the very first run when no token file exists yet.

sos/policies/auth/__init__.py
jcastill: Ideally that data should not be sensitive anymore, and once you get a new token it will be rewritten. I'm also a bit reticent about deleting stuff, but I see your point. Let me think about it a bit more, and lets have Arif's thoughts as well.

sos/policies/auth/__init__.py
jcastill: Indeed, a leftover from my previous attempt. Will remove it in the next iteration

sos/policies/auth/__init__.py
jcastill: I'll create a new function to remove the invalid token

sos/policies/auth/__init__.py
jcastill: I added it to be on the safe side. os.open() covers if we create a new file, but if the file exists and the permissions have somehow changed, then os.chmod() will ensure we go back to the original permissions. Do you think this approach is too strict?

sos/policies/auth/__init__.py
jcastill: Changed, thank you!

sos/policies/auth/__init__.py
jcastill: Actually as you mentioned it being useful below, I'll do this, I think it's a fantastic idea

jcastill:
And new version, this time removing the old or faulty token file. I think I've covered all cases where we should remove it, but let me know if I missed any

sos/policies/auth/__init__.py
jcastill: I've modified both functions so now we force requesting the device code if we get an invalid refresh, and also removing the cached token

sos/policies/auth/__init__.py
```diff
@@ -199,14 +281,69 @@ def _use_refresh_token_grant(self, refresh_token=None):
...
+            bool: True if refresh token was successfully persisted,
+            otherwise False
+        """
+        if self.is_refresh_token_valid():
+            # ToDo: While now we'll use the user's home directory
+            # in the future we may want to offer the possibility
+            # of specifying the directory (making sure the user knows
+            # the security implications of this decision)
+            if not os.path.exists(base_path):
+                os.mkdir(base_path)
```
pmoravec: Worth ensuring default permissions set (when creating (by default) home dir, or whatever-else-equivalent-dir)? Or is it sufficient to rely on umask? (I really dont know)

sos/policies/auth/__init__.py
```diff
@@ -16,43 +16,123 @@
...
+                f"Invalid refresh_expires_at in {OIDC_TOKEN_FILE}: "
+                f"{dt_str!r}"
+            )
+            os.remove(f"{token_file_path}")
+            return None
+
+        if refresh_expires_at - timedelta(seconds=300) <= \
+                datetime.now(timezone.utc):
+            self.ui_log.error("Locally cached offline token has expired "
+                              "or is incorrectly formatted. "
```
pmoravec: I think incorrectly formatted token would be caught above already, and either `return None` is invoked. We should get here only with expired token, hence `or is incorrectly formatted` is redundant and rather misleading?

sos/policies/auth/__init__.py
```diff
@@ -16,43 +16,123 @@
...
+            self.try_read_refresh_token(self._token_dir)
+        if force or not stored_refresh_token:
+            self._request_device_code()
+            self.ui_log.info(
+                "Please visit the following URL to authenticate this"
+                f" device: {self._verification_uri_complete}"
+            )
+            self.poll_for_auth_completion()
+        else:
+            self._use_refresh_token_grant(stored_refresh_token)
```
pmoravec, line 145: Not fully related here but Claude AI review suggests this method calls `refresh_token_res.json()` multiple times - it is better to store that as a json object than iteratively call the method.

Shall I raise a simple PR for that or can you extend it within this PR?

pmoravec (commented):
Just those minor issues spot, LGTM otherwise.

sos/policies/auth/__init__.py
jcastill: I thought it's sufficient to rely on umask. Bear in mind that right now I'm using the home directory and that's in most cases root's one. So I think I rather not touch the dir at all. 
But if we think that it's better to have something like /root/.sos/<token file> then I can rework the permissions to make sure .sos is correct

sos/policies/auth/__init__.py
jcastill: It's misleading, you are right. If it's incorrectly formatted, it should have been picked up in the previous block. I'll modify this error message

sos/policies/auth/__init__.py
jcastill: I can do it in this PR. Thank you Pavel and Claude!

pmoravec:
LGTM!

jcastill:
@pmoravec thank you for the review! Can you add your ack if it's ok with you?
@arif-ali could you review this PR when you have a minute please?

pmoravec (approved):

pmoravec:
> @pmoravec thank you for the review! Can you add your ack if it's ok with you? @arif-ali could you review this PR when you have a minute please?

I wasnt sure if the Draft status means you plan to further change something, hence no ACK given. Assuming the PR is complete, ACK to it.

pmoravec:
Is there a reason to keep the PR in Draft? I think this blocks the others to review and potentially merge it..

jcastill:
> Is there a reason to keep the PR in Draft? I think this blocks the others to review and potentially merge it..

The idea was to give RH QE the opportunity to review it and run their tests and block merging until that was done even with all the acks, but I can move it out of draft to get the second ack.

sos/policies/auth/__init__.py
```diff
@@ -16,43 +16,122 @@
...
+import json
 
-from sos.utilities import TIMEOUT_DEFAULT
+from sos.utilities import TIMEOUT_DEFAULT, path_join
 
 DEVICE_AUTH_CLIENT_ID = "sos-tools"
 GRANT_TYPE_DEVICE_CODE = "urn:ietf:params:oauth:grant-type:device_code"
-
-logger = logging.getLogger("sos")
+OIDC_TOKEN_FILE = ".sos-redhat-oidc-token.json"
```
TurboTurtle: Since this is used in the base, and potentially used elsewhere (granted, in a future version of sos), we shouldn't be vendoring the name here. If it needs to be vendored, do so in the vendor-specific implementation (e.g. `sos.upload.targets.redhat`).

sos/policies/auth/__init__.py
```diff
@@ -16,43 +16,122 @@
...
+            return None
+
+        refresh_token = token_data.get("refresh_token")
+        dt_str = token_data.get("refresh_expires_at")
+        if not refresh_token or not dt_str:
+            self.ui_log.error(
+                "Locally cached offline token is missing required fields "
+                f"in {OIDC_TOKEN_FILE}. We will request a new token."
+            )
+            os.remove(f"{token_file_path}")
```
TurboTurtle: `token_file_path` is already a string, isn't it? Then using an f-string here is redundant.

sos/policies/auth/__init__.py
```diff
@@ -193,20 +274,79 @@ def _use_refresh_token_grant(self, refresh_token=None):
...
+            }
+            token_file_path = path_join(base_path, OIDC_TOKEN_FILE)
+            fd = os.open(
+                token_file_path,
+                os.O_CREAT | os.O_WRONLY | os.O_TRUNC,
+                0o600,
+            )
+            with os.fdopen(fd, 'w', encoding='utf-8') as fileobj:
+                json.dump(token_data, fileobj)
+            os.chmod(token_file_path, 0o600)
```
TurboTurtle: This is redundant in a few places. The same behavior would be achieved via:

```python
with open(token_file_path, "w", encoding="utf-8") as fileobj:
    os.chmod(token_file_path, 0o600)
    json.dump(token_data, fileobj)
```

The `chmod()` is always called, and the behavior of  ` mode='w'` for `open()` matches the flags being set in the fd creation anyway.

sos/upload/targets/redhat.py
TurboTurtle, line 116: ack, homedir sounds reasonable for this kind of operation.

sos/policies/auth/__init__.py
jcastill: Ack, I should have moved it within the redhat class after Pavel's comment when I asked before. I'll go for the simpler name for now

sos/policies/auth/__init__.py
jcastill: Fixed

sos/policies/auth/__init__.py
jcastill: That's a very good point. I've taken your recommendation and extended it a little bit to avoid TOCTOU while opening the file, hopefully it looks bettew now.

jcastill:
Some changes:
- Implemented the changes requested by Jake, hopefully I didn't miss anything.
- Added explicit permissions to os.makedir(), although it may be overkill, but better safe than sorry.
- Created a new function self._remove_token_file to deal better with attempts to remove token file.
- And a couple of other changes dealing with issues when storing the token. 

Mirek will begin testing soon so I expect him to find problems that I didn't think about

sos/policies/auth/__init__.py
```diff
@@ -16,43 +16,133 @@
...
+                f"Unable to remove cached token file {token_file_path}: {err}"
+            )
+
+    def _use_device_code_grant(self, force=False):
+        """
+        Start the device auth flow.
+        """
+        stored_refresh_token = None if force else \
+            self.try_read_refresh_token(self._token_dir)
+        if force or not stored_refresh_token:
```
pmoravec: Technically, `force=True` implies `not stored_refresh_token`, hence a test `if not stored_refresh_token:` would be sufficient. But let it be if this would be the only comment.

sos/policies/auth/__init__.py
```diff
@@ -193,20 +288,70 @@ def _use_refresh_token_grant(self, refresh_token=None):
...
+                        )
+                }
+                token_file_path = path_join(base_path, OIDC_TOKEN_FILE)
+                fd = os.open(token_file_path,
+                             os.O_CREAT | os.O_WRONLY | os.O_TRUNC,
+                             0o600)
+                with os.fdopen(fd, 'w', encoding='utf-8') as fileobj:
+                    json.dump(token_data, fileobj)
+                self.ui_log.info("The new refresh token was successfully"
+                                 f"saved to {token_file_path}")
```
pmoravec: Please add missing space into `successfullysaved`.

sos/policies/auth/__init__.py
```diff
@@ -137,6 +227,11 @@ def _set_token_data(self, token_data):
         else:
             self._refresh_expires_at = datetime.now(timezone.utc) + \
                 timedelta(seconds=self._refresh_expires_in)
+        if not self.persist_refresh_token(self._token_dir):
+            self.ui_log.warning("Token obtained but could not"
+                                f"be persisted locally in {self._token_dir}")
+        else:
+            self.ui_log.info(f"Token data stored in: {self._token_dir}")
```
pmoravec:
```
The new refresh token was successfullysaved to /root/.sos-oidc-token.json
Token data stored in: /root/
```

Is there a reason to log the file once but the corresponding directory in second log? (I would prefer having this coherent, until there is some rationale I miss)

pmoravec:
Just those very minor comments, otherwise it works great!

sos/policies/auth/__init__.py
jcastill: No other reason than me going a little overboard with logging, which leads to a bunch of repeated info most of the time, sorry
I'll push a change with these two lines removed

sos/policies/auth/__init__.py
jcastill: Fixed

sos/policies/auth/__init__.py
jcastill: Fixed

sos/policies/auth/__init__.py
```diff
@@ -16,43 +16,133 @@
...
 
-        self._request_device_code()
-        print(
-            "Please visit the following URL to authenticate this"
-            f" device: {self._verification_uri_complete}"
-        )
-        self.poll_for_auth_completion()
+        if not read_in.strip():
+            self.ui_log.info(f"The {OIDC_TOKEN_FILE} file was empty.")
+            return None
```
pmoravec, line 77: To be coherent with other "token failed to be read, return None" branches, we should call here `self._remove_token_file(token_file_path)`. But from user perspective, it is not needed (here or on a few other places), since such file is overriden by a new valid token.

So I am ok with it as is.

sos/policies/auth/__init__.py
```diff
@@ -137,6 +227,9 @@ def _set_token_data(self, token_data):
         else:
             self._refresh_expires_at = datetime.now(timezone.utc) + \
                 timedelta(seconds=self._refresh_expires_in)
+        if not self.persist_refresh_token(self._token_dir):
+            self.ui_log.warning("Token obtained but could not"
+                                f"be persisted locally in {self._token_dir}")
```
pmoravec, line 232: Nitpick: add space into `notbe` (and apologize for iteratively finding these nitpicks :( ).

pmoravec:
I apologize for yet another never-ending iterative comments, but storing passwords in plaintext, even though these relatively short-lived and within user home dir, it might violate some corporate policies.We should let users know about it (by commenting in man pages, e.g.?) and/or test if home dir or the file itself has proper permissions, like:

```
# Check if file permissions are too permissive
stat_info = os.stat(token_file_path)
if stat_info.st_mode & 0o077:  # Group or other has any access
    self.ui_log.warning(f"Token file {token_file_path} has overly permissive permissions!")
```

? Just a thought..

sos/policies/auth/__init__.py
jcastill: ack, lets leave it this way since as you mentioned the file is overriden

sos/policies/auth/__init__.py
jcastill: Will fix this in the next push

jcastill:
> I apologize for yet another never-ending iterative comments, but storing passwords in plaintext, even though these relatively short-lived and within user home dir, it might violate some corporate policies.We should let users know about it (by commenting in man pages, e.g.?) and/or test if home dir or the file itself has proper permissions, like:
> 
> ```
> # Check if file permissions are too permissive
> stat_info = os.stat(token_file_path)
> if stat_info.st_mode & 0o077:  # Group or other has any access
>     self.ui_log.warning(f"Token file {token_file_path} has overly permissive permissions!")
> ```
> 
> ? Just a thought..

We are not storing passswords, we are storing tokens, and as you mentioned these are short-lived at the moment and potentially even more in the future. 
Also, since sos runs as root, the home dir is always /root. I can add the check for dir permissions if you still think is necessary, or add it in the future if/when we can either specify the home dir if sos was called via sudo, or if we add an option to specify a different directory to store the token. 
About corporate policies, do you mean the end user corporate policies or RH's ones? Because if it's about RH ones, we had a thread some time ago and prodsec gave us their blessing with this approach.

pmoravec:
I meant end user corporate policies, than can vary a lot. We might warn about the storing policy to let users be aware..? Not sure how big sign is enough.. Anyway, I am ok with the PR as is now (up to the typo(s)).

jcastill:
OK so I've added an entry in the man page, the warning you mentioned, and fixed the remaining typo/space issue. Please check the content of the man page because as always I'm not 100% sure of the wording.

pmoravec (approved):

TurboTurtle (approved):
Code looks good to me at this point. I don't really have the capacity to test this locally myself, but I'm assuming there's been some RH-internal validation.

Ack, green light to merge pending sanity-confirmation of such testing.

pmoravec:
I have tested that from functionality point of view, all good.

PR #4296, opened 2026-04-16
[mysql_router] Add dedicated MySQL Router plugin

munirsidd:
[mysql_router] Add dedicated MySQL Router plugin with added support for mysql-router conf and log collection

1. In the current implementation, MySQL Router collection is embedded in the `mysql` plugin, which is gated on MySQL/MariaDB server packages. As a result, on nodes with only MySQL Router installed, the `mysql` plugin never activates, so no router config or logs are collected.
2. The current implementation does not collect conf and logs from `/var/lib/mysql/*-mysql-router/`.

- Remove all `mysqlrouter` references from the `mysql` plugin
- Extract MySQL Router into a new `mysql_router` plugin that:
    - Gates on `mysql-router` / `mysql-router-community` packages
    - Collects information from `/var/lib/mysql/*-mysql-router/` in addition to the previously collected data
---

- [N/A] Are all passwords or private data gathered by this PR [obfuscated](https://github.com/sosreport/sos/wiki/How-to-Write-a-Plugin#how-to-prevent-collecting-passwords)?

### DIFF
--- a/sos/report/plugins/mysql_router.py
+++ b/sos/report/plugins/mysql_router.py
@@ -1,0 +1,45 @@
+# Copyright (C) 2026 Canonical Ltd.,
+# Munir Siddiqui <munir.siddiqui@canonical.com>
+#
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.report.plugins import Plugin, IndependentPlugin
+
+
+class MysqlRouter(Plugin, IndependentPlugin):
+    """
+    Collects configuration and logs for MySQL Router.
+    """
+
+    short_desc = "MySQL Router"
+
+    plugin_name = "mysql_router"
+    profiles = ("services",)
+    packages = ("mysql-router", "mysql-router-community",)
+
+    def setup(self):
+        super().setup()
+
+        self.add_copy_spec([
+            "/etc/mysqlrouter/",
+            "/var/lib/mysql/*-mysql-router/mysqlrouter.conf",
+        ])
+
+        if self.get_option("all_logs"):
+            self.add_copy_spec([
+                "/var/log/mysqlrouter/*",
+                "/var/lib/mysql/*-mysql-router/log/*",
+            ])
+        else:
+            self.add_copy_spec([
+                "/var/log/mysqlrouter/mysqlrouter.log",
+                "/var/lib/mysql/*-mysql-router/log/mysqlrouter.log*",
+            ])
+
+# vim: set et ts=4 sw=4 :
--- a/sos/report/plugins/mysql.py
+++ b/sos/report/plugins/mysql.py
@@ -35,7 +35,6 @@ class Mysql(Plugin):
 
         self.add_copy_spec([
             self.mysql_cnf,
-            "/etc/mysqlrouter/",
             "/var/lib/mysql/grastate.dat",
             "/var/lib/mysql/gvwstate.dat"
         ])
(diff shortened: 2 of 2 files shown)


sos/report/plugins/mysql_router.py
```diff
@@ -0,0 +1,55 @@
...
+    plugin_name = "mysql_router"
+    profiles = ("services",)
+    packages = ("mysql-router", "mysql-router-community",)
+
+    def setup(self):
+        super().setup()
+
+        self.add_copy_spec([
+            "/etc/mysqlrouter/",
+            "/var/lib/mysql/*-mysql-router/mysqlrouter.conf",
```
arif-ali, line 31: is it worth collecting the full folder, and just not the mysqlrouter.conf?

sos/report/plugins/mysql_router.py
```diff
@@ -0,0 +1,55 @@
...
+        ])
+
+        if self.get_option("all_logs"):
+            self.add_copy_spec([
+                "/var/log/mysqlrouter/*",
+                "/var/lib/mysql/*-mysql-router/log/*"
```
arif-ali, line 37: minor nit, add comma at the end

sos/report/plugins/mysql_router.py
```diff
@@ -0,0 +1,55 @@
...
+                "/var/lib/mysql/*-mysql-router/log/mysqlrouter.log*",
+            ])
+
+    def postproc(self):
+        protect_keys = ['password']
+        regex = fr"(^\s*({'|'.join(protect_keys)})\s*=\s*)(.*)"
+        sub = r"\1*********"
+
+        self.do_path_regex_sub(
+            "/etc/mysqlrouter/*",
```
arif-ali, line 51: we may need to also add `/var/lib/mysql/*-mysql-router/`  as well, just in case there is a password in there as well? This file wasn't being obfuscated in the `mysql` plugin, so is this needed here?

arif-ali (changes requested):
A few comments below

sos/report/plugins/mysql_router.py
munirsidd: I don't think the other contents of the the folder (`data/`, `run/`, `start.sh`, `stop.sh`) are too relevant for troubleshooting.

Also, collecting just the logs and conf explicitly might be the safer approach as the folder does contain sensitive files such as `mysqlrouter.key` and `/data/keyring`.

sos/report/plugins/mysql_router.py
munirsidd: Correct, we should add `/var/lib/mysql/*-mysql-router/`, and `/etc/mysqlrouter/` isn't needed

sos/report/plugins/mysql_router.py
munirsidd: Added

sos/report/plugins/mysql_router.py
munirsidd: Rethinking about it: since we're only collecting `log/` and `conf`, and those don't have passwords, I don't think we need the `postproc()` function

arif-ali (approved):

sos/report/plugins/mysql_router.py
```diff
@@ -0,0 +1,45 @@
...
+    plugin_name = "mysql_router"
+    profiles = ("services",)
+    packages = ("mysql-router", "mysql-router-community",)
+
+    def setup(self):
+        super().setup()
+
+        self.add_copy_spec([
+            "/etc/mysqlrouter/",
+            "/var/lib/mysql/*-mysql-router/mysqlrouter.conf",
```
pmoravec, line 31: No credentials or auth tokens in either config..?

munirsidd: Correct, the [mysql-router docs](https://dev.mysql.com/doc/mysql-router/8.0/en/mysql-router-conf-options.html) doesn't suggest any sensitive options in mysqlrouter config

TurboTurtle (approved):

PR #4305, opened 2026-04-20
[Anbox]: Add Anbox plugin

goldberl:
**Add Anbox Plugin**

Create a new plugin for Anbox Cloud (anbox-cloud-appliance) to collect service status, journal logs, and AMC output. The plugin targets all snap-based Anbox Cloud services and includes exclusions for sensitive credential paths.

---

### DIFF
--- a/sos/report/plugins/anbox.py
+++ b/sos/report/plugins/anbox.py
@@ -1,0 +1,131 @@
+# Copyright (C) 2026 Canonical Ltd.,
+# Leah Goldberg <leah.goldberg@canonical.com>
+#
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+import json
+
+from sos.report.plugins import Plugin, IndependentPlugin, PluginOpt
+
+
+class Anbox(Plugin, IndependentPlugin):
+    """
+    Collects diagnostic information for Anbox.
+    """
+
+    short_desc = "Anbox/Anbox Cloud diagnostics"
+
+    plugin_name = "anbox"
+    profiles = ("container", "system")
+    packages = ('anbox-cloud-appliance',)
+
+    services = (
+        "snap.anbox-cloud-appliance.ams",
+        "snap.anbox-cloud-appliance.dashboard",
+        "snap.anbox-cloud-appliance.daemon",
+        "snap.anbox-cloud-appliance.stream-gateway",
+        "snap.anbox-cloud-appliance.stream-agent",
+        "snap.anbox-cloud-appliance.nats-server",
+        "snap.anbox-cloud-appliance.coturn",
+        "snap.anbox-cloud-appliance.openfga",
+    )
+
+    option_list = [
+        PluginOpt('collectdb', default=False,
+                  desc='Collect Anbox database file'),
+    ]
+
+    def setup(self):
+
+        self.add_cmd_output([
+            "amc addon list",
+            "amc application list",
+            "amc auth group list",
+            "amc auth identity list",
+            "amc auth permissions list",
+            "amc config show",  # may have sensitive data
+            "amc image list",
+            "amc info",  # may have sensitive data
+            "amc node list",
+        ])
+
+        try:
+            # Get all AMC instances (JSON)
+            result = self.exec_cmd("amc list --format json")
+
+            try:
+                instances = json.loads(result["output"])
+            except Exception:
+                self._log_error("Failed to parse amc list JSON output")
+                return
+
+            for inst in instances:
+                inst_id = inst.get("id")
+
+                # Skip invalid entries
+                if not inst_id:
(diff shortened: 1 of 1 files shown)


sos/report/plugins/anbox.py
```diff
@@ -0,0 +1,57 @@
...
+class Anbox(Plugin, IndependentPlugin):
+    """
+    Collects diagnostic information for Anbox.
+    """
+
+    short_desc = "Anbox/Anbox Cloud diagnostics"
+
+    plugin_name = "anbox"
+    profiles = ("container", "system")
+
```
arif-ali, line 25: if it's a snap, you could just add `packages = ('anbox-cloud-applicance',)`

sos/report/plugins/anbox.py
```diff
@@ -0,0 +1,57 @@
...
+    )
+
+    def setup(self):
+
+        self.add_cmd_output([
+            "amc application list",
+            "amc ls",
+        ])
+
+        for service in self.get_service_names(
```
arif-ali: normally, if all the services are already listed  in the `saervices` tuple, then there is no need to go through all the services again, as  the the service and journal for the services are automatically collected.

sos/report/plugins/anbox.py
```diff
@@ -0,0 +1,57 @@
...
+        ):
+            self.add_journal(service)
+            self.add_service_status(service)
+
+        self.add_copy_spec([
+            "/var/snap/anbox-cloud-appliance",
+        ])
+
+        self.add_forbidden_path([
+            "/var/lib/ubuntu-advantage/private/anbox-cloud-credentials",
```
arif-ali: if we're not collecting `/var/lib/ubuntu-advantage/`, then is there a need to forbid this? so curious why this is needed?

arif-ali (changes requested):
a few quick comments

sos/report/plugins/anbox.py
goldberl: Great question, I thought I saw that file was collected when I was testing the plugin, but maybe I was just looking in the root folder of the machine. Checking again, you're right, it isn't collected so I will get rid of those lines.

sos/report/plugins/anbox.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def setup(self):
+
+        self.add_cmd_output([
+            "amc application list",
+            "amc ls",
+        ])
+
+        self.add_copy_spec([
+            "/var/snap/anbox-cloud-appliance",
```
arif-ali, line 45: and there are no certs or config files that have passwords? just making sure ;)

sos/report/plugins/anbox.py
```diff
@@ -0,0 +1,48 @@
...
+        ])
```
jcastill, line 46: Do we need to capture any logs for this tool? Or will they be captured in other plugins like snap or juju, perhaps?

sos/report/plugins/anbox.py
goldberl: As we saw, there is only the db file in this /var/snap/anbox-cloud-appliance directory and we may not even need to collect it

sos/report/plugins/anbox.py
goldberl: That's a good question, I will do some more research and find some other commands that would be good to collect logs from

goldberl:
This plugin gets all the logs for all the AMC instanaces.

Additionally, it will obfuscate sensitive data from `amc config show` and `amc info`

For example:
```
ubuntu@sandbox:~/anbox-sos8/sos_commands/anbox$ cat amc_config_show 
config:
  agent.api.fingerprint: ******
  agent.api.token: ******
  agent.api.url: https://10.23.72.133:9033
  application.addons: ""
  application.auto_publish: true
  application.auto_update: true
  application.default_abi: ""
  application.max_published_versions: 3
  container.apt_mirror: ""
  container.default_platform: ""
  container.features: ""
  container.security_updates: true
  core.debug_address: ""
  core.https_address: 10.23.72.133:8444
  core.max_requests_per_second: "1000"
  core.proxy_http: ""
  core.proxy_https: ""
  core.proxy_ignore_hosts: ""
  core.trust_password: ******
  cpu.limit_mode: pinning
  gpu.type: none
  images.allow_insecure: false
  images.auth: true
  images.update_interval: 5m
  images.url: https://images.anbox-cloud.io/stable
  images.version_lockstep: true
  instance.apt_mirror: ""
  instance.default_platform: ""
  instance.features: ""
  instance.security_updates: "true"
  load_balancer.url: ""
  node.queue_size: 100
  node.workers_per_queue: 4
  oidc.allow_insecure: false
  oidc.audience: ""
  oidc.client_id: ""
  oidc.issuer_url: ""
  openfga.api.fingerprint: ******
  openfga.api.token: ******
  openfga.api.url: https://10.23.72.133:8085
  openfga.store.id: 01KPNMC0X4VN764ZEJ24EDG8N4
  registry.filter: ""
  registry.fingerprint: ******
  registry.mode: pull
  registry.update_interval: 1h
  registry.url: ""
  scheduler.strategy: spread
```

arif-ali (approved):

TurboTurtle (approved):

Issue #4288, opened 2026-04-10
Cirrus CI shutdown


pmoravec:
Per the announcement at https://cirruslabs.org/, Cirrus CI shuts down on June 1st 2026. We need to a replacement till then. I will try to do some investigation here and update this comment later on. So far a heads up of a must action.

arif-ali:
That is unfortunate, as per our discussions, I have already been working internally to do GH Actions for CI, I'd be happy to contribute that back here if needs be. But that would still need more work. At the moment, it does Ubuntu builds, unit tests, stage one and stage two. Can easily incorporate others as well, and should be easily doable.

GH free allows 2000 mins, 500 MB artifact storage, and 10GB cache storage; if that is enough then we could just do it with GH. Based on the Cirrus dashboard, the biggest month was 7hrs, which is 420 mins, so could easily use the free tier. The only issue would be the artifacts storage. Each avocado failure could have 50MB of data, but depends how many days we want to keep the artifacts, we could be ok if we limit to like 7 days maybe


The only downside I've found so far is that it doesn't have development builds of the OSs, and therefore we won't see issues in advance (at least in my limited research in the past few days)

TurboTurtle:
Dang, that hurts to read.

I'd be in favor of GH actions if there is already work underway on that front. I don't think we _need_ devel OS builds, even if they are a nice-to-have. I'd assume our downstreams, e.g. Canonical and Red Hat, would be doing sufficient internal testing on pre-releases that could flag most issues ahead of time, and so we'd rely on those getting reported back up to us.

arif-ali:
Initial research suggests it's ubuntu, macros and windows images are available for GH Actions

So many issues and feedbacks asking for Fedora, Alma and such, but GH will not be adding those distros. So we may need to go somewhere else for CI

xtqqczze:
> GH free allows 2000 mins

It's actually unlimited for public repositories:

https://docs.github.com/en/billing/concepts/product-billing/github-actions#how-use-of-github-actions-is-measured

arif-ali:
@xtqqczze good to know

for the wider team. I have completed the migration in our Canonical repo to do all the tests using GH Workflow for Ubuntu 22.04 and 24.04. If we wish to proceed with using GH workflows, at least for Ubuntu this work is now complete, and I can remove the cirrus for Ubuntu stuff, and migrate to GH workflows. Below is a link to the successful run of the workflow.

https://github.com/canonical/sos/actions/runs/24501462363

[...]

PR #4310, opened 2026-04-24
[testing] Migrate CI to GH Workflow

arif-ali:
* CentOS and Fedora as docker containers.
* Fedora for stagetwo depends on py3.14 enablement for avocado.
* CentOS foreman tests to be added later, docker container and foreman
  may not behave as well.
* Tests for systemd and/or journals have been skipped for docker
  containers for the time being.
* Debian will be added later.
* 26.04 enablement is same as Fedora, plus waiting for the runners to
  be available in GH

Related: #4288

---


pmoravec (commented):
Looks great (as far as I my limited knowledge of GH workflows reach..)!

pmoravec:
The failed:

```
(240/307) /__w/sos/sos/sos/tests/report_tests/plugin_tests/logs.py:LogsPluginTest.test_journalctl_collections:  FAIL: /var/tmp/avocado_koakdqmzsos_tests.py.LogsPluginTest/sosreport-LogsPluginTest/sosreport-52a3bbea0876-2026-05-15-ocwynko/sos_commands/logs/journalctl_--no-pager_--boot does not exist (0.60 s)
```

is expected on systems with empty / no journals - see https://github.com/sosreport/sos/blob/main/sos/report/plugins/logs.py#L81-L85 . The https://github.com/sosreport/sos/blob/main/tests/report_tests/plugin_tests/logs.py#L27 must be called conditionally (or never).

arif-ali:
I am trying to work through this, as fedora and centos is container based, so figuring things out, hence still WiP. I'm conscious that we are closing towards EoL of Cirrus

arif-ali:
if anyone knows how to get systemd working in a docker container for fedora/centos, then that would be awesome, that's the last bit to get the stage one tests working under GH workflow/actions. I have tried many things first in my own Canonical repo, and some today. I'll keep researching, and see if I can find anything, but if you already have tackled it, then that could be useful

/cc @jcastill @pmoravec @TurboTurtle

jcastill:
First of all, thank you so much @arif-ali for doing this!
Now, images - from what I know, for centos-stream, an alternative could be to use the bootc images, that should have systemd enabled by default. The two images for centos stream should be:

quay.io/centos-bootc/centos-bootc:stream9
quay.io/centos-bootc/centos-bootc:stream10

Other than that, I think the other way would be to use a modified dockerfile and I'm not sure how feasible is that. I wonder if we could actually create sos-specific images in quay.io that are lightweight and also have systemd enabled, just for testing, but I'm not an expert in containers so maybe anybody else knows if it can be done and it's worth it?

Same for Fedora, these are the ones I've found:

quay.io/fedora/fedora-bootc:43
quay.io/fedora/fedora-bootc:44

Does that help?

pmoravec:
No much knowledge on this topic. I can try to learn & dig in, if necessary - if so, please describe the issue more specifically.

Checking one particular error:

> 110-tests/report_tests/command_priority_tests.py:CommandPriorityTest.test_logs_full_journal_correct_priority: FAIL

It is because `journalctl --no-pager  --reverse` wasnt collected, because `systemd-journald` service was not present (and since `/var/log/journal` was there, cf https://github.com/sosreport/sos/blob/main/sos/report/plugins/logs.py#L81-L85)?

AFAIK the service isnt usually present in containers. So *if* we want to run these journal tests inside containers, we must modify the service-enabling condition (to `is_executable('journalctl')`? that is weird..). Or skip the test if the service isnt present (sounds better approach to me).

arif-ali:
@pmoravec thanks for those ideas.

now we're facing the issue with fedora, that I had with Ubuntu 26.04, i.e. #4313, so we need avocado to fix that before we can move that along

I have changed it to not do stage two tests fedora bits, and just do centos, and move this along, we're already not doing fedora tests in Cirrus, so not going to make much of a difference

I will work out debian side of things for later, as that will require some further work, and wanted folks some time to review this as well as the centos/debian foreman tests

pmoravec:
`full_report_run.py:FullCleanTest` fails again due to no journald inside container, like:

```
[stdlog] 2026-05-20 08:32:09,150 avocado.test stacktrace       L0041 ERROR|
[stdlog] 2026-05-20 08:32:09,150 avocado.test stacktrace       L0043 ERROR| Reproduced traceback from: /__w/sos/sos/sos/.tox/stagetwo_tests/lib/python3.12/site-packages/avocado/core/test.py:594
[stdlog] 2026-05-20 08:32:09,151 avocado.test stacktrace       L0050 ERROR| Traceback (most recent call last):
[stdlog] 2026-05-20 08:32:09,151 avocado.test stacktrace       L0050 ERROR|   File "/__w/sos/sos/sos/tests/sos_tests.py", line 879, in setUp
[stdlog] 2026-05-20 08:32:09,151 avocado.test stacktrace       L0050 ERROR|     super().setUp()
[stdlog] 2026-05-20 08:32:09,151 avocado.test stacktrace       L0050 ERROR|   File "/__w/sos/sos/sos/tests/sos_tests.py", line 533, in setUp
[stdlog] 2026-05-20 08:32:09,151 avocado.test stacktrace       L0050 ERROR|     super().setUp()
[stdlog] 2026-05-20 08:32:09,151 avocado.test stacktrace       L0050 ERROR|   File "/__w/sos/sos/sos/tests/sos_tests.py", line 320, in setUp
[stdlog] 2026-05-20 08:32:09,151 avocado.test stacktrace       L0050 ERROR|     self.pre_sos_setup()
[stdlog] 2026-05-20 08:32:09,151 avocado.test stacktrace       L0050 ERROR|   File "/__w/sos/sos/sos/tests/cleaner_tests/full_report/full_report_run.py", line 39, in pre_sos_setup
[stdlog] 2026-05-20 08:32:09,151 avocado.test stacktrace       L0050 ERROR|     sosfd = journal.stream('sos-testing')
[stdlog] 2026-05-20 08:32:09,151 avocado.test stacktrace       L0050 ERROR|             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[stdlog] 2026-05-20 08:32:09,151 avocado.test stacktrace       L0050 ERROR|   File "/usr/lib64/python3.12/site-packages/systemd/journal.py", line 509, in stream
[stdlog] 2026-05-20 08:32:09,151 avocado.test stacktrace       L0050 ERROR|     fd = stream_fd(identifier, priority, level_prefix)
[stdlog] 2026-05-20 08:32:09,151 avocado.test stacktrace       L0050 ERROR|          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[stdlog] 2026-05-20 08:32:09,151 avocado.test stacktrace       L0050 ERROR| FileNotFoundError: [Errno 2] No such file or directory
[stdlog] 2026-05-20 08:32:09,151 avocado.test stacktrace       L0051 ERROR|
```

I will followup later today..

jcastill:
Is it worth to run a quick test using the systemd enabled containers I mentioned before?

arif-ali:
> Is it worth to run a quick test using the systemd enabled containers I mentioned before?

I thought I did, and had similar issues, I will try again, and see if I get issues again

pmoravec:
> > Is it worth to run a quick test using the systemd enabled containers I mentioned before?
> 
> I thought I did, and had similar issues, I will try again, and see if I get issues again

Thanks for all the work here.

As I was exploring the possibilities a bit (and maybe I found just some well-known-info for others but new info for me, so sorry for that if that's the case), you would need a custom container image like:

```
jobs:
  test:
    runs-on: ubuntu-latest
    container:
      image: ubuntu:22.04
      options: --privileged  # Required for systemd
    steps:
      - run: apt-get update && apt-get install -y systemd python3-pip
      - uses: actions/checkout@v3
      - run: systemctl start systemd-journald
      - run: pytest tests/
```

An alternative is mocking `journald` like via (untested, just a draft):

```
import os
import pytest
from unittest.mock import patch, MagicMock

@pytest.fixture
def mock_journal():
    """Mock systemd journal for CI environments"""
    if os.getenv('CI') == 'true':
        with patch('systemd.journal.stream') as mock_stream:
            mock_stream.return_value = MagicMock()
            yield mock_stream
    else:
        yield None

def test_sos_setup(mock_journal):
    # Your test that uses journal.stream()
    # In CI: uses mock; locally: uses real journal
    pass
```

arif-ali:
I had also done some research, and one way was to use docker to start the container directly with the right options to add systemd and journal stuff. But that could have been convoluted, and would need to run each command using the `docker command`, I wanted to do this in a template fashion, so that the same workflow can be reusable for any OS we decide to choose.

We already have `--privileged --tmpfs /tmp --tmpfs /run --tmpfs /run/lock -v /sys/fs/cgroup:/sys/fs/cgroup:ro` in the options, which is what I found through my extensive research to do, and then tried various steps to try and get systemd (not in this repo, but in our Canonical repo https://github.com/canonical/sos; this is where I did most of the testing before bringing it here)

Maybe the mocking option could be one way. Let's see how the current tests go, and we can potentially look at that

arif-ali:
> Is it worth to run a quick test using the systemd enabled containers I mentioned before?

Based on the current test, the bootc image doesn't work in stageone tests, but the normal one does 🤷🏽, and this is with all the skips for systemd/journal. Something about `krb5-server` not being able to install

jcastill:
What about the option of creating sos-specific centos-stream docker images in quay.io? I imagine that in that scenario we could control the dockerfile and have a combination of the base image + the bootc image (and give us more flexibility in the future if we want to add more things)

arif-ali:
> What about the option of creating sos-specific centos-stream docker images in quay.io? I imagine that in that scenario we could control the dockerfile and have a combination of the base image + the bootc image (and give us more flexibility in the future if we want to add more things)

That's an option, I think we did similar in Cirrus world way before I joined the project here.

Seems like you have to pay to get images on there, unless you have some insider thing that allows you to do this. I would be open to that, as long as other folks here would also be happy with that too.

We are so close in closing this off and getting it to work, I'll see if I can do something similar to these last failures with skipIf decorators or variables to skip these for now, and we can work on the new containers with less pressure on getting this done for the end of the month

jcastill:
> Seems like you have to pay to get images on there, unless you have some insider thing that allows you to do this. I would be open to that, as long as other folks here would also be happy with that too.

OK, let me investigate what options do we have and if there's no free tier, if we can get sponsored. 


> We are so close in closing this off and getting it to work, I'll see if I can do something similar to these last failures with skipIf decorators or variables to skip these for now, and we can work on the new containers with less pressure on getting this done for the end of the month

Agreed, lets finish this up before the deadline in 10 days and then we can work on making it "perfect". Once this PR is merged, I'll open a new one where we can experiment further. 

Thank you again @arif-ali !

arif-ali:
Right, this should be it now, there are issues with snapstore at the moment, but I know the ubuntu side will pass once the snapstore infra is good.

So, reviews from the team would be awesome

.github/workflows/avocado.yaml
```diff
@@ -0,0 +1,153 @@
...
+        uses: actions/download-artifact@v8
+        with:
+          name: sos-build.snap
+
+      - name: Install the package
+        run: |
+          if [ $(command -v apt) ] ; then
+            [[ "$(dpkg -l sos)" ]] && sudo apt -y purge sos ubuntu-server
+            [[ "$(dpkg -l sosreport)" ]] && sudo apt -y purge sosreport ubuntu-server
+          fi
```
pmoravec, line 105: I guess this is rather a nitpick: you remove the sos package when on `deb` / when `apt` command exists. Does it make sense to have the same for snap and rpm as well?


Also, shouldnt the `$(command -v apt)` be rather `"${{ inputs.build }}" == "deb" ` to align with below install branch?

.github/workflows/main-pipeline.yaml
```diff
@@ -0,0 +1,228 @@
...
+        with:
+          filters: |
+            foreman:
+              - '.github/workflows/**'
+              - '**/__init__.py'
+              - '**/apache.py'
+              - '**/foreman_tests.py'
+              - '**/candlepin.py'
+              - '**/pulp.py'
+              - '**/pulpcore.py'
```
pmoravec, line 57: Is my understanding right that this specifies "run foreman tests only when either of these files are updated"? Then `'**/foreman*.py` is missing..?

.github/workflows/avocado.yaml
arif-ali: `$(command -v apt)` covers both scenarios for deb and snap, as we're only doing these on Ubuntu images.

will align the rpm

.github/workflows/main-pipeline.yaml
arif-ali: good catch

pmoravec: A nitpick: `foreman.py` -> `foreman*.py` as there is also `foreman_installer`, `foreman_proxy` and will be `foremanctl` (I will need to figure out how to test that..)

tests/report_tests/basic_report_tests.py
```diff
@@ -48,13 +48,16 @@ def test_tag_summary_created(self):
...
 
     def test_dir_listing_works(self):
-        self.assertFileCollected('sos_commands/boot/ls_-alZR_.boot')
-        boot_ls = self.get_name_in_archive('sos_commands/boot/ls_-alZR_.boot')
-        with open(boot_ls, 'r', encoding='utf-8') as ls_file:
+        self.assertFileCollected('sos_commands/logs/ls_-alZRs_.var.log')
+        var_log_ls = self.get_name_in_archive(
+            'sos_commands/logs/ls_-alZRs_.var.log'
+        )
+        with open(var_log_ls, 'r', encoding='utf-8') as ls_file:
```
pmoravec, line 55: Just curious question, I don't object at all: why the `boot/ls_-alZR_.boot` -> `logs/ls_-alZRs_.var.log` change, please?

pmoravec: Eh, self-answering; /boot` is not in containers, I guess.

tests/sos_tests.py
```diff
@@ -75,6 +75,20 @@ def wrapper(*args, **kwargs):
...
+    def wrapper(*args, **kwargs):
+        if check_if_container():
+            raise TestSkipError('Not running on Physical or VM environment')
+        tst(*args, *kwargs)
+    return wrapper
+
+
+def check_if_container():
+    return (os.path.exists('/.dockerenv') or
+            'docker' in open('/proc/1/cgroup', encoding='utf-8').read())
```
pmoravec, line 89: No podman? Or is this meant purely for GH testing (that apparently uses docker)?

arif-ali: at the moment, I've only gone about on this for GH testing. Happy to add podman here, if needed.

pmoravec:
Comparing tests and builds triggered before and after:

- fedora RPM builds just for x86_64 arch. IMHO sufficient
- Ubuntu 22.04 and 24.04 is used instead of 26.04 (or 25.10). More tests on older version, is it pros or cons? (no strong opinion)
- foreman tested only on ubuntu deb, not on debian or snap. it would be great to have 2 distros, ideally something from Red Hat family, but I am ok with this
- what replaced "Breakage test"s?
- no stage one or two tests on debian..?

Thanks again for all the huge work, @arif-ali !

arif-ali:
> Comparing tests and builds triggered before and after:
> 
> * fedora RPM builds just for x86_64 arch. IMHO sufficient

That's what we were doing before? 

> * Ubuntu 22.04 and 24.04 is used instead of 26.04 (or 25.10). More tests on older version, is it pros or cons? (no strong opinion)

These are LTS releases, and GH native runners are available. 26.04 is still pending to be added by GH, so that's a waiting game. 

For intermediate releases GH doesn't do runners, so we may need to do containers for those. So, some extra work is required there. It would be good to do it, but thought as they are only supported for 9 months, and the latest 25.10 is only supported for another month didn't make sense. 

> * foreman tested only on ubuntu deb, not on debian or snap. it would be great to have 2 distros, ideally something from Red Hat family, but I am ok with this

Totally understand, the RH and Debian both need adding here, but as these will be done in docker, this may take some time to get done, but it's in my plan to do it. 

I'm away some of next week and today, and have other work on my radar, so wasn't going to have time to complete all of it

> * what replaced "Breakage test"s?

The py tests are the breakage tests, which tests python 3.9 and latest 3.x

> * no stage one or two tests on debian..?

Again, docker related, so need to spend extra time to fine tune it all to get these going, it's on my radar to do

> Thanks again for all the huge work, @arif-ali !

No worries at all

pmoravec (approved):
Ack from me, thanks for all the work done.

Not sure if we should wait for @TurboTurtle review as well, to have full consensus here. As the change is quite substantial..

pmoravec:
> > Comparing tests and builds triggered before and after:
> > 
> > * fedora RPM builds just for x86_64 arch. IMHO sufficient
> 
> That's what we were doing before?

Per some other PR and its tests, yes :) but as I wrote, I see no much benefit in those.

> 
> > * Ubuntu 22.04 and 24.04 is used instead of 26.04 (or 25.10). More tests on older version, is it pros or cons? (no strong opinion)
> 
> These are LTS releases, and GH native runners are available. 26.04 is still pending to be added by GH, so that's a waiting game.

LGTM.

> > * foreman tested only on ubuntu deb, not on debian or snap. it would be great to have 2 distros, ideally something from Red Hat family, but I am ok with this
> 
> Totally understand, the RH and Debian both need adding here, but as these will be done in docker, this may take some time to get done, but it's in my plan to do it.

+1, plus I should have this on my radar for the `foremanctl` stuff as well (and feel free to point to my this promise there :) )

> > * what replaced "Breakage test"s?
> 
> The py tests are the breakage tests, which tests python 3.9 and latest 3.x

Ah, I was blind, thx.

TurboTurtle (approved):
Ack. Thanks for taking this on! This was a real curveball from Cirrus, and this is a very strong pivot.

TurboTurtle:
Holding merge for ~24 hours for final comments given the size and scope of this.

PR #4312, opened 2026-04-27
[net_diagram] add plugin for topology visualization

harshith-aa:
---

- [X] Does the subject start with **[net_diagram]**?
- [X] Does the commit contain a **Signed-off-by: Harshith Hari <harshithhari000111@.gmail.com>**?
- [X] Are all passwords or private data gathered by this PR obfuscated?

---

Problem:
On systems with a high number of interfaces, the standard ip a output is overwhelming. It is difficult to visualize the relationships between interfaces and trace network flow.

Solution:
The net_diagram plugin automates the generation of a topology map. It helps create a clear and understandable network diagram.

### DIFF
--- a/sos/report/plugins/net_diagram.py
+++ b/sos/report/plugins/net_diagram.py
@@ -1,23 +1,29 @@
-# Copyright (C) 2026 Harshith Hari <ghari@redhat.com>
-# License: GPLv2
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+#
+# Copyright (C) 2026 Harshith Hari <harshithhari000111@gmail.com>
 
 import os
-import subprocess
-from sos.report.plugins import Plugin, RedHatPlugin
+from sos.report.plugins import Plugin, IndependentPlugin
 
 
-class NetDiagram(Plugin, RedHatPlugin):
-    """
-    CEE-Grade Network Topology Visualizer.
-    Generates a DOT graph and PNG image of the network hierarchy.
+class NetDiagram(Plugin, IndependentPlugin):
+    """This plugin generates a visual network topology diagram (PNG) using
+    Graphviz. It maps relationships between physical NICs, Bonds, Bridges,
+    and Team devices to help support engineers visualize complex network
+    hierarchies.
     """
 
     plugin_name = 'net_diagram'
     profiles = ('network',)
     packages = ('graphviz', 'teamd')
 
-    def setup(self):
-        # 1. Initial DOT Graph Header
+    def collect(self):
         dot_lines = [
             "digraph network {",
             "  rankdir=LR; overlap=false; splines=true;",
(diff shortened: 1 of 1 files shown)


sos/report/plugins/net_diagram.py
```diff
@@ -0,0 +1,141 @@
+"""
+This plugin generates a visual network topology diagram (PNG) using Graphviz.
+It maps relationships between network interfaces to help support engineers visualize complex network hierarchies.
+"""
+# Copyright (C) 2026 Harshith Hari <harshithhari000111@gmail.com>
+# License: GPLv2
+
+import os
+import subprocess
```
TurboTurtle: Don't use `subprocess` directly in plugins. Please use the `Plugin.exec_cmd()` method if you just need the output returned, or `Plugin.collect_cmd()` if you need it returned within the code _and_ saved within the archive.

sos/report/plugins/net_diagram.py
```diff
@@ -0,0 +1,141 @@
...
+            "fontsize=10];",
+            "  edge [penwidth=2, color=\"#444444\"];"
+        ]
+
+        # 2. Capture IPs using standard subprocess
+        ips = {}
+        try:
+            out = subprocess.check_output(
+                ["ip", "-br", "addr"], encoding='utf-8'
+            )
```
TurboTurtle: Here, for example.

```python
addrs = self.exec_cmd('ip -br addr')
for line in addrs['output'].splitlines():
    [...]
```

sos/report/plugins/net_diagram.py
```diff
@@ -0,0 +1,141 @@
...
+        net_path = "/sys/class/net/"
+        if not os.path.exists(net_path):
+            return
+
+        # 3. Process Interfaces and Relationships
+        for iface in os.listdir(net_path):
+            if iface in ('lo', 'bonding_masters'):
+                continue
+
+            path = os.path.join(net_path, iface)
```
TurboTurtle: Use `Plugin.path_join()` here, as it automatically accounts for non-standard sysroots, e.g. when `sos` is run from a rescue environment.

sos/report/plugins/net_diagram.py
```diff
@@ -0,0 +1,141 @@
...
+                                    f'[style=bold, color=blue];'
+                                )
+                except (IOError, OSError):
+                    pass
+
+        # 4. Gateway Detection
+        try:
+            route_out = subprocess.check_output(
+                ["ip", "route", "show"], encoding='utf-8'
+            )
```
TurboTurtle: Same here as earlier, don't use `subprocess` directly in plugins, please.

sos/report/plugins/net_diagram.py
```diff
@@ -0,0 +1,141 @@
...
+
+        try:
+            with open(tmp_dot, "w", encoding='utf-8') as f:
+                f.write(dot_str)
+            subprocess.run(
+                ["dot", "-Tpng", tmp_dot, "-o", tmp_png], check=True
+            )
+            self.add_copy_spec(tmp_png)
+        except (subprocess.SubprocessError, IOError):
+            pass
```
TurboTurtle: Do not write out to `/tmp` as part of plugins.

For this, you'd want to use a pattern like what is in `process` plugin; use `self.collection_file()`, and do this as part of `collect()`, not `setup()`. E.g.

```python
def collect(self):
    [...move the bits generating dot_str here...]
    with self.collection_file('network_topology.dot') as dotf:
        dotf.write(dot_str)
```

For writing directly into the archive path, use `self.get_cmd_output_path()`, such as in the `lvm2` plugin.

sos/report/plugins/net_diagram.py
```diff
@@ -0,0 +1,141 @@
+"""
+This plugin generates a visual network topology diagram (PNG) using Graphviz.
+It maps relationships between network interfaces to help support engineers visualize complex network hierarchies.
+"""
+# Copyright (C) 2026 Harshith Hari <harshithhari000111@gmail.com>
+# License: GPLv2
```
TurboTurtle: Please use our standard header here - you can see it in our [Contribution Guidelines](https://github.com/sosreport/sos/wiki/Contribution-Guidelines#license-header-and-copyright-assignment) wiki entry.

sos/report/plugins/net_diagram.py
```diff
@@ -0,0 +1,141 @@
+"""
+This plugin generates a visual network topology diagram (PNG) using Graphviz.
+It maps relationships between network interfaces to help support engineers visualize complex network hierarchies.
+"""
```
TurboTurtle: This should be part of the plugin docstring, so that it gets consumed by `sos help`.

TurboTurtle (changes requested):
I get the intent of this plugin, however it should conform to our standard plugin design approach. See comments below.

jcastill:
@harshith-aa some notes:
- Make sure you squash all commits into one.
- Once you do that, make sure it's properly signed (you can find instructions [here](https://github.com/sosreport/sos/pull/4312/checks?check_run_id=73239782550) , just modify  step 2 from HEAD~4 to HEAD~1).
- When you move line 3 to the plugin docstring, modify it to fit within 79 chars

One more thing: With this plugin, you are not gathering information about network devices, but rather using a tool to create a graph, right? I.e. the information you are capturing should be already there in other plugins.

And two follow up questions:
a- What's the advantage of running this in customers/users machines instead of, lets say, Red Hat's supportshell (or similar for other distributions)?
b- Is there a reason why we need a new plugin, instead of running it as part of the networking plugin? Perhaps as an extra option that can be enabled in the command line.

sos/report/plugins/net_diagram.py
```diff
@@ -0,0 +1,102 @@
...
+#
+# See the LICENSE file in the source distribution for further information.
+#
+# Copyright (C) 2026 Red Hat Inc., Harshith Hari <harshithhari000111@gmail.com>
+
+import os
+from sos.report.plugins import Plugin, RedHatPlugin
+
+
+class NetDiagram(Plugin, RedHatPlugin):
```
jcastill: Any reason why this plugin will apply only to Red Hat machines?

harshith-aa: I've updated the plugin to be distro-agnostic by inheriting from the base Plugin class. I also added check_enabled() to ensure it's active across different distributions and verified the code with Pylint (10/10). The implementation continues to use self.path_join for sysroot compliance.

jcastill: Use IndependentPlugin in this case (check https://github.com/sosreport/sos/wiki/How-to-Write-a-Plugin#os-specific-plugins for more details)

sos/report/plugins/net_diagram.py
```diff
@@ -0,0 +1,108 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+#
+# Copyright (C) 2026 Red Hat Inc., Harshith Hari <harshithhari000111@gmail.com>
```
TurboTurtle: Did you want to assign the copyright to Red Hat? If you not, you can drop the company bit and keep the copyright line with just yourself as an individual.

sos/report/plugins/net_diagram.py
```diff
@@ -0,0 +1,108 @@
...
+
+import os
+from sos.report.plugins import Plugin
+
+
+class NetDiagram(Plugin):
```
TurboTurtle: You'll need to use `IndependentPlugin` here as well as the base `Plugin` in order for it to be enabled on all hosts.

sos/report/plugins/net_diagram.py
```diff
@@ -0,0 +1,108 @@
...
+    and Team devices to help support engineers visualize complex network
+    hierarchies.
+    """
+
+    plugin_name = 'net_diagram'
+    profiles = ('network',)
+    packages = ('graphviz', 'teamd')
+
+    def setup(self):
+        """Standard setup method to ensure plugin activation."""
```
TurboTurtle: No need to redefine this if we aren't doing anything within `setup()`.

sos/report/plugins/net_diagram.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def check_enabled(self):
+        """Enable this plugin regardless of the distribution."""
+        return True
```
TurboTurtle: Once you use subclass `IndependentPlugin`, this will not be needed.

sos/report/plugins/net_diagram.py
```diff
@@ -0,0 +1,108 @@
...
+                )
+                dot_lines.append(
+                    f'  "{gw_dev}" -> "gateway" [label="default route", '
+                    'color="#27ae60", style=dashed];'
+                )
+
+        # 3. Build Topology from Sysfs relationships
+        net_path = self.path_join("/sys", "class", "net")
+        if os.path.exists(net_path):
+            for iface in os.listdir(net_path):
```
TurboTurtle: Like `Plugin.path_join()` for handling non-`/` sysroots, there are `Plugin.path_exists()` and `Plugin.listdir()` methods available.

TurboTurtle (commented):
Looking close, just a couple overall minor sos-isms to address.

sos/report/plugins/net_diagram.py
harshith-aa: I have Inherited from IndependentPlugin for distro-agnostic support.

sos/report/plugins/net_diagram.py
harshith-aa: I have updated the plugin to fully comply with the requested changes:

TurboTurtle (approved):
LGTM

pmoravec (approved):

PR #4315, opened 2026-04-30
suse: add SLESPolicy and fix OpenSuSEPolicy detection

filanov:
SuSEPolicy.check() always returns False, requiring concrete subclasses
  to override it. OpenSuSEPolicy never did, and no SLESPolicy existed,
  so both openSUSE and SLES systems fell back to GenericLinuxPolicy which
  uses a no-op PackageManager with an empty package dict. This caused all
  package-gated plugins (rdma, roce, infiniband, etc.) to show as inactive
  regardless of what was installed.

  Add SLESPolicy with os_release_name/os_release_id matching SLES
  /etc/os-release values, and fix OpenSuSEPolicy to override check() via
  LinuxPolicy.check() so both policies are correctly detected and
  RpmPackageManager is used for package lookups.

---

### DIFF
--- a/sos/policies/distros/suse.py
+++ b/sos/policies/distros/suse.py
@@ -61,8 +61,8 @@ class SuSEPolicy(LinuxPolicy):
 class OpenSuSEPolicy(SuSEPolicy):
     vendor = "SuSE"
     vendor_urls = [('Community Website', 'https://www.opensuse.org/')]
-    os_release_name = "OpenSuSE"
-    os_release_file = '/etc/SUSE-brand'
+    os_release_name = "openSUSE Leap"
+    os_release_id = "opensuse-leap"
     msg = _("""\
 This command will collect diagnostic and configuration \
 information from this %(os_release_name)s system and installed \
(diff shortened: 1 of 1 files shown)


sos/policies/distros/suse.py
```diff
@@ -82,4 +82,37 @@ def __init__(self, sysroot=None, init=None, probe_runtime=True,
...
+
+No changes will be made to system configuration.
+%(vendor_text)s
+""")
+
+    def __init__(self, sysroot=None, init=None, probe_runtime=True,
+                 remote_exec=None):
+        super().__init__(sysroot=sysroot, init=init,
+                         probe_runtime=probe_runtime,
+                         remote_exec=remote_exec)
```
pmoravec: Why to redefine the same method like in parent?

sos/policies/distros/suse.py
```diff
@@ -82,4 +82,37 @@ def __init__(self, sysroot=None, init=None, probe_runtime=True,
                          probe_runtime=probe_runtime,
                          remote_exec=remote_exec)
 
+    @classmethod
+    def check(cls, remote=''):
+        return LinuxPolicy.check.__func__(cls, remote)
```
pmoravec, line 81: Why the unnecessary re-definition (same applies to `SLESPolicy.check`)?

filanov: It's not the same, i'm not sure about the original reason to have `check` return false, but i saw that without it this suse version will identify the packages and list packages that should exit as if they are not and other that shouldn't match as positive. is there a better way to resolve this?

sos/policies/distros/suse.py
filanov: right, removing

sos/policies/distros/suse.py
pmoravec: Ah I understand - you dont want calling direct parent class method, but "higher ancestor" one. That makes sense, and I think this "weird" way of implementation is required - I cant find any better.

pmoravec (approved):
Sounds good to me.

TurboTurtle (approved):

PR #4316, opened 2026-04-30
[cleaner] Fix short hostname obfuscation

### LINKED ISSUE #3388: sos clean should obfuscate hostnames like dhcp-192-168-0-66
As seen in https://access.redhat.com/solutions/5154441 , dhcp often sets hostnames using the numeric parts of the IP address separated by dashes. The cleaner should obfuscate this kind of hostname because the ip address is getting leaked with dashes instead of dots. This has been frequently flagged in Ubuntu autopkgtests.

The sos cleaner regex for IP address is , looking at https://github.com/sosreport/sos/blob/main/sos/cleaner/parsers/ip_parser.py#L15C1-L22C6 ,

class SoSIPParser(SoSCleanerParser):
    """Handles parsing for IP addresses"""

    name = 'IP Parser'
    regex_patterns = [
        # IPv4 with or without CIDR
        r'((?<!(-|\.|\d))([0-9]{1,3}\.){3}([0-9]){1,3}(\/([0-9]{1,2}))?)'
    ] 

I think we should also consider non dash characters, to catch leaks like an ip address for eg like w.x.y.z. leaking like something-w-x-y-z when dhcp seems to set the hostnames this way.

kicch0u:
Any hostname in /etc/hosts/ not in FQDN format, such as 'my-computer', was being missed by sos clean in /etc/hosts.

Modify control flow in sos/cleaner/preppers/hostname.py to strip trailing comments (denoted by '#') and unconditionally add remaining lines in /etc/hosts to the 'items' list, allowing the hostnames within the lines to be obfuscated.

Add unit tests for short hostname obfuscation. The new unit tests avoid modifying the existing test archive by declaring some mock objects, but I can submit a revision if desired.

Closes: https://github.com/sosreport/sos/issues/3388
Assisted-by: Copilot:claude-sonnet-4.6 <github.com/features/copilot>

---

### DIFF
--- a/sos/cleaner/preppers/hostname.py
+++ b/sos/cleaner/preppers/hostname.py
@@ -50,12 +50,17 @@ class HostnamePrepper(SoSPrepper):
         for line in _hosts.splitlines():
             if line.startswith('#') or 'localhost' in line:
                 continue
+            # strip inline comments before processing
+            line = line.split('#')[0].strip()
+            if not line:
+                continue
             hostln = line.split()[1:]
             for host in hostln:
                 if len(host.split('.')) == 1:
                     self.regex_items['hostname'].add(host)
-                else:
-                    items.append(host)
+                # unconditionally append host to items to avoid
+                # missing short hostnames like 'my-computer'
+                items.append(host)
 
         for domain in self.opts.domains:
             items.append(domain)
--- a/tests/unittests/cleaner_tests.py
+++ b/tests/unittests/cleaner_tests.py
@@ -358,6 +358,88 @@ class PrepperTests(unittest.TestCase):
             self.host_prepper.get_items_for_map('hostname', self.archive)
         )
 
+    def test_hostname_prepper_etc_hosts_short_name_in_items(self):
+        """Short hostnames from /etc/hosts must appear in items so that
+        mapping.add() is called and self.hosts is populated, allowing proper
+        obfuscation."""
+        class MockArchive:
+            is_sos = True
+            is_insights = False
+
+            def get_file_content(self, path):
+                if path == 'sos_commands/host/hostname_-f':
+                    return 'myhost.example.com'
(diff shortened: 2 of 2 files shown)


sos/cleaner/preppers/hostname.py
pmoravec, line 41: Worth to unify the behaviour with `hostname -f` output parsing? E.g. (I havent tested this, though) if `hostname -f` is just `my-hostname`, we should add it to `regex_items` as well?

kicch0u: Thanks for the review! However, I'm not sure I understand your question.

As I understand the code, the command output of 'hostname -f' is already stored by that snippet you highlighted.

The issue I'm trying to address with this PR is that the remaining short hostnames in /etc/hosts aren't being obfuscated at all. If we modify /etc/hosts to contain:
```
fd::1 my-router
192.168.1.1 still-my-router
192.168.1.15 printer
fd::17 ps5
```
sosreport fails to obfuscate those entries entirely. This PR addresses that.

The entry in /etc/hosts that happens to contain the same output as the 'hostname -f' command (juarez in this case):
```
127.0.0.1 localhost
127.0.1.1 juarez

# The following lines are desirable for IPv6 capable hosts
::1     ip6-localhost ip6-loopback
fe00::0 ip6-localnet
ff00::0 ip6-mcastprefix
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
```
already gets obfuscated correctly.

pmoravec: Ah I was blind, my question in fact was "why not to have also `items.append(host)` here, to unify handling `hostname` and `/etc/hosts` records - but it is done via `items.append(domains[0])` already.

pmoravec (approved):

arif-ali (approved):

PR #4318, opened 2026-05-04
[cloud_init] Add missing systemd service units

jonathanbrenes:
Add cloud-init-local, cloud-config, and cloud-final to the services tuple so sos report captures their status alongside the existing cloud-init, cloud-init-main, and cloud-init-network units.

As of cloud-init 24.3, cloud-init uses a single-process optimization with multiple systemd service units acting as shims for each boot stage (see canonical/cloud-init#5489):

- cloud-init-local.service — Local stage
- cloud-init-network.service — Network stage (renamed from cloud-init.service)
- cloud-config.service — Config stage
- cloud-final.service — Final stage
- cloud-init-main.service — Main single-process service

Previously only cloud-init, cloud-init-main, and cloud-init-network were listed, missing the local, config, and final stage services.

### DIFF
--- a/sos/report/plugins/cloud_init.py
+++ b/sos/report/plugins/cloud_init.py
@@ -17,7 +17,8 @@ class CloudInit(Plugin, IndependentPlugin):
 
     plugin_name = 'cloud_init'
     packages = ('cloud-init',)
-    services = ('cloud-init', 'cloud-init-main', 'cloud-init-network')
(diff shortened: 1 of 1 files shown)


jcastill:
@jonathanbrenes thank you for this PR. The issue with the DCO (missing sign off) can be easily fixed. Just do:

```
git commit --amend -s

git push --force
```

That should be enough to sign off your commit.

arif-ali (approved):

pmoravec (approved):

PR #4319, opened 2026-05-06
Include OCP node role and cluster name in RHCOS archive name

asadawar:
## Summary

When `sos report` runs standalone on an RHCOS node, the archive name only contains the short hostname (e.g. `sosreport-master-0-2026-05-06-abcdefg.tar.xz`). In customer environments where a single bastion host is used to collect sosreports from multiple OpenShift clusters, archives from nodes with similar hostnames across clusters are indistinguishable without extraction.

This PR overrides `get_archive_name()` in `RedHatCoreOSPolicy` to include the OCP node role and cluster name in the archive name when running on RHCOS:

**Before:** `sosreport-master-0-2026-05-06-abcdefg.tar.xz`
**After:** `sosreport-master-0-master-mycluster-2026-05-06-abcdefg.tar.xz`

### How it works

- **Node role** is detected from `/etc/machine-config-daemon/currentconfig` using three fallbacks in order:
  1. The `machineconfiguration.openshift.io/role` label
  2. The owning `MachineConfigPool` name from `ownerReferences`
  3. The rendered config name (format `rendered-<role>-<hash>`)

  Testing on a live RHCOS 9.6 node showed the label is not always present — the `ownerReferences` fallback was needed to detect the role.

- **Cluster name** is extracted from `/etc/kubernetes/kubeconfig` by parsing the API server URL (`https://api-int.<clustername>.<basedomain>:6443` or `https://api.<clustername>.<basedomain>:6443`). Falls back to hostname parsing (`<role>-<index>.<clustername>.<basedomain>`) if the kubeconfig is unavailable.

- The detected role and cluster name are prepended to the `--label` field, delegating all archive name formatting to the parent `get_archive_name()` to avoid duplicating logic.
- No API access needed — all values come from local filesystem
- Falls back to standard naming when neither value is available
- Existing `--label` and `--case-id` options continue to work as before

Note: `sos collect` already handles node role labeling via the `ocp` cluster profile's `set_node_label()`, but that only works through the collector. This change bridges the gap for standalone `sos report` runs.

## Test plan

- [x] All existing unit tests pass (32 passed, 2 skipped)
- [x] Added 16 new unit tests covering:
  - `_get_node_role()`: label detection, ownerReferences fallback, rendered name fallback, missing file, bad JSON, empty metadata
  - `_get_cluster_name()`: extraction from kubeconfig (`api-int` and `api` prefixes), fallback to hostname, short hostname, complex FQDN
  - `get_archive_name()`: end-to-end with role+cluster, user label preservation, role-only, fallback, label restoration
- [x] Verified on a live RHCOS 9.6 (OCP 4.21) master node — role detected as `master` via ownerReferences, cluster name `sharedocp421` from kubeconfig
- [ ] Manual verification with full `sos report` run on an RHCOS node

RFE: https://redhat.atlassian.net/browse/RFE-9249

### DIFF
--- a/sos/policies/distros/redhat.py
+++ b/sos/policies/distros/redhat.py
@@ -8,6 +8,7 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
+import json
 import os
 import sys
 import re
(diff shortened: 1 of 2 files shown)


asadawar:
Updated based on testing against a live RHCOS 9.6 (OCP 4.21) node:

- **Node role detection**: The `machineconfiguration.openshift.io/role` label was not present in `currentconfig` on the test node. Added two fallbacks: first checks the owning `MachineConfigPool` from `ownerReferences`, then parses the rendered config name (`rendered-<role>-<hash>`).
- **Cluster name**: Now extracted from `/etc/kubernetes/kubeconfig` (handles both `api-int.` and `api.` prefixes). Falls back to hostname parsing if unavailable.
- Added `Assisted-by` tag per AI_POLICY.md.
- Flake8 and DCO fixes.

asadawar:
**A note on HyperShift (Hosted Control Planes):**

This PR targets standard OCP deployments where RHCOS nodes run as full cluster members (master/worker) with the Machine Config Operator managing node configuration. In this setup, `/etc/machine-config-daemon/currentconfig` and `/etc/kubernetes/kubeconfig` are present on every node across all platforms (AWS, Azure, GCP, vSphere, Baremetal, SNO, platform:none).

HyperShift is a different OCP deployment model where the control plane runs as pods on a separate management cluster, and only the worker nodes are actual RHCOS machines. On these HyperShift worker nodes, the MCD stores its config at `/etc/mcd-currentconfig.json` instead of `/etc/machine-config-daemon/currentconfig`.

This PR does not cover HyperShift workers — if the file is not found, it gracefully falls back to standard naming. If HyperShift support is desired, I can open a follow-up PR to add `/etc/mcd-currentconfig.json` as a fallback path. Let me know if that would be useful.

pmoravec (approved):

TurboTurtle (approved):

PR #4322, opened 2026-05-07
[azure] Refactor IMDS collection with pretty-printed JSON

jonathanbrenes:
Replace the curl-based IMDS metadata collection with a Python-based approach using `urllib`. This produces pretty-printed JSON output (4-space indentation) for improved readability in the sos report, matching the pattern used by the GCP plugin.

Update the IMDS API version from `2023-07-01` to `2025-04-07`, the latest supported version as documented at:
https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service

Collect the full instance metadata (`/metadata/instance`) instead of only `/metadata/instance/compute`, which now also includes network information.

Tested on RHEL 8.10, 9.7, 10.1, and SAP-HA 8.10 Azure VMs.

---

### DIFF
--- a/sos/report/plugins/azure.py
+++ b/sos/report/plugins/azure.py
@@ -9,8 +9,16 @@
 # See the LICENSE file in the source distribution for further information.
 
 import glob
+import json
 import os
+from urllib.error import URLError
+from urllib.request import Request, urlopen
+
 from sos.report.plugins import Plugin, UbuntuPlugin, RedHatPlugin
+
+
+IMDS_URL = ("http://169.254.169.254/metadata/instance?"
+            "api-version=2025-04-07&format=json")
 
 
 class Azure(Plugin, UbuntuPlugin):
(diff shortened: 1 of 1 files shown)


arif-ali:
looks good to me, you need to sign your commit as per the contribution guidelines

`git commit -s --amend` should do it, and then force push

jonathanbrenes:
> looks good to me, you need to sign your commit as per the contribution guidelines
> 
> `git commit -s --amend` should do it, and then force push

@arif-ali  it should be done. Please check

sos/report/plugins/azure.py
```diff
@@ -40,11 +48,28 @@ def setup(self):
...
-            'curl -s -H Metadata:true --noproxy "*" '
-            '"http://169.254.169.254/metadata/instance/compute?'
-            'api-version=2023-07-01&format=json"'
-        ), suggest_filename='instance_metadata.json')
+    def collect(self):
+        with self.collection_file('instance_metadata.json') as mfile:
+            try:
+                metadata = self._get_imds_metadata()
+                mfile.write(json.dumps(metadata, indent=4))
+            except RuntimeError as err:
```
pmoravec, line 57: We rely on `json.loads` and `json.dumps` won't raise an exception like `TypeError` or `JSONDecodeError` that won't be caught here - is it safe enough? I.e. can we rely on the remote server it always reply with a valid JSON object? (I guess so, just want to be assured).

jonathanbrenes: Yes, this is safe. The Azure IMDS endpoint guarantees valid JSON in the response body when it returns HTTP 200 with `format=json` — any failure mode (service unavailable, timeout, non-200 status) is already caught and wrapped as `RuntimeError`. If IMDS somehow returned invalid JSON on a 200, that would indicate a critical Azure host infrastructure failure well beyond sos's scope.

Additionally, `json.dumps()` cannot raise on the output of `json.loads()` — the round-trip is always safe for native Python types — and the base plugin framework in [`_collect_manual()`](https://github.com/sosreport/sos/blob/main/sos/report/plugins/__init__.py#L3314-L3316) catches any unhandled `Exception` anyway:

```python
def _collect_manual(self):
    try:
        self.collect()
    except Exception as err:
        self._log_error(f"Error during plugin collections: {err}")
```

So even a hypothetical `JSONDecodeError` would be logged gracefully without crashing the report.

This is also the exact same pattern used by the [GCP plugin](https://github.com/sosreport/sos/blob/main/sos/report/plugins/gcp.py#L63-L85), which handles its metadata collection identically.

pmoravec:
Thanks for explanation, sounds good to me.

Woudl you be so kind and rebase your PR to the current `main`, to resolve the merge conflict due to https://github.com/jonathanbrenes/sos/commit/777eda42ad147360a9dc14d4e3b91e67f0123785?

pmoravec (approved):

TurboTurtle (approved):

PR #4327, opened 2026-05-14
fix: merge skip_plugins from config and cmdline instead of replacing

filanov:
When -n is passed on the cmdline, the config file's skip-plugins list was being replaced rather than merged, causing previously-skipped plugins to run unexpectedly.

---

### DIFF
--- a/sos/component.py
+++ b/sos/component.py
@@ -254,6 +254,8 @@ class SoSComponent():
                     oplugopts = [oopt for oopt in oplugopts
                                  if not oopt.startswith(cstring)]
                 setattr(opts, opt, list(set(val) | set(oplugopts)))
+            elif opt in ('skip_plugins', 'enable_plugins'):
+                setattr(opts, opt, list(set(val) | set(getattr(opts, opt))))
             elif val != opts.arg_defaults[opt]:
                 setattr(opts, opt, val)
 


sos/component.py
```diff
@@ -254,6 +254,8 @@ def apply_options_from_cmdline(self, opts):
                     oplugopts = [oopt for oopt in oplugopts
                                  if not oopt.startswith(cstring)]
                 setattr(opts, opt, list(set(val) | set(oplugopts)))
+            elif opt == 'skip_plugins':
+                setattr(opts, opt, list(set(val) | set(getattr(opts, opt))))
```
pmoravec: Doesnt the same affect `-o` and `-e` options and shouldnt we treat them the same way?

Let me test it (later today).

filanov: `-o` should override everything because the user intend to run only those plugins that he selected, so i think that the behavior is correct there.
`-e` is a strange flag from my point of view, it means that the plugin didn't pass needed checks in order to run and the user still want to run it as part of the global batch, regardless i the issue exist there and i will push a fix for that as well

pmoravec: `-o` could be also interpreted as a concatenation of plugins the user wants to run only - i.e. I can call `sos report -o host,networking`. Should this *extend* the option in config file, or *replace* it? Both interpretations are valid, and I dont know what we should consider as the "right behaviour".

`-e` is really rather used in testing, or when something gets broken (what breaks the plugin enablement condition), but the "extend" approach should be the right approach for sure.

filanov: `-e` i agree 
`-o` i would suggest to leave it as is for this patch

do you think that the current change is good to go?

pmoravec (approved):
I am ok with `-o` on cmdline overwriting the config option (and not concatenating to it), until Jake or Arif disagrees.

TurboTurtle (approved):
LGTM.

I agree that `-o` should override in all cases.

Issue #4189, opened 2025-12-20
Support for foremanctl


pmoravec:
`sos report` should collect data from `foremanctl`, a [foreman deployment using Podman and Ansible](https://github.com/theforeman/foremanctl). I am [planning](https://github.com/theforeman/foremanctl/issues/49#issuecomment-3558161005) to implement a sos plugin for it, but I have a dilema.

My original plan was to follow same approach we did with `aap_containerized` plugin, where one plugin replaces several other `aap_*` plugins. But `foremanctl` puts into containers stuff being collected by quite different plugins `foreman` (but NOT `foreman-proxy`), `candlepin` and `pulpcore`, plus "supporting" plugins for `redis` and `postgresql`. Having one `foremanctl` plugin might hunt too many different tails, plus the plugin run time might be quite big (pulpcore and foreman plugins are usually lengthy to run).

An alternative is to enhance the current plugins to support their containerised deployments. But that is not great either: it would add complexity to the current plugins' code (esp. `foreman` one).

As a reference, here is the list of containers under a freshly installed `foremanctl`:

```
# podman ps --sort names
CONTAINER ID  IMAGE                                    COMMAND               CREATED      STATUS                PORTS                   NAMES
2e17744d9934  quay.io/foreman/candlepin:foreman-3.16   /usr/libexec/tomc...  3 hours ago  Up 3 hours (healthy)                          candlepin
df7671bdd6e3  quay.io/foreman/foreman:3.16             /usr/libexec/fore...  4 hours ago  Up 4 hours            3000/tcp                dynflow-sidekiq-orchestrator
58593d0ad109  quay.io/foreman/foreman:3.16             /usr/libexec/fore...  4 hours ago  Up 4 hours            3000/tcp                dynflow-sidekiq-worker
2053b0d7021e  quay.io/foreman/foreman:3.16             /usr/libexec/fore...  4 hours ago  Up 4 hours            3000/tcp                dynflow-sidekiq-worker-hosts-queue
3ed6c8b24505  quay.io/foreman/foreman:3.16             /bin/sh -c /usr/s...  2 hours ago  Up 2 hours            3000/tcp                foreman
5106a2355af9  quay.io/sclorg/postgresql-13-c9s:latest  run-postgresql        2 hours ago  Up 2 hours (healthy)  5432/tcp                postgresql
ebc5ed7e3176  quay.io/foreman/pulp:foreman-3.16        pulp-api              4 hours ago  Up 4 hours                                    pulp-api
[...]

PR #4330, opened 2026-05-15
[foreman*|pulp|candlepin] support foremanctl deployments

pmoravec:
A merge-able implementation of #4189. When reviewing all stuff being collected in `foreman` plugin, I did remove some parts that are not further relevant (passenger, qpid).

I am kindy asking @evgeni / @adamruzicka / @ehelms for a review of the `TODO` questions as I dont know how (or if) to collect some stuff.

All other plugin changes (pulp, candlepin,..) included, as well as a new `foremanctl` plugin is implemented.

---

### DIFF
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -2216,7 +2216,7 @@ class Plugin():
                 ocmd = cmd
                 container_cmd = (ocmd, container)
                 cmd = self.fmt_container_cmd(container, cmd, runtime=runtime,
-                                             runas=runas)
+                                             runas=runas, env=env)
                 if not cmd:
                     self._log_debug(f"Skipping command '{ocmd}' as the "
                                     f"requested container '{container}' does "
@@ -2733,7 +2733,7 @@ class Plugin():
                 return _default
             if self.container_exists(container, runtime) or runas is not None:
                 cmd = self.fmt_container_cmd(container, cmd, quotecmd,
-                                             runtime, runas)
+                                             runtime, runas, env=_env)
             else:
                 self._log_info(f"Cannot run cmd '{cmd}' in container "
                                f"{container}: no such container is running.")
@@ -2958,7 +2958,7 @@ class Plugin():
                     self.add_cmd_output(cmd, **kwargs)
 
     def fmt_container_cmd(self, container, cmd, quotecmd=False, runtime=None,
-                          runas=None):
+                          runas=None, env=None):
         """Format a command to be executed by the loaded ``ContainerRuntime``
         in a specified container
 
@@ -2979,6 +2979,9 @@ class Plugin():
                             the container really runs (we dont keep them atm)
         :type runas: ``str``
 
+        :param env:         OS environment variables to set in the container
+        :type env: ``dict``
+
         :returns: The command to execute so that the specified `cmd` will run
                   within the `container` and not on the host
         :rtype: ``str``
@@ -2986,7 +2989,7 @@ class Plugin():
         _runtime = self._get_container_runtime(runtime)
         if _runtime and (self.container_exists(container, runtime) or
            runas is not None):
-            return _runtime.fmt_container_cmd(container, cmd, quotecmd)
+            return _runtime.fmt_container_cmd(container, cmd, quotecmd, env)
         return ''
 
     def is_module_loaded(self, module_name):
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -2109,7 +2109,7 @@ class Plugin():
                 suggest_filename=suggest_filename
             )
 
-    def add_cmd_output(self, cmds, suggest_filename=None,
+    def add_cmd_output(self, cmds, suggest_filename=None, stdin=None,
                        root_symlink=None, timeout=None, stderr=True,
                        chroot=True, runat=None, env=None, binary=False,
                        sizelimit=None, pred=None, subdir=None,
@@ -2137,6 +2137,9 @@ class Plugin():
 
         :param timeout: Timeout in seconds to allow each `cmd` to run for
         :type timeout: ``int``
+
+        :param stdin: Supply stdin from this parameter
+        :type stdin: ``str``
 
         :param stderr: Should stderr output be collected
         :type stderr: ``bool``
@@ -2221,13 +2224,14 @@ class Plugin():
                     continue
             self._add_cmd_output(cmd=cmd, suggest_filename=suggest_filename,
                                  root_symlink=root_symlink, timeout=timeout,
-                                 stderr=stderr, chroot=chroot, runat=runat,
-                                 env=env, binary=binary, sizelimit=sizelimit,
-                                 pred=pred, subdir=subdir, tags=tags,
-                                 changes=changes, foreground=foreground,
-                                 priority=priority, cmd_as_tag=cmd_as_tag,
-                                 to_file=to_file, container_cmd=container_cmd,
-                                 runas=runas, snap_cmd=snap_cmd)
+                                 stdin=stdin, stderr=stderr, chroot=chroot,
+                                 runat=runat, env=env, binary=binary,
+                                 sizelimit=sizelimit, pred=pred, subdir=subdir,
+                                 tags=tags, changes=changes,
+                                 foreground=foreground, priority=priority,
+                                 cmd_as_tag=cmd_as_tag, to_file=to_file,
+                                 container_cmd=container_cmd, runas=runas,
+                                 snap_cmd=snap_cmd)
 
     def add_cmd_tags(self, tagdict):
         """Retroactively add tags to any commands that have been run by this
@@ -2384,7 +2388,7 @@ class Plugin():
         self._log_debug(f"added string as '{filename}'")
 
     def _collect_cmd_output(self, cmd, suggest_filename=None,
-                            root_symlink=False, timeout=None,
+                            root_symlink=False, timeout=None, stdin=None,
                             stderr=True, chroot=True, runat=None, env=None,
                             binary=False, sizelimit=None, subdir=None,
                             changes=False, foreground=False, tags=[],
@@ -2401,6 +2405,7 @@ class Plugin():
                                         archive
             :param root_symlink:        Create a symlink in the archive root
             :param timeout:             Time in seconds to allow a cmd to run
+            :param stdin:               Supply stdin from this parameter.
             :param stderr:              Write stderr to stdout?
             :param chroot:              Perform chroot before running cmd?
             :param runat:               Run the command from this location,
@@ -2466,7 +2471,7 @@ class Plugin():
         start = time()
 
         result = sos_get_command_output(
-            cmd, timeout=timeout, stderr=stderr, chroot=root,
+            cmd, timeout=timeout, stdin=stdin, stderr=stderr, chroot=root,
             chdir=runat, env=_env, binary=binary, sizelimit=sizelimit,
             poller=self.check_timeout, foreground=foreground,
             to_file=out_file, tac=tac, runas=runas
@@ -2507,7 +2512,8 @@ class Plugin():
                     result = sos_get_command_output(
                         cmd, timeout=timeout, chroot=False, chdir=runat,
                         env=env, binary=binary, sizelimit=sizelimit,
-                        poller=self.check_timeout, to_file=out_file, tac=tac,
+                        stdin=stdin, poller=self.check_timeout,
+                        to_file=out_file, tac=tac,
                     )
                     run_time = time() - start
             self._log_debug(f"could not run '{cmd}': command not found")
@@ -2567,7 +2573,7 @@ class Plugin():
         return result
 
     def collect_cmd_output(self, cmd, suggest_filename=None,
-                           root_symlink=False, timeout=None,
+                           root_symlink=False, timeout=None, stdin=None,
                            stderr=True, chroot=True, runat=None, env=None,
                            binary=False, sizelimit=None, pred=None,
                            changes=False, foreground=False, subdir=None,
@@ -2587,6 +2593,9 @@ class Plugin():
 
         :param timeout:             Time in seconds to allow a cmd to run
         :type timeout: ``int``
+
+        :param stdin:               Supply stdin from this parameter
+        :type stdin: ``str``
 
         :param stderr:              Write stderr to stdout?
         :type stderr: ``bool``
@@ -2637,13 +2646,13 @@ class Plugin():
 
         return self._collect_cmd_output(
             cmd, suggest_filename=suggest_filename, root_symlink=root_symlink,
-            timeout=timeout, stderr=stderr, chroot=chroot, runat=runat,
-            env=env, binary=binary, sizelimit=sizelimit, foreground=foreground,
-            subdir=subdir, tags=tags, runas=runas
+            timeout=timeout, stdin=stdin, stderr=stderr, chroot=chroot,
+            runat=runat, env=env, binary=binary, sizelimit=sizelimit,
+            foreground=foreground, subdir=subdir, tags=tags, runas=runas
         )
 
-    def exec_cmd(self, cmd, timeout=None, stderr=True, chroot=True,
-                 runat=None, env=None, binary=False, pred=None,
+    def exec_cmd(self, cmd, timeout=None, stderr=True, stdin=None, chroot=True,
+                 runat=None, env=None, binary=False, pred=None, sizelimit=None,
                  foreground=False, container=False, quotecmd=False,
                  runas=None, runtime=None):
         """Execute a command right now and return the output and status, but
@@ -2661,6 +2670,9 @@ class Plugin():
         :param stderr:              Write stderr to stdout?
         :type stderr: ``bool``
 
+        :param stdin:               Supply stdin from this parameter
+        :type stdin: ``str``
+
         :param chroot:              Perform chroot before running cmd?
         :type chroot: ``bool``
 
@@ -2676,6 +2688,9 @@ class Plugin():
 
         :param pred:                A predicate to gate execution of the `cmd`
         :type pred: ``SoSPredicate``
+
+        :param sizelimit:           Maximum amount of output in MB to save
+        :type sizelimit: ``int``
 
         :param foreground:          Run the `cmd` in the foreground with a TTY
         :type foreground: ``bool``
@@ -2725,7 +2740,8 @@ class Plugin():
 
         return sos_get_command_output(cmd, timeout=timeout, chroot=root,
                                       chdir=runat, binary=binary, env=_env,
-                                      foreground=foreground, stderr=stderr,
+                                      foreground=foreground, stdin=stdin,
+                                      stderr=stderr, sizelimit=sizelimit,
                                       runas=runas)
 
     def _add_container_file_to_manifest(self, container, path, arcpath, tags):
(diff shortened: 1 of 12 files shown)


sos/report/plugins/foreman.py
```diff
@@ -41,15 +42,28 @@ class Foreman(Plugin):
...
         # production:
         # ..
         #   host: some.hostname
         production_scope = False
         try:
             foreman_db = '/etc/foreman/database.yml'
-            with open(foreman_db, 'r', encoding='UTF-8') as dfile:
-                foreman_lines = dfile.read().splitlines()
+            if self.container:
+                dbout = self.exec_cmd(f"podman exec foreman cat {foreman_db}")
```
evgeni: there is no `/etc/foreman/database.yml'` inside the container, we set the database  url via environment vars/secrets

sos/report/plugins/foreman.py
```diff
@@ -92,22 +109,24 @@ def setup(self):
             "/etc/foreman/registry-auth.json",
         ])
 
+        # TODO: is it required in foremanctl?
```
evgeni: i'd keep it

sos/report/plugins/foreman.py
```diff
@@ -92,22 +109,24 @@ def setup(self):
...
         _hostname = _hostname.strip()
         _host_f = self.exec_cmd('hostname -f')['output']
         _host_f = _host_f.strip()
 
-        self.add_copy_spec([
-            "/var/log/foreman/production.log",
-            f"/var/log/{self.apachepkg}*/foreman-ssl_*_ssl.log"
-        ], sizelimit=500)
+        self.add_copy_spec(["/var/log/foreman/production.log"],
+                           container=self.container, sizelimit=500)
```
evgeni: there is no var log blah in the container, we log to stdout which gets collected by the journal on the host as the `foreman.service`

sos/report/plugins/foreman.py
```diff
@@ -92,22 +109,24 @@ def setup(self):
...
 
         # Allow limiting these
         self.add_copy_spec([
             "/etc/foreman/",
-            "/etc/sysconfig/foreman",
-            "/etc/sysconfig/dynflowd",
-            "/etc/default/foreman",
+            "/etc/sysconfig/foreman", # TODO: is it required at all?
+            "/etc/sysconfig/dynflowd", # TODO: is it required at all?
+            "/etc/default/foreman", # TODO: is it required at all?
```
evgeni: nuke it

sos/report/plugins/foreman.py
```diff
@@ -126,30 +148,21 @@ def setup(self):
             f"/var/log/{self.apachepkg}*/katello-reverse-proxy_access_ssl.log*"
         ])
 
+        # TODO: is either of 4 cmds required in foremanctl?
```
evgeni: probably not

sos/report/plugins/foreman.py
```diff
@@ -126,30 +148,21 @@ def setup(self):
...
-        ], recursive=True)
-
-        self.add_cmd_output(
-            'qpid-stat -b amqps://localhost:5671 -q \
-                    --ssl-certificate=/etc/pki/katello/qpid_router_client.crt \
-                    --ssl-key=/etc/pki/katello/qpid_router_client.key \
-                    --sasl-mechanism=ANONYMOUS',
-            suggest_filename='qpid-stat_-q'
-        )
+            '/root/ssl-build', # TODO: what to collect in foremanctl?
```
evgeni: /var/lib/foremanctl(/certs)

sos/report/plugins/foreman.py
```diff
@@ -126,30 +148,21 @@ def setup(self):
...
+            '/usr/share/foreman/config/hooks', # TODO: is it required at all?
```
evgeni: gone

sos/report/plugins/foreman.py
```diff
@@ -126,30 +148,21 @@ def setup(self):
...
+            '/var/lib/foreman/red_hat_inventory',
```
evgeni: this won't exist inside the container

sos/report/plugins/foreman.py
pmoravec: I see such file present, though apparently with some default content only.

How to retrieve the database url + credentials from the container? Or can we rely on (so far) the container runs on localhost:5432 ?

sos/report/plugins/foreman.py
pmoravec: Ack, now I see the `podman exec foreman cat /var/log/foreman/production.log` is almost empty..

sos/report/plugins/foreman.py
evgeni: yeah, it might be there, but it won't have any useful content :)

`podman exec foreman env | grep DATABASE_URL` should give you something workable. but that's certainly not a nice interface.

sos/report/plugins/foreman.py
pmoravec: Yeah, that was my impression as well. Originally, it was used for troubleshooting DNS issues where foreman didnt ping its own hostname. But there is an installer check (afaik), for standalone deployments, and this kind of issues should not happen in foremanctl.

For legacy reasons, I will leave it there just for standalone deployments (just in case..).

sos/report/plugins/foreman_proxy.py
```diff
@@ -33,8 +40,9 @@ def setup(self):
 
         self.add_copy_spec([
             "/etc/foreman-proxy/",
-            "/etc/smart_proxy_dynflow_core/settings.yml",
             "/var/log/foreman-proxy/*log*",
```
evgeni: as with foreman itself, this is now routed to the journal

sos/report/plugins/foreman.py
adamruzicka: `podman inspect foreman | jq '.[] | .Config | .Env'` not the greatest of interfaces either, but eh

sos/report/plugins/foreman.py
```diff
@@ -92,22 +109,27 @@ def setup(self):
...
+            _host_f = self.exec_cmd('hostname -f')['output']
+            _host_f = _host_f.strip()
+            self.add_cmd_output([
+                'foreman-selinux-relabel -nv',
+                f'ping -c1 -W1 {_hostname}',
+                f'ping -c1 -W1 {_host_f}',
+                'ping -c1 -W1 localhost'
+            ])
+
+        self.add_journal(units="foreman.service", sizelimit=500)
```
adamruzicka: Ideally also all of these
```
# systemctl list-units | grep dynflow-sidekiq@ | awk '{ print $1 }'
dynflow-sidekiq@orchestrator.service
dynflow-sidekiq@worker-hosts-queue.service
dynflow-sidekiq@worker.service
```

Or maybe even better, logs of all the services forming foreman.target?
```
# systemctl list-dependencies foreman.target | grep -Po '[a-zA-Z].*'
foreman.target
candlepin.service
dynflow-sidekiq@orchestrator.service
dynflow-sidekiq@worker-hosts-queue.service
dynflow-sidekiq@worker.service
foreman-proxy.service
foreman-recurring@daily.service
foreman-recurring@daily.timer
foreman-recurring@hourly.service
foreman-recurring@hourly.timer
foreman-recurring@monthly.service
foreman-recurring@monthly.timer
foreman-recurring@weekly.service
foreman-recurring@weekly.timer
foreman.service
postgresql.service
pulp-api.service
pulp-content.service
pulp-worker@1.service
pulp-worker@2.service
pulp-worker@3.service
pulp-worker@4.service
pulp-worker@5.service
pulp-worker@6.service
redis.service
pulp-worker.target
pulp-worker@1.service
pulp-worker@2.service
pulp-worker@3.service
pulp-worker@4.service
pulp-worker@5.service
pulp-worker@6.service
```

sos/report/plugins/foreman.py
```diff
@@ -288,19 +299,36 @@ def collect_foreman_db(self):
...
+            if dyn != 'foreman_tasks_tasks' and self.container:
+                cmdout = self.exec_cmd(
+                    _cmd, timeout=600, sizelimit=100,
+                    env=self.env, container=self.dbcontainer)
+                self.add_cmd_output(
+                    'podman exec -i foreman /usr/libexec/dynflow-expand',
+                    suggest_filename=dyn,
+                    timeout=600,
+                    stdin=cmdout['output'],
+                )
```
adamruzicka, line 341: Would it be possible to pipe the two together in one go? I'm pretty sure it could be done in a hack way, this question is more about doing it cleanly.

sos/report/plugins/foreman.py
pmoravec:
```
# find /var/lib/foremanctl/
/var/lib/foremanctl/
/var/lib/foremanctl/parameters.yaml
/var/lib/foremanctl/foreman-oauth-consumer-key
/var/lib/foremanctl/foreman-oauth-consumer-secret
/var/lib/foremanctl/.installed
#
# find / -name certs
/etc/pki/tls/certs
/etc/pki/httpd/certs
/etc/openldap/certs
/etc/ssl/certs
/root/certificates/certs
.. (and plenty of /var/lib/containers/storage/overlay subdirs)
#
```

The `/root/certificates/certs` seems to have foreman(ctl) generated certs. So worth rather recursively listing this dir..?

evgeni: Trust me, the certs will be in var lin the next time ;)

sos/report/plugins/foreman.py
pmoravec: the pod env sounds good, thx

sos/report/plugins/foreman_proxy.py
pmoravec: TY, fixed!

sos/utilities.py
```diff
@@ -340,66 +344,79 @@
...
+                    'status': p.returncode,
+                    'output': stdout,
+                    'truncated': reader.is_full
+                }
+            finally:
+                if stdin_writer is not None:
+                    stdin_writer.join()
+                    if stdin_writer.error:
+                        log.warning(f"Passing stdin to command {command} "
+                                    f"raised an error: {stdin_writer.error}")
```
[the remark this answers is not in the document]
pmoravec: Hi @arif-ali and @TurboTurtle , does the CodeQL error makes sense to you? IMHO a red herring, since we log just command (logged else-where,here also to stdout) and `(BrokenPipeError, ValueError)` exception details.

sos/report/plugins/foreman.py
pmoravec: Cool, collecting it in the new foremanctl plugin, then.

sos/report/plugins/foreman.py
evgeni: you could also do `podman inspect foreman --format '{{json .Config.Env}}'` and then parse the result with `json.loads` :)

sos/report/plugins/foreman.py
pmoravec: We collect `self.add_journal(units="dynflow-sidekiq@*")` on line 179. `candlepin`, `foreman-*` and `pulp-*` shall be rather collected by the respective plugins (so I will extend the `foreman_proxy` one, candlepin and pulp is on the TODO list).

both redis and postgresql journals make sense to collect only in foremanctl, right? It makes bit more sense to collect them in foreman plugin as a special use case, since redis and postgresql plugins dont collect the journals by default.

sos/report/plugins/foreman.py
pmoravec: @TurboTurtle some thoughts? I know there are some know issues when collecting commands with pipes, so we try to avoid that as much as we can.

Moreover, plain:

```
self.add_cmd_output(f'podman exec postgresql {_cmd} | podman exec -i foreman /usr/libexec/dynflow-expand', env=self.env,suggest_filename=f'piped_{dyn}', timeout=600)
```

does not work (`psql: invalid option -- 'i'`), regardless if I add ` -` to tail of the 2nd `podman` command, and:

```
podman exec -i foreman /usr/libexec/dynflow-expand < <(podman exec postgresql {_cmd})
```

gets stuck.

I dont like the proposed approach, but dont see anything better (that would work).

pmoravec:
Kindly asking @arif-ali and/or @TurboTurtle for a review - I would like to fix this regression till the next release scheduled for the next week.

sos/report/plugins/__init__.py
```diff
@@ -2637,13 +2646,13 @@ def collect_cmd_output(self, cmd, suggest_filename=None,
...
-            subdir=subdir, tags=tags, runas=runas
+            timeout=timeout, stdin=stdin, stderr=stderr, chroot=chroot,
+            runat=runat, env=env, binary=binary, sizelimit=sizelimit,
+            foreground=foreground, subdir=subdir, tags=tags, runas=runas
         )
 
-    def exec_cmd(self, cmd, timeout=None, stderr=True, chroot=True,
-                 runat=None, env=None, binary=False, pred=None,
+    def exec_cmd(self, cmd, timeout=None, stderr=True, stdin=None, chroot=True,
+                 runat=None, env=None, binary=False, pred=None, sizelimit=None,
```
arif-ali, line 2655: we've added sizelimit here, do  we need to add docs below for both stdin and sizelimit

sos/report/plugins/candlepin.py
```diff
@@ -71,10 +90,24 @@ def setup(self):
             "/var/log/candlepin/candlepin.log"
         ], sizelimit=0)
 
+        if self.container:
+            self.add_journal(units="candlepin.service", sizelimit=500)
```
arif-ali: is is worth adding more lines if `--all-logs` option has been defined?

sos/report/plugins/foreman.py
```diff
@@ -92,22 +119,30 @@ def setup(self):
...
+                'ping -c1 -W1 localhost'
+            ])
 
         self.add_copy_spec([
             "/var/log/foreman/production.log",
             f"/var/log/{self.apachepkg}*/foreman-ssl_*_ssl.log"
         ], sizelimit=500)
+        # foremanctl logs to host's journal instead of production.log
+        if self.container:
+            self.add_journal(units="foreman.service", sizelimit=500)
```
arif-ali: same as above, is it worth adding more logs if `--all-logs` has been specified

sos/report/plugins/foremanctl.py
```diff
@@ -0,0 +1,42 @@
...
+
+    plugin_name = 'foremanctl'
+    profiles = ('sysmgmt',)
+    packages = ('foremanctl', )
+
+    def setup(self):
+        self.add_copy_spec([
+            "/etc/foremanctl/inventory",
+            "/var/lib/foremanctl/parameters.yaml",
+            "/var/log/foremanctl/foremanctl.*log",
```
arif-ali, line 27: and there is any rotated logs, that can be there? maybe behind `--all-logs`?

arif-ali (commented):
overall looks good to me, just a few comments

sos/report/plugins/foremanctl.py
pmoravec: They are already included in the glob:

```
# ls -tr1 /var/log/foremanctl/foremanctl*log
/var/log/foremanctl/foremanctl.2026-04-13-01-05-14.log
/var/log/foremanctl/foremanctl.2026-04-13-01-06-14.log
/var/log/foremanctl/foremanctl.2026-04-13-01-07-03.log
/var/log/foremanctl/foremanctl.2026-04-13-01-07-28.log
/var/log/foremanctl/foremanctl.2026-05-15-19-05-33.log
/var/log/foremanctl/foremanctl.2026-05-15-19-05-46.log
/var/log/foremanctl/foremanctl.2026-05-15-19-06-33.log
/var/log/foremanctl/foremanctl.2026-05-29-11-26-17.log
/var/log/foremanctl/foremanctl.log
#
```

sos/report/plugins/candlepin.py
```diff
@@ -71,10 +90,25 @@ def setup(self):
...
+        self.add_copy_spec([
+                "/etc/candlepin/candlepin.conf",
+                "/etc/candlepin/broker.xml",
+        ], container=self.container)
+
+        if self.container:
+            self.add_cmd_output(
+                'podman exec candlepin sh -c '
+                '"du -sh /var/lib/candlepin/*/*"'
+            )
```
TurboTurtle: Does this hit issues if you do something like:

```python
self.add_cmd_output(
    'du -sh /var/lib/candlepin/*/*',
    container=self.container
)
```

That way we leverage the container runtime abstraction, and can run it outside of a container without the if/else conditional?

sos/report/plugins/foreman.py
```diff
@@ -126,30 +165,11 @@ def setup(self):
...
-            'passenger-status --show backtraces',
-            'passenger-memory-stats',
-            f'ping -c1 -W1 {_hostname}',
-            f'ping -c1 -W1 {_host_f}',
-            'ping -c1 -W1 localhost'
-        ])
+        if not self.container:
+            self.add_dir_listing(['/root/ssl-build',
+                                  '/var/lib/foreman/red_hat_inventory'],
+                                 recursive=True)
```
TurboTurtle, line 171: Minor style nit: over-indented here.

sos/report/plugins/pulpcore.py
```diff
@@ -82,42 +84,91 @@ def separate_value(line, sep=':'):
...
+                for item in env_list
+                if isinstance(item, str)
+                and item.startswith("PULP_DATABASES__default__")
+                and '=' in item
+            )
+
+        pod_inspect = self.exec_cmd(
+                          f"podman inspect {self.pulp_container} "
+                          "--format='{{json .Config.Env}}'"
+                      )
```
TurboTurtle: Minor style nit: over-indented

TurboTurtle (commented):
Overall, looks good.

One question about calling `podman exec` in an a_c_o (for `exec_cmd()` no issue), and then some super minor style nits.

If no change for the a_c_o call, then I won't block on the style nits.

sos/report/plugins/candlepin.py
pmoravec: We have tried several simplier approaches that all fail, until this one. The problem is, '*' is often wrongly evaluated on the host, not inside the container - all these fail with `du: cannot access '/var/lib/candlepin/*/*': No such file or directory`:

```
podman exec candlepin du -hs /var/lib/candlepin/*/*
podman exec candlepin du -hs '/var/lib/candlepin/*/*'
podman exec candlepin du -hs /var/lib/candlepin/\*/\*
```

**Anyway**, invoking `sh -c` works :) I will improve this. Though still the `sh -c` is required only in container, so.. no big simplification gained.

TurboTurtle: Just a thought. What if within `ContainerRuntime.fmt_container_cmd()` we allow podman (and docker presumably) to specify the `-t` option to exec? Does that remove the need to explicitly call `sh` here?

pmoravec: I tried to pass `foreground` from `a_c_o` into `-t` option in `ContainerRuntime.fmt_container_cmd()`, but I got:

```
# cat /var/tmp/sosreport-pmoravec-foremanctl-2026-06-13-vzbnsba/sos_containers/candlepin/sos_commands/candlepin/du_-hs_.var.lib.candlepin 
du: cannot access '/var/lib/candlepin/*/*': No such file or directory
#
```

arif-ali (approved):
lgtm

TurboTurtle (approved):

PR #4332, opened 2026-05-19
Add preset for Canonical OpenStack with Sunbeam

### LINKED ISSUE #4331: Issues with preset on Debian and Ubuntu
When trying to create my own presets on Ubuntu 20.04 using sos v4.10.0 with the --add-preset pname, the --list-presets option does not recognize it.  It creates a file in /etc/sos/presets.d/pname (with no .json extension) but it does not recognize it.

Trying to execute the preset with --preset pname does not work either.

`> sudo sos report --add-preset pname --desc "a description" --note "a more detailed description" --only-plugins=host,kernel,memory

sos report (version 4.10.0)

Added preset 'pname' with options report --desc a description --note a more detailed description --only-plugins=host,kernel,memory

> ls -l /etc/sos/presets.d/
total 4
-rw-r--r-- 1 root root 1701 May 19 22:15 pname
> sudo sos report --list-presets

sos report (version 4.10.0)

The following presets are available:

         name: none
  description: Do not load a preset
         note: Use to disable automatically loaded presets

         name: minimal
  description: Small and quick report that reduces sos report resource consumption
         note: May be useful for low-resource systems, but may not provide sufficient data for analysis

> sudo sos report --preset pname
Unknown preset: 'pname'

sos report (version 4.10.0)

The following presets are available:

         name: none
  description: Do not load a preset
         note: Use to disable automatically loaded presets

         name: minimal
  description: Small and quick report that reduces sos report resource consumption
         note: May be useful for low-resource systems, but may not provide sufficient data for analysis

> `

Looking at the code it seem like  load_presets() is only called in redhat.py — it is never called in the Ubuntu/Debian policy.

I can confirm that this does not happens on AlmaLinux.

arif-ali:
Also loads presets for both Ubuntu and Debian, so that local presets can also be loaded.

Closes: #4331

---

### DIFF
--- a/sos/presets/ubuntu/__init__.py
+++ b/sos/presets/ubuntu/__init__.py
@@ -1,0 +1,36 @@
+# Copyright (C) 2026 Canonical Ltd., Arif Ali <arif-ali@ubuntu.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+from sos.options import SoSOptions
+from sos.presets import PresetDefaults
+
+
+UBUNTU = "ubuntu"
+UBUNTU_DESC = "Ubuntu"
+
+SUNBEAM = "sunbeam"
+SUNBEAM_DESC = "Canonical OpenStack"
+SUNBEAM_OPTS = SoSOptions(
+    all_logs=True,
+    plugopts=[
+        'sunbeam.juju-allow-login = True',
+        'kubernetes.all = True',
+        'kubernetes.describe = True',
+        'kubernetes.kubelogs = True',
+        'kubernetes.podlogs = True',
+    ])
+
+UBUNTU_PRESETS = {
+    SUNBEAM: PresetDefaults(name=SUNBEAM, desc=SUNBEAM_DESC,
+                            opts=SUNBEAM_OPTS),
+    UBUNTU: PresetDefaults(name=UBUNTU, desc=UBUNTU_DESC),
+}
+
+# vim: set et ts=4 sw=4 :
--- a/sos/policies/distros/debian.py
+++ b/sos/policies/distros/debian.py
@@ -44,6 +44,7 @@ class DebianPolicy(LinuxPolicy):
         self.package_manager = DpkgPackageManager(chroot=self.sysroot,
                                                   remote_exec=remote_exec)
         self.valid_subclasses += [DebianPlugin]
+        self.load_presets()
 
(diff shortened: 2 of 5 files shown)


sos/presets/ubuntu/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+SUNBEAM = "sunbeam"
+SUNBEAM_DESC = "Canonical OpenStack"
+SUNBEAM_OPTS = SoSOptions(
+    all_logs=True,
+    plugopts=[
+        'sunbeam.juju-allow-login=True',
+        'kubernetes.all = True',
+        'kubernetes.describe = True',
+        'kubernetes.kubelogs = True',
+        'kubernetes.podlogs = True',
```
pmoravec, line 27: Are the spaces around `=` allowed? When I mimic the same in Red Hat Satellite preset:

```
SAT_OPTS = SoSOptions(log_size=100, plugopts=['apache.log = on', 'networking.traceroute = True'])
```

I get error:

```
# sos report -l

sos report (version 4.11.0)

no such option "log " for plugin (apache)
#
```

arif-ali: I thought I tested it, obviously not :)

agreed, it's an issue, but maybe that's an issue in itself, and we should solve that. I tested the below patch, which works for me now. So this way it truly behaves like a variable, and we can have spaces. 

```diff
diff --git a/sos/report/__init__.py b/sos/report/__init__.py
index 20f314c0..05073d1b 100644
--- a/sos/report/__init__.py
+++ b/sos/report/__init__.py
@@ -960,15 +960,17 @@ class SoSReport(SoSComponent):
             for plugname, plug in self.loaded_plugins:
                 if plugname in opts:
                     for opt in opts[plugname]:
+                        opt = opt.strip()
                         if opt not in plug.options:
                             self.soslog.error(f'no such option "{opt}" for '
                                               f'plugin ({plugname})')
                             self._exit(1)
                         try:
-                            plug.options[opt].set_value(opts[plugname][opt])
+                            plug.options[opt].set_value(
+                                opts[plugname][opt].strip())
                             self.soslog.debug(
                                 f"Set {plugname} plugin option to "
-                                f"{plug.options[opt]}")
+                                f"{plug.options[opt].strip()}")
                         except Exception as err:
                             self.soslog.error(err)
                             self._exit(1)
```

pmoravec: It wont work on cmdline args (obviusly), will work in presets and I guess also in config file - sounds good to me!

Let add a commit for that.

pmoravec (approved):
Maybe splitting the commit to preset and to plugopt enhancement would be beneficial, but I am ok as is.

(leaving on Arif to decide if it is worth splitting, will merge after 24 hours otherwise)

TurboTurtle (approved):

PR #4334, opened 2026-05-20
[cleaner] Match uppercase IPv6 addresses

### LINKED ISSUE #4163: [cleaner] Obfuscate IPv6 addresses with capital A-F letters as well
Per https://www.rfc-editor.org/rfc/rfc4291.html#page-4 , an IPv6 address contains hexadecimal digits, which can be uppercase as well. So https://github.com/sosreport/sos/blob/main/sos/cleaner/parsers/ipv6_parser.py#L28-L30 should `s/a-f/a-fA-F/g`.

Honestly, I havent seen IPv6 address in capitals, but that is not my well known domain. Hence raising issue so far if there is a consensus we should enhance the cleaner this way.

puneetdixit200:
## Summary
- match IPv6 cleaner regexes case-insensitively so uppercase hex digits are obfuscated
- add parser-only regression coverage for the uppercase case and existing false-positive patterns

Closes: #4163

## Validation
- red check before parser change: `python3 -m unittest tests.unittests.cleaner_tests.IPv6ParserRegexTests.test_ipv6_parser_uppercase_hex`
- `python3 -m unittest tests.unittests.cleaner_tests.IPv6ParserRegexTests`
- `python3 -m compileall -q sos/cleaner/parsers/ipv6_parser.py tests/unittests/cleaner_tests.py`
- `uvx flake8 sos tests`
- `git diff --check`
- Docker/Linux: `python -m unittest tests.unittests.cleaner_tests.IPv6ParserRegexTests tests.unittests.cleaner_tests.CleanerParserTests`
- Docker/Linux: `python -m unittest discover -s tests/unittests -p "*_tests.py"`
- Docker/Linux: `tox -e unit_tests` (187 passed, 2 skipped)

Assisted-by: OpenAI GPT-5 <https://openai.com>

### DIFF
--- a/sos/cleaner/parsers/ipv6_parser.py
+++ b/sos/cleaner/parsers/ipv6_parser.py
@@ -26,9 +26,12 @@ class SoSIPv6Parser(SoSCleanerParser):
         # is not extracted from a log message such as 'SomeFuncUsed::ADiffFunc'
         # that come components may log with. Further, we optionally try to grab
         # a trailing prefix for the network bits.
-        r"(?<![:\\.\\-a-z0-9])((([0-9a-f]{1,4})(:[0-9a-f]{1,4}){7})|"
-        r"(([0-9a-f]{1,4}(:[0-9a-f]{0,4}){0,5}))([^.])::(([0-9a-f]{1,4}"
-        r"(:[0-9a-f]{1,4}){0,5})?)(\/\d{1,3})?)(?!([a-z0-9]|:[a-z0-9]))"
+        r"(?<![:\\.\\-a-zA-Z0-9])"
+        r"((([0-9a-fA-F]{1,4})(:[0-9a-fA-F]{1,4}){7})|"
+        r"(([0-9a-fA-F]{1,4}(:[0-9a-fA-F]{0,4}){0,5}))"
+        r"([^.])::(([0-9a-fA-F]{1,4}"
+        r"(:[0-9a-fA-F]{1,4}){0,5})?)(\/\d{1,3})?)"
+        r"(?!([a-zA-Z0-9]|:[a-zA-Z0-9]))"
     )
     parser_skip_files = [
         'etc/dnsmasq.conf.*',
--- a/tests/unittests/cleaner_tests.py
+++ b/tests/unittests/cleaner_tests.py
@@ -302,6 +302,14 @@ class CleanerParserTests(unittest.TestCase):
         self.assertNotEqual(t4, t4_test,
                             f"Parser did not match and obfuscate '{t4}'")
 
+    def test_ipv6_parser_uppercase_hex(self):
+        line = 'testing 2001:DB8::ABCD as an uppercase address'
+        _test = self.ipv6_parser.parse_line(line)[0]
+        self.assertNotEqual(
+            line, _test,
+            f"Parser did not match and obfuscate '{line}'"
+        )
+
     def test_ipv6_no_match_signature(self):
         modstr = '2D:4F:6E:55:4F:E8:5E:D2:D2:A3:73:62:AB:FD:F9:C5:A5:53:31:93'
         mod_test = self.ipv6_parser.parse_line(modstr)[0]


sos/cleaner/parsers/ipv6_parser.py
```diff
@@ -28,7 +28,8 @@ class SoSIPv6Parser(SoSCleanerParser):
         # a trailing prefix for the network bits.
         r"(?<![:\\.\\-a-z0-9])((([0-9a-f]{1,4})(:[0-9a-f]{1,4}){7})|"
         r"(([0-9a-f]{1,4}(:[0-9a-f]{0,4}){0,5}))([^.])::(([0-9a-f]{1,4}"
-        r"(:[0-9a-f]{1,4}){0,5})?)(\/\d{1,3})?)(?!([a-z0-9]|:[a-z0-9]))"
+        r"(:[0-9a-f]{1,4}){0,5})?)(\/\d{1,3})?)(?!([a-z0-9]|:[a-z0-9]))",
+        re.IGNORECASE
```
pmoravec: Good point we should detect upper-case letters `A-F` in an IPv6 address. For performance reasons, wont be faster to have `s/a-f/a-fA-F/` inside the regular expression, rather than using the re.IGNORECASE?

I will run some perf.comparison, if I spot some difference..

pmoravec: OK, running kind of:

```
                with open(filepath, 'r', encoding='utf-8', errors='ignore') as f:
                    match_count += sum(len(pattern.findall(line)) for line in f)
```

for all files in 10 "benchmark" sosreports, the proposed re.IGNORECASE change took `23m42.766s` while extending `a-f` to `a-fA-F` and `a-zA-Z` directly, such `pattern` needed `16m9.950s` only.

Honestly, I expected much smaller difference, but in either way, please extend the RE directly.

tests/unittests/cleaner_tests.py
```diff
@@ -326,35 +327,38 @@ def test_too_short_username(self):
...
-            self.assertNotIn(name, result,
-                             f"Username '{name}' was not obfuscated in:"
-                             f" {result}")
+
+class IPv6ParserRegexTests(unittest.TestCase):
+
+    def parse_ipv6_line(self, line):
+        with TemporaryDirectory() as workdir:
+            parser = SoSIPv6Parser(config={}, workdir=workdir)
+            return parser.parse_line(line)[0]
```
pmoravec: Why not extending the `CleanerParserTests` class? And why the extra `TemporaryDirectory` requirement?

tests/unittests/cleaner_tests.py
```diff
@@ -326,35 +327,38 @@ def test_too_short_username(self):
...
+
+    def test_ipv6_parser_uppercase_hex(self):
+        line = 'testing 2001:DB8::ABCD as an uppercase address'
+        _test = self.parse_ipv6_line(line)
+        self.assertNotEqual(
+            line, _test,
+            f"Parser did not match and obfuscate '{line}'"
+        )
+
+    def test_ipv6_parser_ignores_module_signature(self):
```
pmoravec: This method functionally tests the same as `test_ipv6_no_match_signature`, no..?

tests/unittests/cleaner_tests.py
```diff
@@ -326,35 +327,38 @@ def test_too_short_username(self):
...
+        line = '2D:4F:6E:55:4F:E8:5E:D2:D2:A3:73:62:AB:FD:F9:C5:A5:53:31:93'
+        _test = self.parse_ipv6_line(line)
+        self.assertEqual(
+            line, _test,
+            "Parser matched module signature, and should not"
+        )
+
+    def test_ipv6_parser_ignores_log_false_positive(self):
```
pmoravec: .. and this `test_ipv6_no_match_log_false_positive`.

IMHO just the first test makes sense, to verify the cleaner detects an IPv6 address with capitals.

pmoravec (approved):
ACK, thanks for the patience.

Please squash the commits such that they dont iteratively change the same code - either squash all three commits into one, or have one commit for the cleaner change and one for the tests change.

arif-ali (approved):
LGTM, as suggested by Pavel, please squash commits

PR #4336, opened 2026-05-21
[Cleaner] Ensure Obfuscation of Domains in SSSD conf files

bryanfraschetti:

---

### DIFF
--- a/sos/cleaner/preppers/hostname.py
+++ b/sos/cleaner/preppers/hostname.py
@@ -8,6 +8,9 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
+import glob
+import os
+
 from sos.cleaner.preppers import SoSPrepper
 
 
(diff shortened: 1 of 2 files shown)


bryanfraschetti:
Hmm. I've stumbled on something strange here
I was developing the tests locally and all were passing except `test_hostname_prepper_sssd_conf_values_in_items`. After mocking `host_prepper._get_conf_files` since the files aren't actually on disk, all were passing. In CI `test_hostname_prepper_sssd_fqdn_fully_obfuscated` failed so I cleared `.tox` and `/var/tmp/sos_avocado_testing/` and now I reproduce the failure locally so there must some weird caching of either test results or host/domain names

sos/cleaner/preppers/hostname.py
```diff
@@ -25,6 +28,85 @@ class HostnamePrepper(SoSPrepper):
...
+            return None
+        # Handle URI-style values, e.g. ldap://host.example.com:389
+        if '://' in value:
+            value = value.split('://', 1)[1]
+        value = value.split('/')[0]
+        if value.startswith('['):
+            value = value.strip('[]')
+        elif ':' in value:
+            value = value.split(':', 1)[0]
+        return value.strip()
```
pmoravec: Can't `urlparse(value).hostname` help here? Dont reinvent wheel until you need some specific one (which might be the case here..?)

sos/cleaner/preppers/hostname.py
```diff
@@ -25,6 +28,85 @@ class HostnamePrepper(SoSPrepper):
...
+        items = []
+
+        paths = self._get_conf_files(archive)
+
+        for path in sorted(paths):
+            content = archive.get_file_content(path)
+            if not content:
+                continue
+            for line in content.splitlines():
+                line = line.split('#', 1)[0].split(';', 1)[0].strip()
```
pmoravec, line 92: maybe an explanation or an example of this split/strip helps..?

What if a user comments out whole line with a valid hostname? Technically the semantics of the config file is "no such (commented) hostname ever exists". Practically, the comment can be auxiliary and the hostname can be very valid (occurring in past logs, e.g.).

Not sure if my comment about `^#..` lines is much valid, or if it is worth adding complexity here for such a corner case (well, maybe simple `while line starts with '#', strip the char` can be prepped..?).

pmoravec:
Ack up to the few comments. Ack to your easter egg / name in tests :)

sos/cleaner/preppers/hostname.py
bryanfraschetti: I actually don't need this function, so I'll remove it

sos/cleaner/preppers/hostname.py
bryanfraschetti: That's true, I was treating comments as not having any sensitive content, but there may be previous entries that have temporarily been commented out or left there for legacy's sake. I think `while line starts with '#', strip the char` makes sense

I think it is less likely that inline comments will have any personally identifiable information and is more likely to be unstructured text. I'll update the PR so that we don't ignore lines that start with a comment

pmoravec (approved):

TurboTurtle (approved):

PR #4340, opened 2026-05-25
[cleaner] Ignore compiled regexes matching static regex_pattern

### LINKED ISSUE #4335: [cleaner] Most of compiled regexes are redundantly parsed twice
`mapping.compiled_regexes` is a union of individual regular expressions identified e.g. in `/etc/sos/cleaner/default_mapping` or by a prepper. If such a string *also* matches cleaner's `regex_pattern`, we check the string "twice" - once via `_parse_line_with_compiled_regexes` (https://github.com/sosreport/sos/blob/main/sos/cleaner/parsers/__init__.py#L96) - which is redundant here, since subsequent `_parse_line` (https://github.com/sosreport/sos/blob/main/sos/cleaner/parsers/__init__.py#L98) will detect it either way.

So, `mapping.compiled_regexes` and also mapping.compiled_search` do contain redundant strings, making the search redundantly complex.

Shouldn't we add items to them only when parser's `regex_pattern` doesnt match them?

This can bring huge performance improvement, since most of `default_mapping` records (that feed `compiled_regexes`) - being accumulated over the time - come from a `regex_pattern` match.

@TurboTurtle , is my idea sane? Should we enhance https://github.com/sosreport/sos/blob/main/sos/cleaner/mappings/__init__.py#L150 :

```
if item not in self._regexes_made:
```

to something like:

```
if item not in self._regexes_made and not parser.regex_pattern.search(item):
```

? (mapping isnt aware of parser, though, we would have to extend it in some way).

pmoravec:
Items that match both `mapping.compiled_regexes` and `parser.regex_pattern` are currently checked twice (by `_parse_line_with_compiled_regexes` and `_parse_line`). That is redundant.

Add to `mapping.compiled_regexes` only the items that don't match `parser.regex_pattern`, to make the `compiled_regex` simplier and faster to check.

Resolves: #4335
Closes: #4340

---

### DIFF
--- a/sos/cleaner/mappings/__init__.py
+++ b/sos/cleaner/mappings/__init__.py
@@ -35,7 +35,7 @@ class SoSMap():
 
     _token_split_re = re.compile(r'[^a-z0-9]+')
 
-    def __init__(self, workdir):
+    def __init__(self, workdir, _static_regex=re.compile(r'(?!)')):
         self.initializing = True
         self.dataset = {}
         self._regexes_made = set()
@@ -44,6 +44,7 @@ class SoSMap():
         self._simple_tokens = set()
         self._complex_items = []
         self._regex_dict = {}
+        self._static_regex = _static_regex
         self.cname = self.__class__.__name__.lower()
         # workdir's default value '/tmp' is used just by avocado tests,
         # otherwise we override it to /etc/sos/cleaner (or map_file dir)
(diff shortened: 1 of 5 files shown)


pmoravec:
I will provide some performance improvements stats once the tests complete. The main advantage is on use cases with big `default_mapping` - where majority of items match `parser.regex_pattern`, but still adding them to a huge `mapping.compiled_regexes`.

pmoravec:
This PR improves performance on average by 12% (with empty initial `default_mapping`) / 22% (with "own" initial `default_mapping`) / 37% (with "all mappings from 10 different sosreports" `default_mapping`).

So let say by 20% on the very most average / typical execution.

pmoravec:
Hi @TurboTurtle and @arif-ali , I would like to include this PR in the 4.11.2 to be tagged in 2 weeks. Could you please review this PR, brining 20% cleaner performance improvement?

Thank you.

arif-ali (approved):

PR #4347, opened 2026-05-28
[report] Add --pack-dir to pack verbose dirs into tarballs

### LINKED ISSUE #4346: Add `--compress-dir` option to compress verbose collected directories into tarballs
A new sos report option, --compress-dir, that packs one or more already-collected directories into a single compressed tarball in place within the report (e.g. proc/fs/  -> proc/fs.tar.xz) and removes the original tree. The rest of the archive is untouched.

gaurang1988:
- Add --pack-dir to pack collected directories into a single tar in
    place (e.g. proc/fs/ -> proc/fs.tar) and drop the original tree
- Avoids slow extraction of dirs like /proc/fs/ with millions of
    files: the top-level archive yields one member instead of millions
- Tar is left uncompressed; the whole report is already compressed
    when packaged, so compressing again only duplicates that work
- Accepts multiple dirs, comma-separated or by repeating the flag
- Runs after clean (obfuscation applies) and before packaging
-  Fixes: https://github.com/sosreport/sos/issues/4346
---

### DIFF
--- a/sos/archive.py
+++ b/sos/archive.py
@@ -602,6 +602,48 @@ class FileCacheArchive(Archive):
             replacements = 0
         return replacements
 
+    def tar_subdirs(self, paths):
+        """Replace one or more collected directories with a single tarball.
+
+        For each host path in ``paths`` that was collected into this archive,
+        the corresponding directory tree under the archive root is packed into
+        a single uncompressed tarball alongside it (e.g. ``proc/fs/`` becomes
+        ``proc/fs.tar``) and the original tree is removed. The rest of the
+        archive is left untouched.
+
+        The tarball is deliberately not compressed here: the final report
+        archive is compressed as a whole, so compressing again would only
+        duplicate that work and cannot shrink already-compressed data. The
+        point is to collapse a directory of many small files into one member,
+        so extracting the top-level archive does not have to recreate every
+        file on disk. This backs the report ``--pack-dir`` option, used
+        for verbose trees such as ``/proc/fs/``.
+
+        :param paths: Iterable of host directory paths to pack in place
+        :returns: A list of the host paths that were successfully packed
+        """
+        packed = []
+        for path in paths:
+            dest = self.dest_path(path.rstrip(os.sep))
+            if not os.path.isdir(dest):
+                self.log_info(f"skipping --pack-dir '{path}': not collected"
+                              " as a directory in the archive")
+                continue
+            tarpath = f"{dest}.tar"
+            try:
+                with tarfile.open(tarpath, mode="w") as tar:
+                    tar.add(dest, arcname=os.path.basename(dest))
+                shutil.rmtree(dest)
+                self.log_info(f"packed collected directory '{path}' into "
+                              f"'{tarpath}'")
+                packed.append(path)
+            except Exception as err:
+                self.log_error(f"Could not pack directory '{path}': {err}")
+                # leave the directory in place on failure
+                if os.path.exists(tarpath):
+                    os.unlink(tarpath)
+        return packed
+
     def finalize(self, method):
         self.log_info(f"finalizing archive '{self._archive_root}' using method"
                       f" '{method}'")
--- a/sos/cleaner/archives/__init__.py
+++ b/sos/cleaner/archives/__init__.py
@@ -8,6 +8,7 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
+import json
 import logging
 import os
 import shutil
@@ -74,6 +75,7 @@ class SoSObfuscationArchive():
         self.soslog = logging.getLogger('sos')
         self.ui_log = logging.getLogger('sos_ui')
         self.skip_list = self._load_skip_list()
+        self.packed_dirs = []
         self.is_extracted = False
         self._load_self()
         self.archive_root = ''
@@ -157,6 +159,17 @@ class SoSObfuscationArchive():
         try:
             rel_name = os.path.relpath(filename, start=self.extracted_path)
             if self.should_skip_file(rel_name):
+                return
+            if self.is_packed_dir_tarball(rel_name):
+                # tarballs produced by the report '--pack-dir' option hold
+                # collected data the user explicitly asked to retain. We can't
+                # obfuscate their contents in place, but unlike other binary
+                # files we must not drop them, or that collected data would be
+                # silently lost from the cleaned report.
+                self.log_warn(
+                    f"Keeping packed directory tarball '{rel_name}'; its "
+                    "contents are not obfuscated"
+                )
                 return
             if (not self.keep_binary_files and
                     self.should_remove_file(rel_name)):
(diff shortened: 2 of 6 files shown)


pmoravec:
How much it prolongs execution during `sos report`? (show time diff on an example sosreport)

How much time it will save on extraction? And how much time is needed to additionally untar the tar-ed paths? (that will usually not be needed, I understand, but would like the all pros and cons)?

Have you considered `sos clean` / `sos report --clean` will drop the tarball by default, after marking it as binary data? (should the PR detect the `packed_dirs` during cleaner, and untar+clean+tar ? or skip removing the dir (with the potential of *not* obfuscating sensitive data there?)?

TurboTurtle:
I'll continue my comments here instead of the related issue #4346 

In echo'ing @pmoravec's comments, I still do not follow the use case. Do you have a practical example of a large `/proc` filesystem causing extremely long extraction times? Even for high file counts, I can't see this producing extraction times that are impractical for reviewing engineers. Beyond that; from my Support days, I would have preferred to extract once and get everything, than potentially need to do multiple extractions as part of a workflow.

adilger:
On a 2-socket 128-core system, , an idle system has over 1300 processes running, which result in:
```
$ sudo find /proc | wc -l
422748
$ sudo find /sys | wc -l
149552
```
files in each report.  Many of the running processes are from the kernel and per CPU core, so will continue to grow. 

This is by no means the largest system today, compared to 256-core DGX servers with 8 GPUs, 8 NVMe, 8 NICs.
When loaded with running containers there may be 10k+ processes and millions of files in /proc and /sys. 

Having multiple sosreports extracted creates a huge number of files, and extracting and deleting them all is time consuming.

cfaber:
I agree with Andreas. In my experience, the initial extraction phase can take 30 minutes or longer due to the enormous number of files in /proc and /sys.

This is generally not an issue on typical single-user systems, where these directories are of reasonable size. However, on busy large-scale systems the file count becomes extremely high, and the problem is significantly worse when collecting sosreports from dozens (or more) of such large or high-core nodes.

While the data in /proc and /sys can occasionally be useful for diagnosis, it is typically only needed after the initial extraction has completed.

pmoravec:
I understand and acknowledge the use case, thanks for the numbers.

So "just" the second point remains: to fix that, one needs to extend `sos clean` utility to either:
- untar (and re-tar after cleanup) the dirs in `packed_dirs` (probably safer option? or can we rely on users *not* tar-ing data that might be sensitive? since this untar would usually take a lot of time..)
- or at least prevent deleting the files at https://github.com/sosreport/sos/blob/main/sos/cleaner/archives/__init__.py#L161-L162 .

@TurboTurtle , do you see the second option as sufficient, relying on user's safe usage of this feature..? (or would be a warning in manpages sufficient "coverage"?)

gaurang1988:
Hi, I can go ahead and make changes according to 
>or at least prevent deleting the files at >https://github.com/sosreport/sos/blob/main/sos/cleaner/archives/__init__.py#L161-L162 .

@TurboTurtle please let me know if any objections.

gaurang1988:
@pmoravec @TurboTurtle please review the PR.

sos/cleaner/archives/__init__.py
```diff
@@ -158,6 +158,17 @@ def obfuscate_arc_file(self, filename):
             rel_name = os.path.relpath(filename, start=self.extracted_path)
             if self.should_skip_file(rel_name):
                 return
+            if self.is_packed_dir_tarball(filename):
+                # tarballs produced by the report '--pack-dir' option hold
+                # collected data the user explicitly asked to retain. We can't
+                # obfuscate their contents in place, but unlike other binary
+                # files we must not drop them, or that collected data would be
+                # silently lost from the cleaned report.
+                self.log_info(
```
pmoravec: Could you change it to a warning, please? To heads up the user also on UI that cleaner is skipping its work on the dir.

I am ok with no obfuscation being made here, as the intention of using the `--pack-dir` option is for small "safe" files. But to guide (and *warn*) users about potential uncleaned content, the warning makes sense.

sos/cleaner/archives/__init__.py
```diff
@@ -587,4 +598,30 @@ def should_remove_file(self, fname):
...
+                            archive
+        :type filename:     ``str``
+
+        :returns:   ``True`` if the file is an uncompressed tarball
+        :rtype:     ``bool``
+        """
+        if not filename.endswith('.tar'):
+            return False
+        try:
+            return tarfile.is_tarfile(filename)
```
pmoravec: This will skip dropping *any* tarball, right? So e.g.:

```
mkdir /etc/foreman
hostname > /etc/foreman/hostname.example
tar /etc/foreman/backup.tar /etc/foreman   #  yeah, *this* backup makes no sense, but.. just a quick example

sos report --clean -o foreman,host,networking --batch --build
```

will keep `etc/foreman/backup.tar` with packed sensitive `hostname` - without mentioning it in `--pack-dir`.

Idea of enhancement:

packed_dirs = < archive's manifest item 'packed_dirs', per    self.report_md.add_list('packed_dirs', packed) >

return packed_dirs and filename_with_path.rstrip('.tar') in packed_dirs

First condition is required for cleaning reports with no packed_dirs. You need to pass filename with full relative path here for the 2nd test.

sos/archive.py
```diff
@@ -602,6 +602,48 @@ def do_file_sub(self, path, regexp, subst):
             replacements = 0
         return replacements
 
+    def tar_subdirs(self, paths):
+        """Replace one or more collected directories with a single tarball.
+
+        For each host path in ``paths`` that was collected into this archive,
+        the corresponding directory tree under the archive root is packed into
+        a single uncompressed tarball alongside it (e.g. ``proc/fs/`` becomes
+        ``proc/fs.tar``) and the original tree is removed. The rest of the
```
pmoravec, line 611: Great point to keep the tarball in the relevant relative directory! That is very intuitive.

sos/archive.py
```diff
@@ -602,6 +602,48 @@ def do_file_sub(self, path, regexp, subst):
...
+        so extracting the top-level archive does not have to recreate every
+        file on disk. This backs the report ``--pack-dir`` option, used
+        for verbose trees such as ``/proc/fs/``.
+
+        :param paths: Iterable of host directory paths to pack in place
+        :returns: A list of the host paths that were successfully packed
+        """
+        packed = []
+        for path in paths:
+            dest = self.dest_path(path.rstrip(os.sep))
```
pmoravec, line 627: The `dest_path` causes that also leading `/` are properly trimmed. Such that both `--pack-dir=proc/fs` and `--pack-dir=/proc/fs` works properly - I like this outcome

pmoravec:
Please squash the commits into one before we will merge.

Let me review the PR now..

pmoravec (approved):

gaurang1988:
> Please squash the commits into one before we will merge.

Done

pmoravec (approved):

gaurang1988:
@pmoravec when can we expect to merge this?

pmoravec:
Our policy is two approvals from the core team as required - I asked @TurboTurtle and @arif-ali for a review.

TurboTurtle (approved):

PR #4351, opened 2026-05-29
[landscape] Adding path: /var/lib/landscape/client/

MoonC432:
The directory includes client side configs, metadata etc useful for support troubleshooting.

---

### DIFF
--- a/sos/report/plugins/landscape.py
+++ b/sos/report/plugins/landscape.py
@@ -53,7 +53,8 @@ class Landscape(Plugin, UbuntuPlugin):
             "/etc/landscape/client.conf",
             "/etc/landscape/service.conf",
             "/etc/landscape/service.conf.old",
-            "/var/lib/landscape/landscape-oops/*/OOPS-*"
+            "/var/lib/landscape/landscape-oops/*/OOPS-*",
+            "/var/lib/landscape/client/",
         ])
 
         if not self.get_option("all_logs"):


arif-ali:
Just checking if the files in there have any passwords, secrets or certs?

MoonC432:
I did check those files for sensitive info but nothing stood out. There is a `keystone_token` and an `exchange_token`  in one of the `bpickle` filewhich seems like a uuid of some kind but the `token` threw me off. I don't think its a sensitive info but a second opinion would be nice. I didn't mess with it for now.

pmoravec:
Not knowing Landscape at all, but what I have searched:

- `keystone_token` – Used for authentication with the Landscape server (OpenStack Keystone).
- `exchange_token` – Used for secure communication between the client and server.

So both should be treated as secrets..?

MoonC432:
The syntax looks a lot like uuid, the ones we see in openstack (128-bit random), and the file itself contains hardware information. I think its a uuid for the token that is stored in some database elsewhere. I can just resubmit obfuscating those just in case. What do you think?

pmoravec:
(devils advocate) some secrets can be in the form of uuid.. :) I dont know, I would defer to some Landscape expert (that I dont have or know).

arif-ali:
So looking at the some of these files in that folder on my machine, I do see `/var/lib/landscape/client/broker.bpickle` that does have what looks like a hashed password in it, something after `secure-ids64`. See below, where I have added `<snip>` after `secure-ids64` is what I am looking at

```
> cat /var/lib/landscape/client/broker.bpickle | tail -n 1 | tr ';' '\n'

<snip>
u12:registrationdu11:insecure-idi984
u9:secure-ids64:<snip>
u14:system-versioni0
```

Also `/var/lib/landscape/client/monitor.bpickle`  does have what also looks like hashed password. I would suggest get some feedback from Engineering or an SME on this, just so that we can get this right.

MoonC432:
@arif-ali I asked Mark, and he said those files are good to be retained. None of it contains any sensitive info or PII.

arif-ali (approved):
Based on the fact our SME has confirmed no issues, then it's good to go

pmoravec (approved):

PR #4352, opened 2026-06-01
[plugins] Resolve symlink targets from real parent path

srivathsa729:
Relative symlink targets must be resolved from the real directory that contains the symlink, not from the visible path used to reach it.

When a copied sysfs path contains symlinked parent components, resolving the relative target with dirname(srcpath) can produce invalid paths such as:

/sys/devices/system/cpu/devices/system/cpu/cpu0

Resolve the parent directory with realpath() before joining relative symlink targets. This avoids repeated failed stat messages from plugins that walk sysfs paths, including processor, scsi, md, and hardware.

Pablo Prado reported the issue and proposed this fix.

Reported-by: Pablo Prado <pablo.prado@oracle.com>

---

### DIFF
--- a/sos/report/plugins/__init__.py
+++ b/sos/report/plugins/__init__.py
@@ -1362,7 +1362,12 @@ class Plugin():
     def _copy_symlink(self, srcpath):
         # the target stored in the original symlink
         linkdest = os.readlink(srcpath)
-        dest = os.path.join(os.path.dirname(srcpath), linkdest)
+        srcdir = os.path.dirname(srcpath)
+        realdir = os.path.realpath(srcdir)
+        if os.path.isabs(linkdest):
+            dest = linkdest
+        else:
+            dest = os.path.join(realdir, linkdest)
         # Absolute path to the link target. If SYSROOT != '/' this path
         # is relative to the host root file system.
         absdest = os.path.normpath(dest)
@@ -1375,7 +1380,6 @@ class Plugin():
         if os.path.isabs(linkdest):
             # Canonicalize the link target path to avoid additional levels
             # of symbolic links (that would affect the path nesting level).
-            realdir = os.path.realpath(os.path.dirname(srcpath))
             reldest = os.path.relpath(linkdest, start=realdir)
             # trim leading /sysroot
             if self.use_sysroot():
--- a/tests/unittests/plugin_tests.py
+++ b/tests/unittests/plugin_tests.py
@@ -289,6 +289,44 @@ class PluginTests(unittest.TestCase):
         self.mp._do_copy_path("not_here_tests")
         self.assertEqual(self.mp.archive.m, {})
 
+    def test_copy_relative_symlink_from_symlinked_parent(self):
+        tmpdir = tempfile.mkdtemp()
+        self.addCleanup(shutil.rmtree, tmpdir)
+
+        sysdir = os.path.join(tmpdir, "sys")
+        real_parent = os.path.join(
+            sysdir, "bus", "cpu", "drivers", "processor"
+        )
+        os.makedirs(real_parent)
+
+        target = os.path.join(sysdir, "devices", "system", "cpu", "cpu0")
+        os.makedirs(os.path.dirname(target))
+        with open(target, "w", encoding="utf-8") as target_file:
+            target_file.write("cpu0\n")
+
+        visible_cpu = os.path.join(
+            sysdir, "devices", "system", "cpu", "cpu3"
+        )
+        os.makedirs(visible_cpu)
+        os.symlink("../../../../bus/cpu",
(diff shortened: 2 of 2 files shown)


pmoravec:
I believe you have a valid issue and aiming to resolve it properly, but.. what I am doing wrong when trying to reproduce?

```
# cd /tmp/pr4352
# file $(find | sort)
.:                                       directory
./sys:                                   directory
./sys/bus:                               directory
./sys/bus/cpu:                           directory
./sys/bus/cpu/drivers:                   directory
./sys/bus/cpu/drivers/processor:         directory
./sys/bus/cpu/drivers/processor/cpu0:    symbolic link to ../../../../devices/system/cpu/cpu0
./sys/devices:                           directory
./sys/devices/system:                    directory
./sys/devices/system/cpu:                directory
./sys/devices/system/cpu/cpu0:           ASCII text
./sys/devices/system/cpu/cpu3:           directory
./sys/devices/system/cpu/cpu3/subsystem: symbolic link to ../../../../bus/cpu
#
```

Then modifying a plugin to collect that dir, I get collected the dir exactly as-is..?

(further, I smell there might be an issue/regression when collecting a symlink within a container (and within its sysroot) that targets outside the sysroot - I must think more about that use case)

srivathsa729:
To reproduce you need to collect the nested symlink reached through the symlinked
parent, not only the symlinked directory itself.

The failing path is:

  `/tmp/pr4352/sys/devices/system/cpu/cpu3/subsystem/drivers/processor/cpu0`

Here `subsystem` is a symlink to `../../../../bus/cpu`, so the real parent of `cpu0` is:

  `/tmp/pr4352/sys/bus/cpu/drivers/processor`

The `cpu0` symlink target is:

  `../../../../devices/system/cpu/cpu0`

With the old code, that target is resolved from the visible parent path:

  `/tmp/pr4352/sys/devices/system/cpu/cpu3/subsystem/drivers/processor`

which normalizes to the invalid path:

  `/tmp/pr4352/sys/devices/system/cpu/devices/system/cpu/cpu0`

With this patch, the target is resolved from `realpath(dirname(srcpath))`, which
gives the correct path:

  `/tmp/pr4352/sys/devices/system/cpu/cpu0`

So the issue is not with collecting `cpu3/subsystem` itself as a symlink. The issue
appears when sos later reaches symlinks below that path, such as `subsystem/drivers/processor/cpu0`.

On the container/sysroot case: yes, that is worth checking separately. My intent
here is only to change the base directory used for relative symlink targets from
the visible parent path to the real parent path, because the current behavior
creates invalid paths for sysfs layouts like the above.

pmoravec:
Ah-a, finally I have a reproducer. With the above structure, let collect e.g. `/tmp/pr4352/sys/devices` (but not whole `/tmp/pr4352/sys`), like:

```
# echo ":/tmp/pr4352/sys/devices" > /etc/sos/extras.d/symlink.txt
# python3 bin/sos report -o sos_extras --batch --build
# cd $(ls -td /var/tmp/sos* 2>/dev/null | head -1)
# cd tmp/pr4352
# file $(find . | sort)
.:                                       directory
./sys:                                   directory
./sys/bus:                               directory
./sys/bus/cpu:                           directory
./sys/devices:                           directory
./sys/devices/system:                    directory
./sys/devices/system/cpu:                directory
./sys/devices/system/cpu/cpu0:           ASCII text
./sys/devices/system/cpu/cpu3:           directory
./sys/devices/system/cpu/cpu3/subsystem: symbolic link to ../../../../bus/cpu
#
```

tricky to identify the reproducer :) let me dig more..

sos/report/plugins/__init__.py
```diff
@@ -1362,7 +1362,12 @@ def do_regex_find_all(self, regex, fname):
     def _copy_symlink(self, srcpath):
         # the target stored in the original symlink
         linkdest = os.readlink(srcpath)
-        dest = os.path.join(os.path.dirname(srcpath), linkdest)
+        srcdir = os.path.dirname(srcpath)
+        realdir = os.path.realpath(srcdir)
```
pmoravec, line 1366: Why not simply

```
realdir = os.path.realpath(os.path.dirname(srcpath))
```

as the `srcdir` is unused later on? (well, this is more clear code, so it does make a sense..)

pmoravec:
The patch and unit test sounds good to me, should be idempotent to different sysroot in containers.

pmoravec (approved):

sos/report/plugins/__init__.py
srivathsa729: You are right, we can reduce these two lines to one line, but I kept it separated to make it easier to read, I can simply if you want.

TurboTurtle (approved):

PR #4353, opened 2026-06-01
[dhcp] Update DHCP plugin to correctly support Debian/Ubuntu DHCP servers

goldberl:
**Update DHCP plugin to correctly support Debian/Ubuntu DHCP servers**

The previous Ubuntu plugin assumed the presence of BusyBox udhcpd,
which is not used on modern Debian or Ubuntu systems.

This patch replaces the Ubuntu-specific logic with a Debian/Ubuntu
plugin that collects data for the DHCP servers actually used on
these platforms: isc-dhcp-server and dnsmasq.

Obsolete udhcpd paths have been removed.

Signed-off-by: Leah Goldberg leah.goldberg@canonical.com

---

### DIFF
--- a/sos/report/plugins/dhcp.py
+++ b/sos/report/plugins/dhcp.py
@@ -6,7 +6,7 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.report.plugins import Plugin, RedHatPlugin, UbuntuPlugin
+from sos.report.plugins import Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin
 
 
 class Dhcp(Plugin):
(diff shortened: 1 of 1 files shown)


sos/report/plugins/dhcp.py
```diff
@@ -6,7 +6,7 @@
 #
 # See the LICENSE file in the source distribution for further information.
 
-from sos.report.plugins import Plugin, RedHatPlugin, UbuntuPlugin
+from sos.report.plugins import DebianPlugin, Plugin, RedHatPlugin, UbuntuPlugin
```
pmoravec: Nitpick (that can be inore if the only comment): it is nice to have the list of `*Plugin`s logically ordered, like `Plugin, RedHatPlugin, DebianPlugin, UbuntuPlugin` ("distro-graphically" ordered), anyway.. this is really a detail.

sos/report/plugins/dhcp.py
```diff
@@ -29,16 +29,32 @@ def setup(self):
...
-    packages = ('udhcpd',)
+    packages = ('isc-dhcp-server', 'dnsmasq')
 
     def setup(self):
         super().setup()
         self.add_copy_spec([
-            "/etc/default/udhcpd",
-            "/etc/udhcpd.conf"
+            # ISC DHCP server
+            "/etc/dhcp",
```
pmoravec, line 46: No `/etc/dhcpd.conf` file (to collect)? It is possible, havent checked a Debian/Ubuntu system..

jcastill: It seems that isc-dhcp-server only puts the ipv4 and ipv6 files in that dir:

```
$ dpkg -L isc-dhcp-server | grep etc
[...]
/etc/dhcp
/etc/dhcp/dhcpd.conf
/etc/dhcp/dhcpd6.conf
```
But better if we have confirmation

goldberl: Right that's what I found too. `dhcpd.conf` and `dhcpd6.conf` is collected as part of `/etc/dhcp`.

pmoravec (approved):

goldberl:
Just talked to @arif-ali and we decided it makes sense to keep `udhcpd` in the plugin because older releases still use it and we don't want anything to break during SRUs.

arif-ali (approved):

pmoravec (approved):

PR #4364, opened 2026-06-18
[jvm] New plugin for collecting info about running Java Virtual Machines

fthevenet:
This plugin collects information about all instances of the Java Virtual Machine running on the system. that are based on the "Hotspot" JVM (as made available by the OpenJDK project).

To achieve this, it uses the 'jcmd' utility that is bundled as part of all OpenJDK version currently supported. The plugin makes a first call to `jcmd -l` to list all running JVMs, then attempt to call `jcmd VM.version`, `jcmd VM.info` and  `jcmd System.map` for all detected processes.

Because not all versions of the JVM supports all of these commands, the plugin retrieves a list of supported operation for each of those (using `jcmd $pid help`) and only invokes the commands that the target supports.

The plugin relies on a copy of the jcmd utility being available on the path, as part of OpenJDK 1.8 of higher. It does not matter that the version for jcmd available on the path is different from that of the JVMs it is used to query.

The plugin will attempt to sanitize the output of the `VM.info` by pruning out the values for the properties passed to the JVM using the `-D` argument on the command line.
These properties may be used to communicate secrets to an instance on startup, but because they can be user defined, it is not realistic to hope catch those using simple  heuristics; the safe approach is to obfuscate the values for all of them.

---

### DIFF
--- a/sos/report/plugins/jvm.py
+++ b/sos/report/plugins/jvm.py
@@ -1,0 +1,108 @@
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
+import os
+import pwd
+from sos.report.plugins import PluginOpt, Plugin, IndependentPlugin
+
+
+class Jvm(Plugin, IndependentPlugin):
+    """
+    This plugin collects information about all instances of the Java
+    Virtual Machine running on the system. that are based on the "Hotspot"
+    JVM (as made available by the OpenJDK project).
+
+    To achieve this, it uses the 'jcmd' utility that is bundled as part of
+    all OpenJDK versions currently supported. The plugin makes a first call
+    to `jcmd -l` to list all running JVMs, then attempt to call `jcmd
+    VM.version`, `jcmd VM.info` and  `jcmd System.map` for all detected
+    processes.
+
+    Because not all versions of the JVM support all of these commands, the
+    plugin retrieves a list of supported operations for each of those (using
+    `jcmd $pid help`) and only invokes the commands that the target
+    supports.
+
+    The plugin relies on a copy of the jcmd utility being available on the
+    path, as part of OpenJDK 1.8 of higher. It does not matter that the
+    version for jcmd available on the path is different from that of the
+    JVMs it is used to query.
+
+    The plugin will attempt to sanitize the output of the `VM.info` by
+    pruning out the values for the properties passed to the JVM using the
+    `-D` argument on the command line.
+    These properties may be used to communicate secrets to an instance on
+    startup, but because they can be user defined, it is not realistic to
+    hope catch those using simple heuristics; the safe approach is to
+    obfuscate the values for all of them.
+    """
+
+    short_desc = 'Collect information about running Java Virtual Machines'
+
+    plugin_name = 'jvm'
+    profiles = ('java',)
+    commands = ('jcmd',)
+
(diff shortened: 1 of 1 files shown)


sos/report/plugins/jvm.py
```diff
@@ -0,0 +1,97 @@
...
+                main_class = proc_info[1]
+                self._log_info(main_class)
+                if main_class != 'jdk.jcmd/sun.tools.jcmd.JCmd':
+                    # Get target process user to run jcmd as (or attaching to the VM may fail).
+                    user_id = os.stat(f'/proc/{pid}').st_uid
+                    user_name = pwd.getpwuid(user_id).pw_name
+                    # Get a list of commands supported by the target VM.
+                    jcmd_help_output = self.exec_cmd(f'/usr/bin/jcmd {pid} help',
+                                                     runas=user_name,
+                                                     timeout=self.get_option('jcmd_timeout'))
```
pmoravec: Rather nitpick: from performance point of view (of tiny impact), isnt it better to populate `timeout=self.get_option('jcmd_timeout')` as a variable, ahead of any cycle, and set `timeout=timeout` directly here and few lines below? Rather than invoking a method with deterministic output inside a (nested) loop.

pmoravec:
Please squash the two commits into one, ACK otherwise.

sos/report/plugins/jvm.py
fthevenet: Thanks for the review.
I've done as you suggested.

pmoravec (approved):

pmoravec:
Please fix the pylint/flake8 issues https://github.com/sosreport/sos/actions/runs/27812085765/job/82515918139?pr=4364 (why this wasnt found in the first version of the PR? I would raise the points altogether and not such iteratrively :( ).

fthevenet:
> why this wasnt found in the first version of the PR? I would raise the points altogether and not such iteratrively :( 

It is as if the  pylint/flake8 CI jobs are only run _after_ a reviewer has approved the changes...
That doesn't seem right, does it?

arif-ali:
> > why this wasnt found in the first version of the PR? I would raise the points altogether and not such iteratrively :(
> 
> It is as if the pylint/flake8 CI jobs are only run _after_ a reviewer has approved the changes... That doesn't seem right, does it?

Hey @fthevenet, its your first contribution to the project, so that is really appreciated, and thanks for your contribution. Typically the workflow has to be approved by one of the maintainers for first time contributors.

Based on my experience, further PRs from you will not require approval in the workflow

arif-ali (approved):
pylint and flake is now good, I suspect everything else will also go through

fthevenet:
> > > why this wasnt found in the first version of the PR? I would raise the points altogether and not such iteratrively :(
> > 
> > 
> > It is as if the pylint/flake8 CI jobs are only run _after_ a reviewer has approved the changes... That doesn't seem right, does it?
> 
> Hey @fthevenet, its your first contribution to the project, so that is really appreciated, and thanks for your contribution. Typically the workflow has to be approved by one of the maintainers for first time contributors.
> 
> Based on my experience, further PRs from you will not require approval in the workflow

Hi @arif-ali 
What I meant is that I would have expected the linter jobs to be run on commits, like the build jobs, so that it has a chance to catch problems as early as possible, without having to waste one round of review.

fthevenet:
@arif-ali Oh, and thanks for the review, btw :slightly_smiling_face:

arif-ali:
The CI requires us to manually approve for first time contributors. I agree, it would be good to do that automatically. But that also then prevents any security concerns from any bombardment of PRs, and constantly CI being run.

The linter does run on the jobs, but has to be approved on the platform same as build, functional and unit tests

pmoravec (approved):

Issue #4362, opened 2026-06-13
[FR] Add `--all-projects` to lxc commands


WizardBit:
Some commands here can benefit from having `--all-projects` appended for more output.

https://github.com/sosreport/sos/blob/1f6661ba8b532b4a70cbe479939264974659df3e/sos/report/plugins/lxd.py#L26-L40

pmoravec:
Can you identify particular commands that should be enhanced that way?

I guess your idea is just to append the `--all-project` option to the commands, and not call them once without the option and once with the option..?

Would you be able to prepare a PR for this, or would you prefer some help here?

arif-ali:
Not all of the commands have `--all-projects` as an option. I will come back to this to review and add, or hand it over to one of our team members at Canonical

goldberl:
I'll work on this

arif-ali:
This is done, so closing

PR #4368, opened 2026-06-23
[lxd] include --all-projects in relevant commands

goldberl:
### [lxd] include --all-projects in relevant commands

Update LXD plugin commands to collect resources from all LXD
projects rather than only the default project.
    
Some LXD versions do not support the --all-projects option for
all resource types. Include both variants of the affected
commands so that sosreport can collect the available data across
a wider range of LXD releases.

Addresses #4362 

---

### DIFF
--- a/sos/report/plugins/lxd.py
+++ b/sos/report/plugins/lxd.py
@@ -23,13 +23,22 @@ class LXD(Plugin, UbuntuPlugin):
 
     def setup(self):
 
+        # Some LXD versions do not support the --all-projects option for
+        # all resource types. Include both variants of the affected
+        # commands so that sosreport can collect the available data across
+        # a wider range of LXD releases.
         lxc_cmds = [
+            "lxc image list local: --all-projects",
+            "lxc list local: --all-projects",
+            "lxc network list local: --all-projects",
+            "lxc profile list local: --all-projects",
+            "lxc operation list local: --all-projects",
             "lxc image list local:",
             "lxc list local:",
             "lxc network list local:",
             "lxc profile list local:",
-            "lxc storage list local:",
             "lxc operation list local:",
+            "lxc storage list local:",
             "lxc info local:",
             "lxc alias list",
             "lxc remote list",


sos/report/plugins/lxd.py
```diff
@@ -24,12 +24,12 @@ class LXD(Plugin, UbuntuPlugin):
     def setup(self):
 
         lxc_cmds = [
-            "lxc image list local:",
-            "lxc list local:",
-            "lxc network list local:",
-            "lxc profile list local:",
+            "lxc image list local: --all-projects",
+            "lxc list local: --all-projects",
+            "lxc network list local: --all-projects",
```
MggMuggins, line 33: Not all versions of `lxc` currently deployed in the wild support `--all-projects` for all these resource types; support was just added for networks in `5.21.4`.

I agree that sosreport should collect these (I just did a [similar change](https://github.com/canonical/lxd/commit/f31245cceaddb8042a6ab43c1079312094182821) in `lxd.buginfo`).

Trying to detect what's supported based on version seems fraught. I don't see a reasonable way to define a "fallback" for [`add_cmd_output`](https://sos.readthedocs.io/en/main/plugins.html#sos.report.plugins.Plugin.add_cmd_output). I'm thinking the strategy that's most likely to work is just to add both variants (one with and one without `--all-projects`), even though the duplication or extra code is a bit irrirating. Thoughts?

goldberl: Thanks @MggMuggins, that's a good point. I tested on `5.21.4 LTS`. 

@arif-ali Do you think we would ever backport this change to a release that uses an earlier lxc version? 

I'll look into a possible fallback route as well.

arif-ali: good points here.

We got a couple of scenarios here

LXD is typically a snap package, so hence in the majority of the situations would see it being autorefreshed, so would typically get newer options too with `lxc` commands. However, anyone still using older LTS releases such as 5.0, 4.0 and 3.0, we're not going to have that.

It does make sense to have both then, so that we don't break legacy, but we also get the extra useful commands for the ones we do have `--all-projects` options.

@goldberl to your question, yes, when the next one is released, and most likely 4.12.2, this will be brought into Ubuntu, hence the updated collection will be used later in the year

goldberl: Thanks for the feedback, added both commands (with and without --all-projects)

MggMuggins (commented):
LGTM!

pmoravec (approved):
ACK (purely from sos point of view).

pmoravec:
Again the irrelevant `hyperv` `Input/output error` test failure, I rescheduled the job.

arif-ali (approved):

PR #4371, opened 2026-06-24
[ceph/NVMe-oF] Add a new plugin for Ceph NVMe-oF

pponnuvel:
Fixes #SET-2413.

---

### DIFF
--- a/sos/report/plugins/ceph_nvmeof.py
+++ b/sos/report/plugins/ceph_nvmeof.py
@@ -1,0 +1,62 @@
+# Copyright (C) 2026 Canonical Ltd., Ponnuvel Palaniyappa <pponnuvel@gmail.com>
+
+# This file is part of the sos project: https://github.com/sosreport/sos
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# version 2 of the GNU General Public License.
+#
+# See the LICENSE file in the source distribution for further information.
+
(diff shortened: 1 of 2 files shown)


sos/report/plugins/ceph_nvmeof.py
```diff
@@ -0,0 +1,62 @@
...
+            self.add_copy_spec([
+                "/var/log/ceph/**/ceph-nvmeof*.log",
+            ])
+        else:
+            self.add_copy_spec([
+                "/var/log/ceph/**/ceph-nvmeof*.log*",
+            ])
+
+        self.add_copy_spec([
+            "/var/lib/ceph/**/nvmeof.*/ceph-nvmeof.conf",
```
arif-ali, line 59: just double checking, anything in this config file that could have a password, secret or anything like that?

pponnuvel: No, it doesn't.

It does contain paths to cert files and keys. But nothing in the file itself.

Full list of options can be see at: https://github.com/ceph/ceph/blob/main/src/pybind/mgr/cephadm/templates/services/nvmeof/ceph-nvmeof.conf.j2

pmoravec:
Again HyperV `Input/output error`, independent issue I reported at https://github.com/sosreport/sos/issues/4374 .

pmoravec (approved):
ACK from both sos and ceph content perspective.

arif-ali (approved):

